Issue #4, opened 2015-05-18
Support disabling/enabling rules from source code comments


jpsim:
For example:

```swift
// swiftlint:disable_rule:line_length
... some long line
// swiftlint:enable_rule:line_length
```

aarondaub:
@jpsim do we want to make this purely additive or subtractive (additive would be having no rules on by default, subtractive would be having all rules on by default)? I don't think we should support both - that would complicate things unnecessarily. 

How do you feel about using introspection to map YAML properties to rules? Another option would be to add a property to each Rule for a YAML identifier.

segiddins:
I like rubocops approach -- sensible defaults, can override in either direction.

regexident:
I would vote for an approach based on a context-stack akin to [clang analyzer](http://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-via-pragmas):

Instead of simple serial switches one would be able to have elaborate nested configurations.

And instead of binary `enable_rule`/`disable_rule` switches one would be offered to choose from the same options that clang offers: `ignored`, `note`, `remark`, `warning`, `error`, `fatal`.

    // swiftlint: push
    // swiftlint: ignored some_rule
    
    ... some code
    
    // swiftlint: pop

Or a more complex scenario:

    // swiftlint: push
    // swiftlint: warning rule_1
    // swiftlint: error rule_2
    // swiftlint: error rule_3
    
    ... some code
    
    // swiftlint: push
    // swiftlint: ignored rule_3
    
    ... some code
    
    // swiftlint: pop
        
    ... some code
    
    // swiftlint: pop

With sensible defaults of course.

People should be fairly familiar with this approach by now and can just apply their knowledge of `clang-analyzer` to `swiftlint`, making `swiftlint` feel kinda like an community-based extension to `clang-analyzer`. With Apple being more than likely to come up with a stack-based equivalent for `#pragma clang diagnostic` for Swift in the future, migrating from one stack-based system to another would be much easier than from a stack-less one.

aarondaub:
:+1: stack approach. Not sure if we should add the non-binary options in this issue or open another.

regexident:
I'll add a separate issue for non-binary warning types if desired.

aarondaub:
@regexident I have an implementation of the stack approach in #24. It'll be _pretty_ easy to add options for enabling/disabling rules within a LinterContext (but I'll open a new PR for that if/when #24 is merged)

regexident:
[...]

Issue #3, opened 2015-05-18
Support configuration file


jpsim:
Something like a `.swiftlint.yml` or similar that could be used to override rule parameters or turn off certain rules globally. See #1.

aarondaub:
What is the support for YAML like? Should we limit it to the JSON subset, or is YAML well enough supported?

skywinder:
+1 it will awesome to have `.swiftlint.yml` with custom rules.
as a good example of this approach is  `.rubocop.yml` for ruby linter https://github.com/bbatsov/rubocop

chriseidhof:
I played around a bit, and came up with a possible solution. Maybe this is too complicated, so feel free to build something simpler ;).

For every `Rule`, we can make a separate struct which holds the configurable values. For example, for the TodoRule, we might want to configure the words:

```swift
struct TodoRuleConfiguration {
    var words = ["TODO", "FIXME"]
}
```

Now, if we want to capture the fact that `words` is an option, we can create a struct `ConfigurationOption` which allows us to describe the options name (e.g. this is what people specify in the YAML) and a way to apply it to a configuration of type `A`. Because we don't really know what type will come in, we will try to do a cast within the apply function, and return an updated `A` if this succeeeds. Otherwise we return nil (which means: couldn't parse option). This should probably return a `Result` rather than an optional.

```swift
struct ConfigurationOption<A> {
    let name: String
    let apply: (A, Any) -> A?
}
```

For example, one option would be the `keywords` option:

```swift
let keywordsOption = ConfigurationOption<TodoRuleConfiguration>(name: "keywords", apply: { (var copy, value) in
    if let words = value as? [String] {
      copy.words = words
      return copy
    }
    return nil
})
```

Now, to say that a configuration (e.g. `TodoRuleConfiguration`) has options, we can create a protocol:

```swift
protocol Configurable {
    var options: [ConfigurationOption<Self>] { get }
}

extension TodoRuleConfiguration: Configurable {
    var options = [keywordsOption]
}
```

And finally, now that we have all of that in place, we can write a simple `applyConfiguration` function, which takes a rule, a list of (untyped) option values (this can be a YAML dict, or JSON dict, or whatever) and returns a new rule with all the options applied. Again, this might also produce errors, not sure how we should handle them (hence the TODO).

```swift
func applyConfiguration<A: Configurable>(var rule: A, optionValues: [String:Any]) -> A {
    for option in rule.options {
        if let value = optionValues[option.name] {
[...]

Issue #1, opened 2015-05-18
Unable to specify what rules to validate against


aarondaub:
We should be able to specify what to use - or perhaps specify what rules NOT to use. We should also probably be able to get a list of current rules.

aarondaub:
I'll open a PR for this when I'm back at my laptop this evening.

jpsim:
That would be great! The design I'm considering is to have a `.swiftlint.yml` file or similar (I've opened up #2 to track that).

I'd also like a syntax like
```swift
// swiftlint:disable_rule:line_length
... some long line
// swiftlint:enable_rule:line_length
```

I've opened up #4 to track *that*. PRs are definitely welcome, or even discussion around the design.

aarondaub:
I'm having an issue constraining objects to Rule because it has a type alias. Trying to think of a good solution.

eneko:
I support the idea of a `.swiftlint.yml` configuration file to disable or configure some rules, like line_length. Any work completed on that yet?

In addition, I'd like to suggest for the same configuration file to include options to specify which project folders to include or exclude. This will allow for easy excluding of Pods or third party libraries that developers shouldn't be concerned about (unless working on those projects).

PR #7, opened 2015-05-18
WIP: Add Configuration file and support for removing certain Rules

aarondaub:
@jpsim I haven't verified that this works as I can't get the repo to build. I'll update this PR once I have seen that this actually does what I want it to :)

### DIFF
--- a/Source/SwiftLintFramework/Configuration.swift
+++ b/Source/SwiftLintFramework/Configuration.swift
@@ -1,0 +1,47 @@
+//
+//  Configuration.swift
+//  SwiftLint
+//
+//  Created by Aaron Daub on 2015-05-18.
+//  Copyright (c) 2015 Realm. All rights reserved.
+//
+
+import Foundation
+
+struct Configuration {
+  static let defaultConfigurationName: String = ".swiftlint.yml"
+
+  lazy var configurationString: String {
+    return String(contentsOfFile: self.fileName, encoding: NSUTF8StringEncoding, error: nil)
+  }
+  
+  let fileName: String
+  
+  init(fileName: String? = nil) {
+    self.fileName = fileName ?? Configuration.defaultConfigurationName
+  }
+  
+ private var settings: [String] {
+    return self.configurationString.componentsSeparatedByString("\n")
+  }
+  
+  // Returns an array of values for a given key
+  private func settingsFor(key prefix: String) -> [String] {
+    return filter(self.settings) {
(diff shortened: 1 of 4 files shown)


aarondaub:
This (hypothetically) addresses issues #1, #3, #4.

Source/SwiftLintFramework/Linter.swift
```diff
@@ -31,13 +32,37 @@ public struct Linter {
...
+      TypeNameRule.validateFile(file),
+      VariableNameRule.validateFile(file),
+      TypeBodyLengthRule.validateFile(file),
+      FunctionBodyLengthRule.validateFile(file),
+      NestingRule.validateFile(file)
+    ]
+    
+    return map(self.configuration) {
+      filter(allRules) {
+        return !self.configuration.shouldIgnore($0)
```
aarondaub: Should we remove the negation by changing the interface to `shouldValidateWith` or something equivalent?

Source/SwiftLintFramework/Configuration.swift
```diff
@@ -0,0 +1,47 @@
...
+struct Configuration {
+  static let defaultConfigurationName: String = ".swiftlint.yml"
+
+  lazy var configurationFileContents: String? {
+    return String(contentsOfFile: self.fileName, encoding: NSUTF8StringEncoding, error: nil)
+  }
+  
+  let fileName: String
+  
+  init(fileName: String? = nil) {
```
jpsim: This should be:

```swift
init(fileName: String = Configuration.defaultConfigurationName)
```

jpsim:
We should really be using a robust YAML parser, as the basic string parsing done here will break down pretty quickly.

I can only find https://github.com/behrang/yaml.swift and https://github.com/brynbellomy/SwiftConfig for Swift, and I haven't looked into them yet, so they may not be what we need. Some research needs to be done there. There's also https://github.com/mirek/YAML.framework which is Objective-C but looks a bit more mature.

jpsim:
Also, thanks for working on this and for posting a WIP PR! It's so great to see so much community interest in contributing to this so quickly.

aarondaub:
Yeah, I wrote this very haphazardly yesterday evening. I'm going to focus on getting the Rules to be settable (still hung up on the issue we discussed earlier) and then totally re-do this PR

aarondaub:
@jpsim I think we should close this in favour of PR 12 and then later open one specifically for the configuration file.

PR #9, opened 2015-05-18
Have the reason include the correct length (instead of the hard coded…

aarondaub:
… length) for FileLengthRule and LineLengthRule


jpsim:
Well these rules are really enforcing that the length is at most (minimum parameter value), and you can violate that rule to different extents. So I think the original messages convey the appropriate intent.

Issue #8, opened 2015-05-18
Rule instances only have static properties/methods


aarondaub:
Since all functions/properties on Rule instances are static, we have to validate each rule against the file then remove irrelevant results.

Would it be better to construct an array of Rules, filter that, then call map on the resulting subarray?

aarondaub:
After finally being able to build the project and seeing what pain my ad-configurationFile PR wrought onto the repo I think this is necessary to let the Rules to apply be determined dynamically.

aarondaub:
Addressed by PR 11

PR #11, opened 2015-05-19
Have the Rule protocol define its members as non-static so that we ca…

aarondaub:
…n pass around instances of Rule

@jpsim

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
 ## 0.1.0
 
+Have Rule define its members as non-static so that we can use instances of structs conforming to Rule
+
 First Version!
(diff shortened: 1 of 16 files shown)


aarondaub:
This addresses my issue #8. I believe it is necessary to make a fix for #3 (and later #4) possible.

jpsim:
I agree that being able to pass around rule instances will make passing them around much simpler.

Could you add an entry to `CHANGELOG.md`? Otherwise this looks good.

aarondaub:
@jpsim how do you want to handle merging? Should we squash the commits?

jpsim:
Merged. For future notice, please add changelog entries as I did here: https://github.com/realm/SwiftLint/commit/a196c6acdb2768890405ca5202ac2e92d113986f

PR #12, opened 2015-05-19
Add ability to toggle Rules on and off.

aarondaub:
@jpsim this addresses #1. It branches off of ad-unstaticifyRules as I needed those changes.

My solution to the type alias problem may be controversial and I'm definitely up for alternative suggestions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,11 +15,15 @@
   FunctionBodyLength, Nesting, TypeBodyLength, TypeName, VariableName.  
   [JP Simard](https://github.com/jpsim)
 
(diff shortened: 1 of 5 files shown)


aarondaub:
@jpsim re our discussion regarding Rule_ -> ParameterizedRule:. Should ASTRule conform to Rule or ParameterizedRule? Will ASTRules ever NOT have parameters? If so, should we have a ParameterizedASTRule or just have those have an empty array of parameters?

aarondaub:
Fixed merge conflicts. ASTRules conform to ParameterizedRule, is that the correct behaviour?

jpsim:
Closing in favor of #24, at @aarondaub's request.

PR #19, opened 2015-05-20
Rule explanation

chriseidhof:
Opening this up for discussion.

- [x] What should be the name of the command? Currently it's `swiftlint explain`, maybe `swiftlint rules` is better?
- [x] What should the formatting be? Markdown (for easy website generation)? ANSI-formatted (for easy command-line printing)?
- [x] Add more explanations to more rules


Source/swiftlint/StructuredText.swift
```diff
@@ -0,0 +1,79 @@
...
+        var copy = self
+        copy.removeRange(0..<count)
+        return copy
+    }
+}
+
+
+extension String {
+    var chomped: String {
+        return String(reverse(self).dropWhile { $0 == "\n" }.reverse())
```
jpsim: Could you use `String.lines()` for this instead? https://github.com/realm/SwiftLint/blob/0.1.0/Source/SwiftLintFramework/String+SwiftLint.swift#L13-L20

Source/SwiftLintFramework/ASTRule.swift
```diff
@@ -16,3 +16,10 @@ protocol ASTRule: Rule {
         kind: SwiftDeclarationKind,
         dictionary: XPCDictionary) -> [StyleViolation]
 }
+
+public protocol RuleExample {
+    var ruleName: String { get }
```
jpsim: These first two members (ruleName and ruleDescription) should probably go in `Rule` instead.

Source/SwiftLintFramework/Rules/TodoRule.swift
```diff
@@ -20,4 +22,21 @@ struct TodoRule: Rule {
                 reason: "TODOs and FIXMEs should be avoided")
         }
     }
+
+    public var ruleName = "Todo Rule"
+
+    public var ruleDescription = "This rule checks whether you removed all TODOs and FIXMEs."
+
+    public var correctExamples = [
```
jpsim: I'm afraid this is where mixing tests with usage examples start to get confusing. It's not like the "correctExamples" in this case are recommended style, they're more just tests where we don't want the rule to trigger.

Maybe just changing the semantics a bit to be more like "these examples trigger the rule" and "these examples don't".

Source/SwiftLintFramework/Linter.swift
```diff
@@ -13,23 +13,28 @@ import SourceKittenFramework
...
-                FileLengthRule().validateFile(file),
-                TodoRule().validateFile(file),
-                ColonRule().validateFile(file),
-                TypeNameRule().validateFile(file),
-                VariableNameRule().validateFile(file),
-                TypeBodyLengthRule().validateFile(file),
-                FunctionBodyLengthRule().validateFile(file),
-                NestingRule().validateFile(file)
-            ], [], +)
+        return rules.flatMap { $0.validateFile(self.file) }
```
jpsim: remove `self`

Source/swiftlint/StructuredText.swift
jpsim: Oh, actually I'm a dummy. But you should use `stringByTrimmingTrailingCharactersInSet()` defined in `String+SourceKitten`.

Source/swiftlint/StructuredText.swift
```diff
@@ -0,0 +1,79 @@
...
+    static func red(string: String) -> String {
+        return AnsiCode.RedForeground.wrap(string)
+    }
+    
+    static func green(string: String) -> String {
+        return AnsiCode.GreenForeground.wrap(string)
+    }
+}
+
+enum StructuredText {
```
jpsim: consider moving these types to their own files

chriseidhof:
I think it would be nice to extract some of the other examples too, but they don't just test for correct/incorrect, but also for the severity. (e.g. line length). Maybe we can do that in a separate PR?

chriseidhof:
Hey @jpsim I think this might be ready for merging. Would you mind taking another look at it?

jpsim:
Thank you so much, @chriseidhof! I made a few tweaks and merged in #34.

chriseidhof:
Awesome, great final changes!

On Tue, May 26, 2015 at 12:45 AM, JP Simard <notifications@github.com>
wrote:

> Thank you so much, @chriseidhof <https://github.com/chriseidhof>! I made
> a few tweaks and merged in #34
> <https://github.com/realm/SwiftLint/pull/34>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/realm/SwiftLint/pull/19#issuecomment-105329734>.
>



-- 
Chris Eidhof

Issue #30, opened 2015-05-22
Status code is 0 when high-severity errors found.


pawrsccouk:
In the example file ok.swift, containing: 
```
import UIKit
class Foo {
    var _badVariable = 0
}
```
The bash command: 
```
swiftlint lint
echo "Return Status = $?"
```
Returns 
```
Finding Swift files in current directory...
Linting 'ok.swift' (1/1)
/tmp/ok.swift:3: error: Name Format Violation (High Severity): Variable name should only contain alphanumeric characters: '_badVariable'
Done linting! Found 1 violation in 1 file.
Return Status = 0
```

The return status of a shell script which generates error messages should not be 0. Xcode gives the following message at the end of the build process:
```
Command /bin/sh emitted errors but did not return a nonzero exit code to indicate failure
```
Currently the swiftlint command-line application only returns a non-zero exit code when swiftlint itself had errors, not when it discovers style violations in the source files it looks at. Please can you change it so that it returns a non-zero exit status when it finds style violations?

jpsim:
You're right that SwiftLint should exit with a non-zero code when it finds violations, but only when errors are found, not warnings.

I'd appreciate a PR doing this, if you're interested.

pawrsccouk:
I'll have a look.

jpsim:
You could keep track of the number of violations above `ViolationSeverity.Medium` [here](https://github.com/realm/SwiftLint/blob/0.1.0/Source/swiftlint/Lint.swift#L28) and return `.failure(())` instead of `.success(())`.

pawrsccouk:
Done. It passes the SwiftLintFramework unit tests and lints against one of my projects successfully. If you want any more testing, or want me to make further changes please give me a shout.

segiddins:
Closed by https://github.com/realm/SwiftLint/pull/31/files.

PR #31, opened 2015-05-22
Return an error status code if serious violations found

pawrsccouk:
To fix issue #30 

This mingles errors due to swiftlint issues with errors in the files swiftlint is auditing. We lose information doing that, but otherwise we'd need to re-architect the Command structure and I don't think that's possible right now.

This is my first go at creating GitHub pull requests, so please let me know if I'm doing anything wrong.

### DIFF
--- a/Source/SwiftLintFramework/ViolationSeverity.swift
+++ b/Source/SwiftLintFramework/ViolationSeverity.swift
@@ -28,8 +28,12 @@ public enum ViolationSeverity: Int, Printable, Comparable {
         }
     }
 
+    public var isError: Bool {
+        return self > Medium
+    }
+
     public var xcodeSeverityDescription: String {
-        return self <= Medium ? "warning" : "error"
+        return self.isError ? "error" : "warning"
     }
 }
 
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/ViolationSeverity.swift
```diff
@@ -28,8 +28,12 @@ public enum ViolationSeverity: Int, Printable, Comparable {
         }
     }
 
+    public var isError: Bool {
+        return self > Medium
+    }
+
     public var xcodeSeverityDescription: String {
-        return self <= Medium ? "warning" : "error"
+        return self.isError ? "error" : "warning"
```
jpsim: please omit the `self` keyword here. (it'd be nice to add a rule for this at some point :smile:)

jpsim:
Thanks for doing this! The PR looks good, other than my one point about `self`.

Could you also add an entry to the CHANGELOG under "Enhancements" so we can properly attribute the change to you?

pawrsccouk:
Thanks for the feedback. I've removed the 'self' reference and updated the change log.
Hopefully the changes should appear in this request now.

jpsim:
Thanks!

PR #35, opened 2015-05-26
Excluding Pods files by default (fix #6)

### LINKED ISSUE #6: Don't check cocoapods
Maybe make it optional by specifying it in the config file (#3), but it's probably a sensible default as well. Unless this already shouldn't be happening and I messed it up :grin:

GlaucoNeves:
A start point to make SwiftLint configurable to chose which folders to scan.

### DIFF
--- a/Source/swiftlint/LintCommand.swift
+++ b/Source/swiftlint/LintCommand.swift
@@ -54,7 +54,10 @@ struct LintCommand: CommandType {
 
 func recursivelyFindSwiftFilesInDirectory(directory: String) -> [String] {
     let subPaths = fileManager.subpathsOfDirectoryAtPath(directory, error: nil) as? [String]
-    return map(subPaths) { subPaths in
+    let filteredSubPaths = subPaths?.filter {
+        return !$0.hasPrefix("Pods")
+    }
+    return map(filteredSubPaths) { subPaths in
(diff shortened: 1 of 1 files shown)


jpsim:
Thanks for the PR, but I'd really rather have the ability to exclude folders explicitly rather than have some hardcoded exceptions for `Pods/` and `Carthage/`.

GlaucoNeves:
Agreed. Do you already have an idea of how this should be configured? Maybe passed as parameters when invoked the SwiftLint? I can take a look in that.

jpsim:
Yes, a Commandant parameter would be appropriate for this I think.

eneko:
Any updates on this? I think this belongs to the `.swiftlint.yml` configuration file, is there any PR for that yet?

jpsim:
> Any updates on this? I think this belongs to the .swiftlint.yml configuration file, is there any PR for that yet?

not yet!

jakecraige:
Since the config has been merged, should this be closed?

jpsim:
> Since the config #99 has been merged, should this be closed?

Yes, I'd say so. `Pods/` and `Carthage/` directories still aren't excluded by default, but it's now trivial to do by adding those directories to a `.swiftlint.yml`

PR #37, opened 2015-05-26
Add caching of Structure and SyntaxMap.

nschum:
There is no need to read them individually for every rule.
This cuts execution time by more than 50%.

I've designed the change to be minimally-invasive. Alternatively, I think it might be cleaner to pass a pre-filled custom struct into `ASTRule` instead of `SourceKitten.File`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@
   with examples.  
   [Chris Eidhof](https://github.com/chriseidhof)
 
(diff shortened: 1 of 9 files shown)


jpsim:
Thanks, yeah the initial implementation was very naive, but the tool ran quickly enough in my tests for it not to bother me too much.

I agree it would be cleaner to make a type for this which could get passed around instead of `File`, but this is something we can do later on.

Nice work! :tada:

PR #39, opened 2015-05-26
Add ControlStatementRule

andreamazz:
Hi @jpsim 
I've added the lint rule for `if`, `for`, `while` and `do` while statements as mentioned in one of the Todos in the test suite.

### DIFF
--- a/Source/SwiftLintFramework/Rules/ControlStatementRule.swift
+++ b/Source/SwiftLintFramework/Rules/ControlStatementRule.swift
@@ -1,0 +1,59 @@
+//
+//  ControlStatementRule.swift
+//  SwiftLint
+//
+//  Created by Andrea Mazzini on 26/05/15.
+//  Copyright (c) 2015 Realm. All rights reserved.
+//
+
+import SourceKittenFramework
+
+public struct ControlStatementRule: Rule {
+    public init() {}
+
+    public let identifier = "control_statement"
+
+    public func validateFile(file: File) -> [StyleViolation] {
+        let ifStatement = file.matchPattern("\\s{0,}(if)\\s{0,}\\(",
+            withSyntaxKinds: [.Keyword]).map{ return ($0, "if") }
+        let forStatement = file.matchPattern("\\s{0,}(for)\\s{0,}\\(",
+            withSyntaxKinds: [.Keyword]).map{ return ($0, "for") }
+        let whileStatement = file.matchPattern("\\s{0,}(while)\\s{0,}\\(",
+            withSyntaxKinds: [.Keyword]).map{ return ($0, "while") }
+        return (ifStatement + forStatement + whileStatement).map { violation in
+            return StyleViolation(type: .ControlStatement,
+                location: Location(file: file, offset: violation.0.location),
+                severity: .Low,
+                reason: "\(violation.1) statements shouldn't wrap their conditionals in parentheses.")
+        }
+    }
+
+    public let example = RuleExample(
+        ruleName: "Control Statement",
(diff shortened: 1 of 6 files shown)


Source/SwiftLintFramework/Rules/ControlStatementRule.swift
```diff
@@ -0,0 +1,62 @@
...
+    public let identifier = "control_statement"
+
+    public func validateFile(file: File) -> [StyleViolation] {
+        let statement = file.matchPattern("\\s{0,}(if|for|while)\\s{0,}\\(",
+            withSyntaxKinds: [.Keyword])
+        return statement.map { range in
+            return StyleViolation(type: .ControlStatement,
+                location: Location(file: file, offset: range.location),
+                severity: .Low,
+                reason: "if,for,while,do statements shouldn't wrap their conditionals in parentheses.")
```
segiddins: should probably use the regex match for the specific keyword instead of listing them all

Source/SwiftLintFramework/StyleViolationType.swift
```diff
@@ -16,6 +16,7 @@ public enum StyleViolationType: String, Printable {
     case TODO               = "TODO or FIXME"
     case Colon              = "Colon"
     case Nesting            = "Nesting"
+    case ControlStatement   = "Control Statement"
```
segiddins: `Control Statement Parentheses`?

jpsim:
This is really great, @andreamazz! Thanks for the PR :sparkles: 

You'll have to resolve the merge conflict before we can merge this.

Also, `verifyRule()` has a `commentDoesntViolate` parameter which verifies that triggering examples don't trigger a violation when commented out, so you should use that instead of explicitly commenting out your examples.

Source/SwiftLintFramework/Rules/ControlStatementRule.swift
```diff
@@ -0,0 +1,66 @@
...
+    public let identifier = "control_statement"
+
+    public func validateFile(file: File) -> [StyleViolation] {
+        let ifStatement = file.matchPattern("\\s{0,}(if)\\s{0,}\\(",
+            withSyntaxKinds: [.Keyword]).map{ return ($0, "if") }
+        let forStatement = file.matchPattern("\\s{0,}(for)\\s{0,}\\(",
+            withSyntaxKinds: [.Keyword]).map{ return ($0, "for") }
+        let whileStatement = file.matchPattern("\\s{0,}(while)\\s{0,}\\(",
+            withSyntaxKinds: [.Keyword]).map{ return ($0, "while") }
+        return (ifStatement + forStatement + whileStatement).map { violation in
```
andreamazz: Should we separate `while` from `do-while` statements?

andreamazz:
Fixed the conflict :+1:

Source/SwiftLintFramework/Rules/ControlStatementRule.swift
jpsim: No, this seems fine to me.

jpsim:
Looks great, @andreamazz! Could you add an entry to CHANGELOG.md under "Enhancements"?

andreamazz:
Sure! Updated.

jpsim:
Fantastic, thanks for the hard work @andreamazz!

jpsim:
FYI, there were a few false positives with this rule, especially when using tuples in control statements, and I've attempted to address those in #48.

andreamazz:
Right, tuples did not cross my mind :sweat_smile: 
Thanks :+1:

Issue #20, opened 2015-05-20
Add ability to skip certain files


mcdappdev:
Add an option that allows the user to select files to skip when scanning for errors

aarondaub:
Related to: #6

jpsim:
This was done in #99 via the `excluded` YAML entry.

Issue #16, opened 2015-05-20
Allow specifying a directory / file patterns to check


segiddins:

aarondaub:
Related: #6 #20

jpsim:
Completed by #44.

Issue #43, opened 2015-05-27
TrailingNewlineRule.swift doesn't seem to work


rajohns08:
I've tried adding multiple ending new lines and having 0 extra new lines at the end of my file, but I never get a warning for this. I know the swiftlint library is active though, because I'm getting other warnings.

jpsim:
You're absolutely right! I'll look into why this is the case.

jpsim:
The problem was that we weren't specifying a line in the warning, so Xcode didn't pick it up. Fixed in #46.

rajohns08:
Has this fix been pushed to brew yet? Running a brew upgrade didn't update the package.

jpsim:
Nope, there's a big pile of fixes as you can see from the [CHANGELOG](https://github.com/realm/SwiftLint/blob/master/CHANGELOG.md) that I'd like to include in a release shortly. I'll cut a release today or tomorrow.

PR #44, opened 2015-05-27
Added a command line option to specify a path to lint

larslockefeer:
This pull request adds a command-line option to specify an (absolute or relative) path to a file or directory to lint.

This enables use cases such as:

* Linting only the file that you are currently working on
* Linting only modified files in an Xcode Run Script build phase
* Linting modified files as a pre-commit hook

Usage examples:

* `swiftlint lint --path Source/swiftlint/LintCommand.swift`
* `swiftlint lint --path Source/SwiftLintFramework`

This PR relates to issues #6, #16 and #20


jpsim:
Thanks for the PR, @larslockefeer! This is definitely on the right track, but needs a bit of work:

1. Please run SwiftLint on `LintCommand.swift` itself, as a number of your additions are causing violations. I'll be sure to add this to our CONTRIBUTING.md document once we add one :).
2. Please add a single newline at the end of the file. We have a SwiftLint rule for this, but it's not working at the moment (see #43).
3. I'd rather you used [`NSString.absolutePathRepresentation(rootDirectory:)`](https://github.com/jpsim/SourceKitten/blob/0.4.3/Source/SourceKittenFramework/String%2BSourceKitten.swift#L66-L76) which is already available in SourceKittenFramework, or at least `NSString.absolutePath` rather than combining `path` to `currentDirectoryPath`

Otherwise, I have a few minor comments that I'll add inline to your PR now.

Again, thanks for doing this!

Source/swiftlint/LintCommand.swift
```diff
@@ -20,51 +20,94 @@ struct LintCommand: CommandType {
...
+struct LintOptions: OptionsType {
+    let path: String
+    
+    static func create(path: String) -> LintOptions {
+        return LintOptions(path: path)
+    }
+    
+    static func evaluate(m: CommandMode) -> Result<LintOptions, CommandantError<()>> {
+        return create
+            <*> m <| Option(key: "path", defaultValue: "", usage: "theff path to the file or directory to lint")
```
jpsim: `theff` should be `the`

Source/swiftlint/LintCommand.swift
```diff
@@ -20,51 +20,94 @@ struct LintCommand: CommandType {
...
-                if violation.severity.isError {
-                    numberOfSeriousViolations++
+        return LintOptions.evaluate(mode).flatMap { options in
+            return self.lint(options.path)
+        }
+    }
+    
+    func lint(path: String) -> Result<(), CommandantError<()>> {
+        let filesToLint = filesToLintAtPath(path)
+        if  filesToLint.count > 0 {
```
jpsim: there's an extra space here

larslockefeer:
Thanks for the feedback @jpsim!

I have processed your comments and synced with the upstream again.

I have discovered one additional issue after syncing with the upstream: files are now linted twice. I suspect that it is due to `fileManager.allFilesRecursively` yielding both `contentsOfDirectoryAtPath` and `subpathsOfDirectoryAtPath` but I haven't looked into that in detail yet.

Also, I see that the tests failed. On my machine, they ran fine so this could be an issue with Travis. I see the following error in their console output:

```
The command "script/cibuild" exited with 2.
```

segiddins:
Failure looks unrelated, I restarted the build.

Source/swiftlint/LintCommand.swift
```diff
@@ -20,36 +20,87 @@ struct LintCommand: CommandType {
...
+    }
+
+    static func evaluate(m: CommandMode) -> Result<LintOptions, CommandantError<()>> {
+        return create
+            <*> m <| Option(key: "path", defaultValue: "", usage: "the path to the file or" +
+                        " directory to lint")
+    }
+}
+
+func filesToLintAtPath(path: String) -> [String] {
```
jpsim: Seems like this function could be simplified:

```swift
func filesToLintAtPath(path: String) -> [String] {
    let absolutePath = path.absolutePathRepresentation()
    var isDirectory: ObjCBool = false
    if fileManager.fileExistsAtPath(absolutePath, isDirectory: &isDirectory) {
        if isDirectory {
            return fileManager.allFilesRecursively(directory: absolutePath).filter {
                $0.isSwiftFile()
            }
        } else if standardizedPath.isSwiftFile() {
            return [standardizedPath]
        }
    }
    return []
 }
 ```

jpsim:
This is great, @larslockefeer. My final comment is about perhaps simplifying `filesToLintAtPath(_:)`. It would also be great if you could add an entry to `CHANGELOG.md`. Thanks for this.

larslockefeer:
Thanks again @jpsim, I have processed your feedback.

jpsim:
Hurray! Great work.

PR #47, opened 2015-05-28
[LineLengthRule] Increase default parameters

### LINKED ISSUE #28: Line length to 120
On a 27-inch display, you can comfortable fit two 120 wide files at 14pt with sidebars and all. On a 13" rMBP, you can fit one. I've had my guide set at 120 for years. Personally, I think 100 is overly narrow. Almost all of the warnings for one of my projects was length being over 100 but less 120.

I'm guessing there's a reason you picked 100 though. I'd love to know why!

I'm tempted to ask for a way to configure it, but I think one of the biggest benefits of this is we can all start doing things the same as a community.

jpsim:
Fixes #28. After further consideration, I don't think a default of 100 was tenable for even this project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
   [aarondaub](https://github.com/aarondaub)
   [#21](https://github.com/realm/SwiftLint/issues/21)
 
+* Increase default `LineLengthRule` parameters to 120, 150, 180, 200, 250.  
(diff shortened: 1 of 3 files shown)


segiddins:
:-1:

jpsim:
hmmm, yeah we should really just be prioritizing #3 instead.

PR #65, opened 2015-05-29
[IntegrationTests] Added testSwiftLintLints(). Yo dawg.

jpsim:
As noted in the test comment: This is as close as we're ever going to get to a self-hosting linter.

### DIFF
--- a/Source/SwiftLintFrameworkTests/IntegrationTests.swift
+++ b/Source/SwiftLintFrameworkTests/IntegrationTests.swift
@@ -1,0 +1,26 @@
+//
+//  IntegrationTests.swift
+//  SwiftLint
+//
+//  Created by JP Simard on 5/28/15.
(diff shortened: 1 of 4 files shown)


Source/SwiftLintFrameworkTests/IntegrationTests.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+
+import Foundation
+import SourceKittenFramework
+import SwiftLintFramework
+import XCTest
+
+class IntegrationTests: XCTestCase {
+    func testSwiftLintLints() {
+        // This is as close as we're ever going to get to a self-hosting linter.
```
segiddins: not really 'self-hosting'

jpsim: Yeah, because I'm taking that claim *extremely* seriously.

PR #75, opened 2015-06-03
add ReturnArrowWhitespaceRule

akirahrkw:
Hi @jpsim and everyone
I've added a rule which checks whether you have 1 space before return arrow and return type or not
ex:
 Good: func abc() -> Int {}
 Bad:  func abc()->Int {}

But I'm not sure whether this rule makes sense for you or not. 
and I think this rule is not on github's swift style guide though, we have our own swift style guide for our dev team, https://github.com/burpple/swift-style-guide (which is not completed yet:) )
so that would be nice if I can contribute and apply some of our rules to this project.

thanks:)

### DIFF
--- a/Source/SwiftLintFramework/Rules/ReturnArrowWhitespaceRule.swift
+++ b/Source/SwiftLintFramework/Rules/ReturnArrowWhitespaceRule.swift
@@ -1,0 +1,58 @@
+//
+//  ReturningWhitespaceRule.swift
+//  SwiftLint
+//
+//  Created by Akira Hirakawa on 2/6/15.
+//  Copyright (c) 2015 Realm. All rights reserved.
+//
+
+import SourceKittenFramework
+
+public struct ReturnArrowWhitespaceRule: Rule {
+    public init() {}
+
+    public let identifier = "return_arrow_whitespace"
+
+    public func validateFile(file: File) -> [StyleViolation] {
+        // space doesn't include \n so that "func abc()->\n" can pass validation
+        let space = "[ \\f\\r\\t\\v]"
+        let spaceRegex = "(\(space){0}|\(space){2,})"
+
+        // ex: func abc()-> Int {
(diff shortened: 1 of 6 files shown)


Source/SwiftLintFrameworkTests/StringRuleTests.swift
```diff
@@ -70,6 +70,12 @@ class StringRuleTests: XCTestCase {
             commentDoesntViolate: false)
     }
 
+    func testLinesShouldContainReturnArrowWhitespace() {
+        verifyRule(ReturnArrowWhitespaceRule().example,
+            type: .ReturnArrowWhitespace,
+            commentDoesntViolate: false)
```
jpsim: I think this should be true (or just omitted)

Source/SwiftLintFramework/Rules/ReturnArrowWhitespaceRule.swift
```diff
@@ -0,0 +1,58 @@
...
+        triggeringExamples: [
+            "func abc()->Int {}\n",
+            "func abc()->[Int] {}\n",
+            "func abc()-> Int {}\n",
+            "func abc() ->Int {}\n",
+            "func abc()  ->  Int {}\n",
+            "var abc = {(param: Int) ->Bool in }\n",
+            "var abc = {(param: Int)->Bool in }\n"
+        ],
+        showExamples: false
```
jpsim: Why not show these examples?

jpsim:
This looks great, thanks! I just have very minor feedback. It'd be nice if you could add an entry for this in the CHANGELOG too!

akirahrkw:
@jpsim thanks for your feedback!
I will push the change soon, but should I add an entry on 0.1.1 part in CHANGELOG? or better to make 0.1.2?

jpsim:
Ah, I just created an empty "master" section to the changelog, please add your entry there.

akirahrkw:
thanks:) i just pushed my commits!

jpsim:
Great work! Thanks!

PR #76, opened 2015-06-03
Option to ignore one or all rules for a file by adding a comment

evermeer:
With this change you can disable all rules for a file by adding the following line to your swift file:
// SwiftLint ignore all
or you can disable a specific rule by adding something like:
// SwiftLint ignore variable_name
Where variable_name is the identifier of the rule

### DIFF
--- a/Source/SwiftLintFramework/Linter.swift
+++ b/Source/SwiftLintFramework/Linter.swift
@@ -31,7 +31,14 @@ public struct Linter {
     ]
 
     public var styleViolations: [StyleViolation] {
-        return rules.flatMap { $0.validateFile(file) }
+        return rules.flatMap {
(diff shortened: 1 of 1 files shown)


segiddins:
This only seems to be per-file, and thus has no way to only ignore a certain section of code?

jpsim:
This would conflict with the current proposal described in #4. Although the approach of filtering the results could still work with file regions rather than per-file.

evermeer:
When #4 will be implemented soon, then this PR can be closed without merge.

keith:
A similar but different version of this was implemented in https://github.com/realm/SwiftLint/pull/111

jpsim:
Thanks for pointing that out, @keith. Closing since #111 allows this behavior and more.

Issue #60, opened 2015-05-28
Validate whitespace around operator definitions


jpsim:
Use whitespace around operator definitions.

Good: `func <|< <A>(lhs: A, rhs: A) -> A`
Bad:  `func <|<<A>(lhs: A, rhs: A) -> A`

PR #82, opened 2015-06-16
Use whitespace around operator definitions.

akirahrkw:
this PR is about "Use whitespace around operator definitions."
https://github.com/realm/SwiftLint/issues/60

thanks:)

### DIFF
--- a/Source/SwiftLintFramework/Location.swift
+++ b/Source/SwiftLintFramework/Location.swift
@@ -48,7 +48,7 @@ Returns true if `lhs` Location is equal to `rhs` Location.
 
 :returns: True if `lhs` Location is equal to `rhs` Location.
 */
-public func ==(lhs: Location, rhs: Location) -> Bool {
+public func == (lhs: Location, rhs: Location) -> Bool {
     return lhs.file == rhs.file &&
         lhs.line == rhs.line &&
         lhs.character == rhs.character
--- a/Source/SwiftLintFramework/Rules/OperatorFunctionWhitespaceRule.swift
+++ b/Source/SwiftLintFramework/Rules/OperatorFunctionWhitespaceRule.swift
@@ -48,10 +48,10 @@ public struct OperatorFunctionWhitespaceRule: ASTRule {
             let location = Location(file: file, offset: offset)
             let startAdvance = advance(file.contents.startIndex, nameOffset)
             let endAdvance = advance(startAdvance, nameLength)
-            let range = Range<String.Index>(start: startAdvance, end: endAdvance)
+            let range = Range(start: startAdvance, end: endAdvance)
             let definition = file.contents.substringWithRange(range)
 
-            let ope1 = ["/", "=", "-", "+", "!", "*", "|", "^", "~", "?", "."].map({ "\\\($0)" })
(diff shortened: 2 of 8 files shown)


Source/SwiftLintFramework/Location.swift
```diff
@@ -48,7 +48,7 @@ Returns true if `lhs` Location is equal to `rhs` Location.
 
 :returns: True if `lhs` Location is equal to `rhs` Location.
 */
-public func ==(lhs: Location, rhs: Location) -> Bool {
+public func == (lhs: Location, rhs: Location) -> Bool {
```
jpsim: :smile:

Source/SwiftLintFramework/Rules/OperatorFunctionWhitespaceRule.swift
```diff
@@ -0,0 +1,87 @@
...
+        }
+        var violations = [StyleViolation]()
+        if let nameOffset = flatMap(dictionary["key.nameoffset"] as? Int64, { Int($0) }),
+            let nameLength = flatMap(dictionary["key.namelength"] as? Int64, { Int($0) }),
+            let offset = flatMap(dictionary["key.offset"] as? Int64, { Int($0) }) {
+
+            let location = Location(file: file, offset: offset)
+            let startAdvance = advance(file.contents.startIndex, nameOffset)
+            let endAdvance = advance(startAdvance, nameLength)
+            let range = Range<String.Index>(start: startAdvance, end: endAdvance)
```
jpsim: `<String.Index>` should be inferred, is it not?

Source/SwiftLintFramework/Rules/OperatorFunctionWhitespaceRule.swift
```diff
@@ -0,0 +1,87 @@
...
+            let definition = file.contents.substringWithRange(range)
+
+            let ope1 = ["/", "=", "-", "+", "!", "*", "|", "^", "~", "?", "."].map({ "\\\($0)" })
```
jpsim: nice use of `map`. Can you use trailing closure syntax here?

jpsim:
This is really nice work, @akirahrkw! I just had a few minor comments, let me know what you think, then we'll merge this in.

Source/SwiftLintFramework/Rules/OperatorFunctionWhitespaceRule.swift
akirahrkw: ah yes, this is inferred:) I will remove it

Source/SwiftLintFramework/Rules/OperatorFunctionWhitespaceRule.swift
akirahrkw: it means .map({"\\\($0)"}), doesn't it?
yeah I will use it.

akirahrkw:
thanks for your feedback. I just pushed the change:)

jpsim:
I made a few changes to your work in #83 to support multibyte characters & specify only a single space on each side of the operator. So I'm closing this in favor of #83. Very nice work on this, @akirahrkw!

akirahrkw:
got it and thank you:)

PR #83, opened 2015-06-17
Use whitespace around operator definitions.

jpsim:
Minor tweaks & refactoring from #82. All credit to @akirahrkw! :clap:

### DIFF
--- a/Source/SwiftLintFramework/Rules/ForceCastRule.swift
+++ b/Source/SwiftLintFramework/Rules/ForceCastRule.swift
@@ -26,8 +26,7 @@ public struct ForceCastRule: Rule {
         ruleName: "Force Cast Rule",
         ruleDescription: "This rule checks whether you don't do force casts.",
         nonTriggeringExamples: [
-            "NSNumber() as? Int\n",
-            "// NSNumber() as! Int\n",
+            "NSNumber() as? Int\n"
         ],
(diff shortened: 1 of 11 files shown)


Source/SwiftLintFramework/Rules/ForceCastRule.swift
```diff
@@ -26,8 +26,7 @@ public struct ForceCastRule: Rule {
         ruleName: "Force Cast Rule",
         ruleDescription: "This rule checks whether you don't do force casts.",
         nonTriggeringExamples: [
-            "NSNumber() as? Int\n",
```
segiddins: why this change?

jpsim: Because of `commentDoesntViolate` in [`verifyRule()`](https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFrameworkTests/TestHelpers.swift#L18).

segiddins: right, but this is a non-triggering example?

jpsim: `verifyRule()` will treat all triggering examples as non-triggering when commented out. So `// NSNumber() as! Int` was already being tested because `NSNumber() as! Int` is a triggering example. So this line was duplicating a test.

Issue #66, opened 2015-05-29
Validate all public declarations are fully documented


segiddins:
Allow the ACL level to be customized

keith:
I've been implementing this on a fork of swiftlint. It got complicated since we (possibly not everyone) don't want to require comments on things that are implementing a protocol.

jpsim:
That's a valid concern. Ideally, this feature would take into account "parent" documentation (e.g. if subclassing, or implementing a protocol method).

keith:
We have a working implementation of this which works like this:

First we cache all declared protocols in given paths in a json file with the format: (in theory we could do this each time, it doesn't take that long, but this allows us to save some time linting when nothing has changed)

```json
{ "protocol": "path/to/containing/file" }
```

Then while parsing the file we:

1. Grab the type's inherited types
1. Grab all members that should be commented
1. Exclude those that are commented
1. Look at the inherited type's members
1. If there's a matching signature, a comment isn't needed

The issue for us comes when your type inherits from a protocol that we don't know about from the list we generated above. In our codebase we mostly do something like this:

```swift
struct Foo {
  /**
  Stuff
  */
  func bar() {}

  ...
}

extension Foo: SomeFrameworkProtocol {
  func thingThatDoesNotNeedComment() {}
}
```

Because of this we decided the easiest way to handle these cases which also don't need comments is to exclude everything inside the extension/declaration if one of the inherited types matches a "blacklist" that we came up with. This blacklist is regexes which mainly contains `^UI\w+(DataSource|Delegate)\w*$` and then a few third party things we have that declare protocols. I'd love a better way to handle this situation since this case isn't flagged:

```swift
extension Foo: UITableViewDelegate {
  func someRandomUncommentedFunction() {}
}
```

But we also want to be able to do this linting without compiling the project.

jpsim:
This was done as two complementary rules: `missing_docs` and `valid_docs`.

PR #91, opened 2015-08-11
Only validate examples produce their own violations

keith:
When adding new general rules, examples from other rules could violate your new rule. Since we're already testing rules practically through the integration tests we should only validate that examples produce the expected violation types.

This became an issue when I was working on https://github.com/realm/SwiftLint/issues/66 since all the current examples don't have docstrings (since they're not relevant to the rules they're testing).


keith:
Thinking about this a little more it might be easier if verifyRule just created the rule itself and used that instead of using the styleViolations from `Linter` as it does now. Let me know what you think about that and I'll change it to work that way instead.

jpsim:
I can see how this definitely becomes a problem as we add more rules. Thanks for the PR.

PR #94, opened 2015-08-15
Improve trailing newline performance

keith:
This sidesteps the previous method of reversing the entire string from every file by only checking the minimum necessary number of trailing characters for each file.

In our tests linting of our entire project with ~400 files using only this rule, (I disabled the others for testing) the lint time went from minutes to seconds.

Before:

screen shot 2015-08-14 at 14 03 08

After:

screen shot 2015-08-14 at 14 05 31

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@
   [JP Simard](https://github.com/jpsim)
   [#78](https://github.com/realm/SwiftLint/issues/78)
 
+* Improve performance of `TrailingNewlineRule`  
+  [Keith Smiley](https://github.com/keith)
+
 ##### Bug Fixes
 
 * None.
--- a/Source/SwiftLintFramework/Rules/TrailingNewlineRule.swift
+++ b/Source/SwiftLintFramework/Rules/TrailingNewlineRule.swift
@@ -14,16 +14,20 @@ public struct TrailingNewlineRule: Rule {
     public let identifier = "trailing_newline"
 
     public func validateFile(file: File) -> [StyleViolation] {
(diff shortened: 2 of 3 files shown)


jpsim:
I'm appalled at the initial performance of this. That's just horrible. Thanks for improving this so much.

It'd be great if these changes could go in `countOfTailingCharactersInSet` instead, since they appear to be generally applicable. Please add a changelog entry too.

keith:
The only problem with porting this back to `countOfTailingCharactersInSet` is that we're only checking the minimum necessary (2) for the case we care about. I think to get the count you'd have to do the reversing as it is now. Because of the performance implications of that function I've actually removed it entirely on our fork. The only other place it's used is the trailing whitespace line rule, which has the same performance problems. Was just waiting for https://github.com/jpsim/SourceKitten/pull/60

jpsim:
I think modifying `countOfTailingCharactersInSet` is a better approach though, and the performance would probably be even better than the `split` way if using the lazy reversed generator:

```swift
func countOfTrailingCharactersInSet(characterSet: NSCharacterSet) -> Int {
    let utf16View = utf16
    var count = 0
    for char in reverse(utf16View) {
        if !characterSet.characterIsMember(char) {
            break
        }
        count++
    }
    return count
}
```

keith:
Good idea! I'll benchmark this tomorrow!

keith:
I've just benchmarked this using just this rule.

Original:

original

Lazy reverse:

lazy

This current fix:

minimum

jpsim:
:+1: thanks for doing this, and the benchmarks too!

keith:
Awesome thanks! More stuff incoming!

PR #96, opened 2015-08-17
Remove SourceKittenFramework through Makefile

keith:
While working on SwiftLint, if you ended up editing the SourceKitten
framework you end up getting really cryptic errors about missing types
from both sides. Since we're building this framework when we build this
project and we're installing it on `make install` we should remove it
with the uninstall.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,7 @@ install: package
 
 uninstall:
 	rm -rf "$(FRAMEWORKS_FOLDER)/SwiftLintFramework.framework"
+	rm -rf "$(FRAMEWORKS_FOLDER)/SourceKittenFramework.framework"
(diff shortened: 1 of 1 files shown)


segiddins:
:+1:

jpsim:
`SourceKittenFramework.framework` is nested in `SwiftLintFramework.framework` when installing SwiftLint, so if there's a `SourceKittenFramework.framework` next to `SwiftLintFramework.framework`, it got there by other means (e.g. installing SourceKitten). So this fix isn't correct.

However, the problem still stands that frameworks are installed in a shared location, which can cause conflicts like the one you've identified. The appropriate fix is to store the frameworks in a non-shared location (e.g. `<current path to frameworks>/<app name>/<app version>/`).

PR #99, opened 2015-08-24
Configure SwiftLint via a YAML file

### LINKED ISSUE #1: Unable to specify what rules to validate against

jpsim:
Fixes #1 and #3.

### DIFF
--- a/Source/SwiftLintFramework/Configuration.swift
+++ b/Source/SwiftLintFramework/Configuration.swift
@@ -1,0 +1,103 @@
+//
+//  Configuration.swift
+//  SwiftLint
+//
+//  Created by JP Simard on 2015-08-23.
+//  Copyright (c) 2015 Realm. All rights reserved.
+//
+
+import Yaml
+
+extension Yaml {
+    var arrayOfStrings: [Swift.String]? {
+        return array?.flatMap { $0.string }
+    }
+}
+
+public struct Configuration {
+    public let disabledRules: [String] // disabled_rules
+    public let included: [String]      // included
+    public let excluded: [String]      // excluded
+
+    public var rules: [Rule] {
+        return allRules.filter { !disabledRules.contains($0.identifier) }
+    }
+
+    public init?(disabledRules: [String] = [], included: [String] = [], excluded: [String] = []) {
+        self.disabledRules = disabledRules
+        self.included = included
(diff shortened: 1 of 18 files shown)


jpsim:
@keith I assume this would be useful for you, I'd love if you could share your thoughts on the implementation.

Source/SwiftLintFramework/Configuration.swift
```diff
@@ -0,0 +1,103 @@
...
+                fputs("config error: '\(duplicateRule.0)' is listed \(duplicateRule.1) times\n",
+                    stderr)
+            }
+            return nil
+        }
+    }
+
+    public init?(yaml: String) {
+        guard let yamlConfig = Yaml.load(yaml).value else {
+            return nil
```
keith: Should this print some error about an invalid config?

keith: Ah I see this error below

Source/swiftlint/LintCommand.swift
```diff
@@ -21,36 +21,43 @@ struct LintCommand: CommandType {
...
             if options.useSTDIN {
                 let standardInput = NSFileHandle.fileHandleWithStandardInput()
                 let stdinData = standardInput.readDataToEndOfFile()
                 let stdinNSString = NSString(data: stdinData, encoding: NSUTF8StringEncoding)
                 if let stdinString = stdinNSString as? String {
-                    let violations = Linter(file: File(contents: stdinString)).styleViolations
-                    print("\n".join(violations.map { $0.description }))
+                    let file = File(contents: stdinString)
+                    let linter = Linter(file: file, configuration: configuration)
+                    print("\n".join(linter.styleViolations.map { $0.description }))
```
keith: Not sure which beta you're on but this `.join` will have to change eventually

keith:
Awesome! This looks great! :+1:

Source/swiftlint/LintCommand.swift
jpsim: The `swift-2.0` branch is targeting Xcode 7 beta 5. I'll have to update all the dependencies to beta 6 before I can do the same to SwiftLint... ugh.

PR #100, opened 2015-08-25
Add path or file based exclusion to the linter

nwest:
Taking a stab at #6.

It would be pretty easy to add regex based exclusion too.


jpsim:
This is great, @nwest! I actually wrote similar functionality in #99 just yesterday (although you configure excluded paths via a YAML file rather than a command-line option). It'd be great to compare these two approaches and stick with the more efficient one. What do you think?

nwest:
Ah dang! I wish I would have noticed that.

I think your yaml based config is more robust. Will we be able to switch up the config without committing updates? (e.g for CI)

jpsim:
> Will we be able to switch up the config without committing updates? (e.g for CI)

No, I don't intend on backporting YAML configuration support to Swift 1.2 unless there's strong demand.

Let me know if you think there's still good reason to keep this PR open, otherwise please close @nwest. I hope you'll continue to contribute to SwiftLint in the future!

PR #106, opened 2015-08-28
[Configuration] allow configuring parameterized rules from the configuration file

jpsim:
This is awkwardly implemented at the moment because we're starting to duplicate the set of rules in quite a few places, but this works for now.

```ruby
line_length:
  - 120 # Very Low
  - 140 # Low
  - 160 # Medium
  - 180 # High
  - 200 # Very High
```

Thoughts @segiddins @keith ?

### DIFF
--- a/Source/SwiftLintFramework/Rule.swift
+++ b/Source/SwiftLintFramework/Rule.swift
@@ -9,6 +9,7 @@
 import SourceKittenFramework
 
 public protocol Rule {
+    init()
     var identifier: String { get }
     func validateFile(file: File) -> [StyleViolation]
     var example: RuleExample { get }
@@ -16,6 +17,7 @@ public protocol Rule {
 
 public protocol ParameterizedRule: Rule {
     typealias ParameterType
+    init(parameters: [RuleParameter<ParameterType>])
     var parameters: [RuleParameter<ParameterType>] { get }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,10 +14,14 @@
 ##### Enhancements
 
 * Configure SwiftLint via a YAML file:
-  Supports `disabled_rules`, `included` and `excluded`.
+  Supports `disabled_rules`, `included`, `excluded` and passing parameters to
+  parameterized rules.
   Pass a configuration file path to `--config`, defaults to `.swiftlint.yml`.  
   [JP Simard](https://github.com/jpsim)
+  [#1](https://github.com/realm/SwiftLint/issues/1)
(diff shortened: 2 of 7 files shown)


keith:
Personally I question the need for the different levels of severity. But I also think projects should have 0 warnings. Regardless 5 of them may be a little more than needed?

I think that if someone only provides a single number for these it should actually correspond to very high, and work backwards to very low. In our project I would just use very high at 110 for line length for example.

jpsim:
> Personally I question the need for the different levels of severity.

Same here, to be honest. The original idea was that you may want to warn for some levels, but fail/error for more severe cases. Or you may want to set a minimum severity level you care about depending on the situation? But I think in practice, that complicates usage of SwiftLint more than helping.

Maybe just moving to 2 severity levels (warning & error, which is all that Xcode can present anyway) would be the best way forward?

I understand your intentions in proposing going backwards in the list of parameters, but I think that could get confusing.

keith:
I think warning vs error is a good distinction. I get that reversing is confusing, I'm not sure what the best way to handle this when someone just adds a single parameter. That is pretty confusion from the users side.

keith:
Actually if there are only 2 levels I think it makes a lot of sense to be an error unless you add multiple. With 5 reversing is definitely complicated.

Source/SwiftLintFramework/Rule.swift
```diff
@@ -16,6 +17,7 @@ public protocol Rule {
 
 public protocol ParameterizedRule: Rule {
     typealias ParameterType
+    init(parameters: [RuleParameter<ParameterType>])
     var parameters: [RuleParameter<ParameterType>] { get }
```
keith: Can this be removed now?

jpsim: not really, we still need a way to access the parameters.

jpsim:
This has been rebased off master with the changes from #114 so there are now only two severity levels, warning and error.

jpsim:
Marking this as work-in-progress until I can refactor this a bit to bundle a config file in the SwiftLintFramework.framework bundle and define the default rule parameters in there rather than in code. Thanks @segiddins for the idea.

This will require the ability to "merge" configuration files (framework bundled file + custom file).

jpsim:
Turns out that refactoring to bundle a config file in the framework and adding the ability to merge config files is a bit out of scope for this PR, and is holding back a release, so I'll merge this now and create a new GitHub issue to track doing this refactoring in the future.

keith:
Sounds good!

PR #107, opened 2015-08-28
Document rule identifiers in README

jakecraige:
This removes the `todo` and fills it in
with the names of the currently defined rules.


jpsim:
why remove `todo` from in there? it's a rule just like the others

jakecraige:
Whoops! I got confused. Adding back.

jpsim:
Thanks!

PR #110, opened 2015-08-29
Allow capitalized var names for OptionSetType

wfleming:
There is at least one semi-common case new to Swift 2 where capitalized names should probably be allowed: in [OptionSetType](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Reference/Swift_OptionSetType_Protocol/index.html) classes, the standard pattern is to use `static let = Foo` to declare each option for the type.

I don't know of any common usages outside of `OptionSetType` where this is expected, since for anything like this you're probably either using `OptionSetType` or `enum`. But the context of the enclosing class isn't easily available in the rule right now, as far as I could tell, and I feel like keeping the rule a bit more general is probably better anyway.

One downside of this implementation is that it would *also* pass a capitalized "static var Foo", which is probably not desirable. It's not clear to me how feasible addressing this is: it seems like SourceKitten might need to be updated to be more precise with `SwiftDeclarationKind` values? Using `static var` at all is an anti-pattern as far as I'm concerned, though, so hopefully this isn't a huge deal. Perhaps there should be a rule identifying `static var` and considering it a violation? (I'd be happy to take a look at that either on top of this PR or separately if you think that's worthwhile.)

### DIFF
--- a/Source/SwiftLintFramework/Rules/VariableNameRule.swift
+++ b/Source/SwiftLintFramework/Rules/VariableNameRule.swift
@@ -16,11 +16,14 @@ public struct VariableNameRule: ASTRule {
         identifier: "variable_name",
         name: "Variable Name",
         description: "Variable name should only contain alphanumeric characters, " +
-        "start with a a lowercase character and be between 3 and 40 characters in length.",
+          "start with a a lowercase character and be between 3 and 40 characters in length. " +
+          "In an exception to the above, variable names may start with a capital letter when " +
+          "they are declared static and immutable.",
         nonTriggeringExamples: [
             "let myLet = 0",
             "var myVar = 0",
-            "private let _myLet = 0"
+            "private let _myLet = 0",
+            "class Abc { static let MyLet = 0 }"
         ],
         triggeringExamples: [
             "let MyLet = 0",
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Rules/VariableNameRule.swift
```diff
@@ -78,11 +79,14 @@ public struct VariableNameRule: ASTRule {
...
-        "start with a a lowercase character and be between 3 and 40 characters in length.",
+        "start with a a lowercase character and be between 3 and 40 characters in length. " +
+        "In an exception to the above, variable names may start with a capital letter when " +
+        "they are declared static and immutable.",
         nonTriggeringExamples: [
             "let myLet = 0",
             "var myVar = 0",
-            "private let _myLet = 0"
+            "private let _myLet = 0",
+            "static let MyLet = 0"
```
wfleming: This is the reason the tests are failing: it looks like when a `static let` is at a global level, SourceKitten doesn't actually consider it static? (Which is pretty smart, actually.)

I can fix this here by making it `"class Abc { static let MyLet = 0 }"`, but then it's inconsistent with the other examples. Let me know what your preferred way of dancing around this kind of thing is.

jpsim: `class Abc { static let MyLet = 0 }` is fine.

jpsim:
Sorry it's taken me so long to look at this. This seems good! Could you please update this to address the merge conflicts @wfleming? Thanks.

wfleming:
@jpsim No worries about the time. Merge conflicts addressed.

Source/SwiftLintFrameworkTests/ASTRuleTests.swift
```diff
@@ -97,6 +97,10 @@ class ASTRuleTests: XCTestCase {
         }
     }
 
+    func testPermittedCapitalizedVariableNames() {
```
jpsim: This test is unnecessary because the rule's `nonTriggeringExamples` are tested automatically.

jpsim:
This looks great, @wfleming! I also think this is the first time ever that I've suggested someone remove a unit test they wrote for their functionality :wink:.

Could you please add an entry to the changelog for this too, so others will know about it?

Then I'm happy to merge :smile:.

wfleming:
Ha, fair enough: I'm trying to remember now why I thought I should also add the unit test. I'll push the changes shortly.

wfleming:
@jpsim Pushed. For changelog I marked it as an an enhancement since I felt like this wasn't really a bug, just a choice to support a different stylistic usage that seems common in the community.

jpsim:
Thanks for the PR, very nicely done.

PR #111, opened 2015-08-30
Disable/re-enable rules from within source code comments. Fixes #4.

### LINKED ISSUE #4: Support disabling/enabling rules from source code comments

jpsim:
First draft addressing #4. Still need to write some tests, but seems to be working from what I can tell from manually testing. /cc @keith @segiddins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@
   names to start with an underscore.
   [JP Simard](https://github.com/jpsim)
 
+* Disable and re-enable rules from within source code comments using
+  `// swiftlint:disable:$IDENTIFIER` and `// swiftlint:enable:$IDENTIFIER`.  
+  [JP Simard](https://github.com/jpsim)
+  [#4](https://github.com/realm/SwiftLint/issues/4)
+
 ##### Bug Fixes
 
 * None.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
   [JP Simard](https://github.com/jpsim)
 
 * Disable and re-enable rules from within source code comments using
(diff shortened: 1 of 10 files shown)


Source/SwiftLintFramework/File+SwiftLint.swift
```diff
@@ -9,9 +9,51 @@
...
+    case Enable = "enable"
+    case Disable = "disable"
+}
+
+public typealias Command = (CommandAction, String, Int)
 
 extension File {
+    public func regions() -> [Region] {
+        let nsStringContents = contents as NSString
+        let commands = matchPattern("swiftlint:(enable|disable):force_cast")
```
segiddins: shouldn't `force_cast` be `.+` and then captured to determine which rule it is?

CHANGELOG.md
```diff
@@ -18,6 +18,11 @@
   names to start with an underscore.
   [JP Simard](https://github.com/jpsim)
 
+* Disable and re-enable rules from within source code comments using
+  `// swiftlint:disable:$IDENTIFIER` and `// swiftlint:enable:$IDENTIFIER`.
```
segiddins: i'd prefer `// swiftlint:disable force_unwrap`

Source/SwiftLintFramework/File+SwiftLint.swift
jpsim: the rule capturing was already being done by `NSScanner`, but I had accidentally hardcoded `force_cast` in here as part of my testing.

CHANGELOG.md
jpsim: that *would* make it more similar to rubocop, and I have no main reason to keep that separator a colon... so ok!

jpsim:
Good enough! :shipit:

PR #114, opened 2015-08-31
ViolationSeverity is now only `.Warning` and `.Error`

### LINKED ISSUE #113: Reduce ViolationSeverity to two members: Warning and Error
As discussed in #106, 5 severity levels is overkill, especially since Xcode can only (really) display 2: warning and error. So this should simplify things quite a bit.

jpsim:
Fixes #113.


keith:
Awesome! 👍🏻

PR #115, opened 2015-08-31
Update README .swiftlint.yml to use correct 'trailing_newline' rule identifier

daniel-beard:


jpsim:
oh hai, Daniel! thanks.

PR #117, opened 2015-09-02
Update for SourceKitten File lines

keith:
This depends on https://github.com/jpsim/SourceKitten/pull/62

This updates all of SwiftLint's calls to the now removed `lines()` function. Instead we're using the immutable cached property on File. This way we don't have to compute these as many times.

Right now this PR is pointing to my fork and the sha of that branch, I'll remove this change once that other PR is figured out.


Source/SwiftLintFramework/Rules/FileLengthRule.swift
```diff
@@ -19,7 +19,7 @@ public struct FileLengthRule: ParameterizedRule {
     ]
 
     public func validateFile(file: File) -> [StyleViolation] {
-        let lines = file.contents.lines()
+        let lines = file.lines
```
jpsim: this can just be `let lineCount = file.lines.count`

keith: Nice, updating

keith:
I've updated this to use the correct version of SourceKitten again, although I'm not sure how you want to handle that since it's pointing to master and not a specific release right now.

jpsim:
The Cartfile should just point to `github jpsim/SourceKitten, "master"` until a release is made.

keith:
I've updated the Cartfile.resolved to match master there, since master is already in the Cartfile. I saw some weirdness about the result library locally so lets see if this build passes.

keith:
Looks like I hit this carthage issue https://github.com/Carthage/Carthage/issues/135 locally because I changed the submodule URL locally, I deleted that commit and everything looks good on CI though.

Source/SwiftLintFramework/Rules/FileLengthRule.swift
```diff
@@ -19,14 +19,14 @@ public struct FileLengthRule: ParameterizedRule {
     ]
 
     public func validateFile(file: File) -> [StyleViolation] {
-        let lines = file.contents.lines()
+        let lines = file.lines.count
```
jpsim: please this variable to `lineCount` and we're in business!

jpsim:
Other than `lineCount`, :+1:

Source/SwiftLintFramework/Rules/FileLengthRule.swift
keith: Haha, I was just thinking this name was no longer obvious. Updated.

PR #118, opened 2015-09-03
Add strict linting flag

keith:
This flag makes SwiftLint exit with 1 even if there are only warnings. This is useful for CI to make sure that nothing was introduced in the current build.


jpsim:
Looks great, thanks

PR #119, opened 2015-09-03
Print violations to stderr

keith:
This means you can do something like:

```sh
swiftlint lint 1>/dev/null
```

If you want to hide the progress output, such as on CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,9 @@
   warnings.  
   [Keith Smiley](https://github.com/keith)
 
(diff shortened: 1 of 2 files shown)


jpsim:
Have you confirmed that Xcode still renders these warnings and errors in the IDE itself, even when printed to stderr?

keith:
Yep, these do still show up:

screen shot 2015-09-03 at 10 50 18

jpsim:
Cool, thanks.

PR #121, opened 2015-09-03
Optimize trailing whitespace rule

keith:
This updates the rule to no longer reverse each line in each file.
Instead the last character in each line is individually checked for
whitespace.

Before:

before

After:

after

### DIFF
--- a/Source/SwiftLintFramework/String+SwiftLint.swift
+++ b/Source/SwiftLintFramework/String+SwiftLint.swift
@@ -11,6 +11,18 @@ import SourceKittenFramework
 import SwiftXPC
 
 extension String {
+    func hasTrailingWhitespace() -> Bool {
+        if isEmpty {
+            return false
+        }
+
+        if let character = utf16.suffix(1).first {
+            return NSCharacterSet.whitespaceCharacterSet().characterIsMember(character)
+        }
+
+        return false
+    }
+
     func isUppercase() -> Bool {
         return self == uppercaseString
     }
--- a/Source/SwiftLintFramework/String+SwiftLint.swift
+++ b/Source/SwiftLintFramework/String+SwiftLint.swift
@@ -27,10 +27,6 @@ extension String {
         return self == uppercaseString
     }
 
-    func countOfTailingCharactersInSet(characterSet: NSCharacterSet) -> Int {
-        return String(characters.reverse()).countOfLeadingCharactersInSet(characterSet)
-    }
-
     public var chomped: String {
         return stringByTrimmingCharactersInSet(NSCharacterSet.newlineCharacterSet())
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,8 @@
 
 ##### Bug Fixes
 
-* None.
+* Improve performance of `TrailingWhitespaceRule`.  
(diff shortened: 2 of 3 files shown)


jpsim:
Why not something like this?

```swift
extension String {
    func hasTrailingWhitespace() -> Bool {
        if isEmpty {
            return false
        }
        guard let lastCharacter = String(self[endIndex.predecessor()]).utf16.first else {
            return false
        }
        return NSCharacterSet.whitespaceCharacterSet().characterIsMember(lastCharacter)
    }
}
```

hyperspacemark:
@jpsim Getting the last character is even simpler in Swift 2.

```swift
extension String {
    func hasTrailingWhitespace() -> Bool {
        if isEmpty {
            return false
        }
        guard let lastCharacter = utf16.suffix(1).first else {
            return false
        }
        return NSCharacterSet.whitespaceCharacterSet().characterIsMember(lastCharacter)
    }
```

keith:
I've updated this to use the suffix approach. Much nicer. I didn't move it to a string extension yet. I'm happy to do that if we want to but I'm not sure we should add more stuff to our string extension unless we expect to have more callers in the future.

keith:
With this new change this rule only takes 7ms on the same files as the tests above!

jpsim:
Thanks for the tip, @hyperspacemark. That looks much nicer. I was worried that `utf16.suffix(1)` would be slower than `String(self[endIndex.predecessor()]).utf16` because it has to calculate the utf16 view for the entire string rather than just the last character, but in practice, it seems to be ~10% faster, so :+1:

jpsim:
I'd rather have it as an extension on String though, which would make it easier to reuse later elsewhere if needed.

keith:
:+1: moved it to the string extension

hyperspacemark:
> I was worried that utf16.suffix(1) would be slower than String(self[endIndex.predecessor()]).utf16 because it has to calculate the utf16 view for the entire string rather than just the last character, but in practice, it seems to be ~10% faster, so :+1:

That is surprising. My guess is that `utf16` is a lazy sequence under the hood and since the only access on it is via `suffix(1)`, it only has to compute the last `UTF16View`.

Source/SwiftLintFramework/String+SwiftLint.swift
```diff
@@ -11,12 +11,20 @@ import SourceKittenFramework
 import SwiftXPC
 
 extension String {
-    func isUppercase() -> Bool {
-        return self == uppercaseString
+    func hasTrailingWhitespace() -> Bool {
+        if self.isEmpty {
```
jpsim: can omit `self.`

Source/SwiftLintFramework/String+SwiftLint.swift
```diff
@@ -11,12 +11,20 @@ import SourceKittenFramework
...
+            return false
+        }
+
+        if let character = self.utf16.suffix(1).first {
```
jpsim: can omit `self.`

jpsim:
> My guess is that utf16 is a lazy sequence

surprising indeed

keith:
Updated without `self.`

jpsim:
Perfect! :+1: Fantastic speed up.

PR #122, opened 2015-09-03
Allow newlines in function return arrow

jpsim:
This prevented SwiftLint from linting as of #118. /cc @keith

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,9 @@
 * Improve performance of `TrailingWhitespaceRule`.  
   [Keith Smiley](https://github.com/keith)
 
+* Allow newlines in function return arrow.  
+  [JP Simard](https://github.com/jpsim)
+
 
 ## 0.1.2: FabricSoftenerRule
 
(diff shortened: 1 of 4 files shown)


keith:
Oops! Sorry about that, shouldn't CI have failed for that PR?

keith:
:+1:

jpsim:
> shouldn't CI have failed for that PR?

Yeah, I'm also weirded out as to why the integration tests didn't fail: https://travis-ci.org/realm/SwiftLint/builds/78498728#L2679

jpsim:
hmmm, I don't know why this is failing on travis, it's passing for me locally...

keith:
I'm seeing this failing locally, it looks like `func abc()  ->  Int {}\n` is the failing example.

keith:
I'm not sure what `\h` is but just reverting your regex change and adding those 2 examples passes for me locally `¯\_(ツ)_/¯`

jpsim:
`\h` is horizontal spacing, but is a fairly recent addition to regex syntax. I'll revert that part.

jpsim:
I think I sort of figured this weirdness out... Depending on the system, `\v` matches `\n`, while `\h` may or may not be supported.

PR #124, opened 2015-09-04
Remove Location's comparison of strings

keith:
This fixes an issue with tests and regions where comparing the start and
end would return incorrect results. Really the locations just need to
compare the line and character numbers to be useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,9 @@
 * Allow newlines in function return arrow.  
   [JP Simard](https://github.com/jpsim)
 
+* Don't compare text when comparing `Location`s.  
+  [Keith Smiley](https://github.com/keith)
+
 
 ## 0.1.2: FabricSoftenerRule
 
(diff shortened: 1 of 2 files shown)


jpsim:
I don't understand the motivation for this change. Say you compare pos1 (`0/0` in `a.swift`) to pos2 (`0/0` in `b.swift`), this change will mean that both `pos1 == pos2` and `pos1 < pos2` will be false, which will make `pos1 > pos2` evaluate to true, which is definitely not what we want.

keith:
To me a position is a more generic thing than this, more like a range than a specific file position. Maybe there's a better way to solve this problem though. The problem was in https://github.com/realm/SwiftLint/pull/126 where the tests did not pass because the regions start and end location did not share the same file as the warnings being produced by this new rule.

jpsim:
Is that really why the new rule was failing? I can't think of scenarios where a single violation would span multiple files.

keith:
Ah, the issue was the ambiguity in the file initializers. So I can close this now, although I do still think Location shouldn't be tied to files.

PR #126, opened 2015-09-04
Add rule for header comments

keith:
This adds a rule to warn when files have copyright boilerplate comments. This depends on https://github.com/realm/SwiftLint/pull/124 for the tests to run successfully. There's also the issue that the integration tests for SwiftLint don't work with this. Maybe this would need to wait for the new defaults to disable this rule by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,9 @@
 * Violations are now printed to stderr.  
   [Keith Smiley](https://github.com/keith)
 
+* Add rule for header comment boilerplate.  
+  [Keith Smiley](https://github.com/keith)
+
 ##### Bug Fixes
 
 * Improve performance of `TrailingWhitespaceRule`.  
--- a/Source/SwiftLintFramework/Rule.swift
+++ b/Source/SwiftLintFramework/Rule.swift
@@ -35,5 +35,6 @@ public let allRules: [Rule] = [
     TypeBodyLengthRule(),
     FunctionBodyLengthRule(),
     NestingRule(),
-    ControlStatementRule()
+    ControlStatementRule(),
+    HeaderCommentRule(),
 ]
--- a/Source/SwiftLintFramework/Rules/HeaderCommentRule.swift
+++ b/Source/SwiftLintFramework/Rules/HeaderCommentRule.swift
@@ -1,0 +1,46 @@
+//
+//  HeaderCommentRule.swift
+//  SwiftLint
+//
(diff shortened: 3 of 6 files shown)


jpsim:
A few thoughts:

1. I don't think is a rule that should be on by default. The overwhelming majority of code written out there (not to mention Apple's default) will have copyright restrictions. If this is something important to you, maybe we should instead prioritize having a way to support opt-in rules rather than just opt-out. Or alternatively, custom rules defined in `.swiftlint.yml`.
2. This rule is implemented in a very brittle way. For example, doesn't work at all when the line has leading whitespace or if using block-comments (`/*...*/`). Instead, this rule should be limited to searching for a pattern of `\bCopyright\b` only in comment syntax tokens. See how [TodoRule.swift](https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules/TodoRule.swift) does this. Actually, that's also not 100% correct, PR incoming.

keith:
> I don't think is a rule that should be on by default.

Totally agree.

> Or alternatively, custom rules defined in `.swiftlint.yml`

I think a nice way to do this, outside of having to regex all files would be nice but I'm not sure how that would work.

> This rule is implemented in a very brittle way.

This was intentionally simple to only catch the obvious case. Regexing an entire file for something from the template didn't seem necessary.

jpsim:
@keith thoughts on this a few weeks later? I'm inclined to leave this out for now until we have support for more custom rules or a set of disabled-by-default rules.

keith:
If you don't want this in the repo we can close this. I'd rather not do this locally via regex if that ever is supported because that seems heavy handed for what' it's trying to do. I'm also skeptical that we'll be able to come up with a way to customize rules on this level without having to maintain a fork, which is a pain.

jpsim:
I'd be happy to merge in arbitrary rules that may not be generally applicable, if we supported opt-in rules rather than only having opt-out rules. This should be relatively easy to do, and would avoid having lots of forks of SwiftLint.

keith:
Yea definitely. That also might be unsustainable though, depending on how many people want random rules like this for their style guide. I'm happy to do whatever you want to do here.

jpsim:
> That also might be unsustainable though

Perhaps longer term, but it's sustainable until it's not :wink:... so until this becomes a problem I'm happy to merge in off-by-default rules.

jpsim:
Closing until we can support off-by-default rules.

Issue #125, opened 2015-09-04
Need for firstRegion


keith:
I was looking through the region code and noticed the [`firstRegion`](https://github.com/realm/SwiftLint/blob/b932cd2b242ccaecac8476f7600efbd4cb2cdc2d/Source/SwiftLintFramework/File%2BSwiftLint.swift#L21-L26). What's the need for this region with no `disabledRuleIdentifiers`?

jpsim:
[`Linter.styleViolations`](https://github.com/realm/SwiftLint/blob/56fe7b0d3b4db5fe95299b80798a99599967d9fe/Source/SwiftLintFramework/Linter.swift#L22-L27) expects all style violations to be contained in a region.

keith:
Is this true though? This guard will return true if that array is empty?

jpsim:
You're right! Feel free to refactor that in a PR.

keith:
Awesome, incoming.

PR #127, opened 2015-09-04
Remove unnecessary firstRegion property

keith:
See https://github.com/realm/SwiftLint/issues/125


jpsim:
This shouldn't have a changelog entry because it has no impact on behavior whatsoever.

jpsim:
Otherwise, :+1:

keith:
Removed!

PR #128, opened 2015-09-04
Generalize TodoRule

jpsim:
as prompted by my own recommendations... doh! https://github.com/realm/SwiftLint/pull/126#issuecomment-137846688


keith:
Hadn't submitted a PR for this yet but this also doesn't cover the Google `TODO(keith):` version.

jpsim:
How about we relax this rule even more to just be `\b(TODO|FIXME)\b` limited to comments?

keith:
:+1:

keith:
Looks good :+1:

PR #133, opened 2015-09-19
added custom reporters: xcode (default), json, csv. fixes #42

### LINKED ISSUE #42: Support custom reporters
It would be cool if there will be a Jenkins integration

jpsim:
Fixes #42.

Any interest in this from you, @keith or @jakecraige?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,12 @@
 * Violations are now printed to stderr.  
   [Keith Smiley](https://github.com/keith)
 
+* Custom reporters are now supported. Specify a value for the `reporter:` key in
+  your configuration file. Available reporters are `xcode` (default), `json`,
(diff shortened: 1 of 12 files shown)


jakecraige:
But I already wrote my clang parser!! https://github.com/thoughtbot/hound-swift/blob/master/lib/swift_lint/violation.rb#L3-L9

Now that this exists I'll probably switch to the JSON format since I trust parsing of that more than my large regex.

Thanks for this!

jpsim:
Speaking of which, how are those efforts progressing? Anything I can do to help?

jakecraige:
Things are going well. It's pretty much good to go but we still need to do some internal testing before releasing it into public beta. 

With Swift 2 being out and having configuration, that should be happening soon. :)

PR #140, opened 2015-09-30
Remove duplicated Line declaration

keith:
This typealias is public from sourcekitten, so we don't need to redeclare it here.


jpsim:
:+1:

PR #141, opened 2015-09-30
Remove allRules

keith:
Now that the Configuration struct is creating a list of rules based on optional yaml, this list duplicates the same behavior. By removing it you no longer have to maintain duplicate lists of rules.

Passing nil as yaml isn't my favorite. If anyone has a better suggestion I'd love to hear it.


jpsim:
Passing nil as YAML is fine here I think. Doesn't trigger any yuck alarms for me :stuck_out_tongue:

PR #142, opened 2015-09-30
Update rule descriptions to not include specifics

keith:
Now that configuration of rules has been added, having these hard coded lengths is no longer valid.

### DIFF
--- a/Source/SwiftLintFramework/Rules/FileLengthRule.swift
+++ b/Source/SwiftLintFramework/Rules/FileLengthRule.swift
@@ -40,7 +40,7 @@ public struct FileLengthRule: ParameterizedRule {
 
     public let example = RuleExample(
         ruleName: "File Line Length Rule",
-        ruleDescription: "Files should be less than 400 lines.",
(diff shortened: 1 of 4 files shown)


Source/SwiftLintFramework/Rules/FileLengthRule.swift
```diff
@@ -40,7 +40,7 @@ public struct FileLengthRule: ParameterizedRule {
 
     public let example = RuleExample(
         ruleName: "File Line Length Rule",
-        ruleDescription: "Files should be less than 400 lines.",
```
jpsim: In `validateFile(_:)`, I'm using `parameters.first!.value` and I think we should do the same for rule examples.

keith: Since this is a `let` we don't have access to parameters at this point.

jpsim: Right, ok then this PR is fine as-is.

PR #144, opened 2015-10-01
Update yaml parsing to allow single value configurations

keith:
This updates the `arrayOfStrings` and `arrayOfInts` methods to also
accept a single value if they are not parsed as an array. This lets us
do this for our configuration:

    line_length: 110

Instead of:

    line_length:
      - 110

If we only actually need a single value. The same goes for source
directories and other string keys.

### DIFF
--- a/Source/SwiftLintFramework/Configuration.swift
+++ b/Source/SwiftLintFramework/Configuration.swift
@@ -10,10 +10,10 @@ import Yaml
 
 extension Yaml {
     var arrayOfStrings: [Swift.String]? {
-        return array?.flatMap { $0.string }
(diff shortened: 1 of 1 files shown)


jpsim:
Great! Nice and simple.

keith:
Thanks! Not sure if want a changelog entry for this, but I can submit a new PR for one if we do.

jpsim:
> Thanks! Not sure if want a changelog entry for this, but I can submit a new PR for one if we do.

Sure. We also need some documentation around passing rule parameter values in the README, since we don't have any at all at the moment. Would you be up for doing that?

keith:
Yep! I'm working on splitting up those rules for configuration now, but I'll do this soon.

PR #152, opened 2015-10-05
Merge stdin/non-stdin code paths, only log reports to stdout, chunk json/csv reports

### LINKED ISSUE #151: Custom reporters only work when specifying the --use-stdin command
- Custom reporters do not work when calling just the `swiftlint` command with a folder that contains a valid `.swiftlint.yml` file

jpsim:
Fixes #151.

### DIFF
--- a/Source/SwiftLintFramework/Configuration.swift
+++ b/Source/SwiftLintFramework/Configuration.swift
@@ -6,6 +6,7 @@
 //  Copyright (c) 2015 Realm. All rights reserved.
 //
 
+import Foundation
 import Yaml
 
 extension Yaml {
@@ -112,7 +113,7 @@ public struct Configuration {
                 let yamlContents = try NSString(contentsOfFile: fullPath,
                     encoding: NSUTF8StringEncoding) as String
                 if let _ = Configuration(yaml: yamlContents) {
-                    print("Loading configuration from '\(path)'")
+                    fputs("Loading configuration from '\(path)'\n", stderr)
                     self.init(yaml: yamlContents)!
                 } else {
                     self.init()!
--- a/Source/swiftlint/LintCommand.swift
+++ b/Source/swiftlint/LintCommand.swift
@@ -28,10 +28,9 @@ struct LintCommand: CommandType {
                 let stdinData = standardInput.readDataToEndOfFile()
                 let stdinNSString = NSString(data: stdinData, encoding: NSUTF8StringEncoding)
                 if let stdinString = stdinNSString as? String {
-                    let file = File(contents: stdinString)
-                    let linter = Linter(file: file, configuration: configuration)
-                    print(linter.generateReport())
-                    return .Success()
+                    return lint([File(contents: stdinString)],
+                        configuration: configuration,
+                        strict: options.strict)
                 }
                 return .Failure(CommandantError<()>.CommandError())
             }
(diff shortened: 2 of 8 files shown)


Source/SwiftLintFramework/Configuration.swift
```diff
@@ -112,7 +113,7 @@ public struct Configuration {
                 let yamlContents = try NSString(contentsOfFile: fullPath,
                     encoding: NSUTF8StringEncoding) as String
                 if let _ = Configuration(yaml: yamlContents) {
-                    print("Loading configuration from '\(path)'")
+                    fputs("Loading configuration from '\(path)'\n", stderr)
```
keith: I didn't change this one originally since it didn't really seem like an "error" thoughts?

jpsim: I don't think custom reporters are particularly useful when format-breaking logs are printed to the same file descriptor. E.g. it's tricky to parse JSON from stdout if you first have to filter out the non-JSON parts.

Source/swiftlint/LintCommand.swift
```diff
@@ -41,47 +40,63 @@ struct LintCommand: CommandType {
...
+        }
+        let numberOfSeriousViolations = violations.filter({ $0.severity == .Error }).count
+        let violationSuffix = (violations.count != 1 ? "s" : "")
+        let filesSuffix = (files.count != 1 ? "s." : ".")
+        fputs(
+            "Done linting!" +
+            " Found \(violations.count) violation\(violationSuffix)," +
+            " \(numberOfSeriousViolations) serious" +
+            " in \(files.count) file\(filesSuffix)\n",
+            stderr
```
keith: same comment about this one and stderr

jpsim: Same response regarding custom reporters :smile:

Source/swiftlint/LintCommand.swift
```diff
@@ -41,47 +40,63 @@ struct LintCommand: CommandType {
...
-                print("Linting '\(filename)' (\(index + 1)/\(filesToLint.count))")
-                let file = File(path: path)!
-                for violation in Linter(file: file, configuration: configuration).styleViolations {
-                    fputs("\(violation)\n", stderr)
-                    numberOfViolations++
-                    if violation.severity == .Error {
-                        numberOfSeriousViolations++
-                    }
-                }
+                fputs("Linting '\(filename)' (\(index + 1)/\(files.count))\n", stderr)
```
keith: The problem with changing all of these to stderr is that it defeats the purpose of this in the first place. I made that change so we could filter out all the `Linting...` lines by redirecting just stdout to dev null

jpsim: I think you have this flipped, since the opposite is now true: all informational logs can now be silenced by `2> /dev/null`, so you can now pipe the result of swiftlint like this: `swiftlint > violations.json` or `swiftlint > violations.csv`.

keith: It seems kind of weird to direct errors to stdout and output to stderr?

jpsim: I see what you mean, but in the case of a linter, the main content *are* the violations. [Rubocop](https://github.com/bbatsov/rubocop) also prints its violations to stdout, for example.

PR #157, opened 2015-10-07
Fix TrailingNewLine rule line index

### LINKED ISSUE #156: TrailingNewLine rule location
For trailing newline rule, we give a line location of `file.lines.count + 1`.
This means that in this case, violations can't be mapped correctly to file contents.
For example:

```
1 +
2 +import Foundation
3 +
4 +class TestClass : NSObject {
5 +
6 +}
7 +
8 +
9 +
```

The trailing newline output gives us a line number of `10` for this file. Should we change this to output `file.lines.count` (9) instead?

daniel-beard:
Fixes #156


jpsim:
The failing test indicates why this is currently `+1`. In reality, I think this should be `max(file.lines.count, 1)`.

jpsim:
Much better! Thanks for the fix!

daniel-beard:
:+1: We are using swiftlint extensively, thanks for all the hard work :)

PR #165, opened 2015-10-22
Make TypeNameRule Parameterized

roman-dzieciol:
TypeNameRule now accepts configuration for minimum and maximum type name length.
Example yml configuration:
```
type_name:
 - 2
 - 100
```

### DIFF
--- a/Source/SwiftLintFramework/Configuration.swift
+++ b/Source/SwiftLintFramework/Configuration.swift
@@ -145,7 +145,11 @@ public struct Configuration {
         }
         rules.append(TodoRule())
         rules.append(ColonRule())
-        rules.append(TypeNameRule())
+        if let params = yaml?[.String(TypeNameRule().identifier)].arrayOfInts {
+            rules.append(TypeNameRule(parameters: ruleParametersFromArray(params)))
+        } else {
+            rules.append(TypeNameRule())
+        }
         rules.append(VariableNameRule())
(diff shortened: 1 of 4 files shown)


Tableau-David-Potter:
This looks like a decent implementation.  This same thing should be done for variable names.  I implemented something similar but much more heavy weight, so I like the simplicity of this.

I have a couple comments:
- It isn't obvious what the values represent because the parsing logic only supports numbers as implemented right now.  It may clear enough, though.  Something to think about.
- Other parameterized rules specify the first as a warning and the second as an error.  This one is different.  That difference may or may not be important.

@jpsim what do you think?

Tableau-David-Potter:
Hmm, as I think about this more, I don't really like that I can't disable parts of the rule.  My solution was to implement a min and a max for both types and variables, which is really heavy-weight, but at least the user gets a lot of control.

By the way @roman-dzieciol, the likely reason this is failing the continuous integration is because you have lines that are longer than 100 characters.

jpsim:
@roman-dzieciol thanks for the PR! It's not clear to me however why you'd want to limit how short lines should be. Could you explain your rationale here?

Also, if there are valid, general reasons why a minimum line length rule should be enforced, it'd be preferable to extract that as a separate rule so it can be disabled & parameterized separately.

Thanks!

jpsim:
Closing since there's been no activity since November. Feel free to re-open a PR if you'd like to revive this.

PR #167, opened 2015-10-24
Add support for the guard statement to ControlStatementRule.

Tableau-David-Potter:
This change allows SwiftLint to catch parentheses around conditions in guard statements which were added in Swift 2.


Tableau-David-Potter:
It looks like the Integration Tests are failing.  This change is super simple, so not sure what to do here.

Tableau-David-Potter:
I found the problem.  This should pass now.

jpsim:
Thanks for the PR, @Tableau-David-Potter!

PR #170, opened 2015-10-26
Opening brace rule

S2dentik:
Add rule for opening brace, one space and same line when in declaration and same for closures.


keith:
Looks like you got a commit and some files in this diff that you weren't expecting.

S2dentik:
Well, I have added some changes that are needed like multiline comment instead of single line inside checks, and some new functions that help filter the checking when inside strings and comments, but outside some files like .DS_Store they are needed.

PR #171, opened 2015-10-27
Add OpeningBrace Rule

S2dentik:
Added support for opening brace after to be located one space after declaration and on the same line.


Source/SwiftLintFramework/Configuration.swift
```diff
@@ -152,13 +152,15 @@ public struct Configuration {
         } else {
             rules.append(TypeBodyLengthRule())
         }
-        if let params = yaml?[.String(FunctionBodyLengthRule().identifier)].arrayOfInts {
+        if let params   = yaml?[.String(FunctionBodyLengthRule().identifier)].arrayOfInts {
```
jpsim: why was this extra space added?

jpsim:
This looks good, @S2dentik! Could you please add an entry to the CHANGELOG.md and address the extra spacing that was added? Thanks!

keith:
How does/should this affect the case where you have to put the opening bracket on another line because of the line length restrictions?

```swift
func foo...max...chars
{
```

It would probably be ok if it forced an arg on the next line?

jpsim:
How do you suggest we codify that, @keith?

keith:
That's a great question :smile: posing the question right now because that could generate unintended warnings if we're not handling it.

S2dentik:
@keith, I guess line length rule should handle this, well not exactly this, it should leave the choice to the programmer on how to change the code based on the required style, but it shouldn't affect this rule if it is being checked, as it is superior.

PR #176, opened 2015-11-01
Added rule reference on command output to enable batch processing

zippy1978:
Hi,

In this pull request I added a Rule reference to StyleViolation in order to be able to retrieve the rule matching a violation when generating a report.

I also added a rule id field on the 3 reporters.

Finally: I slightly modified the 'rules' command output in order to make it easier to process as well.

My goal is to integrate SwiftLint to a SonarQube plugin for Swift

### DIFF
--- a/Source/SwiftLintFramework/Reporters/CSVReporter.swift
+++ b/Source/SwiftLintFramework/Reporters/CSVReporter.swift
@@ -23,7 +23,8 @@ public struct CSVReporter: Reporter {
             "character",
             "severity",
             "type",
-            "reason"
+            "reason",
+            "rule_id"
         ]
         return (keys + violations.flatMap(arrayForViolation)).joinWithSeparator(",")
     }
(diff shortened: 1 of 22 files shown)


jpsim:
This looks great, @zippy1978! I have two thoughts:

1. Instead of passing a `Rule` as part of the `StyleViolation` initializer, why not just the identifier as a `String`?
2. At this point we're duplicating a lot of information in the `ViolationType` along with the `ruleName` in `RuleExample` and the rule identifier. We should probably consolidate those instead of continuing to duplicate that information.

At the very least, I'd like this PR to do 1), maybe 2) can be done at a later time.

zippy1978:
OK,

So I replaced the rule: Rule? reference in StyleViolation by a simple String ruleId: String?
Is that ok for 1) ?

For 2), I wish I can help but I'm not sure to get what you mean : do you want to move some RuleExample  properties (ruleName and ruleDescription) to Rule ?

jpsim:
This is great for now, @zippy1978. Thanks for the PR!

zippy1978:
Thank you !

Do you have any idea of the next release date ?

jpsim:
I hope to get a release out today, as it's been just over a month since the last one.

jpsim:
[0.3.0: Wrinkly Rules](https://github.com/realm/SwiftLint/releases/tag/0.3.0) is now out and submitted to Homebrew: https://github.com/Homebrew/homebrew/pull/45650.

PR #177, opened 2015-11-03
Add OpeningBrace Rule

S2dentik:
Add rule for checking if opening brace is on the same line as declaration and after one space. (Remade branch not to force ugly updates on commit history).
As of what @keith said, I have commented on the commit I have closed, please answer whether I should change anything in here.


jpsim:
Thanks for the new rule, @S2dentik!

PR #184, opened 2015-11-04
Fixed disable/enable rules documentation in the README

Noobish1:
The disable/enabled documentation in the README was incorrect, it should have 2 slashes instead of 3.


jpsim:
Thanks! I'll also fix this to support 3 documentation style comments though.

PR #185, opened 2015-11-04
Perform major refactor of rule description, identifiers and examples

### LINKED ISSUE #183: Remove StyleViolationType
This mostly just duplicates rules 1 for 1, so I really can't see any reason to keep it around.

jpsim:
Fixes #183. This deduplicates a lot of information and drastically helps with readability & rule maintenance IMO.

Do you have any thoughts on this @keith? One thing I'm not convinced about is the fact that `Rule.description` is a static method, but my rationale there is to avoid needing an instance of the type when you just need its description.


keith:
I think it makes sense to be static, even though it makes the caller's syntax a little less nice.

jpsim:
Cool, any other feedback?

keith:
Nope, code looks good to me!

--
Keith Smiley

PR #186, opened 2015-11-04
move ASTRule function implementations into a protocol extension

jpsim:
I should have done this when moving to Swift 2.0.


keith:
Nice!

PR #194, opened 2015-11-11
Apply isSwiftFile filter on using scriptInputFiles

norio-nomura:
`SCRIPT_INPUT_FILE_*` contains not only Swift files.
ExtraBuildPhase also does not check file type on providing them.


jpsim:
:+1:

Issue #187, opened 2015-11-05
Control Statement Parentheses Violation not detecting all violations


daniel-beard:
swiftlint version 0.3.0

Reproducible case:

```
import Foundation

class TestClass: TestObject {
    
    func enterUserFirstLastName(firstName: String,lastName: String) {
        let clearTextButton = app.buttons[kClearText]
        let firstNameField = app.tables.textFields[kFirstName]
        let lastNameField = app.tables.textFields[kLastName]
        firstNameField.tap()
        if(clearTextButton.exists) {
            clearTextButton.tap()
        }
        firstNameField.typeText(firstName)
        lastNameField.tap()
        if(clearTextButton.exists) {
            clearTextButton.tap()
        }
        lastNameField.typeText(lastName)
    }
}
```

Swiftlint output:

```
Linting Swift files in current working directory
Linting 'test.swift' (1/1)
/Users/dbeard/iphone/test.swift:1: warning: Leading Whitespace Violation: File shouldn't start with whitespace: currently starts with 1 whitespace characters
/Users/dbeard/iphone/test.swift:5: warning: Trailing Whitespace Violation: Line #5 should have no trailing whitespace
/Users/dbeard/iphone/test.swift:11:9: warning: Control Statement Parentheses Violation: if statements shouldn't wrap their conditionals in parentheses.
Done linting! Found 3 violations, 0 serious in 1 file.
```

Swiftlint picks up the first violation, but does not detect the second control statement violation.

Issue #189, opened 2015-11-08
ControlStatementRule failing on acceptable use of parentheses


mattrubin:
`ControlStatementRule` is showing a warning on the following piece of code. I think this is an acceptable use of parens in an `if` statement, and the rule should be updated to accept it.

```swift
if (minValue...maxValue).contains(value) {
    // Do something...
}
```
> Control Statement Parentheses Violation: if statements shouldn't wrap their conditionals in parentheses. (control_statement)

jpsim:
Agreed. PRs welcome if you'd like to address this "false positive".

mmorier:
I also notice this issue for the following case where the cause is the same.
```swift
if (isA || isB) && (isC || isD) {
    // Do something...
}
```
I'll try to find a way to fix it and make a PR.

jpsim:
> I'll try to find a way to fix it and make a PR.

Great! Thanks for your help.

PR #195, opened 2015-11-11
Improve ControlStatementRule for issues #187 and #189

mmorier:
I fix false positive for `ControlStatementRule` that we have notice in issue #189. To do that, I had to check condition's depth.

I also fix issue #187 with this PR in remove closing curly bracket from regex.

I had some function in `String+Swiftlint` to make easier substring and index search.

### DIFF
--- a/Source/SwiftLintFramework/Rules/ControlStatementRule.swift
+++ b/Source/SwiftLintFramework/Rules/ControlStatementRule.swift
@@ -77,7 +77,10 @@ public struct ControlStatementRule: Rule {
             return true
         }
 
-        let lastClosingParenthesePosition = content.lastIndexOf(")")
+        guard let lastClosingParenthesePosition = content.lastIndexOf(")") else {
+            return false
+        }
+
         var depth = 0
         var index = 0
         for char in content.characters {
--- a/Source/SwiftLintFramework/Rules/ControlStatementRule.swift
+++ b/Source/SwiftLintFramework/Rules/ControlStatementRule.swift
@@ -18,6 +18,8 @@ public struct ControlStatementRule: Rule {
         nonTriggeringExamples: [
             "if condition {\n",
             "if (a, b) == (0, 1) {\n",
+            "if (a || b) && (c || d) {\n",
+            "if (min...max).contains(value) {\n",
             "if renderGif(data) {\n",
             "renderGif(data)\n",
             "for item in collection {\n",
@@ -28,11 +30,13 @@ public struct ControlStatementRule: Rule {
             "while condition {\n",
             "} while condition {\n",
             "do { ; } while condition {\n",
-            "switch foo {\n",
+            "switch foo {\n"
         ],
         triggeringExamples: [
             "if (condition) {\n",
             "if(condition) {\n",
+            "if ((a || b) && (c || d)) {\n",
+            "if ((min...max).contains(value)) {\n",
             "for (item in collection) {\n",
             "for (var index = 0; index < 42; index++) {\n",
             "for(item in collection) {\n",
@@ -44,7 +48,7 @@ public struct ControlStatementRule: Rule {
             "} while(condition) {\n",
             "do { ; } while(condition) {\n",
             "do { ; } while (condition) {\n",
-            "switch (foo) {\n",
+            "switch (foo) {\n"
         ]
     )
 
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/String+SwiftLint.swift
```diff
@@ -38,6 +38,26 @@ extension String {
...
+            return substringWithRange(Range(start: subStart, end: subEnd))
+        }
+    }
+
+    func substring(from: Int, length: Int? = nil) -> String {
+        let length = length ?? characters.count
+        return self[from..<from + length]
+    }
+
+    public func lastIndexOf(search: String) -> Int {
```
jpsim: why not return `-> Int?`, returning `nil` instead of `-1` when not found?

Source/SwiftLintFramework/String+SwiftLint.swift
```diff
@@ -38,6 +38,26 @@ extension String {
...
+
+    subscript (range: Range<Int>) -> String {
+        get {
+            let subStart = startIndex.advancedBy(range.startIndex, limit: endIndex)
+            let subEnd = subStart.advancedBy(range.endIndex - range.startIndex, limit: endIndex)
+            return substringWithRange(Range(start: subStart, end: subEnd))
+        }
+    }
+
+    func substring(from: Int, length: Int? = nil) -> String {
```
jpsim: You can avoid calculating the count of all characters by doing this:

```swift
func substring(from: Int, length: Int? = nil) -> String {
    if let length = length {
        return self[from..<from + length]
    }
    return substringFromIndex(startIndex.advancedBy(from, limit: endIndex))
}
```

Source/SwiftLintFramework/Rules/ControlStatementRule.swift
```diff
@@ -64,5 +69,28 @@ public struct ControlStatementRule: Rule {
                     "parentheses.")
                 }
         }
+
+    }
+
+    private func isFalsePositive(content: String, syntaxKind: SyntaxKind?) -> Bool {
```
jpsim: very cool implementation!

Source/SwiftLintFramework/Rules/ControlStatementRule.swift
```diff
@@ -64,5 +69,28 @@ public struct ControlStatementRule: Rule {
...
+        if syntaxKind != .Keyword {
+            return true
+        }
+
+        let lastClosingParenthesePosition = content.lastIndexOf(")")
```
jpsim: This algorithm is pretty wasteful right now if `content.lastIndexOf(")")` is `-1`. If you change that function to return `nil` on not found, you could have `guard let lastClosingParenthesePosition = content.lastIndexOf(")") else { return false }` here.

jpsim:
I like what you're doing here! I just had a few minor comments that I'd like you to address.

Could you also please add an entry to the CHANGELOG.md? Thanks!

Source/SwiftLintFramework/String+SwiftLint.swift
mmorier: Good point, dunno why I'm not use optional benefit for this function. I'll change it now.

Source/SwiftLintFramework/String+SwiftLint.swift
mmorier: Two more lines to write but more efficient, I'll change it.

Source/SwiftLintFramework/Rules/ControlStatementRule.swift
mmorier: well seen, I change to early exit with guard statement.

Source/SwiftLintFramework/Rules/ControlStatementRule.swift
mmorier: thanks for your review. I'm just push my changes.

mmorier:
Thank you. I like coded on this tool:-)
I forgot to add entries to the `CHANGELOG.md`. Wait a minute, I'll make it for my both PR.

jpsim:
This is great. Thanks for the hard work, @mmorier. I've made you a SwiftLint contributor which should make it easier for you to contribute in the future :smile:

mmorier:
Thanks!

Issue #197, opened 2015-11-12
Unknown rule in `.swiftlint.yml` makes SwiftLint ignore the file and run


dpassage:
If you have an unknown rule in the `disabled` section of your `.swiftlint.yml` file, swiftlint prints a nice message - and then proceeds to run with an empty configuration.

Found this because I'd added lines to ignore new rules in version 0.3.0 but our CI server had 0.2.0 installed. Many hundreds of lines of output resulted as SwiftLint descended into all our dependencies...

Happy to work on a PR. @jpsim what do you think the right behavior is - bail out entirely, or just ignore the unknown rule?

jpsim:
> ignore the unknown rule?

I'd say this would be the right move. Thanks for offering to make a PR for this!

jpsim:
Fixed in #198.

PR #198, opened 2015-11-13
Filter out rules which don't exist rather than failing the configuration

dpassage:
While I'm at it, only print the list of valid rules once, not once per bogus rule.

This should address #197.

Thanks again for a great tool!

### DIFF
--- a/Source/SwiftLintFrameworkTests/ConfigurationTests.swift
+++ b/Source/SwiftLintFrameworkTests/ConfigurationTests.swift
@@ -32,8 +32,6 @@ class ConfigurationTests: XCTestCase {
     }
 
     func testDisabledRules() {
-        XCTAssert(Configuration(yaml: "disabled_rules:\n  - a") == nil,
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFrameworkTests/ConfigurationTests.swift
```diff
@@ -51,4 +49,22 @@ class ConfigurationTests: XCTestCase {
         XCTAssert(duplicateConfig == nil, "initializing Configuration with duplicate rules in " +
             " YAML string should fail")
     }
+
+    func testDisabledRulesWithUnknownRule() {
+        let validRule = "nesting"
+        let bogusRule = "no_sprites_with_elf_shoes"
```
jpsim: now we have to make this rule :wink: :mans_shoe:

jpsim:
Thanks for the great PR! Concise, smart modifications to the tests, really nice work :clap:

dpassage:
You're welcome, happy to help! :tada:

Issue #135, opened 2015-09-23
Colon rule not triggered when badly formatted assignment operator is used in declaration statement


mgrebenets:
Colon rule is not triggered if assignment operator is used in declaration statement and that assignment operator is badly formatted, e.g. no spaces around `=`.

```swift
// This triggers colon rule
let x :Int = 0
let x2:Int = 0

// This does not trigger colon rule
let y :Int=0
let y2:Int=0
```

tom-sparo:
Are you sure?

<img width="709" alt="screen shot 2015-10-17 at 10 45 40 am" src="https://cloud.githubusercontent.com/assets/12956264/10559665/4574db7a-74bc-11e5-8d69-66a5c7344e78.png">

mmorier:
Fixed in master. Reopen if you stil have the problem.

PR #199, opened 2015-11-15
Fix colon rule issue #135

mmorier:
`ColonRule` does not trigger all colon bad format especially when no spaces surround `=` operator.

Regex was good but there are too much restrictions on SyntaxKind. Only [.Identifier, .Typeidentifier] was accepted instead of all array begin with these kinds such as:
* [.Identifier, .Typeidentifier, .String] for string value assignment
* [.Identifier, .Typeidentifier, .Number] for number value assignment
* [.Identifier, .Typeidentifier, .Identifier, .Identifier] for enumeration assignment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,11 @@
   [JP Simard](https://github.com/jpsim)
   [#190](https://github.com/realm/SwiftLint/issues/190)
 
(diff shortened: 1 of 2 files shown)


jpsim:
:+1: thanks for the fix!

PR #203, opened 2015-11-17
add trailing semicolon rule

jpsim:


Source/SwiftLintFrameworkTests/StringRuleTests.swift
```diff
@@ -96,4 +96,8 @@ class StringRuleTests: XCTestCase {
     func testStatementPosition() {
         verifyRule(StatementPositionRule.description)
     }
+
+    func testTrailingSemicolon() {
+        verifyRule(TrailingSemicolonRule.description, commentDoesntViolate: false)
```
segiddins: comments shouldn't violate, we want to encourage good grammar in comments!

jpsim: It was trickier to implement, but it's done now. Thanks for the push!

Source/SwiftLintFrameworkTests/TestHelpers.swift
```diff
@@ -29,15 +28,14 @@ extension XCTestCase {
             ruleDescription.triggeringExamples.flatMap({
                 violations($0, ruleDescription).map({$0.ruleDescription})
             }),
-            Array(count: ruleDescription.triggeringExamples.count, repeatedValue: ruleDescription))
+            Array(count: ruleDescription.triggeringExamples.count, repeatedValue: ruleDescription)
+        )
 
         if commentDoesntViolate {
```
segiddins: maybe should also test that, if a comment does violate, the violation still happens inside a comment?

PR #209, opened 2015-11-17
Test comments cause violations when they should

### LINKED ISSUE #208: test that violation still happens when commented if commentDoesntViolate is false in verifyRule
Thanks @segiddins for the idea in https://github.com/realm/SwiftLint/pull/203#discussion_r45130602

segiddins:
Closes #208

### DIFF
--- a/Source/SwiftLintFrameworkTests/StringRuleTests.swift
+++ b/Source/SwiftLintFrameworkTests/StringRuleTests.swift
@@ -46,7 +46,7 @@ class StringRuleTests: XCTestCase {
     }
 
     func testFileShouldntStartWithWhitespace() {
-        verifyRule(LeadingWhitespaceRule.description, commentDoesntViolate: false)
+        verifyRule(LeadingWhitespaceRule.description, commentDoesntViolate: true)
     }
 
     func testLinesShouldntContainTrailingWhitespace() {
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFrameworkTests/StringRuleTests.swift
```diff
@@ -46,7 +46,7 @@ class StringRuleTests: XCTestCase {
     }
 
     func testFileShouldntStartWithWhitespace() {
-        verifyRule(LeadingWhitespaceRule.description, commentDoesntViolate: false)
+        verifyRule(LeadingWhitespaceRule.description, commentDoesntViolate: true)
```
jpsim: the parameter has a default value of `true` so you can just omit it.

segiddins: can we add a rule for that?

Source/SwiftLintFrameworkTests/TestHelpers.swift
```diff
@@ -33,9 +33,18 @@ extension XCTestCase {
 
         if commentDoesntViolate {
             let commentedViolations = ruleDescription.triggeringExamples.flatMap {
-                violations("/** " + $0, ruleDescription)
+                violations("/**\n  " + $0 + "\n */", ruleDescription)
             }
             XCTAssertEqual(commentedViolations, [])
+        } else {
+            let commentedViolations = ruleDescription.triggeringExamples.flatMap {
```
jpsim: you could reuse the `commentedViolations` and `XCTAssertEqual(commentedViolations, Array(count: commentDoesntViolate  ? 0 : ruleDescription.triggeringExamples.count, repeatedValue: ruleDescription)`.

jpsim:
:+1: thanks Sam!

PR #212, opened 2015-11-18
add force try rule

jpsim:
Made SwiftLint better in the process! :smile:

### DIFF
--- a/Source/SwiftLintFramework/Rules/ForceTryRule.swift
+++ b/Source/SwiftLintFramework/Rules/ForceTryRule.swift
@@ -1,0 +1,32 @@
+//
+//  ForceTryRule.swift
+//  SwiftLint
+//
+//  Created by JP Simard on 2015-11-17.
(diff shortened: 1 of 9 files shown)


Source/SwiftLintFramework/Rules/ForceTryRule.swift
```diff
@@ -0,0 +1,32 @@
...
+
+public struct ForceTryRule: Rule {
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "force_try",
+        name: "Force Try",
+        description: "Force tries should be avoided.",
+        nonTriggeringExamples: [
+            "func a() throws {}; do { try a() } catch {}"
```
segiddins: Should also make sure try! In a string doesn't trigger

jpsim: #214

PR #216, opened 2015-11-18
Fix an issue where logs would be printed asynchronously

### LINKED ISSUE #200: Command /bin/sh emitted errors but did not return a nonzero exit code to indicate failure
I'm experiencing an issue with SwiftLint. It breaks the build even if I have lint warning but no lint errors. I run the following script in Xcode build phases:
```shell
if which swiftlint >/dev/null; then
    swiftlint
else
    echo "SwiftLint does not exist, download from https://github.com/realm/SwiftLint"
fi
```
In terminal, logs seems good when SwiftLint is launched only but it emitted errors when I build my project with xctool in Terminal. I suspect issue comes from a file which have 77 lint warning but don't know why it emitted shell error. Any idea?

Here a subset of Xcode messages which shows strange logs `waLinting` and `169 files. rning: Control`:
```
[...]/InfiniLog.swift:20:7: warning: Control Statement Violation: if statements shouldn't wrap their conditionals in parentheses. (control_statement)
[...]/InfiniLog.swift:28:7: waLinting 'InfiniUtil.swift' (120/169)
Linting 'QuizzCategoriesButton.swift' (121/169)
[...]
Linting 'PlayerControlsViewTests.swift' (169/169)
Done linting! Found 191 violations, 0 serious in 169 files.
rning: Control Statement Violation: if statements shouldn't wrap their conditionals in parentheses. (control_statement)
[...]
[...]/PlayerControls/PlayerControlsView.swift:352:8: warning: Todo Violation: TODOs and FIXMEs should be avoided (todo)
Command /bin/sh emitted errors but did not return a nonzero exit code to indicate failure
```

jpsim:
Fixes #200. I think. /cc @mmorier @keith

Code mostly copied over from https://github.com/Carthage/Carthage/blob/741a1e0553c770088518edadbcf1c5c40a919344/Source/carthage/Extensions.swift#L19-L49.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,10 @@
   [Mickael Morier](https://github.com/mmorier)
   [#135](https://github.com/realm/SwiftLint/issues/135)
 
+* Fix an issue where logs would be printed asynchronously over each other.  
+  [JP Simard](https://github.com/jpsim)
+  [#200](https://github.com/realm/SwiftLint/issues/200)
+
 
 ## 0.3.0: Wrinkly Rules
 
--- a/Source/SwiftLintFramework/Extensions/QueuedPrint.swift
+++ b/Source/SwiftLintFramework/Extensions/QueuedPrint.swift
@@ -1,0 +1,34 @@
+//
+//  QueuedPrint.swift
+//  SwiftLint
+//
+//  Created by JP Simard on 2015-11-17.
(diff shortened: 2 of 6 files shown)


keith:
Seems good `¯\_(ツ)_/¯`

mmorier:
Great

Issue #213, opened 2015-11-18
Wrong line number and/or index is reported.


crimsonwoods:
Wrong location of detected violation is reported when a file contains Japanese comment or string literal.

Reproducible input:
```Swift
import Foundation

class Sample {
    func sample_for_reproduction(param: AnyObject?) -> Int {
        // 日本語コメント <- Japanese
        return param as! Int // force cast violation
    }   
}
```

I think that the cause of this issue is misuse of `offset` parameter supplied to `Location` class, maybe.
In `NSString.lineAndCharacterForByteOffset`, `offset` is being used as a byte-offset. On the other hand, the offset that is being used to instantiate new `Location` object is a character-offset.

jpsim:
Thanks for reporting this. I can confirm your assessment of the issue.

crimsonwoods:
I wrote a patch like below to fix this issue. but I don't know what the patch is good or not.

```diff
--- a/Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
@@ -64,18 +64,17 @@ extension String {
 
 extension NSString {
     public func lineAndCharacterForByteOffset(offset: Int) -> (line: Int, character: Int)? {
-        return byteRangeToNSRange(start: offset, length: 0).flatMap { range in
-            var numberOfLines = 0, index = 0, lineRangeStart = 0, previousIndex = 0
-            while index < length {
-                numberOfLines++
-                if index > range.location {
-                    break
-                }
-                lineRangeStart = numberOfLines
-                previousIndex = index
-                index = NSMaxRange(lineRangeForRange(NSRange(location: index, length: 1)))
+        let range = NSRange(location: offset, length: 0)
+        var numberOfLines = 0, index = 0, lineRangeStart = 0, previousIndex = 0
+        while index < length {
+            numberOfLines++
+            if index > range.location {
+                break
             }
-            return (lineRangeStart, range.location - previousIndex + 1)
+            lineRangeStart = numberOfLines
+            previousIndex = index
+            index = NSMaxRange(lineRangeForRange(NSRange(location: index, length: 1)))
         }
+        return (lineRangeStart, range.location - previousIndex + 1)
     }
 }
```

jpsim:
That seems right to me, @crimsonwoods! I'd appreciate a PR and ideally some unit tests with your patch. Would you be able to do that?

crimsonwoods:
[...]

PR #217, opened 2015-11-18
Fix an issue: wrong line number and/or index is reported

crimsonwoods:
This PR will fix issue #213.

For a file containing multi-byte-characters (like Japanese-Hiragana, Kanji and more), `offset` parameter supplied to `NSString.lineAndCharacterForByteOffset` should be handled as a character-offset.

### DIFF
--- a/Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
@@ -63,7 +63,7 @@ extension String {
 }
 
 extension NSString {
-    public func lineAndCharacterForByteOffset(offset: Int) -> (line: Int, character: Int)? {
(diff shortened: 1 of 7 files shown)


jpsim:
Thanks for the PR!!! Shouldn't we rename `lineAndCharacterForByteOffset` to `lineAndCharacterForCharacterOffset` though?

crimsonwoods:
We should do. I add a change to rename soon. And I might rebase this PR.

crimsonwoods:
`lineAndCharacterForByteOffset` is renamed to `lineAndCharacterForCharacterOffset`.
And this PR is rebased.

@jpsim Would you review again?

jpsim:
Looks good to me, how about you @keith?

keith:
:shipit:

Issue #70, opened 2015-05-29
Validate use of $0 only if it's "simple"


chriseidhof:
e.g. if it's the first thing that's switched on, or if the closure is really simple (not many tokens).

jpsim:
:+1:

segiddins:
Maybe only allow for single-line closures by default?

chriseidhof:
Yeah... I often do this pattern:

```swift
blablaFunc {
    switch $0 {
        // multiple cases
    }
}
```

But maybe that's very specific. I guess the important thing is that the `$0` is very close to the corresponding opening brace. But having them only in single-line closures is a good first step.

keith:
I've implemented a rule for this, but it's pretty simple. It just validates that all `$0`s have a `{` before them and a `}` after on the same line. This does mean that it doesn't catch all cases, something like:

```swift
foo.map {
     bar.something { $0 }
}
```

Wouldn't fail, but we feel like most of those cases will be caught by our line length as well. I can submit this upstream if the leniency is alright.

jpsim:
@keith could you make a PR with your single-line rule? I can't guarantee we'd merge it right away, but it would be a good starting point, at least to stir a conversation.

keith:
Yep, I'll do that soon.

keith:
Submitted: https://github.com/realm/SwiftLint/pull/218

PR #218, opened 2015-11-18
Add multiline closure argument rule

keith:
This rule forces users to name arguments for multiline closures.

This is the first draft of https://github.com/realm/SwiftLint/issues/70

See the first commit for the rule. The second is just fixes for it.

### DIFF
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -59,10 +59,10 @@ extension File {
         let syntax = syntaxMap
         let matches = regex(pattern).matchesInString(contents, options: [], range: range)
         return matches.map { match in
-            let tokensInRange = syntax.tokens.filter {
-                NSLocationInRange($0.offset, match.range) ||
+            let tokensInRange = syntax.tokens.filter { token in
+                NSLocationInRange(token.offset, match.range) ||
                     NSLocationInRange(match.range.location,
-                        NSRange(location: $0.offset, length: $0.length))
+                        NSRange(location: token.offset, length: token.length))
             }.map { $0.type }
             let kindsInRange = tokensInRange.flatMap(SyntaxKind.init)
             return (match.range, kindsInRange)
@@ -87,10 +87,10 @@ extension File {
         let syntax = syntaxMap
         let matches = regex(pattern).matchesInString(contents, options: [], range: range)
         return matches.filter { match in
-            let tokensInRange = syntax.tokens.filter {
-                NSLocationInRange($0.offset, match.range) ||
+            let tokensInRange = syntax.tokens.filter { token in
+                NSLocationInRange(token.offset, match.range) ||
                     NSLocationInRange(match.range.location,
-                        NSRange(location: $0.offset, length: $0.length))
+                        NSRange(location: token.offset, length: token.length))
             }
             for token in tokensInRange {
                 if NSIntersectionRange(NSRange(location: token.offset,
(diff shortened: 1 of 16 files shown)


segiddins:
if at all possible, it might be preferable to make this single statement closures instead of single-line

jpsim:
Lol, of course the rule would violate itself:

```
/Users/travis/build/realm/SwiftLint/Source/SwiftLintFramework/Rules/MultilineClosureArgumentRule.swift:14:32: error: Force Try Violation: Force tries should be avoided. (force_try)
/Users/travis/build/realm/SwiftLint/Source/SwiftLintFramework/Rules/MultilineClosureArgumentRule.swift:25:1: warning: Multi-line closure argument Violation: Multi-line closures should not use $0 (closure_argument)
/Users/travis/build/realm/SwiftLint/Source/SwiftLintFramework/Rules/MultilineClosureArgumentRule.swift:32:1: warning: Multi-line closure argument Violation: Multi-line closures should not use $0 (closure_argument)
/Users/travis/build/realm/SwiftLint/Source/SwiftLintFramework/Rules/MultilineClosureArgumentRule.swift:41:1: warning: Multi-line closure argument Violation: Multi-line closures should not use $0 (closure_argument)
/Users/travis/build/realm/SwiftLint/Source/SwiftLintFramework/Rules/MultilineClosureArgumentRule.swift:42:1: warning: Multi-line closure argument Violation: Multi-line closures should not use $0 (closure_argument)
```

keith:
Yea I can disable those if we want to move forward with this. And fix the description.

jpsim:
I think we have a difference of stylistic opinion regarding when it's appropriate to use `$0`.

I'd be much more inclined to say it's for single *statements* like @segiddins suggested rather than requiring the closure's braces to be on the same line.

I actually consider all of SwiftLint's current uses of the dollar sign shorthand to be appropriate in that sense.

Unfortunately, I have no idea how to detect statements.

jpsim:
By the way, it'd be trivial to update this rule to only apply in the appropriate context:

```diff
-    private static let regex = try! NSRegularExpression(pattern:
-        "(^[^\\{\\n]*\\$0|\\$0[^\\}\\n]*$)", options: .AnchorsMatchLines)
-
     public func validateFile(file: File) -> [StyleViolation] {
-        let range = NSRange(location: 0, length: file.contents.utf16.count)
-        let matches = MultilineClosureArgumentRule.regex.matchesInString(file.contents,
-            options: [], range: range)
-
-        return matches.map { match in
-            return StyleViolation(ruleDescription: self.dynamicType.description,
-                location: Location(file: file, offset: match.range.location))
+        let pattern = "(^[^\\{\\n]*(\\$0)|(\\$0)[^\\}\\n]*$)"
+        return file.matchPattern(pattern, withSyntaxKinds: [.Identifier]).map { match in
+            StyleViolation(ruleDescription: self.dynamicType.description,
+                           location: Location(file: file, offset: match.location))
         }
     }
```

jpsim:
@keith would you like to revive this now that we have opt-in rules?

keith:
Sure. I'll try and get to the feedback shortly.

jpsim:
FWIW I wouldn't enable it for this project.

scottrhoyt:
@keith do you have any updates here? Or should we close this PR out?

keith:
I'll reopen this when I have time to make the fixes!

jpsim:
I'll actually take a stab at reviving this if you don't mind @keith. I think it's a great rule to have, even on by default, but keeping the default number of lines a bit higher (4-5?).

keith:
Have at it! I think that a single line is a nice rule, but maybe that could be customizable?

jpsim:
> Have at it! I think that a single line is a nice rule, but maybe that could be customizable?

Yeah, if this goes in, it should certainly be customizable.

jpsim:
Hmm, on second thought, I don't think I can do this as easily as I had expected... There's no `SwiftDeclarationKind` for closures, so they're a bit hard to detect.

PR #223, opened 2015-11-20
Use `cp -Rf` instead of `cp -rf` in Makefile

norio-nomura:
`-r` option of `cp` copies symbolic links as normal files.
As the result of that, `SwiftLintFramework.framework` that distributed by Homebrew contains a lot of duplicated files inside that.
`-R` option copies symbolic links as symbolic links.

On my testing, this fix reduces the size of `swiftlint-0.4.0.el_capitan.bottle.tar.gz` from 17.2MB to 4.3MB.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ installables: clean bootstrap
 
 prefix_install: installables
 	mkdir -p "$(PREFIX)/Frameworks" "$(PREFIX)/bin"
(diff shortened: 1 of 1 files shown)


norio-nomura:
https://github.com/Carthage/Carthage/pull/932

jpsim:
Sweet, thanks!

norio-nomura:
:+1:

PR #225, opened 2015-11-21
add Valid Docs rule

jpsim:
I've been meaning to write a rule like this for a while, to prevent doc comments from getting out of sync with their declarations, or preventing typos.

This rule aims to validate that documentation comments correctly reflect the declaration.

This currently only checks if all function parameters are present in the docs as `- parameter ...:`, but this rule could be expanded to validate the following things:

- [x] if a function `throws`, that this be marked in the comment as `- Throws:`.
- [x] if a function returns something, that the comment either start with the word "Returns" or that `- returns:` be present.
- [x] the opposite of all the above in the absence of those things in the declaration.

/cc @keith

### DIFF
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -9,7 +9,7 @@
 import SourceKittenFramework
 import SwiftXPC
 
-private func regex(pattern: String) -> NSRegularExpression {
+internal func regex(pattern: String) -> NSRegularExpression {
     // all patterns used for regular expressions in SwiftLint are string literals which have
     // been confirmed to work, so it's ok to force-try here.
     // swiftlint:disable force_try
--- a/Source/SwiftLintFramework/Extensions/QueuedPrint.swift
+++ b/Source/SwiftLintFramework/Extensions/QueuedPrint.swift
@@ -19,14 +19,22 @@ private let outputQueue: dispatch_queue_t = {
     return queue
 }()
 
-/// A thread-safe version of Swift's standard print().
+/**
+ A thread-safe version of Swift's standard print().
+
+ - parameter object: Object to print.
+*/
 public func queuedPrint<T>(object: T) {
     dispatch_async(outputQueue) {
         print(object)
     }
 }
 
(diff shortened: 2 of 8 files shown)


ashfurrow:
👍

jpsim:
Thanks for the feedback @ashfurrow!

Thomvis:
Perhaps it would make sense to have two versions of this rule: one that only checks if the doc and declaration are in sync and one that additionally checks for missing documentation.

I'm thinking the second rule would be too noisy for projects where documentation is not always required. But there still is a lot of value in it for where documentation exists, hence rule 1. 

Also, is this rule smart enough to see that a function is defined in a protocol (where it is documented) and therefore does not need to have docs in all occurrences?

stephencelis:
When a function has more than one parameter, I tend to use this formatting:

``` swift
/// docs
///
/// - Parameters:
///   - first: this is void
///   - second: this is also void
public func a(first: Void, second: Void)
```

I find it's more human-readable.

jpsim:
Thanks for the feedback @Thomvis!

> Perhaps it would make sense to have two versions of this rule: one that only checks if the doc and declaration are in sync and one that additionally checks for missing documentation.

This rule (`valid_docs`) just checks existing documentation. I'm also working on another rule (`missing_docs`, #201) which verifies that all public declarations be documented.

This way, you'll be able to disable or parameterize them separately.

I agree that the missing docs rule might create too much noise in many projects, so I'm thinking of having that one off by default.

The missing docs rule doesn't currently count inherited documentation, which is one of the reasons it's still a work in progress.

jpsim:
@stephencelis I agree that your format is nice and legible. Does Xcode support that documentation formatting though?

stephencelis:
It sure does :)

jpsim:
Neat! Well I learned something today. We should support that in jazzy, so I created https://github.com/realm/jazzy/issues/388 to track that. It should be easy to adapt this rule for that syntax.

jpsim:
> It should be easy to adapt this rule for that syntax.

Well, not so easy after all :grimacing:. This rule now just skips validating parameters for batched parameters in comments. Still, I think this is ready to be merged.

Issue #214, opened 2015-11-18
Add verifyRule(...) option to check string doesn't trigger a violation


jpsim:
Most rules shouldn't trigger when included in a string literal. Similar to how `verifyRule(...)` checks that the triggering examples don't trigger when included in a comment, we should verify the same thing for string literals.

PR #232, opened 2015-11-23
verify that triggering examples in a string literal don't trigger violations

jpsim:
Addresses #214.

### DIFF
--- a/Source/SwiftLintFrameworkTests/TestHelpers.swift
+++ b/Source/SwiftLintFrameworkTests/TestHelpers.swift
@@ -23,32 +23,51 @@ private func violations(string: String, _ description: RuleDescription) -> [Styl
     return violations(string, config: Configuration(disabledRules: disabledRules)!)
 }
 
+extension String {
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFrameworkTests/TestHelpers.swift
```diff
@@ -23,32 +23,51 @@ private func violations(string: String, _ description: RuleDescription) -> [Styl
...
+                violations($0, ruleDescription)
+            }).isEmpty
+        )
+
+        // Triggering examples violate
         XCTAssertEqual(
-            ruleDescription.nonTriggeringExamples.flatMap({violations($0, ruleDescription)}),
-            []
+            ruleDescription.triggeringExamples.flatMap({ violations($0, ruleDescription) }).count,
+            ruleDescription.triggeringExamples.count
```
segiddins: this change will make it harder to debug

jpsim: unlikely given #226

PR #240, opened 2015-11-28
Variable name violation should be ignored if the entire name is in capitals

### LINKED ISSUE #161: Variable name violation should be ignored if the entire name is in capitals
A lot of times there are variable names where the name is an acronym, even in apple code such as RSSI from CoreBluetooth which stands for Relative Signal Strength Indication. of course you could call it relativeSignalStrengthIndication but it seems impractically long. I think people agree that when acronyms are used, the entire name is in capitals. 
I think writing the entire variable name in capitals is unlikely something done by mistake and usually has a meaning.

marcelofabri:
Fixes #161.

### DIFF
--- a/Source/SwiftLintFramework/Rules/VariableNameRule.swift
+++ b/Source/SwiftLintFramework/Rules/VariableNameRule.swift
@@ -16,13 +16,15 @@ public struct VariableNameRule: ASTRule {
         identifier: "variable_name",
         name: "Variable Name",
         description: "Variable name should only contain alphanumeric characters and " +
-          "start with a a lowercase character. In an exception to the above, variable " +
-          "names may start with a capital letter when they are declared static and immutable.",
+          "start with a a lowercase character or should only contain capital letters. " +
+          "In an exception to the above, variable names may start with a capital letter " +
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Rules/VariableNameRule.swift
```diff
@@ -16,13 +16,15 @@ public struct VariableNameRule: ASTRule {
         identifier: "variable_name",
         name: "Variable Name",
         description: "Variable name should only contain alphanumeric characters and " +
-          "start with a a lowercase character. In an exception to the above, variable " +
-          "names may start with a capital letter when they are declared static and immutable.",
+          "start with a a lowercase character or should only contain capital letters. " +
```
jpsim: "a a". I know it was already like that but as you're already modifying this line, I'd appreciate if you could fix this typo. Thanks!

jpsim:
:+1: thanks for the fix!

Source/SwiftLintFramework/Rules/VariableNameRule.swift
marcelofabri: :+1:

PR #241, opened 2015-11-28
Automatically correct certain violations

jpsim:
This is a first stab at #5. ~~Not super helpful at the moment because there's no reporting of what was corrected, which we'll want to add before merging this in~~ **edit: done**.

Since this sets the direction for future autocorrect behavior, feedback is greatly appreciated!

Turns out this is kind of tricky to do in a robust way. Specifically: modifying files on disk can cause previous violation offsets to become invalid, or new violations to be created, or previous violations to vanish. For this reason, I've opted for `--autocorrect` to not lint in the normal sense, not reporting any violations.

Remaining:

- [x] Report corrections
- [x] Don't perform corrections if a rule is disabled globally via `disabled_rules`
- [x] Don't perform corrections if a rule is disabled locally via `// swiftlint:disable`
- [x] Decide if a dedicated `autocorrect` command is preferable to an `--autocorrect` flag on the `lint` command (requires moving a lot more of the logic currently in `swiftlint.LintCommand` into `SwiftLintFramework.Linter`, which is probably a good thing to do anyway)


jpsim:
@keith it'd be great to get your feedback on this oine.

Source/SwiftLintFramework/Models/Linter.swift
```diff
@@ -10,6 +10,14 @@ import Foundation
 import SwiftXPC
 import SourceKittenFramework
 
+private func correctableRules() -> [CorrectableRule] {
```
keith: Maybe instead of keeping a separate list we have to maintain we could `flatMap` enabled rules from the configuration `as? CorrectableRule` ?

keith:
I'm totally on board with this approach!

keith:
Re the last check mark, I think I'd prefer a separate autocorrect command.

Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
```diff
@@ -60,4 +57,10 @@ extension String {
         }
         return nil
     }
+
+    internal func nsrangeToIndexRange(nsrange: NSRange) -> Range<Index> {
```
keith: I really want this to be in the standard lib :disappointed:

segiddins: file a :satellite: !

Source/SwiftLintFramework/Rules/TrailingNewlineRule.swift
```diff
@@ -8,7 +8,19 @@
 
 import SourceKittenFramework
 
-public struct TrailingNewlineRule: Rule {
+extension String {
+    public func trailingNewlineCount() -> Int? {
```
keith: This could probably be `private` unless you can think of another place we might use it.

Source/SwiftLintFramework/Rules/TrailingNewlineRule.swift
```diff
@@ -19,20 +31,37 @@ public struct TrailingNewlineRule: Rule {
...
     }
+
+    public func correctFile(file: File) -> [Correction] {
+        guard let count = file.contents.trailingNewlineCount() where count != 1 else {
+            return []
+        }
+        let region = file.regions().filter {
+            $0.contains(Location(file: file.path, line: max(file.lines.count, 1)))
+        }.first
+        if region?.isRuleDisabled(self) == true {
```
keith: We could probably check this region thing before calling `trailingNewlineCount`? Just to save some cycles. Also maybe we should be verifying that we should actually fix this at a higher level? Not sure how that would work though.

Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
keith: :trollface:

Source/SwiftLintFrameworkTests/TestHelpers.swift
```diff
@@ -23,6 +23,31 @@ private func violations(string: String, _ description: RuleDescription) -> [Styl
...
+        return
+    }
+    let corrections = Linter(file: file).correct()
+    XCTAssertEqual(corrections.count, 1)
+    XCTAssertEqual(file.contents, expected)
+    do {
+        let corrected = try NSString(contentsOfFile: path, encoding: NSUTF8StringEncoding) as String
+        XCTAssertEqual(corrected, expected)
+    } catch {
+        XCTFail("coun't read file at path '\(path)': \(error)")
```
keith: spleling

Source/SwiftLintFramework/Rules/TrailingWhitespaceRule.swift
```diff
@@ -8,13 +8,14 @@
...
 import SourceKittenFramework
 
-public struct TrailingWhitespaceRule: Rule {
+public struct TrailingWhitespaceRule: CorrectableRule {
     public static let description = RuleDescription(
         identifier: "trailing_whitespace",
         name: "Trailing Whitespace",
         description: "Lines should not have trailing whitespace.",
         nonTriggeringExamples: [ "//\n" ],
-        triggeringExamples: [ "// \n" ]
```
keith: wouldn't have this diff if you added trailing `,`s on the last item :trollface:

Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
zwaldowski: I see this method being used by the regex matchers. Since `NSString` is based around UTF-16 code units and not grapheme clusters, wouldn't this give wrong results for certain kinds of strings?

jpsim: You're right! :scream: I just fixed this. Nice catch!

Source/SwiftLintFramework/Rules/TrailingNewlineRule.swift
jpsim: `trailingNewlineCount` is more efficient than computing the regions in a file. Not sure what you mean by fixing this at a higher level though.

Source/SwiftLintFramework/Models/Linter.swift
jpsim: good idea, I think that should work.

Source/SwiftLintFramework/Rules/TrailingNewlineRule.swift
keith: All rules happen to be checking if they're disabled for regions. I think ideally `correctFile` wouldn't even be called if that were the case. But that's tough because you'd have to codify what part of the file you aimed to fix.

Source/SwiftLintFramework/Models/Linter.swift
jpsim: whoa, much nicer.

Source/SwiftLintFramework/Rules/TrailingNewlineRule.swift
jpsim: we could add file region computation to the file cache, but that's out of scope for this PR IMO.

jpsim:
@keith when you get the chance, I'd love your input on #253 and this one, which has now been rebased on top of that visitor PR. Now autocorrect has a dedicated command. Damn that was a lot of work to do right.

keith:
:shipit:

jpsim:
Thanks for reviewing, @keith!

Gerst20051:
👍

PR #242, opened 2015-11-29
Prefer Swift constructors to legacy ones

marcelofabri:
What do you think, @jpsim?

### DIFF
--- a/Source/SwiftLintFramework/Rules/LegacyConstructorRule.swift
+++ b/Source/SwiftLintFramework/Rules/LegacyConstructorRule.swift
@@ -7,7 +7,6 @@
 //
 
 import SourceKittenFramework
-import SwiftXPC
 
 public struct LegacyConstructorRule: Rule {
     public init() {}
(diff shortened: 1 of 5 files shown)


Source/SwiftLintFramework/Rules/LegacyConstructorRule.swift
```diff
@@ -0,0 +1,50 @@
...
+        triggeringExamples: [
+            "CGPointMake(10, 10)",
+            "CGSizeMake(10, 10)",
+            "CGRectMake(0, 0, 10, 10)",
+            "CGVectorMake(10, 10)",
+            "NSMakeRange(10, 1)",
+        ]
+    )
+
+    public func validateFile(file: File) -> [StyleViolation] {
```
jpsim: I have a few suggestions:

1. It's more performant to only scan the file for matches once, which can be done by merging all the regular expressions into one.
2. Instead of excluding comments and strings, we should check if the match is a `SyntaxKind.Identifier` which is more semantically correct, even though it probably won't make a difference in practice.

So this function could be refactored to this:

```swift
    let constructors = ["CGRectMake", "CGPointMake", "CGSizeMake", "CGVectorMake",
        "NSMakeRange"]

    let pattern = "\\b(" + constructors.joinWithSeparator("|") + ")\\b"

    return file.matchPattern(pattern, withSyntaxKinds: [.Identifier]).map {
            location: Location(file: file, offset: $0.location))
    }
}
```

jpsim:
This is great @marcelofabri! I only had a few minor suggestions regarding the implementation.

Could you also add an entry for this to the changelog?

Thanks for doing this.

Source/SwiftLintFramework/Rules/LegacyConstructorRule.swift
```diff
@@ -0,0 +1,50 @@
+//
+//  LegacyConstructorRule.swift
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 29/11/15.
+//  Copyright © 2015 Realm. All rights reserved.
+//
+
+import SourceKittenFramework
+import SwiftXPC
```
jpsim: We don't use SwiftXPC in this file, so this import statement can be removed.

marcelofabri:
:tada:

jpsim:
:+1: awesome work!

PR #245, opened 2015-12-03
add Realm Object rule

jpsim:
I wrote this quickly so it's pretty messy, but it shouldn't have any false positives and might correct some otherwise misconfigured Realm models. /cc @bdash

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#161](https://github.com/realm/SwiftLint/issues/161)
 
+* Add Realm Object rule to validate Realm Swift models.  
+  [JP Simard](https://github.com/jpsim)
+  [realm/realm-cocoa#2929](https://github.com/realm/realm-cocoa/issues/2929)
+
 ##### Bug Fixes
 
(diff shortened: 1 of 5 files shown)


bdash:
This seems like it would generate false-positives in many codebases unless it skipped ignored properties.

jpsim:
There are several limitations with the current implementation that make me think this rule should be disabled by default:

**1) Doesn't validate Object or List properties**

Object properties would be challenging to validate because we don't know what types are Realm models. If we had valid compiler arguments for the file being linted, we'd be able to resolve those types and determine if they subclass RealmSwift.Object, but as things are now, we can't tell.

List properties could feasibly be validated with moderately extra effort though.

**2) Doesn't work with inferred types**

Because SourceKit doesn't apply type inference (that's a significant chunk of the compilation process), the current implementation just validates explicitly defined types.

**3) Doesn't consider ignored properties**

If you have a `let name: String` as an ignored property, this rule will treat that as incorrect, yielding a false positive.

**4) Doesn't know if inheriting from RealmSwift.Object or some other \*.Object type**

We could check for `import RealmSwift` within the same file which would be required for that code to compile, but I'm not sure how I feel about that.

jpsim:
This rule is much too limited in scope to be useful I think. Closing.

PR #250, opened 2015-12-05
Use thoughtbot/Curry to automatically curry the LintOptions initializer

jpsim:
Commandant's evaluation needs a curried initializer to use that operator syntax and Curry does this nicely for us, so we don't have to do it manually. Thanks @gfontenot!


Source/swiftlint/LintCommand.swift
```diff
@@ -167,14 +168,9 @@ struct LintOptions: OptionsType {
...
 
-    static func create(path: String)(useSTDIN: Bool)(configurationFile: String)(strict: Bool)
-        (useScriptInputFiles: Bool) -> LintOptions {
-        return LintOptions(path: path, useSTDIN: useSTDIN, configurationFile: configurationFile,
-            strict: strict, useScriptInputFiles: useScriptInputFiles)
-    }
-
     static func evaluate(mode: CommandMode) -> Result<LintOptions, CommandantError<()>> {
-        return create
+        let curriedInit = curry(self.init)
```
jpsim: has to be split off into its own function to avoid crippling the compiler due to complex expression evaluation.

PR #251, opened 2015-12-06
move commands into Source/swiftlint/Commands

jpsim:


jpsim:
this make adding an `Extensions` directory and an additional `autocorrect` command more manageable.

PR #253, opened 2015-12-06
Consolidate lint files

jpsim:
This is setting up for a dedicated autocorrect command, and is a lot nicer to work with IMO

### DIFF
--- a/Source/swiftlint/Extensions/Configuration+CommandLine.swift
+++ b/Source/swiftlint/Extensions/Configuration+CommandLine.swift
@@ -12,9 +12,10 @@ import Result
 import SourceKittenFramework
 import SwiftLintFramework
 
+private let fileManager = NSFileManager.defaultManager()
 private let inputFileKey = "SCRIPT_INPUT_FILE_COUNT"
 
-func scriptInputFiles() -> Result<[String], CommandantError<()>> {
+private func scriptInputFiles() -> Result<[String], CommandantError<()>> {
     func getEnvironmentVariable(variable: String) -> Result<String, CommandantError<()>> {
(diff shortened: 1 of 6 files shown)


Source/swiftlint/Extensions/Configuration+CommandLine.swift
```diff
@@ -0,0 +1,94 @@
...
+    func getFiles(path: String, action: String, useSTDIN: Bool) ->
+                  Result<[File], CommandantError<()>> {
+        if useSTDIN {
+            let standardInput = NSFileHandle.fileHandleWithStandardInput()
+            let stdinData = standardInput.readDataToEndOfFile()
+            let stdinNSString = NSString(data: stdinData, encoding: NSUTF8StringEncoding)
+            if let stdinString = stdinNSString as? String {
+                return .Success([File(contents: stdinString)])
+            }
+            return .Failure(.UsageError(description: "stdin isn't a string"))
```
segiddins: might want to mention UTF-8 in this error

jpsim:
Ok, I ended up making a lot more changes in here because it would have been too much work to split it up. Specifically a `Configuration.visitLintableFiles(...)` visitor function which will more easily allow performing other operations on the same lintable files (e.g. autocorrecting).

Source/swiftlint/Extensions/Configuration+CommandLine.swift
jpsim: thanks for the tip, done

jpsim:
Since this is just an internal refactor, I'll merge this one now. @keith if you have feedback, I'll do my best to incorporate it afterwards.

PR #260, opened 2015-12-13
Exclude files from included folders.

michallaskowski:
Allows to exclude specific files and folders via `excluded` from `included` paths.

Reason behind this:
In one of my projects I use [sbconstants](https://github.com/paulsamuels/SBConstants), which produces a file that does not pass lint. I would like to store the generated file in the same directory with my Storyboards, for example `Source/Interface`, and to be able to exclude the `Source/Interface` directory while linting other files in `Source`.


jpsim:
This is great, thanks!

PR #272, opened 2015-12-17
Updated README with a list of rules coming from `swiftlint rules`

jschmid:


keith:
I'm not sure we want to keep these 2 data sources in sync. This seems particularly error prone to me. Maybe we remove this and just have advice for how users can see the real list locally?

jschmid:
I agree with you. This was just a quick fix. 

See my updated commit.

keith:
Looks good to me! Mind squashing these commits? After that we should be good to merge!

jschmid:
There you go !

keith:
Thanks!

jpsim:
The sample YAML in the README was never meant to be a canonical source of all rules, but it kind of just became that over time. This is much better, thanks!

PR #274, opened 2015-12-17
Improve autocorrect for OpeningBraceRule.

inamiy:
This is an improvement for `OpeningBraceRule` to conform to `CorrectableRule` protocol.
Many of Allman indent-style will be converted to 1TBS :wink:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,18 @@
+## Master
+
+##### Breaking
+
+* None.
+
+##### Enhancements
+
+* Improve autocorrect for OpeningBraceRule.  
+  [Yasuhiro Inami](https://github.com/inamiy)
+
+##### Bug Fixes
+
+* None.
+
 ## 0.5.2: Snuggle™
 
 ##### Breaking
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -126,3 +126,17 @@ extension File {
         lines = contents.lines()
     }
 }
+
+extension File {
+    internal func ruleEnabledViolatingRanges(violatingRanges: [NSRange], forRule rule: Rule)
+        -> [NSRange] {
+        let fileRegions = self.regions()
+        let violatingRanges = violatingRanges.filter { range in
+            let region = fileRegions.filter {
+                $0.contains(Location(file: self, offset: range.location))
+            }.first
+            return region?.isRuleEnabled(rule) ?? true
+        }
+        return violatingRanges
+    }
+}
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -125,12 +125,10 @@ extension File {
         contents = string
         lines = contents.lines()
     }
-}
 
-extension File {
     internal func ruleEnabledViolatingRanges(violatingRanges: [NSRange], forRule rule: Rule)
         -> [NSRange] {
-        let fileRegions = self.regions()
+        let fileRegions = regions()
         let violatingRanges = violatingRanges.filter { range in
             let region = fileRegions.filter {
                 $0.contains(Location(file: self, offset: range.location))
--- a/Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
+++ b/Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
@@ -9,7 +9,7 @@
 import Foundation
 import SourceKittenFramework
 
-let whitespaceAndNewlineCharacterSet = NSCharacterSet.whitespaceAndNewlineCharacterSet()
+private let whitespaceAndNewlineCharacterSet = NSCharacterSet.whitespaceAndNewlineCharacterSet()
 
 extension File {
     private func violatingOpeningBraceRanges() -> [NSRange] {
--- a/Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
+++ b/Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
@@ -9,15 +9,12 @@
 import Foundation
 import SourceKittenFramework
 
-private let violatingPattern = "((?:[^( ]|[\\t\\n\\f\\r (] )\\{)"
-private let correctString = " {"
-
-private let whitespaceAndNewlineCharacterSet = NSCharacterSet.whitespaceAndNewlineCharacterSet()
+let whitespaceAndNewlineCharacterSet = NSCharacterSet.whitespaceAndNewlineCharacterSet()
 
 extension File {
     private func violatingOpeningBraceRanges() -> [NSRange] {
         return matchPattern(
-            violatingPattern,
+            "((?:[^( ]|[\\s(][\\s]+)\\{)",
             excludingSyntaxKinds: SyntaxKind.commentAndStringKinds()
         )
     }
@@ -39,6 +36,15 @@ public struct OpeningBraceRule: CorrectableRule {
             "func abc()\n\t{ }",
             "[].map(){ $0 }",
             "[].map( { } )"
+        ],
+        corrections: [
+            "struct Rule{}\n": "struct Rule {}\n",
+            "struct Rule\n{\n}\n": "struct Rule {\n}\n",
+            "struct Rule\n\n\t{\n}\n": "struct Rule {\n}\n",
+            "struct Parent {\n\tstruct Child\n\t{\n\t\tlet foo: Int\n\t}\n}\n":
+                "struct Parent {\n\tstruct Child {\n\t\tlet foo: Int\n\t}\n}\n",
+            "[].map(){ $0 }\n": "[].map() { $0 }\n",
+            "[].map( { } )\n": "[].map({ } )\n",
         ]
     )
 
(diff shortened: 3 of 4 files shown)


jpsim:
Thanks for this! It's exciting to see community-provided corrections, this is the first one!

Could you please write tests by adding entries to the `corrections` dictionary of the `RuleDescription`?

An entry in the changelog would also be appreciated, as described in [CONTRIBUTING.md](https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md).

Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
```diff
@@ -9,7 +9,21 @@
 import Foundation
 import SourceKittenFramework
 
-public struct OpeningBraceRule: Rule {
+private let violatingPattern = "((?:[^( ]|[\\t\\n\\f\\r (] )\\{)"
```
jpsim: since this is only used in `File.violatingOpeningBraceRanges()`, it could just be inlined into that method.

Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
```diff
@@ -9,7 +9,21 @@
...
+private let correctString = " {"
```
jpsim: since this is only used in `OpeningBraceRule.writeToFile()`, it could just be inlined into that method.

Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
```diff
@@ -29,12 +43,68 @@ public struct OpeningBraceRule: Rule {
...
-        return file.matchPattern(pattern, excludingSyntaxKinds: excludingKinds).map {
+        return file.violatingOpeningBraceRanges().map {
             StyleViolation(ruleDescription: self.dynamicType.description,
                 location: Location(file: file, offset: $0.location))
         }
     }
+
+    public func correctFile(file: File) -> [Correction] {
+        let fileRegions = file.regions()
+        let violatingRanges = file.violatingOpeningBraceRanges().filter { range in
```
jpsim: We're now doing this region check in two places: here and in `TrailingSemicolonRule`. We should add a `File.filterEnabledRange(ranges: [NSRange], forRule: Rule)` method.

Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
```diff
@@ -29,12 +43,68 @@ public struct OpeningBraceRule: Rule {
...
+        let correctStringCount = correctString.characters.count
+        var correctedContents = file.contents
+        var adjustedRanges = [NSRange]()
+        var previousLengthDelta = 0
+
+        for violatingRange in violatingRanges {
+            guard let range = file.contents.nsrangeToIndexRange(violatingRange) else {
+                continue
+            }
+            let capturedString = file.contents[range]
```
jpsim: I think this could be simplified quite a bit:

```swift
let capturedString = file.contents[range]

// if "struct Command{" is violated with violating string = "d{",
// adjust range to only replace "{"
let shouldAdjust = capturedString.characters.count == 2 &&
    capturedString.rangeOfCharacterFromSet(whitespaceAndNewlineCharacterSet) == nil
let adjustment = Int(shouldAdjust)

previousLengthDelta = violatingRange.length - correctStringCount - adjustment
let adjustedRange = NSRange(
    location: violatingRange.location - previousLengthDelta + adjustment,
    length: correctStringCount - adjustment
)
adjustedRanges += [adjustedRange]
```

inamiy:
Thanks for review!
I improved codes and added autocorrect tests, so please take a try :pray:

BTW, there was a bug where SwiftLint failed to catch `struct Rule\n\n\n{ }`, so I tweaked original regex a little.

Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
inamiy: I added `File.ruleEnabledViolatingRanges()` in 4a8e997.
Method name is probably better this way.

Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
inamiy: Done in f5fa186 (fixing regex too).

Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
inamiy: Done in f5fa186.

Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
inamiy: Done in f5fa186.

CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+## 0.5.2:
```
jpsim: This should be `## Master`. I've pushed an empty top-level section to our `master` branch, so you should be able to rebase it over that.

CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
...
+
+##### Breaking
+
+* None.
+
+##### Enhancements
+
+* Improve autocorrect for OpeningBraceRule.
```
jpsim: Please add 2 trailing spaces on the last line describing the change, as outlined in [CONTRIBUTING.md](https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md).

CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
...
+  [Yasuhiro Inami](https://github.com/inamiy)
+  [#274](https://github.com/realm/SwiftLint/issues/274)
```
jpsim: This should be the issue that's being resolved, not the PR number, as outlined in [CONTRIBUTING.md](https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md).

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -126,3 +126,17 @@ extension File {
         lines = contents.lines()
     }
 }
+
+extension File {
+    internal func ruleEnabledViolatingRanges(violatingRanges: [NSRange], forRule rule: Rule)
```
jpsim: This function looks good, but why is it in a separate extension? Please just move to the main extension.

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -126,3 +126,17 @@ extension File {
...
+        -> [NSRange] {
+        let fileRegions = self.regions()
```
jpsim: Please omit `self` unless required, as outlined in GitHub's Swift style guide: https://github.com/github/swift-style-guide#only-explicitly-refer-to-self-when-required

Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
```diff
@@ -25,16 +36,89 @@ public struct OpeningBraceRule: Rule {
...
             "[].map( { } )"
+        ],
+        corrections: [
+            "struct Rule{}\n": "struct Rule {}\n",
+            "struct Rule\n{\n}\n": "struct Rule {\n}\n",
+            "struct Rule\n\n\t{\n}\n": "struct Rule {\n}\n",
+            "struct Parent {\n\tstruct Child\n\t{\n\t\tlet foo: Int\n\t}\n}\n":
+                "struct Parent {\n\tstruct Child {\n\t\tlet foo: Int\n\t}\n}\n",
+            "[].map(){ $0 }\n": "[].map() { $0 }\n",
+            "[].map( { } )\n": "[].map({ } )\n",
```
jpsim: If I was a user of SwiftLint, I'm not sure I'd want this correction, since it leaves the brackets asymmetrical.

Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
```diff
@@ -25,16 +36,89 @@ public struct OpeningBraceRule: Rule {
...
+
+    private func writeToFile(file: File, violatingRanges: [NSRange]) -> [Correction] {
+        var correctedContents = file.contents
+        var adjustedOffsets = [Int]()
+
+        for violatingRange in violatingRanges.reverse() {
+            let (contents, adjustedRange) =
+            correctContents(correctedContents, violatingRange: violatingRange)
+
+            correctedContents = contents
```
jpsim: why have the `contents` variable at all? It seems like the line above could be `let (correctedContents, adjustedRange) = ...` to avoid this intermediate variable.

Source/SwiftLintFramework/Rules/TrailingSemicolonRule.swift
```diff
@@ -38,13 +38,10 @@ public struct TrailingSemicolonRule: CorrectableRule {
...
 
     public func correctFile(file: File) -> [Correction] {
-        let fileRegions = file.regions()
-        let violatingRanges = file.violatingTrailingSemicolonRanges().filter { range in
-            let region = fileRegions.filter {
-                $0.contains(Location(file: file, offset: range.location))
-            }.first
-            return region?.isRuleEnabled(self) ?? true
-        }
+        let violatingRanges = file.ruleEnabledViolatingRanges(
```
jpsim: :+1:

jpsim:
This is looking really good @inamiy! But I do have quite a few comments :grimacing: 

> BTW, there was a bug where SwiftLint failed to catch struct Rule\n\n\n{ }, so I tweaked original regex a little.

:+1: :100:

Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
inamiy: `let (correctedContents, adjustedRange) = ...` didn't work because it is using `let` thus is different from previously declared `correctedContents`.
Also, `(correctedContents, let adjustedRange) = ...` didn't work too, so I decided to use temporary value here.

Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
inamiy: Should we *not* correct this example? 
(It so, `swiftlint autocorrect` will not improve this warning.)

CHANGELOG.md
inamiy: Done in 1ba7b1f.

CHANGELOG.md
inamiy: Done in 1ba7b1f.

CHANGELOG.md
inamiy: Removed PR line in 1ba7b1f.

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
inamiy: Done in ff3dd5b.

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
inamiy: Done in ff3dd5b.

Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
inamiy: Maybe I can add example `"[].map( { })\n": "[].map({ })\n"` too as a contrast.
In this case, we rather *want* to autocorrect :grin:.

I personally think we should autocorrect for `"[].map( { } )\n": "[].map({ } )\n"` as well,
and provide companion `TrailingBraceRule.swift` later.

Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
jpsim: Ok!

Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
jpsim: Ok, a TrailingBraceRule could work.

inamiy:
@jpsim Thanks for merge! :+1:

jpsim:
Thank you for doing this, and addressing my feedback!

Source/SwiftLintFramework/Rules/OpeningBraceRule.swift
inamiy: Improved this issue in https://github.com/realm/SwiftLint/pull/280!

PR #276, opened 2015-12-18
Add Function Naming Case Rule

squarefrog:
This commit adds a new rule for ensuring method names start with a lower case letter.

I'm unsure where new contributions should be added to the `CHANGELOG.md` as there is no `Master` heading anymore.

I hope this is useful.

### DIFF
--- a/Source/SwiftLintFramework/Rules/FunctionNamingCaseRule.swift
+++ b/Source/SwiftLintFramework/Rules/FunctionNamingCaseRule.swift
@@ -1,0 +1,33 @@
+//
+//  FunctionNamingCaseRule.swift
+//  SwiftLint
+//
+//  Created by Paul Williamson on 18/12/2015.
+//  Copyright © 2015 Realm. All rights reserved.
+//
+
+import SourceKittenFramework
+
+public struct FunctionNamingCaseRule: Rule {
+    public static let description = RuleDescription(
+        identifier: "function_naming_case",
+        name: "Function Naming Case",
+        description: "Function names should begin with a lower case letter.",
+        nonTriggeringExamples: [
+            "func doSomething() { }"
+        ],
+        triggeringExamples: [
+            "func DoSomething() { }"
+        ]
+    )
+
+    public func validateFile(file: File) -> [StyleViolation] {
+        let pattern = "func\\s+[A-Z]"
+        let excludingKinds = SyntaxKind.commentAndStringKinds()
+
+        return file.matchPattern(pattern, excludingSyntaxKinds: excludingKinds).map {
+            StyleViolation(ruleDescription: self.dynamicType.description,
+                location: Location(file: file, offset: $0.location))
+        }
+    }
+}
--- a/Source/SwiftLintFramework/Models/Configuration.swift
+++ b/Source/SwiftLintFramework/Models/Configuration.swift
@@ -130,6 +130,7 @@ public struct Configuration {
             ControlStatementRule(),
             ForceCastRule(),
             ForceTryRule(),
+            FunctionNamingCaseRule(),
             LeadingWhitespaceRule(),
             LegacyConstructorRule(),
             NestingRule(),
--- a/Source/SwiftLintFrameworkTests/StringRuleTests.swift
+++ b/Source/SwiftLintFrameworkTests/StringRuleTests.swift
@@ -105,4 +105,8 @@ class StringRuleTests: XCTestCase {
     func testLegacyConstructor() {
         verifyRule(LegacyConstructorRule.description)
     }
+
+    func testFunctionNamingCase() {
+        verifyRule(FunctionNamingCaseRule.description)
+    }
 }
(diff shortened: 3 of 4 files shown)


Source/SwiftLintFramework/Rules/FunctionNamingCaseRule.swift
```diff
@@ -0,0 +1,33 @@
...
+            "func doSomething() { }"
+        ],
+        triggeringExamples: [
+            "func DoSomething() { }"
+        ]
+    )
+
+    public func validateFile(file: File) -> [StyleViolation] {
+        let pattern = "func\\s+[A-Z]"
+        let excludingKinds = SyntaxKind.commentAndStringKinds()
```
jpsim: instead of excluding kinds, this should look for identifiers

jpsim: See [FunctionBodyLengthRule](https://github.com/realm/SwiftLint/blob/0.5.1/Source/SwiftLintFramework/Rules/FunctionBodyLengthRule.swift#L35-L50) for an example that does this.

squarefrog: :+1: OK I will make that change. Have a great weekend!

keith:
This is probably going to have some of the same issues as some of the other naming rules. Where `URLWithString` would fail here. We may want to come up with a solution to this.

jpsim:
> This is probably going to have some of the same issues as some of the other naming rules. Where URLWithString would fail here. We may want to come up with a solution to this.

Honestly my personal style preference in those cases is to not make exceptions, so I'd call such a variable/function `urlWithString`. Or if I had a type representing iPhones, I'd call that `IPhone`. Maybe I'm a weirdo for preferring this though.

squarefrog:
Would you like me to hold off trying to fix this until a solution for `URLWithString` is found?

keith:
I don't think we should hold off unless there's an obvious solution to it. Possibly verifying that more than just the first character is uppercase. That would probably hide some invalid cases though. Users can always disable this if they don't follow this style!

squarefrog:
My thought then would be match uppercase followed by lowercase. 

`MyMethod()` matches where `URLWithString` does not. Obviously this would be trivial to work around with `MYMethod`, but if you do that you're perhaps beyond help :-1:

keith:
That sounds good to me. Thoughts @jpsim ?

jpsim:
Sure, starting with either lowercase or 2+ uppercase sounds fine to me.

wesbillman:
The same rule might be good for variable names as well..

For example:
```swift
init?(JSONDictionary: JSONObject) {
```

jpsim:
@squarefrog will you be updating this based on feedback provided?

squarefrog:
To be honest I'm struggling to get the first case passing using identifiers. I did some work that seemed promising then I haven't touched a computer for two weeks :)

jpsim:
Sorry I never gave you more guidance on this @squarefrog, that's on me!

Even so, I'll be closing this PR since it hasn't seen any traction in a few months.

For academic purposes, here's how you could have adapted this rule to only apply to functions:

```diff
diff --git a/Source/SwiftLintFramework/Rules/FunctionNamingCaseRule.swift b/Source/SwiftLintFramework/Rules/FunctionNamingCaseRule.swift
index af0608a..5e441b0 100644
--- a/Source/SwiftLintFramework/Rules/FunctionNamingCaseRule.swift
+++ b/Source/SwiftLintFramework/Rules/FunctionNamingCaseRule.swift
@@ -7,8 +7,9 @@
 //
 
 import SourceKittenFramework
+import SwiftXPC
 
-public struct FunctionNamingCaseRule: Rule {
+public struct FunctionNamingCaseRule: ASTRule {
     public static let description = RuleDescription(
         identifier: "function_naming_case",
         name: "Function Naming Case",
@@ -21,13 +22,35 @@ public struct FunctionNamingCaseRule: Rule {
         ]
     )
 
-    public func validateFile(file: File) -> [StyleViolation] {
-        let pattern = "func\\s+[A-Z]"
-        let excludingKinds = SyntaxKind.commentAndStringKinds()
-
-        return file.matchPattern(pattern, excludingSyntaxKinds: excludingKinds).map {
-            StyleViolation(ruleDescription: self.dynamicType.description,
-                location: Location(file: file, offset: $0.location))
+    public func validateFile(file: File, kind: SwiftDeclarationKind,
+                             dictionary: XPCDictionary) -> [StyleViolation] {
+        let functionKinds: [SwiftDeclarationKind] = [
+            .FunctionAccessorAddress,
+            .FunctionAccessorDidset,
+            .FunctionAccessorGetter,
+            .FunctionAccessorMutableaddress,
+            .FunctionAccessorSetter,
+            .FunctionAccessorWillset,
+            .FunctionConstructor,
+            .FunctionDestructor,
+            .FunctionFree,
+            .FunctionMethodClass,
+            .FunctionMethodInstance,
+            .FunctionMethodStatic,
+            .FunctionOperator,
+            .FunctionSubscript
+        ]
+        if !functionKinds.contains(kind) {
+            return []
         }
+        guard let name = dictionary["key.name"] as? String,
+            offset = (dictionary["key.offset"] as? Int64).flatMap({ Int($0) }) where
+            name.substringToIndex(name.startIndex.successor()).isUppercase() else {
+                return []
+        }
+        return [
+            StyleViolation(ruleDescription: self.dynamicType.description,
+                location: Location(file: file, byteOffset: offset))
+        ]
     }
 }
```

Feel free to reopen a new PR if you ever want to get back to this.

PR #280, opened 2015-12-19
Add ClosingBraceRule.

inamiy:
This is an improvement for #274 to support closing-brace rule.

`[].map({ } )` will be autocorrected to `[].map({ })`.


jpsim:
:+1: excellent work

PR #283, opened 2015-12-21
Add autocorrect for ColonRule.

brianpartridge:
This change adds autocorrect support for ColonRule so now all your colons will be properly positioned.  The changes include:
- Improved regular expressions (support for more type identifiers, lazy evaluation of type identifiers).
- More tests for new type identifier support.
- Improved support for rejecting strings and comments.

The only thing I dislike about this is that we need to take 2 passes over the file (once for each regex). I'm not sure if there is a way to avoid this without adding false positives to the validation step though.

:rainbow:

### DIFF
--- a/Source/SwiftLintFramework/Rules/ColonRule.swift
+++ b/Source/SwiftLintFramework/Rules/ColonRule.swift
@@ -77,7 +77,7 @@ public struct ColonRule: CorrectableRule {
     )
 
     public func validateFile(file: File) -> [StyleViolation] {
-        let pattern = (patterns() as NSArray).componentsJoinedByString("|")
+        let pattern = patterns().joinWithSeparator("|")
 
         return validMatchesInFile(file, withPattern: pattern).flatMap { range in
             return StyleViolation(ruleDescription: self.dynamicType.description,
--- a/Source/SwiftLintFramework/Rules/ColonRule.swift
+++ b/Source/SwiftLintFramework/Rules/ColonRule.swift
@@ -8,7 +8,7 @@
 
 import SourceKittenFramework
 
-public struct ColonRule: Rule {
+public struct ColonRule: CorrectableRule {
     public static let description = RuleDescription(
         identifier: "colon",
         name: "Colon",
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Rules/ColonRule.swift
```diff
@@ -41,19 +49,112 @@ public struct ColonRule: Rule {
...
+            "func abc(def:  Void) {}\n": "func abc(def: Void) {}\n",
+            "func abc(def :Void) {}\n": "func abc(def: Void) {}\n",
+            "func abc(def : Void) {}\n": "func abc(def: Void) {}\n",
+            "func abc(def: Void, ghi :Void) {}\n": "func abc(def: Void, ghi: Void) {}\n"
         ]
     )
 
     public func validateFile(file: File) -> [StyleViolation] {
-        let pattern = "\\w+\\s+:\\s*\\S+|\\w+:(?:\\s{0}|\\s{2,})\\S+"
+        let pattern = (patterns() as NSArray).componentsJoinedByString("|")
```
jpsim: You could use `SequenceType`'s `joinWithSeparator` here to avoid the cast to `NSArray`.

jpsim:
This is all sorts of awesome! Thank you so much.

I'm on my phone right now so my ability to review is a bit limited though. One thing to make sure of is that the corrections's locations should make sense after they're applied. This might already be the case, but I just want to make sure :). 

Otherwise I really like this implementation, I think we could take hints from this and apply to other correctable rules (like using `reduce`).

Thanks so much!

brianpartridge:
Thanks. Take you time.  I tried to take inspiration from the other CorrectableRules, but there wasn't a good example for ones using regex replacements. Perhaps some of this can be refactored into `File` down the line.

My goal was to keep a balance between efficiency and location accuracy.  For example, I go over the matches in reverse to ensure that one match doesn't impact the range of an 'earlier' match.  This is accurate from some testing, but I may be missing some edge cases. The alternative is to keep matching and fixing the first match in a file until there are none left, but then you're running a regex on each file a potentially large number of times. This is especially true for codebases that need a lot of this formatting automation and are running this for the first time.

Source/SwiftLintFramework/Rules/ColonRule.swift
brianpartridge: Will do.

Issue #249, opened 2015-12-05
.swiftlint.yml not working


tspecht:
Hi guys,

sorry for this probably totally stupid question but I'm not able to get the .swiftlint.yml file to work for excluding rules :( Here's the content of .swiftlint.yml I'm using:

```
disabled_rules: # rule identifiers to exclude from running
  - trailing_whitespace
included: # paths to include during linting. `--path` is ignored if present. takes precendence over `excluded`.
  - MyCodeFolder/Path/
excluded: # paths to ignore during linting. overridden by `included`.
  - Carthage
  - Pods
# parameterized rules can be customized from this configuration file
line_length: 110
# parameterized rules are first parameterized as a warning level, then error level.
type_body_length:
  - 300 # warning
  - 400 # error
# parameterized rules are first parameterized as a warning level, then error level.
variable_name_max_length:
  - 40 # warning
  - 60 # error
# parameterized rules are first parameterized as a warning level, then error level.
variable_name_min_length:
  - 3 # warning
  - 2 # error
```

When I run `swiftlint lint` or `swiftlint lint --config .swiftlint.yml` the file gets not honored ... any ideas?

jpsim:
Your YAML [validates](http://www.yamllint.com/) and when I paste it into SwiftLint's `.swiftlint.yml` it gets parsed properly when I run `swiftlint` from that root.

Are you saying that `Loading configuration from '.swiftlint.yml'` doesn't get printed to the console? Then it must be a path error. Did you perhaps name the file something slightly different (e.g. `.swiftlint.yaml`)?

tspecht:
Yeah, I don't see any log about a configuration being loaded. I checked with `ls -alh` and the file seems to have both correct access rights as well as the correct naming:
`-rw-r--r--   1 myUser  staff   807B Dec  5 16:52 .swiftlint.yml`
In addition I made sure that the file also lies in the very same directory I'm executing `swiftlint lint` from. Is there any way I also could disable rules using command line arguments?

tspecht:
Hmm sorry for the confusion, I guess my brew didn't correctly update swiftlint ... my bad, sorry again.

jeanpierrevg:
Hi, i have the same issue. You fixed it by using brew install swiftlint ? When i try to do that, since I am on OX 10.11.12 it won't install. When I install the swiftlint as a package I can run the swiftlint but it does not take into account the Exclude folders such as Pods ...any ideas?

toshi0383:
[...]

PR #287, opened 2015-12-23
Better failing on yaml parser error

norio-nomura:
- Add printing error when `Yaml.load(_:)` returns error
This will help finding error from YAML parser without using third-party YAML validator.

- Change to fail on YAML parser error if configuration file is required
This will help user to noticing YAML parser error.

This will reduce issues such as #249

### DIFF
--- a/Source/SwiftLintFramework/Models/Configuration.swift
+++ b/Source/SwiftLintFramework/Models/Configuration.swift
@@ -120,6 +120,8 @@ public struct Configuration {
                 queuedPrintError("Loading configuration from '\(path)'")
                 self.init(yaml: yamlContents)!
                 return
+            } else {
+                failIfRequired()
             }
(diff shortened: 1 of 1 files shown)


jpsim:
:+1: thanks! I'm sure this will be helpful to many.

norio-nomura:
Thank you for review and merged.
It would be better that could fail on YAML parser error when using `.swiftlint.yml` implicitly without `--config` option.
That case currently treated as `optional: Bool = true` on calling `init(path:optional:)` as following:
```swift
extension Configuration {
    init(commandLinePath: String) {
        self.init(path: commandLinePath, optional: !Process.arguments.contains("--config"))
    }
```

jpsim:
Ah, I hadn't caught that. Good point. I agree.

PR #288, opened 2015-12-23
Optimize `invalidDocOffsets(_:)`

norio-nomura:
Change guard condition to fails earlier

This commit reduces the duration of linting KeychainAccess from:
```
swiftlint lint  5.37s user 0.14s system 93% cpu 5.889 total
```
to:
```
swiftlint lint  3.74s user 0.13s system 91% cpu 4.238 total
```

the duration of linting Carthage is reduced from:
```
swiftlint lint  33.48s user 1.86s system 82% cpu 42.645 total
```
to:
```
swiftlint lint  29.61s user 1.81s system 79% cpu 39.377 total
```

### DIFF
--- a/Source/SwiftLintFramework/Rules/ValidDocsRule.swift
+++ b/Source/SwiftLintFramework/Rules/ValidDocsRule.swift
@@ -14,11 +14,12 @@ extension File {
         let substructure = (dictionary["key.substructure"] as? XPCArray)?
             .flatMap { $0 as? XPCDictionary } ?? []
         let substructureOffsets = substructure.flatMap(invalidDocOffsets)
-        guard let kind = (dictionary["key.kind"] as? String).flatMap(SwiftDeclarationKind.init),
-            offset = dictionary["key.offset"] as? Int64,
+        guard let kind = (dictionary["key.kind"] as? String).flatMap(SwiftDeclarationKind.init)
+            where kind != .VarParameter,
+            let offset = dictionary["key.offset"] as? Int64,
             bodyOffset = dictionary["key.bodyoffset"] as? Int64,
             comment = getDocumentationCommentBody(dictionary, syntaxMap: syntaxMap)
-            where kind != .VarParameter && !comment.containsString(":nodoc:") else {
+            where !comment.containsString(":nodoc:") else {
                 return substructureOffsets
         }
         let declaration = contents[Int(offset)..<Int(bodyOffset)]


jpsim:
This is great! I didn't even know you could have multiple `where` conditions! I learned something very useful today! :blush:

norio-nomura:
:smile:

norio-nomura:
I think there is room for more performance optimization in SwiftLint and SourceKitten, but that will require breaking changes on SourceKitten and I don't know how that affects to another projects.
And I had read [the post of Realm Blog](https://realm.io/news/swift-opensource/#open-source-swift--realm).

I think I should wait to doing them until the new action appears that hinted in the post. :smiley:

jpsim:
Yeah, there's definitely more room for performance improvements. I don't know if anything we said in the blog post will lead to perf improvements though! I'd expect that using SourceKit in process rather than through XPC calls would speed things up a bit, but probably not all that much.

norio-nomura:
Yes, I expect using SourceKit in process.

By the way, now I'm surprised that swift-2.2 produce faster binaries.:zap:
I builded SwiftLint 26dba4e (with some patch to SourceKitten) using swift-2.2-SNAPSHOT-2015-12-22-a.
the duration of linting KeychainAccess is:
```
swiftlint lint  2.71s user 0.12s system 83% cpu 3.395 total
```
and linting Carthage is:
```
swiftlint lint  21.35s user 1.67s system 77% cpu 29.687 total
```
I didn't think about performance improvements of Swift-2.2.

jpsim:
Whoa... those performance gains from Swift 2.2 are crazy! I can't wait! :smile:

norio-nomura:
FYI, [How to use swift-2.2-SNAPSHOT in xcodebuild](https://gist.github.com/norio-nomura/ea63b61ab942fff2e2b3)

On building SwiftLint, it needs patch for SourceKitten as following:
```sh
sed -i "~" "s/@asmname/@_silgen_name/" Carthage/Checkouts/SourceKitten/Source/SourceKittenFramework/sourcekitd.swift
```
then it can be built by following:
```sh
XCODE_XCCONFIG_FILE=/path/to/swift-latest.xcconfig make install
```

It needs `swift-2.2-SNAPSHOT-2015-12-22-a` or later because that contains fix of [a bug SR-329](https://bugs.swift.org/browse/SR-329).

PR #291, opened 2015-12-25
Fix multibyte handling in many rules

### LINKED ISSUE #279: "Variable Name" rules don't handle multibyte characters
Having a warning:
> warning: Variable Name Min Length Rule Violation: Variable name should be 3 characters or more: currently 2 characters (variable_name_min_length)

The warning appears on several lines of code in my swift file, but I was able to reproduce this warning only on one code sample in a separate swift file:
 
```swift
protocol SampleProtocol {
    func longMethodNameWithCompletion(completion: () -> Void)
    func longMethodNameWithCompletionAndId(id: Int, completion: () -> Void)
    func longMethodNameWithParameters(param1: Int, param2: Int, strParam: String, completion: () -> Void)
    func yetAnotherMethodNameWithSomeParameters(fromTime: NSDate, toTime: NSDate, accountKey: AccountKey, completion: () -> Void)
}
```
The warning is on the `longMethodNameWithParameters` line.

`.swiftlint.yml`:
```yml
disabled_rules: 
  - line_length
  - trailing_newline
``` 

The file  contains only structs and protocols with markdown comments. Removing `line_length` form `.swiftlint.yml`doesn't help.
Any thoughts?

jpsim:
Fixes #279. /cc @norio-nomura @keith


Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -131,7 +131,7 @@ extension File {
         let fileRegions = regions()
         let violatingRanges = violatingRanges.filter { range in
             let region = fileRegions.filter {
-                $0.contains(Location(file: self, offset: range.location))
+                $0.contains(Location(file: self, characterOffset: range.location))
```
norio-nomura: That is not *character* more accurately.
But that is not only here. :wink:
It may hard to keep consistency without using correct parameter types such as `String.UTF16Index`

norio-nomura:
:+1:

Source/SwiftLintFramework/Models/Location.swift
```diff
@@ -26,7 +26,18 @@ public struct Location: CustomStringConvertible, Comparable {
         self.character = character
     }
 
-    public init(file: File, offset: Int) {
+    public init(file: File, byteOffset offset: Int) {
+        self.file = file.path
+        if let lineAndCharacter = file.contents.lineAndCharacterForByteOffset(offset) {
+            line = lineAndCharacter.line
+            character = lineAndCharacter.character
```
norio-nomura: I found that `character` is distance of utf16index, not `String.characters.count`.
But, it seems Xcode count columns by utf16index instead of characters on displaying annotation as following:
screenshot 2015-12-25 12 27 48
So, I stop fix that.

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
jpsim: Good point. I still think I'll keep it as is though.

Source/SwiftLintFramework/Models/Location.swift
jpsim: Interesting.

norio-nomura: I found that
```
"👨‍👩‍👧‍👧".characters.count // 4
```
and above example is very edge case and Xcode fails to position the column indicator sometimes.
screenshot 2015-12-25 13 26 39

By fixing https://github.com/jpsim/SourceKitten/pull/129:
screenshot 2015-12-25 15 18 15

Issue #222, opened 2015-11-19
Add a command to disable a rule for a single line


jpsim:
For example:

```swift
// swiftlint:disable:next force_cast
let forceCastNoWarning = NSNumber() as! Int
let forceCastWithWarning = NSNumber() as! Int
```

Although I'm not convinced on the naming. Maybe `// swiftlint:disable:X` where `X` is an integer number of lines for which to disable the rule?

It would make it more convenient to temporarily disable rules for occasional violations.

segiddins:
Why not just allow putting it at the end of the line, like rubocop does?

jpsim:
> Why not just allow putting it at the end of the line, like rubocop does?

so `// swiftlint:disable rule_name` would start a region if preceded by whitespace, but start and close a single-line region otherwise?

segiddins:
Yeah. Not necessarily saying it's better, and it might clash with the line length rule

richellis:
+1 for the suggestion about putting it at the end of the line. Perhaps an abbreviated syntax to mitigate line length issues, such as `//sl: rule_name`

mpurland:
+1 for this suggestion

jpsim:
Done in #292.

PR #292, opened 2015-12-25
Support command comment modifiers

jpsim:
addresses #222.

I'm happy to reconsider the naming or expand the functionality a bit if others think it would be useful (like specifying the number of lines on which this applies). /cc @realm/swiftlint

### DIFF
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -11,18 +11,22 @@ import SourceKittenFramework
 import SwiftXPC
 
 internal func regex(pattern: String) -> NSRegularExpression {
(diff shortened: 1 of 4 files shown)


Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -115,12 +119,12 @@ extension File {
     }
 
     public func write(string: String) {
-        guard let stringData = string.dataUsingEncoding(NSUTF8StringEncoding) else {
-            fatalError("can't encode '\(string)' with UTF8")
-        }
         guard let path = path else {
             fatalError("file needs a path to call write(_:)")
         }
+        guard let stringData = string.dataUsingEncoding(NSUTF8StringEncoding) else {
```
jpsim: moved this down because it's a more expensive check than `let path = path`.

PR #297, opened 2015-12-27
Add valid_docs rule handle multibyte characters.

### LINKED ISSUE #295: valid_docs rule doesn't handle multibyte characters
test case:
```sh
/// docs👨‍👩‍👧‍👧
/// - returns: false
public func no() -> Bool { return false }
```

norio-nomura:
Fix #295

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@
   [Norio Nomura](https://github.com/norio-nomura)
   [#294](https://github.com/realm/SwiftLint/issues/294)
 
-* valid_docs rule doesn't handle multibyte characters.  
+* The `valid_docs` rule now handles multibyte characters.  
   [Norio Nomura](https://github.com/norio-nomura)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,10 @@
   [JP Simard](https://github.com/jpsim)
   [#279](https://github.com/realm/SwiftLint/issues/279)
 
+* valid_docs rule doesn't handle multibyte characters.
```
jpsim: This should describe the fix rather than the previous bug: ```The `valid_docs` rule now handles multibyte characters.```.

jpsim:
Other than the changelog entry, :+1:. Nice catch.

norio-nomura:
:pray:

jpsim:
This needs to be rebased

norio-nomura:
I rebased. :bow:

PR #300, opened 2015-12-27
test violation locations

jpsim:
We've had several issues in the past due to violations being reported at the wrong location. This PR introduces a *violation marker* (`↓`) in rule description trigger strings that gets stripped out when linting the string, but is used to identify where violations should occur.

With this approach, we should be able to insert arbitrary multibyte characters while confirming that the violation is still triggered at the appropriate location. I'll make a new PR doing just that once this one has been reviewed. /cc @norio-nomura


Source/SwiftLintFrameworkTests/TestHelpers.swift
```diff
@@ -65,6 +70,18 @@ extension XCTestCase {
 
         // Triggering examples violate
         XCTAssertEqual(triggers.flatMap({ violations($0, ruleDescription) }).count, triggers.count)
+        let triggersWithMarkers = triggers.filter { $0.containsString(violationMarker) }
+
+        // Triggering examples with violation markers violate at the marker's location
+        for trigger in triggersWithMarkers {
+            let firstViolation = violations(trigger, ruleDescription).first!
```
keith: Maybe we should combine this and the `triggers.flatMap` above to only calculate the violations once?

keith:
This is great! Might be worth combining those from my comment and :ship:

norio-nomura:
:+1:

Source/SwiftLintFrameworkTests/TestHelpers.swift
jpsim: Ok done in c8fa4e0. Hopefully the perf gain makes the loss of readability worthwhile in time.

keith: If you don't think it will we can revert it. Either way is fine with me.

jpsim:
Thanks for reviewing @keith and @norio-nomura

PR #301, opened 2015-12-27
Recursive Configs

### LINKED ISSUE #299: Separate config for tests (or any sub directory)
There are some cases where I would like to disable certain rules for a sub directory without disabling them for the entire project or excluding that directory. For example, in general I would like to disable force casting and force try's, but I would like to leave these as valid for test code where the exceptions they cause can be considered test failures.

One approach to this might be to allow a `.swiftlint.yml` to be placed in a subdirectory where it will override the global `.swiftlint.yml` (or potentially augment it).

What do you think?

scottrhoyt:
Closes #299

Ok, so here is my first stab at the recursive folder configurations. I think that using a recursive folder search method for the entire linting process would greatly optimize nested configurations, but I implemented this using the simplest algorithm with the current linting search method.

Basically, the configuration to use in linting a file is now a function of the file. To find this configuration, we recursively search backwards from directory of the file until we either find the current config file, reach the root of the lint search, or we find a configuration file to `merge` with. In this initial implementation, `merge` is simply overriding the config.

**NOTE** I primarily put the additional code in an extension to avoid exceeding the line limit for the `Configuration` type, though I suppose it has some encapsulation value to it.

Here are the things that I am still planning to add to this PR:
- [x] Config option to use nested configs (defaults to false for performance)
- [x] Refactor initializing `Configuration` from a yml to be more efficient (it's getting called a lot more)
- [x] Suppress spamming "Loading configuration from " messages intelligently

Let me know what you think so far and I'll continue adding commits to this PR.


Source/SwiftLintFramework/Extensions/QueuedPrint.swift
```diff
@@ -34,9 +34,12 @@ public func queuedPrint<T>(object: T) {
 A thread-safe, newline-terminated version of fputs(..., stderr).
 
 - parameter string: String to print.
+- parameter silent: True suppresses output
 */
-public func queuedPrintError(string: String) {
-    dispatch_async(outputQueue) {
-        fputs(string + "\n", stderr)
+public func queuedPrintError(string: String, silent: Bool = false) {
```
jpsim: Instead of a `silent` parameter that just bypasses the function like this, I'd rather just not call the function in the first place. Could you please revert this change?

Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
```diff
@@ -69,4 +69,16 @@ extension String {
         }
         return nil
     }
+
+    var stringByDeletingLastPathComponent: String {
```
jpsim: For methods on NSString, I prefer not masking them behind a `Swift.String` interface, since that hides the fact that the string has to be cast to NSString. Especially since SwiftLint (and SourceKitten) make heavy use of NSString in its implementation, you then end up casting an NSString into a `Swift.String` to use these methods, which then cast it back to an NSString.

For these reasons, I prefer casting at the call site.

Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
```diff
@@ -69,4 +69,16 @@ extension String {
...
+        return (self as NSString).stringByDeletingLastPathComponent
+    }
+
+    func stringByAppendingPathComponent(pathComponent: String) -> String {
+        return (self as NSString).stringByAppendingPathComponent(pathComponent)
+    }
+
+    public func absolutePathStandardized() -> String {
```
jpsim: It's possible that we never actually need `absolutePathRepresentation`, and that `stringByStandardizingPath` would be a sufficient replacement.

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -89,23 +94,36 @@ public struct Configuration {
     }
 
     public init?(yaml: String) {
-        let yamlResult = Yaml.load(yaml)
```
jpsim: Was this done intentionally? I think this incorrectly reverts the change introduced in a1f734e, which prints more informative errors when encountering misconfigured YAML.

Source/SwiftLintFramework/Extensions/QueuedPrint.swift
scottrhoyt: Sure, but because the new code results in a new configuration load for every file under a nested configuration, reverting this will result in dozens of "Loading configuration from " messages when using nested configs. I found this really obfuscates the logs. Another option would be to suppress the logging inside the `init` function, but that didn't seem right to me. My intention here was to centralize the logging logic in one place so that the logger could decide how to process. Because we wouldn't have to hunt down selective logging in methods, this could potentially ease the transition to more robust logging if needed. So let me know which of the three solutions you'd like:
 1. Revert to original, which is the simplest but logs will be muddled with new load messages
 2. Put the selective logging logic in the `init` function which requires no change to logging functions, but makes `init` more complicated and sets the precedent of distributing logging logic to non-logging objects.
 3. Put the selective logging logic in the logger which solves the above problems but results in a degenerative trivial silent logging case

Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
scottrhoyt: That motivation makes a lot of sense to me. My thinking was mostly DRY, and how it would make it trivial to swap this out project-wide with a Swift `String` native approach if desired. But I'd be happy to revert.

Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
scottrhoyt: Certainly possible. I didn't do much testing other than realizing `stringByStandardizingPath` was necessary and following the pattern from [here](https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Extensions/NSFileManager%2BSwiftLint.swift#L18).

Source/SwiftLintFramework/Models/Configuration.swift
scottrhoyt: Yes, it was intentional. I refactored the initialization code so that it doesn't `Yaml.load` twice and `init` twice every time you initialize from a *.yml. This was previously done to detect errors in loading the *.yml and revert to a clean config if possible. The new approach separates the loading of the *yml (and the printing of the loading errors) into a separate method so that `Yaml.load` and `init` are only called once per initialization. It still preserves the error printing.

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -89,23 +94,36 @@ public struct Configuration {
...
+        if let yamlConfig = yamlResult.value {
+            return yamlConfig
+        } else {
+            if let error = yamlResult.error {
+                queuedPrint(error)
+            }
+            return nil
+        }
+    }
+
```
scottrhoyt: ^^ This is the refactored init code.

Source/SwiftLintFramework/Extensions/QueuedPrint.swift
jpsim: > Put the selective logging logic in the init function which requires no change to logging functions, but makes init more complicated and sets the precedent of distributing logging logic to non-logging objects.

This is what I was trying to say. The logging in Configuration's initializer is already unfortunate but in fact, it should be relatively straightforward to move it to the `swiftlint` CLI app instead, which would address the need to silence some configuration loads without polluting the model object with logging code.

Ideally, all logs would be in `swiftlint` with none in `SwiftLintFramework`.

scottrhoyt: Ok, I will revert this and instead put the conditional logging inside of `Configuration.init`. Stay tuned! :)

scottrhoyt:
Logging is now suppressed in `Configuration.init(...)`.

Source/SwiftLintFramework/Models/Configuration.swift
jpsim: I see, thanks for taking the time to explain :smile:

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -118,9 +136,12 @@ public struct Configuration {
         do {
             let yamlContents = try NSString(contentsOfFile: fullPath,
                 encoding: NSUTF8StringEncoding) as String
-            if let _ = Configuration(yaml: yamlContents) {
-                queuedPrintError("Loading configuration from '\(path)'")
-                self.init(yaml: yamlContents)!
+            if let yamlConfig = Configuration.loadYaml(yamlContents) {
+                if !silent {
```
jpsim: :+1: much better IMO

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -188,4 +209,37 @@ public struct Configuration {
...
+            NSFileManager.defaultManager().fileExistsAtPath(configSearchPath) {
+            return merge(Configuration(path: configSearchPath, optional: false, silent: true))
+        }
+
+        // If we are not at the root path, continue down the tree
+        if path != rootPath {
+            return configForPath(path.stringByDeletingLastPathComponent)
+        }
+
+        // If nothing else, return ourselves
```
jpsim: singular? `return self`

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -188,4 +209,37 @@ public struct Configuration {
...
+        return self
+    }
+
+    private func merge(config: Configuration) -> Configuration {
```
jpsim: might be worth adding a note to eventually expand the logic to merge configurations rather than replace the current one.

jpsim:
This is looking quite good, @scottrhoyt! Please don't let my many comments discourage you :wink:.

It would be nice to have some tests for this. The rest of the project is pretty well tested IMO, so it'd be nice to keep that up here.

It might make sense to introduce a `Fixtures/` directory in `SwiftLintFrameworkTests/` since much of this functionality relies on directory structure. Otherwise, some straightforward unit tests could be sufficient.

scottrhoyt:
@jpsim No worries. I can definitely sympathize with being a test nut. I wrote some tests for for this in the best way I could come up with. All in all, writing the tests took longer than the code (I hate when that happens), but that was mostly because of the things I needed to support the tests. This is what I did:
  - Enabled code coverage generation for the project
  - Added the `isEqual` method to the `Rule` and `ParameterizedRule` protocols via extensions
  - Wrote conformance for `RuleParameter` and `Configuration` to `Equatable`
  - Wrote tests for all of the above
  - Wrote a test for `Configuration.merge`
  - Created a mock project structure in the `Resources` folder of the testing bundle with mock Swift files and mock *.yml's
  - Tested recursive configuration search with the mock project structure.

Let me know what you think.

Source/SwiftLintFramework/Protocols/Rule.swift
```diff
@@ -13,12 +13,35 @@ public protocol Rule {
...
+
 public protocol CorrectableRule: Rule {
     func correctFile(file: File) -> [Correction]
 }
+
+// MARK: - == Implementations
+
+func == (lhs: [Rule], rhs: [Rule]) -> Bool {
+    if lhs.count == rhs.count {
+        return zip(lhs, rhs).map { $0.isEqualTo($1) }.reduce(true) { $0 && $1 }
```
jpsim: I think you can avoid creating the intermediate closures here:

```swift
zip(lhs, rhs).map(Rule.isEqualTo).reduce(true, &&)
```

Source/SwiftLintFrameworkTests/Resources/ProjectMock/Level1/Level2/.swiftlint.yml
```diff
@@ -0,0 +1,9 @@
+disabled_rules:
+  - force_try
+included:
+  - "everything"
+exluded:
+  - "the place where i commited many coding sins"
+line_length: 10000000000
+reporter: "json"
+use_nested_configs: true
```
jpsim: missing trailing newline here

Source/SwiftLintFrameworkTests/Resources/ProjectMock/.swiftlint.yml
```diff
@@ -0,0 +1,9 @@
+disabled_rules:
+  - force_cast
+included:
+  - "everything"
+exluded:
+  - "the place where i commited many coding sins"
```
jpsim: typo: s/commited/committed/

Source/SwiftLintFrameworkTests/Resources/ProjectMock/Level1/Level2/.swiftlint.yml
```diff
@@ -0,0 +1,9 @@
+disabled_rules:
+  - force_try
+included:
+  - "everything"
+exluded:
+  - "the place where i commited many coding sins"
```
jpsim: same typo: typo: s/commited/committed/

SwiftLint.xcodeproj/xcshareddata/xcschemes/swiftlint.xcscheme
```diff
@@ -26,7 +26,8 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      shouldUseLaunchSchemeArgsEnv = "YES">
+      shouldUseLaunchSchemeArgsEnv = "YES"
+      codeCoverageEnabled = "YES">
```
jpsim: :+1:

Source/SwiftLintFrameworkTests/ConfigurationTests.swift
```diff
@@ -6,7 +6,8 @@
 //  Copyright © 2015 Realm. All rights reserved.
 //
 
-import SwiftLintFramework
+@testable import SwiftLintFramework
+import SourceKittenFramework
```
jpsim: is this import used anywhere? I can't see.

Source/SwiftLintFrameworkTests/ConfigurationTests.swift
```diff
@@ -6,7 +6,8 @@
 //  Copyright © 2015 Realm. All rights reserved.
 //
 
-import SwiftLintFramework
+@testable import SwiftLintFramework
```
jpsim: I think it's fine to make `isEqualTo` part of the public API, therefore unnecessary to import this as `@testable`

Source/SwiftLintFramework/Protocols/Rule.swift
```diff
@@ -13,12 +13,35 @@ public protocol Rule {
     func validateFile(file: File) -> [StyleViolation]
 }
 
+extension Rule {
```
jpsim: Is it not possible to make `Rule` conform to `Equatable` and implement `==`? Would be a bit more Swifty than `isEqualTo`. Or `ParameterizedRule` for that matter.

Source/SwiftLintFrameworkTests/Resources/ProjectMock/Level1/Level2/.swiftlint.yml
scottrhoyt: Got it. I'll clean these files up. Thanks.

Source/SwiftLintFrameworkTests/ConfigurationTests.swift
scottrhoyt: Just for `File`.

Source/SwiftLintFrameworkTests/ConfigurationTests.swift
scottrhoyt: I think that's probably ok. I get slightly nervous with equality stuff because it's easy for someone to add a member to the respective type without including it in the member-wise check. You can somewhat mitigate this by using Swift's limited reflection to write a test to assert that the number of members hasn't changed. But I often just keep this stuff out of public API until the data structures have stabilized or it's otherwise needed.

Source/SwiftLintFramework/Protocols/Rule.swift
scottrhoyt: Well it would be possible. The problem is that making `Rule` conform to `Equatable` will impose a `Self` requirement on `Rule`, therefore you won't be able to use it as a type (e.g. `[Rule]`) and only as a generic constraint. This exact case is covered really well in the 2015 WWDC session on Protocol Orientated Programming, and implementing `isEqualTo` for situations like this was the recommended solution. Now, that still doesn't stop us from implementing `==` as well using `isEqualTo`, and I originally did that, but then took it out as because these sort of semantics should probably be discussed before implemented just because it's not 100% clear. In addition to the WWDC video, there's a pretty decent discussion of it [here](http://www.llamagraphics.com/content/combining-swift-protocols). I can definitely implement it with `Equatable`, but I'll have to modify all the cases that the code uses `Rule` as a type.

Source/SwiftLintFrameworkTests/ConfigurationTests.swift
jpsim: Ah, :+1: then

Source/SwiftLintFrameworkTests/ConfigurationTests.swift
jpsim: Keeping as-is is fine.

Source/SwiftLintFramework/Protocols/Rule.swift
jpsim: Keeping as-is is fine then.

jpsim:
This is looking really good, could you please add an entry to the CHANGELOG following [these guidelines](https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md) and add some basic documentation in the README?

scottrhoyt:
@jpsim CHANGELOG and README updated. Let me know if you need anything else.

scottrhoyt:
I pulled master into my branch to resolve the conflicts in `README` and `CHANGELOG`

Carthage/Checkouts/SourceKitten
```diff
@@ -1 +1 @@
-Subproject commit 0ed186552d8f9f3e128ab382581102c55c2d3c73
```
scottrhoyt: @jpsim might want to make sure this is as intended? Seems to have been switched during merging master.

jpsim:
Many many thanks for this work @scottrhoyt! I made a few tweaks and merged in #307.

Issue #310, opened 2016-01-03
Always fail when a YAML file was found but couldn't be parsed


jpsim:

scottrhoyt:
@jpsim I believe this is good to close?

jpsim:
Yes, done in #311. Thanks.

PR #311, opened 2016-01-03
fail if YAML doesn't parse

jpsim:
Addresses #310 /cc @NachoSoto

### DIFF
--- a/Source/SwiftLintFramework/Models/Configuration.swift
+++ b/Source/SwiftLintFramework/Models/Configuration.swift
@@ -124,11 +124,9 @@ public struct Configuration: Equatable {
 
     public init(path: String = ".swiftlint.yml", optional: Bool = true, silent: Bool = false) {
         let fullPath = (path as NSString).absolutePathRepresentation()
-        let failIfRequired = {
-            if !optional { fatalError("Could not read configuration file at path '\(fullPath)'") }
-        }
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -143,10 +141,10 @@ public struct Configuration: Equatable {
                 configPath = fullPath
                 return
             } else {
-                failIfRequired()
+                fail()
```
NachoSoto: Makes sense!

NachoSoto:
Awesome :+1:

Source/SwiftLintFramework/Models/Configuration.swift
jpsim: Ah yeah, that could work too, although the point of the local `fail()` function is to keep the failure-related logic in there anyway, so I'll keep it as is.

jpsim:
Thanks for reviewing!

Source/SwiftLintFramework/Models/Configuration.swift
scottrhoyt: @NachoSoto @jpsim That is now the [current behavior](https://github.com/realm/SwiftLint/blob/69477dab924ff61854c90903b30090ff74b799de/Source/SwiftLintFramework/Models/Configuration.swift#L107-L109).

PR #314, opened 2016-01-04
Dictionary Parameterization of rules

### LINKED ISSUE #231: Special case id/exceptions for NameFormatViolation
We tend to have properties called `id` on objects and we always get the `NameFormatViolation` error because of it. Generally I like this rule but it's annoying when it does it for ID. I'd propose 2 solutions, one temporary, one more permanent.

1. Special case `id` so that it won't trigger this violation. It's a pretty common thing I'd assume and I haven't seen anyone take issue with something called `id` before.
2. Add configuration for "exceptions" to this rule and the user could then add "id" to the list of exceptions.

screen shot 2015-11-23 at 16 04 26

Thoughts?

### LINKED ISSUE #303: Switch to dictionary format for parameterized rules
In addition to not being self documenting, the current array format of parameterizing config values in `.swiftlint.yml` is proving to be a hurdle to allowing more customization for rules (see #191 #130 #231 ). It's also complicating providing a better way to `merge` nested configs for #301 .

I propose we switch the format to a dictionary format. e.g.:

    variable_name_max_length:
      warning: 40
      error: 60
      exclude: "id"

This would require an overhaul to how `ParameterizedRule` works. It would also make breaking changes for previous versions of the `.swiftlint.yml` file, so it probably shouldn't happen until at least the 1.0 release. However, I'd be willing to start looking at how to best accomplish this.

scottrhoyt:
# Dictionary Parameterization

Closes #303

Here is my idea on how to implement dictionary parameterization for rules. Sorry it is such a beast of a change (and hard to review), but it was necessary IMO to accomplish the task. There are some pretty big changes in here, so I understand if we need to refine this further or even reject this course.

## Merge Notes

I merged my recursive configuration branch into this branch before it was integrated into master. In addition to that I have merged master several times to keep up with the other changes. Somehow this resulted in the recursive config commits being duplicated, though the diff's still look correct. This could be a result of the combination of how I merged and the cherry picking that was done to do the final integration.

## Overview

This was my thought process as I developed this:

 * With the increased complexity in rule configuration, it didn't make sense to have `Configuration` store the knowledge on how to configure each rule anymore. This should be encapsulated in the `Rule` itself.
 * A new protocol, `ConfigurableRule`, was created to facilitate this.
 * When encapsulating the configuration in `ConfigurableRule`, I didn't want to create a dependency on `Yaml`, I chose instead for `ConfigurableRule` to have a failable initializer taking an `AnyObject` instead. `AnyObject` was chosen over the alternative, `[String: AnyObject]`, to support the trivial config cases (e.g. `line_length: 100`).
 * I wrote `Yaml` extensions to flatten the `Yaml.Dictionary`s into `[String: AnyObject]`s
 * I then refactored `Configuration` to extract Yaml parsing to `YamlParser`
 * I also didn't think it made sense anymore to have `Configuration` store the master rule list, so I extracted this to the global `masterRuleList` using a rather trivial `RuleList` container I created.
 * I wrote protocol extensions for `ParameterizedRule` to make all legacy parameterized rules automatically conform to `ConfigurableRule`
 * `Configuration` now constructs it's rule array by iterating through a `RuleList` (defaulting to `masterRuleList`). If a rule is a `ConfigurableRule` and a configuration exists for this rule, the rule will be instantiated by passing it's configuration to `init?(config:)`. If this fails, the rule is not a `ConfigurableRule`, or a configuration does not exist, the `Rule` is instantiated by calling it's default `init`.
 * I added the `init()` requirement to the `Rule` protocol to support the initialization of `Rule`s with their default values.
 * I added the trivial case `init() { }` to all non-`ConfigurableRule`s to support this.
 * I wrote unit tests for most of the above.
 * I adapted `VariableNameMinLengthRule` as an example of how to create a `ConfigurableRule`. The new implementation supports the same syntax as before, but now also supports configuration via `warning` and `error` parameters as well as a new `excluded` array parameter to exclude certain variable names from the check (e.g. `id`). Closes #231.

        variable_name_min_length:
          warning: 3
          error: 4
          excluded:
            - id

## Thoughts

 * I'm not crazy about the `AnyObject` parameter for [configuration](https://github.com/realm/SwiftLint/blob/sh-dictionary-parameters/Source/SwiftLintFramework/Protocols/Rule.swift#L35). I feel like it's fighting the strong type system a bit. But on the other hand, it seems like a necessary evil to support the use cases we would like. Also, this follows in the well-tread footsteps of serializing objects, and it makes it relatively easy to support any serialization technique for configurations (e.g. json, xml, or plist). The plist option is particularly nice because it would support a tighter integration of SwiftLint into Xcode.
 * Because we store `Rules` in a `[Rule]` for `Configuration` and I needed to implement equality testing (`isEqualTo` and `Equatable`) for `Configuration` to support unit tests, I needed a way of dealing with two `ConfigurableRule`s being compared while being downcasted to `Rule`s. This result in a somewhat [ugly](https://github.com/realm/SwiftLint/blob/sh-dictionary-parameters/Source/SwiftLintFramework/Protocols/Rule.swift#L18-L24) change to the implementation of `Rule.isEqualTo` that tests for upcasting to `ConfigurableRule`s. This feels like fighting the type system, but I couldn't come up with a way around it. However, this is now a potential trap if more rule protocols are added.
 * There appears to be some sort of bug in how SwiftXPC overrides comparing two `[String]`s that results in two empty `[String]`s not being equal. I had to therefore test for equality in a different [way](https://github.com/realm/SwiftLint/blob/sh-dictionary-parameters/Source/SwiftLintFramework/Rules/VariableNameMinLengthRule.swift#L78) in `VariableNameMinLengthRule`.
 * This is a beginning, but we can further this work by writing additional protocols and methods to support common configuration schemes to ease new `ConfigurableRule` creation. We can also further restrict configuration to remove the `AnyObject` requirement or create a new type to facilitate this.
 * Another project I have contributed to, [ObjectMapper](https://github.com/Hearst-DD/ObjectMapper), can potentially be used to add some nice syntactical sugar to the syntax of `ConfigurableRule`s.
 * I originally wrote the `Yaml` flattening as extensions, but it might make more sense to put those in `YamlParser`.


scottrhoyt:
Also, if we go this route, `ParameterizedRule` can be deprecated. It already wasn't being used much at all since all conforming rules were referenced directly. Every `ParameterizedRule` can be rewritten as a `ConfigurableRule`, I just didn't take the time to do that except for `VariableNewMinLengthRule`.

jpsim:
Thanks for the PR, it will take some time to review :grimacing:

scottrhoyt:
No problem. Just post any questions you have about implementation details or design decisions, and I'll be happy to discuss.

scottrhoyt:
Hey @jpsim. I was wondering if you had any thoughts on this yet? I just don't want to let it go too long because of merge conflicts. If you are ready to take another look, I can clean up whatever conflicts have arisen.

Source/SwiftLintFramework/Extensions/Yaml+SwiftLint.swift
```diff
@@ -0,0 +1,75 @@
...
+            }
+            return newDict
+        } else if self.count == 0 || self == .Null {
+            return [:]
+        }
+
+        return nil
+    }
+
+    var flatArray: [AnyObject]? {
```
jpsim: One liner: `var flatArray: [AnyObject]? { return array?.map { $0.flatValue } }`

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -133,65 +105,40 @@ public struct Configuration: Equatable {
...
-
-    private static func parameterRulesFromYAML(yaml: Yaml? = nil) -> [Rule] {
-        let intParams: (Rule.Type) -> [RuleParameter<Int>]? = {
-            (yaml?[.String($0.description.identifier)].arrayOfInts).map(ruleParametersFromArray)
+    public static func rulesFromDict(dict: [String: AnyObject]? = nil,
+                          ruleList: RuleList = masterRuleList) -> [Rule] {
+        var rules = [Rule]()
+        for rule in ruleList.list.values {
+            let identifier = rule.description.identifier
+            if rule is ConfigurableRule.Type,
```
jpsim:
```swift
if let ConfigurableRuleType = rule as? ConfigurableRule, ruleConfig = dict?[identifier] {
  if let configuredRule = ConfigurableRuleType.init(config: ruleConfig) {
    rules.append(configuredRule)
  } else {
    queuedPrintError("Invalid config for '\(identifier)'. Falling back to default.")
    rules.append(rule.init())
  }
} else {
  rules.append(rule.init())
}
```

Source/SwiftLintFramework/Models/RuleParameter.swift
```diff
@@ -14,6 +14,10 @@ public struct RuleParameter<T: Equatable>: Equatable {
         self.severity = severity
         self.value = value
     }
+
+    static func ruleParametersFromArray<T>(array: [T]) -> [RuleParameter<T>] {
```
jpsim: Makes total sense to move here, :+1:

Source/SwiftLintFramework/Protocols/Rule.swift
```diff
@@ -32,13 +43,40 @@ extension ParameterizedRule {
     }
 }
 
+extension ParameterizedRule where Self: ConfigurableRule, ParameterType == Int {
```
jpsim: I think this can be generalized to avoid the constraint on `Int`. What happens when you replace the references to `Int` in this extension to `ParameterType`?

Source/SwiftLintFramework/Protocols/Rule.swift
```diff
@@ -32,13 +43,40 @@ extension ParameterizedRule {
...
+
+    public init?(config: AnyObject) {
+        if let array = Self.arrayOfInts(config) {
```
jpsim: A `guard` might be more semantic here:

```swift
guard let array = Self.arrayOfInts(config) else {
  return nil
}
self.init(parameters: RuleParameter<Int>.ruleParametersFromArray(array))
```

Source/SwiftLintFramework/Rules/VariableNameMinLengthRule.swift
```diff
@@ -1,27 +1,40 @@
 //
-//  VariableNameMinLengthRule.swift
+//  VariableNameMinLengthNewRule.swift
```
jpsim: VariableNameMinLengthNewRule?

Also, did you accidentally re-attribute this file to yourself?

Source/SwiftLintFrameworkTests/ConfigurationTests.swift
```diff
@@ -12,16 +12,14 @@ import XCTest
...
     func testInit() {
-        XCTAssert(Configuration(yaml: "") != nil,
-            "initializing Configuration with empty YAML string should succeed")
-        XCTAssert(Configuration(yaml: "a: 1\nb: 2") != nil,
-            "initializing Configuration with valid YAML string should succeed")
-        XCTAssert(Configuration(yaml: "|\na") == nil,
-            "initializing Configuration with invalid YAML string should fail")
+        XCTAssert(Configuration(dict: [:]) != nil,
+            "initializing Configuration with empty Dictionary should succeed")
+        XCTAssert(Configuration(dict: ["a" : 1, "b" : 2]) != nil,
```
jpsim: Please use no space to the left, and a single space to the right, of colons when using dictionary literals to stay consistent with the style of the rest of the project.

Source/SwiftLintFrameworkTests/Yaml+SwiftLintTests.swift
```diff
@@ -0,0 +1,62 @@
...
+//
+//  Created by Scott Hoyt on 12/28/15.
+//  Copyright © 2015 Realm. All rights reserved.
+//
+
+import XCTest
+import Yaml
+@testable import SwiftLintFramework
+
+class YamlSwiftLintTests: XCTestCase {
```
jpsim: Yaml is sufficiently tested in https://github.com/behrang/YamlSwift IMO, so these tests (and therefore `test.yml` too) are unnecessary.

jpsim: D'oh! I didn't read this right. This tests flattening YAML. These tests are fine. I said nothing.

Source/SwiftLintFrameworkTests/YamlParserTests.swift
```diff
@@ -0,0 +1,30 @@
+//
```
jpsim: Ah, *this* file is superfluous since Yaml is tested in its own project. Please remove.

SwiftLint.xcodeproj/project.pbxproj
```diff
@@ -10,13 +10,19 @@
 		02FD8AEF1BFC18D60014BFFB /* ExtendedNSStringTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 02FD8AEE1BFC18D60014BFFB /* ExtendedNSStringTests.swift */; };
 		1F11B3CF1C252F23002E8FA8 /* ClosingBraceRule.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1F11B3CE1C252F23002E8FA8 /* ClosingBraceRule.swift */; };
 		24E17F721B14BB3F008195BE /* File+Cache.swift in Sources */ = {isa = PBXBuildFile; fileRef = 24E17F701B1481FF008195BE /* File+Cache.swift */; };
+		3B1150CA1C31FC3F00D83B1E /* Yaml+SwiftLint.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3B1150C91C31FC3F00D83B1E /* Yaml+SwiftLint.swift */; };
```
jpsim: I've tried to keep this sorted as much as possible. Could you please make sure this stays at least as sorted as it was before?

Source/SwiftLintFramework/Extensions/Yaml+SwiftLint.swift
```diff
@@ -0,0 +1,75 @@
...
+        case .Bool(let myBool):
+            return myBool
+        case .Int(let myInt):
+            return myInt
+        case .Double(let myDouble):
+            return myDouble
+        case .String(let myString):
+            return myString
+        case .Array:
+            return self.flatArray! // This is valid because .Array will always flatten
```
jpsim: elsewhere in the project, we omit references `self` if they're not necessary.

jpsim:
@scottrhoyt I really like this. It makes a lot more sense than our current approach. I like the way you built it too.

I've commented a lot of nitpicky stuff as line notes, but taking a step back and looking at the big picture, there's just one thing that jumps out at me for now: can we consolidate the `ConfigurableRule` and `ParameterizedRule` protocols? Is seems like we should just need one of these, no?

Otherwise, once you've had a chance to rebase against `master` and address my feedback, I'll take another look.

Source/SwiftLintFrameworkTests/YamlParserTests.swift
scottrhoyt: This is mostly testing that the `YamlParserError` that I introduced works as expected. Would you like to not test this? Or even switch parsing to return an optional instead of throwing?

SwiftLint.xcodeproj/project.pbxproj
scottrhoyt: Best I can tell, the build files are still sorted by file UUID. Is this not what you are seeing, or are you using a different sort?

Source/SwiftLintFramework/Rules/VariableNameMinLengthRule.swift
scottrhoyt: Yeah, this was done accidentally as I originally created a new rule `VariableNameMinLengthNewRule` to ensure the results of both rules were the same. Then I consolidated by deleting the old rule. I fixed and reattributed.

scottrhoyt:
Thanks, I'm glad you like it. I think it's a good direction as well. I fixed most of the suggestions and asked for follow up on a couple.

Yes, there is now no need for `ParamterizedRule` and `ConfigurableRule`. `ParameterizedRule` was left as is as a demonstration of the new approach, and also because I didn't want to rewrite all the old `ParameterizedRule`s until I was sure we wanted to go the direction. But as I mentioned above, if we go this route, `ParameterizedRule` can be deprecated or removed outright.

Once I get that clarification on the couple of outstanding edits, I will rebase against `master`.

Source/SwiftLintFrameworkTests/YamlParserTests.swift
jpsim: My mistake again. I had completely missed `YamlParserError` when reviewing. This looks good.

Source/SwiftLintFramework/Rules/VariableNameMinLengthRule.swift
```diff
@@ -42,15 +55,28 @@ public struct VariableNameMinLengthRule: ASTRule, ParameterizedRule {
...
+                        reason: "Variable name should be \(parameter.value) characters " +
+                        "or more: currently \(charCount) characters")]
+                }
             }
             return []
         } ?? []
     }
+
+    public func isEqualTo(rule: ConfigurableRule) -> Bool {
+        if let rule = rule as? VariableNameMinLengthRule {
```
jpsim: good place for a `guard` here

Source/SwiftLintFrameworkTests/ConfigurationTests.swift
```diff
@@ -33,11 +31,11 @@ class ConfigurationTests: XCTestCase {
     }
 
     func testDisabledRules() {
-        let disabledConfig = Configuration(yaml: "disabled_rules:\n  - nesting\n  - todo")!
+        let disabledConfig = Configuration(dict: ["disabled_rules" :  ["nesting", "todo"]])!
```
jpsim: dict colon spacing

Source/SwiftLintFrameworkTests/ConfigurationTests.swift
```diff
@@ -46,21 +44,20 @@ class ConfigurationTests: XCTestCase {
         XCTAssertEqual(expectedIdentifiers, configuredIdentifiers)
 
         // Duplicate
-        let duplicateConfig = Configuration( yaml: "disabled_rules:\n  - todo\n  - todo")
+        let duplicateConfig = Configuration(dict: ["disabled_rules" :  ["todo", "todo"]])
```
jpsim: dict colon spacing

Source/SwiftLintFrameworkTests/ConfigurationTests.swift
```diff
@@ -46,21 +44,20 @@ class ConfigurationTests: XCTestCase {
...
         XCTAssert(duplicateConfig == nil, "initializing Configuration with duplicate rules in " +
-            " YAML string should fail")
+            " Dictionary should fail")
```
jpsim: previous issue, but double space in the failing message

Source/SwiftLintFrameworkTests/ConfigurationTests.swift
```diff
@@ -138,11 +135,28 @@ class ConfigurationTests: XCTestCase {
     }
 
     func testDoNotUseNestedConfigs() {
-        var config = Configuration(yaml: "use_nested_configs: false\n")!
+        var config = Configuration(dict: ["use_nested_configs" : false])!
```
jpsim: dict colon spacing

Source/SwiftLintFrameworkTests/Yaml+SwiftLintTests.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    override func setUp() {
```
jpsim: not necessary to override `setUp` just to forward to `super`.

SwiftLint.xcodeproj/project.pbxproj
jpsim: Hmm, now I'm not sure either. As long as they're sorted in the Xcode UI I'm happy

jpsim:
> Yes, there is now no need for ParamterizedRule and ConfigurableRule. ParameterizedRule was left as is as a demonstration of the new approach, and also because I didn't want to rewrite all the old ParameterizedRules until I was sure we wanted to go this direction. But as I mentioned above, if we go this route, ParameterizedRule can be deprecated or removed outright--and probably should be removed--before merging.

Yup, let's remove `ParameterizedRule` altogether then!

scottrhoyt:
Most of this is cleaned up now. My thinking on removing `ParameterizedRule`s is to initially just move the protocol extension methods into the rule classes themselves and keep the existing initialization implementation. Then we can either wait until more configuration options are added to each rule (which will necessitate changing how `init` works), or open a separate branch for removing any redundant configuration? Alternatively I can embark upon the conversion now, but that might delay this merge by a couple of days due to my current work load.

scottrhoyt:
Also, with regards to the rebase, I must confess I haven't rebased much, but I'm familiar with the concept and the intended results of cleaning up this commit history. However, I was pretty sure rebasing after pushing to a remote was a no no. Is that not the case? Or should we use a work around like creating a fresh branch to do the rebase and then pushing to github with that, creating a new PR from there, and deleting this branch?

jpsim:
I don't want to take up more of your time, so how about you rebase this to address the merge conflicts and I can do a final review pass as-is before we embark on deprecating `ParameterizedRule`.

> Also, with regards to the rebase, I must confess I haven't rebased much, but I'm familiar with the concept and the intended results of cleaning up this commit history. However, I was pretty sure rebasing after pushing to a remote was a no no. Is that not the case? Or should we use a work around like creating a fresh branch to do the rebase and then pushing to github with that, creating a new PR from there, and deleting this branch?

Some projects discourage rebasing because it does in a sense rewrite history, but I value legible commit history in SwiftLint enough to prefer rebasing over verbose merge noise.

To rebase, first make sure that your local `master` branch is up to date with this remote, then from your branch you can run `git rebase master` and use `git mergetool` to address merge conflicts as they occur.

scottrhoyt:
Great. I'll get that done later today. Something to put in the backlog after this merge will be a new system for documenting rules: so that we can provide documentation via web and command line for the new expanded configuration options. I suppose that will be contingent on what the standards are for creating `ConfigurableRule`s.

jpsim:
> Great. I'll get that done later today. Something to put in the backlog after this merge will be a new system for documenting rules: so that we can provide documentation via web and command line for the new expanded configuration options. I suppose that will be contingent on what the standards are for creating ConfigurableRules.

This is one downside to `ConfigurableRule`'s untyped approach rather than `ParameterizedRule`, which was type-safe.

jpsim:
@scottrhoyt let me know if you're too busy to get to this, I don't mind taking care of the rebase.

scottrhoyt:
Sorry, I'm finding the rebase a bit challenging, but let me try to work through it one more time. I'll let you know if I need some assistance.

scottrhoyt:
Ok, so I completed the rebase. (That was a bit of a doozy. I hate *.xcodeproj). So what should I do now? Should I do a force push to github or push the rebase into a different branch?

scottrhoyt:
Force pushed.

jpsim:
@scottrhoyt this is looking good! Could you update the README and CHANGELOG to reflect these changes?

scottrhoyt:
Yeah, will do.

scottrhoyt:
@jpsim , as far as updating the README, what I am thinking is to just include a new section, "Creating A Rule", to explain how to create `ConfigurableRule`s. Since no functionality should have changed on the user side, we should be good with just that for now. However, we will need to figure out how we want to go about documenting individual rule configuration options going forward. I think a good place to start that work would be in the next stage when we remove `ParameterizedRule` altogether. What are your thoughts?

jpsim:
Works for me! Although "Creating a Rule" should probably be worked into the current [Rules section of the contributing.md document](https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#rules).

Again, let me know if you'd like help rebasing this since there are now merge conflicts due to merging a few PRs over the last few days. This is a downside of working on an actively developed project, I'm afraid :grimacing:.

scottrhoyt:
@jpsim, please review these last two commits while I rebase and adjust the CHANGELOG and README. 

d65d873 fixed an omission to fully reimplementing `VariableNameMinLengthRule` as a `ConfigurableRule`.

0d9c03a refactored the `ParameterizedRule` conformance to `ConfigurableRule` by having `ParameterizedRule` inherit from `ConfigurableRule`. IMO this is a bit cleaner way to implement this.

jpsim:
> 0d9c03a refactored the ParameterizedRule conformance to ConfigurableRule by having ParameterizedRule inherit from ConfigurableRule. IMO this is a bit cleaner way to implement this.

This makes sense. It will make even more sense when we merge `ConfigurableRule` with `ParameterizedRule`.

Source/SwiftLintFramework/Protocols/Rule.swift
```diff
@@ -9,26 +9,56 @@
...
+            return array
+        } else if let number = obj as? ParameterType {
+            return [number]
+        }
+        return nil
+    }
+
+    public func isEqualTo(rule: ConfigurableRule) -> Bool {
+        if let rule = rule as? Self {
+            return self.parameters == rule.parameters
```
jpsim: can omit `self` here probably?

Source/SwiftLintFramework/Protocols/Rule.swift
```diff
@@ -9,26 +9,56 @@
...
+    static func arrayOf(obj: AnyObject?) -> [ParameterType]? {
+        if let array = obj as? [ParameterType] {
+            return array
+        } else if let number = obj as? ParameterType {
+            return [number]
+        }
+        return nil
+    }
+
+    public func isEqualTo(rule: ConfigurableRule) -> Bool {
```
jpsim: whole function can be `return parameters == (rule as? Self)?.parameters`?

scottrhoyt:
@jpsim Incorporated latest feedback and updated CHANGELOG and CONTRIBUTING. Let me know if there is anything else.

jpsim:
This looks ok to me, but it's a large change, so I'd love to get some feedback from @realm/swiftlint thoughts?

nicolai86:
I don't like that rule configuration parsing is mixed with the validation and correcting of rules. Maybe splitting these concerns would make sense down the line. 
E.g. A rule taking a configuration struct, and a configuration parser needs to be declared together with the rule in the masterRule list, which encapsulates the configuration parsing.

However it's a good step to making the rules better configurable! :+1:

jpsim:
Cool, this looks like it's good to merge to me. @nicolai86 feel free to decouple that in a PR, or even just file an issue detailing what you think should be done so it doesn't slip through the cracks.

jpsim:
Great job @scottrhoyt! I know this was a lot of work, and especially dealing with all my nitpicks, but it turned out great :tada:

scottrhoyt:
@nicolai86 I completely agree with you on that and would love to see some more separation of responsibility as well as perhaps bringing back some type safety to configurations. I think this can be part of the work of standardizing how rules can be configured, how to handle documenting rule configurations, and providing the boilerplate for common rule configs like `warning` and `error` levels. I think we can probably go a lot further in extracting more functionality out of `Configuration` as well.

On the flip side, I do think this is a necessary first step to open the door to more advanced configurations. I also think it is a step forward, single responsibility-wise, from the relatively monolithic `Configuration` object that previously parsed YAML, stored the master rule list, configured rules, and does the file system traversals. Also, as a replacement for `ParameterizedRule`, I like that `ConfigurableRule` now has benefit used as a protocol where `ParameterizedRule` wasn't referenced as a protocol but just provided a standard template.

@jpsim no problem! Thanks for your feedback and allowing me to contribute in a big way to a great project.

jakecraige:
This is great. :+1: Thanks for all the work @scottrhoyt

scottrhoyt:
No problem!

norio-nomura:
Great! :tada:

PR #318, opened 2016-01-05
Fixed trailingNewlineCount() returning wrong count

JohnEstropia:
`let start = endIndex.advancedBy(-2, limit: startIndex)` 
basically limited this method to count only up to 2 lines...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@
 
 ##### Bug Fixes
 
-* None.
+* AutoCorrect for TrailingNewlineRule only removes at most one line.
+  [John Estropia](https://github.com/JohnEstropia)
 
 
 ## 0.5.5: Magic Drying Fluff Balls™
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ##### Bug Fixes
 
-* AutoCorrect for TrailingNewlineRule only removes at most one line.
+* AutoCorrect for TrailingNewlineRule only removes at most one line.  
   [John Estropia](https://github.com/JohnEstropia)
 
 
--- a/Source/SwiftLintFramework/Rules/TrailingNewlineRule.swift
+++ b/Source/SwiftLintFramework/Rules/TrailingNewlineRule.swift
@@ -10,11 +10,8 @@ import SourceKittenFramework
 
 extension String {
     private func trailingNewlineCount() -> Int? {
-        let start = endIndex.advancedBy(-2, limit: startIndex)
-        let range = Range(start: start, end: endIndex)
-        let substring = self[range].utf16
         let newLineSet = NSCharacterSet.newlineCharacterSet()
-        let slices = substring.split(allowEmptySlices: true) { !newLineSet.characterIsMember($0) }
+        let slices = self.utf16.split(allowEmptySlices: true) { !newLineSet.characterIsMember($0) }
 
         return slices.last?.count
     }
@@ -34,7 +31,8 @@ public struct TrailingNewlineRule: CorrectableRule {
         ],
         corrections: [
             "let a = 0": "let a = 0\n",
-            "let a = 0\n\n": "let a = 0\n"
+            "let b = 0\n\n": "let b = 0\n",
+            "let c = 0\n\n\n\n": "let c = 0\n"
         ]
     )
 
--- a/Source/SwiftLintFramework/Rules/TrailingNewlineRule.swift
+++ b/Source/SwiftLintFramework/Rules/TrailingNewlineRule.swift
@@ -9,11 +9,19 @@
 import SourceKittenFramework
 
 extension String {
+    private func countOfTrailingCharactersInSet(characterSet: NSCharacterSet) -> Int {
+        var count = 0
+        for char in utf16.lazy.reverse() {
+            if !characterSet.characterIsMember(char) {
+                break
+            }
+            count++
+        }
+        return count
+    }
+    
     private func trailingNewlineCount() -> Int? {
-        let newLineSet = NSCharacterSet.newlineCharacterSet()
-        let slices = self.utf16.split(allowEmptySlices: true) { !newLineSet.characterIsMember($0) }
-
-        return slices.last?.count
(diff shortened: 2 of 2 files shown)


keith:
This behavior was intentional. The only thing that matters for this rule is that there's a single trailing newline. Mainly count == 0 || count > 1. We want to keep the condition this way because splitting the entire file dirastically decreases performance.

keith:
The downside is of course that if you don't fix the error correctly it might be non obvious that you need to decrease it to just a single newline. I think this trade off is worth it.

keith:
Also just to be clear the test case you added should pass without this change correct?

JohnEstropia:
@keith I don't think this was intentional because when I added 
```
"let c = 0\n\n\n\n": "let c = 0\n"
```
to the `corrections` dictionary, the test case started to fail. I also find it weird that with `"\n\n\n\n"`, `trailingNewlineCount()` still returns `2`.

JohnEstropia:
@keith If performance was the reason for this, would it be sufficient to just scan the string for newlines starting at the end without splitting the file?

jpsim:
Thanks for the fix! Although this does fix the correction case when there are more than 2 trailing new lines, as @keith said, this does sacrifice much performance.

Instead, how about using `lazy.reverse().countOfLeadingCharactersInSet(newLineSet)`? With `String.countOfLeadingCharactersInSet(_:)` being defined at https://github.com/jpsim/SourceKitten/blob/master/Source/SourceKittenFramework/String+SourceKitten.swift#L536-L546

JohnEstropia:
@jpsim Yup, same line of thought I had :) I'll update the code

jpsim:
Of course, I'm making up fake code that won't compile in my previous, but this would work:

```swift
extension String {
    private func countOfTrailingCharactersInSet(characterSet: NSCharacterSet) -> Int {
        var count = 0
        for char in utf16.lazy.reverse() {
            if !characterSet.characterIsMember(char) {
                break
            }
            count++
        }
        return count
    }

    private func trailingNewlineCount() -> Int? {
        return countOfTrailingCharactersInSet(NSCharacterSet.newlineCharacterSet())
    }
}
```

jpsim:
@JohnEstropia or @keith could you please benchmark before and after this change? I'd want to make sure we don't regress too much, if at all.

JohnEstropia:
@jpsim
Using `lazy` made it just a wee bit slower than the current implementation.

Old method:
```
Test Case '-[SwiftLintFrameworkTests.ExtendedNSStringTests testPerformance1]' measured [Time, seconds] average: 0.000, relative standard deviation: 177.109%, values: [0.000178, 0.000018, 0.000012, 0.000012, 0.000011, 0.000011, 0.000011, 0.000010, 0.000010, 0.000011], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100
```

New method with lazy:
```
Test Case '-[SwiftLintFrameworkTests.ExtendedNSStringTests testPerformance2]' measured [Time, seconds] average: 0.000, relative standard deviation: 95.958%, values: [0.000102, 0.000021, 0.000018, 0.000018, 0.000017, 0.000018, 0.000017, 0.000017, 0.000017, 0.000017], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100
```

New method without lazy:
```
Test Case '-[SwiftLintFrameworkTests.ExtendedNSStringTests testPerformance3]' measured [Time, seconds] average: 0.000, relative standard deviation: 61.541%, values: [0.000033, 0.000010, 0.000009, 0.000009, 0.000009, 0.000009, 0.000009, 0.000009, 0.000009, 0.000009], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100
```

The non-lazy one seems orders of magnitude faster (probably at the cost of memory for bigger files). But these are all probably negligible numbers so ill stick with the `lazy.reverse()` one.

keith:
I'll benchmark this on our project today as well.

CHANGELOG.md
```diff
@@ -10,7 +10,8 @@
 
 ##### Bug Fixes
 
-* None.
+* AutoCorrect for TrailingNewlineRule only removes at most one line.
```
jpsim: missing 2 trailing spaces here. see https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes

jpsim:
The very high standard deviation values in your benchmarks (177%, 95%, 61%) and short sub-milisecond duration times makes me doubt the benchmarks a bit. To trust these number's we should get the stddev under 10%, which could probably be accomplished just by making more passes with larger strings, e.g. doing more work in the benchmark.

keith:
I built this locally, disabled all rules except this one, and tested it on a portion of our project that has ~350 files. Running `time swiftlint lint` the difference between this branch and master was negligible. Probably still worth the testing @jpsim mentioned.

norio-nomura:
I think this PR has no problems about performance.
Because `TrailingNewlineRule.validateFile`'s cost is 7ms in 26651ms by profiling `swiftlint lint --path <Carthage Repo>` on Instruments as following:
screenshot 2016-01-06 11 52 40

`TrailingNewlineRule` is evaluated only once on every `.swift` files, so that can not cause large performance impact as same as rules that are evaluated on every line in `.swift` files.

CHANGELOG.md
JohnEstropia: @jpsim Fixed!

scottrhoyt:
I tried this out locally as well, and it doesn't appear to have a significant adverse effect on performance.

jpsim:
Thanks everyone for helping to get this in!

PR #325, opened 2016-01-08
Fix regex for ignoring closure parameter on detecting return value

### LINKED ISSUE #324: valid_docs triggers a warning on tuples
Hi,

A warning is shown by the `valid_docs` rule when a function returns a tuple.
For example: 
```swift
	/**
	Awesome test function

	- returns: super tuple
	*/
	func tupleAsReturnValue() -> (r: CGFloat, g: CGFloat, b: CGFloat) {
		return (r: 1, g: 2, b: 3)
	}
```

There is no warning when the tuple is a parameter.

Thanks a lot for your work !
Regards

norio-nomura:
fix #324

### DIFF
--- a/Source/SwiftLintFramework/Rules/ValidDocsRule.swift
+++ b/Source/SwiftLintFramework/Rules/ValidDocsRule.swift
@@ -43,7 +43,7 @@ func delcarationReturns(declaration: String, kind: SwiftDeclarationKind) -> Bool
     if SwiftDeclarationKind.variableKinds().contains(kind) { return true }
 
     let outsideBraces = NSMutableString(string: declaration)
-    regex("(\\s*->\\s*).*[^(]\\)").replaceMatchesInString(outsideBraces, options: [],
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Rules/ValidDocsRule.swift
```diff
@@ -135,6 +135,7 @@ public struct ValidDocsRule: Rule {
             "\n///- parameter param2: this is void too" +
             "\npublic func no(param: (Void -> Void)?, param2: String->Void) {}",
             "/// docs👨‍👩‍👧‍👧\n/// - returns: false\npublic func no() -> Bool { return false }",
+            "/// docs\n/// - returns: tuple\npublic func no() -> (Int, Int) {return (1, 2)}",
```
NachoSoto: :heart_eyes:

jpsim:
Awesome! Great work as always @norio-nomura

norio-nomura:
:pray:

PR #329, opened 2016-01-08
Add autocorrect for CommaRule

nicolai86:
Another tiny addition to make `autocorrect` more useful for existing codebases.


jpsim:
Great work! It's so nice to see autocorrect making its way to more and more rules. It really makes SwiftLint much more useful.

PR #330, opened 2016-01-08
FunctionBodyLengthRule should not count comment lines

### LINKED ISSUE #258: FunctionBodyLengthRule should not count comment lines
lines that start with a "//" after removing all leading whitespaces should not be treated as line inside a function. Otherwise the obvious way to fix violations of that rule would be to write less comments.

marcelofabri:
Fixes #258.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
 
 ##### Enhancements
 
+* FunctionBodyLengthRule now does not count comment lines.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#258](https://github.com/realm/SwiftLint/issues/258)
+
 * `ConfigurableRule` protocol allows for improved rule configuration. See
   `CONTRIBUTING` for more details.  
   [Scott Hoyt](https://github.com/scottrhoyt)
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -91,6 +91,25 @@ extension File {
         }
     }
 
+    public func syntaxKindsByLine(startLine: Int? = nil,
+                                  endLine: Int? = nil) -> [(Int, [SyntaxKind])] {
+        let contents = self.contents as NSString
+        let syntax = syntaxMap
+
+        return lines.flatMap { line in
+            if let startLine = startLine, endLine = endLine
+                where startLine > line.index || endLine < line.index {
+                return nil
+            }
+
+            let kindsInRange = syntax.tokens.filter { token in
+                let tokenLine = contents.lineAndCharacterForByteOffset(token.offset)
+                return tokenLine?.line == line.index
+                }.map({ $0.type }).flatMap(SyntaxKind.init)
+            return (line.index, kindsInRange)
+        }
+    }
+
     //Added by S2dent
     /**
     This function returns only matches that are not contained in a syntax kind
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -105,7 +105,7 @@ extension File {
             let kindsInRange = syntax.tokens.filter { token in
                 let tokenLine = contents.lineAndCharacterForByteOffset(token.offset)
                 return tokenLine?.line == line.index
-                }.map({ $0.type }).flatMap(SyntaxKind.init)
+            }.map({ $0.type }).flatMap(SyntaxKind.init)
             return (line.index, kindsInRange)
         }
     }
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -106,6 +106,12 @@ extension File {
         for kindAndLine in kindsWithLines {
             results[kindAndLine.0] = (results[kindAndLine.0] ?? []) + [kindAndLine.1]
         }
+
+        for line in lines
+            where line.index >= (startLine ?? 0) && line.index <= (endLine ?? Int.max) {
+            results[line.index] = results[line.index] ?? []
+        }
+
         return Array(zip(results.keys, results.values))
     }
 
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -91,7 +91,7 @@ extension File {
         }
     }
 
-    public func syntaxKindsByLine(startLine: Int? = nil,
+    internal func syntaxKindsByLine(startLine: Int? = nil,
                                   endLine: Int? = nil) -> [(Int, [SyntaxKind])] {
         let contents = self.contents as NSString
         let kindsWithLines = syntaxMap.tokens.map { token -> (Int, SyntaxKind) in
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -92,22 +92,19 @@ extension File {
     }
 
     public func syntaxKindsByLine(startLine: Int? = nil,
-                                  endLine: Int? = nil) -> [(Int, [SyntaxKind])] {
+                              endLine: Int? = nil) -> [(Int, [SyntaxKind])] {
         let contents = self.contents as NSString
-        let syntax = syntaxMap
-
-        return lines.flatMap { line in
-            if let startLine = startLine, endLine = endLine
-                where startLine > line.index || endLine < line.index {
-                return nil
-            }
-
-            let kindsInRange = syntax.tokens.filter { token in
-                let tokenLine = contents.lineAndCharacterForByteOffset(token.offset)
-                return tokenLine?.line == line.index
-            }.map({ $0.type }).flatMap(SyntaxKind.init)
-            return (line.index, kindsInRange)
+        let kindsWithLines = syntaxMap.tokens.map { token -> (Int, SyntaxKind) in
+            let tokenLine = contents.lineAndCharacterForByteOffset(token.offset)
+            return (tokenLine!.line, SyntaxKind(rawValue: token.type)!)
+        }.filter { line, token in
+            return line >= startLine && line <= endLine
         }
+        var results = [Int: [SyntaxKind]]()
+        for kindAndLine in kindsWithLines {
+            results[kindAndLine.0] = (results[kindAndLine.0] ?? []) + [kindAndLine.1]
+        }
+        return Array(zip(results.keys, results.values))
     }
 
     //Added by S2dent
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -92,13 +92,15 @@ extension File {
     }
 
     internal func syntaxKindsByLine(startLine: Int? = nil,
-                                  endLine: Int? = nil) -> [(Int, [SyntaxKind])] {
+                                    endLine: Int? = nil) -> [(Int, [SyntaxKind])] {
         let contents = self.contents as NSString
-        let kindsWithLines = syntaxMap.tokens.map { token -> (Int, SyntaxKind) in
+        let kindsWithLines = syntaxMap.tokens.map { token -> (Int, SyntaxToken) in
             let tokenLine = contents.lineAndCharacterForByteOffset(token.offset)
-            return (tokenLine!.line, SyntaxKind(rawValue: token.type)!)
+            return (tokenLine!.line, token)
         }.filter { line, token in
             return line >= (startLine ?? 0) && line <= (endLine ?? Int.max)
+        }.map { (line, token) -> (Int, SyntaxKind) in
+            return (line, SyntaxKind(rawValue: token.type)!)
         }
         var results = [Int: [SyntaxKind]]()
         for kindAndLine in kindsWithLines {
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -92,13 +92,13 @@ extension File {
     }
 
     public func syntaxKindsByLine(startLine: Int? = nil,
-                              endLine: Int? = nil) -> [(Int, [SyntaxKind])] {
+                                  endLine: Int? = nil) -> [(Int, [SyntaxKind])] {
         let contents = self.contents as NSString
         let kindsWithLines = syntaxMap.tokens.map { token -> (Int, SyntaxKind) in
             let tokenLine = contents.lineAndCharacterForByteOffset(token.offset)
             return (tokenLine!.line, SyntaxKind(rawValue: token.type)!)
         }.filter { line, token in
-            return line >= startLine && line <= endLine
+            return line >= (startLine ?? 0) && line <= (endLine ?? Int.max)
         }
         var results = [Int: [SyntaxKind]]()
         for kindAndLine in kindsWithLines {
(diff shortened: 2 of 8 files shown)


Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -90,6 +90,25 @@ extension File {
...
+        return lines.flatMap { line in
+            if let startLine = startLine, endLine = endLine
+                where startLine > line.index || endLine < line.index {
+                return nil
+            }
+
+            let kindsInRange = syntax.tokens.filter { token in
+                let tokenLine = contents.lineAndCharacterForByteOffset(token.offset)
+                return tokenLine?.line == line.index
+                }.map({ $0.type }).flatMap(SyntaxKind.init)
```
jpsim: indentation is off here

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 ##### Enhancements
 
+* FunctionBodyLengthRule now does not count comment lines.
```
jpsim: Seems to be 3 trailing spaces rather than 2.

Source/SwiftLintFrameworkTests/ASTRuleTests.swift
```diff
@@ -148,6 +148,10 @@ class ASTRuleTests: XCTestCase {
             ruleDescription: FunctionBodyLengthRule.description,
             location: Location(file: nil, line: 1, character: 1),
             reason: "Function body should span 40 lines or less: currently spans 41 lines")])
+        let longerFunctionBodyWithComments = "func abc() {" +
+            Repeat(count: 40, repeatedValue: " // this is a comment\n").joinWithSeparator("") +
+        "}\n"
```
jpsim: indentation is off here

jpsim:
Thanks for doing this @marcelofabri, this makes the rule much more useful!

I was a bit worried about the impact this would have on performance, and turns out it's a fairly large one :grimacing:.

As a reference point, the `testSwiftLintLints` integration test takes 2.817s to run on the `master` branch right now.

With this PR, it takes 46.092s :scream:.

By changing the `syntaxKindsByLine` function to this, that goes down to 6.815s:

```swift
public func syntaxKindsByLine(startLine: Int? = nil,
                              endLine: Int? = nil) -> [(Int, [SyntaxKind])] {
    let contents = self.contents as NSString
    let kindsWithLines = syntaxMap.tokens.map { token -> (Int, SyntaxKind) in
        let tokenLine = contents.lineAndCharacterForByteOffset(token.offset)
        return (tokenLine!.line, SyntaxKind(rawValue: token.type)!)
    }.filter { line, token in
        return line >= startLine && line <= endLine
    }
    var results = [Int: [SyntaxKind]]()
    for kindAndLine in kindsWithLines {
        results[kindAndLine.0] = (results[kindAndLine.0] ?? []) + [kindAndLine.1]
    }
    return Array(zip(results.keys, results.values))
}
```

However, I still think that's too much of a performance regression, especially for larger projects.

One thing we could do to minimize the performance impact here would be to only count the number of comment-only lines *after* confirming that the function exceeds the maximum line length when accounting for all lines, which is much faster. For example, in `FunctionBodyLengthRule.validateFile(...)`:

```swift
where endLine - startLine > limit && lineCount(file, startLine: startLine, endLine: endLine) > limit {
```

Actually we could move that to its own function:

```swift
...
where exceedsLineCountExcludingComments(file, startLine, endLine, limit)
...
private func exceedsLineCountExcludingComments(file: File, _ start: Int, _ end: Int,
                                               _ limit: Int) -> Bool {
    return end - start > limit && lineCount(file, startLine: start, endLine: end) > limit
}
```

In this case, SwiftLint goes back to linting in 2.866s, although that's not exactly a great benchmark now because SwiftLint's functions never exceed 40 lines, so the benchmark doesn't hit the slow path.

Could you please apply these changes to help reduce the performance impact? If you find different ways to achieve similar performance gains, I'm certainly open to that too!

marcelofabri:
@jpsim I think that if we called `syntaxKindsByLine` only once (with the full file) we'd have a performance gain. However, I don't see an easy way now to do it. We could hack it by adding an associated object on the `File` object, but we probably should do something better.

Anyway, I have committed the changes you proposed. I was worried about performance too :grimacing: 

I know that it deserves another issue, but we should think about adding performance tests to the project, even linting other projects.

marcelofabri:
By using the associated objects approach, I was able to make it run in ~ 3.401s (without the `exceedsLineCountExcludingComments ` shortcut), against 2.744s with the shortcut.

norio-nomura:
Calling counting line can be reduced by changing position of parameter enumeration.
from:
```swift
// enumerate parameters
for parameter in parameters.reverse() {
    // count lines
    let lineCountExcludingComments =…
    // check violation
    if lineCountExcludingComments > parameter.value {
        …
    }
}
```
to:
```swift
// count lines
let lineCountExcludingComments =…
// enumerate parameters
for parameter in parameters.reverse() {
    // check violation
    if LineCountExcludingComments > parameter.value {
        …
    }
}
```

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -90,6 +90,22 @@ extension File {
         }
     }
 
+    public func syntaxKindsByLine(startLine: Int? = nil,
+                                  endLine: Int? = nil) -> [(Int, [SyntaxKind])] {
+        let contents = self.contents as NSString
+        let kindsWithLines = syntaxMap.tokens.map { token -> (Int, SyntaxKind) in
+            let tokenLine = contents.lineAndCharacterForByteOffset(token.offset)
+            return (tokenLine!.line, SyntaxKind(rawValue: token.type)!)
```
norio-nomura: `SyntaxKind`s of whole the file are created here on every checking functions in the file.
`.filter` should be applied before creation for reducing them

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -90,6 +90,22 @@ extension File {
         }
     }
 
+    public func syntaxKindsByLine(startLine: Int? = nil,
```
norio-nomura: `internal`?

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -90,6 +90,22 @@ extension File {
...
+                                  endLine: Int? = nil) -> [(Int, [SyntaxKind])] {
+        let contents = self.contents as NSString
+        let kindsWithLines = syntaxMap.tokens.map { token -> (Int, SyntaxKind) in
+            let tokenLine = contents.lineAndCharacterForByteOffset(token.offset)
+            return (tokenLine!.line, SyntaxKind(rawValue: token.type)!)
+        }.filter { line, token in
+            return line >= (startLine ?? 0) && line <= (endLine ?? Int.max)
```
norio-nomura: For better performance, tokens should be filtered by range instead of line.

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
jpsim: :+1:

jpsim:
> I know that it deserves another issue, but we should think about adding performance tests to the project, even linting other projects.

How do you recommend we do that? XCTest supports performance tests, but those only really matter when running on a given device. What we really want is to compare with running on the previous version of the test running on the exact same device, although I don't think there's any elegant way to set that up.

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
marcelofabri: Could you expand on this? I'm not sure how I'd do that.

marcelofabri:
@jpsim I haven't thought about it a lot, but I was thinking about XCTest performance tests. I'm still not sure why it wouldn't work. Is it because Travis won't have a baseline to compare to?

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
marcelofabri: This is because currently `.filter` needs the line. If we filtered first, `lineAndCharacterForByteOffset` would be called twice.

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
norio-nomura: Sorry, it does not easy as I thought.
It needs extending SourceKitten for implementing that.

norio-nomura: > It needs extending SourceKitten for implementing that.

I meant that for best performance implementation.

For better performance implementation.
`startLine` and `endLine` calculation is not needed in `validateFile`. It would be better filtering by `bodyOffset` and `bodyLength` as like [here](https://github.com/realm/SwiftLint/blob/master/Source%2FSwiftLintFramework%2FExtensions%2FFile%2BSwiftLint.swift#L85-L88).

Calling `lineAndCharacterForByteOffset` is heavy.
With this PR, the duration of linting KeychainAccess on my Mac using Instruments is 6159ms.
`lineAndCharacterForByteOffset` uses 2867ms on that:
screenshot 2016-01-12 00 44 06
I think we can reduce calls to `lineAndCharacterForByteOffset` by that. :wink:

jpsim:
> @jpsim I haven't thought about it a lot, but I was thinking about XCTest performance tests. I'm still not sure why it wouldn't work. Is it because Travis won't have a baseline to compare to?

XCTest performance tests compare the test output with a previously recorded *baseline* that's associated with a unique device identifier (in this case, the Mac on which these tests are running). This wouldn't work on Travis because they use VMs to run tests, and they're not guaranteed to have the same unique device ID. Even if/when they did, the fact that these are running on shared hardware means that we can't really trust a single run of performance tests.

I'm thinking a different way to measure performance in PRs would be to install the latest version of SwiftLint through homebrew and compare the total time to lint SwiftLint from the latest release and the current PR. Ideally Travis would be able to post back a comment to the PR with the before & after lint times for us to review. Even better would be to have a way to run SwiftLint in a benchmarking mode that logs out the time spent in each rule.

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
jpsim: @marcelofabri it might be better to eagerly create the `SyntaxKind`s by mapping twice:

```swift
let kindsWithLines = syntaxMap.tokens.map { token -> (Int, SyntaxToken) in
    let tokenLine = contents.lineAndCharacterForByteOffset(token.offset)
    return (tokenLine!.line, token)
}.filter { line, token in
    return line >= (startLine ?? 0) && line <= (endLine ?? Int.max)
}.map { (line, token) -> (Int, SyntaxKind) in
    return (line, SyntaxKind(rawValue: token.type)!)
}
```

jpsim:
@marcelofabri this is looking good. What are the next steps (if any) you want to take with this PR?

norio-nomura:
Sorry, my saying too much about performance. :bow:
I also think this is looking good.

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -90,6 +90,22 @@ extension File {
         }
     }
 
+    internal func syntaxKindsByLine(startLine: Int? = nil,
+                                  endLine: Int? = nil) -> [(Int, [SyntaxKind])] {
```
jpsim: indentation is off here

marcelofabri:
@jpsim @norio-nomura I've added a new property on `File+Cache` to cache this and I'm pretty happy with the performance now.

Could you guys please take another look and share your thoughts?

Source/SwiftLintFrameworkTests/ASTRuleTests.swift
```diff
@@ -148,6 +148,10 @@ class ASTRuleTests: XCTestCase {
             ruleDescription: FunctionBodyLengthRule.description,
             location: Location(file: nil, line: 1, character: 1),
             reason: "Function body should span 40 lines or less: currently spans 41 lines")])
+        let longerFunctionBodyWithComments = "func abc() {" +
+            Repeat(count: 40, repeatedValue: " // this is a comment\n").joinWithSeparator("") +
+            "}\n"
+        XCTAssertEqual(violations(longerFunctionBodyWithComments), [])
```
norio-nomura: Sorry, I had overlooked that the tests is not enough.
Maybe it should pass following tests:
```swift
        let longFunctionBodyWithComments = "func abc() {" +
            Repeat(count: 40, repeatedValue: "\n").joinWithSeparator("") +
            "// comment only line should be ignored.\n" +
            "}\n"
        XCTAssertEqual(violations(longFunctionBodyWithComments), [])

        let longerFunctionBodyWithComments = "func abc() {" +
            Repeat(count: 41, repeatedValue: "\n").joinWithSeparator("") +
            "// comment only line should be ignored.\n" +
            "}\n"
        XCTAssertEqual(violations(longerFunctionBodyWithComments), [StyleViolation(
            ruleDescription: FunctionBodyLengthRule.description,
            location: Location(file: nil, line: 1, character: 1),
            reason: "Function body should span 40 lines or less: currently spans 41 lines")])

        let longFunctionBodyWithMultilineComments = "func abc() {" +
            Repeat(count: 40, repeatedValue: "\n").joinWithSeparator("") +
            "/* multi line comment only line should be ignored.\n*/\n" +
            "}\n"
        XCTAssertEqual(violations(longFunctionBodyWithMultilineComments), [])
        
        let longerFunctionBodyWithMultilineComments = "func abc() {" +
            Repeat(count: 41, repeatedValue: "\n").joinWithSeparator("") +
            "/* multi line comment only line should be ignored.\n*/\n" +
            "}\n"
        XCTAssertEqual(violations(longerFunctionBodyWithMultilineComments), [StyleViolation(
            ruleDescription: FunctionBodyLengthRule.description,
            location: Location(file: nil, line: 1, character: 1),
            reason: "Function body should span 40 lines or less: currently spans 41 lines")])
```

marcelofabri: Shouldn't `longerFunctionBodyWithComments`and `longerFunctionBodyWithMultilineComments` *not* trigger any violations?

marcelofabri: Forget that, I didn't realize that the comments were outside `Repeat`. My bad!

marcelofabri: That tests actually caught a bug: the number of lines that the function spans in the reason. Should we update it to subtract the comment only lines? This might be confusing for someone who isn't very familiar with the rule.

Also, if we follow that idea, any ideas on how to return the "fixed"  line count on `exceedsLineCountExcludingComments`? I was thinking about returning a tuple, but I'm not sure if that's the best way.

norio-nomura: > This might be confusing for someone who isn't very familiar with the rule.

I think it would be better `reason` explaining that comment only lines are ignored.

> That tests actually caught a bug

As I tested, SourceKitten returns one token for one multiline comment block. But your code expects that every tokens are in one line. So, counting ignored comment only lines is not correct on multiline comment.

jpsim: We should fill in missing lines in `syntaxKindsByLines`, which will also help avoid counting whitespace-only lines in the function body line count.

marcelofabri: I've committed ec2be78 which should address these issues. As @jpsim (implicitly) suggested, the rule now ignores whitespace-only lines.

Also, I had to separate the tests to another file (`FunctionBodyLengthRuleTests`), because the integration tests were failing.

jpsim:
@marcelofabri @norio-nomura this PR seems fine to me now, other than a rebase, what else needs to be done?

norio-nomura:
Maybe making the reason message clearer by adding explains that comment and whitespace only lines are excluded.
Everyone does not know how the rules counting the lines.

marcelofabri:
I've updated the reason message to

> Function body should span 40 lines or less: currently spans 41 lines (already ignoring comment and whitespace only ones)

What do you guys think?

jpsim:
Maybe this instead?

> Function body should span 40 lines or less excluding comments and whitespace: currently spans 41 lines.

jpsim:
@scotthoyt I think we should disable codecov's automated comments and just have it as a GitHub status check instead. The automated comments are a bit too noisy for my taste.

marcelofabri:
@jpsim Much better! Already updated on 75a4f2f.

jpsim:
:clap: thanks!!!

norio-nomura:
:tada:

marcelofabri:
:100:

scottrhoyt:
@jpsim, will do.

PR #332, opened 2016-01-09
Fix typo in CommaRule.swift

yurrriq:
The description implied the opposite of the rule.


Source/SwiftLintFramework/Rules/CommaRule.swift
```diff
@@ -13,7 +13,7 @@ public struct CommaRule: CorrectableRule {
     public static let description = RuleDescription(
         identifier: "comma",
         name: "Comma Spacing",
-        description: "One space before and no after must be present next to any comma.",
+        description: "One space after and no before must be present next to any comma.",
```
jpsim: If we're discussing wording, this should probably say "No space before and one after...".

yurrriq: Definitely.

PR #333, opened 2016-01-09
Add SourceKit response cache for reducing calls to SourceKit

norio-nomura:
This depends on https://github.com/jpsim/SourceKitten/pull/135
By applying this, the duration of linting Carthage is reduced from:
```
swiftlint lint  27.85s user 1.98s system 77% cpu 38.538 total
```
to:
```
swiftlint lint  21.15s user 1.26s system 80% cpu 27.743 total
```


jpsim:
This is great! Definitely the right thing to do. We should have done this a long time ago.

norio-nomura:
:pray:

PR #335, opened 2016-01-10
Reduce calls to `String.characters.count` in `LineLengthRule.validateFile(_:)`

norio-nomura:
By applying this, the duration of linting Carthage is reduced from:
```
swiftlint lint  20.07s user 1.10s system 84% cpu 25.023 total
```
to:
```
swiftlint lint  18.27s user 1.05s system 84% cpu 22.769 total
```


jpsim:
:+1: you're really on a roll!

PR #193, opened 2015-11-11
Add --use-script-input-files

norio-nomura:
Option to lint files provided by `SCRIPT_INPUT_FILE_*` environment variables.

I created  a Xcode plugin "[ExtraBuildPhase](https://github.com/norio-nomura/ExtraBuildPhase)" that runs SwiftLint on every building without adding "Run Script" build phase to targets in Xcode Project.

I want to resolve a problem that on running SwiftLint by ExtraBuildPhase.
When ExtraBuildPhase runs SwiftLint on each target, SwiftLint does lint all source files under the Project directory. That cause every sources are been lint multiple time as number of targets in Project. :disappointed: 

For resolving this problem, I added functionality to ExtraBuildPhase that provide SourceBuildPhase targets(source files) by `SCRIPT_INPUT_FILE_*` environment variables to shell script for running SwiftLint.
https://github.com/norio-nomura/ExtraBuildPhase/releases/tag/0.3
By applying this PR and using `--use-script-input-files`, SwiftLint does lint only sources that builded by target on Xcode with ExtraBuildPhase.

You can try [swiftlint-use-script-input-files branch of ExtraBuildPhase](https://github.com/norio-nomura/ExtraBuildPhase/tree/swiftlint-use-script-input-files) that enabled using `--use-script-input-files` option by default.

Thanks,

PR #336, opened 2016-01-11
README: added note about --use-script-input-files

### LINKED ISSUE #334: [Feature request] Make autocorrect take an optional list of files
*Writing from my phone, so apologies if this is actually already possible and I didn't see it*.

I was thinking of adding a build phase that calls `swiftlint autocorrect` before actually linting, but I wouldn't want it to modify any file, only those that I've touched (to avoid too many changes but also to speed it up). This could by done by passing the list of "dirty" files on the `git`  checkout.

Other tools like `clang-tidy` can do this on a line-by-line basis too, which would be harder but definitely great!

NachoSoto:
This was introduced in #193, now it's documented! I also added `autocorrect` to the list of available commands in the `README`.

Closes #334.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,14 +56,20 @@ APM.
 $ swiftlint help
 Available commands:
 
-   help      Display general or command-specific help
-   lint      Print lint warnings and errors for the Swift files in the current directory (default command)
-   rules     Display the list of rules and their identifiers
-   version   Display the current version of SwiftLint
+   autocorrect  Automatically correct warnings and errors
+   help         Display general or command-specific help
+   lint         Print lint warnings and errors for the Swift files in the current directory (default command)
+   rules        Display the list of rules and their identifiers
+   version      Display the current version of SwiftLint
 ```
 
 Run `swiftlint` in the directory containing the Swift files to lint. Directories
 will be searched recursively.
+
+If you want to specify a list of files when using `lint` or `autocorrect`
+(like modified files in the working tree based on `git ls-files -m`)
+you can do so by passing the option `--use-script-input-files` and setting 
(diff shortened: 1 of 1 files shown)


NachoSoto:
Actually I should have tested this before writing the docs. It's not completely clear to me what the format of the environment variable is supposed to be. I see that it will read `SCRIPT_INPUT_FILE_COUNT` as well as `SCRIPT_INPUT_FILE_0`, `SCRIPT_INPUT_FILE_1`, ..., `SCRIPT_INPUT_FILE_${SCRIPT_INPUT_FILE_COUNT}`.

Wouldn't it have been easier to make it accept just one environment variable with a space-delimited string of paths? That would make it easier to use by being able to do something like:
```bash
export SCRIPT_INPUT_FILES=$(git ls-files -m)
```

keith:
I'm not sure where this mechanism for sourcing files came from but it seems to be used in more places than just SwiftLint. Carthage implements this as well.

Also at least one possible problem with the space delimited approach would be having to deal with spaces in file paths.

ikesyo:
The mechanism comes from Xcode's "Run Script" build phase and its "Input Files".

jpsim:
:+1: thanks for this @NachoSoto!

jpsim:
Oops, merged this a bit too quickly. Will adapt the docs to better reflect usage.

norio-nomura:
I can't find any official documentation about `SCRIPT_INPUT_FILE_*` from apple, but found that the senior employee of Apple wrote about that in [2006](http://lists.apple.com/archives/Xcode-users/2006/Jan/msg00086.html), Xcode was version 2.2.

PR #337, opened 2016-01-11
update recently added section on script input files in README

jpsim:
to more accurately reflect usage, ExtraBuildPhase plugin and further reading.

/cc @NachoSoto @norio-nomura

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,10 +66,16 @@ Available commands:
 Run `swiftlint` in the directory containing the Swift files to lint. Directories
 will be searched recursively.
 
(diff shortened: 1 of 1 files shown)


NachoSoto:
Perfect 👍
I'll write a small bash command to define these variable based on `git ls-files -m` and I can create another PR documenting it as an example if you'd like.

norio-nomura:
:+1:

jpsim:
@NachoSoto cool! I'm wondering if baking that into SwiftLint might be even better?

NachoSoto:
@jpsim maybe! I'll post here and we shall see :)

PR #339, opened 2016-01-11
escape strings when using CSV reporter

jpsim:
oops

### DIFF
--- a/Source/SwiftLintFramework/Reporters/CSVReporter.swift
+++ b/Source/SwiftLintFramework/Reporters/CSVReporter.swift
@@ -8,6 +8,16 @@
 
 import Foundation
 
+extension String {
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Reporters/CSVReporter.swift
```diff
@@ -8,6 +8,16 @@
 
 import Foundation
 
+extension String {
+    private func escapedForCSV() -> String {
+        let escapedString = stringByReplacingOccurrencesOfString("\"", withString: "\"\"")
```
segiddins: Shouldn't this be `"\\""`?

jpsim: indeed

Issue #53, opened 2015-05-28
Validate the number of functions in a type


jpsim:
Types should contain 10 functions or less.

keith:
Would the functions being `public/internal/private` affect this number? Also would this also ideally include or exclude extensions? In just type definitions ignoring extensions this should be pretty easy to implement.

jpsim:
I don't think ACL would affect this rule, actually, because that only affects usage complexity rather than implementation complexity.

jpsim:
I no longer think we should actually do this. It could easily lead to encouraging poor design patterns.

PR #342, opened 2016-01-12
Add type function count rule 

bkobilansky:
Initial work on #53

Howdy folks, I have a couple of questions. First, I assumed this should be parameterized. What's our take on the default values? 8 for a warning, 10 for an error? 

Adding this also rule caused several lint failures around the codebase. How do we want to deal with that? For now, I disabled the rule for the failing types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ##### Enhancements
 
+* Add TypeFunctionCountRule.  
+  [Brandon Kobilansky](https://github.com/bkobilansky)
+  [#53](https://github.com/realm/SwiftLint/issues/53)
+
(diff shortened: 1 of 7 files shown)


jpsim:
Even though I originally wrote that ticket, I'm not sure this is actually a good idea to enforce.

bkobilansky:
Yeah, seems like the rules for type length and function length are a more flexible and sort of solve this.

If this doesn't provide value, feel free to close the issue and PR. I'm just looking for easy rules to implement. :)

jpsim:
Yeah, let's close this and I'll also close the issue I had originally opened. Sorry for the trouble @bkobilansky, I hope you contribute again in the future!

Issue #319, opened 2016-01-06
Add Rules to Support Ray Wenderlich's Swift Style Guide


glennrfisher:
It would be great if SwiftLint supported [Ray Wenderlich's Swift Style Guide](https://github.com/raywenderlich/swift-style-guide):

- [x] Variable Names: Use camel case.
- [ ] Function Names: Use camel case.
- [x] Class Names: Use upper camel case.
- [ ] Function Arguments: Prefer named parameters, unless the context is very clear.
- [ ] Enumerations: Use upper camel case.
- [ ] Spacing: Indent using 2 spaces. *Should be parametrizable, and probably 4 by default.*
- [x] Opening Braces: Always open on the same line as the statement.
- [ ] Closing Braces: Close on a new line.
- [ ] Spacing Between Functions: Leave exactly one blank line between functions. *This should probably also apply between structs, classes, etc.* *Should be parametrizable.*
- [ ] Access Modifiers: Don't add modifiers such as `internal` when they're already the default.
- [ ] Access Modifiers: Don't repeat the access modifier when overriding a method.
- [ ] Protocol Conformance: When adding protocol conformance to a class, prefer adding a separate class extension for the protocol methods.
- [ ] Computed Properties: If a computed property is read-only, omit the get clause. The get clause is only required when a set clause is provided.
- [ ] Closure Expressions: Use trailing closure syntax only if there's a single closure expression parameter at the end of the argument list.
- [ ] Types: Prefer Swift's native types over Objective-C types (e.g. use `Double` instead of `NSNumber`).
- [x] Constants: Always use `let` instead of `var` if the value of the variable will not change. *Supported by compiler.*
- [ ] Naming Optionals: Avoid names like `optionalString` or `maybeView`, since optionalness is already captured in the variable's type. (Prefer using the same name: `if let subview = subview { ... }`.)
- [ ] Initializing CGGeometry: Use native Swift struct initializers (e.g. `let bounds = CGRect(...)` instead of `let bounds = CGRectMake(...)`).
- [ ] CGGeometry: Prefer struct-scope constants `CGRect.infinite`, `CGRect.null`, etc. over global constants `CGRectInfinite`, `CGRectNull`, etc.
- [ ] Syntactic Sugar: Prefer shortcut versions (`var deviceModels: [String]`) instead of the full generics syntax (`var deviceModels: Array<String>`).
- [ ] For Loops: Prefer the `for-in` style instead of the `for-condition-increment` style.
[...]

PR #344, opened 2016-01-12
add LegacyConstantRule

aamctustwo:
Adds `LegacyConstantRule` from #319 (Add Rules to Support Ray Wenderlich's Swift Style Guide).

CGGeometry: Prefer struct-scope constants CGRect.infinite, CGRect.null,
etc. over global constants CGRectInfinite, CGRectNull, etc.


CHANGELOG.md
```diff
@@ -25,6 +25,10 @@
   [Aaron McTavish](https://github.com/aamctustwo)
   [#202](https://github.com/realm/SwiftLint/issues/202)
 
+* Add LegacyConstantRule.
```
jpsim: missing 2 trailing spaces as described in https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes.

Source/SwiftLintFramework/Rules/LegacyConstantRule.swift
```diff
@@ -0,0 +1,80 @@
+//
+//  LegacyConstantRule.swift
+//  SwiftLint
+//
+//  Created by Aaron McTavish on 12/01/2016.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import SourceKittenFramework
+import CoreGraphics
```
jpsim: I don't think you need this import.

Source/SwiftLintFramework/Rules/LegacyConstantRule.swift
```diff
@@ -0,0 +1,80 @@
...
+            "CGSizeZero": "CGSize.zero",
+            "CGRectNull": "CGRect.null"
+        ]
+
+        let description = self.dynamicType.description
+        var corrections = [Correction]()
+        var contents = file.contents
+
+        for (pattern, template) in patterns {
+            let matches = file.matchPattern(pattern, excludingSyntaxKinds: [.Comment])
```
jpsim: Why not filter on `.Identifier` like you're doing in `validateFile`? As it stands, the string references above would get corrected and that's certainly not what we want.

And here you're scanning the entire file for each pattern, whereas `validateFile` just does it once for all patterns.

jpsim:
Thanks for this @aamctustwo! Merging in #357.

Issue #321, opened 2016-01-06
Enforce explicit self


czechboy0:
Due to the Swift evolution proposal 009 being rejected, but still having a pretty big following, I think it'd be very valuable for certain teams to be able to enforce explicit self (like in Objective-C). [Here's](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20160104/005478.html) the reasoning for not changing the language from the core team and [here's](https://github.com/apple/swift-evolution/blob/master/proposals/0009-require-self-for-accessing-instance-members.md) the proposal itself, where you can find reasons why the proposed change makes sense for some teams.

The gist of the decision was that explicit self makes sense for a certain group of people and those should use a lint tool (suggested by the core Swift team) instead of the language being changed. Thus I think SwiftLint supporting this rule would help it to be used by all those devs.

NachoSoto:
I'd love this, even if it's not enabled by default.

czechboy0:
Yeah I'm assuming people would cherry-pick it into their SwiftLint config file only if they wanted it enforced in their project. Unfortunately I'm not yet familiar enough with SwiftLint myself so I'm hoping for someone experienced to at least chime in about how difficult it'd be to add. (Also I'm studying for exams atm, so this time right here officially counts as procrastination.)

keith:
We also have this in our style guide and would love to see this in swiftlint. I'll look into an approach for this today.

jpsim:
I'd love to see this as an opt-in rule.

czechboy0:
Are opt-in rules already supported by SwiftLint?

keith:
No they're not. That's something that's come up multiple times before and
We have been meaning to add. Pull requests welcome!

czechboy0:
Understood. Does that mean that this issue will be blocked until opt-in rules are implemented? I'd be happy to use the rule from a feature branch in the meantime if so.

keith:
It definitely would be. We wouldn't be able to ship this on by default.

jpsim:
It would be trivial to introduce opt-in rules.

1. Introduce an `OptInType` protocol to tag opt-in rules
2. Make opt-in rules conform to `OptInType`
3. In [Configuration's initializer](https://github.com/realm/SwiftLint/blob/0.5.5/Source/SwiftLintFramework/Models/Configuration.swift#L91-L93), filter out opt-in rules that haven't been explicitly added to `enabled_rules` in the YAML file

AliSoftware:
[...]

PR #345, opened 2016-01-12
support opt-in rules

### LINKED ISSUE #256: enabled_rules option in .swiftlint.yml?
- It would be helpful to have an `enabled_rules` option in the `.swiftlint.yml` config file that works as a whitelist. 
- I'm using swiftlint as part of a CI process, so by using the `disabled_rules` option, each time I upgrade the swift lint binary, and there have been new rules added, I have to disable more rules (otherwise swiftlint will generate warnings / errors for rules I don't necessarily want).

jpsim:
Closes #256 /cc @keith 

We could rules conform to this that are either not universally accepted, or applicable (e.g. #245 & #321) and rules that have too high of a false positive rate (e.g. #201 & #206).

### DIFF
--- a/Source/SwiftLintFramework/Models/Configuration.swift
+++ b/Source/SwiftLintFramework/Models/Configuration.swift
@@ -37,6 +37,7 @@ public struct Configuration: Equatable {
     }
 
     public init?(disabledRules: [String] = [],
+                 enabledRules: [String] = [],
                  included: [String] = [],
                  excluded: [String] = [],
                  reporter: String = "xcode",
@@ -48,7 +49,6 @@ public struct Configuration: Equatable {
         self.useNestedConfigs = useNestedConfigs
 
         // Validate that all rule identifiers map to a defined rule
-
         let validRuleIdentifiers = Configuration.rulesFromDict().map {
             $0.dynamicType.description.identifier
         }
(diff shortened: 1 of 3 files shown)


keith:
Awesome! :+1:

daniel-beard:
This is great! So `enabled_rules:` in the `yml` file acts as a whitelist?

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -78,14 +76,17 @@ public struct Configuration: Equatable {
         }
         self.disabledRules = validDisabledRules
 
-        self.rules = rules.filter {
-            !validDisabledRules.contains($0.dynamicType.description.identifier)
+        self.rules = rules.filter { rule in
+            let id = rule.dynamicType.description.identifier
+            if validDisabledRules.contains(id) { return false }
+            return enabledRules.contains(id) || !(rule is OptInRule)
```
daniel-beard: I'd prefer not to have the `OptInRule` or possibly a different mechanism for it. It makes `enablesRules` only useful for opting in to non-default rules, rather a definitive whitelist.

jpsim: I see. That's not what I was going for, and not what I outlined in https://github.com/realm/SwiftLint/issues/256#issuecomment-163393662.

What you want is a definitive list of rules to run, ignoring `enabled_rules` and `disabled_rules`, which now reading back that comment I linked to, was the motivation for a `rules` config option.

I see value in that, but it does have to be weighed against the complexity of having a white list, a black list, and a list list :disappointed: .

daniel-beard: I get it, although I got a different meaning from @scottrhoyt 's comment which seems to imply either a white list or black list, but not both (or opt in).

PR #328, opened 2016-01-08
Update Commandant to 0.8.2

ikesyo:
https://github.com/realm/SwiftLint/pull/312#issuecomment-168558053 is fixed. :zap:

Issue #340, opened 2016-01-12
`ConditionalBindingCascadeRule` hangs up on linting KeychainAccess


norio-nomura:
It can be reproduced by
`curl https://raw.githubusercontent.com/kishikawakatsumi/KeychainAccess/master/Examples/Playground-iOS.playground/section-1.swift |swiftlint lint --use-stdin`

norio-nomura:
SwiftLint hangs up in `matchesInString(_:options:range:)`.
Maybe it needs to change `ConditionalBindingCascadeRule` from `Rule` to `ASTRule` for supporting the triggering test case:
```swift
"if let a = b, c = d.indexOf({$0 == e}), let f = g {",
```
and avoiding infinite loop in `matchesInString`.

jpsim:
Yes, this is a major problem. I'm looking into it now.

norio-nomura:
@jpsim On re-writing, It also needs to consider #347

PR #346, opened 2016-01-13
Fix Homebrew bottling error

norio-nomura:
`brew install --verbose --build-bottle swiftlint` failed by the error on https://github.com/Homebrew/homebrew/pull/47688

See also: https://github.com/realm/SwiftLint/pull/328#issuecomment-170214980

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ TEMPORARY_FOLDER?=/tmp/SwiftLint.dst
 PREFIX?=/usr/local
 BUILD_TOOL?=xcodebuild
 
-XCODEFLAGS=-workspace 'SwiftLint.xcworkspace' -scheme 'swiftlint' DSTROOT=$(TEMPORARY_FOLDER)
+XCODEFLAGS=-xcconfig settings-for-all-projects.xcconfig -workspace 'SwiftLint.xcworkspace' -scheme 'swiftlint' DSTROOT=$(TEMPORARY_FOLDER)
 
 BUILT_BUNDLE=$(TEMPORARY_FOLDER)/Applications/swiftlint.app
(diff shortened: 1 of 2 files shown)


jpsim:
I 100% don't understand this, and defer to you as the expert in these matters! Would this be necessary for other Commandant projects like SourceKitten and Carthage?

jpsim:
Nevermind, I understand this a lot better now after reading https://github.com/Homebrew/homebrew/issues/20233 :smile:. HUGE thanks for the fix @norio-nomura! I hope to cut a SwiftLint release soon so we'll see if it works soon enough.

norio-nomura:
I think the error of `install_name_tool` is not predictable one.
On this time, it happened with Commandant. But on next, it may happens with another frameworks.

By applying this, installed image size of SwiftLint will increase from 15556Kbytes to 15728Kbytes.
```sh
% du -d 1 -k /usr/local/SwiftLint/
15728	/usr/local/SwiftLint//with346
15556	/usr/local/SwiftLint//without346
```
That is the cost for preventing future fails.

:-)

Above was written before receiving https://github.com/realm/SwiftLint/pull/346#issuecomment-171123760
Thanks for understanding :smile: 

BTW, I think #340 is show stopper. I can't complete linting KeychainAccess (my lightest benchmark) since 0d1c3d2

jpsim:
> By applying this, installed image size of SwiftLint will increase from 15556Kbytes to 15728Kbytes.

That's quite acceptable I think.

> BTW, I think #340 is show stopper. I can't complete linting KeychainAccess (my lightest benchmark) since 0d1c3d2

Oh right, I had forgotten about that. Let's take the conversation there.

Issue #347, opened 2016-01-13
`ConditionalBindingCascadeRule` does not check `where` clause­


norio-nomura:
`let` is needed after `where` clause­.
So, `ConditionalBindingCascadeRule` should not trigger on following example:
```swift
"guard let a = b where a, let c = d {"
```

jpsim:
Addressed in #348.

PR #348, opened 2016-01-13
fix infinite regex search & 'where' false positive in ConditionalBindingCascadeRule

### LINKED ISSUE #340: `ConditionalBindingCascadeRule` hangs up on linting KeychainAccess

jpsim:
Fixes #340 and #347 /cc @norio-nomura

### DIFF
--- a/Source/SwiftLintFramework/Rules/ConditionalBindingCascadeRule.swift
+++ b/Source/SwiftLintFramework/Rules/ConditionalBindingCascadeRule.swift
@@ -22,7 +22,8 @@ public struct ConditionalBindingCascadeRule: Rule {
             "if let a = b, \n c = d \n {",
             "if let a = b { if let c = d {",
             "if let a = b { let c = d({ foo in ... })",
-            "guard let a = b, c = d else {"
+            "guard let a = b, c = d else {",
+            "guard let a = b where a, let c = d else {"
         ],
         triggeringExamples: [
             "if let a = b, let c = d {",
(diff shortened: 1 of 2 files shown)


jpsim:
A small performance win too!

* `testSwiftLintLints` in master: 2.246 seconds
* `testSwiftLintLints` in this branch: 2.061 seconds

Source/SwiftLintFramework/Rules/ConditionalBindingCascadeRule.swift
```diff
@@ -36,12 +37,10 @@ public struct ConditionalBindingCascadeRule: Rule {
...
     public func validateFile(file: File) -> [StyleViolation] {
-        let conditionalBindingKeywords = ["if", "guard"]
-        let pattern =  "^(" +
-                        conditionalBindingKeywords.joinWithSeparator("|") +
-                        ")(\\s*?)let((.|\\s)*?),(\\s*?)let((.|\\s)*?)\\{"
-        return file.matchPattern(pattern,
-            excludingSyntaxKinds: SyntaxKind.commentAndStringKinds()).map {
+        return file.matchPattern("^(if|guard)(.*?)let(.*?),(.*?)let(.*?)\\{",
+            excludingSyntaxKinds: SyntaxKind.commentAndStringKinds()).filter {
+                !(file.contents as NSString).substringWithRange($0).containsString("where")
```
jpsim: I'm sure this could lead to some missed violations, but I'm not too concerned with that right now.

jpsim:
Merging because it addresses a critical issue. @norio-nomura please let me know if you can think of a better way to handle `where` cases.

norio-nomura:
That's enough now! :+1:

PR #354, opened 2016-01-13
ParameterizedRule Migration

### LINKED ISSUE #353: ParameterizedRule Migration
Migrate `ParameterizedRule`s to `ConfigurableRule`s.

@realm/swiftlint feel free to collect your thoughts here for the time being. I'll open a WIP PR soon.

scottrhoyt:
Closes #353

### DIFF
--- a/Source/SwiftLintFramework/Protocols/Rule.swift
+++ b/Source/SwiftLintFramework/Protocols/Rule.swift
@@ -32,6 +32,7 @@ public protocol ConfigurableRule: Rule {
     func isEqualTo(rule: ConfigurableRule) -> Bool
 }
 
+@available(*, deprecated=0.5.6, message="Use ConfigurableRule instead.")
 public protocol ParameterizedRule: ConfigurableRule {
     typealias ParameterType: Equatable
     init(parameters: [RuleParameter<ParameterType>])
--- a/Source/SwiftLintFramework/Protocols/Rule.swift
+++ b/Source/SwiftLintFramework/Protocols/Rule.swift
@@ -37,30 +37,6 @@ public protocol ViolationLevelRule: ConfigurableRule {
     var error: RuleParameter<Int> { get set }
 }
 
-@available(*, deprecated=0.5.6, message="Use ConfigurableRule instead.")
-public protocol ParameterizedRule: ConfigurableRule {
-    typealias ParameterType: Equatable
-    init(parameters: [RuleParameter<ParameterType>])
-    var parameters: [RuleParameter<ParameterType>] { get }
-}
-
-// Default implementation for ConfigurableRule conformance
-extension ParameterizedRule {
-    public init?(config: AnyObject) {
-        guard let array = [ParameterType].arrayOf(config) else {
-            return nil
-        }
-        self.init(parameters: RuleParameter<ParameterType>.ruleParametersFromArray(array))
-    }
-
-    public func isEqualTo(rule: ConfigurableRule) -> Bool {
-        if let rule = rule as? Self {
-            return parameters == rule.parameters
-        }
-        return false
-    }
-}
-
 public protocol CorrectableRule: Rule {
     func correctFile(file: File) -> [Correction]
 }
--- a/Source/SwiftLintFramework/Protocols/Rule.swift
+++ b/Source/SwiftLintFramework/Protocols/Rule.swift
@@ -46,7 +46,7 @@ public protocol CorrectableRule: Rule {
 public extension ViolationLevelRule {
     public init?(config: AnyObject) {
         self.init()
-        if let config = [Int].arrayOf(config) where config.count > 0 {
+        if let config = [Int].arrayOf(config) where !config.isEmpty {
             warning = RuleParameter(severity: .Warning, value: config[0])
             if config.count > 1 {
                 error = RuleParameter(severity: .Error, value: config[1])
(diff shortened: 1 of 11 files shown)


scottrhoyt:
@jpsim I took my first stab at migrating `ParameterizedRule`s to `ConfigurableRules` using a new protocol `ViolationLevelRule` that inherits from `ConfigurableRule`. I actually really like how it has improved the readability of the rule code and reduced boilerplate, but let me know what you think.

Source/SwiftLintFramework/Protocols/Rule.swift
```diff
@@ -32,6 +32,12 @@ public protocol ConfigurableRule: Rule {
     func isEqualTo(rule: ConfigurableRule) -> Bool
 }
 
+public protocol ViolationLevelRule: ConfigurableRule {
+    var warning: RuleParameter<Int> { get set }
+    var error: RuleParameter<Int> { get set }
+}
+
+@available(*, deprecated=0.5.6, message="Use ConfigurableRule instead.")
```
jpsim: lol, it's cute that you want to deprecate this "properly", but you should really just remove `ParameterizedRule` entirely.

We've removed things without prior deprecation notice in SwiftLint before, and I'm comfortable continuing to do that freely until SwiftLint hits 1.0.

A cursory search of GitHub seems to indicate that no one outside this project is using the protocol: https://github.com/search?l=swift&q=ParameterizedRule&ref=simplesearch&type=Code&utf8=%E2%9C%93

Source/SwiftLintFramework/Protocols/Rule.swift
```diff
@@ -59,6 +65,37 @@ public protocol CorrectableRule: Rule {
     func correctFile(file: File) -> [Correction]
 }
 
+// MARK: - ViolationLevelRule conformance to ConfigurableRule
+
+public extension ViolationLevelRule {
+    public init?(config: AnyObject) {
+        self.init()
+        if let config = [Int].arrayOf(config) where config.count > 0 {
```
jpsim: `where !config.isEmpty`

jpsim:
This is a fine approach, I just had a few minor comments, but otherwise you should continue down that direction.

Please note that this is the last PR blocking a 0.6.0 release :smile:

Source/SwiftLintFramework/Protocols/Rule.swift
scottrhoyt: Got it. It was more for myself to practice using that attribute and flag warnings for visibility. Planning on removing.

scottrhoyt:
@jpsim check out the last couple of changes. If nothing else is needed, ready to merge.

scottrhoyt:
P.S. I love seeing more deletions than additions while maintaining functionality. :+1:

Source/SwiftLintFramework/Protocols/Rule.swift
```diff
@@ -32,33 +32,46 @@ public protocol ConfigurableRule: Rule {
...
+            if let warningNumber = config["warning"] as? Int {
+                warning = RuleParameter(severity: .Warning, value: warningNumber)
+            }
+            if let errorNumber = config["error"] as? Int {
+                error = RuleParameter(severity: .Error, value: errorNumber)
+            }
+        } else {
             return nil
         }
-        self.init(parameters: RuleParameter<ParameterType>.ruleParametersFromArray(array))
```
jpsim: I don't think `ruleParametersFromArray` is used anymore at all, can you please remove it?

scottrhoyt: Yep.

Source/SwiftLintFramework/Protocols/Rule.swift
```diff
@@ -32,33 +32,46 @@ public protocol ConfigurableRule: Rule {
     func isEqualTo(rule: ConfigurableRule) -> Bool
 }
 
-public protocol ParameterizedRule: ConfigurableRule {
-    typealias ParameterType: Equatable
-    init(parameters: [RuleParameter<ParameterType>])
-    var parameters: [RuleParameter<ParameterType>] { get }
+public protocol ViolationLevelRule: ConfigurableRule {
+    var warning: RuleParameter<Int> { get set }
```
jpsim: do you need the `set`s here? If you move `self.init()` from the `ViolationLevelRule` extension to happen after setting the `warning` and `error` properties, I don't think you'll need the `set`s here.

scottrhoyt:
So what is odd is that as soon as I removed `ruleParametersFromArray`, I got compiler errors thrown off by `MissingDocsRule` which not only uses this method, but is a `ParameterizedRule`!!! I have no idea how this compiled on my machine or the CI server after I removed `ParameterizedRule`. It's very weird. Anyway, moved the PR back to WIP because I need to think a bit on how to best deal with `RuleParameter`s with `ParameterTypes` different than `Int`.

Source/SwiftLintFramework/Protocols/Rule.swift
scottrhoyt: I think you do. `init?(config:)` calls `init()` first to load the default values for a `Rule` before overwriting those values with something that might be in the `config`. This is important to accommodate configurations that aren't fully specified (e.g. `variable_name_min_length: 3`). Because it is a delegating initializer, this needs to be done before you access `self`. The only way around this would be to make `init?(config:)` non-delegating. Which means we can't support both configured values and default values at the same time, it becomes an either/or. At least that is my thinking on it. Could be wrong.

jpsim: you're right, let's keep as-is then.

scottrhoyt:
The problem is solved for the time being by making `MissingRulesDoc` a `ConfigurableRule` straight up and keeping most of the implementation the same. Not super happy with that, but it gets us moving forward.

jpsim:
Great work! :clap:

PR #366, opened 2016-01-14
Add Codecov

### LINKED ISSUE #364: Add CodeCov support.
Hey @jpsim. How do you feel about adding [CodeCov](https://codecov.io/) support to display code coverage metrics on the README and potentially provide another check on how PRs affect test coverage? I've done the travis integration before if you want me to take a look at it.

scottrhoyt:
Closes #364 

@jpsim I'm pretty sure this will do it. We might need to fiddle with the settings [here](https://codecov.io/github/realm/SwiftLint) to get it exactly how we want. You can change stuff like what info is put into PR comments, commit statuses, and such. Also, I don't think it will be needed, but you can give Codecov the identity of a team member to make those API calls. Check it out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,5 @@ script: script/cibuild | xcpretty
 notifications:
   email: false
   slack: realmio:vPdpsG9NLDo2DNlbqtcMAQuE
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ SwiftLint hooks into [Clang](http://clang.llvm.org) and
 of your source files for more accurate results.
 
 ![Test Status](https://travis-ci.org/realm/SwiftLint.svg?branch=master)
+[![codecov.io](https://codecov.io/github/realm/SwiftLint/coverage.svg?branch=master)](https://codecov.io/github/realm/SwiftLint?branch=master)
 
 ![](screenshot.png)
 


scottrhoyt:
You can also enable Codecov as a required check for a PR merge.

jpsim:
> You can also enable Codecov as a required check for a PR merge.

I think we should do this. Can you set that up please?

scottrhoyt:
Will do as soon as this gets merged into `master`. It's a GitHub setting and it will need to see the reports being available for `master` before being an option. In addition, you can disable push access to `master` so that code only gets in through a PR.

jpsim:
> Will do as soon as this gets merged into master. It's a GitHub setting and it will need to see the reports being available for master before being an option.

Awesome! Merging.

> In addition, you can disable push access to master so that code only gets in through a PR.

No actually, GitHub just has a setting to disable *force* pushing to `master`. :cry:

scottrhoyt:
Ah, got it. I don't have access to change any repo settings. However, this is where you can play around with it:

screen shot 2016-01-15 at 8 15 46 am

screen shot 2016-01-15 at 8 16 05 am

jpsim:
Oh cool, so you were right! By requiring status checks, this prevents even regular pushes from happening on `master`. Here are the branch protection settings now:

image

What's the difference between the `codecov/patch` and `codecov/project` status checks?

scottrhoyt:
The project check is a success when the PR results in the same or higher coverage. It is a failure otherwise.

The patch check looks at just the changed lines of code and calculates their coverage. This is then compared against a threshold (currently set to 50% [here](https://codecov.io/github/realm/SwiftLint/features/status)) and passes if >=, fails otherwise.

jpsim:
Thanks for explaining that to me, it makes sense. Let's keep both status checks enabled, but not required, as it might occasionally be appropriate to merge PRs that actually reduce test coverage.

Unrelated, were you successful in disabling Codecov's automatic commenting?

scottrhoyt:
Project is strict, and I usually go with Patch.

However, there are times that Patch can be misleading. For example, if I replace these untested lines of code:

```swift
let a = [1, 2]
var sum = 0
for num in a {
  sum += num
}
```

with this untested line:

```swift
let sum = [1, 2].reduce(0) { + }
```

... the project coverage will increase, but the patch coverage will be 0%.

scottrhoyt:
I altered the verbosity of the comment to look like:

screen shot 2016-01-15 at 10 19 35 am

... to see if that was acceptable. However, I can disable it entirely if that is still too much clutter.

jpsim:
> Project is strict, and I usually go with Patch.

I still think we should keep both status checks enabled, but not required.

> However, I can disable it entirely if that is still too much clutter.

Yes, please just disable entirely, this information will still be available via GitHub's status checks and I think that's sufficient.

scottrhoyt:
:+1:

PR #367, opened 2016-01-14
Fixing a minor typo

seanhussey:
At least, I think it's a typo. I would probably want to exclude a file names ```EcludedFile.swift```, too.


jpsim:
Thanks!

PR #374, opened 2016-01-15
Added a force unwrapping rule

Argent:
The rule will create a warning for all force unwrappings of swift optionals


scottrhoyt:
Thanks for this PR! This might be interesting as one of the new `OptInRule`s instead of being disabled by default. I think the fork you've been working on is at least a week old, and there have been some big changes introduced. You might want to rebase the changes onto the current master. Do you know if this will also flag implicitly unwrapped optionals?

Argent:
Thanks for the quick reply. An OptInRule makes sense since this one is a rather strict rule. I will rebase the branch to check for the new features.
No, it will not be triggered by implicitly unwrapped optionals because the rule only checks `.Identifier` tokens and not `.TypeIdentifier`

jpsim:
Nice work @Argent! I agree with @scottrhoyt that this would make the most sense as an opt-in rule.

Once you've made that change, rebased (there are merge conflicts currently), and added a changelog entry, this will be good to go!

norio-nomura:
I feel that it produces too many false positive.
screenshot 2016-01-18 16 23 21

Argent:
@jpsim @scottrhoyt Thanks for the suggestion about the `OptInRule`.

I rebased the branch and changed the rule to an `OptInRule`

@norio-nomura The default matching regex options were changed to `.DotMatchesLineSeparators` after I initially created my code, which caused all of these false positives. I changed the regex and it is now working again

norio-nomura:
I tested with updated version. Less false positive and better performance.
I found only one false positive on linting SwiftLint work tree:
screenshot 2016-01-18 19 28 15

Some locations are incorrect.
Nesting parentheses case:
screenshot 2016-01-18 19 21 06

Matches in substring case:
screenshot 2016-01-18 19 12 33

Personally, it is good enough.

While browsing the violations in Xcode, I feel that it would be better if  location points `!`.
Because sometimes identifier is far from `!`.

Argent:
@norio-nomura Thanks for your input. I think I came up with a (rather complicated) expression that gets rid of the problem of unintentionally tagging some explicitly unwrapped optionals.
I also changed the location of the violation to point to the `!`

I added the changes in a new commit so you can see the differences (can be squashed later)

norio-nomura:
It seems 49866b3 has performance issue.
swiftlint hangs up by `curl https://raw.githubusercontent.com/thoughtbot/Curry/v1.3.0/Source/Curry.swift | swiftlint lint --use-stdin --config ~/.swiftlint-test.yml`.

`.swiftlint-test.yml`:
```yml
enabled_rules:
  - empty_count
  - missing_docs
  - force_unwrapping
```

Argent:
@norio-nomura Yes you are right. I guess the expression is too complex

norio-nomura:
It also produces duplicated violations:
screenshot 2016-01-18 23 04 30
That is minor issue than the performance issue.

Argent:
@norio-nomura I reverted the regex to the previous state, it was way to complex for this task. I also added tests for the rule

norio-nomura:
Sorry, I didn't notice that test did not run.

I tested 55f33ef. It is a little bit slow, but good.

> it was way to complex for this task

I agree. If avoiding false positive that I mentioned, It will need to rewriting to `ASTRule`.
screenshot 2016-01-20 00 33 28

I think it is reasonable as regex based rule.

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#258](https://github.com/realm/SwiftLint/issues/258)
 
+* `ForceUnwrappingRule` was added as an `OptInRule` to issue warnings for all forced unwrappings
```
jpsim: Please adjust the formatting of your changelog entry to match https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes.

* Period at the end
* Two trailing spaces
* Include the issue number, not the PR number

Source/SwiftLintFramework/Rules/ForceUnwrappingRule.swift
```diff
@@ -0,0 +1,49 @@
...
+
+    public func validateFile(file: File) -> [StyleViolation] {
+        // Get all identifiers from syntax map
+        let identifiers = file.syntaxMap.tokens.filter({
+                $0.type == SyntaxKind.Identifier.rawValue
+            }).flatMap({
+                file.contents.substringWithByteRange(start: $0.offset, length: $0.length)
+            })
+        // Check if there is any identifier followed by a '!'
+        let violations = Set(identifiers).flatMap({
```
jpsim: This is going to be terribly inefficient, since you're scanning the entire file for a regex match for each identifier in the file.

norio-nomura: FYI, the durations of linting Carthage on my Mac,
cc41abb:
```
swiftlint lint --config ~/.swiftlint-test.yml  21.63s user 1.33s system 83% cpu 27.463 total
```
55f33ef:
```
swiftlint lint --config ~/.swiftlint-test.yml  27.60s user 1.56s system 84% cpu 34.440 total
```

jpsim:
Continuing in #421.

PR #377, opened 2016-01-16
TypeBodyLengthRule should not count comment and whitespace lines

### LINKED ISSUE #369: TypeBodyLengthRule should not count comment and whitespace lines
If FunctionBodyLengthRule doesn't (#258 and #330), TypeBodyLengthRule shouldn't as well.

marcelofabri:
Fixes #369.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ##### Enhancements
 
+* TypeBodyLengthRule now does not count comment lines.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#369](https://github.com/realm/SwiftLint/issues/369)
+
 * FunctionBodyLengthRule now does not count comment lines.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#258](https://github.com/realm/SwiftLint/issues/258)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,11 @@
 
 ##### Enhancements
 
-* TypeBodyLengthRule now does not count comment lines.  
+* TypeBodyLengthRule now does not count comment or whitespace lines.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#369](https://github.com/realm/SwiftLint/issues/369)
 
-* FunctionBodyLengthRule now does not count comment lines.  
+* FunctionBodyLengthRule now does not count comment or whitespace lines.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#258](https://github.com/realm/SwiftLint/issues/258)
 
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -181,4 +181,31 @@ extension File {
         }
         return violatingRanges
     }
+
+    internal func numberOfCommentOnlyLines(startLine: Int, endLine: Int) -> Int {
+        let commentKinds = Set(SyntaxKind.commentKinds())
+
+        return syntaxKindsByLines.filter { line, kinds -> Bool in
+            guard line >= startLine && line <= endLine else {
+                return false
+            }
+
+            return kinds.filter { !commentKinds.contains($0) }.isEmpty
+        }.count
+    }
+
+    internal func lineCount(startLine: Int, endLine: Int) -> Int {
+        let commentedLines = numberOfCommentOnlyLines(startLine, endLine: endLine)
+        return endLine - startLine - commentedLines
+    }
+
+    internal func exceedsLineCountExcludingComments(start: Int, _ end: Int,
+                                                    _ limit: Int) -> (Bool, Int) {
+        if end - start <= limit {
+            return (false, end - start)
+        }
+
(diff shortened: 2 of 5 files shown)


norio-nomura:
:+1: other than change log entry

marcelofabri:
Fixed

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -181,4 +181,31 @@ extension File {
         }
         return violatingRanges
     }
+
+    internal func numberOfCommentOnlyLines(startLine: Int, endLine: Int) -> Int {
```
jpsim: I thought these functions also excluded whitespace only lines? It doesn't seem like they do. I'm not sure what I'm missing.

marcelofabri: They do, but probably they name could be changed to reflect that.

`syntaxKindsByLine()` adds the whitespace only lines with an empty `SyntaxKind` array. Then `numberOfCommentOnlyLines` keeps these lines because of
```swift
kinds.filter { !commentKinds.contains($0) }.isEmpty
```
as `kinds` will be empty, `.filter` won't have any effect, keeping the line on the resulting array of the outer `.filter`.

jpsim: Right, that line could use a short comment explaining that, and these functions should be renamed accordingly.

marcelofabri:
Renamed the functions in 1ad7e76

jpsim:
:+1:

marcelofabri:
I've rebased this with master. Is there anything else that needs to be done to get this merged?

norio-nomura:
I think it would be better if it follows the 04267a4's example.

norio-nomura:
You can check compilation time by `make display_compilation_time` as following:
```
% make display_compilation_time
…
5089.2ms	…/ASTRuleTests.swift:139:10	@objc func testTypeBodyLengths()
```

marcelofabri:
OK, I'l try to refactor that later today

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 ##### Enhancements
 
+* TypeBodyLengthRule now does not count comment lines.
```
jpsim: "comment or whitespace lines"? Same with the one for `FunctionBodyLengthRule` below.

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -181,4 +181,32 @@ extension File {
...
+        return endLine - startLine - ignoredLines
+    }
+
+    internal func exceedsLineCountExcludingCommentsAndWhitespace(start: Int, _ end: Int,
+                                                                 _ limit: Int) -> (Bool, Int) {
+        if end - start <= limit {
+            return (false, end - start)
+        }
+
+        let count = lineCount(start, endLine: end)
```
jpsim: Removing the `lineCount()` function and having the following inline would help improve code locality:

`let count = end - start - numberOfCommentAndWhitespaceOnlyLines(start, endLine: end)`

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -181,4 +181,32 @@ extension File {
         }
         return violatingRanges
     }
+
+    internal func numberOfCommentAndWhitespaceOnlyLines(startLine: Int, endLine: Int) -> Int {
```
jpsim: `private`?

marcelofabri:
:sparkles:

norio-nomura:
:+1:

jpsim:
:+1: nicely done!

norio-nomura:
:tada:

jpsim:
Hmm, I just noticed that the tests now take much longer to run than before, and it's because `testTypeBodyLengths` alone takes ~30 seconds :grimacing:. See https://travis-ci.org/realm/SwiftLint/builds/104029577#L628-L706

@marcelofabri could you please look a bit more into this to see if the tests are doing tons of unnecessary work, or if it's the rule that's quite slow?

norio-nomura:
@jpsim I did notice about the test issue and investigated.
That caused by the test matches to worst case of using `SourceKitten.NSString.CacheContainer`.
On the investigating process, I found some optimization points that can be reduce the test duration.
Maybe I can file the PR for that today.

norio-nomura:
Sorry, it seems #399 did not reduce the duration of test as https://travis-ci.org/realm/SwiftLint/builds/104243015#L583. :disappointed:

marcelofabri:
I've investigated a little and found that it seems that most of time is spent in `NSString. lineAndCharacterForByteOffset` and `File.syntaxMap`.

I'm afraid that's as far as I got. If you guys have any ideas on how to make that faster, I'd be happy to implement.

norio-nomura:
Current `testTypeBodyLengths` is validating all rules.
We can reduce time by validating only `TypeBodyLengthRule`.

marcelofabri:
That apparently worked:

> Test Case '-[SwiftLintFrameworkTests.ASTRuleTests testTypeBodyLengths]' passed (0.886 seconds).

PR #382, opened 2016-01-18
rewrite a few functions to shed ~10s off compile times

jpsim:
It turns out that operators are expensive!

Running `xcodebuild -workspace SwiftLint.xcworkspace -scheme SwiftLintFramework clean build`:

before: 64.43s user 7.17s system 237% cpu 30.142 total
after: 39.46s user 6.94s system 236% cpu 19.615 total

Big thanks to @irace for writing up http://irace.me/swift-profiling which motivated me to do this :)


norio-nomura:
Nice!

scottrhoyt:
Awesome! It's funny how no matter the language, the argument for/against operators just won't go away.

Issue #268, opened 2015-12-14
Adding --strict does not log or display linter violations as errors


tamarnachmany:
It seems that the purpose of the --strict flag in SwiftLint is to display or log linter violations as errors. I'd like to use that functionality but I am still seeing everything that has a ```.Warning``` level of severity displayed in Xcode or logged on the command line as a warning. 

I have attempted to add this flag to a script which runs linting on certain project files, and also to linting on the command line.

In our build script I've tried:
* swiftlint lint --config "${SRCROOT}/.swiftlint.yml" --path $file --strict 
* swiftlint lint --strict --config "${SRCROOT}/.swiftlint.yml" --path $file
* swiftlint lint --config "${SRCROOT}/.swiftlint.yml" --path $file --strict 1
* swiftlint lint --strict 1 --config "${SRCROOT}/.swiftlint.yml" --path $file

```file``` is a variable in our build script.

On the command line I've tried the same thing, but without a reference to any particular file, and with different file path specification syntax, since I don't believe SRCROOT works correctly there.

The build script displays these violations as warnings and the command on the command line prints "warning........" for each violation. 

Has anyone else seen this issue? 

Thanks!

**Comment by jpsim** [comment]
The only thing that the `--strict` flag does is to change the exit status code. It's like Xcode's "treat warnings as errors" setting.

Warnings are still warnings, but unlike regular SwiftLint operation which exits `0` if only warnings were found, `--strict` will make the presence of warnings *or* errors exist with `1`:

```shell
$ echo "let abc:Int = 0" | swiftlint lint --use-stdin
<nopath>:1:5: warning: Colon Violation: Colons should be next to the identifier when specifying a type. (colon)
Done linting! Found 1 violation, 0 serious in 1 file.
$ echo $?
0
$ echo "let abc:Int = 0" | swiftlint lint --use-stdin --strict
<nopath>:1:5: warning: Colon Violation: Colons should be next to the identifier when specifying a type. (colon)
Done linting! Found 1 violation, 0 serious in 1 file.
$  echo $?
1
```

tamarnachmany:
In your view, what is the best way to display all violations as errors?

One thing I played around with was modifying the source code to add a 'severity' parameter to the style violation initializer (the same way it's being done in the force casting rule implementation) and to pass in .Error to the severity param, which did not work.

jpsim:
[...]

Issue #259, opened 2015-12-11
Global Let should allow capitalization of first charater


Marc-Hayes:
I have found that it is common, though not truly consistent, to have the first letter of global lets to be capitalized rather then lowercase.

My personal use case for this is declaration of NSNotificationCenter names in the same style as Apple.

```swift
public let TopLevelFooBarNotificationIdentifier: String = "com.foo.bar"

public class FooBar {
...
NSNotificationCenter.defaultCenter().postNotificationName(TopLevelFooBarNotificationIdentifier, object: nil, userInfo: nil)
...
}

public class NotFooBar {
...
NSNotificationCenter.defaultCenter().addObserver(self, selector: "keyboardVisible", name: UIKeyboardWillShowNotification, object: nil)
NSNotificationCenter.defaultCenter().addObserver(self, selector: "doTheThing:", name: TopLevelFooBarNotificationIdentifier, object: nil)
...
}
```

scottrhoyt:
I agree that this has also come up for me, though I can't decide how I feel about it. I've used capitals for the global cases just to be consistent with some of Apple's global constants. But Apple is also inconsistent. Sometimes it's the capital lettering, other times its the c-style `kConstant`, and so on. So I also think I'd be just as happy adhering to a more consistent standard.

scottrhoyt:
A workaround for this is now possible via the `excluded` parameter for naming rules.

Marc-Hayes:
Thank you for letting me know!

Issue #191, opened 2015-11-10
Issues with the Variable and Type Name Rule


kmikael:
Here are some cases that are correct imho, but are wrongly flagged by this rule, so that me and my team have to disable the variable name rule:

`error: Name Format Violation: Variable name should start with a lowercase character: 'URL' (variable_name)`

There should be exceptions to this rule like:

- URLs: `var URL: NSURL?`
- Global constants should start with capital letters according to the Cocoa conventions: `let ErrorDomain = "…"`, or for example notification names like `let SomethingDidChangeNotification = "SomethingDidChangeNotification"`

`warning: Name Format Violation: Variable name should be between 3 and 40 characters in length: 'n' (variable_name)`

- Sometimes, short variable names like `i` and `n` are appropriate imho, in mathematics contexts or very short closures.
- Also, when considering Cocoa notification constant names, 40 characters is also pretty restrictive. Here are some examples: `PKPassLibraryRemotePaymentPassesDidChangeNotification`, `NSUndoManagerDidOpenUndoGroupNotification`, `MPMoviePlayerReadyForDisplayDidChangeNotification` and many of the constants own our app are similarly longer and descriptive.

Also I find the 40 character restriction on type names also too limiting considering Cocoa. For example `MotorizedVehicleDescriptionCollectionViewCell` is totally valid in my opinion.

jpsim:
I'd welcome PRs addressing any of the concerns you've brought up that you believe should be generally applicable. I'd consider your recommendations for acronym variable names & capitalized global constants to be reasonable.

As for short variable names (`i` and `n`) or very long constant names (`PKPassLibraryRemotePaymentPassesDidChangeNotification`), I'd consider those to not be generally applicable, but rather just exceptions to the rule. For these cases, I'd suggest you temporarily disable the rule around those declarations (using `// swiftlint:disable variable_name` and `// swiftlint:enable variable_name`).

scottrhoyt:
Closing this because #391 provides a workaround.

PR #391, opened 2016-01-21
Configuration Structs

### LINKED ISSUE #130: Ability to customize severity / color of individual rules
I know there are several other tickets regarding customization of rules (e.g., #26) but I don't think this is covered:

It would be ideal to be able to customize not only rule parameters (e.g., the length of a line before the line-length-warning is triggered), but also *color*. Right now we have two colors: Yellow for warnings and red for errors. Severity tends to be in two states ("it's yellow ignore it" and "it's red I have to fix it").

Ideally, I could introduce a pale green (for, say, "FIXME" and "TODO" rules). This makes a clear distinction between to-do items that probably can be ignored, versus warnings that really should *not* be ignored.

### LINKED ISSUE #371: Configure severity of rules
I would like to be able to configure rules to give me a warning instead of an error and the other way around. 

For example force_cast and force_try are today errors, I don't want them to be but I might not want to disable them.

### LINKED ISSUE #388: Make TypeNameRule conform to ConfigurableRule
All its options are currently hardcoded: https://github.com/realm/SwiftLint/blob/9d138ad691f55fd016f883eb4044649b150582b7/Source/SwiftLintFramework/Rules/TypeNameRule.swift

scottrhoyt:
# Configuration Structs

This PR expands upon the work in #314 and #354 by providing `struct`-based configurations for the common patterns and providing default conformance to `ConfigurableRule` via these structures.

## Key Benefits

* All `Rule`s now have some sort of configuration available.
  * Basic rules are now configurable in their violation severity. (`SeverityConfig`)
  * Rules with multiple severity levels are configurable in one or both levels. (`SeverityLevelsConfig`)
  * Name evaluation rules are configurable in their `min_length` and `max_length` (both `warning` and `error`) as well as with `excluded` names. (`NameConfig`)
* Rule configurations can be partly or fully customized.
* Performance has been increased by ~2% (vs `master`) by folding `VariableNameMinLengthRule` and `VariableNameMaxLengthRule` into `VariableNameRule`.
* Test coverage has been increased.

## Key Drawbacks

* This introduces some additional verbosity because `ConfigProviderRule`s need to access their configuration via `config` and it's properties. This can be somewhat mitigated with more convenience computed properties.

## Example YAML

The following configuration is now possible.

```yaml
force_cast: Warning      # supports severity config case-insensitive
todo:
  severity: error        # severity can be configure key/value as well
line_length: 120         # only warning configured
file_length:
  warning: 500
  error: 1200
function_body_length:    # warning and error via array still supported
  - 40
  - 60
type_name:
  min_length: 4          # only warning
  max_length:            # warning and error
    warning: 40
    error: 50
  excluded: iPhone       # excluded via string
variable_name:
  min_length:            # only min_length
    error: 4             # only error
  excluded:              # excluded via string array
    - id
    - URL
    - GlobalAPIKey
```
## Open Issues Addressed

* Closes #388
* Closes #371
* Closes #130
* Workaround for #268 (via `severity`)
* Workaround for #259 (via `excluded`)
* Workaround for #191 (via `excluded`)

## Design Decisions

* Used `struct`s for `RuleConfig`s to encapsulate configuration via `Dictionary`s and also promote maximum reusability (e.g. `NameConfig` uses two `SeverityLevelsConfig`s).
* Left `ConfigurableRule` in place as a catch-all for non-standard configs (e.g. `MissingDocsRule`)
* Used `RuleConfig.setConfig(config:) throws` instead of a failable initializer for ingesting configuration dictionaries. This is to support the coexistence of default values and configured values so that partial configurations can be provided via the YAML.
* Switched `ConfigurableRule.init?(config:)` to `ConfigurableRule.init(config:) throws`to gracefully propograte these `ConfigurationError`s.

## TODO

- [x] Update README
- [x] Update CHANGELOG
- [x] Update CONTRIBUTING

/cc @realm/swiftlint


Source/SwiftLintFramework/Protocols/RuleConfig.swift
```diff
@@ -0,0 +1,14 @@
...
+//  Created by Scott Hoyt on 1/19/16.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import Foundation
+
+public protocol RuleConfig {
+    mutating func setConfig(config: AnyObject) throws
+    func isEqualTo(ruleConfig: RuleConfig) -> Bool
+}
```
norio-nomura: It can use:
```swift
extension RuleConfig where Self: Equatable {
    public func isEqualTo(ruleConfig: RuleConfig) -> Bool {
        if let ruleConfig = ruleConfig as? Self {
            return self == ruleConfig
        }
        return false
    }
}
```

Source/SwiftLintFramework/Rules/VariableNameRule.swift
```diff
@@ -42,19 +54,28 @@ public struct VariableNameRule: ASTRule {
...
-                return [StyleViolation(ruleDescription: description,
-                    severity: .Error,
-                    location: location,
-                    reason: "Variable name should only contain alphanumeric characters: '\(name)'")]
-            } else if kind != SwiftDeclarationKind.VarStatic && nameIsViolatingCase(name) {
-                return [StyleViolation(ruleDescription: description,
-                    severity: .Error,
-                    location: location,
-                    reason: "Variable name should start with a lowercase character: '\(name)'")]
+            if !config.excluded.contains(name) {
```
norio-nomura: `guard` ?

norio-nomura:
Great!

Source/SwiftLintFramework/Protocols/RuleConfig.swift
scottrhoyt: Great add.

Source/SwiftLintFramework/Rules/VariableNameRule.swift
scottrhoyt: Perfect use-case for a guard here as well as in `VariableNameRule`. I always forget about those. :)

scottrhoyt: Never mind on `TypeNameRule` given how it's currently implemented. The `where` makes sense.

Source/SwiftLintFramework/Protocols/RuleConfig.swift
scottrhoyt: Done!

norio-nomura: :+1:

scottrhoyt:
This might be a good time to add `ViolationSeverity.Info` so that it is an option for severity configuration.

jpsim:
Awesome! At first glance, I like this direction, but I might not be able to review in depth for a few days. I'm certainly happy to see this address so many issues users have faced regarding configuration!

Source/SwiftLintFramework/Models/ConfigurationError.swift
```diff
@@ -0,0 +1,13 @@
...
+//  ConfigurationError.swift
+//  SwiftLint
+//
+//  Created by Scott Hoyt on 1/19/16.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import Foundation
+
+public enum ConfigurationError: ErrorType {
```
jpsim: Why introduce this if there's only one member, and it's never checked? `ConfigurableRuleType.init(config:)` is only ever used with `try?`, which type-erases the error in a sense.

Source/SwiftLintFramework/Rules/RuleConfigs/NameConfig.swift
```diff
@@ -0,0 +1,71 @@
...
+
+import Foundation
+
+public struct NameConfig: RuleConfig, Equatable {
+    var minLength: SeverityLevelsConfig
+    var maxLength: SeverityLevelsConfig
+    var excluded: Set<String>
+
+    var minLengthThreshold: Int {
+        return Swift.max(minLength.warning, minLength.error)
```
jpsim: Why use the full `Swift.max` qualified name rather than just `max`?

Source/SwiftLintFramework/Rules/RuleConfigs/SeverityConfig.swift
```diff
@@ -0,0 +1,39 @@
...
+}
+
+public func == (lhs: SeverityConfig, rhs: SeverityConfig) -> Bool {
+    return lhs.severity == rhs.severity
+}
+
+// MARK: - ViolationSeverity extensions
+
+public extension ViolationSeverity {
+    init?(unnormalized: String) {
```
jpsim: This feels out of place. Since it's only used once, in this file, I think you should just pass `ViolationSeverity(rawValue: value.lowercaseString.capitalizedString)` at the call site in `setConfig` above.

jpsim:
This is pretty awesome! I'm happy with almost all of this, other than where noted, and the remaining checklist items from the original PR comment.

It shouldn't be too much work to rebase this on `master` for now, since it's only conflicting with #377 AFAICT, but #397 will probably cause much more rebase work, so we should probably merge this one first.

Source/SwiftLintFramework/Models/ConfigurationError.swift
scottrhoyt: This was something I thought about for a bit. Basically what was the best way to support configuration structures failing and provide optional configuration values functionality. To use the failable initializers approach that was previously used in `ConfigurableRule` would have required each rule to have its own `RuleConfig` type because you could only declare the default values in the structure. The second approach was to split initialization with default values and configuration with optional values into `init` and `setConfig`. Since we needed a way to indicate failure in `setConfig`, we could either throw or return some sort of success/failure flag (e.g. Bool, `Result`, etc.). Throwing felt the most Swifty to me, and I needed something to throw, so that's how this came to be.

There's certainly a case to be made that if we use this approach, we should introduce some finer-grain errors, possibly merge this with YamlParsingError, and probably change `Configuration.init?` to `Configuration.init() throws`. That's all possible in this PR, but in my head I was scoping it to the rework that needs to be done on logging and errors.

Source/SwiftLintFramework/Rules/RuleConfigs/NameConfig.swift
scottrhoyt: I'll change that. This was a leftover from when I originally had the `minLength` and `maxLength` properties shortened to `min` and `max` so I needed to namespace those calls.

Source/SwiftLintFramework/Rules/RuleConfigs/SeverityConfig.swift
scottrhoyt: That was what was originally done, but it greatly exceeded the `line_length` we have set for this project. I justified using this as the technique to shorten it because if this functionality breaks, it's mostly likely because something was changed with the `enum` so it made sense to me to encapsulate it there. I don't feel strongly about this though and would be fine added it as a method to SeverityConfig or putting it back at the call site and breaking that line up more.

scottrhoyt:
Great. I'll make those changes today so we can get this in.

Source/SwiftLintFramework/Models/ConfigurationError.swift
scottrhoyt: For the time being, I made this *marginally* more useful by not squashing the error in `Configuration.rulesFromDict`, though nothing is really being done with it. Let me know if you want to combine this with `YamlParserError`. I like that we have one less error type with that approach. But I don't like that a `ConfigProviderRule` would even know about something that knows about the `Yaml` subsystem.

scottrhoyt:
@jpsim take a look at this. Only outstanding issue is the use `ConfigurationError`. Let me know what you think about README and CONTRIBUTING. Is the provided info on customizing through `.swiftlint.yml` and contributing a `ConfigProviderRule` clear enough?

scottrhoyt:
FWIW, unit test took ~4 seconds before the rebase and ~16 seconds after on my computer. Probably related to the work with not counting comments and whitespace lines, but the tests don't feel like unit tests anymore.

CONTRIBUTING.md
```diff
@@ -22,37 +22,56 @@ those test cases in the unit tests directly. This makes it easier to understand
...
+configuration object via the `config` property:
+
+* The object provided must conform to `RuleConfig`.
+* There are several provided `RuleConfig`s that cover the common patterns like
+  configuring violation severity, violation severity levels, and evaluating
+  names.
+* If none of the provided `RuleConfig`s are applicable, you can create one
+  specifically for your rule.
+
+See [`ForceTryRule`](https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules/ForceCastRule.swift)
```
jpsim: Mentions `ForceTryRule` but links to `ForceCastRule`.

Source/SwiftLintFramework/Rules/VariableNameRule.swift
```diff
@@ -41,18 +53,29 @@ public struct VariableNameRule: ASTRule {
     public func validateFile(file: File, kind: SwiftDeclarationKind,
                              dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
         return file.validateVariableName(dictionary, kind: kind).map { name, offset in
+            guard !config.excluded.contains(name) else {
```
jpsim: In general, I think `if A {}` is preferable to `guard !A else {}`... we could even write a rule for this :wink:

Source/SwiftLintFramework/Rules/VariableNameRule.swift
```diff
@@ -41,18 +53,29 @@ public struct VariableNameRule: ASTRule {
...
                 return [StyleViolation(ruleDescription: description,
                     severity: .Error,
                     location: Location(file: file, byteOffset: offset),
                     reason: "Variable name should start with a lowercase character: '\(name)'")]
+            } else if let severity = severity(forLength: name.characters.count) {
+                return [StyleViolation(ruleDescription: self.dynamicType.description,
+                    severity: severity,
+                    location: Location(file: file, byteOffset: offset),
+                    reason: "Variable name should be between \(config.minLengthThreshold) " +
+                                "and \(config.maxLengthThreshold) characters long: '\(name)'")]
```
jpsim: indentation

jpsim:
> Only outstanding issue is the use ConfigurationError.

Actually your explanation makes sense: `throws` propagates errors, and better to throw an error in a domain under our control than a random `NSError`. :+1:

> Let me know what you think about README and CONTRIBUTING. Is the provided info on customizing through .swiftlint.yml and contributing a ConfigProviderRule clear enough?

Very useful and informative :+1: Thanks for adding the opt-in segment too!

Other than my last few comments, big :+1: on this!

CONTRIBUTING.md
scottrhoyt: Good catch!

scottrhoyt:
@jpsim I think we are all good here now.

jpsim:
:+1: :100: kudos. Can you make sure to close all the issues resolved by this?

scottrhoyt:
No problem. The 3 that are 100% addressed have been closed. What do you want to do with the 3 that this provided a solid workaround for?

jpsim:
I think this PR is the best way to address #191, despite not being exactly what the user wanted, so I'd close that one as resolved. The other 2 can remain open.

norio-nomura:
Wow! :tada:

tamarnachmany:
:pizza:

derpoliuk:
Thanks for PR!

Is there anyway to set severity for all rules? I want to make sure that build fails if **any** SwiftLint rule is violated.

ikesyo:
@derpoliuk What you want would be the `--strict` option of `lint` command.

```
[--strict]
	fail on warnings
```

derpoliuk:
@ikesyo thanks for very quick answer! 

I know about `--strict` option (should've addressed it in my first comment), but it doesn't provide pretty output (as discussed in #268). Though it's a working solution.

PR #393, opened 2016-01-21
fix infinite recursion when using nested config files. Fixes #368

### LINKED ISSUE #368: Segmentation fault with nested configs and autocorrect.
Not 100% sure on this one. Was getting a segmentation fault using `autocorrect`, and I removed `enable_nesting_configs: true` and it went away. There might be something to investigate here.

jpsim:
/cc @scottrhoyt 

Also replaced all the hardcoded `.swiftlint.yml` strings while I was in there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,9 @@
 
 ##### Bug Fixes
 
-* None.
+* Fix crash caused by infinite recursion when using nested config files.  
+  [JP Simard](https://github.com/jpsim)
+  [#368](https://github.com/realm/SwiftLint/issues/368)
 
 ## 0.6.0: Steam Cycle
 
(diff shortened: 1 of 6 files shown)


scottrhoyt:
Love getting those literals out of there. The sanity check against getting to the `/` directory is a great add as well. We should never get to `/` in the search though (unless you are linting the `/` path). I believe the more direct solution here is to set `Configuration.rootPath` in the `autocorrect` command how it is set in the `lint` command [here](https://github.com/realm/SwiftLint/blob/cc41abbdfa79fa53483530afdc12c0fb04562ee5/Source/swiftlint/Commands/LintCommand.swift#L24). :+1:

scottrhoyt:
Closes #368

jpsim:
Ah, you're right @scottrhoyt, that's definitely the correct fix. Updated. Thanks for reviewing!

scottrhoyt:
No problem. Yeah, it probably makes sense to be setting root path in `Configuration.init` given how critical it is to the nested configs option. But ideally it would be nice to refactor the directory traversal out into a new consumer of `Configuration` or perhaps into `Linter`, though I'd prefer the former for SRP purposes.

PR #398, opened 2016-01-23
Prepare codebase for Swift Package Manager support

jpsim:
Avoids Swift 2.2 deprecations and adds missing imports needed by SPM.


scottrhoyt:
Looks good to me!

jpsim:
Merging since a quick glance seems to indicate it should cause conflicts with #391.

PR #399, opened 2016-01-23
Reduce creation of `Location`

norio-nomura:
Performance improvement is small on real usage.
But, it reduces the duration of `make test`
from:
```
…
✓ testTypeBodyLengths (15.942 seconds)
…
Executed 70 tests, with 0 failures (0 unexpected) in 22.355 (22.393) seconds
```
to:
```
…
✓ testTypeBodyLengths (8.197 seconds)
…
Executed 70 tests, with 0 failures (0 unexpected) in 14.455 (14.500) seconds
```

This is that I mentioned on https://github.com/realm/SwiftLint/pull/377#issuecomment-174112575.


jpsim:
:+1:

PR #401, opened 2016-01-23
Testing only TypeBodyLengthRule on testTypeBodyLengths

marcelofabri:
As commented on #377.

Also, IMO this probably doesn't reserve an entry in the changelog, so I haven't added it.

### DIFF
--- a/Source/SwiftLintFrameworkTests/ASTRuleTests.swift
+++ b/Source/SwiftLintFrameworkTests/ASTRuleTests.swift
@@ -141,31 +141,42 @@ class ASTRuleTests: XCTestCase {
     }
 
     func testTypeBodyLengths() {
+        let identifier = TypeBodyLengthRule.description.identifier
+        let disabledRules = allRuleIdentifiers.filter { $0 != identifier }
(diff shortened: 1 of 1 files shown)


jpsim:
Why not use the standard test approach here of having these go through `verifyRule()`, which already does this?

marcelofabri:
Probably because `triggeringExamples` and `nonTriggeringExamples` would be large? Why didn't this rule called `verifyRule` before?

jpsim:
> Probably because triggeringExamples and nonTriggeringExamples would be large? Why didn't this rule called verifyRule before?

I think the tests for this rule were originally written before the introduction of `verifyRule`. If `triggeringExamples` and `nonTriggeringExamples` are too large to have inline with the rule declaration, we could still construct a `RuleDescription` in this test and invoke `verifyRule` with it.

jpsim:
Thanks for this, @marcelofabri but I've decided to do a more comprehensive clean up of rules tests in #408, so this won't be needed.

marcelofabri:
:+1:

Issue #402, opened 2016-01-23
Cyclomatic complexity rule


garnett:
Surprisingly haven't found this rule https://en.wikipedia.org/wiki/Cyclomatic_complexity
There is "nested ifs" rule, but cyclomatic complexity is different. Looks like it should be pretty easy to implement (just count if, guard, switch, etc.)

It probably should be opt in (or with high triggering values), because from my experience it is considered to be pretty "hardcore" to obey.

@jpsim if you think it might be useful I would be interested in implementing that

scottrhoyt:
This is functionality that I would find very useful while developing, but I think this is more the domain of a static analyzer than a linter.

garnett:
@scottrhoyt I agree with you, but the thing is that `SwiftLint` is only project in the wild right now which allows "analysing"  and IMHO `lint` word in it's name confuses people (me). For example, https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules/EmptyCountRule.swift looks to me like static analysis too, but there is not other project/repo where you can put it right now.

daniel-beard:
I think it would be nice for swiftlint to have a cyclomatic complexity rule, so that for iOS dev it's covered for both Objective-C and swift (OCLint has a rule for this: https://github.com/oclint/oclint/blob/master/oclint-rules/rules/size/CyclomaticComplexityRule.cpp)

jpsim:
>  if you think it might be useful I would be interested in implementing that

Would be tremendously useful. Go for it!

garnett:
Implemented!

PR #405, opened 2016-01-24
Cyclomatic complexity rule

garnett:
Implements #402 
Provided implementation should not be considered as exhaustive, e.g. it does not count semantic langauge features (e.g. `fallthrough` should reduce comlexity of `switch` by one)
http://stackoverflow.com/questions/30240236/cyclomatic-complexity-of-switch-case-statement

Values for warning and error where chosen based on this thread: http://programmers.stackexchange.com/questions/194061/cyclomatic-complexity-ranges


jpsim:
Awesome work! I just made a few minor edits in #407 and will merge soon. Thanks for your contribution :tada:

PR #406, opened 2016-01-24
[WIP] Generate assertion on every (non)triggering examples in `verifyRule()`

norio-nomura:
It makes failed triggering/non-triggering example locatable on Xcode's Issue Navigator.
e.g.:
screenshot 2016-01-24 20 02 59


jpsim:
This is interesting, I'm not sure how I feel about it yet. Would it not be possible to make `Trigger` conform to string literal convertible so we can continue to use strings in the declaration?

jpsim:
Thinking about this more, I'm pretty sure it's impossible to make `Trigger` conform to `StringLiteralConvertible` while preserving the line & file where it was declared. But the benefits of this approach outweigh the marginally more verbose rule definitions IMO.

@scottrhoyt what do you think?

norio-nomura:
At next morning, I also feel that too verbose. :dizzy_face: 
I want to think another approach that is less verbose and in proportion.

norio-nomura:
After tried few another approaches, I can't get better implementation than this PR's.

By this PR, I want to make detecting easy which example failed **without debugger**.
For getting that, I tried…
- Using `Triggers` struct instead of `[String]`. That can't locate the exact line of example.
- Above + locating string from .swift using SourceKitten API. That is too complicated and can't locate exact line of example.

I think this PR is straightforward approach.

But after knowing which example failed, we mostly need to using debugger anyway. :stuck_out_tongue:

jpsim:
Thanks for sharing your thought process @norio-nomura. I could go either way on this one, so I'd like to hear what @scottrhoyt thinks.

scottrhoyt:
Hmm. It's nice functionality to have, but it does only get you a step further when dealing with these cases. The extra verbosity is a real cost. I can't say I feel strongly about it either way. I'm trying to think of alternative approaches.

Here's an idea that might not be good:

```swift
extension String {
   var doesTrigger: Trigger { ... }
   var doesNotTrigger: Trigger { ... }
}

examples: [
// Does Not Trigger
  "// notaTODO:\n".doesNotTrigger,
  "// notaFIXME:\n".doesNotTrigger,

// Does Trigger
  "// ↓TODO:\n".doesTrigger,
  "// ↓FIXME:\n".doesTrigger,
  "// ↓TODO(note)\n".doesTrigger,
  "// ↓FIXME(note)\n".doesTrigger,
  "/* ↓FIXME: */\n".doesTrigger,
  "/* ↓TODO: */\n".doesTrigger,
  "/** ↓FIXME: */\n".doesTrigger,
  "/** ↓TODO: */\n".doesTrigger
]

extension RuleDescription {
   func triggerViolations() -> [StyleViolation]
}

extension Rule {
  func verifyTriggers() -> [StyleViolation]
}

XCTAssertEqual(rule.dynamicType.triggerViolations(), rule.verifyTriggers())
```

I don't know if that would get us any further, but I was just trying to come up with some additional value from the verbosity. In this case it would be a little extra syntactical sugar and readability.

norio-nomura:
Thanks for reviews.

I did think about "How can I remove those repeating `Trigger()`?" yesterday.
Maybe I should think "How can we make better those test?".

I'll try to use @scottrhoyt's suggestion!

jpsim:
The `String.doesTrigger` extensions won't allow preserving the line on which the strings are defined, but `String.doesTrigger(file: String = __FILE__, line: Int = __LINE__)` would, although at that point we might as well just stick with the `Trigger` initializer currently in this PR.

scottrhoyt:
That makes sense. This approach is probably the best way of achieving this then.

scottrhoyt:
Ah, if you want to make the string extensions work, they just need to be functions with default values instead of computed properties:

screen shot 2016-01-25 at 3 29 18 pm

jpsim:
That's what I said in https://github.com/realm/SwiftLint/pull/406#issuecomment-174712284, but noted that at that point, we're better off with the current `Trigger` initializers from this PR.

scottrhoyt:
Ah yes, sorry read on my email earlier and missed that.

Yeah, it's not really any better--just a style preference.

norio-nomura:
Updated.

jpsim:
I'm sorry to say I strongly prefer `Trigger("")` to this :cry:

norio-nomura:
No problems! :wink: 
By the way, how do you think merging `nonTriggeringExamples` and `triggeringExamples` into `examples`?
It has a benefit on following case:
```swift
public struct NestingRule: ASTRule, ConfigProviderRule {
…
        examples: ["class", "struct", "enum"].flatMap { kind in [
            doesNotTrigger("\(kind) Class0 { \(kind) Class1 {} }\n"),
            doesTrigger("\(kind) A { \(kind) B { ↓\(kind) C {} } }\n"),
        ]} + [
            doesNotTrigger("func func0() {\nfunc func1() {\nfunc func2() {\nfunc func3() {\nfunc func4() { " +
            "func func5() {\n}\n}\n}\n}\n}\n}\n"),
            doesNotTrigger("enum Enum0 { enum Enum1 { case Case } }"),
            doesTrigger("func func0() {\nfunc func1() {\nfunc func2() {\nfunc func3() {\nfunc func4() { " +
            "func func5() {\n↓func func6() {\n}\n}\n}\n}\n}\n}\n}\n"),
        ]
```

jpsim:
With that specific example, being able to reuse the map is nice, but it's harder to read which examples trigger and which ones don't, something that's currently quite clear with `nonTriggeringExamples` and `triggeringExamples`.

norio-nomura:
Thanks.
I have reset to 9cb3b45.

norio-nomura:
After I saw #425, I close this.

Issue #376, opened 2016-01-15
SwiftLint Performance Benchmarking


scottrhoyt:
This issue has popped up in a number of instances. As we create more sophisticated rules, we need to be mindful of the performance implications and the requirements of being a linting tool. To aid in this, it would be nice to automate some performance benchmarks on CI builds. I started this discussion to evaluate proposals.

keith:
This would be great. I spent a huge amount of time investigating optimization points when I first implemented SwiftLint in our project. It would be a shame if anyone's work on that was silently regressed.

jpsim:
The best way I could come up to do automated performance testing would be to run a separate Travis job that builds SwiftLint from `master` and the PR commit that triggered the job, and runs both against a fixed version tagged checkout of an OSS Swift project with a reasonable number of files to lint (Carthage comes to mind).

Ideally, we'd add a `--benchmark` flag to the lint command that would log the time spent in each file or rule.

scottrhoyt:
One idea I had was to make sure the current release of SwiftLint it installed:

```sh
brew update
brew outdated swiftlint || brew upgrade swiftlint
```

Then write a shell script to lint a benchmark project 3 times, taking the average time. This benchmark could be the SwiftLint project itself or some community standard like [Alamofire](https://github.com/Alamofire/Alamofire) or the [Swift Standard Library](https://github.com/apple/swift/tree/master/stdlib).

From there, we can use the package built by the `make package` that is already being run to install the version to test:

```sh
sudo installer -pkg swiftlint.pkg -target / -allowUntrusted
```

And then run the same benchmark, comparing the results?

Do you guys think this would work?

P.S. it looks like we should be fine with `sudo` privileges on Travis.

jpsim:
I don't think benchmarking a pull request against the latest homebrew release is a viable option, because at that point you're not just evaluating the PR, but all other changes since the release. I think the baseline needs to be against `master`.

Benchmarking multiple times and averaging the results *may* make sense, we'll have to look at the standard deviation between runs and if it's consistently small, it shouldn't be necessary to do multiple runs at all.

scottrhoyt:
I agree that benchmarking against `master` is the ideal way to do it. It just seemed harder to set up and harder to potentially associate with a particular PR. It will also add a good amount of work to the build to package master every time (~2x). While the comparison against the last release should be easier (at the expense of some false positives) and alert us to large problems. 

[...]

PR #410, opened 2016-01-24
release 0.7.0

jpsim:


jpsim:
Some numbers:

| Time to...         | 0.6.0 | 0.7.0 |
| ------------------ | ----- | ----- |
| build              | 35.6s | 22.6s |
| test               | 7.2s  | 9.3s  |
| lint Carthage 0.11 | 16.3s | 19.4s |

So we regressed a bit on performance with this release, so it would be wise to implement some performance tests (#376) sooner rather than later.

NachoSoto:
I wonder if the increased lint time is due to new rules? In that case it would be more fair to compare with the same set of rules :)

jpsim:
I also tried with `excluded:\n  - type_body_length\n  - function_body_length` with minimal impact.

Issue #394, opened 2016-01-22
Initialize regexes once per Rule


scottrhoyt:
I've come across this while implementing #123. There are quite a few rules that compile a regex for every file that they lint. Given that these regexes are constant and the construction cost is non-trivial, I believe we can get a small performance bump from storing the precompiled regex once for each rule that needs one.

jpsim:
How would one "precompile" a regex?

scottrhoyt:
Sorry, old semantics. I just mean initializing them once and holding them for the life of the `Rule`. Instantiating these things can be relatively expensive. Given my cursory testing, it would probably only be ~2% performance increase, but given it's such a small change, it might be worth it, but maybe not. Doubtless there is more to gain in optimizing the patterns and rules themselves, but this just seemed like a quick win.

jpsim:
Ah, I see what you mean. Yes, I'd be in favor of only generating the regexes once per run. Maybe as static variables on the rules that need them? Would be trivial to do.

scottrhoyt:
Yeah, something like that I think. Should speed up testing a bit too.

scottrhoyt:
I'm going to close this because my testing has not indicated a significant speedup via initializing regex's differently. It is quite possible that there are optimizations happening behind the scenes that make this unnecessary. However, there is still the smaller task of cleaning up and standardizing how `NSRegularExpression`s are created.

keith:
I tested this a while back and it seems like the framework caches recently used regexes

PR #416, opened 2016-01-25
Custom Rules

### LINKED ISSUE #123: Generic regex rules
As we've been expanding our usage of SwiftLint we've added a lot of simple regex rules to conform to our style guide. I'm not sure what the long term plans are to handle custom rules but it might be nice to provide a way for users to add regex rules in their `.swiftlint.yml`. For example:

```yml
regex:
  - (^[^\\s]+\\s+@objc|@objc[^\\n_])
```

This would create multiple rules (once https://github.com/realm/SwiftLint/pull/106 is merged) at runtime.

There are a few issues I can see with this, specifically around them not being customizable. I'm not sure how the warning/error message would be generated and things like `NSRegularExpression` options wouldn't be unusable.

I just wanted to get some feedback on this idea before putting time into it. Let me know your thoughts!

scottrhoyt:
Closes #123 

This PR adds the ability for end users to define their own custom rules. The rules are defined in `.swiftlint.yml` as follows:

```yaml
custom_rules:
  no_scott:                # rule identifier
    name: "No Scott Rule"  # rule name
    regex: "([s,S]cott)"   # matching pattern 
    match_kinds:           # SyntaxKinds to match
      - comment
      - string
    message: "There should be no Scotts here"
    severity: error
```

Which will generate the following:

screen shot 2016-01-25 at 7 22 01 am

The identifier and regex are required. All other fields are optional. There is no limit to the number of `custom_rules` that can be defined other than performance degradation.

### Testing

The new code is all tested, but I could use some suggestions for a testing strategy to more rigorously test `CustomRules`.

### TODO

- [x] Upgrade testing if necessary
- [x] Update README
- [x] Update CHANGELOG

/cc @realm/swiftlint


jpsim:
This is AWESOME. :clap:

I like the direction, implementation & usage. As for testing, I think it's preferable to test the mechanism rather than to offer a way to test user-provided custom rules, so the tests you have here are sufficient IMO.

I'd say the only things left are CHANGELOG + README.

Source/SwiftLintFramework/Extensions/NSRegularExpression+SwiftLint.swift
```diff
@@ -0,0 +1,20 @@
...
+//
+
+import Foundation
+
+public extension NSRegularExpression {
+    public convenience init(pattern: String) throws {
+        try self.init(pattern: pattern,
+            options: [.AnchorsMatchLines, .DotMatchesLineSeparators])
+    }
+    public static func forcePattern(pattern: String) -> NSRegularExpression {
```
jpsim: This deprecates the `regex()` function? Shouldn't this be `internal`?

jpsim: There's also a comment on `regex()` stating something like "safe to force cast because all regexes used in SwiftLint are verified to work" or something which would make sense to restate here IMO.

Source/SwiftLintFramework/Extensions/SyntaxKind+SwiftLint.swift
```diff
@@ -9,6 +9,15 @@
 import SourceKittenFramework
 
 extension SyntaxKind {
+    // swiftlint:disable:next function_body_length
```
jpsim: unnecessary disable command?

Source/SwiftLintFramework/Rules/CustomRules.swift
```diff
@@ -0,0 +1,71 @@
...
+// MARK: - CustomRules
+
+public struct CustomRules: Rule, ConfigProviderRule {
+
+    public static let description = RuleDescription(
+        identifier: "custom_rules",
+        name: "Custom Rules",
+        description: "Create custom rules by providing a regex string. " +
+          "Optionally specify what syntax kinds to match against, the severity " +
+          "level, and what message to display")
```
jpsim: please add a terminating period here

Source/SwiftLintFramework/Rules/CustomRules.swift
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        return config.customRuleConfigs.flatMap {
+            self.validate(file, withConfig: $0)
+        }
+    }
+
+    private func validate(file: File, withConfig config: RegexConfig) -> [StyleViolation] {
+        return file.matchPattern(config.regex).filter {
+                !config.matchKinds.intersect($0.1).isEmpty
```
jpsim: would be good to document this behavior in the custom rules documentation.

Source/SwiftLintFramework/Extensions/NSRegularExpression+SwiftLint.swift
scottrhoyt: Yeah, I was planning on dealing with this all in a branch dedicated to optimizing regex usage as talked about in #394. I can make it internal, but then I will need to `@testable import SwiftLintFramework` in the tests, and it seemed like we might be staying away from that?

Source/SwiftLintFramework/Extensions/SyntaxKind+SwiftLint.swift
scottrhoyt: Good catch. I'll delete that. It was a holdover from a more complicated implementation of this.

Source/SwiftLintFramework/Rules/CustomRules.swift
scottrhoyt: :+1:

Source/SwiftLintFramework/Rules/CustomRules.swift
scottrhoyt: Do you mean how `matchKinds` is used?

jpsim: Yes

Source/SwiftLintFramework/Rules/RuleConfigs/RegexConfig.swift
```diff
@@ -0,0 +1,63 @@
...
+        if let matchKinds = [String].arrayOf(configDict["match_kinds"]) {
+            self.matchKinds = Set( try matchKinds.map { try SyntaxKind(shortName: $0) })
+        }
+        if let severityString = configDict["severity"] as? String {
+            try severityConfig.setConfig(severityString)
+        }
+    }
+}
+
+public func == (lhs: RegexConfig, rhs: RegexConfig) -> Bool {
```
NachoSoto: I think this should be `==(...)` (without the space). I'm only bringing this up because I've noticed that there's not a `SwiftLint` rule for this, so I'm wondering whether my "convention" is different than most people's, or whether maybe this is a good rule suggestion :)

scottrhoyt: Yeah, I don't feel strongly either way about this one. I was just following the lead here:
https://github.com/github/swift-style-guide/blob/master/README.md#use-whitespace-around-operator-definitions

scottrhoyt:
@jpsim `CHANGELOG` and `README` updated. For lack of a better example, I opted to be a bit silly in the Defining Custom Rules section, but if you have something better to use let me know. Also, maybe it makes sense to link to this section from the Rules section? Let me know what you think.

README.md
```diff
@@ -186,6 +186,53 @@ variable_name:
...
+  no_hiding_in_strings:
+    regex: "([n,N]inja)"
+    match_kinds: string
+```
+
+This is what the output would look like:
+
+![](custom-rule.png)
+
+By default, custom rules will match against all syntax kinds. You can filter
```
jpsim: I still don't think this sufficiently conveys the logic used to match syntax kinds. Consider making this clearer, such as:

> You can filter the matches by providing one or more `match_kinds`, which will reject matches that include syntax kinds that are not present in this list. Here are all the possible syntax kinds:

jpsim:
:+1: other than my last comment. Excellent work! :clap:

Source/SwiftLintFramework/Extensions/NSRegularExpression+SwiftLint.swift
jpsim: This is fine.

scottrhoyt:
Thanks! Should be good to go now.

norio-nomura:
Great! :clap:

PR #419, opened 2016-01-26
Remove old VariableName*LengthRules.swift files

norio-nomura:
These are unused since ee3fe42.


jpsim:
Oops, thanks!

scottrhoyt:
Oops. Good catch. I caught that in the custom rules work, but I guess I dropped the commit in a rebase or something. Thanks!

norio-nomura:
:smile:

PR #420, opened 2016-01-26
Fix ValidDocs rule

### LINKED ISSUE #375: Valid docs rule not covering all closure parameter cases.
I've noticed that some cases are not being covered by the `validDocsRule` and SwiftLint generates wrong warnings, for instance:
* `func foobar (closure1: () -> Foo, closure2: (foobar: Bar) -> Foo, closure3: (foobar: Bar) -> Foo) {`
* `func footer (closure1: (Foo) -> Void, onError closure2: ((NSError) -> Void)? = nil) {`

I'll add a pull request with a fix for this ASAP.

diogoguimaraes:
Closes #375 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@
 
 ##### Bug Fixes
 
-* None.
+* Fix several false positives in `ValidDocsRule`.  
+  [diogoguimaraes](https://github.com/diogoguimaraes)
+  [#375](https://github.com/realm/SwiftLint/issues/375)
 
 ## 0.7.0: Automatic Permanent Press
 
--- a/Source/SwiftLintFramework/Rules/ValidDocsRule.swift
+++ b/Source/SwiftLintFramework/Rules/ValidDocsRule.swift
@@ -39,9 +39,10 @@ func superfluousOrMissingThrowsDocumentation(declaration: String, comment: Strin
     return declaration.containsString(" throws ") == !comment.containsString("- throws:")
 }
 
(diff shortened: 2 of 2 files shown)


diogoguimaraes:
@jpsim can you help me with the `codecov` please?

jpsim:
Thanks for this! The code coverage isn't significantly affected so that's fine, but this code is kind of duplicated at this point, so it'd be good to combine, which would also incidentally help code coverage.

Nice work!

diogoguimaraes:
Thanks! 
Do you think we should do this before merging this PR?
Regarding the "combine" could you be more specific?

jpsim:
I made a few minor changes: combining some duplicate code and adding a changelog entry. This will be good to go once it passes CI.

Thanks again!

diogoguimaraes:
Thanks again!
Do you have an ETA to the next release?

jpsim:
0.7.0 was accepted into Homebrew yesterday: https://github.com/Homebrew/homebrew/pull/48425#issuecomment-174798049

Do you need the fix from this PR in a release? If so, I could push out a 0.7.1 release today, although I can't say when Homebrew will accept it. The last few submissions each took a few days.

diogoguimaraes:
That would be great! Thank you.

jpsim:
0.7.1 has been cut: https://github.com/realm/SwiftLint/releases/tag/0.7.1

and submitted to homebrew: https://github.com/Homebrew/homebrew/pull/48497

PR #421, opened 2016-01-26
Add opt-in force unwrapping rule

jpsim:
Continued from #374 /cc @scottrhoyt @norio-nomura @Argent


jpsim:
I think this would only really make sense in certain parts of a project, if ever, which is why it's an opt-in rule.

jpsim:
This is an opt-in rule, and probably fine as-is. There's just the one false positive in the SwiftLint codebase mentioned in https://github.com/realm/SwiftLint/pull/374#issuecomment-172890579.

norio-nomura:
:+1:

PR #429, opened 2016-01-27
Add `NSRegularExpression.cached(pattern:)`

norio-nomura:
All rules enabled linting Carthage 0.11 with this commit:
- Persistent Bytes on termination is reduced from 584.67MB to 556.62MB.
- Duration is reduced from 22sec to 21sec.

Related: #394

### DIFF
--- a/Source/SwiftLintFramework/Extensions/NSRegularExpression+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/NSRegularExpression+SwiftLint.swift
@@ -8,13 +8,21 @@
 
 import Foundation
 
+private var regexCache = [String: NSRegularExpression]()
+
 public extension NSRegularExpression {
-    public convenience init(pattern: String) throws {
-        try self.init(pattern: pattern,
+    public static func cached(pattern pattern: String) throws -> NSRegularExpression {
+        if let result = regexCache[pattern] {
+            return result
+        }
+
+        let result = try NSRegularExpression(pattern: pattern,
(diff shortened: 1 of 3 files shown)


scottrhoyt:
:+1:

Source/SwiftLintFramework/Extensions/NSRegularExpression+SwiftLint.swift
```diff
@@ -8,13 +8,21 @@
 
 import Foundation
 
+private var regexCache = [String: NSRegularExpression]()
+
 public extension NSRegularExpression {
-    public convenience init(pattern: String) throws {
-        try self.init(pattern: pattern,
+    public static func cached(pattern pattern: String) throws -> NSRegularExpression {
```
jpsim: Why not keep the old API (`NSRegularExpression.init(pattern:)`)? We should probably start being more conservative about API breaking changes if we want to prepare for 1.0 :smile:.

jpsim: Yeah, I see now, we can't have this functionality as an initializer. My bad. :+1:

jpsim:
:+1: 

Although this is technically a breaking change of the publicly accessible API, we haven't really been enforcing that so far, so let's merge this. But it'd be good to start thinking about that moving forward.

jpsim:
And of course, thanks so much @norio-nomura for digging into this and getting us some nice perf improvements :smile: !

norio-nomura:
Thanks for reviews and merge! :pray:

PR #431, opened 2016-01-27
Disable SeverityLevelsConfig error ruleParam if only warning is set

robinkunde:
SeverityLevelsConfig is used for both rules where the warning threshold must logically be lower than the error threshold and vice versa.

In order to prevent the generation of invalid configs via implicit definition in yaml, the error ruleParameter should be disable in such cases. For example, the README contains this code right now:

```yaml
# rules that have both warning and error levels, can set just the warning level
# implicitly
line_length: 110
```

If 110 is larger than whatever default value SwiftLint define for error, this would be an invalid config.

This commit also updates the test names since ViolationLevel is no more.

### DIFF
--- a/Source/SwiftLintFramework/Rules/RuleConfigs/NameConfig.swift
+++ b/Source/SwiftLintFramework/Rules/RuleConfigs/NameConfig.swift
@@ -14,11 +14,11 @@ public struct NameConfig: RuleConfig, Equatable {
     var excluded: Set<String>
 
     var minLengthThreshold: Int {
-        return max(minLength.warning, minLength.error)
+        return max(minLength.warning, minLength.error ?? minLength.warning)
     }
 
     var maxLengthThreshold: Int {
-        return min(maxLength.warning, maxLength.error)
+        return min(maxLength.warning, maxLength.error ?? maxLength.warning)
     }
 
     public init(minLengthWarning: Int,
--- a/Source/SwiftLintFramework/Rules/RuleConfigs/SeverityLevelsConfig.swift
+++ b/Source/SwiftLintFramework/Rules/RuleConfigs/SeverityLevelsConfig.swift
@@ -10,24 +10,25 @@ import Foundation
 
 public struct SeverityLevelsConfig: RuleConfig, Equatable {
     var warning: Int
-    var error: Int
+    var error: Int?
 
     var params: [RuleParameter<Int>] {
-        return [RuleParameter(severity: .Error, value: error),
+        if let error = error {
+            return [RuleParameter(severity: .Error, value: error),
                 RuleParameter(severity: .Warning, value: warning)]
+        }
+        return [RuleParameter(severity: .Warning, value: warning)]
     }
 
     mutating public func setConfig(config: AnyObject) throws {
         if let config = [Int].arrayOf(config) where !config.isEmpty {
             warning = config[0]
-            if config.count > 1 {
-                error = config[1]
-            }
+            error = (config.count > 1) ? config[1] : nil
         } else if let config = config as? [String: AnyObject] {
             if let warningNumber = config["warning"] as? Int {
                 warning = warningNumber
-            }
-            if let errorNumber = config["error"] as? Int {
+                error = config["error"] as? Int
+            } else if let errorNumber = config["error"] as? Int {
                 error = errorNumber
             }
         } else {
--- a/Source/SwiftLintFrameworkTests/RuleTests.swift
+++ b/Source/SwiftLintFrameworkTests/RuleTests.swift
@@ -13,7 +13,7 @@ import SourceKittenFramework
 struct RuleWithLevelsMock: ConfigProviderRule {
     var config = SeverityLevelsConfig(warning: 2, error: 3)
 
-    static let description = RuleDescription(identifier: "violation_level_mock",
+    static let description = RuleDescription(identifier: "severity_level_mock",
         name: "",
         description: "")
     func validateFile(file: File) -> [StyleViolation] { return [] }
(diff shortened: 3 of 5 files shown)


Source/SwiftLintFrameworkTests/RuleTests.swift
```diff
@@ -58,7 +58,7 @@ class RuleTests: XCTestCase {
         XCTAssertFalse([RuleMock1(), RuleMock2()] == [RuleMock1()])
     }
 
-    func testViolationLevelRuleInitsWithConfigDictionary() {
+    func testSecurityLevelRuleInitsWithConfigDictionary() {
```
jpsim: This should be `testSeverityLevelRuleInitsWithConfigDictionary`. All the other tests changed in this file now also say "Security" where they should say "Severity".

jpsim:
Thanks for the PR, @robinkunde! This is fine, but having `error` be an `Optional<Int>` would be more appropriate.

robinkunde:
@jpsim That's much Swiftier solution! I've amended the PR accordingly.

Source/SwiftLintFramework/Rules/RuleConfigs/NameConfig.swift
```diff
@@ -14,11 +14,19 @@ public struct NameConfig: RuleConfig, Equatable {
     var excluded: Set<String>
 
     var minLengthThreshold: Int {
-        return max(minLength.warning, minLength.error)
```
jpsim: This can just be `return max(minLength.warning, minLength.error ?? minLength.warning)`

Source/SwiftLintFramework/Rules/RuleConfigs/NameConfig.swift
```diff
@@ -14,11 +14,19 @@ public struct NameConfig: RuleConfig, Equatable {
...
+        if let minLengthError = minLength.error {
+            return max(minLength.warning, minLengthError)
+        } else {
+            return minLength.warning
+        }
     }
 
     var maxLengthThreshold: Int {
-        return min(maxLength.warning, maxLength.error)
```
jpsim: This can just be `return min(maxLength.warning, maxLength.error ?? maxLength.warning)`

Source/SwiftLintFramework/Rules/RuleConfigs/SeverityLevelsConfig.swift
```diff
@@ -10,22 +10,31 @@ import Foundation
 
 public struct SeverityLevelsConfig: RuleConfig, Equatable {
     var warning: Int
-    var error: Int
+    var error: Int?
 
     var params: [RuleParameter<Int>] {
-        return [RuleParameter(severity: .Error, value: error),
+        if error == nil {
```
jpsim: You can avoid the force unwrapping by flipping this around:

```swift
if let error == error {
    return [RuleParameter(severity: .Error, value: error),
        RuleParameter(severity: .Warning, value: warning)]
}
return [RuleParameter(severity: .Warning, value: warning)]
```

Source/SwiftLintFramework/Rules/RuleConfigs/SeverityLevelsConfig.swift
```diff
@@ -10,22 +10,31 @@ import Foundation
...
-            if config.count > 1 {
+            if config.count == 1 {
+                error = nil
+            } else {
                 error = config[1]
             }
         } else if let config = config as? [String: AnyObject] {
             if let warningNumber = config["warning"] as? Int {
                 warning = warningNumber
+                if config["error"] == nil {
```
jpsim: instead of adding another check here, just set the following:

```swift
error = config["error"] as? Int
```

jpsim:
Sorry for all the nitpicky comments, @robinkunde, this is definitely going in the right direction.

Please also add a changelog entry following our [guidelines](https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes). This is technically a breaking change.

robinkunde:
No worries, I appreciate the pointers. I've updated the commit.

Source/SwiftLintFramework/Rules/RuleConfigs/SeverityLevelsConfig.swift
```diff
@@ -10,24 +10,25 @@ import Foundation
...
-            }
+            error = (config.count > 1) ? config[1] : nil
         } else if let config = config as? [String: AnyObject] {
             if let warningNumber = config["warning"] as? Int {
                 warning = warningNumber
-            }
-            if let errorNumber = config["error"] as? Int {
+                error = config["error"] as? Int
+            } else if let errorNumber = config["error"] as? Int {
                 error = errorNumber
```
scottrhoyt: I'm pretty sure this can just be replace with:

```swift
if let warningNumber = config["warning"] as? Int {
    warning = warningNumber
}
error = config["error"] as? Int
```

What do you think?

scottrhoyt:
Thanks for the PR, it's looking good!

Source/SwiftLintFramework/Rules/RuleConfigs/SeverityLevelsConfig.swift
jpsim: Yes, that's what I tried to convey in my original message.

robinkunde: I thought about it, but it would change the behavior of the method in cases of config:

- being an empty dictionary
- containing keys besides 'warning' and 'error' (because of typos, for example)

The first one can be addressed easily enough with a where clause. The second case might be ok too, since we're documenting this as a breaking change. However, since there's no formal config validator, I wasn't quite comfortable going that route.

What do you think?

jpsim: An empty dictionary or dictionaries containing keys other than `error` and `warning` are configuration errors. I added validation for this in #438.

jpsim:
Thanks for your work here, @robinkunde. Merging in #438.

PR #433, opened 2016-01-28
Fix Makefile regex to grep compilation time output correctly

toshi0383:
Now we can catch e.g. "10.0ms", but not "0.0ms".


jpsim:
Thanks @toshi0383, nice catch! Will merge in #442.

Issue #392, opened 2016-01-21
Improve output of "rules" command


nmccann:
I'm wondering if there are currently any plans to improve the output of the `swiftlint rules` command to help with discoverability. While trying to integrate with SwiftLint, I found myself constantly referring to the source code of the various rules to determine what their defaults were, what the triggered or didn't trigger them etc. The recent addition of opt-in rules complicates things even further.

The current output of the rules command is limited to the name of the rule. It's not clear whether or not the rule is enabled by default, or if it has to be opted into.

Some proposals, roughly ordered in (estimated) level of effort:
* Group opt-in rules and display them in a way that is makes them visually distinct from default rules
* Show description of rule after it's identifier
* Show parameters for configurable rules
** Alternatively, make it obvious whether or not a rule is configurable and require that users drill down for more details with something like `swiftlint rules variable_name_min_length  --help`

If this is something that would be useful, I'd be interested in helping out.

Thank you for your time,
Noah

marcelofabri:
I like how [fastlane](https://github.com/fastlane/fastlane) handles this:

```
$ fastlane actions | head
+--------------------------------+-------------------------------------------------------------------------------+---------------------------+
|                                                         Available fastlane actions                                                         |
+--------------------------------+-------------------------------------------------------------------------------+---------------------------+
| Action                         | Description                                                                   | Author                    |
+--------------------------------+-------------------------------------------------------------------------------+---------------------------+
| add_git_tag                    | This will add an annotated git tag to the current branch                      | lmirosevic                |
| appaloosa                      | Upload your app to Appaloosa Store                                            | Appaloosa                 |
| appetize                       | Create or Update apps on Appetize.io                                          | giginet                   |

...

| xcodebuild                     | Use the `xcodebuild` command to build and sign your app                       | dtrenz                    |
| xctest                         | Runs tests on the given simulator                                             | dtrenz                    |
| xctool                         | Run tests using xctool                                                        | KrauseFx                  |
+--------------------------------+-------------------------------------------------------------------------------+---------------------------+
  Total of 140 actions

Get more information for one specific action using `fastlane action [name]`
```

And then you can run `fastlane action [name]`:

```
fastlane action create_pull_request
Loading documentation for create_pull_request:

+-----------------------------------------------+
|              create_pull_request              |
+-----------------------------------------------+
| This will create a new pull request on GitHub |
|                                               |
| Created by seei                               |
+-----------------------------------------------+

[...]

PR #435, opened 2016-01-28
tabular rules command

jpsim:
Just an early proof of concept to discuss what we want out of #392. Currently outputs this:

```
+-----------------------------+--------+---------------------------------------------------------------------------------------------------------+
| identifier                  | opt-in | description                                                                                             |
+-----------------------------+--------+---------------------------------------------------------------------------------------------------------+
| closing_brace               | no     | Closing brace with closing parenthesis should not have any whitespaces in the middle.                   |
| colon                       | no     | Colons should be next to the identifier when specifying a type.                                         |
| comma                       | no     | There should be no space before and one after any comma.                                                |
| conditional_binding_cascade | no     | Repeated `let` statements in conditional binding cascade should be avoided.                             |
| control_statement           | no     | if,for,while,do statements shouldn't wrap their conditionals in parentheses.                            |
| custom_rules                | no     | Create custom rules by providing a regex string. Optionally specify what syntax kinds to match again... |
| cyclomatic_complexity       | no     | Complexity of function bodies should be limited.                                                        |
| empty_count                 | yes    | Prefer checking `isEmpty` over comparing `count` to zero.                                               |
| file_length                 | no     | Files should not span too many lines.                                                                   |
| force_cast                  | no     | Force casts should be avoided.                                                                          |
| force_try                   | no     | Force tries should be avoided.                                                                          |
| force_unwrapping            | yes    | Force unwrapping should be avoided.                                                                     |
| function_body_length        | no     | Functions bodies should not span too many lines.                                                        |
| leading_whitespace          | no     | Files should not contain leading whitespace.                                                            |
| legacy_constant             | no     | Struct-scoped constants are preferred over legacy global constants.                                     |
| legacy_constructor          | no     | Swift constructors are preferred over legacy convenience functions.                                     |
| line_length                 | no     | Lines should not span too many characters.                                                              |
| missing_docs                | yes    | Public declarations should be documented.                                                               |
| nesting                     | no     | Types should be nested at most 1 level deep, and statements should be nested at most 5 levels deep.     |
| opening_brace               | no     | Opening braces should be preceded by a single space and on the same line as the declaration.            |
| operator_whitespace         | no     | Operators should be surrounded by a single whitespace when defining them.                               |
| return_arrow_whitespace     | no     | Return arrow and return type should be separated by a single space or on a separate line.               |
| statement_position          | no     | Else and catch should be on the same line, one space after the previous declaration.                    |
| todo                        | no     | TODOs and FIXMEs should be avoided.                                                                     |
| trailing_newline            | no     | Files should have a single trailing newline.                                                            |
| trailing_semicolon          | no     | Lines should not have trailing semicolons.                                                              |
| trailing_whitespace         | no     | Lines should not have trailing whitespace.                                                              |
| type_body_length            | no     | Type bodies should not span too many lines.                                                             |
| type_name                   | no     | Type name should only contain alphanumeric characters, start with an uppercase character and span be... |
| valid_docs                  | no     | Documented declarations should be valid.                                                                |
| variable_name               | no     | Variable names should only contain alphanumeric characters and start with a lowercase character or s... |
+-----------------------------+--------+---------------------------------------------------------------------------------------------------------+
```

I'm putting this up so we can discuss the direction of what we want this to look like. It's too early to review the implementation, so please don't comment on that :grimacing:.

A few open questions:

1. How should we render the configurability of rules, if at all?
2. How should we take into account the `.swiftlint.yml` configuration, if at all?
3. What columns are missing from this table?
4. Should we allow CLI flags to be passed to the `rules` command, if so which ones?
5. Is a table the right format for displaying this information, if so is this styling appropriate?
6. Should the output adapt to the width of the user's console? It currently doesn't.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,11 @@
   [Scott Hoyt](https://github.com/scottrhoyt)
   [#355](https://github.com/realm/SwiftLint/issues/355)
 
+* The `rules` command now prints a table containing values for: `identifier`,
+  `opt-in`, `correctable`, `enabled in your config` & `configuration`.  
+  [JP Simard](https://github.com/jpsim)
+  [#392](https://github.com/realm/SwiftLint/issues/392)
+
 ##### Bug Fixes
 
 * Fix more false positives in `ValidDocsRule`.  
--- a/Source/SwiftLintFramework/Protocols/Rule.swift
+++ b/Source/SwiftLintFramework/Protocols/Rule.swift
@@ -32,7 +32,11 @@ public protocol ConfigurableRule: Rule {
     func isEqualTo(rule: ConfigurableRule) -> Bool
 }
 
-public protocol ConfigProviderRule: ConfigurableRule {
+public protocol _ConfigProviderRule: ConfigurableRule {
+    var configDescription: String { get }
+}
+
+public protocol ConfigProviderRule: _ConfigProviderRule {
     typealias ConfigType: RuleConfig
     var config: ConfigType { get set }
 }
@@ -54,6 +58,10 @@ public extension ConfigProviderRule {
             return config.isEqualTo(rule.config)
         }
         return false
+    }
+
+    public var configDescription: String {
+        return config.consoleDescription
     }
 }
 
--- a/Source/SwiftLintFramework/Protocols/RuleConfig.swift
+++ b/Source/SwiftLintFramework/Protocols/RuleConfig.swift
@@ -11,6 +11,7 @@ import Foundation
 public protocol RuleConfig {
     mutating func setConfig(config: AnyObject) throws
     func isEqualTo(ruleConfig: RuleConfig) -> Bool
+    var consoleDescription: String { get }
 }
 
 extension RuleConfig where Self: Equatable {
--- a/Source/SwiftLintFramework/Rules/CustomRules.swift
+++ b/Source/SwiftLintFramework/Rules/CustomRules.swift
@@ -12,6 +12,7 @@ import SourceKittenFramework
 // MARK: - CustomRulesConfig
 
 public struct CustomRulesConfig: RuleConfig, Equatable {
+    public var consoleDescription: String { return "N/A" }
     public var customRuleConfigs = [RegexConfig]()
 
     public init() {}
--- a/Source/SwiftLintFramework/Rules/RuleConfigs/NameConfig.swift
+++ b/Source/SwiftLintFramework/Rules/RuleConfigs/NameConfig.swift
@@ -9,6 +9,11 @@
 import Foundation
 
 public struct NameConfig: RuleConfig, Equatable {
+    public var consoleDescription: String {
+        return "(min_length) \(minLength.shortConsoleDescription), " +
+            "(max_length) \(maxLength.shortConsoleDescription)"
+    }
+
     var minLength: SeverityLevelsConfig
     var maxLength: SeverityLevelsConfig
     var excluded: Set<String>
--- a/Source/SwiftLintFramework/Rules/RuleConfigs/RegexConfig.swift
+++ b/Source/SwiftLintFramework/Rules/RuleConfigs/RegexConfig.swift
@@ -21,6 +21,10 @@ public struct RegexConfig: RuleConfig, Equatable {
         return severityConfig.severity
     }
 
+    public var consoleDescription: String {
+        return "\(severity.rawValue.lowercaseString): \(regex.pattern)"
+    }
+
     public var description: RuleDescription {
         return RuleDescription(identifier: identifier,
             name: name ?? identifier,
--- a/Source/SwiftLintFramework/Rules/RuleConfigs/SeverityConfig.swift
+++ b/Source/SwiftLintFramework/Rules/RuleConfigs/SeverityConfig.swift
@@ -9,6 +9,10 @@
 import Foundation
 
 public struct SeverityConfig: RuleConfig, Equatable {
+    public var consoleDescription: String {
+        return severity.rawValue.lowercaseString
+    }
+
     var severity: ViolationSeverity
 
     public init(_ severity: ViolationSeverity) {
--- a/Source/SwiftLintFramework/Rules/RuleConfigs/SeverityLevelsConfig.swift
+++ b/Source/SwiftLintFramework/Rules/RuleConfigs/SeverityLevelsConfig.swift
@@ -9,6 +9,23 @@
 import Foundation
 
 public struct SeverityLevelsConfig: RuleConfig, Equatable {
+    public var consoleDescription: String {
+        let errorString: String
+        if let errorValue = error {
+            errorString = ", error: \(errorValue)"
+        } else {
+            errorString = ""
+        }
+        return "warning: \(warning)" + errorString
+    }
+
+    public var shortConsoleDescription: String {
+        if let errorValue = error {
+            return "w/e: \(warning)/\(errorValue)"
+        }
+        return "w: \(warning)"
+    }
+
     var warning: Int
     var error: Int?
 
--- a/Source/swiftlint/Commands/RulesCommand.swift
+++ b/Source/swiftlint/Commands/RulesCommand.swift
@@ -30,24 +30,29 @@ private struct TextTableColumn {
     }
 }
 
+private func fence(strings: [String], separator: String) -> String {
+    return separator + strings.joinWithSeparator(separator) + separator
+}
+
 private struct TextTable {
     let columns: [TextTableColumn]
 
     func render() -> String {
-        let widths = columns.map({ $0.width })
-        let separator = "+" + widths.map({ width in
-            return Repeat(count: width + 2, repeatedValue: "-").joinWithSeparator("")
-        }).joinWithSeparator("+") + "+"
-        let header = "|" + columns.map({
-            " \($0.header.withPadding($0.width)) "
-        }).joinWithSeparator("|") + "|"
-        let result = separator + "\n\(header)\n" + separator + "\n"
+        let joint = "+"
+        let verticalSeparator = "|"
+        let horizontalSeparator = "-"
+        let separator = fence(columns.map({ column in
+            Repeat(count: column.width + 2, repeatedValue: horizontalSeparator)
+                .joinWithSeparator("")
+        }), separator: joint)
+        let header = fence(columns.map({ " \($0.header.withPadding($0.width)) " }),
+            separator: verticalSeparator)
         let values = (0..<columns.first!.values.count).map({ rowIndex in
-            return "|" + columns.map({ column in
-                return " \(column.values[rowIndex].withPadding(column.width)) "
-            }).joinWithSeparator("|") + "|"
+            fence(columns.map({ column in
+                " \(column.values[rowIndex].withPadding(column.width)) "
+            }), separator: verticalSeparator)
         }).joinWithSeparator("\n")
-        return result + values + "\n" + separator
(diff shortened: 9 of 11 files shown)


marcelofabri:
My 2 cents:

> How should we render the configurability of rules, if at all?
I think we should have a command like `swiflint rule cyclomatic_complexity`, which would show in more details the rule, including its configuration options.

> How should we take into account the .swiftlint.yml configuration, if at all?
In the command mentioned above, we should have a column with the current value.

> What columns are missing from this table?
Maybe we could show triggering and non-triggering example on the `swiflint rule {rule}` command.

> Should we allow CLI flags to be passed to the rules command, if so which ones?
I can't think of anything.

> Is a table the right format for displaying this information, if so is this styling appropriate?
Yes

> Should the output adapt to the width of the user's console? It currently doesn't.
I don't think it's necessary.

scottrhoyt:
I this it looks good and much easier to read. To display rule configurability, I'd supplement this with a `configurable` column and `swiftlint rule <rule_name>` command for drill-down help. At least for the common `RuleConfigs` this shouldn't be that hard to implement, though we will have to come up with a way to display it nicely in a nested fashion. Perhaps even having each rule config provide a sample YAML might be the easiest first start? Going this route though we might want to give second thought to whether we want to keep the multiple config options around for severity levels. Documenting them has always been verbose.

scottrhoyt:
Another similar approach: we could add a `RuleDocs` folder to the project. New rules must be accompanied by a `new-rule-identifier.yml` file in the rule docs. The drill down help just prints out that file in a nice way. That might be an easy solution and allow for a lot of flexibility in documentation plus copy-paste ease of use for the user.

jpsim:
Some good ideas here, thanks for the feedback @marcelofabri and @scottrhoyt!

> I think we should have a command like `swiflint rule cyclomatic_complexity`, which would show in more details the rule, including its configuration options.

This is a good idea. How would we render the configuration options, though? It's not just a [flat list of dictionary keys anymore](https://github.com/realm/SwiftLint#configuration).

> Maybe we could show triggering and non-triggering example on the `swiflint rule {rule}` command.

This is a good idea. The initial version did this, but we were a bit careless in how it was rendered, it was really difficult to read.

Ideally, we'd print them with syntax highlighting (which SourceKitten gets us) and have clear separators between them. Otherwise it all blends together and is hard to read.

I agree with your other comments @marcelofabri.

---

@scottrhoyt I'd be very hesitant to have rule documentation elsewhere than its `.swift` file. Right now, rule metadata, implementation and tests are all in one place, which is really nice IMO. It makes it easier for anyone to contribute since it's very localized.

> I'd be happy to contribute this in a separate PR, we would just need to decide if it should be part of `swiftlint` or `SwiftLintFramework`. `swiftlint` makes sense to me.

FYI: We currently only have tests for `SwiftLintFramework`, so we'd need to update the project's configuration to allow testing `swiftlint`.

scottrhoyt:
> I'd be very hesitant to have rule documentation elsewhere than its .swift file. Right now, rule metadata, implementation and tests are all in one place, which is really nice IMO. It makes it easier for anyone to contribute since it's very localized.

I guess I separated those into two cases. `RuleConfig`s should probably be better documented for consumers of `SwiftLintFramework`, but that documentation would focus on what they expect from the `AnyObject`s being passed to them. `swiftlint` should be better documented in terms of end-user YAML configuration, and that was the simplest solution I could think of for that. Now I understand that YAML parsing is actually being handled by `SwiftLintFramework` right now. However, I would be in favor of stripping YAML knowledge completely out of the framework and putting that in the user interface layer because I think that's a user interface choice and API consumers of `SwiftLintFramework` would probably rather interact with it in terms of data type instead of configuration files.

scottrhoyt:
I suppose you could support both cases by having `RuleConfig`s able to return an `AnyObject` template of their configuration and have `swiftlint` translate that into YAML for displaying to the end user.

jpsim:
One of the reasons I'd prefer keeping YAML support in SwiftLintFramework itself is to eventually support using SwiftLintFramework in Swift scripts. For example, say you wanted to have custom rules for your project/organization written in Swift but those rules didn't make sense to contribute back to the main SwiftLint project, you could use `#!/bin/env swift\nimport SwiftLintFramework` to run your own makeshift version of `swiftlint`, including using a `.swiftlint.yml` file in your project.

Just food for thought.

scottrhoyt:
I see. That use case makes a lot of sense and could be hugely valuable. But I might add that if you are going to write your script in swift and your custom rules in swift, you might also want to write your configuration in swift. This idea would be in alignment with how PackageManager works even though another format could have been chosen for the package file.

On another note, it's funny to think just how much of `NSCoding` we have been duplicating to be Swifty. Though we are in plentiful company in wrestling with how to deal with data serialization in the Swiftiest way possible. :)

jpsim:
> But I might add that if you are going to write your script in swift and your custom rules in swift, you might also want to write your configuration in swift. This idea would be in alignment with how PackageManager works even though another format could have been chosen for the package file.

That's a good point. All I'm saying is that it seems strictly positive to keep YAML parsing in SwiftLintFramework; I don't think we lose anything by keeping it in there.

Unless we move completely away from YAML and towards a `Package.swift` approach. `LintConfiguration.swift`. Would be tricky because then we'd have to compile that Swift at runtime, it gets pretty messy.

> On another note, it's funny to think just how much of NSCoding we have been duplicating to be Swifty. Though we are in plentiful company in wrestling with how to deal with data serialization in the Swiftiest way possible. :)

I don't know what you're referring to here. We haven't been serializing/deserializing anything to binary, only human-readable formats.

scottrhoyt:
I only meant that we are just dealing with the classic data serialization problem that has been solved a bunch of times before with the like of `NSCoding`/`NSKeyedArchiver`, `NSPropertyListSerialization`/`NJSONSerielization`, and many other 1st and 3rd party libraries like `YamlSwift`. Considering that I think the easiest way to output dynamic rule configurations would probably be to just serialize a template configuration (maybe even the default config) back out to YAML, it might be worth taking a look at a library that already does this.

scottrhoyt:
Also, if you were to consume 'SwiftLintFramework' directly to create a custom tool with custom rules in swift, you always have the option of initializing the rules yourself and setting their configuration in swift. That use case would be better supported by a 'Configuration' that was a bit leaner and didn't have the ties to YAML.

jpsim:
> you always have the option of initializing the rules yourself and setting their configuration in swift. That use case would be better supported by a 'Configuration' that was a bit leaner and didn't have the ties to YAML.

I agree with the first part but not the second. It's better to have both options available to you and choose the most appropriate initializer.

marcelofabri:
> This is a good idea. How would we render the configuration options, though? It's not just a flat list of dictionary keys anymore.

Maybe we should take advantage that we're not on 1.0 yet and require configurations to be explicit. This way, we'll also have more predictable configurations, making easier to render them (besides making the `.swiftlint.yml` less error prone).

But we probably would have to add methods (on `RuleConfig`?) to get documentation from each available configuration (a description and default value?).

I'm not really very familiar on how configurations work on SwiftLint, so there's a chance I wrote something non-sense. :sweat_smile:

scottrhoyt:
Yeah, basically we'd be reversing what we do now and adding serialization to YAML through `AnyObject` intermediaries. The default configurations are almost fully specified examples (`excluded`, `CustomRuleConfig`, and `RegexConfig` being notable outliers--so you'd have to deal with those cases). The upshot is that you'd also get documentation of default values which would be nice. Restricting configurations to be fully specified and eliminating some of the shortcuts (like arrays for warning and error levels and severity via a string) would make our lives easier at the expense of user experience. It would also make #438 unnecessary, but the cost is too great IMHO. Part of me still leans towards the RuleDocs solution because of its simplicity, but it has the drawback that @jpsim brings up about fracturing the self documenting nature of the rules.

jpsim:
I think I have this at a stage where it's useful, but would love to get some eyes on this. Here's what `swiftlint rules` now prints:

```
+-----------------------------+--------+-------------+------------------------+--------------------------------------------------+
| identifier                  | opt-in | correctable | enabled in your config | configuration                                    |
+-----------------------------+--------+-------------+------------------------+--------------------------------------------------+
| closing_brace               | no     | yes         | yes                    | warning                                          |
| colon                       | no     | yes         | yes                    | warning                                          |
| comma                       | no     | yes         | yes                    | warning                                          |
| conditional_binding_cascade | no     | no          | yes                    | warning                                          |
| control_statement           | no     | no          | yes                    | warning                                          |
| custom_rules                | no     | no          | yes                    | N/A                                              |
| cyclomatic_complexity       | no     | no          | yes                    | warning: 10, error: 20                           |
| empty_count                 | yes    | no          | no                     | error                                            |
| file_length                 | no     | no          | yes                    | warning: 400, error: 1000                        |
| force_cast                  | no     | no          | yes                    | error                                            |
| force_try                   | no     | no          | yes                    | error                                            |
| force_unwrapping            | yes    | no          | no                     | warning                                          |
| function_body_length        | no     | no          | yes                    | warning: 40, error: 100                          |
| function_parameter_count    | no     | no          | yes                    | warning: 5, error: 8                             |
| leading_whitespace          | no     | no          | yes                    | warning                                          |
| legacy_constant             | no     | yes         | yes                    | warning                                          |
| legacy_constructor          | no     | yes         | yes                    | warning                                          |
| line_length                 | no     | no          | yes                    | warning: 100, error: 200                         |
| missing_docs                | yes    | no          | no                     | N/A                                              |
| nesting                     | no     | no          | yes                    | warning                                          |
| opening_brace               | no     | yes         | yes                    | warning                                          |
| operator_whitespace         | no     | no          | yes                    | warning                                          |
| return_arrow_whitespace     | no     | no          | yes                    | warning                                          |
| statement_position          | no     | yes         | yes                    | warning                                          |
| todo                        | no     | no          | yes                    | warning                                          |
| trailing_newline            | no     | yes         | yes                    | warning                                          |
| trailing_semicolon          | no     | yes         | yes                    | warning                                          |
| trailing_whitespace         | no     | yes         | yes                    | warning                                          |
| type_body_length            | no     | no          | yes                    | warning: 200, error: 350                         |
| type_name                   | no     | no          | yes                    | (min_length) w/e: 3/0, (max_length) w/e: 40/1000 |
| valid_docs                  | no     | no          | yes                    | warning                                          |
| variable_name               | no     | no          | yes                    | (min_length) w/e: 3/2, (max_length) w/e: 40/60   |
+-----------------------------+--------+-------------+------------------------+--------------------------------------------------+
```

I opted not to print the rule description in this table because 1) it's printed as part of `swiftlint rules <rule id>` since this morning (#456) 2) it's quite long and hard to in table form read and 3) printing multiline strings in a text table like this makes the code way more complex (see f715252).

It was also important to me to keep the output within a certain width. It's currently 131 characters wide, which should display fine on most screens, although I had to make some concessions (`(min_length) w/e: 3/2, (max_length) w/e: 40/60` is certainly difficult to read).

scottrhoyt:
@jpsim, I think this is a great improvement! I'm in favor.

scottrhoyt:
Also, a quick search didn't reveal any commonly available libraries for tabular console output in Swift. Might be an interesting side project. :)

jpsim:
Thanks for your feedback @scottrhoyt. I'm not 100% happy with this, but let's get this in on the basis of being better than what we have now, and maybe this code being icky will motivate someone to improve it later.

I'll add a changelog entry and get this in.

jpsim:
It'd be great to have unit tests for the `swiftlint` target so we could have tested this... It's definitely silly to turn on test code coverage generation for the target since it's not tested at all.

scottrhoyt:
:+1:

PR #436, opened 2016-01-28
Add whitelist only rules

daniel-beard:
- This is my proposal for adding a `whitelist_rules` option to the `.swiftlint.yml` configuration.
- The idea is that if the `whitelist_rules` option appears, it overrides all other rule configurations, and *only* the rules specified in the `whitelist_rules` array are enabled.
- This is intended for environments that have very specific Swift style guide rules, and where updating SwiftLint currently introduces a number of rules (that may not have been agreed on) and breaks the CI process.
- This is a proposal, I'm open to other ideas for naming or implementation, thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@
 
 ##### Enhancements
 
-* None.
+* Add `whitelist_rules` rule whitelists in config files  
+  [Daniel Beard](https://github.com/daniel-beard)
+  [#256](https://github.com/realm/SwiftLint/issues/256)
 
 ##### Bug Fixes
 
--- a/README.md
+++ b/README.md
@@ -134,12 +134,18 @@ identifiers.
 Configure SwiftLint by adding a `.swiftlint.yml` file from the directory you'll
 run SwiftLint from. The following parameters can be configured:
 
+Rule inclusion:
+- `disabled_rules`: Disable rules from the default enabled set.
+- `opt_in_rules`: Some rules are opt-in.
+- `whitelist_rules`: Can not be specified alongside `disabled_rules` or
+`opt_in_rules`. Acts as a whitelist, only the rules specified in this list will be enabled.
+
 ```yaml
 disabled_rules: # rule identifiers to exclude from running
   - colon
   - comma
   - control_statement
-enabled_rules: # some rules are only opt-in
+opt_in_rules: # some rules are only opt-in
   - empty_count
   - missing_docs
   # Find all the available rules by running:
--- a/Source/SwiftLintFramework/Models/Configuration.swift
+++ b/Source/SwiftLintFramework/Models/Configuration.swift
@@ -38,7 +38,8 @@ public struct Configuration: Equatable {
     }
 
     public init?(disabledRules: [String] = [],
-                 enabledRules: [String] = [],
+                 optInRules: [String] = [],
+                 whitelistRules: [String] = [],
                  included: [String] = [],
                  excluded: [String] = [],
                  reporter: String = "xcode",
@@ -78,17 +79,44 @@ public struct Configuration: Equatable {
         }
         self.disabledRules = validDisabledRules
 
-        self.rules = rules.filter { rule in
-            let id = rule.dynamicType.description.identifier
-            if validDisabledRules.contains(id) { return false }
-            return enabledRules.contains(id) || !(rule is OptInRule)
+        // white_list rules take precendence over all else.
+        if !whitelistRules.isEmpty {
+
+            if !disabledRules.isEmpty || !optInRules.isEmpty {
+                queuedPrintError("'disabled_rules' or 'opt_in_rules' cannot be used in " +
+                    "combination with 'whitelist_rules'")
+                return nil
+            }
+
+            self.rules = rules.filter { rule in
+                let id = rule.dynamicType.description.identifier
+                return whitelistRules.contains(id)
+            }
+        } else {
+            self.rules = rules.filter { rule in
+                let id = rule.dynamicType.description.identifier
+                if validDisabledRules.contains(id) { return false }
+                return optInRules.contains(id) || !(rule is OptInRule)
+            }
         }
     }
 
     public init?(dict: [String: AnyObject]) {
+
+        // Deprecation warning for "enabled_rules"
+        if dict["enabled_rules"] != nil {
+            queuedPrint("'enabled_rules' has been renamed to 'opt_in_rules' " +
+                "and will be completely removed in a future release.")
+        }
+
+        // Use either new 'opt_in_rules' or deprecated 'enabled_rules' for now.
+        let optInRules = dict["opt_in_rules"] as? [String] ??
+            dict["enabled_rules"] as? [String] ?? []
+
         self.init(
             disabledRules: dict["disabled_rules"] as? [String] ?? [],
-            enabledRules: dict["enabled_rules"] as? [String] ?? [],
+            optInRules: optInRules,
+            whitelistRules: dict["whitelist_rules"] as? [String] ?? [],
             included: dict["included"] as? [String] ?? [],
             excluded: dict["excluded"] as? [String] ?? [],
             reporter: dict["reporter"] as? String ?? XcodeReporter.identifier,
--- a/Source/SwiftLintFrameworkTests/ConfigurationTests.swift
+++ b/Source/SwiftLintFrameworkTests/ConfigurationTests.swift
@@ -32,6 +32,36 @@ class ConfigurationTests: XCTestCase {
         XCTAssertEqual(config.reporterFromString.identifier, "xcode")
     }
 
+    func testWhitelistRules() {
+        let whitelist = ["nesting", "todo"]
+        let config = Configuration(dict: ["whitelist_rules":  whitelist])!
+        let configuredIdentifiers = config.rules.map {
+            $0.dynamicType.description.identifier
+        }
+        XCTAssertEqual(whitelist, configuredIdentifiers)
+    }
+
+    func testOtherRuleConfigurationsAlongsideWhitelistRules() {
+        let whitelist = ["nesting", "todo"]
+        let enabledRulesConfigDict = [
+            "opt_in_rules": ["line_length"],
+            "whitelist_rules": whitelist
+        ]
+        let disabledRulesConfigDict = [
+            "disabled_rules": ["variable_name"],
+            "whitelist_rules": whitelist
+        ]
+        let combinedRulesConfigDict = enabledRulesConfigDict.reduce(disabledRulesConfigDict) {
+            var d = $0; d[$1.0] = $1.1; return d
+        }
+        var config = Configuration(dict: enabledRulesConfigDict)
+        XCTAssertNil(config)
+        config = Configuration(dict: disabledRulesConfigDict)
+        XCTAssertNil(config)
+        config = Configuration(dict: combinedRulesConfigDict)
+        XCTAssertNil(config)
+    }
+
     func testDisabledRules() {
         let disabledConfig = Configuration(dict: ["disabled_rules":  ["nesting", "todo"]])!
         XCTAssertEqual(disabledConfig.disabledRules,
--- a/Source/SwiftLintFrameworkTests/TestHelpers.swift
+++ b/Source/SwiftLintFrameworkTests/TestHelpers.swift
@@ -22,13 +22,6 @@ func violations(string: String, config: Configuration = Configuration()) -> [Sty
     return Linter(file: file, configuration: config).styleViolations
 }
 
-private func violations(string: String, _ description: RuleDescription) -> [StyleViolation] {
-    let disabledRules = allRuleIdentifiers.filter { $0 != description.identifier }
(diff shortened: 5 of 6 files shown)


Source/SwiftLintFrameworkTests/ConfigurationTests.swift
```diff
@@ -32,6 +32,17 @@ class ConfigurationTests: XCTestCase {
         XCTAssertEqual(config.reporterFromString.identifier, "xcode")
     }
 
+    func testWhitelistRules() {
```
jpsim: This test is doing a lot of unnecessary work. The following is equivalent:

```swift
let whilelist = ["nesting", "todo"]
let config = Configuration(dict: ["whitelist_rules":  whitelist])!
let configuredIdentifiers = config.rules.map {
    $0.dynamicType.description.identifier
}
XCTAssertEqual(whitelist, configuredIdentifiers)
```

daniel-beard: @jpsim updated

jpsim:
I'm fine with this design. I think the tricky part is documentation, because now we have `enabled_rules`, `disabled_rules` and `whitelist_rules`, which can easily lead to confusion.

Maybe the solution is simply to reconsider the naming of these YAML keys to something more obvious (not that I have any suggestions).

Or maybe I'm overestimating the potential for confusion?

At the very least, I think swiftlint should fail hard if `whitelist_rules` is set in combination with `enabled_rules` or `disabled_rules`, since that's certainly an error.

daniel-beard:
I agree it's confusing, if anyone has any different naming suggestions that would be great.

daniel-beard:
I'll work on updating so that if `enabled_rules` or `disabled_rules` is set, we fail.

scottrhoyt:
Here's my two cents:

* 3 valid configurations
  * `disabled_rules` (blacklist)
  * `disabled_rules` + `opt_in_rules` (blacklist + opt in)
  * `enabled_rules` (whitelist)

jpsim:
> * 3 valid configurations
>  * `disabled_rules` (blacklist)
>  * `disabled_rules` + `opt_in_rules` (blacklist + opt in)
>  * `enabled_rules` (whitelist)

A few comments:

* `disabled` and `enabled` are obvious counterparts, which is why I prefer it over `disabled` and `opt_in`
* it's perfectly valid to have no `disabled_rules` but have `enabled_rules`, so that's also a valid configuration

scottrhoyt:
Very true on the second point, I'll update that list.

For the first point, because `disabled_rules` is a blacklist (disregarding even opt in rules), it makes sense to me to have its counterpart, `enabled_rules`, be the whitelist. So `opt_in_rules` was just my suggestion to take over the role `enabled_rules` previously had.

jpsim:
Fair enough. `disabled_rules`, `opt_in_rules` and `whitelist_rules` now seems most appropriate.

It'd be interesting to see what other linters do (e.g. rubocop) and to bring this up in the SwiftLint Slack channel to check in with users.

daniel-beard:
@jpsim I updated the names to `disabled_rules`, `opt_in_rules` and `whitelist_rules`

Source/SwiftLintFrameworkTests/TestHelpers.swift
```diff
@@ -25,7 +25,7 @@ func violations(string: String, config: Configuration = Configuration()) -> [Sty
 private func violations(string: String, _ description: RuleDescription) -> [StyleViolation] {
```
jpsim: this whole function can now be simplified to just this:

```swift
private func violations(string: String, _ description: RuleDescription) -> [StyleViolation] {
    return violations(string, config: Configuration(whitelistRules: [description.identifier])!)
}
```

Actually, it's so simple now, I'd actually vote for removing this function altogether:

```diff
diff --git a/Source/SwiftLintFrameworkTests/TestHelpers.swift b/Source/SwiftLintFrameworkTests/TestHelpers.swift
index 7da6d1f..becdaa3 100644
--- a/Source/SwiftLintFrameworkTests/TestHelpers.swift
+++ b/Source/SwiftLintFrameworkTests/TestHelpers.swift
@@ -22,13 +22,6 @@ func violations(string: String, config: Configuration = Configuration()) -> [Sty
     return Linter(file: file, configuration: config).styleViolations
 }
 
-private func violations(string: String, _ description: RuleDescription) -> [StyleViolation] {
-    let disabledRules = allRuleIdentifiers.filter { $0 != description.identifier }
-    let enabledRules = allRuleIdentifiers.filter { $0 == description.identifier }
-    let config = Configuration(disabledRules: disabledRules, optInRules: enabledRules)!
-    return violations(string, config: config)
-}
-
 private func assertCorrection(before: String, expected: String) {
     guard let path = NSURL(fileURLWithPath: NSTemporaryDirectory(), isDirectory: true)
         .URLByAppendingPathComponent(NSUUID().UUIDString + ".swift").path else {
@@ -63,15 +56,16 @@ extension String {
 extension XCTestCase {
     func verifyRule(ruleDescription: RuleDescription, commentDoesntViolate: Bool = true,
                     stringDoesntViolate: Bool = true) {
+        let config = Configuration(whitelistRules: [ruleDescription.identifier])!
         let triggers = ruleDescription.triggeringExamples
         let nonTriggers = ruleDescription.nonTriggeringExamples
 
         // Non-triggering examples don't violate
-        XCTAssert(nonTriggers.flatMap({ violations($0, ruleDescription) }).isEmpty)
+        XCTAssert(nonTriggers.flatMap({ violations($0, config: config) }).isEmpty)
 
         var violationsCount = 0
         for trigger in triggers {
-            let triggerViolations = violations(trigger, ruleDescription)
+            let triggerViolations = violations(trigger, config: config)
             violationsCount += triggerViolations.count
             // Triggering examples with violation markers violate at the marker's location
             let markerLocation = (trigger as NSString).rangeOfString(violationMarker).location
@@ -86,19 +80,19 @@ extension XCTestCase {
 
         // Comment doesn't violate
         XCTAssertEqual(
-            triggers.flatMap({ violations("/*\n  " + $0 + "\n */", ruleDescription) }).count,
+            triggers.flatMap({ violations("/*\n  " + $0 + "\n */", config: config) }).count,
             commentDoesntViolate ? 0 : triggers.count
         )
 
         // String doesn't violate
         XCTAssertEqual(
-            triggers.flatMap({ violations($0.toStringLiteral(), ruleDescription) }).count,
+            triggers.flatMap({ violations($0.toStringLiteral(), config: config) }).count,
             stringDoesntViolate ? 0 : triggers.count
         )
 
         // "disable" command doesn't violate
         let command = "// swiftlint:disable \(ruleDescription.identifier)\n"
-        XCTAssert(triggers.flatMap({ violations(command + $0, ruleDescription) }).isEmpty)
+        XCTAssert(triggers.flatMap({ violations(command + $0, config: config) }).isEmpty)
 
         // corrections
         ruleDescription.corrections.forEach(assertCorrection)
```

daniel-beard: Will do

daniel-beard:
@jpsim updated.

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -78,17 +79,33 @@ public struct Configuration: Equatable {
...
-        self.rules = rules.filter { rule in
-            let id = rule.dynamicType.description.identifier
-            if validDisabledRules.contains(id) { return false }
-            return enabledRules.contains(id) || !(rule is OptInRule)
+        // Whitelist rules take precendence over all else.
+        if !whitelistRules.isEmpty {
+
+            // Error if enabled_rules or disabled_rules is set while whitelist_rules are.
+            if !disabledRules.isEmpty || !optInRules.isEmpty {
+                queuedPrintError("Other rulesets can not exist alongside whitelist_rules")
```
jpsim: I think explicitly referencing `disabled_rules` and `opt_in_rules` would be clearer:

> 'disabled_rules' or 'opt_in_rules' cannot be used in combination with 'whitelist_rules'

daniel-beard:
@jpsim updated error message.

jpsim:
If we rename `enabled_rules` to `opt_in_rules`, it'd be nice to start with just a deprecation, followed by a complete removal in a future version. This project's own `.swiftlint.yml` should be updated to use `opt_in_rules` too.

A changelog entry and documentation in the readme to reflect these changes will eventually be needed, but I understand if you'd prefer waiting until the dust settles a bit more on this PR before diving into that.

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -78,17 +79,45 @@ public struct Configuration: Equatable {
...
+                return optInRules.contains(id) || !(rule is OptInRule)
+            }
         }
     }
 
     public init?(dict: [String: AnyObject]) {
+
+        // Deprecation warning for "enabled_rules"
+        if dict["enabled_rules"] != nil {
+            queuedPrint("'enabled_rules' is deprecated, please see 'opt_in_rules' " +
```
daniel-beard: Wasn't sure if this was the best way to convey a deprecation warning.

README.md
```diff
@@ -134,12 +134,18 @@ identifiers.
 Configure SwiftLint by adding a `.swiftlint.yml` file from the directory you'll
 run SwiftLint from. The following parameters can be configured:
 
+Rule inclusion:
```
daniel-beard: I'm sure there is probably a better way to list these rules, I'm bad at describing things :(

daniel-beard: Open to suggestions.

jpsim: Seems fine to me!

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -78,17 +79,45 @@ public struct Configuration: Equatable {
...
         self.disabledRules = validDisabledRules
 
-        self.rules = rules.filter { rule in
-            let id = rule.dynamicType.description.identifier
-            if validDisabledRules.contains(id) { return false }
-            return enabledRules.contains(id) || !(rule is OptInRule)
+        // white_list rules take precendence over all else.
+        if !whitelistRules.isEmpty {
+
+            // Error if opt_in_rules or disabled_rules is set while whitelist_rules are.
```
jpsim: comment is superfluous given the error message belwo

daniel-beard: Removed.

daniel-beard:
Uh, I think codecov went down because I removed a comment. That doesn't seem right...

jpsim:
Don't worry about the `codecov/project` check, it's unimportant. More of a guide than a requirement.

daniel-beard:
Ah, good.

jpsim:
This looks good to me. @scottrhoyt @norio-nomura @marcelofabri any objections or feedback on this?

The only thing left is to add a changelog entry.

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -78,17 +79,44 @@ public struct Configuration: Equatable {
...
+            }
         }
     }
 
     public init?(dict: [String: AnyObject]) {
+
+        // Deprecation warning for "enabled_rules"
+        if dict["enabled_rules"] != nil {
+            queuedPrint("'enabled_rules' has been renamed to 'opt_in_rules' " +
+                "and will be completely removed in a future release.")
```
scottrhoyt: This is a tiny nitpick, but might be better to say something like "the behavior of `enabled_rules` will be changing in a future release. Use `opt_in_rules` instead.`" I only say this because it might be confusing for someone to think `enabled_rules` is going away completely and then is told to use it for something else. Unless that is we decided to use a different label for the whitelist altogether?

scottrhoyt:
As soon as I realized we were going with `whitelist_rules`, I realized this looks great. :)

daniel-beard:
Added a changelog entry.

CHANGELOG.md
```diff
@@ -9,7 +9,9 @@
 
 ##### Enhancements
 
-* None.
+* Add `whitelist_rules` rule whitelists in config files  
+  [Daniel Beard](https://github.com/daniel-beard)
+  [#436](https://github.com/realm/SwiftLint/issues/436)
```
jpsim: This should be the issue number for the issue this PR resolves, not the PR number. This is explained in CONTRIBUTING.md.

jpsim:
@daniel-beard once you've addressed the last minor issue in the changelog, could you please squash these commits? Then we'll merge this.

scottrhoyt:
Thanks for the PR, @daniel-beard !

daniel-beard:
@jpsim, should I point to this issue? #256 Or open a new one?

daniel-beard:
Also, do you want a new branch with squashed commits, or is it ok to force-push?

jpsim:
Please point to #256 (I've just reopened it). Also yes please force-push to this branch.

daniel-beard:
Updated and squashed.

jpsim:
Thanks for pushing this through @daniel-beard! :clap: :tada:

daniel-beard:
Looking forward to contributing more in the future! :D

Issue #415, opened 2016-01-25
Limit number of method/closure parameters rule


garnett:
Time to time I raise this issue with colleagues: there is no proven number to stick, but IMHO anything with 5 or more parameters should be considered as suspicious (might be a good place to use "Introduce Parameter object" refactoring).

I see it as an AST rule that raises warning for methods/functions/closures with 5 or more arguments (configurable). 

Edit: probably it should ignore `init` methods though (value objects usually have tons of parameters)

scottrhoyt:
Interesting idea. Do you think this should or shouldn't count parameters with default values?

garnett:
@scottrhoyt that's a good question indeed :+1: 
Just thinking out loud: declaration of method with 4+ parameters is not often a problem, developer has hard times when he reads place where such method is called or invokes such method. Assuming that I would say that we can:
* make requirement less strict for method containing optional arguments (math to be discussed)
* completely ignore parameters with default values because they can be totally ignored when calling method

scottrhoyt:
Without thinking about it too much, I think I would vote for option number 2 since it is simpler and clearer. Ultimately I think it is more important for this rule to protect potential callers as opposed to the function writer.

PR #447, opened 2016-01-29
Limit parameters count

garnett:
Very early version of #415.
Implementation depends on answer here: https://github.com/jpsim/SourceKitten/issues/151.


Source/SwiftLintFramework/Rules/ParametersListLengthRule.swift
```diff
@@ -0,0 +1,88 @@
...
+            }
+
+            let substructure = dictionary["key.substructure"] as? [SourceKitRepresentable] ?? []
+            let paramsCount = substructure.filter {
+                guard let subDict = $0 as? [String: SourceKitRepresentable],
+                    key = subDict["key.kind"] as? String  else {
+                        return false
+                }
+
+                return key == "source.lang.swift.decl.var.parameter"
```
jpsim: You could avoid the hardcoded string literal by doing `return SwiftDeclarationKind(rawValue: key) == . VarParameter` instead.

jpsim:
I'm excited to see you getting started on this! Let me know if you have any questions moving forward.

Source/SwiftLintFramework/Rules/ParametersListLengthRule.swift
garnett: nice, had no idea there were constants for that :)

garnett:
@jpsim assuming my code example here https://github.com/jpsim/SourceKitten/issues/151#issuecomment-176894084 do you think there are "good" :tm: way to extract function parameters from AST?

I'm inclining to simply get function declaration string and parse it manually. Moreover I have to do it anyway because of default parameters (they are not represented in SourceKit AFAIK

jpsim:
> @jpsim assuming my code example here jpsim/SourceKitten#151 (comment) do you think there are "good"  way to extract function parameters from AST?

Yeah, it's spot on. You're using an `ASTRule` and only continuing the validation process when substructures have `.VarParameter`s. :+1:

garnett:
@jpsim but apparently it triggers false positive for this simple snippet:
```
private func foo(dictionary: [Int: Int]) -> Int {
    dictionary.flatMap { foo in return foo }
    return 1
}
```

produces 1 level-nested structure which will count ```foo``` parameters and return 2 not 1:

  "key.substructure" : [
    {
      "key.kind" : "source.lang.swift.decl.function.free",
      "key.offset" : 22,
      "key.nameoffset" : 27,
      "key.namelength" : 27,
      "key.bodyoffset" : 63,
      "key.bodylength" : 73,
      "key.accessibility" : "source.lang.swift.accessibility.private",
      "key.substructure" : [
        {
          "key.offset" : 31,
          "key.nameoffset" : 0,
          "key.namelength" : 0,
          "key.length" : 10,
          "key.typename" : "[Int: Int]",
          "key.name" : "dictionary"
        },
        {
          "key.offset" : 103,
          "key.nameoffset" : 0,
          "key.namelength" : 0,
          "key.length" : 3,
          "key.typename" : "foo",
          "key.name" : "foo"
        }
      ],
      "key.name" : "foo(_:)",
      "key.length" : 115
    }
  ],
  "key.offset" : 0,
  "key.length" : 138
}

jpsim:
In your example, `foo` *is* a parameter. It's a parameter to the closure. If you want to filter just parameters to *functions*, you can filter `.VarParameter`s that have a `key.offset` between its enclosing function's `key.offset` and `key.bodyoffset`.

In your example, you'd only end up with `dictionary` as parameters for functions because its `key.offset` is 17, which falls between its parent function's `key.offset` of 8 and `key.bodyoffset` of 49. Whereas `foo` isn't considered because its `key.offset` is 71, which is outside the 8..<49 range.

garnett:
@jpsim thanks for this :+1:  I had a feeling I'm missing something

Source/SwiftLintFramework/Rules/ParametersListLengthRule.swift
```diff
@@ -0,0 +1,105 @@
...
+            }
+
+            return []
+    }
+
+    private func allFunctionParameters(structure: [SourceKitRepresentable],
+        range: Range<Int>) -> Int {
+
+            var count = 0
+            for e in structure {
```
garnett: usage of `for` loop is intentional it allows to return early in case method reached end of current function declaration

Source/SwiftLintFramework/Rules/ParametersListLengthRule.swift
```diff
@@ -0,0 +1,105 @@
...
+
+                if SwiftDeclarationKind(rawValue: key) == .VarParameter {
+                    count += 1
+                }
+            }
+            return count
+    }
+
+    private func defaultFunctionParameters(file: File, range: Range<Int>) -> Int {
+        let funcDeclaration = file.contents[range]
```
garnett: based on assumption default parameter has `=` indicating it's value

garnett:
@jpsim ready to be reviewed!

Source/SwiftLintFramework/Rules/ParametersListLengthRule.swift
```diff
@@ -0,0 +1,105 @@
...
+            let allParameters = allFunctionParameters(substructure, range: funcDeclarationRange)
+            let defaultParameters = defaultFunctionParameters(file, range: funcDeclarationRange)
+
+            let parametersCount = allParameters - defaultParameters
+
+            for parameter in config.params where parametersCount > parameter.value {
+                let offset = Int(dictionary["key.offset"] as? Int64 ?? 0)
+                return [StyleViolation(ruleDescription: self.dynamicType.description,
+                    severity: parameter.severity,
+                    location: Location(file: file, characterOffset: offset),
```
norio-nomura: Could you please use `Location(file: file, byteOffset: offset)` instead of `Location(file: file, characterOffset: offset)`?
Because `dictionary["key.offset"]` is byte offset.

Source/SwiftLintFramework/Rules/ParametersListLengthRule.swift
norio-nomura: Could you please use `NSString.substringWithByteRange(start:, length:)` instead of subscript? Because `dictionary["key.nameoffset"]` and `dictionary["key.namelength"]` are byte offset and byte length.

Source/SwiftLintFramework/Rules/ParametersListLengthRule.swift
```diff
@@ -0,0 +1,104 @@
...
+            let defaultParams = defaultFuncParameters(file, offset: nameOffset, length: length)
+
+            let parametersCount = allParams - defaultParams
+
+            for parameter in config.params where parametersCount > parameter.value {
+                let offset = Int(dictionary["key.offset"] as? Int64 ?? 0)
+                return [StyleViolation(ruleDescription: self.dynamicType.description,
+                    severity: parameter.severity,
+                    location: Location(file: file, byteOffset: offset),
+                    reason: "{Parameters list should have \(config.warning) or less parameters: " +
```
jpsim: Typo: `{Parameters`

Source/SwiftLintFramework/Rules/CyclomaticComplexityRule.swift
```diff
@@ -43,7 +43,7 @@ public struct CyclomaticComplexityRule: ASTRule, ConfigProviderRule {
             let offset = Int(dictionary["key.offset"] as? Int64 ?? 0)
             return [StyleViolation(ruleDescription: self.dynamicType.description,
                 severity: parameter.severity,
-                location: Location(file: file, characterOffset: offset),
+                location: Location(file: file, byteOffset: offset),
```
jpsim: :+1: nice catch!

jpsim:
The implementation here looks great! I'd like us to really nail down the rule name though.

"Parameters List Length" and "Parameter List Length" seem to be used interchangeably. We're not really measuring the string length here, but rather the number of parameters.

I think "Function Parameter Count" might be a more appropriate name.

Otherwise like I said, the implementation looks spot on (as far as I can tell reveling this from my phone :wink:).

garnett:
@jpsim agreed with naming, changed

jpsim:
Thanks @garnett! I just had a few minor changes to add, so I'll merge in #459.

Really nice work on this! Not too many people have ventured into writing `ASTRule` based rules so far, so if you learned anything that you think might be useful to add in the contributing.md documentation for others looking to write rules, please let me know!

norio-nomura:
:clap:

Source/SwiftLintFramework/Rules/CyclomaticComplexityRule.swift
norio-nomura: :+1:

PR #454, opened 2016-01-30
Invalidate caches after correct file

norio-nomura:
Cache entry should be invalidate on every corrections, because caches and content may not match after corrections.


jpsim:
:+1: good thinking.

norio-nomura:
I'm sorry for ran `swiftlint autocorrect` for the first time on testing this PR. :laughing:

PR #455, opened 2016-01-30
Improve performance of `ColonRule`

norio-nomura:
The duration of `ColonRule` on linting Carthage 0.11 is reduced from 2287ms to 1581ms by Instruments.
from:
screenshot 2016-01-30 22 07 06
to:
screenshot 2016-01-30 22 07 40


jpsim:
Wow, nice work!

norio-nomura:
:pray:

Issue #461, opened 2016-02-01
Cyclomatic Complexity is improperly measured for case statements


jpsim:
The following function should have a cyclomatic complexity value of 11, but it's reported as 13.

```swift
class func colorFromCode(code: String?) -> UIColor {
    switch code! {
    case "000": return Cache.couplesColor000
    case "001": return // ...
    case "002": return // ...
    case "003": return // ...
    case "004": return // ...
    case "005": return // ...
    case "006": return // ...
    case "007": return // ...
    case "008": return // ...
    case "009": return // ...
    default: return UIColor.clearColor()
    }
}
```

This reports the following violation:
```
<nopath>:1:7: warning: Cyclomatic Complexity Violation: Function should have complexity 10 or less: currently complexity equals 13 (cyclomatic_complexity)
```

garnett:
@jpsim should be easy to fix:
* `switch` probably should not be counted. (leaves us with 12)
* in references I've read cyclomatic complexity of any function starts with 1 (in this case we can assume that there might be a condition when code in `switch` won't be executed). 

Will look into that

garnett:
From: http://gmetrics.sourceforge.net/gmetrics-CyclomaticComplexityMetric.html
`Start with a initial (default) value of one (1)`

From http://www.mccabe.com/pdf/mccabe-nist235r.pdf:

`An implicit default or fall-through branch, if specified by the language, must be taken into account when calculating complexity`

jpsim:
> Start with a initial (default) value of one (1)

I think you might be misinterpreting this. Yes, CC starts with 1 because you can't have CC of zero. Code that can take zero paths is code that never executes. It's not that you must add one to the number of paths code can take to get its cyclomatic complexity value.

garnett:
@jpsim agreed, I will remove `+1` from the rule.

jpsim:
Done in #463 thanks to @garnett

ajaybomma:
cyclomatic_complexity:
  warning: 25

I changed warning value for the cyclomatic_complexity in .yml file. Warnings gone.

superarts:
I don't think Cyclomatic Complexity is improperly measured, but in some cases, it doesn't make sense to restrict your code under complexity 10. For example, if your business logic asks you to handle 20 errors differently, it is a good practice to use exhaustive `switch` to make it incompatible to future changes, and even you write 20 handlers for these errors, you will still have higher complexity.

[...]

PR #463, opened 2016-02-01
Improve switch statement complexity measure

garnett:
Fixes issue described in https://github.com/realm/SwiftLint/issues/461

### DIFF
--- a/Source/SwiftLintFramework/Rules/CyclomaticComplexityRule.swift
+++ b/Source/SwiftLintFramework/Rules/CyclomaticComplexityRule.swift
@@ -9,6 +9,7 @@
 import Foundation
 import SourceKittenFramework
 
+
 public struct CyclomaticComplexityRule: ASTRule, ConfigProviderRule {
     public var config = SeverityLevelsConfig(warning: 10, error: 20)
 
@@ -20,13 +21,16 @@ public struct CyclomaticComplexityRule: ASTRule, ConfigProviderRule {
         description: "Complexity of function bodies should be limited.",
         nonTriggeringExamples: [
             "func f1() {\nif true {\nfor _ in 1..5 { } }\nif false { }\n}",
-            "func f3() {while true {}}",
+            "func f(code: Int) -> Int {" +
+            "switch code {\n case 0: fallthrough\ncase 0: return 1\ncase 0: return 1\n" +
+            "case 0: return 1\ncase 0: return 1\ncase 0: return 1\ncase 0: return 1\n" +
+            "case 0: return 1\ncase 0: return 1\ndefault: return 1}}"
         ],
         triggeringExamples: [
             "func f1() {\n  if true {\n    if true {\n      if false {}\n    }\n" +
-                "  }\n  if false {}\n  let i = 0\n\n  switch i {\n  case 1: break\n" +
-                "  case 2: break\n  case 3: break\n  default: break\n  }\n\n" +
-                "  for _ in 1...5 {\n    guard true else {\n      return\n    }\n  }\n}\n"
+            "  }\n  if false {}\n  let i = 0\n\n  switch i {\n  case 1: break\n" +
+            "  case 2: break\n  case 3: break\n  default: break\n  }\n\n" +
+            "  for _ in 1...5 {\n    guard true else {\n      return\n    }\n  }\n}\n"
         ]
     )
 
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Rules/CyclomaticComplexityRule.swift
```diff
@@ -36,8 +40,7 @@ public struct CyclomaticComplexityRule: ASTRule, ConfigProviderRule {
             return []
         }
 
-        let substructure = dictionary["key.substructure"] as? [SourceKitRepresentable] ?? []
-        let complexity = measureComplexity(substructure) + 1
+        let complexity = measureComplexity(file, dictionary: dictionary) + 1
```
jpsim: I think this `+ 1` is wrong. See my comment here: https://github.com/realm/SwiftLint/issues/461#issuecomment-178115875

Source/SwiftLintFramework/Rules/CyclomaticComplexityRule.swift
```diff
@@ -51,25 +54,50 @@ public struct CyclomaticComplexityRule: ASTRule, ConfigProviderRule {
...
+
+        let c = (file.contents as NSString)
+            .substringWithByteRange(start: bodyOffset, length: bodyLength) ?? ""
+
+        let fallthroughCount = c.componentsSeparatedByString("fallthrough").count - 1
+        return complexity - fallthroughCount
     }
 
     private let complexityStatements = [
         "source.lang.swift.stmt.foreach",
```
jpsim: it'd be nice to get these into SourceKitten much like `SyntaxKind` and `SwiftDeclarationKind`.

Source/SwiftLintFramework/Rules/CyclomaticComplexityRule.swift
```diff
@@ -20,13 +21,16 @@ public struct CyclomaticComplexityRule: ASTRule, ConfigProviderRule {
...
+            "switch code {\n case 0: fallthrough\ncase 0: return 1\ncase 0: return 1\n" +
+            "case 0: return 1\ncase 0: return 1\ncase 0: return 1\ncase 0: return 1\n" +
+            "case 0: return 1\ncase 0: return 1\ndefault: return 1}}"
         ],
         triggeringExamples: [
             "func f1() {\n  if true {\n    if true {\n      if false {}\n    }\n" +
-                "  }\n  if false {}\n  let i = 0\n\n  switch i {\n  case 1: break\n" +
-                "  case 2: break\n  case 3: break\n  default: break\n  }\n\n" +
-                "  for _ in 1...5 {\n    guard true else {\n      return\n    }\n  }\n}\n"
+            "  }\n  if false {}\n  let i = 0\n\n  switch i {\n  case 1: break\n" +
```
jpsim: we lost the indentation here

Source/SwiftLintFramework/Rules/CyclomaticComplexityRule.swift
```diff
@@ -20,13 +21,16 @@ public struct CyclomaticComplexityRule: ASTRule, ConfigProviderRule {
         description: "Complexity of function bodies should be limited.",
         nonTriggeringExamples: [
             "func f1() {\nif true {\nfor _ in 1..5 { } }\nif false { }\n}",
-            "func f3() {while true {}}",
+            "func f(code: Int) -> Int {" +
+            "switch code {\n case 0: fallthrough\ncase 0: return 1\ncase 0: return 1\n" +
```
jpsim: these should be indented to indicate that they're additions to the string above.

jpsim:
This could use a changelog entry, otherwise I think it's on the right track.

Source/SwiftLintFramework/Rules/CyclomaticComplexityRule.swift
garnett: Looks like good idea to me :+1: maybe I will do that too!

garnett:
@jpsim non related to this particular PR:

Do you think it is worth extracting things like `dictionary["key.substructure"] ` into dictionary extension to provide syntax like `dictionary.substructure`? Maybe this is even more related to SourceKitten.

jpsim:
> Do you think it is worth extracting things like dictionary["key.substructure"] into dictionary extension to provide syntax dictionary.substructure? Maybe this is even more related to SourceKitten.

Yeah, there's a WIP PR on SourceKitten to use proper types rather than dictionaries, which would help greatly with that.

jpsim:
This looks good to me, thanks @garnett!

jpsim:
Merging in #474.

Issue #451, opened 2016-01-30
valid_docs rule failing on `throws` inside a closure in a function.


nlutsenko:
The rule is failing on a given function declaration:
```swift
public func performThings(things: (Void throws -> Result)) -> Result
```

From the code - it looks like we are simply doing a regex capture for the `throws` signature, which is technically not correct, as the documentation should be purely on the `throws` that preceeds the function return parameters.

According to the grammar, where optional pieces are italic - it captures the one inside parameter-clauses, but should only look at the one after all of the parameter clauses.

function-signature → parameter-clauses _throws_ ­­_function-result­_

jpsim:
yup, definitely a bug. thanks for reporting!

jpsim:
Fixed in #464

PR #464, opened 2016-02-01
Fix ValidDocs rule

diogoguimaraes:
This pull request fixes another false positive in the `ValidDocsRule`: 
* `func a() throws -> Bool { return true }`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,9 @@
 
 ##### Bug Fixes
 
-* None.
+* Fix more false positives in `ValidDocsRule`.
+  [diogoguimaraes](https://github.com/diogoguimaraes)
+  [#451](https://github.com/realm/SwiftLint/issues/451)
 
 ## 0.7.2: Appliance Manual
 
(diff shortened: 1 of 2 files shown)


jpsim:
Does this also handle #451?

```swift
public func performThings(things: (Void throws -> Result)) -> Result
```

diogoguimaraes:
Yes it handles #451.
I changed the logic of the regex to capture everything between `) foo -> bar { return }`.

jpsim:
So can you please add something like #451 to the tests to make sure we don't regress? Something like this:

```swift
/// no throws docs
func a(b: (Void throws -> Void)) -> Void {}
```

diogoguimaraes:
Done!

jpsim:
`testValidDocs` fails for me with this change.

jpsim:
Once you address that, could you please also add a changelog entry, referencing #451? Thanks!

diogoguimaraes:
I'll have to make a small change to `superfluousOrMissingThrowsDocumentation` method too so I'll finish this by tomorrow.

jpsim:
ok, thanks!

diogoguimaraes:
Done @jpsim. Can you please validate if it's all ok?
Thanks!

jpsim:
Thanks for these fixes!

PR #472, opened 2016-02-02
Fix trailing_semicolon autocorrect to handle more than one violation per line

### LINKED ISSUE #465: trailing_semicolon autocorrect only corrects a single violation
If you have more than one trailing semicolon on a single line, autocorrect does not correct all instances, only the final one (issue observed on latest master).

Example:
```
print("Test");;
```

`swift autocorrect`

Gives:
```
print("Test");
```

daniel-beard:
- Handle multiple violations that may also contain whitespace (excluding newlines) on a single line.
- Fixes #465


jpsim:
Thanks!

PR #473, opened 2016-02-02
Bring name inline with other CHANGELOG entries

daniel-beard:
- Minor change, bring changelog entry inline with others.
- My bad :/


jpsim:
No worries. Thanks!

Issue #468, opened 2016-02-02
Run unit tests with SPM


jpsim:
The unit tests are only run in the Xcode job at the moment, it'd be good to also run the unit tests with SPM.

norio-nomura:
Watching: [SR-592 Implement SE-0019 Swift Package Manager Testing](https://bugs.swift.org/browse/SR-592)

jpsim:
It should be possible to do before that proposal is implemented. See this for example: https://github.com/PureSwift/SwiftFoundation/tree/develop/Sources/UnitTests

jpsim:
Done in #476.

PR #476, opened 2016-02-03
[SPM] Add `make spm_test`

norio-nomura:
#468 

Notes:
- `swift-DEVELOPMENT-SNAPSHOT-2016-01-25-a` contains `libswiftXCTest.dylib`. But that can not be used on building by SPM because that does not contain module. So, this PR uses `swift-corelibs-xctest` as dependency.
- Use forked [`norio-nomura/swift-corelibs-xctest`](https://github.com/norio-nomura/swift-corelibs-xctest) as dependency because original repository has no version tags yet.
- Test cases are needed conforming to `XCTestCaseProvider`.
- Add `import Foundation` explicitly.
- Change some path operation depending `NSBundle`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,7 @@ Carthage/Build
 SwiftLint.pkg
 SwiftLintFramework.framework.zip
 benchmark_*
+
+# SPM
+.build
+Packages
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ matrix:
   include:
     - script: set -o pipefail && script/cibuild | xcpretty
       env: JOB=Xcode
-    - script: make spm
+    - script: make spm_test
       env: JOB=SPM
       before_install:
         - make swift_snapshot_install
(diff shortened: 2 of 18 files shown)


scottrhoyt:
Love seeing this SPM work being done. Thanks @norio-nomura !

norio-nomura:
:smile: 
Does codecov support matrix builds?

scottrhoyt:
I don't think doing matrix builds will break it, but I also don't think we will be able to see code coverage broken down by the matrix element. Probably, it will just be the last report uploaded, but I cannot be sure of this. What do you have in mind? Doing parallel builds with Xcode and SPM? Maybe there is an option other than Codecov that get's us what we want?

norio-nomura:
Oh, I understand. Codecov shows only coverage for PR, not for builds.
> Doing parallel builds with Xcode and SPM?

It already does parallel builds with Xcode an SPM.
> What do you have in mind?

If we would want to make good use of Codecov, we would disable codecov on SPM explicitly.

jpsim:
This is *amazing*. I'm so happy you did this and got it to work!

Let's get this in right away :smile: :tada: :100:

norio-nomura:
:pray:

PR #483, opened 2016-02-05
SwiftyTextTable

scottrhoyt:
This is a version of SwiftLint replacing the built in `TextTable` with a dependency on the newly-minted [SwiftyTextTable](https://github.com/scottrhoyt/SwiftyTextTable) that I extracted out. The dependency is installed via carthage and submodules. It is SPM ready as well. I followed the pattern applied to Commandant and added the dependency to SwiftLintFramework instead of swiftlint.

If we want to keep the builtin `TextTable` and skip adding a new dependency, that makes sense. But this is what it would look like at any rate.

My hope is that SwiftyTextTable can garner contributors outside of SwiftLint and provide downstream improvements to SwiftLint's command line output.


Source/swiftlint/Extensions/TextTable+SwiftLint.swift
```diff
@@ -0,0 +1,33 @@
...
+//  SwiftLint
+//
+//  Created by Scott Hoyt on 2/4/16.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import SwiftLintFramework
+import SwiftyTextTable
+
+extension TextTable {
```
jpsim: since this is just used in `RulesCommand`, I think we can move it back in there. No need for this separate file.

jpsim:
Hurray! :+1: :tada:

Source/swiftlint/Extensions/TextTable+SwiftLint.swift
scottrhoyt: :+1: Done!

scottrhoyt:
The OS X build is running fine. The SPM build is failing because the current dev snapshot pkg seems to be unavailable.

norio-nomura:
New snapshot has come https://swift.org/download/#latest-development-snapshots

norio-nomura:
Oops, not found.

norio-nomura:
Could you please apply following patch for reflecting configuration to `configuration` column?:
```diff
diff --git a/Source/swiftlint/Commands/RulesCommand.swift b/Source/swiftlint/Commands/RulesCommand.swift
index 6725f0a..5e81d27 100644
--- a/Source/swiftlint/Commands/RulesCommand.swift
+++ b/Source/swiftlint/Commands/RulesCommand.swift
@@ -85,12 +85,14 @@ extension TextTable {
         let sortedRules = ruleList.list.sort { $0.0 < $1.0 }
         for (ruleId, ruleType) in sortedRules {
             let rule = ruleType.init()
+            let configuredRule = configuration.rules.indexOf({
+                $0.dynamicType.description.identifier == ruleId
+            }).map { configuration.rules[$0] }
             addRow(ruleId,
                    (rule is OptInRule) ? "yes" : "no",
                    (rule is CorrectableRule) ? "yes" : "no",
-                   configuration.rules.map({ $0.dynamicType.description.identifier })
-                       .contains(ruleId) ? "yes" : "no",
-                   (rule as? _ConfigProviderRule)?.configDescription ?? "N/A")
+                   configuredRule != nil ? "yes" : "no",
+                   (configuredRule as? _ConfigProviderRule)?.configDescription ?? "N/A")
         }
     }
 }
```

scottrhoyt:
Yeah! :+1:

norio-nomura:
Restarted the SPM build.

norio-nomura:
Thanks for applying patch!

PR #487, opened 2016-02-05
Deallocate memory after every linting

norio-nomura:
- Use `autoreleasepool(_:)`
- Call `File.invalidateCache()`

Instruments on linting Carthage 0.12,
maximum Persistent Bytes is reduced from 508.49MB to 170.74MB.
From:
screenshot 2016-02-05 19 21 17
To:
screenshot 2016-02-05 22 30 32

Total number of allocation does not increase. (If disposed memory was reused, it increases.)

Running Time of calling: `File.invalidateCache()` is 381ms, `autoreleasepool(_:)` is 45ms.
screenshot 2016-02-05 22 26 16

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,9 @@
   [JP Simard](https://github.com/jpsim)
   [#392](https://github.com/realm/SwiftLint/issues/392)
 
+* Reduce maximum memory usage.  
+  [Norio Nomura](https://github.com/norio-nomura)
+
 ##### Bug Fixes
 
 * Fix more false positives in `ValidDocsRule`.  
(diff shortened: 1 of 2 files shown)


scottrhoyt:
@norio-nomura if you wouldn't mind adding a section to the CHANGELOG for this, that would be great. I think such a large performance improvement deserves a mention. :)

@jpsim if you are cool with this, I'd like to get this in as the last PR for 0.8.0.

norio-nomura:
@scottrhoyt Thanks! I added.

norio-nomura:
It seems we need pinning SPM dependencies more strict as seeing SPM fails on Travis-CI.
https://travis-ci.org/realm/SwiftLint/jobs/107378714

jpsim:
:+1:

PR #488, opened 2016-02-05
Emit a build warning if SwiftLint is not installed

berikv:


jpsim:
Works for me, thanks!

Although if the project treats warnings as errors, it won't build if SwiftLint isn't installed.

PR #489, opened 2016-02-06
SwiftyTextTable version pinning

scottrhoyt:


jpsim:
I'm curious why this pinning is either necessary or useful? The `Cartfile`s have mostly just been used as a shortcut to update the submodules to the latest releases so far.

scottrhoyt:
SwiftLint SPM builds were failing. This was because I introduced something into SwiftyTextTable's tests that worked fined under Xcode but failed in SPM. I corrected this in SwiftyTextTable, but because I don't yet have SPM tests running for it, I couldn't guarantee it wouldn't happen again. That's why I pinned the SPM build to a known working version of SwiftyTextTable. It made sense to me to include this pin in the Cartfile as well to be consistent.

scottrhoyt:
I plan on getting SwiftyTextTable SPM testing in Travis shortly so this should be a non-issue at that point.

norio-nomura:
Now `Cartfile.resolved` can be updated only specific dependency since [Carthage 0.12](https://github.com/Carthage/Carthage/pull/972), so we can also go without pinning with avoiding updates `SwiftyTextTable`. But I think pinning in Cartfile is good choice for using young repository. :+1:

SPM resolves dependencies on every builds. We need to treat `Package.swift` as a same role of `Cartfile.resolved` on resolving versions of dependencies.

PR #494, opened 2016-02-06
Update screenshot with new build script

berikv:
Following the update of the suggested "run script", the accompanying screenshot has been updated also.


jpsim:
Thanks!

Issue #440, opened 2016-01-28
Configure reporter using command line flag


guidomb:
I think it would be nice to be able to do something like `swiftlint lint --reporter json` and overwrite what it is configured in `.swiftlint.yml`. This would be pretty useful in CI env to be able run scripts to analyze the output and not forcing devs to use JSON reporter if they want to use another one.

jpsim:
:+1:

norio-nomura:
:+1:

jpsim:
Done in #495

guidomb:
:tada:

PR #495, opened 2016-02-06
allow passing an optional `--reporter` parameter to `swiftlint lint`

jpsim:
Addresses #440 /cc @scottrhoyt @norio-nomura


scottrhoyt:
This looks like a good solution to me right now.

jpsim:
Thanks for reviewing!

norio-nomura:
Nice!

PR #497, opened 2016-02-06
add `--quiet` flag to `lint` and `autocorrect` commands

jpsim:
addresses #386.

depends on #495. /cc @scottrhoyt @norio-nomura 

**update:** no longer depends on #495 since that's since been merged.

Issue #120, opened 2015-09-03
Print informative errors when configuration YAML contains unused information


keith:
Not sure this is worth fixing, but I wanted to bring it up.

If you make a config file like:

```yml
included: Classes
```

Instead of:

```yml
included:
  - Classes
```

SwiftLint will silently ignore the first one. Typically I think this performs the same way in yaml, but regardless it might be nice to add errors for cases where the configuration isn't what we'd expect. The only reason I noticed this is because of the large number of files being linted.

jpsim:
I'd absolutely love to do more error handling wrt yaml files, but haven't yet found an elegant way to do that.

sboc:
Where do I put .swiftlint.yml btw? I cannot get this to work at all. Where do I run swiftlint from?

dpassage:
It looks like `.swiftlint.yml` doesn't work with the latest released version 0.1.2; I got it to work by checking out master and building it myself.

SwiftLint team, would love to get the config file support released! Great tool so far.

jpsim:
> SwiftLint team, would love to get the config file support released! Great tool so far.

Homebrew literally just yesterday started accepting formulas requiring Xcode 7, so we haven't been able to release until now. We'll be releasing shortly.

dpassage:
Hooray! Thanks @jpsim !

jpsim:
Many improvements to YAML error reporting were just merged in #498.

PR #498, opened 2016-02-06
Add more validation to configuration files

jpsim:
Addresses #120. Depends on #497.

I may have gotten a bit carried away with other changes in this PR, so let me know if you'd like me to split this up /cc @scottrhoyt @norio-nomura

### DIFF
--- a/Source/SwiftLintFramework/Models/Configuration.swift
+++ b/Source/SwiftLintFramework/Models/Configuration.swift
@@ -108,6 +108,23 @@ public struct Configuration: Equatable {
         let optInRules = dict[ConfigurationKey.OptInRules.rawValue] as? [String] ??
             dict[ConfigurationKey.EnabledRules.rawValue] as? [String] ?? []
 
+        // Log an error when supplying invalid keys in the configuration dictionary
+        let validKeys = [
+            ConfigurationKey.DisabledRules,
+            .EnabledRules,
+            .Excluded,
+            .Included,
+            .OptInRules,
+            .Reporter,
+            .UseNestedConfigs,
+            .WhitelistRules,
+        ].map({ $0.rawValue }) + masterRuleList.list.keys
+
+        let invalidKeys = Set(dict.keys).subtract(validKeys)
+        if !invalidKeys.isEmpty {
+            queuedPrintError("Configuration contains invalid keys:\n\(invalidKeys)")
+        }
+
         self.init(
             disabledRules: dict[ConfigurationKey.DisabledRules.rawValue] as? [String] ?? [],
             optInRules: optInRules,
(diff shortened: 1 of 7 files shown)


Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -122,7 +157,7 @@ public struct Configuration: Equatable {
             let yamlContents = try NSString(contentsOfFile: fullPath,
                 encoding: NSUTF8StringEncoding) as String
             let dict = try YamlParser.parse(yamlContents)
```
norio-nomura: I sometimes add `print(error)` to `catch` block for knowing the error thrown by `Yam.parse(_:)`…

jpsim: Good point. I just added a log for that.

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -121,41 +156,24 @@ public struct Configuration: Equatable {
...
+            do {
+                let dict = try YamlParser.parse(yamlContents)
+                if !quiet {
+                    queuedPrintError("Loading configuration from '\(path)'")
+                }
+                self.init(dict: dict)!
+                configPath = fullPath
+                return
+            } catch {
+                queuedPrintError("Error parsing YAML: \(error)")
```
norio-nomura: It would be better if YAML parsing error would contains location information.
But it seems YamlSwift does not have such functions.

jpsim: Yup, I think that's all we can do for now.

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -87,30 +97,55 @@ public struct Configuration: Equatable {
...
+            .Included,
+            .OptInRules,
+            .Reporter,
+            .UseNestedConfigs,
+            .WhitelistRules,
+        ].map({ $0.rawValue }) + masterRuleList.list.keys
+
+        let invalidKeys = Set(dict.keys).subtract(validKeys)
+        if !invalidKeys.isEmpty {
+            queuedPrintError("Configuration contains invalid keys:\n\(invalidKeys)")
```
norio-nomura: Should we continue on detecting invalid key?

jpsim: I think so.

norio-nomura:
:+1:

scottrhoyt:
This is a good improvement! :+1:

jpsim:
Did either of you want to review #497 separately? If there are no objections, I'll get that in.

PR #499, opened 2016-02-07
fix inconsistencies between violations & corrections in StatementPositionRule

### LINKED ISSUE #466: Autocorrect reports success even though nothing happens
When I call  `swiftlint autocorrect` 

I get a report like this:

```
.
.
.
/Source/TabBarController.swift:51:25 Corrected Comma Spacing
/Source/TabBarController.swift:50:67 Corrected Comma Spacing
/Source/TabBarController.swift:49:101 Corrected Comma Spacing
Done correcting 30 files!
```

But no file is modified, 

I noticed it happens with
Statement Position and  Comma Spacing, but there might be more.

Other rules get corrected corectly. It would be nice to at least to not display the success message if the rule fails to correct the error. It's a bit confusing.

Thanks

jpsim:
Partly fixes #466. /cc @tkohout @scottrhoyt @norio-nomura 

All rules that share a different regular expression or codepath for detecting violations and snippets to correct are at risk of having this issue, so moving forward, we should take extra care to make sure they're consistent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@
   [Ankit Aggarwal](https://github.com/aciidb0mb3r)
   [#298](https://github.com/realm/SwiftLint/issues/298)
 
+* Fixed inconsistencies between violations & corrections in
+  `StatementPositionRule`.  
+  [JP Simard](https://github.com/jpsim)
+  [#466](https://github.com/realm/SwiftLint/issues/466)
+
 ## 0.8.0: High Heat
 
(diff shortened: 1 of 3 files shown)


jpsim:
I just modified `verifyRule` to make sure that non-triggers are never corrected and that corrections are never no-ops. This should help prevent this kind of inconsistency in the future.

scottrhoyt:
Sounds good. :+1:

jpsim:
Sweet.

PR #508, opened 2016-02-07
make RulesCommand show a configuration description for all rules

jpsim:
Now that all rules are configurable, I think it'd be nice to fold `ConfigurableRule` into `Rule` and get rid of `ConfigurableRule` altogether. /cc @scottrhoyt @norio-nomura


scottrhoyt:
That makes sense to me!

PR #509, opened 2016-02-07
fold ConfigurableRule into Rule

jpsim:
Depends on #508 and as mentioned there, since all rules are now configurable and that this is something we'll likely want to enforce with all rules moving forward, there's no need for a dedicated "configurable rule" protocol, it should just be folded into `Rule`. /cc @scottrhoyt @norio-nomura

### DIFF
--- a/Source/swiftlint/Commands/RulesCommand.swift
+++ b/Source/swiftlint/Commands/RulesCommand.swift
@@ -83,16 +83,21 @@ extension TextTable {
         ]
         self.init(columns: columns)
         let sortedRules = ruleList.list.sort { $0.0 < $1.0 }
-        for (ruleId, ruleType) in sortedRules {
+        for (ruleID, ruleType) in sortedRules {
             let rule = ruleType.init()
-            let configuredRule = configuration.rules.indexOf({
-                $0.dynamicType.description.identifier == ruleId
-            }).map { configuration.rules[$0] }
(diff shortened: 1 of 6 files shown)


Source/swiftlint/Commands/RulesCommand.swift
```diff
@@ -83,16 +83,21 @@ extension TextTable {
...
             let rule = ruleType.init()
-            let configuredRule = configuration.rules.indexOf({
-                $0.dynamicType.description.identifier == ruleId
-            }).map { configuration.rules[$0] }
-            addRow(ruleId,
-                   (rule is OptInRule) ? "yes" : "no",
-                   (rule is CorrectableRule) ? "yes" : "no",
-                   configuredRule != nil ? "yes" : "no",
-                   ((configuredRule ?? rule)  as? _ConfigProviderRule)?.configDescription ?? "N/A")
+            let configuredRule: Rule? = {
```
jpsim: I'm not a fan of the current `indexOf(...).map { rules[$0] }` approach (get index of element matching a predicate, then subscript at that index to get the object again). This alternative is more verbose but feels more direct to me.

norio-nomura: I see.

norio-nomura:
:+1:

PR #511, opened 2016-02-08
Avoid mixing strings in console from stdout and stderr

### LINKED ISSUE #432: Command /bin/sh emitted errors but did not return a nonzero exit code to indicate failure
Originally reported in issue #200, seems like it is happening again in 0.7.1?

norio-nomura:
`Swift.print()` uses buffered stdout by calling `putchar(_:)`.
This avoids mixing strings in console from stdout and stderr.

fix #432


jpsim:
Awesome, thanks!

scottrhoyt:
Thanks! :+1:

PR #516, opened 2016-02-09
Change approach for SPM support, requiring no llbuild flags

norio-nomura:
- Same approach as https://github.com/jpsim/SourceKitten/pull/168
- Switch dependency to my fork for testing without build-delete-build approach.

### DIFF
--- a/Source/Clang_C/module.modulemap
+++ b/Source/Clang_C/module.modulemap
@@ -1,5 +1,0 @@
-module Clang_C {
-  umbrella "."
-  link "clang"
-  module * { export * }
-}
--- a/Source/Clang_C/module.modulemap
+++ b/Source/Clang_C/module.modulemap
@@ -1,5 +1,5 @@
-module Clang_C [system] {
-  header "/Library/Developer/Toolchains/swift-latest.xctoolchain/usr/local/include/clang-c/Documentation.h"
+module Clang_C {
+  umbrella "."
   link "clang"
-  export *
+  module * { export * }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ matrix:
       env: JOB=SPM
       before_install:
         - make swift_snapshot_install
+        - make spm_bootstrap
   exclude:
     - script: placeholder # workaround for https://github.com/travis-ci/travis-ci/issues/4681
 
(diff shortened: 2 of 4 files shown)


jpsim:
It looks like the only change necessary in your fork was to explicitly add `Result` as a dependency. I thought we'd have to delete the `sourcekitten` and `SourceKittenFrameworkTests` directories? Why aren't those being built anymore? I know this happens a lot lately, but I'm confused :grin: 

In any case, this is great!

Also, maybe we should move the bootstrap/teardown code and Clang's module map to a centralized location rather than copying it here and everywhere else that needs `SourceKit` or `Clang_C`? Unfortunately, SPM packages don't have any sort of pre-installation script step. Maybe we can just `curl ...raw.github... | sh`?

norio-nomura:
> Why aren't those being built anymore?

Those are built and linked *successfully*.
https://travis-ci.org/realm/SwiftLint/jobs/107926696#L421

`sourcekitten`'s  case:
By using `-v` option of SPM, I found SPM links modules to executable which listed in their `Package.swift` only. does not link nested dependencies. So, explicitly add nested dependency to the `Package.swift`, then it links `Result` and resolved link error.

`SourceKittenFrameworks`'s case:
Maybe it has been resolved when we added `SourceKittenFrameworks` target in  https://github.com/jpsim/SourceKitten/commit/a5fb2ec2363f9253125a71125cdb141688d4f3cb

> Maybe we can just curl ...raw.github... | sh?

:+1:

jpsim:
Ah I see, thanks for explaining! Let's add `script/spm_bootstrap` and `script/spm_teardown` to the SourceKitten repo then.

I should be able to do that quickly and cut a SourceKitten release so we can target that instead of your fork.

jpsim:
You should now be able to point to jpsim/SourceKitten 0.10.0 and update Makefile to this:

```
spm_bootstrap: spm_teardown
  curl https://raw.githubusercontent.com/jpsim/SourceKitten/master/script/spm_bootstrap | bash -s $(SWIFT_SNAPSHOT)

spm_teardown:
  curl https://raw.githubusercontent.com/jpsim/SourceKitten/master/script/spm_teardown | bash
```

Source/Clang_C/module.modulemap
```diff
@@ -1,5 +1,5 @@
-module Clang_C [system] {
-  header "/Library/Developer/Toolchains/swift-latest.xctoolchain/usr/local/include/clang-c/Documentation.h"
+module Clang_C {
```
jpsim: this file can be removed completely

norio-nomura: :+1:

jpsim:
:+1: much better!

norio-nomura:
Thanks! :pray:

PR #525, opened 2016-02-11
Expand usage of "config" to "configuration" in most places

jpsim:
This has a considerable impact on the public API and touches a large part of the codebase. We decided to do this in #507 (which this PR depends on).

There are two places where we still use the "config" abbreviation: the command line `--config` parameter and the `use_nested_configs` configuration key. I'd actually suggest that we deprecate the configuration key altogether since when would you ever have nested configuration files without wanting to use them? This should just be the standard behavior IMO.

/cc @scottrhoyt @norio-nomura


norio-nomura:
Maybe we should use `custom_rules` for finding that. :wink: 
```yaml
custom_rules:
  config_abbreviation:
    name: "config abbreviation"
    regex: "[cC][oO][nN][fF][iI][gG][^uU]"
    match_kinds:
      - identifier
    message: "config should be disabbreviated!"
    severity: error
```

scottrhoyt:
Haha. Love it. :smile:

jpsim:
rules for all the things! :stuck_out_tongue:

jpsim:
Merging in #527.

PR #527, opened 2016-02-11
deprecate use_nested_configs: always consider nested configuration files

jpsim:
is there a good reason not to do this @scottrhoyt ? This depends on #525


scottrhoyt:
The only consideration at the time was that nested configs are ~O(n) wrt the number of files. This can be reduced to O(n) wrt the number of directories if we use a reclusive directory traversal for lifting as opposed to an array of paths.

jpsim:
True, it's inefficient, but it's also very inexpensive compared to the rest of the heavy computations in the rest of the linting process.

In fact, there's no discernible difference even with a large number of files. Here's `time swiftlint` with Carthage 0.11:

* before this change: `11.14s user 0.63s system 82% cpu 14.245 total`
* after this change: `11.04s user 0.68s system 82% cpu 14.239 total`

There are 946 Swift files and 3714 total files according to `find . -type f | wc -l`.

scottrhoyt:
Great! Let's deprecate it then.

jpsim:
Cool. Then we can merge this once #525 is in.

norio-nomura:
Nice!

PR #528, opened 2016-02-11
move common Commandant options to separate file

jpsim:


jpsim:
I just thought we should do this to reduce duplication

scottrhoyt:
Keep it DRY, baby! :+1:

PR #536, opened 2016-02-13
Fix nested config traversing did not stop on root path

norio-nomura:


jpsim:
Excellent! Is this a critical issue? In other words, should we cut a release with this right away or can this wait?

norio-nomura:
No crashes happens by this.
Bu, on running `swiftlint lint --config ~/.swiftlint.yml` in `~/SwiftLint/`:
```
~/.swiftlint.yml
~/SwiftLint/
~/SwiftLint/.swiftlint.yml
~/SwiftLint/Carthage/Checkouts/Commandant/
```
`~/SwiftLint/Carthage/Checkouts/Commandant/*.swift` are linted by `~/SwiftLint/.swiftlint.yml`.
`--config` options is ignored.
I think this is critical.

norio-nomura:
Hmm, I can't say strongly it is critical after all…
Maybe my usage is not common.

PR #542, opened 2016-02-16
Fix `CommaRule` mismatch between violations and corrections

### LINKED ISSUE #466: Autocorrect reports success even though nothing happens

norio-nomura:
- Add failing correction example from #466 to `CommaRule`  
https://github.com/realm/SwiftLint/issues/466#issuecomment-184474925
- Rewrite `CommaRule.correctFile(_:)`  
Now corrections happen only on violations.
Because `assertCorrection(_:expected:)` expects one correction for each `corrections`,
and the rule added by 93e8e12 causes two corrections on previous logic.
Fix #466
- Simplify the regular expression in `CommaRule.validateFile(_:)`  
- Remove surplus `map` calls in `rangesAndTokensMatching(_:)`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+## Master
+
+##### Breaking
+
+* None.
+
+##### Enhancements
+
+* None.
+
+##### Bug Fixes
+
+* Fix `CommaRule` mismatch between violations and corrections.  
+  [Norio Nomura](https://github.com/norio-nomura)
+  [#466](https://github.com/realm/SwiftLint/issues/466)
+
 ## 0.9.1: Air Duct Cleaning
 
 ##### Breaking
@@ -17,10 +33,6 @@
 * Fix issues with nested `.swiftlint.yml` file resolution.  
   [Norio Nomura](https://github.com/norio-nomura)
   [#543](https://github.com/realm/SwiftLint/issues/543)
-
-* Fix `CommaRule` mismatch between violations and corrections.  
-  [Norio Nomura](https://github.com/norio-nomura)
-  [#466](https://github.com/realm/SwiftLint/issues/466)
 
 ## 0.9.0: Appliance Maintenance
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@
   [Norio Nomura](https://github.com/norio-nomura)
   [#543](https://github.com/realm/SwiftLint/issues/543)
 
+* Fix `CommaRule` did not match between violations and corrections.  
+  [Norio Nomura](https://github.com/norio-nomura)
+  [#466](https://github.com/realm/SwiftLint/issues/466)
+
 ## 0.9.0: Appliance Maintenance
 
 ##### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
   [Norio Nomura](https://github.com/norio-nomura)
   [#543](https://github.com/realm/SwiftLint/issues/543)
 
-* Fix `CommaRule` did not match between violations and corrections.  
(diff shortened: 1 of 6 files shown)


norio-nomura:
It seems more than two spaces after comma does not generate violation:
```swift
let a = [1,       2]
```
But correction happens for that.
I think correction's behavior may be correct.

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@
   [Norio Nomura](https://github.com/norio-nomura)
   [#535](https://github.com/realm/SwiftLint/issues/535)
 
+* Fix `CommaRule` did not match between violations and corrections.
```
jpsim: > Fix `CommaRule` mismatch between violations and corrections.

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -88,7 +88,7 @@ extension File {
             let tokensInRange = syntax.tokens.filter { token in
                 let tokenByteRange = NSRange(location: token.offset, length: token.length)
                 return NSIntersectionRange(matchByteRange, tokenByteRange).length > 0
-            }.map({ $0 })
+            }
```
jpsim: lol, oops

Source/SwiftLintFramework/Rules/CommaRule.swift
```diff
@@ -50,23 +51,60 @@ public struct CommaRule: CorrectableRule, ConfigurationProviderRule {
 
     public func correctFile(file: File) -> [Correction] {
         if validateFile(file).isEmpty { return [] }
-        let pattern = "\\s*\\,\\s*([^\\s])"
 
-        let description = self.dynamicType.description
-        var corrections = [Correction]()
-        var contents = file.contents
+        // captures spaces and comma only
+        let pattern =
```
jpsim: why not share the pattern between `validateFile` and `correctFile`? That should help prevent lint/correct mismatches in the future?

CHANGELOG.md
norio-nomura: Thanks :pray:

Source/SwiftLintFramework/Rules/CommaRule.swift
norio-nomura: I'll do.

norio-nomura:
@jpsim I got a question while I'm implementing following won't generate violation and correction.
A:
```swift
func f(
  a: Int,
  b: Int // comma+newline+spaces
) {
}
```
Should we generate violation and correction for following?
B:
```swift
func f(a: Int
     , b: Int // newline+spaces+comma+space
) {
}
```
In other word, should we exclude line ending character on checking?
B is weird for me. But I can't determine whether it should do or not.

norio-nomura:
Updated.

> Should we generate violation and correction for following?
> B:

I leave it generates on case B.

By applying this PR, The duration `CommaRule` of linting Carthage 0.14 increased from 154ms to 457ms.

jpsim:
Travis is backlogged at the moment, but the only commit not tested is the changelog one, so this is good to merge.

These changes all look good to me. Thanks @norio-nomura!

norio-nomura:
Thanks! :pray:

Issue #574, opened 2016-03-02
autocorrect touches all source files


neilgall:
Running swiftlint autocorrect touches the timestamp on all source files, even if no changes are made. This causes Xcode to have to rebuild everything on the next build, which makes autocorrect less useful as part of default build process.

PR #575, opened 2016-03-02
Do not overwrite files whose contents have not changed

neilgall:
Fixes https://github.com/realm/SwiftLint/issues/574

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 ##### Bug Fixes
 
+* Avoid overwriting files whose contents have not changed.
+  [Neil Gall](https://github.com/neilgall)
+  [#574](https://github.com/realm/SwiftLint/issues/574)
+
 * Fix `CommaRule` mismatch between violations and corrections.  
   [Norio Nomura](https://github.com/norio-nomura)
   [#466](https://github.com/realm/SwiftLint/issues/466)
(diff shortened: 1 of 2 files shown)


norio-nomura:
Nice!
Could you please add changelog entry? [CONTRIBUTING.md](https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes)

jpsim:
Thanks for the PR! This check is fine, but really we shouldn't be generating an empty correction in the first place. Can you please share which situations generate these false corrections? These could be opened as new issues. Thanks!

neilgall:
Sorry, I haven't looked into the reasons why empty corrections are generated. I noticed that swiftlint was touching every file in my project causing long Xcode build times, and simply looked for a way to mitigate this.

jpsim:
Is is possible for you to share which situations generate these false corrections? We can then take it from there to fix those, but I'd consider that a bug.

neilgall:
It seems that any situation generates a false correction. I created an new iOS project in Xcode, corrected the whitespace so swift-lint finds zero violations, then ran `swiftlint autocorrect` with no `.swiftlint.yml` configuration file. The .swift files in the project are overwritten with identical contents.

I haven't looked into the code in depth, but there seems to be a mismatch between the intent of `Linter.correct()` and  `OpeningBraceRule.correctFile()`. `correct()` builds up a list of corrections and invalidates the file's cache if non-empty. `OpeningBraceRule.correctFile()` always invokes `file.write()` via its `writeToFile(_: violatingRanges:)` method, before returning the list of corrections. I suspect the problem lies in OpeningBraceRule.

jpsim:
Ok, thanks for looking into this, and for the PR!

Issue #166, opened 2015-10-23
Use multiple exit status codes


jakecraige:
Currently SwiftLint exits with a 1 error code when it dies because something is invalid(path, config, etc) OR if there are serious violations. What do you think about exiting with a different code on failing because of violations vs. invalid params?

This would be useful when running SwiftLint within another script and making decisions based on the response. Currently I have no way to tell if it's failing because of violations or invalid data.

kylef:
This will also be helpful for any vim syntastic plugin since you'd want to show swiftlint errors different to linting responses.

jpsim:
Agreed! Perhaps errors should exit with code `2` and warnings (in strict mode) could return `3`?

picaso:
Hi! Any timeframe on this enhancement? Thanks.

marcelofabri:
Fixed in #584

PR #584, opened 2016-03-10
use multiple exit codes to distinguish between errors & strict warnings

jpsim:
Addresses #166.

Does this approach make sense to you all? /cc @jakecraige @kylef @scottrhoyt @norio-nomura

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,14 @@
 
 ##### Enhancements
 
-* None.
+* Return different exit codes to distinguish between types of errors:
+  * 0: No errors, maybe warnings in non-strict mode
+  * 1: Usage or system error
+  * 2: Style violations of severity "Error"
+  * 3: No style violations of severity "Error", but violations of severity
+       "warning" with `--strict`  
+  [JP Simard](https://github.com/jpsim)
+  [#166](https://github.com/realm/SwiftLint/issues/166)
 
 ##### Bug Fixes
 
--- a/Source/swiftlint/Commands/LintCommand.swift
+++ b/Source/swiftlint/Commands/LintCommand.swift
@@ -61,8 +61,10 @@ struct LintCommand: CommandType {
                 saveBenchmark("files", times: fileTimes)
                 saveBenchmark("rules", times: ruleTimes)
             }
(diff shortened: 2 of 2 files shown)


jpsim:
| Code | Meaning                                                                                       |
| ---- | --------------------------------------------------------------------------------------------- |
| 0    | No errors, maybe warnings in non-strict mode                                                  |
| 1    | Usage or system error                                                                         |
| 2    | Style violations of severity "Error"                                                          |
| 3    | No style violations of severity "Error", but violations of severity "warning" with `--strict` |

scottrhoyt:
That's funny! I just implemented the same thing. Though slightly differently. I created a new error type to be used as Commandant's `ClientError`

```swift
enum SwiftLintError: ErrorType {
    case ExitCode(Int)
}
```

`LintCommand` was similar to your implementation:

```swift
            if options.strict && !violations.isEmpty {
                return .Failure(.CommandError(.ExitCode(3)))
            } else if numberOfSeriousViolations > 0 {
                return .Failure(.CommandError(.ExitCode(2)))
            }
```

Then in `main.swift` I modified the Commandant error handler to be:

```swift
registry.main(defaultVerb: LintCommand().verb) { error in
    queuedPrintError(String(error))
    switch error {
    case .CommandError(.ExitCode(let code)):
        exit(Int32(code))
    default:
        break
    }
}
```

The effect was the same as what you did, though I suppose it's a solution that might offer room for expansion to more types of errors/handling, but there's no good use case for that currently, so I'm happy to go with your much simpler approach. :+1:

jpsim:
Oh, sorry for stepping on your toes then @scottrhoyt! Your approach is definitely nicer, while mine is far simpler. In this case, I'd prefer keeping things simple until that "doesn't scale" :stuck_out_tongue:

scottrhoyt:
Premature optimization is the root of all evil. :smile: 

I'm all for simple. I'll keep it stashed away incase we go that direction.

jpsim:
> I'm all for simple. I'll keep it stashed away incase we go that direction.

:+1:

jakecraige:
Awesome! Thanks for doing this.

norio-nomura:
Nice!

jinnatan:
Awesome

PR #635, opened 2016-04-23
Prevent EmptyCountRule from triggering on variables named "count"

joshfriend:
With this change, `count` must be a property for the rule to trigger:

```swift
if someArray.count == 0 {
    // This case triggers the rule
}
```

```swift
var count = 0
// Do things to count...
if count == 0 {
    // This case no longer triggers the rule
}
```

This would fix the first false positive mentioned in #206:

> 1. `Int` variable named `count`. In this case, it'd be fine to check `count` against zero.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@
   [Norio Nomura](https://github.com/norio-nomura)
   [#617](https://github.com/realm/SwiftLint/issues/617)
 
+* `EmptyCountRule` no longer triggers when `count` is a variable.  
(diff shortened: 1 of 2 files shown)


jpsim:
Thanks @joshfriend, but the reason I kept the false positive here is to catch it in cases where you're extending `CollectionType`s, where accessing `self.count` without `self`. So I'm inclined not to accept this change, especially because I'm happy to have a higher tolerance for false positives in opt-in rules.

joshfriend:
👍 good point, thanks...

PR #636, opened 2016-04-24
Add autocorrect for ReturnArrowWhitespaceRule

CraigSiemens:


Source/SwiftLintFramework/Rules/ReturnArrowWhitespaceRule.swift
```diff
@@ -37,20 +37,81 @@ public struct ReturnArrowWhitespaceRule: ConfigurationProviderRule {
...
+        let description = self.dynamicType.description
+        var corrections = [Correction]()
+        var contents = file.contents
+
+        let results = matches.reverse().flatMap { range in
+            return regularExpression.firstMatchInString(contents, options: [], range: range)
+        }
+
+        for result in results {
+            guard result.numberOfRanges == 8 else { break }
```
jpsim: would be good to add a comment why 8 is the magic number here

jpsim:
This looks great @CraigSiemens, could you please add a changelog entry following the guidelines in our CONTRIBUTING.md document?

CraigSiemens:
Good point on the magic number. I changed it to be based on the largest replacement index that it will be expecting.

jpsim:
:+1: sweeeet! Big thanks for this @CraigSiemens!

PR #639, opened 2016-04-26
Added `UIEdgeInsetsMake` Legacy Rule

Nadohs:
add check for `UIEdgeInsetsMake` in `LegacyConstructorRule`. `UIEdgeInsets` should be used instead.


jpsim:
👍 thanks!

PR #644, opened 2016-04-28
Better handling `let`/`var` mix in `ConditionalBindingCascadeRule`

### LINKED ISSUE #642: False positive in conditional binding cascade violation
Conditional binding cascade violation warning is triggered when `let` and `var` is mixed. This is a false positive due to the fact that the type statements are required when the binding is heterogeneous.

ex:
``` swift
// triggers conditional_binding_cascade
guard let foo = someOptional(), var bar = someMutableOptional(foo) else {
  return
}
```

norio-nomura:
Fix #642

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+## Master
+
+##### Breaking
+
+* None.
+
+##### Enhancements
+
+* None.
+
+##### Bug Fixes
+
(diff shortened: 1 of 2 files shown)


jpsim:
👍

norio-nomura:
Thanks! :pray:

Lascorbe:
@norio-nomura  thank you for this, it's great 🍻

PR #647, opened 2016-05-02
Improves error messaging for Configuration, particularly with YAML

lyricsboy:
Like a noob, I created a new `.swiftlint.yml` using tabs instead of spaces. The error message from SwiftLint was not super helpful:

```
fatal error: Could not read configuration file at path '.../.swiftlint.yml': file .../SwiftLint/Source/SwiftLintFramework/Models/Configuration.swift, line 153
```

With these changes, various error conditions when loading a Configuration are logged more verbosely. In my example, the new output is:

```
fatal error: Could not read configuration file at path '.../.swiftlint.yml': Error parsing YAML: expected colon, near "included:\n- Something\nline_length:\n- 150\n- 250\n": file .../SwiftLint/Source/SwiftLintFramework/Models/Configuration.swift, line 153
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 ##### Enhancements
 
-* None.
+* Improve error messages for invalid configuration files.
+  [Brian Hardy](https://github.com/lyricsboy)
 
 ##### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


jpsim:
:+1: thanks for this!

PR #649, opened 2016-05-02
Xcode reporter on Readme example

phelgo:
The excellent documentation for this project makes it easy to get up and running in seconds. An exception that I found is that the reporter setup shown on the Readme makes warnings go silent (csv), so a naive copy-paste won't give the expected xcode highlights. I propose setting xcode reporting on this example, while leaving the other options on the comment

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ variable_name:
     - id
     - URL
     - GlobalAPIKey
-reporter: "csv" # reporter type (xcode, json, csv, checkstyle)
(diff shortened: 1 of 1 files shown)


jpsim:
Sure. It's not great either because overriding the default setting with the same value might confuse some people in thinking that they *must* set a value, but that's fine

phelgo:
That's definitely a good point, but I hope this will make it easier for newcomers and it will still be easy to learn more. Thanks for the merge!

PR #244, opened 2015-12-03
Fix `LD_RUNPATH_SEARCH_PATHS`

norio-nomura:
## Fix `LD_RUNPATH_SEARCH_PATHS` of `SwiftLintFramework.framework`

- Remove "$(inherited)"
because that comes from xcconfig as following:
```
@executable_path/../Frameworks
@loader_path/../Frameworks
```
are not unsuitable for framework target.
See also: https://github.com/jspahrsummers/xcconfigs/pull/52

- Add "@loader_path/Frameworks"
Sub framework can be loaded by this.

- Add "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib"
$(DEVELOPER_DIR) is resolved on linking. If Xcode.app has been renamed on packaging environment, that was recorded.
On https://github.com/realm/SwiftLint/releases/download/0.4.0/SwiftLint.pkg:
```
% otool -l pkg/Library/Frameworks/SwiftLintFramework.framework/SwiftLintFramework
…
Load command 32
          cmd LC_RPATH
      cmdsize 104
         path /Applications/Xcode-7.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib (offset 12)
```
This cause #169

See also: https://github.com/jpsim/SourceKitten/pull/110

## Fix `LD_RUNPATH_SEARCH_PATHS` of `SwiftLint.app`

Limit paths used for .pkg installation. Homebrew installation will be supported by changes on `prefix_install` of `Makefile`.
- Remove `@executable_path/.` because that was not used.
- Remove `@executable_path/../Frameworks`, that will be used for Homebrew installation.
- Remove `@executable_path/SwiftLintFramework.framework/Versions/Current/Frameworks` because that was not used.
- Remove `@executable_path/SwiftLintFramework.framework/Versions/Current/Frameworks/SourceKittenFramework.framework/Versions/Current/Frameworks` because it was not used.
- Remove `/Library/Frameworks/SwiftLintFramework.framework/Versions/Current/Frameworks/SourceKittenFramework.framework/Versions/Current/Frameworks` because sub frameworks of `SourceKittenFramework` will be loaded by https://github.com/jpsim/SourceKitten/pull/110

- Remains `/Library/Frameworks/SwiftLintFramework.framework/Versions/Current/Frameworks` and make that first. Because Sub-frameworks contained by `SwiftLintFramework.framework` should be loaded before them distributed other apps.
[...]

PR #650, opened 2016-05-03
Add additional "Runpath Search Paths" values.

vokal-isaac:
Add `@executable_path/../Frameworks` so that the `swiftlint` binary can be run while inside the Xcode-built `.app` bundle.
Add `@executable_path` and `@executable_path/SwiftLintFramework.framework/Versions/Current/Frameworks` so that putting `SwiftLintFramework.framework` in the same directory as the `swiftlint` binary will allow the binary to run.

Basically, this gives a bit more flexibility in how `swiftlint` is installed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 
 * Improve error messages for invalid configuration files.
   [Brian Hardy](https://github.com/lyricsboy)
+  
+* Add Runpath Search Paths so that `swiftlint` can find
+  `SwiftLintFramework.framework` in more places.  
+  [Isaac Greenspan](https://github.com/vokal-isaac)
 
 ##### Bug Fixes
 
--- a/SwiftLint.xcodeproj/project.pbxproj
+++ b/SwiftLint.xcodeproj/project.pbxproj
@@ -1145,7 +1145,7 @@
 			baseConfigurationReference = D0D1213419E878CC005E4BAA /* Mac-Application.xcconfig */;
 			buildSettings = {
 				INFOPLIST_FILE = "Source/swiftlint/Supporting Files/Info.plist";
-				LD_RUNPATH_SEARCH_PATHS = "@executable_path/../Frameworks/SwiftLintFramework.framework/Versions/Current/Frameworks /Library/Frameworks/SwiftLintFramework.framework/Versions/Current/Frameworks /Library/Frameworks";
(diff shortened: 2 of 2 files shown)


norio-nomura:
Thanks for contribution.

This PR breaks compatibility of existing two distribution methods, because we are handling `RPATH` not only with `LD_RUNPATH_SEARCH_PATHS` in `SwiftLint.xcodeproj` but also with `install_name_tool` in `Makefile`.
See: #244 
If you want to add another installation method, please keep compatibility of existing two distribution with understanding following.

#### Current binary distribution and `RPATH`
SwiftLint supports two methods of binary distribution. `SwiftLintFramework.framework` is installed to different path on each methods.
- Homebrew installs `SwiftLintFramework.framework` to `/usr/local/Cellar/swiftlint/<version>/Frameworks/SwiftLintFramework.framework`  
  `/usr/local/bin/swiftlint` has following `RPATH`:  
  - `@executable_path/../Frameworks/SwiftLintFramework.framework/Versions/Current/Frameworks`  
  - `@executable_path/../Frameworks`  
  
  Homebrew also creates `/usr/local/Frameworks/SwiftLintFramework.framework` that contains symbolic links to `/usr/local/Cellar/swiftlint/<version>/Frameworks/SwiftLintFramework.framework`.

- Installer Package installs `SwiftLintFramework.framework` to `/Library/Frameworks/SwiftLintFramework.framework`  
  `/usr/local/bin/swiftlint` has following `RPATH`:  
  - `/Library/Frameworks/SwiftLintFramework.framework/Versions/Current/Frameworks`
  - `/Library/Frameworks`

#### Problem with two versions of SwiftLint by different distributions
By applying this PR, problem will be caused on following scenario:
1. Install SwiftLint by Homebrew
2. Install newer version of SwiftLint by Installer package without uninstalling homebrew installation
3. Launching newer version of `/usr/local/bin/swiftlint` loads older version of `/usr/local/Frameworks/SwiftLintFramework.framework`, and the result is unpredictable.

That can be avoided by changing order of adding entries in `RPATH`, but I think we should keep `RPATH` simple on each of installation.

vokal-isaac:
Thanks for your thorough response.  I'm closing this in favor of https://github.com/realm/SwiftLint/pull/669, which I think fits better with what you describe.

Regarding the problem of two copies of the framework from different installations, it may be worth versioning the framework so that a given instance of the app binary won't use an incompatible version of the framework, regardless of location on the filesystem, but that's sufficiently complicated that it may not be worth touching.

PR #660, opened 2016-05-10
Allow lowercase enum values

koke:
The Swift API Design Guidelines states that:

> Names of types and protocols are UpperCamelCase. Everything else is
> lowerCamelCase.

https://swift.org/documentation/api-design-guidelines/#follow-case-conventions

This PR changes `type_name` so it doesn't force enum values to UpperCamelCase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 ##### Breaking
 
-* None.
+* Now `type_name` allows lowercase enum values to match the Swift API Design
+  Guidelines.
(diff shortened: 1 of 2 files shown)


jpsim:
Actually I think the new Swift convention is to essentially enforce lowerCamelCase for enum elements, so I think we can be even stricter? For now, I think this is fine as a transition period. Thanks!

koke:
Yes, I created #663 to enforce lowerCamelCase for all identifiers

PR #673, opened 2016-05-25
Disable the cascade warning when case is involved

paddlefish:
This is a workaround that prevents invalid warning when the use of a case statement requires multiple let keywords for syntax


jpsim:
:+1: a bit limited, but certainly an improvement. Thanks!

paddlefish:
W00T!  Thanks for a prompt response!

jpsim:
No, thank you for the contribution!

PR #679, opened 2016-05-30
Move embedding frameworks from SwiftLintFramework target to swiftlint target

norio-nomura:
- ~~Change to contain single copy of Swift libraries into `SwiftLintFramework.framework` on `make installables`~~
  - ~~Remove `settings-for-all-projects.xcconfig` and add flags into `XCODEFLAGS` in `Makefile`~~
  - ~~Add `EMBEDDED_CONTENT_CONTAINS_SWIFT=NO` to `XCODEFLAGS` in `Makefile`~~
    ~~This avoids installing Swift libraries to both of `SourceKittenFramework` and `SwiftLintFramework`.~~
  - ~~Add `xcrun swift-stdlib-tool --copy ...` to `installables` target in `Makefile`~~
    ~~This installs Swift libraries to `SwiftLintFramework.framework/Versions/Current/Frameworks`.~~  
- ~~Remove frameworks from `SourceKittenFramework.framework/Versions/Current/Frameworks`~~  
  ~~Because same frameworks are already contained in `SwiftLintFramework.framework/Versions/Current/Frameworks`.~~

- Stop copying libraries to `Frameworks` sub directory in `SwiftLintFramework.framework`
  - Remove "Copy Frameworks" build phase from SwiftLintFramework target
  - Change SwiftLintFramework target to `EMBEDDED_CONTENT_CONTAINS_SWIFT=NO`
- Add packaging processes for `SwiftLintFramework.framework` to swiftlint target
  - Add a Copy File Build Phase that embedding frameworks into `SwiftLintFramework.framework`
  - Add a Shell Script Build Phase that embedding Swift libraries into `SwiftLintFramework.framework`
  `swift-stdlib-tool` is called for copying Swift libraries that imitating usage by Xcode.

By applying this and https://github.com/jpsim/SourceKitten/pull/215, the size of `SwiftLint.pkg` is reduced from 4.8MB to 2.9MB.


Makefile
```diff
@@ -48,6 +52,12 @@ uninstall:
 
 installables: clean bootstrap
 	$(BUILD_TOOL) $(XCODEFLAGS) install
+	rm -rf "$(SWIFTLINTFRAMEWORK_BUNDLE)/Versions/Current/Frameworks/SourceKittenFramework.framework/Versions/Current/Frameworks"
+	xcrun swift-stdlib-tool --copy \
+		--scan-executable "$(SWIFTLINT_EXECUTABLE)" \
+		--scan-folder "$(SWIFTLINTFRAMEWORK_BUNDLE)" \
+		--platform macosx --destination "$(SWIFTLINTFRAMEWORK_BUNDLE)/Versions/Current/Frameworks" \
+		--strip-bitcode
```
jpsim: is it necessary to strip bitcode since we're only building on OS X, where `swiftc` shouldn't generate it?

jpsim:
This is a nice reduction in size! Wouldn't it be possible to do all this in the Xcode configuration though?

Makefile
norio-nomura: I copied options that Xcode doing when `EMBEDDED_CONTENT_CONTAINS_SWIFT=YES`.
It seems Xcode uses it regardless of target's platform.
In current behavior, `swift-stdlib-tool` with `--strip-bitcode` just copies Swift libraries without modification.
It might not be necessary, but I couldn't find any reason to drop it that Xcode using.

norio-nomura:
It would be better to start from https://github.com/jpsim/SourceKitten/pull/203 for doing all in the Xcode configuration.

norio-nomura:
Rewrote to depends on https://github.com/jpsim/SourceKitten/pull/215 and doing all in the Xcode configuration.

As far as I tested, `xcodebuild clean` does not clean Intermediates directory that used on `xcodebuild install`. So, the installations by `make install` or `make prefix_install` may cause containing duplicated framework and Swift libraries that expected to be removed by this PR without removing DerivedData directory.

jpsim:
This looks good to me @norio-nomura, but won't merge in case you're still working on it. Please merge this yourself when you think it's ready.

norio-nomura:
Filed the issue of `xcodebuild clean` to [rdar://26591236: `xcodebuild clean` does not clean "OBJROOT" that used on `xcodebuild install`](https://openradar.appspot.com/radar?id=5044858821541888).

I decided that I won't include workaround for the issue of `xcodebuild clean` into this PR.

norio-nomura:
Thanks for reviews! 🙏

Issue #651, opened 2016-05-04
Support linting cuddled else


ateliercw:
As an alternative to the standard StatementPosition Rule, I'd like to see support for uncuddled else statements, either as a configuration option for statement position, or as a separate mutually exclusive rule.

jpsim:
yup, I'd envision this working as a configuration option on the statement position rule.

ateliercw:
@jpsim I folded the listing for uncuddled else into statement position on my branch, but I closed the PR until I get autocorrect implemented for the alternate style, if you have any comments on the parameter name or implementation let me know

https://github.com/ateliercw/SwiftLint/blob/ms-cuddled-else-rule/Source/SwiftLintFramework/Rules/StatementPositionRule.swift

jpsim:
That looks good to me! Not 100% sure about the wording, but can't think of anything better either.

I could imagine autocorrect for the uncuddled case being difficult to get right with indentation, so I'd accept if autocorrect just skipped that rule if the uncuddled mode was specified.

ateliercw:
@jpsim I changed the options to use snake case to match the rest of the options, and have an autocorrect option that's passing tests

PR #688, opened 2016-06-09
Add uncuddled else support to the statement_position rule

ateliercw:
implements uncuddled else as an option on statement_position including autocorrect, should solve issue #651


Source/SwiftLintFramework/Rules/RuleConfigurations/StatementPositionConfiguration.swift
```diff
@@ -0,0 +1,59 @@
...
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import Foundation
+
+public enum StatmentModeConfiguration: String {
+    case Standard = "standard", UncuddledElse = "uncuddled_else"
+
+    init(value: AnyObject) throws {
+        guard let string = (value as? String)?.lowercaseString else {
```
jpsim: this init is equivalent:

```swift
if let string = (value as? String)?.lowercaseString,
       value = StatmentModeConfiguration(rawValue: string) {
    self = value
} else {
    throw ConfigurationError.UnknownConfiguration
}
```

Source/SwiftLintFramework/Rules/StatementPositionRule.swift
```diff
@@ -41,19 +42,85 @@ public struct StatementPositionRule: CorrectableRule, ConfigurationProviderRule
         ]
     )
 
+    public static let uncuddledDescription = RuleDescription(
+        identifier: "statement_position",
+        name: "Statment Position",
+        description: "Else and catch should on the next line, with equal indentation to the " +
```
jpsim: missing "be":

> Else and catch should be on the next line

Source/SwiftLintFramework/Rules/StatementPositionRule.swift
```diff
@@ -41,19 +42,85 @@ public struct StatementPositionRule: CorrectableRule, ConfigurationProviderRule
...
+        "previous declaration.",
```
jpsim: could you please indent this string continuation? Just like the previous `description` above.

Source/SwiftLintFramework/Rules/StatementPositionRule.swift
```diff
@@ -41,19 +42,85 @@ public struct StatementPositionRule: CorrectableRule, ConfigurationProviderRule
...
                 location: Location(file: file, characterOffset: range.location))
         }
     }
 
-    public func correctFile(file: File) -> [Correction] {
-        let matches = violationRangesInFile(file, withPattern: pattern)
+    func standardViolationRangesInFile(file: File, withPattern pattern: String) -> [NSRange] {
+        return file.matchPattern(pattern).filter { range, syntaxKinds in
+            return syntaxKinds.startsWith([.Keyword])
+            }.flatMap { $0.0 }
```
jpsim: please de-indent this one level.

Source/SwiftLintFramework/Rules/StatementPositionRule.swift
```diff
@@ -68,17 +135,117 @@ public struct StatementPositionRule: CorrectableRule, ConfigurationProviderRule
...
+        }
+    }
+
+    static func uncuddledMatchFilter(contents contents: String, syntaxMap: SyntaxMap) ->
+        ((NSTextCheckingResult) -> Bool) {
+        return { match in
+            let range = match.range
+            guard let matchRange = contents.NSRangeToByteRange(start: range.location,
+                                                               length: range.length) else {
+                                                                return false
```
jpsim: could you indent this just one level deeper than the `guard`? I find SourceKit's default here to be absolutely bananas.

jpsim:
On the whole this looks great @ateliercw! I just have a few comments. An entry in the changelog would be good too.

ateliercw:
@jpsim adding the the change log entry made me realize that "default" is a more sensible name than "standard", and I made the other formatting and code tweaks that you requested.

jpsim:
Really great first contribution, @ateliercw! 👏

Issue #643, opened 2016-04-28
bug: legacy_cggeometry_functions autocorrect generates extraneous argument label


BluMist:
`CGRectContainsRect(area, point)` 
gets autocorrected to:
`area.contains(point: point)
`
the `point:` is extraneous

AviSergey:
Hello. Is there any progress here?
I has the same issue. Swiftlint version is 0.10.0. For example:
`CGRectContainsPoint(frame, point)`
gets autocorrected to:
`frame.contains(point: point)`
where Extraneous argument label 'point:' in call

PR #690, opened 2016-06-10
Fix extraneous arguments label in legacy cggeometry functions

bsarr007:
This PR fix https://github.com/realm/SwiftLint/issues/643.


jpsim:
Looks good to me, thanks!

Issue #646, opened 2016-04-29
Autocorrecting UIEdgeInsetsMake legacy constructor doesn't work with variables as parameters


Noobish1:
Lines like `UIEdgeInsetsMake(topInset, 0, bottomInset, 0)` where variables are used instead of numbers for arguments makes autocorrect fail.

PR #692, opened 2016-06-11
Allow variables in legacy constructor rule

bsarr007:
This PR fix https://github.com/realm/SwiftLint/issues/646.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+## Master
+
+##### Breaking
+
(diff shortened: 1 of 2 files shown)


jpsim:
Great work, thank you! Changelog entry is in the wrong place though.

bsarr007:
@jpsim YES Good catch! it's updated, is it ok?

jpsim:
Yes, thanks for the fantastic contribution!

PR #697, opened 2016-06-15
Remove reference of `CuddledElseRule.swift` from SwiftLint.xcodeproj

norio-nomura:
That should be removed on 7a56338.


jpsim:
thanks!

PR #707, opened 2016-06-23
Fix spelling for `declarationReturns` function

epologee:
While trying to figure out why SwiftLint was telling me my documentation wasn't valid, I noticed a typo in `ValidDocsRule.swift`.

I couldn't figure out why my docs aren't valid, but the least I could do was fix the typo 😄.


jpsim:
> I couldn't figure out why my docs aren't valid, but the least I could do was fix the typo 😄.

😂 thanks!

PR #711, opened 2016-06-24
⚠️ WIP - Native plugins support

zats:
## Into

This feature came to be as a possible solution to our need to run GraphQL queries validation against scheme allowing queries to be an organic part of the `.swift` source file and an external file. We wanted it to be part of the build process. All these factors made swiftlint an ideal candidate. However builtin rule system while powerful doesn't allow for code execution allow us to achieve it.
As a potential use case, I can see external plugins for spell checking of strings and comments that would allow `swiftlint autocorrect` to automatically fix spelling mistakes as a part of the build process etc. Also it would open door to adhoc, domain specific and niche rules without need to contribute it back to swiftlint.
## Plugins infrastructure

Plugin is a bundle with a principle class conforming to the `Rule` type. For now it has to be `class` and not a `struct` for the runtime to be able to address it. I used `NSBundle` infrastructure to load external code. 
Plugin should be linked against `SwiftLintFramework` to be able to address appropriate types which might be the problem if the framework is being updated and plugin is not. I'd like to hear your opinion on this one.
This PR comes with a sample rule that compiles into `.plugin` bundle and can be referenced when running `rules`, `autocorrect` and `lint` commands by adding flag `--plugins ~/swiftlint/plugins/PuppetPlugin.plugin`. This plugin doesn't do match, it allows to specify whether it should fail and what is the severity of the failure. I used it as an example and for integration tests.
## Integration tests

Since plugin infrastructure is a layer I added mostly on top of the command line app and not the framework I found that it is not that easy to test the app itself (or rather I couldn't find an easy way). I decided to go with a completely opaque approach of launching `swiftlint` binary that can be found in the tests bundle folder. But if you think it's better to test the swiftlint.app I'm open to suggestions.
## Open issues

As you can see, PuppetPlugin is configured not to include swift libraries relying on the swiftlint use the same version of swift. Result of running plugin that uses different version of Swift against swiftlint is undefined, and, to be honest is not quite clear to me.

While I understand this feature might not be something authors of the library planned for, I think that added value can enrich the capabilities of swiftlint with features that might be beyond traditional scope of linting. I'd appreciate your feedback on this.


zats:
I see that CI is failing due to missing swift libraries, I tried to add it (by specifying `EMBEDDED_CONTENT_CONTAINS_SWIFT` = `YES`) but it didn't help. Please let me know if there is something particular I should know about travis setup. `make tests` passes locally.

jpsim:
This is a large change and will take some time for me to 1) fully grasp what the end goal is here and 2) how it's been implemented. In the meantime, if you'd like to troubleshoot why the build is failing, I'd get this to run on CI commit-by-commit. You can open a scratch PR clearly labeled as WIP to play around there if you won't want to mess with the commits in this branch.

zats:
Okay, let me know if I can clarify any particular aspects of this PR, I agree it did get quite big.

norio-nomura:
Sorry for absence.
On testing, `@executable_path` is `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/Library/Xcode/Agents/xctest`.
You need based on `@loader_path`.

zats:
Hi @norio-nomura, thanks a lot for your help!

I think the failure has to do with the fact I created "integration tests" target that finds `swiftlint` binary in the test product folder and calls various commands (using `NSTask`) on it directly observing results.
Every attempt to call commands on CI fails with 
```
dyld: Library not loaded: @rpath/libswiftCore.dylib
Referenced from: /Users/travis/Library/Developer/Xcode/DerivedData/SwiftLint-guwhnovfcgsqoagacmxaqgxzqosy/Build/Intermediates/CodeCoverage/Products/Debug/swiftlint
Reason: image not found
```

All the targets I added have `LD_RUNPATH_SEARCH_PATHS` set to `“$(inherited) @executable_path/../Frameworks @loader_path/../Frameworks"`

To be honest, I’m not sure where else can I set this flag :)

Tests/swiftlintTests/Helpers/SwiftLintExecutable.swift
```diff
@@ -0,0 +1,56 @@
...
+    case LintError(String)
+}
+
+struct SwiftLintExecutable {
+
+    private let path: String
+
+    init() {
+        let products = testBundle.bundleURL.URLByDeletingLastPathComponent!
+        let executable = products.URLByAppendingPathComponent("swiftlint")
```
norio-nomura: Changing this to:
```swift
        let executable = products
            .URLByAppendingPathComponent("swiftlint.app/Contents/MacOS/swiftlint")
```
will fix the dyld issue.

zats: Bah, wanted to get as close to the real setup as possible, your suggestion makes sense though, let me try that on CI

zats: Thanks a lot! That fixed the Xcode part of the tests. 👍 
I'll take a look at SPM failure

norio-nomura:
Plugin should have settings as following:
```
EMBEDDED_CONTENT_CONTAINS_SWIFT = NO;
SKIP_INSTALL = YES;
```

zats:
Won't `SKIP_INSTALL` prevent plugin from being archived into a bundle?
Will revert `EMBEDDED_CONTENT_CONTAINS_SWIFT` – that was an unsuccessful attempt to fix `dylib` error you helped with earlier!

norio-nomura:
`SKIP_INSTALL = YES;` will prevent `SwiftLint.pkg` from installing `/Library/Bundle/PuppetPluging.plugin`.
`EMBEDDED_CONTENT_CONTAINS_SWIFT = NO` does not affect to fix of `dyld` issue. Swift Runtime Libraries embedded into `PuppetPluging.plugin` is never used because `swiftlint` already loaded them before loading plugins.

zats:
Sorry, I still don't understand the `SKIP_INSTALL` part.
Given that `PuppetPlugin` is not a dependency of `swiftlint` target (it is
a dependency of the new test target though), would it still be installed as
the part of the `SwiftLint.pkg`?
Also, if I disable `SKIP_INSTALL`, let's say I want to archive and export
just the `PuppetPlugin.plugin` to distribute it, wouldn't this flag prevent
me from doing so?

zats:
Sorry, I still don't understand the `SKIP_INSTALL` part.
Given that `PuppetPlugin` is not a dependency of `swiftlint` target (it is
a dependency of the new test target though), would it still be installed as
the part of the `SwiftLint.pkg`?
Also, if I disable `SKIP_INSTALL`, let's say I want to archive and export
just the `PuppetPlugin.plugin` to distribute it, wouldn't this flag prevent
me from doing so?

norio-nomura:
You can verify that `SwiftLint.pkg` contains `PuppetPlugin.plugin` or not by following:
```console
$ DEVELOPER_DIR=/Applications/Xcode.app make package
…
$ pkgutil --payload-files SwiftLint.pkg
.
./Applications
./Library
./Library/Bundles
./Library/Bundles/PuppetPlugin.plugin
./Library/Bundles/PuppetPlugin.plugin/Contents
./Library/Bundles/PuppetPlugin.plugin/Contents/_CodeSignature
./Library/Bundles/PuppetPlugin.plugin/Contents/_CodeSignature/CodeResources
./Library/Bundles/PuppetPlugin.plugin/Contents/Info.plist
./Library/Bundles/PuppetPlugin.plugin/Contents/MacOS
./Library/Bundles/PuppetPlugin.plugin/Contents/MacOS/PuppetPlugin
./Library/Frameworks
…
```

norio-nomura:
If we should distribute `PuppetPlugin.plugin` with SwiftLint, we will need to select proper location of the plugins.
I can't say it should do or not at now.

zats:
Sorry that was my bad, I added PuppetPlugin as a dependency of swiftlint target. 
Removing it as a dependency (and leaving `SKIP_INSTALL=NO`) prevents it from being added to the `pkg`. 
As for the plugin itself - it is meant as a sample if anyone wants to build their own and also being used in the tests.
Hope this makes sense

zats:
As far as plugins location - originally I though that is something that every user decides for herself.
Let's say I wrote a spellchecker plugin. 
I can put it up on GitHub.
Everyone who is interested in using it can download it, place it in any folder on their computer, let's say `.swiftlint` folder in the root of the project and run `swiftlint autocorrect --plugins .swiftlint/autocorrect.plugin` to make sure that it will be invoked (along with default rules)

norio-nomura:
> Also, if I disable SKIP_INSTALL, let's say I want to archive and export
just the PuppetPlugin.plugin to distribute it, wouldn't this flag prevent
me from doing so?

I understand that.
But are you expecting that plugin developer should fork this repository for developing their plugins?
I guess it would be better for plugin developer if plugin project will be separated into another repository.

norio-nomura:
> I guess it would be better for plugin developer if plugin project will be separated into another repository.

Sorry, my worries may be needless on early stage.

zats:
No need to apologize, I appreciate you take your time to analyze various aspects of this feature.
But you are right, this plugin is not intended to be anything more than a subject of tests and maybe a sample code (although I'm sure if this feature will end up being merged, I'd probably come up with a better example)

zats:
What seems to be the last issue I'm running into with `spm_test` is following:

In Xcode my test target set up to build dependent plugin which archives into `.plugin` bundle with compiled code. Without archived `.plugin` bundle in the test suite folder, tests will fail.
If my understanding is correct `swift build` won't read Xcode project configuration, however I couldn't find a way to express this dependency otherwise. Is the best way to add another step to `spm_test` that would build this dependency explicitly?
Sorry for the noob questions - I'm struggling finding any references regarding this 😳

norio-nomura:
AFAIK, `swift build` does not support bundle creation.
I think we should disable codes for plugins.

zats:
Excuse me I think I missed the point of your last comment, what do you mean by "disable codes for plugin"? 
I think it should be possible to build a plugin in a way that it is going to be compiled into a binary with no need for bundle infrastructure on top, i.e.:

```swift
class SwiftLintPlugin {
  var pluginClass: Class {
     return SpellCheckerPlugin.self
  }
}

class SpellCheckerPlugin: Rule {
  // ....
}
```

norio-nomura:
Sorry for my bad english. 🙇 

> I think we should disable codes for plugins.

I meant "on `make spm_test`."

norio-nomura:
`swift build` can't create `PuppetPlugin.plugin` as loadable bundle by using `NSBundle(path: pluginPath)`.
If we would test loading plugin on `make spm_test`, we will need to build that separately by using `xcodebuild`.

zats:
No worries! English is far from being my first language, too 🙂 
You point regarding `xcodebuild` sounds fair. I'll try to find a way to get plugins to work without relying on bundles.

zats:
Sorry for the lack of work on this, I just pushed a bunch of changes. As you suggested, I excluded plugins tests from when testing using `spm`. 

I also rewrote a plugin infrastructure not to require bundles, although given that `spm` doesn't test this aspect right now, bundle shouldn't be an issue.

And just to make plugin slightly less useless, I replaced with the one who checks the copyrights message covers current year ¯\_(ツ)_/¯

jpsim:
This PR is a lot to chew on. Hoping to meet up with @zats so we can go through it together and I can get a better grasp of everything involved here.

marcelofabri:
@zats @jpsim What's the outcome of this? Is that something we want to revive?

zats:
Sorry, I stopped working on this. Feel free to pick it up if it looks interesting, I'm down to help, but probably won't have time to drive this.

marcelofabri:
@zats Would love if you could take a look at #2529 and see if it'd solve your use case.

jpsim:
Let's continue the conversation in #2529. Sorry @zats that this never got the attention it deserved.

PR #714, opened 2016-07-01
Fix a bug in EmptyCountRule when using a variable named count

pylebecq:
Hello guys, I found out the empty count rule is triggered when declaring this king of method:

```
func testEmptyCount(count: Int) -> Bool {
    return count <= 0
}
```

I tried to fix it.


jpsim:
Sorry, but I believe this false positive is better than missing some violations. From a previous comment on a PR with the same change from a few months ago (#635):

> the reason I kept the false positive here is to catch it in cases where you're extending `CollectionType`s, where accessing `self.count` without `self`. So I'm inclined not to accept this change, especially because I'm happy to have a higher tolerance for false positives in opt-in rules.

Still, thanks for your contribution @pylebecq, I hope you contribute again in the future!

Issue #672, opened 2016-05-23
Discourage calling `init()` directly


joshfriend:
The apple language docs always use `Foo()` style over `Foo.init()`. While both do the same thing, using `.init()` seems superfluous. 

A check for this rule could find instances of the text `.init` and verify that it is not preceded by either `self`, or `super`.

mbrandonw:
You would also want to allow things like `optionalString.flatMap(NSURL.init(string:))`.

joshfriend:
Hmm, yes that would make this more tricky...

norio-nomura:
implemented on #715 and #826

PR #715, opened 2016-07-03
Add ExplicitInitRule

mtaube:
Hey guys, I added a quick rule to discourage the explicit use of ``init`` (https://github.com/realm/SwiftLint/issues/672). The current implementation only catches ``init``s that are prefixed by a well-formed class / struct name.

```
// bad
let abc = Abc.init()

// good
let abc = Abc()
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@
 * Adds `allow_private_set` configuration for the `private_outlet` rule.  
   [Rohan Dhaimade](https://github.com/HaloZero)
 
+* Add ``ExplicitInitRule`` Opt-In rule to discourage calling ``init``
+  directly.  
+  [Matt Taube](https://github.com/mtaube)
+  [#715](https://github.com/realm/SwiftLint/pull/715)
+
 ##### Bug Fixes
 
 * Fixed regex bug in Vertical Whitespace Rule by using SourceKitten instead.
--- a/Source/SwiftLintFramework/Rules/ExplicitInitRule.swift
+++ b/Source/SwiftLintFramework/Rules/ExplicitInitRule.swift
@@ -1,0 +1,75 @@
+//
+//  ExplicitInitRule.swift
+//  SwiftLint
+//
+//  Created by Matt Taube on 7/2/16.
+//  Copyright (c) 2016 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+public struct ExplicitInitRule: ConfigurationProviderRule, CorrectableRule, OptInRule {
+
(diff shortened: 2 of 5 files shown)


jpsim:
Do people really write `Abc.init()` superfluously? Since there are legitimate uses (e.g. `optionalString.flatMap(NSURL.init(string:))`) that would trigger this warning, I'd be inclined to not add this rule.

mtaube:
I've seen some people at work do it, especially developers migrating from Obj-C. Also I think it might be because of Xcode autocomplete combined with laziness. Which is why I want to enforce with this rule on everybody.

What do you think about:
1. Making it opt-in
2. Creating a blacklist for the rule to allow for the ``.init``, like the NSURL. I can't think of other cases where this would trigger false positives.

jpsim:
We could make this opt-in. I'm generally more ok with false positives with opt-in rules.

Don't special case NSURL, there's nothing unique about it when it comes to using its initializer as a function reference.

mtaube:
@jpsim made this rule opt-in / correctable, lmk what you think

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
   [bootstraponline](https://github.com/bootstraponline)
   [#689](https://github.com/realm/SwiftLint/issues/689)
 
+* Add ExplicitInitRule.
```
jpsim: missing two trailing spaces according to the guidelines at https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes

Source/SwiftLintFramework/Rules/ExplicitInitRule.swift
```diff
@@ -0,0 +1,75 @@
+//
+//  TodoRule.swift
```
jpsim: This should be "ExplicitInitRule.swift".

Source/SwiftLintFramework/Rules/ExplicitInitRule.swift
```diff
@@ -0,0 +1,75 @@
...
+//  SwiftLint
+//
+//  Created by JP Simard on 2015-05-16.
```
jpsim: Please change this to your name and the date the file was added.

jpsim:
I'm still pretty dubious of the value of such a rule, but since it's opt-in and if some find it useful, then why not.

@mtaube please address my latest round of comments and rebase this so I can review again.

mtaube:
@jpsim made those changes. Lmk if anything else is needed. Thanks for SwiftLint / Realm by the way – we use both at work.

norio-nomura:
Thanks for doing this! :pray:
I have rewritten this as `ASTRule` for reducing false positive on #826, and merged that.

Issue #696, opened 2016-06-15
Ability to set a threshold of warnings to throw an error


Nitner0C:
I've looked for this in the issues, but haven't found it...

It would be cool to set a warning threshold to raise an error

```yml
warning_threshold: 20
```

jpsim:
This is fairly easy to build, so @Nitner0C I'd encourage you to try implementing this yourself if you're looking for ways to contribute 😄

woodhamgh:
I would love this too!

@jpsim Could you point me in the right direction? Id love to look into contributing :)

woodhamgh:
Ive prototyped something up, but it just counts the warnings thrown by swiftlint. 

Would you know if there is anyway we could pass the number of warnings currently in the project from xcode to swiftlint when we start linting. The reason is

Say we have 5 warnings in the project and then when we lint we get another 15 warnings I would like to throw an error as the total is 20 or above

@jpsim

marcelofabri:
I think this can be closed as #757 was merged

norio-nomura:
🙏

PR #716, opened 2016-07-07
Gw add warning threshold rule

woodhamgh:
# What <h3> 
Rule to check the number of warnings thrown by Swiftlint and to throw an error if the number of warnings is greater than the threshold.

# How <h3> 
After all the checks have been run the the list of violations is passed to this rule which adds a new entry if the number of warnings is above the set threshold.

Can be configured in the config file with
 `warning_threshold: 5`
otherwise the default of 10 is used.

# Who <h3>
https://github.com/woodhamgh

# Why <h3> 
This allows for some warnings to go through without causing build failures but limits the number of warnings within the project. Some workplaces might have their own reasons for this or for how many they are happy to allow through.

Issue where request was made:
https://github.com/realm/SwiftLint/issues/696

### DIFF
--- a/Source/SwiftLintFramework/Models/Linter.swift
+++ b/Source/SwiftLintFramework/Models/Linter.swift
@@ -28,9 +28,13 @@ public struct Linter {
         }
         let regions = file.regions()
         var ruleTimes = [(id: String, time: Double)]()
+        var warningRule: WarningThresholdRule?
         let violations = rules.flatMap { rule -> [StyleViolation] in
             if !(rule is SourceKitFreeRule) && self.file.sourcekitdFailed {
                 return []
+            }
+            if rule is WarningThresholdRule {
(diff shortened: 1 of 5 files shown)


jpsim:
Interesting idea to use rules as general Swiftlint parameterization! That's creative 😉. Though I do think a `Configuration` parameter would be more appropriate here.

woodhamgh:
Ah yup. I will look for an example of that and update this :) thanks!

woodhamgh:
Hey @jpsim I think I have updated everything so it is now a Configuration parameter but I am having trouble working out how I will throw the error in say xcode.

Previously I was just adding a new StyleViolation but I am not sure what is going to be appropriate here.

Any thoughts?

woodhamgh:
https://github.com/realm/SwiftLint/pull/757

PR #722, opened 2016-07-12
 Readme update with link to contributing.md

masters3d:
- I don't think I need to add this to the change log. Just adding a little more discoverability to the contributing doc.


jpsim:
Thanks!

PR #724, opened 2016-07-13
autocorrect fix of locally disabled rules

### LINKED ISSUE #601: Autocorrect corrects parts that are wrapped with swiftlint-disable
Running `swiftlint autocorrect` will fix lines that are wrapped with `swiftlint-disable/enable`.

For instance:

```
 // swiftlint:disable comma
 let cancelOptions = [
     XLFormOptionsObject(value: TestConfig.CancelMode.NotCancelled.rawValue, displayText: "Not cancelled"),
-    XLFormOptionsObject(value: TestConfig.CancelMode.ByThem.rawValue,       displayText: "Cancelled by them"),
-    XLFormOptionsObject(value: TestConfig.CancelMode.ByMe.rawValue,         displayText: "Cancelled by me")
+    XLFormOptionsObject(value: TestConfig.CancelMode.ByThem.rawValue, displayText: "Cancelled by them"),
+    XLFormOptionsObject(value: TestConfig.CancelMode.ByMe.rawValue, displayText: "Cancelled by me")
 ]
 // swiftlint:enable comma
```

SwiftLint 0.9.2

masters3d:
This code needs to be added to all the correctable rules that don't have it, but I think we should probably add this to the File type  or Correctable protocol so that we could just say
```
if currentRuleIsDisabled{
   return []
}
```
closes #601

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 ##### Enhancements
 
 * LeadingWhitespaceRule is now auto correctable.  
-  [masters3d](https://github.com/masters3d)
+  [J. Cheyo Jimenez](https://github.com/masters3d)
 
 * Add included regex for custom rules to control what files are processed.  
   [bootstraponline](https://github.com/bootstraponline)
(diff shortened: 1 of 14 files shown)


masters3d:
I went ahead and generalized this on the correctable protocol. 

I am thinking we should probably do something like this with the protocol. 
```swift
public protocol CorrectableRule: Rule {
    func correctFile(file: File) -> [Correction] // this has a default implementation
    func correctFileConfiguration(file: File) -> [Correction] // this gets implemented
}

extension CorrectableRule {
    private func isRuleDisabled(file: File) -> Bool {}// private default implementation
    func correctFile(file: File) -> [Correction]{
        if isRuleDisabled(file) { return [] }
        correctFileConfiguration(file)
}
```

masters3d:
Closing this:  I am disabling the rules per file based and not per region based. I will update and submit a new PR.

PR #725, opened 2016-07-13
Add JUnit reporter

matthewellis:
I needed to pull test results from SwiftLint into Bamboo CI but the current reporters where not compatible with the parsers available, JUnit seems to be a common format for a few tools so i thought i would put in a pull request to add this to the project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ##### Enhancements
 
+* Add Junit reporter.
+  [Matthew Ellis](https://github.com/matthewellis)
+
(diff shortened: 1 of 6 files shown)


jpsim:
Could you please add tests to `ReporterTests.swit`?

jpsim:
This will also need a changelog entry according to `CONTRIBUTING.md`, and to update the `README.md`. Otherwise looks good. Thanks!

matthewellis:
Hey @jpsim thanks for the feedback,  i have made the changes you asked for.

jpsim:
Looks great, thanks!

Issue #730, opened 2016-07-25
Colon rules acting differently for same situation. (+ custom question)


freak4pc:
Hey There :)

In many places in my code I prefer having my colons attached to the identifier, but not with a single space to the type. I think this requires a custom rule (Something like `identifier(no spaces):(multiple spaces)type`), but I'm still not fluent enough in Swiftlint to write this rule

The odd part is the current rule seems picky on this and sometimes lets these kind of statements pass... I'm not sure what's the difference between the first line that wasn't detected as a "warning" to the other two, which are. 

![screen shot 2016-07-25 at 11 38 05 am](https://cloud.githubusercontent.com/assets/605076/17095369/c79234a8-525c-11e6-81e6-c39153e259b3.png)

Also, if you could guide me on how to write a rule that will allow the following detection (meaning disabled `colon_rules` and writing something custom that allows multiple spaces *after* the colon) I'd be happy to read up on it.

Thank you!

jpsim:
It's odd that only 2 of the 3 warnings are rendered for you.... 🤔

I can't reproduce that:

```
$ swiftlint version
0.11.0
$ cat main.swift
public class Location {
  /// Store Location Address.
  dynamic public var address:    String!

  /// Store Location City.
  dynamic public var city:       String!

  /// Store Location Country.
  dynamic public var country:    String!
}
$ swiftlint
Linting Swift files in current working directory
Linting 'main.swift' (1/1)
main.swift:3:22: warning: Colon Violation: Colons should be next to the identifier when specifying a type. (colon)
main.swift:6:22: warning: Colon Violation: Colons should be next to the identifier when specifying a type. (colon)
main.swift:9:22: warning: Colon Violation: Colons should be next to the identifier when specifying a type. (colon)
Done linting! Found 3 violations, 0 serious in 1 file.
```

---

As for making changes to rules to support this kind of style, I'm all for it!

The way I'd recommend this be done is to make `ColonRule` configurable, maybe with a `flexible_right_spacing` boolean flag that allows one _or more_ spaces to the right of the colon.

freak4pc:
I'm down for trying to write that but I'm still not very acquainted with
how the rule mechanism works.

Gonna read a bit into it and see how well my brain works today 🤓

freak4pc:
Well that was easier than I thought: https://github.com/realm/SwiftLint/pull/733.

[...]

PR #734, opened 2016-07-26
Allow setting flexible_right_spacing for colon rule.

freak4pc:
Resolves issue #730.

### DIFF
--- a/Source/SwiftLintFramework/Rules/ColonRule.swift
+++ b/Source/SwiftLintFramework/Rules/ColonRule.swift
@@ -18,7 +18,7 @@ public struct ColonRule: CorrectableRule, ConfigurationProviderRule {
     public var flexibleRightSpacing = false
 
     public init(configuration: AnyObject) throws {
-        flexibleRightSpacing = configuration["flexible_right_spacing"] as? Int == 1
+        flexibleRightSpacing = configuration["flexible_right_spacing"] as? Bool == true
     }
 
     public var configurationDescription: String {
--- a/Source/SwiftLintFramework/Rules/ColonRule.swift
+++ b/Source/SwiftLintFramework/Rules/ColonRule.swift
@@ -15,6 +15,16 @@ public struct ColonRule: CorrectableRule, ConfigurationProviderRule {
 
     public init() {}
 
+    public var flexibleRightSpacing = false
+
+    public init(configuration: AnyObject) throws {
+        flexibleRightSpacing = configuration["flexible_right_spacing"] as? Int == 1
+    }
+
+    public var configurationDescription: String {
+        return "flexible_right_spacing: \(flexibleRightSpacing)"
+    }
+
     public static let description = RuleDescription(
         identifier: "colon",
         name: "Colon",
@@ -110,19 +120,24 @@ public struct ColonRule: CorrectableRule, ConfigurationProviderRule {
 
     // MARK: - Private
 
-    private let pattern =
-        "(\\w)" +              // Capture an identifier
-        "(?:" +                // start group
-        "\\s+" +               // followed by whitespace
-        ":" +                  // to the left of a colon
-        "\\s*" +               // followed by any amount of whitespace.
-        "|" +                  // or
-        ":" +                  // immediately followed by a colon
-        "(?:\\s{0}|\\s{2,})" + // followed by 0 or 2+ whitespace characters.
-        ")" +                  // end group
-        "(" +                  // Capture a type identifier
-        "[\\[|\\(]*" +         // which may begin with a series of nested parenthesis or brackets
-        "\\S)"                 // lazily to the first non-whitespace character.
+    private var pattern: String {
+        // If flexible_right_spacing is true, match only 0 whitespaces.
+        // If flexible_right_spacing is false or omitted, match 0 or 2+ whitespaces.
+        let spacingRegex = flexibleRightSpacing ? "(?:\\s{0})" : "(?:\\s{0}|\\s{2,})"
+
+        return  "(\\w)" +       // Capture an identifier
+                "(?:" +         // start group
+                "\\s+" +        // followed by whitespace
+                ":" +           // to the left of a colon
+                "\\s*" +        // followed by any amount of whitespace.
+                "|" +           // or
+                ":" +           // immediately followed by a colon
+                spacingRegex +  // followed by right spacing regex
+                ")" +           // end group
+                "(" +           // Capture a type identifier
+                "[\\[|\\(]*" +  // which may begin with a series of nested parenthesis or brackets
+                "\\S)"          // lazily to the first non-whitespace character.
+    }
 
     private func violationRangesInFile(file: File, withPattern pattern: String) -> [NSRange] {
         let nsstring = file.contents as NSString
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ##### Enhancements
 
+* Allow setting `flexible_right_spacing` configuration for the `colon` rule.  
+  [Shai Mishali](https://github.com/freak4pc)
+  [#730](https://github.com/realm/SwiftLint/issues/730)
+
 * Add Junit reporter.  
   [Matthew Ellis](https://github.com/matthewellis)
 
--- a/Tests/SwiftLintFramework/RulesTests.swift
+++ b/Tests/SwiftLintFramework/RulesTests.swift
@@ -85,7 +85,7 @@ class RulesTests: XCTestCase {
             ]
         )
 
-        verifyRule(description, ruleConfiguration: ["flexible_right_spacing": 1])
+        verifyRule(description, ruleConfiguration: ["flexible_right_spacing": true])
     }
     // swiftlint:enable function_body_length
 
--- a/Tests/SwiftLintFramework/RulesTests.swift
+++ b/Tests/SwiftLintFramework/RulesTests.swift
@@ -15,9 +15,79 @@ class RulesTests: XCTestCase {
         verifyRule(ClosingBraceRule.description)
     }
 
+    // swiftlint:disable function_body_length
     func testColon() {
+        // Verify Colon rule with test values for when flexible_right_spacing
+        // is false (default).
         verifyRule(ColonRule.description)
+
+        // Verify Colon rule with test values for when flexible_right_spacing
+        // is true.
+        let description = RuleDescription(
+            identifier: "colon",
+            name: "Colon",
+            description: "Colons should be next to the identifier when specifying a type.",
+            nonTriggeringExamples: [
+                "let abc: Void\n",
+                "let abc: [Void: Void]\n",
+                "let abc: (Void, Void)\n",
+                "let abc: ([Void], String, Int)\n",
+                "let abc: [([Void], String, Int)]\n",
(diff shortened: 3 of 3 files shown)


freak4pc:
Reopened new pull request to trigger another Travis build due to a bug in the previous PR build for some reason.

jpsim:
One of the upstream's dependencies' commit sha (Quick's) must have disappeared. I just updated dependencies in #735, so if you rebase this branch off `master`, you'll inherit those changes and this PR should then pass CI.

freak4pc:
All green lights ! 🎉  
@jpsim feel free to go through this and let me know if anything seems off.

Source/SwiftLintFramework/Rules/ColonRule.swift
```diff
@@ -15,6 +15,16 @@ public struct ColonRule: CorrectableRule, ConfigurationProviderRule {
 
     public init() {}
 
+    public var flexibleRightSpacing = false
+
+    public init(configuration: AnyObject) throws {
+        self.flexibleRightSpacing = configuration["flexible_right_spacing"] as? Int == 1
```
jpsim: Doesn't this work?

```swift
self.flexibleRightSpacing = configuration["flexible_right_spacing"] == true
```

If so, that's a more accurate representation of what we're checking, no?

Source/SwiftLintFramework/Rules/ColonRule.swift
```diff
@@ -15,6 +15,16 @@ public struct ColonRule: CorrectableRule, ConfigurationProviderRule {
...
+    }
+
+    public var configurationDescription: String {
+        return "flexible_right_spacing: \(self.flexibleRightSpacing)"
```
jpsim: we avoid the use of `self.` when it can be omitted: https://github.com/github/swift-style-guide#only-explicitly-refer-to-self-when-required

Source/SwiftLintFramework/Rules/ColonRule.swift
freak4pc: No, this won't work. 

configuration contains `AnyObject` keys. 

Initial option was 
`configuration["flexible_right_spacing"] as? Bool == true`

But it won't convert the YAML `true` to a Swift `true`.

Source/SwiftLintFramework/Rules/ColonRule.swift
freak4pc: 👍

jpsim:
We should test with `flexible_right_spacing` turned on. The test code isn't super well set up to do this at the moment, but it is being done for other rules. For example: https://github.com/realm/SwiftLint/blob/0.11.1/Tests/SwiftLintFramework/RulesTests.swift#L129-L144

freak4pc:
@jpsim I don't mind adding a test case specific for `flexible_right_space` but it will have to have it's own test values inlined. Is that fine?

jpsim:
I think adding another special test with its own set of triggering/non-triggering examples is fine for now, though long term we might want to refactor the testing infrastructure to allow for config-specific tests.

freak4pc:
I'm having some problems with the unit tests.

Tried to just get rid of everything except for basic examples in the triggering examples in `ColonRule` , so now it looks something like this that should always fails: 

```swift
triggeringExamples: [
    "let ↓abc:Void\n",
    "let ↓abc :Void\n",
    "let ↓abc : Void\n"
],
```

I left non-triggering the same as they shouldn't affect the result, but running this still fails tests: 

```swift
verifyRule(ColonRule.description, ruleConfiguration: ["flexible_right_spacing": 1],
                                  commentDoesntViolate: false)
```

I'm not very sure on what's causing this, any guides? 

```
/Users/freak4pc/Work/Libs/SwiftLint/Tests/SwiftLintFramework/TestHelpers.swift:135: error: -[SwiftLintFrameworkTests.RulesTests testColon] : XCTAssertEqual failed: ("Optional(0)") is not equal to ("Optional(3)") - 
/Users/freak4pc/Work/Libs/SwiftLint/Tests/SwiftLintFramework/TestHelpers.swift:61: error: -[SwiftLintFrameworkTests.RulesTests testColon] : XCTAssertEqual failed: ("Optional(0)") is not equal to ("Optional(1)") - 
/Users/freak4pc/Work/Libs/SwiftLint/Tests/SwiftLintFramework/TestHelpers.swift:68: error: -[SwiftLintFrameworkTests.RulesTests testColon] : XCTAssertEqual failed: ("Optional("let abc:  ([Void], String, Int)\n")") is not equal to ("Optional("let abc: ([Void], String, Int)\n")") - 
/Users/freak4pc/Work/Libs/SwiftLint/Tests/SwiftLintFramework/TestHelpers.swift:71: error: -[SwiftLintFrameworkTests.RulesTests testColon] : XCTAssertEqual failed: ("Optional("let abc:  ([Void], String, Int)\n")") is not equal to ("Optional("let abc: ([Void], String, Int)\n")") - 
/Users/freak4pc/Work/Libs/SwiftLint/Tests/SwiftLintFramework/TestHelpers.swift:61: error: -[SwiftLintFrameworkTests.RulesTests testColon] : XCTAssertEqual failed: ("Optional(0)") is not equal to ("Optional(1)") - 
/Users/freak4pc/Work/Libs/SwiftLint/Tests/SwiftLintFramework/TestHelpers.swift:68: error: -[SwiftLintFrameworkTests.RulesTests testColon] : XCTAssertEqual failed: ("Optional("func abc(def:  Void) {}\n")") is not equal to ("Optional("func abc(def: Void) {}\n")") - 
/Users/freak4pc/Work/Libs/SwiftLint/Tests/SwiftLintFramework/TestHelpers.swift:71: error: -[SwiftLintFrameworkTests.RulesTests testColon] : XCTAssertEqual failed: ("Optional("func abc(def:  Void) {}\n")") is not equal to ("Optional("func abc(def: Void) {}\n")") - 
/Users/freak4pc/Work/Libs/SwiftLint/Tests/SwiftLintFramework/TestHelpers.swift:61: error: -[SwiftLintFrameworkTests.RulesTests testColon] : XCTAssertEqual failed: ("Optional(0)") is not equal to ("Optional(1)") - 
/Users/freak4pc/Work/Libs/SwiftLint/Tests/SwiftLintFramework/TestHelpers.swift:68: error: -[SwiftLintFrameworkTests.RulesTests testColon] : XCTAssertEqual failed: ("Optional("let abc:  [([Void], String, Int)]\n")") is not equal to ("Optional("let abc: [([Void], String, Int)]\n")") - 
/Users/freak4pc/Work/Libs/SwiftLint/Tests/SwiftLintFramework/TestHelpers.swift:71: error: -[SwiftLintFrameworkTests.RulesTests testColon] : XCTAssertEqual failed: ("Optional("let abc:  [([Void], String, Int)]\n")") is not equal to ("Optional("let abc: [([Void], String, Int)]\n")") - 
/Users/freak4pc/Work/Libs/SwiftLint/Tests/SwiftLintFramework/TestHelpers.swift:61: error: -[SwiftLintFrameworkTests.RulesTests testColon] : XCTAssertEqual failed: ("Optional(0)") is not equal to ("Optional(1)") - 
/Users/freak4pc/Work/Libs/SwiftLint/Tests/SwiftLintFramework/TestHelpers.swift:68: error: -[SwiftLintFrameworkTests.RulesTests testColon] : XCTAssertEqual failed: ("Optional("let abc:  (Void, String, Int)\n")") is not equal to ("Optional("let abc: (Void, String, Int)\n")") - 
/Users/freak4pc/Work/Libs/SwiftLint/Tests/SwiftLintFramework/TestHelpers.swift:71: error: -[SwiftLintFrameworkTests.RulesTests testColon] : XCTAssertEqual failed: ("Optional("let abc:  (Void, String, Int)\n")") is not equal to ("Optional("let abc: (Void, String, Int)\n")") - 
/Users/freak4pc/Work/Libs/SwiftLint/Tests/SwiftLintFramework/TestHelpers.swift:61: error: -[SwiftLintFrameworkTests.RulesTests testColon] : XCTAssertEqual failed: ("Optional(0)") is not equal to ("Optional(1)") - 
/Users/freak4pc/Work/Libs/SwiftLint/Tests/SwiftLintFramework/TestHelpers.swift:68: error: -[SwiftLintFrameworkTests.RulesTests testColon] : XCTAssertEqual failed: ("Optional("let abc:  Void\n")") is not equal to ("Optional("let abc: Void\n")") - 
/Users/freak4pc/Work/Libs/SwiftLint/Tests/SwiftLintFramework/TestHelpers.swift:71: error: -[SwiftLintFrameworkTests.RulesTests testColon] : XCTAssertEqual failed: ("Optional("let abc:  Void\n")") is not equal to ("Optional("let abc: Void\n")") - 
```

freak4pc:
I have gone as low as this but with same issues 

```swift
        identifier: "colon",
        name: "Colon",
        description: "Colons should be next to the identifier when specifying a type.",
        nonTriggeringExamples: [
            "let abc: Void\n"
        ],
        triggeringExamples: [
            "let ↓abc:Void\n",
            "let ↓abc :Void\n",
            "let ↓abc : Void\n"
        ],
        corrections: [
            "let abc:Void\n": "let abc: Void\n"
        ]
    )
```

freak4pc:
Ah ha, corrections was the issue. Getting this sorted out and updating the PR.

freak4pc:
Ok tests look good from here. @jpsim 

I'm not 100% comfortable with having to force `swiftlint:disable function_body_length`, but without that I can't have the Configuration inside the test unit so I think for the time being that's a minor sacrifice to have. 

We could also move the configuration outside of the function but that feels like an unwanted contamination of RulesTests.swift. 

Another option would be to have `ColonRule.flexibleRightSpacingDescription` but that is also not optimal. 

In general, as you mentioned, this should all be a relatively temporary method until there's time to make a more flexible test solution for configuration-based testing.

Source/SwiftLintFramework/Rules/ColonRule.swift
```diff
@@ -15,6 +15,16 @@ public struct ColonRule: CorrectableRule, ConfigurationProviderRule {
 
     public init() {}
 
+    public var flexibleRightSpacing = false
+
+    public init(configuration: AnyObject) throws {
+        flexibleRightSpacing = configuration["flexible_right_spacing"] as? Int == 1
```
jpsim: did you try changing this to a bool check, which would be more correct IMO?

```swift
flexibleRightSpacing = configuration["flexible_right_spacing"] == true
```

freak4pc: Hey, Yes I did ...
You asked this already and I answered but I can't seem to find that comment, I guess it got lost. 

You can't directly compare configuration["flexible_right_spacing"] since configuration contains `AnyObject`s (`cannot convert value of type 'AnyObject?!' to expected argument type 'Bool') 

I also rememebr I tried something like this but YAML's `true` didn't convert to Swift `Bool` and this was always false-y.  

`flexibleRightSpacing = configuration["flexible_right_spacing"] as? Bool == true`

freak4pc: Uhm interesting. 
`as? Bool == true` works OK now. 

Adding another commit to modify this.

jpsim:
👍 thanks for your sustained work on this @freak4pc!

freak4pc:
My pleasure ! Thanks for merging 😇

PR #737, opened 2016-07-30
Removed ConditionalBindingCascadeRule

### LINKED ISSUE #701: Should `conditional_binding_cascade` be removed?
[SE-0099](https://github.com/apple/swift-evolution/blob/master/proposals/0099-conditionclauses.md) has been accepted and we will be able to freely mix boolean conditions and optional bindings.

The syntax for comma separated bindings without `let` is no longer supported, which means this rule enforces behavior which will lead to more work when moving to a future version of Swift. Should it be removed?

masters3d:
closes #701


freak4pc:
Looks good here +1

jpsim:
Thanks @masters3d! Merged in #741

PR #751, opened 2016-08-11
[CONTRIBUTING.md] Small addendum about submodules & tests

AliSoftware:
### Submodules

Starting to want to contribute so read the `CONTRIBUTING.md` then forked, but saw the `xcworkspace` were showing some xcodeproj projects in red… until I realized I forgot to clone the submodules.

### Unit Tests

Another thing I wondered when implementing my first rule is wether I should write a Unit Test explicitly myself or not.

Especially, given the paragraph about providing Unit Tests via the `triggeringExamples` and `nonTriggeringExamples`, and the fact that there's a `masterRuleList` list containing all the rules already, I wondered if the Unit Tests code weren't already auto-executing the tests for all the rules without me having to write anything additional.

That's only when seeing that my `validateFile` wasn't executed during the Rules Unit Tests that I figured I had to explicitly add one myself, which would explicitly call `verifyRule(MyNewRule.description)`

That's a small question and I quickly figured the answer myself by looking at the code, but still when opening a unknown project for the first time, every pointer & explicit doc helps.

---

So I figured a little reminder in the `CONTRIBUTING.md` might help others avoid the same mistakes :wink:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,7 +29,9 @@ Whenever possible, prefer adding tests via the `triggeringExamples` and
 `nonTriggeringExamples` properties of a rule's `description` rather than adding
 those test cases in the unit tests directly. This makes it easier to understand
 what rules do by reading their source, and simplifies adding more test cases
-over time.
+over time. This way adding a unit test for your new Rule is just a matter of
+adding a test case in `RulesTests.swift` which simply calls
+`verifyRule(YourNewRule.description)`.
 
 ### `ConfigurationProviderRule`
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,10 +1,20 @@
 ## Pull Requests
 
 All changes, no matter how trivial, must be done via pull request. Commits
-should never be made directly on the `master` branch. If you have commit access
-to SwiftLint and believe your change to be trivial and not worth waiting for
-review, you may open a pull request and merge immediately, but this should be
-the exception, not the norm.
+should never be made directly on the `master` branch.
+
+_If you have commit access to SwiftLint and believe your change to be trivial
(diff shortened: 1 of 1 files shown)


masters3d:
I am a little confused by this because this was not my experience when cloning with github web UI. 
I just hit cloned and I was up and running. At first I was working off the the xcode projects instead of the workspace and that didn't really work. Once I switched to the workspace it was a breeze.

AliSoftware:
I personally did a `git clone https://github.com/realm/SwiftLint.git` in my terminal directly (not using the GitHub client app nor any third-party GIT client for the cloning), maybe that's the difference?

masters3d:
Do you have carthage installed? If you look inside the `Makefile` in the root folder you will see this. 

```
archive:
	carthage build --no-skip-current --platform mac
	carthage archive SwiftLintFramework
```
I am thinking carthage is required. I have have carthage installed.

AliSoftware:
I do have Carthage installed.

I don't think it's required though, at least for contributing to the project (like just forking to add a Rule and run the Unit Tests), because SwiftLint dependencies are pulled via submodules, without having to explicitly run Carthage to pull & build them.

Also note that I haven't run `make` on SwiftLint at all. Just forked it, `git clone`'d it, opened the workspace, and already added my own Rule — and a unit test for it — and been able to Command-U on it to try it on. Didn't need to run either `make` or Carthage.

masters3d:
Ah you are right, The web page github UI automatically seems to clone the submodules (makes sense). 
https://github.com/blog/2104-working-with-submodules

jpsim:
Fantastic, thanks for documenting this @AliSoftware! I'm 100% sure that I would have stumbled on this if I was trying to work with SwiftLint for the first time too.

PR #752, opened 2016-08-11
Add content of the todo statement to message #478

### LINKED ISSUE #478: Can the todo rule put the content of the todo statement in the warning?
It'd be incredibly helpful to have the text immediately after a todo statement be placed in the warning, so you don't have to go into each warning to find out what the todo is about.

masters3d:
closes #478

### DIFF
--- a/Source/SwiftLintFramework/Rules/TodoRule.swift
+++ b/Source/SwiftLintFramework/Rules/TodoRule.swift
@@ -41,14 +41,48 @@ public struct TodoRule: ConfigurationProviderRule {
         ]
     )
 
+    private func customMessage(lines: [Line], location: Location) -> String {
+            var reason = self.dynamicType.description.description
+
+            guard let lineIndex = location.line,
+                  let currentLine = lines.filter({ $0.index == lineIndex }).first
+                  else { return reason }
+
+            // customizing the reason message to be specific to fixme or todo
+            var message = currentLine.content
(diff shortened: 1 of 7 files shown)


Source/SwiftLintFramework/Rules/TodoRule.swift
```diff
@@ -41,14 +41,49 @@ public struct TodoRule: ConfigurationProviderRule {
         ]
     )
 
+    // swiftlint:disable conditional_binding_cascade
```
jpsim: this rule was removed, so I'd rather not add this, only to have to remember to remove it at some point down the line.

jpsim:
Other than the comment which I don't think should be added, 👍.

jpsim:
This also needs a rebase.

masters3d:
Please use github's `squash and merge`
I tried squashing to one commit but then I get branch conflicts with `CHANGELOG`

Also Travis is failing because brew's version is not updated.
```
  testSwiftLintLints, failed - Repeated `let` statements in conditional binding cascade should be avoided.
```

jpsim:
The diff shows that something went wrong during merging or rebasing. You'll need to address that first.

I wonder if there's an elegant solution to address Travis having an older version installed...

masters3d:
Yeah, I butchered this. I am going to close it for now

PR #753, opened 2016-08-11
[New Rule] weak_delegate

AliSoftware:
Added a rule to enforce `delegate` instance variables to be declared as `weak` to avoid reference cycles.

### DIFF
--- a/Source/SwiftLintFramework/Rules/WeakDelegateRule.swift
+++ b/Source/SwiftLintFramework/Rules/WeakDelegateRule.swift
@@ -42,9 +42,10 @@ public struct WeakDelegateRule: ASTRule, ConfigurationProviderRule {
         }
 
         // Check if name contains "delegate"
-        let name = (dictionary["key.name"] as? String) ?? ""
-        let isDelegate = name.lowercaseString.rangeOfString("delegate") != nil
-        guard isDelegate else { return [] }
+        guard let name = (dictionary["key.name"] as? String) where
+            name.lowercaseString.containsString("delegate") else {
+                return []
+        }
 
         // Check if non-weak
         let attributes = (dictionary["key.attributes"] as? [SourceKitRepresentable])?
--- a/Source/SwiftLintFramework/Rules/WeakDelegateRule.swift
+++ b/Source/SwiftLintFramework/Rules/WeakDelegateRule.swift
@@ -22,7 +22,9 @@ public struct WeakDelegateRule: ASTRule, OptInRule, ConfigurationProviderRule {
             "class Foo {\n  weak var delegate: SomeProtocol?\n}\n",
             "class Foo {\n  weak var someDelegate: SomeDelegateProtocol?\n}\n",
             "class Foo {\n  weak var delegateScroll: ScrollDelegate?\n}\n",
+            // We only consider properties to be a delegate if it has "delegate" in its name
             "class Foo {\n  var scrollHandler: ScrollDelegate?\n}\n",
+            // Only trigger on instance variables, not local variables
             "func foo() {\n  var delegate: SomeDelegate\n}\n"
         ],
         triggeringExamples: [
--- a/Source/SwiftLintFramework/Rules/WeakDelegateRule.swift
+++ b/Source/SwiftLintFramework/Rules/WeakDelegateRule.swift
@@ -9,7 +9,7 @@
 import Foundation
 import SourceKittenFramework
 
-public struct WeakDelegateRule: ASTRule, OptInRule, ConfigurationProviderRule {
+public struct WeakDelegateRule: ASTRule, ConfigurationProviderRule {
     public var configuration = SeverityConfiguration(.Warning)
 
     public init() {}
--- a/Source/SwiftLintFramework/Rules/WeakDelegateRule.swift
+++ b/Source/SwiftLintFramework/Rules/WeakDelegateRule.swift
@@ -1,0 +1,69 @@
+//
+//  WeakDelegate.swift
+//  SwiftLint
+//
+//  Created by Olivier Halligon on 11/08/2016.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+public struct WeakDelegateRule: ASTRule, OptInRule, ConfigurationProviderRule {
+    public var configuration = SeverityConfiguration(.Warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "weak_delegate",
+        name: "Weak Delegate",
+        description: "Delegates should be weak to avoid reference cycles.",
+        nonTriggeringExamples: [
+            "class Foo {\n  weak var delegate: SomeProtocol?\n}\n",
+            "class Foo {\n  weak var someDelegate: SomeDelegateProtocol?\n}\n",
+            "class Foo {\n  weak var delegateScroll: ScrollDelegate?\n}\n",
+            "class Foo {\n  var scrollHandler: ScrollDelegate?\n}\n",
+            "func foo() {\n  var delegate: SomeDelegate\n}\n"
+        ],
+        triggeringExamples: [
+            "class Foo {\n  var delegate: SomeProtocol?\n}\n",
+            "class Foo {\n  var scrollDelegate: ScrollDelegate?\n}\n",
+            "class Foo {\n  var delegateScroll: ScrollDelegate?\n}\n",
+        ]
+    )
+
+    public func validateFile(file: File,
+                             kind: SwiftDeclarationKind,
+                             dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard kind == .VarInstance else {
+            return []
+        }
+
+        // Check if name contains "delegate"
+        let name = (dictionary["key.name"] as? String) ?? ""
+        let isDelegate = name.lowercaseString.rangeOfString("delegate") != nil
+        guard isDelegate else { return [] }
+
+        // Check if non-weak
+        let attributes = (dictionary["key.attributes"] as? [SourceKitRepresentable])?
+            .flatMap({ ($0 as? [String: SourceKitRepresentable]) as? [String: String] })
+            .flatMap({ $0["key.attribute"] }) ?? []
+        let isWeak = attributes.contains("source.decl.attribute.weak")
+        guard !isWeak else { return [] }
+
+        // Violation found!
+        let location: Location
+        if let offset = (dictionary["key.offset"] as? Int64).flatMap({ Int($0) }) {
+            location = Location(file: file, byteOffset: offset)
+        } else {
+            location = Location(file: file.path)
+        }
+
+        return [
+            StyleViolation(ruleDescription: self.dynamicType.description,
+                severity: configuration.severity,
+                location: location
+            )
+        ]
+    }
+}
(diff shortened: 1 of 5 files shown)


AliSoftware:
Not sure why the job testing in Xcode succeeds but the one using SPM crashes.

It seems to be [related to SourceKitCrashTests](https://travis-ci.org/realm/SwiftLint/jobs/151671637#L705) but not sure it's my code and my fault or SourceKit's instability's fault

_Especially since I didn't use any force-cast `as!` or the like but used `as?` everywhere instead, I'm wondering how my code would have made SourceKit crash, even more so it doesn't crash when run via Xcode and not SPM._

Any help resolving that would be appreciated. Or maybe that's just a matter or relaunching the Travis job and sacrificing a chicken to SourceKit's God, idk?

Source/SwiftLintFramework/Rules/WeakDelegateRule.swift
```diff
@@ -0,0 +1,71 @@
...
+    public func validateFile(file: File,
+                             kind: SwiftDeclarationKind,
+                             dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard kind == .VarInstance else {
+            return []
+        }
+
+        // Check if name contains "delegate"
+        let name = (dictionary["key.name"] as? String) ?? ""
+        let isDelegate = name.lowercaseString.rangeOfString("delegate") != nil
```
masters3d: this should probably be one guard line 
```
guard let _ = name.lowercaseString.rangeOfString("delegate") else { return [] }
```

Source/SwiftLintFramework/Rules/WeakDelegateRule.swift
```diff
@@ -0,0 +1,71 @@
...
+        // Check if non-weak
+        let attributes = (dictionary["key.attributes"] as? [SourceKitRepresentable])?
+            .flatMap({ ($0 as? [String: SourceKitRepresentable]) as? [String: String] })
+            .flatMap({ $0["key.attribute"] }) ?? []
+        let isWeak = attributes.contains("source.decl.attribute.weak")
+        guard !isWeak else { return [] }
+
+        // Violation found!
+        let location: Location
+        if let offset = (dictionary["key.offset"] as? Int64).flatMap({ Int($0) }) {
```
masters3d: Why Int64 and not just Int? (this may be why SMP is failing?)

Source/SwiftLintFramework/Rules/WeakDelegateRule.swift
```diff
@@ -0,0 +1,71 @@
...
+            return []
+        }
+
+        // Check if name contains "delegate"
+        let name = (dictionary["key.name"] as? String) ?? ""
+        let isDelegate = name.lowercaseString.rangeOfString("delegate") != nil
+        guard isDelegate else { return [] }
+
+        // Check if non-weak
+        let attributes = (dictionary["key.attributes"] as? [SourceKitRepresentable])?
```
masters3d: Do you have a way to check if the declaration is Optional and a Protocol from source kitten?  Perhaps you can simplify it by using a regex 

```swift
        let lines = file.lines
        return file.matchPattern(".*[dD]elegate.*:*\\s*\\?").flatMap { range, syntaxKinds in
            let location = Location(file: file, characterOffset: range.location)
            guard let lineIndex = location.line,
                   let currentLine = lines.filter({ $0.index == lineIndex }).first
                   else { return nil }
            guard !currentLine.content.containsString("weak") else { return nil }

            return StyleViolation(ruleDescription: self.dynamicType.description,
                location: location)
        }
    }

```

image

Source/SwiftLintFramework/Rules/WeakDelegateRule.swift
AliSoftware: Because I shamelessly copy pasted that part from an existing rule :blush:

Source/SwiftLintFramework/Rules/WeakDelegateRule.swift
AliSoftware: No a RegEx isn't enough:

1) The RegEx like the one you mentioned would need a negative look-behind expression to ensure the weak keyword isn't present (and this generally means non-optimal CPU-intensive backtracking in the RegEx engine)
2) Not sure if SwiftLint easily allows to match patterns against a whole line…while still being able to limit the matching to instance variable declarations only and not for example local variable?
3) The fact that the declaration is optional and a protocol isn't enough condition. With your code you'll catch some false positives

To be fair at first I tried to implement that using `grep` and not using SwiftLint but had some false positives as it didn't allow me to know enough context about the line. Then I tried a SwiftLint RegEx-based `custom_rule` but hit the limitations explained above. That's why I finally switched to an `ASTRule` to have more control over the analysis and context and be sure to annoys false positives.

Source/SwiftLintFramework/Rules/WeakDelegateRule.swift
AliSoftware: I find `guard let _ = x` less readable than `guard x != nil` personally

Source/SwiftLintFramework/Rules/WeakDelegateRule.swift
AliSoftware: Mmmm wait after re-reading your comment I realize you actually do exclude "weak" in a separate line of the code, that's what I missed and why I didn't understand why you didn't have negative-look-behind in your RegEx.

But still even if we switch to a RegEx rule, I'm not sure I see the advantage of migrating to a RegEx over checking the AST:

* as the AST seems more sure and explicit to me (less likely we forget to match a specific syntax that would represent the same thing, or that our RegEx would be too greedy and catch the "delegate" word in another part of the expression
* For example your regex wouldn't warn on `var delegate: SomeProtocol` if the developer forgot both the `weak` and the optionality of the type… while it should indeed warn for this case too
* I'm not sure a RegEx match using SwiftLint would be able to take the context into account and only match if it's part of a instance variable declaration only? Like, wouldn't your code wrongly match (and warn for) `func foo() { var delegate: SomeDelegate? }` too?
* For example your RegEx would also match and warn for `var scrollHandler: ScrollHandler // maybe we should use delegate: ScrollDelegate? instead?`
* And another example is that your code would consider the lines `var scrollDelegate: ScrollDelegate? // maybe should be weak?` and `var weakLinkDelegate: WeakLinkDelegate?` as valid because they contain "weak" despite not being actually weak.

For all these reasons I prefer the AST analysis, to be sure we consider the right parts of the expression (the name for "delegate" and the attributes for "weak") and the right context "only instance variable")

Source/SwiftLintFramework/Rules/WeakDelegateRule.swift
AliSoftware: I just tested by typecasting directly to `Int` and the cast always fails:


```swift
if let offset = (dictionary["key.offset"] as? Int) {
  // never true
}
```

That's because `dictionary["key.offset"] is a `SourceKitRepresentable` and `Int64` are `SourceKitRepresentable` but it appears plain `Int`s aren't (at least not directly). Hence the first cast to `Int64` then building an `Int` from that `Int64` (like it's done everywhere else in other rules, actually).

jpsim:
@AliSoftware could you please rebase this?

AliSoftware:
Done… But it seems that the SPM build still fails at making SourceKitService crash… (and I don't understand how that's related to my changes in that PR) 😒

jpsim:
I'm investigating that SPM failure now. What's odd is that I can't reproduce it locally...

jpsim:
Huh, I can inconsistenly reproduce this locally...

AliSoftware:
As far as I remember, every time I pushed a commit on that PR and a travis build was trigger, the CI had the error; I don't think it ever passed. But maybe we just got very unlucky on the CI, even if it's a non deterministic failure?

jpsim:
I'm very confused. We obviously can't merge this PR with this failure though. We could of course modify the sourcekit tests to pass by ignoring this new rule, but that would just be ignoring the problem.

AliSoftware:
Could it be because my code is too safe somehow? (I tend to avoid force-casting for example)

Idk how you try to make SourceKitService crash but if it's by passing incorrect SourceKitRepresentable dictionaries to the rules maybe that's it?

Or the CI using an older/different version of the Swift snapshot?

jpsim:
> Could it be because my code is too safe somehow?

😂

> Idk how you try to make SourceKitService crash but if it's by passing incorrect SourceKitRepresentable dictionaries to the rules maybe that's it?

No, we don't actually make SourceKit crash in the unit tests, but rather _pretend_ it crashed: https://github.com/realm/SwiftLint/blob/master/Tests/SwiftLintFramework/SourceKitCrashTests.swift#L39-L47

```swift
let file = File(contents: "A file crashed SourceKitService")
file.sourcekitdFailed = true

var assertHandlerCalled = false
file.assertHandler = { assertHandlerCalled = true }

_ = file.structure
XCTAssertTrue(assertHandlerCalled, "Expects assert handler was called on accessing File.structure")
```

AliSoftware:
What I don't get is that the log report crashes on `testAssetHandlerIsNotCalledOnNormalFile` and similar other tests… which don't depend on the rules at all!

So how come adding my rule in this PR could have an impact on those unit tests if they don't depend on there rules themselves?

jpsim:
I really don't know.

norio-nomura:
I can not reproduce the failures of SourceKitCrashTests on my Mac. 😕

masters3d:
I believe those crashes are related to some random caching weirdness. 
https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Extensions/File%2BCache.swift
I also noticed some random test cases failing for no reason locally. The only thing they had in common is that they touched the cache at some point.

AliSoftware:
Very interesting analysis.

Could the fact that `Cache<T>` doesn't appear to be thread-safe and that the test might be executed in parallel explain such randomness in the crashes through race conditions?

masters3d:
Maybe the cache is invalidated non deterministically. Would `autoreleasepool` have anything to do with it?  I am out of ideas. 
https://github.com/realm/SwiftLint/blob/master/Source/swiftlint/Commands/LintCommand.swift#L44-L53

norio-nomura:
I believe that the failing `SourceKitCrashTests` has been fixed on #837.

Source/SwiftLintFramework/Rules/WeakDelegateRule.swift
```diff
@@ -0,0 +1,72 @@
...
+        guard let name = (dictionary["key.name"] as? String) where
+            name.lowercaseString.containsString("delegate") else {
+                return []
+        }
+
+        // Check if non-weak
+        let attributes = (dictionary["key.attributes"] as? [SourceKitRepresentable])?
+            .flatMap({ ($0 as? [String: SourceKitRepresentable]) as? [String: String] })
+            .flatMap({ $0["key.attribute"] }) ?? []
+        let isWeak = attributes.contains("source.decl.attribute.weak")
```
AliSoftware: I'm not convinced on this. `delegates` usually have to be `weak`, not `unowned`, as a best practice. But I can understand that some people want to allow it anyway. Maybe provide a rule configuration to opt-in for `unowned`?

marcelofabri: I agree that they're usually `weak`, but I just faced a problem where the `delegate` is passed on `init` and I didn't wanted to make it optional. Sure, I could rename it but I still think it is a delegate 😅 

But a rule configuration would be enough for me 💯

jpsim:
Merged in #895

PR #754, opened 2016-08-12
[New Rule] private_outlet

AliSoftware:
Added an Opt-In rule to enforce `@IBOutlet` instance variables to be declared `private`.

Some teams (including mine) want to enforce that `@IBOutlet` are not exposed outside of the declaring type because for separation of concerns principles one might not want to leak UI details to the outside of a class.

### Rationale

I personally find it more clean to expose a `String` property to the outside — even if that property implement a `set` and `get` which in fact link to the internal `label.text` — rather than exposing the `UILabel` itself — because if some time later we change that UILabel to a non-editing UITextField or some other custom UI component for design reasons, the consumer of the class shouldn't bother (that's implementation details of the UI layer not to be exposed to the higher abstraction layer)

### Opt-In rule

That being said, knowing that this rationale depends on each team's policy and that other teams might not have similar policies regarding IBOutlets, and that this rule is debatable depending on how strict you want your team to separate the various layers, I think that this rule should be Opt-In.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@
 
 ##### Enhancements
 
+* Add `PrivateOutletRule` Opt-In rule to enforce `@IBOutlet`
+  instance variables to be `private`.  
+  [Olivier Halligon](https://github.com/AliSoftware)
+  [#754](https://github.com/realm/SwiftLint/pull/754)
+
 * Add `LegacyNSGeometryFunctionsRule` rule. Add `NSSize`, `NSPoint`, and 
   `NSRect` constants and constructors to existing rules.  
   [David Rönnqvist](https://github.com/d-ronnqvist)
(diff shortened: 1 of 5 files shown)


Source/SwiftLintFramework/Models/MasterRuleList.swift
```diff
@@ -73,5 +73,6 @@ public let masterRuleList = RuleList(rules:
     TypeBodyLengthRule.self,
     TypeNameRule.self,
     ValidDocsRule.self,
-    VariableNameRule.self
+    VariableNameRule.self,
+    PrivateOutletRule.self
```
jpsim: this list is in alphabetical order

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 ##### Enhancements
 
+* Add `PrivateOutletRule` Opt-In rule to enforce `@IBOutlet`
+  instance variables to be `private`.  
+  [Olivier Halligon](https://github.com/AliSoftware)
+  [#754](https://github.com/realm/SwiftLint/pull/754)
```
jpsim: Please remove this line. The contributing doc states that only issues _addressed_ by this PR should be listed here: https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes

jpsim:
👍 thanks @AliSoftware! I'll address the my two minor comments in an upcoming commit.

PR #757, opened 2016-08-14
Warning threshold rule

woodhamgh:
# What <h3> 
Rule to check the number of warnings thrown by Swiftlint and to throw an error if the number of warnings is greater than the threshold.

# How <h3> 
New configuration parameter which adds and error into the list of violations if the threshold is broken. Done in the lint command 

Can be configured in the config file with
 `warning_threshold: 5`

# Who <h3>
https://github.com/woodhamgh

# Why <h3> 
This allows for some warnings to go through without causing build failures but limits the number of warnings within the project. Some workplaces might have their own reasons for this or for how many they are happy to allow through.

Issue where request was made:
https://github.com/realm/SwiftLint/issues/696

### DIFF
--- a/Source/SwiftLintFramework/Models/Configuration.swift
+++ b/Source/SwiftLintFramework/Models/Configuration.swift
@@ -76,9 +76,7 @@ public struct Configuration: Equatable {
         }
 
         // set the config threshold to the threshold provided in the config file
-        if let warningThreshold = warningThreshold {
-            self.warningThreshold = warningThreshold
-        }
+        self.warningThreshold = warningThreshold
 
         // white_list rules take precendence over all else.
         if !whitelistRules.isEmpty {
@@ -150,7 +148,7 @@ public struct Configuration: Equatable {
             whitelistRules: defaultStringArray(dict[ConfigurationKey.WhitelistRules.rawValue]),
             included: defaultStringArray(dict[ConfigurationKey.Included.rawValue]),
             excluded: defaultStringArray(dict[ConfigurationKey.Excluded.rawValue]),
-            warningThreshold: dict[ConfigurationKey.WarningThreshold.rawValue] as? Int ?? nil,
+            warningThreshold: dict[ConfigurationKey.WarningThreshold.rawValue] as? Int,
             reporter: dict[ConfigurationKey.Reporter.rawValue] as? String ??
                 XcodeReporter.identifier,
             configuredRules: masterRuleList.configuredRulesWithDictionary(dict)
--- a/Source/SwiftLintFramework/Models/Configuration.swift
+++ b/Source/SwiftLintFramework/Models/Configuration.swift
@@ -20,6 +20,7 @@ private enum ConfigurationKey: String {
     case Reporter = "reporter"
     case UseNestedConfigs = "use_nested_configs" // deprecated
     case WhitelistRules = "whitelist_rules"
+    case WarningThreshold = "warning_threshold"
 }
 
 public struct Configuration: Equatable {
@@ -27,6 +28,7 @@ public struct Configuration: Equatable {
     public let included: [String]             // included
     public let excluded: [String]             // excluded
     public let reporter: String               // reporter (xcode, json, csv, checkstyle)
+    public var warningThreshold: Int?         // warning threshold
     public let rules: [Rule]
     public var rootPath: String?              // the root path to search for nested configurations
     public var configurationPath: String?     // if successfully loaded from a path
@@ -36,6 +38,7 @@ public struct Configuration: Equatable {
                  whitelistRules: [String] = [],
                  included: [String] = [],
                  excluded: [String] = [],
+                 warningThreshold: Int? = nil,
                  reporter: String = XcodeReporter.identifier,
                  configuredRules: [Rule] = masterRuleList.configuredRulesWithDictionary([:])) {
         self.included = included
@@ -70,6 +73,11 @@ public struct Configuration: Equatable {
                 "configuration error: '\(rule.0)' is listed \(rule.1) times"
             }.joinWithSeparator("\n"))
             return nil
+        }
+        
+        // set the config threshold to the threshold provided in the config file
+        if let warningThreshold = warningThreshold {
+            self.warningThreshold = warningThreshold
         }
 
         // white_list rules take precendence over all else.
@@ -128,6 +136,7 @@ public struct Configuration: Equatable {
             .Reporter,
             .UseNestedConfigs,
             .WhitelistRules,
+            .WarningThreshold
         ].map({ $0.rawValue }) + masterRuleList.list.keys
 
         let invalidKeys = Set(dict.keys).subtract(validKeys)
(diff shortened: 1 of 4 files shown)


masters3d:
```
    ✗ testSwiftLintLints, failed - Function body should span 40 lines or less excluding comments and whitespace: currently spans 41 lines

```

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -72,6 +75,11 @@ public struct Configuration: Equatable {
             return nil
         }
 
+        // set the config threshold to the threshold provided in the config file
+        if let warningThreshold = warningThreshold {
```
jpsim: this conditional is unnecessary

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -141,6 +150,7 @@ public struct Configuration: Equatable {
             whitelistRules: defaultStringArray(dict[ConfigurationKey.WhitelistRules.rawValue]),
             included: defaultStringArray(dict[ConfigurationKey.Included.rawValue]),
             excluded: defaultStringArray(dict[ConfigurationKey.Excluded.rawValue]),
+            warningThreshold: dict[ConfigurationKey.WarningThreshold.rawValue] as? Int ?? nil,
```
jpsim: the `?? nil` is unnecessary.

Source/swiftlint/Commands/LintCommand.swift
```diff
@@ -55,7 +55,10 @@ struct LintCommand: CommandType {
             violations += currentViolations
             reporter.reportViolations(currentViolations, realtimeCondition: true)
         }.flatMap { files in
-            reporter.reportViolations(violations, realtimeCondition: false)
+            if isWarningThresholdBroken(configuration, violations: violations) {
+                violations.append(createThresholdViolation(configuration.warningThreshold!))
+            }
+            reporter.reportViolations(violations, realtimeCondition: true)
```
jpsim: why did you change `realtimeCondition` from `false` to `true`?

Source/swiftlint/Commands/LintCommand.swift
```diff
@@ -119,3 +122,27 @@ struct LintOptions: OptionsType {
             <*> mode <| quietOption(action: "linting")
     }
 }
+
+private func isWarningThresholdBroken(configuration: Configuration,
+                                      violations: [StyleViolation]) -> Bool {
+    guard let warningThreshold = configuration.warningThreshold else { return false }
+    let numberOfWarningViolations = violations.filter({ $0.severity == .Warning}).count
+    if numberOfWarningViolations >= warningThreshold {
```
jpsim: these 4 lines should just be `return numberOfWarningViolations >= warningThreshold`.

Source/swiftlint/Commands/LintCommand.swift
```diff
@@ -119,3 +122,27 @@ struct LintOptions: OptionsType {
...
+        identifier: "warning_threshold",
+        name: "Warning Threshold",
+        description: "Number of warnings thrown is above the threshold."
+    )
+    let location = Location(file: "", line: 0, character: 0)
+    return StyleViolation(
+        ruleDescription: description,
+        severity: ViolationSeverity.Error,
+        location: location,
+        reason: "Number of warnings exceeded threshold of \(threshold)")
```
jpsim: Missing period at the end of the string.

Source/swiftlint/Commands/LintCommand.swift
```diff
@@ -119,3 +122,27 @@ struct LintOptions: OptionsType {
...
+private func createThresholdViolation(threshold: Int) -> StyleViolation {
+    let description = RuleDescription(
+        identifier: "warning_threshold",
+        name: "Warning Threshold",
+        description: "Number of warnings thrown is above the threshold."
+    )
+    let location = Location(file: "", line: 0, character: 0)
+    return StyleViolation(
+        ruleDescription: description,
+        severity: ViolationSeverity.Error,
```
jpsim: `severity: .Error` please

jpsim:
Could you please add a changelog entry? Also, a test that verifies that an error is reported if the warning threshold is exceeded would be nice.

Source/swiftlint/Commands/LintCommand.swift
woodhamgh: Hey, when it was false the new error would not show up in the reporting. Im guessing because the report was already generated and so it didn't generate again without True.

Happy to change if you know how this could be done differently

Source/swiftlint/Commands/LintCommand.swift
woodhamgh: done

Source/swiftlint/Commands/LintCommand.swift
woodhamgh: done

Source/swiftlint/Commands/LintCommand.swift
woodhamgh: done, sorry

Source/SwiftLintFramework/Models/Configuration.swift
woodhamgh: done

Source/SwiftLintFramework/Models/Configuration.swift
woodhamgh: done

Source/swiftlint/Commands/LintCommand.swift
jpsim: Could you please double-check that? It should definitely be reported.

woodhamgh: Yup so changed to false and the error no longer shows up. It still adds to the array but then I get 

"Command /bin/sh failed with exit code 2" whereas if it is true I get an error shown and then the done linting comment (in xcode)

jpsim: Ah, I know what's happening... this code should be this:

```swift
if isWarningThresholdBroken(configuration, violations: violations) {
  violations.append(createThresholdViolation(configuration.warningThreshold!))
  if reporter.isRealtime {
    reporter.reportViolations([violations.last!], realtimeCondition: true)
  }
}
reporter.reportViolations(violations, realtimeCondition: false)
```

Though this is pretty confusing and could stand to be refactored, but not in this PR.

woodhamgh: What would you like me to do? Cause i think this func is now at its 40line limit

jpsim: Actually, the following works and keeps the function under 40 lines:

```swift
if isWarningThresholdBroken(configuration, violations: violations) {
  violations.append(createThresholdViolation(configuration.warningThreshold!))
  reporter.reportViolations([violations.last!], realtimeCondition: true)
}
reporter.reportViolations(violations, realtimeCondition: false)
```

woodhamgh: Done. With the tests I cant seem to find anywhere there are tests for the LintCommand.

Would you like me to create a test class in this swiftlint group or make one in the SwiftLintFrameworkTests

jpsim: Oh, I just realized that this is in `LintCommand`, which is in the `swiftlint` target, which isn't tested...

We only have unit tests set up for SwiftLintFramework.

Please disregard my comment about adding a test for this then.

jpsim:
Passed CI, this is going in. Big thank you @woodhamgh!

woodhamgh:
@jpsim Wooh! Big thanks to you too. First open source contribution, super happy! haha

PR #760, opened 2016-08-19
Use path arg as root for include/exclude paths

cfilipov:
SwiftLint assumes paths in the YAML config file are relative to the current directory even when a path is passed as an argument. 

For example, say we have this in our config file:

```YAML
excluded:
  - Frameworks/ThirdParty/Foo
```

Now let's run SwiftLint using Xcode's run button. To do this We configure the Xcode build scheme to pass the following argument on launch:

```
lint --config /Users/cfilipov/MyProject/.swiftlint.yml --path /Users/cfilipov/MyProject
```

This will run SwiftLint using some path in DerivedData as the current working directory. It will read the config file and correctly use `/Users/cfilipov/MyProject` as the path to search for files to lint. However, it will use the current working directory, the derived data directory, when resolving `Frameworks/ThirdParty/Foo` when it would resolve to `/Users/cfilipov/MyProject/Frameworks/ThirdParty/Foo`.

This is not specific to running from Xcode, but this demonstrates a scenario where one could encounter this problem.

This PR fixes the issue, causing SwiftLint to use the path provided in the command line `--path` argument as the root of for the include and exclude paths. If no `--path` is specified then the current directory will be used as before.


Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -181,8 +181,8 @@ public struct Configuration: Equatable {
     public func lintablePathsForPath(path: String,
                                      fileManager: NSFileManager = fileManager) -> [String] {
         let pathsForPath = included.isEmpty ? fileManager.filesToLintAtPath(path) : []
-        let excludedPaths = excluded.flatMap(fileManager.filesToLintAtPath)
-        let includedPaths = included.flatMap(fileManager.filesToLintAtPath)
+        let excludedPaths = excluded.flatMap{fileManager.filesToLintAtPath($0, rootDirectory: self.rootPath)}
```
jpsim: Please omit `self.` since it can be inferred, as per the Swift style guide we follow: https://github.com/github/swift-style-guide#only-explicitly-refer-to-self-when-required

Also, a single space after `{` and before `}`:

```swift
flatMap{ fileManager.filesToLintAtPath($0, rootDirectory: rootPath) }
```

jpsim:
Thanks for this PR @cfilipov! Could you please address my comment and add a changelog entry in the "Breaking Changes" section? Otherwise this is good to merge IMO.

cfilipov:
Looks like I must have screwed something up by merging with upstream/master (I thought the unit tests ran successfully before I pushed). I won't be able to update this PR until I get home tonight.

Source/SwiftLintFramework/Extensions/NSFileManager+SwiftLint.swift
```diff
@@ -9,8 +9,8 @@
 import Foundation
 
 extension NSFileManager {
-    internal func filesToLintAtPath(path: String) -> [String] {
-        let absolutePath = path.absolutePathStandardized()
+    internal func filesToLintAtPath(path: String, rootDirectory: String? = nil) -> [String] {
+        let absolutePath = (path.absolutePathRepresentation(rootDirectory ?? NSFileManager.defaultManager().currentDirectoryPath) as NSString).stringByStandardizingPath
```
jpsim: CI is probably failing because this line is too long.

CHANGELOG.md
```diff
@@ -2,7 +2,9 @@
 
 ##### Breaking
 
-* None.
+* Fixed: SwiftLint assumes paths in the YAML config file are relative to the
+current directory even when `--path` is passed as an argument.
```
jpsim: elsewhere in this doc, multi-line list items are aligned with the first line (i.e. this needs to spaces)

Source/SwiftLintFramework/Extensions/NSFileManager+SwiftLint.swift
cfilipov: I didn't see that, I'll update that as well. What I did see was in [travis](https://travis-ci.org/realm/SwiftLint/jobs/154062676) was this:

```
/Users/travis/build/realm/SwiftLint/Tests/SwiftLintFramework/ConfigurationTests.swift:122:31: method does not override any method from its superclass

        private override func filesToLintAtPath(path: String) -> [String] {
                ~~~~~~~~      ^
```

Which I'm assuming came about from merging with master.

cfilipov: It was both issues. Lines were too long + the above unit test failing to compile. Update coming soon.

jpsim:
Thanks for bringing this back up to speed @cfilipov! I think I'll cut a point release of SwiftLint before merging this, since it's the only breaking change.

cfilipov:
Sounds good @jpsim. Thanks!

PR #761, opened 2016-08-19
Add rule to check for private unit tests

cfilipov:
If an XCTest method is marked private it will be silently skipped, no warning. 

Example:

```Swift
class MyTest: XCTestCase {
    private func testFoo() {
        // this test will not run
    }
    func testBar() {
        // this test will run
    }
}
```

Likewise, if the XCTest subclass is marked private all test methods on that class will be skipped.

```Swift
private class MyOtherTest: XCTestCase {
    // none of the tests from `MyOtherTest` will run
}
```

It is bad practice to _intentionally_ commit a private test for the same reason that it is a bad practice to commit commented-out code. This type of mistake is very easy to miss, so a SwiftLint rule has been created to catch instances of private tests.


Source/SwiftLintFramework/Rules/MissingDocsRule.swift
```diff
@@ -54,17 +54,9 @@ extension File {
 }
 
 public enum AccessControlLevel: String {
```
masters3d: Does this need to be public?

Source/SwiftLintFramework/Rules/PrivateUnitTestRule.swift
```diff
@@ -0,0 +1,182 @@
...
+//  SwiftLint
+//
+//  Created by Cristian Filipov on 8/3/16.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+extension AccessControlLevel {
```
masters3d: I would segregate this extension to only work in this file.

Source/SwiftLintFramework/Rules/PrivateUnitTestRule.swift
```diff
@@ -0,0 +1,182 @@
...
+            "}",
+            "public class FooTest: XCTestCase { " +
+                "func test1() {}; " +
+                "internal func test2() {}; " +
+                "public func test3() {}; " +
+            "}"
+        ],
+        triggeringExamples: [
+            "private ↓class FooTest: XCTestCase { " +
+                "func test1() {}; " +
```
masters3d: its probably better to use `\n` here instead of `;`

Source/SwiftLintFramework/Rules/PrivateUnitTestRule.swift
```diff
@@ -0,0 +1,182 @@
...
+    return className
+}
+
+public class FooTest: NSObject {  }
+
+public struct PrivateUnitTestRule: ASTRule, ConfigurationProviderRule {
+
+    public var configuration: PrivateUnitTestConfiguration = {
+        var configuration = PrivateUnitTestConfiguration(identifier: "private_unit_test")
+        configuration.message = "Unit test marked `private` will be silently skipped."
```
masters3d: What does this mean? `"Unit test marked `private` will be silently skipped."` Skipped from Swift Lint or the actual XCTest will be skipped?

masters3d:
You need to update the `changelog`. 
Does Linux XCTest act the same way?
What about filePrivate? Is the story the same?
https://github.com/apple/swift-evolution/blob/master/proposals/0025-scoped-access-level.md

cfilipov:
> Does Linux XCTest act the same way?

Surprisingly, yes. For Linux, XCTest needs to explictely return all tests, so you would think it would at least result in an error. But if you mark the test as private, it just doesn't get run, also no warning or error.

> What about filePrivate? Is the story the same?

Yes. However, that's a Swift 3 feature, so not part of this PR.

Source/SwiftLintFramework/Rules/MissingDocsRule.swift
cfilipov: > Does this need to be public?

I don't know, it was already public, I didn't change this line, just changed the raw values.

Source/SwiftLintFramework/Rules/MissingDocsRule.swift
```diff
@@ -54,17 +54,9 @@ extension File {
...
-    case Private = "private"
```
jpsim: Actually the distinction between `rawValue` and `sourcekitValue()` was intentional here. We sometimes use `AccessControlLevel.description` in violation messages (I think).

Source/SwiftLintFramework/Rules/PrivateUnitTestRule.swift
```diff
@@ -0,0 +1,182 @@
...
+            }
+
+    }
+
+    /* It's not strictly necessary to check for `private` on classes because a
+     private class will result in `private` on all its members in the AST.
+     However, it's still useful to check the class explicitly because this
+     gives us a more clear error message. If we check only methods, the line
+     number of the error will be that of the function, which may not
+     necessarily be marked `private` but inhereted it from the class access
```
jpsim: "inherited"

jpsim:
Thanks for the PR @cfilipov! Looks great. I just had two minor comments, and this will need to be rebased.

Source/SwiftLintFramework/Rules/MissingDocsRule.swift
cfilipov: I didn't realize that, but indeed it looks like this is the case in the init method of `MissingDocsRule` where the configuration is mapped over. I'll change the rawValue back and add a new init `fromSourcekitValue`. It's worth noting however: `AccessControlLevel` seems to be inconsistent with how it's done in other parts of the code. `KindType` and `SwiftDeclarationKind` both use the sourcekit value strings as their rawValue.

jpsim: It may be better to keep your changes but add explicit conformance to `CustomStringConvertible` and declare `var description: String` with the short/lowercase member names.

cfilipov: I could't actually find specific instances where `description` was used on `AccessControlLevel`, but that's easy to miss since there are several way `description` can be used implicitly that's hard to search for. I went ahead and made it conform to `CustomStringConvertible` as suggested which returns the short strings. I also had to modify `MissingDocsRule.init(configuration:)` because that relied on the rawValue being the shorter version of the string:

```Swift
let acl = array.flatMap(AccessControlLevel.init(description:))
```

Since `AccessControlLevel` needs to be able to be instantiated from the short strings I also head to create an init method for that. Couldn't think of a better name so hopefully `init(description:)` will suffice.

jpsim:
This is fantastic @cfilipov. I really appreciate your work on this.

cfilipov:
Wonderful! Thanks for merging this in! I should probably create a ticket to track the inclusion of the check for `fileprivate` when Swiftlint is ready for Swift 3.

PR #748, opened 2016-08-11
Added disable rule test to autocorrect #601

masters3d:
closes https://github.com/realm/SwiftLint/issues/601

PR #766, opened 2016-08-22
Added disable rule test to autocorrect

### LINKED ISSUE #601: Autocorrect corrects parts that are wrapped with swiftlint-disable

masters3d:
https://github.com/realm/SwiftLint/pull/748
closes #601

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@
   [J. Cheyo Jimenez](https://github.com/masters3d)
   [#772](https://github.com/realm/SwiftLint/issues/772)
 
+* Correctable rules no longer apply corrections if the rule is locally disabled.  
+  [J. Cheyo Jimenez](https://github.com/masters3d)  
+  [#601](https://github.com/realm/SwiftLint/issues/601) 
+
 * Fixed regex bug in Mark Rule where MARK could not be used with only a hyphen
   but no descriptive text: `// MARK: -`.  
   [Ruotger Deecke](https://github.com/roddi)
--- a/Source/SwiftLintFramework/Rules/ColonRule.swift
+++ b/Source/SwiftLintFramework/Rules/ColonRule.swift
@@ -101,17 +101,17 @@ public struct ColonRule: CorrectableRule, ConfigurationProviderRule {
     }
 
     public func correctFile(file: File) -> [Correction] {
-        let matches = violationRangesInFile(file, withPattern: pattern)
+        let violations = violationRangesInFile(file, withPattern: pattern)
+        let matches = file.ruleEnabledViolatingRanges(violations, forRule: self)
         guard !matches.isEmpty else { return [] }
-
         let regularExpression = regex(pattern)
         let description = self.dynamicType.description
         var corrections = [Correction]()
         var contents = file.contents
         for range in matches.reverse() {
+            let location = Location(file: file, characterOffset: range.location)
             contents = regularExpression.stringByReplacingMatchesInString(contents,
                 options: [], range: range, withTemplate: "$1: $2")
-            let location = Location(file: file, characterOffset: range.location)
             corrections.append(Correction(ruleDescription: description, location: location))
         }
         file.write(contents)
--- a/Source/SwiftLintFramework/Rules/CommaRule.swift
+++ b/Source/SwiftLintFramework/Rules/CommaRule.swift
@@ -50,19 +50,19 @@ public struct CommaRule: CorrectableRule, ConfigurationProviderRule {
     }
 
     public func correctFile(file: File) -> [Correction] {
-        let matches = violationRangesInFile(file)
+        let violations = violationRangesInFile(file)
+        let matches = file.ruleEnabledViolatingRanges(violations, forRule: self)
         if matches.isEmpty { return [] }
 
-        var contents = file.contents as NSString
+        var nsstring = file.contents as NSString
         let description = self.dynamicType.description
         var corrections = [Correction]()
         for range in matches.reverse() {
-            contents = contents.stringByReplacingCharactersInRange(range, withString: ", ")
             let location = Location(file: file, characterOffset: range.location)
+            nsstring = nsstring.stringByReplacingCharactersInRange(range, withString: ", ")
             corrections.append(Correction(ruleDescription: description, location: location))
         }
-
-        file.write(contents as String)
+        file.write(nsstring as String)
         return corrections
     }
 
--- a/Source/SwiftLintFramework/Rules/LegacyCGGeometryFunctionsRule.swift
+++ b/Source/SwiftLintFramework/Rules/LegacyCGGeometryFunctionsRule.swift
@@ -105,7 +105,7 @@ public struct LegacyCGGeometryFunctionsRule: CorrectableRule, ConfigurationProvi
                 location: Location(file: file, characterOffset: $0.location))
         }
     }
-
+// swiftlint:disable function_body_length
     public func correctFile(file: File) -> [Correction] {
         let varName = RegexHelpers.varNameGroup
         let twoVars = RegexHelpers.twoVars
@@ -140,18 +140,20 @@ public struct LegacyCGGeometryFunctionsRule: CorrectableRule, ConfigurationProvi
 
         let matches = patterns.map({ pattern, template in
             file.matchPattern(pattern)
+                .filter { !file.ruleEnabledViolatingRanges([$0.0], forRule: self).isEmpty }
                 .filter { $0.1.first == .Identifier }
                 .map { ($0.0, pattern, template) }
         }).flatten().sort { $0.0.location > $1.0.location } // reversed
 
+        if matches.isEmpty { return [] }
+
         for (range, pattern, template) in matches {
+            let location = Location(file: file, characterOffset: range.location)
             contents = regex(pattern).stringByReplacingMatchesInString(contents, options: [],
                                                                        range: range,
                                                                        withTemplate: template)
-            let location = Location(file: file, characterOffset: range.location)
             corrections.append(Correction(ruleDescription: description, location: location))
         }
-
         file.write(contents)
         return corrections
     }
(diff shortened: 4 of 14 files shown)


Source/SwiftLintFramework/Rules/ColonRule.swift
```diff
@@ -103,16 +103,22 @@ public struct ColonRule: CorrectableRule, ConfigurationProviderRule {
...
+        let fileregions = file.regions()
 
         let regularExpression = regex(pattern)
         let description = self.dynamicType.description
         var corrections = [Correction]()
         var contents = file.contents
         for range in matches.reverse() {
+            let location = Location(file: file, characterOffset: range.location)
+            let region = fileregions.filter {$0.contains(location)}.first
+            if region?.isRuleDisabled(self) == true {
```
jpsim: Could be:

```swift
if region?.isRuleDisabled(self) == false {
    contents = regularExpression.stringByReplacingMatchesInString(contents,
        options: [], range: range, withTemplate: "$1: $2")
    corrections.append(Correction(ruleDescription: description, location: location))
}
```

Source/SwiftLintFramework/Rules/ColonRule.swift
```diff
@@ -103,16 +103,22 @@ public struct ColonRule: CorrectableRule, ConfigurationProviderRule {
...
         guard !matches.isEmpty else { return [] }
+        let fileregions = file.regions()
 
         let regularExpression = regex(pattern)
         let description = self.dynamicType.description
         var corrections = [Correction]()
         var contents = file.contents
         for range in matches.reverse() {
+            let location = Location(file: file, characterOffset: range.location)
+            let region = fileregions.filter {$0.contains(location)}.first
```
jpsim: space between `{}`: `{ $0.contains(location) }`

Source/SwiftLintFramework/Rules/PrivateOutletRule.swift
```diff
@@ -0,0 +1,68 @@
...
+//  SwiftLint
+//
+//  Created by Olivier Halligon on 12/08/2016.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+public struct PrivateOutletRule: ASTRule, OptInRule, ConfigurationProviderRule {
```
jpsim: I don't think you meant to add this file.

jpsim:
Rather than duplicate this code in a bunch of different rule implementations, this should be made generic.

masters3d:
Yeah that was my initial implementation but it wasn't granular . 
https://github.com/realm/SwiftLint/pull/724#issuecomment-232510246

Do you think we should go that route? it would be nice to have a default implementation that the correctable protocol would have.

jpsim:
> Do you think we should go that route? it would be nice to have a default implementation that the correctable protocol would have.

Yes, we should reuse the code for this as much as possible.

Source/SwiftLintFramework/Rules/ColonRule.swift
masters3d: `region?.isRuleDisabled(self)` returns false when `region` is nil

```
  if region?.isRuleDisabled(self) == true {  // returns false when region is nil
                continue
            } else {...}
```

plus I was trying to stay away from double negatives ;)
https://twitter.com/NachoSoto/status/762358223122927616

jpsim: Good call. This then?

```swift
if let region = fileregions.filter {$0.contains(location)}.first
    where !region.isRuleDisabled(self) {
    contents = regularExpression.stringByReplacingMatchesInString(contents,
        options: [], range: range, withTemplate: "$1: $2")
    corrections.append(Correction(ruleDescription: description, location: location))
}
```

masters3d: tests fail. when region is nil it doesn't fire but we need it to fire sometimes when there are no regions.

jpsim: Alright, keep it as-is then

jpsim:
This needs a rebase.

masters3d:
This is ready for now. I'll refactor all the correctable rules to pull this logic into the protocol in another PR.

jpsim:
@masters3d I don't see how it's important to land this just for a subset of rules, only to do it more generically later. Can we just do it right the first time?

masters3d:
I guess the important part is the added new test that makes sure new correctable rules do not omit disable regions checks. The rules I changed were the ones failing that test. I didn't touch other correctable rules that passed the test. 

I imagine if we provide a general facility for this checking that we would want to refactor all the correctable rules.
That's the part I think belongs to another PR since I can see it being quite big.

jpsim:
I'd rather do this be done in a single swoop, personally.

masters3d:
Sounds good. Currently all `Rule`s check for disabled regions and correctable rules have similar code twice. I am thinking this logic should probably be moved up to the `Rule` protocol. 
The only tricky part is that the repeated code happens in a for-loop for most rules. 

```swift
        for (eachLastLine, eachSectionCount) in linesSections {
            // Skips violation for areas where the rule is disabled
            let region = file.regions().filter {
                $0.contains(Location(file: file.path, line: eachLastLine.index, character: 0))
            }.first
            if region?.isRuleDisabled(self) == true {
                continue
            }
....

```

```swift
      for currentLine in file.lines {
            // Doesnt correct lines where rule is disabled
            let region = fileRegions.filter {
                $0.contains(Location(file: file.path, line: currentLine.index, character: 0))
            }.first
            if region?.isRuleDisabled(self) == true {
                correctedLines.append(currentLine.content)
                continue
            }
....

```

masters3d:
This should be ready now. I was complicating the issue. I found a method on File that does what I needed.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 
 * Adds `allow_private_set` configuration for the `private_outlet` rule.  
   [Rohan Dhaimade](https://github.com/HaloZero)
+  
+* Fix locally disabled comma rule autocorrect.
```
jpsim: I don't think this message is accurate anymore?

Source/SwiftLintFramework/Rules/CommaRule.swift
```diff
@@ -50,18 +50,18 @@ public struct CommaRule: CorrectableRule, ConfigurationProviderRule {
     }
 
     public func correctFile(file: File) -> [Correction] {
-        let matches = violationRangesInFile(file)
+        let violations = violationRangesInFile(file)
+        let matches = file.ruleEnabledViolatingRanges(violations, forRule: self)
         if matches.isEmpty { return [] }
 
         var contents = file.contents as NSString
```
jpsim: since you're in here, is there a reason the `as NSString` cast, and the `as String` one below, are necessary? Other files don't follow this pattern (e.g. `ColonRule.swift`).

masters3d: yeah, that needs to be NSString because `stringByReplacingCharactersInRange` doesn't show up on String.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 
 * Adds `allow_private_set` configuration for the `private_outlet` rule.  
   [Rohan Dhaimade](https://github.com/HaloZero)
+  
+* Added test to correctable rules to check `swiftlint-disable` before auto-correcting file.
```
jpsim: the command is `swiftlint:disable`

jpsim: Also, I'd call this a bug fix and the fact that you added tests is an implementation detail. The entry should read something like this IMO:

> Correctable rules no longer apply corrections if the rule is locally disabled.

masters3d: sounds good. I think I've now mastered the git command line :)

Source/SwiftLintFramework/Rules/LegacyCGGeometryFunctionsRule.swift
```diff
@@ -140,18 +140,20 @@ public struct LegacyCGGeometryFunctionsRule: CorrectableRule, ConfigurationProvi
 
         let matches = patterns.map({ pattern, template in
             file.matchPattern(pattern)
+                .filter { !file.ruleEnabledViolatingRanges([$0.0], forRule: self).isEmpty }
                 .filter { $0.1.first == .Identifier }
                 .map { ($0.0, pattern, template) }
         }).flatten().sort { $0.0.location > $1.0.location } // reversed
 
+        if matches.isEmpty { return []}
```
jpsim: Closure spacing 😉  https://github.com/realm/SwiftLint/issues/770

Tests/SwiftLintFramework/TestHelpers.swift
```diff
@@ -151,6 +151,14 @@ extension XCTestCase {
         ruleDescription.corrections.forEach(config.assertCorrection)
         // make sure strings that don't trigger aren't corrected
         zip(nonTriggers, nonTriggers).forEach(config.assertCorrection)
+
+        //"disable" command do not correct
+        ruleDescription.corrections.forEach { ( before, _) in
```
jpsim: This is more of a personal style preference, but I like to omit the parentheses in these cases: `{ before, _ in`

jpsim:
Thanks for your continued work on this @masters3d! Nearly there.

norio-nomura:
Looks good to me.
Are there any required changes that block this merging?

masters3d:
none that I am aware

norio-nomura:
@masters3d Thanks for doing this! 🙏

jpsim:
🎉

Issue #749, opened 2016-08-11
lint MARK comment formatting


brennantaylor:
For example enforcing and autocorrecting.

```swift
//MARK:bad
// MARK: good
```

jpsim:
good idea

freak4pc:
I wouldn't mind tackling this if I have some spare time tomorrow.

krodak:
Just added a relevant PR for this

jpsim:
Done in #768.

brennantaylor:
THANKS!

Igor-Palaguta:
Should be same rule applied for TODO comment?

Not clear why for MARK it is required. Xcode works properly even if mark without leading space

PR #767, opened 2016-08-22
[New Rule] MARK formatting rule

krodak:
Added a new rule as requested in following issue:

https://github.com/realm/SwiftLint/issues/749


jpsim:
Thanks @krodak, continuing in #768.

PR #768, opened 2016-08-22
Mark rule

jpsim:
Continued from #767. /cc @krodak

### DIFF
--- a/Source/SwiftLintFramework/Rules/MarkRule.swift
+++ b/Source/SwiftLintFramework/Rules/MarkRule.swift
@@ -1,0 +1,47 @@
+//
+//  MarkRule.swift
+//  SwiftLint
+//
+//  Created by Krzysztof Rodak on 22/08/16.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import SourceKittenFramework
+
+public struct MarkRule: ConfigurationProviderRule {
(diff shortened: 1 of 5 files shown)


Source/SwiftLintFramework/Rules/MarkRule.swift
```diff
@@ -35,10 +35,7 @@ public struct MarkRule: ConfigurationProviderRule {
         let options = ["MARK:[^ ]", "[^ ]MARK: [^-]", "\\sMARK:[^ ]", "MARK:[ ][-][^ ]"]
         let pattern = "(" + options.joinWithSeparator("|") + ")"
 
-        return file.matchPattern(pattern).flatMap { range, syntaxKinds in
```
krodak: @jpsim I based this on `TodoRule.swift` - `isCommentLike` seems like a 'wider' check than just `.Comment`.
I understand that `withSyntaxKinds` is better approach, but maybe other values from `isCommentLike` should also be included?

jpsim: `isCommentLike` _is_ wider, because it also includes doc comments, but `// MARK` comments can't be doc comments, they must be double-slashed. Whereas `TODO` comments could be left in documentation comments.

PR #777, opened 2016-08-25
Allow private(set) on IBOutlets

HaloZero:
Added a configuration to the ```private_outlet``` rule to allow for ```private(set)```. 

```private(set)``` can be useful. Sometimes you can have components that can expose their views for information retrieval, modification of components (say alpha transparency), or maybe setting callbacks on view components for later reuse but shouldn't allow their full write modification. 

I made it a configuration to allow people to control it or not, by default false as the rule is more specifically geared toward private outlets.

### DIFF
--- a/Source/SwiftLintFramework/Rules/PrivateOutletRule.swift
+++ b/Source/SwiftLintFramework/Rules/PrivateOutletRule.swift
@@ -10,7 +10,7 @@ import Foundation
 import SourceKittenFramework
 
 public struct PrivateOutletRule: ASTRule, OptInRule, ConfigurationProviderRule {
-    public var configuration = SeverityConfiguration(.Warning)
+    public var configuration = PrivateOutletRuleConfiguration(allowPrivateSet: false)
 
     public init() {}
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by jpsim** [comment]
This is fantastic! Stellar work @HaloZero 👏 . Tests pass locally for me, so I'm happy to get this in.

File: Source/SwiftLintFramework/Rules/PrivateOutletRule.swift [hunk 1]
```diff
@@ -47,8 +47,13 @@ public struct PrivateOutletRule: ASTRule, OptInRule, ConfigurationProviderRule {
 
         // Check if private
         let accessibility = (dictionary["key.accessibility"] as? String) ?? ""
+        let setterAccesiblity = (dictionary["key.setter_accessibility"] as? String) ?? ""
```
dmoreh: typo: `setterAccessibility`

jpsim: thanks, fixed in 1ed079e

keith:
Should this be configurable at all? It seems like `private(set)` is a sensible default for everyone since it solves the possible problem of allowing external callers to overwrite outlets, and it isn't something you would add accidentally since it isn't a default.

PR #787, opened 2016-08-30
Fix #786: Private unit test rule not scoped to tests

### LINKED ISSUE #786: Private Unit Test Rule: Rule is being evaluated on every function, not just tests
Related to #761:

This is embarrassing... This new rule has a bug where it tries to validate instance methods even if they aren't part of an `XCTestCase` subclass. Unlike #782 I think this is unquestionably *not* what people would expect this rule to do so it should be fixed ASAP. I have a working fix but won't be able to create a pull request until I get home tonight (PST). Here's a brain dump for when I get back to this:

I discovered this when trying to enable this rule for my custom subclass:

```YAML
private_unit_tst: #typo
  severity: error
  regex: "(XCTestCase|MyUnitTestCase)"
```

The regex in that rule is supposed to limit the scope of this rule to classes that inherit from `XCTestCase` or `MyUnitTestCase`. However, I had a typo in the config, so this rule should never have caught any style violations in `MyUnitTestCase`, but it did! 

In fact, in its current form, this rule will detect **any** instance method that begins with the word "test" and is marked `private` as a violation. It just so happens that this is uncommon enough that I never ran into it because methods beginning with "test" outside of unit tests are pretty rare.

Knowing this, the bug is pretty easy to spot in the source code for this rule: The `validateClass()` method checks to see if this class is a test class before checking the visibility modifier, and bails out early if it's not a test class. But `validateClass()` has no way of signaling its caller (`validateFile()`) that it bailed out early because this is not an applicable subclass, so `validateFile()` just continues on looking at the substructure. 

The fix is simple, move the logic for matching the inherited types with the regex to a `isTestClass()` helper function and call that explictely in `validateFile()`. At this point the `validateClass()` implementation becomes trivial so it can be moved into `validateFile()`, just after the `isTestClass()` check.

## Side Note

I noticed that `severity: error` in the YAML example above had no effect unless I also included the `regex` key. I suspect I'm failing when reading that field in the config file and that's probably happening before the `severity` key is read.

Edit: Yep. Missing regex causes `applyConfiguration()` to bail out at the beginning.

This happens before any other parts of the config are applied.
```Swift
guard let configurationDict = configuration as? [String: AnyObject],
    regexString = configurationDict["regex"] as? String else {
        throw ConfigurationError.UnknownConfiguration
}
```

Easy to fix, will include that too.

cfilipov:
See issue #786 for all the details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,8 @@
   [J. Cheyo Jimenez](https://github.com/masters3d)
   [#772](https://github.com/realm/SwiftLint/issues/772)
 
-* Fixed: Private unit test rule not scoped to test classes
+* Fixed: Private unit test rule not scoped to test classes.  
+  Fixed: Private unit test rule config is ignored if regex is missing.  
   [Cristian Filipov](https://github.com/cfilipov)
   [#786](https://github.com/realm/SwiftLint/issues/786)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@
   [J. Cheyo Jimenez](https://github.com/masters3d)
   [#772](https://github.com/realm/SwiftLint/issues/772)
 
+* Fixed: Private unit test rule not scoped to test classes
+  [Cristian Filipov](https://github.com/cfilipov)
+  [#786](https://github.com/realm/SwiftLint/issues/786)
+
 ## 0.12.0: Vertical Laundry
 
 ##### Breaking
--- a/Source/SwiftLintFramework/Rules/PrivateUnitTestRule.swift
+++ b/Source/SwiftLintFramework/Rules/PrivateUnitTestRule.swift
@@ -103,7 +103,19 @@ public struct PrivateUnitTestRule: ASTRule, ConfigurationProviderRule {
         -> [StyleViolation] {
 
             guard kind == .Class else { return [] }
-            let classViolations = validateClass(file, kind: kind, dictionary: dictionary)
+            guard isTestClass(dictionary) == true else { return [] }
+
+            /* It's not strictly necessary to check for `private` on classes because a
+             private class will result in `private` on all its members in the AST.
+             However, it's still useful to check the class explicitly because this
+             gives us a more clear error message. If we check only methods, the line
+             number of the error will be that of the function, which may not
+             necessarily be marked `private` but inherited it from the class access
+             modifier. By checking the class we ensure the line nuber we report for
+             the violation will match the line that must be edited.
+             */
+
+            let classViolations = validateAccessControlLevel(file, dictionary: dictionary)
             guard classViolations.isEmpty else { return classViolations }
 
             let substructure = dictionary["key.substructure"] as? [SourceKitRepresentable] ?? []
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,10 @@
   [J. Cheyo Jimenez](https://github.com/masters3d)
   [#772](https://github.com/realm/SwiftLint/issues/772)
 
+* Fixed: Private unit test rule not scoped to test classes
```
jpsim: missing trailing period and two trailing spaces as defined in [CONTRIBUTING.md](https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes).

Source/SwiftLintFramework/Rules/PrivateUnitTestRule.swift
```diff
@@ -103,7 +114,19 @@ public struct PrivateUnitTestRule: ASTRule, ConfigurationProviderRule {
         -> [StyleViolation] {
 
             guard kind == .Class else { return [] }
-            let classViolations = validateClass(file, kind: kind, dictionary: dictionary)
+            guard isTestClass(dictionary) == true else { return [] }
```
jpsim: could you remove the ` == true` please?

jpsim: actually folding the `kind == .Class` check into `isTestClass(_:)` would be nice.

Source/SwiftLintFramework/Rules/PrivateUnitTestRule.swift
```diff
@@ -119,30 +142,14 @@ public struct PrivateUnitTestRule: ASTRule, ConfigurationProviderRule {
...
-            guard !pathMatch.isEmpty else { return [] }
-            return validateAccessControlLevel(file, dictionary: dictionary)
-
+    private func isTestClass(dictionary: [String: SourceKitRepresentable]) -> Bool {
+        guard let superclass = superclass(dictionary) else { return false }
+        let pathMatch = configuration.regex.matchesInString(
+            superclass,
+            options: [],
+            range: NSRange(location: 0, length: (superclass as NSString).length))
+        guard !pathMatch.isEmpty else { return false }
```
jpsim: couldn't these two lines just be `return !pathMatch.isEmpty`?

Source/SwiftLintFramework/Rules/RuleConfigurations/PrivateUnitTestConfiguration.swift
```diff
@@ -36,13 +36,12 @@ public struct PrivateUnitTestConfiguration: RuleConfiguration, Equatable {
     }
 
     public mutating func applyConfiguration(configuration: AnyObject) throws {
-        guard let configurationDict = configuration as? [String: AnyObject],
-            regexString = configurationDict["regex"] as? String else {
-                throw ConfigurationError.UnknownConfiguration
+        guard let configurationDict = configuration as? [String: AnyObject] else {
```
jpsim: why are the changes in this file necessary?

jpsim:
Thanks for this @cfilipov, the changes mostly look good. Ping me once you've had a chance to address my feedback.

Source/SwiftLintFramework/Rules/RuleConfigurations/PrivateUnitTestConfiguration.swift
jpsim: Never mind, I just saw your side note in #786. Could you please add a bug fix entry in the changelog to reflect this fix? 😄

Source/SwiftLintFramework/Rules/PrivateUnitTestRule.swift
cfilipov: > actually folding the kind == .Class check into isTestClass(_:) would be nice.

This would mean either 1) passing kind to `isTestClass()` or 2) pulling kind out of the dictionary in `isTestClass()` which was already done for us.

Technically `isTestClass()` is already checking `kind == .Class` indirectly by calling `superclass` which does the check (by reading it from the dictionary. I figured I was just being extra-explicit by making the check in `validateFile`, if anything, perhaps this check should just be removed completely? I do personally like the extra check, even if it's unnecessary, but I can go either way.

jpsim: ok, how about doing this:

```swift
guard kind == .Class && isTestClass(dictionary) else {
    return []
}
```

Source/SwiftLintFramework/Rules/RuleConfigurations/PrivateUnitTestConfiguration.swift
cfilipov: It's [in there](https://github.com/cfilipov/SwiftLint/blob/5fd99cd1d77f0ef1f7df1481eb675d6a13ae3f84/CHANGELOG.md). Do you mean calling out the fix for the configuration part as well?

jpsim: yes, the configuration part

Source/SwiftLintFramework/Rules/PrivateUnitTestRule.swift
cfilipov: Sounds good!

Source/SwiftLintFramework/Rules/PrivateUnitTestRule.swift
cfilipov: Indeed. Will update this.

cfilipov:
> Ping me once you've had a chance to address my feedback.

Thanks for the feedback. Sorry for introducing this nasty bug. I'm at work right now so I don't have access to my github SSH key. I can make most of the changes using the web interface, which isn't great, but it works. Unfortunately, it looks like I now have a conflict that must be resolved and I don't think I can fix that through the web UI so that will need to wait until I get home tonight.

jpsim:
This can wait a few hours, no need to apologize!

cfilipov:
@jpsim just pushed an update that should address all the items you brought up.

jpsim:
Many thanks @cfilipov!

Issue #58, opened 2015-05-28
Validate explicit ACL modifiers for top-level declarations


jpsim:
Top-level declarations should use explicit ACL keywords (`public`, `internal`, `private`).

keith:
It looks like there's no difference to sourcekit whether or not internal is specified. Using sourcekitten's structure the only difference between:

```swift
class Foo {}
```

And:

```swift
internal class Foo {}
```

Is the offset. This would have to be handled by examining the file as well.

segiddins:
@keith examining the contents of the declaration would probably suffice for this?

keith:
It seems like the class declaration key's offset starts after the modifiers. So in the second example from above the offset is 9. This could cause issue for cases like:

```swift
internal final class Foo {}
```

Or something crazier like:

```swift
internal
class Foo {}
```

jpsim:
Here's a rough algorithm to do this:

1. find all top-level `internal` type declarations: see https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules/TypeNameRule.swift.
2. find all `internal` modifiers: `File.matchPattern("internal", withSyntaxKinds: [.AttributeBuiltin])`.
3. throw style violations for declarations found in step 1 with no associated `internal` modifier from step 2.

jdmarshall90:
Could this be enhanced to include *all* declarations, and not just top-level ones?

PR #808, opened 2016-09-07
[WIP] Missing Access Control rule

KaiCode2:
Implemented a new rule to ensure access control is explicitly set for all classes, structs and enums.
### Notes
- Tests are currently not passing because not all types within the project have adopted this rule.
- The current implementation is to apply a regex to all lines in project and check for hits. Ideally we would use sourcekitten for the implementation however sourcekit does not output wether the type has been explicitly typed (as default is internal).
- The regex is not iron clad and there are probably edge cases that would set it off incorrectly… So please give review the regex expression as my regex knowledge is pretty minimal.
### TODOs
- [ ] update CHANGELOG.md file
- [ ] incorporate this change across the entire project to make tests pass
- [ ] `git squash`


jpsim:
Thanks for the PR, @KaiCode2! Sorry for the delay in reviewing this, I'm massively behind on SwiftLint notifications 😅 

Have you considered using SourceKitten's parse structure to detect classes, structs and enums rather than attempting to detect them via regex?

For example, we leverage the parse structure to apply certain rules only to class/struct/enum declarations for `TypeNameRule`: https://github.com/realm/SwiftLint/blob/0.13.0/Source/SwiftLintFramework/Rules/TypeNameRule.swift#L48-L53

This leads to a much more reliable rule since you don't accidentally capture say sample code in comments, since we use the Swift compiler's very accurate parser. Trying to parse Swift by using regular expressions is guaranteed to lead to false positives, and generally more complex less maintainable code.

Given this, the missing "TODO" items from your initial comment, and how outdated this PR is, I'll be closing this for now.

However, I hope you'll be able to take this feedback under consideration and submit a PR again soon!

KaiCode2:
@jpsim I initially tried to use sourcekitten to detect access control, however because sourcekit infers unset access control to internal and doesn't emit wether it was explicitly set, I am unable to hook into sourcekit (and thus sourcekitten) for this PR. 

A potential optimization would be to check type declaration line numbers and explicitly apply the regex to those lines (drastically cutting down on the number of lines where this needs to be applied).

jpsim:
> A potential optimization would be to check type declaration line numbers and explicitly apply the regex to those lines (drastically cutting down on the number of lines where this needs to be applied).

Taking this hybrid approach isn't _just_ an optimization, it also means that you leverage the parse structure that _is_ available, leading to more accurate results.

I'm certainly interested in seeing this pursued, which is why we're still tracking this as an open issue: #58.

KaiCode2:
Fantastic, I'll update to make the Regex work with the parse structure. 

Will open a second PR when finished.

Thanks

PR #819, opened 2016-09-20
[wip/feature proposal] Add version specification support with .swiftlint.yml

hiragram:
- [ ] proposal
- [ ] `~>` support 
- [ ] fix lint warning
- [ ] test

**This PR is a feature proposal and work in progress.**

I want to write minimum required version of SwiftLint in `.swiftlint.yml`.
When lint is executed with old binary, it may not support some rules added newer version and skip checking.
This may cause to miss some violation or warning, and I don't like it.

When you write 

```
required_version: 0.13.0
```

in your .swiftlint.yml and run lint on 0.12.0, you see

```
configuration error: Your version 0.12.0 is older than required version 0.13.0
```

I am very new to contribution on SwiftLint so if you have any ideas, please tell me.

### DIFF
--- a/Source/SwiftLintFramework/Models/Version.swift
+++ b/Source/SwiftLintFramework/Models/Version.swift
@@ -25,6 +25,10 @@ public struct Version: CustomStringConvertible, Comparable {
     let minor: Int
     let patch: Int
 
+    private var components: [Int] {
+        return [major, minor, patch]
+    }
+
     init(major: Int, minor: Int = 0, patch: Int = 0) {
         self.major = major
         self.minor = minor
@@ -50,11 +54,11 @@ public struct Version: CustomStringConvertible, Comparable {
         if numbers.count == 3 {
             self.init(major: numbers[0], minor: numbers[1], patch: numbers[2])
         } else if numbers.count == 2 {
-            self.init(major: numbers[0], minor: numbers[1], patch: 0)
+            self.init(major: numbers[0], minor: numbers[1])
         } else if numbers.count == 1 {
-            self.init(major: numbers[0], minor: 0, patch: 0)
+            self.init(major: numbers[0])
         } else {
-            fatalError()
+            return nil
         }
     }
 
@@ -64,16 +68,9 @@ public struct Version: CustomStringConvertible, Comparable {
 }
 
 public func == (left: Version, right: Version) -> Bool {
-    return
-        left.major == right.major &&
-        left.minor == right.minor &&
-        left.patch == right.patch
+    return left.components == right.components
 }
 
 public func < (left: Version, right: Version) -> Bool {
-    if left == right {
-        return false
-    } else {
-        return !(left > right)
-    }
+    return left.components.lexicographicalCompare(right.components)
 }


Source/SwiftLintFramework/Models/Version.swift
```diff
@@ -0,0 +1,88 @@
...
+
+    init(major: Int, minor: Int, patch: Int) {
+        self.major = major
+        self.minor = minor
+        self.patch = patch
+    }
+
+    init(versionString: String) {
+        let numbers = versionString.componentsSeparatedByString(".").map {
+            return Int($0)!
```
scottrhoyt: You can crash hard here. Consider failing the initializer or throwing an error when not parsable instead.

Source/SwiftLintFramework/Models/Version.swift
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        if numbers.count == 3 {
+            self.init(major: numbers[0], minor: numbers[1], patch: numbers[2])
+        } else if numbers.count == 2 {
+            self.init(major: numbers[0], minor: numbers[1], patch: 0)
+        } else if numbers.count == 1 {
+            self.init(major: numbers[0], minor: 0, patch: 0)
+        } else {
+            fatalError()
```
scottrhoyt: Same for `fatalError()`

Source/SwiftLintFramework/Models/Version.swift
```diff
@@ -0,0 +1,88 @@
...
+        self.patch = patch
+    }
+
+    init(versionString: String) {
+        let numbers = versionString.componentsSeparatedByString(".").map {
+            return Int($0)!
+        }
+
+        if numbers.isEmpty {
+            fatalError()
```
scottrhoyt: `fatalError()` is a code smell. Consider failing the initializer or throwing and error instead.

Source/SwiftLintFramework/Models/Version.swift
```diff
@@ -0,0 +1,88 @@
...
+//  Version.swift
+//  SwiftLint
+//
+//  Created by HirayamaYuya on 2016/09/16.
+//  Copyright © 2016年 Realm. All rights reserved.
+//
+
+import Foundation
+
+public struct Version: CustomStringConvertible {
```
scottrhoyt: Since you are implementing `Comparable`, I would conform to it.

Source/SwiftLintFramework/Models/Version.swift
```diff
@@ -0,0 +1,88 @@
...
+            fatalError()
+        }
+    }
+
+    public var description: String {
+        return "\(major).\(minor).\(patch)"
+    }
+}
+
+public func > (left: Version, right: Version) -> Bool {
```
scottrhoyt: `Comparable` just requires `==` and `<`

Source/SwiftLintFramework/Models/Version.swift
```diff
@@ -0,0 +1,88 @@
...
+
+public func < (left: Version, right: Version) -> Bool {
+    if left == right {
+        return false
+    } else {
+        return !(left > right)
+    }
+}
+
+public func != (left: Version, right: Version) -> Bool {
```
scottrhoyt: `Comparable` just requires `==` and `<`

Source/SwiftLintFramework/Models/Version.swift
```diff
@@ -0,0 +1,88 @@
...
+                    fatalError()
+            }
+
+            return Version(versionString: version)
+        }
+    }
+
+    var major: Int
+    var minor: Int
+    var patch: Int
```
scottrhoyt: These should probably be `let` constants.

Source/SwiftLintFramework/Models/Version.swift
```diff
@@ -0,0 +1,88 @@
...
+
+    init(major: Int, minor: Int, patch: Int) {
```
scottrhoyt: It's possible you might want to provide default values for `minor` and `patch` and maybe even `major`:

`init(major: Int, minor: Int = 0, patch: Int = 0)`

Source/SwiftLintFramework/Models/Version.swift
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        if numbers.isEmpty {
+            fatalError()
+        }
+
+        if numbers.count == 3 {
+            self.init(major: numbers[0], minor: numbers[1], patch: numbers[2])
+        } else if numbers.count == 2 {
+            self.init(major: numbers[0], minor: numbers[1], patch: 0)
```
scottrhoyt: If you use default values for `minor` and `patch` this can be shortened.

Source/SwiftLintFramework/Models/Version.swift
```diff
@@ -0,0 +1,88 @@
...
+        } else if numbers.count == 1 {
+            self.init(major: numbers[0], minor: 0, patch: 0)
```
scottrhoyt: If you use default values for `minor` and `patch` this can be shortened.

Source/SwiftLintFramework/Models/Version.swift
```diff
@@ -0,0 +1,88 @@
...
+
+public func == (left: Version, right: Version) -> Bool {
+    return
+        left.major == right.major &&
+        left.minor == right.minor &&
+        left.patch == right.patch
+}
+
+public func < (left: Version, right: Version) -> Bool {
+    if left == right {
```
scottrhoyt:
```swift
return (left.major < right.major) ||
       ((left.major == right.major) && (left.minor < right.minor)) ||
       ((left.major == right.major) && (left.minor == right.minor) && (left.patch < right.patch))
```

Source/SwiftLintFramework/Models/Version.swift
```diff
@@ -0,0 +1,88 @@
...
+        } else {
+            return left.minor > right.minor
+        }
+    } else {
+        return left.major > right.major
+    }
+}
+
+public func == (left: Version, right: Version) -> Bool {
+    return
```
scottrhoyt: if you add `components`:

```swift
return left.components == right.components
```

Source/SwiftLintFramework/Models/Version.swift
```diff
@@ -0,0 +1,88 @@
...
+
+import Foundation
+
+public struct Version: CustomStringConvertible {
+
+    public static var current: Version {
+        get {
+            guard let version = NSBundle(identifier: "io.realm.SwiftLintFramework")?
+                .objectForInfoDictionaryKey("CFBundleShortVersionString") as? String else {
+                    fatalError()
```
scottrhoyt: Don't `fatalError()` here, consider making this return a `Version?` instead.

Source/swiftlint/Commands/LintCommand.swift
```diff
@@ -36,6 +36,15 @@ struct LintCommand: CommandType {
         let reporter = reporterFromString(
             options.reporter.isEmpty ? configuration.reporter : options.reporter
         )
+
+        // Validate that current version is newer than or equals to required version
+        if let required = configuration.requiredVersion {
+            if Version.current < required {
+                return .Failure(.UsageError(description: "configuration error: Your version \(Version.current) is" +
```
scottrhoyt: I'm not sure if this should be a failure or a warning while proceeding to lint. Thoughts @realm/swiftlint ? I wouldn't want to necessarily break people's CI builds if someone updates the swiftlint version or somehow the CI provider's swiftlint version changes, but all the linting still passes.

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -21,6 +21,7 @@ private enum ConfigurationKey: String {
     case UseNestedConfigs = "use_nested_configs" // deprecated
     case WhitelistRules = "whitelist_rules"
     case WarningThreshold = "warning_threshold"
+    case RequiredVersion = "required_version"
```
scottrhoyt: "required_version" makes it sound like the versions need to match exactly. The check later on is a `>`. This should probably be renamed "minimum_version", "swiftlint_minimum_version", or "swiftlint_minimum" with the current implementation. If you add support for the `~>` operator, then maybe just "swiftlint_version". Thoughts @realm/swiftlint ?

hiragram:
@scottrhoyt Thank you for your review!
I was just about to see Carthage's version implementation as reference.
And I absolutely agree with you and your other comments 🙆

hiragram:
After some progress, please let me request you to review again 🙇

scottrhoyt:
No problem! Thanks for the PR. Somehow at least one of my review comments got lost. I was saying that if you implemented a `var components: [Int] = { return [major, minor, patch] }` in `Version` it would clean up the code needed to implement `Comparable` later on.

Source/SwiftLintFramework/Models/Configuration.swift
hiragram: I would like to support `~>` operator.
"swiftlint_version" sounds good 👍

marcelofabri:
Closing in favor of #1134

PR #829, opened 2016-10-07
Added ExplicitSelfRule

IanKeen:
Requires the explicit use of `self.` when accessing instance members
Issue: #321


IanKeen:
Not sure what to do with those failing tests? - looks like its swiftlint integration tests?

scottrhoyt:
I’m away from a computer right now, so I can’t be sure. But usually that means that the swiftlint code is generating errors or warnings while being linted by swiftlint itself. Try linting your local copy to see if this is the case.

IanKeen:
Ahh you might be right - that trailing whitespace always gets me!

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
```diff
@@ -0,0 +1,132 @@
...
+        let memberTypes: [SwiftDeclarationKind] = [
+            .FunctionMethodInstance,
+            .VarInstance,
+        ]
+        let instanceMembers = file.members(memberTypes)
+        let instanceMemberNames = instanceMembers.flatMap { $0["key.name"] as? String }
+
+        //get the members body (functions and calculated properties will have this data)
+        guard
+            let bodyLocation = (dictionary["key.bodyoffset"] as? Int64).flatMap({ Int($0) }),
```
NachoSoto: You probably want `as! Int64?` so that this doesn't ignore wrong types.

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
```diff
@@ -0,0 +1,132 @@
...
+            let bodyLength = (dictionary["key.bodylength"] as? Int64).flatMap({ Int($0) })
```
NachoSoto: Can you use `flatMap(Int.init)` here?

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
```diff
@@ -0,0 +1,132 @@
...
+    public static let description = RuleDescription(
+        identifier: "explicitSelf",
+        name: "Explict Self",
+        description: "Require explicit self for instance members",
+        nonTriggeringExamples: [
+            "class Good1 { let value: Int = 42; var property: Int { return self.value } }",
+            "class Good2 { let value: Int = 42; func function() -> Int { return self.value } }",
+            "class Good3 { let value: Int; init() { self.value = 42 } }",
+        ],
+        triggeringExamples: [
```
NachoSoto: Awesome :+1:

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
```diff
@@ -0,0 +1,132 @@
...
+    public var configuration = SeverityConfiguration(.Error)
+
+    public init() { }
+
+    public static let description = RuleDescription(
+        identifier: "explicitSelf",
+        name: "Explict Self",
+        description: "Require explicit self for instance members",
+        nonTriggeringExamples: [
+            "class Good1 { let value: Int = 42; var property: Int { return self.value } }",
```
NachoSoto: I would add a few more cases here to verify that global variables, members of other types, etc. don't trigger violations.

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
```diff
@@ -0,0 +1,132 @@
...
+            "class Good2 { let value: Int = 42; func function() -> Int { return self.value } }",
+            "class Good3 { let value: Int; init() { self.value = 42 } }",
+        ],
+        triggeringExamples: [
+            "class Bad1 { let value: Int = 42; var property: Int { return value } }",
+            "class Bad2 { let value: Int = 42; func function() -> Int { return value } }",
+            "class Bad3 { let value: Int; init() { value = 42 } }",
```
NachoSoto: I can think of a few other cases:
- Subclassing: accessing a member from the parent class.
- `struct` and `enum` (because why not)
- `extension`: verify that accessing a member from a function in an extension triggers this too.

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
NachoSoto: And maybe same for accessing variables in a scope other than a type (like global values or local variables in a global function scope.
Also `static` functions! (Sorry, just thinking of more as I go hehe)

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
```diff
@@ -0,0 +1,132 @@
...
+        // swiftlint:disable control_statement
+        for (a, b) in zip(suffix, items) {
+            if a == b { return true }
+        }
+        return false
+    }
+
+    func suffix(matches items: [[Generator.Element]]) -> Bool {
+        for test in items
+            where self.suffix(matches: test) { return true }
```
NachoSoto: This can be simpler using `Array.contains`, I think simply `return items.contains(self.suffix)`

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
IanKeen: Keep them coming!! 👍

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
IanKeen: 🤔 `as?` seems to be the convention throughout - I think it would be preferential to a crash anyway

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
IanKeen: Nope, It cant disambiguate (they are all `.init(_ value: Foo)`

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
IanKeen: it calls the extension above, this one is just a convenience to perform multiple checks at once

keith:
Here's 2 interesting examples we might see that I'm not sure are covered here:

```swift
struct Foo {
    var a = 1
}

extension Foo {
    func bar() {
        print(a)
    }
}

Foo().bar()
```

and

```swift
struct Foo {
    var something = 1
}

extension Foo {
    func bar(something: Int) {
        print(something)
    }
}

Foo().bar(something: 1)
```

IanKeen:
ah nice @keith ! @NachoSoto pointed out the first in his comments but the second will be an interesting one 😄 

I'll dive into the suggestions this weekend at some point.

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
NachoSoto: No, it would be preferable to crash over a silent bug. Here you just want to know if the value is not present.

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
NachoSoto: I know. You can use the appropriate function as the parameter to `contas` then. My point is that you don't need the loop, this is a common construct: return `true` if any element meets the criterion, which is what `Array.contains` does.

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
IanKeen: I'm still inclined to leave it this way, every other rule is using code like this right now

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
IanKeen: perhaps I'm misunderstanding what you mean? this function takes an `Array<Array<Generator.Element>>` and neither this function or the one it calls is equivalent to `.contains` - they are both checking the tail of the array for a set of elements in a specific order. can you perhaps show me a snippet of what you mean?

IanKeen:
So I've significantly improved the test coverage thanks to your suggestions guys which has made the code must more robust 👌 

The covered cases are (all tested for class and struct (except subclasses)):

| Member 👇 / Accessed From 👉 	| Init 	| Function 	| Calculated  Property 	| Extension  Calculated Property 	| Extension  Function 	| Static  Property 	| Static  Function 	| Subclass  Property 	| Subclass  Function 	|
|------------------------------	|------	|----------	|----------------------	|--------------------------------	|---------------------	|------------------	|------------------	|--------------------	|--------------------	|
| Property                     	| ✅    	| ✅        	| ✅                    	| ✅                              	| ✅                   	| ✅                	| ✅                	| ✅                  	| ✅                  	|
| Function                     	| ✅    	| ✅        	| ✅                    	| ✅                              	| ✅                   	| ✅                	| ✅                	| ✅                  	| ✅                  	|

as well as some edge cases:
- Function parameter shadowing wont trigger this rule, eg:
```swift
class Edge2 { 
    func value() -> Int { return 42 }
    func function(value: Int) -> Int { return value /* refers to parameter */ } 
}
```
This is handled for both property/function shadowing

Thanks @NachoSoto  and @keith for the suggestions 👍

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
```diff
@@ -0,0 +1,284 @@
...
+        + "class Bad15_B: Bad15_A { var property: Int { return value } }",
+    "class Bad16_A { func value() -> Int { return 42 } }; "
+        + "class Bad16_B: Bad16_A { var property: Int { return value() } }",
+    "class Bad17_A { var value: Int = 42 }; "
+        + "class Bad17_B: Bad17_A { func function() -> Int { return value } }",
+    "class Bad18_A { func value() -> Int { return 42 } }; "
+        + "class Bad18_B: Bad18_A { func function() -> Int { return value() } }",
+]
+
+let nonTriggeringExamples = ["class", "struct"].flatMap(nonTriggering)
```
NachoSoto: :+1:

NachoSoto:
:shipit:

keith:
I just ran this on our codebase and found a few false positives. Here's one:

```swift
struct Foo {
    var bar: Int

    func baz() {
        let bar = 123
        _ = bar
    }
}
```

keith:
Here's another interesting one:

```swift
struct Foo {
    func bar() {
        self.foo.bar()
    }
}
```

keith:
And another:

```swift
struct Foo {
    func baz() {
        let foo = 1
        foo.something()
    }
}

struct Qux {
    func something() {}
}
```

keith:
I think some of these require more semantic understanding that what we have in SwiftLint. So maybe it isn't really feasible, but I can provide a few more examples if needed.

IanKeen:
@keith hmm I can investigate that first one, but the others don't appear to be valid code? I'm not sure it's within the scope of a rule to lint code that otherwise doesn't compile?

IanKeen:
@keith Updated to handle local scope shadowing
re: https://github.com/realm/SwiftLint/pull/829#issuecomment-252540405

keith:
Sorry, I didn't provide full enough examples here, but each of these cases was pulled from our real code and just made smaller for examples.

IanKeen:
Ah right, well if you can turn examples 2 & 3 above into full working snippets I'm happy to take a look 😄

norio-nomura:
I tried to enable this rule in `.swiftlint.yml` of SwiftLint repository and ran integration test.
screenshot 2016-10-12 22 28 31

I feel it generates too much false positives yet.

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
```diff
@@ -0,0 +1,286 @@
...
+
+public struct ExplicitSelfRule: ASTRule, OptInRule, ConfigurationProviderRule {
+    public var configuration = SeverityConfiguration(.Error)
+
+    public init() { }
+
+    public static let description = RuleDescription(
+        identifier: "explicitSelf",
+        name: "Explict Self",
+        description: "Require explicit self for instance members",
```
NachoSoto: The wording seems weird. How about "Explicit `self` for instance members and functions is required".

jpsim:
Given the number of false positives this triggers, I'm glad this is implemented as an opt-in rule.

@IanKeen would you mind rebasing this against `master` so we can make a last push to land this?

On a side note, this type of rule would benefit from being implemented in a new mode of SwiftLint that I've been considering: one that operates _after_ the language Parse phase, actually after the [Semantic Analysis](https://swift.org/compiler-stdlib/#compiler-architecture) phase, which would let SwiftLint know the difference between instance `bar` and local `bar` in @keith's comment above:

```swift
struct Foo {
    var bar: Int

    func baz() {
        let bar = 123
        _ = bar
    }
}
```

This type of mode would require that the module being analyzed 1) already be built by the compiler and 2) that SwiftLint would somehow know all the compiler flags needed to build the module. But if we can find a good user experience for this, it would enable much more powerful rules, from multi-file aware linting (e.g. #836) to static analysis.

This is much more straightforward for using with the Swift Package Manger since it generates a build manifest which includes all the compiler arguments (`.build/debug.yaml`), but it's doable with Xcode too.

The way [jazzy](https://github.com/realm/jazzy) does this is by straight up cleaning and building the Xcode project, parsing the `xcodebuild` log to extract arguments for a module, but that's very time consuming as the project has to be built from scratch to work reliably.

I hope I can sit down at one point and file an issue tracking exactly how building such a mode for SwiftLint could be built. Until then, I'm sharing these loose thoughts here to spur ideas among SwiftLint contributors and users.

IanKeen:
@jpsim 👍 rebased, also added a couple of final touches from some of the suggestions above

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 ##### Enhancements
 
+* Add `ExplicitSelfRule` Opt-In rule that enforces the use of `self.` for instance members
```
jpsim: 80 char hard-wrap, trailing period following by two spaces. Format described here: https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes

Also, we consistently refer to "opt-in" in lowercase throughout the changelog.

Source/SwiftLintFramework/Models/MasterRuleList.swift
```diff
@@ -87,5 +87,6 @@ public let masterRuleList = RuleList(rules:
     ValidIBInspectableRule.self,
     VariableNameRule.self,
     VerticalWhitespaceRule.self,
-    WeakDelegateRule.self
+    WeakDelegateRule.self,
+    ExplicitSelfRule.self
```
jpsim: this list is alphabetized.

IanKeen:
@jpsim weird, should be good now

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
```diff
@@ -0,0 +1,286 @@
...
+                )
+            )
+        }
+        return violations
+    }
+}
+
+extension SyntaxToken {
+    func name(file: File) -> String {
+        return file.contents.substring(self.offset, length: self.length)
```
jpsim: Even though that's the opposite of what this rule is enforcing, this project prefers omitting `self.` wherever possible. Could you please adapt your code to be consistent with this?

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
```diff
@@ -0,0 +1,286 @@
...
+            .substring(self.offset + self.length + 1)
+            .componentsSeparatedByString(")")[0]
+        guard !remainder.isEmpty else { return "\(value)()" }
+
+        let params = remainder
+            .sanitizedParameters()
+            .componentsSeparatedByString(",")
+
+        var signature = ""
+        for param in params {
```
jpsim: this could be refactored as `let signature = reduce("") { ... }`

Source/SwiftLintFramework/Rules/ExplictSelfRule.swift
```diff
@@ -0,0 +1,286 @@
...
+//  SwiftLint
+//
+//  Created by Ian Keen on 2016-10-06.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+func nonTriggering(type: String) -> [String] {
```
jpsim: all the implicitly `internal` members in this file that aren't needed in other files should be marked as `private` or `fileprivate`.

NachoSoto:
@jpsim I'll test this on Monday!

jpsim:
@NachoSoto friendly reminder to test this when you get a chance 👍

jpsim:
@IanKeen sorry for landing more changes recently and causing merge conflicts with this. Can you please rebase this so we can confirm it builds and passes tests on Linux?

Can anyone think of a popular OSS project that already implicitly adheres to this rule (in spirit at least) so that we can run this against a real codebase to make sure it works as intended, without too many false positives?

JohnEstropia:
> Can anyone think of a popular OSS project that already implicitly adheres to this rule
If you still need libraries to test this against, [CoreStore](https://github.com/JohnEstropia/CoreStore) observes this rule.

kimdv:
This would be nice if it would be added! Thumbs up here 👍

kimdv:
Any status on this @IanKeen ? 😄

Elland:
Can we expect this to be merged soon?

sammy-SC:
@IanKeen @jpsim 
Is it alright if I try to continue the work on this PR?

IanKeen:
ugh, _very sorry_ I've had 0 time to jump back on this... I would love it if you wanna take this one @sammy-SC ... work + personal duties has left me no time for anything fun 😢

sammy-SC:
@IanKeen if you don't mind, could you give me push access to this branch? Otherwise I would have to open new PR.

IanKeen:
@sammy-SC no idea how to do that, just added you as a collaborator on the repo 👍

sammy-SC:
@IanKeen Thanks, that's exactly what I had in mind. 

I tried running this rule against https://github.com/JohnEstropia/CoreStore, it produces lots of false negatives, I am looking into it.

keith:
FWIW I've ran it on our codebase of ~70K LOC where we require `self.` and it generated about 4k false positives. For more context, running `ag "\Wself\."` shows that we have 12k `self.` uses.

It seems like a rule that needs this much context should really be run outside of SwiftLint against the AST, since otherwise it will never not produce significant numbers of false positives. Just my 2 cents.

marcelofabri:
I'm closing this for now due to lack of activity and the high number of false positives. Feel free to open another pull request if the issues are fixed.

PR #837, opened 2016-10-11
Fix `SourceKitCrashTests` randomly failed

norio-nomura:
Sometime `ObjectIdentifier(self)` might collide between tests in `SourceKitCrashTests`.
The collision caused fail on `testAssertHandlerIsNotCalledOnNormalFile`.

Reproduced log with printing `cacheKey`:
```terminal.session
Test Suite 'SourceKitCrashTests' started at 2016-10-11 20:54:42.627
Test Case '-[SwiftLintFrameworkTests.SourceKitCrashTests testAssertHandlerIsCalledOnFileThatCrashedSourceKitService]' started.
set: cacheKey: 140575074796400
get: cacheKey: 140575074796400
get: cacheKey: 140575074796400
get: cacheKey: 140575074796400
Test Case '-[SwiftLintFrameworkTests.SourceKitCrashTests testAssertHandlerIsCalledOnFileThatCrashedSourceKitService]' passed (0.003 seconds).
Test Case '-[SwiftLintFrameworkTests.SourceKitCrashTests testAssertHandlerIsNotCalledOnNormalFile]' started.
set: cacheKey: 140575074796400
get: cacheKey: 140575074796400
/Users/norio/github/SwiftLint/Tests/SwiftLintFramework/SourceKitCrashTests.swift:24: error: -[SwiftLintFrameworkTests.SourceKitCrashTests testAssertHandlerIsNotCalledOnNormalFile] : XCTAssertFalse failed - Expects assert handler was not called on accessing File.structure
get: cacheKey: 140575074796400
/Users/norio/github/SwiftLint/Tests/SwiftLintFramework/SourceKitCrashTests.swift:29: error: -[SwiftLintFrameworkTests.SourceKitCrashTests testAssertHandlerIsNotCalledOnNormalFile] : XCTAssertFalse failed - Expects assert handler was not called on accessing File.syntaxMap
get: cacheKey: 140575074796400
/Users/norio/github/SwiftLint/Tests/SwiftLintFramework/SourceKitCrashTests.swift:34: error: -[SwiftLintFrameworkTests.SourceKitCrashTests testAssertHandlerIsNotCalledOnNormalFile] : XCTAssertFalse failed - Expects assert handler was not called on accessing File.syntaxKindsByLines
Test Case '-[SwiftLintFrameworkTests.SourceKitCrashTests testAssertHandlerIsNotCalledOnNormalFile]' failed (0.007 seconds).
Test Case '-[SwiftLintFrameworkTests.SourceKitCrashTests testRulesWithFileThatCrashedSourceKitService]' started.
set: cacheKey: /Users/norio/github/SwiftLint/Tests/SwiftLintFramework/SourceKitCrashTests.swift
Most of rules are skipped because sourcekitd fails.
Test Case '-[SwiftLintFrameworkTests.SourceKitCrashTests testRulesWithFileThatCrashedSourceKitService]' passed (0.002 seconds).
Test Suite 'SourceKitCrashTests' failed at 2016-10-11 20:54:42.646.
	 Executed 3 tests, with 3 failures (0 unexpected) in 0.012 (0.018) seconds
```

### DIFF
--- a/Source/SwiftLintFramework/Extensions/File+Cache.swift
+++ b/Source/SwiftLintFramework/Extensions/File+Cache.swift
@@ -68,7 +68,7 @@ private struct Cache<T> {
 extension File {
 
     private var cacheKey: String {
-        return path ?? "\(ObjectIdentifier(self).hashValue)"
+        return path ?? contents
     }
 
     internal var sourcekitdFailed: Bool {


Source/SwiftLintFramework/Extensions/File+Cache.swift
```diff
@@ -68,7 +68,7 @@ private struct Cache<T> {
 extension File {
 
     private var cacheKey: String {
-        return path ?? "\(ObjectIdentifier(self).hashValue)"
+        return path ?? "\((contents as NSString).hash)"
```
jpsim: why is it necessary to cast to `NSString`? `String.hash` is also available.

norio-nomura: Oh, my bad! Thanks.
I also think we should use `contents` itself instead of their `hash`.

jpsim: Yes! You're right. Though that will add to the memory overhead.

norio-nomura: On most of our usage, `path` will be used. I think the memory overhead can be ignored.

norio-nomura: > On most of our usage, path will be used.

I confirmed that `contents` is used on running tests and `swiftlint lint --use-stdin`.

norio-nomura:
updated.

Source/SwiftLintFramework/Extensions/File+Cache.swift
jpsim: yes, I think it's a far less common usage scenario.

norio-nomura:
Thanks! :pray:

Issue #544, opened 2016-02-16
function_parameter_count should ignore initializers


ivanbruel:
How would you work around models with a good amount of properties to be set on an initializer?
I agree with the number of parameters being limited, but not initializers themselves.

<img width="1316" alt="screen shot 2016-02-16 at 18 54 25" src="https://cloud.githubusercontent.com/assets/4968178/13087230/bed07cf4-d4de-11e5-8ea0-ddad6fdd35fb.png">

jpsim:
Just `// swiftlint:disable:next function_parameter_count` on the line above the initializer?

ivanbruel:
It's not about working around but rather trying to understand why it's a best practice when it comes to initializers.

jpsim:
Ok, thinking about this more I would agree. This was even suggested in the initial issue tracking this rule (https://github.com/realm/SwiftLint/issues/415#issue-128552879). This should be just a matter of removing `FunctionConstructor` from the `functionKinds` used to apply the rule: https://github.com/realm/SwiftLint/blob/0.9.0/Source/SwiftLintFramework/Rules/FunctionParameterCountRule.swift#L101

kaybutter:
I tried applying your fix to the `FunctionParameterCountRule` but I'm still seeing the warning with my project. The issue seems to be that my constructor is being recognized as `FunctionMethodInstance` instead of `FunctionConstructor`. :(

marcelofabri:
This was fixed in https://github.com/realm/SwiftLint/pull/942

PR #841, opened 2016-10-14
Ignore count of parameters in init methods

garnett:
Fixes https://github.com/realm/SwiftLint/issues/544
As was mentioned deleting `FunctionConstructor` from the list does not really work. 

My solution looks pretty dumb, but those two tests prove that it works. I've decided to test for `init(` instead of `init` to cut at least some of false positives (such as `initProperties`) but arguably it is not really necessary.

### DIFF
--- a/Source/SwiftLintFramework/Rules/FunctionParameterCountRule.swift
+++ b/Source/SwiftLintFramework/Rules/FunctionParameterCountRule.swift
@@ -103,7 +103,7 @@ public struct FunctionParameterCountRule: ASTRule, ConfigurationProviderRule {
 
     fileprivate func functionIsInitializer(_ file: File, offset: Int, length: Int) -> Bool {
         if let function = (file.contents as NSString)
-            .substringWithByteRange(start: offset, length: length), function.contains("init") {
+            .substringWithByteRange(start: offset, length: length), function.hasPrefix("init") {
             return true
         }
         return false
--- a/Source/SwiftLintFramework/Rules/FunctionParameterCountRule.swift
+++ b/Source/SwiftLintFramework/Rules/FunctionParameterCountRule.swift
@@ -19,6 +19,9 @@ public struct FunctionParameterCountRule: ASTRule, ConfigurationProviderRule {
         name: "Function Parameter Count",
         description: "Number of function parameters should be low.",
         nonTriggeringExamples: [
+            "init(a: Int, b: Int, c: Int, d: Int, e: Int, f: Int) {}",
+            "`init`(a: Int, b: Int, c: Int, d: Int, e: Int, f: Int) {}",
+            "init?(a: Int, b: Int, c: Int, d: Int, e: Int, f: Int) {}",
             "func f2(p1: Int, p2: Int) { }",
             "func f(a: Int, b: Int, c: Int, d: Int, x: Int = 42) {}",
             "func f(a: [Int], b: Int, c: Int, d: Int, f: Int) -> [Int] {\n" +
@@ -26,7 +29,10 @@ public struct FunctionParameterCountRule: ASTRule, ConfigurationProviderRule {
         ],
         triggeringExamples: [
             "func f(a: Int, b: Int, c: Int, d: Int, e: Int, f: Int) {}",
-            "func f(a: Int, b: Int, c: Int, d: Int, e: Int, f: Int = 2, g: Int) {}"
+            "func f(a: Int, b: Int, c: Int, d: Int, e: Int, f: Int = 2, g: Int) {}",
+            "struct Foo {\n" +
+                "init(a: Int, b: Int, c: Int, d: Int, e: Int, f: Int) {}\n" +
+                "func bar(a: b, c: Int, d: Int, e: Int, f: Int, g: Int) {}}"
         ]
     )
 
@@ -39,6 +45,10 @@ public struct FunctionParameterCountRule: ASTRule, ConfigurationProviderRule {
         let nameOffset = Int(dictionary["key.nameoffset"] as? Int64 ?? 0)
         let length = Int(dictionary["key.namelength"] as? Int64 ?? 0)
         let substructure = dictionary["key.substructure"] as? [SourceKitRepresentable] ?? []
+
+        if functionIsInitializer(file, offset: nameOffset, length: length) {
+            return []
+        }
 
         let minThreshold = configuration.params.map({ $0.value }).min(by: <)
 
@@ -89,6 +99,14 @@ public struct FunctionParameterCountRule: ASTRule, ConfigurationProviderRule {
         return (file.contents as NSString)
             .substringWithByteRange(start: offset, length: length)?
             .characters.filter { $0 == equalCharacter }.count ?? 0
+    }
+
+    fileprivate func functionIsInitializer(_ file: File, offset: Int, length: Int) -> Bool {
+        if let function = (file.contents as NSString)
+            .substringWithByteRange(start: offset, length: length), function.contains("init") {
+            return true
+        }
+        return false
     }
 
     fileprivate let functionKinds: [SwiftDeclarationKind] = [
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@
 
 ##### Enhancements
 
-* None.
+* `FunctionParameterCountRule` ignores initializers.
+  [Denis Lebedev](https://github.com/garnett)
+  [#544](https://github.com/realm/SwiftLint/issues/544)
 
 ##### Bug Fixes
 


norio-nomura:
As far as I tested, value of "key.name" has prefix `init(` whether it is `init()` or `init?()` on both of Swift 2.3 and 3.0.
If checking `name.hasPrefix("init(")`, I guess that false positive scenario will happen only when using following:
```swift
func `init`(…) {…}
```
Output of `sourcekitten structure …`:
```terminal.session
$ TOOLCHAINS=com.apple.dt.toolchain.XcodeDefault sourcekitten structure --text '
struct S { init() {}; init?() { return nil }; func `init`(_: Int) {} }
class C { init() {}; init?() { return nil }; func `init`(_: Int) {} }
enum E { case c; init() { return c }; init?() { return nil }; func `init`(_: Int) {} }
'
{
  "key.substructure" : [
    {
      "key.offset" : 1,
      "key.nameoffset" : 8,
      "key.namelength" : 1,
      "key.bodyoffset" : 11,
      "key.bodylength" : 59,
      "key.substructure" : [
        {
          "key.kind" : "source.lang.swift.decl.function.method.instance",
          "key.offset" : 12,
          "key.nameoffset" : 12,
          "key.namelength" : 6,
          "key.bodyoffset" : 20,
          "key.bodylength" : 0,
          "key.length" : 9,
          "key.name" : "init()"
        },
        {
          "key.kind" : "source.lang.swift.decl.function.method.instance",
          "key.offset" : 23,
          "key.nameoffset" : 23,
          "key.namelength" : 7,
          "key.bodyoffset" : 32,
          "key.bodylength" : 12,
          "key.length" : 22,
          "key.name" : "init()"
        },
        {
          "key.kind" : "source.lang.swift.decl.function.method.instance",
          "key.offset" : 47,
          "key.nameoffset" : 52,
          "key.namelength" : 14,
          "key.bodyoffset" : 68,
          "key.bodylength" : 0,
          "key.substructure" : [
            {
              "key.offset" : 59,
              "key.typename" : "Int",
              "key.nameoffset" : 0,
              "key.namelength" : 0,
              "key.length" : 6
            }
          ],
          "key.name" : "init(_:)",
          "key.length" : 22
        }
      ],
      "key.name" : "S",
      "key.length" : 70
    },
    {
      "key.kind" : "source.lang.swift.decl.class",
      "key.offset" : 72,
      "key.nameoffset" : 78,
      "key.namelength" : 1,
      "key.bodyoffset" : 81,
      "key.bodylength" : 59,
      "key.substructure" : [
        {
          "key.kind" : "source.lang.swift.decl.function.method.instance",
          "key.offset" : 82,
          "key.nameoffset" : 82,
          "key.namelength" : 6,
          "key.bodyoffset" : 90,
          "key.bodylength" : 0,
          "key.length" : 9,
          "key.name" : "init()"
        },
        {
          "key.kind" : "source.lang.swift.decl.function.method.instance",
          "key.offset" : 93,
          "key.nameoffset" : 93,
          "key.namelength" : 7,
          "key.bodyoffset" : 102,
          "key.bodylength" : 12,
          "key.length" : 22,
          "key.name" : "init()"
        },
        {
          "key.kind" : "source.lang.swift.decl.function.method.instance",
          "key.offset" : 117,
          "key.nameoffset" : 122,
          "key.namelength" : 14,
          "key.bodyoffset" : 138,
          "key.bodylength" : 0,
          "key.substructure" : [
            {
              "key.offset" : 129,
              "key.typename" : "Int",
              "key.nameoffset" : 0,
              "key.namelength" : 0,
              "key.length" : 6
            }
          ],
          "key.name" : "init(_:)",
          "key.length" : 22
        }
      ],
      "key.name" : "C",
      "key.runtime_name" : "_TtC8__main__1C",
      "key.length" : 69
    },
    {
      "key.kind" : "source.lang.swift.decl.enum",
      "key.offset" : 142,
      "key.nameoffset" : 147,
      "key.namelength" : 1,
      "key.bodyoffset" : 150,
      "key.bodylength" : 77,
      "key.substructure" : [
        {
          "key.substructure" : [
            {
              "key.kind" : "source.lang.swift.decl.enumelement",
              "key.offset" : 156,
              "key.nameoffset" : 156,
              "key.namelength" : 1,
              "key.length" : 1,
              "key.name" : "c"
            }
          ],
          "key.kind" : "source.lang.swift.decl.enumcase",
          "key.offset" : 151,
          "key.nameoffset" : 0,
          "key.namelength" : 0,
          "key.length" : 6
        },
        {
          "key.kind" : "source.lang.swift.decl.function.method.instance",
          "key.offset" : 159,
          "key.nameoffset" : 159,
          "key.namelength" : 6,
          "key.bodyoffset" : 167,
          "key.bodylength" : 10,
          "key.length" : 19,
          "key.name" : "init()"
        },
        {
          "key.kind" : "source.lang.swift.decl.function.method.instance",
          "key.offset" : 180,
          "key.nameoffset" : 180,
          "key.namelength" : 7,
          "key.bodyoffset" : 189,
          "key.bodylength" : 12,
          "key.length" : 22,
          "key.name" : "init()"
        },
        {
          "key.kind" : "source.lang.swift.decl.function.method.instance",
          "key.offset" : 204,
          "key.nameoffset" : 209,
          "key.namelength" : 14,
          "key.bodyoffset" : 225,
          "key.bodylength" : 0,
          "key.substructure" : [
            {
              "key.offset" : 216,
              "key.typename" : "Int",
              "key.nameoffset" : 0,
              "key.namelength" : 0,
              "key.length" : 6
            }
          ],
          "key.name" : "init(_:)",
          "key.length" : 22
        }
      ],
      "key.name" : "E",
      "key.length" : 86
    }
  ],
  "key.offset" : 0,
  "key.length" : 229
}
```

garnett:
@norio-nomura I've added the test case for scenario you've described and surprisingly it behaves as it should be.
Please check my latest commit, does to what you expect?

Source/SwiftLintFramework/Rules/FunctionParameterCountRule.swift
```diff
@@ -40,6 +45,10 @@ public struct FunctionParameterCountRule: ASTRule, ConfigurationProviderRule {
         let length = Int(dictionary["key.namelength"] as? Int64 ?? 0)
         let substructure = dictionary["key.substructure"] as? [SourceKitRepresentable] ?? []
 
+        if file.contents.substring(nameOffset, length: length).containsString("init(") {
```
norio-nomura: - `nameOffset` is counted by byte based number.
So, this make failing test with triggering example such as:
```swift
            "struct Foo {\n /* 👨‍👩‍👧‍👦👨‍👩‍👧‍👦👨‍👩‍👧‍👦 */" +
                "init(a: Int, b: Int, c: Int, d: Int, e: Int, f: Int) {}\n" +
                "func bar(a: b, c: Int, d: Int, e: Int, f: Int, g: Int) {}}"
```
Please use SourceKittenFrameworks's byte base API such as `NSString.substringWithByteRange(start:length:)` if getting substring from `file.contents`.
- This also fails with `init?(`.

What I meant on my last comment were:
- "key.name" already contains string what we need here.  
   Most of offset conversion between byte and utf16 are relatively high cost operations. We want to avoid using that if we can.
- And that has prefix `init(` whether initializer is failable or not

jpsim:
@garnett thanks for your work on this. If you could address @norio-nomura's last comment, I think we could get this in! 💃

jpsim:
Also a changelog entry could be helpful.

garnett:
@jpsim I've rebased original PR and tried to resolve all comments.

Source/SwiftLintFramework/Rules/FunctionParameterCountRule.swift
```diff
@@ -91,6 +101,14 @@ public struct FunctionParameterCountRule: ASTRule, ConfigurationProviderRule {
             .characters.filter { $0 == equalCharacter }.count ?? 0
     }
 
+    fileprivate func functionIsInitializer(_ file: File, offset: Int, length: Int) -> Bool {
+        if let function = (file.contents as NSString)
```
marcelofabri: I think this cast is not needed

Source/SwiftLintFramework/Rules/FunctionParameterCountRule.swift
```diff
@@ -91,6 +101,14 @@ public struct FunctionParameterCountRule: ASTRule, ConfigurationProviderRule {
...
+            .substringWithByteRange(start: offset, length: length), function.contains("init") {
```
marcelofabri: maybe check `hasPrefix` instead?

Source/SwiftLintFramework/Rules/FunctionParameterCountRule.swift
garnett: If I understood correctly, @norio-nomura was explicit in his comment and suggested using `NSString`. If you search for `(file.contents as NSString)` in the project you'll see that sometimes this conversion is used even if redundant. But I can be wrong here.

Source/SwiftLintFramework/Rules/FunctionParameterCountRule.swift
garnett: good spot, fixed

Source/SwiftLintFramework/Rules/FunctionParameterCountRule.swift
jpsim: this might matter when we add Linux support, but the implicit bridging works fine for now and is cleaner, so let's stick with the cleaner approach.

jpsim:
Continued in #942. Thanks for your sustained effort and work on this @garnett!

Issue #784, opened 2016-08-29
False positive for conditional_returns_on_newline with returnKeyType


mattrubin:
The `conditional_returns_on_newline` rule returns a false positive for the following line:
```swift
if textFieldCell.textField.returnKeyType == .Next {
```
It looks like [the regex used by the rule](https://github.com/realm/SwiftLint/blob/050ccd17972e247571d2a494708eb0a7df59acb7/Source/SwiftLintFramework/Rules/ConditionalReturnsOnNewline.swift#L36) simply checks for the word "return" appearing on the same line as an `if` or `guard` and doesn't ensure that the "return" string is the `return` keyword.

I'm not certain how to improve the regex. Checking for a whitespace character after "return" is not sufficient since `return(true)` and `return[]` are – though ugly – technically valid Swift.

jpsim:
Ah, I should've caught this in code review. `return` should be checked to be a built-in syntax type, rather than an identifier, as it is in your example. Thanks for reporting this @mattrubin!

nikolay-dementiev:
regex line should be something like this: (guard|if)[^\n]*([\s]return)[^\n]\n* (I added "[\s]" before return keyword) 

jpsim I'm Right?

PR #842, opened 2016-10-15
Validates if "return" in ConditionalReturnsOnNewline is a keyword

marcelofabri:
Fixes https://github.com/realm/SwiftLint/issues/784

I haven't found a better way to get a token's string than using `NSString.substringWithRange`, so I might be missing something.

### DIFF
--- a/Source/SwiftLintFramework/Rules/ConditionalReturnsOnNewline.swift
+++ b/Source/SwiftLintFramework/Rules/ConditionalReturnsOnNewline.swift
@@ -51,7 +51,7 @@ public struct ConditionalReturnsOnNewline: ConfigurationProviderRule, Rule, OptI
         }.map {
             StyleViolation(ruleDescription: self.dynamicType.description,
                 severity: self.configuration.severity,
-                location: Location(file: file, byteOffset: $0.0.location))
+                location: Location(file: file, characterOffset: $0.0.location))
         }
     }
 
--- a/Source/SwiftLintFramework/Rules/ConditionalReturnsOnNewline.swift
+++ b/Source/SwiftLintFramework/Rules/ConditionalReturnsOnNewline.swift
@@ -23,22 +23,43 @@ public struct ConditionalReturnsOnNewline: ConfigurationProviderRule, Rule, OptI
             "guard true else {\n return true\n}",
             "guard true,\n let x = true else {\n return true\n}",
             "if true else {\n return true\n}",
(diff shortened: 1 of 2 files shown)


marcelofabri:
Forgot to add a changelog entry, will do that later

Source/SwiftLintFramework/Rules/ConditionalReturnsOnNewline.swift
```diff
@@ -23,22 +23,43 @@ public struct ConditionalReturnsOnNewline: ConfigurationProviderRule, Rule, OptI
...
+            }
+
+            let nsString: NSString = file.contents
+            return !tokens.filter { token in
+                guard SyntaxKind(rawValue: token.type) == .Keyword else {
+                    return false
+                }
+
+                let tokenRange = NSRange(location: token.offset, length: token.length)
+                let tokenString = nsString.substringWithRange(tokenRange)
```
norio-nomura: `SyntaxToken`'s `offset` and `length` are byte based numbers.
We need to use SourceKittenFramework's byte based API's e.g. `NSString.substringWithByteRange(start:length:)`.

Source/SwiftLintFramework/Rules/ConditionalReturnsOnNewline.swift
```diff
@@ -23,22 +23,43 @@ public struct ConditionalReturnsOnNewline: ConfigurationProviderRule, Rule, OptI
...
     )
 
     public func validateFile(file: File) -> [StyleViolation] {
         let pattern = "(guard|if)[^\n]*return[^\n]\n*"
         let excludingKinds = SyntaxKind.commentAndStringKinds()
-        return file.matchPattern(pattern, excludingSyntaxKinds: excludingKinds).map {
+        return file.rangesAndTokensMatching(pattern).filter { range, tokens in
+            let kinds = tokens.flatMap { SyntaxKind(rawValue: $0.type) }
+
+            guard kinds.filter(excludingKinds.contains).isEmpty else {
```
norio-nomura: This check caused false negative when match contains comment or string.
e.g. `"if true { return \"YES\" } else { return \"NO\" }"`.

marcelofabri:
Thanks for the tips, @norio-nomura! I've updated the PR 🚀

marcelofabri:
Done 💯

norio-nomura:
Thanks! 🙏

jpsim:
Nice!

Issue #681, opened 2016-06-02
Rule to enforce new-line after `case pattern:` in a `switch` 


brendanjerwin:
Always require a new-line after the colon.

PR #845, opened 2016-10-15
Add SwitchCaseOnNewline rule

marcelofabri:
This fixes https://github.com/realm/SwiftLint/issues/681

I actually haven't checked if the code is inside a `switch`, ~~so this is currently triggering a false positive: `if let case .someEnum(value) = aFunction([key: 2]) {`.~~

~~I thought about checking if `case` is the first keyword in the line instead of checking if it's inside a `switch`, as I think the later is trickier. Thoughts?~~

Update: I added a check to see if `case` is the first keyword in the line

### DIFF
--- a/Source/SwiftLintFramework/Rules/SwitchCaseOnNewlineRule.swift
+++ b/Source/SwiftLintFramework/Rules/SwitchCaseOnNewlineRule.swift
@@ -26,7 +26,10 @@ public struct SwitchCaseOnNewlineRule: ConfigurationProviderRule, Rule, OptInRul
             "/*case 1: */return true",
             "//case 1:\n return true",
             "let x = [caseKey: value]",
-            "let x = [key: .default]"
+            "let x = [key: .default]",
+            "if case let .someEnum(value) = aFunction([key: 2]) {",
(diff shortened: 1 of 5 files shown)


Source/SwiftLintFramework/Rules/SwitchCaseOnNewlineRule.swift
```diff
@@ -0,0 +1,82 @@
...
+        triggeringExamples: [
+            "case 1: return true",
+            "case let value: return true",
+            "default: return true",
+            "case \"a string\": return false"
+        ]
+    )
+
+    public func validateFile(file: File) -> [StyleViolation] {
+        let pattern = "(case[^\n]*|default):[^\\S\n]*[^\n]+"
```
norio-nomura: Sorry, I forgot to wrote in last review.
I think trailing `+` won't be needed.

marcelofabri:
Thanks for reviewing this! I've addressed on the latest commit 👍

norio-nomura:
Thanks! 😄

jpsim:
Nice work! 🎉

Issue #846, opened 2016-10-15
"Attributes should be on their own line" rule


marcelofabri:
I think we should have a rule that _some_ (that's the tricky) part attributes should be on their own line.

The attributes that I'd like to see in that list are:
- `@discardableResult`
- `@available`
- `@NSApplicationMain`
- `@UIApplicationMain`
- `@IBDesignable`

The attributes that I particularly use on the same line:
- `@GKInspectable`
- `@objc`
- `@nonobjc`
- `@NSCopying`
- `@NSManaged`
- `@testable`
- `@IBAction`
- `@IBOutlet`
- `@IBInspectable`

It seems to me that all the attributes that I use on the same line are in these categories:
1. Applied to a `var` 
2. Applied to `import`
3. `@objc` or `@nonobj`

I'd be OK with `@objc` and `@nonobj` being on their own line for functions/types but I don't think we can just use the "where is this attribute" rule because of `@available`.

keith:
👍🏻 this seems useful. For 3 we also put those on their own line. But that's just us!

marcelofabri:
I've started looking into this, but I'm having trouble because the AST doesn't expose the range for the attribute:

```
$ sourcekitten structure --text "@IBOutlet private var label: UILabel"
{
  "key.substructure" : [
    {
      "key.kind" : "source.lang.swift.decl.var.global",
      "key.offset" : 18,
      "key.attributes" : [
        {
          "key.attribute" : "source.decl.attribute.iboutlet"
        }
      ],
      "key.nameoffset" : 22,
      "key.namelength" : 5,
      "key.length" : 18,
      "key.accessibility" : "source.lang.swift.accessibility.private",
      "key.setter_accessibility" : "source.lang.swift.accessibility.private",
      "key.typename" : "UILabel",
      "key.name" : "label"
    }
  ],
  "key.offset" : 0,
  "key.diagnostic_stage" : "source.diagnostic.stage.swift.parse",
  "key.length" : 36
}
```

I can get it from `sourcekitten syntax`, but the type is `source.lang.swift.syntaxtype.attribute.builtin`, which is used for other things (such as `private`).

```
$ sourcekitten syntax --text "@IBOutlet private var label: UILabel"
[
  {
    "offset" : 0,
    "length" : 9,
    "type" : "source.lang.swift.syntaxtype.attribute.builtin"
  },
  {
    "offset" : 10,
    "length" : 7,
[...]

PR #847, opened 2016-10-16
Adds AttributeRule

marcelofabri:
In short, here's the logic I've implemented:
- If the attribute is `@IBAction` or `@NSManaged`, it should always be on
  the same line as the declaration
- If the attribute has parameters, it should always be on the line above
  the declaration
- Otherwise:
  - if the attribute is applied to a variable, it should be on the same line
  - if it's applied to a type or function, it should be on the line above
  - if it's applied to an import (the only option is `@testable import`), 
    it should be on the same line.  

Check #846 for more details on the idea.

TODO:

- [x] Make tests green on Swift 2.2
- [x] Add tests for Swift 2.2 attributes
- [x] Handle parameter in `@objc`


Source/SwiftLintFramework/Rules/AttributesRule.swift
```diff
@@ -0,0 +1,224 @@
...
+                                      dictionary: dictionary)
+        } else {
+            violations = []
+        }
+
+        return validateFile(file, dictionary: dictionary) + violations
+    }
+
+    private func validateTestableImport(file: File) -> [StyleViolation] {
+        let pattern = "@testable[\n]+\\s*import*"
```
marcelofabri: there's no info in AST about imports:

```
$ sourcekitten structure --text "@testable import Module"                    
{
  "key.offset" : 0,
  "key.length" : 23
}
```

Source/SwiftLintFramework/Rules/AttributesRule.swift
```diff
@@ -0,0 +1,224 @@
...
+
+        let tokens = file.syntaxMap.tokensIn(line.byteRange)
+        let attributesTokens = tokens.flatMap { attributeName($0, file: file) }
+        let isViolation: Bool
+
+        do {
+            let previousAttributes = try attributesFromPreviousLines(lineNumber - 1, file: file)
+
+            let alwaysInSameLineAttributes = ["@IBAction", "@NSManaged"]
+            let alwaysInNewLineAttributes = ["@available"]
```
marcelofabri: maybe these should be configured?

Source/SwiftLintFramework/Rules/AttributesRule.swift
```diff
@@ -0,0 +1,224 @@
...
+
+        return [
+            StyleViolation(ruleDescription: self.dynamicType.description,
+                severity: configuration.severity,
+                location: location
+            )
+        ]
+    }
+
+    private func attributesFromPreviousLines(lineNumber: Int, file: File) throws -> [String] {
```
marcelofabri: I don't particularly like this, but it was the only way I figured out to get the texts for the attributes, as there aren't exposed on AST: https://github.com/realm/SwiftLint/issues/846#issuecomment-254016737

marcelofabri:
When running the tests on `SwiftLintFramework` they pass, but on `swiftlint` they don't 😱

marcelofabri:
The failing cases are:

```
"@GKInspectable\n var maxSpeed: Float"
"@discardableResult func a() -> Int"
"@objc\n @discardableResult func a() -> Int"
"@objc\n\n @discardableResult\n func a() -> Int"
```

I think they're failing because they were only introduced on Swift 3 (at least `@discardableResult` - couldn't find information about `@GKInspectable`)

norio-nomura:
> When running the tests on SwiftLintFramework they pass, but on swiftlint they don't 😱

`swiftlint` scheme has environment settings since 997de45 and 940fa37 and `SwiftLintFramework` scheme does not.
So, testing on Xcode 8.0 or later, `swiftlint` scheme uses SourceKit from Swift 2.3, `SwiftLintFramework` scheme uses from Swift 3.x (default).

marcelofabri:
Thanks for explaining it, now it makes sense!

Should we wait on this until we support Swift 3.0 officially then?

norio-nomura:
> Should we wait on this until we support Swift 3.0 officially then?

IMO, we can use `#if swift(>=3.0)` to disable failing examples.

jpsim:
@marcelofabri thanks for your work on this. What's the next step?

marcelofabri:
I need to detect and handle attributes with parameters (I think only `@objc` would benefit for this now) as IMO attributes with parameters should be on their own lines as well.

jpsim:
> I need to detect and handle attributes with parameters (I think only @objc would benefit for this now) as IMO attributes with parameters should be on their own lines as well.

`@available` is also parameterized.

marcelofabri:
>> I need to detect and handle attributes with parameters (I think only @objc would benefit for this now) as IMO attributes with parameters should be on their own lines as well.

> @available is also parameterized.

`@available` is already handled in a special way: https://github.com/realm/SwiftLint/pull/847/files#diff-cd01cc2805cb6eda59b220cf4a7b0648R185. (Although we could remove that when the parameters are handled)

marcelofabri:
@jpsim This is ready for review. It got waaaay more complex than I thought it would be, so feel free to throw any ideas or suggestions to make it better 💯

CHANGELOG.md
```diff
@@ -43,6 +43,23 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#883](https://github.com/realm/SwiftLint/issues/883)
 
+* Add `AttributesRule` which validates if an attribute (`@objc`, `@IBOutlet`,
+  `@discardableResult`, etc) is on the right position:
```
jpsim: "in the right position"

CHANGELOG.md
```diff
@@ -43,6 +43,23 @@
...
+    - If the attribute is `@IBAction` or `@NSManaged`, it should always be on
+    the same line as the declaration
+    - If the attribute has parameters, it should always be on the line above
+    the declaration
+    - Otherwise:
+      - if the attribute is applied to a variable, it should be on the same line
+      - if it's applied to a type or function, it should be on the line above
+      - if it's applied to an import (the only option is `@testable import`), 
+      it should be on the same line.
+  You can also configure what attributes should be always in a new line or in
```
jpsim: "on a new line or on the same line"

CHANGELOG.md
```diff
@@ -43,6 +43,23 @@
...
+  the same line as the declaration with the `always_in_same_line` and
```
jpsim: `always_on_same_line`

CHANGELOG.md
```diff
@@ -43,6 +43,23 @@
...
+    - If the attribute has parameters, it should always be on the line above
+    the declaration
+    - Otherwise:
+      - if the attribute is applied to a variable, it should be on the same line
+      - if it's applied to a type or function, it should be on the line above
+      - if it's applied to an import (the only option is `@testable import`), 
+      it should be on the same line.
+  You can also configure what attributes should be always in a new line or in
+  the same line as the declaration with the `always_in_same_line` and 
+  `always_in_new_line` keys.
```
jpsim: `always_on_line_above`?

Source/SwiftLintFramework/Rules/AttributesRule.swift
```diff
@@ -0,0 +1,385 @@
...
+        } else {
+            violations = []
+        }
+
+        return validateFile(file, dictionary: dictionary) + violations
+    }
+
+    private func validateTestableImport(file: File) -> [StyleViolation] {
+        let pattern = "@testable[\n]+\\s*import"
+        let excludingKinds = SyntaxKind.commentAndStringKinds()
```
jpsim: `@testable` is always a `source.lang.swift.syntaxtype.attribute.builtin` and `import` is always a `source.lang.swift.syntaxtype.keyword`, so you should use those rather than excluding comments and strings.

Source/SwiftLintFramework/Rules/AttributesRule.swift
```diff
@@ -0,0 +1,385 @@
...
+
+        let attributesTokens = Set(attributesTokensWithRanges.map { $0.0 })
+        var isViolation = false
+
+        do {
+            let previousAttributesWithParameters = try attributesFromPreviousLines(lineNumber - 1,
+                                                                                   file: file)
+            let previousAttributes = Set(previousAttributesWithParameters.map { $0.0 })
+
+            let alwaysInSameLineAttributes = configuration.alwaysInSameLine
```
jpsim: `alwaysInSameLineAttributes`

Source/SwiftLintFramework/Rules/AttributesRule.swift
```diff
@@ -0,0 +1,385 @@
...
+            let alwaysInNewLineAttributes =
```
jpsim: `alwaysOnLineAboveAttributes`?

Source/SwiftLintFramework/Rules/AttributesRule.swift
```diff
@@ -0,0 +1,385 @@
...
+                                                line: line, file: file)
+
+            if !attributesTokens.intersect(alwaysInNewLineAttributes).isEmpty {
+                isViolation = true
+            } else if !previousAttributes.intersect(alwaysInSameLineAttributes).isEmpty {
+                isViolation = true
+            } else {
+
+                // ignore whitelisted attributes
+                let operation: Set<String> -> Set<String> -> Set<String> =
```
jpsim: `operation` is a very generic name for this. Can you find a better name?

Source/SwiftLintFramework/Rules/AttributesRule.swift
```diff
@@ -0,0 +1,385 @@
...
+            location = Location(file: file, byteOffset: offset)
+        } else {
+            location = Location(file: file.path)
+        }
+
+        return [
+            StyleViolation(ruleDescription: self.dynamicType.description,
+                severity: configuration.severityConfiguration.severity,
+                location: location
+            )
```
jpsim: open parens hugs the parameters, so should the closing one.

Source/SwiftLintFramework/Rules/AttributesRule.swift
```diff
@@ -0,0 +1,385 @@
...
+            "@testable import SourceKittenFramework",
+            "@objc(foo_x)\n var x: String",
+            "@available(iOS 9.0, *)\n@objc(abc_stackView)\n let stackView: UIStackView",
+            "@objc(abc_addSomeObject:)\n @NSManaged func addSomeObject(book: SomeObject)",
+            "@objc(ABCThing)\n @available(iOS 9.0, *)\n class Thing"
+        ]
+
+        #if swift(>=3.0)
+            let swift3Only = [
+            "@GKInspectable var maxSpeed: Float",
```
jpsim: indentation

Source/SwiftLintFramework/Rules/AttributesRule.swift
```diff
@@ -0,0 +1,385 @@
...
+            "@objc(foo_x) ↓var x: String",
+            "@available(iOS 9.0, *) @objc(abc_stackView)\n ↓let stackView: UIStackView",
+            "@objc(abc_addSomeObject:) @NSManaged\n ↓func addSomeObject(book: SomeObject)",
+            "@objc(abc_addSomeObject:)\n @NSManaged\n ↓func addSomeObject(book: SomeObject)",
+            "@available(iOS 9.0, *)\n @objc(ABCThing) ↓class Thing"
+        ]
+
+        #if swift(>=3.0)
+            let swift3Only = [
+            "@GKInspectable\n ↓var maxSpeed: Float",
```
jpsim: indentation

Source/SwiftLintFramework/Rules/CyclomaticComplexityRule.swift
```diff
@@ -40,7 +40,7 @@ public struct CyclomaticComplexityRule: ASTRule, ConfigurationProviderRule {
 
     public func validateFile(file: File, kind: SwiftDeclarationKind,
                              dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
-        if !functionKinds.contains(kind) {
+        if !SwiftDeclarationKind.functionKinds().contains(kind) {
```
jpsim: sweet refactoring!

Source/SwiftLintFramework/Rules/AttributesRule.swift
marcelofabri: I couldn't find a better name 😅
Maybe we should drop this clever way and just have an if with two paths?

Source/SwiftLintFramework/Rules/AttributesRule.swift
jpsim: Yes

Source/SwiftLintFramework/Rules/AttributesRule.swift
jpsim: maybe call this `let excludeWhitelistedAttributes`?

marcelofabri: I don't know, the more I think the more I realize this is too clever. Are you ok with splitting it in two paths?

jpsim: sure

Source/SwiftLintFramework/Rules/AttributesRule.swift
```diff
@@ -119,16 +119,18 @@ public struct AttributesRule: ASTRule, ConfigurationProviderRule {
...
                 // ignore whitelisted attributes
-                let operation: Set<String> -> Set<String> -> Set<String> =
-                    attributeShouldBeOnSameLine ? Set.union : Set.subtract
-
-                let attributesAfterWhitelist = operation(
-                        operation(attributesTokens)(
-                            previousAttributes.intersect(alwaysOnNewLineAttributes)
-                        )
-                    )(attributesTokens.intersect(alwaysOnSameLineAttributes))
+                let attributesAfterWhitelist: Set<String>
```
jpsim: Yeah, this is definitely clearer!

jpsim:
I'm not sure this is working correctly. Try running this on [realm/realm-cocoa](https://github.com/realm/realm-cocoa) for example.

Getting a violation here: https://github.com/realm/realm-cocoa/blob/v2.1.1/RealmSwift/Tests/TestCase.swift#L39

```swift
@discardableResult
func realmWithTestPath(configuration: Realm.Configuration = Realm.Configuration()) -> Realm {
```

and here: https://github.com/realm/realm-cocoa/blob/v2.1.1/RealmSwift/Tests/TestCase.swift#L45

```swift
override class func setUp() {
```

In total there are 627 violations right now, and I don't think all of them are legit.

marcelofabri:
@jpsim I'll take a look at this, thanks for letting me know!

marcelofabri:
@jpsim I was able to reduce the violations to these:

```
/Users/marcelofabri/Downloads/realm-cocoa-2.1.1/RealmSwift/List.swift:36:19: warning: Attributes Violation: Attributes should be on their own lines in functions and types, but on the same line as variables and imports. (attributes_rule)
/Users/marcelofabri/Downloads/realm-cocoa-2.1.1/RealmSwift/Object.swift:291:19: warning: Attributes Violation: Attributes should be on their own lines in functions and types, but on the same line as variables and imports. (attributes_rule)
/Users/marcelofabri/Downloads/realm-cocoa-2.1.1/RealmSwift/Object.swift:295:19: warning: Attributes Violation: Attributes should be on their own lines in functions and types, but on the same line as variables and imports. (attributes_rule)
/Users/marcelofabri/Downloads/realm-cocoa-2.1.1/RealmSwift/Object.swift:302:19: warning: Attributes Violation: Attributes should be on their own lines in functions and types, but on the same line as variables and imports. (attributes_rule)
/Users/marcelofabri/Downloads/realm-cocoa-2.1.1/RealmSwift/Object.swift:309:19: warning: Attributes Violation: Attributes should be on their own lines in functions and types, but on the same line as variables and imports. (attributes_rule)
/Users/marcelofabri/Downloads/realm-cocoa-2.1.1/RealmSwift/Object.swift:315:19: warning: Attributes Violation: Attributes should be on their own lines in functions and types, but on the same line as variables and imports. (attributes_rule)
/Users/marcelofabri/Downloads/realm-cocoa-2.1.1/RealmSwift/Object.swift:324:19: warning: Attributes Violation: Attributes should be on their own lines in functions and types, but on the same line as variables and imports. (attributes_rule)
/Users/marcelofabri/Downloads/realm-cocoa-2.1.1/RealmSwift/Object.swift:360:19: warning: Attributes Violation: Attributes should be on their own lines in functions and types, but on the same line as variables and imports. (attributes_rule)
/Users/marcelofabri/Downloads/realm-cocoa-2.1.1/RealmSwift/Object.swift:365:19: warning: Attributes Violation: Attributes should be on their own lines in functions and types, but on the same line as variables and imports. (attributes_rule)
/Users/marcelofabri/Downloads/realm-cocoa-2.1.1/Realm/Swift/RLMSupport.swift:23:25: warning: Attributes Violation: Attributes should be on their own lines in functions and types, but on the same line as variables and imports. (attributes_rule)
```

which are all `func`s with `@objc` or `@nonobjc` in the same line as the declaration.

I'll commit it shortly.

jpsim:

marcelofabri:
The tests are failing on Linux because of 

```swift
guard let config = makeConfig(ruleConfiguration, ruleDescription.identifier) else {
            XCTFail()
            return
        }
```

Any ideas? 😬

Source/SwiftLintFramework/Rules/AttributesRule.swift
```diff
@@ -0,0 +1,309 @@
+//
+//  AttributesRule.swift
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 10/15/16.
```
jpsim: 12/12/16?

Source/SwiftLintFramework/Rules/AttributesRulesExamples.swift
```diff
@@ -0,0 +1,124 @@
+//
+//  AttributesRulesExamples.swift
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 09/12/16.
```
jpsim: 12/12/16?

Source/SwiftLintFramework/Rules/RuleConfigurations/AttributesConfiguration.swift
```diff
@@ -0,0 +1,52 @@
+//
+//  AttributesConfiguration.swift
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 26/11/16.
```
jpsim: 12/12/16?

jpsim:
Personally, I'd keep `@objc` and `@nonobjc` on the same line, so I'm trying to configure Realm's `.swiftlint.yml` to do so, but violations are still being reported:

```yaml
attributes_rule:
  always_on_same_line:
    - '@objc'
    - '@nonobjc'
```

am I missing something?

marcelofabri:
@jpsim I've checked why violations are still being reported and almost all of them seems to be because the attribute had a parameter. Another one was because there was two properties without blank lines between them, so the algorithm thought that the attributes from both of them belonged to the last one:

```
@objc private var object: RLMWeakObjectHandle?
@objc private var property: RLMProperty?
```

I'll try to fix these soon. 
(Ugh, this rule is complicated 😅)

Source/SwiftLintFramework/Rules/AttributesRule.swift
marcelofabri: this is right actually 😅

marcelofabri:
@jpsim OK, so I had another look on this and fixed a bug in e4a556d. The remaining violations of this rule on Realm with the updated `.swiftlint.yml` are because of `@objc` with parameters in a new line, when the configuration says they should be on the same line always. For example:

```swift
@objc(RealmSwiftObjectUtil)
public class ObjectUtil
```

We could update the rule implementation to prioritize the "attributes with parameters should be on their own line" over the `always_on_same_line` configuration (and maybe `always_on_new_line`?), but that might be inconvenient for a few people.

In reality, there're so many cases that this rule is probably the most complex one in SwiftLint, so I understand if you feel we shouldn't merge this or make it opt-in.

jpsim:
> In reality, there're so many cases that this rule is probably the most complex one in SwiftLint, so I understand if you feel we shouldn't merge this or make it opt-in.

I'm thinking opt-in might make sense for those reasons.

marcelofabri:
Also, when `@objc` has a name, it's actually another type: `source.decl.attribute.objc.name` (vs `source.decl.attribute.objc`). Maybe we should use that instead of the user-facing type?

marcelofabri:
I've updated it to be opt-in and also renamed the identifier to `attributes` as no other rule has `_rule` in the identifier.

Source/SwiftLintFramework/Rules/AttributesRule.swift
```diff
@@ -26,7 +26,7 @@ public struct AttributesRule: ASTRule, ConfigurationProviderRule {
     public init() {}
 
     public static let description = RuleDescription(
-        identifier: "attributes_rule",
+        identifier: "attributes",
```
jpsim: 👌

marcelofabri:
Let's just wait for CI to be sure 😅

marcelofabri:
Giphy

jpsim:
🎉 congrats @marcelofabri, I know this was a massive undertaking.

PR #849, opened 2016-10-20
Adding ValidIBInspectableRule rule

### LINKED ISSUE #756: Rule for invalid @IBInspectable
I've been thinking of adding a rule flagging invalid uses of `@IBInspectable` such as:
- variables with unsupported types (only a [handful of types](http://stackoverflow.com/q/28749598/130066) are supported by IB)
- variables with inferred types (unsupported by IB)

Do we want to add add rules that detect likely bugs rather than just stylistic issues? The contributing guide does not answer this question.

marcelofabri:
Fixes #756 

Add `ValidIBInspectableRule` rule that checks if `@IBInspectable` declarations
  are valid. An `@IBInspectable` is valid if:
- It's declared as a `var` (not `let`)
- Its type is explicit (not inferred)
- Its type is one of the 
  [supported types](http://help.apple.com/xcode/mac/8.0/#/devf60c1c514)

### DIFF
--- a/Source/SwiftLintFramework/Rules/ValidIBInspectableRule.swift
+++ b/Source/SwiftLintFramework/Rules/ValidIBInspectableRule.swift
@@ -11,6 +11,7 @@ import SourceKittenFramework
 
 public struct ValidIBInspectableRule: ASTRule, ConfigurationProviderRule {
     public var configuration = SeverityConfiguration(.Warning)
+    private static let supportedTypes = ValidIBInspectableRule.createSupportedTypes()
 
     public init() {}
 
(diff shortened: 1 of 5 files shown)


Source/SwiftLintFramework/Rules/ValidIBInspectableRule.swift
```diff
@@ -0,0 +1,130 @@
...
+        // Check if IBInspectable
+        let attributes = (dictionary["key.attributes"] as? [SourceKitRepresentable])?
+            .flatMap({ ($0 as? [String: SourceKitRepresentable]) as? [String: String] })
+            .flatMap({ $0["key.attribute"] }) ?? []
+        let isIBInspectable = attributes.contains("source.decl.attribute.ibinspectable")
+        guard isIBInspectable else {
+            return []
+        }
+
+        // if key.setter_accessibility is nil, it's a `let` declaration
```
marcelofabri: I've assumed this from experimentation - is there a better way to check if it's a `let`?

norio-nomura: Casting to `String` is not needed.
>  is there a better way to check if it's a let?

`guard let _ = ` ? I can't say which is better.

Source/SwiftLintFramework/Rules/ValidIBInspectableRule.swift
```diff
@@ -0,0 +1,130 @@
...
+        // Variable should be of one of the supported types
+        guard let type = dictionary["key.typename"] as? String
+            where supportedTypes().contains(type) else {
+                return violation(dictionary, file: file)
+        }
+
+        return []
+    }
+
+    private func supportedTypes() -> [String] {
```
norio-nomura: Do we need to generate on every access?

marcelofabri:
I've addressed the feedback on 0697cf7. Thanks for taking a look at this 😊

norio-nomura:
Thanks for doing this! 🙏

PR #851, opened 2016-10-21
Adding SyntacticSugar rule

marcelofabri:
I've implemented the "Syntactic Sugar" from https://github.com/raywenderlich/swift-style-guide#syntactic-sugar (also mentioned in #319).

It should be easy to make this rule correctable, but I think it's better to do that on a later PR.

Also, should this be opt-in? (It's not currently)

### DIFF
--- a/Source/SwiftLintFramework/Extensions/File+Cache.swift
+++ b/Source/SwiftLintFramework/Extensions/File+Cache.swift
@@ -78,7 +78,8 @@ extension File {
         }
         set {
             if newValue {
-                responseCache.values[cacheKey] = Optional<[String: SourceKitRepresentable]>.None
+                let value: [String: SourceKitRepresentable]? = nil
+                responseCache.values[cacheKey] = value
             } else {
                 responseCache.values.removeValueForKey(cacheKey)
             }
--- a/Source/SwiftLintFramework/Rules/SyntacticSugarRule.swift
+++ b/Source/SwiftLintFramework/Rules/SyntacticSugarRule.swift
@@ -1,0 +1,55 @@
+//
+//  SyntacticSugarRule.swift
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 21/10/16.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import SourceKittenFramework
+import Foundation
+
+public struct SyntacticSugarRule: Rule, ConfigurationProviderRule {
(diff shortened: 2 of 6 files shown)


Source/SwiftLintFramework/Extensions/File+Cache.swift
```diff
@@ -77,7 +77,8 @@ extension File {
         }
         set {
             if newValue {
-                responseCache.values[cacheKey] = Optional<[String: SourceKitRepresentable]>.None
+                let value: [String: SourceKitRepresentable]? = nil
+                responseCache.values[cacheKey] = value
```
marcelofabri: I've tried just using `responseCache.values[cacheKey] = nil` or `responseCache.values[cacheKey] = .None` but a test started to fail 😬

norio-nomura: Yeah, we are using `Optional.None` as failed state of response from SourceKit.

Source/SwiftLintFramework/Rules/SyntacticSugarRule.swift
```diff
@@ -0,0 +1,55 @@
...
+            "func x(a: ↓Array<Int>, b: Int) -> [Int: Any]",
+            "func x(a: [Int], b: Int) -> ↓Dictionary<Int, String>",
+            "func x(a: ↓Array<Int>, b: Int) -> ↓Dictionary<Int, String>",
+        ]
+    )
+
+    public func validateFile(file: File) -> [StyleViolation] {
+        let types = ["Optional", "ImplicitlyUnwrappedOptional", "Array", "Dictionary"]
+
+        let pattern = "\\b(" + types.joinWithSeparator("|") + ")\\b\\s*<.*?>"
```
norio-nomura: Second `\\b` would not be needed.

marcelofabri:
I've updated the regex in ab7ede0 🚀 

Thanks for reviewing this 💯

norio-nomura:
Could you please resolve conflicts?

marcelofabri:
Done!

norio-nomura:
> > should this be opt-in?

> IMO, I can accept this rule as opt-out.

@realm/swiftlint: How do you think of that this rule should be opt-in or opt-out?

jpsim:
This looks great, and I think it's fine to keep on-by-default. I'll address the merge conflict and make an new PR.

jpsim:
Merged in #896.

PR #854, opened 2016-10-22
Add configuration for trailing_whitespace to ignore comments

### LINKED ISSUE #576: swiftlint set trailing_whitespace warning for comments
This rule is fine for a code, but why does it lint a comments too?
screenshot

jaherhi:
Fixes #576 

Rule `trailing_whitespace` can now be configured to ignore comments using `ignores_comments`, its default value is `true`

### DIFF
--- a/Source/SwiftLintFramework/Rules/TrailingWhitespaceRule.swift
+++ b/Source/SwiftLintFramework/Rules/TrailingWhitespaceRule.swift
@@ -12,7 +12,8 @@ import SourceKittenFramework
 public struct TrailingWhitespaceRule: CorrectableRule, ConfigurationProviderRule,
                                       SourceKitFreeRule {
 
-    public var configuration = TrailingWhitespaceConfiguration(ignoresEmptyLines: false)
+    public var configuration = TrailingWhitespaceConfiguration(ignoresEmptyLines: false,
+                                                               ignoresComments: true)
 
     public init() {}
 
@@ -20,14 +21,15 @@ public struct TrailingWhitespaceRule: CorrectableRule, ConfigurationProviderRule
         identifier: "trailing_whitespace",
         name: "Trailing Whitespace",
         description: "Lines should not have trailing whitespace.",
-        nonTriggeringExamples: [ "//\n" ],
-        triggeringExamples: [ "// \n" ],
-        corrections: [ "// \n": "//\n" ]
+        nonTriggeringExamples: [ "let name: String\n", "//\n", "// \n" ],
+        triggeringExamples: [ "let name: String \n" ],
+        corrections: [ "let name: String \n": "let name: String\n" ]
     )
 
     public func validateFile(file: File) -> [StyleViolation] {
         let filteredLines = file.lines.filter {
-            $0.content.hasTrailingWhitespace() &&
+            !(configuration.ignoresComments && $0.content.isComment()) &&
+                $0.content.hasTrailingWhitespace() &&
                 (!configuration.ignoresEmptyLines ||
                     // If configured, ignore lines that contain nothing but whitespace (empty lines)
                     !$0.content.stringByTrimmingCharactersInSet(.whitespaceCharacterSet()).isEmpty)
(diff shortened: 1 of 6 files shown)


marcelofabri:
I think you should use `SyntaxKind`s instead of checking for `//`. Also, I think this implementation won't work in "real" examples because it only checks the beginning of strings, i.e. this would fail:

```swift
class Example {
  // empty but with trailing spaces          
}
```

You probably can use something like [this](https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Extensions/File%2BSwiftLint.swift#L222-L225) to check if the line only contains comments instead.

jaherhi:
Thanks for the suggestion @marcelofabri! I managed to change my implementation to use `SyntaxKind`, although I'm afraid I don't get what you mean with the example, and I'm trying to see if that could happen using `SyntaxKind` too.

In your example, checking for `//` at the beginning of the string would detect that second line as a comment and, by default, ignore it even if it has trailing spaces, wouldn't it?

Source/SwiftLintFramework/Rules/TrailingWhitespaceRule.swift
```diff
@@ -9,25 +9,31 @@
...
+        nonTriggeringExamples: [ "let name: String\n", "//\n", "// \n",
+            "let name: String //\n", "let name: String // \n" ],
+        triggeringExamples: [ "let name: String \n" ],
+        corrections: [ "let name: String \n": "let name: String\n" ]
     )
 
     public func validateFile(file: File) -> [StyleViolation] {
         let filteredLines = file.lines.filter {
-            $0.content.hasTrailingWhitespace() &&
+            let commentKinds = Set(SyntaxKind.commentKinds())
```
norio-nomura: `commentKinds` is not need to be `Set`.

Source/SwiftLintFramework/Rules/TrailingWhitespaceRule.swift
```diff
@@ -9,25 +9,31 @@
...
+            let lineSyntaxKinds = file.syntaxKindsByLines[$0.index]
+            let lineContainsComment = !lineSyntaxKinds.filter { commentKinds.contains($0) }.isEmpty
```
norio-nomura: `lineContainsComment` should be calculated only if `$0.content.hasTrailingWhitespace()` and `configuration.ignoresComments` are `true`.
Please pay attention to reduce calculation. 🕵🏻

jaherhi:
Changes made to address the feedback 🙂 Thank you!

Source/SwiftLintFramework/Rules/TrailingWhitespaceRule.swift
```diff
@@ -9,25 +9,34 @@
...
         let filteredLines = file.lines.filter {
-            $0.content.hasTrailingWhitespace() &&
+            let commentKinds = SyntaxKind.commentKinds()
+            if $0.content.hasTrailingWhitespace() && configuration.ignoresComments,
+                let lastSyntaxKind = file.syntaxKindsByLines[$0.index].last
+                where commentKinds.contains(lastSyntaxKind) {
+                return false
+            }
+
+            return $0.content.hasTrailingWhitespace() &&
```
norio-nomura: On non violating case, `$0.content.hasTrailingWhitespace()` is evaluated twice. The closure should return `false` first if `$0.content.hasTrailingWhitespace()` returns `false`.

Source/SwiftLintFramework/Rules/TrailingWhitespaceRule.swift
```diff
@@ -46,6 +55,14 @@ public struct TrailingWhitespaceRule: CorrectableRule, ConfigurationProviderRule
         var corrections = [Correction]()
         let fileRegions = file.regions()
         for line in file.lines {
+            let commentKinds = SyntaxKind.commentKinds()
+            if line.content.hasTrailingWhitespace() && configuration.ignoresComments,
```
norio-nomura: Same with my comment on `validateFile(_:)`.
This loop does not need to evaluate following operations if `line.content.hasTrailingWhitespace()` returns `false`.
```swift
if !line.content.hasTrailingWhitespace() {
    correctedLines.append(line.content)
    continue
}
```

jaherhi:
Changes made. Thanks for your comments and patience, @norio-nomura!

I can't think of a way to fix that without manually checking if the comment contains `*/`and whitespace after it. Any suggestion, if that's a case that should work properly?

norio-nomura:
We can continue with filing issue of the false negative and another PR. 😉 
I want to merge this PR. But I forgot about `CHANGELOG.md`.
Could you please add a changelog entry, following the guidelines in `CONTRIBUTING.md`?

jaherhi:
`CHANGELOG.md`updated. I rebased the branch to avoid conflicts after updating it.

norio-nomura:
Thanks! 🙏

jpsim:
Awesome! 🎉

Issue #850, opened 2016-10-20
What is the purpose of opt_in_rules?


castus:
I know that this question might by stupid, but I don't get the idea behind opt_in_rules. Can someone explain me what is the purpose of opt_in_rules?
Thanks

marcelofabri:
`opt_in_rules` are disabled by default (you have to explicitly add them on your configuration file). 

From what I've seen there're a few reasons on why a rule should be opt-in:

* A rule that can have many false positives (`empty_count`)
* A rule that is too slow
* A rule that is not general consensus or only useful in some cases (`force_unwrapping`, `missing_docs`)

jpsim:
That's exactly right.

castus:
Thank you for such a good answer!

PR #856, opened 2016-10-24
added reasons why a new rule should be opt in

masters3d:
this is a modification of https://github.com/realm/SwiftLint/issues/850
all credit goes to @marcelofabri


norio-nomura:
Nice! 👍

Issue #661, opened 2016-05-10
Add Support/Usage For AppCode


maarek:
It would be beneficial to some if SwiftLint added documentation around setting up SwiftLint for use in AppCode from Linting to Autocorrect.

jpsim:
I don't use AppCode, so I can't help too much here.

bestimmaa:
Would be very nice! Is there a way to import these kind of reports in AppCode? Any similar projects doing something like this?

brennantaylor:
If you add swiftlint to a `run script phase` inside the project the messages console will see the warnings and errors and you can click on them to jump to the right file+line number.

![image](https://cloud.githubusercontent.com/assets/2840772/15866206/8c6f0436-2c92-11e6-8222-11cc522ac5ac.png)

yeswolf:
[Here you go](https://plugins.jetbrains.com/plugin/9175) - brings all warnings/errors into the AppCode editor (autocorrect is not yet implemented as a quick-fix). Contributions are welcome [here](https://github.com/bealex/SwiftLintAppCode). 
@jpsim It would be great to have the corresponding section in the README if possible.

jpsim:
I'd be happy to accept a PR that would link to https://plugins.jetbrains.com/plugin/9175 or https://github.com/bealex/SwiftLintAppCode for informing how to integrate SwiftLint with AppCode!

yeswolf:
Added a PR. Quick-fix for Autocorrect is already implemented, so I believe this issue could be closed.

norio-nomura:
Thanks! 🙏

PR #858, opened 2016-10-25
Add information about SwiftLint plugin for AppCode into README.md

yeswolf:
Information about AppCode plugin which integrates SwiftLint (after the conversation here #661)


norio-nomura:
Thanks! 🙏

jpsim:
:tada:

PR #859, opened 2016-10-27
Smarter checking for conditionals

HaloZero:
Smarter checking for conditionals in order to avoid false positives.


igz:
After this merges, we can remove `disable conditional_returns_on_newline` from https://github.com/lyft/Lyft-iOS/pull/5093

HaloZero:
My bad, it seems that the examples triggered were already fixed by the addition of some other code. 

Would it make sense to include these cases without the modifications to the rule?

norio-nomura:
No, I don't. I think existing examples already covered.
Thanks,

Issue #57, opened 2015-05-28
Validate implicit getters


jpsim:
Read-only properties and subscripts should avoid using the `get` keyword.

garnett:
`cat file.swift`

```swift

struct Thing {

    var foo: Int {
      get { return 3 }
      set { _abc = newValue }
    }

    var bar: Int { return 42 }

    var baz: Int {
      get { return 42 }
    }
}
```

 sourcekitten structure --file file.swift:

```
{
  "key.substructure" : [
    {
      "key.kind" : "source.lang.swift.decl.struct",
      "key.offset" : 1,
      "key.nameoffset" : 8,
      "key.namelength" : 5,
      "key.bodyoffset" : 15,
      "key.bodylength" : 162,
      "key.accessibility" : "source.lang.swift.accessibility.internal",
      "key.substructure" : [
        {
          "key.kind" : "source.lang.swift.decl.var.instance",
          "key.offset" : 21,
          "key.nameoffset" : 25,
          "key.namelength" : 3,
          "key.bodyoffset" : 35,
          "key.bodylength" : 58,
          "key.accessibility" : "source.lang.swift.accessibility.internal",
          "key.setter_accessibility" : "source.lang.swift.accessibility.internal",
          "key.typename" : "Int",
          "key.name" : "foo",
          "key.length" : 12
        },
        {
          "key.kind" : "source.lang.swift.decl.var.instance",
          "key.offset" : 100,
          "key.nameoffset" : 104,
          "key.namelength" : 3,
          "key.bodyoffset" : 114,
          "key.bodylength" : 11,
          "key.accessibility" : "source.lang.swift.accessibility.internal",
          "key.length" : 12,
          "key.typename" : "Int",
          "key.name" : "bar"
        },
        {
          "key.kind" : "source.lang.swift.decl.var.instance",
          "key.offset" : 132,
          "key.nameoffset" : 136,
          "key.namelength" : 3,
          "key.bodyoffset" : 146,
          "key.bodylength" : 29,
          "key.accessibility" : "source.lang.swift.accessibility.internal",
          "key.length" : 12,
          "key.typename" : "Int",
          "key.name" : "baz"
        }
      ],
      "key.name" : "Thing",
      "key.length" : 177
    }
  ],
  "key.offset" : 0,
  "key.diagnostic_stage" : "source.diagnostic.stage.swift.parse",
  "key.length" : 179
}
```

[...]

PR #862, opened 2016-10-30
Add ImplicitGetterRule

marcelofabri:
Partially fixes https://github.com/realm/SwiftLint/issues/57

It seems that SourceKit doesn't provide information on `subscript`, which is weird because there is a `FunctionSubscript` case on `SyntaxKind`.

``` swift
struct Thing {

    subscript(index: Int) -> Int {
      return 2
    }
}
```

``` bash
sourcekitten structure --file file.swift
```

``` json
{
  "key.substructure" : [
    {
      "key.offset" : 0,
      "key.nameoffset" : 7,
      "key.namelength" : 5,
      "key.bodyoffset" : 14,
      "key.bodylength" : 58,
      "key.substructure" : [
        {
          "key.offset" : 30,
          "key.nameoffset" : 0,
          "key.namelength" : 0,
          "key.length" : 10,
          "key.typename" : "Int",
          "key.name" : "index"
        }
      ],
      "key.name" : "Thing",
      "key.length" : 73
    }
  ],
  "key.offset" : 0,
  "key.length" : 74
}
```

### DIFF
--- a/Source/SwiftLintFramework/Rules/ImplicitGetterRule.swift
+++ b/Source/SwiftLintFramework/Rules/ImplicitGetterRule.swift
@@ -1,0 +1,119 @@
+//
+//  ImplicitGetterRule.swift
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 29/10/16.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+private func classScoped(value: String) -> String {
+    return "class Foo {\n  \(value)\n}\n"
+}
+
+public struct ImplicitGetterRule: ASTRule, ConfigurationProviderRule {
+    public var configuration = SeverityConfiguration(.Warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "implicit_getter",
+        name: "Implicit Getter",
+        description: "Computed read-only properties should avoid using the get keyword.",
+        nonTriggeringExamples: [
+            classScoped("var foo: Int {\n get {\n return 3\n}\n set {\n _abc = newValue \n}\n}"),
+            classScoped("var foo: Int {\n return 20 \n} \n}"),
+            classScoped("static var foo: Int {\n return 20 \n} \n}"),
+            classScoped("static foo: Int {\n get {\n return 3\n}\n set {\n _abc = newValue \n}\n}"),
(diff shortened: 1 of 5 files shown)


norio-nomura:
False positive pattern with local scoped type:
```swift
class Foo {
  var foo: Int {
    struct Bar {
      var bar: Int {
        get { return 1 } // violation produced on this line
        set { _ = newValue }
      }
    }
    return Bar().bar
  }
}
```

marcelofabri:
Nice catch! Do you have any tips on how to deal with this?

norio-nomura:
If I need to write this rule, I guess I would do:
1. Search keyword `get` from `SyntaxMap`
2. Search deepest structure containing the `get`, like as [`Structure.kindsFor(_:)`](https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Extensions/Structure%2BSwiftLint.swift#L19)
3. Check condition of the structure for violation

The key of avoiding false positive is 2.
I guess 1 would be faster than using `ASTRule`.

Source/SwiftLintFramework/Rules/ImplicitGetterRule.swift
```diff
@@ -0,0 +1,135 @@
...
+
+            return StyleViolation(ruleDescription: self.dynamicType.description,
+                severity: configuration.severity,
+                location: location
+            )
+        }
+    }
+
+    private func variableDeclarationsFor(byteOffset: Int, structure: Structure) ->
+                                                             [[String : SourceKitRepresentable]] {
```
marcelofabri: I'm not a fan of breaking line here, but I couldn't think of other ways to respect the 100 line limit 😬

marcelofabri:
Sorry for taking to take a look at this, but I've implemented your suggestion on 772e702, @norio-nomura. Thanks again 🙌

jpsim:
Awesome work! 🎉

frank61003:
My project can run in my old Macbook Pro but show the error in my new Mac mini. I check my Xcode(11.4.1) and swift version(4.2) are all the same. how to solve this issue?  I use cocoa pod to install swiftlint and the version is 0.27. (pod 'SwiftLint', '0.27'). I check the document many times but still can't find the answer.

https://stackoverflow.com/questions/61947496/how-to-slove-implicit-getter-violation-computed-read-only-subscripts-should-avo

Issue #221, opened 2015-11-18
Add version pinning à la Podfile.lock


ZevEisenberg:
To enforce multiple team members running Swiftlint, the .swiftlint.yml file should include a specifier for the version of SwiftLint that it is expecting to be used with. Then the swiftlint build script (or versions of swiftlint after the first to support such a feature) could check the version as the first step when running, and throw a warning or error if there is a mismatch.

(Aside: Safari/OS X keeps wanting to autocorrect `swiftlint` to `swxftlint`, which as far as I can tell is not a word, and only appears on the Internet here: https://github.com/realm/SwiftLint/issues/169. Anyone else have this problem?)

jpsim:
Thanks for the request! We'd probably go with a minimum supported version rather than requiring an exact version match.

marcelofabri:
Do you guys think it's worth to add a dependency to deal with comparing versions?

Here are some that I found:
https://github.com/mrackwitz/Version
https://github.com/AlexanderNey/SemanticVersioning
https://github.com/trifia/LXSemVer

Or using `NSString` is enough for us? [This SO thread](http://stackoverflow.com/questions/1978456/compare-version-numbers-in-objective-c/) lists some of the problems we might have, such as `1.4.5` would be considered greater than `1.5`.

```swift
if version.compare(currentVersion, options: .NumericSearch) == .OrderedDescending {

}
```

jpsim:
If we need a dependency, I have no trouble adding it.

But that's far from the most challenging decision to make to enable this feature.

Notably, what should be pinned?

The SwiftLint version? Homebrew only allows installing the latest version of a formula, unless resorting to its 'versions' tap. And it's cumbersome for a user to just be told "you don't have the right version of SwiftLint, run this command and try again".

So if not SwiftLint, should we version the rules and keep all previous versions of a rule in our code base, only applying the pinned ones? That would complicate the project significantly, bloat the resulting binary and I'm not sure how useful it would be to end users.

The more I think of this, the more `whitelist_rules` seems like the right design to accomplish the goals that motivated this feature request.

scottrhoyt:
I agree with all of that, @jpsim .

marcelofabri:
I don't think `whitelist_rules` is the right call to solve this. Particularly, it needs that people actively add new rules (and discover them!). Although new rules may trigger warnings on an existing codebase, I'd like to see these violations first and the choose whether I'd embrace the rule and fix the violations or just ignore the rule.

[...]

PR #865, opened 2016-11-04
Add fallback if swiftlint fails

dostrander:
- Currently swiftlint fails if you are on a old version of it. There should be a fallback when it fails and some type of recommended next step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ in the IDE. Just add a new "Run Script Phase" with:
 
 ```bash
 if which swiftlint >/dev/null; then
-  swiftlint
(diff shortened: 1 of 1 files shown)


jpsim:
This isn't quite right. For one, `swiftlint || echo "..."` will always have a return code of 0.

Even if that were addressed, always logging "Is SwiftLint installed and up to date?" no matter of the _reason_ of the failure is misleading.

Seems like overall, this PR aims to address a larger issue, which is one of version enforcement and compatibility, for which discussions around potential strategies to resolve are actively happening in #221. I'd prefer focusing our energies on finding a good solution.

Issue #227, opened 2015-11-22
Add rule for proper indentation with preconfigured number of spaces


Johennes:
It would be great to check for a consistent indentation using e.g. 2 or 4 space for each level of nesting. Here are some violating examples:

``` swift
class Foo {
let bar: String
}
```

``` swift
class Foo {
    func bar() {
      return "foo"
    }
}
```

In addition, it'll also be nice to check for a consistent indentation of multi-line array/dictionary or string constructs. Violating examples:

``` swift
let a = ["foo",
"foo",
"foo"]
```

``` swift
let a = "foo"
+ "bar"
+ "foobar"
```

ecnepsnai:
Any update on this @jpsim ?  Would really value this rule.

jpsim:
It's up for grabs, no one has started work on this as far as I know.

arsonik:
+1

qasim:
Working on this over at https://github.com/realm/SwiftLint/compare/master...qasim:qasim/add-indentation. ☺️

salutis:
@qasim Any updates on this one?

ldiqual:
@Johennes @salutis @ecnepsnai I've created a PR addressing the first part of this issue. Please feel free to 👍 if you'd like to express interest! https://github.com/realm/SwiftLint/pull/873

fredpi:
The new `indentation_width` opt-in rule is now available (see https://github.com/realm/SwiftLint/pull/2765), so maybe we can close this issue? Or should we wait until there's also a more advanced / flexible indentation rule?

jpsim:
Let's close this. Anyone should feel empowered to file new, more specific issues about further enhancements. Thanks @fredpi!

keith:
FYI it looks like this rule is incompatible with this style:

```swift
        XCTAssertEqual(foo,
                       "blah")
```

which we use when the first line would overflow our column length limit

fredpi:
@keith That's by design, as the rule suggests the following style in case of overflow:

```swift
XCTAssertEqual(
   foo,
   "blah"
)
```

jpsim:
It' be awesome to add a configuration option to support the first style too.

patrickbdev:
@fredpi @jpsim was support added for the first style?
It'd be useful for instances like:
```
[textField,
 callButton].forEach(addSubview)
```

fredpi:
[...]

PR #873, opened 2016-11-10
Added IndentationRule

ldiqual:
Here's a proposal to fix https://github.com/realm/SwiftLint/issues/227

The rule enforces that code indentation uses 4 spaces or one tab, and not a mix of both. Let me know if you'd like to enforce 2/4 spaces or tabs from the configuration. Also let me know if you'd like this rule to be opt-in.

I don't believe there's a nice way to autocorrect this.

I'm also not 100% sure that triggering and non-triggering examples are properly validated because I get a bunch of errors that are unrelated to my rule.

Most notably:

image

The current SwiftLint codebase has about 230 validation issues because the indentation is inconsistent. The most common source of error is something like this:

```
private func getFiles(path: String, action: String, useSTDIN: Bool, quiet: Bool,
                          useScriptInputFiles: Bool) -> Result<[File], CommandantError<()>> {
```

I'm proposing to format it this way:

```
private func getFiles(
    path: String,
    action: String,
    useSTDIN: Bool, quiet: Bool,
    useScriptInputFiles: Bool
) -> Result<[File], CommandantError<()>> {
```

Or at the very least:

```
private func getFiles(path: String, action: String, useSTDIN: Bool, quiet: Bool,
    useScriptInputFiles: Bool) -> Result<[File], CommandantError<()>> {
```

Let me know how you want to deal with this kind of stuff.

### DIFF
--- a/Source/SwiftLintFramework/Models/MasterRuleList.swift
+++ b/Source/SwiftLintFramework/Models/MasterRuleList.swift
@@ -57,6 +57,7 @@ public let masterRuleList = RuleList(rules:
     ForceUnwrappingRule.self,
     FunctionBodyLengthRule.self,
     FunctionParameterCountRule.self,
+    IndentationRule.self,
     LeadingWhitespaceRule.self,
     LegacyCGGeometryFunctionsRule.self,
     LegacyConstantRule.self,
--- a/Source/SwiftLintFramework/Rules/IndentationRule.swift
+++ b/Source/SwiftLintFramework/Rules/IndentationRule.swift
@@ -1,0 +1,81 @@
+//
+//  IndentationRule.swift
+//  SwiftLint
+//
+//  Created by Lois Di Qual on 11/9/16.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+public struct IndentationRule: ConfigurationProviderRule, SourceKitFreeRule {
+    public var configuration = SeverityLevelsConfiguration(warning: 400, error: 1000)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "indentation",
+        name: "Indentation",
+        description: "Source code should have consistent indentation. " +
+            "Accepted indentations are tab characters or 4 spaces.",
+        nonTriggeringExamples: [
+            "\t",
+            "    ",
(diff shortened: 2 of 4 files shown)


ldiqual:
@jpsim @norio-nomura Let me know you'd like me to proceed with this

jpsim:
The fact that this codebase has 230 violations based on the way this rule is implemented might be your first sign that this may not be the right way to do this.

Indentation semantics are actually quite complex when you think of it. Just look at the complexity of [`lib/IDE/Formatting.cpp`](https://github.com/apple/swift/blob/master/lib/IDE/Formatting.cpp) in the Swift compiler. They're not perfect by any means, but they do take lots of different situations under consideration.

ldiqual:
@jpsim Thanks a lot for following up!

I think https://github.com/realm/SwiftLint/issues/227 can be broken down in two points:

 1. Make sure that all indentation is consistent, which means each "tab" has to be 1-tab, 2-space, or 4-space depending on the configuration.
 2. Enforce a specific indentation style for nesting. For instance make sure that nested code has one more level of indentation than the parent.

While I agree that point 2. requires a more complex approach, this PR only intends to address point 1. I believe point 2. could be addressed in a separate rule (eg: `nested_indentation`).

Regarding violations in SwiftLint, while I believe that it'd be reasonable to make SwiftLint indentation 4-space only, we could definitely make the rule opt-in so that it doesn't affect SwiftLint itself or projects that would upgrade to a new version of SwiftLint.

Let me know what you think.

jpsim:
I'm not comfortable reviewing this, if other contributors want to take a stab at giving feedback, I encourage you to do so. I just haven't thought enough of what indentation-related rules in SwiftLint should look like.

Coeur:
Xcode auto-formatting will align parameters, regardless of the count of spaces, like in:

    let url = URL(fileURLWithPath: "/",
                  isDirectory: true)

This makes `spaceCount % 4` irrelevant, and the pull should be denied for that in its current state.

Alignment like that also makes the usage of spaces a requirement for compatibility. So a clearer rule could be to simply warn on tabulations.

marcelofabri:
What's the current status of this?

marcelofabri:
Closing this by now. Feel free to open another PR if someone wants to revive this.

PR #876, opened 2016-11-12
Comma rule does not get triggered in all situations

sadefigu:
Improving comma rule to support expressions that begin with the comma. (fixes a gap when there are consecutive comma violations)

For example, if the expression is 'let a = [1,2,3]', the previous comma rule regexp would consume '1,2' and the remaining ',3' would not be consumed by the regexp because it does not support something that begins with a comma.


jpsim:
Thanks for the PR, @sadefigu!

PR #877, opened 2016-11-14
`switch_case_on_newline` bugfixes

### LINKED ISSUE #874: switch_case_on_newline should ignore comments
Triggering example:

```swift
switch myValue {
case .myCase: // error from network
    break
default:
    break
}
```

Expectation would be to ignore comments on the same line as the case.

marcelofabri:
Fixes #874 and #878


jpsim:
Thanks 🙏

PR #879, opened 2016-11-16
closure_spacing accepts empty bodies with a space

### LINKED ISSUE #875: closure_spacing should ignore empty bodies
Triggering example:

```swift

class CustomResponse: Response {
        override init() { } // triggers a warning
}
```

The expectation would be that empty bodies with a single space does not trigger a warning.

marcelofabri:
Fixes #875

### DIFF
--- a/Source/SwiftLintFramework/Rules/ClosureSpacingRule.swift
+++ b/Source/SwiftLintFramework/Rules/ClosureSpacingRule.swift
@@ -21,7 +21,9 @@ public struct ClosureSpacingRule: Rule, ConfigurationProviderRule, OptInRule {
         description: "Closure expressions should have a single space inside each brace.",
         nonTriggeringExamples: [
             "[].map ({ $0.description })",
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Rules/ClosureSpacingRule.swift
```diff
@@ -90,9 +92,12 @@ public struct ClosureSpacingRule: Rule, ConfigurationProviderRule, OptInRule {
         let matchedUpBraces = matchBraces(validBraces(file))
 
         var violationRanges = matchedUpBraces.filter {
-            //removes enclosing brances to just content
+            // removes enclosing brances to just content
             let content = file.contents.substring($0.location + 1, length: $0.length - 2)
-            if content.isEmpty { return false } // case when {} is not a closure
+            if content.isEmpty || content == " " {
```
marcelofabri: this is kind of ugly, but I didn't want to make multiple spaces valid, only `{}` and `{ }`

PR #882, opened 2016-11-20
Add NimbleOperatorRule

### LINKED ISSUE #881: Add a NimbleOperatorRule
We use [Nimble](https://github.com/Quick/Nimble) on our unit tests, and we prefer using its [operator overloads](https://github.com/Quick/Nimble/#operator-overloads) instead of functions:

```swift
// triggers
expect(seagull.squawk).toNot(equal("Hi"))
expect(10).to(beGreaterThan(2))

// doesn't trigger
expect(seagull.squawk) != "Hi!"
expect(10) > 2
```

Creating a rule to enforce this shouldn't be too hard. However, even if it's opt-in I know it's a very specific use case. Would this be a welcome rule addition?

marcelofabri:
Fixes #881

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,12 @@
 
 ##### Enhancements
 
-* None.
+* Add `NimbleOperatorRule` opt-in rule that enforces using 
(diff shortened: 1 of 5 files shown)


jpsim:
Merged in #899.

PR #886, opened 2016-11-21
Addded TrailingCommaRule

### LINKED ISSUE #883: Trailing Comma Rule
In collections, it's possible to keep a trailing comma in the end:

```swift
let array = [
  1,
  2,
  3,
]
```

I've seen people that like this pattern (because the last line wouldn't change in a diff if another item is added later) and people who don't (because it's an extra unneeded symbol).

We could have a configuration so both styles (they're exclusive) could be validated in the same rule.

marcelofabri:
Fixes #883.

How would be the best way to handle both styles (mandatory comma and forbidden comma)? A configuration (if so, how should the rule description be)? Another rule?

I made this rule opt-in as even SwiftLint codebase isn't consistent. Thoughts?

PS: The changelog entry is missing on purpose as I thought it'd be better to think on how to handle both styles first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,12 @@
 * `closure_spacing` rule now accepts empty bodies with a space.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#875](https://github.com/realm/SwiftLint/issues/875)
+  
+* Add `TrailingCommaRule` to enforce/forbid trailing commas in arrays and
+  dictionaries. The default is to forbid them, but this can be changed with
+  the `mandatory_comma` configuration.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#883](https://github.com/realm/SwiftLint/issues/883)
 
 ##### Bug Fixes
 
--- a/Source/SwiftLintFramework/Models/MasterRuleList.swift
+++ b/Source/SwiftLintFramework/Models/MasterRuleList.swift
@@ -79,6 +79,7 @@ public let masterRuleList = RuleList(rules:
     SwitchCaseOnNewlineRule.self,
     SyntacticSugarRule.self,
     TodoRule.self,
+    TrailingCommaRule.self,
     TrailingNewlineRule.self,
     TrailingSemicolonRule.self,
     TrailingWhitespaceRule.self,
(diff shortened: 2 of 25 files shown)


jpsim:
I think this rule can be on-by-default as long as it's also configurable.

marcelofabri:
But how would I handle the configurations in the examples? They're very different according to the specified configuration.

jpsim:
> But how would I handle the configurations in the examples? They're very different according to the specified configuration.

True, but that wouldn't be unprecedented. We have other rules where we reuse the convenient `RuleDescription` examples for tests with the default configuration, but have more explicit tests for custom configurations. For example, see [RulesTests.testColon()](https://github.com/realm/SwiftLint/blob/0.13.0/Tests/SwiftLintFramework/RulesTests.swift#L19-L89) where we first run the standard `verifyRule(ColonRule.description)` test, followed by passing in a custom `RuleDescription` and configuration: `verifyRule(description, ruleConfiguration: ["flexible_right_spacing": true])`.

Could do the same here, although there's an argument to be made that we should have a nicer way to specify different test examples for each configuration to test.

marcelofabri:
@jpsim Another (small) problem with the current description behavior is that `swiftlint rules [rule]` would always show the examples for the default configuration, but it could be interesting to show for the provided one instead. However I do feel that this feature is mostly unused and few people are aware of it.

CHANGELOG.md
```diff
@@ -36,6 +36,12 @@
 * `closure_spacing` rule now accepts empty bodies with a space.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#875](https://github.com/realm/SwiftLint/issues/875)
+  
+* Add `TrailingCommaRule` to enforce/forbid trailing commas in arrays and
+  dictionaries. The default is to forbid them, but this can be changed with
+  the `mandatory_comma` configuration.
```
jpsim: just two trailing spaces: https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes

CHANGELOG.md
```diff
@@ -36,6 +36,12 @@
 * `closure_spacing` rule now accepts empty bodies with a space.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#875](https://github.com/realm/SwiftLint/issues/875)
+  
+* Add `TrailingCommaRule` to enforce/forbid trailing commas in arrays and
+  dictionaries. The default is to forbid them, but this can be changed with
+  the `mandatory_comma` configuration.    
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#886](https://github.com/realm/SwiftLint/issues/886)
```
jpsim: link to the issue this resolves, not the PR adding it: https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes

PR #895, opened 2016-11-25
[New Rule] weak_delegate

jpsim:
This is #753 rebased on `master`. @AliSoftware should this be opt-in?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#57](https://github.com/realm/SwiftLint/issues/57)
 
-* Add `WeakDelegateRule` opt-in rule to enforce delegate instance variables to
-  be `weak`.  
+* Add `WeakDelegateRule` rule to enforce delegate instance variables to be
+  marked as `weak`.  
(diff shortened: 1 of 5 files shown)


AliSoftware:
Oh I thought this rule was already merged a while ago, totally forgot it was still not integrated in master.

Not sure it's worth making it opt-in, seems like a good rule to enforce for everybody by default.
It seems that I stated it was opt-in in the CHANGELOG but didn't actually make it opt-in in the code, right? So maybe just fixing the changelog entry would be the way to go.

AliSoftware:
Maybe before merging this, it could worth renaming the rule, as https://github.com/realm/SwiftLint/pull/753#discussion_r83897262 suggests that some would want to add a rule configuration to this rule to allow `unowned` delegates in a future PR… but couldn't think of a generic-enough name for it, so… I'll let you decide. Maybe `weak_delegate` is OK after all even if a future rule configuration allows `unowned` too…

jpsim:
> Maybe `weak_delegate` is OK after all even if a future rule configuration allows `unowned` too…

I think that's fine. I've removed "opt-in" from the changelog entry and will now merge this.

PR #913, opened 2016-11-30
Added a rule to enforce where colons can go in Dictionary type declar…

codeman9:
…ations and literals.


jpsim:
Instead of creating a new rule for this, I think we should just expand the current ColonRule to support dictionaries. What do you think?

codeman9:
I was thinking about that, too. Would you think another configuration option would be appropriate? Something like "apply_to_dictionaries: true"?

jpsim:
That works.

codeman9:
Not intentional. I will update. Thanks!

PR #923, opened 2016-12-02
Update weak delegate rule to only match delegate suffix

philwebster:
The rule was warning on lines like the following line in our tests:

`var createActivityDelegateMethodCalled = false`

I'm not sure if anyone names their delegates with `delegate` prefixed or in the middle (which this change would miss), but this seems like a reasonable modification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 ##### Bug Fixes
 
 * Fix `weak_delegate` rule reporting a violation for variables containing
-  but not ending in `delegate`.
+  but not ending in `delegate`. 
   [Phil Webster](https://github.com/philwebster)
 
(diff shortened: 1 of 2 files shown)


jpsim:
Can you please add a changelog entry? Probably under "Bug Fixes" would be best.

@AliSoftware this seem reasonable to you?

philwebster:
Done. Let me know if there's anything else!

AliSoftware:
:+1:

marcelofabri:
Maybe we could add an entry to `nonTriggeringExamples` to make sure we don't break this in the future?

jpsim:
Looks good @philwebster, but please ping me once Travis has had a chance to catch up, and as long as everything's green we'll merge.

PR #926, opened 2016-12-04
Adding a code of conduct file.

stupergenius:


jpsim:
Thanks for the PR! However, there's a certain logic to keeping this document centralized on Realm's website, as we link to in the README. It means that if we have to make adjustments, they only need to be done in one place. That's why we consistently follow this approach across all of our open source projects. So I don't think this is something we want to be doing at this time. I hope you'll contribute to SwiftLint again!

stupergenius:
Yeah. I can see that. Thanks!

PR #930, opened 2016-12-06
Add number_separator opt-in rule

### LINKED ISSUE #924: Use underscore separation for large integers
Any number greater than 4 characters should use ``_`` for separation.

```swift
// bad
let num = 10000000

// good
let num = 10_000_000
```

marcelofabri:
Fixes #924 

Maybe we should add a `min_length` configuration that defaults to 4? (So `1000` would be valid, but `10000` wouldn't)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,10 @@
 
 ##### Enhancements
 
-* None.
+* Add `number_separator` opt-in rule that enforces that underscores are
+  used as thousand separators in large numbers.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#924](https://github.com/realm/SwiftLint/issues/924)
 
 ##### Bug Fixes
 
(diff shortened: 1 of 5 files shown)


jpsim:
This is great! Though I'm wondering how this handles other number literals, like hex or exponentials.

marcelofabri:
Totally forgot about those 😅

Should we apply this rule only on decimals or does it make sense for the other types?

```
A binary number, with a 0b prefix
An octal number, with a 0o prefix
A hexadecimal number, with a 0x prefix
```

jpsim:
I'd say this rule only makes sense with decimals.

marcelofabri:
I agree, but I don't think I've ever used the other literals. Should we trigger if they contain an underscore or just ignore them completely? 🤔

I'd say we should skip completely, because the grouping rules for each one of them could be different.

PR #935, opened 2016-12-07
Skip declarations in protocols in weak_delegate rule

### LINKED ISSUE #934: Minor weak_delegate rule bugs
Hi there! I love the new `weak_delegate` rule. It's super useful! We were however able to find a couple of cases where the `weak_delegate` rule is erroneously calling out a `var`:

The first is in a protocol:
```
protocol P {
  var delegate: AnyObject? { get set }
}
```
`weak` is not allowed in that context, so this rule should probably skip delegate-like properties in protocols. The next is a variable that refers to a primitive:
```
class C {
  var delegateCalled = false // Used in the context of a test mock object
}
```

In both cases, there is no action in the code that can be taken to fix the issue. This can be reproduced by just dropping those 2 declarations into a file called test.swift and running `swiftlint lint --path test.swift` with `swiftlint` version 0.13.2. Thanks!

marcelofabri:
Fixes #934

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@
 * Fix `weak_delegate` rule reporting a violation for variables containing
   but not ending in `delegate`.  
   [Phil Webster](https://github.com/philwebster)
+  
+* Fix `weak_delegate` rule reporting a violation for variables in protocols
+  declarations.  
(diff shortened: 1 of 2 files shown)


jpsim:
💃

Issue #390, opened 2016-01-20
Distribution through Cocoapods


nmccann:
Hi,

I've been using SwiftLint for a few days now but found a few problems with how it is installed. A few of the problems (from my perspective):
- Requires action from new developers (either `brew install` or manually installing package)
- Difficult to specify a version that should be used
- Installed globally, which isn't good for shared environments like self-hosted CI (Bamboo). Particularly if you don't have permissions to do such an installation.

I have two suggestions:
- Provide a zipped copy of the binary and framework along with the package, source etc. that is currently included in each release.
- Provide a pod spec that would download the binary, similar to what is used by [R.swift](https://github.com/mac-cain13/R.swift)

Users who aren't interested in using Cocoapods could just directly download the zipped binary for the version they want.

I've got this working with a podspec in a private repo and a zipped binary/framework that I manually generated using `make prefix-install`. Having to do all of that manually isn't ideal though because it makes it harder to maintain. And I feel like other people could benefit from this functionality.

I think this would also partially solve #302 

Please let me know if there are any issues/concerns with this approach. If there aren't any, I could try to make a PR that handles automatic creation of the zipped binary during the release process.

Thanks,
Noah

jpsim:
> Requires action from new developers (either brew install or manually installing package)
> Difficult to specify a version that should be used
> Installed globally, which isn't good for shared environments like self-hosted CI (Bamboo). Particularly if you don't have permissions to do such an installation.

All of these can be addressed for your use case (and #302) by bundling a pre-built version of SwiftLint with your project, similar to what [Jazzy does with SourceKitten](https://github.com/realm/jazzy/tree/master/lib/jazzy/SourceKitten).

If you keep this directory structure, `swiftlint` will always link the framework in `../Frameworks`:
```.
├── Frameworks
│   └── SwiftLintFramework.framework
└── bin
    └── swiftlint
```

> Provide a zipped copy of the binary and framework along with the package, source etc. that is currently included in each release.
> Provide a pod spec that would download the binary, similar to what is used by R.swift

I have a hard time seeing us adding a third or fourth installation method for SwiftLint (the existing two being `brew` and `.pkg`).

A Podspec might make sense if anyone wants to use `SwiftLintFramework` in a CocoaPods project, however, so I'd merge a PR adding that.

nmccann:
[...]

PR #936, opened 2016-12-07
Add SwiftLint CLI podspec for binary distribution via CocoaPods

jpsim:
Addresses #390.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ matrix:
   include:
     - script:
         - set -o pipefail && script/cibuild | xcpretty
-        - pod lib lint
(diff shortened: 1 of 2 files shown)


jpsim:
Stuff we should do once we confirm this works:

1. A release is necessary for this to work. Specifying `pod 'SwiftLint', git: '...', branch: '...'` doesn't work with this podspec approach.
2. Update `make release` to build `portable_zip`
3. Update installation options/instructions in the README
4. Add the podspecs to the list of places where the version must be updated
5. Update the `Releasing.md` instructions to account for the pod trunk pushes

PR #941, opened 2016-12-07
Fix `traling_comma` when using commas in comments.

### LINKED ISSUE #940: trailing_comma rule false positive for comments
Hi,

the following code will create a false positive in 0.13.1:

```
let example = [ "a",
                "b"
                // "c",
        ]
```

marcelofabri:
Fixes #940

### DIFF
--- a/Source/SwiftLintFramework/Rules/TrailingCommaRule.swift
+++ b/Source/SwiftLintFramework/Rules/TrailingCommaRule.swift
@@ -68,16 +68,17 @@ public struct TrailingCommaRule: ASTRule, ConfigurationProviderRule {
             return []
         }
 
-        if let (startLine, _) = file.contents.bridge().lineAndCharacter(forByteOffset: bodyOffset),
-            let (endLine, _) = file.contents.bridge().lineAndCharacter(forByteOffset: lastPosition),
+        let contents = file.contents.bridge()
+        if let (startLine, _) = contents.lineAndCharacter(forByteOffset: bodyOffset),
+            let (endLine, _) = contents.lineAndCharacter(forByteOffset: lastPosition),
             configuration.mandatoryComma && startLine == endLine {
-            // shouldn't trigger if mandatory comma style and is a single-line declaration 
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Rules/TrailingCommaRule.swift
```diff
@@ -99,6 +102,22 @@ public struct TrailingCommaRule: ASTRule, ConfigurationProviderRule {
             )
         ]
     }
+
+    private func trailingCommaIndex(_ contents: String, file: File, offset: Int) -> Int? {
+        // swiftlint:disable:next force_try
+        let regex = try! NSRegularExpression(pattern: ",", options: [.ignoreMetacharacters])
+        let range = NSRange(location: 0, length: contents.characters.count)
```
jpsim: `contents.characters.count` is not always equal to `(contents as NSString).length`

Source/SwiftLintFramework/Rules/TrailingCommaRule.swift
```diff
@@ -99,6 +102,22 @@ public struct TrailingCommaRule: ASTRule, ConfigurationProviderRule {
...
+        let ranges = regex.matches(in: contents, options: [], range: range).map { $0.range }
+
+        // skip commas in comments
+        return ranges.filter {
```
jpsim: I wonder if there's a more efficient way to do this...

jpsim:
Thanks for tackling this!

Source/SwiftLintFramework/Rules/TrailingCommaRule.swift
marcelofabri: I don't think this will be an issue since in most cases `ranges` would be empty. We could cache the regex tough.

jpsim: nah, I think it's fine as-is.

jpsim:

jpsim:
🎉

PR #943, opened 2016-12-08
Add a dictionary colon rule. Also make it so that the description tha…

codeman9:
…t is printed matches the configuration.

Redo rule on current master.

### DIFF
--- a/Tests/SwiftLintFrameworkTests/RulesTests.swift
+++ b/Tests/SwiftLintFrameworkTests/RulesTests.swift
@@ -9,7 +9,8 @@
 import SwiftLintFramework
 import XCTest
 
-// swiftlint:disable:next type_body_length
+// swiftlint:disable type_body_length
+// swiftlint:disable file_length
 class RulesTests: XCTestCase {
 
     func testClosingBrace() {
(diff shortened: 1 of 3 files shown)


Tests/SwiftLintFrameworkTests/RulesTests.swift
```diff
@@ -9,7 +9,8 @@
 import SwiftLintFramework
 import XCTest
 
-// swiftlint:disable:next type_body_length
+// swiftlint:disable type_body_length
```
jpsim: This could be: `// swiftlint:disable type_body_length file_length`

jpsim: make sure you're using SwiftLint 0.13.2, since support for this syntax was only recently added

codeman9: Hmmm...I did a `brew upgrade swiftlint` and it updated to 0.13.2, but I'm still getting an error and a warning when I put this on a single line.

jpsim:
I'm not terribly thrilled about the way this is implemented, but unfortunately can't really give concrete feedback as to why. Maybe another contributor can chime in and help review?

codeman9:
Since my other pull request makes the custom rules correctable, this particular rule as a .swift rule vs a custom regex becomes less valuable for us. In fact, my other pull request #944 uses the regex for this rule in its correctable test to show that this rule isn't as valuable as part of the ColonRule.swift.

codeman9:
I think I'm going to decline/close this PR. I've been doing more work on getting the custom rules to work the way we want and have implemented a regex that matches this need there as well as being correctable. See PR #944

PR #944, opened 2016-12-08
Make custom rules correctable with a template in the configuration.

codeman9:


codeman9:
Any thoughts on this?

codeman9:
By using these changes, we would be able to change the implementation of many of the existing `CorrectableRule`s to just return a pattern and a template and pass through this somehow. Maybe via another protocol adherence of some sort. That way, the rules that are basically regex matchers can all be very easily implemented.

I have written a few tests for some of the existing rules by just taking their regex pattern, their template, and triggering examples and passing them through these changes and all seems well.

masters3d:
I actually think that it would be better to have full plug-in support instead of making Custom Rules more like built in rules. I'd be in favor of something like #711 and phase out custom rules. Plug-In rules would be first class rules with out the limitations of Custom Rules.

abaire:
Looks like both this and #711 have been abandoned are there any plans to allow extensible correctable rules?

Dschee:
Hi @abaire, I know this is an old one and I'm not sure what others think about this feature, but to me at least the idea doesn't sound that wrong, if the custom rules can be made correctable in a sensible way without making the code base overly complex, I'd be in favor of this change. But: From the code changes alone I really can't see how this is intended to work since there's no description in this pull requests first post and also there's no changes in the README nor a new CHANGELOG entry. Could you please further explain how this is supposed to work?

PR #950, opened 2016-12-08
Add RedundantStringEnumValueRule

### LINKED ISSUE #946: Add rule for redundant string enum values
```swift
enum Numbers: String {
    case one = "one"
    case two = "two"
    case three = "three"
}
```

is redundant. Could be simply

```swift
enum Numbers: String {
    case one
    case two
    case three
}
```

marcelofabri:
Fixes #946

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@
 
 * Add `EmojiReporter`: a human friendly reporter.  
   [Michał Kałużny](https://github.com/justMaku)
+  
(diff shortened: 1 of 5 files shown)


jpsim:
😍. It's amazing having contributors like you 😊 .

Will review shortly.

jpsim:
I made some improvements in 43380ce, but it was already pretty amazing that you were able to build this so quickly!

Once that passes CI, we can merge this.

PR #904, opened 2016-11-27
Add FileHeaderRule

marcelofabri:
Fixes #844

PR #953, opened 2016-12-09
Add rule to ban dynamic inline functions

dduan:
When a class method is marked as both `dynamic` and `@inline(__always)`, its
dispatch behavior is not well defined. Therefore this specific combination
should be discouraged.

### DIFF
--- a/Source/SwiftLintFramework/Rules/DynamicInlineRule.swift
+++ b/Source/SwiftLintFramework/Rules/DynamicInlineRule.swift
@@ -1,0 +1,81 @@
+//
+//  DynamicInlineRule.swift
+//  SwiftLint
+//
+//  Created by Daniel Duan on 12/08/16.
+//  Copyright © 2015 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+public struct DynamicInlineRule: ASTRule, ConfigurationProviderRule {
+
+    public var configuration = SeverityConfiguration(.error)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "dynamic_inline",
+        name: "Dynamic Inline",
+        description: "avoid using 'dynamic' and '@inline(__always)' together.",
+        nonTriggeringExamples: [
+            "class C {\ndynamic func f() {}}",
+            "class C {\n@inline(__always) func f() {}}",
+            "class C {\n@inline(never) dynamic func f() {}}"
+        ],
+        triggeringExamples: [
+            "class C {\n@inline(__always) dynamic func f() {}\n}",
+            "class C {\n@inline(__always) public dynamic func f() {}\n}",
+            "class C {\n@inline(__always) dynamic internal func f() {}\n}",
+            "class C {\n@inline(__always)\ndynamic func f() {}\n}",
+            "class C {\n@inline(__always)\ndynamic\nfunc f() {}\n}"
+        ]
+    )
+
+    public func validateFile(_ file: File, kind: SwiftDeclarationKind,
+        dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        // Look for functions with both "inline" and "dynamic". For each of these, we can get offset
+        // of the "func" keyword. We can assume that the nearest "@inline" before this offset is
+        // the attribute we are interested in.
+        guard functionKinds.contains(kind),
+            case let attributes = dictionary.enclosedSwiftAttributes,
+            attributes.contains("source.decl.attribute.dynamic"),
+            attributes.contains("source.decl.attribute.inline"),
+            let funcByteOffset = (dictionary["key.offset"] as? Int64).flatMap({ Int($0) }),
+            let funcOffset = (file.contents as NSString)
+                .byteRangeToNSRange(start: funcByteOffset, length: 0)?
+                .location,
+            case let inlinePattern = regex("@inline"),
+            case let range = NSRange(location: 0, length: funcOffset),
+            let inlineMatch = inlinePattern.matches(in: file.contents, range: range).last,
+            inlineMatch.range.location != NSNotFound,
+            case let attributeRange = NSRange(location: inlineMatch.range.location,
+                length: funcOffset - inlineMatch.range.location),
+            case let alwaysInlinePattern = regex("@inline\\(\\s*__always\\s*\\)"),
+            alwaysInlinePattern.firstMatch(in: file.contents, range: attributeRange) != nil
+        else {
+            return []
+        }
+        return [StyleViolation(ruleDescription: type(of: self).description,
+                    severity: configuration.severity,
+                    location: Location(file: file, characterOffset: funcOffset))]
+    }
+
+    fileprivate let functionKinds: [SwiftDeclarationKind] = [
+        .functionAccessorAddress,
+        .functionAccessorDidset,
+        .functionAccessorGetter,
+        .functionAccessorMutableaddress,
+        .functionAccessorSetter,
+        .functionAccessorWillset,
+        .functionConstructor,
+        .functionDestructor,
+        .functionFree,
+        .functionMethodClass,
+        .functionMethodInstance,
+        .functionMethodStatic,
+        .functionOperator,
+        .functionSubscript
+    ]
+}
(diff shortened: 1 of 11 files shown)


dduan:
I'm not sure what this error on CI means:

```
testSwiftLintLints, failed - Files should have consistent header comments.
  /Users/travis/build/realm/SwiftLint/Source/SwiftLintFramework/Rules/DynamicInlineRule.swift:1
```

aamctustwo:
@dduan The file header comments where you specify copyright and what-not need to be consistent. It's a fairly new rule that was added in PR #904 `Add FileHeaderRule`. I suspect it is the date that needs reformatting and that you should use the © character rather than `(c)`.

marcelofabri:
@dduan Thanks for this. I believe this could use AST instead to be more reliable, as https://github.com/realm/SwiftLint/pull/847 does.

One example:
```bash
$ sourcekitten structure --text "class C { @inline(__always) dynamic func f() {} }"
```
```json
{
  "key.substructure" : [
    {
      "key.bodylength" : 39,
      "key.nameoffset" : 6,
      "key.substructure" : [
        {
          "key.bodylength" : 0,
          "key.nameoffset" : 41,
          "key.offset" : 36,
          "key.length" : 11,
          "key.name" : "f()",
          "key.bodyoffset" : 46,
          "key.attributes" : [
            {
              "key.attribute" : "source.decl.attribute.dynamic"
            },
            {
              "key.attribute" : "source.decl.attribute.inline"
            }
          ],
          "key.kind" : "source.lang.swift.decl.function.method.instance",
          "key.namelength" : 3
        }
      ],
      "key.length" : 49,
      "key.runtime_name" : "_TtC8__main__1C",
      "key.name" : "C",
      "key.bodyoffset" : 9,
      "key.kind" : "source.lang.swift.decl.class",
      "key.offset" : 0,
      "key.namelength" : 1
    }
  ],
  "key.offset" : 0,
  "key.length" : 49
}
```

The only issue is that it doesn't provide the parameter to `@inline`, but you could parse that from the source code itself. What do you think?

dduan:
@marcelofabri there's one more issue with this approach: SourceKit doesn't include function attributes in method's offset. In the example you gave, `key.offset` for the method is here:
```
class C { @inline(__always) dynamic func f() {} }
                                    ^
```

as opposed to the ideal location:

```
class C { @inline(__always) dynamic func f() {} }
          ^
```

In other words, simply inspect function-like kinds such as `source.lang.swift.decl.function.method.instance` is not enough. We have to start with whatever the parent structure is and somehow find the attributes location there. Only then can we start parsing.

I'm not sure the complexity is worth the benefit we gain from this altertive approach.

marcelofabri:
I think we could at least handle `dynamic` with AST, as it would enable using different combinations of the attributes and you could remove the ACLs from the regex.

Source/SwiftLintFramework/Rules/DynamicInlineRule.swift
```diff
@@ -0,0 +1,46 @@
+//
+//  DynamicInlineRule.swift
+//  SwiftLint
+//
+//  Created by Daniel Duan on 2016/12/08.
```
marcelofabri: you'll need to change the date format to `mm/dd/yy`: `12/08/16`

dduan:
@marcelofabri hmm, how would that work? We find a `dynamic` attribute in the AST, then we still need to see if there's an `@incline(__always)` to go along with it, right?

marcelofabri:
You could still use the regex to find the `@inline(__always)` function, then use something like [this](https://github.com/realm/SwiftLint/blob/64f67ccb4ea2380197508ca66d3c6d683239dae3/Source/SwiftLintFramework/Extensions/Structure%2BSwiftLint.swift#L19) to find the function and check if it contains the `dynamic` attribute.

It seems more complicated than it actually is. The advantage is that something like this would work:
```swift
@inline(__always) dynamic private func f()
```

The dynamic and ACL order wouldn't matter.

dduan:
Respectfully, that's not faster, simpler nor more correct.

marcelofabri:
While I agree that it's not faster nor simpler, I'd say it's more correct. The previous example I gave was a bad one - I didn't see that the regex deals with ACLs on both positions.

However, the regex doesn't deal with other attributes. For example:
```swift
class Foo {
    @inline(__always) static dynamic private func g() { }
    @inline(__always) @objc dynamic private func h() { }
}
```

The rule currently isn't triggered on these functions, but it should. I hope this makes sense.

dduan:
@marcelofabri fair enough, let me take another stab at this.

dduan:
@marcelofabri k, here's the new algorithm: we use AST to find functions with both `dynamic` and `inline` attributes. Then we can assume from the beginning of the file to the `func` keyword's offset, the last `@inline` occurrence is associated with this function. We find *it* with regex and then see if it wants `(__always)`.

Source/SwiftLintFramework/Rules/DynamicInlineRule.swift
```diff
@@ -0,0 +1,79 @@
...
+        dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        // Look for functions with both "inline" and "dynamic". For each of these, we can get offset
+        // of the "func" keyword. We can assume that the nearest "@inline" before this offset is
+        // the attribute we are interested in.
+        guard functionKinds.contains(kind),
+            let array = dictionary["key.attributes"] as? [Any],
+            case let attributes = array.flatMap({ ($0 as? [String: String])?["key.attribute"] }),
+            attributes.contains("source.decl.attribute.dynamic"),
+            attributes.contains("source.decl.attribute.inline"),
+            let funcOffset = (dictionary["key.offset"] as? Int64).flatMap({ Int($0) }),
```
jpsim: `funcOffset` is actually in bytes, not in `NSString` offsets that `matches(in:range:)` expects. So this will break if there are multi-byte characters such as emoji in the file's contents before the function.

You can use SourceKitten's [`NSString.location(fromByteOffset:)`](https://github.com/jpsim/SourceKitten/blob/0.15.3/Source/SourceKittenFramework/String+SourceKitten.swift#L104-L131) to convert.

Source/SwiftLintFramework/Rules/DynamicInlineRule.swift
```diff
@@ -0,0 +1,79 @@
...
+    )
+
+    public func validateFile(_ file: File, kind: SwiftDeclarationKind,
+        dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        // Look for functions with both "inline" and "dynamic". For each of these, we can get offset
+        // of the "func" keyword. We can assume that the nearest "@inline" before this offset is
+        // the attribute we are interested in.
+        guard functionKinds.contains(kind),
+            let array = dictionary["key.attributes"] as? [Any],
+            case let attributes = array.flatMap({ ($0 as? [String: String])?["key.attribute"] }),
```
jpsim: There's a private [`extractAttributes(_:)`](https://github.com/realm/SwiftLint/blob/0.13.2/Source/SwiftLintFramework/Rules/OverriddenSuperCallRule.swift#L92-L98) function elsewhere that might be worth putting in a shared location and reusing, since we extract attributes like this in a few rules already (`OverriddenSuperCallRule`, `PrivateOutletRule`, `ValidIBInspectableRule` and `WeakDelegateRule`).

dduan: done (in a separate commit arranged prior to the one for the rule).

Source/SwiftLintFramework/Rules/DynamicInlineRule.swift
dduan: done (had to indirectly use the function you pointed out since it's not public).

jpsim:
I'd like you to know that I learnt two tips from reviewing this PR!

1. Constraining generic extensions on String can't be done, but you can use `ExpressableByStringLiteral` in some cases.
2. If you have a long list of conditional bindings with a non-optional in the middle, you can use `if case let nonOpt = ...`.

dduan:
@jpsim @marcelofabri @aamctustwo Thank you for all the help ❤️

dduan:
@jpsim Ideally the constraint there should be `where Key == String, Value: SourceKitRepresentable`. Neither can be done at current Swift version. Here's hoping the language team and community can get these resolved soon :P

PR #954, opened 2016-12-09
variable_name rule should extend to all identifiers

### LINKED ISSUE #663: variable_name rule should extend to all identifiers
The [Swift API Design Guidelines](https://swift.org/documentation/api-design-guidelines/#follow-case-conventions) state that:

> Names of types and protocols are UpperCamelCase. Everything else is
> lowerCamelCase.

The existing `variable_name` rule only checks that variables are lowerCamelCase, but other identifiers, like functions or methods aren't checked:

``` swift
// No error
func Function(param: Int) -> Int {
    return param
}

// error: Variable Name Violation: Variable name should start with a lowercase character: 'Function' (variable_name)
let Function: Int -> Int = { $0 }
```

I'd propose renaming `variable_name` to `identifier_name` and check all the other identifier types as well.

Related: #660

marcelofabri:
Fixes #663

I haven't applied all the rules to functions because I didn't think it was a good idea as their limits probably should be different.

This is a breaking change, I'm not sure how we should handle this. Perhaps we should keep a `variable_name` rule to avoid breaking configs?


jpsim:

Tests/SwiftLintFrameworkTests/RulesTests.swift
```diff
@@ -397,6 +397,7 @@ extension RulesTests {
             // ("testForceUnwrapping", testForceUnwrapping),
             ("testFunctionBodyLength", testFunctionBodyLength),
             ("testFunctionParameterCountRule", testFunctionParameterCountRule),
+//            ("testIdentifierName", testIdentifierName),
```
marcelofabri: I've commented it because `testVariableName` was commented too

jpsim: That's fine for now. When we merge this, we should add this test to #966.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@
   Guidelines but will shortly.  
   [JP Simard](https://github.com/jpsim),
   [Norio Nomura](https://github.com/norio-nomura)
+  
+* `variable_name` rule is now `identifier_rule` as it validates other
```
jpsim: `identifier_name`. Also, it'd be nice to keep `VariableNameRule` around, logging a warning to the console if used in comment commands or the config file, proxying its implementation to `IdentifierNameRule`. Would lead to a nicer deprecation path.

marcelofabri: But wouldn't it make the rule run twice?

marcelofabri:
I've changed this to WIP as we need to figure out the deprecation part. However, I have no clue why the tests are failing on Linux 😭

marcelofabri:
I was able to reproduce the crash on a local docker instance. It can be reduce to this:

```swift
let name = "matchPattern(_:excludingSyntaxKinds:excludingPattern:exclusionMapping:)"
let nameCharacterSet = CharacterSet(charactersIn: name)
```


Stacktrace:
```
* thread #1: tid = 40, 0x0000000000000000, name = 'repl_swift', stop reason = signal SIGSEGV: invalid address (fault address: 0x0)
  * frame #0: 0x0000000000000000
    frame #1: 0x00007ffff427fa8a libFoundation.so`_CFCharacterSetInitWithCharactersInString + 266
    frame #2: 0x00007ffff43a5736 libFoundation.so`function signature specialization <Arg[0] = Owned To Guaranteed and Exploded> of Foundation.NSCharacterSet.init (charactersIn : Swift.String) -> Foundation.NSCharacterSet + 150
    frame #3: 0x00007ffff43a3ace libFoundation.so`Foundation.NSCharacterSet.__allocating_init (charactersIn : Swift.String) -> Foundation.NSCharacterSet + 78
    frame #4: 0x00007ffff46493e5 libFoundation.so`Foundation.CharacterSet.init (charactersIn : Swift.String) -> Foundation.CharacterSet + 53
    frame #5: 0x00007ffff7f6dddf $__lldb_expr7`main + 127
    frame #6: 0x00000000004009b0 repl_swift`swift_once + 16
```

There are already two tickets on Swift bug tracking: [SR-3215](https://bugs.swift.org/browse/SR-3215) and [SR-3282](https://bugs.swift.org/browse/SR-3282).

It seems that the crash happens with any string with more then 63 characters.

jpsim:
Can you just not include those examples in the rule description when compiling on Linux?

marcelofabri:
Not really, because it's failing to lint a real file in `IntegrationTests.testSwiftLintLints` 😓

marcelofabri:
A note to future self and all others: you can run `lldb` on a binary built with `swift test`:

1. Run docker interactively: `docker run --privileged -it -v `pwd`:/SwiftLint norionomura/sourcekit:302`
2. Run `cd /SwiftLint && swift test` to see it crashing
3. Run `lldb .build/debug/SwiftLintPackageTests.xctest` to start `lldb`
4. In `lldb`, type `r` to start the tests. When they crash, you can get a stack trace with `bt`.

Source/SwiftLintFramework/Extensions/CharacterSet+LinuxHack.swift
```diff
@@ -0,0 +1,36 @@
+//
+//  CharacterSet+LinuxHack.swift
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 17/12/16.
```
jpsim: mm/dd/yy 🔥 🇺🇸

marcelofabri: 🤧

Source/SwiftLintFramework/Extensions/CharacterSet+LinuxHack.swift
```diff
@@ -0,0 +1,36 @@
...
+    }
+}
+
+extension CharacterSet {
+    init(fixedCharactersIn string: String) {
+        #if os(Linux)
+            // workaround for https://bugs.swift.org/browse/SR-3215 and
+            // https://bugs.swift.org/browse/SR-3282
+            let sets = split(string, 63).map { CharacterSet(charactersIn: $0) }
+            let finalSet = sets.reduce(CharacterSet()) { (acc, set) -> CharacterSet in
```
jpsim: This is a minor suggestion, but:

```diff
-let finalSet = sets.reduce(CharacterSet()) { (acc, set) -> CharacterSet in
-    acc.union(set)
-}
-
-self = finalSet
+self = sets.reduce(CharacterSet()) { final, set in final.union(set) }
```

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -178,11 +178,14 @@ extension File {
         return matches.filter { !$0.intersectsRanges(exclusionRanges) }
     }
 
-    internal func validateVariableName(_ dictionary: [String: SourceKitRepresentable],
-                                       kind: SwiftDeclarationKind) -> (name: String, offset: Int)? {
+    internal func validateIdentifierName(_ dictionary: [String: SourceKitRepresentable],
+                                     kind: SwiftDeclarationKind) -> (name: String, offset: Int)? {
```
jpsim: indentation

Source/SwiftLintFramework/Rules/IdentifierNameRule.swift
```diff
@@ -0,0 +1,134 @@
...
+            "private ↓let _i = 0",
+            "enum Foo { case ↓MyEnum }",
+            "↓func IsOperator(name: String) -> Bool"
+        ]
+    )
+
+    private func nameIsViolatingCase(_ name: String) -> Bool {
+        let secondIndex = name.characters.index(after: name.startIndex)
+        let firstCharacter = name.substring(to: secondIndex)
+        if firstCharacter.isUppercase() {
```
jpsim: This might benefit from using `guard`s:

```diff
-if firstCharacter.isUppercase() {
-    if name.characters.count > 1 {
-        let range = secondIndex..<name.characters.index(after: secondIndex)
-        let secondCharacter = name.substring(with: range)
-        return secondCharacter.isLowercase()
-    }
-    return true
-}
-return false
+guard firstCharacter.isUppercase() else { return false }
+guard name.characters.count > 1 else { return true }
+let range = secondIndex..<name.characters.index(after: secondIndex)
+let secondCharacter = name.substring(with: range)
+return secondCharacter.isLowercase()
```

Source/SwiftLintFramework/Rules/IdentifierNameRule.swift
```diff
@@ -0,0 +1,134 @@
...
+                ]
+            }
+
+            return []
+        } ?? []
+    }
+
+    private func isOperator(name: String) -> Bool {
+        let operators = ["/", "=", "-", "+", "!", "*", "|", "^", "~", "?", ".", "%", "<", ">", "&"]
+        return !operators.filter { name.hasPrefix($0) }.isEmpty
```
jpsim: Prefer passing function references over simple closures:

```diff
-return !operators.filter { name.hasPrefix($0) }.isEmpty
+return !operators.filter(name.hasPrefix).isEmpty
```

marcelofabri: Nice catch! 💯

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
marcelofabri: that was probably intentional to avoid breaking the return 😅 
either way, I've moved this to the rule itself, since IMO it's not related to `File` in any obvious way.

Source/SwiftLintFramework/Extensions/CharacterSet+LinuxHack.swift
marcelofabri: 👍

Source/SwiftLintFramework/Rules/IdentifierNameRule.swift
```diff
@@ -0,0 +1,147 @@
...
+
+    private func isOperator(name: String) -> Bool {
+        let operators = ["/", "=", "-", "+", "!", "*", "|", "^", "~", "?", ".", "%", "<", ">", "&"]
+        return !operators.filter(name.hasPrefix).isEmpty
+    }
+
+    private func validateName(_ dictionary: [String: SourceKitRepresentable],
+                              kind: SwiftDeclarationKind) -> (name: String, offset: Int)? {
+        let kinds = SwiftDeclarationKind.variableKinds() +
+            SwiftDeclarationKind.functionKinds() + [.enumelement]
```
marcelofabri: this probably should check the Swift version to add or not `.enumelement`

jpsim:
Maybe this rule shouldn't apply to `@available(*, unavailable, renamed: ...)` declarations since if they're acting as migration helpers from Swift 2 naming to Swift 3, they'll report violations. For example (_as caught by osscheck!_):

```swift

// MARK: - migration support
extension Request {
    @available(*, unavailable, renamed: "editorOpen(file:)")
    public static func EditorOpen(_: File) -> Request { fatalError() }
    ...
}
```

marcelofabri:
> Maybe this rule shouldn't apply to @available(*, unavailable, renamed: ...)

While I agree that it would be useful, I'm not convinced that it's worth the extra complexity. There's no way to get the `unavailable` attribute from AST and the offsets we have don't contain the `@available` attribute.

In both Realm and SourceKitten it'd be pretty easy to just ignore the rule in these declarations since they're isolated inside extensions.

jpsim:
@marcelofabri what's next for this? Is this just blocked on me reviewing? Are there certain high-level points I should consider when reviewing this?

marcelofabri:
If you're cool with not verifying the ` @available(*, unavailable, renamed: ...)` annotation, this is ready for review. 

I don't remember any important points, this is basically renaming the rule and adding more types to it. Also, some violations are skipped for functions. And this also validates that enum cases are lowercase on Swift 3 (I think there was an issue about it somewhere).

CHANGELOG.md
```diff
@@ -939,7 +939,10 @@ This release has seen a phenomenal uptake in community contributions!
 
 ##### Enhancements
 
-* None.
+* `variable_name` rule is now `identifier_name` as it validates other
```
jpsim: This is now in the wrong section

jpsim:
This definitely has a sizable impact on lint times, but that's to be expected. Hopefully with caching, no one feels the pain from this too much...

---

On my MacBook Pro:

Linting Alamofire with this PR took 2.01s vs 1.98s on master **(1% slower)**
Linting swift with this PR took 8.43s vs 6.84s on master **(23% slower)**
Linting Aerial with this PR took 0.31s vs 0.31s on master **(0% slower)**
Linting SourceKitten with this PR took 0.87s vs 0.83s on master **(4% slower)**
Linting Sourcery with this PR took 1.78s vs 1.49s on master **(19% slower)**
Linting ios-oss with this PR took 10.54s vs 10.21s on master **(3% slower)**
Linting Moya with this PR took 0.33s vs 0.29s on master **(13% slower)**
Linting firefox-ios with this PR took 12.21s vs 11.14s on master **(9% slower)**
Linting Nimble with this PR took 1.13s vs 1.14s on master **(0% faster)**
Linting Quick with this PR took 0.4s vs 0.36s on master **(11% slower)**
Linting realm-cocoa with this PR took 1.96s vs 1.7s on master **(15% slower)**
Linting WordPress-iOS with this PR took 6.78s vs 3.3s on master **(105% slower)**

PR #957, opened 2016-12-10
Enabling more rules in SwiftLint itself

marcelofabri:
This enables some recently added opt-in rules in SwiftLint source code.

I've enabled `nimble_operator` and `private_outlet` just as another check against false negatives, as they should never trigger in this project. They've only added `0,012`s combined on my rMBP.

I was thinking about enabling `switch_case_on_newline`  and `conditional_returns_on_newline` too, but there are several violations (23 and 31, respectively) in the current codebase and I'm not sure if we want to enforce them.

I don't this this is worth a changelog entry ¯\_(ツ)_/¯

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -6,6 +6,12 @@ excluded:
 opt_in_rules:
   - empty_count
   - file_header
+  - explicit_init
+  - closure_spacing
+  - overridden_super_call
+  - redundant_nil_coalescing
+  - private_outlet
+  - nimble_operator
 
 file_header:
   required_pattern: |
--- a/Source/SwiftLintFramework/Extensions/File+Cache.swift
+++ b/Source/SwiftLintFramework/Extensions/File+Cache.swift
@@ -31,7 +31,7 @@ private var syntaxKindsByLinesCache = Cache({ file in file.syntaxKindsByLine() }
 private var syntaxTokensByLinesCache = Cache({ file in file.syntaxTokensByLine() })
 
 private typealias AssertHandler = () -> ()
-private var assertHandlers = [String: AssertHandler?]()
+private var assertHandlers = [String: AssertHandler]()
 
 private var _allDeclarationsByType = [String: [String]]()
 private var queueForRebuild = [Structure]()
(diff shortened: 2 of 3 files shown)


.swiftlint.yml
```diff
@@ -6,6 +6,12 @@ excluded:
 opt_in_rules:
   - empty_count
   - file_header
+  - explicit_init
+  - closure_spacing
+  - overridden_super_call
+  - redundant_nil_coalescing
+  - private_outlet
+  - nimble_operator
```
jpsim: What's the rationale in enabling rules that 100% don't apply to SwiftLint?

The `nimble_operator` rule is only useful if you're using Nimble in your project...

Source/SwiftLintFramework/Extensions/File+Cache.swift
```diff
@@ -88,7 +88,7 @@ extension File {
 
     internal var assertHandler: (() -> ())? {
         get {
-            return assertHandlers[cacheKey] ?? nil
+            return assertHandlers[cacheKey]
```
jpsim: Is this change safe? Aren't there cases where we put `nil` as a valid value in `assertHandlers`?

.swiftlint.yml
marcelofabri: I've commented about this on the PR body, but the idea is to add an extra level of integration test, to make sure these rules aren't being trigger where they shouldn't. This applies to the `private_outlet` rule as well.

As they only add 0,012s, I thought it was useful to add this sanity check.

Source/SwiftLintFramework/Extensions/File+Cache.swift
marcelofabri: I don't think so, because I've changed the type of the dictionary to be `[String: AssertHandler]` (it was `[String: AssertHandler?]`). We also only set the value in the setter. 

That `??` was to prevent a double optional, which would be weird.

.swiftlint.yml
jpsim: Ah, yes. That's great. Sorry I didn't understand that when reading the PR comment.

Source/SwiftLintFramework/Extensions/File+Cache.swift
jpsim: Seems fine then.

jpsim:
Thanks!

PR #958, opened 2016-12-11
Add empty_parentheses_with_trailing_closure rule

### LINKED ISSUE #885: Parenthesis in trailing closure rule
When using trailing closures, one can add `()` after the method call, but that should be discouraged.

```swift
let x = [1, 2]

// shouldn't trigger
_ = x.map { $0 + 1 }

// should trigger
_ = x.map() { $0 + 1 }
```

Note that there are cases when parenthesis are allowed (when there're more than 1 parameter):

```swift
// shouldn't trigger
_ = x.reduce(0) { $0 + $1 }
```

marcelofabri:
Fixes #885

### DIFF
--- a/Source/SwiftLintFramework/Rules/EmptyParenthesesWithTrailingClosureRule.swift
+++ b/Source/SwiftLintFramework/Rules/EmptyParenthesesWithTrailingClosureRule.swift
@@ -1,0 +1,84 @@
+//
+//  EmptyParenthesesWithTrailingClosureRule.swift
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 11/12/16.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+public struct EmptyParenthesesWithTrailingClosureRule: ASTRule, ConfigurationProviderRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "empty_parentheses_with_trailing_closure",
+        name: "Empty Parentheses with Trailing Closure",
+        description: "When using trailing closures, empty parenthesis should be avoided " +
+                     "after the method call.",
+        nonTriggeringExamples: [
+            "[1, 2].map { $0 + 1 }\n",
+            "[1, 2].map({ $0 + 1 })\n",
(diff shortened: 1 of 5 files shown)


Source/SwiftLintFramework/Rules/EmptyParenthesesWithTrailingClosureRule.swift
```diff
@@ -0,0 +1,84 @@
...
+        ],
+        triggeringExamples: [
+            "[1, 2].map↓() { $0 + 1 }",
+            "[1, 2].map↓( ) { $0 + 1 }\n",
+            "[1, 2].map↓() { number in\n number + 1 \n}\n",
+            "[1, 2].map↓(  ) { number in\n number + 1 \n}\n"
+        ]
+    )
+
+    public enum Kind: String {
```
marcelofabri: this is the second place where this enum is needed and the 3rd where `"source.lang.swift.expr.call"` is used. Should we create a shared enum? If so, where?

Source/SwiftLintFramework/Rules/EmptyParenthesesWithTrailingClosureRule.swift
```diff
@@ -0,0 +1,84 @@
...
+        guard let offset = (dictionary["key.offset"] as? Int64).flatMap({ Int($0) }),
+            let length = (dictionary["key.length"] as? Int64).flatMap({ Int($0) }),
+            let nameOffset = (dictionary["key.nameoffset"] as? Int64).flatMap({ Int($0) }),
+            let nameLength = (dictionary["key.namelength"] as? Int64).flatMap({ Int($0) }),
+            let bodyLength = (dictionary["key.bodylength"] as? Int64).flatMap({ Int($0) }),
+            bodyLength > 0 else {
+                return []
+        }
+
+        let rangeStart = nameOffset + nameLength
```
marcelofabri: using bodyOffset and bodyLength were giving incorrect ranges ¯\_(ツ)_/¯

Source/SwiftLintFramework/Rules/EmptyParenthesesWithTrailingClosureRule.swift
jpsim: Yes! This should live in SourceKittenFramework though.

marcelofabri:
🎉 🚢

PR #960, opened 2016-12-11
Add closure_parameter_position rule

### LINKED ISSUE #931: Closure parameters on the same line as opening brace
Closure parameters should be on the same line as opening brace.

```swift
// triggers
[1, 2].map {
    number in
    return number + 1
}

// ok
[1, 2].map { number in
    return number + 1
}
```

Real world example: https://github.com/realm/SwiftLint/pull/916#discussion_r90507668

marcelofabri:
Fixes #931


jpsim:

Source/SwiftLintFramework/Extensions/SwiftExpressionCallKind.swift
```diff
@@ -0,0 +1,23 @@
...
+//  SwiftExpressionCallKind.swift
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 11/12/16.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import Foundation
+
+public enum SwiftExpressionCallKind: String {
```
jpsim: Can we make this `internal` since we'll probably want to move to SourceKitten in the near future?

FYI, all the "Expression" kinds are listed here: https://github.com/apple/swift/blob/swift-3.0.1-RELEASE/tools/SourceKit/lib/SwiftLang/SwiftLangSupport.cpp#L143-L147

At this moment, that's `call`, `argument`, `array`, `dictionary` and `object_literal`.

Since we already use some of those in existing SwiftLint rules, it'd be nice to rename this `SwiftExpressionKind`, update the other rules to use this, and move to SourceKittenFramework.

marcelofabri: I think we'd need to rethink our implementations if we did that. They only work because of the `other` case, as these kinds can be nested in other kinds (`struct` for example). If we added the other cases and removed the `other` one, some rules could start failing.

For example, `trailing_comma` in this example: 
```swift
struct Bar {
  let foo = [1: 2, 2: 3, ]
},
```

This enum also can't be `internal` because it's used as `associatedtype` on `ASTRule`s, in functions that are `public`.

jpsim: It's fine to keep `public`, especially since SwiftLint is pre-1.0 (_still_ 😅). But this really _does_ belong in SourceKittenFramework (as a `public` API, of course), and we certainly _should_ move it there before SwiftLint gets to 1.0 with a CLI and framework interface we want to maintain following SemVer.

To be clear, you're just opposed to making this `internal` at the moment, right? We can still rename to `SwiftExpressionKind` and add the other members?

marcelofabri: We could rename and add the other members, but we either:
a) will have to keep the `other` case to keep the rules working
b) change [validateFile](https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Protocols/ASTRule.swift#L22-L32) implementation to look into the subitems no matter the `KindType`.

jpsim: we can keep the `other` case.

marcelofabri: I think that is fine by now, but wouldn't it be weird once this enum is on `SourceKittenFramework`?

jpsim: yes, but that refactoring can happen later.

jpsim: IMHO, `other` should be handled by a failing `SwiftExpressionKind(rawValue:)` initialization.

PR #963, opened 2016-12-11
Add Linux Support

jpsim:


marcelofabri:
Maybe we should add a note about testing on Linux on https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#rules?

marcelofabri:
@jpsim I think we should add an explanation about the `allTests` and `XCTMain` stuff, as people will probably forget to update those as it's not usual on a mac OS environment.

marcelofabri:
This is awesome BTW 💯🎉😍

Source/SwiftLintFramework/Rules/RuleConfigurations/PrivateUnitTestConfiguration.swift
```diff
@@ -13,8 +13,8 @@ public struct PrivateUnitTestConfiguration: RuleConfiguration, Equatable {
     public let identifier: String
     public var name: String?
     public var message = "Regex matched."
-    public var regex = NSRegularExpression()
-    public var included = NSRegularExpression()
+    public var regex = NSRegularExpression.forcePattern("placeholder")
```
marcelofabri: maybe make them optional? 😬

jpsim: Seems obvious, but I couldn't figure it out earlier. Fixed now.

jpsim:
YOLO

PR #965, opened 2016-12-12
Add `empty_parameters` rule

### LINKED ISSUE #573: Add rule prefer `() -> ` over `Void -> `
Related to #570, we talked about another this.
In apple/swift, we can't find `Void ->`.
#570 can also be avoided by this.

marcelofabri:
Fixes #573

I'm not a big fan of this rule name though 😅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#964](https://github.com/realm/SwiftLint/issues/964)
 
+* Add `empty_parameters` rule to validate usage of `() -> ` over `Void -> `.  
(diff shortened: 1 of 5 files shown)


jpsim:
`VoidParametersRule`?

marcelofabri:
I thought about that, but should rules describe what they catch or what it's the "right way"? 💭

jpsim:
I actually like the "Empty Parameters" rule name.

marcelofabri:
Let's keep then! I'll rebase the changelog

jpsim:
I'll mark this as `[WIP]` until you can adapt to ignore "middle types" as discussed in https://github.com/realm/SwiftLint/pull/967#discussion_r91881943

jpsim:
🎉

PR #967, opened 2016-12-12
Void return rule

### LINKED ISSUE #964: Rule request: prefer `-> Void` over `-> ()`
Similar to #573

marcelofabri:
Fixes #964

### DIFF
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -161,9 +161,12 @@ extension File {
         }.map { $0.0 }
     }
 
+    internal typealias MatchMapping = (NSTextCheckingResult) -> NSRange
+
     internal func matchPattern(_ pattern: String,
                                excludingSyntaxKinds: [SyntaxKind],
-                               excludingPattern: String) -> [NSRange] {
+                               excludingPattern: String,
+                               exclusionMapping: MatchMapping = { $0.range }) -> [NSRange] {
         let contents = self.contents.bridge()
         let range = NSRange(location: 0, length: contents.length)
         let matches = matchPattern(pattern, excludingSyntaxKinds: excludingSyntaxKinds)
@@ -171,7 +174,7 @@ extension File {
             return []
         }
         let exclusionRanges = regex(excludingPattern).matches(in: self.contents, options: [],
-                                                              range: range).ranges()
+                                                              range: range).map(exclusionMapping)
         return matches.filter { !$0.intersectsRanges(exclusionRanges) }
     }
 
--- a/Source/SwiftLintFramework/Rules/VoidReturnRule.swift
+++ b/Source/SwiftLintFramework/Rules/VoidReturnRule.swift
@@ -21,25 +21,26 @@ public struct VoidReturnRule: Rule, ConfigurationProviderRule {
         nonTriggeringExamples: [
             "let abc: () -> Void = {}\n",
             "func foo(completion: () -> Void)\n",
-            "func foo: (ConfigurationTests) -> () throws -> Void)\n",
-            "func foo: (ConfigurationTests) ->   () throws -> Void)\n",
-            "func foo: (ConfigurationTests) ->() throws -> Void)\n",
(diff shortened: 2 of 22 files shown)


Tests/SwiftLintFrameworkTests/ConfigurationTests.swift
```diff
@@ -266,6 +266,7 @@ extension XCTestCase {
 }
 
 extension ConfigurationTests {
+    // swiftlint:disable:next void_return
     static var allTests: [(String, (ConfigurationTests) -> () throws -> Void)] {
```
marcelofabri: I'm not sure we can handle this case, so I've disabled the rule. I could change to `(ConfigurationTests) -> Void throws -> Void`, but then https://github.com/realm/SwiftLint/pull/965 would trigger

jpsim:
Has the community settled on a style preference here? I'd love to see the discussion around that.

marcelofabri:
There's a post by Erica Saud (http://ericasadun.com/2015/05/11/swift-vs-void/) in which see links a post on devforums (https://devforums.apple.com/message/1133616#1133616) where an Apple employee says:

> FWIW, we've recently decided to standardize on () -> Void (generally, () for parameters and Void for return types) across all of our documentation.

Tests/SwiftLintFrameworkTests/ConfigurationTests.swift
jpsim: I think this rule should be opt-in until we can address this case.

jpsim:
Thanks for finding that thread, very useful.

Tests/SwiftLintFrameworkTests/ConfigurationTests.swift
marcelofabri: I don't think it'd be too hard to address this case, it'd be more a question of choosing the expected style for the type in the middle of a curried declaration. I think the current declaration would be ok, what do you think?

jpsim: I think types in the middle could be either `()` or `Void`, and to adapt both `void_return` and `empty_parameters` would just ignore types in the middle.

marcelofabri: Seems good to me!

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -161,17 +161,20 @@ extension File {
         }.map { $0.0 }
     }
 
+    internal typealias MatchMapping = (NSTextCheckingResult) -> NSRange
```
jpsim: Why did you add this, the `exclusionMapping` parameter and replace the `ranges()` call later? Seems like the previous code does the exact same thing?

marcelofabri: Because I needed to get the capture group (`$0.rangeAt(1)`), and not the full match. I'll comment on the relevant line.

I've created the typealias just to save space to avoid an ugly formatting.

Source/SwiftLintFramework/Rules/VoidReturnRule.swift
```diff
@@ -0,0 +1,56 @@
...
+        ]
+    )
+
+    public func validateFile(_ file: File) -> [StyleViolation] {
+        let kinds = SyntaxKind.commentAndStringKinds()
+        let pattern = "->\\s*\\(\\s*\\)\\s*(?!->)"
+        let excludingPattern = "(\(pattern))\\s*(throws\\s+)?->"
+
+        return file.matchPattern(pattern, excludingSyntaxKinds: kinds,
+                                 excludingPattern: excludingPattern) { $0.rangeAt(1) }.flatMap {
```
marcelofabri: here's where I use the capture group

jpsim: oh, I see! I searched for the parameter name and didn't see it invoked anywhere, but that's because you're using trailing closure syntax. 👍

jpsim:
💃

PR #971, opened 2016-12-12
Add ProhibitedSuperRule

### LINKED ISSUE #970: Rule: Methods calling super that should not
Some of Apple's methods should not call super when overriden, according to their documentation. These methods should be flagged. Example:

```swift
class foo: UIViewController {
  override func loadView() {
    super.loadView() // <- Should be flagged
  }
}
```

aamctustwo:
Resolves #970: `Rule: Methods calling super that should not`.

### Summary

Adds an opt-in `ProhibitedSuperRule` that catches:
- [`NSFileProviderExtension.providePlaceholder(at:completionHandler:)`](https://developer.apple.com/reference/uikit/nsfileproviderextension/1623483-provideplaceholder)
- [`NSTextInput.doCommand(by:)`](https://developer.apple.com/reference/appkit/nstextinputclient/1438256-docommand)
- [`NSView.updateLayer()`](https://developer.apple.com/reference/appkit/nsview/1483580-updatelayer)
- [`UIViewController.loadView()`](https://developer.apple.com/reference/uikit/uiviewcontroller/1621454-loadview)

Rule is based on [`OverriddenSuperCallRule`](https://github.com/realm/SwiftLint/blob/0.13.2/Source/SwiftLintFramework/Rules/OverriddenSuperCallRule.swift).

### DIFF
--- a/Source/SwiftLintFramework/Rules/ProhibitedSuperRule.swift
+++ b/Source/SwiftLintFramework/Rules/ProhibitedSuperRule.swift
@@ -55,15 +55,13 @@ public struct ProhibitedSuperRule: ConfigurationProviderRule, ASTRule, OptInRule
                              kind: SwiftDeclarationKind,
                              dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
         guard let offset = dictionary["key.bodyoffset"] as? Int64,
-            let name = dictionary["key.name"] as? String
-            else { return [] }
-
-        let substructure = (dictionary["key.substructure"] as? [SourceKitRepresentable]) ?? []
-        guard kind == .functionMethodInstance &&
+            let name = dictionary["key.name"] as? String,
+            kind == .functionMethodInstance &&
             configuration.resolvedMethodNames.contains(name) &&
             dictionary.enclosedSwiftAttributes.contains("source.decl.attribute.override")
             else { return [] }
 
+        let substructure = (dictionary["key.substructure"] as? [SourceKitRepresentable]) ?? []
         let callsToSuper = extractCallsToSuper(name, substructure: substructure)
 
         if !callsToSuper.isEmpty {
--- a/Source/SwiftLintFramework/Rules/ProhibitedSuperRule.swift
+++ b/Source/SwiftLintFramework/Rules/ProhibitedSuperRule.swift
@@ -1,0 +1,90 @@
+//
+//  ProhibitedSuperRule.swift
+//  SwiftLint
+//
+//  Created by Aaron McTavish on 12/12/16.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import SourceKittenFramework
+
+public struct ProhibitedSuperRule: ConfigurationProviderRule, ASTRule, OptInRule {
+    public var configuration = ProhibitedSuperConfiguration()
+
+    public init() { }
+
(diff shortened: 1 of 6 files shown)


Source/SwiftLintFramework/Rules/ProhibitedSuperRule.swift
```diff
@@ -0,0 +1,90 @@
...
+    )
+
+    public func validateFile(_ file: File,
+                             kind: SwiftDeclarationKind,
+                             dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard let offset = dictionary["key.bodyoffset"] as? Int64,
+            let name = dictionary["key.name"] as? String
+            else { return [] }
+
+        let substructure = (dictionary["key.substructure"] as? [SourceKitRepresentable]) ?? []
```
jpsim: could move this variable declaration down, and merge both `guard` groups together

Source/SwiftLintFramework/Rules/ProhibitedSuperRule.swift
```diff
@@ -0,0 +1,90 @@
...
+//  ProhibitedSuperRule.swift
+//  SwiftLint
+//
+//  Created by Aaron McTavish on 12/12/16.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import SourceKittenFramework
+
+public struct ProhibitedSuperRule: ConfigurationProviderRule, ASTRule, OptInRule {
```
jpsim: why should this be an opt-in rule? seems like it should be enabled by default...

aamctustwo: I had considered it, but had decided to make it opt-in as the rule is based on method signature and some of the signatures are not particularly unique. Without adding additional checks to see if the class is a descendent of the correct object, I was worried there would likely be some false positives. I thought that might be a good addition in a future PR.

But if you feel that those developers should either change the names of their methods or mark it as excluded in the configuration, I'm happy to make it opt-in. :smile_cat:

jpsim: Yeah actually, I wrote this before realizing that this rule didn't check for superclass inheritance to make sure that only the correct methods were being flagged, so opt-in is probably best. 👌

Source/SwiftLintFramework/Rules/ProhibitedSuperRule.swift
```diff
@@ -55,15 +55,13 @@ public struct ProhibitedSuperRule: ConfigurationProviderRule, ASTRule, OptInRule
...
-
-        let substructure = (dictionary["key.substructure"] as? [SourceKitRepresentable]) ?? []
-        guard kind == .functionMethodInstance &&
+            let name = dictionary["key.name"] as? String,
+            kind == .functionMethodInstance &&
             configuration.resolvedMethodNames.contains(name) &&
             dictionary.enclosedSwiftAttributes.contains("source.decl.attribute.override")
             else { return [] }
 
+        let substructure = (dictionary["key.substructure"] as? [SourceKitRepresentable]) ?? []
```
jpsim: this could also be folded into the `guard` statement above since it falls back on `[]` if the conditional cast fails, which causes to return `[]` anyway

PR #969, opened 2016-12-12
Fix XML Reporters not Escaping Characters

fabb:
Fixes #968.

PR #979, opened 2016-12-13
Fix XML Reporters not Escaping Characters

jpsim:
Continued from #969.


jpsim:
Travis is massively backed up and I've confirmed that all tests pass locally, so I'm merging this.

PR #986, opened 2016-12-15
Fix invalid CSS in HTML reporter

### LINKED ISSUE #981: HTML Reporter - Invalid CSS Styling
At the moment, the HTML report has an invalid CSS style within it. It is attempting to defining a background color in the table header to have an alpha value. This is the [currently offending line](https://github.com/realm/SwiftLint/blob/b36ca2f071d55df7af8152b0ab0717696801de81/Source/SwiftLintFramework/Reporters/HTMLReporter.swift#L63).

Currently, CSS color properties do not support defining an RGBA color from an 8 hexidecimal characters value (only using the `rgba` functions). See [CSS Color Module Level 3 Recommendation](https://www.w3.org/TR/css3-color/#rgba-color).

Defining RGBA colors from hex is tentatively slated for the future. See [CSS Color Module Level 4 Editor's Draft](https://drafts.csswg.org/css-color/#hex-notation).

We should either use a non-transparent color there or we should use the `rgba` function to generate the desired color.

aamctustwo:
Resolves #981 `HTML Reporter - Invalid CSS Styling`.

### Summary

Updates the HTML reporter template to use the `rgba` method for defining the semi-transparent table header color in the CSS Styling. The color itself has not been changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,10 @@
   [JP Simard](https://github.com/jpsim)
   [#976](https://github.com/realm/SwiftLint/issues/976)
 
+* Fix invalid CSS in HTML reporter template.  
+  [Aaron McTavish](https://github.com/aamctustwo)
+  [#981](https://github.com/realm/SwiftLint/issues/981)
+
 ## 0.13.2: Light Cycle
 
 ##### Breaking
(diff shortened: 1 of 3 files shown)


marcelofabri:
Thanks for doing this 🚀

PR #985, opened 2016-12-15
make VoidReturnRule correctable

jpsim:
/cc @marcelofabri

PR #988, opened 2016-12-15
Making empty_parameters correctable

marcelofabri:
Totally based on #985 😊

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,7 +88,8 @@
   [JP Simard](https://github.com/jpsim)
   [#964](https://github.com/realm/SwiftLint/issues/964)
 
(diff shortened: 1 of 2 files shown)


jpsim:
> Totally based on #985 😊

👍. As intended! I was about to suggest to you that #985 could be used to do this, but you're too fast!

These two PRs has reminded me that we could 1) reuse some of the code used in correctable rules and 2) make them easier to write. But that's not pressing.

marcelofabri:
yeah, I think it's even worse on `ASTRule`s. I was thinking about creating a `CorrectableASTRule` protocol to reuse some code and try to apply that on `EmptyParenthesesWithTrailingClosureRule` which should be an easy one to implement.

PR #990, opened 2016-12-15
Make operator_usage_whitespace correctable

marcelofabri:


Source/SwiftLintFramework/Rules/OperatorUsageWhitespaceRule.swift
```diff
@@ -78,10 +100,61 @@ public struct OperatorUsageWhitespaceRule: OptInRule, ConfigurationProviderRule
...
+                .trimmingCharacters(in: .whitespaces)
+            let correction = " " + operatorContent + " "
+
+            return (range, correction)
+        }
+    }
+
+    public func correctFile(_ file: File) -> [Correction] {
+        let violatingRanges = violationRanges(file: file).filter { range, _ in
+            return !file.ruleEnabledViolatingRanges([range], forRule: self).isEmpty
```
jpsim: why is `ruleEnabledViolatingRanges` called for every violation range? Couldn't you just do this?

```swift
let violatingRanges = file.ruleEnabledViolatingRanges(violationRanges(file: file), forRule: self)
```

Source/SwiftLintFramework/Rules/OperatorUsageWhitespaceRule.swift
```diff
@@ -78,10 +100,61 @@ public struct OperatorUsageWhitespaceRule: OptInRule, ConfigurationProviderRule
...
+
+    private func writeToFile(_ file: File, violatingRanges: [(NSRange, String)]) -> [Correction] {
+        var correctedContents = file.contents
+        var adjustedLocations = [Int]()
+
+        for (violatingRange, correction) in violatingRanges.reversed() {
+            if let indexRange = correctedContents.nsrangeToIndexRange(violatingRange) {
+                correctedContents = correctedContents
+                    .replacingCharacters(in: indexRange, with: correction)
+                adjustedLocations.insert(violatingRange.location, at: 0)
```
jpsim: Just an FYI, this logic is incorrect. I know it's what I used in yesterday's PR making void return correctable, but this won't successfully adjust the correction locations as characters are removed/added in the file.

It's not a _huge_ deal because we just report these locations in the `swiftlint autocorrect` output log and they're not rendered in Xcode or anything, which is why I was too lazy to do this correctly in my last PR.

marcelofabri: I didn't know that the locations should be the updated ones, thanks to let me know 😅

jpsim:
Sorry for the delay.

Issue #900, opened 2016-11-25
Rule Request: Sorted import statements


jpsim:
I like having my import statements sorted alphabetically for consistency. Not sure if I'm just a weirdo, or if others would benefit from this rule. But it'd be pretty easy to write.

marcelofabri:
Closed in #1071.

PR #991, opened 2016-12-15
Add rule requiring imports be sorted

sberrevoets:
Adds an opt-in rule to requiring imports to be sorted, as requested in #900.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1005](https://github.com/realm/SwiftLint/issues/1005)
 
+* Add a rule that enforces alphabetical sorting of imports.  
+  [Scott Berrevoets](https://github.com/sberrevoets)
+  [#900](https://github.com/realm/SwiftLint/issues/900)
+
 ##### Bug Fixes
 
 * `FunctionParameterCountRule` also ignores generic initializers.  
@@ -79,10 +83,6 @@
   [Norio Nomura](https://github.com/norio-nomura)
 
 ##### Enhancements
-
-* Add a rule that enforces alphabetical sorting of imports.  
-  [Scott Berrevoets](https://github.com/sberrevoets)
-  [#900](https://github.com/realm/SwiftLint/issues/900)
 
 * Now builds and passes most tests on Linux using the Swift Package Manager with
   Swift 3. This requires `libsourcekitdInProc.so` to be built and located in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,10 @@
 
 ##### Enhancements
 
+* Add a rule that enforces alphabetical sorting of imports.  
+  [Scott Berrevoets](https://github.com/sberrevoets)
+  [#991](https://github.com/realm/SwiftLint/pull/991)
+
 * Now builds and passes most tests on Linux using the Swift Package Manager with
   Swift 3. This requires `libsourcekitdInProc.so` to be built and located in
   `/usr/lib`, or in another location specified by the `LINUX_SOURCEKIT_LIB_PATH`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,7 @@
 * Add a rule that enforces alphabetical sorting of imports.  
   [Scott Berrevoets](https://github.com/sberrevoets)
   [#991](https://github.com/realm/SwiftLint/pull/991)
+  [#900](https://github.com/realm/SwiftLint/issues/900)
 
 * Now builds and passes most tests on Linux using the Swift Package Manager with
   Swift 3. This requires `libsourcekitdInProc.so` to be built and located in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,7 +82,6 @@
 
 * Add a rule that enforces alphabetical sorting of imports.  
   [Scott Berrevoets](https://github.com/sberrevoets)
-  [#991](https://github.com/realm/SwiftLint/pull/991)
   [#900](https://github.com/realm/SwiftLint/issues/900)
 
 * Now builds and passes most tests on Linux using the Swift Package Manager with
--- a/Source/SwiftLintFramework/Rules/SortedImportsRule.swift
+++ b/Source/SwiftLintFramework/Rules/SortedImportsRule.swift
@@ -27,31 +27,29 @@ public struct SortedImportsRule: ConfigurationProviderRule, OptInRule {
     )
 
     public func validateFile(_ file: File) -> [StyleViolation] {
-        let pattern = "import\\s+\\w+"
+        let importRanges = file.matchPattern("import\\s+\\w+",
+                                             withSyntaxKinds: [.keyword, .identifier])
+        let contents = file.contents.bridge()
 
-        var previousMatch = ""
-        return file.matchPattern(pattern).flatMap { range, kinds in
-            if kinds.count != 2 || kinds[0] != .keyword || kinds[1] != .identifier {
-                return nil
-            }
+        let importLength = 6
+        let modulesAndOffsets: [(String, Int)] = importRanges.map { range in
+            let moduleRange = NSRange(location: range.location + importLength,
+                                      length: range.length - importLength)
+            let moduleName = contents.substring(with: moduleRange)
+                .trimmingCharacters(in: .whitespacesAndNewlines)
+            let offset = NSMaxRange(range) - moduleName.bridge().length
+            return (moduleName, offset)
+        }
 
-            let fullMatch = file.contents.bridge().substring(with: range)
-            let moduleStartIndex = (fullMatch.lastIndexOf(" ") ?? -1) + 1
-            let moduleLength = range.length - moduleStartIndex
-            let moduleRange = NSRange(location: moduleStartIndex, length: moduleLength)
-            let moduleNameMatch = fullMatch.bridge().substring(with: moduleRange)
+        let modulePairs = zip(modulesAndOffsets, modulesAndOffsets.dropFirst())
+        let violatingOffsets = modulePairs.flatMap { previous, current in
+            return current < previous ? current.1 : nil
+        }
 
-            defer { previousMatch = moduleNameMatch }
-
-            if moduleNameMatch > previousMatch {
-                return nil
-            }
-
-            let characterOffset = range.location + moduleStartIndex
-            let location = Location(file: file, characterOffset: characterOffset)
-            return StyleViolation(ruleDescription: type(of: self).description,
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
 
 ##### Enhancements
 
+* Add a rule that enforces alphabetical sorting of imports.
```
marcelofabri: missing the 2 trailing whitespaces

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
...
+  [Scott Berrevoets](https://github.com/sberrevoets)
```
marcelofabri: can you add the issue number too?

aamctustwo:
@sberrevoets 🤔 Maybe it could treat comment lines as separators to allow you to group the import statements and then alphabetise within each group? This is the most common use case I've seen for not having them all alphabetised at the file level. Or maybe that's a configurable option?

sberrevoets:
To be honest, I've never seen that style. I've seen groupings within imports before (by having a blank line between groups), but that was mostly in the Objective-C world where every individual file had to be imported, and still at the top of the file.

If we were to go down that path I think I'd use a blank line as group separator, or even any line that doesn't start with `import`.

jpsim:
> It's hard to determine exactly which import is in violation if the list is not sorted. For example, in this list:
>
> ```swift
> import ZZZ
> import AAA
> import BBB
> import CCC
> ```

Why not just report violations for items that aren't "greater" than the ones before it?

It would generate these violations:

```swift
import ZZZ
↓import AAA
import BBB
import CCC
```

```swift
import AAA
import BBB
import ZZZ
↓import CCC
```

```swift
import BBB
↓import AAA
import ZZZ
↓import CCC
```

Which IMHO is generally more useful than the current algorithm.

sberrevoets:
That would work, though in your first example marking `AAA` would still be a little weird since really it's `ZZZ` that's out of order. But I agree that it would still be more useful than the current implementation, so I'll change the code to behave like that.

sberrevoets:
@jpsim Updated to mark violating lines the way you suggested.

@marcelofabri I noticed you changed the example I based mine on so I changed this PR to use `characterOffset` as well.

marcelofabri:
@sberrevoets Thanks for updating this. I've also opened https://github.com/realm/SwiftLint/pull/997 to prevent us to do merge something like that again 😅

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
 
 ##### Enhancements
 
+* Add a rule that enforces alphabetical sorting of imports.  
+  [Scott Berrevoets](https://github.com/sberrevoets)
+  [#991](https://github.com/realm/SwiftLint/pull/991)
```
jpsim: reference the issue number this solves (#900): https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes

Source/SwiftLintFramework/Rules/SortedImportsRule.swift
```diff
@@ -0,0 +1,51 @@
...
+        ],
+        triggeringExamples: [
+            "import AAA\nimport ZZZ\nimport BBB\nimport CCC",
+            "import AAA\nimport ZZZ\nimport BBB\nimport CCC\nimport AAA"
+        ]
+    )
+
+    public func validateFile(_ file: File) -> [StyleViolation] {
+        let pattern = "import\\s+(\\w+)"
+        let excludingKinds = SyntaxKind.commentAndStringKinds()
```
jpsim: instead of _excluding_ comments and strings, you should be matching a single keyword followed by a single identifier, as highlighted by getting the syntax information from imports:

```shell
$ sourcekitten syntax --text "import AAA"                
[
  {
    "type" : "source.lang.swift.syntaxtype.keyword",
    "offset" : 0,
    "length" : 6
  },
  {
    "type" : "source.lang.swift.syntaxtype.identifier",
    "offset" : 7,
    "length" : 3
  }
]
```

Source/SwiftLintFramework/Rules/SortedImportsRule.swift
```diff
@@ -0,0 +1,51 @@
...
+
+    public static let description = RuleDescription(
+        identifier: "sorted_imports",
+        name: "Sorted Imports",
+        description: "Imports should be sorted.",
+        nonTriggeringExamples: [
+            "import AAA\nimport BBB\nimport CCC\nimport DDD"
+        ],
+        triggeringExamples: [
+            "import AAA\nimport ZZZ\nimport BBB\nimport CCC",
```
jpsim: You should be inserting the `↓` character where you expect the violation to be reported.

Source/SwiftLintFramework/Rules/SortedImportsRule.swift
marcelofabri: There's a similar check in https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules/AttributesRule.swift#L67-L72, it you need inspiration ;)

Tests/SwiftLintFrameworkTests/RulesTests.swift
```diff
@@ -263,6 +263,10 @@ class RulesTests: XCTestCase {
         verifyRule(ReturnArrowWhitespaceRule.description)
     }
 
+    func testSortedImports() {
```
marcelofabri: you should add this to `allTests` in the end of this file to make it run on Linux as well

sberrevoets:
@jpsim @marcelofabri Thanks for the patience and guiding me through! I've incorporated your comments in the latest updates.

Source/SwiftLintFramework/Rules/SortedImportsRule.swift
```diff
@@ -0,0 +1,53 @@
...
+        nonTriggeringExamples: [
+            "import AAA\nimport BBB\nimport CCC\nimport DDD"
+        ],
+        triggeringExamples: [
+            "import AAA\nimport ZZZ\nimport ↓BBB\nimport CCC"
+        ]
+    )
+
+    public func validateFile(_ file: File) -> [StyleViolation] {
+        let pattern = "import (\\w+)"
```
marcelofabri: Sorry to nitpick, but could you change the regex to `import\\s+\\w+`, since the rule shouldn't assume that there's only one space? Also, the capture group doesn't seem to be used, so it can be remove to make the regex faster.

This would probably make the offset calculation harder, but IMO we can trigger in the `import` itself (i.e. the regex match itself). What do you think?

marcelofabri: Actually, one the current implementation would give incorrect results if you had one import with multiple spaces and another one with just one, since the whole match is used for comparison.

sberrevoets: The previous implementation allowed for any number of spaces, but for exactly the reason you pointed out I decided to stick with 1 since I don't see why someone would (intentionally) use more than 1 space.

However, I like the idea of the module name triggering the warning, so I've pushed a commit that adds some complexity in finding the module name. It assumes the last character before it is a space, which means `import\nMyModule` will fail, but IMO that's an acceptable compromise. If you think we should just trigger on `import` I can live with that as well.

I've removed the capture group from the regex as well.

Source/SwiftLintFramework/Rules/SortedImportsRule.swift
```diff
@@ -0,0 +1,58 @@
...
+            let moduleRange = NSRange(location: moduleStartIndex, length: moduleLength)
+            let moduleNameMatch = fullMatch.bridge().substring(with: moduleRange)
+
+            defer { previousMatch = moduleNameMatch }
+
+            if moduleNameMatch > previousMatch {
+                return nil
+            }
+
+            let characterOffset = range.location + moduleCharacterOffset
```
marcelofabri: shouldn't this use `moduleStartIndex` as well? Then you can remove `moduleCharacterOffset` as well

Source/SwiftLintFramework/Rules/SortedImportsRule.swift
```diff
@@ -0,0 +1,57 @@
...
+        nonTriggeringExamples: [
+            "import AAA\nimport BBB\nimport CCC\nimport DDD"
+        ],
+        triggeringExamples: [
+            "import AAA\nimport ZZZ\nimport ↓BBB\nimport CCC"
+        ]
+    )
+
+    public func validateFile(_ file: File) -> [StyleViolation] {
+        let pattern = "import\\s+\\w+"
```
jpsim: What do you think of this implementation instead?

```swift
let importRanges = file.matchPattern("import\\s+\\w+",
                                     withSyntaxKinds: [.keyword, .identifier])
let nsstringContents = file.contents.bridge()
let importLength = 6
let modulesAndOffsets: [(String, Int)] = importRanges.map { range in
    let rangeAfterImport = NSRange(location: range.location + importLength,
        length: range.length - importLength)
    let module = nsstringContents.substring(with: rangeAfterImport)
        .trimmingCharacters(in: .whitespacesAndNewlines)
    let offset = NSMaxRange(range) - module.bridge().length
    return (module, offset)
}
let modulePairs = zip(modulesAndOffsets, modulesAndOffsets.dropFirst())
let violatingOffsets = modulePairs.flatMap { previous, current in
    return current < previous ? current.1 : nil
}
return violatingOffsets.map {
    StyleViolation(ruleDescription: type(of: self).description,
                          location: Location(file: file, characterOffset: $0))
}
```

It avoids the `flatMap`/`previousMatch` side-effect mutation, avoids some of the bridging operations, leverages the existing `matchPattern(_:withSyntaxKind:)` method and avoids the cutsey use of `defer`.

sberrevoets: To be completely honest, for me personally the hit on readability wouldn't be worth the benefits in this case:

- Not needing `previousMatch` is definitely nice, no argument there
- The bridging is reduced to once for every `import` - which is a decent win but probably not very noticeable considering the matches are short and there aren't many of them
- I wasn't aware of `matchPattern(_:withSyntaxKind:)`, but it appears that would even work with the current implementation (which I agree would definitely be nicer!)
- I personally don't have a huge problem with the `defer`, but I can see how that can require a double-take

That said, I'm totally onboard with following the general coding practices of the existing codebase (which your proposal seems more in line with) so I've pushed another commit with that implementation.

jpsim: I actually agree with all your points, I didn't share that sample thinking it was necessarily superior, just had some different ideas and I wanted to spark a discussion.

marcelofabri:
Merged in #1071. Thanks again for doing this, @sberrevoets! 💯

PR #992, opened 2016-12-15
Add unused_closure_parameter rule

### LINKED ISSUE #982: Unused parameter in closure should be replaced with _
Unused parameter in a closure should be replaced with `_`:

```swift
// should trigger
[1, 2].map { number in
    return 3
}

// ok
[1, 2].map { _ in
    return 3
}

[1, 2].map { number in
    return number + 1
}

[1, 2].map { $0 + 1 }
```

marcelofabri:
Fixes #982

This currently doesn't catch some violations:

```swift
// 1. If there's a variable with the same name in another object
[1, 2].map { number in
  return otherThing.number
}

// 2. If there're nested closures and the inner one shadows a variable from the other one
[1, 2].map { number in
  return [1, 2].map { number in
    return number + 1
  }
}
```

1 seems to be possible to be fixed (I just don't have any good ideas). The other one though I'd say is much harder.

Maybe this should be opt-in since it's not a wide-spread convention AFAIK?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#626](https://github.com/realm/SwiftLint/issues/626)
 
+* Add `unused_closure_parameter` rule that validates if all closure parameters
+  are being used. If any of them isn't, it should be replaced by `_`.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#982](https://github.com/realm/SwiftLint/issues/982)
+
 ##### Bug Fixes
 
 * Fix `weak_delegate` rule reporting a violation for variables containing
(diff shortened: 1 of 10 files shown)


jpsim:
Do you have a job? Or do you just write awesome SwiftLint rules for a living? 😛

jpsim:
I made some small changes in d2b9c72, I hope you don't mind. I feel like this could _easily_ be made correctable. Do you think you can do that in this PR, or should we merge this as-is?

marcelofabri:
> Do you have a job? Or do you just write awesome SwiftLint rules for a living? 😛

I'll take that as a compliment 😂

> I made some small changes in d2b9c72, I hope you don't mind. I feel like this could easily be made correctable. Do you think you can do that in this PR, or should we merge this as-is?

I think we can merge this now and later make it correctable. I feel we should create a better way to have correctable `ASTRule`s to take advantage of the methods in `ASTRule` protocol.

jpsim:
Sounds good! Let's get this in.

PR #994, opened 2016-12-16
Fixing false positives in operator_usage_whitespace

### LINKED ISSUE #989: operator_usage_whitespace doesn't trigger if operand has parentheses
Examples:

```swift
let v8 = Int8(1)  << 6
let v8 = 1 <<  (6)
```

marcelofabri:
Fixes #989

### DIFF
--- a/Source/SwiftLintFramework/Rules/OperatorUsageWhitespaceRule.swift
+++ b/Source/SwiftLintFramework/Rules/OperatorUsageWhitespaceRule.swift
@@ -30,7 +30,9 @@ public struct OperatorUsageWhitespaceRule: OptInRule, ConfigurationProviderRule 
             "let foo = 1 + \n  2\n",
             "let range = 1...3\n",
             "let range = 1 ... 3\n",
-            "let range = 1..<3\n"
+            "let range = 1..<3\n",
+            "#if swift(>=3.0)\n",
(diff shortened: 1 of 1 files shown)


Source/SwiftLintFramework/Rules/OperatorUsageWhitespaceRule.swift
```diff
@@ -61,18 +66,19 @@ public struct OperatorUsageWhitespaceRule: OptInRule, ConfigurationProviderRule
...
         let patterns = spaces.map { first, second in
-            variableOrNumber + first + operators + second + variableOrNumber
+            leadingVariableOrNumber + first + operators + second + trailingVariableOrNumber
         }
         let pattern = "(?:\(patterns.joined(separator: "|")))"
 
-        let genericPattern = "<.+?>"
-        let validRangePattern = variableOrNumber + zeroSpaces + rangePattern +
-            zeroSpaces + variableOrNumber
+        let genericPattern = "<(?:\(oneSpace)|\\S)+?>" // not using dot to avoid matching new line
```
marcelofabri: this still can prevent some violations if there's `>` on the same line, for example:
```swift
method(1 <<   (6), otherParam: 1 > 2)
```

🐛

jpsim:
👍

PR #996, opened 2016-12-17
make UnusedClosureParameterRule correctable

jpsim:
I know @marcelofabri expressed an interest in improving how correctable rules are written in https://github.com/realm/SwiftLint/pull/992#issuecomment-267716268, but as I ran this rule on Realm and noticed that there were lots of violations, I realized it would take the same amount of time to manually fix the violations than it would take to make this rule correctable. So here we are...

How's this for now, @marcelofabri?


marcelofabri:
👯

PR #997, opened 2016-12-17
Add offset tests with multi-byte strings

marcelofabri:
As @jpsim mentioned in https://github.com/realm/SwiftLint/pull/995, this will catch problems with multi-byte strings


jpsim:
I thought we already had something like this... maybe I'm thinking of some tests in SourceKitten instead.

This is crashing on Linux 😱

jpsim:
Also interesting that these tests wouldn't have caught the problems fixed in #995.

marcelofabri:
> This is crashing on Linux 😱
From what I've seen, several methods are crashing on Linux when using emojis 😱. Mainly `replacingOccurrences(of:with)` and the ones in `cleanedContentsAndMarkerOffsets`. Should I just disable these new tests on Linux? Also, do you know a way of getting the stacktrace to find out the cause of a crash? My take on this so far has been adding `print` statements and running the tests again 😱😅  

> Also interesting that these tests wouldn't have caught the problems fixed in #995.
Actually, these would catch the one in `NimbleOperatorRule`. Either way, we probably should do a check and add missing violation markers on rules.

marcelofabri:
I haven't updated `OperatorUsageWhitespaceRule` because #990 already does that

jpsim:
Thanks for adding markers to all sorts of unmarkered rules!

I see you added a bunch of markers to the corrections, however I don't think these are validated in any way...

Also useful would be to test adding that family of emoji families (_err, basket of emojiables_) to corrections, since it's even more important to get the locations right when we're _editing and overwriting_ the file than when we're just reporting violations.

---

I think disabling on Linux is fine, due to those crashes. It'd be cool if we could add a comment linking to a Swift bug (`SR-XYZ`) explaining why these are disabled.

marcelofabri:
> I see you added a bunch of markers to the corrections, however I don't think these are validated in any way...

Actually, they are: https://github.com/realm/SwiftLint/blob/master/Tests/SwiftLintFrameworkTests/TestHelpers.swift#L95-L98

> Also useful would be to test adding that family of emoji families (err, basket of emojiables) to corrections (...)

Good idea 💯 

> I think disabling on Linux is fine, due to those crashes. It'd be cool if we could add a comment linking to a Swift bug (SR-XYZ) explaining why these are disabled.

I haven't found any Swift bug directly related to this, only a thread in the mailing list which seems to be the same problem: https://lists.swift.org/pipermail/swift-corelibs-dev/Week-of-Mon-20160523/000677.html.

jpsim:
> Actually, they are

Right-O! 😅

marcelofabri:
Adding tests for corrections already caught a 🐛: a crash in https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules/StatementPositionRule.swift#L229 😱

jpsim:
That's great!

marcelofabri:
I'll just fill the bug reports and mention them in code before merging this. The first one: https://bugs.swift.org/browse/SR-3448

jpsim:
Excellent. Thanks for being thorough

PR #998, opened 2016-12-18
Add unused_enumerated rule

### LINKED ISSUE #619: Add 'for (_, foo) in bar.enumerate()' rule
This would be useful as I see this in code reviews every now and again. 

People no longer need to use the index returned from enumerate, and xcode tells them to ignore "_" the variable. 

Instead, enumerate should be removed. 

Not sure if I should file a radar for this to be added directly to xcode, so opening it here first for discussion.

marcelofabri:
Fixes #619

### DIFF
--- a/Source/SwiftLintFramework/Rules/UnusedEnumeratedRule.swift
+++ b/Source/SwiftLintFramework/Rules/UnusedEnumeratedRule.swift
@@ -23,7 +23,8 @@ public struct UnusedEnumeratedRule: ASTRule, ConfigurationProviderRule {
             "for (_, foo) in bar.enumerated().something() { }\n",
             "for (_, foo) in bar.something() { }\n",
             "for foo in bar.enumerated() { }\n",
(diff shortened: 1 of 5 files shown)


Source/SwiftLintFramework/Rules/UnusedEnumeratedRule.swift
```diff
@@ -0,0 +1,98 @@
...
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "unused_enumerated",
+        name: "Unused Enumerated",
+        description: "When the index is not used, .enumerated() can be removed.",
+        nonTriggeringExamples: [
+            "for (idx, foo) in bar.enumerated() { }\n",
```
jpsim: I think it'd be useful to add a `for (idx, _) in bar.enumerated() { }\n` non-triggering example.

PR #1000, opened 2016-12-18
Updates docker image to use Swift 3.0.2

marcelofabri:
I don't think this will fix any of the issues we've been facing with Foundation on Linux, but it can't hurt.

Thanks to @norio-nomura for keeping these images updated 🙇


jpsim:
> Thanks to @norio-nomura to keeping these images updated 🙇

A million times yes! 👏 💯

PR #1003, opened 2016-12-18
	Add dictionary colon validation to colon rule

### LINKED ISSUE #603: Create a rule enforcing colon position for dictionaries
Not sure if this is possible, but i'd love a rule that enforces colon position for dictionaries, I like dictionaries to be defined as `[Key : Value]` with a space on either side of the colon, this rule could be for dictionary instances and also for dictionary type definitions (e.g `[String : AnyObject]`).

marcelofabri:
Fixes #603 

This probably should be updated to be auto-correctable too before merging. Just wanted to validate the approach first 😊


jpsim:
At first glance, I think I'm fine with this direction.

codeman9:
Have you tried a test with something like: ["string": "string"] and ["string:string": "string"] ? Those should not trigger.

marcelofabri:
@codeman9 I have added those in 9fae824. Thanks for the suggestion!

codeman9:
One more that you may want to check:

```
let tapRecognizer = UITapGestureRecognizer(target: self, action: #selector(handleTapGestureRecognizer(sender:)))
```

Source/SwiftLintFramework/Rules/RuleConfigurations/ColonConfiguration.swift
```diff
@@ -0,0 +1,41 @@
...
+import Foundation
+
+public struct ColonConfiguration: RuleConfiguration, Equatable {
+    private(set) var severityConfiguration = SeverityConfiguration(.warning)
+    private(set) var flexibleRightSpacing = false
+    private(set) var applyToDictionaries = true
+
+    public var consoleDescription: String {
+        return severityConfiguration.consoleDescription +
+        ", flexible_right_spacing: \(flexibleRightSpacing)" +
```
jpsim: indentation

Source/SwiftLintFramework/Rules/ColonRule.swift
```diff
@@ -9,27 +9,16 @@
...
+    public var configuration = ColonConfiguration()
 
-    public var configurationDescription: String {
-        return configuration.consoleDescription +
-            ", flexible_right_spacing: \(flexibleRightSpacing)"
-    }
+    public init() {}
 
     public static let description = RuleDescription(
         identifier: "colon",
```
jpsim: The description needs to be updated.

> Colons should be next to the identifier when specifying a type and next to the key in dictionary literals.

Source/SwiftLintFramework/Rules/ColonRule.swift
```diff
@@ -161,3 +208,89 @@ public struct ColonRule: CorrectableRule, ConfigurationProviderRule {
...
+                return !isCorrect
+            }
+
+            return colon != ": " && !colon.hasPrefix(":\n")
+        }
+    }
+
+    fileprivate func dictionaryViolations(in file: File,
+                                          dictionary: [String: SourceKitRepresentable])
+                                                                                    -> [NSRange] {
```
jpsim: elsewhere in the codebase that we need to split lines at the return type, we align the return arrow vertically with the parameters:

```swift
fileprivate func dictionaryViolations(in file: File,
                                      dictionary: [String: SourceKitRepresentable])
                                      -> [NSRange] {
```

Source/SwiftLintFramework/Rules/ColonRule.swift
```diff
@@ -161,3 +209,89 @@ public struct ColonRule: CorrectableRule, ConfigurationProviderRule {
...
+                    return nil
+            }
+
+            return NSRange(location: offset, length: length)
+        }
+
+        let even = ranges.enumerated().flatMap { $0 % 2 == 0 ? $1 : nil }
+        let odd = ranges.enumerated().flatMap { $0 % 2 != 0 ? $1 : nil }
+
+        return zip(even, odd).map { range1, range2 -> NSRange in
```
jpsim: I feel like `evenRange, oddRange` would be better parameter names?

Source/SwiftLintFramework/Rules/ColonRule.swift
```diff
@@ -161,3 +209,89 @@ public struct ColonRule: CorrectableRule, ConfigurationProviderRule {
...
+
+        let ranges = violationRangesInFile(file, kind: kind, dictionary: dictionary)
+        return ranges.map {
+            StyleViolation(ruleDescription: type(of: self).description,
+                           severity: configuration.severityConfiguration.severity,
+                           location: Location(file: file, byteOffset: $0.location))
+        }
+    }
+
+    private func violationRangesInFile(_ file: File, kind: SwiftExpressionKind,
```
jpsim: this function name could make it clearer that it only detects dictionary violations

Source/SwiftLintFramework/Rules/ColonRule.swift
```diff
@@ -161,3 +209,89 @@ public struct ColonRule: CorrectableRule, ConfigurationProviderRule {
         }
     }
 }
+
+// MARK: Dictionary Colon Rule
+extension ColonRule {
+
+    public func validateFile(_ file: File, kind: SwiftExpressionKind,
```
jpsim: this function name could make it clearer that it only detects dictionary violations

Source/SwiftLintFramework/Rules/ColonRule.swift
```diff
@@ -94,42 +94,90 @@ public struct ColonRule: CorrectableRule, ConfigurationProviderRule {
...
+                return nil
+            }
+            return (range: range, kind: ColonKind.dictionary)
+        }
+
+        return (violations + dictViolations).sorted { $0.range.location < $1.range.location }
+    }
+}
+
+// MARK: Type Colon Rule
```
jpsim: elsewhere in SwiftLint's codebase, we tend to consistently have a single empty newline after `// MARK`s.

Source/SwiftLintFramework/Rules/ColonRule.swift
```diff
@@ -94,42 +94,90 @@ public struct ColonRule: CorrectableRule, ConfigurationProviderRule {
...
+            "let abc = [Void↓:  Void]()\n": "let abc = [Void: Void]()\n",
+            "let abc = [Void↓ :  Void]()\n": "let abc = [Void: Void]()\n",
+            "let abc = [1: [3↓ : 2], 3: 4]\n": "let abc = [1: [3: 2], 3: 4]\n",
+            "let abc = [1: [3↓ : 2], 3↓:  4]\n": "let abc = [1: [3: 2], 3: 4]\n"
         ]
     )
 
     public func validateFile(_ file: File) -> [StyleViolation] {
-        return violationRangesInFile(file, withPattern: pattern).flatMap { range in
+        let violations = violationRangesInFile(file, withPattern: pattern).flatMap { range in
```
jpsim: The `violationRangesInFile(_:withPattern:)` function name should make it clearer that it only detects type violations.

marcelofabri:
I was going with more clearer function names, but got bitten by the 100-char line length limit. Now that we merged #1010, we can have better names 😝

Source/SwiftLintFramework/Rules/ColonRule.swift
jpsim: right... a comment then

marcelofabri: that was the reason I put that in an extension, to group the dictionary-related functions. I'll add a comment either way 👍

jpsim: the extension makes it clearer at the declaration site. The comment... doesn't help clarifying the call site either 😓

jpsim: but a comment at the call site does... ugh. anyway, no great solutions to any of this. do what you think is best.

marcelofabri: yeah, mixing `ASTRule` with regular `Rule` is weird ATM. I even thought about making this a new rule.

PR #1004, opened 2016-12-19
Closure end indentation rule

### LINKED ISSUE #326: [Rule suggestion] Indentation for closure end
Xcode still doesn't know how to indent this one correctly, and I'd love it if `SwiftLint` could detect it AND correct it, because it really kills my OCD.

I think this only happens when the closure starts in a "overly" indented line. A very common example of this is when using a chained operator, like in RAC:

``` swift
SignalProducer(values: [1, 2, 3])
    .map { $0 + 1 }
    .startWithNext { number in
        print(number)
}
```

That's how Xcode indents the closing `}`, which is really annoying. Instead, it should be aligned:

``` swift
SignalProducer(values: [1, 2, 3])
    .map { $0 + 1 }
    .startWithNext { number in
        print(number)
    }
```

I'm happy to implement this myself (I'd love to finally dive in `SwiftLint`'s source), just wanted to make sure other people agree with this too :)

marcelofabri:
Fixes #326

This still have some issues:
- [x] It should be opt-in. It's not currently to make it easier to test.
- [x] Some false positives are being trigged, mainly related to opening brace on different line than the method call. For example:
```swift
return file.matchPattern(pattern, excludingSyntaxKinds: kinds,
                                          excludingPattern: excludingPattern).map {
    return StyleViolation(ruleDescription: type(of: self).description,
                          location: Location(file: file, characterOffset: $0.location))
}
```
- [x] It should check (and skip) if the closure begins and ends on the same line, not the body offset. This should prevent this false violation:
```swift
 return file.matchPattern(pattern, excludingSyntaxKinds: kinds,
                                 excludingPattern: excludingPattern) { ↓$0.rangeAt(1) }.flatMap {
            let parensRegex = NSRegularExpression.forcePattern(parensPattern)
            return parensRegex.firstMatch(in: file.contents, options: [], range: $0)?.range
}
```

The main issue here is that I'd need the starting point of the method related to the closure, not the full chained one (which is what SourceKits gives me). Any ideas on how to get it?

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -14,6 +14,7 @@ opt_in_rules:
   - nimble_operator
   - attributes
   - operator_usage_whitespace
+  - closure_end_indentation
 
 file_header:
   required_pattern: |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,16 @@
 
 ##### Enhancements
 
-* None.
+* Add `closure_end_indentation` opt-in rule that validates closure closing 
+  braces according to this logic:
+  * If the method call is chained breaking lines on each method 
+  (`.` is on a new line), the closing brace should be in the same 
+  indentation as the `.`.
+  * Else, the closing brace should be in the same indentation as 
+  the beginning of the statement (in the first line).  
+
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#326](https://github.com/realm/SwiftLint/issues/326)
 
 ##### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,13 +6,13 @@
 
 ##### Enhancements
 
-* Add `closure_end_indentation` opt-in rule that validates closure closing 
-  braces according to this logic:
-  * If the method call is chained breaking lines on each method 
-  (`.` is on a new line), the closing brace should be in the same 
-  indentation as the `.`.
-  * Else, the closing brace should be in the same indentation as 
-  the beginning of the statement (in the first line).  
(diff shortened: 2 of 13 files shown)


jpsim:
I'm not sure I stylistically agree with the current semantics (the style changes you've already made). Would need to write down more explicit rules just in sentences/point-form explaining when to place closing braces and where.

marcelofabri:
So, my interpretation of the comments in #326 was:

* If the method call is chained breaking lines on each method (e.g. `.` is on a new line), the closing brace should be in the same indentation as the `.`.
* Else, the closing brace should be in the same indentation as the beginning of the statement (in the first line).

jpsim:
Yeah, that sounds good. It was hurting my head a bit to try to think of when you'd apply which rules... this is why we need machines to decide this for us! 😅

marcelofabri:
🤖 > 👨‍💻

jpsim:
> 🤖 > 👨‍💻

Is it too late to rename SwiftLint to this? 😂 

But seriously, this is the exact opposite of the point I was trying to make in my SwiftLint presentation from last week ([slides](https://speakerdeck.com/jpsim/watch-your-language-the-road-to-clean-code-with-swiftlint)), that you shouldn't bend to the will of the machines and humans know better...

But there are times when the machines pay closer attention to details than we can.

marcelofabri:
Maybe the issue is that it's harder to come up with logic rules to describe what "feels right". Our intuition is much more complex than I thought! I've learned the hard way when developing https://github.com/realm/SwiftLint/pull/847.

jpsim:
> Our intuition is much more complex than I thought

This. Though it _is_ definable.

jpsim:
🎉 thanks Marcelo!

PR #1007, opened 2016-12-19
Add check to ignore urls for line_length rule

### LINKED ISSUE #384: Allow urls to form long lines without long line warnings.
Some urls are just long and shortening them isn't ideal.

jaherhi:
Fixes #384

The rule now detects urls using `NSDataDetector` and replaces them with empty strings, so they can be ignored when checking the length of a line.


CHANGELOG.md
```diff
@@ -6,7 +6,9 @@
 
 ##### Enhancements
 
-* None.
+* Add check to ignore urls in `line_length` rule.
```
jpsim: t o o   m u c h   t r a i l i n g   w h i t e s p a c e 

Just two: see https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes

Source/SwiftLintFramework/Rules/LineLengthRule.swift
```diff
@@ -87,4 +92,14 @@ public struct LineLengthRule: ConfigurationProviderRule, SourceKitFreeRule {
         return modifiedString
     }
 
+    private func stripUrls(fromSourceString sourceString: String) -> String {
+        let types = NSTextCheckingResult.CheckingType.link.rawValue
+        let range = NSRange(location: 0, length: sourceString.characters.count)
```
jpsim: this should be `sourceString.bridge().length` since the number of characters does not always equal the NSString length value.

jaherhi:
Sure, I'll change it so it can be configured 🙂

jaherhi:
The rule is now configurable using `ignore_urls`. It's disabled by default. I also updated the branch to solve some conflicts.

I'm having difficulties to fix the errors Travis shows when testing with SPM and Linux. Any advice on how to proceed with that?

marcelofabri:
@jaherhi I think you need to `import Foundation` on LineLengthRule.swift. 
You can test it on your machine with `swift test`.

jaherhi:
Yeah, that fixed it, at least locally on my machine. I should have thought about that 😅 Thanks a lot @marcelofabri

marcelofabri:
So, it turns out we forgot that `NSDataDetector` is available only in the Objective-C Foundation, not in https://github.com/apple/swift-corelibs-foundation which is what the Linux build uses. Maybe we should log a warning if this configuration is set on Linux to say it's unsupported?

jaherhi:
I added a check to ignore this configuration on Linux, although I'm not sure if this is the right thing to do since there will be differences when using SwiftLint on Linux.

Would it make sense to use `NSRegularExpression` to detect urls on Linux? That way the rule may have more or less the same behaviour, and the main difference would be the way urls are detected.

marcelofabri:
Well, using a regex may lead to different behaviors as well since we don't know exactly how `NSDataDetector` works. Feel free to implement this though! :)

Source/SwiftLintFramework/Rules/LineLengthRule.swift
```diff
@@ -87,4 +91,25 @@ public struct LineLengthRule: ConfigurationProviderRule, SourceKitFreeRule {
...
 
+    private func stripUrls(fromSourceString sourceString: String) -> String {
+        let range = NSRange(location: 0, length: sourceString.bridge().length)
+        // Workaround for Linux until NSDataDetector is available
+        #if os(Linux)
+            // Regex pattern from http://daringfireball.net/2010/07/improved_regex_for_matching_urls
+            let pattern = "(?i)\\b((?:[a-z][\\w-]+:(?:/{1,3}|[a-z0-9%])|www\\d{0,3}[.]|[a-z0-9.\\-]+[.][a-z]{2,4}/)" +
+                "(?:[^\\s()<>]+|\\(([^\\s()<>]+|(\\([^\\s()<>]+\\)))*\\))+(?:\\(([^\\s()<>]+|(\\([^\\s()<>]+\\)))*" +
+                "\\)|[^\\s`!()\\[\\]{};:'\".,<>?«»“”‘’]))"
+            guard let regex = try? NSRegularExpression(pattern: pattern, options: []) else {
```
marcelofabri: you could use `regex(pattern)` here so it can be cached (and you know it it's a valid regex)

CHANGELOG.md
```diff
@@ -106,6 +106,11 @@
 
 ##### Enhancements
 
+* Add check to ignore urls in `line_length` rule when `ignores_urls`
```
jpsim: this is now in the wrong section since 0.15.0 has been released. please move up.

CHANGELOG.md
```diff
@@ -106,6 +106,11 @@
...
+  configuration is enabled.
```
jpsim: just two trailing spaces here

Source/SwiftLintFramework/Rules/LineLengthRule.swift
```diff
@@ -6,10 +6,11 @@
 //  Copyright © 2015 Realm. All rights reserved.
 //
 
+import Foundation
 import SourceKittenFramework
 
 public struct LineLengthRule: ConfigurationProviderRule, SourceKitFreeRule {
-    public var configuration = SeverityLevelsConfiguration(warning: 120, error: 200)
+    public var configuration = LineLengthConfiguration(warning: 120, error: 200, ignoresUrls: false)
```
jpsim: please keep acronyms capitalized (`ignoresURLs`)

Source/SwiftLintFramework/Rules/LineLengthRule.swift
```diff
@@ -87,4 +91,23 @@ public struct LineLengthRule: ConfigurationProviderRule, SourceKitFreeRule {
         return modifiedString
     }
 
+    private func stripUrls(fromSourceString sourceString: String) -> String {
```
jpsim: please keep acronyms capitalized (`stripURLs`)

jpsim: also, for Swift 3 API guidelines, this should probably be `stripURLs(from string: String) -> String` or a `fileprivate` extension on `String` as a computed variable `strippingURLs`.

Source/SwiftLintFramework/Rules/RuleConfigurations/LineLengthConfiguration.swift
```diff
@@ -0,0 +1,48 @@
...
+
+import Foundation
+
+public struct LineLengthConfiguration: RuleConfiguration, Equatable {
+    public var consoleDescription: String {
+        return length.consoleDescription + ", ignores urls: \(ignoresUrls)"
+    }
+
+    var length: SeverityLevelsConfiguration
+    var ignoresUrls: Bool
```
jpsim: please keep acronyms capitalized (`ignoresURLs`)

Tests/SwiftLintFrameworkTests/RuleConfigurationTests.swift
```diff
@@ -10,8 +10,127 @@ import SourceKittenFramework
 @testable import SwiftLintFramework
 import XCTest
 
+// swiftlint:disable file_length
```
jpsim: The number of tests this adds as well as the silenced SwiftLint length violations should be an indicator that it might be time to move line length rule and configuration tests into their own test file. Could you please do that? Also please remember to update `LinuxMain.swift`.

jaherhi:
Those commits shouldn't be there 😕 I'll fix it and push again

Tests/LinuxMain.swift
```diff
@@ -18,6 +18,8 @@ XCTMain([
     testCase(FileHeaderRuleTests.allTests),
     testCase(FunctionBodyLengthRuleTests.allTests),
     testCase(IntegrationTests.allTests),
+    testCase(LineLengthConfigurationTests),
```
marcelofabri: these should be `ClassName.allTests`

jpsim:
Thanks for your continued work on this @jaherhi 🎉

PR #1010, opened 2016-12-20
Changing the default line length limit to 120

### LINKED ISSUE #1008: Line length and [String: SourceKitRepresentable]
Our current line length limit is 100, which makes creating functions that receives `[String: SourceKitRepresentable]` as parameter hard, since it's a long (33-char) type.

For example, how should one format this method?

```swift
    fileprivate func dictionaryViolations(in file: File,
                                          dictionary: [String: SourceKitRepresentable])
                                                                                    -> [NSRange] {
```

I find myself using worse choices of name method names or weird formatting because of this limit. I propose we should either (or both 😛):

1. Increase the limit a bit, to 120 for example.
2. Create a `typealias SouceKitDictionary =  [String: SourceKitRepresentable]` to save 11 characters.

Thoughts?

marcelofabri:
Fixes #1008.

I particularly don't like breaking lines on `(` or `->` in function declarations, so I've changed the ones I've found now that we have a larger line length limit.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -25,3 +25,4 @@ file_header:
                     \/\/  Created by .*? on \d{1,2}\/\d{1,2}\/\d{2}\.
                     \/\/  Copyright © \d{4} Realm\. All rights reserved\.
                     \/\/
+line_length: 120
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ##### Breaking
 
-* None.
+* `file_length` rule now has a default value of `120` for warnings.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#1008](https://github.com/realm/SwiftLint/issues/1008)
 
 ##### Enhancements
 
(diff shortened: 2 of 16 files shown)


.swiftlint.yml
```diff
@@ -25,3 +25,4 @@ file_header:
                     \/\/  Created by .*? on \d{1,2}\/\d{1,2}\/\d{2}\.
                     \/\/  Copyright © \d{4} Realm\. All rights reserved\.
                     \/\/
+line_length: 120
```
marcelofabri: I've added this here to prevent the `swiftlint` build phase to generate warnings (since it's using the installed version, which probably is the latest stable one).

CHANGELOG.md
```diff
@@ -2,7 +2,9 @@
 
 ##### Breaking
 
-* None.
+* `file_length` rule now has a default value of `120` for warnings.
```
jpsim: Though not really a breaking change, I do approve of having it in this section to spread the word...

.swiftlint.yml
jpsim: Maybe we could log a message if a configuration has line length explicitly set to 120, so that more people will be aware that this is SwiftLint's new default?

marcelofabri: I don't feel it'd be very useful as some people may see to not relying on the defaults as a good practice?

jpsim: yeah, maybe. I'm a bit torn, so happy to go with your preference on this one.

PR #1015, opened 2016-12-21
Switch to a relative path within the HTML report output

JamieEdge:
Currently the HTML report output uses the full path to each source file. In the case of using this in a CI environment, the path can be quite long and adds a lot of noise to the report.

This change displays the file paths relative to the directory that SwiftLint was run within (only for HTML). An example of the output is below.

Relative Path Example


CHANGELOG.md
```diff
@@ -19,7 +19,10 @@
 
 * `operator_usage_whitespace` rule is now correctable.  
   [Marcelo Fabri](https://github.com/marcelofabri)
-  
+
+* HTML reports now display a relative path to files.
```
marcelofabri: this is missing 2 trailing whitespaces

JamieEdge: Sorry - fixed now.

JamieEdge:
Thanks - conflicts now resolved.

marcelofabri:
Thanks 🚢

PR #1022, opened 2016-12-21
Fix false positive in `empty_parentheses_with_trailing_closure` rule

### LINKED ISSUE #1021: false positive with empty_parentheses_with_trailing_closure
@marcelofabri 

```swift
UIView.animateWithDuration(0.3, animations: {
    self.disableInteractionRightView.alpha = 0
}, completion: { _ in
    ()
})

### PR DESCRIPTION
**Pull request by marcelofabri**
Fixes #1021

### DIFF
--- a/Source/SwiftLintFramework/Rules/EmptyParenthesesWithTrailingClosureRule.swift
+++ b/Source/SwiftLintFramework/Rules/EmptyParenthesesWithTrailingClosureRule.swift
@@ -24,7 +24,12 @@ public struct EmptyParenthesesWithTrailingClosureRule: ASTRule, ConfigurationPro
             "[1, 2].map({ $0 + 1 })\n",
             "[1, 2].reduce(0) { $0 + $1 }",
             "[1, 2].map { number in\n number + 1 \n}\n",
-            "let isEmpty = [1, 2].isEmpty()\n"
+            "let isEmpty = [1, 2].isEmpty()\n",
+            "UIView.animateWithDuration(0.3, animations: {\n" +
+            "   self.disableInteractionRightView.alpha = 0\n" +
+            "}, completion: { _ in\n" +
+            "   ()\n" +
+            "})"
         ],
         triggeringExamples: [
             "[1, 2].map↓() { $0 + 1 }",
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Source/SwiftLintFramework/Rules/EmptyParenthesesWithTrailingClosureRule.swift [hunk 1]
```diff
@@ -24,7 +24,12 @@ public struct EmptyParenthesesWithTrailingClosureRule: ASTRule, ConfigurationPro
             "[1, 2].map({ $0 + 1 })\n",
             "[1, 2].reduce(0) { $0 + $1 }",
             "[1, 2].map { number in\n number + 1 \n}\n",
-            "let isEmpty = [1, 2].isEmpty()\n"
+            "let isEmpty = [1, 2].isEmpty()\n",
+            "UIView.animateWithDuration(0.3, animations: {\n" +
+            "   self.disableInteractionRightView.alpha = 0\n" +
+            "}, completion: { _ in\n" +
+            "   ()\n" +
```
marcelofabri: this is valid Swift 😳
weird and useless, but valid 😱

BluMist: thanks for fixing this "useless" case, it is valid Swift :P
we tend to use () in our switch statement for no-op instead of "break" since break can mean breaking out of a loop

marcelofabri: hah, I thought this was just a sample of a more complicated case - but good to know that it's actually useful!

jpsim: that's interesting! We didn't mean any offense by that, in fact it might be better to do than what I usually do in these cases, which is to add a `/* no-op */` comment.

BluMist: no offense taken, just explaining how we actually hit this issue in the first place

PR #1030, opened 2016-12-22
Fix false positive in `operator_usage_whitespace`

### LINKED ISSUE #1028: 'operator_usage_whitespace' rule conflicts with Image Literal
When using image literal, It's should not triggered

marcelofabri:
Fixes #1028

### DIFF
--- a/Source/SwiftLintFramework/Rules/OperatorUsageWhitespaceRule.swift
+++ b/Source/SwiftLintFramework/Rules/OperatorUsageWhitespaceRule.swift
@@ -32,7 +32,9 @@ public struct OperatorUsageWhitespaceRule: OptInRule, CorrectableRule, Configura
             "let range = 1 ... 3\n",
             "let range = 1..<3\n",
             "#if swift(>=3.0)\n",
-            "array.removeAtIndex(-200)\n"
+            "array.removeAtIndex(-200)\n",
+            "let name = \"image-1\"\n",
+            "button.setImage(#imageLiteral(resourceName: \"image-1\"), for: .normal)\n"
         ],
         triggeringExamples: [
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Rules/OperatorUsageWhitespaceRule.swift
```diff
@@ -105,7 +107,7 @@ public struct OperatorUsageWhitespaceRule: OptInRule, CorrectableRule, Configura
             zeroSpaces + trailingVariableOrNumber
         let excludingPattern = "(?:\(genericPattern)|\(validRangePattern))"
 
-        let kinds = SyntaxKind.commentAndStringKinds()
+        let kinds = SyntaxKind.commentAndStringKinds() + [.objectLiteral]
```
jpsim: I would suspect that other rules that exclude comments and strings would also need to exclude object literals.

This is why checking for explicit matching kinds is always preferable to excluding.

marcelofabri: I agree, but I'm not sure what we should do. Maybe rename `commentAndStringKinds` and make it return `.objectLiteral` as well?

jpsim: yes, basically "non-code kinds", and check to see if there are even more kinds that fall into that category.

jpsim: I've looked at all the other rules using `commentAndStringKinds()` and I don't think any of them would benefit from also including image literals. So let's keep it just here for now.

PR #1035, opened 2016-12-22
Vertical parameter alignment rule

### LINKED ISSUE #1033: Rule Request: Vertical Parameter Alignment
When refactoring, it's often very easy to mess up the vertical alignment of parameters in function declarations and calls:

```swift
func validateFunction(_ file: File, kind: SwiftDeclarationKind,
```

We could (easily?) have a correctable rule for this.

marcelofabri:
Fixes #1033 

This isn't correctable yet because I wasn't sure if it'd be ok to use spaces to correct the indentation. I heard some people use tabs 🤷‍♂️

Also #1033 mentions checking function calls, but I think it's better to do it later as it's much more complex IMO.

### DIFF
--- a/Source/SwiftLintFramework/Rules/VerticalParameterAlignmentRule.swift
+++ b/Source/SwiftLintFramework/Rules/VerticalParameterAlignmentRule.swift
@@ -1,0 +1,86 @@
+//
+//  VerticalParameterAlignmentRule.swift
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 22/12/16.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+public struct VerticalParameterAlignmentRule: ASTRule, ConfigurationProviderRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "vertical_parameter_alignment",
+        name: "Vertical Parameter Alignment",
+        description: "Function parameters should be aligned vertically if they're in multiple lines in a declaration.",
+        nonTriggeringExamples: [
+            "func validateFunction(_ file: File, kind: SwiftDeclarationKind,\n" +
+            "                      dictionary: [String: SourceKitRepresentable]) { }\n",
+            "func validateFunction(_ file: File, kind: SwiftDeclarationKind,\n" +
+            "                      dictionary: [String: SourceKitRepresentable]) -> [StyleViolation]\n",
+            "func foo(bar: Int)\n",
+            "func foo(bar: Int) -> String \n",
+            "func validateFunction(_ file: File, kind: SwiftDeclarationKind,\n" +
+            "                      dictionary: [String: SourceKitRepresentable])\n" +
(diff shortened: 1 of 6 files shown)


Source/SwiftLintFramework/Rules/VerticalParameterAlignmentRule.swift
```diff
@@ -0,0 +1,86 @@
+//
+//  VerticalParameterAlignmentRule.swift
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 22/12/16.
```
jpsim: 🇺🇸 📆 🔥

marcelofabri: obviously 😂

jpsim:
Some minor ideas you may or may not want to incorporate:

```diff
diff --git a/Source/SwiftLintFramework/Rules/VerticalParameterAlignmentRule.swift b/Source/SwiftLintFramework/Rules/VerticalParameterAlignmentRule.swift
index f9f6dc3..6484c90 100644
--- a/Source/SwiftLintFramework/Rules/VerticalParameterAlignmentRule.swift
+++ b/Source/SwiftLintFramework/Rules/VerticalParameterAlignmentRule.swift
@@ -59,17 +59,15 @@ public struct VerticalParameterAlignmentRule: ASTRule, ConfigurationProviderRule
             let paramStart = regex(pattern).firstMatch(in: file.contents,
                                                        options: [], range: nameRange)?.rangeAt(1).location,
             let (startLine, startCharacter) = contents.lineAndCharacter(forCharacterOffset: paramStart),
-            let (endLine, _) = contents.lineAndCharacter(forByteOffset: nameOffset + nameLength - 1),
+            let endLine = contents.lineAndCharacter(forByteOffset: nameOffset + nameLength - 1)?.0,
             endLine > startLine else {
                 return []
         }
 
-        let linesRange = (startLine + 1)...endLine
-        let violationLocations = linesRange.flatMap { lineIndex -> Int? in
-            let line = file.lines[lineIndex - 1]
+        let violationLocations = (startLine..<endLine).flatMap { lineIndex -> Int? in
             guard let paramLocation = regex("\\S").firstMatch(in: file.contents, options: [],
-                                                              range: line.range)?.range.location,
-                let (_, paramCharacter) = contents.lineAndCharacter(forCharacterOffset: paramLocation),
+                                                              range: file.lines[lineIndex].range)?.range.location,
+                let paramCharacter = contents.lineAndCharacter(forCharacterOffset: paramLocation)?.1,
                 paramCharacter != startCharacter else {
                     return nil
             }
```

marcelofabri:
I actually prefer the current style if it's ok. Even if it's smarter because it avoids the `+1` and `-1`, I think it's less readable (it took me a while to realize there wasn't a bug).

jpsim:
> I actually prefer the current style if it's ok.

It's absolutely ok! My sample basically abuses the fact that line indexes are 1-indexed.

jpsim:
If you end up closing #1033 after merging this, make sure you open a new issue to track improving this rule by also checking function _calls_.

Issue #1025, opened 2016-12-22
Add CI job to run on real OSS projects


jpsim:
We've already seen in a handful of PRs recently that running on [Realm Swift](https://github.com/realm/realm-cocoa) was beneficial in highlighting false positives.

There are a number of other open source projects that use SwiftLint that we could run new rules and bug fixes against, to help minimize the number of false positives.

* [Realm Swift](https://github.com/realm/realm-cocoa)
* [Kickstarter for iOS](https://github.com/kickstarter/ios-oss)
* [Facebook SDK Swift](https://github.com/facebook/facebook-sdk-swift)
* [ReduxKit](https://github.com/ReduxKit/ReduxKit)
* [Aerial](https://github.com/JohnCoates/Aerial)
* [LTMorphingLabel](https://github.com/lexrus/LTMorphingLabel)
* [Moya](https://github.com/Moya/Moya)
* [SourceKitten](https://github.com/jpsim/SourceKitten)

If we do this, these integration tests should never fail a build, and would ideally follow in Danger/Codecov's approach of posting a constantly-updating GitHub comment reporting the violations in these projects. (`GitHubReporter` anyone?)

Proactive contributors could also "fix" these style violations in these OSS projects! Everybody wins 😄.

marcelofabri:
I really like this idea! 💯 

I only wish CI was faster to handle this as well as it deserves 😬

marcelofabri:
We could even make the report with checkboxes so we can mark if the violation is a valid one or not ✅

jpsim:
Could have a "Regressions" section too, which would obviously be more important to address than say a valid violation of an unreleased rule.

marcelofabri:
We could run the projects against a stable SwiftLint version and the current one. Then we can flag the violations that only happen in the new version (and are from a rule that is also in the stable version) as potential regressions. I say potential because we could been improving an existing rule to flag more cases (such as `comma_rule`).

jpsim:
> We could run the projects against a stable SwiftLint version and the current one.

If I recall correctly, that's the approach that @scottrhoyt had started to take when working on performance tests (see https://github.com/realm/SwiftLint/issues/376#issuecomment-268702693).

Architecturally, I think that's slightly the wrong approach, and that it'd be preferable to compare `master` against the PR, but on the other hand, it's that kind of "search for perfection" that stalled the work on automated performance testing last time around, so I'm a lot more open to tackling this problem in more incremental steps.

[...]

PR #1036, opened 2016-12-23
Initial Danger integration

### LINKED ISSUE #843: Setup Danger?
[People](https://github.com/realm/SwiftLint/pull/842#issuecomment-253991177) seems to forget about adding an entry to CHANGELOG. We could add [Danger](http://danger.systems) to validate that automatically.

We could even borrow [CocoaPods's Dangerfile](https://github.com/CocoaPods/CocoaPods/blob/master/Dangerfile#L18-L32).

marcelofabri:
Fixes #843

@jpsim We need to create a bot account and a token with it, then add it as an env var (`DANGER_GITHUB_API_TOKEN`)

From `danger init`:

```

Step 2: Creating a GitHub account

In order to get the most out of Danger, I'd recommend giving her the ability to post in
the code-review comment section.

IMO, it's best to do this by using the private mode of your browser. Create an account like
SwiftlintBot, and don't forget a cool robot avatar.

Here are great resources for creative commons images of robots:
 -> https://www.flickr.com/search/?text=robot&license=2%2C3%2C4%2C5%2C6%2C9
 -> https://www.google.com/search?q=robot&tbs=sur:fmc&tbm=isch&tbo=u&source=univ&sa=X&ved=0ahUKEwjgy8-f95jLAhWI7hoKHV_UD00QsAQIMQ&biw=1265&bih=1359
SwiftlintBot will need access to your repo. Simply because the code is not available for the public
to read and comment on.

Note: Holding cmd ( ⌘ ) and double clicking a link will open it in your browser.

Cool, please press return when you have your account ready (and you've verified the email...)


Step 3: Configuring a GitHub Personal Access Token

Here's the link, you should open this in the private session where you just created the new GitHub account
 -> https://github.com/settings/tokens/new

For token access rights, I need to know if this is for an Open Source or Closed Source project
? [ Open / Closed ]
> 
Using: open
For Open Source projects, I'd recommend giving the token the smallest scope possible.
This means only providing access to public_repo in the token.

This token limits Danger's abilities to just writing comments on OSS projects. I recommend
this because the token can quite easily be extracted from the environment via pull requests.

It is important that you do not store this token in your repository, as GitHub will automatically revoke it when pushed.

👍, please press return when you have your token set up...


Step 4: Add Danger for your CI

I'd recommend adding before_script: bundle exec danger to the script section of your .travis.yml file.
You shouldn't use after_success, after_failure, after_script as they cannot fail your builds.

OK, I'll give you a moment to do this...


Final step: exposing the GitHub token as an environment build variable.

As you have an Open Source repo, this token should be considered public, otherwise you cannot
run Danger on pull requests from forks, limiting its use.
In order to add an environment variable, go to:
 -> https://travis-ci.org/marcelofabri/SwiftLint/settings

The name is DANGER_GITHUB_API_TOKEN and the value is the GitHub Personal Access Token.
Make sure to have "Display value in build log" enabled.
This is the last step, I can give you a second...
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
       env: JOB=SPM
       os: osx
       osx_image: xcode8.2
-      before_script: bundle exec danger
+      before_script: bundle install --jobs=3 --retry=3 --without documentation --path bundle && bundle exec danger
     - script: make docker_test
       env: JOB=Linux
       sudo: required
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,9 @@ matrix:
       env: JOB=SPM
       os: osx
       osx_image: xcode8.2
-      before_script: bundle install --jobs=3 --retry=3 --without documentation --path bundle && bundle exec danger
+      before_script:
+        - bundle install --jobs=3 --retry=3 --without documentation --path bundle
+        - bundle exec danger
     - script: make docker_test
       env: JOB=Linux
       sudo: required
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ matrix:
       env: JOB=SPM
       os: osx
       osx_image: xcode8.2
+      before_script: bundle exec danger
     - script: make docker_test
       env: JOB=Linux
       sudo: required
--- a/Dangerfile
+++ b/Dangerfile
@@ -12,9 +12,9 @@ if !git.modified_files.include?("CHANGELOG.md") && has_app_changes
     markdown <<-MARKDOWN
 Here's an example of your CHANGELOG entry:
 ```markdown
-* #{github.pr_title}#{' '}
-[#{github.pr_author}](https://github.com/#{github.pr_author})
-[#issue_number](https://github.com/realm/SwiftLint/issues/issue_number)
+* #{github.pr_title}#{'  '}
+  [#{github.pr_author}](https://github.com/#{github.pr_author})
+  [#issue_number](https://github.com/realm/SwiftLint/issues/issue_number)
 ```
 *note*: There are two invisible spaces after the entry's text.
 MARKDOWN
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,0 +1,26 @@
+# Warn when there is a big PR
+warn("Big PR") if git.lines_of_code > 500
+
+# Sometimes its a README fix, or something like that - which isn't relevant for
+# including in a CHANGELOG for example
+has_app_changes = !git.modified_files.grep(/Source/).empty?
+has_test_changes = !git.modified_files.grep(/Tests/).empty?
+
+# Add a CHANGELOG entry for app changes
+if !git.modified_files.include?("CHANGELOG.md") && has_app_changes
+  fail("Please include a CHANGELOG entry to credit yourself! \nYou can find it at [CHANGELOG.md](https://github.com/realm/SwiftLint/blob/master/CHANGELOG.md).")
+    markdown <<-MARKDOWN
+Here's an example of your CHANGELOG entry:
+```markdown
+* #{github.pr_title}#{' '}
+[#{github.pr_author}](https://github.com/#{github.pr_author})
+[#issue_number](https://github.com/realm/SwiftLint/issues/issue_number)
+```
+*note*: There are two invisible spaces after the entry's text.
+MARKDOWN
+end
+
+# Non-trivial amounts of app changes without tests
+if git.lines_of_code > 50 && has_app_changes && !has_test_changes
+  warn "This PR may need tests."
+end
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,7 @@ Packages
 
 # macOS
 .DS_Store
+
+# Bundler
+.bundle/
+bundle/
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,4 @@
+# frozen_string_literal: true
+source "https://rubygems.org"
+
+gem "danger"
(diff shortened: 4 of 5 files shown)


marcelofabri:
I've done this a first step towards #1025.

jpsim:
> I've done this a first step towards #1025.

I'm not _exactly_ sure how this helps with #1025. You're thinking we should run OSS-check as a "Danger job" in order to piggy-back off its auto-updating GitHub comment reporting capabilities?

marcelofabri:
Yes! My idea would be having a custom formatter as you hinted and then reading it and commenting with danger. Or we could use the JSON formatter as [danger-swiftlint](https://github.com/ashfurrow/danger-swiftlint) does and format it (with links to the files in their repos for example).

We'd need to make danger run in the end though. Or we could have two Dangerfiles - one for PR validation (in this PR), which could be run in any job; and another one that runs after the OSS-check.

marcelofabri:
We can have something like this:

<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%">
          <a href="https://github.com/realm/realm-cocoa">Realm Swift</a>
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td><g-emoji alias="no_entry_sign" fallback-src="https://assets-cdn.github.com/images/icons/emoji/unicode/1f6ab.png">🚫</g-emoji></td>
      <td>
<strong>Colon Rule Violation</strong>: Colons should be next to the identifier when specifying a type and next to the key in dictionary literals.  <br>  <a href="https://github.com/realm/realm-cocoa/blob/3c50fab2ade055d10e288688712f6fea557ba973/RealmSwift/Tests/ObjectCreationTests.swift#L463">RealmSwift/Tests/ObjectCreationTests.swift#L463</a>
</td>
    </tr>
  </tbody>
</table>

Unfortunately, checkboxes don't work inside tables, so we'd need another way to mark a violation as legit.

Actually, even if it did danger would delete it on a new build 😬

jpsim:
How hard would it be to build our own auto-updating GitHub comment reporting, which would actually allow the format we want? It doesn't seem too challenging, but I may be ignorant of the complexity involved...

marcelofabri:
Danger does support any markdown text. Or are you thinking about something else?

jpsim:
Ah, I see. I thought it was necessary to use the table format. But that still doesn't help us avoid overwriting "approved" violations.

marcelofabri:
Yeah, we could get away with that by using GitHub API and applying our own logic in `Dangerfile` though.

jpsim:
I've created @SwiftLintBot, a new token for that account and added it in Travis's `DANGER_GITHUB_API_TOKEN` env vars. Then I reran the SPM build that had previously failed, which has now succeeded, but I was expecting Danger to post a comment, which it did not. Although maybe that's normal because none of the code paths causing `fail` or `warn` in the Dangerfile were hit, so Danger doesn't comment in those cases?

marcelofabri:
Yes, that's the expected behavior. You can run it locally against a PR that would fail to check:

```
danger pr https://github.com/realm/SwiftLint/pull/1032
```

Dangerfile
```diff
@@ -0,0 +1,26 @@
...
+has_app_changes = !git.modified_files.grep(/Source/).empty?
+has_test_changes = !git.modified_files.grep(/Tests/).empty?
+
+# Add a CHANGELOG entry for app changes
+if !git.modified_files.include?("CHANGELOG.md") && has_app_changes
+  fail("Please include a CHANGELOG entry to credit yourself! \nYou can find it at [CHANGELOG.md](https://github.com/realm/SwiftLint/blob/master/CHANGELOG.md).")
+    markdown <<-MARKDOWN
+Here's an example of your CHANGELOG entry:
+```markdown
+* #{github.pr_title}#{' '}
```
jpsim: Why not this?

```diff
-* #{github.pr_title}#{' '}
-[#{github.pr_author}](https://github.com/#{github.pr_author})
-[#issue_number](https://github.com/realm/SwiftLint/issues/issue_number)
+* #{github.pr_title}  
+  [#{github.pr_author}](https://github.com/#{github.pr_author})
+  [#issue_number](https://github.com/realm/SwiftLint/issues/issue_number)
```

.travis.yml
```diff
@@ -13,6 +13,7 @@ matrix:
       env: JOB=SPM
       os: osx
       osx_image: xcode8.2
+      before_script: bundle install --jobs=3 --retry=3 --without documentation --path bundle && bundle exec danger
```
jpsim: maybe break this up into a YAML array for legibility?

```yaml
before_script:
  - bundle install --jobs=3 --retry=3 --without documentation --path bundle
  - bundle exec danger
```

Dangerfile
marcelofabri: because I shamelessly copied this from CocoaPods' Dangerfile 😂

.travis.yml
marcelofabri: 👍

Dangerfile
marcelofabri: I've kept the `#{' '}` to make sure we don't miss this if someone uses an editor that strips trailing whitespaces.

jpsim: that's just one space not two, though?

jpsim:
We should probably add `.bundle/` and `bundle/` to `.gitignore`?

Dangerfile
acecilia: Just curious, why the two spaces at the end of the `github.pr_title` line?

jpsim: To force a newline.

jpsim: It was taken from CocoaPods.

PR #1040, opened 2016-12-23
Add class_delegate_protocol rule

### LINKED ISSUE #1039: Rule Request: delegate protocols should be reference types
Based on #972.

Protocol declarations that end with `Delegate`should be reference types (either by using `: class` or marking as `@objc`).

marcelofabri:
Fixes #1039 

This could flag a valid declaration when the protocol inherits form another protocol that is `: class` or `@objc`. However, I feel like this should be default as it should be rare enough and you can always mark the child protocol as `: class` as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@
 * Add more helpful reason strings to TrailingCommaRule.  
   [Matt Rubin](https://github.com/mattrubin)
 
+* Add `class_delegate_protocol` rule that warns against protocol declarations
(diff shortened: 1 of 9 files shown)


jpsim:
Perhaps we could make an exception if the protocol inherits from another protocol matching `(NS|UI)\w+Delegate`? Since that covers the vast majority of the class protocols people would inherit from?

marcelofabri:
Makes sense. It'd say that we could event just check if it ends with `Delegate` as well.

PR #1041, opened 2016-12-23
`syntactic_sugar` rule now doesn't flag declarations that can't be fixed

### LINKED ISSUE #928: syntactic_sugar flags var declarations that can't be fixed
I have this line:

```swift
fileprivate var currentIndex: Array<OnboardingPage>.Index?
```

because I want a variable of the same type as the index of an array of `OnboardingPage`s. However, if I change it to `[OnboardingPage].Index?`, I get this from Xcode:

```
path/to/OnboardingViewController.swift:35:51: error: consecutive declarations on a line must be separated by ';'
    fileprivate var currentIndex: [OnboardingPage].Index?
                                                  ^
                                                  ;
path/to/OnboardingViewController.swift:35:51: error: expected declaration
    fileprivate var currentIndex: [OnboardingPage].Index?
                                                  ^
path/to/OnboardingViewController.swift:13:13: note: in declaration of 'OnboardingViewController'
final class OnboardingViewController: UIViewController {
```

I suppose this should also be considered a Swift bug, but SwiftLint shouldn't report issues that can't be fixed. For now, I'll use `// swiftlint:disable:next syntactic_sugar`

marcelofabri:
Fixes #928

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@
 * Ignore close parentheses on `vertical_parameter_alignment` rule.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1042](https://github.com/realm/SwiftLint/issues/1042)
+  
+* `syntactic_sugar` rule now doesn't flag declarations that can't be fixed.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#928](https://github.com/realm/SwiftLint/issues/928)
 
 ## 0.15.0: Hand Washable Holiday Linens 🎄
 
--- a/Source/SwiftLintFramework/Rules/SyntacticSugarRule.swift
+++ b/Source/SwiftLintFramework/Rules/SyntacticSugarRule.swift
@@ -26,7 +26,9 @@ public struct SyntacticSugarRule: Rule, ConfigurationProviderRule {
             "let x: Int!",
             "extension Array { \n func x() { } \n }",
             "extension Dictionary { \n func x() { } \n }",
-            "let x: CustomArray<String>"
+            "let x: CustomArray<String>",
+            "var currentIndex: Array<OnboardingPage>.Index?",
+            "func x(a: [Int], b: Int) -> Array<Int>.Index"
         ],
         triggeringExamples: [
             "let x: ↓Array<String>",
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -39,6 +39,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1005](https://github.com/realm/SwiftLint/issues/1005)
 
+* `syntactic_sugar` rule now doesn't flag declarations that can't be fixed.
```
jpsim: more of a bug fix, wouldn't you say?

marcelofabri: yeah, I think so. I'll update it.

jpsim:
You could probably also fix the false positives reported from using `Optional<T>.self`, such as

```swift
unsafeBitCast(nonOptionalT, to: Optional<T>.self)
```

There are quite a few of these false positives in Realm: https://github.com/realm/realm-cocoa/blob/v2.1.2/RealmSwift/LinkingObjects.swift#L160

Source/SwiftLintFramework/Rules/SyntacticSugarRule.swift
```diff
@@ -35,20 +37,38 @@ public struct SyntacticSugarRule: Rule, ConfigurationProviderRule {
...
     public func validateFile(_ file: File) -> [StyleViolation] {
         let types = ["Optional", "ImplicitlyUnwrappedOptional", "Array", "Dictionary"]
-
         let pattern = "\\b(" + types.joined(separator: "|") + ")\\s*<.*?>"
+        let kinds = SyntaxKind.commentAndStringKinds()
+        let contents = file.contents.bridge()
+
+        return file.matchPattern(pattern, excludingSyntaxKinds: kinds).flatMap { range in
+
+            // avoids trigering when referring to an associatedtype
```
jpsim: "avoids trigering" -> "avoid triggering"

marcelofabri:
> You could probably also fix the false positives reported from using `Optional<T>.self`

I've handled them in 0bd1647, but I don't know if `.self` is the only case that should be handled.

jpsim:
There's one more false positive type in Realm: https://github.com/realm/realm-cocoa/blob/v2.1.2/RealmSwift/Object.swift#L331

```swift
if type is Optional<String>.Type
```

Can be fixed with:

```diff
-                if file.matchPattern("\\s*\\.self", withSyntaxKinds: [.keyword],
-                                     range: restOfFileRange).first?.location == start {
+                if let match = file.matchPattern("\\s*\\.(?:self|Type)", range: restOfFileRange).first,
+                    match.0.location == start, match.1 == [.keyword] || match.1 == [.identifier] {
                     return nil
```

jpsim:
🎉

PR #1044, opened 2016-12-23
Use HTML template for reporter

### LINKED ISSUE #1024: Use an HTML template file rather than large strings in Swift source
The HTML reporter has this wall of strings in an array:

image

Which is unreadable, challenging to maintain, and we need to fight the Swift compiler to avoid type inference fits.

We should just move this template to a file and do string replacement to inject our dynamic content. No need for fancy dependencies like a templating library, our needs are simple enough for straight up string replacement.

marcelofabri:
Fixes #1024

### DIFF
--- a/Source/SwiftLintFramework/Reporters/HTMLReporter.swift
+++ b/Source/SwiftLintFramework/Reporters/HTMLReporter.swift
@@ -65,14 +65,20 @@ public struct HTMLReporter: Reporter {
     private static func loadTemplate() -> String {
         let path = templatePath()
 
-        // swiftlint:disable:next force_try
-        return try! String(contentsOfFile: path)
+        #if os(Linux)
+            // swiftlint:disable:next force_try
+            let data = try! Data(contentsOf: URL(fileURLWithPath: path))
+            return String(data: data, encoding: .utf8)!
+        #else
+            // swiftlint:disable:next force_try
+            return try! String(contentsOfFile: path)
+        #endif
     }
 
     private static func templatePath() -> String {
         #if SWIFT_PACKAGE
             let components = Array(NSString(string: #file).pathComponents.dropLast() + ["Templates", "template.html"])
-            let path = NSString.path(withComponents: components)
+            let path = components.joined(separator: "/")
         #else
             // swiftlint:disable:next nesting
             class DummyClass { }
--- a/Source/SwiftLintFramework/Reporters/HTMLReporter.swift
+++ b/Source/SwiftLintFramework/Reporters/HTMLReporter.swift
@@ -30,144 +30,76 @@ public struct HTMLReporter: Reporter {
                               dateString: formatter.string(from: Date()))
     }
 
-    // swiftlint:disable:next function_body_length
     internal static func generateReport(_ violations: [StyleViolation], swiftlintVersion: String,
                                         dateString: String) -> String {
         let rows = violations.enumerated().reduce("") { rows, indexAndViolation in
-            return rows + generateSingleRow(for: indexAndViolation.1, at: indexAndViolation.0 + 1)
+            return rows + generateSingleRow(for: indexAndViolation.1, at: indexAndViolation.0 + 1,
+                                            initialIndentation: 16)
         }
 
         let fileCount = Set(violations.flatMap({ $0.location.file })).count
         let warningCount = violations.filter({ $0.severity == .warning }).count
         let errorCount = violations.filter({ $0.severity == .error }).count
 
-        return [
-            "<!doctype html>\n",
-            "<html>\n",
-            "\t<head>\n",
-            "\t\t<meta charset=\"utf-8\" />\n",
-            "\t\t<meta name=\"viewport\" content=\"width=device-width, initial-scale=1.0\" />\n",
-            "\t\t\n",
-            "\t\t<style type=\"text/css\">\n",
-            "\t\t\tbody {\n",
-            "\t\t\t\tfont-family: Arial, Helvetica, sans-serif;\n",
-            "\t\t\t\tfont-size: 0.9rem;\n",
-            "\t\t\t}\n",
-            "\t\t\t\n",
-            "\t\t\ttable {\n",
-            "\t\t\t\tborder: 1px solid gray;\n",
-            "\t\t\t\tborder-collapse: collapse;\n",
-            "\t\t\t\t-moz-box-shadow: 3px 3px 4px #AAA;\n",
-            "\t\t\t\t-webkit-box-shadow: 3px 3px 4px #AAA;\n",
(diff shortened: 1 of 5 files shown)


Source/SwiftLintFramework/Reporters/HTMLReporter.swift
```diff
@@ -30,144 +30,82 @@ public struct HTMLReporter: Reporter {
...
+            return try! String(contentsOfFile: path)
+        #endif
+    }
+
+    private static func templatePath() -> String {
+        #if SWIFT_PACKAGE
+            let components = Array(NSString(string: #file).pathComponents.dropLast() + ["Templates", "template.html"])
+            let path = components.joined(separator: "/")
+        #else
+            // swiftlint:disable:next nesting
```
jpsim: Could make a `Bundle.SwiftLintFramework` public extension and re-use from [`VersionCommand.swift`](https://github.com/realm/SwiftLint/blob/0.15.0/Source/swiftlint/Commands/VersionCommand.swift#L13) instead?

```swift
extension Bundle {
    public static let SwiftLintFramework = Bundle(identifier: "io.realm.SwiftLintFramework")!
}
```

I think that would allow combining the SwiftPM & Darwin code paths too... 🤔

Source/SwiftLintFramework/Reporters/HTMLReporter.swift
```diff
@@ -30,144 +30,82 @@ public struct HTMLReporter: Reporter {
...
+    private static func generateReport(parameters: [String: String]) -> String {
+        var report = loadTemplate()
+        for (key, value) in parameters {
+            report = report.replacingOccurrences(of: "$$\(key)$$", with: value)
+        }
+
+        return report
+    }
+
+    private static func loadTemplate() -> String {
```
jpsim: This should be a global `let` constant so it's just resolved once on first access. Useful if using SwiftLintFramework and invoking the HTML reporter multiple times.

Source/SwiftLintFramework/Reporters/HTMLReporter.swift
```diff
@@ -30,144 +30,82 @@ public struct HTMLReporter: Reporter {
...
 
-    private static func generateSingleRow(for violation: StyleViolation, at index: Int) -> String {
+    private static func generateSingleRow(for violation: StyleViolation, at index: Int,
+                                          initialIndentation: Int) -> String {
         let severity: String = violation.severity.rawValue.capitalized
         let location = violation.location
         let file: String = (violation.location.relativeFile ?? "<nopath>").escapedForXML()
         let line: Int = location.line ?? 0
         let character: Int = location.character ?? 0
+        let indentation = String(repeating: " ", count: initialIndentation)
```
jpsim: was anyone counting on this being tabs?

marcelofabri: I don't think so, but I had to change the test stub too.

Source/SwiftLintFramework/Reporters/HTMLReporter.swift
marcelofabri: Sounds good!

Source/SwiftLintFramework/Reporters/HTMLReporter.swift
jpsim: Probably fine.

Source/SwiftLintFramework/Reporters/HTMLReporter.swift
marcelofabri: No luck in combining the paths, as it returns `nil` when running with SPM, probably because it doesn't use dynamic frameworks AFAIK.

Do you still think it's worth to get the bundle from identifier when not using SPM?

jpsim: Thought process:

"Getting bundles on Linux can't be broken! We use it to print the version and that surely works"

[goes to confirm]

```
.build/debug/swiftlint version
fatal error: unexpectedly found nil while unwrapping an Optional value
Current stack trace:
0    libswiftCore.so                    0x00007fc4f9554fc0 swift_reportError + 117
1    libswiftCore.so                    0x00007fc4f9567df0 _swift_stdlib_reportFatalError + 63
2    libswiftCore.so                    0x00007fc4f93756a3 <unavailable> + 0
3    libswiftCore.so                    0x00007fc4f94d2dfd <unavailable> + 0
4    libswiftCore.so                    0x00007fc4f93756a3 <unavailable> + 0
5    libswiftCore.so                    0x00007fc4f94946b0 specialized _fatalErrorMessage(StaticString, StaticString, StaticString, UInt, flags : UInt32) -> Never + 96
6    swiftlint                          0x000000000042c086 <unavailable> + 0
7    libpthread.so.0                    0x00007fc4f858cae9 <unavailable> + 0
8    libswiftCore.so                    0x00007fc4f9565b20 swift_once + 90
9    swiftlint                          0x000000000042c16a <unavailable> + 0
10   swiftlint                          0x000000000042c1e9 <unavailable> + 0
11   swiftlint                          0x000000000042c6b6 <unavailable> + 0
12   swiftlint                          0x0000000000447285 <unavailable> + 0
13   swiftlint                          0x000000000044b0c1 <unavailable> + 0
14   swiftlint                          0x000000000044865f <unavailable> + 0
15   swiftlint                          0x0000000000448fc4 <unavailable> + 0
16   swiftlint                          0x0000000000448ac1 <unavailable> + 0
17   swiftlint                          0x0000000000416b91 <unavailable> + 0
18   swiftlint                          0x0000000000416cd7 <unavailable> + 0
19   libdispatch.so                     0x00007fc4f974c2a7 <unavailable> + 0
20   libdispatch.so                     0x00007fc4f975b152 <unavailable> + 0
21   libdispatch.so                     0x00007fc4f978361f <unavailable> + 0
22   libpthread.so.0                    0x00007fc4f858570a <unavailable> + 0
23   libc.so.6                          0x00007fc4f73de7c0 clone + 109
Illegal instruction
```

💥

marcelofabri: 😂

marcelofabri: Actually it's not in Linux that is broken, it's when using SPM.

jpsim:
I guess this can't work at all with SwiftPM actually... There's nothing that's telling `swift build` to make a bundle and to include the template (or `Info.plist` for that matter) in distributions...

marcelofabri:
Yeah, that makes sense. We should revisit this once SPM adds support for bundle: https://bugs.swift.org/browse/SR-2866.

jpsim:
Thanks for the effort on this nonetheless, @marcelofabri!

PR #1046, opened 2016-12-23
add `substructure` member to Dictionary+SwiftLint.swift

jpsim:
Kind of insane that it took us this long to do this, given the amount of times we've collectively had to write duplicate code that did this.


marcelofabri:
👏

Issue #1038, opened 2016-12-23
Rule Request: Educated Sorting


jpsim:
There area few arrays in SwiftLint's own source that are alphabetically sorted, like [masterRuleList](https://github.com/realm/SwiftLint/blob/0.15.0/Source/SwiftLintFramework/Models/MasterRuleList.swift#L43-L108) and the [Linux `allTests` extensions](https://github.com/realm/SwiftLint/blob/0.15.0/Tests/SwiftLintFrameworkTests/RulesTests.swift#L347-L409).

The same thing could apply to enum members, dictionaries, constants, etc.

For example, SourceKitten's [SwiftDeclarationKind](https://github.com/jpsim/SourceKitten/blob/0.16.0/Source/SourceKittenFramework/SwiftDeclarationKind.swift) enum members are all sorted.

It'd be awesome if SwiftLint could automatically infer which sets of declarations were _mostly_ sorted and trigger a warning.

For example, if an array of 10 items only has 2 out-of-order elements, odds are that these are just mis-ordered. This likelihood goes up as the number of total items increases and the number of out-of-order elements decreases.

This could be configurable by minimum thresholds, and probably be made correctable in simple cases.

JamieEdge:
Hey, so I've started work on a rule for this (see the related commit above), however when I run unit tests (adding the below code to `RulesTests.swift`), I'm finding the values that are sent in the dictionary in the method call `validateFile(file:kind:dictionary:)` to be very different to 'real' ones.

```
    func testEducatedSorting() {
        verifyRule(EducatedSortingRule.description)
    }
```

Any idea why the body offset and body length come through as out of range values during unit testing? I may just be misunderstanding the purpose of these variables.

```
(lldb) po file.contents.characters.count
59

(lldb) po bodyOffset
93

(lldb) po bodyLength
27
```

jpsim:
The offsets and lengths returned from SourceKit are all in bytes. For multi-byte characters in Strings (e.g. non-ASCII such as emoji), `characters.count != bytes`.

SourceKitten has a number of conversion methods to help with this, most of which are in [`String+SourceKitten.swift`](https://github.com/jpsim/SourceKitten/blob/0.16.0/Source/SourceKittenFramework/String+SourceKitten.swift). Such as [`String.byteRangetoNSRange(start:length:)`](https://github.com/jpsim/SourceKitten/blob/0.16.0/Source/SourceKittenFramework/String+SourceKitten.swift#L268-L276).

[...]

PR #1050, opened 2016-12-24
Add inferred_sorting rule

JamieEdge:
This branch adds the `inferred_sorting` rule for arrays (https://github.com/realm/SwiftLint/issues/1038). It doesn't support auto-correction (yet) as this needs a bit more work on it to preserve original formatting and be efficient. It is opt-in due to the amount of issues that it could flag on an existing codebase.

As the PR also fixes all violations, I kept the actual changes in a parent branch, so you can see a cleaner diff: https://github.com/realm/SwiftLint/compare/master...JamieEdge:educated-sorting-rule

Configuration options:
* threshold: The amount that the array already needs to be sorted before it can cause a violation. Defaults to 0.25 (25%).
* min_items: The minimum number of items in the array before it can cause a violation. Defaults to `3`.
* case_sensitive: Determines if case should be considered when sorting array items. Defaults to `true`.


marcelofabri:
I haven't taken a deep look into this yet, but the number of ` // swiftlint:disable educated_sorting` that we needed in SwiftLint codebase might indicate that this rule might need higher default values.

Also, the build is failing on Linux because of an implicit conversion of `String` to `NSString`, as `substringWithByteRange` is a method on `NSString`. Use `contents.bridge()` to convert to an `NSString`.

jpsim:
Yeah when I was picturing this rule, I was thinking of a default min length threshold of 8 or 10.

JamieEdge:
Sure, I'll increase the minimum length and fix the Linux build issue. Thanks for pointing these things out.

On the point about how many disables there are: a lot of these weren't actually required. I just placed them on all of the rule descriptions as they intentionally aren't ordered alphabetically. Would you prefer for me to remove those and just place it where a violation occurs?

jpsim:
I'd prefer the rule be structured so that false positives be very rare. This rule isn't stylistic or enforcing a convention, but rather helping prevent small mistakes.

Also would prefer the name "Inferred Sorting" rather than "Educated Sorting".

JamieEdge:
Sure, so if I adjust the rule to 8+ items and a 80%+ 'sortedness' would that be ideal? I did originally use higher values but for testing lowered the values as I couldn't sort my 4 item arrays (and just wasn't sure where to put the values back to).

jpsim:
> Sure, so if I adjust the rule to 8+ items and a 80%+ 'sortedness' would that be ideal?

That seems right, but I'd want to double check that we don't get any false positives with a handful known open source projects.

JamieEdge:
All requested changes are now made to this branch. I went with 10+ items and 75% threshold based upon the SwiftLint source code.

A workaround I had to put in place was for Swift bug [SR-530](https://bugs.swift.org/browse/SR-530) where String comparison (and therefore sorting) isn't the same across Darwin/Linux.

JamieEdge:
Hi @jpsim @marcelofabri, just chasing this PR to see if there is anything you want me to fix generally to save merging conflicts too many times?

jpsim:
@JamieEdge could you please rebase this on top of the latest `master`? Ideally removing the merge commits too. I'd love to see OSSCheck run on this so we can see if this triggers any false positives, then I can take a closer look to review. Thanks for your continued work on this!

marcelofabri:
Continued on #1655.

PR #1054, opened 2016-12-24
redundant_nil_coalescing rule is now correctable

marcelofabri:


Source/SwiftLintFramework/Rules/RedundantNilCoalescingRule.swift
```diff
@@ -30,10 +30,15 @@ public struct RedundantNilCoalescingRule: OptInRule, ConfigurationProviderRule {
...
-            "var myVar: Int?; myVar ?? 0"
+            "var myVar: Int?; myVar ?? 0\n"
         ],
         triggeringExamples: [
-            "var myVar = nil; myVar ↓?? nil"
+            "var myVar: Int? = nil; myVar↓ ?? nil\n",
+            "var myVar: Int? = nil; myVar↓??nil\n"
+        ],
+        corrections: [
+            "var myVar Int? = nil; let foo = myVar↓ ?? nil\n": "var myVar Int? = nil; let foo = myVar\n",
```
jpsim: Missing colons on this line and the one below it. Should be `var myVar: Int?`

jpsim:
Looks good other than my one comment.

PR #1056, opened 2016-12-24
Implement OSSCheck on CI

jpsim:
Work in progress to address #1025.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
       env: JOB=SPM
       os: osx
       osx_image: xcode8.2
-      before_script:
+      after_success:
         - bundle install --jobs=3 --retry=3 --without documentation --path bundle
         - bundle exec danger
     - script: make docker_test
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ SwiftLint.pkg
 SwiftLintFramework.framework.zip
 benchmark_*
 portable_swiftlint.zip
+osscheck/
 
 # SPM
 .build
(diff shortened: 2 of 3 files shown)


jpsim:
@marcelofabri what do you think of this as a first version of OSSCheck?

marcelofabri:
Are you Santa? 🎅🎁 

This is an awesome first step. I'd only prefer that we did the building and running on another script and then Danger would only be used to report the differences. Does that make sense? This way we could easily run OSSCheck locally as well.

marcelofabri:
Also using the JSON report might be better than the Xcode one.

jpsim:
> I'd only prefer that we did the building and running on another script and then Danger would only be used to report the differences. Does that make sense? This way we could easily run OSSCheck locally as well.

That's how I had first prototyped this, but the fact that this mangles the local git environment (by checking out `master`, pulling, and updating submodules) meant that I wanted to keep that to a CI-only script for now.

> Also using the JSON reporter might be better than the Xcode one.

Not with the way this is currently built, since it relies on each violation being in a single line, doing a set difference on the array of lines to report fixed/introduced violations.

> Are you Santa? 🎅🎁

Yes. 💃

.travis.yml
```diff
@@ -1,25 +1,13 @@
 language: generic
 matrix:
   include:
-    - script:
```
marcelofabri: commenting just as a reminder that this should be added again before merging

jpsim:
I'd say this is a fine starting point. Let's refine and improve over time. 😄

marcelofabri:
🎉

PR #1057, opened 2016-12-25
Make empty_parentheses_with_trailing_closure correctable

marcelofabri:


jpsim:
> This PR may need tests.

Hmm... I wonder if we can "teach" Danger that some of our tests are actually defined in `RuleDescription` 🤔

marcelofabri:
> Hmm... I wonder if we can "teach" Danger that some of our tests are actually defined in RuleDescription 🤔
Well, we still need to add them in Tests when adding the rule 🤔

jpsim:
Yes, but only for new rules.

marcelofabri:
We can check if the modified files are just existing rules and skip the warning in this case.

PR #1067, opened 2016-12-25
Add object_literal rule

### LINKED ISSUE #1060: Rule request: prefer object literals
Prefer image and color literals over `UIImage(named:)` and `UIColor` inits.

marcelofabri:
Fixes #1060

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,12 @@
   [Rafael Machado](https://github.com/rakaramos/)
   [#1089](https://github.com/realm/SwiftLint/issues/1089)
 
+* Add `object_literal` opt-in rule that warns agains using image and color
+  inits that can be replaced for `#imageLiteral` or `#colorLiteral` in
+  Swift 3.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#1060](https://github.com/realm/SwiftLint/issues/1060)
+
 ##### Bug Fixes
 
 * Ignore close parentheses on `vertical_parameter_alignment` rule.  
--- a/Source/SwiftLintFramework/Rules/ObjectLiteralRule.swift
+++ b/Source/SwiftLintFramework/Rules/ObjectLiteralRule.swift
@@ -93,11 +93,8 @@ public struct ObjectLiteralRule: ASTRule, ConfigurationProviderRule, OptInRule {
     }
 
     private func validateColorKinds(arguments: [[String: SourceKitRepresentable]], file: File) -> Bool {
-        for dictionary in arguments {
-            let kinds = kindsFor(dictionary, file: file)
-            if kinds != Set(arrayLiteral: .number) {
-                return false
-            }
+        for dictionary in arguments where kindsFor(dictionary, file: file) != [.number] {
+            return false
         }
 
         return true
--- a/Source/SwiftLintFramework/Rules/ObjectLiteralRule.swift
+++ b/Source/SwiftLintFramework/Rules/ObjectLiteralRule.swift
@@ -24,15 +24,20 @@ public struct ObjectLiteralRule: ASTRule, ConfigurationProviderRule, OptInRule {
             "let color = #colorLiteral(red: 0.9607843161, green: 0.7058823705, blue: 0.200000003, alpha: 1)",
             "let image = UIImage(named: aVariable)",
             "let image = UIImage(named: \"interpolated \\(variable)\")",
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -33,6 +33,12 @@
 * `empty_parentheses_with_trailing_closure` rule is now correctable.  
   [Marcelo Fabri](https://github.com/marcelofabri)
 
+* Add `object_literal` opt-in rule that warns agains using image and color
+  inits that can be replaced for `#imageLiteral` or `#colorLiteral` in
+  Swift 3.
```
jpsim: missing trailing spaces here

marcelofabri:
Hah, the first PR that #1087 helped us! 🙃

jpsim:
Why should this rule be opt-in? Following the guidelines in the [README](https://github.com/realm/SwiftLint#rules):

> 1. A rule that can have many false positives (e.g. `empty_count`)
> 2. A rule that is too slow
> 3. A rule that is not general consensus or is only useful in some cases
>   (e.g. `force_unwrapping`, `missing_docs`)

As best I can tell, 1 and 2 don't apply. I haven't seen any discussion about 3, but I'm not sure if there are any stylistic arguments _against_ using literals, other than the fact that they're a novel-ish concept and not everyone's accustomed to using them. But if that's the only argument, it's ok for a tool like SwiftLint to be opinionated and encourage the use of cool language features like these.

Source/SwiftLintFramework/Rules/ObjectLiteralRule.swift
```diff
@@ -0,0 +1,115 @@
...
+        return names.flatMap { name in
+            [
+                name,
+                name + ".init"
+            ]
+        }
+    }
+
+    private func validateColorKinds(arguments: [[String: SourceKitRepresentable]], file: File) -> Bool {
+        for dictionary in arguments {
```
jpsim: Consider this?

```swift
for dictionary in arguments where kindsFor(dictionary, file: file) != [.number] {
    return false
}

return true
```

Maybe it's just me, but in general, it seems odd to use the `arrayLiteral` initializers explicitly. They're designed to be used to actually convert array literals.

marcelofabri: IRRC there could be more than one kind in each dictionary, i.e. an operation as `120 / 255.0`

I agree that it's kind of odd to use that initializer. I could change to use the one that receives an array.

jpsim:
> * It's harder to read on everywhere but Xcode
> * It's harder to verify if the value is right (mainly for colors)

You think? Quick GitHub searches for [#imageLiteral](https://github.com/search?utf8=%E2%9C%93&q=imageLiteral&type=Code&ref=searchresults) and [#colorLiteral](https://github.com/search?utf8=%E2%9C%93&q=colorLiteral&type=Code&ref=searchresults) seem very readable to me in source code, and no less hard to validate than their initializer versions.

marcelofabri:
The color one is harder IMO because they use values from 0-1 and we're used to think in terms of 0-255.

jpsim:
I'm still not convinced that this shouldn't be enabled by default, but don't feel super strongly about it. We can always change it later, as going from opt-in to not is less breaking than the other way around.

Source/SwiftLintFramework/Rules/ObjectLiteralRule.swift
marcelofabri: I've just read in that init header: `Do not call this initializer directly.`. Oops 😅

jpsim: `[.number]` in my example is actually a `Set<SyntaxKind>`! Not an `Array<SyntaxKind>`. Because `Set` conforms to `ArrayLiteralConvertible`, it can _literally_ be converted from an an array literal.

marcelofabri:
Yeah, we can add it as opt-in and then revisit it later

marcelofabri:
Also IIRC, this new syntax is also only supported on Swift 3 in projects.

jpsim:
They were added in Xcode 7.1: https://developer.apple.com/swift/blog/?id=33

marcelofabri:
Yeah, but I think they're were a playground-only feature at that time

marcelofabri:
https://github.com/apple/swift-evolution/blob/master/proposals/0039-playgroundliterals.md

PR #1069, opened 2016-12-26
Re-enable testVariableName and testTypeName on Linux

marcelofabri:


Source/SwiftLintFramework/Extensions/CharacterSet+LinuxHack.swift
```diff
@@ -0,0 +1,35 @@
...
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import Foundation
+
+extension CharacterSet {
+    func isSuperset(ofCharactersIn string: String) -> Bool {
+        #if os(Linux)
+            // workaround for https://bugs.swift.org/browse/SR-3485
+            let chars = Set(Array(string.characters))
```
jpsim: why convert to an array, only to convert to a set? Why not `let chars = Set(string.characters)`?

PR #1070, opened 2016-12-26
Add redundant_void_return rule

### LINKED ISSUE #1066: Rule request: redundant Void return
Returning `Void` (or `()`) is redundant.

```swift
func foo() -> Void { }
```

marcelofabri:
Fixes #1066


jpsim:
👍 👏

PR #1071, opened 2016-12-26
Sorted imports

marcelofabri:
Continued from #991

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,7 +135,6 @@
 
 * Add a rule that enforces alphabetical sorting of imports.  
   [Scott Berrevoets](https://github.com/sberrevoets)
-  [#991](https://github.com/realm/SwiftLint/pull/991)
   [#900](https://github.com/realm/SwiftLint/issues/900)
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -38,6 +38,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1066](https://github.com/realm/SwiftLint/issues/1066)
 
+* Add a rule that enforces alphabetical sorting of imports.
```
jpsim: Opt-in

marcelofabri:
I squashed all commits because I thought that was @sberrevoets's idea, but it seems it removed his name from the history. I'm sorry 😰

jpsim:
It's ok. His name will forever be credited in the changelog. The only place that matters :)

sberrevoets:
Believe it or not, I was actually able to not open my laptop for a few days so didn't get a chance to look finish this up. Thanks for taking this over the finish line @marcelofabri!

jpsim:
I applaud your self-restraint

Issue #868, opened 2016-11-07
Keep track of already linted/autocorrected files.


nhgrif:
I'm using SwiftLint with a code base of around 30,000 lines of code (and growing).  We have set up a build phase to run SwiftLint (and SwiftLint autocorrect) on every build.  Some devs on the project are reporting that this takes a significantly long amount of time even when they have not modified any files.

On every run, SwiftLint could keep track of which files it checked or corrected and when.  Then, on a future run, SwiftLint will determine if the .swiftlint.yml or the Swift file has been modified since the last run.  If neither has been modified, SwiftLint will skip the file.

On a version update from SwiftLint, it should clear out these records so the first run of a new version of SwiftLint, it re-checks every file.

scottrhoyt:
This sounds a bit out of scope. If you just don't lint files that have been modified from the last linting, then you will loose any warnings/errors generated by those. Simply running swiftlint twice in a row would therefore always not emit any errors or warnings. You could solve this by tracking only files that have listed successfully, but my point is that I think this gets complicated quickly--particularly if you want to clear the linting status of every project on `swiftlint` configuration changes and upgrades.

There are a few other solutions you can consider:

* If the codebase contains any 3rd party libraries or files/directories that seldom change, consider excluding them in your `.swiftlint.yml`.
* Run `swiftlint lint --benchmark` to generate benchmark files. You can use these files to determine if there are any rules you can disable or files you can exclude to speed up the process.
* Some users have tied in `swiftlint` to a pre-commit hook. Doing this would have the desired result of linting only changed files. However, you lose the feedback in your build stage. This also would not re-lint every file if your configuration changed or `swiftlint` was upgraded. I believe if you search the issues hear you will find some notes on that approach.

nhgrif:
> rules you can disable

Can you elaborate on this?

scottrhoyt:
If you run `swiftlint lint --benchmark`, files will be generated informing you how much time was spent per-rule and per-file. Have a look at the rule benchmarks and see if there are any rules that are taking a significant amount of time. If you decide you want to not run that rule, you can add it to the `disabled_rules` key in your configuration if it is a default rule or remove it from `opt_in_rules` if it is an opt-in rule.

nhgrif:
Will the `--benchmark` command also work with autocorrect?

scottrhoyt:
No. But `autocorrect` generally shouldn't be run automatically anyway.

[...]

Issue #1045, opened 2016-12-23
Crash when printing version when building with Swift Package Manager


jpsim:
"Getting bundles on Linux can't be broken! We use it to print the version and that surely works"

[goes to confirm]

```
.build/debug/swiftlint version
fatal error: unexpectedly found nil while unwrapping an Optional value
Current stack trace:
0    libswiftCore.so                    0x00007fc4f9554fc0 swift_reportError + 117
1    libswiftCore.so                    0x00007fc4f9567df0 _swift_stdlib_reportFatalError + 63
2    libswiftCore.so                    0x00007fc4f93756a3 <unavailable> + 0
3    libswiftCore.so                    0x00007fc4f94d2dfd <unavailable> + 0
4    libswiftCore.so                    0x00007fc4f93756a3 <unavailable> + 0
5    libswiftCore.so                    0x00007fc4f94946b0 specialized _fatalErrorMessage(StaticString, StaticString, StaticString, UInt, flags : UInt32) -> Never + 96
6    swiftlint                          0x000000000042c086 <unavailable> + 0
7    libpthread.so.0                    0x00007fc4f858cae9 <unavailable> + 0
8    libswiftCore.so                    0x00007fc4f9565b20 swift_once + 90
9    swiftlint                          0x000000000042c16a <unavailable> + 0
10   swiftlint                          0x000000000042c1e9 <unavailable> + 0
11   swiftlint                          0x000000000042c6b6 <unavailable> + 0
12   swiftlint                          0x0000000000447285 <unavailable> + 0
13   swiftlint                          0x000000000044b0c1 <unavailable> + 0
14   swiftlint                          0x000000000044865f <unavailable> + 0
15   swiftlint                          0x0000000000448fc4 <unavailable> + 0
16   swiftlint                          0x0000000000448ac1 <unavailable> + 0
17   swiftlint                          0x0000000000416b91 <unavailable> + 0
18   swiftlint                          0x0000000000416cd7 <unavailable> + 0
19   libdispatch.so                     0x00007fc4f974c2a7 <unavailable> + 0
20   libdispatch.so                     0x00007fc4f975b152 <unavailable> + 0
21   libdispatch.so                     0x00007fc4f978361f <unavailable> + 0
22   libpthread.so.0                    0x00007fc4f858570a <unavailable> + 0
23   libc.so.6                          0x00007fc4f73de7c0 clone + 109
Illegal instruction
```

💥 

---

[...]

PR #1073, opened 2016-12-27
Cache implementation

marcelofabri:
Will fix https://github.com/realm/SwiftLint/issues/868

TODO:

- [x] Ignore cache if SwiftLint version has changed
- [x] Ignore cache if configuration has changed
- [x] Expose cache path in CLI and configuration
- [x] Ignore cache flag
- [x] Keep benchmarks when reading from cache
- [x] Tests


Source/swiftlint/Commands/LintCommand.swift
```diff
@@ -40,13 +40,24 @@ struct LintCommand: CommandProtocol {
...
             useSTDIN: options.useSTDIN, quiet: options.quiet,
             useScriptInputFiles: options.useScriptInputFiles) { linter in
             let start: NSDate! = options.benchmark ? NSDate() : nil
             var currentViolations: [StyleViolation] = []
+            var readFromCache = false
+            var fileHash: Int?
             autoreleasepool {
-                if options.benchmark {
+                if let file = linter.file.path,
+                    case let hash = linter.file.contents.hash,
```
marcelofabri: I'm not sure if this is enough or we should use something more sophisticated 🤔

jpsim:
Do you understand why osscheck is saying that some violations were fixed by this PR?

marcelofabri:
> Do you understand why osscheck is saying that some violations were fixed by this PR?

Not really 🙃

marcelofabri:
I've seen the log and that's because there was a crash while getting the version number (https://github.com/realm/SwiftLint/issues/1045)

Source/SwiftLintFramework/Models/Version.swift
```diff
@@ -12,9 +12,11 @@ public struct Version {
...
-        let value = Bundle(identifier: "io.realm.SwiftLintFramework")!
-            .object(forInfoDictionaryKey: "CFBundleShortVersionString")!
-        // swiftlint:disable:next force_cast
-        return Version(value: value as! String)
+        if let value = Bundle(identifier: "io.realm.SwiftLintFramework")?
+            .object(forInfoDictionaryKey: "CFBundleShortVersionString") as? String {
+            return Version(value: value)
+        }
+
+        return Version(value: "0.15.0")
```
marcelofabri: this should be reverted or done in a way that `make set_version` updates it as well

Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -15,7 +15,7 @@ public enum LinterCacheError: Error {
     case differentConfiguration
 }
 
-public struct LinterCache {
+public class LinterCache {
```
marcelofabri: I had to change it to be a class to make sure all `Linter`s used the same reference

marcelofabri:
I think this is ready for review, other than missing changelog entry and README update (and the version workaround)

Source/SwiftLintFramework/Models/Version.swift
marcelofabri: Done in caee985

Source/SwiftLintFramework/Models/ViolationSeverity.swift
```diff
@@ -9,6 +9,10 @@
 public enum ViolationSeverity: String, Comparable {
     case warning
     case error
+
+    public init?(identifier: String) {
```
jpsim: Why this init when `rawValue` does the exact same thing?

Source/SwiftLintFramework/Models/Version.swift
```diff
@@ -0,0 +1,22 @@
+//
+//  Version.swift
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 27/12/16.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import Foundation
```
marcelofabri: 👍 for not reading from the bundle, but I think it'd be better if we kept this namespaced inside a `struct` so we can avoid using `String` everywhere. This would also make it easier to implement version pinning if we want.

jpsim:
Could you please extract the `Version.swift` related changes into a separate PR? It would help reduce the scope of review for these changes.

Source/swiftlint/Extensions/LinterCache+CommandLine.swift
```diff
@@ -0,0 +1,44 @@
...
+        }
+    }
+
+}
+
+private func cacheUrl(options: LintOptions, configuration: Configuration) -> URL? {
+    guard !options.ignoreCache else {
+        return nil
+    }
+    let path = (options.cachePath.isEmpty ? configuration.cachePath : options.cachePath) ?? ".swiftlint_cache.json"
```
jpsim: I think we should be storing the cache in `~/Library/Application Support/SwiftLint` instead of the working directory from which we run `swiftlint`.

script/Version.swift.template
```diff
@@ -0,0 +1,15 @@
+//
+//  Version.swift
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 27/12/16.
```
jpsim: 📆 🇺🇸 🔥

Source/SwiftLintFramework/Models/Version.swift
```diff
@@ -0,0 +1,15 @@
+//
+//  Version.swift
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 27/12/16.
```
jpsim: 📆 🇺🇸 🔥

Source/SwiftLintFramework/Models/Version.swift
```diff
@@ -0,0 +1,15 @@
...
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+public struct Version {
+    public let value: String
+
+    public static let current: Version = {
```
jpsim: Can't you do `public static let current = Version(value: "0.15.0")`?

Source/swiftlint/Commands/LintCommand.swift
```diff
@@ -137,5 +143,9 @@ struct LintOptions: OptionsProtocol {
             <*> mode <| Option(key: "reporter", defaultValue: "",
                                usage: "the reporter used to log errors and warnings")
             <*> mode <| quietOption(action: "linting")
+            <*> mode <| Option(key: "cache-path", defaultValue: "",
+                               usage: "the cache that should be used when linting")
```
jpsim: "the location of the cache used when linting"

Source/swiftlint/Commands/VersionCommand.swift
```diff
@@ -9,9 +9,9 @@
 import Commandant
 import Foundation
 import Result
+import SwiftLintFramework
 
-private let version = Bundle(identifier: "io.realm.SwiftLintFramework")!
-    .object(forInfoDictionaryKey: "CFBundleShortVersionString")!
+private let version = Version.current.value
```
jpsim: We can just use `Version.current.value` directly in `VersionCommand.run(_:)`, so you can remove this constant.

Source/swiftlint/Extensions/LinterCache+CommandLine.swift
```diff
@@ -0,0 +1,44 @@
...
+
+    func save(options: LintOptions, configuration: Configuration) {
+        if let url = cacheUrl(options: options, configuration: configuration) {
+            try? save(to: url)
+        }
+    }
+
+}
+
+private func cacheUrl(options: LintOptions, configuration: Configuration) -> URL? {
```
jpsim: please keep acronyms capitalized: `cacheURL(...)`

Source/swiftlint/Extensions/LinterCache+CommandLine.swift
marcelofabri: but would it be a global cache?

jpsim: you could still have a single cache per root path.

marcelofabri: Do you have any ideas on how to do this? I've thought about several ways, but each one of them have big disadvantages:

1. Keep an index file where we map root path to a cache file named from an UUID. I think this is my favorite option so far.
2. Use MD5 to generate the cache file name, but it's not easy to do it in Swift.
3. Use `.hashValue` on the root path name to generate the cache name, but I don't know how much we can trust on it.

We also need to have different defaults for Linux and Darwin, since Linux doesn't have `~/Library/Application Support`. It doesn't even implement `FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask)`.

jpsim: We could sanitize the root path string to make sure it's a valid file name.

For example: http://stackoverflow.com/a/39435590/373262

```swift
extension String {
    var stringForFilePath: String {
        // characterSet contains all illegal characters on OS X and Windows
        let characterSet = NSCharacterSet(charactersInString: "\"\\/?<>:*|")
        // replace "-" with character of choice
        return componentsSeparatedByCharactersInSet(characterSet).joinWithSeparator("-")
    }
}
```

marcelofabri: But then wouldn't we open the possibility of having a name with more than [255 characters](http://apple.stackexchange.com/questions/86611/does-os-x-enforce-a-maximum-filename-length-or-character-restriction)?

jpsim: Yes... so maybe we use the hash, and hope we're not super super super super unlucky to get a collision?

marcelofabri: I don't know if it's worth, but we could store the root path and ignore the cache if it changed (i.e. a collision happened)

jpsim: probably not worth worrying about collisions

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -47,6 +48,8 @@ public struct Configuration: Equatable {
     public let rules: [Rule]
     public var rootPath: String?              // the root path to search for nested configurations
     public var configurationPath: String?     // if successfully loaded from a path
+    public var hash: Int?
```
jpsim: Why not make `Configuration` conform to `Hashable`? http://swiftdoc.org/v3.0/protocol/Hashable/

Would require naming this variable as `hashValue`.

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -179,6 +185,7 @@ public struct Configuration: Equatable {
             }
             self.init(dict: dict)!
             configurationPath = fullPath
+            hash = yamlContents.hash
```
jpsim: instead of hashing the yaml contents, we could hash the properties of this struct, meaning that changes to the configuration file like reordering items or adding comments wouldn't invalidate the cache.

jpsim: this would also allow `Configuration` to conform to `Hashable` and for caching to work with configurations created without a YAML source.

Source/SwiftLintFramework/Models/Linter.swift
```diff
@@ -78,6 +84,11 @@ public struct Linter {
             deprecatedToValidIdentifier[key] = value
         }
 
+        if let cache = cache, let file = file.path {
```
jpsim: `let path = file.path` would disambiguate from the property

Source/SwiftLintFramework/Models/Linter.swift
```diff
@@ -86,8 +97,34 @@ public struct Linter {
         return (violations, ruleTimes)
     }
 
-    public init(file: File, configuration: Configuration = Configuration()!) {
+    private func cachedStyleViolations(_ benchmark: Bool = false) -> ([StyleViolation], [(id: String, time: Double)])? {
```
jpsim: `benchmark` should be a required argument label to conform to Swift 3 API Design Guidelines.

Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -0,0 +1,125 @@
...
+        entry["hash"] = hash
+
+        lock.lock()
+        var filesCache = (cache["files"] as? [String: Any]) ?? [:]
+        filesCache[file] = entry
+        cache["files"] = filesCache
+        lock.unlock()
+    }
+
+    public func violations(for file: String, hash: Int) -> [StyleViolation]? {
```
jpsim: weakly typed argument. Label should be `forFile file: String`.

Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -0,0 +1,125 @@
...
+    case differentVersion
+    case differentConfiguration
+}
+
+public final class LinterCache {
+    private var cache: [String: Any]
+    private let lock = NSLock()
+
+    public init(currentVersion: Version = .current, configurationHash: Int? = nil) {
+        cache = [String: Any]()
```
jpsim: why not just this?

```swift
cache = [
    "version": currentVersion.value,
    "configuration_hash: configurationHash,
    "files": [:]
]
```

Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -0,0 +1,125 @@
...
+
+    public convenience init(contentsOf url: URL, currentVersion: Version = .current,
+                            configurationHash: Int? = nil) throws {
+        let data = try Data(contentsOf: url)
+        let json = try JSONSerialization.jsonObject(with: data, options: [])
+        try self.init(cache: json, currentVersion: currentVersion,
+                      configurationHash: configurationHash)
+    }
+
+    public func cacheFile(_ file: String, violations: [StyleViolation], hash: Int) {
```
jpsim: Doesn't conform to Swift 3 API Design Guidelines, and implementation could be simplified:

```swift
public func cache(violations: [StyleViolation], forFile file: String, fileHash: Int) {
    lock.lock()
    var filesCache = (cache["files"] as? [String: Any]) ?? [:]
    filesCache[file] = [
        "violations": violations.map(dictionaryForViolation),
        "hash": fileHash
    ]
    cache["files"] = filesCache
    lock.unlock()
}
```

Adds an insignificant amount of work inside the lock.

Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    public func save(to url: URL) throws {
+        lock.lock()
+        let json = toJSON(cache)
+        lock.unlock()
+        try json.write(to: url, atomically: true, encoding: .utf8)
+    }
+
+    private func dictionaryForViolation(_ violation: StyleViolation) -> [String: Any] {
```
jpsim: Swift 3 API Design Guidelines: `dictionary(for violation: StyleViolation)`

Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -0,0 +1,125 @@
...
+            "severity": violation.severity.rawValue,
+            "type": violation.ruleDescription.name,
+            "rule_id": violation.ruleDescription.identifier,
+            "reason": violation.reason
+        ]
+    }
+}
+
+extension StyleViolation {
+    fileprivate static func fromCache(_ cache: [String: Any], file: String) -> StyleViolation? {
```
jpsim: Swift 3 API Design Guidelines: `from(cache: [String: Any], file: String) -> StyleViolation?`

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -35,6 +35,7 @@ private enum ConfigurationKey: String {
     case swiftlintVersion = "swiftlint_version"
     case useNestedConfigs = "use_nested_configs" // deprecated
     case warningThreshold = "warning_threshold"
+    case cachePath = "cache_path"
```
jpsim: the rest of this list is sorted

Source/SwiftLintFramework/Models/Configuration.swift
marcelofabri: Mainly because I thought it'd be too much work to make `Rule` conform to `Hashable` as well. Do you think it's worth the extra complexity?

jpsim: A `Rule` is uniquely identified by its `identifier`, so its `hashValue` can just be `identifier.hashValue` and this can be implemented as a protocol extension.

marcelofabri: What about its configuration?

jpsim: Yeah, ok thinking more about it, `Hashable` implies `Equatable`, and the protocol semantics require that all user-visible data from two equal types be indistinguishable, so makes sense not to conform to it.

marcelofabri: So, to make sure we're in the same page, do you still think we should make `Configuration` conform to `Hashable`?

jpsim: No.

Source/SwiftLintFramework/Models/Configuration.swift
jpsim: Forget the bit about conforming to `Hashable`, but do you think you could adapt this to use the has value from `dict` at least? Would allow writing comments and reordering YAML contents without throwing away the cache.

jpsim:
What in the world happened here? https://circleci.com/gh/realm/SwiftLint/82

```
Building branch
Cloning Alamofire
Linting 5 iterations of Alamofire with branch: 1..2..3..4..5
Cloning Aerial
Linting 5 iterations of Aerial with branch: 1..2..3..4..5
Cloning SourceKitten
Linting 5 iterations of SourceKitten with branch: 1..2..3..4..5
Cloning Moya
Linting 5 iterations of Moya with branch: 1..2..3..4..5
Cloning realm-cocoa
Linting 5 iterations of realm-cocoa with branch: 1..2..3..4..5
Switched to branch 'master'
From github.com:realm/SwiftLint
   1d2b3fd..480acd7  master     -> origin/master
Building master
<unknown>:0: error: build had 1 command failures
error: exit(1): /Applications/Xcode-8.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swift-build-tool -f /Users/distiller/SwiftLint/.build/debug.yaml
Linting 5 iterations of Alamofire with master: 1..2..3..4..5
Linting 5 iterations of Aerial with master: 1..2..3..4..5
Linting 5 iterations of SourceKitten with master: 1..2..3..4..5
Linting 5 iterations of Moya with master: 1..2..3..4..5
Linting 5 iterations of realm-cocoa with master: 1..2..3..4..5
```

image

marcelofabri:
> What in the world happened here?

I have no idea 😱

Source/SwiftLintFramework/Models/Configuration.swift
marcelofabri: done in c28e308, but had to bridge to `NSDictionary`

jpsim:
Stepping back a bit. Wouldn't we be better off just building the cache at the very end of the lint process, from a single thread, avoiding the need for locking? Having this at the `Linter` level is nice architecturally, but maybe we should split that up into a new `Linter` model that coordinates many `SingleFileLinter`s (aka what is `Linter` today).

jpsim:
LOL. I understand why OSSCheck is reporting massive perf regressions. It somehow can't build `master`, so it's running this PR thinking that it's `master` after things have been cached from the first few `branch` runs. So #1147 is biting us here.

But also, we should disable the cache when running OSSCheck.

marcelofabri:
> Stepping back a bit. Wouldn't we be better off just building the cache at the very end of the lint process, from a single thread, avoiding the need for locking?

Maybe, this implementation started before we had multithread lint. We could measure the lock cost to see if it's worth the refactoring and the "architectural cost".

jpsim:
> Maybe, this implementation started before we had multithread lint. We could measure the lock cost to see if it's worth the refactoring and the "architectural cost".

Let's get this in, and look into avoiding locking later.

Awesome work! 🎉

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -179,6 +185,7 @@ public struct Configuration: Equatable {
             }
             self.init(dict: dict)!
             configurationPath = fullPath
+            hash = dict.bridge().hashValue
```
jpsim: This is returning a value of `3` for Realm. Seems suspicious: https://github.com/realm/realm-cocoa/blob/master/.swiftlint.yml

marcelofabri: maybe we should revert to use the String hash then?

jpsim: Nah, apparently `NSDictionary` hash just returns the count. I've tested `dict.description.hashValue` locally and that seems to work surprisingly well.

jpsim: Works if items are re-ordered, and if comments are added/edited.

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -269,7 +276,8 @@ private func validKeys(ruleList: RuleList) -> [String] {
         .swiftlintVersion,
         .useNestedConfigs,
         .warningThreshold,
-        .whitelistRules
+        .whitelistRules,
+        .cachePath
```
jpsim: This list is also otherwise alphabetized.

Source/SwiftLintFramework/Models/Linter.swift
```diff
@@ -86,8 +97,34 @@ public struct Linter {
...
+            ruleTimes = rules.flatMap { rule in
+                let id = type(of: rule).description.identifier
+                return (id, fractionedTime)
+            }
+        }
+
+        return (cachedViolations, ruleTimes)
+    }
+
+    public init(file: File,
```
jpsim: this declaration doesn't need to be broken up in multiple lines.

Source/swiftlint/Extensions/LinterCache+CommandLine.swift
```diff
@@ -0,0 +1,66 @@
...
+private func cacheURL(options: LintOptions, configuration: Configuration) -> URL? {
+    guard !options.ignoreCache else {
+        return nil
+    }
+
+    let path = options.cachePath.isEmpty ? configuration.cachePath : options.cachePath
+    return path.map(URL.init(fileURLWithPath:)) ?? defaultCacheURL(options: options)
+}
+
+private func defaultCacheURL(options: LintOptions) -> URL {
```
jpsim: Much of this should be moved into SwiftLintFramework IMHO, so that the cache is used even if not going through the SwiftLint CLI.

marcelofabri: It'd be used, you'd just need to setup by yourself since we depend on `LintOptions` to resolve the cache path

jpsim: Only if overriding the default. Wouldn't be much different than the path we pass to Configuration: https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Models/Configuration.swift#L162

marcelofabri: can we do that on a later moment? 😬

jpsim: Sure.

jpsim:
🎉  this is gonna be 'uuuge 🎉

marcelofabri:
Awesomeee 💯

PR #1081, opened 2016-12-29
Fix false positives on closure rules on Swift 2.3

### LINKED ISSUE #1019: False positive on closure_parameter_position with Swift 2.3
Hey,

the following code triggers a parameter position violation for the completion closure :

```
UIView.animateWithDuration(0.3, animations: {
  self.disableInteractionRightView.alpha = 0
  }, completion: { _ in
    self.disableInteractionRightView.hidden = true
})
```

Do you have any idea why ?

marcelofabri:
This should fix #1019 

Maybe we should have a way of running the tests on all Swift versions 🤔

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#928](https://github.com/realm/SwiftLint/issues/928)
 
+* Fix false positives on `closure_parameter_position` and 
+  `unused_closure_parameter` rules on Swift 2.3.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#1019](https://github.com/realm/SwiftLint/issues/1019)
+
 ## 0.15.0: Hand Washable Holiday Linens 🎄
 
 ##### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,7 @@
   [#928](https://github.com/realm/SwiftLint/issues/928)
 
 * Fix false positives on `closure_parameter_position` and 
-  `unused_closure_parameter` rules on Swift 2.3.  
+  `unused_closure_parameter` rules with Swift 2.3.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1019](https://github.com/realm/SwiftLint/issues/1019)
 
--- a/Source/SwiftLintFramework/Extensions/Dictionary+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/Dictionary+SwiftLint.swift
@@ -27,7 +27,13 @@ extension Dictionary where Key: ExpressibleByStringLiteral {
             }
 
             if SwiftDeclarationKind(rawValue: kindString) == .varParameter {
-                return [subDict]
+                let nestedParameters = subDict.enclosedVarParameters
+                // on Swift 2.3, a closure parameter is inside another .varParameter and not inside an .argument
+                if nestedParameters.isEmpty {
+                    return [subDict]
+                } else {
+                    return nestedParameters
+                }
             } else if SwiftExpressionKind(rawValue: kindString) == .argument {
                 return subDict.enclosedVarParameters
             }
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -52,6 +52,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#928](https://github.com/realm/SwiftLint/issues/928)
 
+* Fix false positives on `closure_parameter_position` and 
+  `unused_closure_parameter` rules on Swift 2.3.
```
jpsim: "with Swift 2.x" instead of "on Swift 2.3"

Source/SwiftLintFramework/Extensions/Dictionary+SwiftLint.swift
```diff
@@ -27,7 +27,13 @@ extension Dictionary where Key: ExpressibleByStringLiteral {
             }
 
             if SwiftDeclarationKind(rawValue: kindString) == .varParameter {
-                return [subDict]
+                let nestedParameters = subDict.enclosedVarParameters
```
jpsim: Shouldn't this switch on the current Swift version instead?

Source/SwiftLintFramework/Models/SwiftVersion.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+
+import SourceKittenFramework
+
+enum SwiftVersion {
+    case two
+    case three
+
+    static let current: SwiftVersion = {
+        let file = File(contents: "#sourceLocation()")
```
jpsim: We should propose that a request type be added to SourceKit to get the current Swift version. But this is good for now.

Source/SwiftLintFramework/Extensions/Dictionary+SwiftLint.swift
```diff
@@ -27,7 +27,13 @@ extension Dictionary where Key: ExpressibleByStringLiteral {
...
+                // on Swift 2.3, a closure parameter is inside another .varParameter and not inside an .argument
```
jpsim: Generally "with Swift X.Y", not "on Swift X.Y".

marcelofabri:
@jpsim We can also create another scheme in Xcode changing the value of `TOOLCHAINS` env var (it's already being set to `${TOOLCHAINS}`). In that scheme we can disable the Swift 3 tests.

Besides the integration tests (that I've disabled), the only two currently failing tests are:
* `attributes` rule: there's some history on the PR that added this rule, but basically some attributes were just added on Swift 3
* `number_separator` rule: it seems that with Swift 3, SourceKit reports the negative sign as part of the number but Swift 2.3 doesn't, so it's only a matter of where the violation marker should be.

Source/SwiftLintFramework/Models/SwiftVersion.swift
```diff
@@ -23,4 +23,11 @@ enum SwiftVersion {
 
         fatalError("Unexpected Swift version")
     }()
+
+    var nameKey: String {
```
jpsim: This is now used in just one place, keep it there? This shouldn't be a widely used thing.

marcelofabri: 👍

marcelofabri: Actually this is being used on `enclosedVarParameters` as well 🤔

marcelofabri: But I could use "key.typename" directly there, as it's used inside the switch. What do you think would be better?

jpsim: I think `nameKey` isn't representative of the conditional value it returns. Copying the code inline avoids coming up with an accurate name for it. So either come up with a good name, or skirt around the issue :-)

marcelofabri: I've created a private `nameKey(for version: SwiftVersion)` function where it's really needed. I hope it's a good enough name as it's private 😅

jpsim: Yup!

PR #1085, opened 2016-12-29
Fix adding LICENSE to portable_zip

### LINKED ISSUE #1083: Unable to read license file when installing via CocoaPods
The SwiftLint Podspec is pointing to the license file in the repository, but maybe something about also using a .zip instead of pointing to the repository itself is confusing CocoaPods. Maybe just specify MIT, without pointing to a file?

Podfile:

```ruby
# Uncomment the next line to define a global platform for your project
# platform :ios, '9.0'

target 'Licensator' do
  # Comment the next line if you're not using Swift and don't want to use dynamic frameworks
  use_frameworks!

  pod 'SwiftLint'

end
```

results of `pod install`:

```
Analyzing dependencies
Downloading dependencies
Using SwiftLint (0.15.0)
Generating Pods project
Integrating client project
Sending stats
Pod installation complete! There is 1 dependency from the Podfile and 1 total pod installed.

[!] Unable to read the license file `/Users/zev/Desktop/Licensator/Pods/SwiftLint/LICENSE` for the spec `SwiftLint (0.15.0)`

[!] Unable to read the license file `/Users/zev/Desktop/Licensator/Pods/SwiftLint/LICENSE` for the spec `SwiftLint (0.15.0)`
```

CocoaPods version: 1.1.1

marcelofabri:
Fixes #1083

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ SWIFTLINT_EXECUTABLE=$(BUILT_BUNDLE)/Contents/MacOS/swiftlint
 
 FRAMEWORKS_FOLDER=/Library/Frameworks
 BINARIES_FOLDER=/usr/local/bin
+LICENSE_PATH="$(shell pwd)/LICENSE"
 
 OUTPUT_PACKAGE=SwiftLint.pkg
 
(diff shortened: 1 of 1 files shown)


jpsim:
Yes, thanks! I tried to fix this in 7abbd84, but evidently that wasn't a complete fix.

PR #1086, opened 2016-12-30
Update installation options on README

### LINKED ISSUE #939: Add CocoaPods CLI usage instructions
Support was added in #936, but since we couldn't confirm it worked until a release, I didn't add usage instructions to the readme. This ticket tracks doing so once we've confirmed the distribution approach works.

marcelofabri:
Fixes #939

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,18 +19,34 @@ unacceptable behavior to [info@realm.io](mailto:info@realm.io).
 
 ## Installation
 
-Using [Homebrew](http://brew.sh/):
+### Using [Homebrew](http://brew.sh/):
 
 ```
 brew install swiftlint
 ```
 
+### Using [CocoaPods](https://cocoapods.org):
+
+Simply add the following line to your Podfile:
+
+```ruby
+pod 'SwiftLint'
+```
+
+This will download the SwiftLint binaries and dependencies in Pods/ during your next 
+`pod install` execution and will allow you to invoke it via `${PODS_ROOT}/SwiftLint/swiftlint` 
+in your Script Build Phases.
+
+### Using a pre-built package:
+
 You can also install SwiftLint by downloading `SwiftLint.pkg` from the
 [latest GitHub release](https://github.com/realm/SwiftLint/releases/latest) and
 running it.
 
+### Compiling from source:
+
 You can also build from source by cloning this project and running
-`git submodule update --init --recursive; make install` (Xcode 8.0 or later).
+`git submodule update --init --recursive; make install` (Xcode 8.2 or later).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,18 +19,34 @@ unacceptable behavior to [info@realm.io](mailto:info@realm.io).
...
 
+### Using [CocoaPods](https://cocoapods.org):
+
+Simply add the following line to your Podfile:
+
+```ruby
+pod 'SwiftLint'
+```
+
+This will download the SwiftLint binaries and dependencies in Pods/ during your next
```
jpsim: "Pods/" should be in backticks

README.md
```diff
@@ -19,18 +19,34 @@ unacceptable behavior to [info@realm.io](mailto:info@realm.io).
...
+
 You can also install SwiftLint by downloading `SwiftLint.pkg` from the
 [latest GitHub release](https://github.com/realm/SwiftLint/releases/latest) and
 running it.
 
+### Compiling from source:
+
 You can also build from source by cloning this project and running
-`git submodule update --init --recursive; make install` (Xcode 8.0 or later).
+`git submodule update --init --recursive; make install` (Xcode 8.2 or later).
```
jpsim: We still support compiling with Xcode 8.0, so this shouldn't need to be bumped to 8.2.

README.md
```diff
@@ -19,18 +19,34 @@ unacceptable behavior to [info@realm.io](mailto:info@realm.io).
...
+
+Simply add the following line to your Podfile:
+
+```ruby
+pod 'SwiftLint'
+```
+
+This will download the SwiftLint binaries and dependencies in Pods/ during your next 
+`pod install` execution and will allow you to invoke it via `${PODS_ROOT}/SwiftLint/swiftlint` 
+in your Script Build Phases.
```
jpsim: Perhaps we should also say the following two things:

1. That this is the preferred installation method to install a specific version of SwiftLint to be used, since it supports installing a pinned version rather than simply the latest (which is the case with Homebrew).
2. This will add the SwiftLint binaries, its dependencies' binaries and the Swift binary library distribution to the `Pods/` directory, so checking in this directory to SCM such as git is discouraged.

README.md
marcelofabri: For some reason I thought we had to make changes when Xcode 8.2 was released, but I've taken a quick look and couldn't find anything.

marcelofabri:
> Do we want to add the <details> tags like SwiftGen does?

I don't particularly like that because it's an extra step for people reading the README. It's also harder to compare the options.

PR #1087, opened 2016-12-30
Run tests with Swift 2.3 set for SourceKitten

marcelofabri:
This should help preventing issues like #1019.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,8 +41,8 @@ bootstrap:
 	script/bootstrap
 
 test: clean bootstrap
-	$(BUILD_TOOL) $(XCODEFLAGS) test
 	$(BUILD_TOOL) $(SWIFT_2_XCODEFLAGS) test
+	$(BUILD_TOOL) $(XCODEFLAGS) test
 
 clean:
 	rm -f "$(OUTPUT_PACKAGE)"
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,11 @@ XCODEFLAGS=-workspace 'SwiftLint.xcworkspace' \
 	DSTROOT=$(TEMPORARY_FOLDER) \
 	OTHER_LDFLAGS=-Wl,-headerpad_max_install_names
 
+SWIFT_2_XCODEFLAGS=-workspace 'SwiftLint.xcworkspace' \
+	-scheme 'swiftlint with Swift 2.3' \
+	DSTROOT=$(TEMPORARY_FOLDER) \
+	OTHER_LDFLAGS=-Wl,-headerpad_max_install_names
+
 BUILT_BUNDLE=$(TEMPORARY_FOLDER)/Applications/swiftlint.app
 SWIFTLINTFRAMEWORK_BUNDLE=$(BUILT_BUNDLE)/Contents/Frameworks/SwiftLintFramework.framework
 SWIFTLINT_EXECUTABLE=$(BUILT_BUNDLE)/Contents/MacOS/swiftlint
@@ -37,6 +42,7 @@ bootstrap:
 
 test: clean bootstrap
 	$(BUILD_TOOL) $(XCODEFLAGS) test
+	$(BUILD_TOOL) $(SWIFT_2_XCODEFLAGS) test
 
 clean:
 	rm -f "$(OUTPUT_PACKAGE)"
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -7,6 +7,11 @@ XCODEFLAGS=-workspace 'SwiftLint.xcworkspace' \
 	DSTROOT=$(TEMPORARY_FOLDER) \
 	OTHER_LDFLAGS=-Wl,-headerpad_max_install_names
 
+SWIFT_2_XCODEFLAGS=-workspace 'SwiftLint.xcworkspace' \
```
marcelofabri: I've tried to change as little as possible in the Makefile 😅

marcelofabri:
Also I have the feeling that this might mess with code coverage. Perhaps I should change the order to run the new scheme first?

Source/SwiftLintFramework/Rules/NumberSeparatorExamples.swift
```diff
@@ -0,0 +1,77 @@
...
+                "let foo = \(sign)1000e2",
+                "let foo = \(sign)1__000",
+                "let foo = \(sign)1.0001",
+                "let foo = \(sign)1_000_000.000000_1",
+                "let foo = \(sign)1000000.000000_1"
+            ]
+        }
+    }
+
+    static func triggeringExamples() -> [String] {
```
jpsim: These should be `static let` so they're just calculated once:

* `triggeringExamples`
* `swift2TriggeringExamples`
* `corrections`
* `swift2corrections`

I'd also suggest renaming for consistency:

* `swift2TriggeringExamples`
* `swift3TriggeringExamples`
* `swift2corrections`
* `swift3corrections`

SwiftLint.xcodeproj/xcshareddata/xcschemes/swiftlint.xcscheme
```diff
@@ -38,6 +38,17 @@
                BlueprintName = "SwiftLintFrameworkTests"
                ReferencedContainer = "container:SwiftLint.xcodeproj">
             </BuildableReference>
+            <SkippedTests>
+               <Test
+                  Identifier = "AttributesRuleTests/testSwift2Attributes()">
```
jpsim: No such test exists...

SwiftLint.xcodeproj/xcshareddata/xcschemes/swiftlint.xcscheme
```diff
@@ -38,6 +38,17 @@
...
+               </Test>
+               <Test
+                  Identifier = "RulesTests/testNumberSeparatorWithSwift2()">
```
jpsim: No such test exists...

SwiftLint.xcodeproj/xcshareddata/xcschemes/swiftlint.xcscheme
marcelofabri: they did but I've refactored them 😅 
why do I still think Xcode would do the right thing? 😬

Source/SwiftLintFramework/Rules/NumberSeparatorExamples.swift
marcelofabri: Do you think `AttributesRuleExamples` should be changed to use `static let` as well?

jpsim: Yes.

jpsim: Basically if the value doesn't change across invocations, it should be a static or global let, unless the memory overhead of preserving that value across invocations is too high. Otherwise, that code will always be run even though the result won't change.

marcelofabri: It's just that this is only called by the tests and when creating the `RuleDescription` which is `static` on the rule, so I thought it wasn't worth the memory overhead. But I don't have any strong opinions on this.

jpsim: Hmmm, ok good points, let's keep as static funcs! But the renaming would still be nice.

marcelofabri: Too late 😆

PR #1092, opened 2016-12-30
Ignore computed delegates

### LINKED ISSUE #1089: Don't trigger WeakDelegateRule if it is computed property
In my project i have a factory of AppDelegates, in order to simplify resolving, i've added computed properties, which look like this:

```swift
var normalLaunchAppDelegate: NormalLaunchAppDelegate {
    return resolve()
}
```

I'm pretty sure that in this case WeakDelegateRule shouldn't trigger.

rakaramos:
Fixes #1089

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,6 @@
 
 ##### Enhancements
 
-* Make `weak_delegate` ignore computed properties  
-  [Rafael Machado](https://github.com/rakaramos/)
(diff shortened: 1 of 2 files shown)


marcelofabri:
Thanks @rakaramos for doing this! 🎉🇧🇷

jpsim:
Thanks @rakaramos! Though the `flatMap({ Int($0) })` is unnecessary. I'll remove in an upcoming commit.

PR #1094, opened 2016-12-30
Danger improvements

jpsim:
These were originally done in #1056, but were accidentally removed when I rebased 😅.

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -27,7 +27,12 @@ end
 
 # Run OSSCheck if there were app changes
 if has_app_changes
-  @repos = ['realm/realm-cocoa', 'jpsim/SourceKitten']
+  @repos = [
+    'JohnCoates/Aerial',
+    'jpsim/SourceKitten',
+    'Moya/Moya',
+    'realm/realm-cocoa'
+  ]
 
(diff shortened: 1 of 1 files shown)


Dangerfile
```diff
@@ -58,16 +66,28 @@ if has_app_changes
...
-    branch = File.read("osscheck/branch_reports/#{repo_name}.txt").split(/\n+/).reject { |c| c.empty? }
-    master = File.read("osscheck/master_reports/#{repo_name}.txt").split(/\n+/).reject { |c| c.empty? }
+    @repo_name = repo.partition('/').last
+    def non_empty_lines(path)
+      File.read(path).split(/\n+/).reject { |c| c.empty? }
+    end
+    branch = non_empty_lines("osscheck/branch_reports/#{@repo_name}.txt")
+    master = non_empty_lines("osscheck/master_reports/#{@repo_name}.txt")
+    @repo = repo
+    def convert_to_link(string)
```
jpsim: This is fragile, but should work for now. An improvement would be to use a more structured reporter, as @marcelofabri and mentioned again in https://github.com/realm/SwiftLint/issues/1025#issuecomment-269105939)

PR #1097, opened 2016-12-31
add Benchmark and BenchmarkEntry structs

jpsim:
and generally improve the file/rule benchmarking code. Even though this is more code, I think it's nicer, more object-oriented and easier to reason about.

### DIFF
--- a/Source/swiftlint/Helpers/Benchmark.swift
+++ b/Source/swiftlint/Helpers/Benchmark.swift
@@ -28,7 +28,7 @@ struct Benchmark {
     }
 
     mutating func record(file: File, from start: Date) {
-        record(id: file.path ?? "<no path>", time: -start.timeIntervalSinceNow)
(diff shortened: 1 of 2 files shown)


Source/swiftlint/Helpers/Benchmark.swift
```diff
@@ -7,6 +7,45 @@
...
+        self.name = name
+    }
+
+    // swiftlint:disable:next variable_name
+    mutating func record(id: String, time: Double) {
+        entries.append(BenchmarkEntry(id: id, time: time))
+    }
+
+    mutating func record(file: File, from start: Date) {
+        record(id: file.path ?? "<no path>", time: -start.timeIntervalSinceNow)
```
marcelofabri: the previous code used `<nopath>`, is this an intentional change?

jpsim: oops, no. I wrote it from memory. Will revert.

PR #1101, opened 2016-12-31
Add compiler_protocol_init rule

### LINKED ISSUE #1096: Rule request: avoid calling init(arrayLiteral:)
`ExpressibleByArrayLiteral` protocol declares `init(arrayLiteral elements: Element...)`, but according to the [documentation](https://developer.apple.com/reference/swift/set/1539368-init) it shouldn't be called directly. 

The types that implement this protocol are:
* Array
* ArraySlice
* ContiguousArray
* IndexPath
* NSArray
* NSCountedSet
* NSMutableArray
* NSMutableOrderedSet
* NSMutableSet
* NSOrderedSet
* NSSet
* SBElementArray
* Set

List obtained from https://developer.apple.com/reference/swift/expressiblebyarrayliteral

marcelofabri:
Fixes #1096


Source/SwiftLintFramework/Rules/ExpressibleByArrayLiteralInitRule.swift
```diff
@@ -0,0 +1,76 @@
...
+            let offset = (dictionary["key.offset"] as? Int64).flatMap({ Int($0) }),
+            let length = (dictionary["key.length"] as? Int64).flatMap({ Int($0) }),
+            let range = file.contents.bridge().byteRangeToNSRange(start: offset, length: length) else {
+                return []
+        }
+
+        return [range]
+    }
+
+    private let initCallNames: Set<String> = {
```
aamctustwo: Should this be made `static` as well? It doesn't look like this will ever produce a different `Set`.

jpsim:
There are a handful of other literal syntax protocols that may need to be handled in the same way:

* ExpressibleByNilLiteral
* ExpressibleByBooleanLiteral
* ExpressibleByFloatLiteral
* ExpressibleByIntegerLiteral
* ExpressibleByUnicodeScalarLiteral
* ExpressibleByExtendedGraphemeClusterLiteral
* ExpressibleByStringLiteral
* ExpressibleByStringInterpolation
* ExpressibleByDictionaryLiteral

marcelofabri:
> There are a handful of other literal syntax protocols that may need to be handled in the same way

Done in 7d6600a

jpsim:
Is there a reason this rule shouldn't be called `literal_syntax_init`/`LiteralSyntaxInitRule `?

Also, I didn't actually _confirm_ that the protocols listed in https://github.com/realm/SwiftLint/pull/1101#issuecomment-269927526 have the same remark about their initializers. Did you confirm that it was the case?

Source/SwiftLintFramework/Rules/CompilerProtocolInitRule.swift
```diff
@@ -0,0 +1,156 @@
...
+            StyleViolation(ruleDescription: type(of: self).description,
+                           severity: configuration.severity,
+                           location: Location(file: file, characterOffset: $0.location))
+        }
+    }
+
+    private func violationRangesInFile(_ file: File,
+                                       kind: SwiftExpressionKind,
+                                       dictionary: [String: SourceKitRepresentable]) -> [NSRange] {
+        guard kind == .call,
```
jpsim: There are a few things that could be improved in this function IMO. Especially since many variables are computed in every loop iteration even though they can't change.

Consider incorporating some of these changes:

```diff
         guard kind == .call,
-            let name = dictionary["key.name"] as? String else {
-                return []
+            let name = dictionary["key.name"] as? String,
+            let offset = (dictionary["key.offset"] as? Int64).flatMap({ Int($0) }),
+            let length = (dictionary["key.length"] as? Int64).flatMap({ Int($0) }) else {
+            return []
         }
 
+        let arguments = dictionary.enclosedArguments.flatMap({ $0["key.name"] as? String })
+
         for compilerProtocol in ExpressibleByCompiler.allProtocols {
-            guard compilerProtocol.initCallNames.contains(name),
-                case let arguments = dictionary.enclosedArguments.flatMap({ $0["key.name"] as? String }),
+            if compilerProtocol.initCallNames.contains(name),
                 compilerProtocol.matchArguments(arguments),
-                let offset = (dictionary["key.offset"] as? Int64).flatMap({ Int($0) }),
-                let length = (dictionary["key.length"] as? Int64).flatMap({ Int($0) }),
-                let range = file.contents.bridge().byteRangeToNSRange(start: offset, length: length) else {
-                    continue
+                let range = file.contents.bridge().byteRangeToNSRange(start: offset, length: length) {
+                return [range]
             }
-
-            return [range]
         }
```

Source/SwiftLintFramework/Rules/CompilerProtocolInitRule.swift
```diff
@@ -0,0 +1,156 @@
...
+
+            return [range]
+        }
+
+        return []
+    }
+}
+
+private struct ExpressibleByCompiler {
+    let protocolName: String
```
jpsim: `private`?

Source/SwiftLintFramework/Rules/CompilerProtocolInitRule.swift
```diff
@@ -0,0 +1,156 @@
...
+        initCallNames = Set(types.flatMap { [$0, "\($0).init"] })
+    }
+
+    static let allProtocols = [byArrayLiteral, byNilLiteral, byBooleanLiteral,
+                               byFloatLiteral, byIntegerLiteral, byUnicodeScalarLiteral,
+                               byExtendedGraphemeClusterLiteral, byStringLiteral,
+                               byStringInterpolation, byDictionaryLiteral]
+
+    func matchArguments(_ arguments: [String]) -> Bool {
+        for item in self.arguments {
```
jpsim: I don't feel strongly about this one, so ignore if you don't like:

```swift
return self.arguments.first(where: { $0 == arguments }) != nil
```

Source/SwiftLintFramework/Rules/CompilerProtocolInitRule.swift
marcelofabri: The only one that would actually be computed in every loop would be `arguments` (since the other ones shouldn't really return `nil`). However, computing it earlier actually makes the rule a bit slower because most of times a `.call` won't be one of the forbidden ones, so we'd be computing `arguments` unnecessarily.

Source/SwiftLintFramework/Rules/CompilerProtocolInitRule.swift
marcelofabri: I feel like the procedural style is more readable in this case 😅

marcelofabri:
> Is there a reason this rule shouldn't be called literal_syntax_init/LiteralSyntaxInitRule?

Mainly because `ExpressibleByStringInterpolation` does not contain `Literal` in its name 😅

> Also, I didn't actually confirm that the protocols listed in #1101 (comment) have the same remark about their initializers. Did you confirm that it was the case?

I've checked and some of them contain the note, some don't. IMO that is probably a documentation omission and I don't see why one would call these methods directly.

jpsim:
I had proposed that rule name based on Swift Evolution proposal [SE-0115](https://github.com/apple/swift-evolution/blob/master/proposals/0115-literal-syntax-protocols.md), which refers to these protocols as "Literal Syntax Protocols".

CHANGELOG.md
```diff
@@ -70,6 +70,13 @@
   [JP Simard](https://github.com/jpsim)
   [#1077](https://github.com/realm/SwiftLint/issues/1077)
 
+* Add `compiler_protocol_init` rule that flags usage of initializers 
+  declared in protocols used by compiler such as `ExpressibleByArrayLiteral`
```
jpsim: used by the compiler

Issue #1077, opened 2016-12-28
Parallel linting 


marcelofabri:
See this conversation on Twitter: https://twitter.com/merowing_/status/812939171878858752

![10849975_10152591681202799_3947848605304345534_n](https://cloud.githubusercontent.com/assets/833072/21513042/fac65634-cc9a-11e6-9838-df3ab983dead.jpg)

*Picture for illustration purposes only*

jpsim:
The gains would be much higher if we parallelized linting multiple _files_ at once more than parallelizing _rules_.

marcelofabri:
Closed in #1103 🎉

PR #1103, opened 2016-12-31
Make getting style violations thread-safe and parallelize linting

jpsim:
Addresses #1077. Thread sanitizer seems happy with this when running unit tests and linting a few projects I could find, including the Swift standard library.

Generally speeds up linting by ~30% to ~140% on my 2013 MacBook Pro, though the gains should be higher on more recent CPUs.

* **apple/swift:** 218s to 166s (31% faster)
* **realm/realm-cocoa:** 9.6s to 5.6s (71% faster)
* **Carthage/Carthage:** 96s to 53s (80% faster)
* **jpsim/SourceKitten:** 4.2s to 2.1s (100% faster)
* **realm/SwiftLint:** 6.8s to 3.3s (136% faster)

I had initially added a `--parallel` flag to the lint command, but have since removed that option since I haven't run into any issues with it so far, and couldn't come up with a compelling reason to hide it behind an opt-in flag.

### DIFF
--- a/Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
@@ -60,6 +60,6 @@ extension Array {
                 result.append(jobIndexAndResults)
             }
         }
-        return result.sorted { $0.0 < $1.0 }.flatMap { $0.1 }
+        return result.sorted { $0.0 < $1.0 }.map { $0.1 }
     }
 }
--- a/Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
@@ -6,6 +6,7 @@
 //  Copyright © 2016 Realm. All rights reserved.
 //
 
+import Dispatch
 import Foundation
 
 extension Array where Element: NSTextCheckingResult {
@@ -41,5 +42,45 @@ extension Array {
             dictionary[key] = (dictionary[key] ?? []) + [element]
             return dictionary
         }
+    }
+
+    public func parallelForEach(block: @escaping ((Element) -> Void)) {
+        _ = parallelMap(transform: block)
+    }
+
+    func parallelFlatMap<T>(transform: @escaping ((Element) -> [T])) -> [T] {
+        return parallelMap(transform: transform).flatMap { $0 }
+    }
+
+    func parallelMap<T>(transform: @escaping ((Element) -> T)) -> [T] {
+        let count = self.count
+        let maxConcurrentJobs = ProcessInfo.processInfo.activeProcessorCount
+
+        guard count > 1 && maxConcurrentJobs > 1 else {
+            // skip GCD overhead if we'd only run one at a time anyway
+            return map(transform)
+        }
+
+        var result = [(Int, [T])]()
+        result.reserveCapacity(count)
+        let group = DispatchGroup()
+        let uuid = NSUUID().uuidString
+        let jobCount = Int(ceil(Double(count) / Double(maxConcurrentJobs)))
+
+        let queueLabelPrefix = "io.realm.SwiftLintFramework.map.\(uuid)"
+        let resultAccumulatorQueue = DispatchQueue(label: "\(queueLabelPrefix).resultAccumulator")
+
+        for jobIndex in stride(from: 0, to: count, by: jobCount) {
+            let queue = DispatchQueue(label: "\(queueLabelPrefix).\(jobIndex / jobCount)")
+            queue.async(group: group) {
+                let jobElements = self[jobIndex..<Swift.min(count, jobIndex + jobCount)]
+                let jobIndexAndResults = (jobIndex, jobElements.map(transform))
+                resultAccumulatorQueue.sync {
+                    result.append(jobIndexAndResults)
+                }
+            }
+        }
+        group.wait()
+        return result.sorted { $0.0 < $1.0 }.flatMap { $0.1 }
     }
 }
--- a/Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
@@ -44,43 +44,22 @@ extension Array {
         }
     }
 
-    public func parallelForEach(block: @escaping ((Element) -> Void)) {
-        _ = parallelMap(transform: block)
-    }
-
     func parallelFlatMap<T>(transform: @escaping ((Element) -> [T])) -> [T] {
         return parallelMap(transform: transform).flatMap { $0 }
     }
 
     func parallelMap<T>(transform: @escaping ((Element) -> T)) -> [T] {
-        let count = self.count
-        let maxConcurrentJobs = ProcessInfo.processInfo.activeProcessorCount
-
-        guard count > 1 && maxConcurrentJobs > 1 else {
-            // skip GCD overhead if we'd only run one at a time anyway
-            return map(transform)
-        }
-
-        var result = [(Int, [T])]()
+        var result = [(Int, T)]()
         result.reserveCapacity(count)
-        let group = DispatchGroup()
-        let uuid = NSUUID().uuidString
-        let jobCount = Int(ceil(Double(count) / Double(maxConcurrentJobs)))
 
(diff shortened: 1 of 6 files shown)


Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
```diff
@@ -42,4 +43,44 @@ extension Array {
...
+            return map(transform)
+        }
+
+        var result = [(Int, [T])]()
+        result.reserveCapacity(count)
+        let group = DispatchGroup()
+        let uuid = NSUUID().uuidString
+        let jobCount = Int(ceil(Double(count) / Double(maxConcurrentJobs)))
+
+        let queueLabelPrefix = "io.realm.SwiftLintFramework.map.\(uuid)"
```
krzysztofzablocki: curious whether there is a reason you didn't use `DispatchQueue.concurrentPerform` here?

Source/SwiftLintFramework/Extensions/File+Cache.swift
```diff
@@ -33,35 +33,96 @@ private var syntaxTokensByLinesCache = Cache({ file in file.syntaxTokensByLine()
 internal typealias AssertHandler = () -> Void
 private var assertHandlers = [String: AssertHandler]()
 
-private var _allDeclarationsByType = [String: [String]]()
-private var queueForRebuild = [Structure]()
+private struct RebuildQueue {
+    private let lock = NSLock()
```
marcelofabri: I'm not sure here, but wouldn't using a queue be faster?

Source/swiftlint/Commands/LintCommand.swift
```diff
@@ -16,6 +17,8 @@ struct LintCommand: CommandProtocol {
     let verb = "lint"
     let function = "Print lint warnings and errors (default command)"
 
+    private static let violationsAccumulatorQueue = DispatchQueue(label: "io.realm.swiftlint.violationsAccumulator")
```
marcelofabri: I know probably no one will use multiple linters at once, but shouldn't this queue be per instance?

marcelofabri:
I got an exception on one of my runs:

```
Linting 'SchemaTests.swift' (29/41)
[2]    95692 illegal hardware instruction  
/parallel_swiftlint/swiftlint lint
```

marcelofabri:
And another one:

```
swiftlint(95787,0x7000043ed000) malloc: *** error for object 0x7f8e41915d18: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug
[2]    95787 abort      parallel_swiftlint/swiftlint lint 
```

marcelofabri:
If it helps, I'm running on a 6-Core 2013 Mac Pro.

marcelofabri:
Actually this just happens with the `--benchmark` option.

Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
jpsim: I didn't use it because I wasn't aware of it! Thanks for sharing 😊 

The docs are non-existent, unfortunately: https://developer.apple.com/reference/dispatch/dispatchqueue/2016088-concurrentperform

But doing a bit of digging, the Dispatch Swift overlay [seems to implement that function](https://github.com/apple/swift-corelibs-libdispatch/blob/012f48b05b0aa62b95f034236e98685175fc4d2c/src/swift/DispatchStubs.cc#L75) with `dispatch_apply`, which _is_ well documented here: https://developer.apple.com/reference/dispatch/1452846-dispatch_apply_f

It seems to be exactly what I've reimplemented here, so I think I'll refactor to use that instead.

Source/SwiftLintFramework/Extensions/File+Cache.swift
jpsim: What makes you think that?

Source/swiftlint/Commands/LintCommand.swift
jpsim: Done in 253e1f4.

jpsim:
Yeah, I had neglected to make recording benchmarks thread-safe. Done in 253e1f4.

Source/SwiftLintFramework/Extensions/File+Cache.swift
marcelofabri: I had the impression that `NSLock`s are really expensive, but can't find any good references right now

norio-nomura:
I have gotten crashes twice on linting:
https://gist.github.com/norio-nomura/e8dc253954ec8299e1e647aad3d99398

jpsim:
Are those crashes reproducible?

jpsim:
@norio-nomura what project(s) were you linting when you experienced these crashes?

norio-nomura:
I experienced on Carthage. It happens rarely. But it happened first and second try with this PR.

jpsim:
Hmmm, I ran this with Thread Sanitizer enabled against Carthage with all submodules recursively cloned 5 times in a row without hitting any issues. Xcode 8.2 with just a bare `swiftlint` command, no extra configuration files or CLI arguments.

The only change I've made was to enable all opt-in rules:

```diff
--- a/Source/SwiftLintFramework/Models/Configuration.swift
+++ b/Source/SwiftLintFramework/Models/Configuration.swift
@@ -91,11 +91,7 @@ public struct Configuration: Equatable {
                 return whitelistRules.contains(type(of: rule).description.identifier)
             }
         } else {
-            rules = configuredRules.filter { rule in
-                let id = type(of: rule).description.identifier
-                if validDisabledRules.contains(id) { return false }
-                return optInRules.contains(id) || !(rule is OptInRule)
-            }
+            rules = configuredRules
         }
```

Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
```diff
@@ -42,4 +43,23 @@ extension Array {
...
+
+        let queueLabelPrefix = "io.realm.SwiftLintFramework.map.\(NSUUID().uuidString)"
+        let resultAccumulatorQueue = DispatchQueue(label: "\(queueLabelPrefix).resultAccumulator")
+        DispatchQueue.concurrentPerform(iterations: count) { index in
+            let jobIndexAndResults = (index, transform(self[index]))
+            resultAccumulatorQueue.sync {
+                result.append(jobIndexAndResults)
+            }
+        }
+        return result.sorted { $0.0 < $1.0 }.flatMap { $0.1 }
```
norio-nomura: Is the `flatMap` correct?

Source/SwiftLintFramework/Extensions/File+Cache.swift
```diff
@@ -33,35 +33,96 @@ private var syntaxTokensByLinesCache = Cache({ file in file.syntaxTokensByLine()
...
         let key = file.cacheKey
-        if let value = values[key] {
-            return value
+        lock.lock()
+        if let cachedValue = values[key] {
+            lock.unlock()
+            return cachedValue
         }
+        lock.unlock()
         let value = factory(file)
```
norio-nomura: `factory()` invocation might be happen in parallel for same key `file` because it `unlocked()` while running `factory()`.

Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
jpsim: No, should've been changed to `map` in 8774a92. Fixed in 53d0cf6. Thanks!

Source/SwiftLintFramework/Extensions/File+Cache.swift
jpsim: That shouldn't be a problem when `factory` is already thread-safe. But anyway, I've refactored to perform the factory while locked in 8906da6, to make sure this keeps working if factory is ever not thread-safe.

jpsim:
I appreciate everyone's reviews on this! Thanks @norio-nomura, @marcelofabri and special guest @krzysztofzablocki 🙏.

I think I'll be merging this despite @norio-nomura's crashes, since neither of us appear to be able to reproduce them now.

Source/SwiftLintFramework/Extensions/File+Cache.swift
norio-nomura: Thanks for change.

I thought that it should be locked for functionality of `Cache` even if `factory` would be already thread-safe. Because cached contents possibly be reference type, and previous implementation could return difference instance for same key.

PR #1107, opened 2017-01-02
Add command to generate rules documentation

### LINKED ISSUE #1078: Show description of rule
It would be nice to show the description of the rule when running the `swiftlint rules` command.

marcelofabri:
Will fix #1078

This produces a markdown looking like [this](https://gist.github.com/marcelofabri/7f568c40030bb672e62f72911e0392fc).

Next steps:

- [x] Add a target on `Makefile` to run this command and output it to a file.
- [ ] List rules configuration. In order to do this right, we'd need to change a bit the `RuleConfiguration` protocol to provide more information about the parameters.
- [ ] Add a `detailedDescription: String?` on `RuleDescription` that would be used to generate this. All descriptions currently were thought for console and Xcode usage and we should go into more details in the docs. @jpsim did an awesome job on the [0.14 release blog post](https://realm.io/news/swiftlint-0-14-rules-cocoapods-swiftpm-linux/) that could be used as an inspiration.
- [x] Add a Danger check to validate that the markdown was updated if a new rule is being created.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
   when you run `swiftlint rules`.  
   [Marcelo Fabri](https://github.com/marcelofabri)
 
+* Add [rules documentation](Rules.md) generation.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#1078](https://github.com/realm/SwiftLint/issues/1078)
+
 ##### Bug Fixes
 
 * None.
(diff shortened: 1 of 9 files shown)


aamctustwo:
It would also be good if this generated a Table of Contents for the top of the Markdown file. As we keep adding rules, this is going to be become one loooooong file.

marcelofabri:
We probably should use Sourcery for this instead

marcelofabri:
> We probably should use Sourcery for this instead

Actually, that won't work 😅

marcelofabri:
🚢

jpsim:
Oh my, I'm reviewing this now finally and this is soooo cool. This is a very nice addition to the project.

I'm very happy you did this. I was expecting to have more nitpicks, but really I just have #1660.

marcelofabri:
🎉

I hope we can inprove this even more in the future with rule configurations and better descriptions for rules 🙌

PR #1108, opened 2017-01-02
Enable more opt-in rules in SwiftLint

marcelofabri:


marcelofabri:
Totally agree. I've enable it more as an extra integration test since there were only 2 violations.

PR #1110, opened 2017-01-02
Configurable number separator

### LINKED ISSUE #1109: Configure number_separator with minimum length
From https://github.com/realm/SwiftLint/pull/1108#pullrequestreview-14820745:

---

It'd be nice if `number_separator` was configurable with a minimum length, since I personally would prefer `1000` over `1_000`, but starting at 5 digits would prefer using underscores (`10_000`).

marcelofabri:
Fixes #1109


Source/SwiftLintFramework/Rules/RuleConfigurations/NumberSeparatorConfiguration.swift
```diff
@@ -0,0 +1,42 @@
+//
+//  NumberSeparatorConfiguration.swift
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 01/02/17.
+//  Copyright © 2017 Realm. All rights reserved.
+//
+
+import Foundation
```
jpsim: Is it necessary to import Foundation?

PR #1112, opened 2017-01-03
Make NimbleOperatorRule correctable

VojtaStavik:
Hey guys! I had over 500 warnings in my project from the new nimble_operator rule and I didn't want to correct it manually. I made the rule correctable for the most common use cases.

### DIFF
--- a/Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
+++ b/Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
@@ -8,7 +8,7 @@
 
 import SourceKittenFramework
 
-public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule {
+public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule, CorrectableRule {
     public var configuration = SeverityConfiguration(.warning)
 
     public init() {}
@@ -39,9 +39,20 @@ public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule {
             "↓expect(10).to(beLessThanOrEqualTo(10))\n",
             "↓expect(x).to(beIdenticalTo(x))\n",
             "expect(10) > 2\n ↓expect(10).to(beGreaterThan(2))\n"
+        ],
+        corrections: [
+            "↓expect(seagull.squawk).toNot(equal(\"Hi\"))\n": "expect(seagull.squawk) != \"Hi\"\n",
+            "↓expect(\"Hi!\").to(equal(\"Hi!\"))\n": "expect(\"Hi!\") == \"Hi!\"\n",
+            "↓expect(12).toNot(equal(10))\n": "expect(12) != 10\n",
+            "↓expect(value1).to(equal(value2))\n": "expect(value1) == value2\n",
+            "↓expect(value1).to(equal(10))\n": "expect(value1) == 10\n",
+            "↓expect(10).to(beGreaterThan(8))\n": "expect(10) > 8\n",
+            "↓expect(10).to(beGreaterThanOrEqualTo(10))\n": "expect(10) >= 10\n",
+            "↓expect(10).to(beLessThan(11))\n": "expect(10) < 11\n",
+            "↓expect(10).to(beLessThanOrEqualTo(10))\n": "expect(10) <= 10\n",
+            "↓expect(x).to(beIdenticalTo(x))\n": "expect(x) === x\n"
         ]
     )
-
     public func validateFile(_ file: File) -> [StyleViolation] {
         let operators = ["equal", "beIdenticalTo", "beGreaterThan",
                          "beGreaterThanOrEqualTo", "beLessThan", "beLessThanOrEqualTo"]
@@ -59,5 +70,19 @@ public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule {
                 severity: configuration.severity,
                 location: Location(file: file, characterOffset: $0.0.location))
         }
+    }
+    public func correctFile(_ file: File) -> [Correction] {
+        let variable = "\\s*(.*?)\\s*"
+        let patterns = [
+            "expect\\(\(variable)\\)\\.to\\(equal\\(\(variable)\\)\\)": "expect($1) == $2",
+            "expect\\(\(variable)\\)\\.toNot\\(equal\\(\(variable)\\)\\)": "expect($1) != $2",
+            "expect\\(\(variable)\\)\\.to\\(beIdenticalTo\\(\(variable)\\)\\)": "expect($1) === $2",
+            "expect\\(\(variable)\\)\\.toNot\\(beIdenticalTo\\(\(variable)\\)\\)": "expect($1) !== $2",
+            "expect\\(\(variable)\\)\\.to\\(beGreaterThan\\(\(variable)\\)\\)": "expect($1) > $2",
+            "expect\\(\(variable)\\)\\.to\\(beGreaterThanOrEqualTo\\(\(variable)\\)\\)": "expect($1) >= $2",
+            "expect\\(\(variable)\\)\\.to\\(beLessThan\\(\(variable)\\)\\)": "expect($1) < $2",
+            "expect\\(\(variable)\\)\\.to\\(beLessThanOrEqualTo\\(\(variable)\\)\\)": "expect($1) <= $2"
+        ]
+        return file.correctLegacyRule(self, patterns: patterns)
     }
 }
--- a/Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
+++ b/Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
@@ -6,6 +6,7 @@
 //  Copyright © 2016 Realm. All rights reserved.
 //
 
+import Foundation
 import SourceKittenFramework
 
 public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule, CorrectableRule {
--- a/Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
+++ b/Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
@@ -45,44 +45,112 @@ public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule, Correcta
             "↓expect(\"Hi!\").to(equal(\"Hi!\"))\n": "expect(\"Hi!\") == \"Hi!\"\n",
             "↓expect(12).toNot(equal(10))\n": "expect(12) != 10\n",
             "↓expect(value1).to(equal(value2))\n": "expect(value1) == value2\n",
+            "↓expect(   value1  ).to(equal(  value2.foo))\n": "expect(value1) == value2.foo\n",
             "↓expect(value1).to(equal(10))\n": "expect(value1) == 10\n",
             "↓expect(10).to(beGreaterThan(8))\n": "expect(10) > 8\n",
             "↓expect(10).to(beGreaterThanOrEqualTo(10))\n": "expect(10) >= 10\n",
             "↓expect(10).to(beLessThan(11))\n": "expect(10) < 11\n",
             "↓expect(10).to(beLessThanOrEqualTo(10))\n": "expect(10) <= 10\n",
-            "↓expect(x).to(beIdenticalTo(x))\n": "expect(x) === x\n"
+            "↓expect(x).to(beIdenticalTo(x))\n": "expect(x) === x\n",
+            "expect(10) > 2\n ↓expect(10).to(beGreaterThan(2))\n": "expect(10) > 2\n expect(10) > 2\n"
         ]
     )
+
+    typealias Operators = (to: String?, toNot: String?)
+    typealias MatcherFunction = String
+
+    fileprivate let operatorsMapping: [MatcherFunction: Operators] = [
+        "equal": (to: "==", toNot: "!="),
+        "beIdenticalTo": (to: "===", toNot: "!=="),
+        "beGreaterThan": (to: ">", toNot: nil),
+        "beGreaterThanOrEqualTo": (to: ">=", toNot: nil),
+        "beLessThan": (to: "<", toNot: nil),
+        "beLessThanOrEqualTo": (to: "<=", toNot: nil)
+    ]
+
     public func validateFile(_ file: File) -> [StyleViolation] {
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
```diff
@@ -39,9 +39,20 @@ public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule {
...
+            "↓expect(value1).to(equal(value2))\n": "expect(value1) == value2\n",
+            "↓expect(value1).to(equal(10))\n": "expect(value1) == 10\n",
+            "↓expect(10).to(beGreaterThan(8))\n": "expect(10) > 8\n",
+            "↓expect(10).to(beGreaterThanOrEqualTo(10))\n": "expect(10) >= 10\n",
+            "↓expect(10).to(beLessThan(11))\n": "expect(10) < 11\n",
+            "↓expect(10).to(beLessThanOrEqualTo(10))\n": "expect(10) <= 10\n",
+            "↓expect(x).to(beIdenticalTo(x))\n": "expect(x) === x\n"
         ]
     )
-
```
jpsim: we typically separate functions with a single empty newline

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
```diff
@@ -60,4 +71,18 @@ public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule {
                 location: Location(file: file, characterOffset: $0.0.location))
         }
     }
+    public func correctFile(_ file: File) -> [Correction] {
```
jpsim: we typically separate functions with a single empty newline

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
```diff
@@ -60,4 +71,18 @@ public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule {
...
+        let variable = "\\s*(.*?)\\s*"
+        let patterns = [
+            "expect\\(\(variable)\\)\\.to\\(equal\\(\(variable)\\)\\)": "expect($1) == $2",
```
jpsim: we should keep the list of supported operators in one place, since we also use them in `validateFile(_:)`.

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
```diff
@@ -60,4 +71,18 @@ public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule {
...
+            "expect\\(\(variable)\\)\\.toNot\\(equal\\(\(variable)\\)\\)": "expect($1) != $2",
+            "expect\\(\(variable)\\)\\.to\\(beIdenticalTo\\(\(variable)\\)\\)": "expect($1) === $2",
+            "expect\\(\(variable)\\)\\.toNot\\(beIdenticalTo\\(\(variable)\\)\\)": "expect($1) !== $2",
+            "expect\\(\(variable)\\)\\.to\\(beGreaterThan\\(\(variable)\\)\\)": "expect($1) > $2",
+            "expect\\(\(variable)\\)\\.to\\(beGreaterThanOrEqualTo\\(\(variable)\\)\\)": "expect($1) >= $2",
+            "expect\\(\(variable)\\)\\.to\\(beLessThan\\(\(variable)\\)\\)": "expect($1) < $2",
+            "expect\\(\(variable)\\)\\.to\\(beLessThanOrEqualTo\\(\(variable)\\)\\)": "expect($1) <= $2"
+        ]
+        return file.correctLegacyRule(self, patterns: patterns)
```
jpsim: If we reuse the `File.correctLegacyRule(_:patterns:)` function, we should rename it to something more appropriate.

Also, this function validates different syntax kinds than `validateFile(_:)` does:

```swift
return matchPattern(pattern).filter { range, kinds in
    return kinds.first == .identifier &&
        !ruleEnabledViolatingRanges([range], forRule: rule).isEmpty
}...
```

This needs to be aligned so that we don't accidentally correct things that wouldn't have triggered a lint violation, or miss any.

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
VojtaStavik: Actually, I wanted to do that but testSwiftLintAutoCorrects was failing because of that. I'm not sure what exactly is happening here.

screen shot 2017-01-03 at 19 15 07

marcelofabri: probably there were empty spaces in the line (Xcode adds them by default)

Make sure trimming is enabled:

captura de tela 2017-01-03 as 16 26 53

VojtaStavik: Ha! I didn't know about that. Fixed. Thank you 🎉

VojtaStavik:
This was more fun than I expected 😄  It should be fixed @jpsim .

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
```diff
@@ -39,25 +40,118 @@ public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule {
...
         ]
     )
 
-    public func validateFile(_ file: File) -> [StyleViolation] {
-        let operators = ["equal", "beIdenticalTo", "beGreaterThan",
-                         "beGreaterThanOrEqualTo", "beLessThan", "beLessThanOrEqualTo"]
-        let operatorsPattern = "(" + operators.joined(separator: "|") + ")"
-        let pattern = "expect\\((.(?!expect\\())+?\\)\\.to(Not)?\\(\(operatorsPattern)\\("
-        let excludingKinds = SyntaxKind.commentKinds()
+    typealias Operators = (to: String?, toNot: String?)
```
jpsim: `fileprivate`?

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
```diff
@@ -39,25 +40,118 @@ public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule {
...
+    typealias MatcherFunction = String
```
jpsim: `fileprivate`?

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
```diff
@@ -39,25 +40,118 @@ public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule {
...
+
+        return file.matchPattern(pattern)
+            .filter { _, kinds in
+                kinds.filter(excludingKinds.contains).isEmpty && kinds.first == .identifier
+            }.map { $0.0 }
+    }
+
+    public func correctFile(_ file: File) -> [Correction] {
+        let matches = violationMatchesRanges(in: file)
+            .filter { file.ruleEnabledViolatingRanges([$0], forRule: self).isEmpty == false }
```
jpsim: Use `!` instead of `== false`?

```swift
.filter { !file.ruleEnabledViolatingRanges([$0], forRule: self).isEmpty }
```

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
```diff
@@ -39,25 +40,118 @@ public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule {
...
+        guard matches.isEmpty == false else { return [] }
```
jpsim: Use `!` instead of `== false`?

```swift
guard !matches.isEmpty else { return [] }
```

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
```diff
@@ -39,25 +40,118 @@ public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule {
...
+            for (functionName, operatorCorrections) in operatorsMapping {
+                guard let correctedString = contents.replace(function: functionName,
+                                                             with: operatorCorrections,
+                                                             in: range)
+                else {
+                    continue
+                }
+
+                contents = correctedString
+                corrections.append(Correction(ruleDescription: description,
```
jpsim: spacing:

```swift
corrections.append(Correction(ruleDescription: description,
                              location: Location(file: file, characterOffset: range.location)))
```

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
```diff
@@ -39,25 +40,118 @@ public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule {
...
+                                              location:  Location(file: file, characterOffset: range.location)))
+                break matchesLoop
```
jpsim: This will break out of the `matchesLoop` as soon as the first correction is performed. Doesn't seem to be what you want.

I think what you want is:

```swift
for range in matches.sorted(by: { $0.location > $1.location }) {
    for (functionName, operatorCorrections) in operatorsMapping {
        if let correctedString = contents.replace(function: functionName, with: operatorCorrections,
                                                  in: range) {
            contents = correctedString
            let location = Location(file: file, characterOffset: range.location)
            corrections.insert(Correction(ruleDescription: description, location: location), at: 0)
            break
        }
    }
}
```

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
jpsim: also, since you're iterating over the matches in last-to-first order (i.e. in reverse), to have the corrections in first-to-last order, you'd need to `corrections.insert(..., at: 0)` instead.

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
```diff
@@ -39,25 +40,118 @@ public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule {
...
+    fileprivate func replace(function name: NimbleOperatorRule.MatcherFunction,
+                             with operators: NimbleOperatorRule.Operators,
+                             in range: NSRange) -> String? {
+
+        let anything = "\\s*(.*?)\\s*"
+
+        let toPattern = ("expect\\(\(anything)\\)\\.to\\(\(name)\\(\(anything)\\)\\)", operators.to)
+        let toNotPattern = ("expect\\(\(anything)\\)\\.toNot\\(\(name)\\(\(anything)\\)\\)", operators.toNot)
+
+        var correctedString: String? = nil
```
jpsim: Initializing an optional variable with nil is redundant.

```swift
var correctedString: String?
```

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
```diff
@@ -39,25 +40,118 @@ public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule {
...
+
+        for (pattern, operatorString) in [toPattern, toNotPattern] {
+            guard let operatorString = operatorString else {
```
jpsim: This should never be `nil`. Please force-unwrap instead.

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
```diff
@@ -39,25 +40,118 @@ public struct NimbleOperatorRule: ConfigurationProviderRule, OptInRule {
...
+                continue
+            }
+
+            let expression = regex(pattern)
+            if let _ = expression.matches(in: self, options: [], range: range).first {
```
jpsim: Please use `if ... != nil` rather than `if let _ = ...`. We ought to have a rule for this.

And while we're at it, instead of getting the first item only to discard it, might as well check `isEmpty`:

```swift
if !expression.matches(in: self, options: [], range: range).isEmpty {
```

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
VojtaStavik: I didn't know the order for Corrections matters. I thought it would be OK because each correction has its specific location. I'll fix that.

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
VojtaStavik: Actually, that's what I wanted. There will always be **zero or one** correction for each match. So once we applied the correction, we can just to the next range.

jpsim: That's not what your code is doing. It just corrects a single range from the `matches`, since the break exits the `matchesLoop`, not the inner one.

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
VojtaStavik: Operator can be nil. For example:
``` "beLessThanOrEqualTo": (to: "<=", toNot: nil) ```
There's no correction for ```toNot(beLessThanOrEqualTo(a))```. I was thinking about that but changing it to ```> a``` also changes the meaning of the expectation and I didn't want to do that.

jpsim: awesome, I hadn't realized. Keep as-is then! 👌

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
jpsim: The order for corrections doesn't really matter all that much, it's just used for reporting, but might as well keep it correct-ish.

Source/SwiftLintFramework/Rules/NimbleOperatorRule.swift
VojtaStavik: Ooooh, I see it now :man_facepalming:  Sorry. What I mean to use was ```continue matchesLoop```. It just accidentally worked with this bug, too. I'll fix it.

jpsim: I'm pretty sure saying `continue [outer-label]` from the inner-loop is equivalent to saying `break`, without the need to label your loops.

VojtaStavik: You're right. I used the labeled loop in the earlier versions of the function and my mind got stuck to it.

VojtaStavik:
Please take a look @jpsim . It should be fixed.

Issue #655, opened 2016-05-06
Exclude enum cases from type_name rule


phelgo:
The `type_name` rule is currently being applied to enum cases. While the capital and alphanumerical requirements apply perfectly to enum cases, the minimum length may not. Examples like `Direction.Up`, `Sixe.XL`, `Cardinal.SW` should be allowed, and it feels cumbersome to add all these exceptions to the `.swiflint.yml`

chadmoone:
I'm not sure about this.  Potentially some default exceptions should be added for common two-letter words (`up`, `in`, `id` etc).  But your other examples are acronyms/abbreviations, which most styleguides discourage to begin with.   Personally, I don't see why enum cases should be treated differently than anything else in this regard.  If your application really needs to have several two-letter enum types, you can always customize the rule to change the minimum allowed to 2.

jpsim:
as of #660, we allow lowercase names for enum cases.

Issue #654, opened 2016-05-05
Add support for lowerCamelCase enum cases


bmunge:
The accepted Swift evolution guideline for Swift 3.0 standardizes lowerCamelCase enum values.
[Source](https://github.com/apple/swift-evolution/blob/master/proposals/0006-apply-api-guidelines-to-the-standard-library.md)

Unless I'm missing something, the current type_name rule does not allow a way to ignore the uppercase requirement for enum cases. I think support needs to be added to the type_name rule to exclude enum cases from the capitalization requirement, or from the rule entirely.

jpsim:
as of #660, we allow lowercase names for enum cases.

PR #1115, opened 2017-01-04
TypeNameRuleExamples: added test to verify that enum cases are lowercase in Swift 3

NachoSoto:
To my surprise, the `type_name` rule is not currently validating this. I've found some issues referring to this (#655, #654, #663, and the change in #660), but it's not clear to me whether not validating this was on purpose.

So I figured I'd ask this question by sending the failing test. I'd love to have this rule validate this, but I wonder if there's a reason why this was not enabled.

Thanks.


marcelofabri:
This will actually be fixed by #954. You can see the a similar test here: https://github.com/realm/SwiftLint/pull/954/files#diff-eb77d039fb569db405aad5349c41349aR47

That took a while because we wanted to make sure to have a deprecation warning and not just breaking old configurations, which is now possible thanks to #974.

jpsim:
Closing in favor of #954.

NachoSoto:
Awesome, thank you!

PR #1119, opened 2017-01-05
Add Unused optional binding rule

### LINKED ISSUE #1116: Rule request:  if ... != nil over if let _ = ....
From [this comment](https://github.com/realm/SwiftLint/pull/1112#discussion_r94672579):

> Please use `if ... != nil` rather than `if let _ = ....` We ought to have a rule for this.

rakaramos:
Closes #1116

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,10 @@
   [Aaron McTavish](https://github.com/aamctustwo)
   [#769](https://github.com/realm/SwiftLint/issues/769)
 
+* Add `unused_optional_binding` that will check for optional bindings not being used.  
+  [Rafael Machado](https://github.com/rakaramos/)
+  [#1116](https://github.com/realm/SwiftLint/issues/1116)
+
 ##### Bug Fixes
 
 * Ignore close parentheses on `vertical_parameter_alignment` rule.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,6 +103,9 @@
   [JP Simard](https://github.com/jpsim)
   [#1105](https://github.com/realm/SwiftLint/issues/1105)
 
+* Fix `variable_name` and `type_name` rules on Linux.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+
 ## 0.15.0: Hand Washable Holiday Linens 🎄
 
 ##### Breaking
@@ -110,9 +113,6 @@
 * `line_length` rule now has a default value of `120` for warnings.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1008](https://github.com/realm/SwiftLint/issues/1008)
-
-* Fixing `variable_name` and `type_name` rules on Linux.  
-  [Marcelo Fabri](https://github.com/marcelofabri)
 
 ##### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1060](https://github.com/realm/SwiftLint/issues/1060)
 
+* Now `number_separator` rule can be configured with a minimum length.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#1109](https://github.com/realm/SwiftLint/issues/1109)
+
 ##### Bug Fixes
 
 * Ignore close parentheses on `vertical_parameter_alignment` rule.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,15 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1109](https://github.com/realm/SwiftLint/issues/1109)
 
+* Make many operations in SwiftLintFramework safe to call in multithreaded
+  scenarios, including accessing `Linter.styleViolations`.  
+  [JP Simard](https://github.com/jpsim)
+  [#1077](https://github.com/realm/SwiftLint/issues/1077)
+
+* Speed up linting by processing multiple files and rules concurrently.  
+  [JP Simard](https://github.com/jpsim)
+  [#1077](https://github.com/realm/SwiftLint/issues/1077)
+
 ##### Bug Fixes
 
 * Ignore close parentheses on `vertical_parameter_alignment` rule.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 
 ##### Enhancements
 
+* Permit unsigned and explicitly-sized integer types in `valid_ibinspectable`  
+  [Daniel Duan](https://github.com/dduan)
+
 * Make `nimble_operator` rule correctable.  
   [Vojta Stavik](https://github.com/VojtaStavik)
 
@@ -98,7 +101,24 @@
   [Aaron McTavish](https://github.com/aamctustwo)
   [#769](https://github.com/realm/SwiftLint/issues/769)
 
-* Add `unused_optional_binding` that will check for optional bindings not being used.  
+* Add check to ignore urls in `line_length` rule when `ignores_urls`
+  configuration is enabled.  
+  [Javier Hernández](https://github.com/jaherhi)
+  [#384](https://github.com/realm/SwiftLint/issues/384)
+
+* Add `shorthand_operator` rule that validates that shorthand operators should
+  be used when possible.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#902](https://github.com/realm/SwiftLint/issues/902)
+
+* Allow specifying a `swiftlint_version` configuration key which will log a
+  warning if the current running version of SwiftLint is different than this
+  value.  
+  [JP Simard](https://github.com/jpsim)
+  [#221](https://github.com/realm/SwiftLint/issues/221)
+
+* Add `unused_optional_binding` rule that will check for optional bindings 
+  not being used.  
   [Rafael Machado](https://github.com/rakaramos/)
   [#1116](https://github.com/realm/SwiftLint/issues/1116)
 
@@ -127,6 +147,13 @@
 
 * Fix `variable_name` and `type_name` rules on Linux.  
   [Marcelo Fabri](https://github.com/marcelofabri)
+
+* Fix linting directories with names ending with `.swift`.  
+  [JP Simard](https://github.com/jpsim)
+
+* Fix running `swiftlint version` when building with Swift Package Manager.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#1045](https://github.com/realm/SwiftLint/issues/1045)
 
 ## 0.15.0: Hand Washable Holiday Linens 🎄
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1096](https://github.com/realm/SwiftLint/issues/1096)
 
+* Add `large_tuple` configurable rule that validates that tuples shouldn't
+  have too many members.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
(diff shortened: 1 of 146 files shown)


rakaramos:
@jpsim not really, sorry! I just went through the issues and grab this one to learn/contribute 😅 
But I think you are right, maybe it should be opt-in, I can see arguments going forever on `let _` over `!= nil`

Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
```diff
@@ -0,0 +1,73 @@
...
+            "}\n"
+        ],
+        triggeringExamples: [
+            "if let ↓_ = Foo.optionalValue {\n" +
+            "}\n",
+            "if let a = Foo.optionalValue, let ↓_ = Foo.optionalValue2 {\n" +
+            "}\n",
+            "guard let a = Foo.optionalValue, let ↓_ = Foo.optionalValue2 {\n" +
+            "}\n",
+            "if let (first, second), let ↓_ = Foo.optionalValue2 = getOptionalTuple() {\n" +
```
marcelofabri: can you add an assignment to `let (first, second)` so it can be valid Swift?

Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
```diff
@@ -0,0 +1,73 @@
...
+
+    public func validateFile(_ file: File,
+                             kind: StatementKind,
+                             dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard kind == .if || kind == .guard else {
+                return []
+        }
+
+        return violationRanges(file: file).map {
+            print((file.contents as NSString).substring(with: $0))
```
marcelofabri: this`print` should be removed (and it's actually breaking the Linux build)

Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
```diff
@@ -0,0 +1,73 @@
...
+            return StyleViolation(ruleDescription: type(of: self).description,
+                           severity: configuration.severity,
+                           location: Location(file: file, characterOffset: $0.location))
+        }
+    }
+
+    private func violationRanges(file: File) -> [NSRange] {
+        let kinds = SyntaxKind.commentAndStringKinds()
+        let underscorePattern = "(\\b_\\b)"
```
marcelofabri: you should remove the capture group as you're not using it

Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
```diff
@@ -40,33 +40,29 @@ public struct UnusedOptionalBindingRule: ASTRule, ConfigurationProviderRule {
...
         }
     }
 
-    private func violationRanges(file: File) -> [NSRange] {
+    private func violationRanges(file: File, in range: NSRange) -> [NSRange] {
         let kinds = SyntaxKind.commentAndStringKinds()
         let underscorePattern = "\\b_\\b"
         let parenthesesPattern = "\\([^)]*\\)"
-        let pattern = underscorePattern + "|" + parenthesesPattern
+        let pattern = underscorePattern
```
marcelofabri: this assignment is not needed, you can just use `underscorePattern` in the call below

CHANGELOG.md
```diff
@@ -117,6 +117,10 @@
   [JP Simard](https://github.com/jpsim)
   [#221](https://github.com/realm/SwiftLint/issues/221)
 
+* Add `unused_optional_binding` that will check for optional bindings not being used.
```
marcelofabri: Missing the word "rule" after `unused_optional_binding`

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -166,10 +166,11 @@ extension File {
     internal func matchPattern(_ pattern: String,
                                excludingSyntaxKinds: [SyntaxKind],
                                excludingPattern: String,
+                               range: NSRange? = nil,
                                exclusionMapping: MatchMapping = { $0.range }) -> [NSRange] {
         let contents = self.contents.bridge()
         let range = NSRange(location: 0, length: contents.length)
```
marcelofabri: you need to remove this line because it'll shadow the `range`parameter

Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
```diff
@@ -0,0 +1,67 @@
...
+        guard kind == .if || kind == .guard,
+            let offset = (dictionary["key.offset"] as? Int64).flatMap({ Int($0) }),
+            let length = (dictionary["key.length"] as? Int64).flatMap({ Int($0) }),
+            let range = file.contents.bridge().byteRangeToNSRange(start: offset, length: length) else {
+            return []
+        }
+
+        return violationRanges(file: file, in: range).map {
+            return StyleViolation(ruleDescription: type(of: self).description,
+                           severity: configuration.severity,
```
marcelofabri: the parameters should be vertically aligned

Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
```diff
@@ -0,0 +1,65 @@
...
+            "}\n"
+        ],
+        triggeringExamples: [
+            "if let ↓_ = Foo.optionalValue {\n" +
+            "}\n",
+            "if let a = Foo.optionalValue, let ↓_ = Foo.optionalValue2 {\n" +
+            "}\n",
+            "guard let a = Foo.optionalValue, let ↓_ = Foo.optionalValue2 {\n" +
+            "}\n",
+            "if let (first, second) = getOptionalTuple(), let ↓_ = Foo.optionalValue2 = getOptionalTuple() {\n" +
```
jpsim: This trigger isn't valid Swift: `let ↓_ = Foo.optionalValue2 = getOptionalTuple()`

rakaramos:
@jpsim Ops! Sorry, PR is now updated.

Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
```diff
@@ -0,0 +1,69 @@
...
+            "}\n",
+            "if let a = Foo.optionalValue, let ↓_ = Foo.optionalValue2 {\n" +
+            "}\n",
+            "guard let a = Foo.optionalValue, let ↓_ = Foo.optionalValue2 {\n" +
+            "}\n",
+            "if let (first, second) = getOptionalTuple(), let ↓_ = Foo.optionalValue {\n" +
+            "}\n",
+            "if let (first, _) = getOptionalTuple(), let ↓_ = Foo.optionalValue {\n" +
+            "}\n",
+            "if let (_, _) = getOptionalTuple(), let ↓_ = Foo.optionalValue {\n" +
```
jpsim: Why shouldn't `if let (_, _) = getOptionalTuple()` trigger a violation?

Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
```diff
@@ -0,0 +1,69 @@
...
+            "}\n"
+        ]
+    )
+
+    public func validate(file: File,
+                         kind: StatementKind,
+                         dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard kind == .if || kind == .guard,
```
jpsim: Perhaps we should include `.while` in this?

Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
rakaramos: @jpsim good point! Haven't thought of it...Any suggestions on how to properly validate tuples?

Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
jpsim: @rakaramos feel free to make a new PR for this if you like.

Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
jpsim: @rakaramos feel free to make a new PR for this if you like.

jpsim:
I just had two small comments to improve this rule, but these can be added later. I'd like to cut a release soon, and would rather land this in its current state.

Thanks @rakaramos!

freak4pc:
Hey, was just wondering, how come this eventually became a default rule? This doesn't seem like a very community-spread standard, but more of a preference. @jpsim

jpsim:
This PR started with the rule being on by default, and I brought up whether or not there was a strong case for that in the community, @rakaramos said that wasn't really considered, and then we never revisited the subject.

I happen to agree fairly strongly about this stylistic rule, but it is certainly a personal preference.

To change it at this point, I'm afraid the burden of proof falls on making a case that this is actively _not_ a common convention. Which is admittedly kind tricky to determine.

However, if you'd like to propose a new more formal process around determining opt-in vs default for new rules by requiring that a reference to an external discussion or survey, I'm definitely open to that.

freak4pc:
I have not seen this at any "common" style guides such as 
https://github.com/github/swift-style-guide
https://github.com/raywenderlich/swift-style-guide
Eric'a book

etc. mention `guard optional != nil` preferable to `guard let _ = optional`

I couldn't find any fors and against this, which is why its a bit weird it wasn't optional to begin with.

I think, if speaking of a process to determine, it needs that rule would be unanimous as a standard to all swift devs without a shred-of-doubt, since otherwise it makes developers think they're wrong, when perhaps they only have a different preference/taste/coding style. 

Don't think we need to necessarily revert this as it's already out, but just sharing thoughts here.

What do you think? @jpsim

marcelofabri:
I don't think this is the case for this rule, but I'd say that SwiftLint should aim to have less optional rules and should be more opineted (as Rubocop is). It's ok to disable rules or change their configurations.

marcelofabri:
Anyways, there's at least one issue asking to make this rule opt-in, we should move the discussion there.

freak4pc:
Opinionated is absolutely fine but it needs to be opinionated based on community feedback IMHO. I agree disabling a rule is straightforward (which is what we did eventually), But perhaps some voting process on a rule suggestion could be more fitting in the future ? 

e.g. Suggesting -> Upvotes & Downvotes stage -> Evaluate demand -> Implement 

It's just a suggestion. Anyways putting the link to the other discussion here: 
https://github.com/realm/SwiftLint/issues/1446

ArtSabintsev:
hey,

I love this rule, but found it hard to get around without calling `// swiftlint:disable unused_optional_binding` in the following situation:

```swift
if let _ = try? F().g() { 
}
``` 

where g is a throwing function on F.

I'm trying to avoid having to do a do-catch loop. Is there a way around this that I'm not aware of?

SDGGiesbrecht:
@ArtSabintsev,

Try this:

```swift
if (try? F().g()) != nil {
}
```

Note: The extra parentheses are (unfortunately) necessary to keep `try?` from overreaching itself.

freak4pc:
I seriously don't see how that is more readable ... 
Anyways since there hasn't been any progress around this I've just disabled the rule... But based on the several issues opened around this, it seems I'm not the only one who doesn't prefer this syntax

ArtSabintsev:
@SDGGiesbrecht Thanks! Forgot to try parenthesis! Still going to keep it disabled.

@freak4pc Our team likes this rule, and have followed its guidance. However, I do agree that it has some issues around it (see my `_ = try?` comment).

I'm fine with it being on by default.

SDGGiesbrecht:
@freak4pc

> I seriously don’t see how that is more readable ...

I did not mean to say it was more readable, only that it is what the rule is asking you to do.

I am still on the fence about this rule. Some days I like it, other days it annoys me. The example above definitely falls under the latter. But that is why the maintainers graciously gave us the ability to use a configuration file.

PR #1120, opened 2017-01-06
reset assertion handler in test

jpsim:
specifically, after running `testRulesWithFileThatCrashedSourceKitService`.

Otherwise, other tests could fail if they're linting the same file, which can happen today if the integration tests happen to run afterwards.

/cc @marcelofabri @rakaramos


rakaramos:
👍 🎉 
thanks @jpsim and @marcelofabri

PR #1122, opened 2017-01-06
use FileManager.DirectoryEnumerator to implement filesToLintAtPath(...)

jpsim:
Just a minor refactoring. /cc @marcelofabri

### DIFF
--- a/Source/SwiftLintFramework/Extensions/NSFileManager+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/NSFileManager+SwiftLint.swift
@@ -19,7 +19,7 @@ extension FileManager: LintableFileManager {
             .absolutePathRepresentation(rootDirectory: rootPath).bridge()
             .standardizingPath
         return enumerator(atPath: absolutePath)?.flatMap { element in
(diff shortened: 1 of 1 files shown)


Source/SwiftLintFramework/Extensions/NSFileManager+SwiftLint.swift
```diff
@@ -18,27 +18,11 @@ extension FileManager: LintableFileManager {
...
-        if isDirectory {
-            do {
-                return try subpathsOfDirectory(atPath: absolutePath)
-                    .map(absolutePath.bridge().appendingPathComponent).filter {
-                        $0.bridge().isSwiftFile()
-                    }
-            } catch {
-                fatalError("Couldn't find files in \(absolutePath): \(error)")
+        return enumerator(atPath: absolutePath)?.flatMap { element in
+            if let element = element as? String, element.hasSuffix(".swift") {
```
marcelofabri: `element.isSwiftFile()`?

jpsim: Sure. I didn't do it before to avoid the `bridge()`, but might as well use the method we have for this.

PR #1123, opened 2017-01-06
Add shorthand_operator rule

### LINKED ISSUE #902: Rule Request: prefer assign operators
Prefer assign operators (`+=`, `-=`, `*=`, `/=`) over doing the operation and assigning:

```swift
var foo = 0

// should trigger
foo = foo + 2

// shouldn't trigger
foo += 2
```

A concrete example: https://github.com/realm/SwiftLint/pull/893#discussion_r89646409

marcelofabri:
Fixes #902

This is a bit slow right now (running against Realm):

```
0,173: opening_brace
0,204: colon
0,293: mark
0,304: comma
0,526: generic_type_name
0,569: shorthand_operator
```

Maybe we should make it opt-in because of this.

### DIFF
--- a/Source/SwiftLintFramework/Rules/ShorthandOperatorRule.swift
+++ b/Source/SwiftLintFramework/Rules/ShorthandOperatorRule.swift
@@ -6,6 +6,7 @@
 //  Copyright © 2017 Realm. All rights reserved.
 //
 
+import Foundation
 import SourceKittenFramework
 
 public struct ShorthandOperatorRule: ConfigurationProviderRule {
@@ -22,7 +23,9 @@ public struct ShorthandOperatorRule: ConfigurationProviderRule {
             [
                 "foo \(operation)= 1",
                 "foo \(operation)= variable",
-                "foo \(operation)= bar.method()"
+                "foo \(operation)= bar.method()",
+                "self.foo = foo \(operation) 1",
+                "foo = self.foo \(operation) 1"
             ]
         },
         triggeringExamples: allOperators.flatMap { operation in
@@ -33,7 +36,9 @@ public struct ShorthandOperatorRule: ConfigurationProviderRule {
                 "↓foo = 1 \(operation) foo\n",
                 "↓foo = aVariable \(operation) foo\n",
                 "↓foo = bar.method() \(operation) foo\n",
-                "↓foo = bar.method(param: 1, otherParam: 2) \(operation) foo\n"
+                "↓foo = bar.method(param: 1, otherParam: 2) \(operation) foo\n",
+                "↓foo.aProperty = foo.aProperty \(operation) 1\n",
+                "↓self.aProperty = self.aProperty \(operation) 1\n"
             ]
         }
     )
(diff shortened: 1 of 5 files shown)


jpsim:
How does this handle `self.foo = foo + 1`?

marcelofabri:
It shouldn't trigger, but I can add a test later to be sure

marcelofabri:
It was triggering, but I fixed on b9b60ba. One tradeoff I had to do because of performance and complexity is that this doesn't match expressions in multiple lines and also requires that the expression is on a new line (i.e. there're only whitespaces before it).

With these limitations, performance is much better:

```
0,047: vertical_parameter_alignment
0,050: valid_docs
0,053: control_statement
0,057: type_name
0,073: redundant_void_return
0,126: empty_parentheses_with_trailing_closure
0,134: closure_end_indentation
0,173: opening_brace
0,201: shorthand_operator <=
0,205: colon
0,273: comma
0,284: mark
0,533: generic_type_name
````

File: Source/SwiftLintFramework/Rules/ShorthandOperatorRule.swift [hunk 1]
```diff
@@ -0,0 +1,108 @@
...
+        let otherOperand = "\(spaces).+?\(spaces)"
+
+        let pattern1 = "^\(spaces)(\(operand))\(spaces)=\(spaces)(\\1)\(spaces)\(operators)"
+        let pattern2 = "^\(spaces)(\(operand))\(spaces)=\(otherOperand)\(operators)\(spaces)(\\3)"
+
+        return "\(pattern1)|\(pattern2)"
+    }()
+
+    // swiftlint:disable:next force_try
+    private static let regex = try! NSRegularExpression(pattern: pattern, options: [.anchorsMatchLines])
```
jpsim: we should extend our `regex` function to take in options so we don't have to silence these force tries everywhere we want different options.

Source/SwiftLintFramework/Rules/ShorthandOperatorRule.swift
```diff
@@ -0,0 +1,108 @@
...
+
+                return contents.NSRangeToByteRange(start: range.location, length: range.length)
+            }
+
+            guard byteRanges[0] != nil || byteRanges[2] != nil else {
+                return nil
+            }
+
+            let kindsInCaptureGroups = byteRanges.map { range in
+                range.flatMap { file.syntaxMap.tokensIn($0).flatMap { SyntaxKind(rawValue: $0.type) } } ?? []
```
jpsim: this line could afford to be broken up, it's hard to read as is

Source/SwiftLintFramework/Rules/ShorthandOperatorRule.swift
marcelofabri: it's complicated, how would you keep the options in the cache?

jpsim: not _that_ complicated. We store a bitmask of the options as the key in a multimap.

marcelofabri: I'll create a ticket for it so it can be done in another PR.

PR #1125, opened 2017-01-06
Allow regex variable names to include a period

nickdomenicali:
The legacy constructor doesn't autocorrect when there are periods in the parameters.

This should be recognized and autocorrected.
```NSMakeRange(0, self.length)``` 

A few other legacy constructors that weren't autocorrected and could be added to future tests.

`CGPointMake(0, y)` == `CGPoint(x: 0, y: y)`
`CGPointMake(CGRectGetWidth(bounds), y)` == `CGPoint(x: bounds.width, y: y)`
`CGRectMake(0, 0, view.frame.size.width, height)` == `CGRect(x: 0, y: 0, width: view.frame.size.width, height: height)`

### DIFF
--- a/Source/SwiftLintFramework/Helpers/RegexHelpers.swift
+++ b/Source/SwiftLintFramework/Helpers/RegexHelpers.swift
@@ -8,7 +8,7 @@
 
 struct RegexHelpers {
     // A single variable
-    static let varName = "[a-zA-Z_][a-zA-Z0-9_]+"
+    static let varName = "[a-zA-Z_\.][a-zA-Z0-9_\.]+"
 
     // A single variable in a group (capturable)
     static let varNameGroup = "\\s*(\(varName))\\s*"

### CODE REVIEW

File: Source/SwiftLintFramework/Helpers/RegexHelpers.swift [hunk 1]
```diff
@@ -8,7 +8,7 @@
 
 struct RegexHelpers {
     // A single variable
-    static let varName = "[a-zA-Z_][a-zA-Z0-9_]+"
+    static let varName = "[a-zA-Z_\.][a-zA-Z0-9_\.]+"
```
marcelofabri: I don't think you should need to add `.` in the first range, as a variable can't start with `.`

nickdomenicali: There could be an enum or static var on a number.

Not saying this code make sense, but it compiles:
```extension Int {
    static var header = 50
}
NSMakeRange(.header, 0)
```

nickdomenicali: It also doesn't work with simple operations `NSMakeRange(length + 2, 0)`.

Is this something I should add a new test file for or where would tests belong? I just started looking at this code so it might take a few days to get up to speed but I'll add some regex/autocorrect tests if that makes sense?

Coeur: If the initial dot is for extensions, then clearly declare it as an optional prefix:

    "\.?[a-zA-Z_][a-zA-Z0-9_\.]+"

Also you will want to document your pattern as lenient, as it doesn't catch invalid cases of consecutive dots or final dots.

jpsim:
Thanks for the PR, but given the lack of activity, I'll be closing this. Feel free to file a new PR if you wish to revive this.

PR #1130, opened 2017-01-06
Refactor regex functions to accept options

### LINKED ISSUE #1127: Refactor NSRegularExpression methods to accept options
we should extend our `regex` function to take in options so we don't have to silence these force tries everywhere we want different options.

---

From https://github.com/realm/SwiftLint/pull/1123#discussion_r94995715

marcelofabri:
Fixes #1127

### DIFF
--- a/Source/SwiftLintFramework/Extensions/NSRegularExpression+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/NSRegularExpression+SwiftLint.swift
@@ -13,22 +13,32 @@ public typealias NSRegularExpression = RegularExpression
 public typealias NSTextCheckingResult = TextCheckingResult
 #endif
 
-private var regexCache = [String: NSRegularExpression]()
+private var regexCache = [RegexCacheKey: NSRegularExpression]()
+
+private struct RegexCacheKey: Hashable {
+    let pattern: String
+    let options: NSRegularExpression.Options
+
+    var hashValue: Int {
+        return pattern.hashValue ^ options.rawValue.hashValue
+    }
+
+    static func == (lhs: RegexCacheKey, rhs: RegexCacheKey) -> Bool {
+        return lhs.options == rhs.options && lhs.pattern == rhs.pattern
+    }
+}
 
 extension NSRegularExpression {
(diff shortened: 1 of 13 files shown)


Source/SwiftLintFramework/Extensions/NSRegularExpression+SwiftLint.swift
```diff
@@ -13,22 +13,32 @@ public typealias NSRegularExpression = RegularExpression
 public typealias NSTextCheckingResult = TextCheckingResult
 #endif
 
-private var regexCache = [String: NSRegularExpression]()
+private var regexCache = [RegexCacheKey: NSRegularExpression]()
+
+private struct RegexCacheKey: Hashable {
```
marcelofabri: I thought it was easier to just create a type to wrap the key then implementing a full multimap 😅

Source/SwiftLintFramework/Extensions/NSRegularExpression+SwiftLint.swift
```diff
@@ -13,22 +13,32 @@ public typealias NSRegularExpression = RegularExpression
...
 
-        let result = try NSRegularExpression(pattern: pattern,
-            options: [.anchorsMatchLines, .dotMatchesLineSeparators])
-        regexCache[pattern] = result
+        let result = try NSRegularExpression(pattern: pattern, options: options)
+        regexCache[key] = result
         return result
     }
-
-    internal static func forcePattern(_ pattern: String) -> NSRegularExpression {
```
marcelofabri: I've removed this since I don't get why we should have it and the free `regex` function if they do the same thing.

Source/SwiftLintFramework/Extensions/NSRegularExpression+SwiftLint.swift
jpsim: absolutely!

Source/SwiftLintFramework/Extensions/NSRegularExpression+SwiftLint.swift
jpsim: 👍

PR #1132, opened 2017-01-07
Report lint duration changes in Danger

jpsim:
Addresses #376.

Also adds Alamofire to osscheck and improves Danger log output. /cc @marcelofabri @scottrhoyt

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,6 @@ matrix:
       env: JOB=SPM
       os: osx
       osx_image: xcode8.2
-      after_success:
-        - bundle install --jobs=3 --retry=3 --without documentation --path bundle
-        - bundle exec danger
     - script: make docker_test
       env: JOB=Linux
       sudo: required
--- a/Dangerfile
+++ b/Dangerfile
@@ -38,6 +38,8 @@ if has_app_changes
   ]
 
   @commits = {}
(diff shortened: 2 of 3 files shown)


jpsim:
Hmmm, unfortunately Travis doesn't provide very repeatable lint times...

As of f4a8986, no changes were made to SwiftLint, yet runs were ~50% slower than `master`:

image

We're already linting 5 times and taking the average, so linting more wouldn't lead to more reproducible results...

I'm not sure how we can make this more reliable.

marcelofabri:
Maybe we could try Circle CI to see if it's more reliable? 😬

jpsim:
Heh, what do you know. Much more reliable on Circle CI!

image

However, one of my assumptions was incorrect... disabling parallelization doesn't have a major impact on performance in this case, probably because Circle (and likely Travis too) must just run their VMs with a single virtual processor.

I'll try something else to attempt to simulate a performance regression.

marcelofabri:
Maybe you could disable cache in `File+Cache` instead? Enabling opt-in rules made the number of violations grow so much that Danger couldn't post to GitHub:

```
  message: body is too long (maximum is 65536 characters) // See: https://developer.github.com/v3/issues/comments/#edit-a-comment
```

marcelofabri:
👏

PR #1133, opened 2017-01-07
Permit more Int types in valid_ibinspectable

dduan:
Unsigned and/or explicitly-sized integer types are valid with `@IBInspectable`.
SwiftLint should not produce lint warnings againt these.


dduan:
Thank you @jpsim!

PR #1134, opened 2017-01-07
Version pinning and SPM fix

jpsim:
SPM fix extracted from #1073.


jpsim:
Fixes #221 and #1045.

PR #1135, opened 2017-01-07
Fix false positive on vertical_parameter_alignment

### LINKED ISSUE #1129: False positive on vertical_parameter_alignment with default parameters in multiple lines
This shouldn't trigger:

```swift
internal func regex(_ pattern: String,
                    options: NSRegularExpression.Options = [.anchorsMatchLines,
                                                            .dotMatchesLineSeparators]) -> NSRegularExpression 
```

marcelofabri:
Fixes #1129

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -150,6 +150,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1045](https://github.com/realm/SwiftLint/issues/1045)
 
+* Fix false positive on `vertical_parameter_alignment` rule when breaking line
+  in a default parameter declaration.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#1129](https://github.com/realm/SwiftLint/issues/1129)
+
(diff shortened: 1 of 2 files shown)


marcelofabri:
@jpsim Can we enable Circle for forks as well? 🤔

jpsim:
> Can we enable Circle for forks as well? 🤔

I didn't know that needed to be explicitly enabled. Could you please open a new dummy PR (maybe with a tiny change to the app's source) to make sure Circle and OSSCheck run as expected? I think I may have seen a log from Danger saying that it didn't run due to a commit push before a PR not actually being tied to a PR.

marcelofabri:
I've just force pushed this again to make sure it runs 🚀

jpsim:
> I've just force pushed this again to make sure it runs 🚀

Could you please open a new dummy PR too? I want to make sure a commit pushed prior to a PR existing still runs Danger correctly.

marcelofabri:
> Could you please open a new dummy PR too? 

Done: https://github.com/realm/SwiftLint/pull/1137

PR #1139, opened 2017-01-10
update SwiftLintFramework/Extensions to follow Swift 3 API Design Guidelines

jpsim:
1 out of 7 Swift 3 API Guidelines PRs. /cc @marcelofabri

PR #1140, opened 2017-01-10
update SwiftLintFramework/Models to follow Swift 3 API Design Guidelines

jpsim:
~~Depends on #1139.~~ 2 out of 7 Swift 3 API Guidelines PRs. /cc @marcelofabri

### DIFF
--- a/Source/SwiftLintFramework/Models/Configuration.swift
+++ b/Source/SwiftLintFramework/Models/Configuration.swift
@@ -86,7 +86,7 @@ public struct Configuration: Equatable {
         let validDisabledRules = disabledRules.filter(validRuleIdentifiers.contains)
 
         // Validate that rule identifiers aren't listed multiple times
-        if containsDuplicatedRuleIdentifiers(validDisabledRules: validDisabledRules) {
+        if containsDuplicateIdentifiers(validDisabledRules) {
             return nil
         }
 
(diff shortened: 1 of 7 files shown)


Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -233,7 +237,7 @@ private func validateRuleIdentifiers(configuredRules: [Rule], disabledRules: [St
     return validRuleIdentifiers
 }
 
-private func containsDuplicatedRuleIdentifiers(_ validDisabledRules: [String]) -> Bool {
+private func containsDuplicatedRuleIdentifiers(validDisabledRules: [String]) -> Bool {
```
marcelofabri: maybe change to `containsDuplicatedRuleIdentifiers(disabledRulesIdentifiers:)`?

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -269,7 +273,7 @@ private func validKeys(ruleList: RuleList) -> [String] {
     ].map({ $0.rawValue }) + ruleList.allValidIdentifiers()
 }
 
-private func warnAboutDeprecations(_ dict: [String: Any],
+private func warnAboutDeprecations(dict: [String: Any],
```
marcelofabri: `configurationDictionary` instead of `dict`?

Source/SwiftLintFramework/Models/Configuration.swift
jpsim: There's nothing specific to the IDs being _disabled_ in this function, so I'll just change to `containsDuplicateIdentifiers(_ identifiers: [String])`.

Source/SwiftLintFramework/Models/Configuration.swift
jpsim: `configuration` is probably sufficient, given that `Dictionary` is implied by the type.

Source/SwiftLintFramework/Models/Configuration.swift
marcelofabri: sounds good!

Source/SwiftLintFramework/Models/Configuration.swift
marcelofabri: I was in doubt on `configuration` vs `configurationDictionary`  because as we have a `Configuration` type I'd expect that a `configuration` parameter would be of that type.

PR #1157, opened 2017-01-11
Fix false positive on shorthand_operator rule

### LINKED ISSUE #1156: Inappropriate shorthand operator violation
`self.flashModeSelectionView.alpha = 1.0 - self.flashModeSelectionView.alpha`.
I can't find an easy conversion for this. I think this should be out of the rule. I mean no check for it because it has a `-` before on the right side of the `=`.
Am I right?

marcelofabri:
Fixes #1156

### DIFF
--- a/Source/SwiftLintFramework/Rules/ShorthandOperatorRule.swift
+++ b/Source/SwiftLintFramework/Rules/ShorthandOperatorRule.swift
@@ -33,27 +33,37 @@ public struct ShorthandOperatorRule: ConfigurationProviderRule {
                 "↓foo = foo \(operation) 1\n",
                 "↓foo = foo \(operation) aVariable\n",
                 "↓foo = foo \(operation) bar.method()\n",
+                "↓foo.aProperty = foo.aProperty \(operation) 1\n",
+                "↓self.aProperty = self.aProperty \(operation) 1\n"
+            ]
+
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Rules/ShorthandOperatorRule.swift
```diff
@@ -33,27 +33,37 @@ public struct ShorthandOperatorRule: ConfigurationProviderRule {
...
     )
 
-    private static let allOperators = ["+", "-", "/", "*"]
+    private static let allOperators = ["-", "/"] + commutativeOperators
+    private static let commutativeOperators = ["+", "*"]
 
     private static let pattern: String = {
-        let escapedOperators = allOperators.map { "\\\($0)" }.joined()
-        let operators = "[\(escapedOperators)]"
+        let escaped = { (operators: [String]) -> String in
```
aamctustwo: You use this pattern to escape characters in OperatorFunctionWhitespaceRule and OperatorUsageWhitespaceRule (and possibly others). Might be worth refactoring this into a helper?

marcelofabri: I don't think it's complex enough to justify the additional complexity

Issue #1170, opened 2017-01-12
`--enable-all-rules` option


norio-nomura:
It would be useful for dog fooding or benchmarking SwiftLint, if `--enable-all-rules` enables all rules over `disabled_rules`, `opt_in_rules` and `whitelist_rules` in configuration file.
Currently, I'm using `~/.swiftlint-test.yml` to enable all opt-in rules on testing SwiftLint.
This option will free me from keeping my `~/.swiftlint-test.yml` updated.

jpsim:
👍

norio-nomura:
I hope `OptInRule` would be enabled by `--enable-all-rules` too.

jpsim:
I thought my PR did that. I'll take another look at it later.

norio-nomura:
Implemented in #1172. Thanks! 🙏

PR #1172, opened 2017-01-12
Add CLI option to run all rules

jpsim:
Addresses #1170. /cc @norio-nomura

### DIFF
--- a/Source/SwiftLintFramework/Models/Configuration.swift
+++ b/Source/SwiftLintFramework/Models/Configuration.swift
@@ -39,6 +39,7 @@ public struct Configuration: Equatable {
 
     public init?(disabledRules: [String] = [],
                  optInRules: [String] = [],
+                 enableAllRules: Bool = false,
                  whitelistRules: [String] = [],
                  included: [String] = [],
                  excluded: [String] = [],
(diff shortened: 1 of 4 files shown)


norio-nomura:
Sorry, I failed to explain on the issue.
I expected that `OptInRule` will also be enabled by `--enable-all-rules` option.

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -98,8 +99,10 @@ public struct Configuration: Equatable {
         // set the config threshold to the threshold provided in the config file
         self.warningThreshold = warningThreshold
 
-        // white_list rules take precendence over all else.
-        if !whitelistRules.isEmpty {
+        // Precedence is enableAllRules > whitelistRules > everything else
+        if enableAllRules {
+            rules = configuredRules
```
norio-nomura: Do `configuredRules` contain rules that is `OptInRule`?

norio-nomura:
> I expected that OptInRule will also be enabled by --enable-all-rules option.

It seems `OptInRule` is not enabled by `--enable-all-rules`
```terminal.session
➜  12:57:34 $ cat .swiftlint-opt-in.yml
opt_in_rules:
 - attributes
➜  12:57:40 $ time swiftlint lint --benchmark --enable-all-rules --config .swiftlint-opt-in.yml >/dev/null 2>&1
swiftlint lint --benchmark --enable-all-rules --config .swiftlint-opt-in.yml   24.95s user 1.82s system 111% cpu 24.091 total
➜  12:58:13 $ wc -l benchmark_rules_2017_01_12_12_58_13.txt
      56 benchmark_rules_2017_01_12_12_58_13.txt
➜  12:58:37 $ time swiftlint lint --benchmark --enable-all-rules >/dev/null 2>&1
swiftlint lint --benchmark --enable-all-rules > /dev/null 2>&1  48.75s user 2.30s system 168% cpu 30.282 total
➜  12:59:27 $ wc -l benchmark_rules_2017_01_12_12_59_26.txt
      55 benchmark_rules_2017_01_12_12_59_26.txt
```

jpsim:
@norio-nomura I think the issue is that the cache gets re-used, which takes the results from the previous run with fewer rules.

jpsim:
Oops, no the issue was that the value from the CLI wasn't being passed to the Configuration initializer. Fixed now.

PR #1179, opened 2017-01-12
Unused optional binding

rakaramos:


jpsim:
This looks good, but I'm interested to see if OSSCheck flags some new violations or performance regressions with this change.

rakaramos:
@jpsim me too 👍 
that's a dirty regex :sneezing_face:

jpsim:
@SwiftLintBot says we're good! 👌 ⛵️

PR #1190, opened 2017-01-13
Fix number separator violation swiftlint warnings

benasher44:
I have a WIP branch (PR coming soon) to add support for linting multiple files as CLI args, but I ran into these `swiftlint` warnings while working on it. I figured I fix these and get them out of the way first.

### DIFF
--- a/Source/SwiftLintFramework/Rules/FileLengthRule.swift
+++ b/Source/SwiftLintFramework/Rules/FileLengthRule.swift
@@ -9,7 +9,7 @@
 import SourceKittenFramework
 
 public struct FileLengthRule: ConfigurationProviderRule, SourceKitFreeRule {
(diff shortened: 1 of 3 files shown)


jpsim:
I think you need to update your version of SwiftLint, as we fixed this in #1110.

benasher44:
Oh! You're right. Sorry about that!

benasher44:
Ah. I went to upgrade it, and I'm using 0.15.0. I install it with brew, and it looks like 0.16.0 isn't available there yet (http://braumeister.org/formula/swiftlint). It'd be great to see the new version there soon!

jpsim:
Feel free to add your voice to https://github.com/Homebrew/homebrew-core/pull/8719

Issue #810, opened 2016-09-08
Support list of files as argument


gzfrancisco:
I think the tool can be improved if the command can take a list of files so we can check only those files.

We can do something like:

```
$ swiftlint MyClass.swift AnotherClass.swift script.swift
```

jpsim:
That would indeed be nice. This could be done with Commandant by using an `Argument` option, which would look like this:

```shell
$ swiftlint lint -- MyClass.swift AnotherClass.swift script.swift
```

benasher44:
I've started working on this in an effort to make SwiftLint work with pre-commit as part of pre-commit/pre-commit#451. I'm making decent progress, but I have a question/issue. What does `rootPath` mean in the context of multiple paths? Would the current working directory suffice if more than one path is specified, or would it be more appropriate to calculate nearest common ancestor of all of the paths? It seems like the latter would be better. Thoughts?

PH9:
Is this should be close issue. It is already merged

benasher44:
Ah yep this is merged

Issue #1184, opened 2017-01-13
Investigate cases in which cache is stale


dalemy:
We have just updated to 0.16 in our codebase and I'm working through the new rules. First up it's `sorted_imports`. For this, we have a build phase which runs `${PODS_ROOT}/SwiftLint/swiftlint` as per the guidelines in the README file. I've worked through part of our codebase (it takes a while as we have well over a hundred thousand lines of Swift code) and sorted the imports. I then re-ran the linter, to clear out the issues I'd already fixed and make it easier to keep track of my progress. 

At this point, the linter reported the remaining issues, but also reported several files I'd already fixed as having the issue. Manual inspection confirmed that the imports had already been sorted correctly. 

I ignored the issue for the time being and continued through the real issues, eventually reaching the end. Rerunning left only the false positives. No matter how many times I re-ran the linter, it would keep these false positives. 

However, changing our script in the build phase to `${PODS_ROOT}/SwiftLint/swiftlint line --no-cache` resulted in the linter finding lots of files I hadn't got to, and removed all the false positives. 

I have no idea what's going on behind the scenes, but the new cache added in 0.16 seems to be broken.

jpsim:
It's quite possible that the approach used to determine if a file has changed is insufficient. We're currently hashing the file, so maybe the hash isn't affected by a re-ordering of lines?

marcelofabri:
It seems that the hash changes when re-ordering lines: 

```
  1> let str1 = "import Foo\nimport Bar" 
str1: String = "import Foo\nimport Bar"
  2> let str2 = "import Bar\nimport Foo" 
str2: String = "import Bar\nimport Foo"
  3> str1.hashValue
$R0: Int = -300990755179970577
  4> str2.hashValue
$R1: Int = 7298368255116804847
```

dalemy:
Any further progress on this one? We are waiting on 0.17 for a few fixes, and while we could go ahead with it currently, we would need to always disable the cache due to these problems. Obviously that's undesirable for everyone.

marcelofabri:
Not that I'm aware of, but feel free to investigate it deeper! A repro case would be very welcome 😉

jpsim:
> It seems that the hash changes when re-ordering lines

Sorry, it was just a guess. I'll update the issue title to not be so misleading 😉

norio-nomura:
It is a mistake to use `hash` to detect changes in` String`, because `hash` does not use the entire contents of` String`.

An example of `hash` collision:
[...]

Issue #1093, opened 2016-12-30
Rule Request: prefer 'joined()' over 'joined(separator: "")'


jpsim:
since that's the default separator.

aamctustwo:
Does SourceKitten provide sufficient detail to make this rule more general? In other words, detect any instance where a function is being used with its default arguments.

marcelofabri:
I don't think so, we'd have to keep a list with known functions and default parameters.

jpsim:
> Does SourceKitten provide sufficient detail to make this rule more general?

No, it doesn't, because binary Swift modules don't actually contain that information 😱 

For example, looking at the generated interface for the Swift standard library in Xcode itself, you get this:

```swift
extension Array where Element == String {
    public func joined(separator: String = default) -> String
}
```

Notice the `default` in the place where you'd expect to see `""`.

I actually _think_ (I could be wrong) that this has something to do with Swift's [planned ABI stability](https://github.com/apple/swift/blob/master/docs/ABI.rst). But I'm not sure.

marcelofabri:
Closed by @ornithocoder in #1746 🎉

PR #1191, opened 2017-01-14
Support for passing multiple paths to swiftlint lint and autocorrect

benasher44:
This implements #810:

TODO:

- [x] What does `rootPath` mean now (especially in terms of caching) when multiple files are passed? (disabled with a warning)
- [x] Okay to get drop the `--path` option? (decided to keep)
- [x] Add tests. (filed CLI arg testing follow up ticket  #1213)
- [x] Add CHANGELOG entry.

This + pre-commit/pre-commit#467 (personal motivation for this PR) + hooks.yaml (maybe in a follow up PR) makes this work with pre-commit using SwiftPM. I've tested this branch locally with pre-commit, and it works great!

I'd appreciate thoughts and feedback on the above TODOs and existing changes :)

### DIFF
--- a/Source/swiftlint/Extensions/Configuration+CommandLine.swift
+++ b/Source/swiftlint/Extensions/Configuration+CommandLine.swift
@@ -48,15 +48,15 @@ private func autoreleasepool(block: () -> Void) { block() }
 
 extension Configuration {
 
-    func visitLintableFiles(path: String, action: String, useSTDIN: Bool = false,
+    func visitLintableFiles(paths: [String], action: String, useSTDIN: Bool = false,
                             quiet: Bool = false, useScriptInputFiles: Bool, forceExclude: Bool,
                             cache: LinterCache? = nil, parallel: Bool = false,
                             visitorBlock: @escaping (Linter) -> Void) -> Result<[File], CommandantError<()>> {
-        return getFiles(path: path, action: action, useSTDIN: useSTDIN, quiet: quiet, forceExclude: forceExclude,
+        return getFiles(paths: paths, action: action, useSTDIN: useSTDIN, quiet: quiet, forceExclude: forceExclude,
                         useScriptInputFiles: useScriptInputFiles)
         .flatMap { files -> Result<[Configuration: [File]], CommandantError<()>> in
             if files.isEmpty {
-                let errorMessage = "No lintable files found at path '\(path)'"
+                let errorMessage = "No lintable files found at paths: '\(paths.joined(separator: ", "))'"
                 return .failure(.usageError(description: errorMessage))
             }
             return .success(Dictionary(grouping: files, by: configuration(for:)))
@@ -105,7 +105,7 @@ extension Configuration {
     }
 
     // swiftlint:disable function_parameter_count
-    fileprivate func getFiles(path: String, action: String, useSTDIN: Bool, quiet: Bool, forceExclude: Bool,
+    fileprivate func getFiles(paths: [String], action: String, useSTDIN: Bool, quiet: Bool, forceExclude: Bool,
                               useScriptInputFiles: Bool) -> Result<[File], CommandantError<()>> {
         if useSTDIN {
             let stdinData = FileHandle.standardInput.readDataToEndOfFile()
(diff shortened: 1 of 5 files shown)


Source/swiftlint/Extensions/Configuration+CommandLine.swift
```diff
@@ -112,29 +112,30 @@ extension Configuration {
...
+        return visitLintableFiles(paths: options.paths, action: "Linting", useSTDIN: options.useSTDIN,
                                   quiet: options.quiet, useScriptInputFiles: options.useScriptInputFiles,
                                   cache: cache, parallel: true, visitorBlock: visitorBlock)
     }
 
     // MARK: AutoCorrect Command
 
     init(options: AutoCorrectOptions) {
-        self.init(commandLinePath: options.configurationFile, rootPath: options.path, quiet: options.quiet)
+        self.init(commandLinePath: options.configurationFile, rootPath: "", quiet: options.quiet)
```
benasher44: This is temporary to get this to compile and test the multiple paths changes.

Source/swiftlint/Extensions/LinterCache+CommandLine.swift
```diff
@@ -45,7 +45,7 @@ private func cacheURL(options: LintOptions, configuration: Configuration) -> URL
 }
 
 private func defaultCacheURL(options: LintOptions) -> URL {
-    let rootPath = options.path.bridge().absolutePathRepresentation()
+    let rootPath = ""//options.path.bridge().absolutePathRepresentation()
```
benasher44: This is temporary to get this to compile and test the multiple paths changes.

marcelofabri:
> What does `rootPath` mean now (especially in terms of caching)?

In terms of caching, its hash is used as the default cache file name.

benasher44:
Yep. Sorry that TODO is poorly worded. I mean that I'm not sure what it has any meaning in this context. There isn't a useful `rootPath` when you pass multiple files as CLI args, since, even if you calculated the nearest commmon ancestor of all of the paths, you could end up with something like `~` as the `rootPath`. I think it makes sense to just print that caching is disabled in this instance.

benasher44:
Does that seem reasonable? Or is there some way I'm missing that caching behavior should be preserved in this instance?

marcelofabri:
I think either disabling or concatenating all files before getting the hash would be ok.

benasher44:
Okay thanks! I went with concatenating all the files to start. If we're okay dropping `--path` in favor of this, I'll move forward with tests and changelog.

marcelofabri:
> If we're okay dropping `--path` in favor of this, I'll move forward with tests and changelog.

IMHO it'd be better if we could keep `--path` too to avoid breaking any existing setups.

Source/swiftlint/Extensions/LinterCache+CommandLine.swift
```diff
@@ -45,15 +45,15 @@ private func cacheURL(options: LintOptions, configuration: Configuration) -> URL
 }
 
 private func defaultCacheURL(options: LintOptions) -> URL {
-    let rootPath = options.path.bridge().absolutePathRepresentation()
+    let cacheSignature = options.paths.map { $0.bridge().absolutePathRepresentation() }.joined(separator: "")
```
marcelofabri: `""` is the default separator, you should use just `.joined()` (see https://github.com/realm/SwiftLint/issues/1093)

benasher44:
Okay! I've been thinking about concatenating the paths as a cache signature, and it feels a bit odd to me. It seems like it'd be better to disable and say up front that a cache path should be provided, which I think would avoid inconsistent/unexpected caching behavior.

benasher44:
That is, only for the case of multiple paths.

marcelofabri:
> Okay! I've been thinking about concatenating the paths as a cache signature, and it feels a bit odd to me. It seems like it'd be better to disable and say up front that a cache path should be provided, which I think would avoid inconsistent/unexpected caching behavior.

Seems good for me!

benasher44:
Where do I find the tests that test CLI args? In Xcode, I only see the ones that test SwiftLintFramework.

marcelofabri:
> Where do I find the tests that test CLI args?

There're no tests for this 🙃

benasher44:
Ha! Well, there is the note from Danger that says:
>Linting Alamofire with this PR took 3.3s vs 3.33s on master (0% faster)
I'm guessing this is referring to running `swiftlint`? I could add a case that runs it with `--path`, multiple paths, and both. Right?

benasher44:
Oh I see it's just there in the Dangerfile. Hm okay. Any recommendations then? I could file a task to include some smoke tests for basic CLI options?

marcelofabri:
> I could file a task to include some smoke tests for basic CLI options?

Good idea!

---

This seems good to me, but I'd like to get a 👍 from @jpsim on this since it deals with the CLI

benasher44:
Filed #1213

jpsim:
We need to fix our caching strategy before we do this. We need to encapsulate all the things that can affect the cache.

benasher44:
Is there an open issue that summarizes the issues and/or possible solutions? I have some thoughts after working on this, but I think it'd be best to chime in there.

marcelofabri:
There are https://github.com/realm/SwiftLint/issues/1240 and #1184, but I'm afraid there may be more problems that aren't mentioned on them.

benasher44:
@marcelofabri or @jpsim could either of you open a mega thread/issue?

jpsim:
#1184 should be that umbrella issue

benasher44:
Awesome. Thanks!

benasher44:
Updated to only do automatic caching for single directories. Does that make this a breaking change?

jpsim:
Hi @benasher44, sorry this has been on hiatus for a while. I'd like to see us fully address our caching model (#1184) before we make more changes that could affect it.

benasher44:
@jpsim no problem! I'd appreciate another ping here once it's resolved, and then I'll rebase and get this PR back in order. Thanks!

marcelofabri:
This could be revisited now that #1530 was merged

benasher44:
Awesome! Will try to revive it this week! Thanks!

benasher44:
@marcelofabri I've updated with master, and I think this is getting close to being ready to go. With caching issues being resolved, this is definitely more straightforward.

The next thing to figure out is `Configuration.rootPath`. It doesn't appear to be used anywhere, since it was previously more important for caching. The only place I see it used it to compare `Configuration` instances. In the new world of multiple paths, `rootPath` doesn't really make sense (as it stands now), unless we think it's important to calculate the nearest common ancestor path of all paths passed to `swiftlint`. Thoughts? Maybe I should remove `rootPath`?

shepting:
This will be awesome. Looking forward to seeing it merge.

esteluk:
Is there any appetite for reviving this one? Happy to resolve the conflicts if it's no longer on @benasher44's radar.

benasher44:
I'll take a crack at updating this PR this week

benasher44:
Okay this is working again. I don't feel good about having no tests. Any advice on the testing front?

esteluk:
Works great for me!

The output for `swiftlint help lint` is a bit confusing though - is there a way to tidy it up?

```
[[""]]
	list of paths to the files or directories to lint
```

esteluk:
I think there might also be a problem with using --path now? 
My script runs `swiftlint --path foo/bar.swift --config foo/.swiftlint.yml` which throws:
`Missing argument for --config`

benasher44:
@esteluk thanks for that great feedback! I believe I've fixed the bug you found, so it'd be great if you could give this latest update another go.

For the help message issue, I've filed an issue on Commandant: Carthage/Commandant/issues/128

esteluk:
Sorry, had a play and it seems to work for my purposes at least! Not sure if the maintainers can take another look? :)

jpsim:
I think this is a good time to merge this @benasher44. Sorry for the massive inexcusable delay. Mind rebasing first to address the merge conflict? Thanks!

marcelofabri:
Thanks, @benasher44 🎉

benasher44:
🎉 🎉 Thanks for your help @marcelofabri and @jpsim!!

PR #1193, opened 2017-01-15
Trailing closure rule

### LINKED ISSUE #54: Validate trailing closure syntax
Trailing closure syntax should be used whenever possible.

marcelofabri:
Fixes #54

I'm thinking maybe this should be opt-in because it's not clear what style the community prefers. 

* GitHub's style guide doesn't pick a style
*  [Ray Wenderlich's one](https://github.com/raywenderlich/swift-style-guide#closure-expressions) says that trailing closures are preferred
* [Erica Sadun](http://ericasadun.com/2015/11/17/a-handful-of-swift-style-rules-swiftlang/) says that “When a trailing closure argument is functional, use parentheses. When it is procedural, use braces.”

I'm curious to see the OSS check on this rule. ~One improvement that could be made is ignore a violation if there're other closure parameters in the call.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,10 @@
 * Improve `file_header` rule description.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1492](https://github.com/realm/SwiftLint/issues/1492)
+* Add `trailing_closure` opt-in rule that validates that trailing
+  closure syntax should be used whenever possible.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#54](https://github.com/realm/SwiftLint/issues/54)
 
 ##### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,7 @@
 * Improve `file_header` rule description.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1492](https://github.com/realm/SwiftLint/issues/1492)
+
 * Add `trailing_closure` opt-in rule that validates that trailing
   closure syntax should be used whenever possible.  
   [Marcelo Fabri](https://github.com/marcelofabri)
--- a/Rules.md
+++ b/Rules.md
@@ -77,6 +77,7 @@
 * [Switch Case on Newline](#switch-case-on-newline)
 * [Syntactic Sugar](#syntactic-sugar)
 * [Todo](#todo)
+* [Trailing Closure](#trailing-closure)
 * [Trailing Comma](#trailing-comma)
 * [Trailing Newline](#trailing-newline)
 * [Trailing Semicolon](#trailing-semicolon)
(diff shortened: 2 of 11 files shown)


jpsim:
I'm not sure about this one.

marcelofabri:
> I'm not sure about this one.

About this being a rule at all or being a default rule? I feel like it should be opt-in since it should trigger lots of warnings on projects and there's no definitive decision from the community.

jpsim:
Even as opt-in I think it needs work. For example, I wouldn't consider accessing the property after a call to a closure to require it to use trailing closure syntax, such as the first OSSCheck violation reported here:

```swift
XCTAssertEqual(1, schema.objectSchema.filter({ $0.className == "SwiftStringObject" }).count)
```

marcelofabri:
I think that could be a configuration, since I can see teams preferring one style over another.

jpsim:
> Danger found 1046 violations with this PR. Due to GitHub's max issue comment size, the number shown has been truncated to 133.

😳 . Should definitely be opt-in.

Coeur:
does it support the case where we would have an ambiguity between two func?

```
func test(one: (String) -> Void) {}
func test(two: (Int) -> Void) {}
// error: ambiguous use of 'test'
//test { _ in }
// must not warn with SwiftLint
test(one: { _ in })
```

marcelofabri:
@Coeur not really, and I'm not sure we'd able to address that. Another reason to make this opt-in.

jpsim:
@marcelofabri what would you like to do with this?

marcelofabri:
I can revive this, just want to be sure what the behavior should be regarding https://github.com/realm/SwiftLint/pull/1193#issuecomment-274981646

Mazyod:
Not sure if you guys already considered this, but just putting putting it out there. Trailing closure syntax doesn't work within a conditional statement:

```
  4> if a.filter { $0 == 1 }.isEmpty { 
  5.     print("no ones!") 
  6. } 
error: repl.swift:4:15: error: anonymous closure argument not contained in a closure
if a.filter { $0 == 1 }.isEmpty {
              ^

error: repl.swift:4:24: error: consecutive statements on a line must be separated by ';'
if a.filter { $0 == 1 }.isEmpty {
                       ^
                       ;


  4> if a.filter({ $0 == 1 }).isEmpty { 
  5.     print("no ones!") 
  6. } 
```

Coeur:
@marcelofabri if the question is:
> XCTAssertEqual(1, schema.objectSchema.filter({ $0.className == "SwiftStringObject" }).count)
or:
> XCTAssertEqual(1, schema.objectSchema.filter { $0.className == "SwiftStringObject" }.count)

Then second syntax is fine. But again, I would personally not opt-in to this rule, because of the false positive this rule would cause like in the example I gave two months ago, so my opinion is biased.

marcelofabri:
@Mazyod the rule already deals with it

jpsim:
Sounds good @marcelofabri!

jpsim:
Should we file an issue to allow parameterizing the rule to allow/disallow https://github.com/realm/SwiftLint/pull/1193#issuecomment-274981646?

marcelofabri:
Filled https://github.com/realm/SwiftLint/issues/1718.

PR #1199, opened 2017-01-16
Ignore overrides for `variable_name`

### LINKED ISSUE #1169: variable_name should not apply to overriding properties
When I'm overriding a property, I don't have the freedom to change its name. For example, when overriding a UIKit property with a very long name, I can't avoid an error:

```swift
class MyViewController: UIViewController {
    override var preferredInterfaceOrientationForPresentation: UIInterfaceOrientation {
        return .landscapeLeft
    }
}
```

> Variable Name Violation: Variable name should be between 3 and 40 characters long: 'preferredInterfaceOrientationForPresentation' (variable_name)

It doesn't make sense for SwiftLint to ask me to change the property name. If I did, it wouldn't be overriding the superclass property anymore.

aamctustwo:
Resolves #1169 `variable_name should not apply to overriding properties`.

### Summary

Updated `variable_name` to ignore overrides.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 * Improve `unused_optional_binding` rule on tuples check.  
   [Rafael Machado](https://github.com/rakaramos/)
 
+* Updated `variable_name` to ignore overrides.  
+  [Aaron McTavish](https://github.com/aamctustwo)
+  [#1169](https://github.com/realm/SwiftLint/issues/1169)
+
 ##### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


aamctustwo:
@marcelofabri @jpsim Any idea what happened to Danger on this build? 99% faster with this little change makes no sense to me. Also some of the violations "fixed" in other rules that are not the `variable_name` rule. Neither of those things appear to be correct about this PR.

marcelofabri:
@aamctustwo I have no idea, but I triggered it again and it seems we're good now!

PR #1202, opened 2017-01-16
Add minimum_fraction_length to number_separator

### LINKED ISSUE #1200: Configure number_separator for minimum length for either side of decimal
I ran into this while working on a project using coordinates. While I would like the minimum length of 5, it should apply to either side of the decimal.

So taking the example from #1109 I find `1000.01` more readable than `1_000.01` or `10.0003` more readable than `10.000_3`

Maybe even be able to ignore after the decimal.

Is this something others would be interested in as well? Because then I'd look into making a PR.

bjarkehs:
This adds `minimum_fraction_length` as a configuration to `number_separator`. If specified it overrules `minimum_length`.

Closes #1200

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
   [#1169](https://github.com/realm/SwiftLint/issues/1169)
 
 * Update `number_separator` rule to allow for specifying
-  minimum length of fraction.
+  minimum length of fraction.  
   [Bjarke Søndergaard](https://github.com/bjarkehs)
   [#1200](https://github.com/realm/SwiftLint/issues/1200)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@
   [Aaron McTavish](https://github.com/aamctustwo)
   [#1169](https://github.com/realm/SwiftLint/issues/1169)
 
+* Update `number_separator` rule to allow for specifying
+  minimum length of fraction.
+  [Bjarke Søndergaard](https://github.com/bjarkehs)
+  [#1200](https://github.com/realm/SwiftLint/issues/1200)
+
 ##### Bug Fixes
 
 * Fix false positives on `shorthand_operator` rule.  
--- a/Source/SwiftLintFramework/Rules/NumberSeparatorRule.swift
+++ b/Source/SwiftLintFramework/Rules/NumberSeparatorRule.swift
@@ -10,7 +10,7 @@ import Foundation
 import SourceKittenFramework
 
 public struct NumberSeparatorRule: OptInRule, CorrectableRule, ConfigurationProviderRule {
-    public var configuration = NumberSeparatorConfiguration(minimumLength: 0)
+    public var configuration = NumberSeparatorConfiguration(minimumLength: 0, minimumFractionLength: nil)
 
     public init() {}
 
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,11 @@
   [Aaron McTavish](https://github.com/aamctustwo)
   [#1169](https://github.com/realm/SwiftLint/issues/1169)
 
+* Update `number_separator` rule to allow for specifying
+  minimum length of fraction.
```
marcelofabri: missing two trailing whitespaces

Source/SwiftLintFramework/Rules/NumberSeparatorRule.swift
```diff
@@ -113,10 +113,18 @@ public struct NumberSeparatorRule: OptInRule, CorrectableRule, ConfigurationProv
         return prefixes.filter { lowercased.hasPrefix($0) }.isEmpty
     }
 
-    private func isValid(number: String, reversed: Bool) -> (Bool, String) {
+    private func isValid(number: String, reversed: Bool, isFraction: Bool) -> (Bool, String) {
```
marcelofabri: you can remove the `reversed` parameter now, since it's always equivalent to `!isFraction`

Source/SwiftLintFramework/Rules/RuleConfigurations/NumberSeparatorConfiguration.swift
```diff
@@ -9,13 +9,23 @@
 public struct NumberSeparatorConfiguration: RuleConfiguration, Equatable {
     private(set) var severityConfiguration = SeverityConfiguration(.warning)
     private(set) var minimumLength: Int
+    private(set) var minimumFractionLength: Int?
 
     public var consoleDescription: String {
-        return severityConfiguration.consoleDescription + ", minimum_length: \(minimumLength)"
+        let minimumFractionLengthDescription: String
+        if let minimumFractionLength = self.minimumFractionLength {
```
marcelofabri: I don't think you need to use `self.` here

Source/SwiftLintFramework/Rules/NumberSeparatorRule.swift
bjarkehs: Ah right, for some reason my tests wouldn't pass locally even without changing anything. So I decided not to touch it.

Source/SwiftLintFramework/Rules/RuleConfigurations/NumberSeparatorConfiguration.swift
bjarkehs: Yeah, my bad. Code style from current project 😄

marcelofabri:
This seems good to me, but as we're complicating the configuration would you mind adding some tests to make sure we don't break it in the future?

bjarkehs:
Yeah that sounds good. I'll give it a try.

bjarkehs:
So I'm not entirely sure how to write the tests. But I added some triggering and non-triggering examples.

marcelofabri:
> So I'm not entirely sure how to write the tests. But I added some triggering and non-triggering examples.

I hope you don't mind, but I've pushed dc2ebdf which does some improvements. Basically:

* Extract `testNumberSeparator` to another class to avoid the 200-line body length limit error
* Add violation marks on triggering examples
* Add corrections to the tested rule configurations

bjarkehs:
That's perfectly fine, I'll just have a look at your commit to see how its done for the future 👍 Thanks for the help.

mdznr:
I cannot get this to work (version 0.33.1). If I run `swiftlint autocorrect`, it adjusts something like

```swift
NSColor(srgbRed: 0.4355, green: 0.2602, blue: 0.7565, alpha: 1)
```

to 

```swift
NSColor(srgbRed: 0.435_5, green: 0.260_2, blue: 0.756_5, alpha: 1)
```


Config file:
```yml
# Don’t force the underscore thousands separator if the number if less than 5 characters. For example, “2019” shouldn’t be corrected to “2_019”.
number_separator:
  minimum_length: 5
  minimum_fraction_length: 7 # This doesn’t seem to work as of SwiftLint 0.33.1

custom_rules:
```

jpsim:
> Why doesn't this configuration item show up on the website?

Because nothing was being included in the configuration description when the field was unset. Fixed in 54a24638a2402a90358f5f4964476cf666dc1cde.

revolter:
Oh, I thought that it extracts the configuration items automatically (e.g. using reflection). Thanks for the quick fix!

PR #1205, opened 2017-01-17
Update `legacy_constant` to support `CGFloat.pi`

### LINKED ISSUE #1198: Prefer `CGFloat.pi` to `CGFloat(M_PI)`
Swift supports `CGFloat.pi` for a CGFloat version of π as opposed to a Double. It also appears that `M_PI` is being deprecated.

It would be good in the obvious cases where the developer wants a CGFloat version of π to detect and autocorrect to use the new Swift-y constant, avoid the typecast, and make the code more readable.

This may make sense as an extension of `legacy_constant`?

For reference:
- https://github.com/apple/swift/pull/6764
- https://github.com/apple/swift/pull/6802

aamctustwo:
Resolves #1198 `Prefer CGFloat.pi to CGFloat(M_PI)`.

### Summary

- Moved examples for `legacy_constant` into its own file, similar to `type_name`, to allow for Swift 2 vs Swift 3 configurations
- Added `CGFloat.pi` and `Float.pi` support to `legacy_constant`
  - Adjusted validation to only look for word boundary at the beginning rather than also anchoring at the end, to support finding a multiple word pattern such as `CGFloat(M_PI)`
  - Adjusted validation to ensure that the set of SyntaxKind found is a subset of the set of SyntaxKind desired, rather than equal as arrays (as you now have two identifiers rather than only one being matched on occasion).
  - Adjust correction to ensure initial word boundary
- Added Swift 2 specific tests for `legacy_constant` in addition to the regular Swift 3 ones

### DIFF
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -245,7 +245,7 @@ extension File {
         typealias RangePatternTemplate = (NSRange, String, String)
         let matches: [RangePatternTemplate]
         matches = patterns.flatMap({ pattern, template -> [RangePatternTemplate] in
-            return match(pattern: "\\b" + pattern).filter { range, kinds in
+            return match(pattern: pattern).filter { range, kinds in
                 return kinds.first == .identifier &&
                     !ruleEnabled(violatingRanges: [range], for: legacyRule).isEmpty
             }.map { ($0.0, pattern, template) }
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -245,7 +245,7 @@ extension File {
         typealias RangePatternTemplate = (NSRange, String, String)
         let matches: [RangePatternTemplate]
         matches = patterns.flatMap({ pattern, template -> [RangePatternTemplate] in
-            return match(pattern: pattern).filter { range, kinds in
+            return match(pattern: "\\b" + pattern).filter { range, kinds in
                 return kinds.first == .identifier &&
                     !ruleEnabled(violatingRanges: [range], for: legacyRule).isEmpty
             }.map { ($0.0, pattern, template) }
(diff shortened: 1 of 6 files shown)


aamctustwo:
@marcelofabri @jpsim Could you re-run the Travis build without cache? They're all passing for me locally.

marcelofabri:
@aamctustwo It's failing when using the `swiftlint with Swift 2.3` scheme.

Source/SwiftLintFramework/Rules/LegacyConstantRule.swift
```diff
@@ -15,66 +15,63 @@ public struct LegacyConstantRule: CorrectableRule, ConfigurationProviderRule {
...
+            description: "Struct-scoped constants are preferred over legacy global constants.",
+            nonTriggeringExamples: nonTriggeringExamples,
+            triggeringExamples: triggeringExampes,
+            corrections: corrections
+        )
+    }()
 
-        let pattern = "\\b(" + constants.joined(separator: "|") + ")\\b"
+    private static let legacyConstants: [String] = {
+        switch SwiftVersion.current {
```
marcelofabri: could this be refactored to `return Array(legacyPatterns.keys)`?

Source/SwiftLintFramework/Rules/LegacyConstantRule.swift
```diff
@@ -15,66 +15,63 @@ public struct LegacyConstantRule: CorrectableRule, ConfigurationProviderRule {
...
+        case .two:
+            return LegacyConstantRuleExamples.swift2Patterns
+        case .three:
+            return LegacyConstantRuleExamples.swift3Patterns
         }
+    }()
+
+    public func validate(file: File) -> [StyleViolation] {
+        let pattern = "\\b(" + LegacyConstantRule.legacyConstants.joined(separator: "|") + ")"
```
marcelofabri: I know this was already in this way, but you can use a non-capture group here for a performance boost

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -245,7 +245,7 @@ extension File {
         typealias RangePatternTemplate = (NSRange, String, String)
         let matches: [RangePatternTemplate]
         matches = patterns.flatMap({ pattern, template -> [RangePatternTemplate] in
-            return match(pattern: pattern).filter { range, kinds in
+            return match(pattern: "\\b" + pattern).filter { range, kinds in
```
marcelofabri: shouldn't this be changed on the caller?

Issue #1211, opened 2017-01-18
shorthand_operator false positive


keith:
```swift
var foo = [3, 4]
var bar = [1, 2]
foo = bar + foo
```

This generates a warning to use `+=` instead, which doesn't work here because of the ordering.

keith:
@marcelofabri you might be interested in this one.

marcelofabri:
Fixed in #1212.

keith:
Thanks!

Issue #1183, opened 2017-01-13
Shorthand Operator False Positive - ternaries


dalemy:
Not sure where it's going wrong, but here is a sample:

    var angle = atan2(a.y - b.y, a.x - b.x)
    angle = someCheck ? angle : -angle

marcelofabri:
Fixed in #1212.

PR #1212, opened 2017-01-19
Disable commutative operations on shorthand_operator rule

### LINKED ISSUE #1182: Shorthand Operator False Positive - String concatenation
The `+` operator is commutative when working with numbers, but when used for string concatenation it is not. For example, the following code is a violation with the rules currently:

    var helloWorld = "world!"
    helloWorld = "Hello, " + helloWorld

marcelofabri:
Fixes #1182, #1183, #1211


aamctustwo:
Should we add some of the false positives in those Issues to the `nonTriggeringExamples` to help prevent future regressions?

PR #1215, opened 2017-01-19
Fix bool negation false positive for force_unwrap

### LINKED ISSUE #918: force_unwrapping: false positive when using ! as boolean operator
this trivial example `var test = (!false)` would trigger an error condition for force_unwrapping

aamctustwo:
Resolves #918 `force_unwrapping: false positive when using ! as boolean
operator`.


jpsim:
Thanks!

Issue #1185, opened 2017-01-13
Sorted Imports should ignore case (or have it customisable)


dalemy:
I'd expect the following to be valid:

    import AcmeCo
    import ADAL
    import AfterFramework

aamctustwo:
:+1: For adding a customisation option to ignore case. I'd prefer to keeping the case-sensitive sorting as the default however.

marcelofabri:
@dalemy Would you like to try to implement this?

dalemy:
I'd like to implement the various things I've suggested, but getting open source permissions from the company and the time to do it isn't usually possible. And almost never on the time scale that you would expect. For now, assume that I'm unable to take this one.

keith:
I've submitted a non-configurable fix for this: https://github.com/realm/SwiftLint/pull/1218

PR #1218, opened 2017-01-19
Lowercase imports before sorting

keith:
Currently if you have imports that differ in case, they are not sorted
as you would expect.

Fixes https://github.com/realm/SwiftLint/issues/1185

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#663](https://github.com/realm/SwiftLint/issues/663)
 
+* Fix `sorted_imports` rule to sort ignoring case.  
+  [Keith Smiley](https://github.com/keith)
+  [#1185](https://github.com/realm/SwiftLint/issues/1185)
+
 ##### Enhancements
 
 * Performance improvements to `generic_type_name`,
--- a/Source/SwiftLintFramework/Rules/SortedImportsRule.swift
+++ b/Source/SwiftLintFramework/Rules/SortedImportsRule.swift
@@ -19,7 +19,9 @@ public struct SortedImportsRule: ConfigurationProviderRule, OptInRule {
         name: "Sorted Imports",
         description: "Imports should be sorted.",
         nonTriggeringExamples: [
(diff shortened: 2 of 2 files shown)


keith:
Hmmm. These numbers seem to be a little more than I would expect..

jpsim:
The longer benchmarks are more reliable and well within margins (3% for swift stdlib, 6% for Realm), so I'm not concerned about those.

keith:
I can fix the new issue by bumping that up a line if that works for this fix?

keith:
The simple fix for this case would be to sort those SourceKitten imports like this:

```
import Foundation
#if SWIFT_PACKAGE
import SourceKit
#endif
import SWXMLHash
```

jpsim:
I'm uneasy about this, since it changes our definition of what is "sorted".

Going back to the conversation in Slack:

> it's always important to keep in mind why these rules exist. In this case, sorting imports helps find out if a file imports a module, which is served just as well if it's case insensitive, and gets in the way of the user less

Although I can see how this can be interpreted as being okay with changing the definition of what sorted means, what I actually meant was that we should consider a list that's sorted either case sensitive or case insensitive to be sorted, and the rule still serves its purpose.

keith:
But I guess the problem here is that allowing both would mean that it would be acceptable for your imports to not be "sorted" for the way you wanted. It doesn't seem like a single consumer would want both ascii sorting, and "normal" sorting to be acceptable.

jpsim:
so maybe it's a matter of adding a `case_insensitive: true` key to the rule's configuration?

keith:
I'm happy to do that if that's the way you want to go, but besides backwards compatibility I do think it's strange to even allow "sorting" to be determined by the ascii order of characters for this use case.

keith:
@jpsim any solution here you'd be happy with?

jpsim:
Alright, let's do this as-is. Could you please just update the changelog and merge?

keith:
Changelog is updated here? Did you want me to add something else?

keith:
Moved it to the breaking section in the changelog

keith:
Thanks!

PR #1222, opened 2017-01-22
Make mark rule faster

marcelofabri:
By grouping the regexes when linting.


jpsim:
We need a better way to report benchmarks when changes are being done to a single rule...

Maybe we could detect that only changes to `Rules/*.swift` were made, and only run _those_ rules when benchmarking. This could help get rid of some of the variance from other rules in our measurements.

Issue #1238, opened 2017-01-23
SwiftLint 0.16.1 incorrectly converting legacy constant for Swift 2.3


raid5:
I have a project using Swift 2.3 and have installed version 0.16.1 of SwiftLint. When I run `swiftlint autocorrect`, it's incorrectly replacing `CGFloat(M_PI)` with `CGFloat.pi`. In the logs I see the output of `Corrected Legacy Constant` where it made the change to my code and the project won't compile due to this error.

marcelofabri:
I've tried to reproduce it and couldn't. The implementation seems correct too. 

The only reason that could make this happen seems to be the wrong toolchain being used (see https://github.com/jpsim/SourceKitten#how-is-sourcekit-resolved).

Using `TOOLCHAINS=com.apple.dt.toolchain.Swift_2_3 swiftlint autocorrect` should work. Can you test this?

jpsim:
I think the issue might be that since `CGFloat(M_PI)` is valid in both Swift 2 _and_ 3, that it doesn't need to be fenced in a version compile-guard (`#if swift(>=3.0)`), so if you're running SwiftLint in Swift 3 mode, it'll gladly "correct" it to something that doesn't compile in Swift 2.x.

It's not a great experience, but I'd still consider it a "user error".

raid5:
@marcelofabri, yes specifying the `TOOLCHAINS` works and does not make the adjustment to `CGFloat(M_PI`. I wasn't aware this needed to be done as I figured SwiftLint would know what version of Swift to target.

marcelofabri:
In general this shouldn't be needed. Do you have more than one Xcode installed?

jpsim:
> I wasn't aware this needed to be done as I figured SwiftLint would know what version of Swift to target.

SwiftLint doesn't check the version specified in your Xcode project, or a `.swift-version` file at the root of your repo. It _does_ leave code alone if it wouldn't compile under its current Swift version (determined via https://github.com/jpsim/SourceKitten#how-is-sourcekit-resolved).

> In general this shouldn't be needed. Do you have more than one Xcode installed?

This can happen even with just a single version of Xcode installed, since Xcode 8.x ships with both Swift 3 and Swift 3 toolchains

raid5:
@marcelofabri, yes I have 7.3.1 and 8.2.1 as I need both for client work.

raid5:
@jpsim, that makes sense. I may suggest adding something about that to the README.

[...]

PR #1239, opened 2017-01-23
add documentation on working with multiple swift versions in README

jpsim:
Suggested in #1238. /cc @raid5


raid5:
Thanks for this update!

marcelofabri:
I feel like this could be improved from an user view. Some questions that aren't obvious:

* What value should I set an env var?
* In which scenarios would I override the default toolchain?

jpsim:
Good feedback @marcelofabri, I hope I've addressed it with my last commit.

Issue #1175, opened 2017-01-12
unused_closure_parameter autocorrect is too eager


allen-zeng:
When autocorrecting `unused_closure_parameter`, swiftlint also removes the type annotation. This isn't going to work in a situation where the type annotation is necessary, for example, when generics needs to use it for type inference. That is, the autocorrection will make the code fail to compile.

```swift
func genericsFunc<T>(closure: (T) -> Void) {
    print("\(T.self)")
}

genericsFunc(closure: { (int: Int) -> Void in
    // this will print "Int"
})
```

Autocorrect replaces the entire parameter list with an underscore, which isn't valid syntax. It should instead retain the type, and replace the name with an underscore:

```swift
genericsFunc(closure: { (_) -> Void in
    // autocorrected version, it's wrong!
})

genericsFunc(closure: { (_: Int) -> Void in
    // this will compile and still print "Int"
})
```

marcelofabri:
This is kind of hard, because if we always retain the type, I'd say in at least 95% of the cases it could be removed. On the other hand, `autocorrect` shouldn't break any code 🙃

allen-zeng:
Does swiftlint actually work out whether the closure parameter types can be removed? If it doesn't, the correct autocorrection behaviour is to have leave the type annotation. If it's not even supplied then that's a different story.

```swift
func plainFunction(closure: (Int) -> Void) {
    // stuff
}

plainFunction { (param: Int) -> Void in // this will be corrected to { (_: Int) -> Void in }
    // stuff
}

plainFunction { param -> Void in // this will be corrected to { _ -> Void in }
    // stuff
}
```

marcelofabri:
@allen-zeng Do you want to try to submit a PR?

allen-zeng:
Sure! I'll see what I can do :)

marcelofabri:
I took a look at this today and the problem is that `key.nameoffset` and `key.namelength` are 0 in this case:

```json
{
	"key.nameoffset": 0,
	"key.typename": "Int",
	"key.length": 8,
	"key.name": "int",
	"key.kind": "source.lang.swift.decl.var.parameter",
	"key.namelength": 0,
	"key.offset": 25
}
```

We could however check if there's a `key.typename` and avoid correcting it in this case.

allen-zeng:
Yeah I noticed that and thought it was weird. I'm guessing this is a bug in SourceKitten?

[...]

PR #1247, opened 2017-01-25
retain closure parameter types when they are specified during autocorrect

allen-zeng:
This fixes the autocorrect behaviour of unused closure parameters. See #1175 

The function `nameKey(for:)` is removed. It appears that for both Swift 2 and 3, `key.name` is populated correctly.

### DIFF
--- a/Source/SwiftLintFramework/Rules/UnusedClosureParameterRule.swift
+++ b/Source/SwiftLintFramework/Rules/UnusedClosureParameterRule.swift
@@ -30,14 +30,18 @@ public struct UnusedClosureParameterRule: ASTRule, ConfigurationProviderRule, Co
                 "return { migration, schemaVersion in\n" +
                 "rlmMigration(migration.rlmMigration, schemaVersion)\n" +
                 "}\n" +
-            "}"
+            "}",
+            "genericsFunc { (a: Type, b) in\n" +
+                "a + b\n" +
+            "}\n"
         ],
         triggeringExamples: [
             "[1, 2].map { ↓number in\n return 3\n}\n",
             "[1, 2].map { ↓number in\n return numberWithSuffix\n}\n",
             "[1, 2].map { ↓number in\n return 3 // number\n}\n",
             "[1, 2].map { ↓number in\n return 3 \"number\"\n}\n",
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Rules/UnusedClosureParameterRule.swift
```diff
@@ -83,8 +97,7 @@ public struct UnusedClosureParameterRule: ASTRule, ConfigurationProviderRule, Co
 
         return parameters.flatMap { param -> (NSRange, String)? in
             guard let paramOffset = param.offset,
-                let paramLength = param.length,
-                let name = param[nameKey(for: .current)] as? String,
+                let name = param["key.name"] as? String,
```
marcelofabri: you can use `param.name` here

Source/SwiftLintFramework/Rules/UnusedClosureParameterRule.swift
```diff
@@ -93,6 +106,8 @@ public struct UnusedClosureParameterRule: ASTRule, ConfigurationProviderRule, Co
                 return nil
             }
 
+            let paramLength = name.characters.count
```
marcelofabri: you should use `name.bridge().length` instead as `name.characters.count` can return a different value

allen-zeng: Either approach causes SourceKitten to crash [here](https://github.com/jpsim/SourceKitten/blob/b12e1442854eb6d283a4f7e7c09f2207d1652906/Source/SourceKittenFramework/String+SourceKitten.swift#L131). I'll look into why tomorrow.

marcelofabri: This is a known bug: https://github.com/realm/SwiftLint/issues/1006 😱

allen-zeng: Thanks! Saves me from having to look into it :D

jpsim: Should still use `name.bridge().length`.

allen-zeng:
Happy to help! Appreciate your guidance as well! 😃

Issue #136, opened 2015-09-26
SourceKit doesn't return some variable declarations


jakecraige:
I very well could not be understanding swifltint correctly but I'm expecting this to give me a VariableNameRule error and I'm not seeing one. I've tried the latest master and homebrew release

``` sh
printf -- "func fooBar() { let x = 1 }\n" > example.swift; swiftlint lint --path example.swift

$ Linting Swift files at path example.swift
$ Linting 'example.swift' (1/1)
$ Done linting! Found 0 violations, 0 serious in 1 file.
```

When If I do this, I see the error:

``` sh
printf -- "let x = 1 }\n" > example.swift; swiftlint lint --path example.swift

$ Linting Swift files at path example.swift
$ Linting 'example.swift' (1/1)
$ /tmp/example.swift:1:1: warning: Name Format Violation: Variable name should be between 3 and 40 $ $ characters in length: 'x'
$ Done linting! Found 1 violation, 0 serious in 1 file.
```

Is this expected?

jpsim:
Since SwiftLint uses the AST obtained via SourceKit, it will share the same limitations. In this case, SourceKit doesn't identify "x" in `func fooBar() { let x = 1 }` as a variable.

jakecraige:
Interesting. Shall I leave this issue open since it's weird behavior or close it since it's kind of our of our hands unless we want to write some custom AST or work around it?

jpsim:
Sure, let's leave this open, but the only real way to make progress on this will be to move away from using SourceKit, file a radar with Apple about it, or fix it in SourceKit itself if/when its source is released when Swift goes open-source later this year (which is unlikely).

marcelofabri:
I've created [SR-5057](https://bugs.swift.org/browse/SR-5057) to track this.

marcelofabri:
This should be automatically fixed since https://github.com/apple/swift/pull/11431 was merged. I'll keep this open until that PR is included in an official release (probably 4.1?).

marcelofabri:
This was included in Swift 4.1 🎉

PR #1248, opened 2017-01-25
Navigate substructure even for different kinds on ASTRule

### LINKED ISSUE #1237: unused_optional_binding misses ifs inside functions
This example does not trigger the unused_optional_binding:

```swift
func foo() {
    if let _ = bar {
    }
}
```

If you add this test case:

```swift
"func foo() {\nif let _ = bar {\n}\n"
```

You can see that the tests fail to pass. I believe this is because the `if` in this case is nested in the substructure of the func, which isn't handled by the [current logic](https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift#L50)

// @rakaramos

marcelofabri:
Fixes #1237

I expect oss-check to introduce some violations that weren't being caught before.

### DIFF
--- a/Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
+++ b/Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
@@ -24,7 +24,9 @@ public struct UnusedOptionalBindingRule: ASTRule, ConfigurationProviderRule {
             "if let (_, second) = getOptionalTuple() {\n" +
             "}\n",
             "if let (_, asd, _) = getOptionalTuple(), let bar = Foo.optionalValue {\n" +
-            "}\n"
+            "}\n",
+            "if foo() { let _ = bar() }\n",
+            "if foo() { _ = bar() }\n"
         ],
         triggeringExamples: [
             "if let ↓_ = Foo.optionalValue {\n" +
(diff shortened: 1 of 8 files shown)


Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
```diff
@@ -40,23 +42,31 @@ public struct UnusedOptionalBindingRule: ASTRule, ConfigurationProviderRule {
...
 
     public func validate(file: File,
                          kind: StatementKind,
                          dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        let conditionKind = "source.lang.swift.structure.elem.condition_expr"
         guard kind == .if || kind == .guard,
-            let offset = dictionary.offset,
-            let length = dictionary.length,
-            let range = file.contents.bridge().byteRangeToNSRange(start: offset, length: length) else {
+            let elements = dictionary.elements?.filter({ $0.kind == conditionKind }) else {
```
marcelofabri: this was needed to make sure that the pattern was in the condition, not inside an `if` body

keith:
Thanks for submitting this one! I'll test this locally shortly, FWIW the `RedundantNilCoalescingRule` also has the non-recursive issue on master as well.

marcelofabri:
@keith I think that might be another issue because that rule isn't an `ASTRule`

keith:
Sorry, wrong rule, I meant the `RedundantOptionalInitializationRule` which is 😄

keith:
It looks like with this test case in `RedundantOptionalInitializationRule` :

```
            "func foo() {\nvar myVar: Optional<Int>↓=nil\n}\n"
```

This still fails on this branch

keith:
Ah, it looks like there's some duplicate custom logic in that rule that we'll need to fix as well.

marcelofabri:
> Ah, it looks like there's some duplicate custom logic in that rule that we'll need to fix as well.

From a quick look I haven't found a motive to fail. My bet is that it's the same problem as #136.

marcelofabri:
In fact, running `sourcekitten structure`:


```json
{
  "key.substructure" : [
    {
      "key.offset" : 0,
      "key.nameoffset" : 5,
      "key.length" : 47,
      "key.name" : "foo()",
      "key.bodyoffset" : 12,
      "key.kind" : "source.lang.swift.decl.function.free",
      "key.namelength" : 5,
      "key.bodylength" : 34
    }
  ],
  "key.offset" : 0,
  "key.length" : 48
}
```

keith:
Nope, I think this fixes what it meant to. This just didn't fix what I was hoping, but that's not the fault of this code, just my misunderstanding!

jpsim:
🚢 ❗️

PR #1250, opened 2017-01-26
Run clean before starting the docker run

allen-zeng:
I stumbled across the error described [here](https://github.com/realm/SwiftLint/issues/1175#issuecomment-275108850) when I was trying to run `docker_test`. The [solution](https://github.com/realm/SwiftLint/issues/1175#issuecomment-275111430) is simple enough.

It's probably a good idea run clean before docker run anyway.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,7 @@ test: clean bootstrap
 clean:
 	rm -f "$(OUTPUT_PACKAGE)"
 	rm -rf "$(TEMPORARY_FOLDER)"
+	rm -rf "./.build"
 	$(BUILD_TOOL) $(XCODEFLAGS) -configuration Debug clean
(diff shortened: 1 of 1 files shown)


jpsim:
I'd rather not do this, as it means we don't benefit from incremental compilation when running `docker_test` 😞 

The only time I've ever seen an `open: Is a directory` error is when I run `swift test` from macOS, followed by `swift test` from the Docker container. I think the problem here is that the `.xctest` "bundle" is a directory on macOS, but a file on Linux.

More generally, I don't think SwiftPM aims to have compatible directory structures across platforms, so that's to be expected, but it happens to work well for everything else AFAICT.

I think a preferable fix would be to remove the `xctest` directory _only if it is a directory_ when running the `docker_test` make job.

allen-zeng:
Good call. I'll close this.

PR #1251, opened 2017-01-26
Fix redundant_void_return execution if return type starts with Void~

sora0077:
`redundant_void_return` matches wrong pattern like an example below.

```swift
struct Response<T> {}
typealias VoidResponse = Response<Void>

func foo() -> VoidResponse {}
// -> after autocorrect
func foo()Response {}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,9 @@
   [Allen Zeng](https://github.com/allen-zeng)
   [#1175](https://github.com/realm/SwiftLint/issues/1175)
 
+* Fix `redundant_void_return` matches if return type starts with Void~.
+  [Hayashi Tatsuya](https://github.com/sora0077)
+
 ## 0.16.1: Commutative Fabric Sheets
 
 ##### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,7 @@
   [Allen Zeng](https://github.com/allen-zeng)
   [#1175](https://github.com/realm/SwiftLint/issues/1175)
 
-* Fix `redundant_void_return` matches if return type starts with Void~.
+* Fix `redundant_void_return` matches if return type starts with Void~.  
   [Hayashi Tatsuya](https://github.com/sora0077)
 
 ## 0.16.1: Commutative Fabric Sheets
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Rules/RedundantVoidReturnRule.swift
```diff
@@ -41,7 +42,7 @@ public struct RedundantVoidReturnRule: ASTRule, ConfigurationProviderRule, Corre
         ]
     )
 
-    private let pattern = "\\s*->\\s*(?:Void|\\(\\s*\\))"
+    private let pattern = "\\s*->\\s*(?:Void(?=(\\W|$))|\\(\\s*\\))"
```
marcelofabri: maybe adding a word boundary (`\b`) would be better?

sora0077: Thanks! I'll fix it.

Source/SwiftLintFramework/Rules/RedundantVoidReturnRule.swift
```diff
@@ -41,7 +42,7 @@ public struct RedundantVoidReturnRule: ASTRule, ConfigurationProviderRule, Corre
         ]
     )
 
-    private let pattern = "\\s*->\\s*(?:Void|\\(\\s*\\))"
+    private let pattern = "\\s*->\\s*(?:Void(?=\\b)|\\(\\s*\\))"
```
marcelofabri: you can remove the lookahead to make it simpler: `"\\s*->\\s*(?:Void\\b|\\(\\s*\\))"`

CHANGELOG.md
```diff
@@ -41,6 +41,9 @@
   [Allen Zeng](https://github.com/allen-zeng)
   [#1175](https://github.com/realm/SwiftLint/issues/1175)
 
+* Fix `redundant_void_return` matches if return type starts with Void~.
```
marcelofabri: You need to add two trailing spaces as described in https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes

Source/SwiftLintFramework/Rules/RedundantVoidReturnRule.swift
sora0077: Oh! Thanks!!

PR #1252, opened 2017-01-26
Ignore unused_closure_parameter on inline closures

### LINKED ISSUE #1161: unused_closure_parameter false positive on lazy var
When a closure is used in this way, `swiftlint` raise a warning about the variable `lbl` not being used.

```swift
    lazy var label: UILabel = { (lbl: UILabel) -> UILabel in
        lbl.backgroundColor = UIColor.red
        return lbl
    }(UILabel())
```

marcelofabri:
Fixes #1161


jpsim:
👍 💯

PR #1255, opened 2017-01-27
Update README for Cocoapods script

bjarkehs:
The script doesn't work out of the box if pods are not updated. In our project we don't want the build to fail because people haven't run pod update.

Another issue is also that the current script doesn't have `""` around the path, which causes it to fail when there are spaces in ${PODS_ROOT}.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,9 +80,17 @@ fi
 ![](assets/runscript.png)
 
 Alternatively, if you've installed SwiftLint via CocoaPods the script should look like this:
-
 ```bash
-${PODS_ROOT}/SwiftLint/swiftlint
+"${PODS_ROOT}/SwiftLint/swiftlint"
+```
+
+or if you don't want the build to fail if `pod install` hasn't been run:
+```bash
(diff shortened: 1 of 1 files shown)


marcelofabri:
Thanks for doing this, but I don't think that not keeping the dependencies sync'ed a good idea. In fact, CocoaPods adds a build phase itself to guarantee that:

#### `[CP] Check Pods Manifest.lock`
```bash
diff "${PODS_ROOT}/../Podfile.lock" "${PODS_ROOT}/Manifest.lock" > /dev/null
if [ $? != 0 ] ; then
    # print error to STDERR
    echo "error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation." >&2
    exit 1
fi
```


However, adding the quotes around the path is a good idea. Could you send another PR doing just that?

Thanks! 🚀

bjarkehs:
Yeah sure. The only reason I ran into the issue with swiftlint is because we keep it out of the repo, whereas other dependencies are kept in the repo. So if you just download the code and build, then it fails due to swiftlint not being present.

I'll create the new PR 👍

PR #1256, opened 2017-01-27
Update README for Cocoapods script

bjarkehs:
The path should be surrounded by `""` to fix issues with spaces in `PODS_ROOT` path.

E.g. right now the script would fail if swiftlint is located in path: `/Users/me/Awesome Project/Pods/Swiftlint/swiftlint`

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ fi
 Alternatively, if you've installed SwiftLint via CocoaPods the script should look like this:
 
 ```bash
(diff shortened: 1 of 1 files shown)


bjarkehs:
This is the fix for path also implemented in #1255 , but without making assumptions about how the script should handle missing `swiftlint`.

marcelofabri:
Thanks! 💯

PR #1259, opened 2017-01-28
Fix false positive on large_tuple when using generics inside a tuple

### LINKED ISSUE #1257: large_tuple: false positive on tuples with generics having 2 & more type parameters
I have a warning: Large Tuple Violation: Tuples should have at most 2 members. (large_tuple)
on this:
`func getDictionaryAndInt() -> (Dictionary<Int, String>, Int)?`

But there are only 2 members in the tuple.

swiftlint version: 0.16.1

marcelofabri:
Fixes #1257


jpsim:
👍 💯

PR #1260, opened 2017-01-29
Add for_where rule

### LINKED ISSUE #1228: Rule Request: Use where in loop instead of if
For better readable code in my opinion it would be nice with a where rule for loops like. 

I think it will make it more readable and there is less lines. Of course if there is multiple if or and if else it should be ignored

```swift
for user in users where user.id == "1" {
// Awesome code
}
```

instead of
```swift
for user in users {
    if user.id == "1" {
        // Awesome code
   }
}
```

marcelofabri:
Fixes #1228

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,11 @@
   [Jon Shier](https://github.com/jshier)
   [#1261](https://github.com/realm/SwiftLint/issues/1261)
 
+* Add `for_where` rule that validates that `where` is used in a `for` loop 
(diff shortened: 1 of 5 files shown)


jpsim:
Isn't it odd that OSSCheck didn't uncover any violations of this?

marcelofabri:
> Isn't it odd that OSSCheck didn't uncover any violations of this?

I think it's because of #1248

jpsim:
Merged in f0c8fba3b6343d1b0272817b2bd0e66c2b91f944. Thanks!

PR #1267, opened 2017-02-01
Disable automatic caching

benasher44:
As mentioned in #1184, disabling automatic caching would make `swiftlint`'s caching situation simpler as well as serve a few goals:

1. Running `swiftlint lint`from the CLI now behaves the same across multiple invocations (i.e. won't hit a cache the second time without a `.swiftlint.yml` specifying a `cache_path`).
2. This makes caching simpler in the case of passing multiple paths as arguments for #1191, since no cache key/url has to be determined based on the paths being linted in the case where there is no `cache_path` specified.
3. Caching is a feature that makes the most sense in the context of a project, in which case there should be a `.swiftlint.yml` (`cache_path` can be easily set), so automatic caching shouldn't be needed (again just reducing complexity).

I realize that this might be a bit premature, but I thought it might be useful to have the scope of the change out in the open. Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1061](https://github.com/realm/SwiftLint/issues/1061)
 
+* Disable automatic caching  
+  [Ben Asher](https://github.com/benasher44)
+
 ##### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


benasher44:
I think this would unblock #1191.

benasher44:
Closing after further discussion in #1184

PR #1276, opened 2017-02-02
Dramatically speed-up cache by using file modification dates instead of hashes

victorpimentel:
Right now Swiftlint uses `String.hash` to detect if a file has changed, and if the hash has not changed, it avoids executing all rules and returns previously cached violations.

`Swift.hash`, as the rest of Foundation hashes methods, is not a particularly strong hash function, and it results in many hash collisions, which results in stale cache (#1184). For example, in our project when we sort the imports the hashes of the resulting files are always the same 😢 

This pull request fixes that by avoiding all hashing and instead **it detects file changes using the file modification dates** given by the system. This method is much more reliable and also faster, because we don't need to read the file contents at all if the file did not change.

For our iOS app (@tuenti) this change makes Swiftlint a breeze to use in our daily developer lives. For a fully cached run these are roughly the times:

**Before:**

```shell
→ time swiftlint

Done linting! Found 0 violations, 0 serious in 643 files.

real	0m7.322s
user	0m17.235s
sys	0m1.250s
```

**After:**

```shell
→ time swiftlint
Done linting! Found 0 violation, 0 serious in 0 files.

real	0m4.031s
user	0m2.415s
sys	0m0.587s
```

### Notes
- It does not print the name of unchanged files if they had no violations, which I think it's not a big deal since the integration with Xcode and other CI tools will still work. However, if the project does not contain any lintable files, Swiftlint will act like now: failing with a error message. In any case, this may be a breaking change because the output of the tool is different.
- When the times do not make sense (because the user changes the time, for example), we avoid using the cache.
- I'm storing the dates in JSON as time intervals, which is not the most portable thing between machines. I tried to use SwiftDate to store instead the iso8601 date representation but SwiftDate is not Linux compatible yet. I may be wrong, but I don't think this is a big problem since this file is a cache usually not committed in projects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,8 @@
   `_ = foo()`.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1232](https://github.com/realm/SwiftLint/issues/1232)
+* Accept global and local variables in `implicit_getter` rule.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
 
 ##### Bug Fixes
 
--- a/Source/SwiftLintFramework/Rules/ImplicitGetterRule.swift
+++ b/Source/SwiftLintFramework/Rules/ImplicitGetterRule.swift
@@ -13,7 +13,7 @@ private func classScoped(_ value: String) -> String {
     return "class Foo {\n  \(value)\n}\n"
 }
 
-public struct ImplicitGetterRule: Rule, ConfigurationProviderRule {
+public struct ImplicitGetterRule: ConfigurationProviderRule {
     public var configuration = SeverityConfiguration(.warning)
 
     public init() {}
@@ -47,7 +47,8 @@ public struct ImplicitGetterRule: Rule, ConfigurationProviderRule {
         triggeringExamples: [
             classScoped("var foo: Int {\n ↓get {\n return 20 \n} \n} \n}"),
             classScoped("var foo: Int {\n ↓get{\n return 20 \n} \n} \n}"),
-            classScoped("static var foo: Int {\n ↓get {\n return 20 \n} \n} \n}")
+            classScoped("static var foo: Int {\n ↓get {\n return 20 \n} \n} \n}"),
+            "var foo: Int {\n ↓get {\n return 20 \n} \n} \n}"
         ]
     )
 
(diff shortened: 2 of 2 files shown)


Source/SwiftLintFramework/Extensions/NSFileManager+SwiftLint.swift
```diff
@@ -31,4 +32,18 @@ extension FileManager: LintableFileManager {
             return nil
         } ?? []
     }
+
+    public func isFile(atPath path: String, modifiedSince: Date) -> Bool {
+        // If there is any problem reading the modification date, prefer to lint the file
+        do {
+            let attr = try attributesOfItem(atPath: path)
+            if let fileModificationDate = attr[FileAttributeKey.modificationDate] as? Date {
+                return fileModificationDate > modifiedSince
```
phoney: Actually I commented on this earlier. I think the date comparison should be != and not >. It's always possible to reset the clock to an earlier time, on purpose or by accident. I think some of your other logic checking for inconsistent last run date could probably be removed if you simply linted the file if the dates don't match.

Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -13,12 +13,31 @@ public enum LinterCacheError: Error {
...
+    case inconsistentLastRunDate
 }
 
 public final class LinterCache {
     private var cache: [String: Any]
     private let lock = NSLock()
 
+    public var lastRunDate: Date? {
+        get {
+            lock.lock()
```
phoney: Is it possible to write an unlocker class that would own a NSLock instance and unlock it in its deinit method, like a C++ smart pointer? Having these unlock()s everywhere seems messy and error prone.

phoney:
Is there an explanation for why Wordpress-IOS is slower? Does it have a lot of files or a lot of folders?

Out of curiosity does the tool give different times when run as sudo?

marcelofabri:
> Is there an explanation for why Wordpress-IOS is slower?

It might be just an issue with Circle CI. I'd not treat these results as 100% reliable.

Source/SwiftLintFramework/Extensions/NSFileManager+SwiftLint.swift
victorpimentel: That would be another strategy (honestly I don't have a strong opinion in either direction), but for that we would need to store the previous file modification date for every file. Right now this solution just stores the last run date, so it cannot use a !=. That would be completely doable, though.

I found easier to just be conservative about using the cache or not: in any case this is a just a cache, so we always can throw it out and start again. The only issue that I can think of is in an aberrant situation where the clock is constantly changing. Again, I'm ignorant about if that could really happen :/

Source/SwiftLintFramework/Models/LinterCache.swift
victorpimentel: Yeah, and also we could use a safe dictionary where the operations used that lock. However, I wanted to reduce the scope of this change.

Also, I though that this particular attribute would be a really nice place to use defer to ensure unlocking the lock

victorpimentel:
> It might be just an issue with Circle CI. I'd not treat these results as 100% reliable.

Yep, Circle CI times depend on the load that their servers has in that particular moment. In my experience that's the usual issue with performance tests: it's really hard to ensure the same exact conditions in every run.

That was why I was asking if someone was so nice to run this pull request against big apps like ours, where this change really makes a difference.

jpsim:
Nice work, I was hoping someone would take the time to do this :smile:. I have a few points of feedback though:

---

There are many API breaking changes that aren't reflected in the "Breaking" section of the changelog. Here's what I found:

```md
* Add `isFile(atPath:modifiedSince:)` function requirement to
  `LintableFileManager` protocol.  
  [Victor Pimentel](https://github.com/victorpimentel)

* Replace `Configuration.lintableFiles(inPath:)` function with
  `Configuration.lintableFiles(withPaths:cache:fileManager:)`.  
  [Victor Pimentel](https://github.com/victorpimentel)

* Remove `fileHash` parameter from `LinterCache.cache(violations:forFile:)` and
  `LinterCache.violations(forFile:)` functions.  
  [Victor Pimentel](https://github.com/victorpimentel)
```

---

This doesn't complete the cache. Necessarily need to be addressed in this PR, but could.

There's still potential for stale caches when getting a hash collision from a modified configuration file.

We're also still not considering command line arguments in the cache.

---

Some of the breaking API changes are clear to me why they're necessary, and neither is the change in output behavior.

jpsim:
I've started investigating a different way to do this in the [jp-modified-time](https://github.com/realm/SwiftLint/tree/jp-modified-time) branch. It's not meant to replace your work, but rather to explore different options. I'd appreciate if you could take a look and see if there are concepts you want to apply to your PR.

victorpimentel:
@jpsim Sorry, I forgot that SwiftLint is also a framework.

Those API changes are due to the architecture of the framework, more exactly due to the File class from SourceKitten (I think that's yours 😛 ). All those objects read the whole contents of the file in the init, so that means that right now you cannot have instances of that class without reading that file. The main speedup in these changes is due to avoiding having to read the contents of those files. Maintaining the same output for those cases is very doable, but it will probably require further refactoring.

In any case for my particular use that is not the important change that I need, I just **need** to have a reliable cache because as it stands it is painful using Swiftlint in our daily tasks. Honestly, avoiding to read files is not a big deal to me compared with the stale cache.

Please tell me how do you want to proceed, and which tradeoffs should we accept. I will happily help (in a new pull request/branch or modifying this one) if the next version or two end up with a more reliable cache 😄

jpsim:
Sorry for the delay @victorpimentel, yeah the incorrect/stale cache issue is certainly much more pressing than avoiding opening the files. Do you think you can revive this and move it forward?

victorpimentel:
It's a pity that the linting times could not be improved in the new pull request, but I suppose it's better to make one change at a time.

Maybe we can tackle it in a further pull request, for our use case (linting projects with hundreds/thousands of lines), the gains are noteworthy.

PR #1245, opened 2017-01-25
Extract benchmark to a custom script

marcelofabri:
Fixes #1244

PR #1283, opened 2017-02-03
small Danger/oss-check improvements

jpsim:


marcelofabri:
What do you think about cherry-picking https://github.com/realm/SwiftLint/pull/1248/commits/e69c1f3e795ff13b3d95e50a90a8cca23972004a to this PR as well?

jpsim:
I think #1245 introduced some false reports in OSSCheck (see the `missing_docs` fixes/regressions above).

marcelofabri:
I've seen that before as well, but have no idea why sometimes it happens. Maybe something with the docs-related rules?

PR #1284, opened 2017-02-03
disable docs rules in Swift 2.3 and later

jpsim:
Related to #728.

### DIFF
--- a/Source/SwiftLintFramework/Rules/MissingDocsRule.swift
+++ b/Source/SwiftLintFramework/Rules/MissingDocsRule.swift
@@ -155,6 +155,10 @@ public struct MissingDocsRule: OptInRule {
     )
 
     public func validate(file: File) -> [StyleViolation] {
+        guard SwiftVersion.current == .two else {
+            warnMissingDocsRuleDisabledOnce
+            return []
+        }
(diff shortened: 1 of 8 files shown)


Source/SwiftLintFramework/Rules/MissingDocsRule.swift
```diff
@@ -155,6 +155,10 @@ public struct MissingDocsRule: OptInRule {
     )
 
     public func validate(file: File) -> [StyleViolation] {
+        guard SwiftVersion.current == .two else {
+            warnMissingDocsRuleDisabledOnce
```
marcelofabri: this is very clever 😅

jpsim: Swift gives a `dispatch_once` fixit recommending to do this for Swift 3. It's what we do for SourceKit initialization in `Request.swift` too.

marcelofabri: I don't think it's bad, but I definitely took a while to understand what's going on. I didn't know it was in a fixit, so maybe it'll become more popular.

PR #1286, opened 2017-02-04
Reimplement switch_case_on_newline as an ASTRule

### LINKED ISSUE #1268: switch_case_on_newline should support `where` in a new line
Example:
```swift
enum SampleError: Error, Equatable {
	
	case boom
	case bang
	case ouch
	case badaboom(code: Int, description: String)
	
	static func == (lhs: SampleError, rhs: SampleError) -> Bool {
		
		switch (lhs, rhs) {
			
		case let (.badaboom(code: lhsErrorCode, description: _), .badaboom(code: rhsErrorCode, description: _))
			where lhsErrorCode == rhsErrorCode:
			
			return true
			
		case (.boom, .boom),
		     (.bang, .bang), (.ouch, .ouch):
			
			return true
			
		default:
			
			return false
			
		}
		
	}
	
}
```

1. false positive in first case - `case let (.badaboom(...`.
2. no warning on the second line of second case: `(.bang, .bang), (.ouch, .ouch):`.

swiftlint version: 0.16.1

marcelofabri:
Fixes #1268

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,15 @@
 
 ##### Breaking
 
-* None.
+* `variable_name` rule (`VariableNameRule`) is now `identifier_name`
+  (`IdentifierNameRule`) as it validates other identifiers as well.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#663](https://github.com/realm/SwiftLint/issues/663)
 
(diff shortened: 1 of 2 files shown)


marcelofabri:
The false negatives are a consequence of #1248

PR #1287, opened 2017-02-05
Accept global and local variables in `implicit_getter` rule

marcelofabri:


jpsim:
Almost all the violations reported by OSSCheck are due to using `@inline(__always)` or other attributes, which require the explicit `get` keyword. I've filed #1309 to address this, but it shouldn't prevent us from merging this now.

PR #1288, opened 2017-02-06
Add RuleKind to RuleConfiguration

### LINKED ISSUE #1282: Rules should explain their reason for existing
For example, why does the `empty_count` rule exist? Is it purely a stylistic preference? Is the compiler able to better optimize `isEmpty`?

Why do we have `shorthand_operator`? Is there some official Swift guide somewhere that states shortand syntax is more idiomatic?

These are the kinds of things I'd like to know when choosing what rules to disable or opt-in. This could potentially lead to less disabled rules in general.

The simplest thing would be for rules to be able to include a rationale field that can be viewed somehow, likely using the `rules` command.

If we wanted to get a little more fancy, maybe rules could be annotated with attributes such as `stylistic` for rules that are purely subjective, and provide no objective benefit either way. `idiomatic` could be used for rules that are considered idiomatic Swift. `stability` might be applied to rules that avoid stability issues such as `weak_delegate`.

marcelofabri:
Partially fixes #1282

In some of the rules I was in doubt between 2 kinds, so this can/should be validated.


jpsim:
I definitely like this direction, that's the right way to do this IMO./

norio-nomura:
> In some of the rules I was in doubt between 2 kinds, so this can/should be validated.

If Kind would be protocol, rule could be adapt to multiple kinds.

marcelofabri:
> If Kind would be protocol, rule could adapt to multiple kinds.

My initial feeling is that from an user perspective it's better that a rule can belong to only one kind, even if some of them are in a "gray area".

Source/SwiftLintFramework/Protocols/Rule.swift
```diff
@@ -8,10 +8,19 @@
...
+    case style
+    case metrics
+    case performance
+}
+
 public protocol Rule {
     init() // Rules need to be able to be initialized with default values
     init(configuration: Any) throws
     static var description: RuleDescription { get }
+    static var kind: RuleKind { get }
```
masters3d: What about using an option set? I think the advantage is that a rule can be classifed with multiple "labels" without those labels having to be protocols. I like the idea of protocols as labels too. 
The emum restriction kind of reminds me of  http://www.jessesquires.com/enums-as-configs/

marcelofabri: I still think that from an user point of view, it's better that a rule can only belong to one category.

This way we could easily in the future allow enabling/disabling full categories that once. That would be tricky if a rule was in two groups: what would be the expected behavior?

marcelofabri:
@jpsim Is this something that we want to move forward with? 

Revisiting this, I'd add `kind` in `RuleConfiguration` instead.

Source/SwiftLintFramework/Protocols/Rule.swift
masters3d: What about config options that change behavior?

marcelofabri: I'm not sure I understand, can you elaborate?

masters3d: Rules with configurations that make the rule be a different category depending on what config is given.

marcelofabri: Can you think of any examples?

masters3d: Colon, Comma.

marcelofabri: for me they'd be style rules no matter the configuration

marcelofabri:
I've rebased this and moved the definition to `RuleDescription`. Basically now we should:

- [ ] Validate that the kinds (do they meaningfully divide the rules?)
- [ ] Validate each rule kind
- [ ] Choose what to do with custom rules. Currently it's always flagged as `.style`. Options:
    * Allow `kind` to be specified in configuration? (What if it's not specified?)
    * Create a new kind `.custom`?

PR #1295, opened 2017-02-09
[WIP] Fixed sorted_imports rule to take into account all import declaration variants

### LINKED ISSUE #1269: sorted_imports rule is not caring about different import flavors
Imports can declare which kind of type wants to imports, see [Import Declaration from Apple Docs](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Declarations.html#//apple_ref/doc/uid/TP40014097-CH34-ID354). Additionally imports could be marked with `@testable`. 

Personally I'd like having `@testable` at the end of imports list and grouping imports with `kind` declaration by `kind` after normal imports 

1. I would expect that next case don't throw. 
   ```swift
   import UIKit
   @testable import MyApp
   ```

2. I would expect that next case throws
    ```swift
    import AVKit
    import struct MyApp.MyStruct
    import enum MyApp.MyEnum
    import class Foundation.NSArray
    import UIKit
    ```

3. Next case would be the valid version of previous case. Actually it is not throwing because `sorted_imports` rule is case sensitive. I would also expect it to be case insensitive as pointed out in this issue [#1185](https://github.com/realm/SwiftLint/issues/1185)
    ```swift
    import AVKit
    import UIKit
    import class Foundation.NSArray
    import class MyApp.MyClass
    import enum MyApp.MyEnum
    import struct MyApp.MyStruct
    ```

4. I would expect next case don't throw
    ```swift
    import AVKit
    import test
    import class Foundation.NSArray
    @testable import MyApp
    ```

m-revetria:
Fixes #1269 & #1185.
Additionally, added two new rules: `duplicated_imports` and `imports_at_top`

### Added
* `duplicated_imports` rule which checks for duplicated imports
* `imports_at_top` rule which checks if all imports are placed at top of the file regardless of comments

### Fixes
* `sorted_imports` rule: 
   - `ignore_case` option added to configuration, by default the rule will be case insensitive.
   - Grouping imports by importation kind (`class`, `enum`, `func`, etc., [Import Declaration from Apple Docs](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Declarations.html#//apple_ref/doc/uid/TP40014097-CH34-ID354))
   - Checking if `@testable` imports are at the end of the import list.

   ```swift
   // Example of a non-triggering import list
   import AAA
   import bbb

   import enum AAA.Enum
   import enum bbb.Enum

   import func AAA.Func

   import protocol bbb.Protocol

   import struct AAA.Struct

   import typealias AAA.Typealias

   import var CCC.var

   @testable import bbb
   @testable import CCC
   ```


jpsim:
Thanks for the PR @m-revetria! What's the state of this now? The OSSCheck comment has lots of noise so I'd rather not bang my head trying to make sense of all this if this is still work in progress.

m-revetria:
@jpsim this is ready to go. I fixed some issues that were appearing like I forgot to update the changelog or some conflicts with the master branch.

I'm not sure how to proceed with all these warnings while linting other libraries. What should I do in those cases?

Thanks

m-revetria:
Hello @jpsim, is any progress on reviewing this? Is anything I should do in order to go forward?

jpsim:
I'm just reviewing the output from OSSCheck now, and it mostly looks acceptable. Here are a few violations this introduced that I don't think are quite right:

* https://github.com/apple/swift/blob/dfd3b744d896872a3cd6c41f87b6650f379d94ae/stdlib/public/SDK/Dispatch/Dispatch.swift#L14:8
* https://github.com/apple/swift/blob/dfd3b744d896872a3cd6c41f87b6650f379d94ae/stdlib/public/SDK/Foundation/Data.swift#L33:12
* https://github.com/apple/swift/blob/dfd3b744d896872a3cd6c41f87b6650f379d94ae/stdlib/public/SDK/ObjectiveC/ObjectiveC.swift#L15:8
* https://github.com/apple/swift/blob/dfd3b744d896872a3cd6c41f87b6650f379d94ae/stdlib/public/SDK/SafariServices/SafariServices.swift#L14:8
* https://github.com/apple/swift/blob/dfd3b744d896872a3cd6c41f87b6650f379d94ae/stdlib/public/SDK/SpriteKit/SpriteKit.swift#L14:8
* https://github.com/apple/swift/blob/dfd3b744d896872a3cd6c41f87b6650f379d94ae/stdlib/public/SDK/XPC/XPC.swift#L14:8
* https://github.com/apple/swift/blob/dfd3b744d896872a3cd6c41f87b6650f379d94ae/stdlib/public/SDK/XCTest/XCTest.swift#L16:8
* https://github.com/kickstarter/ios-oss/blob/7a7acec0535f98625a1330cedaaf858542585d85/Library/UIAlertController.swift#L3:8
* https://github.com/Moya/Moya/blob/fd2a33236166bb1b1f95686c4a4671c8e0e302ff/Sources/Moya/Image.swift#L5:5

I'm also wondering if it'd be useful to allow configuring whether `@testable` imports should go before or after, as different projects prefer different settings.

mlilback:
It should ignore sorting order for rules in #ifdefs or an option to do so. For example

```
#ifdef os(OSX)
import AppKit
#else
import UIKit
#endif
```

m-revetria:
Hi @jpsim, @mlilback, sorry for the long time to reply. 

I think we can allow developers to set what to do with `@testable` in 3 diff way:
* All of them should go before other imports:
   ```swift
   @testable import MyModule
   import AVKit
   ```
* All of them should go after other imports:
   ```swift
   import UIKit
   @testable import MyModule
   ```
* The `@testable` is ignored and imports are then sorted jsut by their name
   ```swift
   import AVKit
   @testable import MyModule
   import UIKit
   ```

Regarding nested imports in `#if` macros, I'll add an option to just ignore them. By default, I would let the rule to check those imports as well.

What do you think?

jpsim:
The suggestions in your last comment make sense to me @m-revetria!

SwiftLint.xcodeproj/xcshareddata/xcschemes/swiftlint.xcscheme
```diff
@@ -84,6 +84,14 @@
...
             isEnabled = "YES">
          </CommandLineArgument>
+         <CommandLineArgument
+            argument = "--path /Users/remer/Workspace/ios/swiftlint-test"
+            isEnabled = "YES">
+         </CommandLineArgument>
+         <CommandLineArgument
+            argument = "--config /Users/remer/Workspace/ios/swiftlint-test/.swiftlint.yml"
+            isEnabled = "YES">
+         </CommandLineArgument>
```
Coder-256: I think you might have forgot to remove this...

Coder-256:
Are there any plans to implement this? I think it would still be useful.

Dschee:
@m-revetria Do you still have plans to complete this implementation? I'd be happy to review changes.

If @m-revetria doesn't have the time to continue this, is there anyone else who would post a new PR with his own solution to fix https://github.com/realm/SwiftLint/issues/1269 (you could decide to use the code from this PR or begin yourself).

Also please note, that in my opinion this PR actually should be 5 separate PRs:

1. Adding a `case_sensitive` option to the `sorted_imports` rule
2. Adding an `testable_order` option with the three possible values `above`, `below` and `mixed`
4. A new `duplicated_imports` rule
5. A new `imports_at_top` rule

PR #1296, opened 2017-02-09
Line length rule configuration console description

mlwelles:
I forgot to change the console description for line line rule configuration when I added the ignores_function_declarations and ignores_comments in my last pull request. This fixes.


jpsim:
I updated the console description in #1300. Thanks!

PR #1311, opened 2017-02-15
Fix stale cache by using file modification dates instead of hashes

victorpimentel:
This pull request merges the efforts from #1276 and the [jp-modified-time](https://github.com/realm/SwiftLint/tree/jp-modified-time) branch by @jpsim. I tried to follow all the changes made by @jpsim and basically fixed unit tests. Also, I am not avoiding reading unchanged files, which brings the times back to their original times 😢 

Right now Swiftlint uses `String.hash` to detect if a file has changed, and if the hash has not changed, it avoids executing all rules and returns previously cached violations.

`Swift.hash`, as the rest of Foundation hashes methods, is not a particularly strong hash function, and it results in many hash collisions, which results in stale cache (#1184). For example, in our project when we sort the imports the hashes of the resulting files are always the same 😢

This pull request fixes that by avoiding all hashing and instead **it detects file changes using the file modification dates** given by the system. This method is much more reliable.

### Notes
- The location for the cache file changes from Application Support to the Caches directory.
- I'm storing the dates in JSON as time intervals, which is not the most portable thing between machines. I tried to use SwiftDate to store instead the iso8601 date representation but SwiftDate is not Linux compatible yet. I may be wrong, but I don't think this is a big problem since this file is a cache usually not committed in projects.
- In order to test the LinterCache file, I need to expose the FileManager dependency so that I can replace it in tests. I don't know what is the recommended way for this project, I just did it in the less possible intrusive way for the rest of the code (so no more breaking API changes were needed).


marcelofabri:
LGTM

victorpimentel:
I did update the branch with small conflicts in the CHANGELOG.md. I'm still suffering a bit this bug, will be nice if it were merged back :)

victorpimentel:
Sorry to reiterate this, but, @jpsim, what is the status on this? Is this change considered too risky? Any changes that I should do to be less risky? Will this issue be addressed in a different way in a future release?

I still suffer from this cache issue from time to time, and I would like to help fixing it :/

drodriguez:
@marcelofabri seems like a good fix for the cache problems.

marcelofabri:
Merged in #1530 💯

PR #1315, opened 2017-02-17
[WIP] Unhandled Throw Rule

ArtSabintsev:
This rules is for scenarios when a throwing function is defined, but an exception is never explicitly handled. 

Violating Throwing Function:
```swift
func f() throws {
    // Developer does stuff here, but forgets to call `throw anError`
}
```

I opted to not add a suggested correction (e.g., `CorrectableRule` protocol conformation), as I believe this scenario occurs when developers forget to handle an exception, rather than having second thoughts and opting to change the method signature.

I originally posted this at https://bugs.swift.org/browse/SR-3818 to be added to the language/compiler. As it seems to be a bit complicated to implement at a lower level, I've opted to take matters into my own hands and make it a SwiftLint rule.


Source/SwiftLintFramework/Rules/UnhandledThrow.swift
```diff
@@ -0,0 +1,71 @@
...
+            "func f() throws {\n}\n",
+            "func f() throws {\n \n}\n",
+            "func f() throws -> Any {}\n",
+            "func f() throws -> Any { }\n",
+            "func f() throws -> Any {\n}\n",
+            "func f() throws -> Any {\n \n}\n",
+            "func f() throws { var x = 0 }\n",
+            "func f() throws {\n var x = 0 \n}\n",
+            "func f() throws -> Any { var x = 0 }\n",
+            "func f() throws -> Any {\n var x = 0 \n}\n"
```
aamctustwo: What about a case like the one below? I would see this as a triggering example as the function `foo` will never throw.

```swift
func foo() throws {
  do {
     try bar()
  } catch {}
}
```

ArtSabintsev: Handled it in the latest PR due to errors throw via Danger.

aamctustwo: I just tried again with that commit. It isn't triggering the rule, but it should. Try adding this to your array of triggering examples:

`"func foo() throws {\n\t do {\n\t\t try bar() \n\t } catch {} \n}"`

ArtSabintsev: Ah, ok, I understand. Will have a fix for it soon or in the morning. Thanks!

marcelofabri:
I haven't reviewed it fully yet, but I'd argue that overriden functions should be ignored.

Ideally, functions from protocols would too, but that seems harder.

ArtSabintsev:
@marcelofabri Already working on the solution for protocols. As for functions marked with `override`, I don't really agree that it should be exempt by default. When you override a function, you are agreeing to fulfill a contract that is established by the parent. By not handling exceptions in the child, you are breaking with the contract. In this scenario, maybe an alternative method ought to be used, or a thrown explicitly handled?

I know SwiftLint can handle conditional options on a rule in the YAML file, so one can be added to this rule to disable overridden functions in an opt-in manner.

ArtSabintsev:
Just pushed a build that should work. Going through CI now.

See attached screenshot.
regex-solution

ArtSabintsev:
Tests pass - woohoo!

Also, I have no idea how UnhandledThrow breaks https://github.com/Quick/Nimble/blob/16a9a31918596b03ad88e6ec9a5cd7f90f0e4d93/Tests/NimbleTests/Matchers/AllPassTest.swift#L6:12. Can't replicate it on my end.

jpsim:
This could definitely be improved to avoid using regular expressions.

For example, the Nimble false positive is caused by the `allTests` extension later in the file: https://github.com/Quick/Nimble/blob/514358cadbe7012a3e870eb700ec626a486fae52/Tests/NimbleTests/Matchers/AllPassTest.swift#L6:12

```swift
final class AllPassTest: XCTestCase, XCTestCaseProvider {
    static var allTests: [(String, (AllPassTest) -> () throws -> Void)] {
...
```

So this rule isn't super robust at the moment. I don't think I'm comfortable merging this.

Refactoring to avoid regular expressions and only making sure `throws` is marked on the function and not a closure signature in the return type or parameter, and constraining the call to `try` or `throw` to within the function body would be preferable.

jpsim:
Transcript from my Slack convo with @ArtSabintsev:

> arthur [10:00 AM]
> I kind of reverse engineered other rules to put this one together, since I couldn’t find a helpful-enough guide to put it together
>
> [10:01]
> so what’s the alternative to using regexp?
>
> jpsim [10:01 AM]
> let me dig around to find the closest rule to what you're doing
>
> arthur [10:01 AM]
> thanks
>
> [10:01]
> i’ll get to it tonight (after work) (edited)
>
> jpsim [10:01 AM]
> looking at `ASTRule` rules would be a step in the right direction
>
> arthur [10:02 AM]
> ok thanks!
>
> jpsim [10:03 AM]
> `FunctionBodyLengthRule` is a good start, since you can recursively check functions, you can check if the function itself is marked as `throws`, and you have access to its whole body range: https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules/FunctionBodyLengthRule.swift
>
> arthur [10:04 AM]
> sweet ok - thanks!
>
> jpsim [10:05 AM]
> also `OverriddenSuperCallRule` does something very similar, where if the function name matches a list of known methods that should usually call `super`, it makes sure the body invokes it: https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules/OverriddenSuperCallRule.swift
>
> arthur [10:06 AM]
> great - I’ll frankenstein something together from those three rules, and then smooth it all out
>
> jpsim [10:06 AM]
> lol, thanks!
>
> arthur [10:06 AM]
> thank you :slightly_smiling_face:
>
> jpsim [10:06 AM]
> adding some examples that would've previously triggered false positives in the current implementation of the rule would be appreciated too!
>
> [10:07]
> for example: nested throwing functions, `throws` in function parameters, `throws` in return type, etc.
>
> arthur [10:07 AM]
> yup - I’ll do that as well
>
> jpsim [10:08 AM]
> Thanks!

marcelofabri:
@jpsim @ArtSabintsev what's the status of this PR?

ArtSabintsev:
I haven't had time to get back to it yet. I'd love for this rule to become a thing.

marcelofabri:
@ArtSabintsev thanks for the update! I'll keep it open for now.

ArtSabintsev:
It's been over a year since I made the first commit. Since then, I haven't really had time to implement this rule. 

@marcelofabri feel free to close this issue.

marcelofabri:
OK, closing it. @ArtSabintsev feel free to reopen/open another PR if needed.

PR #1329, opened 2017-02-21
Add more mark tests: missing colon, titlecase, lowercase

xinsight:
First, just want to say, thanks for `swiftlint` - it's awesome.

I added a few extra checks for the `mark` rule: check for a missing colon and check for titlecase and lowercase spelling. For example: 

```
// Mark: does not work
// mark: does not work
// MARK - does not work
```

I updated the description so that user is provided with a valid MARK: usage. I find it useful, but maybe there is a better way to communicate a resolution to the end-user.

Thanks!


jpsim:
Thanks for the PR!

However, this introduced a false positive: https://github.com/mozilla-mobile/firefox-ios/blob/f71760eef423a46ac3a65338a080401cb2cf6678/StorageTests/TestSQLiteHistory.swift#L670:60

Could you please fix that, add a regression test to make sure it doesn't happen again and ping me again when resolved?

Thanks!

xinsight:
@jpsim: I tweaked the regexp's and added some more tests. Let me know if you have any other comments. Thanks!

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
 
 ##### Enhancements
 
+* Add lowercase and missing colon checks to the `mark` rule.
```
marcelofabri: missing two trailing whitespaces

Source/SwiftLintFramework/Rules/MarkRule.swift
```diff
@@ -24,11 +24,14 @@ public struct MarkRule: CorrectableRule, ConfigurationProviderRule {
     public static let description = RuleDescription(
         identifier: "mark",
         name: "Mark",
-        description: "MARK comment should be in valid format.",
+        description: "MARK comment should be in valid format. e.g. // MARK: ok",
```
marcelofabri: I don't really like this change because there're other valid formats (e.g. with a `-`).

marcelofabri: but I'm fine if @jpsim thinks it's an improvement

jpsim: I think we could also include the `-` case:

```swift
"MARK comment should be in valid format. e.g. '// MARK: ...' or '// MARK: - ...'"
```

xinsight: Happy to add it the extended help. (I personally love how the `-` becomes a line in the Xcode navigator.)

xinsight:
@jpsim: i merged in master. Anything else that needs to be done?

CHANGELOG.md
```diff
@@ -31,6 +31,12 @@
...
```
marcelofabri: this is the wrong section now

marcelofabri:
Merged in #1456

Issue #1326, opened 2017-02-21
Make trailing_comma rule autocorrectable


wongzigii:
When comma is nested with Dictionary,  autocorrect command will not correct properly.

````swift
let inner = JSON([
     "some_field": "1" + "2",  // warnings
])



let inner = [
     "some_field": "1" + "2", // warnings
]
````

Expect:
````swift
let inner = [
     "some_field": "1" + "2"
]
````

sammy-SC:
@wongzigii Hello,

I believe `trailing_comma` rule doesn't support `autocorrect`.

marcelofabri:
I'll reopen this to track adding support to `autocorrect` to this rule.

sammy-SC:
@marcelofabri I would like to get on it if no one else is working on it.

marcelofabri:
I don't think anyone is working at it right now, so feel free to grab it! 💯 

Just a reminder: this rule supports two styles according to the configuration, so this should be respected when implementing autocorrect.

marcelofabri:
Fixed on #1335 🎉

PR #1335, opened 2017-03-01
Add autocorrect to `trailing_comma`

sammy-SC:
implements #1326 

cc @marcelofabri 

I will be happy to implement any changes you suggest.


marcelofabri:
@sammy-SC I haven't fully reviewed this, but you should add `corrections` to the `RuleDescription` initializer so this will be unit tested.

Source/SwiftLintFramework/Rules/TrailingCommaRule.swift
```diff
@@ -46,13 +50,21 @@ public struct TrailingCommaRule: ASTRule, ConfigurationProviderRule {
...
                          dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        if let (index, reason) = violationIndexAndReason(in: file, kind: kind, dictionary: dictionary) {
+            return violations(file: file, byteOffset: index, reason: reason.rawValue)
+        } else {
+            return []
+        }
+    }
 
+    private func violationIndexAndReason(in file: File, kind: SwiftExpressionKind,
+                                         dictionary: [String: SourceKitRepresentable]) -> CommaRuleViolation? {
```
mgrebenets: I'm new to this codebase, but isn't returning empty array semantically the same as returning `nil`?

I mean, when returning optional, it first will be unwrapped and then checked with `isEmpty` anyways, so why not return empty array... I guess that was the rationale behind original design of this API.

jpsim:
@sammy-SC thanks for the PR! Could you please add some `corrections` entries as @marcelofabri requested? Thanks!

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
 
 ##### Enhancements
 
+* Make `trailing_comma` rule autocorrectable.
```
jpsim: requires two trailing spaces as described in CONTRIBUTING.md: https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes

sammy-SC:
@jpsim i'm working on it, i plan to do it within next few days.

sammy-SC:
@jpsim I added `corrections` to `RuleDescription` however I encountered following problem I need a little help with.

For string `/* 👨‍👩‍👧‍👦👨‍👩‍👧‍👦👨‍👩‍👧‍👦 */\nlet example = [ 1,\n 2\n // 3,\n]` 
the original implementation of `TrailingCommaRule` returns that violation happens on index 103. 

String has 105 characters in UTF8 but only 49 in unicode. 
My implementation of `func correct(file: File) -> [Correction]` therefore crashes on 
```
fatal error: cannot increment beyond endIndex
```

Also I found that in `TrailingSemicolonRule` the returned violation index is unicode aware.

Is it possible that `TrailingCommaRule` doesn't deal with unicode characters correctly?

If violation index 103 is correct, how should I go about removing it from unicode string?

Thank you

marcelofabri:
@sammy-SC I think the reason is that the violations are reported in byte offsets, and you need character offsets. There're methods to convert from one to another, you can take a look at `RedundantVoidReturnRule` for example. 

I don't think the current (`master`) implementation of this rule is wrong regarding offsets, because we also test violations with these multi-byte emojis.

BTW, we also needs tests for the corrections when using `mandatoryComma: true`. You can dig into `NumberSeparatorRuleTests` for inspiration on how to do this.

SDGGiesbrecht:
Byte offsets are the same as UTF‐16 units.

Use `String.UTF16View`:

```swift
// Information you started with:

var string = "👨‍👩‍👧‍👦👨‍👩‍👧‍👦, 👨‍👩‍👧‍👦"
let utf16Index = string.utf16.index(of: 0x2C /* Comma */)!

// To remove just the comma:
string.unicodeScalars.remove(at: utf16Index.samePosition(in: string.unicodeScalars)!)
/* (The forced unwrap will always succeed for the comma.
    It can only fail outside the Basic Multilingual Plane.) */

// To remove the comma and any combining characters associated with it:
string.characters.remove(at: utf16Index.samePosition(in: string)!)
/* (This forced unwrap will also always succeed for the comma.
    It can only fail outside ASCII.) */
```

CHANGELOG.md
sammy-SC: I've fix this.

sammy-SC:
@SDGGiesbrecht @marcelofabri thanks for the advice.

@marcelofabri @jpsim I've added changes you requested.

Source/SwiftLintFramework/Rules/TrailingCommaRule.swift
```diff
@@ -121,4 +141,51 @@ public struct TrailingCommaRule: ASTRule, ConfigurationProviderRule {
...
+        if matches.isEmpty { return [] }
+
+        var correctedContents = file.contents
+        let description = type(of: self).description
+
+        matches.reversed().forEach { offset in
+            let index = correctedContents.utf8.index(correctedContents.utf8.startIndex, offsetBy: offset)
+            let correctedIndex = index.samePosition(in: correctedContents)!
+            if configuration.mandatoryComma {
+                correctedContents.characters.insert(Character(","), at: correctedIndex)
```
marcelofabri: I think you can just use `","` instead of `Character(",")` here

sammy-SC: @marcelofabri thanks, I replaced it.

sammy-SC:
Hello @SDGGiesbrecht  

I implemented the change.

I still needed your advice about `String.UTF16View` 
because `ruleEnabled(violatingRanges: [NSRange], for rule: Rule)` works with `characterOffset`.

If there are any further changes you would like me to make, let me know.

Thank you

SDGGiesbrecht:
@sammy-SC, what you have does the right thing, and you changed what I mentioned perfectly.

When I look closer though, it still does a lot of work it doesn’t need to. And that is my fault for explaining what was going on instead of just answering the question the first time.

I think it will be easiest if I submit a pull request to your fork, that way you can see what I mean in the diffs more clearly than with an explanation.

Thank you for your patience.

***

Also, the world in general kind of owes you an apology for making the naming of string elements/indices more complicated than it ought to be:

| Unicode | Swift (String) | Cocoa (NSString) | SourceKitten/SwiftLint |
| --- | --- | --- | --- |
| “Character”/“Scalar” | “Unicode Scalar” | ∅ | ∅ |
| “Extended Grapheme Cluster” | “Character” | ∅ | ∅ |
| “UTF‐16” Unit | “UTF‐16” Unit | “Character” | “Character” |
| “UTF‐8” Unit | “UTF‐8” Unit | ∅ | “Byte” |

(And [Swift 4](https://github.com/apple/swift/blob/master/docs/StringManifesto.md) will change things yet again...)

SDGGiesbrecht:
@jpsim and @marcelofabri, when @sammy-SC pushes the changes I suggested, you will see a minor change to the test infrastructure. The failure reporting was confusing UTF‐16 units (SourceKitten “characters” with grapheme clusters (Swift “characters”), and was unsafe and inaccurate when emoticons or anything outside the Basic Multilingual Plane occurred in the string.

That bug was probably the cause of all of the confusion in the first place.

Also @jpsim and @marcelofabri, Samuel has addressed the concerns in both of your existing reviews.

SDGGiesbrecht:
@sammy-SC, there are merge conflicts (probably just in the change log). Merge `master` (freshly `pull`ed) into `trailing-comma-autocorrect`, and sort out the conflicts. Let me know if you need help.

sammy-SC:
@SDGGiesbrecht sorry, I didn't notice there was a conflict. 
It is resolved now. 
Thank you very much for the help.

SDGGiesbrecht:
@sammy-SC, Travis‐CI failed on Linux. Since the offending code is mine, I am willing to fix it. Would you give me write access to your fork to make it simpler for me?

sammy-SC:
@SDGGiesbrecht you should have write access now. I did try to fix linux build however without any luck, I will be able to look into it later today.

SDGGiesbrecht:
@sammy-SC, your fix ought to have worked, (although it ought also to have been unnecessary), but for Swift on Linux being a little behind.

My machines are all in the middle of updating, so I made the change in a text editor. We’ll see what CI says.

SDGGiesbrecht:
Grrr...

SDGGiesbrecht:
I’ll come back to this when my updates are finished and I can build locally again.

SDGGiesbrecht:
Sorry for the Git log mayhem. I was having trouble reproducing the failure locally.

@sammy-SC, [these are the cumulative diffs](https://github.com/realm/SwiftLint/pull/1335/files/5c1336c81bf86d8b1e64f94039f88d70fae624d0..d12b15f93d7b22fde2335b93291513b87012e19c) since your last commit (the Linux workaround), in case you are interested.

@jpsim, @marcelofabri, this pull request is now ready.

CHANGELOG.md
```diff
@@ -31,6 +31,14 @@
...
```
marcelofabri: this is on the wrong section now

Source/SwiftLintFramework/Rules/TrailingCommaRule.swift
```diff
@@ -9,14 +9,45 @@
...
+            ]
+        #if !os(Linux)
+            // disabled on Linux because of https://bugs.swift.org/browse/SR-3448 and
+            // https://bugs.swift.org/browse/SR-3449
+            result.append("let foo = [\"אבג\", \"αβγ\", \"🇺🇸\"↓,]\n")
+        #endif
+        return result
+    }()
+    private static let corrections: [String: String] = {
+        let fixed = triggeringExamples.map { $0.replacingOccurrences(of: "↓,", with: "") }
```
marcelofabri: I'm not sure this makes sense. Isn't it just saying that the correction is the same as the triggering example without the marker?

Tests/SwiftLintFrameworkTests/TestHelpers.swift
```diff
@@ -58,10 +58,9 @@ private func render(locations: [Location], in contents: String) -> String {
     var contents = contents.bridge().lines().map { $0.content }
     for location in locations.sorted(by: > ) {
         guard let line = location.line, let character = location.character else { continue }
-        var content = contents[line - 1]
-        let index = content.index(content.startIndex, offsetBy: character - 1)
-        content.insert("↓", at: index)
-        contents[line - 1] = content
+        let content = NSMutableString(string: contents[line - 1])
```
marcelofabri: can you explain the reasoning for this change?

CHANGELOG.md
SDGGiesbrecht: @sammy-SC, @marcelofabri is right.

Source/SwiftLintFramework/Rules/TrailingCommaRule.swift
SDGGiesbrecht: The correction for an excessive comma is to remove that comma. So basically yes.

Creating the correction tests algorithmically instead of writing them out individually allowed me to simply to channel it through workaround for the Linux bug, without having to write the same workaround in several places.

marcelofabri: oh, I didn't realize there was a comma after the marker, sorry!

Tests/SwiftLintFrameworkTests/TestHelpers.swift
SDGGiesbrecht: @marcelofabri, `Location.character` has always been an `Int` that describes the distance into a string in **`UTF‐16` units**. It was probably named `character` based off of `NSString`.

Before my change, this method measured off that distance into the string in **extended grapheme clusters** instead, which Swift calls `Character`, and tried to insert `↓` at that location to mark the location of the failure.

While both measurements have been given the name “character”, they are not the same thing.

For example, “ơ̧̈̂” is only `1` grapheme cluster but `5` UTF‐16 units (o +  ̛ +  ̧ +  ̈ +  ̂).
Also, “👨‍👩‍👧‍👦123” is only `4` grapheme clusters but `5` UTF‐16 units (a pair of surrogates, 1, 2, 3).

(The following example offsets are zero‐based.)

Using mis‐matched offsets can always result in an unexpected position, but whenever an offset points to a UTF‐16 index near the end, there is also the risk of an index out of bounds error. For example, `4` is the UTF‐16 offset of the  ̂ above, but the same offset of `4`, measured in grapheme clusters, is way out of bounds, so the tests crash.
Using the emoji example above, the UTF‐16 offset of `3` points at “2”, but requesting the `3`rd grapheme cluster results in “3” (wrong but no crash). The UTF‐16 offset of `4` points at “3”, but requesting the `4`th cluster grasps at out of bounds (crash).

My fix simply uses the `NSString`‐based, UTF‐16, `Int` offset on—`NSString`. No more mis‐match. I could have used `String.UTF16View`, but that would have incurred unnecessary parsing up and down the string to convert the `Int` to a `String.UTF16Index`.

When @sammy-SC first asked about the out of bounds errors, I assumed he just needed an explanation of Unicode. Later I looked closer and realized the errors had nothing to do with him. It was merely that with the comma nearly always being near the end of the string, this rule is particularly prone to this type of bug, and Samuel just happened the first person the bug messed with. At that point I felt bad and stepped in to untangle it for him. Aside from these string indexing bug fixes, the pull request is all his.

Is that explanation enough?

SDGGiesbrecht:
> Also the submodules probably shouldn't be updated in this PR.

I agree.

@sammy-SC, do you know how to remove those changes?

sammy-SC:
@marcelofabri I removed submodule update from this PR and fixed changelog. 

@SDGGiesbrecht I added you as author to the changelog. Thank you for the help.

SDGGiesbrecht:
The CI test failure with the Swift Package Manager is in the YAML parser tests. I don’t think it has anything to do with this pull request, as none of the changes here even touch the YAML processing.

[Here](https://travis-ci.org/realm/SwiftLint/jobs/220320406#L763) is the line in the log where the failure was reported.

The test failure is in [`testParseInvalidStringThrows`](https://github.com/realm/SwiftLint/blob/d1cebe425fdef6e031d25137759da6e24855d455/Tests/SwiftLintFrameworkTests/YamlParserTests.swift#L25).

The offending function in SwiftLint is [`YamlParser.parse(_:)`](https://github.com/realm/SwiftLint/blob/4008fa38f1eece032600eb943ca73fc48f4cf5cb/Source/SwiftLintFramework/Models/YamlParser.swift#L32).

Other notes:

The Linux tests only succeed because they skip that particular test. It would be [here](https://travis-ci.org/realm/SwiftLint/jobs/220320407#L845) in the log. [Here](https://github.com/realm/SwiftLint/blob/d1cebe425fdef6e031d25137759da6e24855d455/Tests/SwiftLintFrameworkTests/YamlParserTests.swift#L43) is were it has been intentionally skipped in the tests source. There is a note there in the source about a bug. @norio-nomura, you were both the one who reported the bug and the one who wrote the recent changes to the YAML parsing. Do you think this failure might be somehow related to that bug?

On the other hand, the Xcode build passed that test. ??? See the log [here](https://travis-ci.org/realm/SwiftLint/jobs/220320405#L1070).

Any other ideas, Norio?

norio-nomura:
> [Here](https://travis-ci.org/realm/SwiftLint/jobs/220320406#L763) is the line in the log where the failure was reported.

That is caused by strict version description in `Package.swift` of SourceKitten, and that has been fixed on https://github.com/jpsim/SourceKitten/pull/356.
This error will happen until next version of SourceKitten will be released.

jpsim:
I just cut SourceKitten "0.17.2: 😺"

norio-nomura:
@jpsim Thanks!
I restarted the job. Let's wait for the result. 🙂

SDGGiesbrecht:
Thanks @norio-nomura and @jpsim.

PR #1338, opened 2017-03-01
Fix false positive and wrong correction on `unused_closure_parameter` rule.

sora0077:
if parameter named `` `self` `` used in closure, `swiftlint autocorrect` removes parameter allways.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -127,6 +127,8 @@
 * Fix `todo` rule messages when the comment is not on a new line.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1304](https://github.com/realm/SwiftLint/issues/1304)
+* Fix false positive and wrong correction on `unused_closure_parameter` rule.  
+  [Hayashi Tatsuya](https://github.com/sora0077)
 
 * Fix false negative on `unused_closure_parameter` rule.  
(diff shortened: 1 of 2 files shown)


marcelofabri:
@sora0077 it seems that this actually fixed some cases where the rule wasn't catching all violations. Is that a proposital fix? 😅 

This also needs to be rebased before merging.

marcelofabri:
And actually, I'm not 100% sure we should make `self` a special case since it only works because of a bug in the Swift compiler (I remember someone from Apple saying this on a mailing list, but I doubt I'll find the link)

sora0077:
@marcelofabri Thank you for your review.
> And actually I'm not 100% sure we should make self a special case...

Hmm, I agree with your opinion.
What do you think about adding configuration like below.

```yml
unused_closure_parameter:
  excluded:
    - self
    - hoge
```

> it seems that this actually fixed some cases where the rule wasn't catching all violations. 

I'll send another PR about this.

marcelofabri:
Actually, rethink this: maybe we can try to identify an escaped keyword instead of adding a special case for `self`.

Also, I think that what I said about this being a bug in Swift is wrong, I got confused.

sora0077:
@marcelofabri  
Thank you for your review.
I'll close this PR 😄 .

PR #1342, opened 2017-03-04
Add README_KR.md

red282:
SwiftLint is amazing!
I've translated README file in Korean because I hope that SwiftLint get known well to more people in Korea :)


jpsim:
Thank you so much for doing this, @red282! What a wonderful contribution 😄 👍 💯 .

Issue #640, opened 2016-04-26
Enforce spaces in braces


ZevEisenberg:
My team is using a particular curly brace style: extra newlines padding curly braces in classes, structs, enums, and extensions. We do _not_ apply it to functions/conditionals/etc:

``` swift
enum State {

    case Empty
    case Loading
    case Loaded

}

class ViewModel {

    var state: State

    func wazzup() {
        if state == .Empty {
            print("empty")
        }
        else {
            print("not empty")
        }
    }

}
```

My gut feeling is that this is not enforceable with custom rules as they currently exist. I also don't know how broadly something like this would be desired (you could also enforce _not_ having the extra space, if you preferred the opposite style to my team). Is this something that would make sense as part of SwiftLint?

jpsim:
This would definitely make sense as part of SwiftLint. I'd picture a configurable rule enforcing the number of padding newlines for types, with a default value of `1`.

Dschee:
I know the discussion about this rule is quite old but I still agree that an opt-in rule (or two?), where spaces around braces could be configured would be **a win for many projects**. We are actually using a very simple rule which I described in #1518 using a custom_rule & that works fine for us. But I think such an option **should be included into SwiftLint** even if it is opt-in.

Although the rationale for our own style is simple ("if there's indentation, why bother and add a new line – it's clean and readable enough thanks to indentation already, prevent scrolling then!") I can understand that others have **different styles** that may even **vary between type declarations and functions/conditionals**. Note that for the type declarations part there has already been #1345 which was only closed due to inactivity some time ago.

So here's my **suggestion** (which I would also volunteer to implement):

- Add a new rule named `**spaces_around_braces**`
- Add config options for **min/max outer and inner spaces** counts
- Possibility to optionally & separately apply to functions and conditionals
- Allow the config options for spaces counts to specify "any"

Having the rule be this flexible, I think that we can even come up with a **default configuration that is common ground** within the Swift community. The default configuration could for example look like this:

```yaml
- spaces_around_braces
  - opening_brace_padding:
    - outer
      - min: 1
      - max: nil # any count
    - inner
      - min: nil # any count
      - max: 0
  - closing_brace_padding:
    - inner
      - min: nil # any count
      - max: 0
    - outer
      - min: 1
      - max: nil # any count
  - include_type_declarations: true
  - include_function_declarations: true
  - include_other_code: false
```

[...]

PR #1345, opened 2017-03-08
Add spaced_type_declaration rule

dernster:
Added `spaced_type_declaration` rule that warns if either the start or end of a type declaration is not enclosed by two empty lines. 

For example:
```swift
protocol Protocol {
    func somethingAmazing()
}
class Example {
    let value = ""
}
```
should be written as follows:
```swift
protocol Protocol {

    func somethingAmazing()

}

class Example {

    let value = ""

}
```

I have done two commits: the first one is the actual implementation of this rule and the second one is the result of autocorrecting the entire project so that all tests run successfully. Please let me know if there is something I missed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,11 @@
   [Natan Rolnik](https://github.com/NatanRolnik)
   [#1270](https://github.com/realm/SwiftLint/issues/1270)
 
+* Add `spaced_type_declarations` rule that warns if either the start or end
+  of a type declaration is not enclosed by two empty lines.  
+  [Diego Ernst](https://github.com/dernster)
+  [#1345](https://github.com/realm/SwiftLint/pull/1345)
+
 ##### Bug Fixes
 
 * Fix a false positive on `large_tuple` rule when using closures.  
--- a/Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
@@ -10,12 +10,15 @@ import Dispatch
 import Foundation
 
 extension Array where Element: NSTextCheckingResult {
+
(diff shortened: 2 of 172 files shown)


kimdv:
I like this !

marcelofabri:
I'm not sure we should apply this to the project itself. Also, this really should be an opt-in rule.

Can you take a look into https://github.com/realm/SwiftLint/issues/640? This will fix that issue, but there're a few ideas there that might be useful to implement, such as making the rule configurable.

jpsim:
Thanks for the PR, and I definitely agree with @marcelofabri that this rule should be configurable, and that SwiftLint itself would use slightly different settings.

dernster:
Thanks for your comments. I'm thinking of a configuration to specify the number of inner padding lines and the number of external padding lines within a type declaration. They should always be applied symmetrically to the declaration, like this:

dernster:
@marcelofabri @jpsim Thanks for your comments. I'm thinking of a configuration to specify the number of inner padding lines and the number of outer padding lines within a type declaration. They should always be applied symmetrically to the declaration, like this:

```
+ outer_padding_lines
class Example {
+ inner_padding_lines
   var foo: Int
   var bar: String
   func function() {
      ....
   }
+ inner_padding_lines
}
+ outer_padding_lines
```

What do you think? Is it ok if I force a git push to this branch?

marcelofabri:
I think we be really flexible here, since there's not really a convention. I'd propose the configuration be be like:

```yml
outer_padding:
  beginning: 1
  end: 1
inner_padding:
  beginning: 1
  end: 1
```

We could also support something like `outer_padding: 1` to say it's symmetrical. What do you think?

And it's ok to force push this since usually no one works on other's forks.

jpsim:
Closing due to lack of activity.

Issue #517, opened 2016-02-09
Add example .swiftlint.yml files


scottrhoyt:
I've toyed around with the idea of having an `ExampleConfigurations` directory to facilitate configuration file documentation and potentially circulating common configurations. It might even be an imperfect, but lightweight, way of collecting feedback on what default values should be and which custom rules should be considered for including as standard `Rule`s. The more I think about it, it probably makes more sense as a separate repo. I wonder though, is there yet enough variation in community configurations and interest to support this?

jpsim:
We could ask a few popular & actively maintained projects that use SwiftLint to share their workflow. Realm, Lyft & Breather come to mind, I'm sure there are many more.

norio-nomura:
I want `.swiftlint.yml` for benchmarking that runs all rule.

scottrhoyt:
I can probably make that happen, @norio-nomura. I have been playing around with adding serialization to YamlSwift. I have all the cases covered for our needs, however YAML allows some features that make serializing it in the general case a bit hard than I expected at first, but will most likely have it done soon. With that, we can pretty easily do a `swiftlint init --all_rules` that code generate a `.swiftlint.yml` file whitelisting all the `Rule`s.

norio-nomura:
> With that, we can pretty easily do a swiftlint init --all_rules that will generate a .swiftlint.yml file whitelisting all the Rules.

Nice! :+1:

PR #1346, opened 2017-03-08
add experimental rules for Firebase SDK compatibility

iulukaya:


marcelofabri:
This definitely should be an opt-in rule. Also, the headers are not in the project's format (unit tests will fail)

iulukaya:
Thanks Marcelo, I'll definitely make them opt-in rule, and fix the headers

iulukaya:
Hi Marcelo, I realized unit tests forces every file to have Copyright @ xxxx Realm.
As I want to add these rules on behalf of Firebase(Google) that would make it complicated.
Can we have a AUTHORS file that add Google Inc. and Realm
at the root of the repository.

marcelofabri:
I'll leave it for @jpsim

jpsim:
I appreciate the contributions @iulukaya, but I'd like to see these rules made a bit more robust before shipping them with SwiftLint. For example, it seems like it'd be very easy to have false positives for most of these rules right now, because you're just excluding comments and strings rather than checking the specific kinds of method calls, etc.

It also seems as though most of these rules could be written as custom rules in a `.swiftlint.yml` configuration. Perhaps you'd be interested in contributing the first SwiftLint configuration as described in #517? Such as `examples/swiftlint_firebase.yml`.

iulukaya:
Hi @jpsim , I agree with both statements. Looks like I need to check for a certain "method call" inside a certain "function declaration". I was trying to do it in custom yaml rules, but couldn't find enough docs. Not sure how to match or ignore "declaration of method type", it only allows me to check syntax kind. Or how to "exclude".
Maybe if you can point me to available comments inside custom rules (like a reference guide) or at least the file you read (so I can find all available options), or show me one of the rules as an example, I can go from there.
Cheers,

iulukaya:
@jpsim quick ping on this. (also not sure why that particular test is failing on jazzy. It works fine locally.)

iulukaya:
@marcelofabri wdyt?

PR #1351, opened 2017-03-11
Fix false negative on `unused_closure_parameter` rule.

sora0077:


marcelofabri:
@sora0077 Thanks so much for this! Can you add a test that would fail without this change so we don't regress in the future? 💯

sora0077:
@marcelofabri Thank you for your review!
I added test cases.

marcelofabri:
Thanks again 🚀

PR #1355, opened 2017-03-14
Added opt-in IBOutlet Setter Rule

jeffreybergier:
Hi Swiftlint team,

## Problem
I recently came across issues in an iOS project, where IBOutlets were being reset without going through the whole proper process of removing the old object from the view hierarchy and adding the new one into the view hierarchy. Swift doesn't have a native way to prevent outlets from being reset.

I thought this would be a good idea for an Opt-In rule for SwiftLint. Let me know what you think and if you think this rule is a good idea for SwiftLint.

## Solution
This rule checks if the file contains IBOutlets, if it does it looks for any lines of code that attempt to call the setter on that outlet. e.g. self.myLabel = UILabel() would trigger this rule.

While the rule is not perfect, it will help avoid an Interface Builder anti-pattern. In general, if your code is resetting IBOutlets, there is probably a better way to accomplish what you want. 

I added many triggering and non-triggering test cases. They all have a variety of spacing, tabbing and other issues inserted into them to make sure the regular expressions work correctly.

## Pull Request Info
I'm a first time contributor to the Swiftlint project. I apologize in advance if I am doing something incorrect. I would love to be corrected :). I checked the contributing guidelines. This project passes all Xcode tests and `swift test`. Also, script/cibuild appeared to work fine. 

However, when I ran `make docker_test` I received the following error. I've never used docker before, so I'm not quite sure how to resolve the issue. I'm also not sure if thats a deal-breaker for the PR. If it is, let me know.:

```
bash-3.2$ make docker_test
docker run -v `pwd`:/SwiftLint norionomura/sourcekit:302 bash -c "cd /SwiftLint && swift test"
docker: Error parsing reference: "Github/SwiftLint:/SwiftLint" is not a valid repository/tag: invalid reference format.
See 'docker run --help'.
make: *** [docker_test] Error 125
```

## Known Issues
- This rule can only protect against setting IBOutlets within the same file. This rule should be combined with the 'private_outlet' rule in order to make it effective for the whole project.
- 3 TODOs that are commented out in the triggering/non-triggering test cases array. Because this is opt-in and the cases where the rule is triggered are fairly rare and can often be fixed by renaming a local variable, I left them as TODOs for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@
 
 ##### Enhancements
 
+* Add `outlet_setter` opt-in rule
+  that warns when IBOutlet properties are reset in code,
+  rather than set automaticaly by IB.
+  [Jeffrey Bergier](https://github.com/jeffreybergier)
+
 * Performance improvements to `generic_type_name`,
   `redundant_nil_coalescing`, `mark`, `first_where` and
   `vertical_whitespace` rules.  
--- a/Source/SwiftLintFramework/Models/MasterRuleList.swift
+++ b/Source/SwiftLintFramework/Models/MasterRuleList.swift
@@ -122,6 +122,7 @@ public let masterRuleList = RuleList(rules:
     OpeningBraceRule.self,
     OperatorFunctionWhitespaceRule.self,
     OperatorUsageWhitespaceRule.self,
+    OutletSetterRule.self,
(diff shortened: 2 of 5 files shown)


jpsim:
Sorry for the delay. Continued in #1555.

PR #1339, opened 2017-03-02
ExplicitTypeInterfaceRule rule

kimdv:
Hello! 

I don't want to start a war here, so I made this an opt in as default!

To explain myself briefly; I added this rule because I think it makes the code more readable and helps keep an overview of the specific properties. 

As stated before - the rule is made an OPT IN, so people themselves can decide whether to follow the rule or not. 

I have yet to figure out how the test works with the `nonTriggeringExamples`, `triggeringExamples` and `↓`. 

I hope that there is a nice person, who will help me out here, so we can finish the rule and give the people who like it, an opportunity to start using it. :)

Remember to keep it civil. 👍

PR #1356, opened 2017-03-14
ExplicitTypeInterfaceRule rule

marcelofabri:
Continued from #1339


marcelofabri:
Thanks @kimdv for implementing this! 💯

kimdv:
Thanks for helping me out! 😃

jpsim:
I should've paid attention to this a bit more, as I think the name "explicit type interface" doesn't accurately reflect what this does. This is rather an "explicit type annotation". According to Wikipedia, the all-knowing source of absolute truth:

> In computing, an interface is a shared boundary across which two separate components of a computer system exchange information.

Specifying the type explicitly rather than inferring it doesn't change the interface.

Issue #1322, opened 2017-02-20
Request: add --lenient lint parameter to only generate warnings


NachoSoto:
Just like `--strict`, it would be nice to be able to *only* generate warnings.

A use case for me is that I want to run the linter for all files when I run all the tests, but I don't want it to fail, because I want to distinguish a linting (ran separately) problem versus failed tests.

jpsim:
I see how that could be useful. Until we add such an option, you could hack it yourself:

```
swiftlint | sed s/error:/warning:/g
```

jpsim:
added in #1364.

NachoSoto:
Awesome! :tada:

PR #1359, opened 2017-03-15
add --lenient

aaroncrespo:


aaroncrespo:
Don't worry about un-verified I did this on a computer missing my gpg. key.

aaroncrespo:
re: #1322

aaroncrespo:
CI Is failing because type checking something took 190ms over the 150ms max.

aaroncrespo:
Not sure what to do next the failures seem unrelated to the changes in this PR.

marcelofabri:
I think the issue was more about using warnings (and never errors) than the return code.

aaroncrespo:
I've updated the PR:
* convert any error violations to warning violations. I couldn't see an approach that was much different than what I did outside of parameterizing the option to all sorts of types.
* then if lenient is set, I suppress any warning threshold error

marcelofabri:
Integrations tests are failing, can you please take a look at this?

aaroncrespo:
going to resubmit I think a squash failed.

PR #1362, opened 2017-03-17
Add `implicitly_unwrapped_optional` rule

SiarheiFedartsou:
Add `implicitly_unwrapped_optional` rule that warns when using implicitly unwrapped optional,
except cases when this IUO is IBOutlet. 
Also some small fixes to conform to new rule(NSRegularExpression! substituted with NSRegularExpression?).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,10 @@
   [Hayashi Tatsuya](https://github.com/sora0077)
   [#1318](https://github.com/realm/SwiftLint/issues/1318)
 
+* Add `explicit_type_interface` opt-in rule that validates that the properties
+  have an explicit type interface.  
+  [Kim de Vos](https://github.com/kimdv)
+
 ##### Bug Fixes
 
 * Fix crashes when accessing cached regular expressions when linting in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 
 * Add `implicitly_unwrapped_optional` rule
   that warns when using implicitly unwrapped optional,
-  except cases when this IUO is IBOutlet.
+  except cases when this IUO is IBOutlet.  
   [Siarhei Fedartsou](https://github.com/SiarheiFedartsou/)
 
 * Performance improvements to `generic_type_name`,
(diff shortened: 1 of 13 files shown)


SiarheiFedartsou:
It seems that tests failed not because of my changes...

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@
 
 ##### Enhancements
 
+* Add `implicitly_unwrapped_optional` rule
+  that warns when using implicitly unwrapped optional,
+  except cases when this IUO is IBOutlet.
```

Source/SwiftLintFramework/Rules/ImplicitlyUnwrappedOptionalRule.swift
```diff
@@ -0,0 +1,64 @@
...
+    public func validate(file: File, kind: SwiftDeclarationKind,
+                         dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard SwiftDeclarationKind.variableKinds().contains(kind) else {
+            return []
+        }
+
+        guard let typeName = dictionary.typeName  else { return [] }
+        guard typeName.hasSuffix("!") else { return [] }
+
+        let isOutlet = dictionary.enclosedSwiftAttributes.contains("source.decl.attribute.iboutlet")
```
marcelofabri: what do you think about making this configurable? I think by default we should skip outlets, but I know some people who think that IUO shouldn't be used even on outlets.

SiarheiFedartsou:
@marcelofabri Thanks for your feedback! Just updated pull request.

SiarheiFedartsou:
@marcelofabri Any updates?

SiarheiFedartsou:
@marcelofabri  just updated the PR.

marcelofabri:
Thanks for implementing this! 👏

jpsim:
Thanks for the PR @SiarheiFedartsou!

---

> Can you also revert the changes that were made in the project to make it conform to the new rule?
> 
> I don't think we want to enforce it here (at least for now), as there're only a few places that use IUO and I'm not sure the proposed approaches are really better.

👍  thanks for making this point @marcelofabri!

PR #1363, opened 2017-03-18
BodyLengthConfiguration for length rules allowing exclusions

drodriguez:
BodyLengthConfiguration is a new configuration for the length rules (file length, type body length and function body length).

Besides the already existing warning and error configuration options, it allows a third one which applies exclusions to the specific rule. That way, one can keep the rule enabled and with sane defaults, but avoid the rule triggering in a common set of file, types or functions. The main example will be XCTestCase subtypes which might go for more than 400 lines and need disabling comments in each of the files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,11 @@
   have an explicit type interface.  
   [Kim de Vos](https://github.com/kimdv)
 
+* `file_length`, `type_body_length` and `function_body_length` can be configured
+  to exclude certain file names, type names and function names, respectively.
+  [Daniel Rodríguez Troitiño](https://github.com/drodriguez)
+  [#1363](https://github.com/realm/SwiftLint/issues/1363)
+
 * Add `fatal_error_message` opt-in rule that validates that `fatalError()` calls
   have a message.  
   [Kim de Vos](https://github.com/kimdv)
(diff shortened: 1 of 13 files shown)


jpsim:
Hi @drodriguez, sorry for the long delay before acknowledging this, I'm paying down GitHub notification debt and caught this just now.

I like the idea of more flexible+powerful configurations regarding where rules should and shouldn't be applied, but I wonder if we wouldn't be better off with a more general solution that could apply to all rules? For example, a `disabled_in_scope` keyword in `.swiftlint.yml`:

```yaml
line_length:
  disabled_in_scope:
    - "static func create" # allow long curried functions, as found in `Source/swiftlint/Commands/*`
type_body_length:
  disabled_in_scope:
    - "subclass XCTestCase" # allow lots of tests in a single XCTestCase subclass
```

I'm not fully happy with this proposal, but maybe it can spur a conversation about how to best handle this.

Thanks for the PR!

drodriguez:
My initial idea was bigger, but because it was my first approach to SwiftLint code and I wanted this done, I cut the scope a lot.

My initial idea was something similar to what you propose, but a little less stringly typed:

```
line_length:
  disabled:
    - type_pattern: FooBar
    - type_pattern: BarBaz
       only_subtypes: true # false by default
    - func_pattern: myMethod
    - file_name: MyFileNameRegex
```

Each of the items of `disabled` applied to either a file pattern, a type pattern or a function pattern. I was also considering a `enabled` to do a white list approach.

When I started looking at the code, I figured out that in the time I allowed myself, it was going to be impossible, so I decided to be pragmatic, and create the rule specializations.

I agree with you that a more general solution, which applies to every rule, would be the ideal solution, though.

PD: I can still rebase and fix the conflicts in this PR if you want, but I don’t think I would have time to do the more general system any time soon (it is a big change).

Issue #1348, opened 2017-03-10
Rule request: Do not allow fatalErrors without messages


Noobish1:
Had a couple of these in our codebase that we didn't like. fatalErrors should always include an informative message.

jpsim:
This is a great first rule to write, if anyone's interested.

kimdv:
It sounds like a great rule!

marcelofabri:
Implemented on #1365 ✅

PR #1365, opened 2017-03-18
Fatal error message rule

kimdv:
Started on a rule request #1348 

I don't know id this is the most efficient way to do, but I started and now we can continue the work from here 😸

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -20,6 +20,7 @@ opt_in_rules:
   - object_literal
   - number_separator
   - prohibited_super_call
+  - fatal_error_message
 
 file_header:
   required_pattern: |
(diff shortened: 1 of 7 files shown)


marcelofabri:
@kimdv Thanks for contributing! I've made a few changes and made the rule enabled by default. 
I want to check oss-check, but we're having a few issues with it right now (#1369).

marcelofabri:
I've run it locally and it found 506 violations. I still think ideologically this should be a default rule, but maybe it shouldn't since there're so many violations. Thoughts, @jpsim?

kimdv:
@marcelofabri I partly agree. I think it's a good idea to have it enabled by default, because it is good behaviour to write why there is a fatalError. It makes it easier, specially in opensource projects to get a good overview of the code.

marcelofabri:
I thought more about this and I think it should be opt-in until we implement #1379.

PR #1368, opened 2017-03-19
Fix `==` operator in ColonConfiguration

SiarheiFedartsou:
I think such issues can be easily detected by linter. We can add "identical operands in comparison operator" rule. I've tried to add it, but it is not so simple as it seemed to be. As I understand for now SourceKit give access to code structure only(like classes, functions, ifs, whiles etc), but not to expressions. So there is no easy way to parse for example "return A == B" or "if (A == B)" to know what is A and what is B. 
Maybe someone have some ideas how we can write this rule?


marcelofabri:
@SiarheiFedartsou there's no easy way to do this, I guess. The best we can probably do is come up with something using regexes, as in `operator_usage_whitespace` rule. However, it's not 100% reliable. 

I've filled #1371 to track this rule request.

PR #1375, opened 2017-03-20
Fix checkstyle report format

YukiOya:
Hi SwiftLint team,

we use swiftlint on Jenkins using the checkstyle plugin, but since the error element of the generated report does not have a `source` attribute, it is difficult to search for warnings or errors by the rule.

This PR fixes this problem, and fixed the file element to have some errors.

Thank you.

### DIFF
--- a/Source/SwiftLintFramework/Reporters/CheckstyleReporter.swift
+++ b/Source/SwiftLintFramework/Reporters/CheckstyleReporter.swift
@@ -30,12 +30,15 @@ public struct CheckstyleReporter: Reporter {
         let col: Int = violation.location.character ?? 0
         let severity: String = violation.severity.rawValue
         let reason: String = violation.reason.escapedForXML()
+        let identifier: String = violation.ruleDescription.identifier
+        let source: String = identifier.isEmpty ? "" : "swiftlint.rules.\(identifier)".escapedForXML()
         return [
             "\n\t<file name=\"", file, "\">\n",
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Reporters/CheckstyleReporter.swift
```diff
@@ -19,24 +19,34 @@ public struct CheckstyleReporter: Reporter {
...
+    }
+
     private static func generateForSingleViolation(_ violation: StyleViolation) -> String {
-        let file: String = (violation.location.file ?? "<nopath>").escapedForXML()
         let line: Int = violation.location.line ?? 0
         let col: Int = violation.location.character ?? 0
         let severity: String = violation.severity.rawValue
         let reason: String = violation.reason.escapedForXML()
+        let identifier: String = violation.ruleDescription.identifier
+        let source: String = identifier.isEmpty ? "" : "swiftlint.rules.\(identifier)".escapedForXML()
```
marcelofabri: I think we can assume that the identifier is not empty here

YukiOya: Thank you for fast review.

fix it condition.  https://github.com/realm/SwiftLint/pull/1375/commits/f5905072709d94296f34f0580ad982c8aa468c73

marcelofabri:
Thanks! 🚀

PR #1393, opened 2017-03-25
Improve for_where rule

### LINKED ISSUE #1387: Improve for_where rule
If the `if` is too complex, it's better to have it instead of a `where` clause.

Also, that bug that prevents us to recognize local variables sometimes makes the rule trigger false positives.

marcelofabri:
Fixes #1387


jpsim:
Awesome 👍

PR #1394, opened 2017-03-25
apply —lenient to reports emitted while running.

aaroncrespo:
didn't know that some reporters run before all results are gathered. so we apply --lenient at the point of collection per file before `report()` of instead of after when they are aggregated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,7 +98,7 @@
   [Kim de Vos](https://github.com/kimdv)
 
 * Add `--lenient` CLI option to `lint` command. Facilitates running a lint task
-  that doesn't fail a pipeline of other tasks.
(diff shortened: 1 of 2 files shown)


aaroncrespo:
@marcelofabri This is unrelated to Xcode 8.3

NachoSoto:
Is this good to go? :)

CHANGELOG.md
```diff
@@ -112,6 +112,10 @@
 
 ##### Bug Fixes
 
+* Fix `--lenient` reports errors before lint completion.
```
marcelofabri: this is in the wrong section

marcelofabri:
@NachoSoto would you mind testing this for your use case before we merge?

jpsim:
Thanks! Merged to master in bc7fa329.

PR #1396, opened 2017-03-26
Avoid pattern matching false positives

### LINKED ISSUE #1376: False positives and false negatives with unused_optional_binding
```
// Should not trigger, but triggers
if case .some(_) = self {
}

// Should trigger, but doesn’t
if case .some(let _) = self {
}
```

Seems that any usage of `case` is ignored.

Reviewing the rule code, the regular expressions used are quite broad and don’t even check for `let` being or not there.

Something like the following almost get it right:

```
        let underlineOutsideParenthesis = "(?<=let\\s)_\\s*="
        let underlineInsideParenthesis = "(?<=let\\s)\\((\\s*[_,]\\s*)+\\)"
        let pattern = "(("  + underlineOutsideParenthesis + ")|(" + underlineInsideParenthesis + "))"
```

But it doesn’t catch the `case .some(let _)` case.

/cc @marcelofabri

drodriguez:
The rule for unused optional binding was false triggering for pattern matching where the associated value was just an underscore. Additionally, pattern matching mixed with unused optional binding was not triggering in some cases.

New non triggering and triggering examples have been added for the failing cases, and new regular expressions are used to catch all the cases.

Modify the code in `File+SwiftLint` to expose full `NSTextCheckingResult` (and not just ranges) and rewire some pieces to use the new functions.

Fixes #1376

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,9 @@
 
 ##### Bug Fixes
 
-* None.
+* Avoid pattern matching false positives in `unused_optional_binding` rule.  
+  [Daniel Rodríguez Troitiño](https://github.com/drodriguez)
+  [#1376](https://github.com/realm/SwiftLint/issues/1376)
 
 ## 0.17.0: Extra Rinse Cycle
 
--- a/Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
+++ b/Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
@@ -26,7 +26,8 @@ public struct UnusedOptionalBindingRule: ASTRule, ConfigurationProviderRule {
             "if let (_, asd, _) = getOptionalTuple(), let bar = Foo.optionalValue {\n" +
             "}\n",
             "if foo() { let _ = bar() }\n",
-            "if foo() { _ = bar() }\n"
+            "if foo() { _ = bar() }\n",
+            "if case .some(_) = self {}"
         ],
         triggeringExamples: [
             "if let ↓_ = Foo.optionalValue {\n" +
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -10,7 +10,9 @@
 
 ##### Bug Fixes
 
-* None.
+* Avoid pattern matching false positives in `unused_optional_binding` rule.
```

CHANGELOG.md
```diff
@@ -10,7 +10,9 @@
...
+  [Daniel Rodríguez Troitiño](https://github.com/drodriguez)
+  [#1396](https://github.com/realm/SwiftLint/1396)
```
marcelofabri: this should be the number of the issue, not PR

CHANGELOG.md
drodriguez: ✅

CHANGELOG.md
drodriguez: ✅

drodriguez:
I don’t see how to request Travis to retry. The failure seems to be unrelated to the changes to me.

Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
```diff
@@ -73,11 +76,14 @@ public struct UnusedOptionalBindingRule: ASTRule, ConfigurationProviderRule {
...
-        let underlineOutsideParenthesis = "(?<=[^(]\\s)_(?=\\s[^)])"
-        let underlineInsideParenthesis = "\\((\\s*[_,]\\s*)+\\)"
-        let pattern = underlineOutsideParenthesis + "|" + underlineInsideParenthesis
-        return file.match(pattern: pattern,
-                          excludingSyntaxKinds: kinds,
-                          range: range)
+        let letUnderscore = "let\\s+(_\\s*[=,)])"
+        let letUnderscoreTuple = "let\\s+(\\((\\s*[_,]\\s*)+\\)\\s*=)"
+
+        let matches = file.matchesAndSyntaxKinds(matching: letUnderscore, range: range)
```
marcelofabri: Can we combine the regex in one? It should be faster than calling `file.matchesAndSyntaxKinds` twice.

drodriguez: Make the regex more complicated to read, but I changed into using only one.

jpsim:
Merged in c29cf9a4d013c064d211e1b9f4474ece2821839b. Thanks!

PR #1402, opened 2017-03-28
Update optional interpolated strings for Xcode 8.3

keith:
To actually build this will require an update to SoruceKitten. We might want to tag a new release on that repo. Instead of just allowing these optional strings, we may want to provide default values, happy to change this however makes sense.

### DIFF
--- a/Source/SwiftLintFramework/Rules/RuleConfigurations/FileHeaderConfiguration.swift
+++ b/Source/SwiftLintFramework/Rules/RuleConfigurations/FileHeaderConfiguration.swift
@@ -38,9 +38,10 @@ public struct FileHeaderConfiguration: RuleConfiguration, Equatable {
     private static let defaultRegex = regex("\\bCopyright\\b", options: [.caseInsensitive])
 
     public var consoleDescription: String {
-        return severityConfiguration.consoleDescription + ", required_string: \(requiredString)" +
-            ", required_pattern: \(requiredPattern), forbidden_string: \(forbiddenString)" +
-            ", forbidden_pattern: \(forbiddenPattern)"
+        return severityConfiguration.consoleDescription + ", required_string: \(requiredString ?? "None"))" +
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Rules/RuleConfigurations/FileHeaderConfiguration.swift
```diff
@@ -38,9 +38,9 @@ public struct FileHeaderConfiguration: RuleConfiguration, Equatable {
     private static let defaultRegex = regex("\\bCopyright\\b", options: [.caseInsensitive])
 
     public var consoleDescription: String {
-        return severityConfiguration.consoleDescription + ", required_string: \(requiredString)" +
-            ", required_pattern: \(requiredPattern), forbidden_string: \(forbiddenString)" +
-            ", forbidden_pattern: \(forbiddenPattern)"
+        return severityConfiguration.consoleDescription + ", required_string: \(String(describing: requiredString))" +
```
marcelofabri: We actually already have a bug here as it prints `Optional(actual value)`. Can you please change these to have default values (and unwrap the optional if possible)?

keith: Since any of these are optional, should we just replace them with `"None"` in the case that we're missing them?

marcelofabri: I think either `"None"` or `"nil"` would be fine

keith: Updated

jpsim:
Merged in b410c9376a5631a7975c317cf23aaab845172f24 and ca7df3e7be9220578312d8741d60090531607058. Thanks!

keith:
Thanks!

PR #1407, opened 2017-03-31
add opt-in rules for Firebase developers

ulukaya:


ulukaya:
@jpsim WDYT?

masters3d:
This seems like the perfect example for the need of something similar to https://github.com/realm/SwiftLint/pull/711 .

> Also it would open door to adhoc, domain specific and niche rules without need to contribute it back to swiftlint.

ulukaya:
Hi @jpsim and @marcelofabri . Should I keep this in a separate branch or module. What do you suggest?

Source/SwiftLintFramework/Rules/FirebaseConfigActivateRule.swift
```diff
@@ -0,0 +1,65 @@
...
+        description: "Firebase Config should be activated.",
+        nonTriggeringExamples: [
+            "remoteConfig.fetch(withExpirationDuration: TimeInterval(expirationDuration)) {" +
+                " (status, error) -> Void in \n self.remoteConfig.activateFetched() \n }"
+        ],
+        triggeringExamples: [
+            "remoteConfig.fetch(withExpirationDuration: TimeInterval(expirationDuration)) {" +
+                " (status, error) -> Void in \n }"
+        ]
+    )
```
ornithocoder: What if the codebase has an instance method called .fetch which is not related to Firebase? By the code below it seems a violation would be triggered.

Source/SwiftLintFramework/Rules/FirebaseConfigDefaultsRule.swift
```diff
@@ -0,0 +1,55 @@
...
+        nonTriggeringExamples: [
+            "class ViewController: UIViewController {\n" +
+            "  override func viewDidLoad() {\n" +
+            "    super.viewDidLoad() \n" +
+            "    remoteConfig.setDefaults(fromPlist: \"RemoteConfigDefaults\")\n" +
+            "  }\n" +
+            "}"
+        ],
+        triggeringExamples: [
+            "class ViewController: UIViewController {\n" +
```
ornithocoder: Do all the .viewDidLoad methods in the application have to call `.setDefaults(fromPlist:)`? What if the codebase has a View Controller that doesn't require it?

Source/SwiftLintFramework/Rules/FirebaseConfigFetchRule.swift
```diff
@@ -0,0 +1,57 @@
...
+            "  override func viewDidLoad() {\n" +
+            "    super.viewDidLoad() \n" +
+            "    remoteConfig.fetch(withExpirationDuration: TimeInterval(expirationDuration)) {" +
+            "        (status, error) -> Void in \n" +
+            "    }\n" +
+            "  }\n" +
+            "}"
+        ],
+        triggeringExamples: [
+            "class ViewController: UIViewController {\n" +
```
ornithocoder: Same question here.

Source/SwiftLintFramework/Rules/FirebaseCoreRule.swift
```diff
@@ -0,0 +1,57 @@
...
+            "  }\n" +
+            "}"
+        ],
+        triggeringExamples: [
+            "class AppDelegate: UIResponder, UIApplicationDelegate {\n" +
+            "  func application(_ application: UIApplication," +
+            "      didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {\n" +
+            "    return true \n" +
+            "  }\n" +
+            "}"
```
ornithocoder: What if there's an object response for configuring all the 3rd-party software in the application. For instance: Firebase, Mixpanel, etc...

Source/SwiftLintFramework/Rules/FirebaseDynamicLinksCustomSchemeURLRule.swift
```diff
@@ -0,0 +1,69 @@
...
+            "  }\n" +
+            "}"
+        ],
+        triggeringExamples: [
+            "class AppDelegate: UIResponder, UIApplicationDelegate {\n" +
+                "  func application(_ application: UIApplication, open url: URL," +
+                "      sourceApplication: String?, annotation: Any) -> Bool {\n" +
+                "    return true \n" +
+                "  }\n" +
+            "}"
```
ornithocoder: Same question as before.

Source/SwiftLintFramework/Rules/FirebaseDynamicLinksSchemeURLRule.swift
```diff
@@ -0,0 +1,69 @@
...
+            "  }\n" +
+            "}"
+        ],
+        triggeringExamples: [
+            "class AppDelegate: UIResponder, UIApplicationDelegate {\n" +
+            "  func application(_ application: UIApplication," +
+            "      didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {\n" +
+            "    return true \n" +
+            "  }\n" +
+            "}"
```
ornithocoder: Same question as before.

Source/SwiftLintFramework/Rules/FirebaseDynamicLinksUniversalLinkRule.swift
```diff
@@ -0,0 +1,59 @@
...
+            "  }\n" +
+            "}"
+        ],
+        triggeringExamples: [
+            "class AppDelegate: UIResponder, UIApplicationDelegate {\n" +
+                "  func application(_ application: UIApplication, continue userActivity: NSUserActivity," +
+                "      restorationHandler: @escaping ([Any]?) -> Void) -> Bool {\n" +
+                "    return true \n" +
+                "  }\n" +
+            "}"
```
ornithocoder: Same question as before.

Source/SwiftLintFramework/Rules/FirebaseInvitesRule.swift
```diff
@@ -0,0 +1,69 @@
...
+            "  }\n" +
+            "}"
+        ],
+        triggeringExamples: [
+            "class AppDelegate: UIResponder, UIApplicationDelegate {\n" +
+                "  func application(_ application: UIApplication, open url: URL," +
+                "      options: [UIApplicationOpenURLOptionsKey : Any]) -> Bool {\n" +
+                "    return true \n" +
+                "  }\n" +
+            "}"
```
ornithocoder: Same question.

Source/SwiftLintFramework/Rules/FirebaseConfigDefaultsRule.swift
ornithocoder: For instance, this rule triggered a violation on one of [Firefox's View Controllers](https://github.com/mozilla-mobile/firefox-ios/blob/067b6670e8216e7822f94dd3881f25ff481d31b8/Extensions/SendTo/ActionViewController.swift#L19:14). If Firefox were using Firebase, does it mean this View Controller would have to call `.setDefaults(fromPlist:)`?

CHANGELOG.md
```diff
@@ -39,6 +39,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1546](https://github.com/realm/SwiftLint/issues/1546)
 
+* Add opt-in rules for Firebase SDK.
```
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseConfigActivateRule.swift
ulukaya: added the withExpirationDuration argument to check. Being an opt-in rule only used for Firebase devs, that should have a better false positive rate.

Source/SwiftLintFramework/Rules/FirebaseConfigDefaultsRule.swift
ulukaya: Now validates only on instances where remoteConfig.fetch(withExpirationDuration:) is used.

Source/SwiftLintFramework/Rules/FirebaseConfigFetchRule.swift
ulukaya: Now instead of it's availability, it checks that if it's called, it should be called inside viewdidload

Source/SwiftLintFramework/Rules/FirebaseCoreRule.swift
ulukaya: I now check if there is helper function doing it as well. So checking all the way down in AST. Finding if there is a function getting created there, that would do this in it's initialization would be complicated.

I'll leave this rule for the suggested use case, hence it's an opt-in rule, which only returns warning, that should be enough coverage. The rule is towards new devs that could forget to initialize these objects before use. The dev who's already initializing with a special class would already be aware of it, hence won't need the rule or just ignore the warning.

Source/SwiftLintFramework/Rules/FirebaseDynamicLinksCustomSchemeURLRule.swift
ulukaya: Same answer as before.

Source/SwiftLintFramework/Rules/FirebaseDynamicLinksSchemeURLRule.swift
ulukaya: Same answer as before.

Source/SwiftLintFramework/Rules/FirebaseInvitesRule.swift
ulukaya: Same answer as before.

ulukaya:
updated with answers to latest review.

ulukaya:
Thanks @ornithocoder for the review, updated the code wrt that.

Source/SwiftLintFramework/Rules/FirebaseConfigActivateRule.swift
```diff
@@ -0,0 +1,56 @@
...
+
+    public func validate(file: File, kind: SwiftExpressionKind,
+                         dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard SwiftExpressionKind.call == kind else {
+            return []
+        }
+
+        guard dictionary.name!.hasSuffix(".fetch") else {
```
ornithocoder: This will fail in case `name` is `nil`. Consider using guard to make sure `name` exists before using it.

```swift
guard let name = dictionary.name, name.hasSuffix(".fetch") else {
    // code...
}
```

Compare these examples:

```swift
extension Dictionary where Key: ExpressibleByStringLiteral {
    var name: String? {
        return self["name"] as? String
    }
}

func isValidDictionary1(_ dictionary: [String: String]) -> Bool {
    guard let name = dictionary.name, name.hasSuffix(".fetch") else { return false }
    return true
}

func isValidDictionary2(_ dictionary: [String: String]) -> Bool {
    guard dictionary.name!.hasSuffix(".fetch") else { return false }
    return true
}

isValidDictionary1(["name": "test.fetch"])   // true
isValidDictionary1(["aname": "test.fetch"])  // false

isValidDictionary2(["name": "test.fetch"])   // true
isValidDictionary2(["aname": "test.fetch"])  // fatal error: unexpectedly found nil while unwrapping an Optional value
```

Source/SwiftLintFramework/Rules/FirebaseConfigActivateRule.swift
```diff
@@ -0,0 +1,56 @@
...
+            return []
+        }
+
+        guard dictionary.substructure[0].name == "withExpirationDuration" else {
```
ornithocoder: The same logic applies here. if a method called `.fetch` (that's not the Firebase `.fetch`) doesn't take any parameter, it won't have a substructure in `[0]` (or `.first`).

Source/SwiftLintFramework/Rules/FirebaseConfigDefaultsRule.swift
```diff
@@ -0,0 +1,83 @@
...
+        ]
+    )
+
+    public func validate(file: File, kind: SwiftExpressionKind,
+                         dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard SwiftExpressionKind.call == kind else {
+            return []
+        }
+
+        guard dictionary.name!.hasSuffix(".fetch") else {
```
ornithocoder: Same "!" issue as before.

Source/SwiftLintFramework/Rules/FirebaseConfigDefaultsRule.swift
```diff
@@ -0,0 +1,83 @@
...
+            return []
+        }
+
+        guard dictionary.substructure[0].name == "withExpirationDuration" else {
```
ornithocoder: Same issue trying to access the first element of the array (it might not have a first element).

Source/SwiftLintFramework/Rules/FirebaseConfigDefaultsRule.swift
```diff
@@ -0,0 +1,83 @@
...
+        guard first.inheritedTypes.contains("UIViewController")  else {
+            return []
+        }
+
+        for method in first.substructure where
+            SwiftDeclarationKind.functionMethodInstance.rawValue == method.kind &&
+                method.name == "viewDidLoad()" {
+            return validateRecursive(file: file, dictionary: method)
+        }
+        return [StyleViolation(ruleDescription: type(of: self).description,
```
ornithocoder: Does it need to throw this violation here? It recursively looks for a violation - inside the for loop - and if it doesn't find one it shows anyway?

Source/SwiftLintFramework/Rules/FirebaseConfigDefaultsRule.swift
```diff
@@ -0,0 +1,83 @@
...
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "firebase_config_defaults",
+        name: "Firebase Config Defaults",
+        description: "Firebase Config defaults should be set.",
+        nonTriggeringExamples: [
+            "class ViewController: UIViewController {\n" +
```
ornithocoder: Consider adding more examples that do not trigger the violation. Like a `viewDidLoad()` implementation that doesn't have the `.fetch(withExpirationDuration:)` inside.

Source/SwiftLintFramework/Rules/FirebaseConfigFetchRule.swift
```diff
@@ -0,0 +1,85 @@
...
+        ]
+    )
+
+    public func validate(file: File, kind: SwiftExpressionKind,
+                         dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard SwiftExpressionKind.call == kind else {
+            return []
+        }
+
+        guard dictionary.name!.hasSuffix(".fetch") else {
```
ornithocoder: Same issue here.

Source/SwiftLintFramework/Rules/FirebaseConfigFetchRule.swift
```diff
@@ -0,0 +1,85 @@
...
+            return []
+        }
+
+        guard dictionary.substructure[0].name == "withExpirationDuration" else {
```
ornithocoder: Same issue here.

Source/SwiftLintFramework/Rules/FirebaseConfigFetchRule.swift
```diff
@@ -0,0 +1,85 @@
...
+        guard first.inheritedTypes.contains("UIViewController")  else {
+            return []
+        }
+
+        for method in first.substructure where
+            SwiftDeclarationKind.functionMethodInstance.rawValue == method.kind &&
+                method.name == "viewDidLoad()" {
+                    return validateRecursive(file: file, dictionary: method)
+        }
+        return [StyleViolation(ruleDescription: type(of: self).description,
```
ornithocoder: It's throwing a violation even if the recursive analysis fails to find one.

Source/SwiftLintFramework/Rules/FirebaseConfigFetchRule.swift
```diff
@@ -0,0 +1,85 @@
...
+                               severity: configuration.severity,
+                               location: Location(file: file, byteOffset: first.offset ?? 0))]
+    }
+
+    public func validateBaseCase(dictionary: [String : SourceKitRepresentable]) -> Bool {
+        if let kindString = dictionary.kind, SwiftExpressionKind(rawValue: kindString) == .call,
+            let name = dictionary.name, name.hasSuffix(".fetch"),
+                dictionary.substructure[0].name == "withExpirationDuration" {
```
ornithocoder: It might not be there.

Source/SwiftLintFramework/Rules/FirebaseConfigActivateRule.swift
```diff
@@ -0,0 +1,56 @@
...
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "firebase_config_activate",
+        name: "Firebase Config Activate",
+        description: "Firebase Config should be activated.",
+        nonTriggeringExamples: [
+            "remoteConfig.fetch(withExpirationDuration: TimeInterval(expirationDuration)) {" +
```
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseConfigActivateRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseConfigActivateRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseConfigDefaultsRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseConfigDefaultsRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseConfigDefaultsRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseConfigFetchRule.swift
ulukaya: It's the case, they ignored the the other rule of putting fetch(withExperiration) in viewDidload.
So if there is no viewDidLoad at all.
Added an example.

Source/SwiftLintFramework/Rules/FirebaseConfigFetchRule.swift
ulukaya: Done

ulukaya:
@ornithocoder Thanks for the great reviews, updated the code.

Source/SwiftLintFramework/Rules/FirebaseConfigDefaultsRule.swift
ulukaya: Yes, in case viewDidLoad was not implemented/omitted. Added a triggering example to show it.

Source/SwiftLintFramework/Rules/FirebaseConfigFetchRule.swift
ulukaya: Yes, in case viewDidLoad was not implemented/omitted. Added a triggering example to show it.

ulukaya:
Hi @jpsim and @marcelofabri , do you mind taking a look? This seems ready to be merged.

ornithocoder:
You're welcome, @ulukaya. I have some considerations and concerns about this PR, tho.

1. Once these rules are enabled, they will show violations for each and every View Controller that doesn't call Firebase's methods on `.viewDidLoad()`. They will trigger violations even for View Controllers that don't need to make the Firebase calls or those that don't override that method. I understand it's just a warning, but several teams treat warnings as errors.

2. The same is true for the App Delegate. These rules assume all the logic needs to go inside App Delegate in a very specify way. Good practice says App Delegate should not be a dumping area for random code and, for this reason, developers create objects to handle specific tasks such as initializing and configuring 3rd-party apps like Mixpanel, HockeyApp, Firebase, etc... These rules won't catch Firebase being initialized from a different context and will trigger violations.

3. It is true that some teams don't treat warnings as errors, but for a large project, 1) one would have to keep a mental counter of the number of false positives created by these rules, and/or 2) manually check the all the warnings after compilation, and/or 3) enable and disable SwiftLint or these rules every now and then to check the real number of the warnings.

I see SwiftLint as a tool for finding style and convention violations. IMO this PR is more about enforcing a certain template and **testing the implementation** than finding style and convention violations.

These rules would be a nice feature for SwiftLint if SwiftLint had support for 3rd-party plugins/extensions, but it doesn't. There's an issue for implementing this support, tho.

For now, I'd suggest Firebase to have

1. a good documentation covering Firebase setup (including a step by step tutorial), so there's no doubt about what's necessary to implement and how to implement.

2. Quick/Cedar/etc... shared examples that can be added to the developer test suite (for TDD, BDD, simple unit tests, etc...). For instance, when testing a view controller that requires Firebase setup, one would simply add to the VC test: `itBehavesLike("a Firebase view controller")`

3. Xcode file templates, to help developers with boilerplate code

4. Xcode code snippets (like Apple does - see below), to speed up development.

screen shot 2017-06-08 at 12 32 18 am

My 2 cents :-)

ulukaya:
Hey @ornithocoder ,

1) This is completely wrong. We only fire the warning if they already call the firebase remoteconfig method, and only in that specific controller, not everywhere.

2) Rules are meant to be for the novice developer, and we always show the configure method in appdelegate as a best practice. Devs who are already aware where each call should have been made, won't need the rule. This is for saving new Firebase developers from basic mistakes.

3) Firebase is modular, they can pick only the rules depends on the modules of Firebase they use. If they are already a power user, they won't need any of them.

Thanks for the other suggestions, we either already did those or working on them. SwiftLint was more of a experimental/additional project for us to guide new developers. Not a tool to find bugs for power users. I know each of 1-2-3-4 would have a higher impact, but these rules for a new developer trying features and copy/pasting under wrong functions. Still happy to merge only few of these rules that'd have less false positives just to experiment.

We are definitely gonna enable 1-2-3-4, but with the addition of these rules if we can save time of new devs from scratching their head because the feature is not working right, that'd still be a good addition. I'll look into creating more valuable rules in the future, that was just my inital stab on SwiftLint.

Cheers,

ornithocoder:
Hi @ulukaya,

1. My bad. This SwiftLintBot isn't updating the violations (bummer!) and keeps showing the View Controller's violations. I have some additional suggestion then,

In `FirebaseConfigDefaultsRule.swift` and `FirebaseConfigFetchRule.swift` these two non triggering examples.

```swift
"class ViewController: UIViewController {\n" +
"  override func viewDidLoad() {\n" +
"    super.viewDidLoad()\n" +
"  }\n" +
"}"
```

and

```swift
"class ViewController: UIViewController { }"
```

They will make it explicit that `. viewDidLoad()` won't trigger violation, even when the class doesn't override it.

Another change:

```swift
        guard ... else {
            return []
        }

        guard ... else {
            return []
        }

        guard ... else {
            return []
        }

        guard ... else {
            return []
        }

        guard ... else {
            return []
        }
```

can be combined into one guard. This happens in different rules. I'll highlight them using the Code Review tool.

Regards

Source/SwiftLintFramework/Rules/FirebaseConfigDefaultsRule.swift
```diff
@@ -0,0 +1,97 @@
...
+            "    remoteConfig.setDefaults(fromPlist: \"RemoteConfigDefaults\")\n" +
+            "  }\n" +
+            "}",
+            "class ViewController: UIViewController {\n" +
+            "  override func viewDidLoad() {\n" +
+            "    super.viewDidLoad() \n" +
+            "    foo.fetch() { } \n" +
+            "    foo.fetch(fromURL: URL) { } \n" +
+            "  }\n" +
+            "}"
```
ornithocoder: Please consider adding these two examples:

```swift
            "class ViewController: UIViewController {\n" +
                "  override func viewDidLoad() {\n" +
                "    super.viewDidLoad()\n" +
                "  }\n" +
            "}",
            "class ViewController: UIViewController { }",
```

The first one will make sure an "empty" `viewDidLoad()` won't trigger a violation. The second one will make sure a View Controller that doesn't override `viewDidLoad()` won't trigger a violation as well.

Source/SwiftLintFramework/Rules/FirebaseConfigFetchRule.swift
```diff
@@ -0,0 +1,92 @@
...
+        nonTriggeringExamples: [
+            "class ViewController: UIViewController {\n" +
+            "  override func viewDidLoad() {\n" +
+            "    super.viewDidLoad() \n" +
+            "    remoteConfig.fetch(withExpirationDuration: TimeInterval(expirationDuration)) {" +
+            "        (status, error) -> Void in \n" +
+            "    }\n" +
+            "  }\n" +
+            "}"
+        ],
```
ornithocoder: Please consider adding these two examples:

```swift
            "class ViewController: UIViewController {\n" +
                "  override func viewDidLoad() {\n" +
                "    super.viewDidLoad()\n" +
                "  }\n" +
            "}",
            "class ViewController: UIViewController { }",
```

The first one will make sure an "empty" `viewDidLoad()` won't trigger a violation. The second one will make sure a View Controller that doesn't override `viewDidLoad()` won't trigger a violation as well.

Source/SwiftLintFramework/Rules/FirebaseConfigActivateRule.swift
```diff
@@ -0,0 +1,60 @@
...
+            return []
+        }
+
+        guard let name = dictionary.name, name.hasSuffix(".fetch") else {
+            return []
+        }
+
+        guard let param = dictionary.substructure.first, param.name == "withExpirationDuration" else {
+            return []
+        }
```
ornithocoder: These `guard` can be combined into one

```swift
guard
    let a = a,
    let b = b,
    c == something
    else {
        return []
}
```

Source/SwiftLintFramework/Rules/FirebaseConfigDefaultsRule.swift
```diff
@@ -0,0 +1,97 @@
...
+            return []
+        }
+
+        guard let first = file.structure.dictionary.substructure.first else {
+            return []
+        }
+
+        guard first.inheritedTypes.contains("UIViewController")  else {
+            return []
+        }
```
ornithocoder: These `guard` can be combined into one

```swift
guard
    let a = a,
    let b = b,
    c == something
    else {
        return []
}
```

Source/SwiftLintFramework/Rules/FirebaseConfigFetchRule.swift
```diff
@@ -0,0 +1,92 @@
...
+            return []
+        }
+
+        guard let first = file.structure.dictionary.substructure.first else {
+            return []
+        }
+
+        guard first.inheritedTypes.contains("UIViewController")  else {
+            return []
+        }
```
ornithocoder: These `guard` can be combined into one

```swift
guard
    let a = a,
    let b = b,
    c == something
    else {
        return []
}
```

CHANGELOG.md
```diff
@@ -39,6 +39,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1546](https://github.com/realm/SwiftLint/issues/1546)
 
+* Add opt-in rules for Firebase SDK.  
+  [Ibrahim Ulukaya (Google Inc.)](https://github.com/iulukaya)
+
+
```
ornithocoder: No need for this extra line.

Source/SwiftLintFramework/Rules/FirebaseConfigActivateRule.swift
```diff
@@ -0,0 +1,60 @@
...
+        let fetchClosure = dictionary.substructure[1].substructure[2]
+        return validateRecursive(file: file, dictionary: fetchClosure)
+    }
+
+    public func validateBaseCase(dictionary: [String: SourceKitRepresentable]) -> Bool {
+        if let kindString = dictionary.kind, SwiftExpressionKind(rawValue: kindString) == .call,
+            let name = dictionary.name, name.hasSuffix(".activateFetched") {
+            return true
+        }
+        return false
```
ornithocoder: Consider using a `guard` here, for for early return.

Source/SwiftLintFramework/Rules/FirebaseConfigDefaultsRule.swift
```diff
@@ -0,0 +1,97 @@
...
+                method.name == "viewDidLoad()" {
+            return validateRecursive(file: file, dictionary: method)
+        }
+        return [StyleViolation(ruleDescription: type(of: self).description,
+                               severity: configuration.severity,
+                               location: Location(file: file, byteOffset: first.offset ?? 0))]
+    }
+
+    public func validateBaseCase(dictionary: [String : SourceKitRepresentable]) -> Bool {
+        if let kindString = dictionary.kind, SwiftExpressionKind(rawValue: kindString) == .call,
```
ornithocoder: Consider using a `guard` here, for for early return.

Source/SwiftLintFramework/Rules/FirebaseConfigFetchRule.swift
```diff
@@ -0,0 +1,92 @@
...
+                method.name == "viewDidLoad()" {
+                    return validateRecursive(file: file, dictionary: method)
+        }
+        return [StyleViolation(ruleDescription: type(of: self).description,
+                               severity: configuration.severity,
+                               location: Location(file: file, byteOffset: first.offset ?? 0))]
+    }
+
+    public func validateBaseCase(dictionary: [String : SourceKitRepresentable]) -> Bool {
+        if let kindString = dictionary.kind, SwiftExpressionKind(rawValue: kindString) == .call,
```
ornithocoder: Consider using a `guard` here, for for early return.

Source/SwiftLintFramework/Rules/FirebaseCoreRule.swift
```diff
@@ -0,0 +1,59 @@
...
+        ]
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        let appDelegate = file.structure.dictionary.substructure
+        if let first = appDelegate.first, first.inheritedTypes.contains("UIApplicationDelegate") {
```
ornithocoder: Consider using a `guard` here, for for early return.

Source/SwiftLintFramework/Rules/FirebaseCoreRule.swift
```diff
@@ -0,0 +1,59 @@
...
+                SwiftDeclarationKind.functionMethodInstance.rawValue == method.kind &&
+                    method.name == "application(_:didFinishLaunchingWithOptions:)" {
+                return validateRecursive(file: file, dictionary: method)
+            }
+        }
+        return []
+    }
+
+    public func validateBaseCase(dictionary: [String : SourceKitRepresentable]) -> Bool {
+        if let kindString = dictionary.kind, SwiftExpressionKind(rawValue: kindString) == .call,
```
ornithocoder: Consider using a `guard` here, for for early return.

Source/SwiftLintFramework/Rules/FirebaseDynamicLinksCustomSchemeURLRule.swift
```diff
@@ -0,0 +1,66 @@
...
+        ]
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        let appDelegate = file.structure.dictionary.substructure
+        if let first = appDelegate.first, first.inheritedTypes.contains("UIApplicationDelegate") {
```
ornithocoder: Consider using a `guard` here, for for early return.

Source/SwiftLintFramework/Rules/FirebaseDynamicLinksCustomSchemeURLRule.swift
```diff
@@ -0,0 +1,66 @@
...
+            }
+            return [StyleViolation(ruleDescription: type(of: self).description,
+                                   severity: configuration.severity,
+                                   location: Location(file: file, byteOffset: first.offset ?? 0))]
+        }
+        return []
+    }
+
+    public func validateBaseCase(dictionary: [String : SourceKitRepresentable]) -> Bool {
+        if let kindString = dictionary.kind, SwiftExpressionKind(rawValue: kindString) == .call,
```
ornithocoder: Consider using a `guard` here, for for early return.

Source/SwiftLintFramework/Rules/FirebaseDynamicLinksSchemeURLRule.swift
```diff
@@ -0,0 +1,62 @@
...
+        ]
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        let appDelegate = file.structure.dictionary.substructure
+        if let first = appDelegate.first, first.inheritedTypes.contains("UIApplicationDelegate") {
```
ornithocoder: Consider using a `guard` here, for for early return.

Source/SwiftLintFramework/Rules/FirebaseDynamicLinksSchemeURLRule.swift
```diff
@@ -0,0 +1,62 @@
...
+            }
+            return [StyleViolation(ruleDescription: type(of: self).description,
+                                   severity: configuration.severity,
+                                   location: Location(file: file, byteOffset: first.offset ?? 0))]
+        }
+        return []
+    }
+
+    public func validateBaseCase(dictionary: [String : SourceKitRepresentable]) -> Bool {
+        if let kindString = dictionary.kind, SwiftExpressionKind(rawValue: kindString) == .call,
```
ornithocoder: Consider using a `guard` here, for for early return.

Source/SwiftLintFramework/Rules/FirebaseDynamicLinksUniversalLinkRule.swift
```diff
@@ -0,0 +1,64 @@
...
+        ]
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        let appDelegate = file.structure.dictionary.substructure
+        if let first = appDelegate.first, first.inheritedTypes.contains("UIApplicationDelegate") {
```
ornithocoder: Consider using a `guard` here, for for early return.

Source/SwiftLintFramework/Rules/FirebaseDynamicLinksUniversalLinkRule.swift
```diff
@@ -0,0 +1,64 @@
...
+            }
+            return [StyleViolation(ruleDescription: type(of: self).description,
+                                   severity: configuration.severity,
+                                   location: Location(file: file, byteOffset: first.offset ?? 0))]
+        }
+        return []
+    }
+
+    public func validateBaseCase(dictionary: [String : SourceKitRepresentable]) -> Bool {
+        if let kindString = dictionary.kind, SwiftExpressionKind(rawValue: kindString) == .call,
```
ornithocoder: Consider using a `guard` here, for for early return.

Source/SwiftLintFramework/Rules/FirebaseInvitesRule.swift
```diff
@@ -0,0 +1,66 @@
...
+            }
+            return [StyleViolation(ruleDescription: type(of: self).description,
+                                   severity: configuration.severity,
+                                   location: Location(file: file, byteOffset: first.offset ?? 0))]
+        }
+        return []
+    }
+
+    public func validateBaseCase(dictionary: [String : SourceKitRepresentable]) -> Bool {
+        if let kindString = dictionary.kind, SwiftExpressionKind(rawValue: kindString) == .call,
```
ornithocoder: Consider using a `guard` here, for for early return.

Source/SwiftLintFramework/Rules/FirebaseInvitesRule.swift
```diff
@@ -0,0 +1,66 @@
...
+                "      options: [UIApplicationOpenURLOptionsKey : Any]) -> Bool {\n" +
+                "    return true \n" +
+                "  }\n" +
+            "}"
+        ]
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        let appDelegate = file.structure.dictionary.substructure
+        if let first = appDelegate.first, first.inheritedTypes.contains("UIApplicationDelegate") {
```
ornithocoder: Consider using a `guard` here, for for early return.

Source/SwiftLintFramework/Rules/FirebaseConfigDefaultsRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseConfigDefaultsRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseConfigFetchRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseConfigFetchRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseConfigActivateRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseConfigFetchRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseConfigActivateRule.swift
ulukaya: Done

CHANGELOG.md
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseCoreRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseInvitesRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseDynamicLinksUniversalLinkRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseInvitesRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseDynamicLinksUniversalLinkRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseDynamicLinksSchemeURLRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseDynamicLinksSchemeURLRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseDynamicLinksCustomSchemeURLRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseDynamicLinksCustomSchemeURLRule.swift
ulukaya: Done

Source/SwiftLintFramework/Rules/FirebaseCoreRule.swift
ulukaya: Done

ulukaya:
Thanks for the review @ornithocoder , added guard for early return, grouped guards,

ulukaya:
WDYT @ornithocoder ?

marcelofabri:
@ulukaya First of all, sorry for the delay on this one.

I'm not personally comfortable on merging this because it's so tied to Firebase and it's not something that I personally can make a commitment to support (i.e. making sure it still works) in the future.

I'm also worried about setting a precedence here for niche rules, even as opt-in. I know that we kind of already have one of those (`nimble_operator`), but that one is more about style than correctness.

In short, I see 3 possible outcomes for this:

1) Another maintainer makes a commitment to merge this and "be responsible" for it.
2) We come up with a proper 3rd party plugin system
3) You/Firebase creates another tool using SourceKitten and maybe even SwiftLintFramework to implement this. I think this is the best option for the short term.

This is my personal opinion and not necessarily the final decision. I'd welcome everyone to jump in and share their thoughts.

marcelofabri:
I'm closing this for now since no one expressed their thoughts after my last comment.

ulukaya:
Hi @marcelofabri 
@jpsim was ok to merge this as long as rules were opt-in. Hence my commitment to keep these rules uptodate and update this PR consistently. I'd come up with option 2-3 but I'd love to try it out with our developers and get feedback before I commit to that.

It's a little odd that you don't respond for 3 months then you do one comment, when you don't hear few days you close the PR.

Happy to work on this, as long as you guys are still on top of this project. If it's stale, happy to keep this experimental and move on.

marcelofabri:
@ulukaya Sorry if this sent the wrong message.

> @jpsim was ok to merge this as long as rules were opt-in. Hence my commitment to keep these rules uptodate and update this PR consistently.

If @jpsim or any other maintainer wants to reopen this and merge, I'm ok with that. I just can't commit myself to support this in the future. Closing the PR doesn't mean that this will never go in, it just means that I wanted to focus on other things.

> I'd come up with option 2-3 but I'd love to try it out with our developers and get feedback before I commit to that.

I think you can still do that by providing your own SwiftLint binary compiled from your fork.

> It's a little odd that you don't respond for 3 months then you do one comment, when you don't hear few days you close the PR.

I took 3 months to respond because I wasn't comfortable merging this (as I explained in https://github.com/realm/SwiftLint/pull/1407#issuecomment-309304710), so I was waiting to see if others would jump into this first. Maybe this was a mistake and it'd be better if I expressed my concerns since the beginning. Again, sorry about that.

PR #1410, opened 2017-04-02
Prevent false positive in shorthand_operator rule

### LINKED ISSUE #1254: shorthand_operator: false positive: `seconds = seconds * 60 + value`
`seconds = seconds * 60 + value`

We don't want to rewrite it to

```
seconds *= 60
seconds += value
```

as that is simply 2 lines and somehow destroys the code and mind flow.

Swiflint: 0.16.1

sammy-SC:
Fixes #1254.


CHANGELOG.md
```diff
@@ -109,6 +109,10 @@
 
 ##### Bug Fixes
 
+* Prevent false positive in `shorthand_operator` rule.
```
marcelofabri: this is the wrong section now

marcelofabri:
This is a tricky one, because we're actually introducing false negatives, for example: https://github.com/mozilla-mobile/firefox-ios/blob/58ededc6e2a534df3b2d77883319346774fb1da3/Client/Frontend/Browser/Punycode.swift#L150:13

sammy-SC:
@marcelofabri ok, it is a lot more complicated than I thought.
Do you have a recommendation how to go about this problem?

marcelofabri:
Since it's a default rule, I'm inclined to think that it's better to have some false negatives than false positives. You could leave some false negatives examples commented out in the rule, so we can (hopefully) address those in the future.

marcelofabri:
Maybe we could check if there are another operators on the right side of the assignment, they should have a higher precedence than the first one to be a violation.

So
```swift
n = n + i / outputLength // ok
d = d * 65536 + Double(_mantissa.0) // would trigger
```

sammy-SC:
Hey @marcelofabri 

I tried to give this a little more thought, it still has false negatives thought.

I added the false negative as a commented out example https://github.com/realm/SwiftLint/pull/1410/commits/b6133c03cf1dead24441170058fbcc41138c2dfc.

I introduced another regex which accounts for operator precedence. 

Let me know if there is anything else I should change.

Thank you

Source/SwiftLintFramework/Rules/ShorthandOperatorRule.swift
```diff
@@ -42,32 +43,43 @@ public struct ShorthandOperatorRule: ConfigurationProviderRule {
...
+        let operatorsWithoutPrecedence = escaped(["-", "+"])
+        let operatorsWithPrecedence = escaped(["/", "*"])
         let operand = "[\\w\\d\\.]+?"
         let spaces = "[^\\S\\r\\n]*?"
 
-        let pattern = "^\(spaces)(\(operand))\(spaces)=\(spaces)(\\1)\(spaces)\(escapedOperators)"
-        return pattern
+        let pattern1 = "^\(spaces)(\(operand))\(spaces)=\(spaces)(\\1)\(spaces)\(operatorsWithoutPrecedence)"
+        let pattern2 = "^\(spaces)(\(operand))\(spaces)=\(spaces)(\\1)\(spaces)\(operatorsWithPrecedence)\(spaces)\\S+$"
+        return [pattern1, pattern2]
```
marcelofabri: could this be a single regex composed by `pattern1` and `pattern2` with an *or*?

This way the rule would be a bit faster.

sammy-SC: Oh yea, that makes perfect sense :) 

it is now changed.

sammy-SC:
Hey @marcelofabri, 

I combined the regexes into one to improve performance. 
Let me know if there is anything else I could do.

marcelofabri:
Thanks @sammy-SC! 🚀

marcelofabri:
Also filled #1537 with a small bug/enhancement that can be done after.

jpsim:
Looks like the performance regressions reported by OSSCheck are accurate, but worthwhile due to the accuracy. Adds about ~10% to total lint time.

jpsim:
Nicely done! 🎉

sammy-SC:
@jpsim I think @SwiftLintBot is slightly outdated. 
@marcelofabri pointed out the performance issue. 
I initially had 2 regexes even thought it should be only single one. It was addressed here https://github.com/realm/SwiftLint/pull/1410/commits/42476d06bca1335b3be92c99c53fbbde8915f376

If you have an idea how to improve it or believe it is still an issue, I will try to take a look at it.

jpsim:
The OSSCheck script requires the PR branch to be up to date with `master`. This branch was not, so the results can't really be trusted.

PR #1411, opened 2017-04-03
Replace YamlSwift with Yams

### LINKED ISSUE #1412: Replace YamlSwift with Yams
Yams has feature that generating YAML string.
That will help implementing `swiftlint config` subcommand. #532

norio-nomura:
This depends on https://github.com/jpsim/SourceKitten/pull/352 and Yams 0.3.0.
So, it will require next SourceKitten release for building with SwiftPM.
Fixes #1412


norio-nomura:
How can I evaluate the message https://github.com/realm/SwiftLint/pull/1411#issuecomment-291031152 from SwiftLintBot? 🤔
This is my first PR since SwiftLintBot started working.

jpsim:
> How can I evaluate the message #1411 (comment) from SwiftLintBot? 🤔
> This is my first PR since SwiftLintBot started working.

I was surprised to see so many messages. However, upon closer inspection, it looks like there are only two projects with changes, and they're all "fixes" (as in the violations are reported on `SwiftLint@master` but not in this branch). The two projects (Sourcery & Kickstarter) both specify a custom `.swiftlint.yml`, which leads me to believe that this PR parses their yaml configurations differently than `SwiftLint@master` currently does.

jpsim:
I should also say, usually the posts from @SwiftLintBot will report any regression/improvement in performance comparing the PR branch to the current `SwiftLint@master`, as well as any violations that only happen in `master` but not the current PR, or vice-versa, which is useful to know if we're 1) fixing false positives or 2) expanding the scope of a rule or 3) introducing a new rule and want to know its impact, etc.

jpsim:
Looks like Yams (a.k.a. libYAML) is _stricter_ than YamlSwift.

```
Could not read configuration file at path '/path/to/Sourcery/.swiftlint.yml': Error parsing YAML: 20:5: error: scanner: while scanning an alias in line 20, column 4
did not find expected alphabetic or numeric character:
  - *.generated.swift
     ^: file /path/to/SwiftLint/Source/SwiftLintFramework/Models/Configuration.swift, line 162
```

What makes this interesting is that Sourcery (and Kickstarter) have `.swiftlint.yml` files that aren't strictly valid YAML according to libYAML and [yamllint.com](http://www.yamllint.com/).

We're faced with the important decision of choosing between continuing to support this invalid YAML (and therefore not break users' SwiftLint setups), or taking advantage of SwiftLint's pre-1.0 status still and enforcing fully-compliant YAML now rather than later.

norio-nomura:
I see.
Since I forgot to correct the parse error message of the configuration file, I'll add a commit.

norio-nomura:
I pushed commit that makes yaml error message to be locatable in Xcode's Issue Navigator:
screenshot 2017-04-05 12 44 01
It will need https://github.com/jpsim/Yams/pull/34 for displaying correct location.

I hope this will help transition to using strict YAML.

norio-nomura:
I added changelog entries for stricter yaml parsing and locatable configuration error.

@jpsim Thank you for description about SwiftLintBot :pray: That's great!

norio-nomura:
> We're faced with the important decision of choosing between continuing to support this invalid YAML …

AFAIK, libYAML does not have any options that accept such invalid YAML.
If we would continue to support those invalid YAML, we would need to continue using YamlSwift.

marcelofabri:
I think we should enforce valid YAML now and report it as an error in Xcode

norio-nomura:
What about the idea of "Reading with YamlSwift if Yams can not read with errors" ?

jpsim:
> What about the idea of "Reading with YamlSwift if Yams can not read with errors" ?

No, I think a "clean break" is preferable. Speaking to users in the iOS Folks slack organization, there didn't appear to be any pushback from that.

So this PR is fine, but we should remove the Sourcery and Kickstarter checks from OSSCheck until https://github.com/krzysztofzablocki/Sourcery/pull/208 and https://github.com/kickstarter/ios-oss/pull/127 are merged.

jpsim:
Now that krzysztofzablocki/Sourcery#208 and kickstarter/ios-oss#127 have been merged, CI is re-running after a194a40 and I'll merge once that passes. It should clean up the SwiftLintBot comment because SwiftLint won't crash when linting those two projects anymore.

jpsim:
Thanks for this @norio-nomura! 👏

norio-nomura:
Thanks! 🙂

PR #1416, opened 2017-04-04
Fix False Positives in `redundant_discardable_let`.

### LINKED ISSUE #1415: False Positives for `redundant_discardable_let`
`let` is necessary when trying to use an explicit type to disambiguate the a function, but this still triggers a warning.

For example, this is invalid syntax and does not compile:
```swift
_: UInt = 3 - 2
```
This is correct:
```swift
let _: UInt = 3 - 2
```
...but SwiftLint complains about the `let` and suggests switching back to the first.

SDGGiesbrecht:
This fixes #1415.

It allows `let _` if it is followed by exactly `:`.


SDGGiesbrecht:
The Linux build failure has nothing to do with this pull request:

It is a problem with Yaml 3.4.0:
```
/SwiftLint/Packages/Yaml-3.4.0/Sources/Yaml/YAMLTokenizer.swift:52:66: error: use of undeclared type 'NSRegularExpression'
```

Given the way the [dependency on Yaml was declared](https://github.com/realm/SwiftLint/blob/ee90273d79c00bb3fbd0c964b4652fa7cd9b82e6/Package.swift#L14), a hypothesis would be that [Yaml’s update two days ago](https://github.com/behrang/YamlSwift/releases) introduced the Linux incompatibility.

If that is the case, specifying the previous minor version (3.3) should fix the problem. Should I do that here, or would it be better for it to be handled independently elsewhere?

SDGGiesbrecht:
@jpsim, @marcelofabri, unless you want to fix [`master`’s Linux problem](https://github.com/realm/SwiftLint/pull/1416#issuecomment-291668396) here, this pull request is ready.

marcelofabri:
Wouldn't it be better to just update the regex [here](https://github.com/SDGGiesbrecht/SwiftLint/blob/d2155e8b8981da55cd1057e8f5e69c73ae96be35/Source/SwiftLintFramework/Rules/RedundantDiscardableLetRule.swift#L68)? 

I think your proposed fix don't catch spaces, for example `let _ : String = something()`.

SDGGiesbrecht:
I didn’t try to catch the space, since that would overlap with the `colon` rule. I can go back and add it though if you’d like. Is there already an existing SwiftLint method for parsing past Swift whitespace, (which differs from Unicode/ICU whitespace definitions and also includes comments)?

My reasoning for not just adjusting the regular expression was that it is clearer to other developers what the check is and why it is there, along the lines of [`isInBooleanCondition`](https://github.com/SDGGiesbrecht/SwiftLint/blob/d2155e8b8981da55cd1057e8f5e69c73ae96be35/Source/SwiftLintFramework/Rules/RedundantDiscardableLetRule.swift#L73). There are already other issues and pull requests dealing with different kinds of false positives for this same rule. If everyone just adjusts the regular expression, without a full understanding of the reason for each piece of it, there is a much greater risk of accidental side‐effects.

SDGGiesbrecht:
The [Linux problem](https://github.com/realm/SwiftLint/pull/1416#issuecomment-291668396) has been fixed in #1396. Once that is merged, tests will pass.

jpsim:
Since #1411 was merged to `master`, we no longer depend on YamlSwift, and CI should pass again with this PR. Could you please rebase against `master` and resolve the merge conflict with `CHANGELOG.md`? Thanks!

SDGGiesbrecht:
Will do.

jpsim:
Merged in 9b4f495507cf5e21cee249c2cab85b7844b6b685, fb9c482683e0ea5258f464bdb42083c72064de07 and e2ecafea5d272aa4583e8a1ea65567217fbf24ad because I wanted to make a small refactor, I hope you don't mind.

Thanks for the great fix!

SDGGiesbrecht:
Thank you.

PR #1430, opened 2017-04-12
Make `emoji` and `checkstyle` reporter to output report sorted by file name

### LINKED ISSUE #1429: `TestCheckstyleReporter` and `testEmojiReporter` sometimes fail on Swift 3.1 for Linux
Since the enumeration of the dictionary is indefinite order.
e.g. https://travis-ci.org/realm/SwiftLint/jobs/221323293#L561-L571

norio-nomura:
Fixes #1429

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,9 @@
 
 ##### Bug Fixes
 
-* None.
+* `emoji` and `checkstyle` reporter output report sorted by file name.  
+  [norio-nomura](https://github.com/norio-nomura)
+  [#1429](https://github.com/realm/SwiftLint/issues/1429)
 
 ## 0.18.1: Misaligned Drum
 
--- a/Source/SwiftLintFramework/Reporters/CheckstyleReporter.swift
+++ b/Source/SwiftLintFramework/Reporters/CheckstyleReporter.swift
@@ -21,6 +21,7 @@ public struct CheckstyleReporter: Reporter {
             "<?xml version=\"1.0\" encoding=\"utf-8\"?>\n<checkstyle version=\"4.3\">",
             violations
                 .group(by: { ($0.location.file ?? "<nopath>").escapedForXML() })
+                .sorted(by: { $0.key < $1.key })
                 .map(generateForViolationFile).joined(),
             "\n</checkstyle>"
         ].joined()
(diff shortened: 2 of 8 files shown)


norio-nomura:
String comparison result is different between mac and Linux ... 😞 
```terminal.sh-session
$ TOOLCHAINS=org.swift.31120170327a swift
swift_oss_helper command enabled.
warning: (x86_64) /Library/Developer/Toolchains/swift-3.1-RELEASE.xctoolchain/System/Library/PrivateFrameworks/LLDB.framework/Resources/repl_swift empty dSYM file detected, dSYM was created with an executable with no debug info.
Welcome to Apple Swift version 3.1 (swift-3.1-RELEASE). Type :help for assistance.
  1> "Other" < "filename"
$R0: Bool = true
  2>  
```
```terminal.sh-session
$ docker run --privileged -it -v `pwd`:`pwd`:cached -w `pwd` --rm norionomura/sourcekit:31 swift
Welcome to Swift version 3.1 (swift-3.1-RELEASE). Type :help for assistance.
  1> "Other" < "filename"
$R0: Bool = false
  2>  
```

norio-nomura:
[SR-530](https://bugs.swift.org/browse/SR-530)

norio-nomura:
[StringComparable.swift#L111-L112](https://github.com/apple/swift/blob/fc6add388026b54a6b1142bcc3d77231dc9be017/stdlib/public/core/StringComparable.swift#L111-L112):
```swift
  public  // @testable
  func _compareString(_ rhs: String) -> Int {
#if _runtime(_ObjC)
    // We only want to perform this optimization on objc runtimes. Elsewhere,
    // we will make it follow the unicode collation algorithm even for ASCII.
    if _core.isASCII && rhs._core.isASCII {
      return _compareASCII(rhs)
    }
#endif
    return _compareDeterministicUnicodeCollation(rhs)
  }
```
When comparing ASCII character strings, there seems to be a difference depending on whether ObjC runtime is present:
- If present, the same behavior as Foundation
- If not, the behavior defined by Unicode

I think that the value expected for our test results should be switched depending on whether ObjC runtime is present or not.

norio-nomura:
Updated.

jpsim:
Also, way to fix this in Swift! 👏 

(https://github.com/apple/swift/pull/8769)

norio-nomura:
Thanks! 🙏 

> Also, way to fix this in Swift! 👏

Apparently, https://github.com/apple/swift/pull/8769 seems to be incompatible with the String Manifesto.

PR #1433, opened 2017-04-13
fix false positive for empty count violation

rounak:
Having a normal count variable and checking that for `> 0` triggers a false positive


jpsim:
Please see my comment from the last time someone submitted this: https://github.com/realm/SwiftLint/pull/714#issuecomment-230023836

jpsim:
Sorry, but I believe this false positive is better than missing some violations. From a previous comment on a PR with the same change from a few months ago (#635):

> the reason I kept the false positive here is to catch it in cases where you're extending `CollectionType`s, where accessing `self.count` without `self`. So I'm inclined not to accept this change, especially because I'm happy to have a higher tolerance for false positives in opt-in rules.

Still, thanks for your contribution @rounak , I hope you contribute again in the future!

Issue #1408, opened 2017-03-31
Adding custom rule with an exception


jayesh15111988:
I am trying to add a custom rule to enforce no space after opening parentheses. Custom rule is defined as follows
```
no_space_after_opening_parentheses:
        name: "No space after opening parentheses"
        message: "Please avoid using space after opening parantheses"
        regex: "(\(\s+)"        
```

However, I want to avoid Swiftlint from warning me when opening brace is followed by newline. Is there any way to avoid warning by doing something like this?

`excludingPattern: "(\(\n)"`

marcelofabri:
This is not possible right now. Maybe you can do that by being clever with only one regex by now.

marcelofabri:
Let's keep it open to track as an enhancement 💯

norio-nomura:
`\s` includes newline.
Can you try `regex: '(\( +)'` instead?

jayesh15111988:
@norio-nomura Thanks, that worked well. Now SwiftLint will catch space after `(` and before `)`, but newlines will be spared in case developer wants to put long arguments on separate lines.

norio-nomura:
Is this already implemented in #1438?

norio-nomura:
It seems not.

Issue #1353, opened 2017-03-12
Enhancement: Better differentiate uses of "included" and "excluded"


nigelflack:
I'd like to propose an enhancement relating to the "included" and "excluded" configuration keywords.

In the global configuration (i.e. not per-rule configuration) "included" and "excluded" refer to file paths:

```
included: # paths to include during linting. `--path` is ignored if present.
  - Source
excluded: # paths to ignore during linting. Takes precedence over `included`.
  - Carthage
  - Pods
  - Source/ExcludedFolder
  - Source/ExcludedFile.swift
```

In per-rule configuration, "excluded" refers to tokens (I'm not sure if included is supported here):

```
identifier_name:
  min_length: # only min_length
    error: 4 # only error
  excluded: # excluded via string array
    - id
    - URL
    - GlobalAPIKey
```

In custom rules, "included" refers to file paths again, and "excluded" does not appear to be supported. "match_kinds" is used for tokens:

```
custom_rules:
  pirates_beat_ninjas: # rule identifier
    included: ".*.swift" # regex that defines paths to include during linting. optional.
    name: "Pirates Beat Ninjas" # rule name. optional.
    regex: "([n,N]inja)" # matching pattern
    match_kinds: # SyntaxKinds to match. optional.
      - comment
      - identifier
```

I'd propose making these more consistent through updating the names to something like:

include_paths
exclude_paths
include_tokens
exclude_tokens

Any thoughts?

nigelflack:
Any comments from the team on this? Just wondering if I've misinterpreted the settings or whether this is a valid proposal.

Otherwise I might attempt a PR to:

- switch "included" in custom rules to be "include_paths"
- add support for "exclude_paths"

nigelflack:
Just coming back to update this...

I've made a change to add "excluded" support to custom rules, using it for file paths. I don't know how widely custom rules are used - changing the keyword to "include_paths" and "exclude_paths" would be a breaking change to people's config files so I guess it's not ideal. I don't know if there's any mechanism to deprecate config entries?

I still think it would be good to eliminate this overloading of included and excluded though.

Could the keyword for tokens (i.e. "match_kinds") be updated to make that more consistent, and leave included and excluded to always refer to file paths? (if you could migrate people over to a new syntax for token matching with support for both schemes for a while this could be a good approach).

Or should this just be added to a backlog for a 0.2 or 0.5 or 1.0 etc. and accept that it is a breaking change?

PR #1438, opened 2017-04-14
Adds support for excluded in custom rules

nigelflack:
Adds support for excluded in custom rules to specify a regex for files to exclude.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ##### Enhancements
 
+* Adds support for `excluded` in custom rules.  
+  [Nigel Flack](https://github.com/nigelflack)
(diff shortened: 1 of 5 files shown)


marcelofabri:
Maybe we need to rethink the name since this can be confused with https://github.com/realm/SwiftLint/issues/1408. 

I've seen that you actually opened a ticket to discuss the potentially confusing names: https://github.com/realm/SwiftLint/issues/1353

nigelflack:
Yes I'll look at adding some tests. Sorry... first github contribution :-)

And I've updated #1353 with some further thoughts.

marcelofabri:
This looks good to me, but could you add some documentation about this on README?

marcelofabri:
Thanks! 💯

Issue #541, opened 2016-02-16
How to enable only the `max_length` of `variable_name` rule?


dacchan:
Hello. 

In SwiftLint at version v0.6.0 , I want only `variable_name_max_length` rule to be enabled. 
However, I do not know the settings for latest version.

Is there any method to only the `max_length` of `variable_name` rule to be enabled?

`.swiftlint.yml` file.

``` yaml
# at v0.6.0
disabled_rules:
  - force_cast
  - force_try
  - function_body_length
  - line_length
  - nesting
  - todo
  - type_body_length
  - type_name
  - variable_name
  - variable_name_min_length
file_length:
  warning: 500
variable_name_max_length:
  warning: 45
  error: 60
```

norio-nomura:
Could you please use lowest threshold?
```yaml
variable_name:
  max_length:
    warning: 45
    error: 60
  min_length:
    warning: 1
```
You can confirm how configuration is interpreted by `swiftlint`
```sh
swiftlint rules --config ~/.swiftlint-test.yml
Loading configuration from '/Users/norio/.swiftlint-test.yml'
+-----------------------------+--------+-------------+------------------------+--------------------------------------------------+
| identifier                  | opt-in | correctable | enabled in your config | configuration                                    |
+-----------------------------+--------+-------------+------------------------+--------------------------------------------------+
| closing_brace               | no     | yes         | yes                    | warning                                          |
| colon                       | no     | yes         | yes                    | warning                                          |
| comma                       | no     | yes         | yes                    | warning                                          |
| conditional_binding_cascade | no     | no          | yes                    | warning                                          |
| control_statement           | no     | no          | yes                    | warning                                          |
| custom_rules                | no     | no          | yes                    | user-defined                                     |
| cyclomatic_complexity       | no     | no          | yes                    | warning: 10, error: 20                           |
| empty_count                 | yes    | no          | no                     | error                                            |
| file_length                 | no     | no          | yes                    | warning: 400, error: 1000                        |
| force_cast                  | no     | no          | yes                    | error                                            |
| force_try                   | no     | no          | yes                    | error                                            |
| force_unwrapping            | yes    | no          | no                     | warning                                          |
| function_body_length        | no     | no          | yes                    | warning: 40, error: 100                          |
| function_parameter_count    | no     | no          | yes                    | warning: 5, error: 8                             |
| leading_whitespace          | no     | no          | yes                    | warning                                          |
[...]

PR #1444, opened 2017-04-17
Add options to disable some checks for identifier_name rule

jaherhi:
Add options mentioned in https://github.com/realm/SwiftLint/issues/541 for `IdentifierNameRule`, in order to allow names that start with an uppercase character and to exclude non alphanumeric characters.


Tests/SwiftLintFrameworkTests/IdentifierNameRuleTests.swift
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    func testIdentifierNameWithIgnoreStartWithLowercase() {
+        let baseDescription = IdentifierNameRule.description
+        let nonTriggeringExamples = baseDescription.nonTriggeringExamples + [
+            "let MyLet = 0",
+            "enum Foo { case MyEnum }"
+        ]
+        let triggeringExamples = baseDescription.triggeringExamples
+            .filter { !$0.contains("MyLet") && !$0.contains("MyEnum") }
```
jaherhi: Would it make sense to add triggering and non triggering examples to `IdentifierNameRuleExamples` for these two new tests even if they are only used in the tests?

Source/SwiftLintFramework/Rules/IdentifierNameRule.swift
```diff
@@ -70,7 +72,8 @@ public struct IdentifierNameRule: ASTRule, ConfigurationProviderRule {
                 }
             }
 
-            if kind != .varStatic && name.isViolatingCase && !name.isOperator {
+            if (!configuration.ignoresStartWithLowercase || isFunction) &&
```
marcelofabri: maybe extract this condition (inside parenthesis) into a variable?

Tests/SwiftLintFrameworkTests/IdentifierNameRuleTests.swift
marcelofabri: I'm not sure what you mean here, could you elaborate?

jaherhi: I meant adding new static properties in `IdentifierNameRuleExamples.swift` (e.g. `swift3NonTriggeringExamplesWithIgnoreStartWithLowercase` or similar), instead of appending the two examples, `let MyLet = 0` and `enum Foo { case MyEnum }`, to the non triggering examples here, and then having to remove examples that contain `MyLet` and `MyEnum` for the triggering examples.

Source/SwiftLintFramework/Rules/RuleConfigurations/NameConfiguration.swift
```diff
@@ -17,6 +17,8 @@ public struct NameConfiguration: RuleConfiguration, Equatable {
     var minLength: SeverityLevelsConfiguration
     var maxLength: SeverityLevelsConfiguration
     var excluded: Set<String>
+    var allowedSymbols: Set<String>
```
marcelofabri: this configuration is used on `TypeNameRule` and `GenericTypeNameRule` too. Maybe we should add support on those rules as well?

marcelofabri: also, can you update `consoleDescription` to show all variables?

Source/SwiftLintFramework/Rules/RuleConfigurations/NameConfiguration.swift
```diff
@@ -17,6 +17,8 @@ public struct NameConfiguration: RuleConfiguration, Equatable {
...
+    var ignoresStartWithLowercase: Bool
```
marcelofabri: what do you think about `validatesStartWithLowercase` (defaulting to `true`)?

Source/SwiftLintFramework/Rules/IdentifierNameRule.swift
```diff
@@ -70,7 +72,9 @@ public struct IdentifierNameRule: ASTRule, ConfigurationProviderRule {
                 }
             }
 
-            if kind != .varStatic && name.isViolatingCase && !name.isOperator {
+            let requiresCaseCheck = !configuration.ignoresStartWithLowercase || isFunction
```
marcelofabri: why the `isFunction` check? 🤔

jaherhi: It made sense to me not to allow this option for functions, I thought there would be more cases where someone would want to disable this, in order to migrate code with enum cases that start with uppercase for example, than cases where someone would have functions starting with uppercase. Nonetheless, I wouldn't mind changing this if you think it'd be better 🙂

Source/SwiftLintFramework/Rules/RuleConfigurations/NameConfiguration.swift
jaherhi: I think it looks better than my option 😆 I'll change it!

Source/SwiftLintFramework/Rules/RuleConfigurations/NameConfiguration.swift
jaherhi: It wouldn't be difficult to add it to `TypeNameRule`and `GenericTypeNameRule`, but do you think it'll be a common case? I can't remember seeing any code base with types or generic types with symbols

marcelofabri: I was thinking more about keeping them consistente since they all use the same configuration type

jpsim:
Minor changes in 61e9b393

jaherhi:
Thanks for the help and feedback @marcelofabri 🙂

Issue #1432, opened 2017-04-13
False positive in UnusedOptionalBindingRule


mlilback:
This warning is triggered when using the try? syntax for error handling.

```
    enum DummyErrors: Error { case oops }
    func alwaysThrow() throws -> Int {
        throw DummyErrors.oops
    }
    func testSwiftlintFalsePositive() {
        guard let _ = try? alwaysThrow() else { return } // Unused Optional Binding Violation
    }
```

marcelofabri:
I don't think this is necessarily a false positive.

You could do this instead:
```swift
guard try? alwaysThrow() != nil else { return }
```

I don't think anyone thought about `try?` when this rule was created, but I think the same idea applies.

Sega-Zero:
Did you ever tried to compile that code in swift 3? It will never compile:
```swift
func alwaysThrow() throws {
}

guard try? alwaysThrow() != nil else { return }
```
![appdelegate swift 2017-04-22 22-30-50](https://cloud.githubusercontent.com/assets/2203199/25307547/5cf058c6-27ab-11e7-82cb-4cf82d9a6d48.jpg)

marcelofabri:
You need to wrap in parenthesis, I just forgot to do it.

```swift
func alwaysThrow() throws {
}

guard (try? alwaysThrow()) != nil else { return }
```

Sega-Zero:
So, to silence that kind of warning, I better wrap every try? in parenthesis? Yeah, the code will be much cleaner, readable and safe.
I thing that only uglify my code, I better silance that warning rather than using parenthesis.

marcelofabri:
Well, I'd argue that `try?` should in general be avoided so errors are properly treated anyways.

I marked this as a discussion, so more people can jump in and share their thoughts. I'd like to say:

1. You're free to disable this rule in this snippet if it's an occasional thing.
2. Feel free to send a PR to make this behavior customizable, so you can change it on your configuration.
3. Let's avoid ironies and keep the conversation polite, shall we?

Sega-Zero:
Right now I had to disable the rule in the config, 'cause I don't want to put too much of `// swiftlint:disable` in my code. 

`try?` is a 100% normal construct. There are lot of cases when you don't really care if the method throws an error (like in FileManager routines). `throws` doesn't force a programmer to catch all the errors a function may throw, this is an indicator for a programmer, that *something* may have be wrong and it is up to a programmer to process it or ignore it.
[...]

PR #1454, opened 2017-04-22
Added option to ignore optional try in guard statements for unused_optional_binding rule

Sega-Zero:
Added `ignore_optional_try` option to resolve the situation mentioned in #1432.


marcelofabri:
@Sega-Zero This is failing on the `swiftlint with Swift 2.3` scheme. I think you can reproduce on Xcode if you select that scheme.

Sega-Zero:
Thanks, finally figured it out!

marcelofabri:
Merged on #1480

PR #1463, opened 2017-04-25
Fix inaccurate violation report in fileLengthRule and FunctionParameterCountRule

ultimatedbz:
We should be using parameter.value instead of configuration.warning in order to report accurate information about the violation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,9 @@
 * Fix typo in `DiscardedNotificationCenterObserverRule`.  
   [Spencer Kaiser](https://github.com/spencerkaiser)
 
+* Fix inaccurate violation report in fileLengthRule and FunctionParameterCountRule.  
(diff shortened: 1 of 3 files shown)


jpsim:
I don't think this is an improvement, as addressing the violation as reported in the message with `parameter.value` might still produce a violation.

I had used `configuration.warning` on purpose here. @marcelofabri what do you think? If it were up to me, we'd revert this.

ultimatedbz:
I think the change wasn't to not produce a violation. My motivation was to report the violation more accurately. In the case of FileLengthRule.swift, we are comparing `lineCount > parameter.value`. In the case that `lineCount > parameter.value`, would it make more sense to report that the file should contain `parameter.value` # of lines instead of `configuration.warning` # of lines? Otherwise, we should be comparing `lineCount > configuration.warning`?

PR #1465, opened 2017-04-25
Revert #1463

jpsim:
Revert #1463 as suggested in https://github.com/realm/SwiftLint/pull/1463#issuecomment-297106771

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,10 +27,6 @@
 * Fix typo in `DiscardedNotificationCenterObserverRule`.  
   [Spencer Kaiser](https://github.com/spencerkaiser)
 
-* Fix inaccurate violation report in `file_length` and
-  `function_parameter_count` rules.  
-  [ultimatedbz](https://github.com/ultimatedbz)
-
 ## 0.18.1: Misaligned Drum
 
 ##### Breaking
--- a/Source/SwiftLintFramework/Rules/FileLengthRule.swift
+++ b/Source/SwiftLintFramework/Rules/FileLengthRule.swift
@@ -31,7 +31,7 @@ public struct FileLengthRule: ConfigurationProviderRule, SourceKitFreeRule {
             return [StyleViolation(ruleDescription: type(of: self).description,
                 severity: parameter.severity,
                 location: Location(file: file.path, line: lineCount),
-                reason: "File should contain \(parameter.value) lines or less: " +
+                reason: "File should contain \(configuration.warning) lines or less: " +
                         "currently contains \(lineCount)")]
         }
         return []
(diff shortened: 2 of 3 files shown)


ultimatedbz:
(From other PR) I think the change wasn't to not produce a violation. My motivation was to report the violation more accurately. In the case of FileLengthRule.swift, we are comparing lineCount > parameter.value. In the case that lineCount > parameter.value, would it make more sense to report that the file should contain parameter.value # of lines instead of configuration.warning # of lines? Otherwise, we should be comparing lineCount > configuration.warning?

marcelofabri:
Maybe we should use "should/must" for warning/error on message? I feel like #1463 is useful because when adding SwiftLint to a project people could focus first on fixing errors, then warnings.

Does that make sense?

jpsim:
Alright, seems I'm outnumbered on this, which is fine. I wanted to share the motivation explaining why things were done the way they were, since that angle hadn't been discussed in #1463.

I don't think tweaking the "should/must" wording makes a significant difference in usage, so you can do what you think is best there.

marcelofabri:
@jpsim If you feel strongly about this, feel free to reopen/merge it. I'm not strong about any alternatives.

Thanks for raising the issue though, I didn't know it was on purpose.

jpsim:
I think this is a slightly worse user experience, but I appear to be in the minority here, and I don't feel strongly about it if others think otherwise.

My main goal with this and the other comment was in explaining the motivation behind the original implementation.

ultimatedbz:
The following examples should help better explain.

INCORRECT CURRENT OUTPUT:

Warnings
`Function should have 5 parameters or less: it currently has 7`

Errors
`Function should have 5 parameters or less: it currently has 14`

EXPECTED OUTPUT:

Warnings
`Function should have 5 parameters or less: it currently has 7`

Errors
`Function should have 8 parameters or less: it currently has 14`

As you can see in the above example, this PR fixes SwiftLint to correctly output `8` for the error threshold.

And as a reference, you can see this is implemented the same way in these other rules:

https://github.com/ultimatedbz/SwiftLint/blob/67dc8ad50271002038712328773f61f7ae034b0e/Source/SwiftLintFramework/Rules/FunctionBodyLengthRule.swift#L42

https://github.com/ultimatedbz/SwiftLint/blob/67dc8ad50271002038712328773f61f7ae034b0e/Source/SwiftLintFramework/Rules/LargeTupleRule.swift#L69

jpsim:
I personally think that's a terrible user experience.

1. SwiftLint: Function should have 8 parameters or less: it currently has 14
2. User: ok, I've refactored to have 7 parameters, I assume SwiftLint is happy with this now
3. SwiftLint: Function should have 5 parameters or less: it currently has 7
4. User: (╯°□°）╯︵ ┻━┻

ultimatedbz:
@jpsim I thought more about this and I now see your intention in using the warning threshold. Now that I understand your original motivation, I am fine with #1463 being reverted. Might I just suggest that we keep this consistent and make this change across the other rules listed above?

jpsim:
> Might I just suggest that we keep this consistent and make this change across the other rules listed above?

Sure, could you take care of doing this please? @marcelofabri thoughts?

marcelofabri:
I'm fine with any of the approaches

PR #1470, opened 2017-04-26
Added NoExtensionAccessModifierRule

### LINKED ISSUE #1457: Optionally disallow access modifier override on extensions
Instead of
```swift
private extension SomeType {
    func someFuncA() {}
}
```
Prefer
```swift
extension SomeType {
    fileprivate func someFuncA() {}
}
```

masters3d:
Added `no_extension_access_modifier` opt-in rule to disallow access modifiers completely, à la SE-0119.

closes #1457

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,11 @@
   [Samuel Susla](https://github.com/sammy-SC)
   [Jeremy David Giesbrecht](https://github.com/SDGGiesbrecht)
   [#1326](https://github.com/realm/SwiftLint/issues/1326)
-
+  
+* Added `no_extension_access_modifier` opt-in rule to disallow access modifiers completely, à la SE-0119.  
+  [Jose Cheyo Jimenez](https://github.com/masters3d)
+  [#1457](https://github.com/realm/SwiftLint/issues/1457)
+  
 * Add lowercase and missing colon checks to the `mark` rule.  
   [Jason Moore](https://github.com/xinsight)
 
--- a/Source/SwiftLintFramework/Models/MasterRuleList.swift
+++ b/Source/SwiftLintFramework/Models/MasterRuleList.swift
@@ -119,6 +119,7 @@ public let masterRuleList = RuleList(rules:
     MissingDocsRule.self,
     NestingRule.self,
     NimbleOperatorRule.self,
+    NoExtensionAccessModifierRule.self,
     NotificationCenterDetachmentRule.self,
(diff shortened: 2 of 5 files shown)


marcelofabri:
I think this should actually be two separate rules: one for checking for `fileprivate extension` (which would me a default rule) and one for checking what SE-0119 proposed (this would be opt-in).

Also, I think we should use AST for searching extensions.

masters3d:
I thought about two rules but since they are opposite rules I thought it would be better to have it as an option. Otherwise when somebody turns on the optional rule they would have to disable the default rule.

you mean make the rule AST? Does the structure is the AST?

marcelofabri:
> I thought about two rules but since they are opposite rules I thought it would be better to have it as an option. Otherwise when somebody turns on the optional rule they would have to disable the default rule.

I know, but I still think it's better to split in two. Maybe @jpsim has thoughts on this?

> you mean make the rule AST? Does the structure is the AST?

Either making it an `ASTRule` or use `file.structure` instead of `file.match(pattern: "\\s*extension\\s*")`.

masters3d:
What about a generalized top level fileprivate rule with an option to only throw a violation for extensions?

We can then break the SE-119 to its own option rule.

marcelofabri:
> What about a generalized top level fileprivate rule with an option to only throw a violation for extensions?

Actually I wouldn't even provide an option - it should throw for all kinds of types.

masters3d:
Good point. I'm going to steal your algorithm from before. :)

masters3d:
@marcelofabri these have been split up in to two rules as you suggested https://github.com/realm/SwiftLint/pull/1489

marcelofabri:
I've opened [SR-4823](https://bugs.swift.org/browse/SR-4823) because I think we should be able to get `key.accessibility` for `extension`s with SourceKit as we do for type declarations.

masters3d:
Would that block this PR and #1489? We cant get `typealias`es in Sourcekit either. 
https://bugs.swift.org/browse/SR-4828

marcelofabri:
I don't think so, just thought it'd be better if we could get that from SourceKit instead.

masters3d:
Yeah. Absolutely.

marcelofabri:
@masters3d Could you rebase this so (hopefully) we only get the violations introduced by this rule in Danger?

marcelofabri:
Merged in #1538. Thanks for your continued hard work, @masters3d 💯

masters3d:
Thanks!

PR #1471, opened 2017-04-27
Whitespace around let/var groups

Uncommon:
I have this mostly working pretty well, but I'm at a point where I could use a little help.

The goal is to require blank lines around groups of `let`/`var` statements, with comments allowed in those groups as well.

The thing I'm having trouble with is accessors - anything inside an accessor (get/set/didSet) is getting totally ignored. The file structure gives me info about the outer `var` declaration, but then there's no substructure so I'm not getting anything about what's in its accessors. For the `struct` case for example, I do get a substructure, and that works just fine. So how do I find info about statements inside accessors?

Next I need to figure out what to do with `guard` since `guard let` often goes with regular `let`. I figure the options are:
- Always allow `guard` next to let/var groups
- Never allow it
- Allow `guard` if it includes let/var (if that's not too much of a pain to figure out)
- Make it configurable

This also seems like a good candidate for auto-correct. The only thing is that comment lines can make it ambiguous.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,9 @@
   in `guard` statements.  
   [Sega-Zero](https://github.com/Sega-Zero)
   [#1432](https://github.com/realm/SwiftLint/issues/1432)
+* Whitespace around let/var groups.  
+  [Uncommon](https://github.com/Uncommon)
+  [#1471](https://github.com/realm/SwiftLint/issues/1471)
 
 * Add `empty_enum_arguments` correctable rule that warns against using
   silent associated values inside a `case`.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,7 @@
   in `guard` statements.  
   [Sega-Zero](https://github.com/Sega-Zero)
   [#1432](https://github.com/realm/SwiftLint/issues/1432)
+
 * Whitespace around let/var groups.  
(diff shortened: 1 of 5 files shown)


Uncommon:
Rebased and changelog entry added.

Any ideas on how to deal with the accessor situation?

Uncommon:
Since the accessor thing seems to be another manifestation of #136, I'll just punt on it for now. Other than that everything is in order AFAICT.

Uncommon:
So... any reaction to this? Is there something else I should add?

marcelofabri:
@Uncommon is this still WIP? I'm a bit confused about the current status

Uncommon:
Sorry it wasn't clear. I forgot to take the WIP off the title. Since the accessor problem is an issue on its own, I believe this is ready to go.

Uncommon:
Oops, I see cases where `var` followed by `#endif` is triggering, which is undesirable. I'll fix that.

Uncommon:
The `#if`/`#endif` and `@available` cases are fixed, so this is good to go.

Uncommon:
Any other complaints/objections about this PR?

jpsim:
Thanks @Uncommon! Merged in 07b8d13de914972fb689ae8d87fe854a5a3ddd6d to e54dcae5b7e1b41bd3728db4d08a85ba813af04b

PR #1474, opened 2017-04-30
Fix `empty_parameters` rule with Swift 3

### LINKED ISSUE #1460: empty_parameters rule not triggering in Swift 3
The empty_parameters rule does not seem to be updated to Swift 3.

While it triggers a warning to prefer 
`() ->` over `Void ->`, 

it does NOT trigger for 
`(Void) ->`

Since the brackets are required since Swift 3, it never triggers.

marcelofabri:
Fixes #1460


jpsim:
🎉

Jerrot:
Thank you! 👍

PR #1473, opened 2017-04-30
Added ExplicitInternalTopLevelRule

masters3d:
closes https://github.com/realm/SwiftLint/issues/58

Benchmark
```
0.001: file_length
0.001: custom_rules
0.001: leading_whitespace
0.003: trailing_newline
0.018: redundant_discardable_let
0.021: force_cast
0.022: shorthand_operator
0.023: force_try
0.023: closing_brace
0.023: empty_parameters
0.025: return_arrow_whitespace
0.028: trailing_semicolon
0.029: void_return
0.032: syntactic_sugar
0.032: todo
0.033: nimble_operator
0.035: operator_whitespace
0.036: statement_position
0.041: implicit_getter
0.041: legacy_cggeometry_functions
0.041: legacy_nsgeometry_functions
0.041: file_header
0.042: legacy_constructor
0.043: trailing_whitespace
0.044: legacy_constant
0.045: sorted_imports
0.056: first_where
0.059: redundant_nil_coalescing
0.093: empty_count
0.096: explicit_internal_toplevel          <------
0.114: vertical_whitespace
0.117: control_statement
0.170: opening_brace
0.200: number_separator
0.278: comma
0.287: mark
0.288: unused_enumerated
0.292: for_where
0.294: class_delegate_protocol
0.300: operator_usage_whitespace
0.302: weak_delegate
0.306: redundant_string_enum_value
0.313: valid_ibinspectable
0.318: private_outlet
0.319: private_unit_test
0.331: overridden_super_call
0.340: prohibited_super_call
0.348: redundant_optional_initialization
0.357: notification_center_detachment
0.361: dynamic_inline
0.366: fatal_error_message
0.374: discarded_notification_center_observer
0.382: type_body_length
0.386: type_name
0.412: attributes
0.412: explicit_init
0.422: unused_optional_binding
0.449: function_parameter_count
0.454: compiler_protocol_init
0.470: identifier_name
0.514: generic_type_name
0.518: cyclomatic_complexity
0.520: object_literal
0.523: nesting
0.542: vertical_parameter_alignment
0.569: closure_parameter_position
0.571: trailing_comma
0.585: large_tuple
0.598: closure_spacing
0.608: function_body_length
0.616: colon
0.693: redundant_void_return
0.719: line_length
0.725: unused_closure_parameter
0.821: empty_parentheses_with_trailing_closure
0.920: closure_end_indentation

```

PR #1476, opened 2017-04-30
Add explicit_top_level_acl opt-in rule

### LINKED ISSUE #58: Validate explicit ACL modifiers for top-level declarations

marcelofabri:
Continued from #1473

Fixes #58

@masters3d I also took the liberty of renaming the rule so hopefully it's clearer what it enforces (the fact that it only checks for `internal` is an implementation detail).

This will probably have different results from #1473 since it validates all declarations (not only types declarations) - I think this is more aligned with @jpsim's original intention in #58.


jpsim:
How about naming this rule `ExplicitTopLevelACL`? `ExplicitACLInTopLevelDeclaration` is a bit wordy.

marcelofabri:
> How about naming this rule `ExplicitTopLevelACL`? `ExplicitACLInTopLevelDeclaration` is a bit wordy.

Sounds good, will rename and push it.

Source/SwiftLintFramework/Rules/ExplicitACLInTopLevelDeclarationRule.swift
```diff
@@ -0,0 +1,80 @@
...
+
+public struct ExplicitACLInTopLevelDeclarationRule: OptInRule, ConfigurationProviderRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "explicit_acl_in_top_level_declaration",
+        name: "Explicit ACL in Top Level Declaration",
+        description: "Top-level declarations should specify Access Control Labels explicitly.",
```
jpsim: "ACL" stands for "Access Control Level", so perhaps reword to "Top-level declarations should specify Access Control Level keywords explicitly."

jpsim:
🎉

PR #1478, opened 2017-05-01
Add empty_enum_arguments rule

### LINKED ISSUE #1425: Add rule for omitting empty enum with associated value arguments
Currently you can do this:

```swift
enum Foo {
    case bar(String)
}
```

```swift
switch foo {
case .bar(_):
    break
}
```

I think we should add rule to have people remove `(_)`. Note it can also be `(_, _)`.

I've added this regex on our codebase with no false positives: `"\([,\s*_]+\)"`

marcelofabri:
Fixes #1425

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#58](https://github.com/realm/SwiftLint/issues/58)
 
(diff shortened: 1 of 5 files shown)


jpsim:
Awesome! I reviewed all the violations reported by OSSCheck and they look valid. I pushed some small refactors here: bb4c94be

PR #1480, opened 2017-05-01
Unused optional binding rule ignore option

marcelofabri:
Continued from #1454

I've rebased against master and fixed some issues with CHANGELOG and the length of `RulesTests`. Also improved the rule to be a bit safer, checking the type of `try?`.

// cc @Sega-Zero

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,13 @@
   [Sega-Zero](https://github.com/Sega-Zero)
   [#1432](https://github.com/realm/SwiftLint/issues/1432)
   
+  
+* Add lowercase and missing colon checks to the `mark` rule.  
+  [Jason Moore](https://github.com/xinsight)
+  
+* Add opt-in option to `unused_optional_binding` rule to ignore `try?` in `guard` statements  
+  [Sega-Zero](https://github.com/Sega-Zero)
+  [#1432](https://github.com/realm/SwiftLint/issues/1432)
 
 ##### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,15 +40,9 @@
   keyword when it can be omitted inside closures.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1194](https://github.com/realm/SwiftLint/issues/1194)
-* Add opt-in option to `unused_optional_binding` rule to ignore `try?` in `guard` statements  
-  [Sega-Zero](https://github.com/Sega-Zero)
-  [#1432](https://github.com/realm/SwiftLint/issues/1432)
   
-  
-* Add lowercase and missing colon checks to the `mark` rule.  
-  [Jason Moore](https://github.com/xinsight)
-  
-* Add opt-in option to `unused_optional_binding` rule to ignore `try?` in `guard` statements  
+* Add option to `unused_optional_binding` rule to ignore `try?`
+  in `guard` statements.  
   [Sega-Zero](https://github.com/Sega-Zero)
   [#1432](https://github.com/realm/SwiftLint/issues/1432)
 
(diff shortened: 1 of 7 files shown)


Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
```diff
@@ -10,7 +10,7 @@ import Foundation
 import SourceKittenFramework
 
 public struct UnusedOptionalBindingRule: ASTRule, ConfigurationProviderRule {
-    public var configuration = SeverityConfiguration(.warning)
+    public var configuration = UnusedOptionalBindingConfiguration(ignoreOptionalTry: true)
```
jpsim: what's the motivation for enabling this by default?

marcelofabri: Since people are questioning this rule in general, I thought it'd make sense to make it more "permissive" by default but at same time allowing people to keep the old behavior.

CHANGELOG.md
```diff
@@ -40,6 +40,11 @@
   keyword when it can be omitted inside closures.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1194](https://github.com/realm/SwiftLint/issues/1194)
+  
+* Add option to `unused_optional_binding` rule to ignore `try?`
```
jpsim: Since this change causes `try?` to be ignored _by _default, it seems to make sense to reverse the logic in this sentence:

> Add `ignore_optional_try ` option to `unused_optional_binding` rule, `true` by default. Set to `false` to emit a violation when binding the result of `try?` to `_`.

marcelofabri: I'm ok with that. Should I change it or do you prefer to do it?

Tests/SwiftLintFrameworkTests/UnusedOptionalBindingRuleTests.swift
```diff
@@ -0,0 +1,54 @@
...
+//
+
+import Foundation
+@testable import SwiftLintFramework
+import XCTest
+
+class UnusedOptionalBindingRuleTests: XCTestCase {
+
+    func testDefaultConfiguration() {
+        let baseDescription = UnusedOptionalBindingRule.description
```
jpsim: Maybe we could make a lens for `RuleDescription` to simplify this? See http://chris.eidhof.nl/post/lenses-in-swift/

Source/SwiftLintFramework/Rules/UnusedOptionalBindingRule.swift
jpsim: This should be considered a breaking change at the very least.

marcelofabri: I don't have strong opinions here, so I'm ok with reverting the default to `false`

jpsim:
Not sure how I feel about how this is being handled. This changes the semantics of the rule, stops detecting violations that it previously did, has a flag that needs to be set to `false` to change the behavior, which seems like the logic for that should be flipped (i.e. flags should be added to change behavior, not "removed" or "unset").

Tests/SwiftLintFrameworkTests/UnusedOptionalBindingRuleTests.swift
Sega-Zero: https://github.com/narfdotpl/lenso

marcelofabri: I think that might make sense, since this is becoming a common pattern

marcelofabri:
> Not sure how I feel about how this is being handled. This changes the semantics of the rule, stops detecting violations that it previously did, has a flag that needs to be set to false to change the behavior, which seems like the logic for that should be flipped (i.e. flags should be added to change behavior, not "removed" or "unset").

I had this dilemma as well when reviewing this, but decided to change the default because it seems that it's not a common pattern to use `let _ = try? foo()` (at least from our oss-check) and I'd say that the benefits for readability are debatable, so we should provide good defaults. But we need to weight existing users as well as you said, as well as the semantics of the flag. 

I'm cool with any of the options, since the important here IMO is to please both styles.

PR #1482, opened 2017-05-01
Prevent triggering `redundant_optional_initialization` rule  on a `lazy` var

### LINKED ISSUE #1334: Improper autocorrect for optional lazy vars.
I found what is definitely an edge case, but still a valid case of autocorrect producing source that won't compile.

Running autocorrect on the following file
```
class TestClass {
    private lazy var test: Int? = nil
}
```
will produce the output
```
class TestClass {
    private lazy var test: Int?
}
```
Xcode will then let you know `lazy var`s require an initializer.

If the code is
```
class TestClass {
    private lazy var test: Int? = 0
}
```
Everything is fine. It only happens with `nil`. 

Obviously, having a `lazy var` initialize to `nil` makes no sense. No clue why it was written like that. 

I could also see this being a reasonable warning to add. `lazy var x: Any? = nil` is probably a smell.

marcelofabri:
Fixes #1334

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1460](https://github.com/realm/SwiftLint/issues/1460)
 
+* Prevent triggering `redundant_optional_initialization` rule
+  on a `lazy var` since it needs initialization.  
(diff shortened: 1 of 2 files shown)


jpsim:
🎉  thanks for your continued contributions @marcelofabri 👏

PR #1484, opened 2017-05-02
Vertical parameter alignment on call rule

### LINKED ISSUE #1037: Rule Request: Vertical Parameter Alignment in Function Calls
Migrated from #1033:

When refactoring, it's often very easy to mess up the vertical alignment of parameters in function calls:

```swift
validateFunction(file, kind: .function,
                    dictionary: [:])
```

We could (easily?) have a correctable rule for this.

marcelofabri:
Fixes #1037

TODO:
- [x] Add proper examples/tests
- [x] Add CHANGELOG entry
- [x] Think what to do with `MasterRuleList.swift` ~(probably that shouldn't trigger)~
- [x] Check number of violations in oss-check to see if this should be opt-in or default rule (my bet is that it'll trigger LOTS of violations)

### DIFF
--- a/.sourcery/MasterRuleList.stencil
+++ b/.sourcery/MasterRuleList.stencil
@@ -6,6 +6,6 @@
 //  Copyright © 2015 Realm. All rights reserved.
 //
 
-public let masterRuleList = RuleList(rules:
+public let masterRuleList = RuleList(rules: [
 {% for rule in types.structs where rule.name|hasSuffix:"Rule" or rule.name|hasSuffix:"Rules" %}    {{ rule.name }}.self{% if not forloop.last %},{% endif %}
-{% endfor %})
+{% endfor %}])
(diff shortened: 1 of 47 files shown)


marcelofabri:
Definitely opt-in 😂

marcelofabri:
@jpsim I'd love to get this in before cutting a pre-release 😊

marcelofabri:
Giphy

codytwinton:
@marcelofabri - For some reason, when I run:

```bash
swiftlint rules vertical_parameter_alignment_on_call
```

I get this response:

```bash
No rule with identifier: vertical_parameter_alignment_on_call
```

jpsim:
@codytwinton what version of SwiftLint are you running? This rule was first introduced in [0.20.0](https://github.com/realm/SwiftLint/releases/tag/0.20.0).

codytwinton:
I'm using version `0.20.1`

PR #1485, opened 2017-05-02
Enable implicit_return in SwiftLint

marcelofabri:
I'm still not 100% sure about this in some cases. Thoughts, @jpsim?

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -21,6 +21,7 @@ opt_in_rules:
   - number_separator
   - prohibited_super_call
   - fatal_error_message
+  - implicit_return
 
 file_header:
   required_pattern: |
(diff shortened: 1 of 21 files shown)


jpsim:
I don't have a strong preference stylistically here, although I do prefer standardizing one way or the other. I support doing this for now, until I can formally codify my personal preference and write a rule for it.

---

I think something that'd match my closure style preference is loosely:

> Prefer shorthand identifiers (`$0`, etc.) and implicit returns when the closure contents is shorter than 40 characters, after trimming leading+trailing whitespace.

The opposite would also be true:

> Avoid shorthand identifiers and implicit returns when the closure contents is longer than 40 characters, after trimming leading+trailing whitespace.

It's a bit tricky to enforce this exact rule because replacing identifiers & removing/adding explicit return statements changes the length of the closure body's contents.

marcelofabri:
I'm not sure I agree. I was thinking about this and for me implicit returns and shorthand identifiers don't necessarily are related.

For implicit returns, the rule I'd go with is:

> Always use implicit returns when possible, unless you're calling a method with closures.

PR #1489, opened 2017-05-03
Added fileprivateRule

### LINKED ISSUE #1058: Rule request: private is preferred over fileprivate
Because of Swift 3 migration, lots of declarations wouldn't need to be `fileprivate`.

### LINKED ISSUE #1469: [Rule Request] Make all top level fileprivate decl become private
Optional rule to make all the top level `fileprivate` declarations be `private` in order to enforce SE-0025's goal of making `fileprivate` more rare.  

https://github.com/realm/SwiftLint/issues/58

masters3d:
Added `fileprivateRule` to check for top-level usages of `fileprivate` and recommend `private` instead. This is inline with SE-0169's goal "for `fileprivate` to be used rarely". There is a also an "strict" option that will mark every `fileprivate` as a violation.  
closes #1469

benchmark
```
0.001: custom_rules
0.001: file_length
0.001: leading_whitespace
0.003: trailing_newline
0.021: redundant_discardable_let
0.022: force_try
0.023: todo
0.024: shorthand_operator
0.026: return_arrow_whitespace
0.027: syntactic_sugar
0.027: force_cast
0.032: closing_brace
0.033: void_return
0.035: empty_parameters
0.035: legacy_constructor
0.035: statement_position
0.036: file_header
0.039: nimble_operator
0.042: trailing_semicolon
0.043: trailing_whitespace
0.044: legacy_cggeometry_functions
0.045: implicit_getter
0.046: operator_whitespace
0.051: legacy_nsgeometry_functions
0.054: sorted_imports
0.056: legacy_constant
0.057: fileprivate           <------------------------------------------------
0.060: first_where
0.064: redundant_nil_coalescing
0.077: empty_count
0.097: control_statement
0.115: vertical_whitespace
0.170: opening_brace
0.197: number_separator
0.279: for_where
0.283: unused_enumerated
0.299: mark
0.301: operator_usage_whitespace
0.305: class_delegate_protocol
0.307: comma
0.309: redundant_string_enum_value
0.314: valid_ibinspectable
0.317: private_unit_test
0.319: weak_delegate
0.320: private_outlet
0.329: overridden_super_call
0.330: prohibited_super_call
0.352: dynamic_inline
0.353: notification_center_detachment
0.365: redundant_optional_initialization
0.380: discarded_notification_center_observer
0.382: fatal_error_message
0.398: type_name
0.405: type_body_length
0.426: attributes
0.428: explicit_init
0.458: compiler_protocol_init
0.462: function_parameter_count
0.499: identifier_name
0.514: cyclomatic_complexity
0.516: unused_optional_binding
0.517: vertical_parameter_alignment
0.517: nesting
0.533: object_literal
0.546: function_body_length
0.554: generic_type_name
0.586: closure_parameter_position
0.593: trailing_comma
0.599: large_tuple
0.634: colon
0.683: redundant_void_return
0.722: unused_closure_parameter
0.738: closure_spacing
0.752: line_length
0.804: empty_parentheses_with_trailing_closure
0.932: closure_end_indentation
```

### DIFF
--- a/Source/SwiftLintFramework/Rules/FileprivateRule.swift
+++ b/Source/SwiftLintFramework/Rules/FileprivateRule.swift
@@ -1,0 +1,53 @@
+//
+//  FileprivateRule.swift
+//  SwiftLint
+//
+//  Created by Jose Cheyo Jimenez on 05/02/17.
+//  Copyright © 2017 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+public struct FileprivateRule: Rule, ConfigurationProviderRule {
+    public var configuration = FileprivateConfiguration(strict: false)
+
+    public init() {}
+
+    public static let description = FileprivateConfiguration.fileprivateLimited
+
+    public func validate(file: File) -> [StyleViolation] {
+        if !configuration.strict {
+            let toplevel = file.structure.dictionary.substructure.flatMap({ $0.offset })
+            let syntaxTokens = file.syntaxMap.tokens
+            let violationOffsets = toplevel.flatMap { (offSet) -> Int? in
+                let parts = syntaxTokens.partitioned { offSet <= $0.offset }
+                guard let lastKind = parts.first.last,
+                    lastKind.type == SyntaxKind.attributeBuiltin.rawValue,
+                    // Cut the amount of name-look-ups by first checking the char count
+                    lastKind.length == "fileprivate".bridge().length,
+                    // Get the actual name of the attibute
+                    let aclName = file.contents.bridge()
+                            .substringWithByteRange(start:lastKind.offset, length: lastKind.length),
+                    // fileprivate(set) is not possible at toplevel
+                    aclName == "fileprivate"
+                    else { return nil }
+                return offSet
+            }
+            return violationOffsets.map({ StyleViolation(
+                ruleDescription: FileprivateConfiguration.fileprivateLimited,
+                location: Location(file: file, byteOffset: $0))
+            })
+
+        } else { // Mark all fileprivate occurences as a violation
+            let fileprivates = file.match(pattern: "fileprivate", with: [.attributeBuiltin]).flatMap({
+                file.contents.bridge().NSRangeToByteRange(start: $0.location, length: $0.length)
+            }).map({ $0.location })
+            return fileprivates.map({ StyleViolation(
+                ruleDescription: FileprivateConfiguration.fileprivateDisallowed,
+                location: Location(file: file, byteOffset: $0))
+            })
+        }
+
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@
 * Match `(Void)` as return type in the `void_return` rule.  
   [Anders Hasselqvist](https://github.com/nevil)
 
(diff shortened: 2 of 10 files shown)


Source/SwiftLintFramework/Rules/FileprivateRule.swift
```diff
@@ -0,0 +1,47 @@
...
+    public static let description = FileprivateConfiguration.fileprivateLimited
+
+    public func validate(file: File) -> [StyleViolation] {
+        if !configuration.strict {
+            let toplevel = file.structure.dictionary.substructure.flatMap({ $0.offset })
+            let syntaxTokens = file.syntaxMap.tokens
+            let violationOffsets = toplevel.flatMap { (offSet) -> Int? in
+                let parts = syntaxTokens.partitioned { offSet <= $0.offset }
+                guard let lastKind = parts.first.last,
+                    lastKind.length == "fileprivate".bridge().length,
```
masters3d: Full disclosure here: There is another builtin that has the same length as `fileprivate`. `nonmutating` has the same length of characters but `nonmutating` can never be toplevel in the current version of swift AFAIK. We could just `file.substring` to find out exactly what the builtin key is but that would slow down this pass.

marcelofabri:
This doesn't close #1058 because my intention there is broader. For example, this should trigger:

```swift
class Foo {
    fileprivate var bar = ""

    func something() {
        print(bar)
    }
}
```

masters3d:
@marcelofabri That case is triggered in the `strict` mode.  I just didn't think that part should be enforced as default because that could be a valid use case for some people.

marcelofabri:
> @marcelofabri That case is triggered in the strict mode. I just didn't think that part should be enforced as default because that could be a valid use case for some people.

My idea was to trigger only if we find out that the variable is only used in the current type scope (i.e. changing to `private` would still compile). See the following examples. I'd expect a violation on `Example1`, but not on `Example2` - with Swift 3. When we add support for Swift 4, I'd expect a violation on both cases (because of SE-0169). 

Example1:
```swift
class Foo {
    fileprivate var bar = ""

    func something() {
        print(bar)
    }
}
```

Example2:
```swift
class Foo {
    fileprivate var bar = ""
}

extension Foo {
    func printBar() {
        print(bar)
    }
}
```

I know that there's the strict mode, but that comes with false positives as well. 

In a perfect scenario, I'd like this rule to always flag cases where `fileprivate` can be replaced with `private` without any doubts (i.e. can be autocorrectable). Then, with `strict` mode, all usages would be flagged, which would be helpful in Swift 4 as it'd mean "you're using `fileprivate` here, are you sure your abstraction model is right?".

I still think this is useful without implementing the full #1058 issue tough. 

On a related note, I've opened [SR-4823](https://bugs.swift.org/browse/SR-4823) because I think we should be able to get `key.accessibility` for `extension`s with SourceKit as we do for type declarations.

masters3d:
I see. I was thinking `strict` would be for people who do not want their code base to contain `fileprivate` at all.   My thinking is, if people turn on the strict mode and still need `fileprivate`, they can disable it inline for the instances they want. I can remove the strict part for now if though since it is just a glorified regex.

masters3d:
@marcelofabri I rebased this. Let me know if there is anything that would be blocking this.

PR #1494, opened 2017-05-04
Merge master before running OSSCheck on branch

jpsim:
To avoid the branch build being so out of date that OSSCheck reports unrelated differences, which is what happened https://github.com/realm/SwiftLint/pull/1489#issuecomment-298983845 for example.

### DIFF
--- a/script/oss-check
+++ b/script/oss-check
@@ -145,7 +145,11 @@ def generate_reports(branch)
 end
 
 def build(branch)
-  `git fetch && git checkout origin/master` if branch == 'master'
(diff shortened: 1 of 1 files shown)


marcelofabri:
Maybe this had unexpected consequences? I've rebased #1484 and #1478 and none of them are showing violations any more 😬

jpsim:
Yeah, that's weird. I'll revert and if irrelevant regressions/fixes show up in the future, we'll just ask PR authors to rebase until we can address in a way that doesn't break this functionality 😬.

Issue #387, opened 2016-01-19
Enforce consistent attribute ordering


jpsim:
From @AliSoftware:

> Like enforce to consistently use `@IBOutlet private weak var` and not `weak private @IBOutlet var` for example.

Since these are always language keywords, it should be super easy to write a rule that does this. The hard part would be settling on an ordering convention.

AliSoftware:
Du you have an exhaustive list of possible keywords (possibly extracted from SourceKitten) so we could discuss the ordering?

I see `@IBOutlet`/`@IBInspectable` `weak`/`unowned`, `lazy`,`public`/`private`/`internal`,`@objc`… probably more?

jpsim:
I'm sure there are many more. Maybe there's an exhaustive list over at [/apple/swift](/apple/swift)?

ikesyo:
- Modifires: https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/Declarations.html#//apple_ref/swift/grammar/declaration-modifiers
- Attributes: https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/Attributes.html#//apple_ref/doc/uid/TP40014097-CH35-ID347

jpsim:
Thanks!

masters3d:
As far as I can tell, attributes are enforced by the compiler to be declared first. 
These will not compile.
```swift
final public @objc class MyCoolClass: NSObject {} // expected declaration
final @objc public class MyCoolClass: NSObject {} // expected declaration
public @objc final class SomethingCool:NSObject {} // expected declaration
```
This only leaves us with Access Control Labels and other Modifiers to order:
I have a draft implementation for this:
`<@Attibutes> <ACL> <OtherModifiers>`

masters3d:
I've hit a little snag here. All of these modifiers are `source.lang.swift.syntaxtype.attribute.builtin` except:
`class` and `final`. 

`final` is actually`source.lang.swift.syntaxtype.keyword`

```
GRAMMAR OF A DECLARATION MODIFIER

declaration-modifier → class­  convenience­  dynamic­  final­  infix­  lazy­ optional­  override­  postfix­  prefix­  required­  static­  unowned­  unowned­(­safe­)­  unowned­(­unsafe­)­  weak­
declaration-modifier → access-level-modifier­
declaration-modifier → mutation-modifier­
[...]

Issue #27, opened 2015-05-22
Fix it?


soffes:
Not sure if you can provide fix its, but that would be super cool. For some rules (i.e. whitespace only on a line) it would be pretty easy to figure out what to change. Obviously for stuff like `as!` you shouldn't suggest one.

aarondaub:
Related: #5

soffes:
Closing dupe. Sorry.

aarondaub:
@soffes  I actually don't think this is necessarily a dupe as it's a subset. It might make sense to implement this and not #5.

jpsim:
You're right, not a dupe! Do you happen to know if Xcode exposes an API to provide fix-its?

So far, we've been able to avoid making an Xcode plugin and I'd love to continue that in the future :smile:

soffes:
Not sure. Hopefully there's a way without it :)

segiddins:
Pretty sure the fix-it's come from the errors/warnings themselves (I.e. The text)

jpsim:
@segiddins do you know what the format should be to suggest a fix-it to Xcode?

julien-c:
It would be very useful even if it's just a cli command for now: `swiftlint fix`

jpsim:
@julien-c I think you're confusing "Xcode Fix-Its" (this issue), with #5 (the ability to automatically correct violations).

jpsim:
More info on Xcode Fix-Its can be found in Apple's documentation: https://developer.apple.com/library/ios/recipes/xcode_help-source_editor/chapters/Fix-it.html

segiddins:
According to people on the dev tools team, there's no public way to add new fix-its

jpsim:
Then the private way it will be!

masters3d:
@segiddins what is the private way to add fix-it? 

The new xcode 8 extensions are interesting but probably not applicable here. 
[Using and Extending the Xcode Source Editor ](https://developer.apple.com/videos/play/wwdc2016/414/?time=1200) at min20 they start talking about extensions

jpsim:
> what is the private way to add fix-it?

Xcode Plugins, and no, Xcode Source Extensions don't expose a way to do that, so this would have to be done in an [Alcatraz](http://alcatraz.io/)-style Xcode plugin.

SDGGiesbrecht:
I’ve tried to get this to work in my own project to no avail.

[...]

PR #1501, opened 2017-05-08
Added OptIn ModifierOrderRule

### LINKED ISSUE #1472: [rule request] enforce the order of `final` along with ACL of classes
optional rule to enforce order of final of classes in a code base

`internal final class MyClass {}`
or
`final public class MyClass {}`

This could probably be done by using a regex to check both possibilities and suggest one or the other depending on preferences. A regex could also catch multiline declarations

```swift
import Foundation

public
final class MyClass{}
```

`<ws*>public|internal<ws*>final<ws*>class<ws*>`
`<ws*>final<ws*>public|internal<ws*>class<ws*>`

This rule doesn't seek to introduce final, only to make the declaration consistent for readability
This is different than  https://github.com/realm/SwiftLint/issues/228

masters3d:
closes #1472
possibly addresses #387

```
0.001: custom_rules
0.001: file_length
0.001: leading_whitespace
0.003: trailing_newline
0.019: redundant_discardable_let
0.020: closing_brace
0.022: shorthand_operator
0.024: force_try
0.024: todo
0.026: empty_parameters
0.028: trailing_semicolon
0.030: void_return
0.031: return_arrow_whitespace
0.031: force_cast
0.031: statement_position
0.032: syntactic_sugar
0.033: nimble_operator
0.034: operator_whitespace
0.040: implicit_getter
0.042: legacy_constructor
0.043: trailing_whitespace
0.045: first_where
0.045: legacy_cggeometry_functions
0.045: legacy_constant
0.047: file_header
0.051: sorted_imports
0.052: legacy_nsgeometry_functions
0.063: redundant_nil_coalescing
0.082: empty_count
0.105: control_statement
0.120: vertical_whitespace
0.153: opening_brace
0.186: number_separator
0.262: for_where
0.266: modifiers_order          <<<------
0.266: comma
0.273: mark
0.274: unused_enumerated
0.280: valid_ibinspectable
0.294: redundant_string_enum_value
0.296: overridden_super_call
0.297: private_outlet
0.298: weak_delegate
0.299: prohibited_super_call
0.304: class_delegate_protocol
0.306: operator_usage_whitespace
0.308: private_unit_test
0.340: dynamic_inline
0.342: fatal_error_message
0.347: redundant_optional_initialization
0.351: notification_center_detachment
0.355: discarded_notification_center_observer
0.364: type_body_length
0.377: explicit_init
0.386: type_name
0.388: attributes
0.400: unused_optional_binding
0.424: identifier_name
0.431: function_parameter_count
0.444: compiler_protocol_init
0.487: cyclomatic_complexity
0.497: nesting
0.497: vertical_parameter_alignment
0.499: generic_type_name
0.501: object_literal
0.530: closure_parameter_position
0.547: function_body_length
0.555: trailing_comma
0.579: colon
0.580: large_tuple
0.658: closure_spacing
0.669: unused_closure_parameter
0.675: redundant_void_return
0.777: line_length
0.788: empty_parentheses_with_trailing_closure
0.869: closure_end_indentation

```


masters3d:
What do you think? @marcelofabri

marcelofabri:
I haven't got the time to review this properly yet. Sorry about the delay 😬

masters3d:
No problem. Thanks for all the work you already do.

marcelofabri:
Sorry for taking so long, but I think the current implementation is a bit fragile. 

I'd be more comfortable tackling this after https://github.com/apple/swift/pull/12086 and https://github.com/apple/swift/pull/11750 land (hopefully Swift 4.1). 

We also would need to discuss what's the default expected order. We'd probably create modifiers group, for example:
* ACLs
* ACLs for setter
* `final`
* `mutating`/`nonmutating`
* `static`/`class`
...

And somehow it'd be good to customize this order.

Do you have any thoughts on this, @masters3d?

masters3d:
Nice PRs. That would really help this implementation. Aside from those the only other tricky part is static and class because they are keywords. I could never get a straight response as of to why static was a keyword but I think is because in a previous swift there was only ‘type’ to mean both static and a class variable until they got split into the two. Part of the reason why we can’t have a static class variable maybe?

I don’t have a preference in the order. Feel free to update.

Buju77:
any update on this PR? we would like to have the access modifiers always on the first place in our code ;)

eg.`public override func visitWakanda() { ... }`

masters3d:
It looks like we want to make this a swift 4.1+ only feature. That is assuming the above mentioned PRs made it into 4.1, if not then we would have to wait untill swift 5.  If we can live with a fragile implementation  then I can dust this PR sooner. @marcelofabri thoughts?

masters3d:
We could rebase on top of https://github.com/realm/SwiftLint/pull/2038 , This rule would then be in the same boat as: https://github.com/realm/SwiftLint/pull/2036

> "we have to make a decision on how we're going to mark rules that won't work on older Swift compiler versions, as Swift 4.1 includes some SourceKit fixes and improvements that will provide us more information to create rules."

dirtydanee:
Waiting for https://github.com/realm/SwiftLint/pull/2036 to be at least reviewed / ideally merged.

Source/SwiftLintFramework/Rules/ModifiersOrderRule.swift
```diff
@@ -19,6 +19,7 @@ public struct ModifiersOrderRule: ASTRule, OptInRule, ConfigurationProviderRule
         name: "Modifiers Order",
         description: "Modifiers order should be consistent.",
         kind: .style,
+        minSwiftVersion: .fourDotOne ,
```
masters3d: nice!

dirtydanee:
Ready for review.

jpsim:
Can this be called “Modifier Order” basically dropping the plural “s” everywhere?

jpsim:
Where does the term “modifier” come from? As best I can tell, these are referred to as “declaration attributes” in Swift. Perhaps this rule should be called “Attribute Order”?

Sorry for just bikeshedding on the name. I’ve only looked at this on my phone so I can’t easily post my review on the actual code yet.

dirtydanee:
Every feedback is great, and naming is important for sure!

The naming has been kept from the first version of the implementation, maybe @master3d could give us some insights on the decission.

I more than happy to change it to ‘Attribute Order’, sounds more accuarate anyway.

masters3d:
At first I wanted to call it atribute and modifier order because attributes are already enforced by the compiler to come first (most people put them on its own line.

So I changed the name to modifierOrder because there are more than one kind modifer (type, declaration, ACL etc).

With the new revision it will be possible to enforce attributes order (as long as its permitted by the compiler)

I don’t know of a name for both concepts. 


Attributes starts with @

https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Attributes.html


There are declarion modifiers, ACL modifiers, mutating modifiers 

https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Declarations.html#//apple_ref/swift/grammar/declaration-modifiers

jpsim:
Thanks for sharing the thought process @masters3d. Let’s keep the name as is.

Source/SwiftLintFramework/Rules/ModifierOrderRule.swift
```diff
@@ -0,0 +1,251 @@
...
+
+    public func validate(file: File,
+                         kind: SwiftDeclarationKind,
+                         dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+
+        guard let offset = dictionary.offset else {
+            return []
+        }
+
+        let preferedOrderOfModifiers = [.atPrefixed] + configuration.preferedModifierOrder
```
jpsim: Why isn't `atPrefixed` simply part of the `preferedModifierOrder` configuration? Seems odd to do this unconditionally.

Source/SwiftLintFramework/Extensions/SwiftDeclarationAttributeKind+Swiftlint.swift
```diff
@@ -0,0 +1,84 @@
...
+                    .nonmutating
+                ]
+            case .override:
+                return [.override]
+            case .owned:
+                return [.weak]
+            case .final:
+                return [.final]
+            case .typeMethods:
+                return []
```
jpsim: Why is this empty?

dirtydanee: `static` and `class` keywords are not being recognised as declaration attribute, only as a declaration. I have [filed a bug](https://bugs.swift.org/browse/SR-7252), no response so far.

Because somehow these keywords needs to be grouped as well as an attribute, i have decided to declare it as a group, but leave it empty.

Source/SwiftLintFramework/Rules/ModifierOrderRule.swift
dirtydanee: All the attributes / modifiers starting with `@` are enforced by the compiler to be placed first. If the `atPrefixed` group would be in any other place than the first, it would mean all declarations using any modifier with the `@` prefix would fail.

Source/SwiftLintFramework/Extensions/SwiftDeclarationAttributeKind+Swiftlint.swift
masters3d: I've been trying to reach out for the last year about this issue. Is not trivial as far as I can tell. The problem is that swift does treats these as keywords because of the split from the previous keyword called 'type'. Changing this only on the SourceKit representation would be problematic. I don't think there is a way around `class` because of its multiple usages. I do think static should be changed but this requires a change to the swift parser which I do not think anybody would think is worth the effort. https://forums.swift.org/t/should-static-be-a-builtin-and-not-keyword/11162

dirtydanee:
@jpsim Regarding forward compatibility -> Only the currently defined modifier groups are being looked at, and evaluated. If a new modifier would be introduced, `findModifierGroups` would fail to recognise it. A new modifier would mean the need to release a new version of SourceKitten and SwiftLint with an updated list of `SwiftDeclarationAttributeKind` and `SwiftDeclarationAttributeKind.ModifierGroup`.

I wanted to build something more flexible, but somehow i need to be able to match the user defined groups with the ones found in Swift, and could not think of a better way than hard coding the groups 😿

samrayner:
I'd love to enable this but I don't think I can until the warning suggests the correct order. Would be it possible to include an example (or, ideally, the specific fix) in the message?

dirtydanee:
@samrayner Upon creating the the `RuleDescription` `static` variable of any rule, the description variable is declared as a `let` constant. Because the configuration of a rule can be changed after the initialisation of the rule, the description will not be updated. I could of course hard code the default value of the configuration, but in case of an overwrite of the configuration, it would be indicating the wrong order.

ATM SwiftLint does not support overriding the description variable of a rule's description. However, i think it could be a nice addition to the framework. @jpsim what do you think about? Should i dig in deeper to see what needs to be done to support this?

marcelofabri:
@dirtydanee you can provide a reason when creating a StyleViolation

dirtydanee:
@marcelofabri true that! I will create a PR later on today.

SDGGiesbrecht:
@samrayner,

> Presumably we have no hook for providing FixIts along with the Xcode warnings?

That is much desired by many, but we have not found a way yet. There are plenty of details in #27. (Which tells you how long it has been awaited: number 27 out of about 2268.)

samrayner:
Thank you @dirtydanee! This rule rules 😍

PR #1506, opened 2017-05-10
update README with cocoapod config file example

aerickson14:
Added further explanation in the `README` outlining how to setup a base `.swiftlint.yml` file by passing its location in using the `--config` option as part of the build script command.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,10 @@
   in `guard` statements.  
   [Sega-Zero](https://github.com/Sega-Zero)
   [#1432](https://github.com/realm/SwiftLint/issues/1432)
+  
+* Improve documentation for using a specific `.swiftlint.yml` config file 
+  with cocoapods.  
+  [Andrew Erickson](https://github.com/aerickson14)
 
 ##### Bug Fixes
 
--- a/README.md
+++ b/README.md
@@ -40,6 +40,10 @@ in your Script Build Phases.
 This is the recommended way to install a specific version of SwiftLint since it supports
 installing a pinned version rather than simply the latest (which is the case with Homebrew).
 
+The default set of rules will be used to start with, and overwritten by any nested configuration files. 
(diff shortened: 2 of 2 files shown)


marcelofabri:
I'm not sure how this is only related to CocoaPods: you can pass `--config` to `swiftlint lint` no matter what distribution form you're using. 

Can you elaborate the motivation for this change?

aerickson14:
Yep that's a fair point. I had found after installing the SwiftLint via CocoaPods and creating a base configuration in my projects root directory, I didnt realize that the config I had created wasnt being used. This is especially true for Cocoapods since the swiftlint executable gets installed in the `Pods/SwiftLint` directory. I would imagine the majority of SwiftLint installs via CocoaPods will want to be using the `--config` command in their build script.

marcelofabri:
I don't think this is needed, since where the binary is located doesn't matter, what matters is where you run it from.

So running `Pods/SwiftLint/swiftlint` or `swiftlint` (from homebrew) on the project folder should give the same results.

aerickson14:
That's the point though. Because when set up with CocoaPods swiftlint is getting run from `Pods/SwiftLint` which is not somewhere you would store your a project's `.swiftlint.yml` file.

Any time a project sets up SwiftLint as a pod, they will likely want to also provide the path to the `--config` file (which to me wasn't immediately obvious). I was expecting the `.swiftlint.yml` file in the SOURCE_ROOT to be picked up, but it wont unless `--config` is specified.

screenshot 2017-05-21 16 28 29

marcelofabri:
aerickson14 what you described shouldn't happen. I've been using SwiftLint via CocoaPods successfully without passing `--config` in the build phase. If that's not working for you, I suggest you try to create a sample project showing the issue.

jpsim:
I agree that this behavior shouldn't be happening.  If anyone provides steps to reproduce this behavior I'll be happy to consider a fix. Until then, we should close this.

Issue #1310, opened 2017-02-15
Rule request: Line comment formatting rule


joshfriend:
I'd like to have a lint rule which validates the format of line and linline comments similar to [Python's PEP8 inline comment rule](https://www.python.org/dev/peps/pep-0008/#inline-comments). 

### Inline comments
Correct:
```swift
statement()  // some comment
```

Incorrect:
```swift
statement()//some comment
statement() //some comment
statement() // some comment
```

### Single line comments
This rule would be similar to the `MARK` rule added in #768, but would apply to _all_ line comments.

Correct:
```swift
// some comment
statement()
```

Incorrect:
```swift
//some comment
statement()
//  some comment
statement()
```

### Example configuration
This would be my proposal for the default values based on the Python standard and my personal preferences:

```yaml
inline_comment:
  spaces_before: 2
  spaces_after: 1
line_comment:
  spaces_after: 1
```

I think `1` would also be an acceptable default for `spaces_before`.

kimdv:
@joshfriend , I'm currently working on this, and hope I can get a version ready this weekend :)

Uncommon:
Comments sometimes are indented, so I think there should be a way to treat `spaces_after` as a minimum.

joshfriend:
Thats a good point, commented out sections of code would certainly cause multiple violations of this rule. I was mostly hoping to make the spacing before and after `//` consistent throughout code and using a `spaces_after` setting as a minimum would probably not interfere with that goal

maximkrouk:
Maybe it's not a right place to ask(pls redirect me if there is a better place), but can I ban multiline comments using swiftlint? 🙂

PR #1509, opened 2017-05-13
[WIP] Inline comment formatting rule

kimdv:
THIS IS STILL IN PROGRESS!
Fell free to review and make improvements ! 👍 

This is for issue #1310

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -622,6 +622,11 @@
 * Fix non-root configurations logging configuration warnings more than once.  
   [JP Simard](https://github.com/jpsim)
   [#949](https://github.com/realm/SwiftLint/issues/949)
+  
+* Add `inline_comment` rule that validates inline comments.  
+  [Kim de Vos](https://github.com/kimdv)
+  [#1310](https://github.com/realm/SwiftLint/issues/1310)
+
 
 * Fix some overlapping `// swiftlint` commands not being applied.  
   [JP Simard](https://github.com/jpsim)
--- a/Source/SwiftLintFramework/Models/MasterRuleList.swift
+++ b/Source/SwiftLintFramework/Models/MasterRuleList.swift
@@ -55,6 +55,7 @@ public let masterRuleList = RuleList(rules: [
     ImplicitlyUnwrappedOptionalRule.self,
     IsDisjointRule.self,
     JoinedDefaultParameterRule.self,
+    InlineCommentRule.self,
     LargeTupleRule.self,
     LeadingWhitespaceRule.self,
     LegacyCGGeometryFunctionsRule.self,
--- a/SwiftLint.xcodeproj/project.pbxproj
+++ b/SwiftLint.xcodeproj/project.pbxproj
@@ -124,6 +124,7 @@
 		BFF028AE1CBCF8A500B38A9D /* TrailingWhitespaceConfiguration.swift in Sources */ = {isa = PBXBuildFile; fileRef = BF48D2D61CBCCA5F0080BDAE /* TrailingWhitespaceConfiguration.swift */; };
 		C328A2F71E6759AE00A9E4D7 /* ExplicitTypeInterfaceRule.swift in Sources */ = {isa = PBXBuildFile; fileRef = C328A2F51E67595500A9E4D7 /* ExplicitTypeInterfaceRule.swift */; };
 		C3DE5DAC1E7DF9CA00761483 /* FatalErrorMessageRule.swift in Sources */ = {isa = PBXBuildFile; fileRef = C3DE5DAA1E7DF99B00761483 /* FatalErrorMessageRule.swift */; };
+		C3FEAFEC1F3CC70C0097BF0E /* InlineCommentRule.swift in Sources */ = {isa = PBXBuildFile; fileRef = C3FEAFEA1F3CC64A0097BF0E /* InlineCommentRule.swift */; };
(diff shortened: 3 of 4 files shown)


Source/SwiftLintFramework/Rules/InlineCommentRule.swift
```diff
@@ -0,0 +1,69 @@
...
+
+    public static let description = RuleDescription(
+        identifier: "inline_comment",
+        name: "InlineComment",
+        description: "Inline comments should be in valid format.",
+        nonTriggeringExamples: [
+            "var foo: Bool?  // Good"
+        ],
+        triggeringExamples: [
+        "↓var foo1: Date?  //Wrong",
```
Uncommon: The ↓ should probably be at the start of the comment.

kimdv:
@marcelofabri I'm new to regex, so will you please review this and see if there are missing some cases? 😄

marcelofabri:
I wonder if we can avoid triggering in header comments, since the Xcode template uses two spaces after `//`.

kimdv:
@marcelofabri I have just tried with a plain swift file with only headers, and it was not triggering. 

```
Linting 'HeaderTest.swift' (1/1)
```

marcelofabri:
@kimdv I've enabled it on SwiftLint itself and it triggered a lot of false positives.

kimdv:
@marcelofabri i will take a look when I get home in begging of august.

Do you have any optimarions tips to the reggex

kimdv:
@marcelofabri Now i've got some time to look at it! 😄 

I seems the Xcode9 build fails.. Will you check for some false positives. If you find any, please tell what file, thanks! 🚀

Source/SwiftLintFramework/Rules/InlineCommentRule.swift
```diff
@@ -0,0 +1,81 @@
...
+        kind: RuleKind.lint,
+        nonTriggeringExamples: [
+            "// Good\nfunc foo() {\n}",
+            "func foo() {  // Good\n}",
+            "class Foo {var foo = Date()  // Good\n}",
+            "class Foo {var foo: Date?  // Good\n}",
+            "class Foo {var foo = [1,  // Good\n2]\n}"
+        ],
+        triggeringExamples: [
+        "func foo() ↓{  //Wrong\n}",
```
marcelofabri: these violation offsets are really weird. IMO, they should be before the start of a comment (e.g. `↓//`) or where the "error" really is (in this case, I think it's because there's no space between `//` and `Wrong`, so the marker could be `//↓Wrong`).

Source/SwiftLintFramework/Rules/InlineCommentRule.swift
```diff
@@ -0,0 +1,81 @@
...
+
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "inline_comment",
+        name: "InlineComment",
+        description: "Inline comments should be in valid format.",
+        kind: RuleKind.lint,
```
marcelofabri: you can use just `.lint` here

Source/SwiftLintFramework/Rules/InlineCommentRule.swift
marcelofabri: also, I think it'd be nice to add a comment to each line to indicate why that example is invalid and/or improving the rule description and/or adding a custom violation message.

marcelofabri: another thing: why are two spaces required before a comment? Is that a convention documented somewhere? I'd love to see that as a configuration.

kimdv: #1310 have two, from Python's PEP8 inline comment rule. 

But a configuration would be nice! I'll put that on my todo today. 

But should one or two spaces be defualt? What do you think?

marcelofabri: I think it should be just one. It doesn't look like using two spaces in common in the Objective-C/Swift community.

Source/SwiftLintFramework/Rules/InlineCommentRule.swift
```diff
@@ -69,13 +65,18 @@ public struct InlineCommentRule: ConfigurationProviderRule, OptInRule {
...
+                .byteRangeToNSRange(start: syntaxTokens[0].offset + syntaxTokens[0].length, length: 0)
+
+            var commentStart = range
+            commentStart.location -= 1
+            commentStart.location += range.length - 2 // Sets the "pointer" right before the comment starts
 
-            return identifierRange.map { NSUnionRange($0, range) }
+            return identifierRange.map { NSUnionRange($0, commentStart) }
         }
     }
```
kimdv: @marcelofabri is this the right approach ? 😄

kimdv:
hmm.. @marcelofabri the `RulesTests` seems to fail because of `failed - Type body should span 200 lines or less excluding comments and whitespace: currently spans 201 lines`

marcelofabri:
@kimdv Sorry for the delay. Is this still WIP? If not, can you rebase it?

That failure that you've mentioned was fixed in `master`.

kimdv:
@marcelofabri I will make a new pull request. The rebase didn't work for me... Messed everything up 😰😰🙈🙈

I will make a new one. 

But when I try to run Sourcery to generate the rules list and tests I get this message and nothing chnaged

```
SwiftLint kimdevos$ make sourcery
make: `sourcery' is up to date.
```

kimdv:
Work is on #1876

PR #1511, opened 2017-05-14
Add protocol_property_accessors_order rule

### LINKED ISSUE #1504: Rule request: order of `get set` in protocol
When defining a property in a protocol, the order of accessors should be `get set`

```swift
protocol Foo {

    // ok
    var bar: String { get set }

    // not ok
    var bar: String { set get }
}
```

marcelofabri:
Fixes #1504

### DIFF
--- a/Source/SwiftLintFramework/Rules/ProtocolPropertyAccessorsOrderRule.swift
+++ b/Source/SwiftLintFramework/Rules/ProtocolPropertyAccessorsOrderRule.swift
@@ -1,0 +1,68 @@
+//
+//  ProtocolPropertyAccessorsOrderRule.swift
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 15/05/17.
+//  Copyright © 2017 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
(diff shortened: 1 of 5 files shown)


Source/SwiftLintFramework/Rules/ProtocolPropertyAccessorsOrderRule.swift
```diff
@@ -0,0 +1,68 @@
...
+    public func validate(file: File) -> [StyleViolation] {
+        return violationRanges(file: file).map {
+            StyleViolation(ruleDescription: type(of: self).description,
+                           severity: configuration.severity,
+                           location: Location(file: file, characterOffset: $0.location))
+        }
+    }
+
+    private func violationRanges(file: File) -> [NSRange] {
+        return file.match(pattern: "\\bset\\s*get\\b", with: [.keyword, .keyword])
```
marcelofabri: this is pretty dumb because AFAIK the only place where `set get` is valid is inside a protocol property, so I didn't even bother to check if it was inside a property (and inside a protocol)

jpsim:
🎉

Issue #1507, opened 2017-05-11
Extraneous `file.zip` in CocoaPods release


heshamsalman:
Looking through the source in the CocoaPods releases, there's an extraneous `file.zip` that takes up 4.4MB and contains _a duplicate copy of the SwiftLint framework and binary_.

marcelofabri:
I think this is a side effect of https://github.com/realm/SwiftLint/blob/master/SwiftLint.podspec#L9. We could change the regex to not match `file.zip` and try it.

jpsim:
Fixed in 536c24c39dd95bdb91286def32e51c5500243363.

PR #1513, opened 2017-05-15
Excluded file.zip from pod output

heshamsalman:
References: #1507

Hopefully this should remove `file.zip` in the pod output.


jpsim:
Thanks! Merged in 536c24c39dd95bdb91286def32e51c5500243363.

PR #1516, opened 2017-05-15
Reenable cache

jpsim:
Continued from #1311. Addresses #1184.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -171,6 +171,24 @@
   version after important cache-related issues have been addressed.  
   [Marcelo Fabri](https://github.com/marcelofabri)
 
+* Add `modificationDate(forFileAtPath:)` function requirement to
+  `LintableFileManager` protocol.
+  [Victor Pimentel](https://github.com/victorpimentel)
+
+* Replace `configurationHash` parameters in all `LinterCache` initialiazers
+  with `configurationDescription` parameters.
+  [Victor Pimentel](https://github.com/victorpimentel)
+
+* Remove `fileHash` parameter from `LinterCache.cache(violations:forFile:)`
+  and `LinterCache.violations(forFile:)` functions.
+  [Victor Pimentel](https://github.com/victorpimentel)
+
+* Remove `hash` attribute from `Configuration` struct.
+  [Victor Pimentel](https://github.com/victorpimentel)
+
+* Add `configurationDescription` attribute to `Configuration` struct.
+  [Victor Pimentel](https://github.com/victorpimentel)
+
 ##### Enhancements
 
 * Add `implicitly_unwrapped_optional` opt-in rule that warns against using
@@ -315,6 +333,15 @@
 
 * Fix `checkstyle` report format.  
   [Yuki Oya](https://github.com/YukiOya)
+
+* Fix stale cache by using file modification dates instead of hashing whole
+  files.
+  [Victor Pimentel](https://github.com/victorpimentel)
+  [#1184](https://github.com/realm/SwiftLint/issues/1184)
+
+* Move file cache from `ApplicationSupport` to `~/Library/Caches` directory.
+  [Victor Pimentel](https://github.com/victorpimentel)
+  [#1184](https://github.com/realm/SwiftLint/issues/1184)
 
 ## 0.16.1: Commutative Fabric Sheets
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,11 +11,30 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1387](https://github.com/realm/SwiftLint/issues/1453)
 
+* Add `modificationDate(forFileAtPath:)` function requirement to
+  `LintableFileManager` protocol.  
+  [Victor Pimentel](https://github.com/victorpimentel)
+
+* Replace `configurationHash` parameters in all `LinterCache` initializers
+  with `configurationDescription` parameters.  
+  [Victor Pimentel](https://github.com/victorpimentel)
+
+* Remove `fileHash` parameter from `LinterCache.cache(violations:forFile:)`
+  and `LinterCache.violations(forFile:)` functions.  
+  [Victor Pimentel](https://github.com/victorpimentel)
+
+* Remove `hash` attribute from `Configuration` struct.  
+  [Victor Pimentel](https://github.com/victorpimentel)
+
 * Rename `ConditionalReturnsOnNewline` struct to
   `ConditionalReturnsOnNewlineRule` to match rule naming conventions.  
   [JP Simard](https://github.com/jpsim)
 
 ##### Enhancements
+
+* Cache linter results for files unmodified since the previous linter run.  
+  [Victor Pimentel](https://github.com/victorpimentel)
+  [#1184](https://github.com/realm/SwiftLint/issues/1184)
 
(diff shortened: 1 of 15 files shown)


jpsim:
I'd love if you could review this too @victorpimentel.

Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -18,58 +18,67 @@ public enum LinterCacheError: Error {
...
 
-    public func cache(violations: [StyleViolation], forFile file: String, fileHash: Int) {
+    public func cache(violations: [StyleViolation], forFile file: String) {
+        guard let lastModification = fileManager.modificationDate(forFileAtPath: file) else {
+            return
+        }
+
         lock.lock()
-        var filesCache = (cache["files"] as? [String: Any]) ?? [:]
+        var filesCache = (cache[LinterCache.Key.files.rawValue] as? [String: Any]) ?? [:]
```
victorpimentel: Very small code style question: is there any reason to use LinterCache.Key in some lines and just Key in others? It's not important, but I'm reviewing this from my phone so I'm not 100% sure if they refer to another type.

jpsim: I (inconsistently) omitted `LinterCache` within its body since it could be inferred. This one slipped through. Fixed in 43e4ed6.

CHANGELOG.md
```diff
@@ -11,8 +11,27 @@
...
+  [Victor Pimentel](https://github.com/victorpimentel)
+
+* Remove `hash` attribute from `Configuration` struct.
+  [Victor Pimentel](https://github.com/victorpimentel)
+
 ##### Enhancements
 
+* Cache linter results for files unmodified since the previous linter run.
+  [Victor Pimentel](https://github.com/victorpimentel)
+  [#1184](https://github.com/realm/SwiftLint/issues/1184)
```
marcelofabri: missing two trailing spaces on all entries 😅

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -34,9 +32,25 @@ public struct Configuration: Equatable {
     public let rules: [Rule]
     public var rootPath: String?              // the root path to search for nested configurations
     public var configurationPath: String?     // if successfully loaded from a path
-    public var hash: Int?
     public let cachePath: String?
 
+    public var cacheDescription: String {
+        let cacheRulesDescriptions: [String: Any] = rules.reduce([:]) { accu, element in
+            var accu = accu
+            accu[type(of: element).description.identifier] = element.configurationDescription
```
marcelofabri: the `configurationDescription` of `ConditionalReturnsOnNewline` is `"N/A"` for some reason. We probably should change that as well.

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -34,9 +32,25 @@ public struct Configuration: Equatable {
...
+            return accu
+        }
+        let dict: [String: Any] = [
+            "root": rootPath ?? FileManager.default.currentDirectoryPath,
+            "rules": cacheRulesDescriptions
+        ]
+        if let prettyJSONData = try? JSONSerialization.data(withJSONObject: dict),
```
marcelofabri: shouldn't this pass `options` to actually be a pretty? 😆

marcelofabri:
Also, does this handle nested configurations? (It doesn't look like it does)

CHANGELOG.md
jpsim: new computer accidentally has `whitespace = fix` in its `.gitconfig` so it automatically trimmed trailing whitespace when rebasing. Promptly removed that.

Source/SwiftLintFramework/Models/Configuration.swift
jpsim: fixed in #1523.

Source/SwiftLintFramework/Models/Configuration.swift
jpsim: doesn't need to be pretty actually, since I'm just using this for serialization. fixed variable name.

jpsim:
I've updated/rebased this after #1522, #1523 & #1524 were merged. I also updated it to consider custom rules in cache invalidation.

In manual testing, I've confirmed that:

An individual file cache is invalidated when:
- The file is modified, either by changing its contents or running `touch file.swift`

Entire cache is invalidated when:
- `.swiftlint.yml` custom rules are added/removed/modified
- `disabled_rules`, `enabled_rules`, `opt_in_rules`, `whitelist_rules` are added/removed/changed
- a rule's configuration is changed in `.swiftlint.yml`

Cache is not invalidated when:
- contents of `.swiftlint.yml` is reordered
- comments or whitespace are added/removed in `.swiftlint.yml`
- `cache_path`, `excluded`, `included`, `reporter`, `swiftlint_version`, `use_nested_configs` or `warning_threshold` are added/removed/changed

The only thing I can see doesn't work is nested configurations. I haven't really looked into what it would take to support that either. I suspect it wouldn't be too hard though.

Source/SwiftLintFramework/Rules/RuleConfigurations/RegexConfiguration.swift
```diff
@@ -27,6 +27,29 @@ public struct RegexConfiguration: RuleConfiguration, Equatable {
...
+    internal var cacheDescription: String {
+        var dict = [String: Any]()
+        dict["identifier"] = identifier
+        if let name = name {
+            dict["name"] = name
+        }
+        dict["message"] = message
+        dict["regex"] = regex.pattern
+        if let included = included?.pattern {
+            dict["included"] = included
```
marcelofabri: don't `dict["included"] = included?.pattern` have the same effect?

marcelofabri:
> The only thing I can see doesn't work is nested configurations. I haven't really looked into what it would take to support that either. I suspect it wouldn't be too hard though.

Do you want to merge this without supporting nested configurations? If so, I'd suggest to disable caching by default until we do support them.

Source/SwiftLintFramework/Rules/RuleConfigurations/RegexConfiguration.swift
jpsim: that would produce `["included": nil]` rather than `[:]`

marcelofabri: It doesn't 😅

```
var dict = [String: String]()
dict["foo"] = nil
dict["bar"] = "bar"

let data = try! JSONSerialization.data(withJSONObject: dict)
print(String(data: data, encoding: .utf8)!) // {"bar":"bar"}
```

jpsim: 👍  thanks

jpsim:
WIP support for nested configurations: https://github.com/realm/SwiftLint/compare/jp-reenable-cache...jp-wip-cache-support-nested-configs

I have some cleaning up to do and tests to write, but it shows that it can be done fairly easily.

jpsim:
I'd really rather not ship this (again) with known issues or limitations.

samskiter:
Thanks so much for this @jpsim  👍

jpsim:
Superseded by #1530.

Issue #1520, opened 2017-05-16
file_header rule should trigger on first line if there're no comments


rjhodge:
So I have the file header rule working great with one small problem. One developer is notorious for not having a file header. How can I enforce requiring this?

marcelofabri:
Do you want to enforce a certain header format? Take a look how SwiftLint itself does it: https://github.com/realm/SwiftLint/blob/master/.swiftlint.yml#L25-L33

rjhodge:
So I have the header format

```
file_header:
  required_pattern: |
                    \/\/
                    \/\/  .*?\.swift
                    \/\/  Project
                    \/\/
                    \/\/  Created by .*? on \d{1,2}\/\d{1,2}\/\d{2}\.
                    \/\/  Copyright © \d{4} Company\. All rights reserved\.
                    \/\/
```

But he will often put in the following for files

```
import MapKit
import UIKit
```

And this still passes the file_header rule.

marcelofabri:
This rule is opt-in, are you sure you've enabled it as well? It should trigger a warning.

rjhodge:
We are using the whitelist_rules and have it listed as one of the rules. And the rule works great at finding inconsistent headers but does nothing when there is no header.

marcelofabri:
Can you provide a small example that triggers the issue? I couldn't reproduce it:

```
$ swiftlint      
Loading configuration from '.swiftlint.yml'
Linting Swift files in current working directory
Linting 'file.swift' (1/1)
/Users/marcelofabri/Documents/file.swift: warning: File Header Violation: Files should have consistent header comments. (file_header)
Done linting! Found 1 violation, 0 serious in 1 file.
```

**`file.swift`**
```swift
import MapKit
import UIKit
```

**`.swiftlint.yml`**
```yml
opt_in_rules:
  - file_header

file_header:
  required_pattern: |
                    \/\/
                    \/\/  .*?\.swift
                    \/\/  Project
                    \/\/
                    \/\/  Created by .*? on \d{1,2}\/\d{1,2}\/\d{2}\.
                    \/\/  Copyright © \d{4} Company\. All rights reserved\.
                    \/\/
```

marcelofabri:
Maybe the issue is that when there's a comment, the warning is triggered on that line. Otherwise, the warning is only on the file. That could make Xcode behaves differently.

rjhodge:
[...]

PR #1521, opened 2017-05-16
Updated file_header rule to use line 1 character 1 for files that are missing a header.

rjhodge:
Updated file_header rule to use line 1 character 1 for files that are missing a header. #1520

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,7 +62,8 @@
   [Hesham Salman](https://github.com/heshamsalman)
   [#1507](https://github.com/realm/SwiftLint/issues/1507)
 
-* Updated file_header rule to use line 1 character 1 for files that are missing a header
+* Update file_header rule to trigger on the first line when missing a header,
+  so the warning will be shown in Xcode editor.  
   [rjhodge](https://github.com/rjhodge)
   [#1520](https://github.com/realm/SwiftLint/issues/1520)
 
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Rules/FileHeaderRule.swift
```diff
@@ -83,7 +83,7 @@ public struct FileHeaderRule: ConfigurationProviderRule, OptInRule {
                 StyleViolation(
                     ruleDescription: type(of: self).description,
                     severity: configuration.severityConfiguration.severity,
-                    location: Location(file: file.path)
+                    location: Location(file: file.path, line: 1, character: 1) // Force to first line and character
```
marcelofabri: could you remove the `character` parameter since in this case the warning is not related to a particular offset?

CHANGELOG.md
```diff
@@ -62,6 +62,10 @@
   [Hesham Salman](https://github.com/heshamsalman)
   [#1507](https://github.com/realm/SwiftLint/issues/1507)
 
+* Updated file_header rule to use line 1 character 1 for files that are missing a header
```
marcelofabri: What do you think of

> Update `file_header` rule to trigger on the first line when missing a header, so the warning will be shown in Xcode editor.

(this needs to be capped at 80 chars)

marcelofabri:
Thanks! 💯

PR #1523, opened 2017-05-16
use default implementation of configurationDescription for ConditionalReturnsOnNewlineRule

jpsim:


jpsim:
Addresses https://github.com/realm/SwiftLint/pull/1516/files#r116753497.

PR #1524, opened 2017-05-16
refactor CustomRules.validate(file:)

jpsim:


jpsim:
This:

* avoids iterating over configurations twice to apply two filters by merging them
* caches the path range rather than bridging to NSString & creating the NSRange for every iteration
* uses `first(where:)` rather than `filter().first` to avoid iterating over all regions
* uses `contains` rather than `for x in y where z { return false } else { return true }`

PR #1526, opened 2017-05-17
Fall back to line 1 when no line is provided & reporting with `xcode`

jpsim:
this is a generalization of the fix from #1521

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,11 +66,6 @@
   [Hesham Salman](https://github.com/heshamsalman)
   [#1507](https://github.com/realm/SwiftLint/issues/1507)
 
-* Update file_header rule to trigger on the first line when missing a header,
-  so the warning will be shown in Xcode editor.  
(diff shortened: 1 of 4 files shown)


jpsim:
fyi @rhodgex1 / @rjhodge

rjhodge:
Definitely like this solution way better than my one off fix.

marcelofabri:
I don't know how I feel about different reporters having different locations 😬 

In the case of this particular rule, I'd keep the from #1521 as well

jpsim:
> In the case of this particular rule, I'd keep the from #1521 as well

What do you mean by this?

marcelofabri:
Sorry, shouldn't comment on PRs so late! 💤 

I meant that IMO [this](https://github.com/realm/SwiftLint/pull/1526/files#diff-c56abfa701d430bc5b3a2dd78dddc717L86) shouldn't be done. We should always trigger on the first line in that case in `file_header` rule.

PR #1530, opened 2017-05-18
Reenable cache

jpsim:
Significantly extended from #1516 /cc @victorpimentel @marcelofabri


Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -32,9 +32,47 @@ public struct Configuration: Equatable {
...
+        let cacheRulesDescriptions: [String: Any] = rules.reduce([:]) { accu, element in
+            var accu = accu
+            accu[type(of: element).description.identifier] = element.cacheDescription
+            return accu
+        }
+        let dict: [String: Any] = [
+            "root": rootPath ?? FileManager.default.currentDirectoryPath,
+            "rules": cacheRulesDescriptions
+        ]
+        if let jsonData = try? JSONSerialization.data(withJSONObject: dict),
```
norio-nomura: AFAIK, the json string created by `JSONSerialization` is non-deterministic in dictionary key order on Linux.
Should we consider something about it?

jpsim: None of this relies on key order remaining stable.

jpsim: Oh, yes it does of course. The `cache.json` file doesn't rely on ordering, but since this is embedded as a single, that _does_ matter.

jpsim: What do you mean though that the order is non-deterministic?

norio-nomura: `configuration.cacheDescription` is used for key to getting cache in `cache(violations:forFile:configuration:)` and `violations(forFile:configuration:)`.
`cacheDescription` is jsonstring serialized by `JSONSerialization`.
So, getting cache depends on non-deterministic string keys.

norio-nomura: Cache won't hit if key will be changed on every serialization.

norio-nomura: I tried this PR on Linux(norionomura/sourcekit:311).
[`cache.json`](https://gist.github.com/norio-nomura/e3584c423ae0ba61197c0bb542d7d629) contains three variants of `configurationDescription` after three times executing `.build/debug/swiftlint lint --cache-path .` in SwiftLint working copy.

norio-nomura: > What do you mean though that the order is non-deterministic?

I meant that "it is not guaranteed that the output result from a certain input is constant. "
I do not know if the word usage is correct or not. 🙇

SDGGiesbrecht: Yes, @norio-nomura, “non‐deterministic” is precisely the right word.

jpsim: So... any ideas for deterministic serialization? We could simply avoid the issue for now by disabling caching on Linux?

norio-nomura: `Yams.dump(object:)` can do deterministic serialization.
As I tried it now, the order of `Set` does not seem to be constant by Yams.
I do not know if `JSONSerialization` will always serialize` Set` in the same order.

norio-nomura: > But as I tried it now, the order of Set does not seem to be constant by Yams.

Ah, both of `JSONSerialization ` and `Yams` do not serialize `Set` directly and can't serialize `Set`.
They are using results of inserting `Set` into String interpolation.
Inserting `Set` into String interpolation in `public var consoleDescription: String {…}` seems non-deterministic on Linux. 🤔

marcelofabri: Well, `Set`s don't preserve order by definition, so I think this is the expected behavior 😅

norio-nomura: Yup. The fact that the result is always constant with macOS is just such an implementation, it is not guaranteed as a specification.

marcelofabri: But actually, do we use `Set` here? Or just `Dictionary`?

norio-nomura: `Configuration.cacheDescription` > `Rule.cacheDescription` > `RuleConfiguration.consoleDescription` > https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules/RuleConfigurations/AttributesConfiguration.swift#L13-L20

marcelofabri: 🤔 

We could revisit those `consoleDescription` to avoid using `Set`s. In this particular case, it'd be probably even better for the user if we always sorted the `Set`s. Or even don't use them at all.

jpsim: Took a stab at fixing this in 0c09520 & cea6f43.

jpsim:
Will merge. We should do a prerelease of SwiftLint with this and all the many other changes to make sure it doesn't break things in the wild since there's so much.

marcelofabri:
🎉

jpsim:
[0.19.0-rc.1](https://github.com/realm/SwiftLint/releases/tag/0.19.0-rc.1)

PR #1533, opened 2017-05-19
Update danger to 5.2.2

sleekybadger:
@marcelofabri this should fix wrong violations order, that you described here https://github.com/danger/danger/issues/703


jpsim:
Looks good, thanks!

PR #1534, opened 2017-05-19
add `make docker_htop` command

jpsim:
docker provides very little insight into what's happening in its containers, so this command helps keep tabs on what processes are running.


norio-nomura:
Wow! Great!

PR #1536, opened 2017-05-21
Fix false positive in `syntactic_sugar` rule when using nested types

### LINKED ISSUE #1508: False positive on `syntactic_sugar` rule
I have a nested type named `<SomeClass>.Optional` and swiftlint thinks this is `Swift.Optional` so I am getting the following warning:
screen shot 2017-05-12 at 18 42 27

marcelofabri:
Fixes #1508


jpsim:
🎉

PR #1538, opened 2017-05-21
Add NoExtensionAccessModifierRule

### LINKED ISSUE #1457: Optionally disallow access modifier override on extensions

marcelofabri:
Continued from #1470

Fixes #1457

I've rebased this and made some changes on 	12360cc:

* Make the rule an `ASTRule`, so we don't need to rely on extensions being only top level declarations
* Check if the token is actually an ACL identifier. I don't think this would cause issues for us, but better being safe than sorry in this case.
* Small formatting changes

// cc @masters3d

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,11 @@
   [Samuel Susla](https://github.com/sammy-SC)
   [Jeremy David Giesbrecht](https://github.com/SDGGiesbrecht)
   [#1326](https://github.com/realm/SwiftLint/issues/1326)
-
+  
+* Added `no_extension_access_modifier` opt-in rule to disallow access modifiers completely, à la SE-0119.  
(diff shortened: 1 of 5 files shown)


jpsim:
🎉. Small formatting fix in ea4846b.

Should we enable this in our `.swiftlint.yml`? I think we should.

marcelofabri:
My personal preference is not enabling, since in some cases this is helpful to avoid line length limit on a method declaration.

jpsim:
> in some cases this is helpful to avoid line length limit on a method declaration.

True, but I found only two cases of this. See #1543.

PR #1543, opened 2017-05-22
Enable no_extension_access_modifier in .swiftlint.yml

jpsim:
Rule added in #1538.

There are two cases where a function declaration overflowed to two lines, but I think the consistency win is worthwhile here.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -21,6 +21,7 @@ opt_in_rules:
   - number_separator
   - prohibited_super_call
   - fatal_error_message
+  - no_extension_access_modifier
 
 file_header:
   required_pattern: |
--- a/Source/SwiftLintFramework/Models/Command.swift
+++ b/Source/SwiftLintFramework/Models/Command.swift
@@ -9,8 +9,8 @@
 import Foundation
 
 #if !os(Linux)
-private extension Scanner {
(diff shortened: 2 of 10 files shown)


masters3d:
Where you also planning on enabling this? https://github.com/realm/SwiftLint/pull/1476

If so, then https://github.com/realm/SwiftLint/pull/1489
would help limit fileprivate usage on top level declarations thus eliminating the extensions confusion on private vs fileprivate. 

My only issue with no extension modifiers is that it seems to propagate more the usage of fileprivate. Personally I rather have explicit top access labels then just allow private usage at top levels instead of fileprivate  vs not to allowing access modifiers on extensions. 

I do like the consistency thought, I just wanted to give my two cents.

marcelofabri:
My personal preference would actually to create an opposite rule of this:

> If all declarations in an extension have the same ACL, the ACL should be in the extension itself.

But I don't mind if we merge this either.

jpsim:
> If all declarations in an extension have the same ACL, the ACL should be in the extension itself.

I like that, but it has subtle implications when used with `private`/`fileprivate` (no big surprises there).

I'm fine with not enabling this. I guess we're discovering/formalizing the code style we've wanted all along when new rules are introduced 😉.

> Where you also planning on enabling this? #1476

I think that'd be nice.

Issue #1550, opened 2017-05-23
Unneeded lock in cache?


marcelofabri:
I think this lock is unneeded: https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Models/LinterCache.swift#L68

Removing this could bring major performance benefits for the use case of running without any modifications, as it's slower than what I'd expect:

```
$ time Pods/SwiftLint/swiftlint lint --quiet
Pods/SwiftLint/swiftlint lint --quiet  5.96s user 0.26s system 122% cpu 5.065 total
```

marcelofabri:
I've tested it locally and now I get ~2.0s total

norio-nomura:
On checking this, Thread Santizer detects following in `violations(forFile:configuration:)`
```
==================
WARNING: ThreadSanitizer: data race (pid=14103)
  Read of size 8 at 0x7d18000083d8 by thread T5:
    #0 LinterCache.fileManager.getter LinterCache.swift (SwiftLintFramework:x86_64+0x2d2d17)
    #1 LinterCache.violations(forFile : String, configuration : Configuration) -> [StyleViolation]? LinterCache.swift:62 (SwiftLintFramework:x86_64+0x2d9729)
    #2 Linter.cachedStyleViolations(benchmark : Bool) -> ([StyleViolation], [(id : String, time : Double)])? Linter.swift:103 (SwiftLintFramework:x86_64+0x81a55)
    #3 Linter.getStyleViolations(benchmark : Bool) -> ([StyleViolation], [(id : String, time : Double)]) Linter.swift:70 (SwiftLintFramework:x86_64+0x7f287)
    #4 Linter.styleViolations.getter Linter.swift:61 (SwiftLintFramework:x86_64+0x7eced)
    #5 LintCommand.(run(LintOptions) -> Result<(), CommandantError<()>>).(closure #1) LintCommand.swift:40 (swiftlint:x86_64+0x10001c8a6)
    #6 partial apply for LintCommand.(run(LintOptions) -> Result<(), CommandantError<()>>).(closure #1) LintCommand.swift (swiftlint:x86_64+0x10001d0a5)
[...]

PR #1551, opened 2017-05-23
Split cache into read cache and write cache

marcelofabri:
Alternative fix for #1550 as suggested by @norio-nomura 

This makes things slightly more complex, but it reduced by 50-70% the execution time when there's a cache available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,9 @@
 * Cache linter results for files unmodified since the previous linter run.  
   [Victor Pimentel](https://github.com/victorpimentel)
   [JP Simard](https://github.com/jpsim)
+  [Marcelo Fabri](https://github.com/marcelofabri)
   [#1184](https://github.com/realm/SwiftLint/issues/1184)
+  [#1550](https://github.com/realm/SwiftLint/issues/1550)
 
 * Add opt-in configurations to `generic_type_name`, `identifier_name` and
   `type_name` rules to allow excluding non-alphanumeric characters and names 
--- a/Source/SwiftLintFramework/Models/LinterCache.swift
+++ b/Source/SwiftLintFramework/Models/LinterCache.swift
@@ -15,30 +15,47 @@ internal enum LinterCacheError: Error {
 }
 
 public final class LinterCache {
-    private var cache = [String: Any]()
+    private typealias Cache = [String: [String: [String: Any]]]
+
+    private let readCache: Cache
+    private var writeCache = Cache()
     private let lock = NSLock()
-    internal lazy var fileManager: LintableFileManager = FileManager.default
+    internal let fileManager: LintableFileManager
     private let location: URL?
 
-    internal init() {
+    internal init(fileManager: LintableFileManager = FileManager.default) {
         location = nil
+        self.fileManager = fileManager
+        self.readCache = [:]
     }
 
-    internal init(cache: Any) throws {
-        guard let dictionary = cache as? [String: Any] else {
+    internal init(cache: Any, fileManager: LintableFileManager = FileManager.default) throws {
+        guard let dictionary = cache as? Cache else {
             throw LinterCacheError.invalidFormat
         }
 
-        self.cache = dictionary
+        self.readCache = dictionary
         location = nil
+        self.fileManager = fileManager
     }
 
-    public init(configuration: Configuration) {
+    public init(configuration: Configuration,
+                fileManager: LintableFileManager = FileManager.default) {
         location = configuration.cacheURL
         if let data = try? Data(contentsOf: location!),
-            let json = try? JSONSerialization.jsonObject(with: data) {
-            cache = (json as? [String: Any]) ?? [:]
+            let json = try? JSONSerialization.jsonObject(with: data),
+            let cache = json as? Cache {
+            readCache = cache
+        } else {
+            readCache = [:]
         }
+        self.fileManager = fileManager
+    }
+
+    private init(cache: Cache, location: URL?, fileManager: LintableFileManager) {
+        self.readCache = cache
+        self.location = location
+        self.fileManager = fileManager
     }
 
     internal func cache(violations: [StyleViolation], forFile file: String, configuration: Configuration) {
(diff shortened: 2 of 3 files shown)


jpsim:
I like the general idea, but `flushWriteCache()` mutates `readCache` but reads to `readCache` aren't locked, so the mutation could happen in the middle of reading.

Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -65,31 +70,47 @@ public final class LinterCache {
...
+        }
+        flushWriteCache()
         lock.lock()
-        let json = toJSON(cache)
+        let json = toJSON(readCache)
         lock.unlock()
         try json.write(to: url, atomically: true, encoding: .utf8)
     }
 
+    internal func flushWriteCache() {
```
jpsim: `flushWriteCacheToReadCache()` might be a clearer (but more verbose) name?

jpsim: Otherwise it isn't clear that this function is mutating `readCache`.

marcelofabri: Sounds good to me - this method shouldn't be used often, so I'm cool with it being more verbose

jpsim:
> Yes, but this `flushWriteCache` should only be called after the violations were processed. Everything was originally on save, but I had to expose it for unit tests 😅

I guess this brings us to whether or not we want to properly support arbitrary uses of `SwiftLintFramework` outside of just our `swiftlint` CLI. _We_ may not call `save()` while other linters are running, but the public API certainly allows you to do so.

marcelofabri:
> We may not call save() while other linters are running, but the public API certainly allows you to do so.

Agreed, but I think we should be fine if we document this limitation. I think the performance win for `swiftlint` is worth the less safety for other (potential) consumers of the public API.

jpsim:
I agree, but it should be clearly documented.

Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -65,31 +70,47 @@ public final class LinterCache {
...
 
     public func save() throws {
         guard let url = location else {
             throw LinterCacheError.noLocation
         }
+        guard !writeCache.isEmpty else {
+            return
+        }
+        flushWriteCacheToReadCache()
         lock.lock()
```
jpsim: This `lock()` (and subsequent `unlock()`) can be removed if we maintain that:

1) `readCache` can only be mutated in `flushWriteCacheToReadCache()`
2) `flushWriteCacheToReadCache()` can only be called from `save()`
3) `save()` cannot be called concurrently

marcelofabri: Actually, we can make `save()` thread-safe if we think about a way to support the unit tests. Maybe something like this

```swift
func flushWriteCacheToReadCache() {
    readCache = merge()
}

func merge() -> [String: Any] {
  // ...
}

func save() {
   // ...
    let cache = merge()
    let json = toJSON(cache)
   // ...
}
```

(please ignore the bad names and lack of details)

jpsim: I think that's doable. Give it a try and see! 😄

marcelofabri: Done! 💯

Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -15,30 +15,35 @@ internal enum LinterCacheError: Error {
 }
 
 public final class LinterCache {
-    private var cache = [String: Any]()
+    private var readCache = [String: Any]()
+    private var writeCache = [String: Any]()
```
norio-nomura: Changing `readCache` and `writeCache` from `[String: Any]` to `[String: [String: Any]]` boosted performance on my test.

Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -88,30 +100,28 @@ public final class LinterCache {
...
     }
 
-    private func mergeCaches() -> [String: Any] {
+    private func mergeCaches() -> Cache {
         var cache = readCache
+        lock.lock()
         for (key, value) in writeCache {
-            var filesCache = (cache[key] as? [String: Any]) ?? [:]
-            for (file, fileCache) in (value as? [String: Any]) ?? [:] {
+            var filesCache = cache[key] ?? [:]
```
marcelofabri: Yes, but is it worth the extra complexity?


```swift
for (key, value) in writeCache {
    if var filesCache = cache[key] {
        for (file, fileCache) in value {
            filesCache[file] = fileCache
        }
        cache[key] = filesCache
    } else {
        cache[key] = value
    }
}
```

Tests/SwiftLintFrameworkTests/LinterCacheTests.swift
```diff
@@ -77,9 +77,7 @@ class LinterCacheTests: XCTestCase {
     // MARK: Test Helpers
 
     private var cache: LinterCache = {
-        let cache = LinterCache()
-        cache.fileManager = TestFileManager()
-        return cache
+        return LinterCache(fileManager: TestFileManager())
```
jpsim: This doesn't need to be in a closure anymore.

Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -49,12 +66,12 @@ public final class LinterCache {
         let configurationDescription = configuration.cacheDescription
 
         lock.lock()
-        var filesCache = (cache[configurationDescription] as? [String: Any]) ?? [:]
+        var filesCache = (writeCache[configurationDescription]) ?? [:]
```
jpsim: Parens are surperfluous now.

CHANGELOG.md
```diff
@@ -93,6 +93,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1504](https://github.com/realm/SwiftLint/issues/1504)
 
+* Improve cache performance by splitting it into read cache and write cache.
```
jpsim: Just add your name & issue to the item introducing caching? We don't usually include fixes or improvements to unreleased things in the changelog.

marcelofabri:
🚢

PR #1555, opened 2017-05-23
[WIP] Add opt-in IBOutlet Setter Rule

jpsim:
Continued from #1355.

This rule checks if the file contains IBOutlets, if it does it looks for any lines of code that attempt to call the setter on that outlet. e.g. self.myLabel = UILabel() would trigger this rule.

While the rule is not perfect, it will help avoid an Interface Builder anti- pattern. In general, if your code is resetting IBOutlets, there is probably a better way to accomplish what you want.

There are a lot of triggering and non-triggering test cases. Along with a 3 TODOs where the rule can be triggered when it shouldn't. Because this is opt-in and the cases where the rule is triggered are fairly rare and can often be fixed by renaming a local variable, I left them as TODOs for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1504](https://github.com/realm/SwiftLint/issues/1504)
 
+* Add `outlet_setter` opt-in rule that warns when IBOutlet properties are reset
+  in code, rather than set automaticaly by IB.  
+  [Jeffrey Bergier](https://github.com/jeffreybergier)
+
 ##### Bug Fixes
 
 * `emoji` and `checkstyle` reporter output report sorted by file name.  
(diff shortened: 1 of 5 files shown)


jpsim:
@jeffreybergier could you please confirm that this rule still works as intended and that the 4 violations introduced in WordPress above are intended? Looks like to me, but would like your confirmation.

jeffreybergier:
@jpsim it looks like some are valid but pointing to a line above. One looks like its wrong. I probably won't be able to look at this until the weekend. But it looks like there are valid errors that need to be corrected. So don't merge it yet :)

jeffreybergier:
@jpsim and SwiftLint team. I've had this email in my inbox for months. Unfortunately, I'm currently traveling and I don't know when I'll have a chance to look into this. I'm not sure if anyone else is interested in picking it up or if we should just close this issue for now.

Thanks for swiftlint. I love it!

jpsim:
I haven't touched this in a while and don't expect to resume working on this. Anyone's free to pick it up from here.

PR #1561, opened 2017-05-25
Use `git-worktree(1)` in `oss-check`

norio-nomura:
This makes it possible to run `oss-check` even if the repository is not clean.

### DIFF
--- a/script/oss-check
+++ b/script/oss-check
@@ -65,10 +65,6 @@ def fail(str)
 end
 
 def validate_state_to_run
-  repo_clean = `git status --porcelain`.empty?
-
-  fail 'git repo needs to be clean to run oss-check. Aborting.' unless repo_clean
-
   if `git symbolic-ref HEAD --short`.strip == 'master'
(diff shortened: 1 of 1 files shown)


norio-nomura:
SwiftLint will be built in `osscheck/builds/branch` and  `osscheck/builds/master` directories:
```terminal.sh-session
$ git worktree list
/Users/norio/github/SwiftLint                         568a0244 [nn-use-git-worktree-in-oss-check]
/Users/norio/github/SwiftLint/osscheck/builds/branch  568a0244 (detached HEAD)
/Users/norio/github/SwiftLint/osscheck/builds/master  66c060dd (detached HEAD)
```

jpsim:
The biggest downside to this is that we lose incremental builds so we build from scratch every time we run the script and even between the `branch` and `master` builds. Not sure what would be a nice solution to that though.

norio-nomura:
I'll open PR that incremental builds will work on CircleCI.

PR #1564, opened 2017-05-25
Use `kinds(inByteRange:)` in more places

jpsim:
CommaRule, ExplicitTopLevelACLRule, ForceUnwrappingRule, ForWhereRule


jpsim:
Awesome, OSSCheck found things we weren't testing! Will fix and add tests.

PR #1568, opened 2017-05-25
Use prefix(while:) instead of partioned

marcelofabri:
Since we don't run tests on Swift 3.0.x anymore

### DIFF
--- a/Source/SwiftLintFramework/Extensions/Collection+PrefixWhile.swift
+++ b/Source/SwiftLintFramework/Extensions/Collection+PrefixWhile.swift
@@ -1,0 +1,23 @@
+//
+//  Collection+PrefixWhile.swift
+//  SwiftLint
+//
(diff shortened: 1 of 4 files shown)


jpsim:
If we deprecate Swift 3.0.x, we should list it in the changelog, readme and we'll need to bump up the  homebrew formula requirements.

marcelofabri:
> we'll need to bump up the homebrew formula requirements.

that's something that should be done when submitting the next version, right?

jpsim:
It would, but I don't think it's worth deprecating Swift 3.0 for this.

marcelofabri:
Fair enough,

marcelofabri:
Actually, I'm not sure. Since we don't test building on Swift 3.0.x anymore, maybe we'll accidentally break this in the future any way.

Also, this shouldn't matter for most people, I'd assume (since they don't build from source).

jpsim:
🎉

PR #1572, opened 2017-05-26
Adds opt-in multiline parameters rule

ornithocoder:
This commit introduces the opt-in rule that checks if methods and functions parameters - in methods and functions declarations - are all on the same line or one per line. For instance, the dummy function below violates the rule.

```swift
func foo(param1: Int,
         param2: Int, param3: Int) -> (Int) -> Int {
    return { x in x + param1 + param2 + param3 }
}
```

It should be implemented either as:

```swift
func foo(param1: Int, param2: Int, param3: Int) -> (Int) -> Int {
    return { x in x + param1 + param2 + param3 }
}
```

or

```swift
func foo(param1: Int,
         param2: Int,
         param3: Int) -> (Int) -> Int {
    return { x in x + param1 + param2 + param3 }
}
```

Tasks:

- [x] Create rule
- [x] Add entry to the CHANGELOG.md new empty section
- [x] Review violations:
    - [x] Aerial - 7 violations
    - [x] Alamofire - no violations found
    - [x] Firefox - 3 violations
    - [x] Kickstarter - 14 violations
    - [x] Moya - 1 violation
    - [x] Nimble - no violations found
    - [x] Quick - no violations found
    - [x] Realm - 14 violations
    - [x] SourceKitten - 4 violations
    - [x] Sourcery - no violations found
    - [x] Swift - 96 violations
    - [x] WordPress - 2 violations

### DIFF
--- a/Source/SwiftLintFramework/Rules/MultilineParametersRule.swift
+++ b/Source/SwiftLintFramework/Rules/MultilineParametersRule.swift
@@ -1,0 +1,74 @@
+//
+//  MultilineParametersRule.swift
+//  SwiftLint
+//
+//  Created by Ornithologist Coder on 22/05/17.
+//  Copyright © 2017 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+public struct MultilineParametersRule: ASTRule, OptInRule, ConfigurationProviderRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "multiline_parameters",
+        name: "Multiline Parameters",
+        description: "Functions and methods parameters should be either on the same line, or one per line.",
+        nonTriggeringExamples: MultilineParametersRuleExamples.nonTriggeringExamples,
+        triggeringExamples: MultilineParametersRuleExamples.triggeringExamples
+    )
+
+    public func validate(file: File,
+                         kind: SwiftDeclarationKind,
+                         dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard isValidFunction(kind),
+            let offset = dictionary.nameOffset,
+            let length = dictionary.nameLength
+            else {
+                return []
+        }
+
+        var numberOfParameters = 0
+        var linesWithParameters: Set<Int> = []
+
+        for structure in dictionary.substructure {
+            guard
+                let structureOffset = structure.offset,
+                let structureKind = structure.kind, SwiftDeclarationKind(rawValue: structureKind) == .varParameter,
+                let (line, _) = file.contents.bridge().lineAndCharacter(forByteOffset: structureOffset),
+                offset..<(offset + length) ~= structureOffset
+                else {
+                    continue
+            }
+
+            linesWithParameters.insert(line)
+            numberOfParameters += 1
+        }
+
+        guard
+            linesWithParameters.count > 1,
+            numberOfParameters != linesWithParameters.count
+            else {
+                return []
+        }
+
+        return [StyleViolation(ruleDescription: type(of: self).description,
+                               severity: configuration.severity,
+                               location: Location(file: file, byteOffset: offset))]
+    }
+
+    // MARK: - Private
+
+    private func isValidFunction(_ kind: SwiftDeclarationKind) -> Bool {
+        return [
+            SwiftDeclarationKind.functionMethodStatic,
+            SwiftDeclarationKind.functionMethodClass,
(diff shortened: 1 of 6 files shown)


Source/SwiftLintFramework/Rules/MultilineParametersRule.swift
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        return []
+    }
+
+    // MARK: - Private
+
+    private func isValidFunction(_ kind: SwiftDeclarationKind) -> Bool {
+        return [
+            SwiftDeclarationKind.functionMethodStatic,
```
marcelofabri: shouldn't this contain all `source.lang.swift.decl.function.*` kinds?

Source/SwiftLintFramework/Rules/MultilineParametersRule.swift
```diff
@@ -0,0 +1,72 @@
...
+                         kind: SwiftDeclarationKind,
+                         dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard isValidFunction(kind),
+            let offset = dictionary.nameOffset,
+            let length = dictionary.nameLength
+            else {
+                return []
+        }
+
+        var numberOfParameters: Int = 0
```
marcelofabri: you can remove the type here

Source/SwiftLintFramework/Rules/MultilineParametersRule.swift
```diff
@@ -0,0 +1,72 @@
...
+        for structure in dictionary.substructure {
+            guard
+                let structureOffset = structure.offset,
+                let structureKind = structure.kind, SwiftDeclarationKind(rawValue: structureKind) == .varParameter,
+                let (line, _) = file.contents.bridge().lineAndCharacter(forByteOffset: structureOffset)
+                else {
+                    continue
+            }
+
+            guard offset..<(offset + length) ~= structureOffset else { break }
```
marcelofabri: maybe use `continue` (and move to the `guard` above) so we don't rely on the order of parameters from SourceKit?

Source/SwiftLintFramework/Rules/MultilineParametersRule.swift
```diff
@@ -0,0 +1,72 @@
...
+
+            linesWithParameters.insert(line)
+            numberOfParameters += 1
+        }
+
+        if linesWithParameters.count > 1 && numberOfParameters != linesWithParameters.count {
```
marcelofabri: what do you think about using a `guard` here instead?

ornithocoder: Done :-)

Source/SwiftLintFramework/Rules/MultilineParametersRule.swift
ornithocoder: Done. Merged with the previous guard statement.

Source/SwiftLintFramework/Rules/MultilineParametersRule.swift
ornithocoder: Sure, no problem. Done.

Source/SwiftLintFramework/Rules/MultilineParametersRule.swift
ornithocoder: Hmm. Does it make sense? It's up to you and @jpsim. When I implemented this rule I didn't have those functions in mind.

ornithocoder: Maybe it makes sense to include .functionConstructor. What do you think, @marcelofabri?

marcelofabri: I think at least operators should be included too

marcelofabri: For consistency, I'd include all of them and then see if we have any issues

ornithocoder: Cool. I'll do that in a separate commit (instead of amending that one) in this very same pull request. Then we can discard the commit if anything goes south.

marcelofabri: Makes sense, thanks!

ornithocoder: Obrigado você!

ornithocoder: Sweet. Same violations as before ;-)

marcelofabri:
🎉

PR #1575, opened 2017-05-26
Add extension_access_modifier rule

### LINKED ISSUE #1546: [Rule Request] If all extension members are the same suggest extension modifier instead.
> If all declarations in an extension have the same ACL, the ACL should be in the extension itself.

```swift
extension Int {
    fileprivate func someFunc1(){}
    fileprivate func someFunc2(){}
    fileprivate func someFunc3(){}
}
//suggest
private extension Int { 
    func someFunc1(){} 
    func someFunc2(){} 
    func someFunc3(){}
}

```
Do not suggest when:
* all members of the extension are `private` [1]
* all members are the same but there is already an extension modifier
* all members are `internal`

[1] A private modifier on the extension effectively turns all members into fileprivate access. 

reference:
https://github.com/realm/SwiftLint/pull/1543

marcelofabri:
Fixes #1546

**TODO**:

- [x] Tests
- [x] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,11 @@
   as `@discardableResult`.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1525](https://github.com/realm/SwiftLint/issues/1525)
+* Add `extension_access_modifier` opt-in rule that validates that if all 
+  extension declaration have the same Access Control Level, the ACL should
+  be in the extension.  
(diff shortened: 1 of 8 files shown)


marcelofabri:
Should this be a violation? 🤔 

```swift
public extension Foo {
    public ↓var bar: Int { return 1 }
    public ↓func baz() {}
}
```

marcelofabri:
🚢

jpsim:
🎉

PR #1576, opened 2017-05-27
Ignore overriddes in FunctionParameterCountRule

### LINKED ISSUE #1562: function_parameter_count rule should not apply to overridden methods / protocol implementations
When overriding a method, or implementing one from a protocol, I don't have any control over the number of parameters. Therefore I feel this rule should not report in those cases. Instead it should (and presumably does) report at the base class / protocol definition.

frenetisch-applaudierend:
Fixes #1562.

Note, `make docker_test` failed with the following error, but I doubt this is due to the change:

    /usr/bin/ld.gold: fatal error: /Users/username/Development/SwiftLint/.build/debug/SwiftLintPackageTests.xctest: open: Is a directory

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
   or one per line.  
   [Ornithologist Coder](https://github.com/ornithocoder)
 
+* Update `function_parameter_count` rule to ignore overridden methods.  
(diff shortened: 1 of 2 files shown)


marcelofabri:
Thanks 💯

jpsim:
Excellent!

frenetisch-applaudierend:
My pleasure

PR #1581, opened 2017-05-28
Don't trigger `discarded_notification_center_observer` when the observer is returned

### LINKED ISSUE #1525: False positive in DiscardedNotificationCenterObserverRule
Following code is labelled with warning even thought the opaque observer is getting preserved. 

```
return nc.addObserver(forName: .NSSystemTimeZoneDidChange, object: nil, queue: nil, using: { })\n
```

marcelofabri:
Fixes #1525

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,12 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1532](https://github.com/realm/SwiftLint/issues/1532)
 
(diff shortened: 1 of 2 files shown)


jpsim:
Nice fix, although I worry that there are many more cases in which a valid use of `addObserver` that doesn't store to a variable _or_ `return`, such as appending the result to an Array/Set/etc, among many other operations...

Issue #614, opened 2016-04-13
force_unwrapping: False positive on IUO closure property declaration


ldiqual:
```
private var myProperty: (Void -> Void)! // Line 30
```

triggers

```
/path/code.swift:30:50: warning: Force Unwrapping Violation: Force unwrapping should be avoided. (force_unwrapping)
```

pwc3:
I ran into the same thing. Adding a typealias for the block type works around the issue, but in this case, I just added a `// swiftlint:disable:this force_unwrapping` comment.

Issue #977, opened 2016-12-13
Force Unwrapping false positive


AliBilal:
Declaring a local variable inside a function
`var a: [Int]!`
is considered force_unwrapping violation. But we are just defining an implicitly unwrapped optional and not force unwrapping it. Seems it is a false positive. It only occurs for Array type and not plain types. e.g. `var a: Int!` is not considered rule violation.

JamieEdge:
Which version are you using?

I cannot reproduce this with the following example:

![Example](https://cloud.githubusercontent.com/assets/1623997/21456779/75fd39c0-c922-11e6-86fb-d58be2e69e74.png)

AliBilal:
I'm currently using version 0.13.2

Please make sure to define the variable inside a function or method. For example:

`func foo() {
    var x: [Int]!
}`

The `force_unwrapping` rule is an opt-in rule. So please also make sure it is opted in.

marcelofabri:
I was able to reproduce it on `master`.

JamieEdge:
Sorry, moved my test files but not the config! Can now reproduce the issue.

marcelofabri:
I think this is actually a consequence of https://github.com/realm/SwiftLint/issues/136.

```bash
$ sourcekitten structure --text "func foo() { var x: [Int]! }"
{
  "key.substructure" : [
    {
      "key.kind" : "source.lang.swift.decl.function.free",
      "key.offset" : 0,
      "key.nameoffset" : 5,
      "key.namelength" : 5,
      "key.bodyoffset" : 12,
      "key.bodylength" : 15,
      "key.accessibility" : "source.lang.swift.accessibility.internal",
      "key.length" : 28,
      "key.name" : "foo()"
    }
  ],
  "key.offset" : 0,
  "key.diagnostic_stage" : "source.diagnostic.stage.swift.parse",
  "key.length" : 28
}
```

marcelofabri:
Fixed in #1620.

PR #1584, opened 2017-05-29
Add missing triggering cases involving subscript in force_unwrapping …

otaviolima:
…rule

Additionally fixes false positives

Fixes: #614, #977


marcelofabri:
Merged in #1620 🎉

PR #1591, opened 2017-05-31
Automatically generate LinuxMain.swift using Sourcery

jpsim:
This adds 6 tests that were accidentally not being run on Linux:

* LineLengthConfigurationTests.testLineLengthConfigurationInitialiserSetsIgnoresComments
* LineLengthConfigurationTests.testLineLengthConfigurationInitialiserSetsIgnoresFunctionDeclarations
* LineLengthConfigurationTests.testLineLengthConfigurationThrowsOnBadConfigValues
* LineLengthRuleTests.testLineLengthWithIgnoreCommentsEnabled
* LineLengthRuleTests.testLineLengthWithIgnoreFunctionDeclarationsEnabled
* RegionTests.testSeveralRegionsFromSeveralCommands

PR #1592, opened 2017-06-01
Use Sourcery to generate LinuxMain.swift & MasterRuleList.swift

jpsim:
Extends #1591 by also automatically generating `MasterRuleList.swift`


Source/SwiftLintFramework/Models/MasterRuleList.swift
```diff
@@ -100,17 +33,17 @@ public let masterRuleList = RuleList(rules:
     FileHeaderRule.self,
     FileLengthRule.self,
     FirstWhereRule.self,
+    ForWhereRule.self,
```
marcelofabri: shouldn't we keep the sort order as case insensitive?

jpsim: Sourcery returns these already sorted, and I don't think there's a way to re-sort. Is this a problem?

Makefile
```diff
@@ -28,6 +28,16 @@ VERSION_STRING=$(shell /usr/libexec/PlistBuddy -c "Print :CFBundleShortVersionSt
 all: bootstrap
 	$(BUILD_TOOL) $(XCODEFLAGS) build
 
+sourcery:
+	sourcery --sources Tests --templates .sourcery/LinuxMain.stencil --output .sourcery
```
masters3d: If you add to the top of the template the target file, it will override it instead of creating .generated.swift 
`// sourcery:file:Tests/LinuxMain.swift` 
https://twitter.com/ilyapuchka/status/847567291302674434
https://oleb.net/blog/2017/03/keeping-xctest-in-sync/

jpsim: Thanks for the tip! Although I need to shuffle things around to move the auto-generated Sourcery comments after the header anyways, so it doesn't really help in my case.

Source/SwiftLintFramework/Models/MasterRuleList.swift
marcelofabri: no, just wanted to see if it was possible to do it 👍

SwiftLint.xcodeproj/project.pbxproj
```diff
@@ -1216,6 +1220,20 @@
...
+			files = (
+			);
+			inputPaths = (
+			);
+			name = "Run Sourcery";
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "if which sourcery >/dev/null; then\n  make sourcery\nelse\n  echo \"Sourcery not found, install with 'brew install sourcery'\"\nfi";
```
marcelofabri: I know this is a minor issue but:
1. Shouldn't we avoid running `sourcery` on CI?
2. What do you think about making this message a warning?

jpsim: 1. Travis doesn't appear to have Sourcery installed.
2. It shouldn't be an error (or a warning even) if you don't have Sourcery installed but you're building SwiftLint via Xcode. It's only really useful for contributors.

marcelofabri: 👍

Feel free to merge it 🎉

jpsim: Yup, just fixing up the merge conflict now.

PR #1599, opened 2017-06-02
Problem occur also when using lowercased()

felipeplets:


marcelofabri:
Thanks for the pull request, but I don't see how this example is different than the existing one. As far as SourceKit is concerned, the method could even be `.foo()`.

felipeplets:
@marcelofabri thanks for the feedback, it was a lack of understanding from my side on SwiftLint definitions.

marcelofabri:
No worries, I hope you keep contributing 💯

PR #1600, opened 2017-06-02
Enable prohibited_super_call and overridden_super_call rules

marcelofabri:
I don't see why these are not enabled by default - they seem too useful to be opt-in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ##### Breaking
 
-* None.
+* `prohibited_super_call` and `overridden_super_call` rules are 
+   now enabled by default.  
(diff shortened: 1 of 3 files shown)


jpsim:
They're opt-in because we don't validate inheritance. For example, the rule triggers a false-positive violation for this:

```
open class Yolo {
  open func viewDidLoad() {}
}

class NotAVC: Yolo {
  override func viewDidLoad() { /* don't call super */ }
}
```

jpsim:
idk, maybe that's a weak argument though since doing this might be rare enough in practice.

marcelofabri:
Fair enough. Maybe it's worth merging it anyway since these rules can provide much value in the general case, even giving some false positives?

PR #1604, opened 2017-06-03
Fix using incorrect config file

### LINKED ISSUE #1531: Custom config not loaded properly
When linting with custom config file while `.swiftlint.yml` exists in current directory, `disabled_rules` are loaded from `.swiftlint.yml` instead of passed config file.

## How to reproduce (0.18.1):
In folder containing
```
# .swiftlint.yml
# empty
```
```
# .swiftlint-ci.yml
disabled_rules:
  - todo
```
```
// Foo.swift
// TODO
```
Run `swiftlint lint --config .swiftlint-ci.yml`:

>Loading configuration from '.swiftlint-ci.yml'
Linting 'Foo.swift' (1/1)
Foo.swift:1:4: warning: Todo Violation: TODOs should be avoided. (todo)

jpsim:
Fixes #1531.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,12 @@
   [#977](https://github.com/realm/SwiftLint/issues/977)
   [#1614](https://github.com/realm/SwiftLint/issues/1614)
 
+* Fix incorrectly using configuration files named `.swiftlint.yml` when they are
+  located in the same directory as a differently-named, user-provided custom
+  configuration file.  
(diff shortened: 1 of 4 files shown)


jpsim:
I actually think this fix is wrong, or at least incomplete. It won't work if the specified config isn't in the root path (e.g. `swiftlint lint --config src/.swiftlint-ci.yml`).

marcelofabri:
@jpsim I think you're right (see #1631 and the fix in #1644)

Issue #827, opened 2016-10-05
Fix empty_count false-positives


jhanzo:
When using a NSMapTable variable and test number of elements with count statement, Empty Count Violation is thrown whereas it should not. Indeed NSMapTable hasn't any isEmpty property.

XCode 7.3.1, iOS 9.3, swiftlint 0.11.1

<img width="1085" alt="capture d ecran 2016-10-05 a 17 55 07" src="https://cloud.githubusercontent.com/assets/10937985/19121002/371a472e-8b25-11e6-821d-7893750df634.png">

jhanzo:
A workaround would be using : 
```swift
mapTable.objectEnumerator()!.allObjects.isEmpty
```

masters3d:
How about 

```swift
extension NSMapTable {
    var isEmpty:Bool { return count == 0 }
}
```

jhanzo:
Of course, but as a workaround. Because of Swiftlint is an integration tool, you could expect to not being required to add some extra code.

marcelofabri:
Note that this rule is opt-in because of the false positives. You can always disable it.

There's no easy way to know a variable type without compiling the code (as far as I know) and SwiftLint does not compile any code currently.

jpsim:
@marcelofabri is exactly right, the reason why the `empty_count` rule is opt-in is due to these false positives, which can't be avoided unless SwiftLint was able to resolve the type of instances on which `.count` is being accessed, which would be possible with a new mode we're considering for SwiftLint described in https://github.com/realm/SwiftLint/pull/829#issuecomment-263008789.

Until then, be aware of the false-positives, and don't use opt-in rules unless you're aware of their tradeoffs and find them acceptable.

jonasman:
> This is by design as the rule should be able to catch usages of count inside Array extensions for example. See #827 (comment) for more details.

Well i think we make the empty_rule useless just to support an edge case in extensions.
I had to add many disable comments to make it to work and in the end we disabled the rule all together.

DagAgren:
From issue #1080, the rule is triggered on `if count > 0 {`. This does not seem like it should ever be correct. This is not accessing a property, it's using a variable. The check should require that `count` is preceded by a period.

(This would miss cases in subclasses that implement count, but this seems a much smaller corner case, and the ability to use `count` as a variable name should take priority.)

jpsim:
This is intentional, as extensions to collection types implementing `isEmpty` should also prefer using it over comparing `count` to zero.

[...]

PR #1608, opened 2017-06-07
Fixed empty_count rule by adding dot char to regex

danielbyon:
This PR fixes https://github.com/realm/SwiftLint/issues/827 by changing the regex to check for `.count` instead of just `count`.

By only checking for "count", normal variables named `count` comparisons were resulting in false positives.


marcelofabri:
I'm closing this, feel free to reopen/create a new PR if you implement the fix suggested on #827.

PR #1609, opened 2017-06-07
Fixed UnusedClosureParameterRule crash

woosiki:
If closure parameter is unicode, it calculate length 1, but its real length is 3. 
So when running byteRangeToNSRange, it accured crash in String+SourceKitten.swift:131.

screenshot

### DIFF
--- a/Source/SwiftLintFramework/Rules/UnusedClosureParameterRule.swift
+++ b/Source/SwiftLintFramework/Rules/UnusedClosureParameterRule.swift
@@ -50,7 +50,8 @@ public struct UnusedClosureParameterRule: ASTRule, ConfigurationProviderRule, Co
             "[1, 2].something { number, ↓idx in\n return number\n}\n",
             "genericsFunc { (↓number: TypeA, idx: TypeB) in return idx\n}\n",
             "hoge(arg: num) { ↓num in\n" +
-            "}\n"
+            "}\n",
+            "fooFunc { 아 in\n }"
         ],
         corrections: [
             "[1, 2].map { ↓number in\n return 3\n}\n":
@@ -118,7 +119,7 @@ public struct UnusedClosureParameterRule: ASTRule, ConfigurationProviderRule, Co
                 return nil
             }
 
-            let paramLength = name.bridge().length
+            let paramLength = name.lengthOfBytes(using: .utf8)
 
             let matches = regex.matches(in: file.contents, options: [], range: range).ranges()
             for range in matches {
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Rules/UnusedClosureParameterRule.swift
```diff
@@ -118,7 +119,7 @@ public struct UnusedClosureParameterRule: ASTRule, ConfigurationProviderRule, Co
                 return nil
             }
 
-            let paramLength = name.bridge().length
+            let paramLength = name.lengthOfBytes(using: .utf8)
```
marcelofabri: a better fix would be using `let paramLength = param.length` in the `guard`

Tests/SwiftLintFrameworkTests/UnusedClosureParameterRuleTests.swift
```diff
@@ -0,0 +1,26 @@
...
+//  Copyright © 2017 Realm. All rights reserved.
+//
+
+import Foundation
+@testable import SwiftLintFramework
+import XCTest
+
+class UnusedClosureParameterRuleTests: XCTestCase {
+
+    func testDefaultConfiguration() {
```
marcelofabri: this is already tested on `testUnusedClosureParameter` in `RulesTests`

Source/SwiftLintFramework/Rules/UnusedClosureParameterRule.swift
```diff
@@ -50,7 +50,8 @@ public struct UnusedClosureParameterRule: ASTRule, ConfigurationProviderRule, Co
             "[1, 2].something { number, ↓idx in\n return number\n}\n",
             "genericsFunc { (↓number: TypeA, idx: TypeB) in return idx\n}\n",
             "hoge(arg: num) { ↓num in\n" +
-            "}\n"
+            "}\n",
+            "fooFunc { 아 in\n }"
```
marcelofabri: you should use the violation marker (`↓`) to test the offset as well

marcelofabri:
Also, could you add a CHANGELOG entry?

Source/SwiftLintFramework/Rules/UnusedClosureParameterRule.swift
marcelofabri: this actually doesn't work 😅

marcelofabri:
Opened #1619 to continue this. Thanks @woosiki for raising the issue and coming up with the fix! 💯

PR #1621, opened 2017-06-14
Update LinuxMain.swift

marcelofabri:


jpsim:
Anything we could've done to prevent this from slipping?

marcelofabri:
A Danger check and/or a test that checks if the counts are equal (as described here: https://oleb.net/blog/2017/03/keeping-xctest-in-sync/)

PR #1623, opened 2017-06-14
Fix `empty_count` false positive in words with "count"

### LINKED ISSUE #1622: EmptyCountRule catches words that include "count"
[EmptyCountRule.swift:34](https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules/EmptyCountRule.swift#L34)

`let pattern = "count\\s*(==|!=|<|<=|>|>=)\\s*0"`

This should change so that it doesn't match words that _end_ in `count`, only the word `count` itself.

ie. `discount == 0`

marcelofabri:
Fixes #1622

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,10 @@
   [JP Simard](https://github.com/jpsim)
   [#1531](https://github.com/realm/SwiftLint/issues/1531)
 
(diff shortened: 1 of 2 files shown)


jpsim:
🎉

PR #1644, opened 2017-06-24
Fix configuration issue when using custom paths

### LINKED ISSUE #1631: When --path is used, the directory's .swiftlint.yml is ignored
When using SwiftLint 0.20.0, when specifying a path (that is not the current directory) with `--path`, that directory's `.swiftlint.yml` is ignored.

To reproduce it, just create the following 2 files in a subdirectory.

```yml
# subdir/.swiftlint.yml
disabled_rules:
  - todo
```

```swift
// subdir/MyCode.swift
// TODO: ABCDEFG
```

Specifying the path causes the directory's `.swiftlint.yml` to be ignored.
```
$ swiftlint lint --no-cache --quiet # Not path specified, no warning as expected
$ swiftlint lint --no-cache --quiet --path subdir # Specifying the path gets you an unexpected warning
.../subdir/MyCode.swift:2:4: warning: Todo Violation: TODOs should be avoided (ABCDEFG). (todo)
```

marcelofabri:
Fixes #1631 

// @vincentisambart could you try this to be sure it fixes the issue?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1630](https://github.com/realm/SwiftLint/issues/1630)
 
+* Use the directory's .swiftlint.yml when `--path` is used.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#1631](https://github.com/realm/SwiftLint/issues/1631)
+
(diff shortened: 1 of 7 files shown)


vincentisambart:
@marcelofabri I just tried on my machine and it does seem to fix the problem for me, thank you!

PR #1645, opened 2017-06-26
Separating level counting for types and functions in `nesting` rule.

### LINKED ISSUE #1151: Teach NestingRule to ignore types declared in functions
The idea behind types in the nesting rule is to not have a complex interface, but that doesn't apply to types scoped within a single function, such as this example:https://github.com/jpsim/SourceKitten/blob/0.16.0/Source/SourceKittenFramework/SourceDeclaration.swift#L74

Skoti:
1. Separated nesting level counting for types and functions in `nesting` rule (fixes #1151).
2. Enhanced `nesting` rule to search for nested types and functions within closures and statements.
3. Enhanced `nesting` rule to allow for one nested type within a function even if breaking a maximum type level nesting (fixes #1151).

**-> 1**
The previous implementation used one nesting level counter for both types and functions which was producing inconsistent and unexpected behaviours like:
- This was ok (1 nested type):
```
class Example_0 {
    class Example_1 {}
}
```
- This was an error (with default configuration - type is nested at level 2 instead of maximum 1), even though there is only one type (should be type level: 0):
```
func f_0() {
    func f_1() {
        class Example_0 {}
    }
}
```
So if `statement_level` was set to i.e. 0 (no function nesting) then only free global functions could have been written (due to any `type` being counted as a nesting level for functions), so these two level settings could be set to be mutually excluding.

The new implementation counts types and functions separately, so it only matters how many types is nested, no matter if there were some functions in between. The same goes for functions.
This also allows to lower the maximum function level from the default 5 to 2. The previous implementatin must have taken into account the possible first type/extension and the second type.

In the previous implementation `function_level` was called `statement_level` although in `SwiftDeclarationKind` there is no statement cases, so I renamed it to `function_level` to be more appropriate to what it does.

**-> 2**
The previous implementation did not cover searching for nested types and functions within closures and statements (like switch, for, etc.).
This is a new configuration option called `check_nesting_in_closures_and_statements` and it is turned on by default.

**-> 3**
I've also added another configuration option called `always_allow_one_type_in_functions` if one wants to keep the type nesting level constraint, but still be able to declare a type within a function even if breaking the maximum type nesting level. Such type is only accessible in that function so it makes sense to allow for it if needed.
Still the nesting tree is constrained by the maximum function nesting level.
This defaults to false.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,17 +17,21 @@
   [JP Simard](https://github.com/jpsim)
   [#3412](https://github.com/realm/SwiftLint/issues/3412)
 
-* Renamed `statement_level` to `function_level` in `nesting` rule configuration.  
+* Renamed `statement_level` to `function_level` in `nesting` rule
+  configuration.  
   [Skoti](https://github.com/Skoti)
 
-* Separated `type_level` and `function_level` counting in `nesting` rule.  
+* Separated `type_level` and `function_level` counting in `nesting`
+  rule.  
   [Skoti](https://github.com/Skoti)
   [#1151](https://github.com/realm/SwiftLint/issues/1151)
 
 * `function_level` in `nesting` rule defaults to 2 levels.  
   [Skoti](https://github.com/Skoti)
 
-* Added `check_nesting_in_closures_and_statements` in `nesting` rule to search for nested types and functions within closures and statements. Defaults to `true`.  
+* Added `check_nesting_in_closures_and_statements` in `nesting` rule to
+  search for nested types and functions within closures and statements.
+  Defaults to `true`.  
   [Skoti](https://github.com/Skoti)
 
 #### Experimental
@@ -41,6 +45,9 @@
   [Artem Garmash](https://github.com/agarmash)
   [#3286](https://github.com/realm/SwiftLint/issues/3286)
 * Added `always_allow_one_type_in_functions` option in `nesting` rule configuration. Defaults to `false`. This allows to nest one type within a function even if breaking the maximum `type_level`.  
+* Added `always_allow_one_type_in_functions` option in `nesting` rule
+  configuration. Defaults to `false`. This allows to nest one type
+  within a function even if breaking the maximum `type_level`.  
   [Skoti](https://github.com/Skoti)
   [#1151](https://github.com/realm/SwiftLint/issues/1151)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,19 @@
   [JP Simard](https://github.com/jpsim)
   [#3412](https://github.com/realm/SwiftLint/issues/3412)
 
+* Renamed `statement_level` to `function_level` in `nesting` rule configuration.  
+  [Skoti](https://github.com/Skoti)
+
+* Separated `type_level` and `function_level` counting in `nesting` rule.  
+  [Skoti](https://github.com/Skoti)
+  [#1151](https://github.com/realm/SwiftLint/issues/1151)
+
+* `function_level` in `nesting` rule defaults to 2 levels.  
+  [Skoti](https://github.com/Skoti)
+
+* Added `check_nesting_in_closures_and_statements` in `nesting` rule to search for nested types and functions within closures and statements. Defaults to `true`.  
+  [Skoti](https://github.com/Skoti)
+
 #### Experimental
 
 * None.
@@ -27,6 +40,9 @@
   `Never`.  
   [Artem Garmash](https://github.com/agarmash)
   [#3286](https://github.com/realm/SwiftLint/issues/3286)
+* Added `always_allow_one_type_in_functions` option in `nesting` rule configuration. Defaults to `false`. This allows to nest one type within a function even if breaking the maximum `type_level`.  
+  [Skoti](https://github.com/Skoti)
+  [#1151](https://github.com/realm/SwiftLint/issues/1151)
 
 #### Bug Fixes
 
--- a/Source/SwiftLintFramework/Extensions/SwiftDeclarationKind+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/SwiftDeclarationKind+SwiftLint.swift
@@ -34,4 +34,12 @@ extension SwiftDeclarationKind {
         .associatedtype,
         .enum
     ]
+
+    internal static let extensionKinds: Set<SwiftDeclarationKind> = [
+        .extension,
+        .extensionClass,
+        .extensionEnum,
+        .extensionProtocol,
+        .extensionStruct
+    ]
 }
--- a/Source/SwiftLintFramework/Rules/Metrics/NestingRule.swift
+++ b/Source/SwiftLintFramework/Rules/Metrics/NestingRule.swift
@@ -11,16 +11,16 @@ public struct NestingRule: ConfigurationProviderRule {
     public static let description = RuleDescription(
         identifier: "nesting",
         name: "Nesting",
-        description: "Types should be nested at most 1 level deep, " +
-        "and functions should be nested at most 2 levels deep.",
+        description:
+            "Types should be nested at most 1 level deep, and functions should be nested at most 2 levels deep.",
 		kind: .metrics,
 		nonTriggeringExamples: NestingRuleExamples.nonTriggeringExamples,
 		triggeringExamples: NestingRuleExamples.triggeringExamples
     )
 
-    private let omittedStructureKinds: [SwiftStructureKind] =
-        [.declaration(.enumcase), .declaration(.enumelement)]
-        + SwiftDeclarationKind.variableKinds.map { .declaration($0) }
+    private let omittedStructureKinds = SwiftDeclarationKind.variableKinds
+        .union([.enumcase, .enumelement])
+        .map(SwiftStructureKind.declaration)
 
     private struct ValidationArgs {
         var typeLevel: Int = -1
@@ -144,19 +144,6 @@ private enum SwiftStructureKind: Equatable {
             self = .statement(statementKind)
         } else {
             return nil
-        }
-    }
-
-    static func == (lhs: SwiftStructureKind, rhs: SwiftStructureKind) -> Bool {
-        switch (lhs, rhs) {
-        case let (.declaration(lhsKind), .declaration(rhsKind)):
-            return lhsKind == rhsKind
-        case let (.expression(lhsKind), .expression(rhsKind)):
-            return lhsKind == rhsKind
-        case let (.statement(lhsKind), .statement(rhsKind)):
-            return lhsKind == rhsKind
-        default:
-            return false
         }
     }
 }
--- a/Source/SwiftLintFramework/Rules/Metrics/NestingRule.swift
+++ b/Source/SwiftLintFramework/Rules/Metrics/NestingRule.swift
@@ -1,10 +1,10 @@
 import SourceKittenFramework
 
-public struct NestingRule: ASTRule, ConfigurationProviderRule, AutomaticTestableRule {
+public struct NestingRule: ConfigurationProviderRule {
     public var configuration = NestingConfiguration(typeLevelWarning: 1,
                                                     typeLevelError: nil,
-                                                    statementLevelWarning: 5,
-                                                    statementLevelError: nil)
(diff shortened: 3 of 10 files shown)


Skoti:
@marcelofabri, @jpsim 
Hi, could someone please review this pull request? I think it is important to improve and fix the nesting rule.

marcelofabri:
@Skoti could you please rebase this first? 😅

Skoti:
@marcelofabri 
Sure. It's done. There was just one conflict 😄

Skoti:
@marcelofabri 
Any update? 😺

Skoti:
@jpsim, @marcelofabri 
Hi ;)
Could one of you guys please review this? I think this is an important update to the nesting rule.
I'm rebasing like for the third time :P, fortunately only small conflicts ;)

jpsim:
This is kind of moving away from the original intended purposes of the rule.

Purpose 1: limit pyramid-of-doom style indentation
Purpose 2: avoid declaring overly-nested APIs, which hurt their discoverability

Skoti:
@jpsim 
I don't know how did you infer that this is "moving away from the original intended purposes of the rule" but this PR just enhances the current implementation and fixes an incorrect level counting.

Purpose 1: still true, I have even extended it to count type/function nesting level within closures and statements (like switch, for, etc.) => more discoverability, as described in point 2 of my first message.

Purpose 2: still true, additionally now it correctly count nesting level of types and functions and it works more efficiently due to top-bottom approach instead of a bottom-up where same nested levels where visited several times.

To be more precise the previous implementation used one nesting level counter for both types and functions even though its settings allowed to configure it differently (point 1 in my first message). This could easily lead to mutually excluding settings which obviously must be avoided.

If you would like to see this rule to stay with just one counter then only one setting should be available, I don't see the point of two different settings for types and functions if they just don't work correctly in that case.
So either this rule should stay with one counter and shouldn't differentiate between types and functions or it should have two counters to do this correctly what I have introduced in this PR.

I assumed that if nesting rule already has two different settings for types and functions then it must count it separately and in my view this was the purpose of the nesting rule from the very beginning and this PR does not change it. 
Consequently, the pyramid of doom is still detectable. Mind that in this example:

```
func a {
    func a1 {
         struct B {
             func b1 {}
        }
    }
}
```
function nesting level is 2 - there are two functions (b1 and a1) nested in function a, the type is not resetting it, counting is done separately.

Dschee:
@Skoti I know this is an old one and I haven't looked through the entire code, but from my first cursory look I had the feeling that this changes multiple things and discussion might get stuck due to some maintainers not completely aggreeing with one of the changes. Before I have a more detailed look, would you mind separating this into separate PRs so we can merge each aspect of the improvements separately? 

Also having a rationale in the PR will make it easier to understand what this really is about, for example you could provide one previously triggering example and tell why it might not be an expected behavior.

Please note that one thing I can already say regarding the implementation is your usage of string concatenation – could you use multi-line strings instead? For example instead of this:

```swift
"var example: Int = 5 {\n" +
"   didSet {\n" +
"       \(type) Example_0 {\n" +
"           \(type) Example_1 {}\n" +
"       }\n" +
"   }\n" +
"}"
```

You should better use this:

```swift
"""
var example: Int = 5 {
   didSet {
       \(type) Example_0 {
           \(type) Example_1 {}
       }
   }
}
"""
```

Skoti:
@Dschee 
First of all, thanks for taking care of this old pull request! 👍 

Regarding multiline strings - no problem, I will update them in the meantime. 😄 

Now make yourself comfortable 💺 and grab some ☕️ - this will take a while ⏳.
<br/>
> Before I have a more detailed look, would you mind separating this into separate PRs so we can merge each aspect of the improvements separately?

This PR introduces 4 changes:
* top-bottom approach (main change)
* two separate counters (main change - [point 1](https://github.com/realm/SwiftLint/pull/1645#issue-127575068))
* search for nested types and functions within closures and statements (extension to main - [point 2](https://github.com/realm/SwiftLint/pull/1645#issue-127575068))
* allow one nested type within a function (extension to main - [point 3](https://github.com/realm/SwiftLint/pull/1645#issue-127575068))

 **All** of them **depend** on the fact that the rule is now processing a file starting at the root level of the file and not at the most nested part of it (top-bottom approach). Without it the code was less efficient and was making logic for checking nesting level more complicated. 🍝 The current approach (not in this PR) is bottom-up and it is due to the fact of the nesting rule being `ASTRule` and its default `validate` implementation.

So any division we wanted to make would need to include this top-bottom approach change in each smaller PR - otherwise it is a bit of work ⚙️to adjust it to the current approach - and that I would like to avoid without first trying to help you understand this PR changes.
Apart from that there is only one other main change (which probably is the most not clear) regarding two separate counters - and I do not see an advantage of separating this **now** as remaining extensions are fairly easy to add/remove.
If we agree to reuse only some of the code changes or ideas then I can create a new, clear PR - no problem. 🚀 Until then I would prefer to stay with this PR and have a discussion in here. 💬
<br/>

> this changes multiple things and discussion might get stuck due to some maintainers not completely aggreeing with one of the changes

For now, none of the maintainers really explained what is the change they do not like so I am not sure what I should further explain or change. 🤷‍♂️ 
That is why I would prefer to establish what is desirable and what is not and then either update this PR or create a new one with accepted changes only. For instance: implementing top-bottom approach and adding just one extension _without_ introducing main change of two separate counters (point 1). ✂️ ♻️ Although, I will try to explain and justify this change as well :)
<br/>

> Also having a rationale in the PR will make it easier to understand what this really is about, for example you could provide one previously triggering example and tell why it might not be an expected behavior.

I will try to explain it using different examples below. I think the best way to start is that we have a common understanding of the rationale of this PR. 💯 
Apart from the examples below please look at the triggering/non-triggering examples and my explanation/description in the first post.

After all, it would be awesome 🌈 if you could ask some questions or at least point out the things not clear enough, otherwise I can only repeat myself using slightly other words and examples and we will go nowhere. 🛑 
<br/>
## Examples ##
<details><summary>Click to expand</summary><br/>

So let's say we want to allow types to be nested at level 1 (default) and we disallow function nesting. In other words - if we declare a type it can **only** have **flat** functions.
Here is where the first problem emerges. 🌔 If we want to disallow function nesting we might think of using `0` in a configuration but it won't work because we need to account for a type being the first nesting level and we are _forced_ to use `1`. Configuration is not adjusted to a use case.
A sample configuration for current implementation of nesting rule (the one not in this PR) would look like this:

<pre lang="yaml">
nesting:
  type_level:
    warning: 1 # default
  statement_level:
    warning: 1 # default is 5
</pre>
Let's analyze an example below:
<pre lang="swift">
class A {
    func fa1() {
        func fa2() {
        }
    }
}
func f() {
    func f1() {
    }
}
</pre>

This gives us one warning `Nesting Violation: Statements should be nested at most 1 level deep (nesting)` for `func fa2`. This is **correct** as we disallowed function nesting.
⚠️ Unfortunately, a free function below with a nested function is **not triggering** a warning. _Why?_ Because we were _forced_ to account for this one type and now it accidentally allows function nesting in a free function - which is what we wanted to restrict at the very beginning.

🔑 This happens because current nesting rule implementation is oriented towards what is the nesting level of that type/function (in general) instead of how many nested declarations there are of that **particular** kind (type or function). Changes introduced in this PR solve this problem - so from one global nesting counter we have global nesting counters per kind (type/functions).

---
Another problem 🌓  is that we only accounted for one type while it is also allowed to nest types too.
So this code:
<pre lang="swift">
class A {
    class B {
        func fb1() {
        }
    }
}
</pre>
gives a warning for `func fb1`. So we have a type and we cannot declare anything in it 😞 
To fix this we are again _forced_ to increase the nesting level for statements...
This is getting even more wild 🐗  if we allow further type or function nesting. 

🔑 At this point I think you can imagine that if we allowed to nest functions in a nested type (level 3 at least) then we would need to gradually increase the statement warning level which leads to side effects in a form of too many nested functions in parent types (or free functions) - and the _pyramid of doom_ is not detectable.
With statement warning level `3`, code below would **not** yield a warning:
<pre lang="swift">
class A {
    func fa1() {
        func fa2() {
            func fa3() {
            }
        }
    }
}
</pre>
and some would consider this a bit too complicated and would prefer the linter to warn about it - but the nesting rule configuration is not flexible enough.

🔑 This whole dependency between nesting level of types and statements does not make them easy and clear in use. These two settings are **tightly coupled** 🔗 thus you are always forced to account for nested types which makes `statement_level` unusable.
It all comes down to use the formula below to set statement warning level to **at least**:
`{type_level warning} + 1 + {number of nested functions allowed in the most nested type}`
otherwise it will cause the problems described above.

This is making the nesting rule totally unusable for me and I have it disabled in all projects. Simple case to allow for two nested types already requires `3`, `4` or `5` nesting levels for statements and there is no point in increasing these numbers as they already allow for creating a nasty _pyramid of doom_. The configuration is stiff.

---
That is why this PR introduces separate counters. So let's compare **minimal** configurations for slightly exaggerated examples below:
<table>
<tr><th>Example</th><th>Current configuration</th><th>New configuration</th></tr>
<tr><td><pre lang="swift">
class A {
    func fa1() {
        enum E {
            func fe1() {
                func fe2() {
                }
            }
        }
    }
}
</pre></td><td><pre lang="yaml">
nesting:
  type_level:
    warning: 2
  statement_level:  
    warning: 4
</pre></td><td><pre lang="yaml">
nesting:
  type_level:
    warning: 1
  function_level: 
    warning: 2
</pre></td></tr><tr><td><pre lang="swift">
class A {
    struct B {
        func fb1() {
            enum E {
                func fe1() {
                }
            }
        }
    }
}
</pre></td><td><pre lang="yaml">
nesting:
  type_level:
    warning: 3
  statement_level:
    warning: 4
</pre></td><td><pre lang="yaml">
nesting:
  type_level:
    warning: 2
  function_level:      
    warning: 1
</pre></td></tr></table>

So current implementation is more lenient and allows for more side effects than the one in this PR.

One little 🔍 side effect of the **new** implementation is that there is no constant number of functions that can be nested in a type as it depends on whether this type was already declared in a function or not. However, this might also be treated as a **precaution** against _pyramid of doom_.

---
I hope this wall of text did not scared you and you have reached this point 😃 💪. I tried to make it easier by adding some emoji 🤣 
</details>

Skoti:
Hi @Dschee, any chance on getting back to this?

Dschee:
Hi @Skoti, sorry that I didn't have the time to read your comment earlier. I just did though and I feel like we should ask the original author of the Nesting rule for his opinion on this as it changes its idea. I could find [this](https://github.com/realm/SwiftLint/commit/d5914edbaba17fdab21f2e8465c7b374949a6bc9) commit pointing to @jpsim as the original author. Maybe he can give some feedback?

Skoti:
> This issue has been automatically marked as stale because it has not had any recent activity. Please comment to prevent this issue from being closed. Thank you for your contributions!

It is still an issue.

jpsim:
My motivation for the nesting rule was to avoid deeply nested _public_ types in a module's _public_ API. That's because as a user of modules, it get really unwieldy if I have to type `ParentA.ChildB.SubChildC.SubSubChildD(...)` and that should be a sign that this type might be overly nested. I think the same issues can manifest themselves for internal APIs too, but public APIs is where this is most cumbersome to do.

The way I wrote the rule was too simplistic and triggered violations for things that I'd consider totally valid and clean code, like having a function-local type, or a fileprivate/private deeply nested type that's only used in the file where it's defined.

Hopefully this is helpful context.

Skoti:
Thanks for explanation. Your motivation is clear and this PR upholds it :)

> The way I wrote the rule was too simplistic and triggered violations for things that I'd consider totally valid and clean code, like having a function-local type, or a fileprivate/private deeply nested type that's only used in the file where it's defined.

This is what I am also trying to address in this PR.

Is there any chance to move this PR forward?
I would be happy to address any remarks to the solution, or implement a totally different approach if that would be better.

jpsim:
> Is there any chance to move this PR forward?

First step would be rebasing to ease reviewing. Then tag one of the maintainers and hopefully one of us will have time soon to look at it more closely.

Skoti:
@jpsim it's done. I hope this will be reviewed 🚀

jpsim (approved):
Thanks for the PR! Will merge when CI passes.

PR #1655, opened 2017-07-03
[WIP] Inferred sorting rule

marcelofabri:
Continued from  #1050

// cc @JamieEdge


marcelofabri:
This should have triggered the `Rules.md` warning in Danger 🤔

Dangerfile
```diff
@@ -3,15 +3,17 @@ require 'open3'
...
-has_danger_changes = !git.modified_files.grep(/Dangerfile|script\/oss-check|Gemfile/).empty?
-has_rules_changes = !git.modified_files.grep('/Source/SwiftLintFramework/Rules/').empty?
-has_rules_docs_changes = !git.modified_files.grep('Rules.md').empty?
+has_app_changes = !modified_files.grep(/Source/).empty?
+has_test_changes = !modified_files.grep(/Tests/).empty?
+has_danger_changes = !modified_files.grep(/Dangerfile|script\/oss-check|Gemfile/).empty?
+has_build_changes = !modified_files.grep(/Makefile|SwiftLint\.xcodeproj|SwiftLint\.xcworkspace|Package\.swift|Cartfile/).empty?
+has_danger_changes = !modified_files.grep(/Dangerfile|script\/oss-check|Gemfile/).empty?
+has_rules_changes = !modified_files.grep(/Source\/SwiftLintFramework\/Rules/).empty?
+has_rules_docs_changes = !modified_files.grep(/Rules.md/).empty?
```
jpsim: you need to escape this period: `/Rules\.md/`

marcelofabri:
@jpsim What do you think about this? I'm afraid that in practice this rule is not as helpful as I've thought it would be.

masters3d:
I think this rule is not too helpful when applied to arrays.  
It would be somewhat helpful when used on dictionary literals sorted by the key. Another place would be Enum cases.

PR #1656, opened 2017-07-03
Add private_over_fileprivate and strict_fileprivate rules

### LINKED ISSUE #1469: [Rule Request] Make all top level fileprivate decl become private

marcelofabri:
Continued from #1489. Adds `private_over_fileprivate` correctable rule and `strict_fileprivate` opt-in rule.

Fixes #1469, #1058

// cc @masters3d


marcelofabri:
In the end, I think it's better to go with two different rules so it's easier to understand what each of them validates (with examples).

masters3d:
Good call.

PR #1658, opened 2017-07-04
enabled extension_access_modifier

masters3d:
split from #1606

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1078](https://github.com/realm/SwiftLint/issues/1078)
 
+* Enable extension_access_modifier rule.  
+  [masters3d](https://github.com/masters3d)
+
 * Add `private_over_fileprivate` correctable rule to check for top-level usages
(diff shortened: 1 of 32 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,9 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1078](https://github.com/realm/SwiftLint/issues/1078)
 
+* Enable extension_access_modifier rule.
```
marcelofabri: I think this can be confusing, since it's not something that SwiftLint users care about. Can you remove the entry? 😬

Source/swiftlint/Commands/AutoCorrectCommand.swift
```diff
@@ -10,7 +10,7 @@ import Commandant
 import Result
 import SwiftLintFramework
 
-struct AutoCorrectCommand: CommandProtocol {
+internal struct AutoCorrectCommand: CommandProtocol {
```
marcelofabri: I'm not sure about adding `internal` explicitly here 😬 

// thoughts, @jpsim?

marcelofabri:
Closing in favor of #1701.

PR #1661, opened 2017-07-11
various minor edits since 0.19.0

jpsim:
I haven't kept up to date with SwiftLint since 0.19.0, so I decided to review what has gone in, and I had a few minor edits.

### DIFF
--- a/Source/SwiftLintFramework/Rules/VerticalParameterAlignmentOnCallRule.swift
+++ b/Source/SwiftLintFramework/Rules/VerticalParameterAlignmentOnCallRule.swift
@@ -84,7 +84,7 @@ public struct VerticalParameterAlignmentOnCallRule: ASTRule, ConfigurationProvid
                 return []
         }
 
-        var visitedLines: Set<Int> = []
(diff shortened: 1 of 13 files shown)


Source/SwiftLintFramework/Rules/VerticalParameterAlignmentOnCallRule.swift
```diff
@@ -84,7 +84,7 @@ public struct VerticalParameterAlignmentOnCallRule: ASTRule, ConfigurationProvid
                 return []
         }
 
-        var visitedLines: Set<Int> = []
+        var visitedLines = Set<Int>()
```
marcelofabri: Maube we should create rules for this? (To enforce both styles)

marcelofabri:
👏

Source/SwiftLintFramework/Rules/VerticalParameterAlignmentOnCallRule.swift
jpsim: I don't feel that strongly about it.

Issue #1660, opened 2017-07-10
Consider renaming rules_docs command


jpsim:
Underscores in POSIX commands aren't common. Maybe `generate-docs` or `rules-docs`?

For example, Carthage uses [`copy-frameworks`](https://github.com/Carthage/Carthage/blob/0.24.0/Source/carthage/CopyFrameworks.swift#L9) rather than `copy_frameworks` as a command.

marcelofabri:
I'm okay with either options, especially since this shouldn't be used by users in general.

jpsim:
How about just `docs`?

marcelofabri:
I prefer the other options. Just `docs` can be misleading IMO

jpsim:
Addressed in #1662

PR #1662, opened 2017-07-11
rename RulesDocsCommand to GenerateDocsCommand

jpsim:
`rules_docs` -> `generate-docs`

addresses #1660


jpsim:
Hmmm, odd that OSSCheck caught stuff...

Issue #1647, opened 2017-06-29
Set file_length rule only as warning, but still set custom length value 


Bartzy:
Hi,

I want to enable the file_length rule only as a warning, but still set a custom length value for it.
I tried several options but none worked.

Thanks!

marcelofabri:
@Bartzy have you tried setting the error limit to `null`?

Bartzy:
@marcelofabri 
Yes, and then I get this output at the beginning of `swiftlint`:
```
Loading configuration from '.swiftlint.yml'
Invalid configuration for 'file_length'. Falling back to default.
```

Bartzy:
Thanks!

PR #1664, opened 2017-07-11
Add support for optional 'error' severity level in configurations

JamieEdge:
Addresses https://github.com/realm/SwiftLint/issues/1647 and adds validation, so if an invalid value type (e.g. string) is provided in the config, this will now error.


marcelofabri:
Continued on #1691.

PR #1665, opened 2017-07-11
Fix Danger check for updated rules documentation

marcelofabri:


jpsim:
What's going on with OSSCheck's report here? 3 Warnings and 117 Messages... 🤔

jpsim:
Could you please rebase this so that OSSCheck runs properly?

marcelofabri:
Any ideas why this is still showing warnings? 🤔

jpsim:
Hmm, that's frustrating, I thought I had addressed the last OSSCheck issue 😞. I'll have to run locally to investigate.

jpsim:
Can't reproduce the issue locally... I might create a bit of noise as I test this "in production"

jpsim:
That shouldn't stop us from merging this PR though.

PR #1672, opened 2017-07-12
Fix false positive on redundant_discardable_let rule when using while statements

### LINKED ISSUE #1669: while let _ incorrectly changed to while _
Ran autocorrect and SwiftLint changed:
```
while let _ = SplashStyle(rawValue: maxValue) {
    maxValue += 1
}
```
to
```
while _ = SplashStyle(rawValue: maxValue) {
    maxValue += 1
}
```

Expected result: Keep it unchanged.

marcelofabri:
Fix #1669

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,10 @@
 
 ##### Bug Fixes
 
-* None.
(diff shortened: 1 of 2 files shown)


jpsim:
I believe the regressions reported in WordPress are due to #1670.

Issue #1589, opened 2017-05-31
Trigger implicit_getter even if the getter is mutating


RaphaelRO:
In case of a mutating getter, the `get` keyword must be specified to add the `mutating` behavior but this trigger the `implicit_getter` warning.

```swift
struct Counter {
  private var count: Int
  
  init(start: Int = 0) {
    self.count = start
  }
  
  var next: Int? {
    mutating get { // warning: Implicit Getter Violation: Computed read-only properties should avoid using the get keyword. (implicit_getter)
      defer {
        self.count += 1
      }
      
      return self.count
    }
  }
}
```

**Edit:** swiftlint version: 0.19.0

marcelofabri:
Similar to #1309

marcelofabri:
Implemented in #1678.

PR #1678, opened 2017-07-14
Don't trigger implicit_getter when attributes are present

### LINKED ISSUE #1309: Ignore @inline(__always) and other attributes in ImplicitGetterRule
For example: https://github.com/apple/swift/blob/75644d28f9908956f872110f04a1a935887b5e00/stdlib/public/core/Builtin.swift#L406:21

```swift
internal var _objCTaggedPointerBits: UInt {
  @inline(__always) get { return 0 }
}
```

marcelofabri:
Fixes #1309, #1589


jpsim:
🎉

Issue #1586, opened 2017-05-30
Rogue Custom Rules


tiagoalmeida:
There are certain rules that kill all custom rules. I call them Rogue Rules.

## Simple example

Let us start with a simple .swiftlint file, with only one customization:

```
vertical_whitespace: 
  max_empty_lines: 2
```

Imagine that you have a custom rule, like this one:

```
  force_https: # From https://github.com/Twigz/Game
    name: "Force HTTPS over HTTP"
    message: "HTTPS should be favored over HTTP"
    regex: "((?i)http(?!s))"
    match_kinds: string
    severity: warning
```

Then an awesome variable like this one:

`        _ = "http://thavariable.com"
`

This works. The variable produces a warning, everyone is happy.

## THE ROGUE RULE
The moment you bring the Rogue Rule, the panic starts, people running and fire emerges in other rules.

Let me introduce you to my Rogue Rule:

```
  two_enters_before_marks: # ¯\_(ツ)_/¯
    name: "2 Enters before Marks"
    message: "You need two enters before Marks"
    regex: '( *} *\n *\n *// MARK)' # I have no idea what I am doing ¯\_(ツ)_/¯
    severity: "warning"
```

So by now, you must be imagining "YOU ARE THE ROGUE THAT IS KILLING REGEX". Probably you are right. Yeah, I know I have a lot to learn with regex, but it would be cool that people that use websites like [Regex101](https://regex101.com/) and create Rogue Rules wouldn't destroy all the work ever done by humankind.

## Post Script Notes
PS: I am using the version 0.19.0. 
PS2: I imagine that is some error with that regex that destroys everything, but since I have no feedback is kinda hard to spot the error. At least an error on running the script would be cool.

Ok, enough is enough, I am going after that Rogue Rule.

PS3: I started to mess around with SwiftLint code and found out that this rule is created with a niled regex
PS4: Ideally, for me, as a SwiftLinter user it would give an error (like it currently happens when the YAML parser fails) when a custom rule is badly defined. Not saying anything and obliterate the other custom rules is kinda weird.

marcelofabri:
So, the actual issue here is that we should provide better feedback when a regex is invalid?

tiagoalmeida:
When I started the issue I had no idea what was happening.

But after investigating in the code, yes, the issue is related with invalid regex.

This actually leads to two different issues:

- When an invalid regex occurs SwiftLint starts to ignore all custom rules
[...]

PR #1681, opened 2017-07-15
Fix all custom rules not being applied when any rule is configured incorrectly

JamieEdge:
Addresses https://github.com/realm/SwiftLint/issues/1586 by adding handling.


marcelofabri:
Merged in #1683. Thanks @JamieEdge 🎉

PR #1682, opened 2017-07-16
[WIP] Warn if a configured rule is not enabled

### LINKED ISSUE #1350: Warn if a configured rule is not enabled
From https://github.com/realm/SwiftLint/issues/1291#issuecomment-285037132:

> Probably it would make it helpful if SwiftLint could display some kind of warning, when a configuration for opt-in rule is present, but the rule itself is "opted-out" (not enabled).

marcelofabri:
Fixes #1350


marcelofabri:
This will end up conflicting a lot with #1674 to be properly implemented, so closing for now.

PR #1683, opened 2017-07-17
Fix all custom rules not being applied when any rule is configured incorrectly

marcelofabri:
Continued from #1681

// cc @JamieEdge

### DIFF
--- a/.sourcery/LinuxMain.stencil
+++ b/.sourcery/LinuxMain.stencil
@@ -14,7 +14,7 @@ import XCTest
 {% for type in types.classes|based:"XCTestCase" %}
 extension {{ type.name }} {
     static var allTests: [(String, ({{ type.name }}) -> () throws -> Void)] = [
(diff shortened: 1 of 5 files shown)


marcelofabri:
Filled [SR-5477](https://bugs.swift.org/browse/SR-5477) 🙃 

Maybe we should just not run this test on Linux?

jpsim:
This is great, especially the bit to annotate tests to not run on Linux. 😊 

However, why not misconfigure a rule in a cross-platform compatible way, so that we can benefit from this test on Linux without hitting SR-5477 (thanks for filing that, btw).

marcelofabri:
🤔 

I didn't realize that there're other ways to fail inside `RegexConfiguration`. I'll open a new PR.

JamieEdge:
Thanks for adding the tests - sorry I didn't get to it.

PR #1684, opened 2017-07-17
Add unneeded_parentheses_in_closure_argument rule

### LINKED ISSUE #1483: Rule request: Unneeded parentheses in closure argument
```swift
// good
let foo = { (bar: Int) in
    print(bar)
}

// bad
call(arg: { (bar2) in

})
```

We could use a regex (`\{\s*\([^:]+\)\s*in`) to do it and also check for `source.lang.swift.decl.var.parameter` without a `key.typename`.

Stole the idea and regex from @keith

marcelofabri:
Fixes #1483

* Should this be a default rule?
* Should it be configurable (see https://github.com/realm/SwiftLint/issues/1483#issuecomment-298456009)?

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -23,6 +23,7 @@ opt_in_rules:
   - fatal_error_message
   - vertical_parameter_alignment_on_call
   - let_var_whitespace
+  - unneeded_parentheses_in_closure_argument
 
 file_header:
   required_pattern: |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,8 +59,8 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1647](https://github.com/realm/SwiftLint/issues/1647)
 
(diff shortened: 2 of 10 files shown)


Dschee:
I'm for making this default. I still don't understand why Xcode is creating those parentheses in the first place ....

marcelofabri:
Honestly, I think it should be default. But I've done some informal survey with some friends and the majority actually prefer having the parentheses ¯\_(ツ)_/¯ 

Also, I've not implemented any configurations as we've discussed, so it'd be harder for people to bend the rule to what they prefer.

Dschee:
I know that some like them as they are. Some are still using `self.` all over the place. I was one of them once, but embracing the clarity of Swift also let me rethink and I'm happy I've done so. The reason for that were the community standards, now I type less and the code is still clear.

That's why I think we should make it default once Xcode doesn't add them by default. I hope that the time will come sometime. Thanks for implementing this anyways! Good work. 👍

jpsim:
Cool.

PR #1685, opened 2017-07-17
Enable testCustomRuleConfigurationIgnoreInvalidRules on Linux

marcelofabri:
By failing in a way that doesn’t trigger https://bugs.swift.org/browse/SR-5477


jpsim:
👍

PR #1686, opened 2017-07-17
Split Configuration.swift into focused extensions

jpsim:


jpsim:
Motivated by #1674.

Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
```diff
@@ -0,0 +1,128 @@
...
+        case excluded = "excluded"
+        case included = "included"
+        case optInRules = "opt_in_rules"
+        case reporter = "reporter"
+        case swiftlintVersion = "swiftlint_version"
+        case useNestedConfigs = "use_nested_configs" // deprecated
+        case warningThreshold = "warning_threshold"
+        case whitelistRules = "whitelist_rules"
+    }
+
```
marcelofabri: what do you think about creating a custom `subscript` that takes `Key`? Something like https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Models/LinterCache.swift#L156-L160

Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
```diff
@@ -0,0 +1,128 @@
...
+    private static func warnAboutDeprecations(configurationDictionary dict: [String: Any],
+                                              disabledRules: [String] = [],
+                                              optInRules: [String] = [],
+                                              whitelistRules: [String] = [],
+                                              ruleList: RuleList) {
+
+        // Deprecation warning for "enabled_rules"
+        if dict[Key.enabledRules.rawValue] != nil {
```
marcelofabri: maybe this is a good opportunity for removing this?

Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
```diff
@@ -0,0 +1,128 @@
...
+            queuedPrintError("'\(Key.enabledRules.rawValue)' has been renamed to " +
+                "'\(Key.optInRules.rawValue)' and will be completely removed in a " +
+                "future release.")
+        }
+
+        // Deprecation warning for "use_nested_configs"
+        if dict[Key.useNestedConfigs.rawValue] != nil {
```
marcelofabri: and this

Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
jpsim: yeah, I was thinking of doing that in a fresh PR to reduce the scope of this and #1687.

Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
jpsim: it's not a burden to maintain, so I'd rather keep it, especially since I think it's wise to keep support for the deprecation machinery up-to-date to keep it well maintained when we need to deprecate something again in the future.

Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
jpsim: see https://github.com/realm/SwiftLint/pull/1686#discussion_r127802361

Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
marcelofabri: 👍

marcelofabri:
💯

Issue #1352, opened 2017-03-12
Enhancement: Allow inclusion of multiple configuration files


nigelflack:
I'd like to propose adding a mechanism for including modular configuration files. Also requires the ability to name them other than .swiftlint.yml

i.e.:

```
disabled_rules:
  - trailing_whitespace
  - trailing_newline

[...stuff...]

include_subconfig:
  - customconfig1.yml                # some modular configuration (i.e. custom rules)
  - customconfig2.yml                # another modular configuration
```

I'm not sure the best way raise an enhancement request so opened an issue to gauge interest.

acecilia:
Interested! :)

fredpi:
There's still an active PR addressing this: https://github.com/realm/SwiftLint/pull/3058

fredpi:
This is now implemented in #3058 and will be available in the next release!

Issue #1207, opened 2017-01-17
Rules file as argument to support multiple projects


jacobjennings:
It would be nice to specify a rules file as a command line argument, to support having one set of rules for multiple projects.

aamctustwo:
What were you envisioning for this? A commandline argument? A configuration option in the `.swiftlint.yml` file that lists a path to this custom rules document?

jacobjennings:
Either way would be great. I like the configuration option in the .yml idea best

jpsim:
Maybe we could support external file references in `.swiftlint.yml`. Maybe if used in backticks?

```yaml
whitelist_rules: `~/.swiftlint/common_rules.yml`
```

marcelofabri:
I think in general it'd be better to have a `base_configuration` or something like this so this could be solved in a generic way. We'd probably need to implement #676 though.

SDGGiesbrecht:
This isn’t a native solution, but sharing configurations between projects is one of the reasons I started writing [Workspace](https://github.com/SDGGiesbrecht/Workspace).

 If...

1. The only thing you want to configure is to disable rules, and...
2. Your project supports the Swift Package Manager,

...Workspace should already be able to make SwiftLint share its configuration like you want. If your set‐up is more complicated, or if something below does not work properly, raise an [issue](https://github.com/SDGGiesbrecht/Workspace/issues) at the Workspace repository, and I will look into it.

## Steps:

1. Read the summary of [how to use Workspace](https://github.com/SDGGiesbrecht/Workspace/#the-workspace-workflow).

2. [Add Workspace](https://github.com/SDGGiesbrecht/Workspace/#setup) to each project.

3. [Create a configuration file](https://github.com/SDGGiesbrecht/Workspace/blob/master/Documentation/Configuring%20Workspace.md) like [this one](https://github.com/SDGGiesbrecht/SDG/blob/c3bda4d0327d840577257e668323052dc3ff069b/.Workspace%20Configuration.txt#L30-L41) in one of the projects, then link to it like [this](https://github.com/SDGGiesbrecht/Workspace/blob/9d769969a247447be8e4d30c749b0e1b4e518fb3/.Workspace%20Configuration.txt#L15) in the other projects.

4. Remove your own `.swiftlint.yml` from each project to pass control of SwiftLint off to Workspace.

[...]

Issue #606, opened 2016-04-05
.swiftlint.yml in home dir


mvoelkl:
I would like to place a global (and central managed) configuration for all of our projects in `~/.swiftlint.yml`, but that doesn't seem to work. 

Bug (therefore a feature request) or feature?

jpsim:
That's not currently supported, so this is a new feature request.

joedynamite:
Not sure if this should be a separate feature request or an extension of this existing one, but remote configuration files (not just home dir) would be nice as well.

The Ruby linter RuboCop allows this with explicit inheritance http://rubocop.readthedocs.io/en/latest/configuration/#inheriting-configuration-from-a-remote-url

mgrebenets:
Correct me if I'm wrong, but that currently works. Configs are nested.
https://github.com/realm/SwiftLint#nested-configurations

jpsim:
Only configs nested within your project are nested.

jpsim:
or more precisely, configs in or nested in the working directory are applied

ianpartridge:
We would really like to be able to run `swiftlint --config /path/to/.swiftlint.yml`.

This is because we have lots of repos, and use a common git submodule to include files that we need in all of them.

So, we'd do:

```
git submodule init
git submodule update
swiftlint --config mysubmodule/.swiftlint.yml
```

jpsim:
@ianpartridge you can do that, but you do have to specify the `lint` command when passing arguments:

```
swiftlint lint --config mysubmodule/.swiftlint.yml
```

PR #1687, opened 2017-07-17
Add support for merging nested configuration with parent configurations

jpsim:
Continued from @stephanecopin's work in #1674. Depends on #1686.

@stephanecopin I've given you commit access to SwiftLint so that you can continue iterating on this PR directly on this git remote.


stephanecopin:
@jpsim Thanks a lot! I'll continue here.
I'll be pushing some refactorings around between now and tomorrow.
Also, initially I didn't want to change the implementation of `Rule` itself so I left it as it is and created `HashableRule`, but I feel like a `Rule` could be made `Hashable` directly (we could easily use a default implementation using `identifier` of the `ruleDescription` property, was there reason to use `isEqualTo` over the `Equatable` protocol?). What do you think?

jpsim:
> I'll be pushing some refactorings around between now and tomorrow.

👌 

> was there a reason to use a function `isEqualTo` over the `Equatable` protocol?

`Rule.isEqualTo(_:)` compares two different types that conform to the `Rule` protocol, whereas `Equatable` compares two instances of the same type.

stephanecopin:
> Rule.isEqualTo(_:) compares two different types that conform to the Rule protocol, whereas Equatable compares two instances of the same type.

Makes sense 👍 For a moment I thought `Equatable` was like the `isEqual` implementation in Objective-C.

stephanecopin:
I've pushed 2 more commits, which refactors the way `Configuration` objects are initialized.
Rather than having multiple parameters than are mutually exclusive, I've created an enumeration for enforcing this at compile time, and when parsing a `.swiftlint.yml`, failing if necessary when parsing rather than afterwards.
I'm open to suggestions regarding the name of the enumeration and its value, I called it `RulesMode` but I feel like it could better named.
The refactoring also validates for duplicates in `whitelist_rules`, `opt_in_rules` and `disabled_rules`, whereas before this duplication was just done in the `disable_rules` array (was that intentional?)

Aranoledur:
Hi there! Thanks for this feature, right now we've been thinking about it too! I have a question though. Will this merge work if configuration file is passed as a parameter to swiftlint command in --config? I'm searching swiftlint sources for answers, but maybe someone can answer more quickly.

Problem we have:
There is a base config file at /Users/Shared/ at some build machine, but some projects need to override it to support some legacy code, that will be fixed later.

stephanecopin:
Hey @Aranoledur! The new feature will support that use case as well, any `.swiftlint.yml` files found in subdirectories will be complemented to the base configuration, whether the base configuration is specified via `--config` or not.

Aranoledur:
Great news, thanks! Maybe you need some help on this PR?

stephanecopin:
I'm waiting on some feedback for @jpsim (and others probably), if you see any improvements feel free to experiment. The merging of the nested rules should be complete I believe, what could be improved is merging individual rules' configurations but that would require some more involved changes.

marcelofabri:
Actually, from my understanding, this won't @Aranoledur's issue, unless SwiftLint is run from a parent folder:

```
- dumb
   - actualProject
```

Inside `dumb`:
```
swiftlint lint --config ~/SomeSharedFolder/.swiftlint.yml
```

stephane-fueled:
@marcelofabri oh indeed, it could be a nice feature though to have `--config` file merge with `.swiftlint.yml` in the working directory. I can look into that

stephanecopin:
@marcelofabri oh indeed, it could be a nice feature though to have `--config` file merge with `.swiftlint.yml` in the working directory. I can look into that

marcelofabri:
@stephanecopin I actually don't know if that'd be the expected behavior 😬 

I'd assume that if someone is using `--config`, he/she knows what they're doing and don't want any implicit behavior.

stephanecopin:
@marcelofabri true, I agree that could be unexpected.
I guess this behavior could opted-in with a flag or something like that (and maybe output a warning about an ignored `.swiftlint.yml` if one is there, no flag and `--config` is used?). I'm thinking out loud, just thinking of the best to compromise @Aranoledur's use case with others.

marcelofabri:
I think we probably should support that with an "import" feature. Probably just merging configurations is not the best way to tackle that problem. We have several issues opened about it.

Aranoledur:
What is the current behaviour on this? If you set configuration with --config and have nested .yml ?
"import" feature - like importing some configuration to local .yml?

marcelofabri:
> What is the current behaviour on this? If you set configuration with --config and have nested .yml ?

It should consider the custom configuration and merge it with nested ones (but not the one at the same level).

-----

For reference, some issues about this feature: #1352, #1207, #606

Aranoledur:
What about some --baseConfig parameter that will be merged/overrode with config files from working directory?

jpsim:
@Aranoledur could you please continue the conversation in #1352, #1207 or #606, whichever is most relevant to your interests? We should keep this PR focused on the merging of nested configuration files in order to make progress.

Aranoledur:
Sure, sorry for this.

jpsim:
No need to be sorry, you're bringing up great conversation points, I just don't want this PR to steer too far off course, making it harder to review.

stephanecopin:
Indeed, I can't agree more regarding keeping it simple 🙂
As a side note, I see the PR failing in Travis, however I'm unsure why - the push compiles fine, whereas the pr doesn't. Is there something I'm missing? (It also lints fine on my end)
Reading why it fails, `rulesMode` *is* definitely initialized at that point, it's puzzling me 🤔

jpsim:
Yeah, no clue why it failed, on all platforms tested no less... 🤔

jpsim:
Give me some time to digest the idea of this RulesMode thing... I see the motivation and its advantages, but something about that direction (don't think it's the naming) still isn't sitting well with me. I can't put my finger on it though, so no concrete feedback for you just yet.

jpsim:
I'm curious, @marcelofabri what do you think about all this?

stephanecopin:
@jpsim I'm not sure, I feel like it's better than having 4 different parameters, unless using the `enum` looses the flexibility offered by the 4 parameters? It's always something we could do with an `enum`, given the proper `case`s.
For Travis once you've reviewed I can rebase and see if it passes... hopefully it does

Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
```diff
@@ -90,6 +90,45 @@ extension Configuration {
...
+
+        self.init(rulesMode: rulesMode,
+                  included: included,
+                  excluded: excluded,
+                  warningThreshold: warningThreshold,
+                  reporter: reporter,
+                  ruleList: ruleList,
+                  configuredRules: configuredRules,
+                  swiftlintVersion: swiftlintVersion,
+                  cachePath: cachePath ?? cachePath)
```
marcelofabri: is this right: `cachePath ?? cachePath`?

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -10,6 +10,12 @@ import Foundation
 import SourceKittenFramework
 
 public struct Configuration: Equatable {
+    // Represents how a Configuration object can be configured with regards to rules.
+    public enum RulesMode {
```
marcelofabri: I like the idea of this enum, but:
1. Can we think about a better name 🤔 ? (I couldn't)
2. What do you think about making the associated values `Set<String>`?
3. Is there a better name for `default`?

Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
```diff
@@ -43,10 +43,69 @@ extension Configuration {
...
+            return type(of: rule).description.identifier.hashValue
+        }
+    }
+
+    private func mergingRules(with configuration: Configuration) -> [Rule] {
+        switch configuration.rulesMode {
+        case .allEnabled:
+            // Technically not possible yet as it's not configurable in a .swiftlint.yml file,
+            // but implemented for completeness
+            return configuration.rules
```
marcelofabri: I know this case isn't possible today, but shouldn't this merge the configurations? e.g. If I have a custom configuration for a rule in the parent config, but nothing in the child one, it should use the parent's.

Since this is not used anyway, I'd either replace this with a `fatalError` or make clear in the comments that the implementation needs to change if we ever enable that option in a configuration.

Tests/SwiftLintFrameworkTests/ConfigurationTests.swift
```diff
@@ -118,6 +120,20 @@ class ConfigurationTests: XCTestCase {
         XCTAssertEqual(expectedIdentifiers, configuredIdentifiers)
     }
 
+    func testDuplicatedRules() {
+        let duplicateConfig1 = Configuration(dict: ["whitelist_rules": ["todo", "todo"]])
+        XCTAssert(duplicateConfig1 == nil, "initializing Configuration with duplicate rules in " +
```
marcelofabri: `XCTAssertNil`? (the same applies for the whole test)

Tests/SwiftLintFrameworkTests/ConfigurationTests.swift
```diff
@@ -178,7 +194,16 @@ class ConfigurationTests: XCTestCase {
     // MARK: - Testing Nested Configurations
 
     func testMerge() {
-        XCTAssertEqual(projectMockConfig0.merge(with: projectMockConfig2), projectMockConfig2)
+        XCTAssertFalse(projectMockConfig0.rules.contains(where: { $0 is ForceCastRule }))
```
marcelofabri: This is a bit hard to read. Maybe creating a helper method would help?

```swift
private extension Configuration {
   func contains(rule: Rule.Type) -> Bool {
       return rules.contains(where: { $0 is rule })
   }
}
```

Tests/SwiftLintFrameworkTests/ConfigurationTests.swift
```diff
@@ -23,6 +23,8 @@ extension Configuration {
     }
 }
 
+// swiftlint:disable type_body_length
```
marcelofabri: we probably should also split this test case in several ones

marcelofabri:
I've made some comments, but in general I'm happy with the direction.

Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
stephanecopin: Whoops, probably an error copy pasting code around. Will change

Source/SwiftLintFramework/Models/Configuration.swift
stephanecopin: 1. I also thought of `RulesConfiguration`, but it'd be kinda confusing as `RuleConfiguration` is already used. Other that than I can't find of a better name either :/
2. +1 for `Set<String>`, will make the change. We can make the duplicate validation earlier in `Configuration+Parsing`
3. Maybe `overrideDefaults`?

Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
stephanecopin: I'd agree, but the fact that we're not merging the individual rule's configurations yet leads to this code for now.
I can update the comment that the implementation needs to change in order to properly support that feature, if it ever becomes supported

Tests/SwiftLintFrameworkTests/ConfigurationTests.swift
stephanecopin: Definitely, will update

Tests/SwiftLintFrameworkTests/SourceKitCrashTests.swift
```diff
@@ -75,7 +75,7 @@ class SourceKitCrashTests: XCTestCase {
             XCTFail("If this called, rule's SourceKitFreeRule is not properly configured")
         }
         // swiftlint:disable:next force_unwrapping
```
stephanecopin: Will do ;)

stephanecopin:
Thanks! I should be able to fix the comments & rebase today, otherwise I'll do it tomorrow

stephanecopin:
@marcelofabri I've updated the code and rebased, so it should be all good. I've also split up the `ConfigurationTests.swift` in multiple files, let me know what you think.
The build is still not passing for some reason 🤔  I'll try to look more into it as to why later today

marcelofabri:
I'll try to take a look at this later today or tomorrow.

BTW, I think we should release one version before merging this just in case, since it can be an important breaking change.

jpsim:
I've come to really like the `RulesMode` approach! Sorry for the delay in reviewing this. I just merged it after addressing the conflicts and made some minor tweaks. See the following commits:

* de08c7a4 - (HEAD -> master) reduce scope of disabled rule in ConfigurationTests.swift (3 minutes ago) <JP Simard>
* 1e529ea9 - use trailing closure syntax in ConfigurationTests+Nested.swift (4 minutes ago) <JP Simard>
* db95f5d3 - use explicit ACL keywords for HashableRule (10 minutes ago) <JP Simard>
* 428d48cb - make sourcery (26 minutes ago) <JP Simard>
* 6c3a697a - fixup pbxproj (26 minutes ago) <JP Simard>
* 8ffbc64b - merge changelog entries and move to appropriate section (31 minutes ago) <JP Simard>
* f86a99ee - Split ConfigurationTests.swift file in multiple files (33 minutes ago) <Stéphane Copin>
* 3ad4842b - Small changes and refactoring based on github Feedback (34 minutes ago) <Stéphane Copin>
* 8e6c3edc - Add more tests related to duplicated rules / all rules enabled Configurations (34 minutes ago) <Stéphane Copin>
* 9f1f454c - Refactor the way Configuration objects are created with regards to rules (34 minutes ago) <Stéphane Copin>
* 48dbc151 - hard wrap changelog entries & add myself as author (34 minutes ago) <JP Simard>
* 8450418c - add missing tests to allTests (34 minutes ago) <JP Simard>
* a1ab5124 - Add more tests related to nested configurations (34 minutes ago) <Stéphane Copin>
* f5f63c90 - Update tests to reflect nested configuration changes (34 minutes ago) <Stéphane Copin>
* 95d60f93 - Add Changelog entry (34 minutes ago) <Stéphane Copin>
* 97c1dcd0 - Add support for nested configuration (34 minutes ago) <Stéphane Copin>

stephanecopin:
Awesome! Glad to hear it was merged, I can't wait for it to be available in the next release :)

Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
gretzki: As I couldn't find an open issue about the merging of individual rule, I've just opened an issue and suggested a solution here: #2058

PR #1688, opened 2017-07-17
Use cached configurations if available when initializing

jpsim:


jpsim:
Marked as work-in-progress until I can fix some stuff.

jpsim:
ok, should be good now 😅

PR #1690, opened 2017-07-18
Update README.md Using CocoaPods

jirapornpraneet:
Update README.md 
if you've installed SwiftLint via CocoaPods the script should look like this:
 ${PODS_ROOT}/SwiftLint/swiftlint  not should look like this  "${PODS_ROOT}/SwiftLint/swiftlint"


marcelofabri:
I don't think this should change. See #1256.

PR #1691, opened 2017-07-18
Add support for optional 'error' severity level in configurations

### LINKED ISSUE #1647: Set file_length rule only as warning, but still set custom length value

marcelofabri:
Continued from #1664

Fixes #1647

@JamieEdge, I've added some tests and changed the behavior a little so we only set the `error` level to `nil` when it's explicitly set in the configuration.


marcelofabri:
Actually, that behavior was not the right one 😅

PR #1409, opened 2017-04-02
Ignore comment lines in file_length rule

sammy-SC:
Fixes #1165

PR #1702, opened 2017-07-21
Add option to ignore comment lines in file_length rule

### LINKED ISSUE #1165: FileLengthRule rule should not count comment lines
If FunctionBodyLengthRule (#258) and TypeBodyLengthRule (#369) do it then so should FileLengthRule

marcelofabri:
Continued from #1409. Fixes #1165

@sammy-SC, I hope you don't mind, but I've cherry-picked your commits and squashed them to avoid having a confusing history. I also did some small changes, but a important one is that only lines with **only** comments are now excluded.


sammy-SC:
@marcelofabri thank you. I can't see the change where "only lines with only comments are now excluded". Please, could you point it out?

Tests/SwiftLintFrameworkTests/RulesTests.swift
```diff
@@ -96,11 +96,6 @@ class RulesTests: XCTestCase {
         verifyRule(FatalErrorMessageRule.description)
     }
 
-    func testFileLength() {
```
sammy-SC: If this is removed, tests with default settings won't be run, will they?

marcelofabri: This was added to `FileLenghtRuleTests`: https://github.com/realm/SwiftLint/pull/1702/files#diff-d2c68a7ad0b1aca075f32e032d5ea03aR14

Source/SwiftLintFramework/Rules/FileLengthRule.swift
```diff
@@ -28,22 +28,21 @@ public struct FileLengthRule: ConfigurationProviderRule {
     )
 
     public func validate(file: File) -> [StyleViolation] {
-        func getLineCountWithoutComments() -> Int {
+        func lineCountWithoutComments() -> Int {
             let commentKinds = Set(SyntaxKind.commentKinds())
             let lineCount = file.syntaxKindsByLines.filter { kinds in
-                return !kinds.filter { !commentKinds.contains($0) }.isEmpty
+                return !Set(kinds).isSubset(of: commentKinds)
```
marcelofabri: @sammy-SC this is the change ^

PR #1703, opened 2017-07-22
Customize YAML loading for configurations

### LINKED ISSUE #1424: Weird behavior when specifying "on" in `identifier_name` excluded list
```
identifier_name:
  excluded:
    - to
    - id
    - ok
    - on
```
This stopped working since 0.18.1. Swiftlint now acts as if the excluded list doesn't exist. However, when I remove `on` from this list, it works as expected (but still giving a warning about the use of `on` in my codebase of course).

 I think this is the result of the switch in YAML parsing, because "on" is a reserved word in yaml used to specify a boolean. I've tried putting it in quotes around `"on"` but that didn't help.

### LINKED ISSUE #1512: Use environment variables in .swiftlint.yml file
Is it possible to use global vars from project in the .swiftlint.yml file?
I want to create a template for all projects and have a subfolder in these, that I want to exclude like this:

```
  - $(PROJECT_NAME)/Extensions
```

is this possible? because this does not work :-( 
thanks for help

marcelofabri:
This enables:

1. Using environment variables inside a configuration (fixes #1512)
2. Treating reserved keywords as strings (`yes`, `no`, `on` and `off`) instead of booleans (fixes #1424)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,17 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1719](https://github.com/realm/SwiftLint/issues/1719)
 
+* Allow using environment variables in a configuration file in the form of
+  `${SOME_VARIABLE}`. The variables will be expanded when the configuration
+  is first loaded.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#1512](https://github.com/realm/SwiftLint/issues/1512)
+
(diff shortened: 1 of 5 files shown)


jpsim:
I'd love to hear @norio-nomura's thoughts on this, and whether or not this is the "right" way to do this.

marcelofabri:
@jpsim Do you have any thoughts on this PR? Should I just merge it?

Issue #1220, opened 2017-01-20
Support for '*' when search for files excluded?


sunkehappy:
I want to ignore some files that start with _(because I'm using [mogenerator](https://github.com/rentzsch/mogenerator)). But it seems that it does nor work:

    excluded: # paths to ignore during linting. Takes precedence over `included`.
      - Carthage
      - Pods
      - oTMS_Mobile/Model/CoreDataModel/_*.swift

Is this an unsupported feature or something wrong I write above?

aamctustwo:
Is this a duplicate of #560 `Exclude files using regular expression.`?

marcelofabri:
I don't think this is exactly a duplicate because this seems to use [glob patterns](https://en.wikipedia.org/wiki/Glob_(programming)) and not regular expressions.

jpsim:
> I don't think this is exactly a duplicate because this seems to use glob patterns and not regular expressions.

Although I think we'd probably just want to stick with one and not the other. Generally the approach is to use globbing syntax when it comes to specifying files, as far as I know.

marcelofabri:
+1 for globbing and just one of them

fcy:
+1 I'd love to have a way to exclude all generated files, we put them under Generated folders. Something like this:

```
excluded:
  - **/Generated
```

Excludes files in any folder called Generated.

jpsim:
I'm all for this. Anyone know of any good file globbing libraries written in Swift that are compatible with CocoaPods, Swift Package Manager and Carthage?

tschob:
Is there any progress on #1712 or this issue?

We also have the case that we want to exclude files which are located in special folders or contain a certain string. Some of our projects are forced to use a unclean folder setup as we can't configure SwiftLint to ignore the files we need to.

#### Our background:
* Agency with a lot of projects
* One **shared** swiftlint.yml which gets copied into the project with each build (source is a submodule)

### Example with the folder `Generated`

#### Desired project setup:
```
/Root
  /App
      /Generated
  /Extensions
    /Watch
      /Generated
    /Share
      /Generated
```

#### Currently forced project setup:
```
/Root
  /App
  /Extensions
  /Generated
    /App
    /Watch
    /Share
```

tschob:
Based on the README it's possible to use Regex for `included` and `excluded` in custom rules:

```yaml
custom_rules:
  pirates_beat_ninjas: # rule identifier
    included: ".*\\.swift" # regex that defines paths to include during linting. optional.
[...]

PR #1712, opened 2017-07-26
Feature/support for excluding with glob

xzeror:
Addressing issue #1220. Added support for excluding files using globs. 
For example, if you want to exclude all test files in project you can add following line into your .swiftlint.yml:
```
excluded:
  - ./**/*Test?.swift
```


jpsim:
Thanks for your work on this! A few thoughts:

1. If we support globbing in one place where we take paths, we should support it everywhere we take paths. This includes the `included` configuration key, maybe other places too, you'll have to look to determine where else it might apply.
2. We don't often add external dependencies to SwiftLint, and doing so should be weighed very carefully. In this case, I have a few concerns:
    1. The dependency is a fork of another project. Is the fork the canonical version of the project? If not, why is the original project not being used?
    2. What signs are there that we can rely on this being actively maintained? Both the fork and the original repo have 0-5 GitHub stars and I can't find any other projects using them.
    3. Tests appear to be minimal, but at least there are some.
    4. What alternatives do we have? Is there a well-regarded C-library that we could wrap in a Swift API?

xzeror:
Hi. 
1. I thought about it. Supporting globs in all places could be done later. This PR resolves original ticket and common problem of excluding test files from SwiftLint when they lay in the same place with code. It would be really great if we can ship this MVP to users.
2. Yeah. Adding it as dependency was not easy =) 
     i. Original project could not be used because the author doesn't want to add Carthage and CocoaPods support. You could read our discussion [here](https://github.com/Bouke/Glob/pull/2)
    ii. I maintain my fork at the moment and expect to do so in the future.
    iii. I'll try to add some tests. What target coverage should I achieve?
    iv. Glob is lightweight wrapper around [glob](http://www.manpages.info/macosx/glob.3.html) stdlib call already. Should we search for another more lightweight library? I don't think so. Should we use direct stdlib call in SwiftLInt? No. We will need wrapper around it to abstract C structures. Should we incorporate Glob.swift into SwiftLint directly? Maybe. That should be maintainers decision.

xzeror:
@jpsim could you review my recent changes please?

shyambhat:
Is there anything I can do to help push this PR forward? 

I have no idea what Glob is, but I really think the ability to exclude files with a regex is invaluable.
We integrated SwiftLint into a five year old codebase and it's been almost a week that we've been fixing existing lint warnings. The hundreds of test files in the project are placed in the same hierarchy as the code being tested and having regexes to exclude these files will save us probably another week of fixing warnings.

tschob:
Are there any plans to continue or replace this PR?

tschob:
@jpsim Regarding 2.)

Do you think using Ruby to interpret the Glob statements is an option?

jpsim:
If this is to be considered, the next steps are to rebase the PR (no merge commits please) and have it pass all CI checks.

norio-nomura:
`Carthage/Checkouts/Glob` should be submodule instead of embedding.

keith:
I've implemented an alternative to this here: https://github.com/realm/SwiftLint/pull/2316

keith:
This has been merged in https://github.com/realm/SwiftLint/pull/2316 and is part of 0.27.0. I think we can close this PR

PR #1713, opened 2017-07-26
Allow fileprivate extensions in PrivateOverFilePrivateRule

jpsim:
semantically, it seems like you ought to be able to say

"all the declarations in this extension are 'fileprivate'"

### DIFF
--- a/Rules.md
+++ b/Rules.md
@@ -7444,6 +7444,26 @@ fileprivate func Something(){}
 ```
 
 ```swift
+fileprivate extension String {
+fileprivate func Something(){}
+}
+```
+
+```swift
+fileprivate extension String {}
+```
+
+```swift
+fileprivate 
+ extension String {}
+```
+
+```swift
+fileprivate extension 
+ String {}
+```
+
+```swift
 class MyClass {
 fileprivate let myInt = 4
 }
@@ -7472,17 +7492,13 @@ fileprivate struct Inner {}
 ```
 
 ```swift
-↓fileprivate extension String {}
-```
-
-```swift
 ↓fileprivate 
- extension String {}
+ enum MyEnum {}
 ```
 
 ```swift
-↓fileprivate extension 
- String {}
+↓fileprivate enum 
+ MyEnum {}
 ```
 
 ```swift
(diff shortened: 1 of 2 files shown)


jpsim:
I'd like your thoughts on this @masters3d 🙏

masters3d:
I think that for Swift 3 this change makes sense but since Swift 4 degraded the meaning of explicit private inside extensions then this would be a little confusing for a rule named `PrivateOverFilePrivateRule`. 

> "There is no solution that will make everyone happy: maintaining the status quo makes “fileprivate” too common and therefore not meaningful when it occurs in source;"
 https://lists.swift.org/pipermail/swift-evolution-announce/2017-April/000357.html

masters3d:
@jpsim perhaps a better approach for a new rule would be `no implicit fileprivate rule inside extensions` but it seems that you are trying to go there way around.

marcelofabri:
I slightly prefer using `private` in that case, so IMO it should be at least a configuration.

jpsim:
> I slightly prefer using `private` in that case, so IMO it should be at least a configuration.

Except that the following makes things _less_ clear IMHO:

```swift
private extension SyncManagementObjectStatus {
    fileprivate init(statusCode: RealmOptional<Int>) {
        /* impl */
    }
}
```

Unless you meant `private` in the inner declaration, which wouldn't work in this case either.

Even in Swift 4, I'd argue that _if_ you want to use `fileprivate` at all, you ought to be able to specify a `fileprivate` extension.

masters3d:
For the example above, I would suggest expanding the ExtensionAccessModifierRule to help prevent mixing private extensions along and fileprivate members since they do overlap and it is confusing. 
Before SE-025 the above examples would have been an error  😿. 

> https://github.com/apple/swift-evolution/blob/master/proposals/0025-scoped-access-level.md#complications-with-private-types
> The compiler should not warn when a broader level of access control is used within a type with more restrictive access, such as internal within a private type. This allows the designer of the type to select the access they would use were they to make the type more widely accessible. (The members still cannot be accessed outside the enclosing lexical scope because the type itself is still restricted, i.e. outside code will never encounter a value of that type.)

marcelofabri:
> Except that the following makes things *less* clear IMHO:

In your example, I agree. But there're cases where it's useful, for example the [first violation on Firefox](https://github.com/mozilla-mobile/firefox-ios/blob/f9380e6e148710db8bdcca910825b743aa5b8afc/Client/Frontend/Browser/SearchViewController.swift#L587:1). 

That's why I think it should be a configuration, even if by default it wouldn't trigger.

marcelofabri:
Closing in favor of #1733.

Issue #1714, opened 2017-07-27
[Rule Request] Prefer new block based KVO API in Swift 3.2 or 4


marcelofabri:
When using Swift 3.2 or 4, the new block based API KVO API is preferred.

<img width="1920" alt="kvo" src="https://user-images.githubusercontent.com/833072/28663476-c2c953ca-72bd-11e7-9b30-9b874541a395.png">

```swift
// should trigger
func observeValue(forKeyPath keyPath: String?, of object: Any?, change: [NSKeyValueChangeKey : Any]?, context: UnsafeMutableRawPointer?) {

}

// ok
let observer = foo.observe(\.value, options: [.new]) { (foo, change) in
    print(change.newValue)
}
```

marcelofabri:
Probably the most trick thing here is to detect the Swift version.

We can the new keypath syntax and the structure to find if it's Swift >= 3.2.

* `file.swift`
```swift
let nameKeyPath = \ForceUser.name
```

* Command
`sourcekitten structure --file file.swift`

* Result with Swift 3.1:
```json
{
  "key.offset" : 0,
  "key.diagnostic_stage" : "source.diagnostic.stage.swift.parse",
  "key.length" : 34,
  "key.diagnostics" : [
    {
      "key.filepath" : "\/Users\/marcelofabri\/dev\/SwiftLint\/file.swift",
      "key.column" : 19,
      "key.description" : "invalid character in source file",
      "key.fixits" : [
        {
          "key.offset" : 18,
          "key.sourcetext" : " ",
          "key.length" : 1
        }
      ],
      "key.line" : 1,
      "key.diagnostic_stage" : "source.diagnostic.stage.swift.parse",
      "key.severity" : "source.diagnostic.severity.error"
    }
  ],
  "key.substructure" : [
    {
      "key.nameoffset" : 4,
      "key.accessibility" : "source.lang.swift.accessibility.internal",
      "key.length" : 33,
      "key.name" : "nameKeyPath",
      "key.kind" : "source.lang.swift.decl.var.global",
      "key.namelength" : 11,
      "key.offset" : 0
    }
  ]
}
```

* Result with Swift 3.2 or 4:
```json
{
  "key.diagnostic_stage" : "source.diagnostic.stage.swift.parse",
  "key.substructure" : [
    {
      "key.nameoffset" : 4,
      "key.accessibility" : "source.lang.swift.accessibility.internal",
      "key.length" : 33,
      "key.name" : "nameKeyPath",
[...]

PR #1716, opened 2017-07-27
Add four to SwiftVersion

marcelofabri:
This will allow rules like #1714 to be implemented.

Credits to @ornithocoder for the idea of using `#if swift`.

### DIFF
--- a/Source/SwiftLintFramework/Models/SwiftVersion.swift
+++ b/Source/SwiftLintFramework/Models/SwiftVersion.swift
@@ -7,16 +7,26 @@
 //
 
 import Foundation
+import SourceKittenFramework
 
 enum SwiftVersion {
     case three
+    case four
 
     static let current: SwiftVersion = {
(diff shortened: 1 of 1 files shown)


jpsim:
Does the same approach not work for `#if swift(>=3.2)`?

Also, kind of a shame that SourceKit itself can't be used to query the Swift version it was built with. There's a "protocol version" request you can make, but that's always been 1.0: https://github.com/apple/swift/blob/master/tools/SourceKit/docs/Protocol.md#protocol-version

marcelofabri:
> Does the same approach not work for `#if swift(>=3.2)`?

Yes, but that is also true for Swift 4. Since they're the same compiler, just with different modes, I don't even know how SourceKit deals with that (maybe it doesn't?).

> There's a "protocol version" request you can make, but that's always been 1.0

Yeah, I've tried that too 😢

jpsim:
Anyway, 👍

PR #1723, opened 2017-07-28
Fix false positive in `empty_enum_arguments` rule when using wildcards and  `where` clauses

### LINKED ISSUE #1722: Matching tuples using wildcards and a "where" clause triggers empty_enum_arguments
The following snippet is interpreted as a violation of the `empty_enum_arguments` rule:

```swift
switch (email, password) {
    case (_, _) where !email.value.isBlank && !password.value.isBlank:
        return .valid(email: .idle, password: .idle)
    default:
        break
}
```

I was expecting a violation of a different rule, or rather that no rules were violated at all.

marcelofabri:
Fixes #1722

### DIFF
--- a/Rules.md
+++ b/Rules.md
@@ -1738,31 +1738,37 @@ Arguments can be omitted when matching enums with associated types if they are n
 
 ```swift
 switch foo {
- case .bar: break
+    case .bar: break
 }
 ```
 
 ```swift
 switch foo {
- case .bar(let x): break
+    .bar(let x): break
 }
 ```
 
 ```swift
 switch foo {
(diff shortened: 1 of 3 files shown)


Rules.md
```diff
@@ -1738,31 +1738,37 @@ Arguments can be omitted when matching enums with associated types if they are n
...
 switch foo {
- case .bar: break
+    case .bar: break
 }
 ```
 
 ```swift
 switch foo {
- case .bar(let x): break
+    .bar(let x): break
```
marcelofabri: good catch! I'll fix it now.

PR #1725, opened 2017-07-29
Fix false positive in `large_tuple` rule when using throwing closure

Liquidsoul:
This is an issue that I have noticed in this [Swiftlint PR](https://github.com/SwiftGen/StencilSwiftKit/pull/54/files#diff-eb34ef67e2507f89acad77e19f598a4aR43).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,9 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1722](https://github.com/realm/SwiftLint/issues/1722)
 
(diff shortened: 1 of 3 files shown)


marcelofabri:
Thanks so much for your contribution, @Liquidsoul 🎉 

I only made a minor change in 556ad4eed3db8b3b60d2d23c83afc70f1339b93c to use a non-capture group instead, which should be slightly faster.

Liquidsoul:
Oh, you're right, thank you for fixing that! 🙇

PR #1729, opened 2017-07-31
Fixed Readme custom rule included/excluded regexp

DZozulya:
Currently, it does not check for dot presence, but for any symbol, so for example, if a filename is `MySuperFileswift` it would also work for this regexp. Same for excluded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,8 +315,8 @@ following syntax:
 ```yaml
 custom_rules:
   pirates_beat_ninjas: # rule identifier
-    included: ".*.swift" # regex that defines paths to include during linting. optional.
-    excluded: ".*Test.swift" # regex that defines paths to exclude during linting. optional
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -315,8 +315,8 @@ following syntax:
 ```yaml
 custom_rules:
   pirates_beat_ninjas: # rule identifier
-    included: ".*.swift" # regex that defines paths to include during linting. optional.
-    excluded: ".*Test.swift" # regex that defines paths to exclude during linting. optional
+    included: ".*\\.swift" # regex that defines paths to include during linting. optional.
```
marcelofabri: shouldn't it be just one backslash?

DZozulya: One backslash would lead to a crash of yaml parser, so it has to be escaped.

PR #1730, opened 2017-07-31
Make `file_header` rule ignore doc comments.

### LINKED ISSUE #1719: file_header should ignore doc comments
My file_header config is:

```
file_header:
  forbidden_pattern: ".?"
```

I.e. forbid file headers. But I have some types with documentation comment on them. E.g.

```
/// This is great tool.
class GreatTool {}
```

By coincidence the doc comment is the first non-blanc text in the file. As a result, I think, the file_header rule interprets the doc comment as file header and emits a warning.

Ideally, file_header should ignore triple-slash comments prepending some declaration.

marcelofabri:
Fixes #1719

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1714](https://github.com/realm/SwiftLint/issues/1714)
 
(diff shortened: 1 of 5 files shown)


marcelofabri:
The Firefox violations shouldn't exist 🤔

PR #1738, opened 2017-08-02
Change XCTFailMessageRule's ID to just 'xctfail_message'

jpsim:


jpsim:
Oops, dangerfile refactor snuck in there when I didn't mean to 😅 

Let me know if you'd rather review without that commit.

Issue #1102, opened 2016-12-31
Warn if no violations were produced in a disabled region


jpsim:
Maybe SwiftLint should warn if a disable command doesn't actually silence a violation?

Due to SwiftLint's architecture, we already detect these violations, we just filter them out (silence them) if they happen to be in a disabled region before passing it back to the linter.

So it shouldn't be _too_ hard to do this.

heidiproske:
❤ yes please! I recently did this (manually) and was able to remove a bunch of unnecessary disable flags. (The project had been refactored and Linting had been disabled for a while so things became out of date.)

PR #1740, opened 2017-08-02
Produce an error when a 'disable' command doesn't silence any violations

jpsim:
Addresses #1102. Needs tests.


marcelofabri:
This will probably need some changes to avoid triggering on a file starting with a shebang.

jpsim:
This is ready to be reviewed now. Was definitely trickier than I expected to get right.

marcelofabri:
I have two questions:

1. Do we really need to add `skipDisableCommandTests` in all these tests?
2. Should we add a way to opt-out from this validation? For example, [SwiftGen](https://github.com/SwiftGen/SwiftGen) tries to be a good citizen and disable some potential warnings in its generated code (so you don't need to manually exclude the file in your configuration file). It can happen that the generated code doesn't actually trigger one of the violations. One example of created template can be seen [here](https://github.com/SwiftGen/templates/blob/f3b291073f541e4cf712d0bf65d5815760918cf6/Tests/Expected/Storyboards-iOS/swift2-context-all-customname.swift#L53). If it's not to much work, I think it'd be nice to be able to disable this "rule" with commands and also from the configuration.

jpsim:
> 1. Do we really need to add `skipDisableCommandTests` in all these tests?

Some are necessary, but others weren't:

* `FileHeaderRuleTests`: these rules deal with text at the very start of the file, which means the tests are invalid if its content is preceded by a disable command.
* `testLeadingWhitespace`: this rule deals with text at the very start of the file, which means the test is invalid if its content is preceded by a disable command.
* `testLetVarWhitespace`: there was a triggering example that needed to be modified to pass with 'disable command' tests. Fixed in 9bc45d88.
* `VerticalWhitespaceRuleTests`: this rule wasn't producing a violation in disabled regions at all rather than expecting the linter to filter out violations from disabled regions. Fixed in 4ca59c48.

jpsim:
> 2. Should we add a way to opt-out from this validation?

I can definitely see lots of valid cases where you shouldn't need to care if a violation was or wasn't triggered in a disabled region. I'm not sure exactly how we should expose this configurability. A few options:

1. Treat `superfluous_disable_command` as a rule, where you can configure its severity, enable/disable it from `.swiftlint.yml`, locally disable it just like any other rule.
2. Treat it as a CLI/config option, e.g. like `cache-path`/`cache_path` or warning threshold.

marcelofabri:
@jpsim I'd treat it as a rule, as treating as a CLI/config option wouldn't help in SwiftGen use case. If we treat it as a rule, we could also see how hard would it be to handle it on `disabled_rules/whitelist_rules` as well.

marcelofabri:
@jpsim I've pushed a commit (62b5231) that turns `SuperfluousDisableCommandRule` into a full rule, supporting disabling from the configuration file, from commands and even changing the severity.

marcelofabri:
🚢

jpsim:
Thanks for continuing with this, @marcelofabri!

Two thoughts:

1. I dislike that the rule doesn't actually validate in the `Rule.validate(_:)` method. Is there any way that we can do this more cleanly?
2. The Xcode9 CI job is failing on Travis. I assume things will need to be updated for Xcode 9 beta 5? We should always strive to keep `master` green, which means either updating for Xcode 9 beat 5 immediately, or disabling the job and filing an issue tracking to re-enable it.

jpsim:
Also, could you please explain 4fed36c772cc5bfc911858b0a33676a65fc4dad4? I don't understand why that's necessary...

marcelofabri:
> I dislike that the rule doesn't actually validate in the Rule.validate(_:) method. Is there any way that we can do this more cleanly?

I'm not a big fan either, but I don't see any way unless we either remove functionality (disabling from comments/configuration) or we refactor this a lot to not validate on `Linter` level, but somehow at the Rule level. Or even implementing the rule disable behavior just for this rule. Do you have any suggestions?

> The Xcode9 CI job is failing on Travis. I assume things will need to be updated for Xcode 9 beta 5? We should always strive to keep master green, which means either updating for Xcode 9 beat 5 immediately, or disabling the job and filing an issue tracking to re-enable it.

I think Xcode 9 b5 has a (small) performance regression which made it take longer to compile some SourceKitten functions, violating the threshold. Since that was removed from SourceKitten, cutting a new release and updating it here should fix the issue. We could also point to master temporarily.

> Also, could you please explain 4fed36c? I don't understand why that's necessary...

For some weird reason, the other initializer (the one that takes a filename) was being used. That made the tests with SPM to fail, since it'd pick the SwiftLint configuration (instead of a "clean" one). I think it didn't affect the Xcode build because it doesn't run the tests on the project directory.

This enforces the correct (IMO) initializer to be used in that case.

jpsim:
> Do you have any suggestions?

Nope.

> We could also point to master temporarily.

I'll cut a SourceKitten release today or tomorrow hopefully.

> For some weird reason, the other initializer (the one that takes a filename) was being used.

That's confusing, since the initializer that takes the configuration values directly is failable, whereas the one that loads from a file isn't. Something worth revisiting later.

PR #1742, opened 2017-08-03
Update changelog entry for xctfail_message

ornithocoder:
Not an opt-in rule. It's enabled by default https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules/XCTFailMessageRule.swift


jpsim:
Thanks!

Issue #1721, opened 2017-07-28
Swiftlint only expects files with type ".swift"


lswith:
I know that swiftlint explicitly expects files of type .swift, but I'd like to pass a file with ".lint-test" instead of ".swift". This is mainly for easy integration with https://github.com/phacility/arcanist.

SDGGiesbrecht:
I’ve seen several issues where quick bug demonstrations were written as scripts like this:

`echo "func doSomething() {}" | swiftlint lint --use-stdin`

Does it work to pipe your file to `swiftlint` like that?

lswith:
The standard way is to create the .lint-test file. I can override the
standard but it seems odd that you can't take any file

marcelofabri:
I don't think we should change the default behavior, but I think it'd be reasonable to offer a way to override it, perhaps in the configuration file.

lswith:
You guys are still not at 1.0, so you can do whatever you want haha.
Honestly I think specifying exact files is the way to go. Checking if
something is a swift file or not shouldn't be the job of the linter.

A good example is to simply pass globs to the ‘--path‘ variable. Something
like ‘--path **/*.swift‘.

marcelofabri:
> Checking if
something is a swift file or not shouldn't be the job of the linter.

I disagree. I think the responsibility of a tool is to provide sensitive defaults. I wouldn't want that, by default, SwiftLint linted Objective-C files for example. Or that we needed to use `--path` all the times.

> A good example is to simply pass globs to the `--path` variable. Something like `--path **/*.swift`.

That could also work, but currently I don't think we even support passing a non-Swift file to `--path`.

lswith:
If the path argument allowed non swift files and the default was all .swift
files I think that would work to both of our cases

SDGGiesbrecht:
@lswith, @marcelofabri 

> The standard way is to create the .lint-test file.

I do not understand what a `.lint-test` file is. I ran a Google search for `".lint-test"`, but it generated no accurate hits. What standard are you referring to?

I added `arcanist` to the search term list and about the 60th hit was the source code for the unit tests for Arcanist itself.

Those files were the equivalent of SwiftLint’s `RuleDescription.(non)TriggeringExamples`, and the files (the one’s I saw described CSS) are not pure source, but are repeatedly interrupted with error expectations (analogous to `↓` in `triggeringExamples`) and end with a JSON configuration to use for the unit test.

[...]

Issue #1693, opened 2017-07-18
Flag to add Highest and Lowest Level Config Files (in Merge Tree)


SDGGiesbrecht:
This takes over from @Aranoledur’s side track in #1687. Since this requires merging of nested configurations, this would be a future, independent enhancement after that is completed.

This is also the resurrection of @jpsim’s [suggestion](https://github.com/realm/SwiftLint/issues/1323#issuecomment-285782712) in #1323, now that the groundwork (merging, etc.) that was missing at the time is nearly finished.

***

I propose adding two flags that insert config files into the merge tree at the very top and very bottom.

Tentative flag names could be:

- `--base-config`: The file provided here would be treated as though it were parent to the `.swiftlint.yml` file in the project’s root directory.
- `--override-config`: The file provided here would be treated as though it were child to each of the deepest nested `.swiftlint.yml` files. (This one needs a better name.)

**Edit:** Here is a diagram of the merge tree:

`--base-config`
&nbsp;&nbsp;&nbsp;&nbsp;↳ `/.swiftlint.yml`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;↳ `--override-config`

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;↳ `/A/.swiftlint.yml`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;↳ `--override-config`

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;↳ `/A/C/.swiftlint.yml`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;↳ `--override-config`

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;↳ `/B/.swiftlint.yml`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;↳ `--override-config`

[...]

Issue #987, opened 2016-12-15
Using the --path option ignores included/excluded paths in the config file


dalemy:
This change was made about 3 weeks ago: https://github.com/realm/SwiftLint/commit/3568245c1917988829e285067a600584e463d400

It makes files passed with --path ignore the config file. However, we would like to be able to lint single files but make it also respect the configuration file. For example, if it's in the ignored files, linting would just succeed without even checking the file. Likewise if it is not in the included files. 

Keeping this behaviour by default seems fine, but if the --config flag is also specified, it could overwrite the default behaviour?

marcelofabri:
There're other reasons to pass a `--config` flag, so I don't think it would be a good idea to change the behavior based on that. We probably could have another flag to keep the old behavior, but IMO you should just filter the files before calling SwifLint.

Maybe @jpsim has more thoughts about this since he wrote that fix.

jpsim:
Before 3568245, if you wanted to lint a single file like `swiftlint lint --file MyIgnoredDir/File.swift`, and `MyIgnoredDir` was ignored in the configuration, you'd get something like "Could not find any files to lint at 'MyIgnoredDir/File.swift'". If your specifying a single file in the command, it seems obvious that you want that file linted, no matter if it's ignored or not in the config...

Vel0x:
So here is the scenario we have (and hopefully it's clear why we would like slight different functionality). 

Our project is quite large, with over a thousand Swift files and well over a hundred thousand lines of Swift code. Our previous configuration was to run Swiftlint across the entire project on a build, however this caused problems since it would take well over 10 seconds to run. Our solution was to write a script which does a `git diff` and then lint only the files which have changed. This dropped the processing time from 10 seconds to a fraction of a second. The down side is that we have to build the inclusion and exclusion lists into the config file AND into this script now, which is something that we would rather avoid. 

Alternatively, if there is a better solution to this scenario, we would be glad to hear it.

dalemy:
So here is the scenario we have (and hopefully it's clear why we would like slight different functionality).

[...]

Issue #799, opened 2016-09-01
Activate rules only for specific configurations or build types


AliSoftware:
As stated in #727 there can be a interest in activating some rules ONLY in specific configurations (Debug vs Release) and/or ONLY if doing an "Analyze" (Cmd-Maj-B) but not a standard "Build" (Cmd-B).

The most interesting of these is probably to only activate rules on Analyze, this way you could still error for force-unwraps, line length or stuff like that in standard Builds, and only warn about cyclomatic complexity and TODOs in Analyze

I guess today one could imagine having a dedicated `.swiftlint-analyze.yml` file and configure its Build Phase to use `.swiftlint.yml` on standard Builds and `.swiftlint-analyze.yml` on Analyze builds, but maybe it would be even cooler if that feature was integrated in SwiftLint directly, like as an additional parameter for each rule or something?

scottrhoyt:
I suppose one way to implement this would be to create new a new severity level called analyze. During normal builds, you could filter out rules with this severity (for rules with multiple severity levels, you could remove this level). However, doing so you would still need to add a flag to `swiftlint` to let it know how to process analyze because I don’t know of any way to determine the type of build being done otherwise. So you would still need different build scripts for a normal vs. analyze build. Because of that, I think the complexity vs. utility of this is debatable because all you gain is a single configuration file.

AliSoftware:
I see your point.

New idea then: ability to easily switch to "Opt-In only" mode in configuration file, to _only_ list the rules we want to enable explicitly, without having to disable all the default ones one by one except the few ones we want.

It would then be way easier to create configurations where you want some rules to be executed for the Build action and others to be executed for Analyze action, by having the `.swiftlint.yml` config file for Build only disable `todo` and the `.swiftlint-analyze.yml` config file used for Analyze to only enable `todo`

---

```sh
if [ "${RUN_CLANG_STATIC_ANALYZER}" == "YES" ]; then
  swiftlint --config ".swiftlint-analyze.yml"
else
  swiftlint
fi
```

```sh
$ cat .swiftlint.yml
disabled_rules:
 - todo
$ cat .swiftlint-analyze.yml
enabled_rules:
 - todo
```

marcelofabri:
@AliSoftware We have `whitelist_rules` which could be what you've described on your last comment. Can we close this?

AliSoftware:
Indeed 👍

PR #1745, opened 2017-08-03
Fix picking the wrong configuration with —path

### LINKED ISSUE #1744: SwiftLint is picking the wrong configuration
I have a `.swiftlint.yml` that I've used to reproduce #1586 on my home folder. When running `swiftlint` inside a subdirectory (that has a configuration file), it's loading the configuration from my home dir.

This happens even if I use `--config`. Tested with 0.21.0.

This only happens when using `--path`. And the weird thing is that SwiftLint even prints that it's using the right configuration:

```
$ swiftlint lint --path file.swift  --config /Users/marcelofabri/dev/SwiftLint/.swiftlint.yml --no-cache
Loading configuration from '/Users/marcelofabri/dev/SwiftLint/.swiftlint.yml'
Linting Swift files at path file.swift
Linting 'file.swift' (1/1)
Invalid configuration for custom rule 'two_enters_before_marks'.
```

marcelofabri:
Fixes #1744 

Pending tests.


SDGGiesbrecht:
@marcelofabri, @jpsim

I looked at this in case it would affect #1748. (Apparently we wrote some of the same code without communicating. Haha...)

Unfortunately, I think there is more to the bug than this. The related bug report (#1744) states that the bug was already manifest in 0.21.0. The code affected by this pull request did not even exist in 0.21.0, so I doubt these changes actually get to the root of the problem. (They will not fix it for any situation that does no merging.)

I advise verifying that 0.21.0 really does exhibit the bug before proceeding.

SDGGiesbrecht:
> ...before proceeding.

I mean before assuming the bug is gone. There is nothing wrong with merging the changes proposed here.

marcelofabri:
> The code affected by this pull request did not even exist in 0.21.0, so I doubt these changes actually get to the root of the problem. (They will not fix it for any situation that does no merging.)

Actually, I this is not related to the merge of configurations. It's just how we look to parent directories if we don't find one configuration in the current one.

marcelofabri:
I'm not sure this actually works with `--config`.

Tests/SwiftLintFrameworkTests/ConfigurationTests.swift
```diff
@@ -195,6 +195,22 @@ class ConfigurationTests: XCTestCase {
...
+        let file = File(path: projectMockSwift0)!
+        XCTAssertEqual(configuration.configuration(for: file), configuration)
+    }
+
+    func testConfigurationWithSwiftFileAsRootAndCustomConfiguration() {
+        let configuration = Configuration(path: projectMockYAML0CustomPath,
+                                          rootPath: projectMockSwift0,
+                                          optional: false, quiet: true)
+        let file = File(path: projectMockSwift0)!
+        XCTAssertEqual(configuration.configuration(for: file), configuration)
```
marcelofabri: this test should pass, right?

SDGGiesbrecht: > @marcelofabri,

For starters, `Configuration`’s `Equatable` conformance is broken on master, so `XCTAssertEqual` is unreliable here. (See the comments in #1748 starting with [this one](https://github.com/realm/SwiftLint/pull/1748#issuecomment-320370564)). Although that may not be the only issue since there are also failures on macOS.

SDGGiesbrecht:
> Actually, I this is not related to the merge of configurations. It's just how we look to parent directories if we don't find one configuration in the current one.

...which was all refactored to integrate with merging (searching now needs to continue past the first configuration found). The edits here are all in `Configuration+Merging.swift`.

marcelofabri:
> ..which was all refactored to integrate with merging (searching now needs to continue past the first configuration found). The edits here are all in Configuration+Merging.swift.

They were refactored just to be in different extensions.

Tests/SwiftLintFrameworkTests/ConfigurationTests.swift
marcelofabri: This happens locally on my Mac. I'm pretty sure it's not a bug of comparison, but a consequence of the bug.

marcelofabri:
No idea why I'm getting different results on Linux and macOS 🤔

SDGGiesbrecht:
@jpsim, @marcelofabri

I’ve thought more about this. I think this is a broader issue than we initially realized. While the bug addressed here predated configuration merging, **merging is now live in master**, and configuration merging broke the very design of `--path` and `--config` (though it was definitely worth it). Now real thought has to go into what they are supposed to do before trying to fix them for the next release. Each had several different uses that were more or less identical before merging, but are now mutually exclusive. It all comes down to the fact that there are now two distinct paths involved, (a) what scope to lint and (b) where is the root of the configuration tree?

## What does `--path` mean?

### A)
User John Doe normally runs `swiftlint lint` in his project root. But at the moment he just wants to see what is wrong in a particular module and filter out the rest. He runs `swiftlint lint --path Sources/MyModule/` expecting to get the same results as normal for that module. (There is a real usage case in #987.)

A possible clearer name might be `--subdirectory`:
`swiftlint lint --subdirectory Sources/MyModule/`.

He needs the configurations to *inherit down to his project root (the working directory) and no further* in order to produce results consistent with his normal broader use of `swiftlint lint`.

Using this to point outside the working directory would be a precondition failure, as it would—unavoidably—inherit all the way from the system root.

### B)
User Jane Doe does not want to have to `cd` to her project root and back again afterward. She wants to be able to specify the location of the project. She runs `swiftlint lint --path Documents/Projects/MyProject/` expecting to get the same results as her CI which runs in the project root. (There is a real usage case in #1744.)

A possible clearer name might be `--project`:
`swiftlint lint --project Documents/Projects/MyProject/`.

She needs the configurations to *inherit down to the specified directory and no further*.

She will think it is a bug if they inherit from her working directory (home), which is outside the project. However, that “bug” was exactly what John Doe above expected, and he will think it is a “bug” if it does not happen.

(Note however that both `--project` and `--subdirectory` might be used together to do what John Doe wants, but from somewhere else in the file system.)

### C)
User Max Mustermann is integrating SwiftLint into an IDE. He wants to be able to feed Swift arbitrary code even if it is still just in some temporary directory and the user has not saved it yet. His IDE runs `swiftlint lint --path tmp/TransientFile.swift` expecting no interference from configurations anywhere in the file system (though it might provide one to the command line directly). (There is a real usage case in #1721.)

A possible clearer name might be `--isolated-file`:
`swiftlint lint --isolated-file tmp/TransientFile.swift`

Contrary to either John or Jane Doe, Max will consider it a bug if a configuration file gets imported from the same directory as the file he provided.

#### A, B and C have mutually exclusive behaviour and can no longer share the same option.

## What does `--config` mean?

### A)
User Erika Mustermann wants to be able to specify distinct configurations under different conditions. She has two configurations: `.swiftlint‐local.yml` and `.swiftlint‐ci.yml`. She specifies which one she wants at the invocation in each script: `swiftlint lint --config .swiftlint‐local.yml` or `swiftlint lint --config .swiftlint‐ci.yml`. She is less strict about the `Tests` directory, so she has a nested configuration there that she expects both invocations to respect. (There is a real usage case in #799.)

This would be covered by `--config-defaults` as implemented in #1748:
`swiftlint lint --config-defaults .swiftlint‐local.yml`
`swiftlint lint --config-defaults .swiftlint‐ci.yml`

She will think it is a bug if her nested file does not merge on top of the file she specifies.

### B)
User Jean Dupont reads the current description for `--config`, “the path to SwiftLint's configuration file” and naturally believes it will redirect SwiftLint away from the `.swiftlint.yml` files littered throughout the project. He uses `--config` with just that intention: `swiftlint lint --config ~/Documents/MyRules.yml`.

This would be covered by the combination of `--config-defaults` and `--ignore-nested-configs` as implemented in #1748:
`swiftlint lint --config-defaults ~/Documents/MyRules.yml --ignore-nested-configs`

He will be very confused if anything is merged and he will think there is a bug. If enough things are overridden, it may even appear to him as though `--config` did nothing at all.

#### A and B have mutually exclusive behaviour and can no longer share the same option.

## Possibilities for Moving Forward

None of the above usage styles work as expect from master right now. Neither would any of them fully work after merging this pull request. (Though it is not very noticeable yet, since so far no one is using SwiftLint on projects full of nested configurations designed to be merged.)

Moving forward one might:

### A)
Do nothing. (Possibly merging this pull request, but nothing more.)
- Brace for mountains of bug reports at the next release.

### B)
Stick to `--config` and `--path`:
- Choose one definition from above to lock down the semantics.
- Make the `help` description more explicit.
- Re‐implement the options so they actually do what they purport to do.
- Inform users who assumed differently that their usage will no longer work.

### C)
Split the options up for each usage, making `--config` and `--path` obsolete:
- Merge #1748 and then implement `--project`, `--subdirectory` and `--isolated-file`.
- Either:
    - a) Remove `--config` and `--path` to prevent confusion, or:
    - b) Keep them around for pseudo‐compatibility and route them each straight into one of the above clearer options.

In my opinion C)‐a) is the best option, but no matter how you decide to proceed, it will be easier to do after #1748 than before it, as #1748 fixes several existing bugs (like the Linux one @marcelofabri encountered) and does most of the work needed for several of these options.

lswith:
This is an interesting use case for both config and path. I think you should have a look at a few popular linters to figure out the best use case for these configurations. This problem has been solved before.

marcelofabri:
> as #1748 fixes several existing bugs (like the Linux one @marcelofabri encountered)

I don't think it's the same bug. If you look the log for this PR, the configurations are totally different (not just the order of rules).

SDGGiesbrecht:
> If you look the log for this PR, the configurations are totally different (not just the order of rules).

Sorry. I did not look at this log that close. You’re right then that there is another problem in addition to the one I mentioned.

The left side is the configuration for SwiftLint, the right side is the configuration for the mock project.

Does `deletingLastPathComponent` actually do anything on Linux? It is ringing a bell as one of the methods that gave me grief because it was not implemented yet. (I wish such methods would crash and say so instead of silently returning `self`, `nil` or something useless.) It may have been fixed in `Swift 3.1`, but it is as good a place to start as any.

marcelofabri:
Should we continue the discussion in #1693?

SDGGiesbrecht:
> Should we continue the discussion in #1693?

They are only loosely related. This is about fixing `--path`, no? (I mentioned `--config` because you did, but in my opinion it is a separate issue.)

I like the code you have written, @marcelofabri. It is a great improvement to predictability. I honestly suspect the only thing still stopping the tests is that `deletingLastPathComponent` has to been done manually in Swift 3 on Linux.

I was just aware that people are already using `--path` in other ways, so a lot of existing usage is broken now, not just the specific case you discovered (the examples did not come out of nowhere; see the linked issues). You do not have to cater to such usage, it is just useful to be aware of it as you work instead of discovering afterward. (In my experience, I often I wish I had known such things sooner.)

As for me, I do not—and probably never will—have a need to use `--path`, so I am satisfied any way you go.

I apologize if this has been frustrating. I truly wish you a nice day, @marcelofabri.

marcelofabri:
> (I mentioned --config because you did, but in my opinion it is a separate issue.)

Thanks for that detailed comment BTW. It made me realize that we might be overcomplicating things (see https://github.com/realm/SwiftLint/issues/1693#issuecomment-321597723).

> I honestly suspect the only thing still stopping the tests is that deletingLastPathComponent has to been done manually in Swift 3 on Linux.
I'm not so sure: we use `deletingLastPathComponent` in other places and quickly browsing https://github.com/apple/swift-corelibs-foundation gave me the impression that it's implemented. I didn't have the time to investigate deeper yet.

> I apologize if this has been frustrating. I truly wish you a nice day, @marcelofabri.

Not at all! I just wanted to keep the discussion here focused on that single issue. I do think we have issues with how configurations work (see all the issues in the previous releases).

marcelofabri:
If I disable the configuration (memory) cache, the tests pass on Linux 🤔

marcelofabri:
So, turns out mutability causes issues 😅 
Fixed in f133f84.

PR #1746, opened 2017-08-03
Add joined_default_parameter opt-in rule

ornithocoder:
to discourage explicit usage of the default separator. Implements #1093.

[x] Adds opt-in rule
[x] Adds to tests (+ Linux)
[x] Adds to changelog
[x] Updates Rules.md

### DIFF
--- a/Rules.md
+++ b/Rules.md
@@ -4850,7 +4850,12 @@ let foo = bar.joined(separator: toto)
 <summary>Triggering Examples</summary>
 
 ```swift
-let foo = bar.joined(separator: "")
+let foo = bar.joined(separator: ↓"")
+```
+
+```swift
+let foo = bar.filter(toto)
+             .joined(separator: ↓"")
 ```
 
 </details>
--- a/Rules.md
+++ b/Rules.md
@@ -41,6 +41,7 @@
 * [Implicit Getter](#implicit-getter)
 * [Implicit Return](#implicit-return)
 * [Implicitly Unwrapped Optional](#implicitly-unwrapped-optional)
+* [Joined Default Parameter](#joined-default-parameter)
 * [Large Tuple](#large-tuple)
 * [Leading Whitespace](#leading-whitespace)
 * [Legacy CGGeometry Functions](#legacy-cggeometry-functions)
@@ -4813,6 +4814,43 @@ let collection: AnyCollection<Int!>
 
 ```swift
 func foo(int: Int!) {}
+```
+
+</details>
+
+
+
+## Joined Default Parameter
+
+Identifier | Enabled by default | Supports autocorrection | Kind 
+--- | --- | --- | ---
+`joined_default_parameter` | Disabled | No | idiomatic
+
+Discouraged explicit usage of the default separator.
+
+### Examples
+
+<details>
+<summary>Non Triggering Examples</summary>
+
+```swift
+let foo = bar.joined()
+```
+
+```swift
+let foo = bar.joined(separator: ",")
+```
+
+```swift
+let foo = bar.joined(separator: toto)
+```
+
+</details>
+<details>
+<summary>Triggering Examples</summary>
+
+```swift
+let foo = bar.joined(separator: "")
 ```
 
 </details>
--- a/Source/SwiftLintFramework/Rules/JoinedDefaultRule.swift
+++ b/Source/SwiftLintFramework/Rules/JoinedDefaultRule.swift
@@ -25,7 +25,9 @@ public struct JoinedDefaultParameterRule: ASTRule, ConfigurationProviderRule, Op
             "let foo = bar.joined(separator: toto)"
         ],
         triggeringExamples: [
-            "let foo = bar.joined(separator: \"\")"
+            "let foo = bar.joined(↓separator: \"\")",
+            "let foo = bar.filter(toto)\n" +
+            "             .joined(↓separator: \"\")"
         ]
     )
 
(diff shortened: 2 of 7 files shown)


ornithocoder:
CI failed because `String+SourceKitten.swift:397` took 225ms (limit is 200ms). Is there an easy way to re-trigger that specific build?

ornithocoder:
Should this one be opt-in or enabled by default @marcelofabri/@jpsim?

marcelofabri:
I think it should be opt-in. By bet is that `joined` can be a common name.

Source/SwiftLintFramework/Rules/JoinedDefaultRule.swift
```diff
@@ -0,0 +1,58 @@
...
+                               severity: configuration.severity,
+                               location: Location(file: file, byteOffset: defaultSeparatorOffset))]
+    }
+
+    private func defaultSeparatorOffset(dictionary: [String: SourceKitRepresentable], file: File) -> Int? {
+        guard
+            let bodyOffset = dictionary.bodyOffset,
+            let bodyLength = dictionary.bodyLength else { return nil }
+
+        let body = file.contents.bridge().substringWithByteRange(start: bodyOffset, length: bodyLength)
```
marcelofabri: what do you think about using the substructure instead?

```json
 {
 	"key.namelength": 10,
 	"key.nameoffset": 1,
 	"key.length": 25,
 	"key.substructure": [{
 		"key.namelength": 9,
 		"key.nameoffset": 12,
 		"key.length": 13,
 		"key.name": "separator",
 		"key.bodyoffset": 23,
 		"key.bodylength": 2,
 		"key.kind": "source.lang.swift.expr.argument",
 		"key.offset": 12
 	}],
 	"key.name": "bar.joined",
 	"key.bodyoffset": 12,
 	"key.bodylength": 13,
 	"key.kind": "source.lang.swift.expr.call",
 	"key.offset": 1
 }
```

It should be more reliable than the current implementation.

ornithocoder: Sure. Consider it done :-)

Source/SwiftLintFramework/Rules/JoinedDefaultRule.swift
```diff
@@ -49,10 +49,14 @@ public struct JoinedDefaultParameterRule: ASTRule, ConfigurationProviderRule, Op
 
     private func defaultSeparatorOffset(dictionary: [String: SourceKitRepresentable], file: File) -> Int? {
         guard
-            let bodyOffset = dictionary.bodyOffset,
-            let bodyLength = dictionary.bodyLength else { return nil }
+            let argument = dictionary.substructure.first(where: { $0.name == "separator" }),
+            let argumentBodyOffset = argument.bodyOffset,
```
marcelofabri: you can also check if `argument.name == "separator"`

ornithocoder: I've just checked in another version which check the number of arguments and the argument name.

Source/SwiftLintFramework/Rules/JoinedDefaultRule.swift
```diff
@@ -0,0 +1,66 @@
...
+                return []
+        }
+
+        return [StyleViolation(ruleDescription: type(of: self).description,
+                               severity: configuration.severity,
+                               location: Location(file: file, byteOffset: defaultSeparatorOffset))]
+    }
+
+    private func defaultSeparatorOffset(dictionary: [String: SourceKitRepresentable], file: File) -> Int? {
+        let arguments = dictionary.substructure.filter { $0.kind == SwiftExpressionKind.argument.rawValue }
```
marcelofabri: you can use `dictionary.enclosedArguments` instead

Source/SwiftLintFramework/Rules/JoinedDefaultRule.swift
```diff
@@ -0,0 +1,66 @@
...
+            let argument = arguments.first,
+            let argumentBodyOffset = argument.bodyOffset,
+            let argumentBodyLength = argument.bodyLength,
+            argument.name == "separator"
+            else {
+                return nil
+        }
+
+        let body = file.contents.bridge().substringWithByteRange(start: argumentBodyOffset, length: argumentBodyLength)
+        return body == "\"\"" ? argumentBodyOffset : nil
```
marcelofabri: you can use `body.isEmpty` here instead

Rules.md
```diff
@@ -4819,6 +4820,43 @@ func foo(int: Int!) {}
...
+
+Identifier | Enabled by default | Supports autocorrection | Kind 
+--- | --- | --- | ---
+`joined_default_parameter` | Disabled | No | idiomatic
+
+Discouraged explicit usage of the default separator.
+
+### Examples
+
+<details>
```
marcelofabri: It looks like the examples are outdated too.

Source/SwiftLintFramework/Rules/JoinedDefaultRule.swift
ornithocoder: Done :-)

Rules.md
ornithocoder: Done :-)

marcelofabri:
@ornithocoder what do you think about making it autocorrectable? 

If you think it's easy, we can do on this PR. Otherwise we can leave it as a future improvement 👍

ornithocoder:
@marcelofabri I'd do in a different PR, if that's OK. Thanks! :-)

marcelofabri:
Sure, no problem. Thanks! 💯

Issue #1323, opened 2017-02-20
Configuration via Command Line Flags?


SDGGiesbrecht:
(Related to #1207, but more specific.)

Is there a way to specify configuration options with command line flags, **and still have them merged with any extant `.swiftlint.yml` files**, or at least some other way of achieving a similar effect?

The reason I want to do this is that I am writing a project manager for Swift, [Workspace](https://github.com/SDGGiesbrecht/Workspace), that stitches together and helps automate various tools like SwiftLint.

There are several configuration options that are necessary for SwiftLint to cooperate with the other tools, such as excluding the Swift Package Manager’s `Packages` folder. Right now I do this by writing a `.swiftlint.yml` file to the project root before running SwiftLint and deleting the file afterward. The problem with the [current set‐up](https://github.com/SDGGiesbrecht/Workspace/blob/master/Documentation/Proofreading.md#swiftlint) is that when a user wants to customize SwiftLint further by writing his or her own `.swiftlint.yml`, he or she must figure out what those basic necessary options are and include them manually.

I am looking for a way of specifying such options without touching any `.swiftlint.yml` files, so that they properly take effect no matter what, and at the same time the user can fully customize SwiftLint.

marcelofabri:
There's no way to do this currently.

jpsim:
Maybe Workspace could ship with its own `.swiftlint.yml` file, stored in the app bundle maybe, and then the invocation of SwiftLint could specify that special config?

Although from what I can tell, Workspace is mostly intended to be built via SPM, which doesn't build bundles but static executables, so you could write the config file to `~/Library/Application Support/Workspace/.swiftlint.yml`?

SDGGiesbrecht:
> Maybe Workspace could ship with its own `.swiftlint.yml` file, stored in the app bundle maybe, and then the invocation of SwiftLint could specify that special config?

@jpsim, That would work, provided:

- the hierarchy of `.swiftlint.yml` files is *merged*, not *overridden* (#676)
- The (flag‐)specified(?) `.swiftlint.yml` could be treated as a *parent* to the `.swiftlint.yml` in the project root, and not an *override*.

***

> ...from what I can tell, Workspace is mostly intended to be built via SPM...

Actually, for now it installs its own Git repository within `~/.Workspace` (a Linux‐safe path) so that it can use resources and update itself easily.

jpsim:
Yes, you're correct on all counts there.

SDGGiesbrecht:
[...]

PR #1748, opened 2017-08-04
Command Line Options for Interacting with the Configuration Merging Process

SDGGiesbrecht:
(From #1693, expanded from @jpsim’s suggestions in #1323.)

#### Adds three command line options:

- `--config-defaults`: the path of an external configuration file to use as the root of the merge tree.
- `--config-overrides`: the path of an external configuration file to append to the end of each branch of the merge tree
- `--ignore-nested-configs`: ignores nested configuration files

(The combination of `--config-defaults` and `--ignore-nested-configs` can be used to make `--config-defaults` stomp anything present in the project, even the root directory.)

#### Commit breakdown:

The interaction with the merge logic is isolated in the second commit. (The first commit simply adds the options to the interface and gets them to `Configuration`.)

#### Other Notes:

I left `--config` alone for now, but I think this pull request would make it obsolete. (In the wake of #1687 it looks like it merges itself with anything it finds, all the way back to the system root, which I doubt anyone expects.)

The read‐me is out of date. It still says nested files trump their parents. I left it alone in case you want it in sync with the last stable release and not master.

In the process of writing this, I discovered that the existing merge process went backwards. A tree like this...
A
↳ B
&nbsp;&nbsp;↳ C
&nbsp;&nbsp;&nbsp;&nbsp;↳ D
...was being merged as A is overridden by D, which is overridden by C, which is overridden by B.
The refactors in this pull request eliminate the problem, but if this pull request is not accepted, that bug will have to be dealt with elsewhere.

Maintainers have been given direct write access.

(@Aranoledur, you were tracking the related issue, so you might want to look at this.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@
   [JP Simard](https://github.com/jpsim)
   [#676](https://github.com/realm/SwiftLint/issues/676)
 
+  * Several command line options allow interaction with the merging process.
+    (See `swiftlint help lint`.)  
+    [Jeremy David Giesbrecht](https://github.com/SDGGiesbrecht)
+    [##1693](https://github.com/realm/SwiftLint/issues/1693)
+
 ##### Enhancements
 
 * Add `xctfail_message` rule to enforce XCTFail
--- a/Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
@@ -18,29 +18,57 @@ extension Configuration {
     }
 
     private func configuration(forPath path: String) -> Configuration {
-        if path == rootPath {
+        if ignoreNested {
+            if let commandLineDefaults = defaults?.value {
+                // --config-defaults is lowest in the tree
+                return commandLineDefaults
+            } else {
+                // when no --config-defaults, the root directory is lowest in the tree
+                return self
+            }
+        }
+
+        var merged = mergeConfigurationFiles(upTo: path)
+        if let commandLineDefaults = defaults?.value {
+            // add --config-defaults underneath
+            merged = commandLineDefaults.merge(with: merged)
+        }
+        if let commandLineOverrides = overrides?.value {
+            // add --config-overrides on top
+            merged = merged.merge(with: commandLineOverrides)
+        }
+        return merged
+    }
+
+    private func mergeConfigurationFiles(upTo path: String) -> Configuration {
+        if path == rootPath || path == "/" {
             return self
         }
 
         let pathNSString = path.bridge()
         let configurationSearchPath = pathNSString.appendingPathComponent(Configuration.fileName)
+        if configurationSearchPath == configurationPath {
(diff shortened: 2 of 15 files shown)


SDGGiesbrecht:
Since I guess I cannot request reviews directly, this is a callout to you, @marcelofabri.

SDGGiesbrecht:
Any idea why only Linux would fail to accurately compare two `Configuration`s?

SDGGiesbrecht:
The code coverage report is misleading. The two affected files in `SwiftLintFramework` improved. Overall coverage fell because 5⁄9 of the affected files are in the `swiftlint` executable and therefore untestable.

SDGGiesbrecht:
The single Linux failure only occurred in 3/4 times with no changes to the relevant code. (See the CI logs for the four separate commits above).

The test is older than this pull request. It is an `XCTAssertEqual()` that compares two `Configuration` files. I manually compared the two configurations in the most recent [log](https://travis-ci.org/realm/SwiftLint/jobs/261171589#L860). The only difference between their descriptions is that two of the active rules appear in a different order in the list (LineLengthRule and DiscardedNotificationCenterObserverRule).

It reminds me of chatter a while ago about a similar issue on Linux to do with caches not matching because set/dictionary iteration order is not deterministic. Since sets are being used in [`mergingRules`](https://github.com/realm/SwiftLint/blob/54d619977c046b45799d5139852b31f50ada067e/Source/SwiftLintFramework/Extensions/Configuration%2BMerging.swift#L60) (which this pull request does not touch), and never sorted, checking the resulting arrays for equality is unpredictable, which is what `Configuration` does in its `==`.

Adding a sort step would slow things down, but I notice `Configuration`’s `Equatable` conformance is never used by Swiftlint itself, but only by the tests, so maybe it would be best to do the sorting there:
```swift
(lhs.rules == rhs.rules)
```
↓
```swift
(lhs.rules.sorted(by: { String(describing: $0) < String(describing: $1) })
    == rhs.rules.sorted(by: { String(describing: $0) < String(describing: $1) }))
```

Thoughts? Or do you want to keep that issue distinct from this pull request?

SDGGiesbrecht:
I will push the above fix for the Linux tests. (If nothing else, for the nice green checkmark.) Since maintainers have write access to the branch, if you decide a different solution would be better, you can go ahead make the change.

SDGGiesbrecht:
Everything is ready now.

SDGGiesbrecht:
Merge conflicts have accumulated since the original proposal. Since this pull request has never received any reviews or comments, the work of updating it does not seem worth it. I am therefore closing it.

(I will wait a week before I actually delete my branch in case there really are others out there who want something like this.)

Aranoledur:
Well, my company need this too, but guys in charge don't think it's a good feature. Have to find another solution

SDGGiesbrecht:
@Aranoledur, you are welcome to pull my changes to your own fork if you want to maintain one. Just do it sooner rather than later, as I will not likely keep my fork around forever.

PR #1755, opened 2017-08-07
Support non Swift files in `--file` argument

### LINKED ISSUE #1721: Swiftlint only expects files with type ".swift"

marcelofabri:
Fixes #1721.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@
   [Ornithologist Coder](https://github.com/ornithocoder)
   [#1093](https://github.com/realm/SwiftLint/issues/1093)
 
+* Files with extensions other than `.swift` can now be used as argument
+  to `--file` when linting or autocorrecting.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#1721](https://github.com/realm/SwiftLint/issues/1721)
+
 ##### Bug Fixes
 
 * Fix false positive on `force_unwrapping` rule when declaring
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Extensions/NSFileManager+SwiftLint.swift
```diff
@@ -21,7 +21,7 @@ extension FileManager: LintableFileManager {
             .standardizingPath
 
         // if path is a file, it won't be returned in `enumerator(atPath:)`
-        if absolutePath.bridge().isSwiftFile() && absolutePath.isFile {
+        if absolutePath.isFile {
```
jpsim: why this change? I think the `lintablePaths` change is sufficient to support the `lint --path file.notswift` scenario.

marcelofabri: I guess that's true, but when exactly would this check be true then? 🤔

jpsim: when specifying `included` and `excluded` configuration values, which IMO doesn't really make sense to specify non-Swift files in there.

jpsim: I'm kinda worried that if we extend `included`/`excluded` to support globbing, this change would attempt to lint non-Swift files if `Sources/**` were specified, for example.

marcelofabri: I've removed it 👍

PR #1756, opened 2017-08-07
Help debug danger/danger#873

jpsim:
by printing `oss-check` output to logs when running Danger to debug https://github.com/danger/danger/issues/873


jpsim:
Unfortunately this didn't trigger danger/danger#873, but I'll merge this in the hopes that a future PR does, and then we can audit the logs for hints.

PR #1760, opened 2017-08-10
Support Optional<()> returns that use syntactic sugar, ()?

ryanbooker:
https://github.com/realm/SwiftLint/issues/1761

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1721](https://github.com/realm/SwiftLint/issues/1721)
 
+* Sugared `Optional<()>` return types,  `()?`, Void?, `()!`, and Void!
+  no longer triggr the `redundant_void_return` rule.  
+  [Ryan Booker](https://github.com/ryanbooker)
+  [#1761](https://github.com/realm/SwiftLint/issues/1761)
+ 
 ##### Bug Fixes
 
 * Fix false positive on `force_unwrapping` rule when declaring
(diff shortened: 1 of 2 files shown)


marcelofabri:
@ryanbooker Can you share more information about the motivation for this change? 

IMO `-> ()?` and `-> ()!` should trigger (`-> Void?` and `-> Void!` being preferred).

jpsim:
Moreover, why would you ever want to return `Void?` or `Void!`? That binary state is (almost?) always better represented by `Bool` or a two-member enum.

ryanbooker:
e.g. `traverse_` on Optional returns `Optional<()>`/`()?`/`Void?`.

The point is it's a valid return type and not the same as `()`/`Void`. It's not about whether you should return `()` or `Void`, that's clearly personal preference. It's a bug to reject `()?` or `Void?` as `Void`.

:)

ryanbooker:
Perhaps there is another way I'm supposed to specify the test, but only the `()?` and `()!` return types, specifically, run afoul of the current regexp. I can add a test specifically for those cases if you like.

ryanbooker:
@marcelofabri In case I wasn't clear. It's simply incorrect at the moment. It's irrelevant whether `()` or `Void` is preferred. That's a settings in the linter.

The bug is that `()?` and `()!` are rejected as being equivalent to `()`/`Void`. They aren't and there are useful abstractions that can return and `Optional<()>`. It should required doing a dance with the `void_return`, `redundant_void_return`, `syntactic_sugar` rules in order to return a perfectly valid and non redundant type.

:)

marcelofabri:
> It's irrelevant whether `()` or `Void` is preferred. That's a settings in the linter.

Actually, this is the whole purpose of this rule. To make sure `Void` is preferred.

What I *think* would be the better approach is to change `redundant_void_return` to allow returning `Void!`, `Void?`, `()?` and `()!`. Although I agree that's not common to return those types, it's not that rule's job to enforce that (it's only concerned about the cases that the whole return type can be omitted). 

Does that make sense, @ryanbooker? Or am I missing something else here?

ryanbooker:
No we’re on the same page now. That’s precisely what I intended to do. I did it in the wrong rule. Now I understand why you’re talking about irrelevant things. Haha. :)

ryanbooker:
Let's try this again. :)

marcelofabri:
Thanks! 💯

jpsim:
👍

PR #1768, opened 2017-08-11
Validate implicit subscript get

### LINKED ISSUE #898: Validate implicit `subscript` getter
Read-only subscripts should avoid using the get keyword.

See #57 and #862 for some history.

marcelofabri:
Fixes #898

This depends on https://github.com/apple/swift/pull/11258 (which is already available on latest development snapshots).

I don't know how should we proceed with this one. Should we merge right away? Or wait for that PR to be on an official release?

This also needs a CHANGELOG entry.


jpsim:
Let's revisit when Swift 4.1 is out?

marcelofabri:
In theory we could make it already available for people using preview toolchains, but I guess no one actually does that. I'll close it by now, but keep the branch alive.

PR #1772, opened 2017-08-12
Add full Build Shell Script for CocoaPods projects

kgleong:


kgleong:
IMO, it's easier for setup to list the entire cocoapods variant.

marcelofabri:
I don't see the need to check if the binary exists when integrating via CocoaPods. If you don't have the binary, the Podfile.lock won't be in sync with the sandbox anyway, which is an error.

kgleong:
ah makes sense, will decline thanks!

PR #1773, opened 2017-08-12
Add instructions on how to use SwiftLint with fastlane

KrauseFx:


README.md
```diff
@@ -108,6 +108,20 @@ To integrate SwiftLint with [Atom](https://atom.io/), install the
...
 APM.
 
+### fastlane
+
+You can use the [official swiftlint fastlane action](https://docs.fastlane.tools/actions/#swiftlint) to run SwiftLint as part of your fastlane process.
+
+```ruby
+swiftlint(
+  mode: :lint,                          # SwiftLint mode: :lint (default) or :autocorrect
+  output_file: "swiftlint.result.json", # The path of the output file (optional)
```
marcelofabri: what do you think about adding `reporter` too so people know that it's not just a matter of using `.json`?

I also think adding `executable` is important since integration via CocoaPods is now available.

KrauseFx: Yeah, I'm ok with changing it. I just took the example we have in the _fastlane_ repo https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/swiftlint.rb#L132

marcelofabri:
Thanks @KrauseFx 🚀

README.md
marcelofabri: I've added them in  1a50451.

KrauseFx:
Thanks for merging 👍

PR #1776, opened 2017-08-14
Improve attribute handling for let_var_whitespace

### LINKED ISSUE #1770: let_var_whitespace should not trigger for @objc
I like the new `let_var_whitespace ` rule but I think it shouldn't trigger for `@objc` because it applies to that member. E.g. the following snippet triggers a violation:

```
    @objc(status)
    var statusObjc: String {
//...
```

What do you think?

Uncommon:
Fixes #1770

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,10 +120,12 @@
   `unneeded_parentheses_in_closure_argument` rule that were generated in some
   cases (mainly when using chained method calls with closures).  
   [Marcelo Fabri](https://github.com/marcelofabri)
+
 * Fix handling of attributes (`lazy`, `objc`, etc.) for the `let_var_whitespace`
   rule.  
   [David Catmull](https://github.com/Uncommon)
   [#1770](https://github.com/realm/SwiftLint/issues/1770)
+  [#1812](https://github.com/realm/SwiftLint/issues/1812)
 
 ## 0.21.0: Vintage Washboard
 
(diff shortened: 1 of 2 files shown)


marcelofabri:
I'm not sure this is the correct fix. You can see in https://github.com/realm/SwiftLint/pull/1776#issuecomment-322221705 that some fixed violations should actually be triggered.

Uncommon:
Hey, good catch. I'll work on that. This means that the handling of `@available` wasn't quite right either.

Uncommon:
I'll see how the CI goes and then rebase to fix the merge conflict.

CHANGELOG.md
```diff
@@ -44,6 +44,11 @@
   [Hossam Ghareeb](https://github.com/hossamghareeb)
   [#1763](https://github.com/realm/SwiftLint/issues/1763)
 
+* Fix handling of attributes (`lazy`, `objc`, etc.) for the `let_var_whitespace`
+  rule.
```
ornithocoder: Hi @Uncommon. Don't forget the two trailing spaces after the description. More info [here](https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes)

Uncommon: Fixed and rebased.

Uncommon:
Turns out it was my fault the tests were taking too long. Fixed.

Uncommon:
Still needs work. It doesn't recognize attributes with underscores, or multiple lines of attributes before a declaration.

Uncommon:
I looked through a bunch of the OS violation changes, and they look good to me.

Uncommon:
Any more comments on this?

marcelofabri:
Can you rebase this? It looks like it also fix #1812, so could you also add that as a non-triggering example and add the issue number to the changelog entry? Thanks!

Uncommon:
Rebased and updated.

Issue #1779, opened 2017-08-16
Limit number of test classes in file to one


ornithocoder:
One of the things I like the most about Quick is the possibility to isolate scenarios in nested blocks to better describe the behavior of the system under test. 

As result, rare are the cases were one would need to implement two or more QuickSpecs in the same test file. A single QuickSpec per file is enough.

rule request, single quick spec per file, opt-in

marcelofabri:
Should this also be applied to `XCTestCase`?

ornithocoder:
@marcelofabri I'm not sure. Since one might want to test the system under test with different parameters, multiple `XCTestCase` classes inside the same file might be needed (one `setUp` and one `tearDown` might now be enough for all the cases). Using Quick, these are different `describes` or `contexts` within the same QuickSpec class.

marcelofabri:
But then one could do the setup in each test or even split the `XCTestCase` into another file. I don't think I've ever seen a project that had two (or more) `XCTestCase` in the same file.

ornithocoder:
IMO setup should never happen inside the test. Not when using Quick/Cedar/etc.. at least. The 'it' block should be only for verifying the expectation.

I can either change this rule to cover XCTestCase or create a new (and similar rule) for XCTestCase.

ornithocoder:
PR updated :-)

marcelofabri:
Closed in #1780

PR #1780, opened 2017-08-16
Add single_test_class opt-in rule

ornithocoder:
Implements #1779 (Limit number of QuickSpecs and XCTestCases in file to one).


Source/SwiftLintFramework/Rules/SingleTestClassRule.swift
```diff
@@ -0,0 +1,68 @@
...
+
+    private func testClasses(in file: File) -> [[String: SourceKitRepresentable]] {
+        return file.structure.dictionary.substructure.filter {
+            !$0.inheritedTypes.filter { testClasses().contains($0) }.isEmpty
+        }
+    }
+
+    private func toViolation(in file: File,
+                             configuration: SeverityConfiguration,
+                             numberOfSpecs: Int) -> ([String: SourceKitRepresentable]) -> StyleViolation? {
```
marcelofabri: change `numberOfSpecs` to `numberOfTestClasses` (or any other name to avoid using `specs`)

Source/SwiftLintFramework/Rules/SingleTestClassRule.swift
```diff
@@ -0,0 +1,68 @@
...
+
+    // MARK: - Private
+
+    private func testClasses(in file: File) -> [[String: SourceKitRepresentable]] {
+        return file.structure.dictionary.substructure.filter {
+            !$0.inheritedTypes.filter { testClasses().contains($0) }.isEmpty
+        }
+    }
+
+    private func toViolation(in file: File,
```
marcelofabri: what do you think about changing this method name? For me, `toSomething` usually means that the method will use `self` as input and convert it to `Something`.

Source/SwiftLintFramework/Rules/SingleTestClassRule.swift
```diff
@@ -0,0 +1,68 @@
...
+                             configuration: SeverityConfiguration,
```
marcelofabri: you don't need to pass the configuration here since it's a property anyway

Source/SwiftLintFramework/Rules/SingleTestClassRule.swift
```diff
@@ -0,0 +1,68 @@
...
+        return { dictionary in
+            guard let offset = dictionary.offset else { return nil }
+            return StyleViolation(ruleDescription: type(of: self).description,
+                                  severity: configuration.severity,
+                                  location: Location(file: file, byteOffset: offset),
+                                  reason: "\(numberOfSpecs) test classes found in this file.")
+        }
+    }
+
+    private func testClasses() -> [String] {
```
marcelofabri: this could be a `let` to avoid being created everytime (or even declared in `testClasses(in:)`).

Source/SwiftLintFramework/Rules/SingleTestClassRule.swift
marcelofabri: Actually, just now I've realized that it returns a function. Do we really need that? I think it just makes harder to read and the benefits are not so evident.

Source/SwiftLintFramework/Rules/SingleTestClassRule.swift
```diff
@@ -0,0 +1,68 @@
...
+        guard classes.count > 1 else { return [] }
+
+        return classes.flatMap(toViolation(in: file, configuration: configuration, numberOfSpecs: classes.count))
+    }
+
+    // MARK: - Private
+
+    private func testClasses(in file: File) -> [[String: SourceKitRepresentable]] {
+        return file.structure.dictionary.substructure.filter {
+            !$0.inheritedTypes.filter { testClasses().contains($0) }.isEmpty
```
marcelofabri: you probably should check if the type is a class as well (just in case ¯\_(ツ)_/¯)

Source/SwiftLintFramework/Rules/SingleTestClassRule.swift
ornithocoder: agree. it's a simple function and doesn't need to be isolated. changing it to a closure.

Source/SwiftLintFramework/Rules/SingleTestClassRule.swift
ornithocoder: Done

Source/SwiftLintFramework/Rules/SingleTestClassRule.swift
ornithocoder: Done :-)

Source/SwiftLintFramework/Rules/SingleTestClassRule.swift
ornithocoder: not applicable anymore.

Source/SwiftLintFramework/Rules/SingleTestClassRule.swift
ornithocoder: not applicable anymore.

marcelofabri:
Thanks @ornithocoder! 🚀

ornithocoder:
You're welcome @marcelofabri! :-) Thank you for the code review!

PR #1788, opened 2017-08-19
Explicit Enum Raw Value Rule

### LINKED ISSUE #1778: Require raw values in enum
The opposite of our current rule that triggers a violation if the same string values as enum member names were used, resulting in redundant duplication, would sometimes be useful if you want to allow refactoring of the Swift API without breaking the API contract.

Mazyod:
Hey, as @marcelofabri suggested, this was a good rule for new contributors to get their hands dirty (assuming I did this with reasonable correctness!).

I feel the PR is self-explanatory, but just to reiterate:
1. Followed the contributing guideline points (except linux tests, will test on linux now)
2. The rule is opt-in, and enforces the existence of `init_expr`
3. Tested enums with associated types, `Int` enums, and `String` enums.

Fixes #1778

### DIFF
--- a/Source/SwiftLintFramework/Rules/ExplicitAssociatedEnumValueRule.swift
+++ b/Source/SwiftLintFramework/Rules/ExplicitAssociatedEnumValueRule.swift
@@ -57,7 +57,7 @@ public struct ExplicitAssociatedEnumValueRule: ASTRule, OptInRule, Configuration
 
         let locs = substructureElements(of: dictionary, matching: .enumcase)
             .flatMap { substructureElements(of: $0, matching: .enumelement) }
-            .flatMap(enumElementsMissingInitExpr(_:))
+            .flatMap(enumElementsMissingInitExpr)
             .flatMap { $0.offset }
 
         return locs
@@ -66,7 +66,7 @@ public struct ExplicitAssociatedEnumValueRule: ASTRule, OptInRule, Configuration
     private func substructureElements(of dict: [String: SourceKitRepresentable],
                                       matching kind: SwiftDeclarationKind) -> [[String: SourceKitRepresentable]] {
         return dict.substructure
-            .filter { $0.kind.flatMap(SwiftDeclarationKind.init(rawValue:)) == kind }
+            .filter { $0.kind.flatMap(SwiftDeclarationKind.init) == kind }
     }
 
     private func enumElementsMissingInitExpr(
--- a/Source/SwiftLintFramework/Rules/ExplicitAssociatedEnumValueRule.swift
+++ b/Source/SwiftLintFramework/Rules/ExplicitAssociatedEnumValueRule.swift
@@ -24,9 +24,7 @@ public struct ExplicitAssociatedEnumValueRule: ASTRule, OptInRule, Configuration
             "enum Numbers: Int {\n case one = 1\n case two = 2\n}\n",
             "enum Numbers: Double {\n case one = 1.1\n case two = 2.2\n}\n",
             "enum Numbers: String {\n case one = \"one\"\n case two = \"two\"\n}\n",
-            "enum Numbers: String {\n case one = \"one\", two = \"two\"\n}\n",
-            "enum Numbers: String {\n case one = \"ONE\"\n case two = \"TWO\"\n}\n",
-            "enum Numbers: String {\n case one = \"ONE\"\n case two = \"two\"\n}\n"
+            "enum Numbers: String {\n case one = \"ONE\"\n case two = \"TWO\"\n}\n"
         ],
         triggeringExamples: [
             "enum Numbers: Int {\n case one = 10, ↓two, three = 30\n}\n",
--- a/Source/SwiftLintFramework/Rules/ExplicitAssociatedEnumValueRule.swift
+++ b/Source/SwiftLintFramework/Rules/ExplicitAssociatedEnumValueRule.swift
@@ -1,0 +1,80 @@
+//
+//  ExplicitAssociatedEnumValueRule.swift
+//  SwiftLint
+//
+//  Created by Mazyad Alabduljaleel on 8/19/17.
+//  Copyright © 2017 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+
+public struct ExplicitAssociatedEnumValueRule: ASTRule, OptInRule, ConfigurationProviderRule {
+    public var configuration = SeverityConfiguration(.warning)
+    
+    public init() {}
+    
+    public static let description = RuleDescription(
+        identifier: "explicit_associated_enum_value",
+        name: "Explicit Associated Enum Value",
+        description: "Enums should be explicitly assigned their associated values.",
+        kind: .idiomatic,
+        nonTriggeringExamples: [
+            "enum Numbers {\n case int(Int)\n case short(Int16)\n}\n",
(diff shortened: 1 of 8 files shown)


Source/SwiftLintFramework/Rules/ExplicitAssociatedEnumValueRule.swift
```diff
@@ -0,0 +1,80 @@
...
+                           severity: configuration.severity,
+                           location: Location(file: file, byteOffset: $0))
+        }
+    }
+
+    private func violatingOffsetsForEnum(dictionary: [String: SourceKitRepresentable], file: File) -> [Int] {
+
+        let locs = substructureElements(of: dictionary, matching: .enumcase)
+            .flatMap { substructureElements(of: $0, matching: .enumelement) }
+            .flatMap(enumElementsMissingInitExpr(_:))
```
ornithocoder: You can safely remove the `(_:)` here.

Source/SwiftLintFramework/Rules/ExplicitAssociatedEnumValueRule.swift
```diff
@@ -0,0 +1,80 @@
...
+            .flatMap { $0.offset }
+
+        return locs
+    }
+
+    private func substructureElements(of dict: [String: SourceKitRepresentable],
+                                      matching kind: SwiftDeclarationKind) -> [[String: SourceKitRepresentable]] {
+        return dict.substructure
+            .filter { $0.kind.flatMap(SwiftDeclarationKind.init(rawValue:)) == kind }
```
ornithocoder: Same here for `(rawValue:)`

Source/SwiftLintFramework/Rules/ExplicitAssociatedEnumValueRule.swift
Mazyod: very cool! Thanks

Source/SwiftLintFramework/Rules/ExplicitAssociatedEnumValueRule.swift
```diff
@@ -0,0 +1,78 @@
...
+import Foundation
+import SourceKittenFramework
+
+public struct ExplicitAssociatedEnumValueRule: ASTRule, OptInRule, ConfigurationProviderRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "explicit_associated_enum_value",
```
marcelofabri: this is not a good name, because when talking about enums, associated values are generally referred to cases with "parameters". Instead, we should use "raw values"

Source/SwiftLintFramework/Rules/ExplicitAssociatedEnumValueRule.swift
```diff
@@ -0,0 +1,78 @@
...
+    )
+
+    public func validate(file: File, kind: SwiftDeclarationKind,
+                         dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard kind == .enum else {
+            return []
+        }
+
+        // Check if it's an associated value enum
+        guard !dictionary.inheritedTypes.isEmpty else {
```
marcelofabri: this is not enough, because enums can implement protocols, for example:

```swift
protocol Foo {}
enum Bar: Foo {
  case bar
}
```

Instead, we should keep a list of possible raw types that support implicit raw values. As far as I know, that means `String` and all numeric types (`Int`, `Float`, `Int64`, ...). Types like `Character` conform to `RawRepresentable`, but they need an explicit value.

Source/SwiftLintFramework/Rules/ExplicitAssociatedEnumValueRule.swift
```diff
@@ -0,0 +1,78 @@
...
+
+        let violations = violatingOffsetsForEnum(dictionary: dictionary, file: file)
+        return violations.map {
+            StyleViolation(ruleDescription: type(of: self).description,
+                           severity: configuration.severity,
+                           location: Location(file: file, byteOffset: $0))
+        }
+    }
+
+    private func violatingOffsetsForEnum(dictionary: [String: SourceKitRepresentable], file: File) -> [Int] {
```
marcelofabri: you don't actually use the `file` parameter here

Source/SwiftLintFramework/Rules/ExplicitAssociatedEnumValueRule.swift
Mazyod: I thought those were associated types, since jpsim called them associated values. But sure, I don't have any attachments to the name :)

marcelofabri: Just for future reference: https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Enumerations.html#//apple_ref/doc/uid/TP40014097-CH12-ID148

Mazyod:
I guess we should add `CodingKey` to that list?
https://github.com/apple/swift/blob/447dce6c1f1e3daeee3c5c084ea2446aa1024f6b/test/decl/protocol/special/coding/enum_coding_key.swift

Mazyod:
After reading more on `CodingKey`, I think we will only need to check against `CodingKey` conformance to trigger this warning in Swift 4.

Source/SwiftLintFramework/Rules/ExplicitEnumRawValueRule.swift
```diff
@@ -0,0 +1,89 @@
...
+        guard kind == .enum else {
+            return []
+        }
+
+        // Check if it's an enum which supports raw values
+        let implicitRawValueTypes: [Any] = [
+            Int.self, Int8.self, Int16.self, Int32.self, Int64.self,
+            UInt.self, UInt8.self, UInt16.self, UInt32.self, UInt64.self,
+            Double.self, Float.self, Float80.self, Decimal.self, NSNumber.self,
+            NSDecimalNumber.self, "NSInteger", String.self
```
marcelofabri: TIL that you can use `NSNumber` and other types as raw values 😅

marcelofabri:
> I guess we should add CodingKey to that list?

I don't think so, because to add an explicit value, you still need to "inherit" from a raw value type.

Source/SwiftLintFramework/Rules/ExplicitEnumRawValueRule.swift
```diff
@@ -0,0 +1,89 @@
...
+        return violations.map {
+            StyleViolation(ruleDescription: type(of: self).description,
+                           severity: configuration.severity,
+                           location: Location(file: file, byteOffset: $0))
+        }
+    }
+
+    private func violatingOffsetsForEnum(dictionary: [String: SourceKitRepresentable]) -> [Int] {
+
+        let locs = substructureElements(of: dictionary, matching: .enumcase)
```
jpsim: Could you please combine the 3 `flatMap`s? I'd also recommend not storing the result in an intermediate variable, since it's unused other than just being returned right away, so doesn't make things any clearer IMHO.

Mazyod: Sure, I can see this logic using a single loop with `guard`s and `continue`

jpsim: Nevermind, I don't think my suggestion would improve things after all.

PR #1789, opened 2017-08-19
No Grouping Extension Rule

### LINKED ISSUE #1767: Rule Request: Warning for extensions used for grouping code
Swift Extensions are often times used in 3 ways (there are more):
1. Retroactive Modeling
2. Protocol Extensions
3. For organization purposes, where grouping a set of methods together within the same file that declared that class. For example, grouping private methods in an extension, or an extension to implement each protocol that the class implements. 

The use of swift extensions for with regards to point 3 above is very trendy right now and I feel strongly that it is incorrect and has drawbacks. A simple `// MARK:` achieves the “code organization” intention of using extensions. 

Reasons to not use extensions for organizing code within the same file:
- Unlike MARKs, they do not show up in the source editor summary dropdown in Xcode. (Screenshot: https://goo.gl/T5vkr7)
- Must scroll through the file to find all protocol conformances
- Extensions obviously can’t override methods, hence this limits how these groups are composed
- Can’t add properties to extensions, so grouping is somewhat false and not always fully contained as they may suggest
- Take more space than MARKs and not as concise
- Lead to longer compile times (I am finding a source for this)
- Myth: Extensions don’t fully force protocol conformance to be in the scope of that extension
- Myth: They do not give ability to implement protocol privately

Objective-c has categories offer ability to be named yet this pattern isn’t trendy there, suddenly its trendy in Swift.

Extensions are awesome, so to be clear, this proposed swift lint rule would only apply to this trendy usage of extensions for the above drawbacks mentioned above.

Mazyod:
Hello again!

This attempts to implement the opt-in rule for warning against grouping extensions (couldn't think of a better name).
1. Traverse the AST once, creating a subtype `Element` which we can use.
2. While traversing, note down the namespace for later checking.
3. Check the `enum`, `struct`, & `class` types against the extension names.

Fixes #1767


Mazyod:
I can see a good configuration for this might be `include_nested_types`.

marcelofabri:
> I can see a good configuration for this might be include_nested_types. Extensions on nested types are helpful for keeping things flat.

I think this can be a future improvement, since (in my experience) it's not common to have extensions on nested types that you own.

Source/SwiftLintFramework/Rules/GroupingExtensionBanRule.swift
```diff
@@ -0,0 +1,95 @@
...
+                elements += findAllElements(in: subDict, ofTypes: types, namespace: [element.name])
+
+                return elements
+            }
+    }
+}
+
+extension GroupingExtensionBanRule {
+
+    struct Element {
```
marcelofabri: what do you think about making this a private top level declaration? This declaration shouldn't be visible for other classes anyway.

Source/SwiftLintFramework/Rules/GroupingExtensionBanRule.swift
```diff
@@ -0,0 +1,95 @@
...
+            .filter { $0.kind == .extension && susceptibleNames.contains($0.name) }
+            .map {
+                StyleViolation(ruleDescription: type(of: self).description,
+                               severity: configuration.severity,
+                               location: Location(file: file, byteOffset: $0.offset))
+            }
+    }
+
+    private func findAllElements(in dictionary: [String: SourceKitRepresentable],
+                                 ofTypes types: [SwiftDeclarationKind],
```
marcelofabri: maybe change `ofTypes` to `of`?

Source/SwiftLintFramework/Rules/GroupingExtensionBanRule.swift
```diff
@@ -0,0 +1,95 @@
...
+                                 namespace: [String] = []) -> [Element] {
+
+        return dictionary.substructure
+            .flatMap { subDict -> [Element] in
```
marcelofabri: do you need to break line here (before `.flatMap`)

Source/SwiftLintFramework/Rules/GroupingExtensionBanRule.swift
marcelofabri: Also, what do you think about changing `types` to `Set<SwiftDeclarationKind>`?

Mazyod:
> Also, I'm not a big fan of this rule name, since "ban" is too negative and strong IMO. Can we think about something else? Maybe NoGroupingExtensionRule?

Fair enough. My personal opinion is that Swiftlint needs a common suffix for discouraged (banned) practices, in order to easily navigate the rules and add consistency to naming. Starting rules with a common word like "No" and "Discourage" might group those rules and hide the actual name "GroupingExtension".

I will proceed with your suggestion now, but thought I shed light on the issue above for possible future enhancements.

Mazyod:
Weekend is finally here! All requested changes are done. Btw, what do you think about the `sourcery` version header? Didn't know what to do with it.

marcelofabri:
@Mazyod could you just rebase to fix the conflict? Thanks!

Mazyod:
TIL how to resolve conflicts right from within github 😅  .. I pulled and tested, now getting file length lint error. I think I should resolve that?

Mazyod:
Adding `// swiftlint:disable type_body_length` to `RulesTests.swift` resolved the `swift test` failure. Should I commit that?

marcelofabri:
> Should I commit that?

Yeah, we can do that by now 🤷‍♂️

marcelofabri:
Also, can you try rebasing to get rid of 4c7f262? 😬

Mazyod:
Ah, I thought you guys just squash and merge. @marcelofabri is that an option?

marcelofabri:
I can do that 👍

Tests/SwiftLintFrameworkTests/RulesTests.swift
```diff
@@ -133,6 +135,10 @@ class RulesTests: XCTestCase {
         verifyRule(FunctionParameterCountRule.description)
     }
 
+    func testNoGroupingExtension() {
```
marcelofabri: just one more thing: we try to keep this list sorted. Can you please move it (and run `make sourcery` again)? 😅

Mazyod: Of course! Sorry about that.

Issue #1709, opened 2017-07-25
Clear cache when used toolchain changes


marcelofabri:
We probably should store the Swift toolchain used for running SwiftLint and clear the cache when it changes.

More information on how SwiftLint deals with multiple Swift versions: https://github.com/realm/SwiftLint#working-with-multiple-swift-versions

jpsim:
You're technically right, but I think this can be super low priority, because the only time this should matter is when both the following conditions are met:

1. No changes are required to your source code to compile across both toolchain versions; and
2. The different toolchains would produce different (valid) violations

I think in practice it's very unlikely that both these conditions will be met simultaneously.

marcelofabri:
Agreed.

jpsim:
As highlighted by #1741, this can be hit more easily than I had originally assumed.

marcelofabri:
Is there a way to know which toolchain SourceKitten is using? Or would this require some changes there as well?

jpsim:
There's no way currently to do that, no 😞

Exposing these declarations/types publicly would be sufficient, but a terrible way, to expose that to SwiftLint:

https://github.com/jpsim/SourceKitten/blob/0.18.0/Source/SourceKittenFramework/library_wrapper_sourcekitd.swift#L9
and
https://github.com/jpsim/SourceKitten/blob/0.18.0/Source/SourceKittenFramework/library_wrapper.swift#L12

marcelofabri:
@jpsim But that actually may not be enough. For example, if someone updates Xcode, the toolchain path will be the same, won't it?

PR #1790, opened 2017-08-20
Invalidate cache when Swift version changes

marcelofabri:
This is a compromise until #1709 is properly resolved.

### DIFF
--- a/Source/SwiftLintFramework/Models/LinterCache.swift
+++ b/Source/SwiftLintFramework/Models/LinterCache.swift
@@ -22,14 +22,18 @@ public final class LinterCache {
     private let lock = NSLock()
     internal let fileManager: LintableFileManager
     private let location: URL?
+    private let swiftVersion: SwiftVersion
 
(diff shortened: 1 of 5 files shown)


Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -69,7 +77,8 @@ public final class LinterCache {
         var filesCache = writeCache[configurationDescription] ?? [:]
         filesCache[file] = [
             Key.violations.rawValue: violations.map(dictionary(for:)),
-            Key.lastModification.rawValue: lastModification.timeIntervalSinceReferenceDate
+            Key.lastModification.rawValue: lastModification.timeIntervalSinceReferenceDate,
+            Key.swiftVersion.rawValue: SwiftVersion.current.rawValue
```
jpsim: Shouldn't this be `swiftVersion.rawValue`?

marcelofabri: Yes, it should. Nice catch 😅

PR #1791, opened 2017-08-21
Fix false positive on `unused_enumerated` rule with complex variable bindings

### LINKED ISSUE #1787: unused_enumerated bug
`for (section, (event, _)) in data.enumerated()`

Data is an array of tuples, and this line of code reports that the item is not used, even though "event" is being used.

marcelofabri:
Fix #1787

This will likely introduce false negatives, so I'm curious to see oss-check running.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1744](https://github.com/realm/SwiftLint/issues/1744)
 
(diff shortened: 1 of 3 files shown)


marcelofabri:
Since this is default rule, I'm merging it even if that means we can get false negatives. It's more important to avoid false positives in this case IMO.

jpsim:
Given that OSSCheck didn't catch anything here, and that the `enumerated()` check is specific to the stdlib variant that returns two members, this is a good fix 👍

PR #1794, opened 2017-08-21
Fix setting Swift version when caching.

jpsim:
See https://github.com/realm/SwiftLint/pull/1790/files#r134323431.


jpsim:
Also relevant is cc9e618 from #1793.

marcelofabri:
> Also relevant is cc9e618 from #1793.

I'm not sure why that test would assume Swift 3 mode.

PR #1796, opened 2017-08-21
Add "Is Disjoint" rule

jpsim:
to encourage using `Set.isDisjoint(with:)` over `Set.intersection(_:).isEmpty`.


jpsim:
We really need to automate that.

PR #1797, opened 2017-08-24
Validate rules docs when running unit tests

### LINKED ISSUE #1734: Generate Rules.md automatically
It would be great if we could automate the generation of `Rules.md` even more, so we wouldn't even need to run its scheme. Maybe somehow running it together with unit tests?

Another option would be generating whenever something is merged in master.

marcelofabri:
Fix #1734

I'm not sure if this deserves a changelog entry.

### DIFF
--- a/Source/SwiftLintFramework/Models/RuleList+Documentation.swift
+++ b/Source/SwiftLintFramework/Models/RuleList+Documentation.swift
@@ -1,0 +1,97 @@
+//
+//  RuleList+Documentation.swift
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 8/24/17.
+//  Copyright © 2017 Realm. All rights reserved.
+//
+
+import Foundation
+
+extension RuleList {
+    public func generateDocumentation() -> String {
+        let rules = list.sorted { $0.0 < $1.0 }.map { $0.value }
+        let rulesText = rules.map(ruleMarkdown)
+        let rulesSummary = rules.map(ruleSummary)
+
+        var text = h1("Rules")
+        text += rulesSummary.joined()
+        text += "--------\n"
+        text += rulesText.joined(separator: "\n\n")
+
+        return text
+    }
+
+    private func ruleSummary(_ rule: Rule.Type) -> String {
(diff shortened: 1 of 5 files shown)


Tests/SwiftLintFrameworkTests/DocumentationTests.swift
```diff
@@ -0,0 +1,31 @@
...
+    func testRulesDocumentationIsUpdated() throws {
+        let docsPath = "\(projectRoot)/Rules.md"
+        let existingDocs = try String(contentsOfFile: docsPath)
+        let updatedDocs = masterRuleList.generateDocumentation()
+
+        XCTAssertEqual(existingDocs, updatedDocs)
+
+        let overwrite = false // set this to true to overwrite existing docs with the generated ones
+        if existingDocs != updatedDocs && overwrite {
+            try updatedDocs.data(using: .utf8)?.write(to: URL(fileURLWithPath: docsPath))
```
marcelofabri: I copied this approach from SourceKitten but I wonder if we shouldn't just always update the docs 🙃

Source/SwiftLintFramework/Models/RuleList+Documentation.swift
```diff
@@ -0,0 +1,97 @@
...
+//  SwiftLint
+//
+//  Created by Marcelo Fabri on 8/24/17.
+//  Copyright © 2017 Realm. All rights reserved.
+//
+
+import Foundation
+
+extension RuleList {
+    public func generateDocumentation() -> String {
```
marcelofabri: I'd prefer to keep it outside `SwiftLintFramework`, but I think it's a reasonable compromise.

Tests/SwiftLintFrameworkTests/DocumentationTests.swift
jpsim: yes. the reason it's parameterized in SourceKitten is that its output is non-deterministic.

jpsim:
Finally re-enabled this test in af14fb25f08b5c63c209f24c966b4e2d4f3c9e7c

PR #1800, opened 2017-08-25
Add pattern_matching_keywords rule

### LINKED ISSUE #202: Add rules from Erica Sadun
Lots of good stuff in here that we could turn into SwiftLint rules: http://ericasadun.com/2015/11/17/a-handful-of-swift-style-rules-swiftlang/

Notably, the following rules should be straightforward to implement:
- [x] The Rule of Conditional Binding Cascades (done in #327)
- [ ] The Rule of Pattern Matching Keywords
- [x] The Rule of `isEmpty` (wip in #206)
- [x] Prefer Swift constructors to legacy ones (done in #242)
- [x] Avoid terminal semicolons even though they compile (done in #203)

marcelofabri:
Closes #202.

This rule should also be applied in other cases of pattern matching, but I guess supporting only `switch`es is reasonable for now.

Let's see how much oss-check complains to decide if this should be opt-in or enabled by default.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -25,6 +25,7 @@ opt_in_rules:
   - let_var_whitespace
   - unneeded_parentheses_in_closure_argument
   - extension_access_modifier
+  - pattern_matching_keywords
 
 file_header:
   required_pattern: |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,11 @@
 
 * Invalidate cache when Swift version changes.  
   [Marcelo Fabri](https://github.com/marcelofabri)
+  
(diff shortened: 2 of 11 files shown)


Uncommon:
Looks like this one is a false positive:

https://github.com/mozilla-mobile/firefox-ios/blob/2e17eee247a278f90ff3762bd6e8d7ca908358ec/Account/FxAPushMessageHandler.swift#L262:61

marcelofabri:
@Uncommon I don't see why. You should use this instead:

```swift
case let (.deviceConnected(lName), .deviceConnected(rName)):
```

Uncommon:
@marcelofabri Ah, I see. Never mind.

PR #1802, opened 2017-08-27
Add `multiple_closures_with_trailing_closure` rule

### LINKED ISSUE #1801: Rule Request: Multiple Closure Arguments with Trailing Closure
### New Issue Checklist

### Rule Request

In many cases, the trailing closure syntax can make code clearer by reducing boilerplate. When passing a single closure argument to a function, that closure’s role is often clear even without its parameter label. When passing two closure arguments, however, the parameter labels crucially distinguish the role of the two closures. With trailing closure syntax, it becomes less clear that the second argument to `UIView.animate` is a completion handler:

```
UIView.animate(withDuration: 1.0, animations: {
    someView.alpha = 0.0
}) { _ in
    someView.removeFromSuperview()
}
```

I propose a rule that would warn when the trailing closure syntax is used when passing two closure arguments to a function. The principle behind this rule is supported by [the Ray Wenderlich Swift style guide](https://github.com/raywenderlich/swift-style-guide#closure-expressions) and Erica Sadun’s blog post [“Swift: The Problem with Trailing Closure Syntax”](https://www.natashatherobot.com/swift-trailing-closure-syntax/). SwiftLint’s existing [Trailing Closure Rule](https://github.com/realm/SwiftLint/blob/master/Rules.md#trailing-closure) already makes an exception when passing multiple closure arguments to a function (`foo.something(param1: { $0 }, param2: { $0 + 1 })` is a non-triggering example for that rule).

The rule’s severity should be configurable to warning or error, defaulting to warning; there are no other obvious configurable parameters. The rule’s kind should be style. The rule should be enabled by default, because this style guideline has community support and generally improves code readability.

Additional examples of how this rule should behave:

**Non-triggering examples:**

```
foo.map { $0 + 1 }
```

```
foo.reduce(0) { $0 + $1 }
```

```
if let foo = bar.map({ $0 + 1 }) {

}
```

```
foo.something(param1: { $0 }, param2: { $0 + 1 })
```

```
UIView.animate(withDuration: 1.0) {
    someView.alpha = 0.0
}
```

**Triggering examples:**

```
foo.something(param1: { $0 }) { $0 + 1 }
```

```
UIView.animate(withDuration: 1.0, animations: {
    someView.alpha = 0.0
}) { _ in
    someView.removeFromSuperview()
}
```

erikstrottmann:
As described in #1801, the Multiple Closures with Trailing Closure rule disallows trailing closure syntax when passing more than one closure argument to a function. This is my first PR for SwiftLint, so I’m certainly open to feedback.

Fixes #1801.

### DIFF
--- a/Source/SwiftLintFramework/Rules/MultipleClosuresWithTrailingClosureRule.swift
+++ b/Source/SwiftLintFramework/Rules/MultipleClosuresWithTrailingClosureRule.swift
@@ -1,0 +1,98 @@
+//
+//  MultipleClosuresWithTrailingClosureRule.swift
+//  SwiftLint
+//
+//  Created by Erik Strottmann on 8/26/17.
+//  Copyright © 2017 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+public struct MultipleClosuresWithTrailingClosureRule: ASTRule, ConfigurationProviderRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "multiple_closures_with_trailing_closure",
+        name: "Multiple Closures with Trailing Closure",
+        description: "Trailing closure syntax should not be used when passing more than one closure argument.",
+        kind: .style,
+        nonTriggeringExamples: [
+            "foo.map { $0 + 1 }\n",
+            "foo.reduce(0) { $0 + $1 }\n",
+            "if let foo = bar.map({ $0 + 1 }) {\n\n}\n",
+            "foo.something(param1: { $0 }, param2: { $0 + 1 })\n",
+            "UIView.animate(withDuration: 1.0) {\n" +
+            "    someView.alpha = 0.0\n" +
+            "}"
+        ],
+        triggeringExamples: [
+            "foo.something(param1: { $0 }) ↓{ $0 + 1 }",
+            "UIView.animate(withDuration: 1.0, animations: {\n" +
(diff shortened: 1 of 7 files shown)


Source/SwiftLintFramework/Rules/MultipleClosuresWithTrailingClosureRule.swift
```diff
@@ -0,0 +1,95 @@
...
+            call.hasTrailingClosure,
+            call.closureArguments.count > 1
+            else {
+                return []
+        }
+
+        return [
+            StyleViolation(ruleDescription: type(of: self).description,
+                           severity: configuration.severity,
+                           location: Location(file: file, byteOffset: call.offset))
```
marcelofabri: Do you think it would be better if we added the violation to the trailing closure? For example:

```swift
foo.something(param1: { $0 }) ↓{ $0 + 1 }
```

erikstrottmann: Makes sense to me. I had initially put the violation at the start of the function call because that's where it is for `trailing_closure`, but I think the start of the trailing closure is more precise.

erikstrottmann:
Rebased and docs regenerated.

Note that I also added `// swiftlint:disable file_length` to `Tests/SwiftLintFrameworkTests/RulesTests.swift`, because this commit together with the recent changes to master cause that file to exceed the 400 line limit.

marcelofabri:
Thanks @erikstrottmann 🚀

PR #1805, opened 2017-08-27
Prints the performance report before other messages

ornithocoder:
This is a minor suggestion, but I believe it would be easier to review the performance report if it was grouped before all the other messages. Specially for changes with dozens or hundreds of messages. I didn't add this to the change log because this is a minor change to the CI script.


marcelofabri:
Thanks!

PR #1806, opened 2017-08-28
Warn if a configured rule is not enabled

### LINKED ISSUE #1350: Warn if a configured rule is not enabled

marcelofabri:
Closes #1350

Pending tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,10 @@
   [Erik Strottmann](https://github.com/erikstrottmann)
   [#1801](https://github.com/realm/SwiftLint/issues/1801)
 
+* Warn if a configured rule is not enabled.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#1350](https://github.com/realm/SwiftLint/issues/1350)
+
 ##### Bug Fixes
 
 * Fix false positive on `force_unwrapping` rule when declaring
--- a/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
@@ -64,7 +64,7 @@ extension Configuration {
         Configuration.warnAboutDeprecations(configurationDictionary: dict, disabledRules: disabledRules,
                                             optInRules: optInRules, whitelistRules: whitelistRules, ruleList: ruleList)
 
-        let configuredRules: [Rule]
+        let configuredRules: [ConfiguredRule]
         do {
             configuredRules = try ruleList.configuredRules(with: dict)
         } catch RuleListError.duplicatedConfigurations(let ruleType) {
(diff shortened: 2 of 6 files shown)


jpsim:
I'm not thrilled about the changes to the public API and the addition of the `ConfiguredRule` type. Is there any way this could be done by extending `Rule` instead?

This also requires manually keeping two states in sync: the `isDefaultConfiguration` flag and whether or not `init(configuration:)` or `configuration.apply(_:)` was invoked on the Rule. It'd be nice if the flag could be automatically derived.

At the very least, here are two suggested patches to the current state of this PR:

```diff
    Return early in validateConfiguredRulesAreEnabled

diff --git a/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift b/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
index cf653582..2a48c439 100644
--- a/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
@@ -167,6 +167,10 @@ extension Configuration {
 
     internal static func validateConfiguredRulesAreEnabled(rulesMode: Configuration.RulesMode,
                                                            configuredRules: [ConfiguredRule]) {
+        if case .allEnabled = rulesMode {
+            return
+        }
+
         let rulesWithConfiguration = configuredRules.flatMap {
             $0.isDefaultConfiguration ? nil : $0.rule
         }
@@ -177,7 +181,7 @@ extension Configuration {
 
             switch rulesMode {
             case .allEnabled:
-                return
+                fatalError("Unreachable")
             case .whitelisted(let whitelist):
                 if Set(whitelist).isDisjoint(with: identifiers) {
                     queuedPrintError("\(message), but it is not present on " +
```

and

```diff
    Reuse RuleDescription
    
    to avoid a dynamic runtime type check

diff --git a/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift b/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
index 2a48c439..288a2263 100644
--- a/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
@@ -176,8 +176,9 @@ extension Configuration {
         }
 
         for rule in rulesWithConfiguration {
-            let identifiers = type(of: rule).description.allIdentifiers
-            let message = "Found a configuration for '\(type(of: rule).description.identifier)' rule"
+            let ruleDescription = type(of: rule).description
+            let identifiers = ruleDescription.allIdentifiers
+            let message = "Found a configuration for '\(ruleDescription.identifier)' rule"
 
             switch rulesMode {
             case .allEnabled:
```

marcelofabri:
The only way I can think of is adding a flag to the `Rule` (or `ConfigurationProviderRule`) protocol, but that'd add a requirement that all rules will need to declare that property.

jpsim:
We'd have to reshuffle quite a few things in order to get it to be automated I think.

Looking at OSSCheck, it looks like this might have a slight impact on performance too. Would you mind confirming if that's the case?

PR #1833, opened 2017-09-10
Switch case alignment rule

austinlu:
Opt-in rule checking that case and default statements should vertically align with the enclosing switch statement itself.


marcelofabri:
Hi @austinlu, thanks for your contribution!

I have a suggestion to make this more reliable though, as you can see that there're false positives in oss-check (e.g. https://github.com/mozilla-mobile/firefox-ios/blob/ffdce75bae212142d4958cd255c5c9ba8f9fd3d8/Client/Frontend/Home/ActivityStreamPanel.swift#L852:5).

Instead of relying in just the syntax map and regexes, we should rely on the structure from SourceKit.

If you install [SouceKitten](https://github.com/jpsim/SourceKitten), you can make some tests. For example, for this file:

```swift
switch foo {
case .bar:
    break
case .bar2:
    break
case .bar3, bar4:
    break
default:
    break
}
```

And running `sourcekitten structure --file file.swift`, you'd get this output:

```json
{
  "key.substructure" : [
    {
      "key.nameoffset" : 0,
      "key.elements" : [
        {
          "key.kind" : "source.lang.swift.structure.elem.expr",
          "key.offset" : 7,
          "key.length" : 3
        }
      ],
      "key.length" : 104,
      "key.substructure" : [
        {
          "key.namelength" : 0,
          "key.nameoffset" : 0,
          "key.elements" : [
            {
              "key.kind" : "source.lang.swift.structure.elem.pattern",
              "key.offset" : 18,
              "key.length" : 4
            }
          ],
          "key.kind" : "source.lang.swift.stmt.case",
          "key.offset" : 13,
          "key.length" : 20
        },
        {
          "key.namelength" : 0,
          "key.nameoffset" : 0,
          "key.elements" : [
            {
              "key.kind" : "source.lang.swift.structure.elem.pattern",
              "key.offset" : 39,
              "key.length" : 5
            }
          ],
          "key.kind" : "source.lang.swift.stmt.case",
          "key.offset" : 34,
          "key.length" : 21
        },
        {
          "key.namelength" : 0,
          "key.nameoffset" : 0,
          "key.elements" : [
            {
              "key.kind" : "source.lang.swift.structure.elem.pattern",
              "key.offset" : 61,
              "key.length" : 5
            },
            {
              "key.kind" : "source.lang.swift.structure.elem.pattern",
              "key.offset" : 68,
              "key.length" : 4
            }
          ],
          "key.kind" : "source.lang.swift.stmt.case",
          "key.offset" : 56,
          "key.length" : 27
        },
        {
          "key.namelength" : 0,
          "key.nameoffset" : 0,
          "key.elements" : [
            {
              "key.kind" : "source.lang.swift.structure.elem.pattern",
              "key.offset" : 84,
              "key.length" : 7
            }
          ],
          "key.kind" : "source.lang.swift.stmt.case",
          "key.offset" : 84,
          "key.length" : 18
        }
      ],
      "key.kind" : "source.lang.swift.stmt.switch",
      "key.offset" : 0,
      "key.namelength" : 0
    }
  ],
  "key.offset" : 0,
  "key.length" : 105
}
```

Then, you can take advantage of the existing infrastructure available in SwiftLint: make the rule conform to `ASTRule` and implement this method:

```swift
public func validate(file: File, kind: StatementKind,
    guard kind == .switch else {
        return []
   }

  // ...
}
``` 

That method will find all `switch`es in the file. Now, you'd need to navigate its own substructure to get the the elements that are `cases` or `default` (`source.lang.swift.stmt.case`). 

After that, you can use `lineAndCharacter(forByteOffset:)` declared in `String+SourceKitten` to get the line and character in that line for a byte offset (which is what we get from SourceKit elements).

Please let me know if have any doubts or are stuck on something. `MultilineParametersRule` or `VerticalParameterAlignmentRule` are probably good rules to look as a reference too.

austinlu:
Thanks for the feedback @marcelofabri, I will look into the SourceKit structure to avoid the false positives.

PR #1835, opened 2017-09-11
Add fallthrough opt-in rule

### LINKED ISSUE #1834: Rule request: fallthrough should be avoided
### New Issue Checklist

### Rule Request

Avoid `fallthrough` inside `switch` statements.

Using `fallthrough` makes code harder to read and, in most cases, can be replaced by combining `cases`. [LinkedIn's Swift Style Guide](https://github.com/linkedin/swift-style-guide#32-access-modifiers) also thinks that (item 3.4.5).

```swift
// should trigger
switch foo {
case .bar:
    fallthrough
case .bar2:
    fallthrough
case .bar3
    something()
}


// shouldn't trigger
switch foo {
case .bar, .bar2, .bar3:
    something()
}
```

Only severity.

Opt-in. In the end, this is still a matter of style and there are situations where you can't just combine cases.

marcelofabri:
Fixes #1834

As `oss-check` reported few violations and almost all of them can be solved by joining cases, I'm enabling it by default.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -26,6 +26,7 @@ opt_in_rules:
   - unneeded_parentheses_in_closure_argument
   - extension_access_modifier
   - pattern_matching_keywords
+  - fallthrough
 
 file_header:
   required_pattern: |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ##### Enhancements
 
-* Add `fallthrough` opt-in rule that flags usage of `fallthrough`.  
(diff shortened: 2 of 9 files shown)


.swiftlint.yml
```diff
@@ -26,6 +26,7 @@ opt_in_rules:
   - unneeded_parentheses_in_closure_argument
   - extension_access_modifier
   - pattern_matching_keywords
+  - fallthrough
```
jpsim: This is unnecessary now that the rule is enabled by default.

marcelofabri: Ops, missed that. Will push a commit to `master` removing it.

PR #1840, opened 2017-09-14
Fix false positive in for_where with if var

### LINKED ISSUE #1838: for_where, false positive for optional binding
This for in loop with optional binding violates the rule "for_where":
for node in json {
    if var data = node.value as? [String: String] {
        //do stuff with data
    }
}

But the rules preferred code doesn't compile:
for node in json where var data = node.value as? [String: String] {
    //do stuff with data
}
Maybe I don't know how to write the correct code.
Or the rule is throwing a false positive.

### Environment
SwiftLint 0.22.0 over CocoaPods
and Swift3 in Xcode 8.3.3

marcelofabri:
Fixes #1838

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,10 @@
   [#1770](https://github.com/realm/SwiftLint/issues/1770)
   [#1812](https://github.com/realm/SwiftLint/issues/1812)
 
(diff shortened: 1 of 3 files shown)


jpsim:
👍

PR #1841, opened 2017-09-16
Add array_init opt-in rule

### LINKED ISSUE #1271: Rule Request: Prefer `Array(foo)` over `foo.map({ $0 })`
Clearer and lots of optimizations

marcelofabri:
Fixes #1271

This is opt-in because there're types that have `.map` and can't be converted to `Array`. However, i expect this to not give lots of false positives.

Also I'm not super happy about the rule name, so feel free to make suggestions!

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -26,6 +26,7 @@ opt_in_rules:
   - unneeded_parentheses_in_closure_argument
   - extension_access_modifier
   - pattern_matching_keywords
+  - array_init
 
 file_header:
   required_pattern: |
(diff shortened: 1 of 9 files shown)


marcelofabri:
ok, we've got some false positives that we can fix before merging:
* https://github.com/mozilla-mobile/firefox-ios/blob/ffcdebeb1c1421ae8e44826ed8304afb27543415/Client/Frontend/Browser/FaviconManager.swift#L54:20
* https://github.com/kickstarter/ios-oss/blob/29ba281398c958565cad429af8a286d86e872b17/Library/DataSource/ValueCellDataSource.swift#L135:28
* https://github.com/apple/swift/blob/94c596e6edfc830aee6ba2d08136ea8aa49ddbce/stdlib/public/core/FlatMap.swift#L47:12

marcelofabri:
🎉

PR #1844, opened 2017-09-17
Move functions to properties and Sets instead of Arrays

marcelofabri:
Running locally I didn't observe a speed gain, but at least we're more consistent.

```
Message: Linting Aerial with this PR took 0.29s vs 0.28s on master (3% slower)
Message: Linting Alamofire with this PR took 2.05s vs 2.08s on master (1% faster)
Message: Linting Firefox with this PR took 8.14s vs 8.1s on master (0% slower)
Message: Linting Kickstarter with this PR took 12.84s vs 12.81s on master (0% slower)
Message: Linting Moya with this PR took 0.95s vs 0.95s on master (0% slower)
Message: Linting Nimble with this PR took 1.13s vs 1.13s on master (0% slower)
Message: Linting Quick with this PR took 0.35s vs 0.35s on master (0% slower)
Message: Linting Realm with this PR took 1.85s vs 1.85s on master (0% slower)
Message: Linting SourceKitten with this PR took 0.68s vs 0.68s on master (0% slower)
Message: Linting Sourcery with this PR took 2.96s vs 2.95s on master (0% slower)
Message: Linting Swift with this PR took 8.97s vs 9.04s on master (0% faster)
Message: Linting WordPress with this PR took 7.42s vs 7.51s on master (1% faster)
```

### DIFF
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -197,7 +197,7 @@ extension File {
      file contents.
      */
     internal func match(pattern: String,
-                        excludingSyntaxKinds syntaxKinds: [SyntaxKind],
+                        excludingSyntaxKinds syntaxKinds: Set<SyntaxKind>,
                         range: NSRange? = nil) -> [NSRange] {
         return match(pattern: pattern, range: range)
             .filter { $0.1.filter(syntaxKinds.contains).isEmpty }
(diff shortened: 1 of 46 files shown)


jpsim:
I'm busy these days, but briefly: I don't think we should jump into this.

I think the `Array` vs `Set` discussion is sort of parallel to the [`Int` vs `UInt` debate](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/TheBasics.html#//apple_ref/doc/uid/TP40014097-CH5-NoLink_32) in API design. Just because Arrays are so much more prevalent than Sets in Swift, you probably should default to it even if you don't need stable ordering of members and you mostly just use it for membership checks, unless Sets provides performance wins, or you need to deduplicate arrays dynamically (for example).

jpsim:
Finally took a good look at this, and I now agree with the changes. Merged in 0be8c855 and 60ffa732. Thanks for pushing back on my pushback 😉

marcelofabri:
🎉

Issue #1822, opened 2017-09-07
sorted_imports should support auto-correction


Johennes:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Enhancement Request

The `sorted_imports` rule should support autocorrection. It seems like a relatively easy task to just cut all `import ...` lines, sort them and then re-insert them wherever the first `import` statement was.

### Environment

* SwiftLint version: 0.22.0
* Installation method used: Homebrew,
* Paste your configuration file:

```yml
whitelist_rules:
  - sorted_imports
```

marcelofabri:
Implemented in #1922.

PR #1845, opened 2017-09-17
Sorted imports auto correction

sammy-SC:
Makes `sorted_imports` corractable.
Feature request: #1822 

I was not sure how to handle this case.

```
import BBB
// comment 
import CCC
import AAA
```
the code above will generate warning from `sorted_imports` but autocorrection won't fix it because I am not sure what to do with the comment. The code above will result into following correction. Same thing happens if there is a code in place of the comment.
```
import BBB
// comment 
import AAA
import CCC
```

Any comments or suggestions are very welcome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
   [Samuel Susla](https://github.com/sammy-sc)
   [#1514](https://github.com/realm/SwiftLint/issues/1514)
 
+* Make `sorted_imports` correctable.  
+  [Samuel Susla](https://github.com/sammy-sc)
+  [#1822](https://github.com/realm/SwiftLint/issues/1822)
+
 * Add `fallthrough` rule that flags usage of `fallthrough`.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1834](https://github.com/realm/SwiftLint/issues/1834)
--- a/Rules.md
+++ b/Rules.md
@@ -9811,6 +9811,13 @@ import labc
 import Ldef
 ```
 
+```swift
+import BBB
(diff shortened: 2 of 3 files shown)


jpsim:
I know @sberrevoets cares about this, so maybe he'd like to help review?

sberrevoets:
@sammy-SC I think your approach is the right one. There will likely be edge cases, but sorting imports inside an import group (where a group spans from the first line that starts with `import` until a line does not start with `import`) seems to most straightforward.

That would cover your case, but also something like this:

```swift
import ABC
#if os(iOS)
import UIKit
#elseif os(macOS)
import AppKit
#endif
```

In that case everything for `os(iOS)` would be sorted separately from everything for `os(macOS)`. Anything more complex than that is probably not worth autocorrecting because this can very complicated.

sammy-SC:
@sberrevoets I agree that warning should work the same way.

I think it should be done in different pull request thought.

I will be happy to change the current implementation, please let me know whether I should do it in this PR or a different one. cc @jpsim

sberrevoets:
If you create an abstraction here that the warning implementation would benefit from, then I think it's OK to do here, but I'll let @jpsim make the final call on that.

sammy-SC:
@sberrevoets @jpsim I took the liberty and changed it here. 
If you want me to open another PR to make that change I can cherry-pick the code without a problem.

rastersize:
@sammy-SC @jpsim Any chance we can get this merged? I just manually correct this in a medium sized project, took a while 😸

marcelofabri:
Closing in favor of #1922

Issue #1823, opened 2017-09-07
Exclusion paths are not evaluated when using script input files


Johennes:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Bug Report

Consider the following example

```bash
$ ls -hal
total 24
drwxr-xr-x   5 jmarbach  SKYDIVE\Domain Users   170B Sep  7 09:49 .
drwxr-xr-x  40 jmarbach  SKYDIVE\Domain Users   1.3K Sep  7 09:43 ..
-rw-r--r--@  1 jmarbach  SKYDIVE\Domain Users    51B Sep  7 09:46 .swiftlint.yml
-rw-r--r--@  1 jmarbach  SKYDIVE\Domain Users    15B Sep  7 09:44 File.swift
-rwxr-xr-x@  1 jmarbach  SKYDIVE\Domain Users   118B Sep  7 09:47 lint.sh
```

```bash
$ cat .swiftlint.yml 
whitelist_rules:
  - colon
excluded:
  - File.swift
```

```bash
$ cat File.swift 
let a : Int = 0
```

```bash
$ cat lint.sh 
#!/bin/bash
export SCRIPT_INPUT_FILE_COUNT=1
export SCRIPT_INPUT_FILE_0=File.swift
swiftlint --use-script-input-files
```

When running swiftlint in the folder, `File.swift` is excluded.

```bash
$ swiftlint lint
Loading configuration from '.swiftlint.yml'
Linting Swift files in current working directory
No lintable files found at path ''
```

However, when running `lint.sh` which triggers swiftlint with `File.swift` as a script input file, the file is *not* excluded.

```bash
./lint.sh 
Loading configuration from '.swiftlint.yml'
Linting 'File.swift' (1/1)
/Users/jmarbach/ios/test/File.swift:1:5: warning: Colon Violation: Colons should be next to the identifier when specifying a type and next to the key in dictionary literals. (colon)
Done linting! Found 1 violation, 0 serious in 1 file.
```

This is problematic when swiftlint is run on staged files as part of a git commit hook.

### Environment

* SwiftLint version: 0.22.0
* Installation method used: Homebrew

jpsim:
Could you please explain how this request is distinct from #591? If what you're requesting is covered in #591, let's close this one as a duplicate.

Johennes:
[...]

PR #1847, opened 2017-09-19
Exclude files also when using the '-—use-script-input-files' option

### LINKED ISSUE #591: excluded should override SCRIPT_INPUT_FILE_*
The `excluded` config variable should override `SCRIPT_INPUT_FILE_*` environment. The frustrating framework situation for swift command line apps has me including the source of a module directly into my project and I don't want to lint it. ExtraBuildPhase puts it in the environment and swiftlint runs on it anyways ignoring my `excluded` rule.

b-ray:
Fixes #591 

I use this to lint files that are staged for a commit as part of a pre-commit-hook (like also mentioned [here](https://github.com/realm/SwiftLint/issues/1823)).
As I don't want to abort the commit if all changed files are filtered, I perform the filter after the check if the `files` array is empty.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1271](https://github.com/realm/SwiftLint/issues/1271)
 
+* Exclude files defined in the configuration also when using the
+  `--use-script-input-files` option.  
+  [Stefan Pühringer](https://github.com/b-ray)
+  [#591](https://github.com/realm/SwiftLint/issues/591)
+
 ##### Bug Fixes
 
 * Improve how `opening_brace` rule reports violations locations.  
--- a/Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
@@ -14,6 +14,18 @@ extension Configuration {
         return lintablePaths(inPath: path).flatMap(File.init(path:))
     }
 
+    public func lintableFiles(ofFiles files: [File]) -> [File] {
+        let excludedPaths = excluded.flatMap {
+            FileManager.default.filesToLint(inPath: $0, rootDirectory: rootPath)
+        }
+        return files.filter {
+            guard let path = $0.path else {
+                return true
+            }
+            return !excludedPaths.contains(path)
+        }
+    }
+
     internal func lintablePaths(inPath path: String,
                                 fileManager: LintableFileManager = FileManager.default) -> [String] {
         // If path is a file, skip filtering with excluded/included paths
--- a/Source/swiftlint/Extensions/Configuration+CommandLine.swift
+++ b/Source/swiftlint/Extensions/Configuration+CommandLine.swift
@@ -67,6 +67,9 @@ extension Configuration {
                 let errorMessage = "No lintable files found at path '\(path)'"
                 return .failure(.usageError(description: errorMessage))
             }
+            if useScriptInputFiles {
+                return .success(lintableFiles(ofFiles: files))
+            }
             return .success(files)
         }.flatMap { files in
             let queue = DispatchQueue(label: "io.realm.swiftlint.indexIncrementer")


jpsim:
Could you please rebase on master to include eb9acae?

b-ray:
Sure, I rebased and pushed it @jpsim

CHANGELOG.md
```diff
@@ -29,6 +29,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1271](https://github.com/realm/SwiftLint/issues/1271)
 
+* Exclude files defined in the configuration also when using the
```
jpsim: > Exclude files defined in the `excluded` value of SwiftLint's configuration when `--use-script-input-files` is specified.

Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
```diff
@@ -14,6 +14,18 @@ extension Configuration {
         return lintablePaths(inPath: path).flatMap(File.init(path:))
     }
 
+    public func lintableFiles(ofFiles files: [File]) -> [File] {
+        let excludedPaths = excluded.flatMap {
+            FileManager.default.filesToLint(inPath: $0, rootDirectory: rootPath)
```
jpsim: This doesn't inject the `LintableFileManager` like we do everywhere else we interact with `FileManager`.

Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
```diff
@@ -14,6 +14,18 @@ extension Configuration {
         return lintablePaths(inPath: path).flatMap(File.init(path:))
     }
 
+    public func lintableFiles(ofFiles files: [File]) -> [File] {
```
jpsim: `Configuration.lintableFiles(ofFiles:)` is not a very descriptive method name...

Source/swiftlint/Extensions/Configuration+CommandLine.swift
```diff
@@ -67,6 +67,9 @@ extension Configuration {
                 let errorMessage = "No lintable files found at path '\(path)'"
                 return .failure(.usageError(description: errorMessage))
             }
+            if useScriptInputFiles {
```
jpsim: This is the wrong place to be doing this...

jpsim:
Here's an alternative implementation which I prefer:

```diff
diff --git a/Source/swiftlint/Extensions/Configuration+CommandLine.swift b/Source/swiftlint/Extensions/Configuration+CommandLine.swift
index ff0cb1c4..6f10dce4 100644
--- a/Source/swiftlint/Extensions/Configuration+CommandLine.swift
+++ b/Source/swiftlint/Extensions/Configuration+CommandLine.swift
@@ -13,7 +13,7 @@ import Result
 import SourceKittenFramework
 import SwiftLintFramework
 
-private func scriptInputFiles() -> Result<[File], CommandantError<()>> {
+private func scriptInputFiles(excluding excludedPaths: [String]) -> Result<[File], CommandantError<()>> {
     func getEnvironmentVariable(_ variable: String) -> Result<String, CommandantError<()>> {
         let environment = ProcessInfo.processInfo.environment
         if let value = environment[variable] {
@@ -33,10 +33,15 @@ private func scriptInputFiles() -> Result<[File], CommandantError<()>> {
         return .failure(.usageError(description: "\(inputFileKey) did not specify a number"))
     }()
 
+    let excludedAbsolutePaths = excludedPaths.map { $0.absolutePathStandardized() }
+
     return count.flatMap { count in
         let inputFiles = (0..<count).flatMap { fileNumber -> File? in
             switch getEnvironmentVariable("SCRIPT_INPUT_FILE_\(fileNumber)") {
             case let .success(path):
+                if excludedAbsolutePaths.contains(path.absolutePathStandardized()) {
+                    return nil
+                }
                 if let file = File(path: path), path.bridge().isSwiftFile() {
                     return file
                 }
@@ -109,7 +114,7 @@ extension Configuration {
             }
             return .failure(.usageError(description: "stdin isn't a UTF8-encoded string"))
         } else if useScriptInputFiles {
-            return scriptInputFiles()
+            return scriptInputFiles(excluding: excluded)
         }
         if !quiet {
             let message = "\(action) Swift files " + (path.isEmpty ? "in current working directory" : "at path \(path)")
```

CHANGELOG.md
jpsim: Also, this is a breaking change, so it should be in the "Breaking" section of the changelog.

b-ray:
Thanks for your feedback!
I agree that we can improve on this, however, the changes you've requested are not working on my machine.

When I have a directory in my `excluded` value, and I specify a few swift files in that directory as the `script-input-files`, swiftlint will still lint them when I apply your diff. It seems it is comparing the full file paths with the directory path in `Configuration+CommandLine.swift:42`, which doesn't work.

Also, if we filter the files at this point, SwiftLint will return with failure if all specified files are actually filtered out, which makes it inconvenient to use for my use-case as a pre-commit-hook.

sirnacnud:
Any chance this will get merged in?

jpsim:
I had a number of comments which were never addressed. @b-ray are you interested in reviving this? Otherwise let's close to indicate to other contributors that this is up for grabs.

jpsim:
@keefmoon no one is working on this as far as I can tell. Given that you're keen on seeing this done, would you be interested in completing the work that was started in this PR yourself?

keefmoon:
@jpsim Yes. I'll take a look. Thanks

Issue #1502, opened 2017-05-08
Rule request: No explicit `== true` or `== false`


mdiep:
I occasionally come across code like this:

```swift
if condition == true { … }
```

which just be written as:

```swift
if condition { … }
```

I'd love to have a rule that disallowed `== true` and `== false`. (And `!= true`, `!= false`, which I haven't seen, but would be far worse.)

jpsim:
Agreed. This is a good starter rule to write if someone's interested in contributing to SwiftLint but finds other requested rules daunting.

Might want to also validate this rule with the boolean literal on the left hand side:

```swift
if true == condition {}
if false != condition {}
```

SDGGiesbrecht:
> This is a good starter rule.

I’m not so sure. It’s more complicated than it looks at first glance.

For example, what about `Bool?`?

```swift
func someFunction() -> Bool? {
    return true
}

if someFunction() == true {
    // Should the rule trigger?
}

if someFunction() {
    // This will not compile.
}

if someFunction ?? false {
    // This would satisfy both the rule and the compiler,
    // but is this really what we want?
}
```

masters3d:
I'd like to add to also throw a violation when comparing to the same object.

```swift
if someObj == someObj {} // always true
if objA != objA {} // always false
```

jpsim:
Ah, cursed optionals strike again! @masters3d could you please file a new ticket for that, as its implementation could be entirely different than the original request.

mohamede1945:
For the case where comparing with `Optional<Bool>` . I think it's clearer to use `?? true|false` operator than just compare with `==  true|false`.
If I look again at @SDGGiesbrecht 's example at the call site.
```swift
if someFunction() == true {
}
```
It's not clear for the reader that someFunction() might return `nil`.

But
```swift
if someFunction() ?? false {
}
```
It's very clear that the function might return `nil` and also clear what to do in that case.

marcelofabri:
You could also use an `extension` (since Swift 3.1):

```swift
extension Optional where Wrapped == Bool {
    var isTrue: Bool {
        switch self {
        case .none:
            return false
        case .some(let value):
            return value
        }
    }
}
```

mohamede1945:
But that extension assumes `nil` is equivalent to `false` which isn't.

marcelofabri:
[...]

PR #1854, opened 2017-09-27
Add no explicit boolean comparison rule

sammy-SC:
implements #1502

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@
 
 ##### Enhancements
 
+* Add `no_explicit_boolean_comparison` rule to prevent people from writing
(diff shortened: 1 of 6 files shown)


marcelofabri:
I think this needs a second look, @sammy-SC. I can't believe that no violations were triggered by this.

There're even [places in SwiftLint itself](https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules/JoinedDefaultRule.swift#L95) that should trigger a warning.

jpsim:
This needs to work with optionals. E.g.

```swift
let a: Bool? = true
if a { // doesn't compile
}
if a == true { // compiles
}
```

PR #1857, opened 2017-09-28
Fix false positive in `class_delegate_protocol` with Swift 4.0.1

### LINKED ISSUE #1856: `class_delegate_protocol` incorrectly triggering with Xcode 9.1 beta (9B37)
### New Issue Checklist

```bash
$ echo 'protocol FooDelegate: class {}' | swiftlint --use-stdin --config swiftlint.yml 
Loading configuration from 'swiftlint.yml'
<nopath>:1:1: warning: Class Delegate Protocol Violation: Delegate protocols should be class-only so they can be weakly referenced. (class_delegate_protocol)
```

### Environment

> 0.22.0
> Homebrew

```yml
whitelist_rules:
  - class_delegate_protocol
```

> No
```
$ xcode-select -p
/Applications/Xcode-beta.app/Contents/Developer
$ xcodebuild -version
Xcode 9.1
Build version 9B37
```

```swift
// This triggers a violation:
protocol FooDelegate: class {}
```

### More info

Upgrading to 9.1 also started showing (actual) violations of the `colon` rule for `protocol` declarations (I think that's be a good thing). Maybe something changed re: the parse tree for protocols?

marcelofabri:
Fixes #1856

// @danj-stripe, can you please check if this fixes your issue?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1838](https://github.com/realm/SwiftLint/issues/1838)
 
+* Fix false positive in `class_delegate_protocol` rule when using Swift 4.0.1.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#1856](https://github.com/realm/SwiftLint/issues/1856)
+
 ## 0.22.0: Wrinkle-free
 
 ##### Breaking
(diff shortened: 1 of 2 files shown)


marcelofabri:
Thanks for the feedback, @danj-stripe! 

I wasn't sure that it'd work for Linux too, that's why I didn't change before. But I've tested on Linux with Swift 4 and it looks like we can unify the implementation again 👍

jpsim:
Nice improvement over my fix!

PR #1859, opened 2017-09-28
Build on macOS and linux

marcomorain:


jpsim:
Let me know if I can help in any way.

jpsim:
Actually, it looks like these worked (other than `pod lib lint` needing an explicit `--swift-version=3.0` and `--swift-version=4.0`). Is this something I can enable today, or should I wait for general CircleCI 2.0 macOS support?

PR #1861, opened 2017-09-29
Add Mint installation instructions

yonaskolb:
This adds [Mint](https://github.com/yonaskolb/Mint) installation instructions to the readme 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,11 @@ Note that this will add the SwiftLint binaries, its dependencies' binaries and t
 library distribution to the `Pods/` directory, so checking in this directory to SCM such as
 git is discouraged.
 
+### Using [Mint](https://github.com/yonaskolb/mint):
+```
+$ mint run realm/SwiftLint
+```
+
 ### Using a pre-built package:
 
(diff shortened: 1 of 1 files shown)


jpsim:
**tl;dr; please move this to after the CocoaPods section**

Thanks for the PR! I think Mint is awesome, but there are a few reasons why I still want to lead people to Homebrew first, then CocoaPods, then Mint.

First, homebrew is still the canonical way to install CLI tools on macOS. It has the added advantage of not requiring to build from source, thus speeding up the process, since Homebrew just downloads the bottled version of the formula. It's also the installation method that we've supported the longest, so it's well tested.

Second, CocoaPods since many people are _already_ using it. It also avoids compiling the binary upon installation, simply downloading it from GitHub releases. It also helps version-lock SwiftLint to projects by specifying the version[s] in the Podfile.

yonaskolb:
Yeah of course, makes sense! Sorry, that was presumptuous of me to add it at the top, I didn't really think about it.
I've moved it to after cocoapods 👍

jpsim:
Thanks!

PR #1863, opened 2017-09-29
Move all Travis jobs to CircleCI

jpsim:
and migrate CircleCI to version 2. This expands on #1859.


jpsim:
I had to run some of these jobs 7 or 8 times to get to a green checkmark due to CircleCI infrastructure reasons. It looks like they still have stability work to do before this is ready for prime time. /cc @marcomorain

However, when these builds worked, everything was super fast. Waaaaay faster than Travis. I can't wait for this to be ready.

PR #1864, opened 2017-09-30
Fix false negatives when using allowed_symbols

marcelofabri:
Previously, we were not generating violations if there's a symbol which is not alphanumeric and is not on `allowed_symbols`.


marcelofabri:
I wonder if that's a sign to drop support to compiling with Swift < 3.1 😬

jpsim:
> I wonder if that's a sign to drop support for compiling with Swift < 3.1 😬 

I'd be fine to make one last release supporting Swift 3.1, then require 3.2.

marcelofabri:
> I'd be fine to make one last release supporting Swift 3.1, then require 3.2.

Sounds good!

PR #1865, opened 2017-09-30
Improve validating isViolatingCase in identifier_name

### LINKED ISSUE #1762: Non ascii operators are rejected by `identifier_name` for not starting with a lowercase letter.
While `["/", "=", "-", "+", "!", "*", "|", "^", "~", "?", ".", "%", "<", ">", "&"]` are specifically allowed as the first character of operators, anything else is rejected.

e.g. using `•` as the compose operator or `§` as function application is relatively common and useful. Including them in `allowed_symbols` doesn't circumvent this behaviour. Should it? The `isOperator` function seems otherwise very narrow in its permissiveness.

I don't really want to disable the entire feature just to allow these functions.

marcelofabri:
Fixes #1762

### DIFF
--- a/Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
@@ -23,11 +23,11 @@ extension String {
     }
 
     internal func isUppercase() -> Bool {
-        return self == uppercased()
+        return CharacterSet.uppercaseLetters.isSuperset(ofCharactersIn: self)
     }
 
     internal func isLowercase() -> Bool {
-        return self == lowercased()
+        return CharacterSet.lowercaseLetters.isSuperset(ofCharactersIn: self)
     }
 
     internal func nameStrippingLeadingUnderscoreIfPrivate(_ dict: [String: SourceKitRepresentable]) -> String {
--- a/Source/SwiftLintFramework/Rules/IdentifierNameRule.swift
+++ b/Source/SwiftLintFramework/Rules/IdentifierNameRule.swift
@@ -74,8 +74,7 @@ public struct IdentifierNameRule: ASTRule, ConfigurationProviderRule {
             }
 
             let requiresCaseCheck = configuration.validatesStartWithLowercase || isFunction
-            if requiresCaseCheck &&
-                kind != .varStatic && name.isViolatingCase && !name.isOperator {
+            if requiresCaseCheck && kind != .varStatic && name.isViolatingCase {
                 let reason = "\(type) name should start with a lowercase character: '\(name)'"
                 return [
                     StyleViolation(ruleDescription: description,
@@ -122,15 +121,11 @@ private extension String {
         guard firstCharacter.isUppercase() else {
             return false
         }
+
         guard characters.count > 1 else {
-            return true
+            return !firstCharacter.isLowercase()
         }
         let secondCharacter = String(self[characters.index(after: startIndex)])
         return secondCharacter.isLowercase()
-    }
-
-    var isOperator: Bool {
-        let operators = ["/", "=", "-", "+", "!", "*", "|", "^", "~", "?", ".", "%", "<", ">", "&"]
-        return !operators.filter(hasPrefix).isEmpty
     }
 }
--- a/Source/SwiftLintFramework/Rules/IdentifierNameRuleExamples.swift
+++ b/Source/SwiftLintFramework/Rules/IdentifierNameRuleExamples.swift
@@ -21,7 +21,9 @@ internal struct IdentifierNameRuleExamples {
         "func isOperator(name: String) -> Bool",
         "func typeForKind(_ kind: SwiftDeclarationKind) -> String",
         "func == (lhs: SyntaxToken, rhs: SyntaxToken) -> Bool",
-        "override func IsOperator(name: String) -> Bool"
+        "override func IsOperator(name: String) -> Bool",
+        "func • <T>(lhs: T, rhs: T) -> Bool",
+        "func __myFunctionThatIWantToDiscourageInSwift() {}"
     ]
 
     static let triggeringExamples = [
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,11 @@
   [JP Simard](https://github.com/jpsim)
   [#1002](https://github.com/realm/SwiftLint/issues/1002)
 
+* Only trigger a violation for not starting with a lowercase letter in
+  `identifier_name` rule if declaration starts with an uppercase letter.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#1762](https://github.com/realm/SwiftLint/issues/1762)
+
 ## 0.22.0: Wrinkle-free
 
 ##### Breaking
--- a/Rules.md
+++ b/Rules.md
@@ -4876,6 +4876,14 @@ func == (lhs: SyntaxToken, rhs: SyntaxToken) -> Bool
 override func IsOperator(name: String) -> Bool
 ```
 
+```swift
+func • <T>(lhs: T, rhs: T) -> Bool
+```
+
+```swift
+func __myFunctionThatIWantToDiscourageInSwift() {}
+```
+
 </details>
 <details>
 <summary>Triggering Examples</summary>


marcelofabri:
I'm having second thoughts if we really should be that much more permissive here.

Source/SwiftLintFramework/Rules/IdentifierNameRule.swift
```diff
@@ -122,15 +121,11 @@ private extension String {
         guard firstCharacter.isUppercase() else {
             return false
         }
+
```
sirlantis: Maybe this code section needs a general overhaul - at the moment it also looks like you can dodge it by starting an identifier with two uppercase letters, like `let UIcolor`. Maybe this could be an alternative:

```
extension String {
    var isViolatingCase: Bool {
        guard let firstUnicodeScalar = unicodeScalars.first else { return false }

        if CharacterSet.lowercaseLetters.contains(firstUnicodeScalar) {
            return false
        } else if CharacterSet.uppercaseLetters.contains(firstUnicodeScalar) {
            guard characters.count > 1 else { return true } // disallow single uppercases letters
            return unicodeScalars.contains(where: CharacterSet.lowercaseLetters.contains)
        } else {
            // Only allow if no letters at all (operator-ish)
            guard characters.count > 1 else { return false } // optimization
            return unicodeScalars.contains(where: CharacterSet.letters.contains)
        }
    }
}
```

```
// okay
"".isViolatingCase
"f".isViolatingCase
"f1".isViolatingCase
"_".isViolatingCase
"__".isViolatingCase
"FOO".isViolatingCase
"F1".isViolatingCase
"SOME_SNAKE_CASE".isViolatingCase

// violating
"F".isViolatingCase
"Foo".isViolatingCase
"_f".isViolatingCase
"_F".isViolatingCase
```

Source/SwiftLintFramework/Rules/IdentifierNameRule.swift
```diff
@@ -122,15 +121,11 @@ private extension String {
...
         guard characters.count > 1 else {
-            return true
+            return !firstCharacter.isLowercase()
         }
         let secondCharacter = String(self[characters.index(after: startIndex)])
         return secondCharacter.isLowercase()
     }
-
-    var isOperator: Bool {
```
sirlantis: I wonder if it would be better to just update this function to match the [Swift Language Reference](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/LexicalStructure.html#//apple_ref/doc/uid/TP40014097-CH30-ID418) (Nate Cook already [translated this to Swift once](https://gist.github.com/natecook1000/c5fb2b8cd0967f53770e)).

Source/SwiftLintFramework/Rules/IdentifierNameRuleExamples.swift
```diff
@@ -21,7 +21,9 @@ internal struct IdentifierNameRuleExamples {
         "func isOperator(name: String) -> Bool",
         "func typeForKind(_ kind: SwiftDeclarationKind) -> String",
         "func == (lhs: SyntaxToken, rhs: SyntaxToken) -> Bool",
-        "override func IsOperator(name: String) -> Bool"
+        "override func IsOperator(name: String) -> Bool",
+        "func • <T>(lhs: T, rhs: T) -> Bool",
+        "func __myFunctionThatIWantToDiscourageInSwift() {}"
```
sirlantis: Shouldn't this actually be in the `triggeringExamples` section?

Source/SwiftLintFramework/Rules/IdentifierNameRule.swift
sirlantis: I think SwiftLint allowed `_f` before though. So in that case we could adjust the last `else` branch to skip non-letters at the beginning and then do a recursive `isViolatingCase` on the rest.


- `_fooBar` -> `fooBar` -> okay
- `_FooBar` -> `FooBar` -> violation
- `___FOO_BAR` -> `FOO_BAR` -> okay

sirlantis:
```
extension String {
    var isViolatingCase: Bool {
        guard let firstUnicodeScalar = unicodeScalars.first else { return false }

        if CharacterSet.lowercaseLetters.contains(firstUnicodeScalar) {
            return false
        } else if CharacterSet.uppercaseLetters.contains(firstUnicodeScalar) {
            guard characters.count > 1 else { return true } // disallow single uppercases letters
            return unicodeScalars.contains(where: CharacterSet.lowercaseLetters.contains)
        } else {
            guard
                characters.count > 1,
                case let casedLetters = CharacterSet.lowercaseLetters.union(.uppercaseLetters),
                let casedLetterIndex = unicodeScalars.index(where: casedLetters.contains) else {
                    // Allow if no cased letters at all (operator-ish)
                    return false
            }

            return substring(from: casedLetterIndex).isViolatingCase
        }
    }
}
```

```
// okay
"".isViolatingCase
"f".isViolatingCase
"f1".isViolatingCase
"_".isViolatingCase
"__".isViolatingCase
"FOO".isViolatingCase
"F1".isViolatingCase
"SOME_SNAKE_CASE".isViolatingCase
"_f".isViolatingCase
"§".isViolatingCase
"•".isViolatingCase
"_fooBar".isViolatingCase
"___FOO_BAR".isViolatingCase

// violating
"F".isViolatingCase
"Foo".isViolatingCase
"_F".isViolatingCase
"_FooBar".isViolatingCase
```

sirlantis: I'm not super happy with `F1` being valid, but to fix that we would have to adjust the "uppercase is okay if at least two characters and all upper case" to a more strict "uppercase is okay if at least 2 *cased* letters". In that case `FO1` would be okay but `F1` not. Don't know if worth the hassle though.

sirlantis: This would be the variant implementing this check:

```
extension String {
    var isViolatingCase: Bool {
        guard let firstUnicodeScalar = unicodeScalars.first else { return false }

        if CharacterSet.lowercaseLetters.contains(firstUnicodeScalar) {
            return false
        } else if CharacterSet.uppercaseLetters.contains(firstUnicodeScalar) {
            guard
                characters.count > 1, // optimization
                case let rest = unicodeScalars.suffix(from: index(after: startIndex)),
                rest.contains(where: CharacterSet.uppercaseLetters.contains) else {
                    // Disallow if only a single uppercase letter
                    return true
            }

            return unicodeScalars.contains(where: CharacterSet.lowercaseLetters.contains)
        } else {
            guard
                characters.count > 1, // optimization
                case let casedLetters = CharacterSet.lowercaseLetters.union(.uppercaseLetters),
                let casedLetterIndex = unicodeScalars.index(where: casedLetters.contains) else {
                    // Allow if no cased letters at all (operator-ish)
                    return false
            }

            return substring(from: casedLetterIndex).isViolatingCase
        }
    }
}
```

```
// okay
"".isViolatingCase
"f".isViolatingCase
"f1".isViolatingCase
"_".isViolatingCase
"__".isViolatingCase
"FOO".isViolatingCase
"F_O".isViolatingCase
"SOME_SNAKE_CASE".isViolatingCase
"_f".isViolatingCase
"§".isViolatingCase
"•".isViolatingCase
"_fooBar".isViolatingCase
"___FOO_BAR".isViolatingCase

// violating
"F".isViolatingCase
"F1".isViolatingCase
"F__".isViolatingCase
"Foo".isViolatingCase
"_F".isViolatingCase
"_FooBar".isViolatingCase
```

Doesn't look too bad, should be fast, and seems quite reasonable to me - what do you think?

Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
```diff
@@ -23,11 +23,11 @@ extension String {
     }
 
     internal func isUppercase() -> Bool {
-        return self == uppercased()
+        return CharacterSet.uppercaseLetters.isSuperset(ofCharactersIn: self)
```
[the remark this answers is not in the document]
marcelofabri: I agree that this might be an issue, but currently we only use this for single characters.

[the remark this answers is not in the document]
SDGGiesbrecht: > currently we only use this for single characters.

If you mean single scalars, it can be ignored, but if you mean single grapheme clusters (what Swift calls `Character`), this will fail.

é (e + ◌́) is one cluster, but two scalars.
```swift
"é".isUppercase() // → true?!?
```
This happens because ◌́ is not a member of `.lowercaseLetters`.

***

The old fails too for Unicode [General Category](https://en.wikipedia.org/wiki/Unicode_character_property#General_Category) Lt, such as “ǅ” (as one scalar)
```swift
"ǅ".isUppercase() // → false?!?
```
This happens because `"ǅ".uppercased()` yields `DŽ`
(Technically Dž is titlecase, not uppercase, but this should be looking for either.)

***

This would be safer (still assuming only one character):
```swift
return unicodeScalars.contains(where: CharacterSet.uppercaseLetters.contains) // Lu or Lt
```
If this is only intended for a single character, maybe extending `Character` or `UnicodeScalar` is clearer than extending `String`?

Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
```diff
@@ -23,11 +23,11 @@ extension String {
...
     }
 
     internal func isLowercase() -> Bool {
-        return self == lowercased()
+        return CharacterSet.lowercaseLetters.isSuperset(ofCharactersIn: self)
```
SDGGiesbrecht: If it really means, “is lowercase” then you need the following (see explanation above):

```swift
return unicodeScalars.contains(where: CharacterSet.lowercaseLetters.contains) // Ll
```

...but that leaves many characters that are neither. Simply using `!isUppercase` is probably what you actually want to be using instead.

(It does not look like this property is used by any other rule, so if you make the adjustments listed below, it can be dropped entirely.)

Source/SwiftLintFramework/Rules/IdentifierNameRule.swift
SDGGiesbrecht: @sirlantis,

@marcelofabri was right. This property is irrelevant. Operators do not need special treatment.

(But yes, if the property were still useful, it would have been good to change it to use the complete operator list.)

Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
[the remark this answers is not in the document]
sirlantis: > If this is only intended for a single character, maybe extending Character or UnicodeScalar is clearer than extending String?

I agree.

Source/SwiftLintFramework/Rules/IdentifierNameRule.swift
sirlantis: @SDGGiesbrecht Oh, good point! What I wanted to avoid is another hard coded list like `_` (since this PR tries to remove one). How do you feel about this one (also incorporating an extension on `Character`)?

```swift
extension Character {
    enum Casing {
        case lower, upper
    }

    var casing: Casing? {
        let unicodeScalars = String(self).unicodeScalars

        if unicodeScalars.contains(where: CharacterSet.lowercaseLetters.contains) {
            return .lower
        } else if unicodeScalars.contains(where: CharacterSet.uppercaseLetters.contains) {
            return .upper
        } else {
            return nil
        }
    }
}

extension String {
    var isViolatingCase: Bool {
        guard let firstCharacter = characters.first else { return false }

        switch firstCharacter.casing {
        case .lower?:
            return false
        case .upper?:
            guard characters.count > 1 else { return true } // optimization
            var numberOfUppercaseLetters = 1

            for character in characters.suffix(from: index(after: startIndex)) {
                switch character.casing {
                case .lower?: return true
                case .upper?: numberOfUppercaseLetters += 1
                case nil: break
                }
            }

            return numberOfUppercaseLetters < 2
        case nil:
            guard
                characters.count > 1, // optimization
                let firstCasedLetter = characters.index(where: { $0.casing != nil }) else {
                    // Allow if no cased letters at all (operator-ish)
                    return false
            }

            return substring(from: firstCasedLetter).isViolatingCase
        }
    }
}
```

Tests:

```swift
// okay
"".isViolatingCase
"f".isViolatingCase
"f1".isViolatingCase
"_".isViolatingCase
"__".isViolatingCase
"FOO".isViolatingCase
"F_O".isViolatingCase
"SOME_SNAKE_CASE".isViolatingCase
"_f".isViolatingCase
"§".isViolatingCase
"•".isViolatingCase
"_fooBar".isViolatingCase
"___FOO_BAR".isViolatingCase
"é".isViolatingCase
"ÉÉ".isViolatingCase
"ß".isViolatingCase
"Ç̌E".isViolatingCase

// violating
"F".isViolatingCase
"F1".isViolatingCase
"F__".isViolatingCase
"Foo".isViolatingCase
"_F".isViolatingCase
"_FooBar".isViolatingCase
"É".isViolatingCase
"Ç̌".isViolatingCase
```

sirlantis: Much more concise version:

```
extension String {
    var isViolatingCase: Bool {
        var numberOfUppercaseLetters = 0

        for character in characters {
            switch character.casing {
            case .lower?:
                return numberOfUppercaseLetters > 0
            case .upper?:
                numberOfUppercaseLetters += 1
            case nil:
                break
            }
        }

        return numberOfUppercaseLetters == 1
    }
}
```

SDGGiesbrecht: Valiant attempt, @sirlantis, but Unicode is more complicated than that. The `Casing` enumeration is insufficient: a `Character` may actually contain `lower`, `upper`, **both** or neither (your `nil`), so `var ǅblahblahblah = ...` winds up being allowed as lowercase. There are a few other quirks leaking through as well, but...

The abundance comments have made the reviews hard to read illegible, so I will continue the conversation in a pull request to @marcelofabri’s branch instead. I also think his original code is very concise; it only needs minor tweaks, not a complete re‐write.

jpsim:
> I realized how widespread the ASCII‐only assumptions are throughout SwiftLint. It does not really support Unicode. Doing so would be nice, but that is a project for another time.

😱. A while back, some efforts were made to make SwiftLint more unicode friendly. We even added [widespread tests](https://github.com/realm/SwiftLint/blob/0.22.0/Tests/SwiftLintFrameworkTests/TestHelpers.swift#L167-L170) in order to help prevent unicode bugs from sneaking in.

I encourage _anyone_ who finds code in SwiftLint that they suspect isn't unicode-correct to let us know by filing an issue.

I'd rather close 10 tickets explaining why code is unicode correct than miss 1 unicode issue, so please don't hesitate to file.

Thanks.

SDGGiesbrecht:
@jpsim,

> I encourage anyone who finds code in SwiftLint that they suspect isn’t unicode‐correct to let us know by filing an issue.

I might take another look, though I am busy.

For the record, there are no crash level bugs I am aware of, and in my experience everything works fine *in between* uses of Unicode characters.

The oddities only multiply when something should/should not occur at/in a Unicode cluster. Many operator rules (such as spacing) do not apply to Unicode operators, case checks (such as what I wanted to help fix here) trip over Unicode quirks, and the reported error location seems unpredictable within a grapheme cluster, depending on the rule.

This time, I started to write violating cases related to this pull request that were just supposed to test `isValiolatingCase`—would it violate or not? So I added the following to `IdentifierNameRule`...
```swift
"↓func Écrire(_ texte: String)".decomposedStringWithCanonicalMapping
```
Since the second scalar, ◌́ is not lowercase, some of the implementations discussed above would have let it pass, despite clearly being a single leading capital.

But when the test failed by reporting the error at the “u” in `func` (byte 1 instead of 0), I realized I was going to have to track a scalar/cluster indexing bug back to its source—who knows where—just to make the test do what it is supposed to. I did not have the time for a bug hunt, so I decided not to bother.

I needed to mention here that I had changed my mind so that no one would wait for me to put the pull request through (It is a significant improvement as is). That is why I wrote what I did. I did not mean it as a complaint and I am sorry if it came across that way.

jpsim:
Thanks for elaborating on what you meant @SDGGiesbrecht, I appreciate your thoroughness as always!

SDGGiesbrecht:
@marcelofabri. Please go ahead and merge. @dcutting just asked about this again in #1762.

marcelofabri:
I'm not really comfortable merging this in the current state because it makes the rules more tolerant, but I'm not sure everyone would want it. We should have a way to keep them stricter if people want to.

Unfortunately, I'm not having much time lately to revive this.

SDGGiesbrecht:
Sounds good, @marcelofabri. I just wasn’t sure it hadn’t been simply forgotten.

marcelofabri:
I'm closing this because I'm not comfortable merging it and honestly don't know what we should do. I have the feeling that combining function + variables + enum elements on the same rule might have been a mistake 😬

PR #1866, opened 2017-09-30
Remove exception for static variables in identifier_name

marcelofabri:
This probably is something that we added on the early days, before the community settles the preferred naming convention for `static` properties. 

From [Swift 3 API Design Guidelines](https://swift.org/documentation/api-design-guidelines/):

> **Follow case conventions**. Names of types and protocols are `UpperCamelCase`. Everything else is `lowerCamelCase`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ##### Breaking
 
-* None.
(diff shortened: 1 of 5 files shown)


marcelofabri:
Huh, that's weird. I ran `oss-check` locally before submitting this PR and it didn't trigger any violations.

PR #1867, opened 2017-09-30
Introduce queuedFatalError

marcelofabri:
`fatalError` prints the full path of the file, which leaks filesystem information from the machine that built the binary. Now that we release via CocoaPods, this is more critical.


jpsim:
👍. Curious if this has an impact on the binary size.

jpsim:
I'd suggest adding a custom rule to prefer `queuedFatalError` over `fatalError` in this project.

PR #1868, opened 2017-10-01
Fix csv reporter to output records with new lines

atetlaw:
The CSV reporter was outputting all fields separated by commas in 1 long record. This PR changes the reporter to out put the header and each violation as separate records using a new line, that could be imported into a spreadsheet.

This is the expected behaviour for CSV I think


jpsim:
Thanks!

PR #1873, opened 2017-10-01
Add custom rule to enforce using queuedFatalError

marcelofabri:
As suggested by @jpsim in https://github.com/realm/SwiftLint/pull/1867#issuecomment-333413014.

This actually already proved itself valuable as it caught one place that I missed!

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -51,3 +51,11 @@ custom_rules:
     message: Rule IDs must be all lowercase, snake case and not end with `rule`
     regex: identifier:\s*("\w+_rule"|"\S*[^a-z_]\S*")
     severity: error
+  fatal_error:
+    name: Fatal Error
+    excluded: "Tests/*"
+    message: Prefer using `queuedFatalError` over `fatalError` to avoid leaking compiler host machine paths.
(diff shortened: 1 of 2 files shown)


.swiftlint.yml
```diff
@@ -51,3 +51,11 @@ custom_rules:
     message: Rule IDs must be all lowercase, snake case and not end with `rule`
     regex: identifier:\s*("\w+_rule"|"\S*[^a-z_]\S*")
     severity: error
+  fatal_error:
+    name: Fatal Error
+    excluded: "Tests/*"
+    message: Prefer using `queuedFatalError` over `fatalError`
```
jpsim: Worth explaining why.

> Prefer using `queuedFatalError` over `fatalError` to avoid leaking compiler host machine paths.

.swiftlint.yml
```diff
@@ -51,3 +51,11 @@ custom_rules:
...
+    regex: fatalError
```
jpsim: Shouldn't this be `'\bfatalError\b'`? Otherwise `someotherfatalError(...)` would trigger?

Issue #1435, opened 2017-04-14
Expand (or duplicate) closure end indentation rule for other cases


keith:
Currently the `closure_end_indentation` rule is really awesome for cases like this:

```swift
foo
    .bar {
    } // this is enforced
```

I think it might be nice to spread this to other identifiers, such as this:

```swift
foo(
    bar
) // this should be enforced to be here
```

And also arrays:

```swift
let foo = [
    1, 2, 3,
] // this has to be here
```

etc.

Thoughts?

ZevEisenberg:
Probably worth duping as a Radar, since it's kind of ridiculous that these aren't the Xcode defaults, IMHO.

jpsim:
Has anyone filed a radar or a swift bug for this?

PR #1875, opened 2017-10-02
Add literal_expression_end_indentation opt-in rule

marcelofabri:
Partially implements #1435.

This is opt-in to be consistent to `closure_end_indentation`. I also expect lots of existing violations in `oss-check`.

I didn't implement the check for function calls because Xcode (9 at least) formats it as expected:

```swift
foo(
    bar
) // this should be enforced to be here
```

Also, I think this rule shouldn't actually validate that the end is on a new line.

PS: I've named it `literal_expression_end_indentation` and not `collection_literal_end_indentation` because we'll have `source.lang.swift.expr.tuple` [soon](https://github.com/apple/swift/pull/12089) and this rule could also support it.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -27,6 +27,7 @@ opt_in_rules:
   - extension_access_modifier
   - pattern_matching_keywords
   - array_init
+  - literal_expression_end_indentation
 
 file_header:
   required_pattern: |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,12 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1870](https://github.com/realm/SwiftLint/issues/1870)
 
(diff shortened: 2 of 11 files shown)


jpsim:
Nice 👍

jpsim:
Is this marked as opt-in because there are lots of violations in OSSCheck? If so, I suggest we enable it by default. If it's opt-in for other reasons (e.g. is prone to false positives) then fine, but that doesn't appear to be the case.

marcelofabri:
I'd say it's opt-in to be consistent with `closure_end_indentation` and because since this goes against Xcode's default behavior, I'm not sure this style is something that everyone agrees on.

SDGGiesbrecht:
> since this goes against Xcode's default behavior

I use Xcode’s re‐indent (⌃I) a lot and would prefer the default not be to have it and SwiftLint fighting back and forth all the time.

PR #1876, opened 2017-10-02
Inline comment formatting opt-in rule

kimdv:
Made ad new PR because something went wrong on the old #1509

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1870](https://github.com/realm/SwiftLint/issues/1870)
 
+* Add `inline_comment` rule that validates inline comments.  
+  Kim de Vos](https://github.com/kimdv)
+  [#1310](https://github.com/realm/SwiftLint/issues/1310)
+
 ##### Bug Fixes
 
 * Improve how `opening_brace` rule reports violations locations.  
--- a/Rules.md
+++ b/Rules.md
@@ -45,6 +45,7 @@
 * [Implicit Getter](#implicit-getter)
 * [Implicit Return](#implicit-return)
 * [Implicitly Unwrapped Optional](#implicitly-unwrapped-optional)
+* [InlineComment](#inlinecomment)
 * [Is Disjoint](#is-disjoint)
 * [Joined Default Parameter](#joined-default-parameter)
 * [Large Tuple](#large-tuple)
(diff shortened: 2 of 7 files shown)


marcelofabri:
There're some violations that I shouldn't be triggered IMO: https://github.com/mozilla-mobile/firefox-ios/blob/a9b288f1c85a55c00c150a8834db1fac6118ef62/Client/Frontend/Browser/TabLocationView.swift#L277:25 for example.

kimdv:
hmm @marcelofabri can't figure out how to solve this 😞 

Can you please help ? 😄

SDGGiesbrecht:
Stepping in because you asked for help, @kimdv...

***

First though, in the related rule request (#1310), @joshfriend asked for the default of 2 spaces before, 1 space after, and for both numbers to be configurable.

As it stands, this pull request flatly enforces 1 before and 1 after. (Which, incidentally, is my preference.) Is the difference intentional?

@joshfriend, I do see you said,

> I think 1 would also be an acceptable default for spaces_before.

How important is it to you to be able to configure it?

***

Back to @kimdv, I agree with every one of the test examples you provided, but as @marcelofabri noted, the violations reported on real code show the rule needs to handle several other usage cases as well.

The first, which @marcelofabri pointed out, is that the rule must always allow the comment to be indented to match the surrounding code.

```swift
func doStuff() {
    doSomething()
    // Something worth noting.
    doSomethingElse()
}
```

The simplest way is probably to only validate the number of spaces before the slashes if non‐whitespace exists before it on the same line.

Another case that should be allowed is where several comments have been carefully aligned with one another.

```swift
enum Stuff {
    case short                           // Notes...
    case mediumLength                    // More notes...
    case somethingWhichIsMuchLongerStill // Still more notes...
}
```

[Here](https://github.com/Alamofire/Alamofire/blob/b8995447518fd57af14c88a47f27434a16f60403/Tests/ServerTrustPolicyTests.swift#L1193:52) is an example from the OSSCheck with alignment spacing before, as well as the following example which comes straight from some of my [own code](https://github.com/SDGGiesbrecht/SDGCornerstone/blob/3844895ec7118de8de7555117318afd2c118d571/Sources/SDGCornerstone/Mathematics/Numbers/Protocols/RealArithmetic.swift#L1020-L1022) and has alignment spacing after. (A true block comment would have been vulnerable to auto‐indent.)

```swift
//   ∞         n + 1     2n − 1
//   ∑    ( (−1)      __θ________ )
// n = 1               (2n − 1)!
```

For these, it is probably sufficient to check the characters at the same position in the lines above and below, and if either is also `//`, then it is not really being used as an inline comment and we let it go without flagging it.

Start by adding these examples to your tests and adjust the rule description to match (something like “`//` should have spaces on either side or be aligned with other comments.”).

Then see if you can implement the logic described above to handle those tests. When you are ready (or if you need more help), report back and someone will look at the implementation more closely.

*Jeg håber, at det hjælper*, @kimdv.

fredpi:
Why is this closed now?

SDGGiesbrecht:
@fredpi,

> Why is this closed now?

@kimdv did not explain. Maybe the learning curve was too high, maybe he decided the effort was not worth it or maybe he no longer had the time.

In any case, it *does not* indicate a rejection of the rule itself. Someone could certainly continue his work (or even start over fresh if that is easier).

*Danke für die Nachfrage.*

PR #1888, opened 2017-10-05
Add override_in_extension rule

### LINKED ISSUE #1884: Rule request: overriden declaration in extension
### New Issue Checklist

### Rule Request

Adding `override` for Swift-only declarations is already not possible. It's only possible for `@objc` declarations.

From [Apple's Programming with Objective-C](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html#//apple_ref/doc/uid/TP40011210-CH6-SW1)

> Because the methods declared in a category are added to an existing class, you need to be very careful about method names.
>
> If the name of a method declared in a category is the same as a method in the original class, or a method in another category on the same class (or even a superclass), the behavior is undefined as to which method implementation is used at runtime.

```swift
// should trigger
extension Foo {
  override var bar: String { return "" }

  override func baz() {}
}
```

Just severity.

I think it should be enabled by default, but I'm afraid that this might cause a lot of violations that are not easily fixed.

---

I have an initial implementation for this already, I just need some time to clean it up.

marcelofabri:
Fixes #1884 

Let's see how many violations `oss-check` reports.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,10 @@
 
 ##### Enhancements
 
-* None.
+* Add `override_in_extension` rule that warns against overriding declarations
+  in a `extension`.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#1884](https://github.com/realm/SwiftLint/issues/1884)
 
 ##### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 ##### Enhancements
 
 * Add `override_in_extension` rule that warns against overriding declarations
-  in a `extension`.  
+  in an `extension`.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1884](https://github.com/realm/SwiftLint/issues/1884)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@
 
 ##### Enhancements
 
-* Add `override_in_extension` rule that warns against overriding declarations
-  in an `extension`.  
+* Add `override_in_extension` opt-in rule that warns against overriding
+  declarations in an `extension`.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1884](https://github.com/realm/SwiftLint/issues/1884)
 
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -10,7 +10,10 @@
 
 ##### Enhancements
 
-* None.
+* Add `override_in_extension` rule that warns against overriding declarations
+  in a `extension`.
```
sirlantis: nit: a**n** extension

Source/SwiftLintFramework/Rules/OverrideInExtensionRule.swift
```diff
@@ -0,0 +1,55 @@
...
+import Foundation
+import SourceKittenFramework
+
+public struct OverrideInExtensionRule: ASTRule, ConfigurationProviderRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "override_in_extension",
```
sirlantis: This sounds assertive (as in *do this*), instead of restrictive. "disallow_override_in_extensions" maybe?

marcelofabri: I'm not sure we have a super consistent naming convention. We already have `force_try`, `force_cast`, `fallthrough` and others that aren't restrictive.

marcelofabri:
> With people often using extensions to organize their code (e.g. a UITableViewDataSource section for a UITableViewController - can also be seen in osscheck) we should either somehow detect this situation (i.e. file also contains class) or make this opt-in.

I was thinking about this, but I was afraid of this situation:

```swift
class Foo: NSObject {
    override var description: String {
        return "1"
    }
}

extension Foo {
    override var description: String {
        return "2"
    }
}
```

However, the compiler already catches this:

> 'description' has already been overridden

So I'll work on this later 👍

jpsim:
> You are right, naming is already inconsistent across the board. So no reason to block it. :)

If someone wants to audit SwiftLint's naming for related things (flags, rule names, configuration names, config params, etc), I'd be very happy to review. If we settle on a convention, we could also enforce this to some extent with unit tests. Semi-related example: every description string should read like a sentence and be terminated by a period, which we could confirm with tests.

PR #1893, opened 2017-10-07
Make vertical_parameter_alignment(_on_call) expand tabs to 4 spaces

sirlantis:
After my pull request on SourceKitten a new parameter
is available for the lineAndCharacter APIs allowing us
to handle tabs when it comes to vertical parameter alignment.

Right now this pull-request hard-codes a width of 4.

### DIFF
--- a/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
@@ -322,4 +322,12 @@ extension File {
                                                                  length: token.length)
         return aclString.flatMap(AccessControlLevel.init(description:)) != nil
     }
+
+    internal var useTabs: Bool {
+        return true
+    }
+
+    internal var indentWidth: Int {
+        return 4
+    }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,10 @@
   [JP Simard](https://github.com/jpsim)
   [#1822](https://github.com/realm/SwiftLint/issues/1822)
 
+* Make `vertical_parameter_alignment(_on_call)` expand tabs to 4 spaces
+  and add `indent-width` to `autocorrect` command.  
+  [Marcel Jackwerth](https://github.com/sirlantis)
+
 ##### Bug Fixes
 
 * Correct equality tests for `Configuration` values. They previously didn't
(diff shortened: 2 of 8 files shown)


sirlantis:
Open to ideas how we could make the hard-coded tab width configurable and whether we should have that in this PR or could move it to a new one (keeping things small).

Since we need this value in at least two rules and two commands (lint + format) maybe it should go into the `.swiftlint.yml` as a global configuration option? Since we don't have many of those I wanted to check back first what you think.

cc @marcelofabri as original author of the files I touched

marcelofabri:
I agree this should be a global setting, either configurable via `.swiftlint.yml` or the command line (maybe both?). // cc @jpsim 

I'd also like to see other rules to implement this as well: `switch_case_alignment`, `closure_end_indentation` and `literal_expression_end_indentation`.

This will enable us to eventually implement autocorrect for these rules. 🎉

sirlantis:
Yeah, I didn't touch the other rules since I thought they don't necessarily "need" to support it. Only parameter alignment forces you to use mix indentation characters (if your codebase uses spaces). But probably best to use it everywhere consistently. :+1:

sirlantis:
@jpsim @marcelofabri Do you have any suggestions on how to pass along the indentWidth information?

I see these options right now:

1. Change `File` (SourceKitten though...) to include `indentWidth` and `tabWidth` (alternatively we could attach a KV `attributes` store)
2. Pass another argument to `Rule(configuration:)` and `ruleConfiguration.apply` (e.g. GlobalConfiguration)
3. Pass another argument to `rule.validate(file:)`

None of them are low-impact, so I'd like to check back with you first. Also still wondering whether I should do that in another PR, since this PR will get bloated.

sirlantis:
For now I moved the `indentWidth` and `useTabs` into the `File+SwiftLint` extension but they are of course not hooked up. But that could at least make it work for now without having to introduce a global variable, and we could figure out how to wire this up later.

Source/SwiftLintFramework/Extensions/File+SwiftLint.swift
```diff
@@ -322,4 +322,12 @@ extension File {
                                                                  length: token.length)
         return aclString.flatMap(AccessControlLevel.init(description:)) != nil
     }
+
+    internal var useTabs: Bool {
```
jpsim: This doesn't appear to be used.

Dschee:
I'm not sure what the current state of this PR is since this is quite old now. @sirlantis Are you still interested in implementing this? I'd be happy to review once rebased.

sirlantis:
@Dschee I'm a bit busy right now but will bump the to-do.

PR #1895, opened 2017-10-10
Add .swift-version file

jpsim:
since CocoaPods requires it when running `pod trunk push`.


jpsim:
Wow. CircleCI 😍. Push -> PR -> Green CI -> Merge all in less than 6 minutes.

Issue #1905, opened 2017-10-15
Rule request: quick_discouraged_focused_test


ornithocoder:
It's quite common - in large test suites - the usage of focused test (`fdescribe`, `fcontext`, and `fit`), especially when doing TDD.

Focused test is a pretty handy feature, but not rarely do developers push code with focused test without realizing it :-(

This rule would trigger a warning for focused tests, to remind developers they have focused tests.

## Non Triggering Examples

```swift
class TotoTests: QuickSpec {
   override func spec() {
       describe("foo") {
           describe("bar") { } 
           context("bar") {
               it("bar") { }
           }
           it("bar") { }
           itBehavesLike("bar")
       }
   }
}
```

## Triggering Examples

```swift
class TotoTests: QuickSpec {
   override func spec() {
       ↓fdescribe("foo") {
       }
   }
}
```

```swift
class TotoTests: QuickSpec {
   override func spec() {
       ↓fcontext("foo") {
       }
   }
}
```

```swift
class TotoTests: QuickSpec {
   override func spec() {
       ↓fit("foo") {
       }
   }
}
```

```swift
class TotoTests: QuickSpec {
   override func spec() {
       describe("foo") {
           ↓fit("bar") { }
       }
   }
}
```

```swift
class TotoTests: QuickSpec {
   override func spec() {
       context("foo") {
           ↓fit("bar") { }
       }
   }
}
```

```swift
class TotoTests: QuickSpec {
   override func spec() {
       describe("foo") {
           context("bar") {
               ↓fit("toto") { }
            }
        }
    }
}
```

```swift
class TotoTests: QuickSpec {
   override func spec() {
       ↓fitBehavesLike("foo")
   }
}
```

jpsim:
Done in #1906.

PR #1904, opened 2017-10-15
Adds [f,x]{describe, context, itBehavesLike} to quick_discouraged_call rule

ornithocoder:
Fixes #1903.

PR #1906, opened 2017-10-15
Adds opt-in quick_discouraged_focused_test rule

ornithocoder:
Implements #1905.

### DIFF
--- a/Source/SwiftLintFramework/Rules/QuickDiscouragedFocusedTestRule.swift
+++ b/Source/SwiftLintFramework/Rules/QuickDiscouragedFocusedTestRule.swift
@@ -1,5 +1,5 @@
 //
-//  QuickDiscouragedCallRule.swift
+//  QuickDiscouragedFocusedTestRule.swift
 //  SwiftLint
 //
 //  Created by Ornithologist Coder on 10/15/17.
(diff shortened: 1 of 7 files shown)


Source/SwiftLintFramework/Rules/QuickDiscouragedFocusedTestRule.swift
```diff
@@ -0,0 +1,130 @@
...
+        return [StyleViolation(ruleDescription: type(of: self).description,
+                               severity: configuration.severity,
+                               location: Location(file: file, byteOffset: offset))]
+    }
+}
+
+private enum QuickFocusedCallKind: String {
+    case fdescribe
+    case fcontext
+    case fit
```
marcelofabri: should this also catch `fitBehavesLike`?

ornithocoder: Definitely! ;-)

ornithocoder: I've also added it as a valid keyword in #1904. Thanks 👍

jpsim:
LGTM! Thanks for the PR 👍

PR #1913, opened 2017-10-18
Fix false positive in `multiline_parameters` rule when parameter is a closure has default value

### LINKED ISSUE #1912: multiline_parameters false positive when closures have default values
### New Issue Checklist

### Bug Report
When there is a default parameter for a closure, linting fails.

```bash
$ echo "func function(\nparam1: Int,\nparam2: @escaping (Int, Int) -> Void = { _, _ in }\n) {\n}" | swiftlint lint --use-stdin
<nopath>:1:6: warning: Multiline Parameters Violation: Functions and methods parameters should be either on the same line, or one per line. (multiline_parameters)
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? **0.23.1**
* Installation method used (Homebrew, CocoaPods, building from source, etc)? **Homebrew or CocoaPods**, happens on both

```yml
disabled_rules:
  - closing_brace
  - cyclomatic_complexity
  - force_cast
  - force_try
  - identifier_name
  - line_length
  - nesting
  - notification_center_detachment
  - todo
  - trailing_comma
  - trailing_whitespace
  - type_name
  - weak_delegate
  - file_length
excluded:
  - Carthage
  - Pods
colon:
  flexible_right_spacing: true
line_length: 400
type_body_length:
  - 2048
  - 2048
function_body_length:
  - 400
  - 400
force_try:
  severity: warning
reporter: "xcode"
large_tuple:
  - 4
  - 4
opt_in_rules:
  - closure_end_indentation
  - closure_spacing
  - empty_count
  - explicit_init
  - fatal_error_message
  - first_where
  - joined_default_parameter
  - multiline_parameters
  - overridden_super_call
  - prohibited_super_call
  - vertical_parameter_alignment_on_call
```

* Which Xcode version are you using (check `xcode-select -p`)? **9.0**

```swift
// This triggers a violation:
func function(
    param1: Int,
    param2: @escaping (Int, Int) -> Void = { _, _ in }
) {
}

func function(
    param1: Int,
    param2: @escaping (Int) -> Void = { _ in }
) {
}
```

ornithocoder:
Fixes #1912.

### DIFF
--- a/Source/SwiftLintFramework/Rules/MultilineParametersRule.swift
+++ b/Source/SwiftLintFramework/Rules/MultilineParametersRule.swift
@@ -40,6 +40,7 @@ public struct MultilineParametersRule: ASTRule, OptInRule, ConfigurationProvider
         for structure in dictionary.substructure {
             guard
                 let structureOffset = structure.offset,
+                structure.typeName != nil,
                 let structureKind = structure.kind, SwiftDeclarationKind(rawValue: structureKind) == .varParameter,
                 let (line, _) = file.contents.bridge().lineAndCharacter(forByteOffset: structureOffset),
                 offset..<(offset + length) ~= structureOffset
--- a/Source/SwiftLintFramework/Rules/MultilineParametersRule.swift
+++ b/Source/SwiftLintFramework/Rules/MultilineParametersRule.swift
@@ -12,6 +12,8 @@ import SourceKittenFramework
 public struct MultilineParametersRule: ASTRule, OptInRule, ConfigurationProviderRule {
     public var configuration = SeverityConfiguration(.warning)
 
+    private typealias ParameterRange = (offset: Int, length: Int)
+
     public init() {}
 
     public static let description = RuleDescription(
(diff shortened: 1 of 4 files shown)


Source/SwiftLintFramework/Rules/MultilineParametersRule.swift
```diff
@@ -40,6 +40,7 @@ public struct MultilineParametersRule: ASTRule, OptInRule, ConfigurationProvider
         for structure in dictionary.substructure {
             guard
                 let structureOffset = structure.offset,
+                structure.typeName != nil,
```
marcelofabri: I don't think this is enough to fix the issue, as you can explicit use a type for the closure parameter:

```swift
func foo(param1: Int,
         param2: Bool,
         param3: @escaping (Int) -> Void = { (x: Int) in }) { }
```

ornithocoder: Yep. You're right. :'-( Do you see a not-so-dirty way to filter out these parameters within parameters?

marcelofabri: Can’t think about anything right now, but I vaguely remind dealing with this before. Maybe one or the rules has this implemented?

ornithocoder: I'll take a look. Thanks :-)

ornithocoder: I can check if the offset of the parameter isn't in the range of other parameters.

ornithocoder: Done. I've pushed another commit that checks the range of the parameter. If the parameter is in the bounds of another parameter, it rejects it. Something like this:  a 1, a 2, and a 3 would be
filtered out since they're inside parameter 3.

```
dictionary.substructure varParameters
----------------------------------------------------------------|
|        |---------| |---------| |----------------------------| |
|        | param 1 | | param 2 | | param 3: |a 1| |a 2| |a 3| | |
|        |---------| |---------| |----------------------------| |
----------------------------------------------------------------|
```

PR #1914, opened 2017-10-19
Adds missing Linux tests

ornithocoder:
No need for changelog.


jpsim:
If you install Sourcery on your development machine, this should be done automatically when you build with Xcode.

ornithocoder:
Yeah. It was my fault, I was using Visual Studio Code and compiling/testing via `swift test`. Then when I was working on another rule (using Xcode) I realized those tests were missing. Thanks!

PR #1927, opened 2017-10-26
Fix caching on Linux

jpsim:
This has never worked for two reasons:

1. We've used dictionaries to represent cache descriptions, which don't guarantee stable ordering of keys across invocations. This is true both on Darwin and Linux, but in practice ordering varies significantly more on Linux.
2. Storing a `TimeInterval` value in a `[String: Any]` dictionary and retrieving it again will not be dynamically castable to `Double` or `TimeInterval` but will be castable to `Int`.


marcelofabri:
Can we have a test that would fail without these changes? 🤔

jpsim:
I haven't been able to come up with one.

Issue #1932, opened 2017-11-06
Rule request: Use min() or max() instead of sorted().first or sorted().last


tomquist:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Rule Request

If this is a new rule request, please ignore all sections below this one, format
this issue's title as `Rule Request: [Rule Name]` and describe:

1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.
First sorting a sequence and then just reading the first or last element is unnecessarily consuming memory and is much slower (`O(n log(n))`) than searching for the minimum element in a sequence (`O(n)`).

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.
```swift
    // Should trigger
    myList.sorted().first
    myList.sorted(by: { $0.description < $1.description }).first
    myList.sorted(by: >).first
    myList.map { $0 + 1 }.sorted().first
    myList.sorted(by: someFunction).first
    myList.map { $0 + 1 }.sorted { $0.description < $1.description }.first
    myList.sorted().last
    myList.sorted().last?.something()
    myList.sorted(by: { $0.description < $1.description }).last
    myList.map { $0 + 1 }.sorted().last
    myList.sorted(by: someFunction).last
    myList.map { $0 + 1 }.sorted { $0.description < $1.description }.last
```
```swift
    // Should not trigger
    let min = myList.min()\n
    let min = myList.min(by: { $0.description < $1.description })
    let min = myList.min(by: >)
    let min = myList.max()
    let min = myList.max(by: { $0 < $1 })
```
3. Should the rule be configurable, if so what parameters should be configurable?
Just severity.
4. Should the rule be opt-in or enabled by default? Why?
The rule may lead to false positives on non-sequence API where there is no min/max function. Similar to `first_where` and `contains_over_first_not_nil` this rule should be opt-in.

jpsim:
Done in #1933.

PR #1933, opened 2017-11-06
Added rule to prefer min()/max() over sorted().first and sorted().last

tomquist:
Implements #1932

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@
 
 ##### Enhancements
 
+* Add `sorted_first_last` opt-in rule to encourage using `min()` or `max()`
+  over `sorted().first` or `sorted().last`.  
+  [Tom Quist](https://github.com/tomquist)
+  [#1932](https://github.com/realm/SwiftLint/issues/1932)
+
 * Add `quick_discouraged_focused_test` opt-in rule which warns against
   focused tests in Quick tests.  
   [Ornithologist Coder](https://github.com/ornithocoder)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 ##### Enhancements
 
+* Make `first_where` and `contains_over_first_not_nil` rules also detect cases
+  where calls to `filter` and `first` are parenthesized.
+  [Tom Quist](https://github.com/tomquist)
+
 * Add `sorted_first_last` opt-in rule to encourage using `min()` or `max()`
   over `sorted().first` or `sorted().last`.  
   [Tom Quist](https://github.com/tomquist)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,9 +59,9 @@
 
 ##### Bug Fixes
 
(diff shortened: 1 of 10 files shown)


SDGGiesbrecht:
@tomquist,

> Use confusing date format which seems to be preferred by SwiftLint

Dass Amerikaner “confusing” sind weiß jeder. ☺ Doch in diesem Fall ist das deutsche Format genauso sinnlos. ☺ Die [ISO](https://de.wikipedia.org/wiki/ISO_8601#Zusammenfassung) weiß was sie tut. ☺

Aber am Ende liegt Realms Hauptsitz in San Fransisco – also machen wir alle trotzdem mit.

jpsim:
Thanks for the PR! We really need to come up with some cleaner utilities to match multiple function calls. This PR contains duplicated code very similar to what's used to detect `.first != nil` and `.filter { }.first`.

Could you possibly refactor `FirstWhereRule`, `ContainsOverFirstNotNilRule` and `SortedFirstLastRule` to share common code?

tomquist:
I gave it a try. Not sure whether this is the best approach though.

Source/SwiftLintFramework/Protocols/CallPairRule.swift
```diff
@@ -0,0 +1,81 @@
...
+        let contents = file.contents.bridge()
+        let structure = file.structure
+
+        let violatingLocations: [Int] = firstRanges.flatMap { range in
+            guard let bodyByteRange = contents.NSRangeToByteRange(start: range.location,
+                                                                  length: range.length),
+                case let firstLocation = range.location + range.length - 1,
+                let firstByteRange = contents.NSRangeToByteRange(start: firstLocation,
+                                                                 length: 1) else {
+                                                                    return nil
```
jpsim: The general style used elsewhere in this project is to indent the body of `if`/`guard`/block statements by one indentation level more than the parent level. I know Xcode makes this difficult, but I'd like to stay consistent with the current style for now.

Same with the call to `methodCall` below.

Source/SwiftLintFramework/Protocols/CallPairRule.swift
```diff
@@ -0,0 +1,81 @@
...
+//  Copyright © 2017 Realm. All rights reserved.
+//
+
+import SourceKittenFramework
+
+public protocol CallPairRule: Rule {}
+
+extension CallPairRule {
+
+    public func validate(file: File,
```
jpsim: It would be very useful to document this function.

jpsim:
Looks like that code sharing fixed missing violations in the first-where rule! It's paying off already. Could you please add a new changelog entry in the bug fixes section covering that?

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
 
 ##### Enhancements
 
+* Make `first_where` and `contains_over_first_not_nil` rules also detect cases
```
marcelofabri: I'd add this under bug fixes

jpsim:
Thanks for the great PR!

tomquist:
Thanks for guiding me!

Issue #1931, opened 2017-11-02
[Rule Request]: Private @IBAction opt-in rule


ornithocoder:
This is a request to introduce a new opt-in rule which warns against public IBActions. 

Triggering example:

```swift
@IBAction func ↓fooButtonTapped(_ sender: UIButton) {
    // ...
}
```

Non-triggering example:

```swift
@IBAction private func fooSegmentedControlChanged(_ sender: UISegmentedControl) {
    // ...
}
```

```swift
private extension Toto {
    @IBAction func fooSegmentedControlChanged(_ sender: UISegmentedControl) {
        // ...
    }
}
```
Workaround via custom rule (snippet from `.swiftlint.yml`) while the rule isn't implemented/merged into master:

```yml
custom_rules:
  private_ibaction:
    name: "Private @IBAction"
    regex: "@IBAction\\s+func"
    message: "@IBAction should be private"
    severity: warning
```

ornithocoder:
@marcelofabri @jpsim let me know if this makes sense to you. In case it does, I'll work on it.

I'm my project we have all our `@IBAction` private, and we test the code by by sending actions (`.sendActions(_:)`) to the controllers (touch up inside, value changed, etc...).

marcelofabri:
I think this would be a nice rule 👍

jpsim:
Sounds reasonable as opt-in.

ornithocoder:
Super. I'll work on it tomorrow morning :-) Thanks guys!

marcelofabri:
Implemented in #1935.

PR #1935, opened 2017-11-07
Adds private_action opt-in rule

ornithocoder:
Implements #1931.


marcelofabri:
Thanks @ornithocoder!

ornithocoder:
Thank you, @marcelofabri!

Issue #1585, opened 2017-05-30
Rule to enforce ordering of access modifies and override keyword


mgrebenets:
Feature request: add a rule that can enforce the ordering of access modifiers and `override` keyword.

E.g. the "override first" rule could be introduced, to ensure that `override` keyword is always put before access modifiers:

```swift
// Override first
override public internal(set) var property: Bool

// Access modifies first
public override var property: Bool
```

marcelofabri:
Subset of #387

masters3d:
This does seem to be the most popular arrangement. I did a draft rule https://github.com/realm/SwiftLint/pull/1501. I am wondering if we should make it so the default is 
`let beforeACL = ["override"]` 
If somebody wants to have another list they would have to add a list to their config file `["override", "final"]`

masters3d:
I updated this [rule](https://github.com/realm/SwiftLint/pull/1501). It can be configured as so:
```
opt_in_rules:
  - modifiers_order

modifiers_order:
  before_acl: ["override"]
  after_acl: ["class"]

```

SD10:
This would be an awesome addition 😃 The inconsistent ordering in my code-base is driving me crazy.

jpsim:
Done in #1501.

superarts:
> This does seem to be the most popular arrangement.
> I did a draft rule #1501. I am wondering if we should make it so the default is
> 
> ```swift
> let beforeACL = ["override"]
> ```
> 
> If somebody wants to have another list they would have to add a list to their config file like so:
> `["override", "final"]`

Any reference of this please? I search in https://github.com/apple/swift and found:

- less than 43 [override public](https://github.com/apple/swift/search?q=%22override+public%22&unscoped_q=%22override+public%22)
- 72+ [public override](https://github.com/apple/swift/search?q=%22public+override%22&unscoped_q=%22public+override%22)

gobetti:
+1 to @superarts question - not really questioning the decision, for some reason I like `override` first but it's not Xcode's default so often this brings up team discussions.

I've just repeated the search suggested above and I got 24 `override public` vs. 75 `public override`
Also 3 `override open` vs. 27 `open override`

PR #1945, opened 2017-11-12
Introduction of enforced order of keywords upon defining access control while overriding

dirtydanee:
I have attempted to introduce a new rule for SwiftLint. The rule is designed to warn against violations of the order of keywords upon defining access control and overriding on any type of property or function.
Using Xcode's autocompletion, will result the definition of the access control to be put first, and the override keyword second. I would to like enforce this order, and make it a standard for every Swift project.

### DIFF
--- a/Rules.md
+++ b/Rules.md
@@ -1,6 +1,7 @@
 
 # Rules
 
+* [Access Control Override Order](#access-control-override-order)
 * [Array Init](#array-init)
 * [Attributes](#attributes)
 * [Block Based KVO](#block-based-kvo)
@@ -118,6 +119,129 @@
 * [Weak Delegate](#weak-delegate)
 * [XCTFail Message](#xctfail-message)
 --------
+
+## Access Control Override Order
+
+Identifier | Enabled by default | Supports autocorrection | Kind 
+--- | --- | --- | ---
+`access_control_override_order` | Enabled | No | lint
+
+Access control property keywords should be followed by the override keyword.
+
+### Examples
+
+<details>
+<summary>Non Triggering Examples</summary>
+
+```swift
+public override init()
+
+```
+
+```swift
+internal override init()
+
+```
+
+```swift
+private override init(){}
+
+```
+
+```swift
+open override var foo: String
+
+```
+
+```swift
+public override var foo: String
+
+```
+
+```swift
+internal override var foo: String
+
+```
+
+```swift
+private override var foo: String
+
+```
+
+```swift
+open override func foo() -> String
+```
+
+```swift
+public override func foo() -> String
+```
+
+```swift
+internal override func foo() -> String
+```
+
+```swift
+private override func foo() -> String
+```
+
+</details>
+<details>
+<summary>Triggering Examples</summary>
+
+```swift
+↓override public init()
+
+```
+
+```swift
+↓override internal init()
+
+```
+
+```swift
+↓override private init(){}
+
+```
+
+```swift
+↓override open var foo: String
+
+```
+
+```swift
+↓override public var foo: String
+
+```
+
+```swift
+↓override internal var foo: String
+
+```
+
+```swift
+↓override private var foo: String
+
+```
+
+```swift
+↓override open func foo() -> String
+```
+
+```swift
+↓override public func foo() -> String
+```
+
+```swift
+↓override internal func foo() -> String
+```
+
+```swift
+↓override private func foo() -> String
+```
+
+</details>
+
+
 
 ## Array Init
 
--- a/Source/SwiftLintFramework/Rules/AccessControlOverrideOrderRule.swift
+++ b/Source/SwiftLintFramework/Rules/AccessControlOverrideOrderRule.swift
@@ -1,0 +1,97 @@
+//
+//  AccessControlOverrideOrderRule.swift
+//  SwiftLint
+//
+//  Created by Daniel.Metzing on 12/11/17.
+//  Copyright © 2017 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+public struct AccessControlOverrideOrderRule: ASTRule, ConfigurationProviderRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "access_control_override_order",
+        name: "Access Control Override Order",
+        description: "Access control property keywords should be followed by the override keyword.",
+        kind: .lint,
+        nonTriggeringExamples: [
+            "public override init()\n",
+            "internal override init()\n",
+            "private override init(){}\n",
+            "open override var foo: String\n",
+            "public override var foo: String\n",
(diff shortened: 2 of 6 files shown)


Source/SwiftLintFramework/Rules/AccessControlOverrideOrderRule.swift
```diff
@@ -0,0 +1,114 @@
...
+
+        return []
+    }
+
+    private func isOverride(attributes: [String]) -> Bool {
+        return attributes.contains("source.decl.attribute.override")
+    }
+
+    private func syntaxTokens(at keywordOffset: Int, in file: File) -> [SyntaxToken]? {
+        return file
```
ornithocoder: How about

```swift
    return file
        .syntaxTokensByLine()
        .flatMap { $0.first { $0.contains { $0.offset == keywordOffset } }
    }
}
```

ornithocoder: `.first` assumes a particular order. Is the order guaranteed?

Source/SwiftLintFramework/Rules/AccessControlOverrideOrderRule.swift
```diff
@@ -0,0 +1,114 @@
...
+                            .contains {
+                                return $0
+                                    .offset == keywordOffset
+                            }
+                    }
+            }
+    }
+
+    private func attributeTokens(from syntaxTokens: [SyntaxToken]) -> [SyntaxToken] {
+        return syntaxTokens.filter { return $0.type == SyntaxKind.attributeBuiltin.rawValue }
```
ornithocoder: You can drop the `return` inside the closure.

Source/SwiftLintFramework/Rules/AccessControlOverrideOrderRule.swift
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    private func textContents(of attributeTokens: [SyntaxToken], in file: File) -> [String] {
+        return attributeTokens.flatMap {
+            return file.contents.bridge().substringWithByteRange(start: $0.offset, length: $0.length)
```
ornithocoder: You can drop the `return` inside the closure

Source/SwiftLintFramework/Rules/AccessControlOverrideOrderRule.swift
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+
+    private func safeOffSet(of token: SyntaxToken, in file: File) -> Int {
+        let start = token.offset
+        let length = token.length
+        let convertedRange = file.contents.byteRangeToNSRange(start: start, length: length)
+        return convertedRange?.location ?? 0
```
ornithocoder: isn't the `convertedRange?.location` the same as `token.offset`?

ornithocoder:
There's rule request #1585 to introduce this.

Source/SwiftLintFramework/Rules/AccessControlOverrideOrderRule.swift
```diff
@@ -0,0 +1,114 @@
...
+//  SwiftLint
+//
+//  Created by Daniel.Metzing on 12/11/17.
+//  Copyright © 2017 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+public struct AccessControlOverrideOrderRule: ASTRule, ConfigurationProviderRule {
```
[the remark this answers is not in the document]
dirtydanee: Hey man, thanks for comments!
I am not sure, i would go with it as a default rule. I think it is not a flaky rule, quite straight forward what should happen and where. Lets see what others think

Source/SwiftLintFramework/Rules/AccessControlOverrideOrderRule.swift
dirtydanee: I was having trouble getting the proper offset when emojis were in the test files. I had to do this conversion, so it gives the right offset even at that time.

Source/SwiftLintFramework/Rules/AccessControlOverrideOrderRule.swift
```diff
@@ -0,0 +1,114 @@
...
+        let attributeTokens = self.attributeTokens(from: syntaxTokens)
+        let contents = self.textContents(of: attributeTokens, in: file)
+
+        guard
+            let accessIndex = contents.index(of: accessControlLevel.description),
+            let overrideIndex = contents.index(of: "override") else {
+                return []
+        }
+
+        if accessIndex > overrideIndex {
```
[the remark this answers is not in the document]
dirtydanee: I had to change a `guard` to a `if let`, but it was good idea, thanks for the suggestion.

Rules.md
```diff
@@ -1,6 +1,7 @@
 
 # Rules
 
+* [Access Control Override Order](#access_control_override_order)
```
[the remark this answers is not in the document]
dirtydanee: thanks for the info, i knew i was missing something.

dirtydanee:
I see this rule has been covered in more detail under [this PR](https://github.com/realm/SwiftLint/pull/1501).  I will decline my PR.

ornithocoder:
Maybe it's worth checkin the status of that one. I'd love to see an access control rule merged into master. This would save a lot of code review time at work :-)

dirtydanee:
@ornithocoder i have reopened, will make the tests pass at least. Than lets see if it should be left open or not.

Source/SwiftLintFramework/Rules/AccessControlOverrideOrderRule.swift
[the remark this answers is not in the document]
dirtydanee: @ornithocoder not much activity on the PR from others. You think i should change it to opt-in? 
I think the implementation is not flaky, would not show false positives and quite straight forward.

Source/SwiftLintFramework/Rules/AccessControlOverrideOrderRule.swift
dirtydanee: There should one, and only one what contains the given keyword, so .first is mostly used to get a single array.

masters3d:
@dirtydanee did you want take over #1585 ? I've be happy to collab on it. Thanks

dirtydanee:
@masters3d looks pretty interesting, and finally we could have an acl rule in place in. 

What is left? I was looking at the state of the PR, and it was postponed until the `Swift 4.1` release. I should make the updates utilising on `Swift 4.1` to stabilise the rule?

masters3d:
Yeah. You probably want to rebase on top of the pending 4.1 compatibility PR. The main thing with 4.1 is that there are more metadata available in sourcekit to make the implementation more stable.

dirtydanee:
@masters3d cool, i will start looking at the current state this week! I will reach out if i am not advancing or having trouble understanding what you have built so far.

PR #1950, opened 2017-11-17
Clean up Makefile

norio-nomura:


norio-nomura:
I'll continue this on #1951.

PR #1952, opened 2017-11-17
Add option to autocorrect command to set indent width

RubenSandwich:
This pull request adds an option to set the indent width when reformating code.

```
$ swiftlint  autocorrect --format --indent-width 2
```

This partially resolves the issue: #319, specifically the part related to:

> Spacing: Indent using 2 spaces. Should be parametrizable, and probably 4 by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ##### Enhancements
 
+* Add `--indent-width` option to `AutoCorrectOptions`.  
+  [RubenSandwich](https://github.com/RubenSandwich)
+  [#319](https://github.com/realm/SwiftLint/issues/319)
+
 * Invalidate cache when Swift patch version changes.  
   [Norio Nomura](https://github.com/norio-nomura)
 
--- a/Source/swiftlint/Commands/AutoCorrectCommand.swift
+++ b/Source/swiftlint/Commands/AutoCorrectCommand.swift
@@ -29,7 +29,7 @@ struct AutoCorrectCommand: CommandProtocol {
             }
             if options.format {
                 let formattedContents = linter.file.format(trimmingTrailingWhitespace: true,
-                                                           useTabs: options.useTabs, indentWidth: 4)
+                                                           useTabs: options.useTabs, indentWidth: options.indentWidth)
                 _ = try? formattedContents
                     .write(toFile: linter.file.path!, atomically: true, encoding: .utf8)
             }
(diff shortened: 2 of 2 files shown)


jpsim:
I wonder if this should be a configuration value instead, paving the way to enforce indentation in linting in the future?

RubenSandwich:
@jpsim Oh I agree absolutely. Unfortunately, I didn't have enough time to look into the codebase on how to implement that as I'm going on vacation tomorrow so I started with this small change. (It also matches with replacing tabs with spaces and vice versa which to my research is limited to the autocorrect tool as well.)

RubenSandwich:
@jpsim Any update on this? I'd like to get this merged and I'll willing to do some extra work to make sure it does as I believe it would be helpful to others. 

I think it is fine as is because my changes followed the pattern set by the useTabs settings, which also only exists in the autoCorrect tool, but I can understand wanting the indentWidth setting to be in the .swiftlint-ci.yml file.

However, if that happens I think these should also happen:
1. Add indentWidth to rules
2. Add useTabs to rules
3. Emit warnings on lint for indentWidth and useTabs rules
4. Use the indentWidth and useTabs rules in autocorrect if detected

Otherwise, I believe people will be confused if the useTabs or indentWidth settings are only used for one of the tools in the toolbox. So because of that, I think this work should be merged as a first step towards the goal of supporting all of those features.

jpsim:
My thoughts on this haven't changed. This should be a configuration setting rather than a CLI flag.

Perhaps:

```yaml
indentation: tabs
# or number of spaces
indentation: 4
# or
indentation: 2
```

We could introduce this in stages similar to what you described in your last comment:

1. Introduce the configuration option
2. Read it when running `autocorrect --format`
3. Fix existing rules that don't behave well with tabs (e.g. `vertical_parameter_alignment` and others)
4. Write new rules that can leverage indentation configurations

Dschee:
@RubenSandwich  @jpsim Isn't this superseded by #1893 by now? We should close this if so.

RubenSandwich:
@Dschee I believe it this was added in [V2.41 ](https://github.com/realm/SwiftLint/blob/0c36874ec1b11e35adea835db0a135d3ff63755a/CHANGELOG.md#0241-dented-tumbler). I don't do much swift anymore so I haven't tried the recent versions of SwiftLint, but I'm closing this pull request anyway as I doubt it reflects the current codebase anymore.

Dschee:
Thanks @RubenSandwich, no worries! 👍

PR #1957, opened 2017-11-25
Refactor sourcery targets

norio-nomura:
Set dependencies to generated targets and execute Sourcery only when necessary.

- Separate `sourcery` target into` MasterRuleList.swift` and `LinuxMain.swift` in `Makefile`  
  Also set dependencies on each targets.
- Separate `Run Sourcery` build phase in `swiftlint` target into `SwiftLintFramework` ~~and `SwiftLintFrameworkTests`~~ target~~s~~ in `xcodeproj`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,13 +32,13 @@ all: build
 
 sourcery: Tests/LinuxMain.swift Source/SwiftLintFramework/Models/MasterRuleList.swift
 
-Tests/LinuxMain.swift: Tests/*/*.swift
+Tests/LinuxMain.swift: Tests/*/*.swift .sourcery/LinuxMain.stencil
 	sourcery --sources Tests --templates .sourcery/LinuxMain.stencil --output .sourcery
 	sed -e 4,11d .sourcery/LinuxMain.generated.swift > .sourcery/LinuxMain.swift
 	sed -n 4,10p .sourcery/LinuxMain.generated.swift | cat - .sourcery/LinuxMain.swift > Tests/LinuxMain.swift
 	rm .sourcery/LinuxMain.swift .sourcery/LinuxMain.generated.swift
 
-Source/SwiftLintFramework/Models/MasterRuleList.swift: Source/SwiftLintFramework/Rules/*.swift
+Source/SwiftLintFramework/Models/MasterRuleList.swift: Source/SwiftLintFramework/Rules/*.swift .sourcery/MasterRuleList.stencil
 	sourcery --sources Source/SwiftLintFramework/Rules --templates .sourcery/MasterRuleList.stencil --output .sourcery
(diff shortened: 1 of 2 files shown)


jpsim:
Can you explain your motivation for these changes? Running Sourcery is very quick, so I'd rather generate _all_ the files whenever invoked, not just a subset.

If you mostly use SwiftPM for development, add/run tests, then open Xcode and just build, without running tests, the existing behavior will ensure that all sourcery targets are updated. This new PR will mean that the `LinuxMain.swift` isn't updated and I don't know what we're gaining.

Thanks!

norio-nomura:
Certainly sourcery is quick, but the rebuilding by Xcode caused by it is slow. The purpose of this PR is to avoid the rebuilding as much as possible.
> This new PR will mean that the LinuxMain.swift isn't updated and I don't know what we're gaining.

I moved `Update Tests/LinuxMain.swift` build phase to `SwiftLintFramework` target.
🤔 But on such flow, should we add a test that confirming `LinuxMain.swift` is updated?

norio-nomura:
> Certainly sourcery is quick, but the rebuilding by Xcode caused by it is slow.

For example, testing without cleaning on my Mac:
```
xcodebuild -workspace SwiftLint.xcworkspace/ -scheme swiftlint test -only-testing:SwiftLintFrameworkTests/AttributesRuleTests
```
takes about 11sec with master:
```terminal.sh-session
➜  11:59:13 git:(master) ✗ time xcodebuild -workspace SwiftLint.xcworkspace/ -scheme swiftlint test -only-testing:SwiftLintFrameworkTests/AttributesRuleTests|xcpretty
…
xcodebuild -workspace SwiftLint.xcworkspace/ -scheme swiftlint test   8.82s user 3.03s system 114% cpu 10.391 total
xcpretty  0.71s user 0.13s system 8% cpu 10.401 total
```
about 6sec with this PR:
```terminal.sh-session
➜  12:02:37 git:(nn-refactor-sourcery-targets) ✗ time xcodebuild -workspace SwiftLint.xcworkspace/ -scheme swiftlint test -only-testing:SwiftLintFrameworkTests/AttributesRuleTests|xcpretty
…
xcodebuild -workspace SwiftLint.xcworkspace/ -scheme swiftlint test   4.05s user 1.35s system 93% cpu 5.792 total
xcpretty  0.40s user 0.09s system 8% cpu 5.803 total
```

jpsim:
I'm having a very hard time understanding why things are faster. Is it because the build phases aren't triggered if the `inputPaths` haven't changed? Sorry if this is obvious, it's flying above my head at the moment.

norio-nomura:
`MAKE(1)` said:
> `make` updates a target if it depends on prerequisite files that have been modified since the target was last modified, or if the target does not exist.

`Run Script` build phase in Xcode project should behave in the same way.

jpsim:
Oh cool, that's what I was missing. Awesome 👍

Issue #1924, opened 2017-10-24
Rule request: Avoid Yoda conditions


marcelofabri:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Rule Request

1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.

In Swift, you need a `var` or `let` keyword to do an assignment inside a condition (`if`, `guard`, etc).
Thus, you only get the disadvantages of [Yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions), but no benefits.

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

```swift
// should trigger
if 42 == foo {}
guard "str" == bar else { return }
while 10 == baz { }

// shouldn't trigger
if foo == 42 {}
guard bar == "str" else { return }
while baz == 10 { }
if foo == bar {}
```

3. Should the rule be configurable, if so what parameters should be configurable?

Just the severity.

4. Should the rule be opt-in or enabled by default? Why?
   See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a
   rule as opt-in.

I think this might cause some false positives, so I'd say opt-in. But it could be enabled by default if we can avoid those false positives.

jpsim:
Good rule, but definitely opt-in.

jpsim:
Done in #1965

PR #1965, opened 2017-12-02
Introduction of Yoda condition checking.

dirtydanee:
This PR aims to implement [#1924][1].

[1]: https://github.com/realm/SwiftLint/issues/1924


coverback:
This is amazing!

But there are some false positives, like this one: https://github.com/mozilla-mobile/firefox-ios/blob/61d2f9e8d8c8de169c564642563d7c2c97f456a8/Client/Frontend/Settings/SearchSettingsTableViewController.swift#L249:9

coverback:
Have you thought about having `nil` here as well? Or should I ask about it in the #1924?

dirtydanee:
Hey @coverback, glad you like it!

I have fixed the false positives, and i agree, the `nil` check is a good idea! I have added support for that case too.

Issue #1649, opened 2017-07-01
Rule request: explicit access control level


jdmarshall90:
See title

I know there is a rule for *top level* ACL declarations, but one for *all* declarations would be nice. This would probably be opt-in, as it would mostly be useful by frameworks and not applications.

masters3d:
Do you have any examples on how explicit non top level declarations would benefit from this requested rule? 
The use case from the top level ACL rule is frameworks.

jdmarshall90:
I don't have any specific examples; this is just more of a personal preference that I like to follow when developing frameworks. It gives more context to anything that you're looking at within the library.

mshibanami:
+1 because I'm a fan of [this opinion](https://stackoverflow.com/a/4113678/4366470):

>Because default access levels vary across languages, and many people program in more than one language. It's easy to become confused, either as the author or as someone reading the code later, thus explicit is nicer to deal with than implicit.

marcelofabri:
@mshibanami but adding them to the top level declarations should be enough to cover that case

mshibanami:
@marcelofabri 

```swift
internal class Hello {
    let v = "world"
}
```

In this case, I think it would be nice if SwiftLint warns me that `v` doesn't have a explicit access modifier. But it seems that `explicit_top_level_acl` doesn't work so.

I'm sorry if I misunderstood.

marcelofabri:
That rule won't trigger a violation because it only validate top level declarations (as `Hello` in your example).

I actually got your point now. Sorry for adding noise to the conversation.

joseprl89:
+1 to this. I've seen many Swift codebases that neglect using ACLs in the internals of a class/enum/struct. I'm unsure why this is so pervasive specifically on iOS, but I'd prefer to ensure that my team does not neglect something as crucial as class encapsulation.

My concern on class encapsulation is due to Swift defaulting to `internal`, and as opposed to Java where that's package protected, internal means your whole target can see that field.

In summary, adding this rule would improve the health of any codebase that opts in, so I'm a bit surprised it's not already baked in.

jpsim:
@joseprl89 since you appear to feel strongly about this, perhaps you're motivated enough to contribute this rule to SwiftLint yourself? I'm happy to provide guidance if you get stuck.

joseprl89:
@jpsim I had some time that I could chip into it, please review #1966 and let me know if I missed anything 👍

jpsim:
Done in #1966.

PR #1966, opened 2017-12-02
Add explicit acl rule to satisfy Issue #1649

joseprl89:
@jpsim Implemented a rule to enforce explicit acls not only on the top level, based on explicit_top_level_acl.

Basically:

```
public class A {
    let b = 5
}
```

Will trigger a warning on the let b = 5 to ensure that proper class encapsulation is not forgotten.

Tests seem to pass OK, and I updated the changelog file as proposed in the contributing guidelines, so I believe everything is in order. Let me know if anything is amiss.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,10 @@
   [JP Simard](https://github.com/jpsim)
   [#1822](https://github.com/realm/SwiftLint/issues/1822)
 
+* Added `explicit_acl` which enforces explicit access control levels.
+  [Josep Rodriguez](https://github.com/joseprl89)
+  [#1822](https://github.com/realm/SwiftLint/issues/1649)
+
 ##### Bug Fixes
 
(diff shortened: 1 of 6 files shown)


joseprl89:
@jpsim I think the thousands of warnings this rule just raised on those sample projects is proof of how pervasive in the swift community it is too ignore the ACLs.

I am uncertain however, why it did run on them, since I overrode OptInRule... Is there anything specific I should do so that it doesn't run on that job?

marcelofabri:
@joseprl89 The oss-check runs all rules to make sure we are properly testing all of them, no matter whether they're opt-in or if the project has disabled some of them.

joseprl89:
In that case, we have a legit 16k warnings on those projects apparently... Looks rather noisy :S

jpsim:
I'm fine with merging this considering it's opt-in. Thanks for your contribution @joseprl89!

joseprl89:
Awesome! Thanks a lot @jpsim!

Looking forward to start using it in production :)

PR #1973, opened 2017-12-14
Required Enum Case Rule

donald-m-ritter:
# Required Enum Case Rule
Allows for Enums that conform to a protocol to require that a specific case be present.

This is primarily for result enums where a specific case is common but cannot be inherited due cases not being
inheritable.

For example: A result enum is used to define all of the responses a client must handle from a specific service call
in an API.

````
enum MyServiceCallResponse: String {
    case unauthorized
    case unknownError
    case accountCreated
}

// An exhaustive switch can be used so any new scenarios added cause compile errors.
switch response {
   case unauthorized:
       ...
   case unknownError:
       ...
   case accountCreated:
       ...
}
````

If cases could be inherited you could put all of the common ones in an enum and then inherit from that enum:

````
enum MyServiceResponse: String {
    case unauthorized
    case unknownError
}

enum MyServiceCallResponse: MyServiceResponse {
    case accountCreated
}
````

Which would result in MyServiceCallResponse having all of the cases when compiled:

```
enum MyServiceCallResponse: MyServiceResponse {
    case unauthorized
    case unknownError
    case accountCreated
}
```

Since that cannot be done this rule allows you to define cases that should be present if conforming to a protocol.

`.swiftlint.yml`
````
required_enum_case:
  MyServiceResponse:
    unauthorized: error
    unknownError: error
````

````
protocol MyServiceResponse {}

// This will now have errors because `unauthorized` and `unknownError` are not present.
enum MyServiceCallResponse: String, MyServiceResponse {
    case accountCreated
}
````

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@
   [Daniel Metzing](https://github.com/dirtydanee)
   [#1924](https://github.com/realm/SwiftLint/issues/1924)
 
+* Add `required_enum_case` opt-in rule which allows enums that
+  conform to protocols to require one or more cases.  Useful for
+  result enums.  
+  [Donald Ritter](https://github.com/donald-m-ritter)
+
 ##### Bug Fixes
 
 * Fix false positives in `control_statement` rule when methods with keyword
--- a/Rules.md
+++ b/Rules.md
@@ -88,6 +88,7 @@
 * [Redundant Optional Initialization](#redundant-optional-initialization)
 * [Redundant String Enum Value](#redundant-string-enum-value)
 * [Redundant Void Return](#redundant-void-return)
+* [Required Enum Case](#required-enum-case)
 * [Returning Whitespace](#returning-whitespace)
 * [Shorthand Operator](#shorthand-operator)
 * [Single Test Class](#single-test-class)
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: Source/swiftlint/Commands/RulesCommand.swift [hunk 1]
```diff
@@ -7,10 +7,10 @@
 //
 
 import Commandant
-#if os(Linux)
-import Glibc
-#else
+#if !os(Linux)
```
jpsim: This change shouldn't be necessary. Could you please revert?

jpsim:
Thanks for the PR. Very creative way of using SwiftLint to fill in gaps in the Swift language. Points for that!

Also, points for thoroughness. I just have a small request, then I think this could be merged.

Source/swiftlint/Commands/RulesCommand.swift
donald-m-ritter: Sure.  I was just trying to get rid of the ordered imports warning.

jpsim: SwiftLint 0.24.0 shouldn't have that warning.

donald-m-ritter:
I made the changes you requested.

jpsim:
Really impressive first PR! Thanks!

fabb:
The description of this rule in Rules.md is quite confusing without the information in this PR

PR #1974, opened 2017-12-15
delete meaningless comment for colon

Cookiezby:


marcelofabri:
This file is autogenerated based on tests. My guess is that the comment was introduced to test offset with multi-byte characters (such as the ones in the comment).

Cookiezby:
Thanks, I will check for that

PR #1975, opened 2017-12-15
delete meaningless comment for rule Colon

Cookiezby:


jpsim:
This is actually useful in testing multi-byte characters, as @marcelofabri mentioned in the other PR.

SDGGiesbrecht:
@jpsim, @Cookiezby

> This is actually useful in testing multi-byte characters.

Maybe changing the comment to something that explains itself would reduce confusion, since it also winds up in the rule demonstrations. Right now the comment is just gibberish equivalent to “uavzqjwrm” in Latin characters (as far as I can tell using a dictionary). Maybe `// マルチバイト文字/multi‐byte characters` or something similar would be an improvement.

It might eventually be nice to be able to mark some tests to be filtered out from the generated examples, since some very valuable tests are also very useless clutter for the examples. However that is out of scope for now.

(I still plan on providing better, realistic Unicode tests when I can find the time. Besides my own code I have already run into entire Swift projects in Chinese, Portuguese, Japanese, German and Korean.)

jpsim:
You're right @SDGGiesbrecht, maybe we could just change it to some multi-byte length emoji graphemes. @Cookiezby could you open a new PR to do that?

marcelofabri:
Do we actually need this? I think this was added before we added emojis to all tests. Probably this test case should be just removed.

PR #1980, opened 2017-12-21
Fix false positives in `unused_closure_parameter` rule

### LINKED ISSUE #1979: False positive in unused_closure_parameter
### New Issue Checklist

```bash
$ swiftlint lint --path test.swift
Linting Swift files at path test.swift
Linting 'test.swift' (1/1)
/Users/jsimard/Downloads/test.swift:3:5: warning: Unused Closure Parameter Violation: Unused parameter "manager" in a closure should be replaced with _. (unused_closure_parameter)
```

### Environment

0.24.0

Homebrew

None.

None.

/Applications/Xcode-9.2.app/Contents/Developer

```swift
import Foundation

({ (manager: FileManager) in
  print(manager)
})(FileManager.default)
```

Also, as reported by @irace:

image

jpsim:
when closure is wrapped in parentheses. Fixes #1979.

Definitely not elegant, but it should help.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1968](https://github.com/realm/SwiftLint/issues/1968)
 
+* Fix false positives in `unused_closure_parameter` rule when closure is wrapped
+  in parentheses.  
+  [JP Simard](https://github.com/jpsim)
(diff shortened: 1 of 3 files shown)


jpsim:
Unfortunately, this doesn't address the false negatives. For example, this won't trigger a violation:

```swift
import Foundation

({ (manager: FileManager) in
  print(123)
})(FileManager.default)
```

PR #1983, opened 2017-12-28
Configurable indent width

nmccann:
This allows for more configurable formatting, using:

```
$ swiftlint autocorrect --format --indent-width 2
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@
   [Daniel Metzing](https://github.com/dirtydanee)
   [#1924](https://github.com/realm/SwiftLint/issues/1924)
   
-* Indentation can now be specified via a configuration file
+* Indentation can now be specified via a configuration file.  
   [Noah McCann](https://github.com/nmccann)
   [RubenSandwich](https://github.com/RubenSandwich)
   [#319](https://github.com/realm/SwiftLint/issues/319)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@
   conditional statement.  
   [Daniel Metzing](https://github.com/dirtydanee)
   [#1924](https://github.com/realm/SwiftLint/issues/1924)
+  
+* Add `--indent-width` option to `AutoCorrectOptions`  
+  [Noah McCann](https://github.com/nmccann)
 
 ##### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,10 @@
   [Daniel Metzing](https://github.com/dirtydanee)
   [#1924](https://github.com/realm/SwiftLint/issues/1924)
   
-* Add `--indent-width` option to `AutoCorrectOptions`  
+* Indentation can now be specified via a configuration file
   [Noah McCann](https://github.com/nmccann)
+  [RubenSandwich](https://github.com/RubenSandwich)
+  [#319](https://github.com/realm/SwiftLint/issues/319)
 
 ##### Bug Fixes
 
--- a/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
@@ -47,13 +47,13 @@ extension Configuration {
             return [String].array(of: object) ?? []
         }
 
-        func defaultIndentationMode(_ object: Any?) -> IndentationMode {
+        func defaultIndentation(_ object: Any?) -> Indentation {
             switch object {
             case let value as Int: return .spaces(count: value)
             case let value as String where value == "tabs": return .tabs
             default:
                 queuedPrintError("Invalid configuration for '\(Key.indentation)'. Falling back to default.")
-                return .spaces(count: 4)
+                return .default
             }
         }
 
@@ -72,7 +72,7 @@ extension Configuration {
         let whitelistRules = defaultStringArray(dict[Key.whitelistRules.rawValue])
         let included = defaultStringArray(dict[Key.included.rawValue])
         let excluded = defaultStringArray(dict[Key.excluded.rawValue])
-        let indentationMode = defaultIndentationMode(dict[Key.indentation.rawValue])
+        let indentation = defaultIndentation(dict[Key.indentation.rawValue])
 
         Configuration.warnAboutDeprecations(configurationDictionary: dict, disabledRules: disabledRules,
                                             optInRules: optInRules, whitelistRules: whitelistRules, ruleList: ruleList)
@@ -101,7 +101,7 @@ extension Configuration {
                   configuredRules: configuredRules,
                   swiftlintVersion: dict[Key.swiftlintVersion.rawValue] as? String,
                   cachePath: cachePath ?? dict[Key.cachePath.rawValue] as? String,
-                  indentationMode: indentationMode)
+                  indentation: indentation)
     }
 
     private init?(disabledRules: [String],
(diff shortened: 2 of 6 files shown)


Source/swiftlint/Commands/AutoCorrectCommand.swift
```diff
@@ -76,5 +78,8 @@ struct AutoCorrectOptions: OptionsProtocol {
             <*> mode <| Option(key: "use-tabs",
                                defaultValue: false,
                                usage: "should use tabs over spaces when reformatting")
+            <*> mode <| Option(key: "indent-width",
+                               defaultValue: 4,
+                               usage: "the width to indent lines when reformatting")
```
nmccann: I'm open to suggestions about the `usage` wording. I found it hard to describe without ending up with a tautology.

jpsim:
Thanks for the PR! Please see #1952 for my feedback on a very similar approach.

nmccann:
@jpsim thanks, I should have known that something similar had been proposed before. I had seen the PR that implemented the `useTabs`, but not that one.

I agree that having this as a configuration is preferable, but I think this (or the linked PR) would be a good first step and valuable in it's own right - without it, the `autocorrect --format` option provides no value for users that don't want the hardcoded indent width. 

A larger refactoring could be implemented in the future that would deprecate this and `useTabs` as command line options.

jpsim:
What makes you think that it would be a large refactoring?

I'm not inclined to add temporary APIs when there's a clear path forward.

nmccann:
Fair enough, my estimate that it would be a large refactoring isn't grounded in much science as I am not familiar with the project - I was mostly basing it on the age of the previous PR.

nmccann:
The latest push needs updates to it's tests, but I'd like to know if I'm on the right track. I'll add comments on the relevant files where I'm unsure about my implementation.

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -17,10 +17,26 @@ public struct Configuration: Hashable {
         case allEnabled
     }
 
+    public enum IndentationMode: Equatable {
```
jpsim: Suggestion: this seems like more of a "configuration" than a "mode. `IndentationConfiguration`

jpsim: Or given that it's nested within Configuration, just `Indentation` actually. `Configuration.Indentation`

jpsim:
Yes, this direction is in line with what I had in mind.

Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
```diff
@@ -45,6 +47,16 @@ extension Configuration {
             return [String].array(of: object) ?? []
         }
 
+        func defaultIndentationMode(_ object: Any?) -> IndentationMode {
```
nmccann: Alternatively this could be extracted to the IndentationMode enum as a failable initializer.

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -59,7 +75,8 @@ public struct Configuration: Hashable {
                  ruleList: RuleList = masterRuleList,
                  configuredRules: [Rule]? = nil,
                  swiftlintVersion: String? = nil,
-                 cachePath: String? = nil) {
+                 cachePath: String? = nil,
+                 indentationMode: IndentationMode = .spaces(count: 4)) {
```
nmccann: I'm not sure if this is a good place to set the default value - the default value is duplicated in a few places like the the Configuration parser, and AutoCorrectCommand, which isn't ideal.

jpsim: Why not have a `static var default = .spaces(count: 4)`?

Source/swiftlint/Commands/AutoCorrectCommand.swift
```diff
@@ -17,6 +17,17 @@ struct AutoCorrectCommand: CommandProtocol {
     func run(_ options: AutoCorrectOptions) -> Result<(), CommandantError<()>> {
         let configuration = Configuration(options: options)
         let cache = options.ignoreCache ? nil : LinterCache(configuration: configuration)
+        let indentWidth: Int
+        let useTabs: Bool
+
+        switch configuration.indentationMode {
+        case .tabs:
+            indentWidth = 4
```
nmccann: I'm not sure it makes sense to have an indentWidth when using tabs, but it's required by the formatter - using the previous hardcoded value seems like a safe bet.

Source/swiftlint/Commands/AutoCorrectCommand.swift
```diff
@@ -29,7 +40,8 @@ struct AutoCorrectCommand: CommandProtocol {
             }
             if options.format {
                 let formattedContents = linter.file.format(trimmingTrailingWhitespace: true,
-                                                           useTabs: options.useTabs, indentWidth: 4)
+                                                           useTabs: useTabs,
```
nmccann: Currently `useTabs` is based entirely on your `Configuration` - but for backwards compatibility, I'm guessing we would want to use the command line option if it is provided. Correct? If that is the case, should I add a deprecation warning to the option?

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -17,10 +17,26 @@ public struct Configuration: Hashable {
...
+            case (_, _): return false
+            }
+        }
+    }
+
     // MARK: Properties
 
     public static let fileName = ".swiftlint.yml"
 
+    public let indentationMode: IndentationMode        // mode to use when indenting
```
nmccann: I chose the name `IndentationMode ` because it was similar to `RulesMode`, is that alright? Or does the `Mode` suffix imply something that isn't valid for this enum?

jpsim: IMO The mode suffix works for the rule mode, but not for the indentation setting.

Source/SwiftLintFramework/Models/Configuration.swift
nmccann: Agreed - that's much nicer

Source/swiftlint/Commands/AutoCorrectCommand.swift
jpsim: Correct.

jpsim:
Let me know when you're ready for another round of reviews @nmccann. I think you're on the right track here.

nmccann:
@jpsim I think I'm ready for another review.

I was unsure to what extent I should test this functionality - if you think additional tests are necessary please let me know and I can add them.

CHANGELOG.md
```diff
@@ -19,6 +19,11 @@
   conditional statement.  
   [Daniel Metzing](https://github.com/dirtydanee)
   [#1924](https://github.com/realm/SwiftLint/issues/1924)
+  
+* Indentation can now be specified via a configuration file
+  [Noah McCann](https://github.com/nmccann)
+  [RubenSandwich](https://github.com/RubenSandwich)
```
nmccann: Thanks @RubenSandwich for the original PR on this, and the detailed breakdown on the next steps necessary - that coupled with @jpsim's suggestions helped this come to fruition.

jpsim:
Thanks! Merged in b161ee00ad6d16a987d25a652f53a7216047c963 to 34d2fb3949f7140ca9ba65e3090bafd6ad61bebd with minor changes.

PR #1984, opened 2017-12-30
 Add `build` to `.PHONY` in Makefile

norio-nomura:
This will make `make` not to be misled by the existence of `build` directory.


jpsim:
Are you sure a `.PHONY` is necessary here at all? I thought it was only needed if there are local directories with conflicting names.

norio-nomura:
I think at least `build` is necessary. Because there is a case that the `build` directory exists due to #1985. 🤔
Is it only desirable to add `build`?

jpsim:
> Is it better to just add `build`?

Yes, let's do that 👍

marcelofabri:
Should we merge this? 🤔

Issue #1907, opened 2017-10-17
Rule Request: Opt-in rule that top-level constants be prefixed by `k`


jpsim:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Rule Request

**1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.**

Some codebases (i.e. Lyft) consistently require that top-level constants be named `kVarName`.

**2. Provide several examples of what _would_ and _wouldn't_ trigger violations.**

```swift
// Wouldn't trigger
private let kMarginSize = 20.0
public let kFeatureEnabled = false

struct MyStruct {
    let marginSize = 20.0
}

// Would trigger
private let marginSize = 20.0
public let featureEnabled = false
```

**3. Should the rule be configurable, if so what parameters should be configurable?**

We _might_ want to make the prefix configurable (e.g. `k` vs a custom value). I think it'd be fine to do this without allowing configurability for the first version and only add it if people express an interest in it.

**4. Should the rule be opt-in or enabled by default? Why?
   See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a
   rule as opt-in.**

Should definitely be opt-in as this is a non-standard stylistic choice.

marcelofabri:
Implemented in #1991.

PR #1991, opened 2018-01-05
Adds opt-in prefixed_toplevel_constant rule

ornithocoder:
Implements #1907.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,11 @@
   [Ornithologist Coder](https://github.com/ornithocoder)
   [#1987](https://github.com/realm/SwiftLint/issues/1987)
 
+* Adds `prefixed_toplevel_constant` opt-in rule which encourages top-level
+  constants to be prefixed by `k`.  
+  [Ornithologist Coder](https://github.com/ornithocoder)
+  [#1907](https://github.com/realm/SwiftLint/issues/1907)
+
 ##### Bug Fixes
 
 * Fix false positives in `control_statement` rule when methods with keyword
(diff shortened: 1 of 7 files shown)


marcelofabri:
@ornithocoder have you tried using (the lack of) `key.setter_accessibility` from the structure instead of using a regex? For example:

```json
{
  "key.substructure" : [
    {
      "key.nameoffset" : 4,
      "key.length" : 10,
      "key.name" : "x",
      "key.kind" : "source.lang.swift.decl.var.global",
      "key.namelength" : 1,
      "key.offset" : 0
    },
    {
      "key.nameoffset" : 15,
      "key.length" : 10,
      "key.name" : "y",
      "key.kind" : "source.lang.swift.decl.var.global",
      "key.offset" : 11,
      "key.namelength" : 1,
      "key.setter_accessibility" : "source.lang.swift.accessibility.internal"
    }
  ],
  "key.offset" : 0,
  "key.length" : 21
}
```

```swift
let x = ""
var y = ""
```

ornithocoder:
@marcelofabri nope, but now I will! Thank you very much!

ornithocoder:
Hmmm. The number of violations went up by ~200.

ornithocoder:
Really interesting, it caught [this](https://github.com/apple/swift/blob/6a34e3e9cd59e81426bc8a8a73684d35da586369/stdlib/private/SwiftPrivateLibcExtras/SwiftPrivateLibcExtras.swift#L40:12) (which makes sense).

marcelofabri:
You could check if `key.bodylength` if is `nil` or not. Although it's probably debatable if global computed properties should trigger or not as they can act as constants.

ornithocoder:
@marcelofabri the numbers are back to normal now :-)

Issue #1874, opened 2017-10-02
Rule Request: Prefer specific matcher over `XCTAssertEqual`/`XCTAssertNotEqual`


marcelofabri:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Rule Request

1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.

`XCTAssertFalse`, `XCTAssertTrue`,`XCTAssertNil`, `XCTAssertNotNil` should be used over `XCTAssertEqual`/`XCTAssertNotEqual`.

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

```swift
// triggers
XCTAssertEqual(foo, true)
XCTAssertEqual(foo, false)
XCTAssertNotEqual(foo, false)
XCTAssertEqual(foo, nil)
XCTAssertNotEqual(foo, nil)


// doesn't trigger
XCTAssertFalse(foo)
XCTAssertTrue(foo)
XCTAssertNil(foo)
XCTAssertNotNil(foo)
XCTAssertEqual(foo, 2)
XCTAssertNotEqual(foo, "false")
```

3. Should the rule be configurable, if so what parameters should be configurable?

Just severity.

4. Should the rule be opt-in or enabled by default? Why?
   See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a rule as opt-in.

Probably as a default rule, even though there's a risk of false positives.

Issue #1785, opened 2017-08-17
vertical_parameter_alignment_on_call false positive with 0.21.0


keith:
Test case:

```swift
a.b(c: d?.e, f: g,
    h: i)
```

Run with:

```
swiftlint lint --path ~/Desktop/test.swift --enable-all-rules
```

If you remove the `?` it lints correctly.

marcelofabri:
Err, for some reason SourceKit only returns two parameters in that case:

```json
{
  "key.diagnostic_stage" : "source.diagnostic.stage.swift.parse",
  "key.substructure" : [
    {
      "key.namelength" : 3,
      "key.nameoffset" : 0,
      "key.length" : 28,
      "key.substructure" : [
        {
          "key.namelength" : 1,
          "key.nameoffset" : 13,
          "key.length" : 4,
          "key.name" : "f",
          "key.bodyoffset" : 16,
          "key.bodylength" : 1,
          "key.kind" : "source.lang.swift.expr.argument",
          "key.offset" : 13
        },
        {
          "key.namelength" : 1,
          "key.nameoffset" : 23,
          "key.length" : 4,
          "key.name" : "h",
          "key.bodyoffset" : 26,
          "key.bodylength" : 1,
          "key.kind" : "source.lang.swift.expr.argument",
          "key.offset" : 23
        }
      ],
      "key.name" : "a.b",
      "key.bodyoffset" : 4,
      "key.bodylength" : 23,
      "key.kind" : "source.lang.swift.expr.call",
      "key.offset" : 0
    }
  ],
  "key.offset" : 0,
  "key.length" : 28
}
```

If I remove the `?`, it returns all 3:

```json
{
  "key.diagnostic_stage" : "source.diagnostic.stage.swift.parse",
  "key.substructure" : [
    {
      "key.namelength" : 3,
      "key.nameoffset" : 0,
      "key.length" : 27,
      "key.substructure" : [
        {
          "key.namelength" : 1,
          "key.nameoffset" : 4,
          "key.length" : 6,
          "key.name" : "c",
          "key.bodyoffset" : 7,
          "key.bodylength" : 3,
          "key.kind" : "source.lang.swift.expr.argument",
          "key.offset" : 4
        },
        {
          "key.namelength" : 1,
          "key.nameoffset" : 12,
          "key.length" : 4,
          "key.name" : "f",
          "key.bodyoffset" : 15,
          "key.bodylength" : 1,
[...]

PR #1995, opened 2018-01-07
Adds xct_specific_matcher opt-in rule

ornithocoder:
Implements #1874.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -545,6 +545,15 @@ The next release will require Swift 4.0 or higher to build.
   [JP Simard](https://github.com/jpsim)
   [#1998](https://github.com/realm/SwiftLint/issues/1998)
 
+#### Enhancements
+
+* Adds `xct_specific_matcher` opt-in rule to enforce specific matchers
+  over `XCTAssertEqual` and `XCTAssertNotEqual`.  
+  [Ornithologist Coder](https://github.com/ornithocoder)
+  [#1874](https://github.com/realm/SwiftLint/issues/1874)
+
+#### Bug Fixes
+
 ## 0.24.1: Dented Tumbler
 
 ##### Breaking
@@ -590,10 +599,6 @@ The next release will require Swift 4.0 or higher to build.
 * Added `explicit_acl` opt-in rule to enforce explicit access control levels.  
   [Josep Rodriguez](https://github.com/joseprl89)
   [#1822](https://github.com/realm/SwiftLint/issues/1649)
-* Adds `xct_specific_matcher` opt-in rule to enforce specific matchers
-  over `XCTAssertEqual` and `XCTAssertNotEqual`.  
-  [Ornithologist Coder](https://github.com/ornithocoder)
-  [#1874](https://github.com/realm/SwiftLint/issues/1874)
 
 ##### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -417,11 +417,6 @@ The next release will require Swift 4.0 or higher to build.
   [Norio Nomura](https://github.com/norio-nomura)
   [#2038](https://github.com/realm/SwiftLint/issues/2038)
 
-* Adds `xct_specific_matcher` opt-in rule to enforce specific matchers
-  over `XCTAssertEqual` and `XCTAssertNotEqual`.  
-  [Ornithologist Coder](https://github.com/ornithocoder)
-  [#1874](https://github.com/realm/SwiftLint/issues/1874)
-
 #### Bug Fixes
 
 * Fixes an issue with the `yoda_condition` rule where the severity would always
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -266,6 +266,10 @@
   setter ACLs on variable declarations.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1869](https://github.com/realm/SwiftLint/issues/1869)
+* Adds `xct_specific_matcher` opt-in rule to enforce specific matchers
+  over `XCTAssertEqual` and `XCTAssertNotEqual`.  
+  [Ornithologist Coder](https://github.com/ornithocoder)
+  [#1874](https://github.com/realm/SwiftLint/issues/1874)
 
 * Add a new `ignores_interpolated_strings` config parameter to the `line_length`
   rule to ignore lines that include interpolated strings from this rule.  
@@ -537,10 +541,7 @@ The next release will require Swift 4.0 or higher to build.
 
 #### Enhancements
 
-* Adds `xct_specific_matcher` opt-in rule to enforce specific matchers
-  over `XCTAssertEqual` and `XCTAssertNotEqual`.  
-  [Ornithologist Coder](https://github.com/ornithocoder)
-  [#1874](https://github.com/realm/SwiftLint/issues/1874)
+* None.
 
 #### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,11 @@
   [Timofey Solonin](https://github.com/biboran)
   [#2422](https://github.com/realm/SwiftLint/issues/2422)
 
+* Adds `xct_specific_matcher` opt-in rule to enforce specific matchers
+  over `XCTAssertEqual` and `XCTAssertNotEqual`.  
+  [Ornithologist Coder](https://github.com/ornithocoder)
+  [#1874](https://github.com/realm/SwiftLint/issues/1874)
+
 #### Bug Fixes
 
 * Fix `comma` rule false positives on object literals (for example, images).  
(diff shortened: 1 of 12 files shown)


Tests/SwiftLintFrameworkTests/RulesTests.swift
```diff
@@ -461,6 +461,10 @@ class RulesTests: XCTestCase {
         verifyRule(XCTFailMessageRule.description)
     }
 
+    func testXCTSpecificMatcher() {
```
marcelofabri: this is already tested on `XCTSpecificMatcherRuleTests`

marcelofabri:
Some of the violations in oss-check are false positives, such as `XCTAssertEqual(manager?.isReachableOnWWAN, false)`, because `XCTAssetFalse` takes a `Bool`, not `Bool?`.

@ornithocoder Do you think it would make sense to try to catch some obvious cases where an optional is used (e.g. optional chaining) and not generate violations? IMO it would make the rule much more usable in the real world.

ornithocoder:
@marcelofabri I can look for "?" on the string. Should I look for "!" as well to make sure it's not being force unwrapped?

The only problem is that the rule will "understand" certain optionals but not all of them. The way it is now, it always throws a violation, no matter if it has a implicit or explicit optional.

For `XCTAssertEqual(manager?.isReachableOnWWAN, false)` the developer can comply to the `xct_specific_matcher` rule by using:

* `XCTAssertFalse(manager!.isReachableOnWWAN)`, or
* `XCTAssertFalse(manager?.isReachableOnWWAN == .some(true))`.

I'll try something with the "?" to see how it goes :-)

marcelofabri:
I think checking for `?` should be enough. 

I agree that it won't solve the issue in all cases, but I think it solves a common pattern when testing methods that return optional or `throws`, for example: https://github.com/kickstarter/ios-oss/blob/ccde34f1b03ddd5ab327254e78ae7473b7473bda/KsApi/models/UserTests.swift#L52:5.

I think it'd be important to handle it if possible because the other alternatives are way worse IMO. Force unwrapping means that the test would crash instead of failing, which can be hard to diagnose. And comparing to `.some(true)` is just ugly 😅

marcelofabri:
An important detail though is that checking for `?` shouldn't be done for `XCTAsset(Not)Nil`.

ornithocoder:
@marcelofabri I noticed something really *interesting* when implementing the latest patch. When the argument is optional, it doesn't show on the structure.

On the example below, it's not optional:

```swift
XCTAssertEqual(manager.isReachableOnWWAN, false)
```

so both arguments show on the structure

```json
{
  "key.substructure" : [
    {
      "key.namelength" : 14,
      "key.nameoffset" : 0,
      "key.length" : 48,
      "key.substructure" : [
        {
          "key.namelength" : 0,
          "key.nameoffset" : 0,
          "key.length" : 25,
          "key.bodyoffset" : 15,
          "key.offset" : 15,
          "key.bodylength" : 25,
          "key.kind" : "source.lang.swift.expr.argument"
        },
        {
          "key.namelength" : 0,
          "key.nameoffset" : 0,
          "key.length" : 5,
          "key.bodyoffset" : 42,
          "key.offset" : 42,
          "key.bodylength" : 5,
          "key.kind" : "source.lang.swift.expr.argument"
        }
      ],
      "key.name" : "XCTAssertEqual",
      "key.bodyoffset" : 15,
      "key.bodylength" : 32,
      "key.kind" : "source.lang.swift.expr.call",
      "key.offset" : 0
    }
  ],
  "key.offset" : 0,
  "key.length" : 49
}
```

But on this example, it is:

```swit
XCTAssertEqual(manager?.isReachableOnWWAN, false)
```

and the optional argument is completely gone:

```json
{
  "key.substructure" : [
    {
      "key.namelength" : 14,
      "key.nameoffset" : 0,
      "key.length" : 49,
      "key.substructure" : [
        {
          "key.namelength" : 0,
          "key.nameoffset" : 0,
          "key.length" : 5,
          "key.bodyoffset" : 43,
          "key.offset" : 43,
          "key.bodylength" : 5,
          "key.kind" : "source.lang.swift.expr.argument"
        }
      ],
      "key.name" : "XCTAssertEqual",
      "key.bodyoffset" : 15,
      "key.bodylength" : 33,
      "key.kind" : "source.lang.swift.expr.call",
      "key.offset" : 0
    }
  ],
  "key.offset" : 0,
  "key.length" : 50
}
```

The following two examples are force-unwrapping the optional argument. Therefore, it's shown on the on the structure.

```swift
XCTAssertEqual(manager!.isReachableOnWWAN, false)
```

```json
{
  "key.substructure" : [
    {
      "key.namelength" : 14,
      "key.nameoffset" : 0,
      "key.length" : 49,
      "key.substructure" : [
        {
          "key.namelength" : 0,
          "key.nameoffset" : 0,
          "key.length" : 26,
          "key.bodyoffset" : 15,
          "key.offset" : 15,
          "key.bodylength" : 26,
          "key.kind" : "source.lang.swift.expr.argument"
        },
        {
          "key.namelength" : 0,
          "key.nameoffset" : 0,
          "key.length" : 5,
          "key.bodyoffset" : 43,
          "key.offset" : 43,
          "key.bodylength" : 5,
          "key.kind" : "source.lang.swift.expr.argument"
        }
      ],
      "key.name" : "XCTAssertEqual",
      "key.bodyoffset" : 15,
      "key.bodylength" : 33,
      "key.kind" : "source.lang.swift.expr.call",
      "key.offset" : 0
    }
  ],
  "key.offset" : 0,
  "key.length" : 50
}
```

```swift
XCTAssertEqual((manager?.isReachableOnWWAN)!, false)
```

```json
{
  "key.substructure" : [
    {
      "key.namelength" : 14,
      "key.nameoffset" : 0,
      "key.length" : 52,
      "key.substructure" : [
        {
          "key.namelength" : 0,
          "key.nameoffset" : 0,
          "key.length" : 29,
          "key.bodyoffset" : 15,
          "key.offset" : 15,
          "key.bodylength" : 29,
          "key.kind" : "source.lang.swift.expr.argument"
        },
        {
          "key.namelength" : 0,
          "key.nameoffset" : 0,
          "key.length" : 5,
          "key.bodyoffset" : 46,
          "key.offset" : 46,
          "key.bodylength" : 5,
          "key.kind" : "source.lang.swift.expr.argument"
        }
      ],
      "key.name" : "XCTAssertEqual",
      "key.bodyoffset" : 15,
      "key.bodylength" : 36,
      "key.kind" : "source.lang.swift.expr.call",
      "key.offset" : 0
    }
  ],
  "key.offset" : 0,
  "key.length" : 53
}
```

I've used this on the implementation not to trigger the violation for equal true/false when the argument is optional.

ornithocoder:
@marcelofabri maybe i'm too tired today, but i'm having troubles to get the argument as string.

For instance, for:

```
foo(a.bar(b)?.c, false)
```

I want `a.bar(b)?.c`, but all I get from offset and length is `a.bar(b)`

```json
{
  "key.substructure" : [
    {
      "key.namelength" : 3,
      "key.nameoffset" : 0,
      "key.length" : 23,
      "key.substructure" : [
        {
          "key.namelength" : 5,
          "key.nameoffset" : 4,
          "key.length" : 8,
          "key.name" : "a.bar",
          "key.bodyoffset" : 10,
          "key.bodylength" : 1,
          "key.kind" : "source.lang.swift.expr.call",
          "key.offset" : 4
        },
        {
          "key.namelength" : 0,
          "key.nameoffset" : 0,
          "key.length" : 5,
          "key.bodyoffset" : 17,
          "key.offset" : 17,
          "key.bodylength" : 5,
          "key.kind" : "source.lang.swift.expr.argument"
        }
      ],
      "key.name" : "foo",
      "key.bodyoffset" : 4,
      "key.bodylength" : 18,
      "key.kind" : "source.lang.swift.expr.call",
      "key.offset" : 0
    }
  ],
  "key.offset" : 0,
  "key.length" : 24
}
```

marcelofabri:
@ornithocoder I wouldn't rely on that behavior (optional chaining not showing up in structure). Check https://github.com/realm/SwiftLint/issues/1785. It was a bug in SourceKit ([SR-5717](https://bugs.swift.org/browse/SR-5717)) which was fixed by apple/swift#11542,

ornithocoder:
Hmmm. Any suggestion of what should I do in this case?

1) Use regular expressions (which I tried before, according the commit history. I don't remember why I walked away from it)
2) Don't handle optionals for now
3) Leave this PR open until we can use apple/swift#11542

Thanks :-)

marcelofabri:
Would it make sense to change the implementation based on Swift version?

For example:

## Swift >= 4.1 

Optional chaining is returned in the structure, so we could use that to get the text and check for `?`.

## Swift < 4.1

Optional chaining is NOT returned in the structure, so we could only validate calls with (at least) 2 arguments in it.

ornithocoder:
it sounds good. I'll do that.

ornithocoder:
I'll have to check why the performance looks so bad. Maybe I should check if the class is a XCTestCase  subclass (but then it would not work for people using XCTest matchers with Quick/Specta/etc...).

marcelofabri:
@ornithocoder I think it'd be ok to test if it's a subclass of `XCTestCase` or `QuickSpec`.

I'd also not trust *too* much on oss-check running on Circle CI. You can run `script/oss-check` locally to get more accurate results.

ornithocoder:
Hi @marcelofabri, sorry about that. I'll rebase and update the title.

ornithocoder:
@marcelofabri this is "ready". I was trying to use it and found some problems that I don't think it's possible to handle. For instance

```swift
var foo: Foo?

foo = // ...

XCTAssertEqual(foo.isValid, true)
```

`foo` is optional and there's no way to know it's optional, therefore the rule suggests to use `XCTAssertTrue` (which is incorrect, since `XCTAssertTrue` doesn't take optional arguments).

marcelofabri:
Merged in #2528 🚀

PR #1996, opened 2018-01-07
Adds correction examples to Rules.md

ornithocoder:
It would be nice to have correction example on `Rules.md`. But feel free to discard this PR if you don't believe it's relevant.

Entries have the following format:

```swift
// Before correction
foo.bar({} )

// After correction
foo.bar({})
```

### DIFF
--- a/Rules.md
+++ b/Rules.md
@@ -733,6 +733,19 @@ Closing brace with closing parenthesis should not have any whitespaces in the mi
 ```
 
 </details>
+<details>
+<summary>Correction Examples</summary>
+
+```swift
+// Before correction
+[].map({ ↓} )
+
+// After correction
+[].map({ })
+
+```
+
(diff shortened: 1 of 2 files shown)


marcelofabri:
I'm not sure this is useful, because that can be implied based on non triggering exemples and triggering examples. @ornithocoder do you know/remember any examples where it'd be hard to "predict" what autocorrect will do?

I'm just worried that this will add more noise on a file that's already too long.

ornithocoder:
The idea came to me when I was looking at a particular rule, a few weeks ago. Unfortunetely I don't remember which one it was.

But you're right, they can indeed be implied based on the other examples. Maybe that one rule had less triggering and non-triggering examples than it had for correction.

If I find the rule again, I'll add more triggering and non-triggering examples to it.

PR #2002, opened 2018-01-10
Removes 'Rule' word from test functions

ornithocoder:
and adds a custom rule to enforce that. No changelog entry needed.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -60,3 +60,9 @@ custom_rules:
     match_kinds:
       - identifier
     severity: error
+  rule_test_function:
+    included: Tests/SwiftLintFrameworkTests/RulesTests.swift
+    name: Rule Test Function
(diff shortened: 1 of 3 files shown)


.swiftlint.yml
```diff
@@ -60,3 +60,9 @@ custom_rules:
     match_kinds:
       - identifier
     severity: error
+  rule_test_function:
+    included: Tests/SwiftLintFrameworkTests/RulesTests.swift
+    name: Rule Test Function
+    message: Rule Test Function mustn't end with `rule`
+    regex: func\s*test\w+(r|R)ule\(\)
+    severity: error
```
jpsim: please keep the terminating newline in this file

jpsim:
Nice. Just one small change needed prior to merging.

.swiftlint.yml
ornithocoder: Sure, no problem :-)

PR #2007, opened 2018-01-16
Fix some autocorrects of `colon` doesn't work

mshibanami:
Currently `colon`'s autocorrect doesn't work if colons of function-calls are in other functions or classes like this:

```swift
func abc() {
    def(ghi:jkl) // ← This isn't corrected to "def(ghi: jkl)".
}

class ABC {
    let x = def(ghi:jkl) // ← This isn't corrected to "let x = def(ghi: jkl)".
}
```

I believe this PR will fix the problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@
 
 #### Bug Fixes
 
-* None.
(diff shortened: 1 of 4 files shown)


jpsim:
Thanks!

PR #2016, opened 2018-01-22
Add `comment_space` opt-in rule. Partially fixes #1310.

### LINKED ISSUE #1310: Rule request: Line comment formatting rule

Coder-256:
The rule checks to see if there is at least 1 space at the beginning and end of comments. See the examples to get a better idea of what triggers the rule.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,11 @@
 
 #### Enhancements
 
-* None.
+* Add `comment_space` opt-in rule which warns against comments that
+  don't begin with spaces.  
+  [Coder-256](https://github.com/Coder-256)
+  [#1310 (partial fix)](https://github.com/realm/SwiftLint/issues/1310)
+
 
 #### Bug Fixes
 
--- a/Rules.md
+++ b/Rules.md
@@ -11,6 +11,7 @@
 * [Closure Spacing](#closure-spacing)
 * [Colon](#colon)
 * [Comma Spacing](#comma-spacing)
+* [Comment Space](#comment-space)
 * [Compiler Protocol Init](#compiler-protocol-init)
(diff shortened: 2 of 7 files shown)


Dschee:
@Coder-256 Are you still interested in this getting merged? I'd be happy to review this PR once it's rebased. Please also make sure to respect what @SDGGiesbrecht said and add something like [this line](https://github.com/mozilla-mobile/firefox-ios/blob/a9b288f1c85a55c00c150a8834db1fac6118ef62/Client/Frontend/Browser/TabLocationView.swift#L277:25) to the non-triggering test examples.

Coder-256:
@Dschee Sorry, I have not been using Swift for any projects recently, and I honestly I don't fully remember much about this PR, I opened it a while ago :grimacing:. Somebody else should pick this up, I apologize for abandoning this.

Dschee:
Oh, too unfortunate. But thanks for keeping us updated!

Issue #2021, opened 2018-01-25
Fix unit tests when linting with Swift 4.1


marcelofabri:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Bug Report

Even after #2020, some unit tests are failing when a Swift 4.1 toolchain is used.

<img width="313" alt="captura de tela 2018-01-25 as 09 50 00" src="https://user-images.githubusercontent.com/833072/35403628-2af0aa08-01b5-11e8-9a90-00f8b31cc70b.png">

```
Executed 345 tests, with 1126 failures (0 unexpected) in 58.262 (58.485) seconds
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? master
* Installation method used (Homebrew, CocoaPods, building from source, etc)? source
* Which Xcode version are you using (check `xcode-select -p`)? 9.2 with `swift-4.1-DEVELOPMENT-SNAPSHOT-2018-01-23-a.xctoolchain` selected

marcelofabri:
`testDetectSwiftVersion` is only failing because I'm compiling with Swift 4.0.3, but running with a 4.1 toolchain. It shouldn't happen after we start using Xcode 9.3.

PR #2022, opened 2018-01-26
Don’t validate local vars in let_var_whitespace rule

marcelofabri:
As part of #2021 

I know that the original idea was to validate local variables, but that seemed not so useful in a real project (SwiftLint itself), triggering a lot of warnings.

Since this didn't work until Swift 4.1 anyways, I've just removed it for now. If people feel that this is useful, we can add it back as a configuration.


Uncommon:
I definitely want the rule to apply to local variables. In fact, that's more important to me (and I wrote this rule). Maybe it could be a config option?

marcelofabri:
@Uncommon feel free to open an issue to track this 💯

PR #2024, opened 2018-01-26
Make fallthrough rule opt-in

### LINKED ISSUE #1892: fallthrough false positive
### New Issue Checklist

### Bug Report

According to [rule request](https://github.com/realm/SwiftLint/issues/1834), the primary goal of marking fallthrough statement is that it can be replaced by combining cases. In most cases.
However there are cases where that's not possible, reasonable or leading to code duplication:
1. when having optional statement to switch on
2. when having a default statement already

The mentioned [linkedin styleguide](https://github.com/linkedin/swift-style-guide#34-switch-statements-and-enums) also mentioned `whenever possible`, but not saying anything on avoidance.

Take a look at the sample below. I have a block of code that should run in any other cases rather than `.needsRequest` plus a small addition for `.denied` case. 

To resolve the warning, I am forced to copy-paste  the default section which is error prone (+ possibly may lead to function_body_length violation).
I think the rule should ignore falling through to default case. It's not that hard to read and a default case already marks the common code for a several case statements.

### Environment

Swiftlint 0.23.0, installed through CocoaPods

```yml
identifier_name:
  excluded:
  - id
  - to
  - fic_UUID
  - fic_sourceImageUUID

excluded:
- Pods
```

  no
  xcode 9

```swift
switch userLocation?.accessLevel {
case .needsRequest?:
    userLocation?.requestAccess()
case .denied?:
    if permissionChecksDone {
        UIApplication.shared.openURL(URL(string: UIApplicationOpenSettingsURLString)!)
    }
    // This triggers a violation:
    fallthrough
default:
    permissionChecksDone = true
    self.updateLocationPage()
}
```

marcelofabri:
Fixes #1892

I think this is the first time we change whether a rule is opt-in or default, so I'm not sure if there's anything else we should do other than adding it as a breaking change in `CHANGELOG`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 #### Breaking
 
-* None.
+* `fallthrough` rule is now opt-in.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#1892](https://github.com/realm/SwiftLint/issues/1892)
 
 #### Enhancements
 
--- a/Rules.md
+++ b/Rules.md
@@ -3033,7 +3033,7 @@ public extension Foo {
 
 Identifier | Enabled by default | Supports autocorrection | Kind 
 --- | --- | --- | ---
-`fallthrough` | Enabled | No | idiomatic
+`fallthrough` | Disabled | No | idiomatic
 
 Fallthrough should be avoided.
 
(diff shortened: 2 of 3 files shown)


jpsim:
Merged in f8318db9.

PR #2036, opened 2018-02-04
[Swift 4.1] Add redundant_set_access_control rule

### LINKED ISSUE #1869: Rule request: redundant set access control
### New Issue Checklist

### Rule Request

When declaring a variable, one can add a redundant ACL for setter.

This just adds extra noise. I don't think it's common enough for the community actively discuss it.

```swift
class Foo {
  // should trigger
  private(set) private var bar1: Int
  internal(set) internal var bar2: Int
  internal(set) var bar3: Int // I'm not so sure about this one
  fileprivate(set) fileprivate var bar4: Int
  public(set) public var bar5: Int
  open(set) open var bar6: Int

  // shouldn't trigger
  private(set) var bar7: Int
  internal(set) public var bar8: Int
}
```

No

Enabled, as I think this is rare enough.

marcelofabri:
Fixes #1869 

The more important thing here is not the rule itself, but we have to make a decision on how we're going to mark rules that won't work on older Swift compiler versions, as Swift 4.1 includes some SourceKit fixes and improvements that will provide us more information to create rules.

We also probably should hold merging this until we have CI running Swift 4.1. I just wanted to open the PR to get feedback on the `minSwiftVersion` approach.


marcelofabri:
This is ready for review! It looks a huge PR, but most of the changes are in Rules.md.

PR #2048, opened 2018-02-12
Update danger and cocoapods

marcelofabri:
Opening as a PR just to make sure it doesn't break anything.


marcelofabri:
It does break 😅

Issue #2045, opened 2018-02-11
Rule Request: Untyped error in catch statement


dirtydanee:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

#### Description

When an `error` is being thrown from a `do - catch`, the type of the error being thrown is unknown. We might would like to handle some specific types of errors in a specific way, but most of the times the `error` is being logged to the console or reported to some third party error tracking system. In order to do that, we do not need to unwrap the `error` without the type itself, it is available for us in the `catch` statement.

The official [Apple documentation](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/ErrorHandling.html) on error handling is without a single example of unwrapping an error without specifying the type. 

#### What would trigger?
```swift
enum Error: Swift.Error {
    case invalidOperation
    case failedOperation
}

func foo() throws {
    throw Error.invalidOperation
}

do {
    try foo()
} catch let error {
    print("Action failed. Error: \(error.localizedDescription)")
}
```
#### What would not trigger?
```swift
enum Error: Swift.Error {
    case invalidOperation
    case failedOperation
}

func foo() throws {
    throw Error.invalidOperation
}
```
Example 1:

```swift
do {
    try foo()
} catch Error.invalidOperation {
    print("Action failed with invalid operation")
} catch Error.failedOperation {
    print("Action failed with failed operation")
} catch {
    print("Action failed for unkown reason. Error: \(error.localizedDescription)")
}
```
Example 2:
```swift
do {
    try foo()
}  catch let error as Error {
   print("Action failed with invalid operation")
} catch {
    print("Action failed for unknown reason. Error: \(error.localizedDescription)")
}
```
Example 3:
```swift
do {
    try foo()
} catch Error.invalidOperation {
   print("Action failed with invalid operation")
}  catch let error as Error {
   print("Action failed with custom error")
} catch {
    print("Action failed for unknown reason. Error: \(error.localizedDescription)")
}
```
#### Configurable?
The rule should not be configurable. 

#### Opt-in or enabled?
This rule could be a personal preference or coding style from various individuals, so it could be kept as opt-in.

marcelofabri:
Implemented in #2045 and #2101

PR #2055, opened 2018-02-17
Untyped error checks in catch statement

dirtydanee:
Implements https://github.com/realm/SwiftLint/issues/2045

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
   to exempt certain types of variables from the rule.  
   [Rounak Jain](https://github.com/rounak)
   [#2028](https://github.com/realm/SwiftLint/issues/2028)
+* Adds `untyped-error-in-catch-rule` opt-in rule to warn against declaring errors without type check in catch statements.  
+  [Daniel Metzing](https://github.com/dirtydanee)
+  [#2045](https://github.com/realm/SwiftLint/issues/2045)
 
 #### Bug Fixes
 
--- a/Rules.md
+++ b/Rules.md
@@ -113,6 +113,7 @@
 * [Type Name](#type-name)
 * [Unneeded Break in Switch](#unneeded-break-in-switch)
 * [Unneeded Parentheses in Closure Argument](#unneeded-parentheses-in-closure-argument)
+* [Untyped error in catch rule](#untyped-error-in-catch-rule)
 * [Unused Closure Parameter](#unused-closure-parameter)
 * [Unused Enumerated](#unused-enumerated)
 * [Unused Optional Binding](#unused-optional-binding)
@@ -16373,6 +16374,96 @@ foo.map { ($0, $0) }.forEach { ↓(x, y) in }
 
 ```swift
 foo.bar { [weak self] ↓(x, y) in }
+```
+
+</details>
+
+
+
+## Untyped error in catch rule
+
+Identifier | Enabled by default | Supports autocorrection | Kind 
+--- | --- | --- | ---
+`untyped_error_in_catch` | Disabled | No | idiomatic
+
+Catch statements should not declare error variables without type casting.
+
+### Examples
+
+<details>
+<summary>Non Triggering Examples</summary>
+
+```swift
+do {
+    try foo() 
+} catch {}
+```
+
+```swift
+do {
+    try foo() 
+} catch Error.invalidOperation {
+} catch {}
+```
+
+```swift
+do {
+    try foo() 
+} catch let error as MyError {
+} catch {}
+```
+
+```swift
+do {
+    try foo() 
+} catch var error as MyError {
+} catch {}
+```
+
+</details>
+<details>
+<summary>Triggering Examples</summary>
+
+```swift
+do {
+    try foo() 
+} ↓catch var error {}
+```
+
+```swift
+do {
+    try foo() 
+} ↓catch let error {}
+```
+
+```swift
+do {
+    try foo() 
+} ↓catch let someError {}
+```
+
+```swift
+do {
+    try foo() 
+} ↓catch var someError {}
+```
+
+```swift
+do {
+    try foo() 
+} ↓catch let e {}
+```
+
+```swift
+do {
+    try foo() 
+} ↓catch(let error) {}
+```
+
+```swift
+do {
+    try foo() 
+} ↓catch (let error) {}
 ```
 
 </details>
--- a/Source/SwiftLintFramework/Models/MasterRuleList.swift
+++ b/Source/SwiftLintFramework/Models/MasterRuleList.swift
@@ -121,6 +121,7 @@ public let masterRuleList = RuleList(rules: [
     TypeNameRule.self,
     UnneededBreakInSwitchRule.self,
     UnneededParenthesesInClosureArgumentRule.self,
+    UntypedErrorInCatchRule.self,
     UnusedClosureParameterRule.self,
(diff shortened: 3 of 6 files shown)


ornithocoder:
Don't forget to add your entry to the change log  :-)

Source/SwiftLintFramework/Rules/UntypedErrorRule.swift
```diff
@@ -0,0 +1,81 @@
...
+//  SwiftLint
+//
+//  Created by Daniel.Metzing on 17/02/18.
+//  Copyright © 2018 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+public struct UntypedErrorRule: ASTRule, OptInRule, ConfigurationProviderRule {
```
marcelofabri: What do you think about changing it to `UntypedErrorInCatchRule`

marcelofabri: And then chaning the rule identifier and name as well

Source/SwiftLintFramework/Rules/UntypedErrorRule.swift
```diff
@@ -0,0 +1,81 @@
...
+        "\\{"     // Start scope character
+    )
+
+    public static let description = RuleDescription(
+        identifier: "untyped_error",
+        name: "Untyped error rule",
+        description: "Catch statements should not declare error variables without type casting",
+        kind: .lint,
+        nonTriggeringExamples: [
+            "do {\n try foo() \n} catch {\n}",
```
marcelofabri: can you format he entries better? Check how they're rendered in Rules.md. I think just improving the indentation of `try foo()` would already help.

Source/SwiftLintFramework/Rules/UntypedErrorRule.swift
```diff
@@ -0,0 +1,81 @@
...
+public struct UntypedErrorRule: ASTRule, OptInRule, ConfigurationProviderRule {
+
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    private static let regularExpression = regex(
+        "catch" + // The catch keyword
+        "(?:"   + // Start of the first non-capturing group
+        "\\s+"  + // At least one any type of whitespace character
```
marcelofabri: A whitespace is not required if parentheses are used: `catch(let error) {`. 

Also, the rule currently wouldn't catch when they're used even with spaces.

Source/SwiftLintFramework/Rules/UntypedErrorRule.swift
```diff
@@ -0,0 +1,81 @@
...
+                         kind: StatementKind,
+                         dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+
+        guard kind == .brace,
+            let offset = dictionary.offset else {
+                return []
+        }
+
+        return file.lines
+            .lazy
```
marcelofabri: Does this really improve performance? I think `lines` will be calculated anyways.

Source/SwiftLintFramework/Rules/UntypedErrorRule.swift
```diff
@@ -0,0 +1,81 @@
...
+            .filter { $0.byteRange.contains(offset) }
+            .reduce([]) { _, line in
```
marcelofabri: wouldn't it be more semantic to use `flatMap` here?

Source/SwiftLintFramework/Rules/UntypedErrorRule.swift
```diff
@@ -0,0 +1,81 @@
...
+        "(?:"   + // Start of the second non-capturing group
+        "\\s*"  + // Zero or unlimited any whitespace character
+        ")"     + // End of the second non-capturing group
+        "\\{"     // Start scope character
+    )
+
+    public static let description = RuleDescription(
+        identifier: "untyped_error",
+        name: "Untyped error rule",
+        description: "Catch statements should not declare error variables without type casting",
```
marcelofabri: Can you add a `.` in the end?

Source/SwiftLintFramework/Rules/UntypedErrorRule.swift
```diff
@@ -0,0 +1,81 @@
...
+        kind: .lint,
```
marcelofabri: can you change to `.idiomatic`? `.lint` is more for validating things that could indicate a programmer's error.

Source/SwiftLintFramework/Rules/UntypedErrorRule.swift
```diff
@@ -0,0 +1,81 @@
...
+        nonTriggeringExamples: [
+            "do {\n try foo() \n} catch {\n}",
+            "do {\n try foo() \n} catch Error.invalidOperation {\n} catch {\n}",
+            "do {\n try foo() \n} catch let error as? MyError {\n} catch {\n}",
```
marcelofabri: this is not a major issue, but I think `catch let error as MyError` should be used. I got a warning while compiling with the optional cast:

```
 warning: cast from '_' to unrelated type 'MyError' always fails
```

Source/SwiftLintFramework/Rules/UntypedErrorRule.swift
```diff
@@ -0,0 +1,81 @@
...
+                let range = NSRange(location: 0, length: line.content.bridge().length)
+                return type(of: self).regularExpression.matches(in: line.content, options: [], range: range)
+                    .map { match -> StyleViolation in
+                        let characterOffset = line.range.lowerBound + match.range.lowerBound
+                        let location = Location(file: file,
+                                                characterOffset: characterOffset)
+                        return StyleViolation(ruleDescription: type(of: self).description,
+                                              severity: configuration.severity,
+                                              location: location,
+                                              reason: configuration.consoleDescription)
```
marcelofabri: Is there a reason why you're iterating through lines?

I have the impression that it'd be faster to make this a regular rule (not `ASTRule`) and just match the regex on the whole file, validating some syntax kinds (e.g. that `catch` is a `keyword`, as well as `let`/`var`).

dirtydanee: @marcelofabri Thanks for the review! This suggestion was a really spot on. I have decided to use ASTRule to be able to solve the issue with the commented out code. However the syntax kind validation works great, simplified the code by tons and should be faster too.

Source/SwiftLintFramework/Rules/UntypedErrorRule.swift
dirtydanee: Updated the naming as you suggested, its for sure more clear what does it.

Source/SwiftLintFramework/Rules/UntypedErrorRule.swift
dirtydanee: Fixed the indentation and removed the `\n` from the `catch` statements scope.

Source/SwiftLintFramework/Rules/UntypedErrorRule.swift
dirtydanee: Updated the pattern to capture the errors with parentheses with and without spaces.

dirtydanee:
@marcelofabri thoughts on the update?

marcelofabri:
@dirtydanee Merged in #2101 🎉

Issue #1466, opened 2017-04-25
fatal error: Loading sourcekitd.framework/Versions/A/sourcekitd failed


PraveenSingh107:
I started facing this fatal error while running the lint command. 

fatal error: Loading sourcekitd.framework/Versions/A/sourcekitd failed: file /tmp/swiftlint-20160824-65136-lo3uim/Carthage/Checkouts/SourceKitten/Source/SourceKittenFramework/library_wrapper.swift, line 59
.git/hooks/pre-commit: line 7: 32011 Illegal instruction: 4  ${SWIFT_LINT} lint --strict --quiet --path ${filename}

Any suggestions why suddenly this fatal error is stopping swiftlint?

jpsim:
Have you installed Xcode? What's the result of `xcode-select -p`?

PraveenSingh107:
Yeah, i've installed xcode. swiftlint was working fine but suddenly it started throwing this error. 
Output of xcode-select -p : /Library/Developer/CommandLineTools

jpsim:
The command line tools located at `/Library/Developer/CommandLineTools` do not contain a Swift toolchain.

Where have you installed Xcode? If it's installed under `/Applications/Xcode.app`, please run:

```
sudo xcode-select -s /Applications/Xcode.app/Contents/Developer
```

If it's installed elsewhere, please replace the `/Applications/Xcode.app` path in the command above to the location where you've installed Xcode.

PraveenSingh107:
Thanks. Issue got resolved

ulkoart:
> The command line tools located at `/Library/Developer/CommandLineTools` do not contain a Swift toolchain.
> 
> Where have you installed Xcode? If it's installed under `/Applications/Xcode.app`, please run:
> 
> ```
> sudo xcode-select -s /Applications/Xcode.app/Contents/Developer
> ```
> 
> If it's installed elsewhere, please replace the `/Applications/Xcode.app` path in the command above to the location where you've installed Xcode.

Thank you so much! advice helped me.

koushikkarmakar:
> The command line tools located at `/Library/Developer/CommandLineTools` do not contain a Swift toolchain.
> 
> Where have you installed Xcode? If it's installed under `/Applications/Xcode.app`, please run:
> 
> ```
> sudo xcode-select -s /Applications/Xcode.app/Contents/Developer
> ```
> 
[...]

PR #2056, opened 2018-02-17
Adds force-exlucde flag to ling/autocorrect commands

### LINKED ISSUE #2051: Feature request: option to force exclusion
### New Issue Checklist

### Feature Request

Hello! I like SwiftLint a lot, so much so that I built a tool called [danger-ruby-swiftlint](https://github.com/ashfurrow/danger-ruby-swiftlint/) to run SwiftLint on CI with [Danger](https://danger.systems/ruby). The way the tool works is it runs SwiftLint against each file that's been changed in a pull request, and reports any violations to GitHub in a comment. However, we've run into a problem, reported [here](https://github.com/ashfurrow/danger-ruby-swiftlint/issues/87).

Here's the scenario:

- We run `swiftlint` individually against each changed file.
- This includes all files, even those that are excluded in `.swiftlint.yml`.
- Those excluded files are still having their violations reported back to GitHub.

To be clear: I don't think this is a bug in SwiftLint. It's just a sort of edge case that I've hit. I had a very similar problem with my very similar [danger-rubocop](https://github.com/ashfurrow/danger-rubocop) tool, but was able to get around it using Rubocop's `--force-exclusion` flag, detailed [here](https://github.com/bbatsov/rubocop/issues/893). This flag, off by default, forces Rubocop to adhere to the exclusions in its config file, even if Rubocop is explicitly invoked with that file.

Could we add something equivalent to Rubocop's `--force-exclusions` flag to SwiftLint?

I totally understand that this probably isn't a high priority. What I'm really looking for here is a 👍 or 👎 on the feature request itself. If you're amenable to it, just provide a high-level description of how you'd like users to specify they want to force exclusions, and I'll send you a pull request.

Thanks for considering it!

ashfurrow:
This adds a `--force-exclude` option to the `lint` and `autocorrect` commands which will force SwiftLint to exclude a file if it matches the `excluded` config options, even if that file is explicitly passed with the `--path` option.

~I could not test locally because I got the following exception:~

<details>
<summary>Fixed this, ran into #1466.</summary>

```sh
➜  Moya git:(master) ✗ swiftlint
Fatal error: Loading sourcekitd.framework/Versions/A/sourcekitd failed: file /Users/ashfurrow/SwiftLint/.build/checkouts/SourceKitten.git-6484296299232452758/Source/SourceKittenFramework/library_wrapper.swift, line 61
[1]    71720 illegal hardware instruction  swiftlint
```

</details>

Tested this locally:

```sh
➜  Moya git:(master) ✗ swiftlint lint --path Sources/Moya/AnyEncodable.swift --force-exclude --config .swiftlint.yml
Linting Swift files at path Sources/Moya/AnyEncodable.swift
No lintable files found at path 'Sources/Moya/AnyEncodable.swift'
```

It fails because it "can't find" any files, but I think that makes sense. 

Fixes #2051.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 * Adds `--force-exclude` option to `lint` and `autocorrect` commands, which will
   force SwiftLint to exclude files specified in the config `excluded` even if
-  they are explicitly specified with `--path`.
+  they are explicitly specified with `--path`.  
   [Ash Furrow](https://github.com/ashfurrow)
   [#2051](https://github.com/realm/SwiftLint/issues/2051)
 * Add a new `excluded` config parameter to the `explicit_type_interface` rule
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,11 @@
 
 #### Enhancements
 
-* None.
+* Adds `--force-exclude` option to `lint` and `autocorrect` commands, which will
+  force SwiftLint to exclude files specified in the config `excluded` even if
+  they are explicitly specified with `--path`.
+  [Ash Furrow](https://github.com/ashfurrow)
+  [#2051](https://github.com/realm/SwiftLint/issues/2051)
 
 #### Bug Fixes
 
--- a/Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
@@ -10,14 +10,14 @@ import Foundation
 import SourceKittenFramework
 
 extension Configuration {
-    public func lintableFiles(inPath path: String) -> [File] {
-        return lintablePaths(inPath: path).flatMap(File.init(pathDeferringReading:))
+    public func lintableFiles(inPath path: String, forceExclude: Bool) -> [File] {
+        return lintablePaths(inPath: path, forceExclude: forceExclude).flatMap(File.init(pathDeferringReading:))
     }
 
-    internal func lintablePaths(inPath path: String,
+    internal func lintablePaths(inPath path: String, forceExclude: Bool,
                                 fileManager: LintableFileManager = FileManager.default) -> [String] {
-        // If path is a file, skip filtering with excluded/included paths
-        if path.isFile {
+        // If path is a file and we're not forcing excludes, skip filtering with excluded/included paths
+        if path.isFile && !forceExclude {
             return [path]
         }
         let pathsForPath = included.isEmpty ? fileManager.filesToLint(inPath: path, rootDirectory: nil) : []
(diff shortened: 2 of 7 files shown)


ashfurrow:
Okay, got it to run but it's not behaving as I'd expect. Will mark the PR as a WIP.

ashfurrow:
Coooooool, had to `brew uninstall swiftlint` first. Ready for review 👍

marcelofabri:
What should happen if you pass `force-exclude` and `--path ADirectory`?

ashfurrow:
That's a good question. I would expect that the other files in the directory would be linted, but any files in `excluded` would not be. In practice, I just tried this and the file _was_ included. Hmm. I'll do some debugging and let you know how things go.

marcelofabri:
@ashfurrow you probably need to filter in https://github.com/realm/SwiftLint/pull/2056/files#diff-27e9905d3d17e50e5263bb7cae0fd09aR24 too.

I'd also make sure that we have tests for what happens if we have `force-exclude` and:

* `--path aFileThatIsInExcluded`
* `--path aFileThatIsNotInExcluded`
* `--path aDirectoryThatIsInExcluded`
* `--path aDirectoryThatIsNotInExcludedButHasChildrenThatAre`

ashfurrow:
Okay so it turns out I didn't quite understand the semantics of how SwiftLint resolves config files. I re-tested your earlier question:

> What should happen if you pass `force-exclude` and `--path ADirectory`?

This does work as I'd expect: SwiftLint picks up the other files but not the ones in `excluded`. I tested this with Moya:

```sh
swiftlint lint --path .  --force-exclude --config .swiftlint.yml
```

And this lints everything but does exclude the files in `excluded`. So it works as-expected. I added the tests you specified too, let me know if there's anything else I can do to wrap this up. And thanks again!

ashfurrow:
Ack! Just saw the failures, will fix.

CHANGELOG.md
```diff
@@ -6,7 +6,11 @@
 
 #### Enhancements
 
-* None.
+* Adds `--force-exclude` option to `lint` and `autocorrect` commands, which will
+  force SwiftLint to exclude files specified in the config `excluded` even if
+  they are explicitly specified with `--path`.
```
marcelofabri: Can you add two trailing whitespaces after `.`?

marcelofabri:
@ashfurrow Sorry for the delay. This looks good! ✅ 

Just added a small comment, but other than that and fixing the conflict, this is ready to be merged.

CHANGELOG.md
ashfurrow: Sure can. Would you like a separate PR that checks for these missing spaces using Danger? Could save you some time chasing after contributors 😄

ashfurrow:
Any update on this? I don’t understand the nature of the CI failure.

marcelofabri:
@ashfurrow I just manually merged this. I had to rebase to get rid of the merge commit and also I had to run Sourcery to make sure the new tests were being run on Linux (https://github.com/realm/SwiftLint/commit/f231e6d24b7aa1f279b1e9746efe1078cccdacac).

Thanks and sorry for the delay!

ashfurrow:
Cool, thanks a lot!

PR #2057, opened 2018-02-20
Fix failed linting of ReporterTests.swift

jaquish:


marcelofabri:
Huh, surprisingly this wasn't failing in CI or for me locally.

Issue #2058, opened 2018-02-21
Improvement: Merge individual rule's configuration when merging config yml files


gretzki:
When merging Swiftlint configuration files, the set of rules are merged according to the rulesMode flag, but not the configuration of the rules that are carried over. (Since #1687)

**Just for clarification, let me define the following terms:**
Having two configurations for one rule, one at the current directory - subsequently called `nested configuration` - and one at a higher level - subsequently called `parent configuration`. When merging the rules configuration for the current directory, the result is subsequently called the `result configuration`.

**Assume the following example:**
The parent configuration specifies:

> cyclomatic_complexity:
>   warning: 13
>   error: 14
>   ignores_case_statements: true

and the nested configuration specifies:

> cyclomatic_complexity:
>   warning: 5

 **My assumption:**
The result configuration should have:

> cyclomatic_complexity:
>   warning: 5
>   error: 14
>   ignores_case_statements: true

**Current implementation**
The result configuration depends if it is an optIn rule. If it was, the result would specify the  warning threshold of the nested configuration, but default values for the other settings:

> cyclomatic_complexity:
>   warning: 5
>   error: 20
>   ignores_case_statements: false

Since it is not an optIn rule, the parent configuration is taken over and the result configuration has

> cyclomatic_complexity:
>   warning: 13
>   error: 14
>   ignores_case_statements: true 

**My proposition**
Both cases are counterintuitive (at least for me). Therefore, I propose the following:
The result configuration should always contain the rule's configuration of the parent, overriden by what has been set by the nested configuration explicitly.

If this is the expected behaviour one can agree on,  I'm more than happy to publish a pull request for the proposed behaviour. Btw. kudos for the excellent work on #1687.

gretzki:
I've opened a feature branch as a proof of concept and to make the changes more easily traceable: #2059 

In summary, the changes are:
1. Rules are initially configured via init(configuration:) in Rule.swift when parsing yml files. There, the set of explicitly set values (compared to default values) can be determined and stored.
WiP branch still uses another place to store the explicitly set values.

2. Configuration+Merging.swift is extended to merge rule's configurations too, when merging the set of rules considering disabled sets and optIn sets. To merge a rule's configuration, a method called `overrideConfiguration(configuration:)` is invoked.
WiP branch only extends the ruleMode .default.

3. Rule is extended with the function `overrideConfiguration(withRule secondRule: Rule) -> Rule`.
It works as follows:
- lookup own RuleConfiguration property A
[...]

PR #2059, opened 2018-02-22
[WIP] proof of concept for merging individual rule's configuration

gretzki:
This PR serves as proof of concept to demonstrate the merging of individual rule's configurations, as proposed here: #2058 .
It is marked work in progress since the discussion of the proposition is still open.

There's a new unit test to check the proposed merging policy. The current implementation fulfills this test.
Nevertheless, there are still some open todos:

- Adapt implementation
  - still depending on extending all ConfigurationProviderRule implementations, although it is not necessary (experimentally applied to CyclomaticComplexityConfiguration.swift)

- Check Reflection as a dependency
  - The current implementation uses reflection to retrieve the `configuration` property of a ConfigurationProviderRule. Since Swift4 does not allow to check conformance with protocols using associated types, that's currently a working solution.

- Fix circleCI checks
- PR on Reflection (if it stays) 
  - target’s “Require Only App-Extension-Safe API” build setting is set to false. But it compiles if activated too

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,6 @@
 [submodule "Carthage/Checkouts/Yams"]
 	path = Carthage/Checkouts/Yams
 	url = https://github.com/jpsim/Yams.git
+[submodule "Carthage/Checkouts/Reflection"]
+	path = Carthage/Checkouts/Reflection
+	url = https://github.com/Zewo/Reflection.git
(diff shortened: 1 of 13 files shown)


Dschee:
@gretzki This PR sounds like a good idea though I'm not sure why we really need the Reflection dependency. Can't we merge the configurations on the YAML level and skip the dependency completely?

PR #2060, opened 2018-02-23
Add config to exempt local vars from ExplicitTypeInterfaceRule

### LINKED ISSUE #2028: [Swift 4.1] Add configuration option to fire `explicit_type_interface` only for properties rather than local variables
### New Issue Checklist

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.24.2
* Installation method used (Homebrew, CocoaPods, building from source, etc)? portable

* Which Xcode version are you using (check `xcode-select -p`)? 9.3

```swift
// This triggers a violation:
func a() {
    let foo = UIView()
}
```

With reference to issue #2027

rounak:
This PR adds a new `excluded` config to this existing rule, to exempt certain kinds of variables from having explicit types. Values can be `local`, `static`, `class` & `instance`.

Would love to add test coverage for this, but didn't know if I should write a new test in `RuleTests` or do it some other way.


CHANGELOG.md
```diff
@@ -6,7 +6,10 @@
 
 #### Enhancements
 
-* None.
+* Add a new `excluded` config parameter to the `explicit_type_interface` rule
+  to exempt certain types of variables from the rule.
```
marcelofabri: you should leave two trailing whitespaces (currently just one)

Source/SwiftLintFramework/Rules/RuleConfigurations/ExplicitTypeInterfaceConfiguration.swift
```diff
@@ -0,0 +1,58 @@
...
+        default:
+            return nil
+        }
+    }
+}
+
+public struct ExplicitTypeInterfaceConfiguration: RuleConfiguration, Equatable {
+    public var severityConfiguration = SeverityConfiguration(.warning)
+
+    var allowedKinds: Set<SwiftDeclarationKind> = [.varInstance,
```
marcelofabri: this should be `public`

Source/SwiftLintFramework/Rules/RuleConfigurations/ExplicitTypeInterfaceConfiguration.swift
```diff
@@ -0,0 +1,58 @@
...
+                                                   .varLocal,
+                                                   .varStatic,
+                                                   .varClass]
+
+    public var consoleDescription: String {
+        return severityConfiguration.consoleDescription + ", allowed kinds: \(allowedKinds)"
```
marcelofabri: can we print the allowedKinds in the same short style that is expect to be in the configuration file?

marcelofabri:
> Would love to add test coverage for this, but didn't know if I should write a new test in RuleTests or do it some other way.

You can create a new file `ExplicitTypeInterfaceConfigurationTests` (take a look at `CyclomaticComplexityConfigurationTests` for example) and another one to test the rule itself (`ExplicitTypeInterfaceRuleTests`).

When you create `ExplicitTypeInterfaceRuleTests`, you should move the existing test in `RulesTests` to the new file, along with new tests that test different configurations. You can take a look at existing tests to see how it's usually done (`ColonRuleTests` would be an example).

Source/SwiftLintFramework/Rules/RuleConfigurations/ExplicitTypeInterfaceConfiguration.swift
```diff
@@ -27,27 +27,37 @@ private extension SwiftDeclarationKind {
...
-    var allowedKinds: Set<SwiftDeclarationKind> = [.varInstance,
-                                                   .varLocal,
-                                                   .varStatic,
-                                                   .varClass]
+    public var allowedKinds = ExplicitTypeInterfaceConfiguration.variableKinds
 
     public var consoleDescription: String {
-        return severityConfiguration.consoleDescription + ", allowed kinds: \(allowedKinds)"
+        return severityConfiguration.consoleDescription +
+        ", excluded: [\(ExplicitTypeInterfaceConfiguration.variableKinds.subtracting(allowedKinds))]"
```
marcelofabri: My comment here was more about how the types will be spelled out.

By just printing `SwiftDeclarationKind`, you'll get something like `source.lang.swift.decl.var.class`, but when configuring the rule you're expected to use just `class`.

rounak:
Thanks for the review @marcelofabri.

Added two new test files as you said, one for config and one for the rule itself. The only issue is since this local variable stuff depends on SourceKit, we can't add concrete test cases (it warns for local vars) until everyone's running Xcode 9.3.

As for the consoleDescription, I didn't quite understand:

> the same short style that is expect to be in the configuration file

I added square brackets around the set, is that what you meant?

Source/SwiftLintFramework/Rules/RuleConfigurations/ExplicitTypeInterfaceConfiguration.swift
rounak: Ah I see, let me do that

marcelofabri:
> The only issue is since this local variable stuff depends on SourceKit, we can't add concrete test cases (it warns for local vars) until everyone's running Xcode 9.3.

Can you add a test excluding another kind (e.g. class variables)?

rounak:
@marcelofabri Done, let me know if it looks okay now.

Tests/SwiftLintFrameworkTests/ExplicitTypeInterfaceTests.swift
```diff
@@ -0,0 +1,47 @@
...
+//  SwiftLint
+//
+//  Created by Rounak Jain on 2/24/18.
+//  Copyright © 2018 Realm. All rights reserved.
+//
+
+import SwiftLintFramework
+import XCTest
+
+class ExplicitTypeInterfaceTests: XCTestCase {
```
marcelofabri: Can you rename this class (and file) to `ExplicitTypeInterfaceRuleTests`?

marcelofabri:
@rounak This looks good! I have one minor nitpick that I added as a comment.

Can you also rebase against master now that I merged #2068? That should fix the CI failure.

Thanks!

PR #2064, opened 2018-02-24
Implemented empty_string rule

iDevid:
Rule description:
Prefer checking `isEmpty` over comparing `string` to \"\".


CHANGELOG.md
```diff
@@ -22,6 +22,9 @@
 
 #### Enhancements
 
+* Adds `empty_string` opt-in rule.
```
ornithocoder: HI @iDevid! Take a look [here](https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes) for the format used on the changelog file. For instance, it must contain two trailing spaces on the last line describing the change.

Tests/SwiftLintFrameworkTests/EmptyStringTests.swift
```diff
@@ -0,0 +1,30 @@
+//
```
ornithocoder: You don't need to include this test file. The two cases listed here are the same from `triggeringExamples`. SwiftLint builds tests for snippets included on `nonTriggeringExamples` and `triggeringExamples` and automatically runs them to make sure the rule does what it says it does.

CHANGELOG.md
```diff
@@ -22,6 +22,9 @@
 
 #### Enhancements
 
+* Adds `empty_string` opt-in rule.  
+[Davide Sibilio](https://github.com/idevid)
```
marcelofabri: please add two spaces before your name (check other entries)

Source/SwiftLintFramework/Rules/EmptyStringRule.swift
```diff
@@ -0,0 +1,40 @@
...
+        triggeringExamples: [
+            "myString ↓== \"\"",
+            "myString ↓!= \"\""
+        ]
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        let pattern = "\\b\\s*(==|!=)\\s*\"\""
+        let excludingKinds = SyntaxKind.commentKinds
+        return file.match(pattern: pattern, excludingSyntaxKinds: excludingKinds).map {
```
marcelofabri: you should use the `match` variation that takes the exact syntax kinds you want to match:

```
return file.match(pattern: pattern, with: [.string])
```

Source/SwiftLintFramework/Rules/EmptyStringRule.swift
```diff
@@ -0,0 +1,40 @@
...
+
+public struct EmptyStringRule: ConfigurationProviderRule, OptInRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "empty_string",
+        name: "Empty String",
+        description: "Prefer checking `isEmpty` over comparing `string` to \"\".",
```
marcelofabri: what do you think about changing this to `"Prefer checking `isEmpty` over comparing `string` to an empty string literal`?"

iDevid: I think that is better

Source/SwiftLintFramework/Rules/EmptyStringRule.swift
iDevid: fixed! Thanks!

iDevid:
There are some problems with osscheck?

marcelofabri:
> Are there some problems with osscheck?

Apparently, there's an issue with SSL on Circle CI: the `cocoapods` job is failing with the same reason and it's not only this PR that is failing. I'll try to investigate tomorrow.

marcelofabri:
Also related: https://github.com/blog/2507-weak-cryptographic-standards-removed

marcelofabri:
I've merged #2068 which should fix the CI failure. Can you please rebase this PR?

CHANGELOG.md
```diff
@@ -22,6 +22,9 @@
 
 #### Enhancements
 
+* Adds `empty_string` opt-in rule.
```
ornithocoder: Please add your entry to the **Master** section of the change log. Version 0.25.0's already out.

marcelofabri:
@iDevid Thanks!

Issue #1901, opened 2017-10-15
Colon rule - apply_to_dictionaries: false is not respected


DG0BAB:
### New Issue Checklist

- ✅ Updated SwiftLint to the latest version
- ✅ I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Bug Report

Since version 0.23.0 the "apply_to_dictionaries: false" for the colon rule  isn't respected anymore. In our Project we specify dictionaries like `[key : value]`. The bug only appears if value is of type Any. This worked fine until version 0.23.0 with "apply_to_dictionaries: false"

### Environment

* SwiftLint version 0.23.1
* Installation method SwiftLint.pkg

```yml
line_length:
  warning: 180
  ignores_comments: true

colon:
  apply_to_dictionaries: false

identifier_name:
  min_length:
    error: 3
  max_length:
    warning: 60
  excluded:     # excluded via string array
    - qa
    - QA
    - id
    - i
    - iD
  allowed_symbols:
    - _

vertical_whitespace:
    max_empty_lines: 5

function_body_length:
    warning: 80
    error: 150

disabled_rules:
    todo

excluded: # paths to ignore during linting. overridden by `included`.
  - Carthage
```

* Using Xcode version 9.1 beta 2 (Build (9B46)

```bash
echo "[String : Any]" | swiftlint lint --no-cache --use-stdin --enable-all-rules
Loading configuration from '.swiftlint.yml'
<nopath>:1:2: warning: Colon Violation: Colons should be next to the identifier when specifying a type and next to the key in dictionary literals. (colon)
```

```swift
// This triggers a violation:
let foo: [String : Any]

// This does not trigger a violation:
let foo: [String : Int]
```

```bash
echo "[String : Int]" | swiftlint lint --no-cache --use-stdin --enable-all-rules
Loading configuration from '.swiftlint.yml'
Done linting! Found 0 violations, 0 serious in 1 file.
```

serges147:
+1

marcelofabri:
`apply_to_dictionaries` is only verified for dictionary literals. In terms of language, literals and dictionary type declarations are different things. Feel free to submit a PR to improve this, but my initial guess is that it might be tricky to get it right.

DG0BAB:
Yeah, you are right. Only dictionary literals are ignored and not dictionary declarations. But that changed with version 0.23.x. In former versions even dictionary declarations were ignored. So currently we are still using version 0.22 as a workaround.

[...]

PR #2067, opened 2018-02-25
[WIP] Ignore colon rule on dictionary type declaration when apply_to_dictionaries configuration flag is false

guidomb:
As described in #1901 colon rule ignores dictionary literals when `apply_to_dictionaries`
is set to `false` but fails to ignore dictionary type declarations. This PR attempts to fix that issue.

This is marked as WIP because some tests still fail and I consider my fix a little bit rough. I'd like to get some help to better understand why some tests are failing and what would be the proper way to fix them. If the general logic is OK then I can clean the code and extract it into a separate function.

Also I'd like to check with a maintainer if the general direction of my fix is OK. This is my first dive in SwiftLint's code base and want to make sure that I'm not missing anything.

As far as I can tell, the colon rule is implemented by matching a regex against each file to check statements where the `:` character is used and there are white spaces to the left followed by an identifier. Both identifier on the left and on the right of the `:` character are captured on a group and then SourceKitten is used to see what type of token the captured identifier represents.

My first idea was to modify the regex expression but I think it would make the expression way more complicated. So the patch I'm proposing checks if the `apply_to_dictionaries` configuration flag is set to `false` and then it filters out matching violations (raised by the regex) if both identifiers to the left and the right of the `:` characters are type identifiers. The left type identifier should be preceded by a `[` (ignoring white spaces) and the right type identifier should be proceeded with a `]` (also ignoring white spaces). This is what I consider to be a valid dictionary type declaration. Is this definition of dictionary type declaration correct? Am I missing any edge cases?

I realize that this may have the performance overhead of capturing false positive by the regexp which would latter be discarded but I think is a reasonable trade off to be made for readability sake. But I guess some performance metrics should be used to decide whether my assumption is correct.

### DIFF
--- a/Source/SwiftLintFramework/Rules/ColonRule+Type.swift
+++ b/Source/SwiftLintFramework/Rules/ColonRule+Type.swift
@@ -40,6 +40,32 @@ internal extension ColonRule {
                 return false
             }
 
+            if !configuration.applyToDictionaries {
+                let (leftColonSideToken, rightColonSideToken) = (syntaxTokens[0], syntaxTokens[1])
+                let bothTokensAreTypeIdentifiers: Bool
+                switch (syntaxKinds[0], syntaxKinds[1]) {
+                case (.typeidentifier, .typeidentifier):
+                    bothTokensAreTypeIdentifiers = true
+                case (.typeidentifier, .keyword):
+                    bothTokensAreTypeIdentifiers = file.isTypeLike(token: rightColonSideToken)
+                default:
+                    bothTokensAreTypeIdentifiers = false
+                }
+                
+                if bothTokensAreTypeIdentifiers {
+                    let openBracketCandidate = nsstring.firstNonWhitespaceCharacter(beforeCharacterAt: leftColonSideToken.offset)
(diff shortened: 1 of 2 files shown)


mattbauch:
> The left type identifier should be preceded by a [ (ignoring white spaces) and the right type identifier should be proceeded with a ] (also ignoring white spaces). This is what I consider to be a valid dictionary type declaration. Is this definition of dictionary type declaration correct? Am I missing any edge cases?

My use case would be for arrays as value.   
But there are more where the right type isn't followed by a `]`

    let a: [String : String?] = [:]
    let b: [String : [String]] = [:]
    let c: [String : [String : String]] = [:]
    let d: [String : (String, String)] = [:]

And with Swift 4.1 you can even have left types that aren't preceded by a `[`

    extension Array: Hashable where Element: Hashable {
        public var hashValue: Int { return 0 }
    }
    let e: [[String] : String] = [:]

    extension Dictionary: Hashable where Key: Hashable, Value: Equatable {
        public var hashValue: Int { return 0 }
    }
    let f: [[String : String] : String] = [:]

dylanmaryk:
Would love to see this finished and merged! Has any progress been made on this?

guidomb:
@dylanmaryk no progress so far, couldn't find the time to move this forward.

PR #2076, opened 2018-03-01
#2075 - New rule - Empty (whitespace) lines in function bodies 

driver733:
As per #2075 
I have implemented a new rule that prohibits empty (whitespace) lines in function bodies.


Tests/SwiftLintFrameworkTests/FunctionBodyWhitespaceLinesRuleTests.swift
```diff
@@ -0,0 +1,48 @@
+//
+//  FunctionBodyLengthRule.swift
```
Dschee: Copy and paste leftover.

Dschee:
@driver733 Could you please add triggering and non-triggering examples and rebase this?

Please note that I have never seen this style of coding and can't imagine this to be considered a "good practice" (and definitely not a _best_ practice) – but that's my personal experience so far and should not prevent this rule from being added as an optional rule. Unless there's rules we don't merge if there's no real "community acceptance"? @jpsim @marcelofabri

Issue #2077, opened 2018-03-02
Rule request: Comment lines in function bodies


driver733:
## New Issue Checklist

- [x]  Updated SwiftLint to the latest version
- [x]  I searched for existing GitHub issues

## Description

In addition, the code, in my opinion, should not have in-line comments. If any kind of explanations is needed, it has to be put in the method or class documentation. Furthermore, if your code requires comments, then it might be necessary to refactor it, so that it would be self-explainatory.

## What would trigger?

```
internal func exceedsLineCountExcludingCommentsAndWhitespace(_ start: Int, _ end: Int, _ limit: Int) -> (Bool, Int) {
        guard end - start > limit else {
            return (false, end - start)
        }
        // I will put this here instead of putting in the method doc, where it belongs.
        let count = end - start - numberOfCommentAndWhitespaceOnlyLines(startLine: start, endLine: end)
        return (count > limit, count)
    }
```
## What would not trigger?

```
internal func exceedsLineCountExcludingCommentsAndWhitespace(_ start: Int, _ end: Int, _ limit: Int) -> (Bool, Int) {
        guard end - start > limit else {
            return (false, end - start)
        }
        let count = end - start - numberOfCommentAndWhitespaceOnlyLines(startLine: start, endLine: end)
        return (count > limit, count)
    }
```
## Configurable?

The rule should not be configurable.

## Opt-in or enabled?

This rule could be a personal preference or coding style from various individuals, so it could be kept as opt-in.

fredpi:
I'm strictly against this (even when done as an opt-in rule). While I agree that inline method comments often can be a doc-comment on top of the method declaration, I don't believe disallowing comments entirely within method is somehow encouragable.

While comments are often used to document code that is poorly written and that would be self-explanatory if implemented properly (one common example being the (mis)use of a `UIView`'s `tag` attribute), there are other use cases for comments:

- Grouping blocks. Probably the most important use of comments in my opinion. Even within a single method, there are cases where you wouldn't want to split things up into other singular methods, but instead group lines together and write a comment above the group which explains the context (consider it some kind of headline).
[...]

PR #2078, opened 2018-03-02
#2077 - New rule for comment lines in function bodies.

driver733:
As per #2077 
I have implemented a new rule that prohibits comment lines in function bodies.


Dschee:
@driver733 Please see my comment [here](https://github.com/realm/SwiftLint/pull/2076#issuecomment-466957957) as the same applies for this PR.

Issue #2079, opened 2018-03-02
Rule Request: Prohibit subclassing


driver733:
## New Issue Checklist

- [x]  Updated SwiftLint to the latest version
- [x]  I searched for existing GitHub issues

## Description

> Implementation inheritance was created as a mechanism for code reuse, and it doesn't fit into OOP at all. Yes, it may look convenient in the beginning, but it is absolutely wrong in terms of object thinking. Just like getters and setters, implementation inheritance turns objects into containers with data and procedures. Of course, it's convenient to copy some of those data and procedures to a new object in order to avoid code duplication. But this is not what objects are about.

Read [this](http://www.yegor256.com/2016/09/13/inheritance-is-procedural.html) for more info.

## What would trigger?

```
class MyType: SuperType {
                public init() { super() }" 
            }
```
## What would not trigger?

```
class MyType: SuperType {
                public init() {  }" 
            }
```
## Configurable?

The rule should not be configurable.

## Opt-in or enabled?

This rule could be a personal preference or coding style from various individuals, so it could be kept as opt-in.

marcelofabri:
I don't get it. You should call `super` on inits in general. The "problem" is not calling `super.init()`, but declaring inheritance.

driver733:
@marcelofabri If a class is not a subclass, then it does not need to call super.
If a class calls `super`, we assume, by this rule, that it is a subclass and highlight it.

marcelofabri:
But this is the wrong assumption. You can have subclasses that don't call super (if they don't override any methods) but you can't have classes that call super and are not subclasses (the compiler doesn't let you do that).

driver733:
@marcelofabri 
`You can have subclasses that don't call super (if they don't override any methods) `
If a class is a subclass it will have `super.init()` as a first statement is at least one of the constructors and that rule will detect that.

`you can't have classes that call super and are not subclasses (the compiler doesn't let you do that).`
Although I agree, I do not understand how this statement is applicable to this rule.

Again, if a class is not a subclass, then it will not have `super` anywhere. If a class is a subclass, it will have `super` at least somewhere, for instance, in one of the constructors.

marcelofabri:
> If a class is a subclass it will have super.init() as a first statement is at least one of the constructors and that rule will detect that.

You don't need to override a constructor in a subclass.

driver733:
@marcelofabri If you introduce a new properties or create a new constructor, you will inevitably end up having a call to the superclass in one of the constructors.
[...]

PR #2081, opened 2018-03-02
#2079 - Implemented the new sublclass rule.

driver733:
As per #2079
I have implemented a new rule that prohibits subclassing.

### DIFF
--- a/.idea/codeStyles/codeStyleConfig.xml
+++ b/.idea/codeStyles/codeStyleConfig.xml
@@ -1,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
+  <state>
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Default" />
+  </state>
+</component>
--- a/Source/SwiftLintFramework/Rules/SubClassRule.swift
+++ b/Source/SwiftLintFramework/Rules/SubClassRule.swift
@@ -1,0 +1,47 @@
+//
+//  SubClassRule.swift
+//  SwiftLint
+//
+//  Created by Mikhail Yakushin on 03/02/18.
+//  Copyright © 2018 Realm. All rights reserved.
+//
+
+import SourceKittenFramework
+
+public struct SubClassRule: ASTRule, ConfigurationProviderRule, OptInRule {
+    public var configuration = SeverityLevelsConfiguration(warning: 0, error: 0)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "subclass",
+        name: "Subclass",
+        description: "Subclassing is prohibited.",
+        kind: .style
+    )
+
+    public func validate(file: File, kind: SwiftDeclarationKind,
+                         dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard SwiftDeclarationKind.functionKinds.contains(kind),
+              let offset = dictionary.offset,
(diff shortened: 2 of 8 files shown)


.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
```
marcelofabri: this shouldn't be commited

Source/SwiftLintFramework/Rules/SubClassRule.swift
```diff
@@ -0,0 +1,47 @@
...
+public struct SubClassRule: ASTRule, ConfigurationProviderRule, OptInRule {
+    public var configuration = SeverityLevelsConfiguration(warning: 0, error: 0)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "subclass",
+        name: "Subclass",
+        description: "Subclassing is prohibited.",
+        kind: .style
```
marcelofabri: you should provide `triggeringExamples` and `nonTriggeringExamples`

marcelofabri: this would be probably `.idiomatic` kind and not `.style`

Source/SwiftLintFramework/Rules/SubClassRule.swift
```diff
@@ -0,0 +1,47 @@
...
+//  SwiftLint
+//
+//  Created by Mikhail Yakushin on 03/02/18.
+//  Copyright © 2018 Realm. All rights reserved.
+//
+
+import SourceKittenFramework
+
+public struct SubClassRule: ASTRule, ConfigurationProviderRule, OptInRule {
+    public var configuration = SeverityLevelsConfiguration(warning: 0, error: 0)
```
marcelofabri: you should use `SeverityConfiguration`

Source/SwiftLintFramework/Rules/SubClassRule.swift
```diff
@@ -0,0 +1,47 @@
...
+    public func validate(file: File, kind: SwiftDeclarationKind,
+                         dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard SwiftDeclarationKind.functionKinds.contains(kind),
+              let offset = dictionary.offset,
+              case let contentsNSString = file.contents.bridge()
+            else {
+            return []
+        }
+
+        if contentsNSString.contains("super.") || contentsNSString.contains("super()") {
```
marcelofabri: this is not the right way of doing this since you can have subclasses that don't call `super`

driver733: @marcelofabri What would be the right way in your opinion? The `SourceKit` does not distinguish the inheritance between the interfaces and superclasses. It only shows that the class inherits from something.

marcelofabri: IMO we should have a rule that triggers if you don't add `final` to a class, as @SDGGiesbrecht [suggested](https://github.com/realm/SwiftLint/issues/2079#issuecomment-371944096).

There's not a lot of value (and in fact I'd argue that it's worst) to trigger a violation on cases where you don't have an alternative, usually dealing with dependencies (either system frameworks like UIKit or 3rd party ones).

Issue #2089, opened 2018-03-08
ThreadSanitizer reports data race at `File.lines.getter` on `xcodebuild test -enableThreadSanitizer YES`


norio-nomura:
```terminal.sh-session
$ DEVELOPER_DIR=/Applications/Xcode.app xcodebuild -workspace SwiftLint.xcworkspace/ -scheme swiftlint -enableThreadSanitizer YES test -only-testing:SwiftLintFrameworkTests/IntegrationTests/testSwiftLintLints
…
Test Case '-[SwiftLintFrameworkTests.IntegrationTests testSwiftLintLints]' started.
Loading configuration from '.swiftlint.yml'
==================
WARNING: ThreadSanitizer: data race (pid=32964)
  Read of size 8 at 0x7b1800019310 by main thread:
  * #0 File.lines.getter File.swift:37 (SourceKittenFramework:x86_64+0x3c537)
    #1 TrailingWhitespaceRule.validate(file:) TrailingWhitespaceRule.swift:32 (SwiftLintFramework:x86_64+0x4be08c)
    #2 protocol witness for Rule.validate(file:) in conformance TrailingWhitespaceRule TrailingWhitespaceRule.swift (SwiftLintFramework:x86_64+0x4c2d80)
    #3 Rule.lint(file:regions:benchmark:superfluousDisableCommandRule:) Linter.swift:71 (SwiftLintFramework:x86_64+0xc1119)
    #4 closure #2 in Linter.getStyleViolations(benchmark:) Linter.swift:138 (SwiftLintFramework:x86_64+0xd3acc)
    #5 partial apply for closure #2 in Linter.getStyleViolations(benchmark:) Linter.swift (SwiftLintFramework:x86_64+0xdc5dc)
    #6 thunk for @callee_owned (@in Rule) -> (@owned LintResult?) Linter.swift (SwiftLintFramework:x86_64+0xd3d9a)
    #7 partial apply for thunk for @callee_owned (@in Rule) -> (@owned LintResult?) Linter.swift (SwiftLintFramework:x86_64+0xdc6c9)
    #8 closure #1 in Array.parallelMap<A>(transform:) Array+SwiftLint.swift:67 (SwiftLintFramework:x86_64+0x4a966e)
    #9 partial apply for closure #1 in Array.parallelMap<A>(transform:) Array+SwiftLint.swift (SwiftLintFramework:x86_64+0x4a99c4)
[...]

PR #2092, opened 2018-03-09
Add `test_tsan` target to Makefile

norio-nomura:


norio-nomura:
This is preparation for #2089.
Since the issue is caused by SourceKitten, it will need updating SourceKitten to fix #2089.
So, after confirming that `test_tsan` fails on CircleCI, disable that step in CircleCI.

jpsim:
Nice work! Although let's not merge until it passes.

norio-nomura:
Rebased and enabled on CircleCI.

jpsim:
Nice! It must have been tricky to figure out exactly what the magic compiler flags needed to be!

PR #2094, opened 2018-03-11
Fix for crash on Linux on sources with surrogate characters

notorca:
Because of the bug https://bugs.swift.org/browse/SR-5971 SwiftLint
crashes when linting on Linux files like https://github.com/xmartlabs/Eureka/blob/master/Example/Example/ViewController.swift

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2050](https://github.com/realm/SwiftLint/issues/2050)
 
+* Fix linux crash on sources with surrogate pair emojis as variable names.  
+  [Cyril Lashkevich](https://github.com/notorca)
+
 ## 0.25.0: Cleaning the Lint Filter
 
 #### Breaking
--- a/Source/SwiftLintFramework/Extensions/CharacterSet+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/CharacterSet+SwiftLint.swift
@@ -1,0 +1,23 @@
+//
+//  CharacterSet+SwiftLint.swift
+//  SwiftLint
+//
+//  Created by Cyril Lashkevich on 3/11/18.
+//  Copyright © 2018 Realm. All rights reserved.
+//
+
+import Foundation
+
+// This is workaround for https://bugs.swift.org/browse/SR-5971
+// Can be removed when 
+// https://github.com/apple/swift-corelibs-foundation/pull/1471 is merged
+internal extension CharacterSet {
+    init(safeCharactersIn string: String) {
+#if os(Linux)
+        self.init()
+        insert(charactersIn: string)
+#else
+        self.init(charactersIn: string)
+#endif
+    }
+}
--- a/Source/SwiftLintFramework/Rules/GenericTypeNameRule.swift
+++ b/Source/SwiftLintFramework/Rules/GenericTypeNameRule.swift
@@ -166,7 +166,7 @@ public struct GenericTypeNameRule: ASTRule, ConfigurationProviderRule {
         }
 
         let allowedSymbols = configuration.allowedSymbols.union(.alphanumerics)
-        if !allowedSymbols.isSuperset(of: CharacterSet(charactersIn: name)) {
+        if !allowedSymbols.isSuperset(of: CharacterSet(safeCharactersIn: name)) {
             return [
                 StyleViolation(ruleDescription: type(of: self).description,
                                severity: .error,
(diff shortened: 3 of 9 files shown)


notorca:
This is temporary solution, until Foundation for linux is not fixed.

marcelofabri:
@notorca Can you add an unit test that would fail without this patch?

notorca:
@marcelofabri done

SDGGiesbrecht:
The technical terms mentioned here have been confused. There is no such thing as a “surrogate character”. (Did you mean “surrogate code point”? But if there is a surrogate code point left after string initialization, then the source code was malformed Unicode to begin with.)

SR‐3215, SR‐3282 report unsolved bugs in that initializer involving only ASCII. On the other hand, I use `CharacterSet` with Unicode often on Linux and have not noticed any issues.

I would like to investigate further to make sure the problem is correctly understood. I will report back.

notorca:
@SDGGiesbrecht you can try this on linux:
```
import Foundation
let cset = CharacterSet(charactersIn:"\u{10000}") 
```

notorca:
Here is the fix for Foundation https://github.com/apple/swift-corelibs-foundation/pull/1471

SDGGiesbrecht:
@notorca, yes, you are correct.

At first I had read [SR‐5971](https://bugs.swift.org/browse/SR-5971) that you linked to, where the hypothesis involved `sizeof(UniChar)`. That hypothesis was clearly wrong.

Then I looked at [`_CFCharacterSetInitWithCharactersInString`](https://github.com/apple/swift-corelibs-foundation/blame/a4ca69eb37dbff36a1f72a125553f6dc7a4385b4/CoreFoundation/String.subproj/CFCharacterSet.c#L1399-L1436), saw that it forwarded to `CFCharacterSetAddCharactersInString` just like your patch here, and that it has been unchanged for two years (i.e. not fixed since last release).

That was why I was suspicious at first.

But you are correct that Foundation’s current implementation `NULL`ifies the pointer before re‐trying with the other initializer. 😱

Both your patches (this one and the other in Foundation) look good to me. Good catch.

SDGGiesbrecht:
@notorca,

From your change log entry:
> Fix linux crash on sources with surrogate pair emojis as variable names

Actually, you have fixed a lot more than that. Not just in variables, but several different token types, and not just emoji, but over 94% of all Unicode scalars.

If you want, you can change it to something like, “Fixed Linux crashes on source files with Unicode scalars from supplementary planes.”

But you can leave it the way you have it if you would rather.

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2050](https://github.com/realm/SwiftLint/issues/2050)
 
+* Fix linux crash on sources with surrogate pair emojis as variable names
```
marcelofabri: requires period and two trailing spaces as described in CONTRIBUTING.md: https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes

Source/SwiftLintFramework/Extensions/CharacterSet+SwiftLint.swift
```diff
@@ -0,0 +1,20 @@
...
+//  Copyright © 2018 Realm. All rights reserved.
+//
+
+import Foundation
+
+public extension CharacterSet {
+    init(safeCharactersIn string: String) {
+#if os(Linux)
+        self.init()
+        self.insert(charactersIn: string)
```
marcelofabri: unneeded `self`

Source/SwiftLintFramework/Extensions/CharacterSet+SwiftLint.swift
```diff
@@ -0,0 +1,20 @@
...
+//  CharacterSet+SwiftLint.swift
+//  SwiftLint
+//
+//  Created by Cyril Lashkevich on 3/11/18.
+//  Copyright © 2018 Realm. All rights reserved.
+//
+
+import Foundation
+
+public extension CharacterSet {
```
marcelofabri: this should be `internal`, not `public`

Source/SwiftLintFramework/Extensions/CharacterSet+SwiftLint.swift
```diff
@@ -0,0 +1,20 @@
...
+    init(safeCharactersIn string: String) {
+#if os(Linux)
```
marcelofabri: can you please link the SR issue here so we can track when that's fixed?

notorca:
@marcelofabri Done,

marcelofabri:
@notorca Thanks!

jpsim:
Nice sleuthing here folks! One thing someone could do if they’d like is to add a custom rule to lint for our uses of `CharacterSet(charactersIn:)` to ensure we use the safe version.

Issue #2071, opened 2018-02-26
Add option to disable swiftlint in-file


fredpi:
For some cases, it would be useful to have a command to disable swiftlint parsing for a specific file (e. g. `// swiftlint:disable:file`.

I know there's already an exclude option in the .yml file, but that one is not always suitable. Consider 3rd-party-frameworks like [SwiftGen](https://github.com/SwiftGen/SwiftGen) that currently always have to specifically disable single rules. A disable command in-file would fix that issue.

marcelofabri:
We probably should follow [rubocop's approach](http://rubocop.readthedocs.io/en/latest/configuration/):

```swift
// swiftlint:disable all
[...]
// swiftlint:enable all
```

@fredpi Would you like to submit a PR implementing this feature?

fredpi:
Yeah, will do a PR with an implementation these days

fredpi:
Didn't forget about this, however, will have to delay this a few days. On a tight schedule currently...

marcelofabri:
No worries!

bj97301:
Hows it going?

fredpi:
@bj97301 As pointed out, takes a bit, but looking good...

bj97301:
Nice.  Looking forward to it :D

marcelofabri:
Closed in #2095 by @fredpi 💯

PR #2095, opened 2018-03-11
Add all keyword

fredpi:
Implements #2071

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@
   [#2045](https://github.com/realm/SwiftLint/issues/2045)
   * Add `all` keyword for use in disable / enable statement:  `//swiftlint:disable all`. It allows disabling SwiftLint entirely for a particular section.`
 * Add `all` keyword for use in disable / enable statement:  `//swiftlint:disable all`. It allows disabling SwiftLint entirely for a particular section.`
+* Add `all` keyword for use in disable / enable statement:
+  `// swiftlint:disable all`.
+  It allows disabling SwiftLint entirely for a particular section.
   [fredpi](https://github.com/fredpi)
   [#2071](https://github.com/realm/SwiftLint/issues/2071)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@
 * Add `all` keyword for use in disable / enable statement:  `//swiftlint:disable all`. It allows disabling SwiftLint entirely for a particular section.`
 * Add `all` keyword for use in disable / enable statement:
   `// swiftlint:disable all`.
-  It allows disabling SwiftLint entirely for a particular section.
+  It allows disabling SwiftLint entirely for a particular section.  
   [fredpi](https://github.com/fredpi)
   [#2071](https://github.com/realm/SwiftLint/issues/2071)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,11 +20,10 @@
   `error` variable.  
   [Daniel Metzing](https://github.com/dirtydanee)
   [#2045](https://github.com/realm/SwiftLint/issues/2045)
-  * Add `all` keyword for use in disable / enable statement:  `//swiftlint:disable all`. It allows disabling SwiftLint entirely for a particular section.`
-* Add `all` keyword for use in disable / enable statement:  `//swiftlint:disable all`. It allows disabling SwiftLint entirely for a particular section.`
+  
 * Add `all` keyword for use in disable / enable statement:
   `// swiftlint:disable all`.
-  It allows disabling SwiftLint entirely for a particular section.  
+  It allows disabling SwiftLint entirely, in-code, for a particular section.  
   [fredpi](https://github.com/fredpi)
   [#2071](https://github.com/realm/SwiftLint/issues/2071)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
   [Daniel Metzing](https://github.com/dirtydanee)
   [#2045](https://github.com/realm/SwiftLint/issues/2045)
   * Add `all` keyword for use in disable / enable statement:  `//swiftlint:disable all`. It allows disabling SwiftLint entirely for a particular section.`
+* Add `all` keyword for use in disable / enable statement:  `//swiftlint:disable all`. It allows disabling SwiftLint entirely for a particular section.`
   [fredpi](https://github.com/fredpi)
   [#2071](https://github.com/realm/SwiftLint/issues/2071)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@
   `error` variable.  
   [Daniel Metzing](https://github.com/dirtydanee)
   [#2045](https://github.com/realm/SwiftLint/issues/2045)
+  * Add `all` keyword for use in disable / enable statement:  `//swiftlint:disable all`. It allows disabling SwiftLint entirely for a particular section.`
+  [fredpi](https://github.com/fredpi)
+  [#2071](https://github.com/realm/SwiftLint/issues/2071)
 
 #### Bug Fixes
 
--- a/README.md
+++ b/README.md
@@ -252,8 +252,8 @@ let hasWarning :String = "" // Warning generated about colons immediately after 
 
 Including the `all` keyword will disable all rules until the linter sees a matching enable comment:
 
-`// swiftlint:disable all [<rule1> <rule2>...]`
-`// swiftlint:enable all [<rule1> <rule2>...]`
+`// swiftlint:disable all`
+`// swiftlint:enable all`
 
 For example:
 
--- a/README.md
+++ b/README.md
@@ -250,6 +250,22 @@ let noWarning :String = "" // No warning about colons immediately after variable
 let hasWarning :String = "" // Warning generated about colons immediately after variable names
 ```
 
+Including the `all` keyword will disable all rules until the linter sees a matching enable comment:
+
+`// swiftlint:disable all [<rule1> <rule2>...]`
+`// swiftlint:enable all [<rule1> <rule2>...]`
+
+For example:
+
+```swift
+// swiftlint:disable all
+let noWarning :String = "" // No warning about colons immediately after variable names!
+let i = "" // Also no warning about short identifier names
+// swiftlint:enable all
+let hasWarning :String = "" // Warning generated about colons immediately after variable names
+let y = "" // Warning generated about short identifier names
+```
+
 It's also possible to modify a `disable` or `enable` command by appending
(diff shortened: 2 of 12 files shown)


README.md
```diff
@@ -250,6 +250,22 @@ let noWarning :String = "" // No warning about colons immediately after variable
 let hasWarning :String = "" // Warning generated about colons immediately after variable names
 ```
 
+Including the `all` keyword will disable all rules until the linter sees a matching enable comment:
+
+`// swiftlint:disable all [<rule1> <rule2>...]`
```
marcelofabri: should we remove the `[<rule1> <rule2>...]`?

Source/SwiftLintFramework/Models/Linter.swift
```diff
@@ -78,11 +78,12 @@ private extension Rule {
             return region?.isRuleEnabled(self) ?? true
         }
 
-        let ruleIDs = Self.description.allIdentifiers +
+        let ruleIds = Self.description.allIdentifiers +
```
marcelofabri: is there a reason for renaming this variable?

Source/SwiftLintFramework/Models/Command.swift
```diff
@@ -84,15 +84,16 @@ public struct Command: Equatable {
...
-            .substring(from: scanner.scanLocation + 1)
-            .components(separatedBy: .whitespaces)
         line = lineAndCharacter.line
         character = lineAndCharacter.character
 
-        let hasModifier = actionAndModifierScanner.scanString(string: ":") != nil
+        let ruleTexts = scanner.string.bridge().substring(
+            from: scanner.scanLocation + 1
+        ).components(separatedBy: .whitespaces)
+        ruleIdentifiers = Set(ruleTexts.map { RuleIdentifier($0) })
```
marcelofabri: `.map(RuleIdentifier.init)`?

Source/SwiftLintFramework/Models/RuleIdentifier.swift
```diff
@@ -0,0 +1,48 @@
...
+import Foundation
+
+public enum RuleIdentifier: Hashable, ExpressibleByStringLiteral {
+    case all
+    case some(identifier: String)
+
+    private static let allStringRepresentation = "all"
+
+    public var hashValue: Int {
+        switch self {
```
marcelofabri: what do you think about using `stringRepresentation.hashValue`?

Source/SwiftLintFramework/Models/RuleIdentifier.swift
```diff
@@ -0,0 +1,48 @@
...
+    public init(_ value: String) {
+        self = value == RuleIdentifier.allStringRepresentation ? .all : .some(identifier: value)
+    }
+
+    public init(stringLiteral value: String) {
+        self = RuleIdentifier(value)
+    }
+
+    public static func == (lhs: RuleIdentifier, rhs: RuleIdentifier) -> Bool {
+        return lhs.hashValue == rhs.hashValue
```
marcelofabri: we shouldn't rely on `hashValue` here. It's better to compare `stringRepresentation`.

Tests/SwiftLintFrameworkTests/DisableAllTests.swift
```diff
@@ -0,0 +1,38 @@
...
+//  Copyright © 2018 Realm. All rights reserved.
+//
+
+import Foundation
+@testable import SwiftLintFramework
+import XCTest
+
+class DisableAllTests: XCTestCase {
+     /// Example violation (identifer_name). Could be replaced with any other single violation.
+    private let violatingPhrase = "let r = 0\n"
```
marcelofabri: can you add other violation triggered by other rule in the example? I want to make sure that it disables several ones, just not one.

Tests/SwiftLintFrameworkTests/DisableAllTests.swift
```diff
@@ -0,0 +1,38 @@
...
+//  Created by Frederick Pietschmann on 3/11/18.
+//  Copyright © 2018 Realm. All rights reserved.
+//
+
+import Foundation
+@testable import SwiftLintFramework
+import XCTest
+
+class DisableAllTests: XCTestCase {
+     /// Example violation (identifer_name). Could be replaced with any other single violation.
```
marcelofabri: not aligned with the declaration below (one extra whitespace in the beginning)

Source/SwiftLintFramework/Models/Region.swift
```diff
@@ -12,12 +12,12 @@ import SourceKittenFramework
 public struct Region: Equatable {
     public let start: Location
     public let end: Location
-    public let disabledRuleIdentifiers: Set<String>
+    public let disabledRuleIdentifiers: Set<RuleIdentifier>
 
-    public init(start: Location, end: Location, disabledRuleIdentifiers: Set<String>) {
+    public init(start: Location, end: Location, disabledRules: Set<RuleIdentifier>) {
```
marcelofabri: I think we should keep the `disabledRuleIdentifiers`

Source/SwiftLintFramework/Models/Region.swift
```diff
@@ -30,17 +30,17 @@ public struct Region: Equatable {
 
     public func isRuleDisabled(_ rule: Rule) -> Bool {
         let identifiers = type(of: rule).description.allIdentifiers
-        return !disabledRuleIdentifiers.isDisjoint(with: identifiers)
+        return disabledRuleIdentifiers.contains(.all) ||
```
marcelofabri: What do you think about:

```
guard !disabledRuleIdentifiers.contains(.all) else {
   return true
}

return Set(disabledRuleIdentifiers.map { $0.stringRepresentation }).isDisjoint(with: identifiers)
```

Tests/SwiftLintFrameworkTests/DisableAllTests.swift
```diff
@@ -0,0 +1,38 @@
...
+//
+//  Created by Frederick Pietschmann on 3/11/18.
+//  Copyright © 2018 Realm. All rights reserved.
+//
+
+import Foundation
+@testable import SwiftLintFramework
+import XCTest
+
+class DisableAllTests: XCTestCase {
```
marcelofabri: I also would like to see tests checking the behavior when `(enable|disable):(next|previous|this)` is used, especially nested with other regions.

marcelofabri:
@fredpi Sorry for the delay! Thanks for the contribution, I just made some comments. Let me know what do you think.

README.md
fredpi: I would leave it since it's still possible to exclude single other rules additionally, then disable the all rule and have the single other rules still be in effect.

Source/SwiftLintFramework/Models/Command.swift
fredpi: This would be ambiguous since there are two inits with string available...

Source/SwiftLintFramework/Models/RuleIdentifier.swift
fredpi: Makes sense

Source/SwiftLintFramework/Models/RuleIdentifier.swift
fredpi: Makes sense

Tests/SwiftLintFrameworkTests/DisableAllTests.swift
fredpi: Done

Source/SwiftLintFramework/Models/Region.swift
fredpi: Rather !isDisjoint... ;)

fredpi:
@marcelofabri Done implementing most of your proposals. For those I didn't implement, I left an explanation why I didn't include it (yet) in the perspective code review section.

CHANGELOG.md
```diff
@@ -11,6 +11,12 @@
   [Rounak Jain](https://github.com/rounak)
   [#2028](https://github.com/realm/SwiftLint/issues/2028)
 
+* Add `all` keyword for use in disable / enable statement:
+  `// swiftlint:disable all`.
+  It allows disabling SwiftLint entirely for a particular section.
```
ornithocoder: Don't forget to include two trailing spaces at the end of the description (check the other entries on the ChangeLog).

fredpi: Done. I originally did so, but Xcode dropped it automatically ;( Used another editor now

Source/SwiftLintFramework/Models/Command.swift
ornithocoder: You can use the qualified symbol name. For instance, `RuleIdentifier.init(_:)` to specify the one without label. This way you can use map - e.g. `.map(RuleIdentifier.init(_:))`.

fredpi: Thanks, implemented that one

README.md
marcelofabri: Since we already cover disabling individual rules above, I'd just remove them here to make it clearer to understand. I don't expect people to use `all` with other rule identifiers commonly.

Source/SwiftLintFramework/Models/RuleIdentifier.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+//  Created by Frederick Pietschmann on 3/5/18.
+//  Copyright © 2018 Realm. All rights reserved.
+//
+
+import Foundation
+
+public enum RuleIdentifier: Hashable, ExpressibleByStringLiteral {
+    case all
+    case some(identifier: String)
```
marcelofabri: what do you think about using `ruleID` instead of `some`? 🤔  

Sorry for not catching this earlier.

README.md
fredpi: Makes sense

Source/SwiftLintFramework/Models/RuleIdentifier.swift
fredpi: `.some` is actually some legacy naming when the enum was still designed with cases `all` and `.some(let identifiers)` (there were many), so it doesn't fit properly now that it only holds one identifier. Yet I wouldn't go to with `ruleID`which would be confusing to use (`RuleIdentifier.ruleID`). Additionally `ruleID` rather describes the associated value than the case itself.

Since `.some` nonetheless doesn't fit well, I suggest using `case explicit(identifier: String)`

marcelofabri:
@fredpi There's only one conflict in the CHANGELOG. Can you please rebase and fix it so we can merge? 💯

fredpi:
@marcelofabri I did the rebase and fixed the conflict.

I also changed the RuleIdentifier.explicit case to .single again, which I think, after a bit of consideration, fits better.

marcelofabri:
@fredpi Thanks for the amazing contribution! 🎉

PR #2117, opened 2018-03-28
Fixed plurality of autocorrect done message

jszumski:
Fixed issue where the autocorrect done message used the plural form of "files" even if only 1 file changed.


marcelofabri:
Thanks 🎉

PR #2124, opened 2018-03-30
New Rule - Explicit Final

Twigz:
Adds a new `explicit_final` rule to require final on all non-open class types, to encourage composition over inheritance.

### DIFF
--- a/Source/SwiftLintFramework/Rules/ExplicitFinalRule.swift
+++ b/Source/SwiftLintFramework/Rules/ExplicitFinalRule.swift
@@ -56,12 +56,10 @@ public struct ExplicitFinalRule: ASTRule, OptInRule, ConfigurationProviderRule {
 
     public func validate(file: File, kind: SwiftDeclarationKind,
                          dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
-        guard kind == .class else { return [] }
-
-        guard let accessibility = dictionary.accessibility else { return [] }
-        guard AccessControlLevel(identifier: accessibility) != .open else { return [] }
-
-        guard !dictionary.enclosedSwiftAttributes.contains("source.decl.attribute.final") else { return [] }
+        guard kind == .class,
+              let accessibility = dictionary.accessibility,
+              AccessControlLevel(identifier: accessibility) != .open,
+              !dictionary.enclosedSwiftAttributes.contains("source.decl.attribute.final") else { return [] }
 
         guard let offset = dictionary.offset else { return [] }
 
--- a/Source/SwiftLintFramework/Rules/ExplicitFinalRule.swift
+++ b/Source/SwiftLintFramework/Rules/ExplicitFinalRule.swift
@@ -1,0 +1,72 @@
+//
+//  ExplicitFinalRule.swift
+//  SwiftLint
+//
+//  Created by Andrew Harrison on 3/30/18.
+//  Copyright © 2018 Realm. All rights reserved.
+//
+
+import Foundation
+import SourceKittenFramework
+
+public struct ExplicitFinalRule: ASTRule, OptInRule, ConfigurationProviderRule {
+
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "explicit_final",
+        name: "Explicit Final",
+        description: "Non-open class types should always specify final to encourage composition over inheritance.",
+        kind: .idiomatic,
+        nonTriggeringExamples: [
+            "final class A {}",
+            "internal final class B {}",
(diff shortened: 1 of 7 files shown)


SDGGiesbrecht:
@Twigz, if the purpose is “to encourage composition [protocols] over inheritance [subclasses]”, why do you make an exception to allow `open`, which is for cross‐module inheritance? `open` is not just the use of inheritance, but the forcing of client modules into using inheritance too.

I imagine this is pull request is based on issue #2079.

Twigz:
@SDGGiesbrecht The exception for `open` is because it is a mutually exclusive modifier to `final`, unlike other access control modifiers. I think that if you wanted to exclude `open` from being allowed, it would be a separate rule.

As far as `open` forcing client modules to use inheritance, that is not a hard requirement. While some objects may be architected that way, `open` is simply a modifier that allows other modules to subclass, it does not explicitly require it.

SDGGiesbrecht:
> As far as open forcing client modules to use inheritance, that is not a hard requirement. While some objects may be architected that way, open is simply a modifier that allows other modules to subclass, it does not explicitly require it.

But it means you have provided a base class for them to subclass instead of a protocol to conform to. If they want to use the implementations your module provides, they have to subclass what you have given them. They cannot conform to your base class as if it were a protocol.

> I think that if you wanted to exclude open from being allowed, it would be a separate rule.

The motivation is the same, the violation is the same, and I can imagine no reason for using one without the other. I think they should be just one rule. Simply modify the message if `open` is there to avoid confusing users. I would suggest:

“Use (`public`) `final`, or create a protocol instead to avoid the restrictions of single inheritance.”  
(Only including the word “`public`” if `open` is present.)

The message as it is in the pull request is very opaque:

> Non-open class types should always specify final to encourage composition over inheritance.

At the very least, using the common Swift terms, “protocols” and “superclasses”, instead of the abstract jargon, “composition”, and “inheritance”, will give many users a better chance of understanding what they are being told to do. Anyone looking at a Swift file likely knows what protocols are. But many will need to resort to a dictionary or encyclopedia to understand what composition refers to here, and they will find many unrelated definitions before they get to the relevant technical one.

But wait and see what @marcelofabri has to say.

Source/SwiftLintFramework/Rules/ExplicitFinalRule.swift
```diff
@@ -0,0 +1,72 @@
...
+            "final class G {\n" +
+            "    final class H {\n" +
+            "        ↓class I { }\n" +
+            "    }\n" +
+            "}"
+        ])
+
+    public func validate(file: File, kind: SwiftDeclarationKind,
+                         dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard kind == .class else { return [] }
```
ornithocoder: How about merging all these guards statements?

Tests/SwiftLintFrameworkTests/RulesTests.swift
```diff
@@ -114,6 +114,10 @@ class RulesTests: XCTestCase {
         verifyRule(ExplicitEnumRawValueRule.description)
     }
 
+    func testExplicitFinal() {
```
ornithocoder: Don't forget the Linux tests :-)

marcelofabri:
Maybe this rule could have a setting to specify which access levels should be validated?

Twigz:
@marcelofabri Sorry, just coming back around to this now. We could add a setting to specify access levels yes, but I'm not sure that makes sense in the case of `open`. A rule entitled `explicit_final` with `open` being added as one of the options would be closer to `disallow_open`, instead of requiring `final`, since `open` is mutually exclusive with `final`. It then becomes an inconsistency among the options, as it expressly disallows it's use, rather than requiring the addition of the `final` keyword. A modification of language for that case would assist the user in understanding why they can't use open, but, to me, disallowing use of an access modifier seems out of scope for a rule that is requiring the use of `final`.

We could rename the rule to `disable_subclassing` with the setting to specify access levels that disallow subclassing, though I would argue then that rule name is slightly confusing, since it would require adding `final` which may be unintuitive from the rule name.

I'll do whatever you think is best in this scenario, just explaining my rationale.

Tests/SwiftLintFrameworkTests/RulesTests.swift
Twigz: Done 👍

Twigz:
@marcelofabri Just pinging for any guidance you might be able to provide here.

kimdv:
Any status on this @Twigz ? 👍

cennydavidsson:
@Twigz @marcelofabri  Any updates on this.

Twigz:
Think I'll just close this then.

PR #2128, opened 2018-03-31
Improved autocorrect behavior of multiple rules when preprocessor macros are present

jszumski:
I audited the autocorrect behavior of correctable AST rules when preprocessor macros are present:
* `EmptyEnumArgumentsRule`: no change needed
* `EmptyParenthesesWithTrailingClosureRule`: fixed with `.unique`
* `ExplicitInitRule`: fixed with `.unique`
* `JoinedDefaultParameterRule`: fixed with `.unique`
* `RedundantOptionalInitializationRule`: fixed with `.unique`
* `TrailingCommaRule`: has issues, see below
* `UnusedClosureParameterRule`: fixed with `.unique`

**Known issue:**

`TrailingCommaRule` seems to be organized a little differently (older?) than the other updated tests.  Adding a new triggering example:

```
class C {\n #if true\n func f() {\n let foo = [1, 2, 3↓,]\n }\n #endif\n}
```

and applying the fix:

```
    private func violationRanges(in file: File,
                                 dictionary: [String: SourceKitRepresentable]) -> [NSRange] {
        return dictionary.substructure.flatMap { subDict -> [NSRange] in
            var violations = violationRanges(in: file, dictionary: subDict)

            if let kindString = subDict.kind,
                let kind = KindType(rawValue: kindString),
                let index = violationIndexAndReason(in: file, kind: kind, dictionary: subDict)?.index {
                violations += [NSRange(location: index, length: 1)]
            }

            return violations
        }.unique
    }
```

fixes the duplicate correction, but ends up causing `testTrailingCommaRuleWithDefaultConfiguration` to fail for the example:

```
"let foo = [1, 2, 3↓,]\n"
```

It looks like linting and correcting don't go through the same code path.

It almost seems as if we want to apply `.unique` to the results of `ASTRule`'s `validate(file:dictionary:)` in one place rather than sprinkling it throughout various rules.  I realize this would break rules that can have multiple valid results at the same location (for example `DiscouragedOptionalCollectionRule`).  Should we have rules opt into the uniqueness restriction where it makes sense? Something like adding this to `Rule`:

```
var violationsAreUnique: Bool { get }
```

If you'd like me to investigate that approach I'm happy to file an issue and tackle it in another PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,11 @@
 * Fix issue where the autocorrect done message used the plural form of "files"
   even if only 1 file changed.  
   [John Szumski](https://github.com/jszumski)
+  
+* Fix autocorrection for several rules (`empty_parentheses_with_trailing_closure`, 
+  `explicit_init`, `joined_default_parameter`, `redundant_optional_initialization` 
+  and `unused_closure_parameter `) when used with preprocessor macros.  
+  [John Szumski](https://github.com/jszumski)
 
 ## 0.25.0: Cleaning the Lint Filter
 
--- a/Rules.md
+++ b/Rules.md
@@ -13769,6 +13769,16 @@ let foo = ["אבג", "αβγ", "🇺🇸"↓,]
 
 ```
 
+```swift
+class C {
+ #if true
+ func f() {
+ let foo = [1, 2, 3↓,]
+ }
+ #endif
+}
+```
+
 </details>
 
 
--- a/Source/SwiftLintFramework/Protocols/ASTRule.swift
+++ b/Source/SwiftLintFramework/Protocols/ASTRule.swift
@@ -15,7 +15,9 @@ public protocol ASTRule: Rule {
 
 public extension ASTRule where KindType.RawValue == String {
     func validate(file: File) -> [StyleViolation] {
-        return validate(file: file, dictionary: file.structure.dictionary)
+        let violations = validate(file: file, dictionary: file.structure.dictionary)
+
+        return violationsAreUnique ? violations.unique : violations
     }
 
     func validate(file: File, dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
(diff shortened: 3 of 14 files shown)


marcelofabri:
> It almost seems as if we want to apply .unique to the results of ASTRule's validate(file:dictionary:) in one place rather than sprinkling it throughout various rules.

Could you open a PR exploring this path? I'm not actually sure that this would fix the issue because autocorrect and linting for ASTRules are usually implemented a bit different.

jszumski:
@marcelofabri I ended up adding it here to take advantage of the triggers already defined.  This PR now adds the property and the previously problematic trigger case for `TrailingCommaRule`.  

I wish we could centralize the calls to `.unique` for correction cases, but it doesn't look possible since they are recognized and corrected immediately before getting reported back to the caller.

jszumski:
@marcelofabri any thoughts on this approach?

marcelofabri:
@jszumski It's on my todo list to take a look at this, sorry for the delay!

marcelofabri:
@jszumski So, I took a look at this.

> It looks like linting and correcting don't go through the same code path.

This is kind of true. 

I don't see a reason to keep the `violationsAreUnique` property. Just adding `unique` in the rules that have auto-correct is enough. One test was failing for me, but it was because of the docs.

[This](https://gist.github.com/marcelofabri/3f9469dc8715a7a617d35484b8f2b828) is the patch that I think we should apply to simplify things.

jpsim:
This looks great. I'll merge with @marcelofabri's suggested patch, which I agree with for now.

I do think it'd be worth refactoring the auto-correction flow to share more common code, including this uniquing. In fact, I think there are lots of places that have evolved in SwiftLint over the years to duplicate code that would be good opportunities to consider consolidating.

Commits where this was merged:

* d2d6c8f6: Improved autocorrect behavior of multiple rules when preprocessor macros are present.
* b03b4f08: Added a new computed property that identifies whether a rule should treat multiple violations at the same location as only one violation.
* 4649df97: Added `violationsAreUnique` checks to existing uses of unique.
* d8283a40: Added new trigger for trailing_comma.
* 49916465: Prevent discouraged_optional_collection from uniquing values since nested collections will trigger multiple times at the start of the first collection.
* 2eed8d46: Apply Marcelo Fabri's patch
* 4d00a8e0: Fix changelog entry location
* abc6e286: Update Rules.md

PR #2131, opened 2018-04-02
Fix readme's `Warning` emoji

SohailKh:
Old warning emoji was making the 'This plugin' text funky.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ Alternatively, if you've installed SwiftLint via CocoaPods the script should loo
 To run `swiftlint autocorrect` on save in Xcode, install the
 [SwiftLintXcode](https://github.com/ypresto/SwiftLintXcode) plugin from Alcatraz.
 
-⚠ ️This plugin will not work with Xcode 8 or later without disabling SIP.
+⚠️This plugin will not work with Xcode 8 or later without disabling SIP.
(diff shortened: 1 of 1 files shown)


marcelofabri:
Thanks!

SDGGiesbrecht:
Thanks for fixing it, @SohailKh.

***

The [commit which introduced that line a year and half ago](https://github.com/realm/SwiftLint/commit/e9b990a43f4ba903d5a6be803bfa961bea9c6d93#diff-04c6e90faac2675aa89e2176d2eec7d8R57) isolated the variation selector from the symbol by a space for some reason.

i.e.:  
`⚠ + [SP] + [VS16]` →  “⚠ ️”  
...instead of...  
`⚠ + [VS16]` → “⚠️”

It is not valid Unicode to combine a variation selector with a space. From [Unicode.org](http://unicode.org/faq/vs.html):

> Q: Can I define my own [variation] sequences?
A: No, that is the equivalent of trying to define an unassigned code point to be your own character.

Since it is difficult to manually inject something into the middle of a grapheme cluster, I doubt this was a human error. @mrh-is, you may want to check and see if your code editor or an input source (keyboard, character map, auto‐replace, etc.) is causing this sort of problem across your other projects too.

PR #2132, opened 2018-04-02
Include named closure parameters in closure end indentation rule

### LINKED ISSUE #2121: Possible closure_end_indentation rule issue
### New Issue Checklist

### Rule issue
I'm wondering whether the following code should be triggering the `closure_end_indentation` rule to fail (it currently doesn't):

```swift
function(
  parameter: a,
  closure: { x in
    // code
},
  anotherClosure: {
    // code
})
```

To me, this reads as a violation of the rule: since the closure's opening brace is on a line with one level of indentation, it seems like the closing brace should have the same level of indentation. The rule reads as follows:

> "Closure end should have the same indentation as the line that started it.

My assumption from reading the rule is that the following code would satisfy it:

```swift
function(
  parameter: a,
  closure: { x in
    // code
  },
  anotherClosure: {
    // code
  })
```

From this, it would appear that currently, the rule treats "the line that started it" as the start of the function declaration, rather than the line on which the opening brace is located.

What do you think? Thanks in advance! cc @marcelofabri

erichoracek:
Fixes #2121

Let me know if you think this belongs as a config parameter to this rule. Thanks in advance for taking a look!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2127](https://github.com/realm/SwiftLint/issues/2127)
 
+* Fixes a case where the `closure_end_indentation` rule wouldn't lint the end
+  indentation of non-trailing closure parameters.
+  [Eric Horacek](https://github.com/erichoracek)
+  [#2121](https://github.com/realm/SwiftLint/issues/2121)]
+
 * Add `redundant_set_access_control` rule to warn against using redundant
   setter ACLs on variable declarations.  
   [Marcelo Fabri](https://github.com/marcelofabri)
@@ -98,11 +103,6 @@ The next release will require Swift 4.0 or higher to build.
   they are explicitly specified with `--path`.  
   [Ash Furrow](https://github.com/ashfurrow)
   [#2051](https://github.com/realm/SwiftLint/issues/2051)
-
-* Fixes a case where the `closure_end_indentation` rule wouldn't lint the end
-  indentation of non-trailing closure parameters.
-  [Eric Horacek](https://github.com/erichoracek)
-  [#2121](https://github.com/realm/SwiftLint/issues/2121)]
 
 * Adds `discouraged_optional_collection` opt-in rule to encourage the use of
   empty collections instead of optional collections.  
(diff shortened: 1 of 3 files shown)


marcelofabri:
IIRC this was a tradeoff that I did when writing the rule to avoid triggering too many warnings in cases like [this]( https://github.com/mozilla-mobile/firefox-ios/blob/590477391c375efd2a4f7959fdd11df4d55a1572/Client/Frontend/Browser/ButtonToast.swift#L155:13) where it's not really super clear what the "best" style is.

Or even in cases [like this](https://github.com/JohnCoates/Aerial/blob/bb4e0ffe1412c477aa4391cf2f58bfaddc53c5af/Aerial/Source/Models/ManifestLoader.swift#L166:9) where fixing the violation would be actually *worse* than the way it's right now.

Maybe we should only make this check if the first parameter is in a new line?

erichoracek:
@marcelofabri good call—I've updated this rule to only lint closure argument identation if the first parameter is on a newline. PTAL!

jpsim:
Looks great overall! I did find one false positive though:

```swift
self.asdf(asdf: .asdf, asdfasdfasdf: {
    asdfasdfasdfasdf })
{ result in
    guard case .success = result else {
        return
    }

↓    self.asdfasdfasdf(asdfasdfasdfasdf: true) {
        asdfasdfASDFasdfasdf
    }
}
```

code obviously obfuscated.

CHANGELOG.md
```diff
@@ -19,6 +19,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2127](https://github.com/realm/SwiftLint/issues/2127)
 
+* Fixes a case where the `closure_end_indentation` rule wouldn't lint the end
+  indentation of non-trailing closure parameters.
```
jpsim: two trailing spaces here please, as per https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes

erichoracek:
@jpsim Thanks for the review! I tried pasting your false positive into the non-triggering examples section but I wasn't able to get it to trigger a test failure. I've pushed it up on a branch here—let me know if I'm missing something:

https://github.com/erichoracek/SwiftLint/commit/1bdf8165397eafd276358778ee1c8e38a80fe7c6

Thanks in advance for taking a look!

jpsim:
Sorry @erichoracek I made a mistake, there's no false positive. It was a false-false-positive 😅 

I'm reviewing the rest now.

PR #2136, opened 2018-04-03
Add LowerACLThanBodyRule

keith:
This new rule validates that if a type/function/variable definition has
an ACL specifier, it is more restrictive than the containing body's
level. This is intended to lint a peculiarity of SE-0025 where it is
stated:

> The compiler should not warn when a broader level of access control is
used within a type with more restrictive access, such as internal within
a private type. This allows the designer of the type to select the
access they would use were they to make the type more widely accessible.

I think this is an anti-goal because when a type is made more open, it
should be a concious decision at that time to make it public. This is of
course an opt-in rule as well. This also has the added benefit of
linting this compiler bug: https://bugs.swift.org/browse/SR-2925


keith:
There are 2 cases where I think it could be argued that this is a downside. The first is the required to fix the bug I linked in the description, but if you have this valid definition:

```swift
public protocol Foo {}
extension Foo {
    public func bar() {}
}
```

You would now have to annotate `extension` with `public` as well. I don't see this as a major downside, but it isn't a place where they're commonly added.

The other case is with using `fileprivate` and `private` interchangeably at the top level. This is currently valid:

```swift
private struct Foo {
    fileprivate var bar: String?
}
```

But this would be required to change to `fileprivate struct Foo` with this rule. Overall I think this makes this case more clear, but it does require more `fileprivate uses`. Also in this case you couldn't just omit `fileprivate` on the property, which you can otherwise.

These are some downsides, but overall I think this would still be worth it for us in particular!

keith:
Hmmm:

```
SwiftLintFrameworkTests.DocumentationTests
  testRulesDocumentationIsUpdated, XCTAssertEqual failed: ("
  /Users/distiller/project/Tests/SwiftLintFrameworkTests/DocumentationTests.swift:25
  ```

        XCTAssertEqual(existingDocs, updatedDocs)

  ``````

**Comment by marcelofabri** [comment]
@keith Running the unit tests locally should fail on the first time, but update the docs. It shouldn't fail after Rules.md is updated.

**Comment by keith** [comment]
Updated!

File: CHANGELOG.md [hunk 1]
```diff
@@ -6,6 +6,10 @@
 
 #### Enhancements
 
+* Add `LowerACLThanParent` rule
```
ornithocoder: Hi @keith, please leave [two trailing spaces](https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes) after the description.

Tests/SwiftLintFrameworkTests/RulesTests.swift
```diff
@@ -98,6 +98,10 @@ class RulesTests: XCTestCase {
         verifyRule(EmptyParametersRule.description)
     }
 
+    func testLowerACLThanParent() {
```
ornithocoder: Please include the Linux tests as well.

CHANGELOG.md
keith: Done!

Tests/SwiftLintFrameworkTests/RulesTests.swift
keith: Done!

jpsim:
Given that this is opt-in, I'm comfortable with the false positives. Nicely implemented @keith!

keith:
Thanks!

keith:
Follow up to make this a bit more useful https://github.com/realm/SwiftLint/pull/2164

PR #2138, opened 2018-04-04
[CircleCI] Use pre-installed ruby-2.3 instead of installing by Homebrew

norio-nomura:


norio-nomura:
This reduces the duration of the cocoapods and osscheck job by roughly 1 minute 30 seconds.

norio-nomura:
🙂

jpsim:
Wow, so much nicer 😍

PR #2143, opened 2018-04-05
Add 4.1.0, 4.1.1 and 4.2.0 to Swift version detection

norio-nomura:


norio-nomura:
It seems https://cocoapods-specs.circleci.com/fetch-cocoapods-repo-from-s3.sh caused failure.
I'll rerun the job later.

PR #2152, opened 2018-04-12
Use SwiftDeclarationAttributeKind instead of string constants

marcelofabri:
`SwiftDeclarationAttributeKind` was added in the latest release of SourceKitten 🎉

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -60,8 +60,8 @@
         "repositoryURL": "https://github.com/jpsim/SourceKitten.git",
         "state": {
           "branch": null,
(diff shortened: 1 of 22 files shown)


jpsim:
Now that SourceKitten 0.20.0 is out, can we update this & merge?

marcelofabri:
https://github.com/jpsim/SourceKitten/issues/512 needs to be released as well

jpsim:
OK, will do.

jpsim:
SourceKitten 0.21.0: https://github.com/jpsim/SourceKitten/releases/tag/0.21.0

PR #2155, opened 2018-04-13
Created Weak View Controller Rules similar to Weak Delegate

coverback:
We noticed that it's often needed (and forgotted) to mark view controller properties as `weak`. This rule is very similar to `weak_delegate`.

### DIFF
--- a/Rules.md
+++ b/Rules.md
@@ -126,6 +126,7 @@
 * [Vertical Whitespace](#vertical-whitespace)
 * [Void Return](#void-return)
 * [Weak Delegate](#weak-delegate)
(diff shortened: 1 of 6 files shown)


marcelofabri:
Is there any particular reason why you've preferred to check the variable name and not the type? I think in this case it'd make more sense.

coverback:
@marcelofabri maybe both? I wanted to first get feedback if this seems useful to others at all. 

And then I also had an idea if a more generic rule like `weak_property` can be introduced with `name_regex` and `type_regex` as parameters, with `/delegate$/` as a default one and which can then deprecate both `weak_delegate` and this one?

PR #2164, opened 2018-04-20
Remove extensions from LowerACLThanBodyRule

keith:
Because of some oddities in the Swift extension ACL rules, linting ACL
inside extensions isn't trivial because of a few cases:

1. Extensions cannot be open, but members inside them can:

```
open extension Foo { // Not valid
    open func bar() {}
}
```

2. Extensions cannot have an ACL modifier if they conform to a protocol:

```
public extension Foo: Bar {} // Not valid
```

3. Extensions that do have an ACL modifier affects the inner body,
making changes to this slightly riskier:

```
public extension Foo {
    func bar() {} // implicitly public
}
```

With this change we just sidestep all these problems, and instead opt to
only lint non extension types. I think this tradeoff is worth it so that
this can be enabled and catch issues in classes/structs/enums

### DIFF
--- a/Source/SwiftLintFramework/Models/AccessControlLevel.swift
+++ b/Source/SwiftLintFramework/Models/AccessControlLevel.swift
@@ -46,3 +46,19 @@ public enum AccessControlLevel: String, CustomStringConvertible {
     }
 
 }
+
+extension AccessControlLevel: Comparable {
+    private var priority: Int {
+        switch self {
+        case .private: return 1
+        case .fileprivate: return 2
+        case .internal: return 3
+        case .public: return 4
+        case .open: return 5
+        }
+    }
+
+    public static func < (lhs: AccessControlLevel, rhs: AccessControlLevel) -> Bool {
+        return lhs.priority < rhs.priority
+    }
+}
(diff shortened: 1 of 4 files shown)


Source/SwiftLintFramework/Models/AccessControlLevel.swift
```diff
@@ -51,10 +51,10 @@ extension AccessControlLevel: Comparable {
...
         switch self {
         case .private: return 1
-        case .fileprivate: return 2
-        case .internal: return 3
-        case .public: return 4
-        case .open: return 5
+        case .fileprivate: return 1
+        case .internal: return 2
+        case .public: return 3
+        case .open: return 4
```
keith: this change might be a little too weird, in which case I can make this private to LowerACLThanBodyRule. For that use case private == fileprivate

keith: I added this Comparable conformance in my last PR, so I'm assuming this rule is still the only consumer, but I'm definitely happy to move it to be private

keith: I moved this into this rule's file

jpsim:
Thanks @keith. I re-added `Comparable` conformance to `AccessControlLevel` though since this was publicly added and is generally useful to consumers of SwiftLintFramework. Kept all your other changes.

keith:
Sounds good!

keith:
Thanks!


--
Keith Smiley

PR #2166, opened 2018-04-24
Add file name matching to file_header rule

nevil:
Add support of inserting a SWIFTLINT_CURRENT_FILENAME placeholder in the required_pattern.
The placeholder will be replaced by the lastPathComponent of the file's path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,8 @@
 #### Enhancements
 
 * Add optional filename verification to the `file_header` rule.
-  The first occurrence in the pattern of the `SWIFTLINT_CURRENT_FILENAME`
-  placeholder is replaced by the name of the validated file.  
+  All occurrences in the pattern of the `SWIFTLINT_CURRENT_FILENAME`
+  placeholder are replaced by the name of the validated file.  
   [Anders Hasselqvist](https://github.com/nevil)
   [#1079](https://github.com/realm/SwiftLint/issues/1079)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
 
 #### Enhancements
 
+* Add optional filename verification to the `file_header` rule.  
+  [Anders Hasselqvist](https://github.com/nevil)
+  [#1079](https://github.com/realm/SwiftLint/issues/1079)
+
 * Updates the `untyped_error_in_catch` rule to support autocorrection.  
   [Daniel Metzing](https://github.com/dirtydanee)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,9 @@
 
 #### Enhancements
 
-* Add optional filename verification to the `file_header` rule.  
+* Add optional filename verification to the `file_header` rule.
+  The first occurrence in the pattern of the `SWIFTLINT_CURRENT_FILENAME`
+  placeholder is replaced by the name of the validated file.  
   [Anders Hasselqvist](https://github.com/nevil)
   [#1079](https://github.com/realm/SwiftLint/issues/1079)
 
--- a/Source/SwiftLintFramework/Rules/FileHeaderRule.swift
+++ b/Source/SwiftLintFramework/Rules/FileHeaderRule.swift
@@ -53,6 +53,10 @@ public struct FileHeaderRule: ConfigurationProviderRule, OptInRule {
             lastToken = token
         }
 
+        // If we have a requiredPattern, replace the filename placeholder if needed
+        // and compile the regex.
+        let requiredRegex = replaceRequiredRegexIfNeeded(configuration, file: file)
+
         var violationsOffsets = [Int]()
         if let firstToken = firstToken, let lastToken = lastToken {
             let start = firstToken.offset
@@ -66,12 +70,12 @@ public struct FileHeaderRule: ConfigurationProviderRule, OptInRule {
                 violationsOffsets.append(firstMatch.range.location)
             }
 
-            if let regex = configuration.requiredRegex,
+            if let regex = requiredRegex,
                 case let matches = regex.matches(in: file.contents, options: [], range: range),
                 matches.isEmpty {
                 violationsOffsets.append(start)
             }
-        } else if configuration.requiredRegex != nil {
+        } else if requiredRegex != nil {
             let location = firstNonCommentToken.map {
                 Location(file: file, byteOffset: $0.offset)
             } ?? Location(file: file.path, line: 1)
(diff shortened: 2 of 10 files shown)


marcelofabri:
Can we do this in a dumber way like having a reserved string instead? It'd be important to avoid having different behaviors on different platforms.

We could use something like `SWIFTLINT_CURRENT_FILENAME`.

nevil:
@marcelofabri 
Thanks for your comment!

I was thinking of that approach but decided against it because the named capture group seemed more attractive to me. I did know that the multi-platform issue would be problematic...

I considered two approaches for using a reserved string:
1) FileHeaderConfiguration would replace SWIFTLINT_CURRENT_FILENAME with a filename regex pattern. The downside is that it is difficult to create a pattern that will work for all files.
But I guess something like ```\s*(.*?\..*?)\s*```.

2) FileHeaderRule would replace SWIFTLINT_CURRENT_FILENAME with the actual filename in the required_string and compile the regex.
The downside is that the regex will be compiled once per file so there is a small slowdown.
Maybe the performance impact is less of a concern than I was thinking?

Is approach 2) what you were thinking of?

nevil:
Hi @marcelofabri 

I believe I implemented along what you were thinking.
Please have a look when you have time.

Thanks!

nevil:
@marcelofabri 
@jpsim 

SwiftLint itself removed all file headers.
Is this feature still wanted, or should I close the pull request?

marcelofabri:
@nevil I think the feature is still useful even if SwiftLint itself doesn't use it

Source/SwiftLintFramework/Rules/RuleConfigurations/FileHeaderConfiguration.swift
```diff
@@ -54,6 +58,8 @@ public struct FileHeaderConfiguration: RuleConfiguration, Equatable {
                                                     options: [.ignoreMetacharacters])
         } else if let requiredPattern = configuration["required_pattern"] {
             self.requiredPattern = requiredPattern
+            // For requiredPattern the regex might be recompiled in FileHeaderRule
```
marcelofabri: Ideally, we'd support `SWIFTLINT_CURRENT_FILENAME` in `forbiddenPattern` as well

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 
 #### Enhancements
 
+* Add optional filename verification to the `file_header` rule.
```
marcelofabri: can you please add an explanation on `SWIFTLINT_CURRENT_FILENAME`? I think it'd be nice to also update the rule description to mention it.

Source/SwiftLintFramework/Rules/RuleConfigurations/FileHeaderConfiguration.swift
```diff
@@ -27,13 +27,17 @@ public struct FileHeaderConfiguration: RuleConfiguration, Equatable {
         }
     }
 
+    /// Used for unit tests of SWIFTLINT_CURRENT_FILENAME when File.path is nil
+    internal let filenameForTest: String = "Default.swift"
```
marcelofabri: what do you think about creating a fixture file for testing this instead? Check https://github.com/realm/SwiftLint/pull/2204 for an example.

Source/SwiftLintFramework/Rules/FileHeaderRule.swift
```diff
@@ -93,6 +97,26 @@ public struct FileHeaderRule: ConfigurationProviderRule, OptInRule {
...
+
+            // Replace SWIFTLINT_CURRENT_FILENAME with the filename.
+            let escapedName = NSRegularExpression.escapedPattern(for: fileName)
+            let replacedPattern = requiredPattern.replacingOccurrences(of: "SWIFTLINT_CURRENT_FILENAME",
+                                                                       with: escapedName)
+            do {
+                return try NSRegularExpression(pattern: replacedPattern,
+                                               options: [.anchorsMatchLines, .dotMatchesLineSeparators])
+            } catch {
+                queuedFatalError("Invalid pattern for required_pattern in 'file_header'")
```
marcelofabri: What do you think about printing the pattern in the error message?

nevil:
@marcelofabri 
Thanks for your comments! I agree with your suggestions and are working on the changes now.

Btw:
Do you think it would be an improvement if I:
- check for the existence of SWIFTLINT_CURRENT_FILENAME already in FileHeaderConfiguration
- remember the existence in a bool property
- Check this property in FileHeaderRule.replaceRequiredRegexIfNeeded() to shortcut and avoid unnecessary string traversals

?

marcelofabri:
@nevil I think it makes sense, but I'd create a method on the configuration instead of exposing the property, like `requiredRegex(for file: File)` and make the `requiredRegex` property private. Same thing for `forbiddenRegex`.

nevil:
@marcelofabri I addressed your comments and also added pattern support for required_string/forbidden_string.

Please let me know if I went wrong somewhere along the way :-)

Source/SwiftLintFramework/Rules/RuleConfigurations/FileHeaderConfiguration.swift
```diff
@@ -48,28 +38,99 @@ public struct FileHeaderConfiguration: RuleConfiguration, Equatable {
...
         if let severityString = configuration["severity"] {
             try severityConfiguration.apply(configuration: severityString)
         }
     }
+
+    private func makeRegex(for file: File, using pattern: String,
+                           options: NSRegularExpression.Options, escapeFileName: Bool) -> NSRegularExpression? {
+        // Recompile the regex for this file
+        guard let fileName = file.path?.bridge().lastPathComponent else {
+            queuedFatalError("Expected to validate a file.")
```
marcelofabri: Should we just return regex without replacing the pattern in this case? I think it'd be better to just fail this rule than to exit the whole program

Source/SwiftLintFramework/Rules/FileHeaderRule.swift
```diff
@@ -9,7 +9,9 @@ public struct FileHeaderRule: ConfigurationProviderRule, OptInRule {
     public static let description = RuleDescription(
         identifier: "file_header",
         name: "File Header",
-        description: "Header comments should be consistent with project patterns.",
+        description: "Header comments should be consistent with project patterns. " +
```
marcelofabri: Can you please provide the `reason` with the old `description` when creating `StyleViolation`s?

This way oss-check won't report diffs and it'll also be easier for people to read the warnings.

CHANGELOG.md
```diff
@@ -11,6 +11,12 @@
...
+  The first occurrence in the pattern of the `SWIFTLINT_CURRENT_FILENAME`
```
marcelofabri: we replace all occurrences, don't we?

marcelofabri:
@nevil Thanks for the update! This is looking great, I just added a few minor comments.

CHANGELOG.md
nevil: Doh!
Sorry for that...

Source/SwiftLintFramework/Rules/RuleConfigurations/FileHeaderConfiguration.swift
nevil: Sure. I wasn't sure what to do here...

Source/SwiftLintFramework/Rules/FileHeaderRule.swift
nevil: I was concerned about all those diffs reported by ops-check.
Thanks for teaching me how to fix!

Btw: Had to do small reformat to avoid function_body_length error due to the two extra "reason" lines added in ```validate(file:)```.

nevil:
@marcelofabri Thanks for guiding me through this PR!

PR #2167, opened 2018-04-24
Add `indented_cases` to `switch_case_alignment` rule

### LINKED ISSUE #2119: Rule Request: Opposite of SwitchCaseAlignmentRule
### New Issue Checklist

### Rule Request

In my codebase (and others i've worked on) I like switch cases to be indented one level extra than the switch itself (i've known others to do the same also), e.g:

```swift
switch enumThing {
    case firstCase:
        // do stuff
    default:
        // do stuff
}
```

This could be implemented as a separate opt-in rule or as a configuration option on the existing `SwitchCaseAlignmentRule`.

freak4pc:
Fixes #2119. 

This PR adds the following configuration: 

```yaml
switch_case_alignment:
    indented_cases: true
```

Thanks and appreciate your review :)
Shai.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,15 +8,16 @@
 
 #### Enhancements
 
+* Add `indented_cases` support to `switch_case_alignment` rule.  
+  [Shai Mishali](https://github.com/freak4pc)
+  [#2119](https://github.com/realm/SwiftLint/issues/2119)
+
 * Add opt-in `modifier_order` to enforce the order of declaration modifiers.
   Requires Swift 4.1 or later.  
   [Jose Cheyo Jimenez](https://github.com/masters3d)
   [Daniel Metzing](https://github.com/dirtydanee)
   [#1472](https://github.com/realm/SwiftLint/issues/1472)
   [#1585](https://github.com/realm/SwiftLint/issues/1585)
-* Add `indented_cases` support to `switch_case_alignment` rule.
-  [Shai Mishali](https://github.com/freak4pc)
-  [#2119](https://github.com/realm/SwiftLint/issues/2119)
 
 * Validate implicit `subscript` getter in `implicit_getter` rule when using
   Swift 4.1 or later.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@
   [Daniel Metzing](https://github.com/dirtydanee)
   [#1472](https://github.com/realm/SwiftLint/issues/1472)
   [#1585](https://github.com/realm/SwiftLint/issues/1585)
+* Add `indented_cases` support to `switch_case_alignment` rule.
+  [Shai Mishali](https://github.com/freak4pc)
+  [#2119](https://github.com/realm/SwiftLint/issues/2119)
 
 * Validate implicit `subscript` getter in `implicit_getter` rule when using
   Swift 4.1 or later.  
--- a/Source/SwiftLintFramework/Rules/RuleConfigurations/SwitchCaseAlignmentConfiguration.swift
+++ b/Source/SwiftLintFramework/Rules/RuleConfigurations/SwitchCaseAlignmentConfiguration.swift
@@ -16,6 +16,20 @@ public struct SwitchCaseAlignmentConfiguration: RuleConfiguration, Equatable {
         return severityConfiguration.consoleDescription + ", indented_cases: \(indentedCases)"
     }
 
+    public var ruleDescription: RuleDescription {
+        let examples = SwitchCaseAlignmentRule.Examples(indentedCases: indentedCases)
+
+        let reason = indentedCases ? "Case statements should be indented within their enclosing switch statement"
+                                   : "Case statements should vertically align with their enclosing switch statement"
+
+        return RuleDescription(identifier: "switch_case_alignment",
+                               name: "Switch and Case Statement Alignment",
+                               description: reason,
+                               kind: .style,
+                               nonTriggeringExamples: examples.nonTriggeringExamples,
+                               triggeringExamples: examples.triggeringExamples)
+    }
+
     public mutating func apply(configuration: Any) throws {
         guard let configuration = configuration as? [String: Any] else {
             throw ConfigurationError.unknownConfiguration
--- a/Source/SwiftLintFramework/Rules/RuleConfigurations/SwitchCaseAlignmentConfiguration.swift
+++ b/Source/SwiftLintFramework/Rules/RuleConfigurations/SwitchCaseAlignmentConfiguration.swift
@@ -1,0 +1,32 @@
+//
+//  SwitchCaseAlignmentRuleConfiguration.swift
+//  SwiftLint
+//
+//  Created by Shai Mishali on 4/23/18.
+//  Copyright © 2018 Realm. All rights reserved.
+//
+
+public struct SwitchCaseAlignmentConfiguration: RuleConfiguration, Equatable {
+    private(set) var severityConfiguration = SeverityConfiguration(.warning)
+    private(set) var indentation: Int = 0
+
+    init() {}
+
+    public var consoleDescription: String {
+        return severityConfiguration.consoleDescription + ", indentation: \(indentation)"
+    }
+
+    public mutating func apply(configuration: Any) throws {
+        guard let configuration = configuration as? [String: Any] else {
+            throw ConfigurationError.unknownConfiguration
+        }
+
+        self.indentation = configuration["indentation"] as? Int ?? 0
+    }
+
+    public static func == (lhs: SwitchCaseAlignmentConfiguration,
+                           rhs: SwitchCaseAlignmentConfiguration) -> Bool {
+        return lhs.indentation == rhs.indentation &&
+               lhs.severityConfiguration == rhs.severityConfiguration
+    }
+}
--- a/Source/SwiftLintFramework/Rules/RuleConfigurations/SwitchCaseAlignmentConfiguration.swift
+++ b/Source/SwiftLintFramework/Rules/RuleConfigurations/SwitchCaseAlignmentConfiguration.swift
@@ -13,7 +13,7 @@ public struct SwitchCaseAlignmentConfiguration: RuleConfiguration, Equatable {
     init() {}
 
     public var consoleDescription: String {
-        return severityConfiguration.consoleDescription + ", indented cases: \(indentedCases)"
+        return severityConfiguration.consoleDescription + ", indented_cases: \(indentedCases)"
     }
 
     public mutating func apply(configuration: Any) throws {
--- a/Source/SwiftLintFramework/Rules/RuleConfigurations/SwitchCaseAlignmentConfiguration.swift
+++ b/Source/SwiftLintFramework/Rules/RuleConfigurations/SwitchCaseAlignmentConfiguration.swift
@@ -8,12 +8,12 @@
 
 public struct SwitchCaseAlignmentConfiguration: RuleConfiguration, Equatable {
     private(set) var severityConfiguration = SeverityConfiguration(.warning)
-    private(set) var indentation: Int = 0
+    private(set) var indentedCases = false
 
     init() {}
 
     public var consoleDescription: String {
-        return severityConfiguration.consoleDescription + ", indentation: \(indentation)"
+        return severityConfiguration.consoleDescription + ", indented cases: \(indentedCases)"
     }
 
     public mutating func apply(configuration: Any) throws {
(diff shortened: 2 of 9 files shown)


marcelofabri:
> Is using a numeric value for indentation acceptable? Is there a better way you can think of to try here?

We have a top level `indentation` key but that's not really super documented or available for the rules. I think using an `indentation` is not the worst thing, but we could also have a boolean like `indent_cases` and we could pick the first case indentation as the "correct" one (as long as it's more indented than the `switch`).

> For triggering and non-triggering examples, I'm not sure its achievable with the RuleDescription stuff (since it flips the triggering and non-triggering examples based on the flag). What's the usual course of action in this case? Adding regular unit tests instead?

That's one of the reasons why sometimes we create rules for things that could be configurations. 

In this case I think having a configuration is fine, so I'd just create a new unit test class, move the existing test to that file and adding new ones using the helpers that we have to replace examples (check `TypeNameRuleTests.swift`)

freak4pc:
Thanks for the tips. I switched to using `indented_cases: true` instead :) 

Could you help with why these two tests are failing? I based this on the `TypeNameRuleTests.swift` you suggested: 

https://github.com/realm/SwiftLint/pull/2167/commits/1248d03a7b133ad423385aac694a6bfc08fe8da3#diff-737cc0107444a3b6a33ca60a1227f333R35

freak4pc:
All green 💚 
Ready for your review when you have time @marcelofabri 

Only question I had open was dynamically changing the rule's description and triggering/non-triggeting examples (based on the config) but I believes that's impossible with the current structure as it's static. Let me know if you have any thoughts on this.

Source/SwiftLintFramework/Rules/SwitchCaseAlignmentRule.swift
```diff
@@ -99,18 +45,124 @@ public struct SwitchCaseAlignmentRule: ASTRule, ConfigurationProviderRule {
...
-                StyleViolation(ruleDescription: type(of: self).description,
-                               severity: configuration.severity,
-                               location: $0)
-            }
+            .filter { $0.character != switchCharacter + indentation }
+            .map(locationToViolation)
+    }
+
+    private func locationToViolation(_ location: Location) -> StyleViolation {
+        return StyleViolation(ruleDescription: type(of: self).description,
```
marcelofabri: can we add a `reason` argument and keep the old reason if `indentedCases == false`? Then we could also use a more specific reason when it's true (e.g. `"Case statements should be indented in relation to the enclosing switch statement"`).

This would also make it easier to review oss-check violations.

Source/SwiftLintFramework/Rules/SwitchCaseAlignmentRule.swift
```diff
@@ -99,18 +45,124 @@ public struct SwitchCaseAlignmentRule: ASTRule, ConfigurationProviderRule {
...
+                              severity: configuration.severityConfiguration.severity,
+                              location: location)
+    }
+}
+
+public extension SwitchCaseAlignmentRule {
```
marcelofabri: this probably should be `internal`

Source/SwiftLintFramework/Rules/SwitchCaseAlignmentRule.swift
freak4pc: Great idea!

Source/SwiftLintFramework/Rules/SwitchCaseAlignmentRule.swift
freak4pc: @marcelofabri I actually think this has to remain public or the Test target won't be able to access the examples.

I'm defining the ACL on the extension instead of the internal struct itself.

marcelofabri: You can use `@testable import SwiftLintFramework` on the test class

freak4pc: Oh duh! Done.

freak4pc:
@marcelofabri Updated.

I also further abstracted the triggering and non-triggering cases in `SwitchCaseAlignmentRule.Examples` based on the configuration so its less confusing to manually select the indented vs. non-indented examples.

Noobish1:
Looks great @freak4pc :) thanks for doing this.

Source/SwiftLintFramework/Rules/SwitchCaseAlignmentRule.swift
```diff
@@ -99,18 +45,132 @@ public struct SwitchCaseAlignmentRule: ASTRule, ConfigurationProviderRule {
...
-                StyleViolation(ruleDescription: type(of: self).description,
-                               severity: configuration.severity,
-                               location: $0)
-            }
+            .filter { $0.character != switchCharacter + indentation }
+            .map(locationToViolation)
+    }
+
+    private func locationToViolation(_ location: Location) -> StyleViolation {
+        return StyleViolation(ruleDescription: configuration.ruleDescription,
```
marcelofabri: there's an optional param named `reason` that can be used instead of duplicating the description in the configuration

Tests/SwiftLintFrameworkTests/SwitchCaseAlignmentRuleTests.swift
```diff
@@ -0,0 +1,36 @@
...
+//
+//  Created by Shai Mishali on 4/24/18.
+//  Copyright © 2018 Realm. All rights reserved.
+//
+
+@testable import SwiftLintFramework
+import XCTest
+
+class SwitchCaseAlignmentRuleTests: XCTestCase {
+    func testSwitchCaseAlignment() {
```
marcelofabri: you can remove `testSwitchCaseAlignment` from `RulesTests` to avoid testing this twice

Source/SwiftLintFramework/Rules/RuleConfigurations/SwitchCaseAlignmentConfiguration.swift
```diff
@@ -0,0 +1,46 @@
...
+                               nonTriggeringExamples: examples.nonTriggeringExamples,
+                               triggeringExamples: examples.triggeringExamples)
+    }
+
+    public mutating func apply(configuration: Any) throws {
+        guard let configuration = configuration as? [String: Any] else {
+            throw ConfigurationError.unknownConfiguration
+        }
+
+        self.indentedCases = configuration["indented_cases"] as? Bool ?? false
```
marcelofabri: this is not applying the severity configuration 🤔 

Check `ColonConfiguration` for example. It'd be good to have unit tests for this class as well if possible

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 #### Enhancements
 
+* Add `indented_cases` support to `switch_case_alignment` rule.
```

Source/SwiftLintFramework/Rules/SwitchCaseAlignmentRule.swift
```diff
@@ -99,18 +45,132 @@ public struct SwitchCaseAlignmentRule: ASTRule, ConfigurationProviderRule {
...
         }
 
+        guard let firstCase = caseLocations.first,
+              let firstCaseCharacter = firstCase.character else {
+            return []
+        }
+
+        // If indent_cases is on, the first case should be indented from its containing switch.
+        if configuration.indentedCases, firstCaseCharacter <= switchCharacter {
+            return caseLocations.map(locationToViolation)
```
marcelofabri: do any of the examples trigger this case? Just by looking at them, it doesn't look like it.

freak4pc: I believe all of them do as the triggering examples are the non-indented examples. 
what's probably not covered by the tests would be following lines - 

e.g. if the first line is ok but the following cases aren't indented (as it would fall in this condition)

Tests/SwiftLintFrameworkTests/SwitchCaseAlignmentRuleTests.swift
freak4pc: What do you mean by this exactly? I sort of "Stole" this from a different rule (vertical spacing IMO)

marcelofabri: There's already a test that test exactly this on `RulesTests.swift`. You should remove it from there to avoid testing the same thing twice

jpsim:
> Sorry for the minor delay in getting to this, hope to get to it in the middle of the week

👍 looking forward to this!

freak4pc:
Gonna get to this today or tomorrow !

freak4pc:
@marcelofabri I think this should cover all of the notes you left + rebases on top of master. Let me know if I missed anything. 

Thanks !

freak4pc:
Uhm the `linux_swift_4` step is failing on an unrelated rule. Quite strange: 

> TestHelpers.swift:270: error: TrailingCommaRuleTests.testTrailingCommaRuleWithDefaultConfiguration : XCTAssertEqual failed: ("2") is not equal to ("1") -

jpsim:
Oops, looks like `master` is actually broken.

freak4pc:
Seems like a Swift 4.0 issues specifically

jpsim:
Fix in #2184.

jpsim:
If you rebase now you'll include the fix.

Source/SwiftLintFramework/Rules/SwitchCaseAlignmentRule.swift
```diff
@@ -91,18 +37,163 @@ public struct SwitchCaseAlignmentRule: ASTRule, ConfigurationProviderRule {
...
-                let (line, char) = contents.lineAndCharacter(forByteOffset: byteOffset) else {
-                    return nil
+                  let (line, char) = contents.lineAndCharacter(forByteOffset: byteOffset) else {
+                return nil
             }
+
             return Location(file: file.path, line: line, character: char)
         }
 
+        guard let firstCase = caseLocations.first,
```
jpsim: Since we don't use `firstCase` anywhere:

```swift
guard let firstCaseCharacter = caseLocations.first?.character else {
    return []
}

File: Source/SwiftLintFramework/Rules/SwitchCaseAlignmentRule.swift [hunk 9]
```diff
@@ -91,18 +37,163 @@ public struct SwitchCaseAlignmentRule: ASTRule, ConfigurationProviderRule {
...
-                StyleViolation(ruleDescription: type(of: self).description,
-                               severity: configuration.severity,
-                               location: $0)
-            }
+            .filter { $0.character != switchCharacter + indentation }
+            .map(locationToViolation)
+    }
+
+    private func locationToViolation(_ location: Location) -> StyleViolation {
+        // swiftlint:disable line_length
```
jpsim: How about this to avoid having to disable the line length rule?

```swift
private func locationToViolation(_ location: Location) -> StyleViolation {
    let reason = """
        Case statements should \
        \(configuration.indentedCases ? "be indented within" : "vertically align with") \
        their enclosing switch statement
        """

    return StyleViolation(ruleDescription: SwitchCaseAlignmentRule.description,
                          severity: configuration.severityConfiguration.severity,
                          location: location,
                          reason: reason)
}
```

Source/SwiftLintFramework/Rules/SwitchCaseAlignmentRule.swift
```diff
@@ -91,18 +37,163 @@ public struct SwitchCaseAlignmentRule: ASTRule, ConfigurationProviderRule {
...
+                              severity: configuration.severityConfiguration.severity,
+                              location: location,
+                              reason: reason)
+    }
+}
+
+extension SwitchCaseAlignmentRule {
+    struct Examples {
+        private let indentedCasesOption: Bool
+        private let violationMarker = "↓"
```
jpsim: Not in this PR, but we should find a way to expose this (preferably not in the public API) for general use in this project.

freak4pc: You Jean the examples themselves ? Expose in what way ?

jpsim: I mean the violation marker (`↓`). Not for the inline use when in triggering/nontriggering example markup (since those uses are designed to be quick to visually grok) but rather for times when we either parse or dynamically insert violation markers.

freak4pc: Yeah I agree - it was a tad strange to copy-paste the marker everywhere but it wasn't available in any internal-scoped place in the module. We can look into it later for sure 👍

Source/SwiftLintFramework/Rules/SwitchCaseAlignmentRule.swift
```diff
@@ -62,10 +61,11 @@ public struct SwitchCaseAlignmentRule: ASTRule, ConfigurationProviderRule {
     }
 
     private func locationToViolation(_ location: Location) -> StyleViolation {
-        // swiftlint:disable line_length
-        let reason = configuration.indentedCases ? "Case statements should be indented within their enclosing switch statement"
-                                                 : "Case statements should vertically align with their enclosing switch statement"
-        // swiftlint:enable line_length
+        let reason = """
+                    Case statements should
```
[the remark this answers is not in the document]
freak4pc: I actually think it should work just fine but I can verify tomorrow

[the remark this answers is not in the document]
jpsim: It was definitely adding newlines. I fixed it in 4bf69d7.

jpsim:
🎉 well done all!

PR #2170, opened 2018-04-24
Fix ControlStatement rule failing to apply when clauses contain commas

allewun:
This scenario wasn't being caught in my codebase:

```swift
if (max(a, b) < c) {
    ...
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,11 @@
 
 #### Bug Fixes
 
+* Fixes an issue with `control_statement` where commas in clauses prevent the
+  rule from applying.  
+  [Allen Wu](https://github.com/allewun)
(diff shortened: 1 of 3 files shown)


jpsim:
@allewun thanks for this! Although you're right that this expands the rule to catch previously missed violations, it also stops detecting some previously detected violations. See all the "This PR fixed a violation" entries in the @SwiftLintBot comment above.

allewun:
@jpsim thanks, I'll take a look at those.

allewun:
@jpsim I think it's a false positive. I added one more test case which reflects the "fixed violations" just to be sure. The violation description has changed, so is that what `oss-check` is picking up? The PR "fixes" the violations because they don't exist by that description anymore.

jpsim:
Sorry for the delay on this. Looks good! 👌

PR #2173, opened 2018-04-26
Fix a typo

BasThomas:


marcelofabri:
@BasThomas that file is auto generated, Please update the text in `ArrayInitRule` and run the unit tests to update the docs 😉

Issue #2106, opened 2018-03-20
SwiftLint Contributor Copyright Header


dabelknap:
Hi, we'd love to contribute patches to your project. But our company does not allow us to put another's copyright header on our work. Here's what our company would allow, does either work for you?

a)
```
//  Copyright © 2018 The SwiftLint Authors. All rights reserved.
```

b)
```
//  Copyright © 2017 Realm. All rights reserved.
//  Copyright 2018 Google LLC
```

jpsim:
Hi @dabelknap, thanks for your interest in contributing! Your suggestions seem very reasonable to me, but as I no longer work at Realm, I've reached out to someone there to see if this would be possible.

jpsim:
@dabelknap can you confirm that removing the file copyright header throughout the project would enable you to contribute?

ps: I'd love to hear what kind of contributions you have in mind 😄.

jpsim:
I have a PR up to do so here: #2179

dabelknap:
@jpsim Yep, removing the headers will work for us! Thanks!

PR #2179, opened 2018-05-04
Remove all file headers

jpsim:
The MIT license doesn't require that all files be prepended with this licensing or copyright information. Realm confirmed that they're ok with this change. This will enable some companies to contribute to SwiftLint and the date & authorship information will remain accessible via git source control.

Addresses #2106.

### DIFF
--- a/.sourcery/LinuxMain.stencil
+++ b/.sourcery/LinuxMain.stencil
@@ -1,11 +1,3 @@
-//
-//  LinuxMain.swift
-//  SwiftLint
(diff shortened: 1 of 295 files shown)


marcelofabri:
Should we use the header_pattern rule to avoid headers entirely?

jpsim:
> Should we use the header_pattern rule to avoid headers entirely?

I haven't found a good way to do this while still allowing the Sourcery header.

PR #2183, opened 2018-05-06
Fix false positive in redundant_set_access_control

### LINKED ISSUE #2182: False positive in Redundant Set Access Control Rule
### New Issue Checklist

- [x] Updated SwiftLint to the latest version (master)

### Bug Report

Private has different meanings when at the top-level vs nested. Top-level effectively means "fileprivate" whereas nested means "private within this scope".

Repro:
```swift
private final class A {
    private(set) var value: Int
    ^ warning: Redundant Set Access Control Rule Violation: Property setter access level shouldn't be explicit if it's the same as the variable access
}
```

In this case, the code is acceptable and we shouldn't require `fileprivate final class A`.

This rule was added in #2036.

marcelofabri:
Fixes #2182 

Not adding a changelog entry because this rule hasn't been officially released it

### DIFF
--- a/Rules.md
+++ b/Rules.md
@@ -12506,6 +12506,12 @@ public var foo: Int
 var foo: Int
 ```
 
+```swift
+private final class A {
+    private(set) var value: Int
+}
+```
+
 </details>
 <details>
 <summary>Triggering Examples</summary>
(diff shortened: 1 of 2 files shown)


jpsim:
Is there a way to keep the violation on fileprivate/fileprivate?

marcelofabri:
@jpsim I think I've found a way! Let's see if osscheck finds anything interesting.

jpsim:
Smart! Well done! :clap:

PR #2185, opened 2018-05-07
Add function_default_parameter_at_end rule

### LINKED ISSUE #2176: Rule request: default parameters at the end
### New Issue Checklist

### Rule Request

From [Swift's API Design Guidelines](https://swift.org/documentation/api-design-guidelines/):

> Prefer to locate parameters with defaults toward the end of the parameter list. Parameters without defaults are usually more essential to the semantics of a method, and provide a stable initial pattern of use where methods are invoked.

Thanks @mokagio for pointing this potential rule [on Twitter](https://twitter.com/mokagio/status/991626605817548801).

```swift
// would trigger
func foo(bar: Int = 0, baz: String) {}

// wouldn't trigger
func foo(baz: String, bar: Int = 0) {}
```

Just severity. 

   rule as opt-in.

Probably enabled by default, depends on how many violation oss-check finds.

marcelofabri:
Fixes #2176

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,12 @@
   [Ornithologist Coder](https://github.com/ornithocoder)
   [#2190](https://github.com/realm/SwiftLint/pull/2190)
 
+* Add `function_default_parameter_at_end` opt-in rule to validate that
+  parameters with defaults are located toward the end of the parameter list in a
+  function declaration.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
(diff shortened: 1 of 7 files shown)


SDGGiesbrecht:
👍 You even made the trailing closure stuff work succinctly. When I mentioned it, I did not know for sure if it was actually feasible.

marcelofabri:
@SDGGiesbrecht It may still give false positives when closures are used via `typealias` and are not `@escaping`, but I thought it was worth the effort. Thanks for pointing that out 👍

PR #2189, opened 2018-05-08
Add empty_xctest_method opt-in rule

ornithocoder:
Implements #2190.

### DIFF
--- a/Source/SwiftLintFramework/Rules/EmptyXCTestMethodRule.swift
+++ b/Source/SwiftLintFramework/Rules/EmptyXCTestMethodRule.swift
@@ -1,0 +1,61 @@
+import Foundation
+import SourceKittenFramework
+
+public struct EmptyXCTestMethodRule: Rule, OptInRule, ConfigurationProviderRule {
+
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "empty_xctest_method",
+        name: "Empty XCTest Method",
+        description: "Empty XCTest method should be avoided.",
+        kind: .idiomatic,
+        nonTriggeringExamples: EmptyXCTestMethodRuleExamples.nonTriggeringExamples,
+        triggeringExamples: EmptyXCTestMethodRuleExamples.triggeringExamples
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        return testClasses(in: file).flatMap { violations(in: file, for: $0) }
+    }
+
+    // MARK: - Private
+
+    private func testClasses(in file: File) -> [[String: SourceKitRepresentable]] {
+        return file.structure.dictionary.substructure.filter { dictionary in
+            guard
+                let kind = dictionary.kind,
+                SwiftDeclarationKind(rawValue: kind) == .class else { return false }
+
+            return !dictionary.inheritedTypes.filter { $0 == "XCTestCase" }.isEmpty
+        }
+    }
+
+    private func violations(in file: File,
+                            for dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        return dictionary.substructure.compactMap { subDictionary -> StyleViolation? in
+            guard
+                let kind = subDictionary.kind,
+                let swiftKind = SwiftDeclarationKind(rawValue: kind),
+                SwiftDeclarationKind.functionKinds.contains(swiftKind),
+                let name = subDictionary.name, isXCTestMethod(name),
+                let offset = subDictionary.offset,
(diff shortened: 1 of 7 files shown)


Source/SwiftLintFramework/Rules/EmptyXCTestMethodRule.swift
```diff
@@ -0,0 +1,61 @@
...
+
+    private func violations(in file: File,
+                            for dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        return dictionary.substructure.compactMap { subDictionary -> StyleViolation? in
+            guard
+                let kind = subDictionary.kind,
+                let swiftKind = SwiftDeclarationKind(rawValue: kind),
+                SwiftDeclarationKind.functionKinds.contains(swiftKind),
+                let name = subDictionary.name, isXCTestMethod(name),
+                let offset = subDictionary.offset,
```
marcelofabri: can we also test that the method has no parameters just to be safe?

ornithocoder: Done. Added here: https://github.com/realm/SwiftLint/pull/2189/commits/7cef22b968c518232cd64d26d032e4392a4b18cc

Source/SwiftLintFramework/Rules/EmptyXCTestMethodRule.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "empty_xctest_method",
+        name: "Empty XCTest Method",
+        description: "Empty XCTest method should be avoided.",
+        kind: .idiomatic,
```
marcelofabri: this probably should be `lint`

Source/SwiftLintFramework/Rules/EmptyXCTestMethodRule.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    // MARK: - Private
+
+    private func testClasses(in file: File) -> [[String: SourceKitRepresentable]] {
+        return file.structure.dictionary.substructure.filter { dictionary in
+            guard
+                let kind = dictionary.kind,
+                SwiftDeclarationKind(rawValue: kind) == .class else { return false }
+
+            return !dictionary.inheritedTypes.filter { $0 == "XCTestCase" }.isEmpty
```
marcelofabri: `contains`?

Source/SwiftLintFramework/Rules/EmptyXCTestMethodRule.swift
```diff
@@ -0,0 +1,62 @@
...
+                SwiftDeclarationKind.functionKinds.contains(swiftKind),
+                let name = subDictionary.name, isXCTestMethod(name),
+                subDictionary.enclosedVarParameters.isEmpty,
+                let offset = subDictionary.offset,
+                let bodyOffset = subDictionary.bodyOffset,
+                let bodyLength = subDictionary.bodyLength,
+                case let bodyContent = file.contents.bridge(),
+                let startLine = bodyContent.lineAndCharacter(forByteOffset: bodyOffset)?.line,
+                let endLine = bodyContent.lineAndCharacter(forByteOffset: bodyOffset + bodyLength)?.line,
+                case let (_, lineCount) = file.exceedsLineCountExcludingCommentsAndWhitespace(startLine, endLine, 0),
```
marcelofabri: I wonder if it'd be better to check the substructure instead 🤔 

This would probably trigger in something like this:

```swift
func testFoo() { XCTAssertFalse(true) }
```

Source/SwiftLintFramework/Rules/EmptyXCTestMethodRule.swift
ornithocoder: Done.

Source/SwiftLintFramework/Rules/EmptyXCTestMethodRule.swift
ornithocoder: Done.

Source/SwiftLintFramework/Rules/EmptyXCTestMethodRule.swift
ornithocoder: Yep, much better. Thanks!

marcelofabri:
Thanks for another great contribution @ornithocoder 🎉

ornithocoder:
Thank you, @marcelofabri!

PR #2192, opened 2018-05-09
Make closure end indentation rule autocorrectable

erichoracek:
I made an attempt at writing an exhaustive test suite, but would love any additional help on making sure this works in all cases. 

Thanks in advance for the review!


jpsim:
Nice to see this rule become correctable! You'll need to use `compactMap` on `ClosureEndIndentationRule.swift:215` to fix the deprecation warning on Swift 4.1. I'll take a look at the rest shortly.

erichoracek:
Good catch! Just pushed up the `compactMap` change.

PR #2195, opened 2018-05-10
Added ignores_default_parameters when calculating parameter count

varunpm1:


jpsim:
Thanks for the PR! Code looks great, I made a small edit to the changelog entry.

PR #2196, opened 2018-05-10
Fix operator usage spacing nested generics false positive

erichoracek:
This was being corrected to:
```swift
let model = Generic<Container<Button>, NSAttributedString > (
```


jpsim:
Thanks for the fix! PS: We generally don't include links to PRs in the changelog, just issues. I'm too lazy to remove it in this case though. 👍

PR #2208, opened 2018-05-16
Improve mark rule autocorrection

erichoracek:
Adds a few additional cases for the MARK rule to autocorrect.

In our project, this made it so that the all mark rule failures were corrected automatically. Before this change, we had over 30 failures that could not be autocorrected.


jpsim:
👍 will merge.

However, we should refactor this rule to avoid regular expressions. The use of regex in SwiftLint has become somewhat of a bottleneck that would be nice to eliminate.

Issue #2193, opened 2018-05-09
Rule Request: Redundant ObjC attribute


jpsim:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Rule Request

We should warn if you have `@objc` along with `@IBInspectable`/`@IBDesignable`/`@IBOutlet`/`@IBAction`, since they all imply `@objc`. Let's confirm this is the case unilaterally before doing this.

##### 1. Why should this rule be added? Share links to existing discussion about what the community thinks about this.

No external discussions about this that I'm aware of.

##### 2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

```swift
// Trigger
@objc @IBInspectable private var rawStringValue: String? { ... }
@IBInspectable @objc private var rawStringValue: String? { ... }
@objc @IBAction private func didPress(_ sender: Any) { ... }
@IBAction @objc private func didPress(_ sender: Any) { ... }
// Wouldn't trigger
@objc private var rawStringValue: String? { ... }
@IBInspectable private var rawStringValue: String? { ... }
@objc private func didPress(_ sender: Any) { ... }
@IBAction private func didPress(_ sender: Any) { ... }
```

##### 3. Should the rule be configurable, if so what parameters should be configurable?

I can't think of any useful configurations at the moment, other than the standard severity configuration.

##### 4. Should the rule be opt-in or enabled by default? Why? See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a rule as opt-in.

Enabled by default if we can confirm that these attribute pairs are redundant in all cases. Opt-in otherwise.

marcelofabri:
This probably should also catch uses of `@objc` inside a structure with `@objcMembers`.

dirtydanee:
If you guys don't mind, i would pick up and implement this rule. 

Just to visualize what @marcelofabri meant with his last comment would be:
```
// Trigger
@objcMembers
class Foo {
    @objc var bar: Any
}

// Not trigger
@objcMembers
class Foo {
    var bar: Any
}
```

Also, would it be okay, if it would be Swift 4.1+ rule?

marcelofabri:
@dirtydanee As far as I know, nobody is working on this yet, so feel free to grab.

[...]

PR #2218, opened 2018-05-22
Redundant objc attribute

dirtydanee:
Implements https://github.com/realm/SwiftLint/issues/2193

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,17 +2,66 @@
 
 #### Breaking
 
+* None.
+
+#### Enhancements
+
+* Decrease default severity of `superfluous_disable_command` to `warning`.  
+  [Frederick Pietschmann](https://github.com/fredpi)
+  [#2250](https://github.com/realm/SwiftLint/issues/2250)
+
+* Don't touch files when running `autocorrect --format` if the contents haven't
+  changed.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#2249](https://github.com/realm/SwiftLint/issues/2249)
+
+#### Bug Fixes
+
+* Fix an issue with `control_statement` where commas in clauses prevented the
+  rule from applying.  
+  [Allen Wu](https://github.com/allewun)
+
+* Fix `explicit_enum_raw_value`, `generic_type_name`, `implicit_return`,
+  `required_enum_case`, `quick_discouraged_call`, `array_init`,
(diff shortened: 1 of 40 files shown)


Source/SwiftLintFramework/Rules/RedundantObjcAttributeRule.swift
```diff
@@ -0,0 +1,198 @@
...
+public struct RedundantObjcAttributeRule: ASTRule, ConfigurationProviderRule {
+
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "redundant_objc_attribute",
+        name: "Redundant @objc Attribute",
+        description: "Objective C attribute is redundant in declaration.",
```
marcelofabri: what do you think about: `Objective-C  attribute (@objc) is redundant in declaration.`?

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 
 #### Enhancements
 
+* Add `redundant_objc_attribute` to warn against declaring Objective C visibility with attributes already impling the objc attribute.
```
marcelofabri: this need to be wrapped in 80 chars

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
 
 * Updates the `untyped_error_in_catch` rule to support autocorrection.  
   [Daniel Metzing](https://github.com/dirtydanee)
+* Add `redundant_objc_attribute` to warn against declaring Objective C visibility with attributes already impling the objc attribute.
```
Coder-256: implying.

Coder-256:
What about `@objcMembers` classes? You should make sure that none of their members have `@objc` because it's already implied, like `@IBInspectable`, etc. except on the class level.

dirtydanee:
@Coder-256 the `@objcMembers` are covered in the rule, like in this triggering example:
```
@objcMembers
class Foo {
   @objc ↓var bar: Any?
   @objc ↓var foo: Any?
   @objc
   class Bar {
       @objc
       var foo: Any?
   }
}
```

dirtydanee:
@marcelofabri Are there more updates required in my implementation?

dirtydanee:
I have managed to completely ruin this branch with my last rebase. Declining PR.

PR #2219, opened 2018-05-22
Making `UntypedErrorInCatchRule` correctable

dirtydanee:
No issue link, just made the existing `UntypedErrorInCatchRule` correctable.

### DIFF
--- a/Source/SwiftLintFramework/Rules/UntypedErrorInCatchRule.swift
+++ b/Source/SwiftLintFramework/Rules/UntypedErrorInCatchRule.swift
@@ -49,17 +49,48 @@ public struct UntypedErrorInCatchRule: OptInRule, ConfigurationProviderRule {
             "do {\n    try foo() \n} ↓catch let e {}",
             "do {\n    try foo() \n} ↓catch(let error) {}",
             "do {\n    try foo() \n} ↓catch (let error) {}"
+        ],
+        corrections: [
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Rules/UntypedErrorInCatchRule.swift
```diff
@@ -49,17 +49,48 @@ public struct UntypedErrorInCatchRule: OptInRule, ConfigurationProviderRule {
...
+
+    fileprivate func violationRanges(in file: File) -> [NSRange] {
+        return file.match(pattern: type(of: self).regularExpression,
+                          with: [.keyword, .keyword, .identifier])
+    }
+}
+
+extension UntypedErrorInCatchRule: CorrectableRule {
+    public func correct(file: File) -> [Correction] {
+         let violations = violationRanges(in: file)
```
marcelofabri: indentation seems off

marcelofabri:
Thanks! 🎉

dirtydanee:
@jpsim good catch, it could lead to cases like that. 

I could update the auto correction only to cases where the variable is `let error`. That should not lead to compilation issues. What do you think?

jpsim:
> I could update the auto correction only to cases where the variable is `let error`. That should not lead to compilation issues. What do you think?

Yes please!

Issue #2214, opened 2018-05-18
Rule Request: [Multiline Function Chains]


erichoracek:
#### Summary

Chained function calls should be either on the same line, or one per line.

This is similar to the existing "Multiline Arguments" and "Multiline Parameters" rules.

#### Q&A

> 1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.

It's easier to follow control flow through long function chains when each call has the same indentation.

> 2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

* Triggering:
```swift
return match(pattern: pattern, with: [.comment]).flatMap { range in
   return Command(string: contents, range: range)
}.flatMap { command in
   return command.expand()
}

let evenSquaresSum = [20, 17, 35, 4]
    .filter { $0 % 2 == 0 }.map { $0 * $0 }
    .reduce(0, +)
```

* Non-triggering
```swift
return match(pattern: pattern, with: [.comment])
    .flatMap { range in
        return Command(string: contents, range: range)
    }
    .flatMap { command in
        return command.expand()
    }

let evenSquaresSum = [20, 17, 35, 4].filter { $0 % 2 == 0 }.map { $0 * $0 }.reduce(0, +)

let evenSquaresSum = [20, 17, 35, 4]
    .filter { $0 % 2 == 0 }
    .map { $0 * $0 }
    .reduce(0, +)
```

> 3. Should the rule be configurable, if so what parameters should be configurable?

No configuration necessary

> 4. Should the rule be opt-in or enabled by default? Why?
   See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a
   rule as opt-in.

Disabled by default

freak4pc:
@erichoracek I like this and it's also useful to me and many friends in the Realm (😉) of Rx. 
Do you want to tackle this or are you OK with me giving it a go?

erichoracek:
@freak4pc Thanks for the interest! I'm pretty close to having a working implementation—feel free to take a shot at this if you'd like too though 😄

freak4pc:
@erichoracek oh not at all! Just wanted
To help out :) feel free to take your time with this

markst:
#490

PR #2221, opened 2018-05-25
Add multiline function chain rule

erichoracek:
See #2214

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,10 +43,10 @@
   autocorrection.  
   [Eric Horacek](https://github.com/erichoracek)
 
-* Fixes a case where the `closure_end_indentation` rule wouldn't lint the end
-  indentation of non-trailing closure parameters.  
+* Add a new `multiline_function_chains` rule to validate indentation of chained
+  function calls.  
   [Eric Horacek](https://github.com/erichoracek)
-  [#2121](https://github.com/realm/SwiftLint/issues/2121)]
+  [#2214](https://github.com/realm/SwiftLint/issues/2121)]
 
 * Improves the `mark` rule's autocorrection.  
   [Eric Horacek](https://github.com/erichoracek)
@@ -93,6 +93,11 @@
   [Eric Horacek](https://github.com/erichoracek)
   [#1341](https://github.com/realm/SwiftLint/issues/1341)
   [#1897](https://github.com/realm/SwiftLint/issues/1897)
+
+* Fixes a case where the `closure_end_indentation` rule wouldn't lint the end
+  indentation of non-trailing closure parameters.  
+  [Eric Horacek](https://github.com/erichoracek)
+  [#2121](https://github.com/realm/SwiftLint/issues/2121)]
 
 * Fix autocorrection for several rules
   (`empty_parentheses_with_trailing_closure`, `explicit_init`,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,8 +43,8 @@
   autocorrection.  
   [Eric Horacek](https://github.com/erichoracek)
 
-* Add a new `multiline_function_chains` rule to validate indentation of chained
-  function calls.  
+* Add a new `multiline_function_chains` rule to validate chained function 
+  calls start either on the same line or one per line.  
   [Eric Horacek](https://github.com/erichoracek)
   [#2214](https://github.com/realm/SwiftLint/issues/2121)]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,11 @@
   autocorrection.  
   [Eric Horacek](https://github.com/erichoracek)
 
+* Fixes a case where the `closure_end_indentation` rule wouldn't lint the end
+  indentation of non-trailing closure parameters.  
+  [Eric Horacek](https://github.com/erichoracek)
+  [#2121](https://github.com/realm/SwiftLint/issues/2121)]
+
 * Add a new `multiline_function_chains` rule to validate chained function 
   calls start either on the same line or one per line.  
   [Eric Horacek](https://github.com/erichoracek)
@@ -93,11 +98,6 @@
   [Eric Horacek](https://github.com/erichoracek)
   [#1341](https://github.com/realm/SwiftLint/issues/1341)
   [#1897](https://github.com/realm/SwiftLint/issues/1897)
-
-* Fixes a case where the `closure_end_indentation` rule wouldn't lint the end
-  indentation of non-trailing closure parameters.  
-  [Eric Horacek](https://github.com/erichoracek)
-  [#2121](https://github.com/realm/SwiftLint/issues/2121)]
 
 * Fix autocorrection for several rules
   (`empty_parentheses_with_trailing_closure`, `explicit_init`,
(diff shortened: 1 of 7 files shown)


erichoracek:
Just took a pass through all of the new violations, as far as I can tell none of them seem spurious 👍

Source/SwiftLintFramework/Rules/MultilineFunctionChainsRule.swift
```diff
@@ -0,0 +1,221 @@
...
+            guard
+                let offset = callDotOffset(file: file, callRange: range),
+                let line = file.contents.bridge().lineAndCharacter(forCharacterOffset: offset)?.line else {
+                    return nil
+            }
+            return (dotLine: line, dotOffset: offset, range: range)
+        }
+
+        let uniqueLines = calls
+            .map { $0.dotLine }
```
marcelofabri: we have an `unique` computed property declared on an `Array` extension, probably we could use it here

Source/SwiftLintFramework/Rules/MultilineFunctionChainsRule.swift
```diff
@@ -0,0 +1,221 @@
...
+        return noLeadingNewlineViolations.map { $0.dotOffset }
+    }
+
+    private static let whitespaceDotRegex = regex("\\s*\\.")
+
+    private func callDotOffset(file: File, callRange: NSRange) -> Int? {
+        guard
+            let range = file.contents.bridge().byteRangeToNSRange(start: callRange.location, length: callRange.length),
+            case let regex = type(of: self).whitespaceDotRegex,
+            let match = regex.matches(in: file.contents, options: [], range: range).last?.range else {
```
marcelofabri: would using `regex.firstMatch` work here?

Source/SwiftLintFramework/Rules/MultilineFunctionChainsRule.swift
```diff
@@ -0,0 +1,221 @@
...
+
+        let subcalls = dictionary.subcalls
+
+        if subcalls.isEmpty, let parentCallName = parentCallName, parentCallName.starts(with: name) {
+            return [NSRange(location: offset, length: length)]
+        }
+
+        return subcalls.flatMap { call -> [NSRange] in
+            guard let range = callRange(file: file, call: call, parentName: name, parentNameOffset: offset) else {
+                return []
```
marcelofabri: shouldn't we return `callRanges(file: file, kind: .call, dictionary: call, parentCallName: name)` here?

Source/SwiftLintFramework/Rules/MultilineFunctionChainsRule.swift
```diff
@@ -0,0 +1,221 @@
...
+            parentNameOffset == nameOffset,
+            let nameLength = call.nameLength,
+            let bodyOffset = call.bodyOffset,
+            let bodyLength = call.bodyLength,
+            let name = contents.substringWithByteRange(start: nameOffset, length: nameLength),
+            parentName.starts(with: name) else {
+                return nil
+        }
+
+        let linkOffset = nameOffset + nameLength
```
marcelofabri: what is a link? 🤔

Source/SwiftLintFramework/Rules/MultilineFunctionChainsRule.swift
```diff
@@ -0,0 +1,221 @@
...
+        return NSRange(location: linkOffset + offsetDifference + bodyLength,
+                       length: linkLength - bodyLength - offsetDifference)
+    }
+
+}
+
+fileprivate extension Dictionary where Key: ExpressibleByStringLiteral {
+    var subcalls: [[String: SourceKitRepresentable]] {
+        return substructure.compactMap { dictionary -> [String: SourceKitRepresentable]? in
+            guard case .call? = dictionary.kind.flatMap({ SwiftExpressionKind(rawValue: $0) }) else {
```
marcelofabri: small nitpick: you can use `dictionary.kind.flatMap(SwiftExpressionKind.init) == .call`

CHANGELOG.md
```diff
@@ -94,6 +94,11 @@
   [#1341](https://github.com/realm/SwiftLint/issues/1341)
   [#1897](https://github.com/realm/SwiftLint/issues/1897)
 
+* Fixes a case where the `closure_end_indentation` rule wouldn't lint the end
```
marcelofabri: this is on the wrong section

marcelofabri:
BTW, I was concerned about the timing regressions reported by oss-check, but it looks like it isn't that bad. Running locally I got this:

```
Message: Linting Aerial with this PR took 0.29s vs 0.28s on master (3% slower)
Message: Linting Alamofire with this PR took 2.27s vs 2.22s on master (2% slower)
Message: Linting Firefox with this PR took 8.33s vs 8.36s on master (0% faster)
Message: Linting Kickstarter with this PR took 11.13s vs 11.07s on master (0% slower)
Message: Linting Moya with this PR took 1.21s vs 1.16s on master (4% slower)
Message: Linting Nimble with this PR took 1.05s vs 0.96s on master (9% slower)
Message: Linting Quick with this PR took 0.32s vs 0.31s on master (3% slower)
Message: Linting Realm with this PR took 2.16s vs 2.12s on master (1% slower)
Message: Linting SourceKitten with this PR took 0.72s vs 0.72s on master (0% slower)
Message: Linting Sourcery with this PR took 3.13s vs 3.03s on master (3% slower)
Message: Linting Swift with this PR took 20.61s vs 20.23s on master (1% slower)
Message: Linting WordPress with this PR took 9.84s vs 9.32s on master (5% slower)
```

Source/SwiftLintFramework/Rules/MultilineFunctionChainsRule.swift
erichoracek: We need to use the last match here to ensure that in cases where a function call follows a property access don't trigger. e.g. 
```swift
self.happeningNewsletterOn = self.updateCurrentUser
    .map { $0.newsletters.happening }.skipNil().skipRepeats()
```
If you'd like I can add a comment, but this is covered in the tests 👍

Source/SwiftLintFramework/Rules/MultilineFunctionChainsRule.swift
erichoracek: In this case, since there's no chained subcall we don't need to descend further on, so we can just bail out and return early. I'll go ahead and add a comment to make it clear 👍

Source/SwiftLintFramework/Rules/MultilineFunctionChainsRule.swift
erichoracek: Was thinking like the "link" of a chain here, but I'll make these more descriptive 👍

erichoracek:
@marcelofabri Thanks for the review! Updated to address the comments.

To address your question about comment blocks—are you thinking about a case where a comment block falls inline within a function chain? e.g.:

```swift
let chain = a
    .b(1, 2, 3)
    /* Comment with a period. */ .c { blah in
        print(blah)
    }
    .d()
```

Looks like this may trigger a false-positive as is, but I can't imagine a case in which it would happen. Do you think it's worth the extra complexity?

marcelofabri:
@erichoracek When I was reviewing it, I thought that a commented out code block could trigger the rule, but I missed that we only check for dots after a call is found. So I agree it's not a big issue 👍

marcelofabri:
Thanks for another amazing contribution, @erichoracek 🎉

PR #2232, opened 2018-06-05
Fix #2231

### LINKED ISSUE #2231: False positive for rule identifier_name in Xcode 10 beta
Hey guys,

I've just received a bug report in one of my open source projects [here](https://github.com/Flinesoft/HandySwift/issues/22) which is related to the `identifier_name` rule. Specifically here's the code where the rule triggers an error:

``` swift
extension String {
    /// The type of allowed characters.
    ///
    /// - Numeric:          Allow all numbers from 0 to 9.
    /// - Alphabetic:       Allow all alphabetic characters ignoring case.
    /// - AlphaNumeric:     Allow both numbers and alphabetic characters ignoring case.
    /// - AllCharactersIn:  Allow all characters appearing within the specified String.
    public enum AllowedCharacters {
        case numeric
        case alphabetic
        case alphaNumeric
        case allCharactersIn(String) // Identifier Name Violation: Enum element name should only contain alphanumeric characters ...
    }
}
```

As you can see it's complaining about – probably – the braces which shouldn't be the case.

I'm using SwiftLint 0.25.1.

Coder-256:
SwiftLint fails with an error in enums with associated types in Xcode 10 due to the `identifier_name` rule. This fixes that error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,10 @@
 
 #### Bug Fixes
 
+* Fix identifier name false positive for enum cases with associated values in Xcode 10.
+  [Jacob Greenfield](https://github.com/Coder-256)
+  [#2231](https://github.com/realm/SwiftLint/issues/2231)
+
 * Update `LowerACLThanParent` rule to not lint extensions.  
   [Keith Smiley](https://github.com/keith)
   [#2164](https://github.com/realm/SwiftLint/pull/2164)
--- a/Source/SwiftLintFramework/Rules/IdentifierNameRule.swift
+++ b/Source/SwiftLintFramework/Rules/IdentifierNameRule.swift
@@ -35,7 +35,7 @@ public struct IdentifierNameRule: ASTRule, ConfigurationProviderRule {
                 return []
             }
 
-            let isFunction = SwiftDeclarationKind.functionKinds.contains(kind) || kind == .enumelement
+            let isFunction = SwiftDeclarationKind.functionKinds.contains(kind)
             let description = Swift.type(of: self).description
 
             let type = self.type(for: kind)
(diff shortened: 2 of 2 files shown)


Source/SwiftLintFramework/Rules/IdentifierNameRule.swift
```diff
@@ -35,7 +35,7 @@ public struct IdentifierNameRule: ASTRule, ConfigurationProviderRule {
                 return []
             }
 
-            let isFunction = SwiftDeclarationKind.functionKinds.contains(kind)
+            let isFunction = SwiftDeclarationKind.functionKinds.contains(kind) || kind == .enumelement
```
ronanrodrigo: Maybe for better naming and understand you could separate in two variables. Like:

```swift
let isFunction = SwiftDeclarationKind.functionKinds.contains(kind)
let isEnumelement = kind == .enumelement
```

Or jus update the `isFunction` to something like `isFunctionOrEnumelement`.

marcelofabri:
We shouldn't change the rule behavior (i.e. `oss-check` shouldn't report any changes). The problem is that `key.name` for an enum element now contains the parameters labels. Apparently, `key.nameoffset` and `key.namelength` only contain the name range (without parameters), so we should use that instead.

Coder-256:
@marcelofabri I'm having some trouble testing locally (many sourcekit errors, which it looks like they're already working on in SourceKitten). Let's see if these changes pass tests.

nickskull:
Can we get this merged in soon? :)

marcelofabri:
So, I've tested this and it looks like SourceKit reports incorrect values for `key.nameoffset` and `key.namelength` is the enum case contains backticks (e.g. \`private\`).

We should fill an issue in bugs.swift.org and meanwhile we can get the substring from `key.name` up to the first `(`. 

To avoid any potential issues, I'd only do that if `kind == .enumcase && SwiftVersion.current > .fourDotOne`.

CHANGELOG.md
```diff
@@ -85,6 +85,10 @@
 
 #### Bug Fixes
 
+* Fix identifier name false positive for enum cases with associated values in Xcode 10.
```
marcelofabri: requires two trailing spaces and lines should be limited to 80 chars as described in CONTRIBUTING.md: https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes

krodak:
Any update on this?

marcelofabri:
Closing this in favor of #2255

Coder-256:
Thank you @marcelofabri!

PR #2234, opened 2018-06-07
Rule tutorial

donald-m-ritter:
Adds a tutorial for writing native rules and adds a rule for unwanted imports.


donald-m-ritter:
I'll fix these issues once I get home from work.

donald-m-ritter:
That's a good point.  I'll add bench marking to the tutorial along with simplifying the rule.  That way it's more comprehensive and still gives some explanation of the internals.  It'll just be a little while before I get time at work to do that.

jpsim:
A tutorial walking through how to build and add a rule is long overdue, so thanks for taking this initiative! Though I encourage you to split off the new rule into a different PR so code reviews on the tutorial don't hold up the new rule, or vice versa.

donald-m-ritter:
I actually would like to get everything in one PR.  I think not only having documentation but also having the tutorial include more on testing, bench marking (now that I've figured out how), and auto correction (which I haven't figured out yet).  

I honestly hadn't considered the performance impact of this implementation until @marcelofabri pointed it out.  If this tutorial really helps I think there may be a larger influx of rules and the additional topics would be needed.

donald-m-ritter:
@marcelofabri @jpsim Sorry for taking so long to finally update this.  I've finally had time at work to add the performance testing and update the rule based on your suggestions.

Dschee:
@donald-m-ritter I'd like to review this, but after having a cursory look here's my first feedback:

1. Please rebase this branch so I can check it out in detail.
2. Couldn't we split this up into two separate PRs so we can discuss the tutorial separately from the new rule?
3. The tutorial looks way too long from my perspective, I know that I would probably just skip it even if I didn't understand how SwiftLint worked. Especially the code examples and related explanations seem too long for my taste. I'd find it more useful if all esteps (like also performance testing) all were mentioned with basic hints and if I were really interested in an example or details, I'd open a `<details><summary>Details</summary>Detailed Explanation with Code</details>` section.

donald-m-ritter:
@Dschee I've rebased the PR.  I'm more than happy to move some information into `details` blocks to help reduce the initial length.

I created the detailed tutorial because it took me a while to understand how to create native rules in the first place.  I also couldn't find any real tutorials elsewhere that not only explained how to create a rule but why it has to be done this way.  But I am hesitant to remove the information because there still is no comprehensive guide to creating a rule.  This tool is really useful and contributing shouldn't be hidden behind hours of trial and error/reverse engineering.

PR #2238, opened 2018-06-09
document linux install steps

mcandre:
clarify how to install SwiftLint in Linux, where Mint, Cocoapods, Homebrew, etc. are unavailable

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,13 +21,24 @@ unacceptable behavior to [info@realm.io](mailto:info@realm.io).
 
 ## Installation
 
-### Using [Homebrew](http://brew.sh/):
+### Ubuntu (from source)
+
+```console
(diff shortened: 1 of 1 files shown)


jpsim:
Thanks for the PR @mcandre. However, I don't think we can merge this just yet. A few thoughts:

* Most developers will use SwiftLint on macOS, so let's keep macOS installation instructions first.
* Most developers on macOS will use official SwiftLint releases without modification, so binary installations will remain most common, so let's keep those instructions first.
* Installing from source is possible not only on Linux but also on macOS, so let's not just call out source installation as a Linux option.

Dschee:
@mcandre Are you planning on updating this based on the feedback from @jpsim?

PR #2239, opened 2018-06-10
Opt-in rule to check for redundant type annotations

sjavora:
This rule checks for redundant type annotations, e.g. `var view: UIView = UIView()`. Violations are only triggered when it is the same type (e.g., `view: UIControl = UIView()` will not trigger).

This is my first rule and my first ever pull request on github, so feedback is welcome.


marcelofabri:
Thanks for the contribution!

PR #2242, opened 2018-06-14
Add `balanced_blank_lines` opt-in rule validating leading and trailing blank lines in bodies

### LINKED ISSUE #2241: Rule Request: Balanced leading and trailing blank lines in curly-braced bodies
### New Issue Checklist

### Rule Request

In our project and others I tend to often see inconsistent use of leading and trailing blank lines, e.g.

```
class C {
    
    var foo = true
}
```

or

```
func foo() {
    print("bar")

}
```

or

```
array.map { element in

    return foo(element)
}
```

IMHO the number of blank leading and trailing blank lines inside a body enclosed by curly braces should be identical. Unbalanced blank lines disrupt the reading flow and make it unnecessarily hard to quickly see the beginning and end of a structure or the internal grouping.

Optionally logic could be added to allow no more than 1 leading/trailing blank lines in class, struct, protocol, enum and extension bodies and no more than 0 leading/trailing blank lines in other bodies such as if, guard, ...

Not ok: See above.

Ok:

```
class C {
    var foo = true
}
```

```
class C {

    var foo = true

}
```

```
func foo() {
    print("bar")
}
```

```
array.map { element in
    return foo(element)
}
```

Yes, the severity level, potentially the maximum number of blank lines surrounding type bodies (such as class) and non-type bodies (such as if).

Opt in, because not everyone may care or agree.

Johennes:
Closes #2241

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,11 @@
   [JP Simard](https://github.com/jpsim)
   [#1420](https://github.com/realm/SwiftLint/issues/1420)
 
+* Add `balanced_blank_lines` opt-in rule validating leading and trailing blank
+  lines in bodies
+  [Johannes Marbach](https://github.com/Johennes)
+  [#2241](https://github.com/realm/SwiftLint/issues/2241)
+
 #### Bug Fixes
 
 * Update `LowerACLThanParent` rule to not lint extensions.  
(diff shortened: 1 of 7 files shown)


Johennes:
The code in here could probably use some simplification but I first wanted to get feedback on whether the approach of combining a regex search with structure analysis would be appropriate here.

Johennes:
Any chance this could be reviewed?

Issue #2154, opened 2018-04-13
explicit_type_interface rule triggered in case let switches


Hans92:
Xcode 9.3 (9E145) 
Swiflint 0.25.1

```
struct SomeError: Error { }
struct SomeError1: Error { }

let error: Error = Error()

switch error {
	case let error as SomeError:
		// some code
	case let error as SomeError1:
		// some code
}
```

On every `case` line I get an error: 
"Explicit Type Interface Violation: Properties should have a type interface (explicit_type_interface)".
I can't add here error type. Swift doesn't get it.

I am not sure if this is a bug or I'm doing something wrong.

marcelofabri:
Maybe we should ignore variables inside cases (`source.lang.swift.stmt.case`)?

This is the structure that SourceKit returns:

```
{
          "key.elements" : [
            {
              "key.kind" : "source.lang.swift.structure.elem.pattern",
              "key.length" : 23,
              "key.offset" : 164
            }
          ],
          "key.kind" : "source.lang.swift.stmt.case",
          "key.length" : 29,
          "key.namelength" : 0,
          "key.nameoffset" : 0,
          "key.offset" : 159,
          "key.substructure" : [
            {
              "key.kind" : "source.lang.swift.decl.var.local",
              "key.length" : 5,
              "key.name" : "error",
              "key.namelength" : 5,
              "key.nameoffset" : 168,
              "key.offset" : 168
            }
          ]
        }
```

Hans92:
Also there is also a problem with `[weak self]` in closures.

marcelofabri:
Feel free to send a PR!

ppamorim:
There is a error on `case .failure(let error):`, Swiftlint warns to set the type, but it is not possible.

Pikacz:
There is still error on `case X.value(let val)` but not on `case .value(let val)`

ppamorim:
@marcelofabri Any type of regression on this? I am getting this error again on 0.43.1.

PR #2244, opened 2018-06-15
Fix `explicit_type_interface` when used in statements

dirtydanee:
Fix https://github.com/realm/SwiftLint/issues/2154

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,10 @@
 
 #### Bug Fixes
 
+* Fix `explicit_type_interface` when used in statements.  
+  [Daniel Metzing](https://github.com/dirtydanee)
+  [#2154](https://github.com/realm/SwiftLint/issues/2154)
+
 * Fix an issue with `control_statement` where commas in clauses prevented the
   rule from applying.  
   [Allen Wu](https://github.com/allewun)
--- a/Tests/SwiftLintFrameworkTests/ExplicitTypeInterfaceRuleTests.swift
+++ b/Tests/SwiftLintFrameworkTests/ExplicitTypeInterfaceRuleTests.swift
@@ -133,12 +133,16 @@ class ExplicitTypeInterfaceRuleTests: XCTestCase {
 
     func testSwitchCaseDeclarations() {
 
-        let nonTriggeringExamples = [
+        guard SwiftVersion.current >= .fourDotOne else {
+            return
+        }
+
+        let nonTriggeringExamples: [String] = [
             "enum Foo {\n"                                      +
             "   case failure(Any)\n"                            +
             "   case success(Any)\n"                            +
             "}\n"                                               +
-            "func bar {\n"                                      +
+            "func bar() {\n"                                      +
             "   let foo: Foo = Foo.success(1)\n"                +
             "   switch foo {\n"                                 +
             "   case .failure(let error): let bar: Int = 1\n"   +
--- a/Tests/SwiftLintFrameworkTests/ExplicitTypeInterfaceRuleTests.swift
+++ b/Tests/SwiftLintFrameworkTests/ExplicitTypeInterfaceRuleTests.swift
@@ -52,10 +52,128 @@ class ExplicitTypeInterfaceRuleTests: XCTestCase {
             "class Foo {\n  ↓static var myStaticVar = 0\n}\n",
             "class Foo {\n  ↓class var myClassVar = 0\n}\n"
         ]
+
         let description = ExplicitTypeInterfaceRule.description
             .with(triggeringExamples: triggeringExamples)
             .with(nonTriggeringExamples: nonTriggeringExamples)
 
         verifyRule(description, ruleConfiguration: ["allow_redundancy": true])
+    }
+
+    func testEmbededInStatements() {
+        let nonTriggeringExamples = [
+            "func foo() {\n"                        +
+            "   var bar: String?\n"                 +
+            "   guard let strongBar = bar else {\n" +
+            "       return\n"                       +
+            "   }\n"                                +
+            "}",
+            "struct SomeError: Error {}\n"          +
+                "var error: Error?\n"               +
+                "switch error {\n"                  +
+                "case let error as SomeError:\n"    +
+                "   break\n"                        +
+                "default:\n"                        +
+                "   break\n"                        +
+            "}"
+        ]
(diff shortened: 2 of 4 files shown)


Source/SwiftLintFramework/Rules/ExplicitTypeInterfaceRule.swift
```diff
@@ -41,7 +44,36 @@ public struct ExplicitTypeInterfaceRule: ASTRule, OptInRule, ConfigurationProvid
...
+}
+
+private extension Dictionary where Key == String, Value == SourceKitRepresentable {
+
+    var containsType: Bool {
+        return typeName != nil
+    }
+
+    var containsWeakSelf: Bool {
+        guard let name = name, name == "self" else {
```
marcelofabri: you could replace this to `guard name == "self" else`

CHANGELOG.md
```diff
@@ -95,6 +95,10 @@
 
 #### Bug Fixes
 
+* Fix `explicit_type_interface` when used in statements.
```
marcelofabri: this is in the wrong section, can you update this before merging?

Source/SwiftLintFramework/Rules/ExplicitTypeInterfaceRule.swift
```diff
@@ -29,8 +29,11 @@ public struct ExplicitTypeInterfaceRule: ASTRule, OptInRule, ConfigurationProvid
                          dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
 
         guard configuration.allowedKinds.contains(kind),
-            !containsType(dictionary: dictionary),
-            let offset = dictionary.offset else {
+            !dictionary.containsType,
+            !dictionary.containsWeakSelf,
```
marcelofabri: shouldn't we try to check if there's a capture group, no matter if it's weak or if it's capturing `self`?

CHANGELOG.md
dirtydanee: Fixed

Source/SwiftLintFramework/Rules/ExplicitTypeInterfaceRule.swift
dirtydanee: Yeah, that would make sense. Thanks for pointing it out!

This case is still failing:
```
 "class Foo {\n"                                         +
            "   func bar() {\n"                                     +
            "       var k = 0\n"                                    +
            "       _ = { [weak self, weak k] in\n"                 +
            "       guard let strongSelf = self else { return }\n"  +
            "       }\n"                                            +
            "   }\n"                                                +
            "}",
```

I am not sure how could I find the capture group, my first guess would be using a regex, something like this:
`=\s+{\s*\[((\s)*\w+\s*\w+(,)*)+]\s+in`

Do you have maybe an other idea how to approach finding the capture group?

dirtydanee: Added a solution for searching capture groups with regex.

Tests/SwiftLintFrameworkTests/ExplicitTypeInterfaceRuleTests.swift
```diff
@@ -36,4 +36,49 @@ class ExplicitTypeInterfaceRuleTests: XCTestCase {
...
+        ]
+        let triggeringExamples = ExplicitTypeInterfaceRule.description.triggeringExamples
+        let description = ExplicitTypeInterfaceRule.description
+            .with(triggeringExamples: triggeringExamples)
+            .with(nonTriggeringExamples: nonTriggeringExamples)
+
+        verifyRule(description)
+    }
+
+    func testWeakSelf() {
```
marcelofabri: this is not just testing `weak self` anymore

Source/SwiftLintFramework/Rules/ExplicitTypeInterfaceRule.swift
```diff
@@ -29,8 +43,11 @@ public struct ExplicitTypeInterfaceRule: ASTRule, OptInRule, ConfigurationProvid
                          dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
 
         guard configuration.allowedKinds.contains(kind),
-            !containsType(dictionary: dictionary),
-            let offset = dictionary.offset else {
+            !dictionary.containsType,
+            let offset = dictionary.offset,
+            !file.captureGroupByteRanges.contains(where: { $0.contains(offset) }),
```
marcelofabri: This will perform the search for the whole file every time 😱 

This is an opt-in rule, so having a worse performance is acceptable, but I'd be awesome if we could think of a way to make it faster.

Performing the regex match just on the parent type would be an option, but it'd have to change a few things. Other option would be performing the whole search first and the navigate the AST manually instead of using `ASTRule`.

dirtydanee: I choose the later one, thanks for the suggestion. Here is a screenshot of the listing speed from *before the change*:

screen shot 2018-07-28 at 20 29 43

dirtydanee:
@marcelofabri I have pushed changes fixing the false negatives as you pointed out on https://github.com/realm/SwiftLint/pull/2321 and rebased to current master.

I see that os check is failing, with the second try on listing Swift. Could you give me some insights on how to fix this issue?

dirtydanee:
In my last commit https://github.com/realm/SwiftLint/pull/2244/commits/31ec32bcd6d49d56b664878dafd5660c73f6564e I added a runtime guard for the Swift version to be higher than 4.1.

Given this simple example:
```swift
func foo() {
  switch foo {
  case var (x, y): ↓let fooBar = 1
  }
}
```

The issue seems for the simple example below `SourceKitten` is not returning any `varLocal` `SwiftDeclarationKind` for `fooBar` in the file's structure running Swift 4.0. In fact, its not returning any declaration if it is inside a `case` statement.

I think the rule makes could be helpful for earlier Swift versions, so I would keep it as it is, but maybe we should document this limitation somehow?

jpsim:
@dirtydanee would you mind opening a new PR with this change? I'd be happy to start a review on it again, but I'd rather start fresh.

dirtydanee:
@jpsim I have reopened under the PR https://github.com/realm/SwiftLint/pull/2588

jpsim:
Fantastic, thank you @dirtydanee. Let's close this and continue there.

PR #2245, opened 2018-06-16
Command-line option to disable merging of nested configurations

aksvenk:
This PR adds an option to disable merging configurations in nested folders.
Usage: `swiftlint lint --disable-nested-config-merge`

**Motivation**
Sometimes nested configurations in a project may have vastly different rules from the parent. 
The current strategy of always merging nested configs might not cut it in the above case. The nested configs would inherit rules that are not expected from the parent.

**Solution**
A command-line flag has been added to disable merging of nested configurations. When the said flag is used, nested configurations are 'disconnected' from the parent and the rule-sets are not merged. 

**Potential use-case**
The change in the project structure below:
- allows nested config files to operate in a standalone manner without being affected by the root
- allows the preparation of an aggregate SwiftLint report which includes the project and its frameworks by running the `swiftlint` command from `Project A`'s root folder

```
Project A
   +-.swiftlint.yml (root)
   |
   +--+-Framework A
   |  +-.swiftlint.yml (nested)
   |
   +--+-Framework B
   |  +-.swiftlint.yml (nested)
   |
   +--+-Framework C
      +-.swiftlint.yml (nested)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@
 
 #### Enhancements
 
+* Add an option to disable nested config merges with the `--disable-nested-config-merge` command-line option
+  [Akshay Venkatesh](https://github.com/aksvenk)
+
 * Add optional filename verification to the `file_header` rule.
   All occurrences in the pattern of the `SWIFTLINT_CURRENT_FILENAME`
   placeholder are replaced by the name of the validated file.  
--- a/Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
@@ -23,7 +23,8 @@ extension Configuration {
             let fullPath = pathNSString.absolutePathRepresentation()
             let config = Configuration.getCached(atPath: fullPath) ??
                 Configuration(path: configurationSearchPath, rootPath: fullPath, optional: false, quiet: true)
-            return merge(with: config)
+            let shouldMergeNestedConfigs = mergeNestedConfigs ?? true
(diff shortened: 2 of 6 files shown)


jpsim:
Thanks for the PR here and sorry for not reviewing sooner. I definitely see the value of disconnecting a nested configuration file from its parents. However, I still think it's useful to make this decision on a per-configuration basis rather than based on the invocation of swiftlint. Besides, I think a useful property of SwiftLint configurations is that you can go to any Swift project that has properly configured SwiftLint and run `swiftlint` in its root and have that do the right thing.

So instead of a CLI flag, how about we add a Configuration key like `root: true` or `root_configuration: true`? In my mind, this would "reset" the nesting and further subdirectories would merge with all parent configurations up to its closest "root".

What do you think?

Dschee:
@aksvenk Are you still interested in this feature? If so, would you mind answering @jpsim's question? Just trying to bring the discussion forward.

aksvenk:
@Dschee sorry about the delay.

@jpsim I get your point, it is heaps more elegant to stop nesting through configuration than command-line arguments. We do gain the capability of locking in a configuration to "what the author intended" on third party code while keeping the nesting behaviour of the subdirectories untouched.

The initial reasoning behind the command-line parameter was to provide this feature instantly to projects that don't have control over code that they nest, but I do agree it is a jackhammer approach.

I'm happy to close this and open another PR with the configuration file based flag.

Do you think it is worth somehow showing the user where the merging of configurations was terminated in each path so as to avoid confusions about rules not being picked up in configurations "outside the root". Or does this feature already exist?

Issue #2250, opened 2018-06-21
Decrease SuperfluousDisableCommandRule default severity


fredpi:
Currently, when the sources contain superfluous disable commands, SwiftLint triggers a build error.  

There are many cases where such behaviour is counterproductive and misaligned with the goal of SwiftLint to increase overall productivity. While the severity of this rule can be manually adjusted ([as seen here for instance](https://github.com/watson-developer-cloud/swift-sdk/pull/792)), I believe the default case should be to trigger a warning instead.

This is especially important for larger projects where it's not uncommon to have build times over a minute. If the build fails because of a superfluous disable  command, e. g. originating from removing one single line of code while a rule like `file_length` is active, this can be super frustrating and takes away valuable time. This is why I think the default severity should be a warning; if some people still prefer the rule to trigger an error, this can still be changed in the config file.

PR #2251, opened 2018-06-21
Decrease SuperfluousDisableCommandRule severity

fredpi:
Corresponding issue: #2250.


marcelofabri:
This seems reasonable for me. @jpsim do you remember if there was any strong motivation for the default to be `error`?

PR #2252, opened 2018-06-22
[Do Not Merge] test using https://github.com/jpsim/SourceKitten/pull/538

norio-nomura:


norio-nomura:
There is no problem with using that but it seems that the speed is considerably slow. 🤔 
I think that speed reduction is not related to https://github.com/jpsim/SourceKitten/pull/538.

marcelofabri:
It could be just a Circle CI instability. Running `oss-check` locally should tell us the truth.

jpsim:
New SourceKitten release: https://github.com/jpsim/SourceKitten/releases/tag/0.21.1

PR #2257, opened 2018-06-24
Don’t overwrite file when formatting if not needed

### LINKED ISSUE #2249: autocorrect --format touches files which it doesn't actually modify
### New Issue Checklist

`swiftlint autocorrect --format --path <mydir>`

This command touches all files it sees, even when this doesn't result in any changes to the files. Without `--format` swiftlint doesn't touch files it doesn't change.

This is bad when the command is run as part of Xcode build. Xcode thinks the file has changed externally (which it hasn't), and I get into conflicts (Xcode asks whether to reload file or keep Xcode version, which is annoying).

Also, I haven't found documentation for `--format`.

So, am I doing it wrong or could `--format` be fixed so it doesn't touch files that don't need any modifications? Thanks!

marcelofabri:
Fixes #2249

@petterikamppuririchie can you please check if this fixes the issue?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@
   [Frederick Pietschmann](https://github.com/fredpi)
   [#2250](https://github.com/realm/SwiftLint/issues/2250)
 
+* Don't touch files when running `autocorrect --format` if the contents haven't
+  changed.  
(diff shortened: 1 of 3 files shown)


petterikamppuririchie:
It sure does! Thanks for the quick fix! 👍

PR #2261, opened 2018-06-24
Fix required_enum_case rule when linting with Swift 4.2

marcelofabri:
This also simplifies the implementation a bit.


jpsim:
Nice!

PR #2269, opened 2018-06-25
Add modifier failure reason

dirtydanee:
As discussed in https://github.com/realm/SwiftLint/pull/1501#event-1611485385

### DIFF
--- a/Source/SwiftLintFramework/Rules/ModifierOrderRule.swift
+++ b/Source/SwiftLintFramework/Rules/ModifierOrderRule.swift
@@ -1,5 +1,8 @@
 import SourceKittenFramework
 
+private typealias ModifierKeyword = String
+private typealias ModifierDescription = (ModifierKeyword, SwiftDeclarationAttributeKind.ModifierGroup)
+
 public struct ModifierOrderRule: ASTRule, OptInRule, ConfigurationProviderRule {
     public var configuration = ModifierOrderConfiguration(preferredModifierOrder: [.override, .acl])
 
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Rules/ModifierOrderRule.swift
```diff
@@ -183,7 +183,9 @@ public struct ModifierOrderRule: ASTRule, OptInRule, ConfigurationProviderRule {
         let filteredPreferedOrderOfModifiers = preferredOrderOfModifiers.filter(modifierGroupsInDeclaration.contains)
         for (index, preferredGroup) in filteredPreferedOrderOfModifiers.enumerated()
             where preferredGroup != modifierGroupsInDeclaration[index] {
-                return [StyleViolation(ruleDescription: type(of: self).description,
+                let reason = " \(preferredGroup) modifier should be before " +
+                             "\(modifierGroupsInDeclaration[index]) modifier."
+                return [StyleViolation(ruleDescription: type(of: self).description.appendDescription(with: reason),
```
marcelofabri: you don't need to change the description, `StyleViolation` has a `reason` parameter that can be used directly.

Source/SwiftLintFramework/Rules/ModifierOrderRule.swift
```diff
@@ -183,9 +183,12 @@ public struct ModifierOrderRule: ASTRule, OptInRule, ConfigurationProviderRule {
         let filteredPreferedOrderOfModifiers = preferredOrderOfModifiers.filter(modifierGroupsInDeclaration.contains)
         for (index, preferredGroup) in filteredPreferedOrderOfModifiers.enumerated()
             where preferredGroup != modifierGroupsInDeclaration[index] {
+                let failureReason = "\(preferredGroup) modifier should be before " +
```
marcelofabri: this is already an improvement, but would it be possible to print the exact modifiers instead of their groups?

If it's not hard, we should do that. Otherwise we can merge as it is and then figure out later.

dirtydanee: I have figured out a not so complex way to get the  exact modifiers instead of the groups. I will push my changes soon to this PR.

Tests/SwiftLintFrameworkTests/ModifierOrderTests.swift
```diff
@@ -144,4 +144,21 @@ class ModifierOrderTests: XCTestCase {
         verifyRule(descriptionOverride,
                    ruleConfiguration: ["preferred_modifier_order": ["override", "acl", "final"]])
     }
+    #if swift(>=4.1.0)
```
marcelofabri: Should we do a runtime check (using `SwiftVersion.current`) instead of at compile time?

PR #2270, opened 2018-06-25
Add "Redundant @objc Attribute" Rule

dirtydanee:
Implements https://github.com/realm/SwiftLint/issues/2193

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,10 @@
   `homebrew-core/Formula/swiftlint.rb` within sandbox.  
   [Norio Nomura](https://github.com/norio-nomura)
 
+* Add `redundant_objc_attribute` to warn against already implied objc attribute.  
+  [Daniel Metzing](https://github.com/dirtydanee)
+  [#2193](https://github.com/realm/SwiftLint/issues/2193)
+
 #### Bug Fixes
 
 * Fix compiler warnings when building with Swift 4.2 introduced in the last
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,10 @@
 
 #### Enhancements
 
-* None.
+* Add `redundant_objc_attribute` to warn against already implied `@objc`
+  attribute.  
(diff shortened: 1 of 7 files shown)


jpsim:
I'm sorry for the long delay in reviewing this. Would you mind rebasing and pinging me again to review? Thanks for the PR!

dirtydanee:
@jpsim rebased to latest master

jpsim:
Looks good! I can see a lot of care went into this, so thank you for that! I made a few refactors, I hope you don't mind. Assuming CI is happy with this, we can merge 😄

Kaspik:
Thanks for doing this guys!! But there is an issue with selectors. Imagine code:
`button.addTarget(self, action: #selector(getPrintablePosterButtonAction), for: UIControl.Event.touchUpInside)`
This requires to have method with @objc selector - `@objc private func getPrintablePosterButtonAction() {}`. Now if it's inside @objcMembers class, this causes warnings.

I know I can suppress them, but I was thinking if there is not any workarounds or checks for these cases. :)

jpsim:
Sounds like that might be a Swift bug actually. Functions that are representable by the objective-c runtime that are encased in an `@objcMembers` class should be automatically treated as `@objc` themselves.

dirtydanee:
I have run a quick test, and the issue seem to be only occurring, if the function is marked as `private`.

I should have some time to look into this week, will update this thread with my findings.

jpsim:
> I have run a quick test, and the issue seem to be only occurring, if the function is marked as `private`.
> 
> I should have some time to look into this week, will update this thread with my findings.

This seems correct, and probably isn't a Swift bug, but rather subtle behavior with `@objcMembers`, where it isn't automatically applied to private members. It should be easy to special case this in the rule.

Kaspik:
Thank you guys, sorry for bringing up these problems, ideal would be to remove `@objcMemebers` from our app but we can't sometimes. 😄

jpsim:
Not at all, we shouldn't change our code just to satisfy linters 😀

PR #2272, opened 2018-06-28
Fix identifier_name check enum-element-with-associated-value error

gnou:
#2255 was the fix of #2231 , but in this(#2255) fix, @marcelofabri set a Swift lang version check, only check >4.1, which makes it ineffectual if you're using Xcode 10 & Swift 4.1, I removed that constraint in my pull request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,8 +27,8 @@
   when linting with Swift 4.2.  
   [Marcelo Fabri](https://github.com/marcelofabri)
 
-* Fix `identifier_name` rule false positives with `enum` when linting
-  using Swift 4.2.  
+* Fix `identifier_name` rule false positives with `enum`-with-associated-value when linting.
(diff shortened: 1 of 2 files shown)


marcelofabri:
@gnou Even if you are using Swift 4.1, it's actually reported as 4.2 by SwiftLint because that's the compiler version. Have you tried compiling master and running against your project?

gnou:
@marcelofabri I did, this works for my project.

marcelofabri:
> I assume it's a bug in Xcode 10.

This was an issue with SourceKitten that should be fixed on master. Can you please double check?

> I don't think swift-lang-version-check is necessary in this case, as long as it's enumelement we're checking, we can safely check and remove () from it.

I agree, but at the same time, I didn't want to penalize users on Swift < 4.2 with a more expensive operation. Also, since Swift 4.2 is still in beta, the behavior might change.

marcelofabri:
Closing this for now.

PR #2274, opened 2018-07-02
Split RulesTests into several test cases, generating them automatically

### LINKED ISSUE #2263: Split RulesTests into several test cases
Xcode 10 introduced parallel testing. It works by giving out different `XCTestCase`s to different runners.

However, `RulesTests` contains a lot of tests that could be easily parallelizable. We need to split it into several test cases in order to take full advantage of parallel testing.

marcelofabri:
Fixes #2263


Makefile
```diff
@@ -35,16 +35,20 @@ VERSION_STRING=$(shell /usr/libexec/PlistBuddy -c "Print :CFBundleShortVersionSt
 
 all: build
 
-sourcery: Tests/LinuxMain.swift Source/SwiftLintFramework/Models/MasterRuleList.swift
+sourcery: Source/SwiftLintFramework/Models/MasterRuleList.swift Tests/SwiftLintFrameworkTests/AutomaticRuleTests.swift Tests/LinuxMain.swift
 
 Tests/LinuxMain.swift: Tests/*/*.swift .sourcery/LinuxMain.stencil
-	sourcery --sources Tests --templates .sourcery/LinuxMain.stencil --output .sourcery
+	sourcery --sources Tests --templates .sourcery/LinuxMain.stencil --output .sourcery --verbose --force-parse generated
```
marcelofabri: I had to add an extra extension to the generated file, otherwise there's no way to force Sourcery to parse it in order to generate LinuxMain.swift

Source/SwiftLintFramework/Protocols/Rule.swift
```diff
@@ -23,6 +23,8 @@ extension Rule {
 
 public protocol OptInRule: Rule {}
 
+public protocol AutomaticTestableRule: Rule {}
```
marcelofabri: In an ideal world, this would be defined in the test target, but I thought that it'd be much more convenient to add the conformance when defining the rule than having to go to a `Rules+AutomaticTestable.swift` in the test target and add an extension there.

Issue #2278, opened 2018-07-04
SwiftLint 0.26.0 is incompatible with Xcode 10 beta 3


Coeur:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
sourcekit: [1:connection-event-handler:10499: 0.0000] Connection interruptsourcekit: [1:updateSemanticEditorDelay:10499: 0.0005] disabling semantic editor for 10 secondssourcekit: [1:pingService:10499: 0.0006] pinging servicesourcekitten: connection to SourceKitService restored!
Connection interrupted
Never call this for file that sourcekitd fails.: file File+Cache.swift, line 127
/Users/antoinecoeur/Library/Developer/Xcode/DerivedData/swift-lint-ddwozybquovlpoecgvjvvuwycpxq/Build/Intermediates.noindex/swift-lint.build/Debug-iphonesimulator/swift-lint.build/Script-37B7F63520ED0A12000AE5CC.sh: line 2: 54316 Abort trap: 6           "${PODS_ROOT}/SwiftLint/swiftlint"
Command PhaseScriptExecution failed with a nonzero exit code
```

### Environment

* SwiftLint 0.26.0: does **not** happen with any previous versions (like 0.25.1)
* Installation method used: CocoaPods 1.5.3
* Xcode 10 beta 3: does **not** happen with any previous versions (like Xcode 10 beta 2)
* Any version of Swift (3.x, 4.0, 4.2) will cause the error
* Any Swift code will cause the error (your project can be empty)
* Podfile:

```yml
source 'https://github.com/CocoaPods/Specs.git'

platform :ios, '9.0'
use_frameworks!

pod 'SwiftLint', '~> 0.26.0'
target 'swift-lint'
```

Sample:

https://github.com/Coeur/swiftlint0.26.0-xcode10beta3

norio-nomura:
closing this as a duplicate of #2273

PR #2280, opened 2018-07-05
Improving changelog for the SourceKitten update

Coeur:
As demonstrated by #2278, the issue is common to all versions of Swift, including 3.x and 4.0.


norio-nomura:
Since [SR-7954](https://bugs.swift.org/browse/SR-7954) also happened on the Linux version, we should write `Swift 4.2` instead of `Xcode 10`.

If #2278 still occurs with `swiftlint` with installed by `brew install swiftlint --HEAD`, please reopen them.

Issue #1896, opened 2017-10-11
Rule Request: multiline_arguments: Allow enforcing only after first closure


ldiqual:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Rule Request

My proposal is to add a new configuration for `multiline_arguments` that would allow enforcing the rule **only after the first closure**.

For instance, I would like to handle this case:

```swift
UIView.animate(withDuration: duration, animations: {
    // animations
}, completion: { _ in
    // completion
})
```

However with the current defaults, I have to write it this way:

```swift
UIView.animate(
    withDuration: duration,
    animations: {
        // animations
    }, completion: { _ in
        // completion
    }
)
```

You would also be able to pass more inline arguments before `animations`.

 >1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.

I believe that the animate + completion case above is common across iOS codebases and that the first style above is a common way to express it.

Ray Wenderlich's swift style guide passes `animations` on the same line: https://github.com/raywenderlich/swift-style-guide#closure-expressions

 >2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

**Doesn't trigger**

```swift
call(a: a, b: {

}, c: {

})
```

```swift
call(a: a, b:b, c: {

}, d: {

})
```

**Triggers**

```swift
call(a: a, b:b,
  c: c, d: {

  }, d: {

  }
)
```

```swift
call(a: a, b: {

}, c: {

}, d: d, e: e)
```

 >3. Should the rule be configurable, if so what parameters should be configurable?

A name for such option could be:

 - `allow_first_closure_on_same_line`
 - `only_enforce_after_first_closure`

Hard to describe the intent in a generic way and I'd love input from you guys.

 >4. Should the rule be opt-in or enabled by default? Why?
   See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a
   rule as opt-in.

The option would be enabled by default for the `multiline_arguments` rule, since this is a common case as described above.

tinder-garricnahapetian:
I think this is a great idea. Would it be hard to implement? I could give it a try. Where would I start?

jpsim:
[...]

PR #2281, opened 2018-07-05
add "allow_first_closure_on_same_line" configuration to "multiline_arguments"

maciesielka:
Adds the functionality described in #1896. I opted for the `allow_first_closure_on_same_line` name for the new configuration as suggested in the rule request.

The gist of the change is to compute violations in the same manner as the existing logic, but remove those that occur before a closure in the first line of arguments if the configuration is set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,8 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2249](https://github.com/realm/SwiftLint/issues/2249)
   
-  * Add `allow_first_closure_on_same_line` configuration to 
-  `multiline_arguments`  
+  * Add `only_enforce_after_first_closure_on_first_line` configuration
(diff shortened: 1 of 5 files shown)


marcelofabri:
This looks good! I just wonder if `only_enforce_after_first_closure` isn't a better description for this than `allow_first_closure_on_same_line`.

maciesielka:
@marcelofabri any strong feelings about the naming options I suggested above?

marcelofabri:
@maciesielka Sorry for the delay. I don't have any strong feelings, but I think `only_enforce_after_first_closure_on_first_line` would be more accurate if we can't think of a better name. Thoughts?

maciesielka:
It's probably a bit verbose, but I think it describes the rule better than what's in the PR. I'll push the change up to use `only_enforce_after_first_closure_on_first_line` when I get a free minute.

Issue #2283, opened 2018-07-06
Rule Request: Use AnyObject instead of class for class-only protocols


Noobish1:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Rule Request

1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.

Apple recommends it, as of Swift 4: https://docs.swift.org/swift-book/LanguageGuide/Protocols.html. And `class` is just a typealias to `AnyObject` as is stated [here](https://github.com/apple/swift-evolution/blob/e3f387d08307ce0af7a0f80979151714c41988bb/proposals/0156-subclass-existentials.md#class-and-anyobject)

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

triggering: `protocol <protocol-name>: class`
non-triggering: `protocol <protocol-name>: AnyObject`

3. Should the rule be configurable, if so what parameters should be configurable?

Nope, we should we using AnyObject in all cases I believe.

4. Should the rule be opt-in or enabled by default? Why?
   See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a
   rule as opt-in.

Nope, we should we using AnyObject in all cases I believe.

ornithocoder:
@Noobish1, are you planning to work on it? Otherwise I will :-)

Noobish1:
I’m not, no. Go ahead and thanks!

Blair.

ornithocoder:
@Noobish1 👍 Thanks!

marcelofabri:
Implemented by @ornithocoder on #2285 🎉

ornithocoder:
@Noobish1, implemented w/ autocorrect.

Noobish1:
Awesome, nice work @ornithocoder.

PR #2285, opened 2018-07-07
Add opt-in anyobject_protocol rule

ornithocoder:
This commit implements #2283.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2249](https://github.com/realm/SwiftLint/issues/2249)
 
+* Add default rule `anyobject_protocol` which suggests using `AnyObject` over
+  `class` for class-only protocols.  
+  [Ornithologist Coder](https://github.com/ornithocoder)
+  [#2283](https://github.com/realm/SwiftLint/pull/2283)
+
 #### Bug Fixes
 
 * Fix an issue with `control_statement` where commas in clauses prevented the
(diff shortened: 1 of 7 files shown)


ornithocoder:
@marcelofabri @norio-nomura hi there! Do you have any idea why the linux tests are failing?

marcelofabri:
@ornithocoder I think I remember something like that when implementing `ClassDelegateProtocolRule`. Maybe there's something useful there.

marcelofabri:
BTW, I'm not 100% convinced this should be a default rule. The number of existing violations is an indication of that.

ornithocoder:
@marcelofabri, I'll take a look on the `ClassDelegateProtocolRule`. Thanks.

The Swift documentation [says](https://docs.swift.org/swift-book/LanguageGuide/Protocols.html#ID281) AnyObject should be used to limit protocol adoption to class types. But I'll change it to opt-in.

ornithocoder:
I'll need more time to investigate the Linux issue. Even the very [first implementation](https://github.com/ornithocoder/personal-fork-swiftlint/blob/01dc00f576237571a0140162fdf597def70f02a7/Source/SwiftLintFramework/Rules/AnyObjectProtocolRule.swift) (which simply checks the `inheritedTypes`) was failing on Linux.

ornithocoder:
🧐

I removed everything from the rule, including the violations. This is what happens where I comment out the triggering examples:

without examples

And this is what happens with the triggering examples:

with examples

cc: @marcelofabri @norio-nomura

ornithocoder:
It's ready for review :-)

Issue #2275, opened 2018-07-02
force-exclude doesn't work


cooksimo:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Bug Report

Running swiftlint with a `path` specified and an exclusion in the config that would exclude that file still lints that file even when `force-exclude` is specified.

```bash
$ swiftlint lint --path path/to/SomeFile.swift --force-exclude
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)?
0.26.0
* Installation method used (Homebrew, CocoaPods, building from source, etc)?
Cocoapods

marcelofabri:
Can you provide more info or even better a sample example?

cooksimo:
Sure, see attached.

[swiftlint-example.zip](https://github.com/realm/SwiftLint/files/2157613/swiftlint-example.zip)

Running `swiftlint --path exclude/ExcludeMe.swift --force-exclude` will still lint the file `exclude/ExcludeMe.swift`

PR #2286, opened 2018-07-09
Filter excluded files after discovery

inket:
This is a PR fixing the long-standing bug (or unexpected behavior) of SwiftLint where excluded files in the configuration are not taken into account when processing files from the `--path` or `--use-script-input-files` parameters.

---

These are the issues and PRs referencing this behavior that I could find:

Issues:
https://github.com/realm/SwiftLint/issues/2275
https://github.com/realm/SwiftLint/issues/987

Previous attempts:
https://github.com/realm/SwiftLint/pull/1847

---

I hope I got it right, and if not, please point me towards the right way for doing it so we can get this fixed once and for all. 👍

### DIFF
--- a/Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
@@ -2,6 +2,13 @@ import Foundation
 import SourceKittenFramework
 
 extension Configuration {
+    public func lintableFiles(in files: [File], forceExclude: Bool) -> [File] {
+        return files.compactMap { file -> File? in
+            guard let filePath = file.path else { return nil }
+            return lintableFiles(inPath: filePath, forceExclude: forceExclude).first
+        }
+    }
+
     public func lintableFiles(inPath path: String, forceExclude: Bool) -> [File] {
         return lintablePaths(inPath: path, forceExclude: forceExclude).compactMap(File.init(pathDeferringReading:))
     }
--- a/Source/swiftlint/Extensions/Configuration+CommandLine.swift
+++ b/Source/swiftlint/Extensions/Configuration+CommandLine.swift
@@ -62,7 +62,7 @@ extension Configuration {
             }
             return .success(Dictionary(grouping: files, by: configuration(for:)))
         }.flatMap { filesPerConfig -> Result<[Configuration: [File]], CommandantError<()>> in
-            guard !useSTDIN else { return .success(filesPerConfig) }
+            guard !useSTDIN, forceExclude else { return .success(filesPerConfig) }
 
             let filteredFilesPerConfiguration = filesPerConfig.compactMap { config, files -> (Configuration, [File])? in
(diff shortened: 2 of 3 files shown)


mathaeus:
What's the status of this? We are running into this problem https://github.com/realm/SwiftLint/issues/2299 which was immediately closed and linked to a ticket which is closed as well which links to a ticket which finally led me here.

Would be nice to finally fix this :-)

cooksimo:
Can we get this merged? This is a really frustrating issue that is polluting our CI process with issues from `Pods` whenever they are updated.

fabb:
Ah ok, so this PR fixes `--force-exclude` flag being ignored, right?

cooksimo:
@marcelofabri @jpsim Can we get an update on this PR? What is the hold up? Is the fix approach ok? If so can we get this merged? It is incredibly frustrating to run in to a bug, have a PR up to fix that bug and then have it sit around for 2 months with not even a single comment.

luvacu:
Hi @jpsim,

I know you might be very busy, but could you please provide feedback about this PR? This is something I, and others, am interested in, and would really like to be merged. Thank you very much for your time!

Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
```diff
@@ -2,6 +2,13 @@ import Foundation
 import SourceKittenFramework
 
 extension Configuration {
+    public func lintableFiles(in files: [File], forceExclude: Bool) -> [File] {
+        return files.compactMap { file -> File? in
```
jpsim: Shouldn't this be

```swift
return files.flatMap { file in
    return file.path.map({ lintableFiles(inPath: $0, forceExclude: forceExclude) }) ?? []
}
```

so that you can pass directories to `--path` or `--paths` and have it still work?

Source/swiftlint/Extensions/Configuration+CommandLine.swift
```diff
@@ -48,6 +48,7 @@ private func autoreleasepool(block: () -> Void) { block() }
 
 extension Configuration {
 
+    // swiftlint:disable:next function_body_length
```
jpsim: Can we avoid making this function exceed the function body length limit? This rule is in place for a reason, it makes it hard to read and understand what the code is doing when the function is this large.

It should be possible to extract the body of the `flatMap` you added into a function, which would certainly help with readability.

jpsim:
I left some comments. I think this implementation regresses when passing directories instead of just files.

luvacu:
Thank you very much @jpsim!
@inket Could you please review and address @jpsim 's feedback? Otherwise, please let me know, and I'll try to apply the fixes myself if that's OK.

Thank you very much!!

inket:
@luvacu You're free to take it

mihai8804858:
Is this PR gonna be merged soon? We're running into same issue with `--use-script-input-files` from `pre-commit` hook and we'd love to see this released!

PR #2289, opened 2018-07-11
Fix literal multiline element false positives

erichoracek:
The `literal_expression_end_indentation` rule was diagnosing multiline literal elements as failing, e.g. the following would trigger a failure:

```swift
[
  .init(
    a: 1,
    b: 2)]
```

While this would not:
```swift
[
  1]
```

This was triggering some larger issues where autocorrection of `literal_expression_end_indentation` could incorrectly remove parts of multiline literal elements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,10 @@
   [Norio Nomura](https://github.com/norio-nomura)
   [SR-7954](https://bugs.swift.org/browse/SR-7954)
 
+* Fix `literal_expression_end_indentation` false positives with multiline 
+  literal elements.  
+  [Eric Horacek](https://github.com/erichoracek)
+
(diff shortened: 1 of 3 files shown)


marcelofabri:
I'm not super sure this is the right answer - I can see value in the current behavior, but we probably should make the other example (below) to trigger as well.

```swift
[
  1]
```

Maybe having a configuration would be better?

Dschee:
I agree with @marcelofabri here, I don't think any of the two provided examples is a false-positive, if anything we could make the current behavior more consistent – either triggering for both or not with a configuration option. @erichoracek Would you volunteer in adding appropriate triggering and non-triggering examples to the tests and update the impelemntation appropriately?

If not, I'm for closing this PR as it is right now since it doesn't fix any bug as implied by the OP.

PR #2292, opened 2018-07-11
Vertical whitespace between cases

### LINKED ISSUE #1517: Rule Request: Vertical Whitespace between Cases
This is a rule request. I'm willing to implement this myself and send a PR, but before doing that I'd like to know if the rule is wanted or not. To explain the desired rule, I'd like to first post the **custom rule** we are already using in our projects:

```yaml
  vertical_whitespace_between_cases:
    included: ".*.swift"
    regex: '[^\n){][ \t]*\n[ \t]*(?:case[^:(\n]+|default):(?:\\s*\/(\/\*).*)?'
    name: "Vertical Whitespace Between Cases"
    message: "Include a vertical whitespace (empty line) between cases in switch statements."
    severity: warning
```

I felt like this rule might be useful for other projects as well, which is why I'm posting it. What do you think about it? For us it helps a lot to prevent common issues and enforce a shared style.

**Rationale**

We had this discussion a few times, the common ground we were all happy with is this: When using switch-case there should be a newline before a new case begins except for the last case, because there's already a closing parentheses and indentation there as visual clues.

A few examples:

```swift
let x = 5

// ❌ not acceptable
switch x {
case 0..<5:
    // some code
case 5..<10:
    // some code
default:
    // some code
}

// ❌not acceptable
switch x {
case 0..<5:
    // some code

case 5..<10:
    // some code

default:
    // some code

}

// ✅ acceptable
switch x {
case 0..<5:
    // some code

case 5..<10:
    // some code

default:
    // some code
}
```

Dschee:
Fixes #1517.

Please note that I made this rule opt-in for now, but I personally think this could also be turned on by default. When I ran the tests with this turned on by default (replacing `OptInRule` with `Rule`) many files were updated (via autocorrect) and looked IMHO better afterwards.


CHANGELOG.md
```diff
@@ -15,6 +15,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2249](https://github.com/realm/SwiftLint/issues/2249)
 
+* Add new rule `vertical_whitespace_between_cases` to warn against crowded
```
ornithocoder: I'd suggest to mention this is a opt-in rule.

Rules.md
```diff
@@ -19543,6 +19544,46 @@ class BBBB {}
...
+    }
+```
+
+</details>
+<details>
+<summary>Triggering Examples</summary>
+
+```swift
+    switch x {
+    case 0..<5:
```
ornithocoder: This triggering examples are not showing where the violation is triggered. Maybe this should be added, to match the other rules.

Source/SwiftLintFramework/Rules/VerticalWhitespaceBetweenCasesRule.swift
```diff
@@ -0,0 +1,89 @@
...
+    private static let violatingToValidExamples: [String: String] = [
+    """
+        switch x {
+        case 0..<5:
+            print("x is valid")
+        default:
+            print("x is invalid")
+        }
+    """: """
+        switch x {
```
ornithocoder: It's missing is the location where the violations are triggered. The examples should have "↓" to indicate where the violation is triggered. The Unit Tests will use the location of the arrow to make sure the violations are triggered on the correct place.

Source/SwiftLintFramework/Rules/VerticalWhitespaceBetweenCasesRule.swift
```diff
@@ -0,0 +1,89 @@
...
+        nonTriggeringExamples: Array(Set(violatingToValidExamples.values)),
+        triggeringExamples: Array(Set(violatingToValidExamples.keys)),
+        corrections: violatingToValidExamples
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        return file.violatingRanges(for: pattern).map {
+            StyleViolation(
+                ruleDescription: type(of: self).description,
+                severity: ViolationSeverity.warning,
```
ornithocoder: Most rules allow developers to set the severity of the violations triggered. One might want to set it to error instead or warning.

CHANGELOG.md
Dschee: Done.

Rules.md
Dschee: Done.

Source/SwiftLintFramework/Rules/VerticalWhitespaceBetweenCasesRule.swift
Dschee: Done.

Source/SwiftLintFramework/Rules/VerticalWhitespaceBetweenCasesRule.swift
Dschee: Done.

Dschee:
@ornithocoder Thank you for your feedback, I tackled all issues – please have another look! :)

Dschee:
I just rebased this to fix merge conflicts and updated to the new `AutomaticTestableRule` feature.

Dschee:
Just rebased to solve merge conflicts.

Source/SwiftLintFramework/Rules/VerticalWhitespaceBetweenCasesRule.swift
```diff
@@ -0,0 +1,149 @@
+import Foundation
+import SourceKittenFramework
+
+private extension File {
+    func violatingRanges(for pattern: String) -> [NSRange] {
+        return match(pattern: pattern, excludingSyntaxKinds: SyntaxKind.commentAndStringKinds)
+    }
+}
+
+private extension Dictionary where Key: StringProtocol, Value: StringProtocol {
```
marcelofabri: you can use `Key == String, Value == String` since you're casting them anyways

Source/SwiftLintFramework/Rules/VerticalWhitespaceBetweenCasesRule.swift
```diff
@@ -0,0 +1,149 @@
...
+            print("x is invalid")
+        }
+    """
+    ]
+
+    private let pattern = "([^\\n{][ \\t]*\\n)([ \\t]*(?:case[^\\n]+|default):[ \\t]*\\n)"
+}
+
+extension VerticalWhitespaceBetweenCasesRule: OptInRule, AutomaticTestableRule {
+    public var configurationDescription: String { return "N/A" }
```
marcelofabri: you shouldn't override this

Source/SwiftLintFramework/Rules/VerticalWhitespaceBetweenCasesRule.swift
```diff
@@ -0,0 +1,149 @@
...
+        switch x {
+        case 0..<5: print("x is low")
+        case 5..<10: print("x is high")
+        default: print("x is invalid")
+        }
+        """
+    ]
+
+    private static let violatingToValidExamples: [String: String] = [
+    """
```
marcelofabri: can you add a few more examples?

Source/SwiftLintFramework/Rules/VerticalWhitespaceBetweenCasesRule.swift
```diff
@@ -0,0 +1,149 @@
...
+        kind: .style,
+        nonTriggeringExamples: Array(Set(violatingToValidExamples.values)) + nonTriggeringExamples,
+        triggeringExamples: Array(Set(violatingToValidExamples.keys)),
+        corrections: violatingToValidExamples.cleanedKeysDict()
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        let patternRegex: NSRegularExpression = regex(pattern)
+
+        return file.violatingRanges(for: pattern).map { violationRange in
```
marcelofabri: this should be done using the AST instead of a regex. You can check `NoFallthroughOnlyRule` which does something similar and the discussion on the PR: https://github.com/realm/SwiftLint/pull/2194

Source/SwiftLintFramework/Rules/VerticalWhitespaceBetweenCasesRule.swift
```diff
@@ -0,0 +1,149 @@
...
+    public init(configuration: Any) throws {}
+
+    public static let description = RuleDescription(
+        identifier: "vertical_whitespace_between_cases",
+        name: "Vertical Whitespace Between Cases",
+        description: "Include a vertical whitespace (empty line) between cases in switch statements.",
+        kind: .style,
+        nonTriggeringExamples: Array(Set(violatingToValidExamples.values)) + nonTriggeringExamples,
+        triggeringExamples: Array(Set(violatingToValidExamples.keys)),
+        corrections: violatingToValidExamples.cleanedKeysDict()
```
marcelofabri: you don't need to remove the markers from the keys here. It's actually better to leave them.

Dschee:
Just rebased this branch onto current master.

Source/SwiftLintFramework/Rules/VerticalWhitespaceBetweenCasesRule.swift
Dschee: Override what?

Source/SwiftLintFramework/Rules/VerticalWhitespaceBetweenCasesRule.swift
Dschee: When I let go off the `cleanedKeysDict` then 6 tests are failing. That's why I'm keeping it...

Source/SwiftLintFramework/Rules/VerticalWhitespaceBetweenCasesRule.swift
Dschee: Thanks for the pointer, looks like I'd need to basically rewrite the entire logic when switching to AstRule for which I don't have the time right now. I'll try to convert it when I have more time. Would you accept it as it is and we improve it later? I could leave a refactor comment ...

Source/SwiftLintFramework/Rules/VerticalWhitespaceBetweenCasesRule.swift
jpsim: Overriding `configurationDescription` and `init(configuration: Any)` prevents the default violation severity configuration versions of these from working. I fixed this in this PR.

jpsim:
Although I agree this would be better as an ASTRule, given that it's opt-in I see no issue with merging with the regex implementation.

*However. That being said, SwiftLint spends most of its time executing regular expressions and we'll need to come up with a strategy for moving away from them if we want to improve SwiftLint's performance.

PR #2296, opened 2018-07-18
File Types Order & Type Contents Order

Dschee:
This fixes #2294 by introducing two new rules:
- `file_types_order`: Specifies how the types within a file should be ordered.
- `type_contents_order`: Specifies the order of subtypes, properties, methods & more within a type.

PR #2291, opened 2018-07-11
Vertical whitespace opening & closing braces

Dschee:
Fixes #1518.

Please note that I made this rule opt-in for now, but I personally think this could also be turned on by default. When I ran the tests with this turned on by default (replacing `OptInRule` with `Rule`) many files were updated (via autocorrect) and looked IMHO better afterwards.

PR #2302, opened 2018-07-20
Multiline Brackets

### LINKED ISSUE #2306: Multiline Brackets: consistent style with other bracket conventions
## Rationale

When trying to enforce the same code style amongst all team members one issue that often arises is the question, how multi-line literals should be styled. What I mean by `multi-line literals` is arrays and dictionaries, which are written very differently by every team member if no style is enforced. While one might opt to accept all styles so long as other style rules are not violated, this rule should enforce a clear style which is oriented by the community-accepted style for other multi-line bracket based code like classes and method calls. For those we already have clear conventions:

```swift
class SomeType {
    let x: Int
    let y: Int

    func sum() -> Int {
        return x = y
    }

    func viewDidLoad() {
        print("x: \(x)")
        print("x: \(x)")
        let sum = sum()

        let evenNums = [x, y].filter { $0 % 2 == 0 }.map { num in
            // some multi-line code
            // some multi-line code
            // some multi-line code
        }
    }
}
```

While there's of course different styles, one standard widely accepted by the community is to either keep everything in one line including the brackets (like in `[x, y].filter { $0 % 2 == 0 }` or to put the opening bracket and closing bracket into their own line (like for the class body, the body of the function `sum`, the body of the function `viewDidLoad` and the body of the `map` call).

The same rationale can be applied to function parameters and function arguments as well to enforce a unified multiline style.

## Example

The requested opt-in rule requires array and dictionary literals to be styled according to the same rules, to be consistent with other multi-line bracket structure style, for example:

```swift
            let trio = ["harry",
                        "ronald",
                        "hermione"]

            let houseCup = ["gryffinder": 460,
                            "hufflepuff": 370,
                            "ravenclaw": 410,
                            "slytherin": 450]
```

This code is not bad per se, but it could lead to line-length problems when the name of the variable is relatively long. Also renaming the variable in one branch and changing the value of the first literal entry in another branch would lead to a merge conflict. To fix these problems and at the same time be consistent with other brackets, here's the style to be enforced by the new rule:

```swift
            let trio = [
                "harry",
                "ronald",
                "hermione"
            ]

            let houseCup = [
                "gryffinder": 460,
                "hufflepuff": 370,
                "ravenclaw": 410,
                "slytherin": 450
            ]
```

This fixes the line-length problem, prevents merge-conflicts when both the first literal object and the literal variable name are changed in different branches and is consistent with other bracket rules as it puts the opening and closing braces into their own line, just like usually done with `{` and `}`.

Dschee:
This fixes #2306.


Dschee:
Same issue here like in #2296, the `offset` simply isn't right when the prefix `/* 👨‍👩‍👧‍👦👨‍👩‍👧‍👦👨‍👩‍👧‍👦 */` is added. Still no idea why this happens, glad if somebody can help!

marcelofabri:
@Dschee I wasn't able to review the PRs yet, but you're likely mixing UTF-8 and UTF-16 positions.

SourceKit always works with UTF-8 byte offsets, so anything that you get from it (structure and syntax map) use that. Regex matching (and some NSString operations) use UTF-16 code units.

Dschee:
I just fixed all issues and rebased this to the latest master. Ready to merge from my point of view! 🎉

Dschee:
@ornithocoder Thank you for reviewing the PR, I just addressed all points, including creating an issue as reference to this rule request: #2306. I hope it looks good now!

CHANGELOG.md
```diff
@@ -23,6 +23,10 @@
   [Ornithologist Coder](https://github.com/ornithocoder)
   [#2283](https://github.com/realm/SwiftLint/issues/2283)
 
+* Add new opt-in rule `multiline_literal_brackets` to warn against multiline
+  literal arrays & dictionaries with surrounding brackets without newline.  
+  [Cihat Gündüz](https://github.com/Dschee)
```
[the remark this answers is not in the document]
Dschee: There wasn't, but now I created an issue (#2306).

Source/SwiftLintFramework/Rules/MultilineParametersBracketsRule.swift
```diff
@@ -0,0 +1,181 @@
...
+        for substructure in substructure.substructure {
+            violations += self.violations(in: substructure, file: file)
+        }
+
+        return violations
+    }
+
+    private func openingBracketViolation(parameters: [[String: SourceKitRepresentable]],
+                                         file: File) -> StyleViolation? {
+        if
```
ornithocoder: Here's another example where `guard` could be used for early return (of `nil`, in this case). Pretty much everything can be guarded in this method, leaving the return on the happy path.

```swift
guard
    let firstParamByteOffset = ....,
    let firstParamByteLength = ....,
    ...,
    case let prefix = ...,
    case let invalidPrefixRegex = ...,
    let invalidMatch = invalidPrefixRegex.firstMatch(in: prefix, options: [], range: prefix.fullNSRange) else { return nil }

return StyleViolation()
````

File: Source/SwiftLintFramework/Rules/MultilineParametersBracketsRule.swift [hunk 3]
```diff
@@ -0,0 +1,181 @@
...
+                )
+            }
+        }
+
+        return nil
+    }
+
+    private func closingBracketViolation(parameters: [[String: SourceKitRepresentable]],
+                                         file: File) -> StyleViolation? {
+        if
```
ornithocoder: Same here.

Source/SwiftLintFramework/Rules/MultilineParametersBracketsRule.swift
Dschee: Fixed.

Source/SwiftLintFramework/Rules/MultilineParametersBracketsRule.swift
Dschee: Fixed, too.

Dschee:
I just integrated @ornithocoder's feedback and fixed a bug.
Is there anything else preventing this from being merged?

marcelofabri:
@Dschee new rules are usually bigger PRs that can take a while for contributors to review.

Dschee:
Just rebased to solve merge conflicts.

Dschee:
@marcelofabri I see, makes sense. Will wait then and rebase from time to time. :)

Dschee:
Just rebased and fixed an issue with missing CHANGELOG entries for two rules.

Dschee:
@marcelofabri Any estimate on when my new rules might be reviewed? I'm waiting over 7 weeks now without any updates, while during this time 6 other new rules were already merged: #2385, #2381, #2369, #2367, #2366, #2342 

Here are my new rule suggestions, I'd appreciate your feedback:
#2302 (this one), #2296, #2292, #2291

Dschee:
Just rebased this branch onto current master.

jpsim:
> Please don't get me wrong, but I'm slowly starting to believe this project isn't actively maintained anymore.

Hi Cihat and first off, thank you very much for your significant contributions these last few months! I really appreciate it. Lots of people rely on SwiftLint and your efforts to improve their experience is impactful.

I want to apologize for the reduced activity from maintainers to review and discuss issues or pull requests lately. This wasn't a conscious decision by anyone but likely more the people with write access collectively being busy with life and spending less time working on SwiftLint.

Despite this, it's a stretch to call SwiftLint unmaintained, there have been 5 releases in the last 5 months, with over 80 changelog entries in that time, including some very significant new features like analyzer rules.

> Apart from the new rule suggestions posted several months ago, I have also posted two bugfix PRs (#2426 and #2418) about 6 weeks ago without any comment. Both of them are very small changes with tests and CHANGELOG entries, so they could have been merged with a minute-long review and a click of a button, but they aren't.

I just spent a few evenings going through my backlog of SwiftLint notifications and reviewed, updated and merged these two PRs and I can tell you it was significantly more involved than a minute-long review and a click of a button. The biggest contributor to SwiftLint's processing time is executing regular expressions, and the two involved in these PRs are particularly complex or expensive.

Not only that, but there were unintended consequences to #2418, where that fixture was inadvertently picked up by Sourcery when generating our `LinuxMain.swift` file which I had to fix in 827410a4cd24e3b3e0c17d4ecc8e377ce6d492c4.

You did nothing wrong here, I could have made the exact same changes, but I'm just pointing out that reviewing community PRs may be more involved than you think.

> I know how time consuming maintaining open source software can be, so if not having enough time is the issue, I'd be happy to **help as a maintainer** of this repo if you accept me as such.

That's **fantastic!** 🎉 I've been impressed with your contributions over the last few months. I just sent you an invite to gain write access to this project.

> We are really relying on an up-to-date linter to enforce code uniformity and code quality here at Jamit Labs in Karlsruhe, Germany and I would rather not start a new competing linter project if I don't have to. If this project however keeps to stay unmaintained like it is right now, I don't see any other way though.

Lots of people use this project and rely on it being accurate, fast, comprehensive and actively maintained. Now that you have write access, hopefully you'll also help with ensuring this, collectively with other maintainers. ✊ 

> I highly appreciate all the work that was done until now, but I hope you understand that having SwiftLint configured to fail on our CI even if a single warning is shown really makes us less productive if we can't fix any false positives at the minimum (and add new autocorrectable rules at the optimum).

💯

jpsim:
Regarding this PR specifically, looks good overall! I made some small refactors and I'll merge once CI is happy.

Dschee:
Oh @jpsim, I just read your text today (I was on vacation last week) and thank you for the detailed explanation. I understand all your points (and actually hoped for this being the reasons behind my changed not being merged) – of cource even small PRs can be bigger than one might think at first. I just wanted to emphasize my point a little more.

Thank you again, as I've stated elsewhere, for making me a maintainer. I will do what I can to help keeping SwiftLint as useful as possible. I agree that SwiftLint needs to stay fast, we didn't have any performance problems on our projects though, so I wasn't aware there existed any. But I will try to keep code running as fast as possible.

jpsim:
> I agree that SwiftLint needs to stay fast, we didn't have any performance problems on our projects though, so I wasn't aware there existed any.

Yeah, I don't think SwiftLint necessary has performance problems for most projects. Running SwiftLint on itself lints 124 rules on 352 files in 2.9s on a clean lint and 300ms on a cached lint. However, running it on a large iOS app project with 5,000 Swift files takes 10x as long in both clean and cached cases, so that's where performance starts becoming something worth improving.

PR #2303, opened 2018-07-20
Add autocorrect for control_statement

colejd:
Hello! This adds autocorrect support for the `control_statement` rule. 

Caveat: 
As of now, it only seems to get the outermost control blocks it can find that break the rule; to fully correct, you need to run autocorrect a few times. If anyone has a way around this, I'd love to get that in here.

Thank you!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@
   to specify that the rule should only be applied to `if` statements.  
   [Cihat Gündüz](https://github.com/Dschee)
   [#2307](https://github.com/realm/SwiftLint/issues/2307)
+* Add autocorrect support for `control_statement`.  
+  [Jonathan Cole](https://github.com/colejd)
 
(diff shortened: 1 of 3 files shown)


colejd:
It’s apparent now that there’s some optimization work that needs to be done. Also, my paren matching depends on `String.Index` being hashable, which isn’t the case in Swift 4. I’ll find fixes for these tonight.

colejd:
It looks like linux_swift_4 just kind of gave up?

colejd:
Okay, this is now in a state I would consider ready for review. Thank you for your time!

colejd:
Ugh. Something I did between "update changelog" and "optimize correction and validation" breaks autocorrect on production code, even though all the tests pass. Will investigate.

Dschee:
@colejd Any updates on this? I'd be happy to review once you consider this ready for review again!

PR #2308, opened 2018-07-23
Add `if_only` option to rule conditional_returns_on_newline

### LINKED ISSUE #2307: Add if-only config option to conditional_returns_on_newline
While I like the basic idea and rationale of the rule `conditional_returns_on_newline` I think it is more true for `if` than it is for `guard`. From my point of view, the "bas smell" of a return in the same line is that the reader of the code could overlook it when only scanning the beginning of the lines. But since `guard` statements need to leave the current scope in their `else` body somehow (e.g. by returning or by throwing an error etc.) seeing a `guard` at the beginning of a line already is a warning that their probably is some kind of return.

This is why I think this rule should be configurable to only show warnings if the return statement is within an if, but not in guards.

Dschee:
Fixes #2307.

### DIFF
--- a/Source/SwiftLintFramework/Rules/ConditionalReturnsOnNewlineRule.swift
+++ b/Source/SwiftLintFramework/Rules/ConditionalReturnsOnNewlineRule.swift
@@ -1,7 +1,7 @@
 import Foundation
 import SourceKittenFramework
 
-public struct ConditionalReturnsOnNewlineRule: ConfigurationProviderRule, Rule, OptInRule, AutomaticTestableRule {
+public struct ConditionalReturnsOnNewlineRule: ConfigurationProviderRule, Rule, OptInRule {
     public var configuration = ConditionalReturnsOnNewlineConfiguration()
 
     public init() {}
--- a/Source/SwiftLintFramework/Rules/ConditionalReturnsOnNewlineRule.swift
+++ b/Source/SwiftLintFramework/Rules/ConditionalReturnsOnNewlineRule.swift
@@ -2,7 +2,7 @@ import Foundation
 import SourceKittenFramework
 
 public struct ConditionalReturnsOnNewlineRule: ConfigurationProviderRule, Rule, OptInRule, AutomaticTestableRule {
-    public var configuration = SeverityConfiguration(.warning)
+    public var configuration = ConditionalReturnsOnNewlineConfiguration()
 
     public init() {}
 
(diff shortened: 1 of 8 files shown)


Source/SwiftLintFramework/Rules/ConditionalReturnsOnNewlineRule.swift
```diff
@@ -30,19 +30,21 @@ public struct ConditionalReturnsOnNewlineRule: ConfigurationProviderRule, Rule,
     )
 
     public func validate(file: File) -> [StyleViolation] {
-        let pattern = "(guard|if)[^\n]*return"
+        let pattern: String = configuration.ifOnly ? "(if)[^\n]*return" : "(guard|if)[^\n]*return"
```
marcelofabri: we usually don't add explicit types

Source/SwiftLintFramework/Rules/ConditionalReturnsOnNewlineRule.swift
```diff
@@ -30,19 +30,21 @@ public struct ConditionalReturnsOnNewlineRule: ConfigurationProviderRule, Rule,
...
+
         return file.rangesAndTokens(matching: pattern).filter { _, tokens in
             guard let firstToken = tokens.first, let lastToken = tokens.last,
                 SyntaxKind(rawValue: firstToken.type) == .keyword &&
                     SyntaxKind(rawValue: lastToken.type) == .keyword else {
                         return false
             }
 
-            return ["if", "guard"].contains(content(for: firstToken, file: file)) &&
+            let searchTokens: [String] = configuration.ifOnly ? ["if"] : ["if", "guard"]
```
marcelofabri: we usually don't add explicit types

Dschee: This was merely to speed up the compiler since it seems to be slow on type-checking ternary operators. But I can remove it, if you want.

Source/SwiftLintFramework/Rules/ConditionalReturnsOnNewlineRule.swift
Dschee: Removed.

Source/SwiftLintFramework/Rules/ConditionalReturnsOnNewlineRule.swift
Dschee: Removed.

Dschee:
@marcelofabri I just removed the `AutomaticTestableRule` and introduced a manual test which covers the configuration as suggested. I hope it looks good now!

PR #2310, opened 2018-07-23
Add options `prefix_pattern` and `suffix_pattern` to `file_name` rule

### LINKED ISSUE #2309: Make rule file_name accept the ending `Extension` for extensions
Currently the rule `file_name` shows a warning when there's a file named `BoolExtension` which looks like the following:

```swift
extension Bool {
    /// Equivalent to `someBool = !someBool`
    ///
    /// Useful when operating on long chains:
    ///
    ///    myVar.prop1.prop2.enabled.toggle()
    mutating func toggle() {
        self = !self // swiftlint:disable:this toggle_bool
    }
}
```

Extensions are either named `<Type>Extension` or `<Type>+MyProject` in the Swift community which I think should both be exceptions in this rule. As far as I could see, there's already an exception implemented for the latter, but not for the first.

I highly recommend to make this acceptance the default behavior as in my opinion these are false positives. But we _could_ also make this a configuration option if really need be ...

Dschee:
Fixes #2309.

### DIFF
--- a/Source/SwiftLintFramework/Rules/FileNameRule.swift
+++ b/Source/SwiftLintFramework/Rules/FileNameRule.swift
@@ -41,7 +41,7 @@ public struct FileNameRule: ConfigurationProviderRule, OptInRule {
         let prefixRegex = regex("\\A\(configuration.prefixPattern)")
         let suffixRegex = regex("\(configuration.suffixPattern)\\z")
 
-        var typeInFileName = fileName.components(separatedBy: CharacterSet(charactersIn: "+.")).first ?? fileName
+        var typeInFileName = fileName.components(separatedBy: ".").first ?? fileName
 
         if let match = prefixRegex.firstMatch(in: typeInFileName, options: [], range: typeInFileName.fullNSRange) {
             if let range = typeInFileName.nsrangeToIndexRange(match.range) {
--- a/Source/SwiftLintFramework/Rules/FileNameRule.swift
+++ b/Source/SwiftLintFramework/Rules/FileNameRule.swift
@@ -43,16 +43,14 @@ public struct FileNameRule: ConfigurationProviderRule, OptInRule {
 
         var typeInFileName = fileName.components(separatedBy: ".").first ?? fileName
 
-        if let match = prefixRegex.firstMatch(in: typeInFileName, options: [], range: typeInFileName.fullNSRange) {
-            if let range = typeInFileName.nsrangeToIndexRange(match.range) {
-                typeInFileName.removeSubrange(range)
-            }
+        if let match = prefixRegex.firstMatch(in: typeInFileName, options: [], range: typeInFileName.fullNSRange),
+            let range = typeInFileName.nsrangeToIndexRange(match.range) {
+            typeInFileName.removeSubrange(range)
         }
 
-        if let match = suffixRegex.firstMatch(in: typeInFileName, options: [], range: typeInFileName.fullNSRange) {
-            if let range = typeInFileName.nsrangeToIndexRange(match.range) {
-                typeInFileName.removeSubrange(range)
-            }
+        if let match = suffixRegex.firstMatch(in: typeInFileName, options: [], range: typeInFileName.fullNSRange),
+            let range = typeInFileName.nsrangeToIndexRange(match.range) {
+            typeInFileName.removeSubrange(range)
         }
 
         let allDeclaredTypeNames = file.structure.dictionary.recursiveDeclaredTypeNames()
--- a/Source/SwiftLintFramework/Rules/FileNameRule.swift
+++ b/Source/SwiftLintFramework/Rules/FileNameRule.swift
@@ -33,7 +33,12 @@ public struct FileNameRule: ConfigurationProviderRule, OptInRule {
             return []
         }
 
-        let typeInFileName = fileName.components(separatedBy: CharacterSet(charactersIn: "+.")).first ?? fileName
+        var typeInFileName = fileName.components(separatedBy: CharacterSet(charactersIn: "+.")).first ?? fileName
+
+        let extensionSuffix = "Extension"
+        if typeInFileName.hasSuffix(extensionSuffix) {
+            typeInFileName = String(typeInFileName.dropLast(extensionSuffix.count))
+        }
 
         let allDeclaredTypeNames = file.structure.dictionary.recursiveDeclaredTypeNames()
         guard !allDeclaredTypeNames.isEmpty, !allDeclaredTypeNames.contains(typeInFileName) else {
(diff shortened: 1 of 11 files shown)


Source/SwiftLintFramework/Rules/RuleConfigurations/FileNameConfiguration.swift
```diff
@@ -6,10 +6,15 @@ public struct FileNameConfiguration: RuleConfiguration, Equatable {
 
     private(set) public var severity: SeverityConfiguration
     private(set) public var excluded: Set<String>
+    private(set) public var prefixPattern: String
```
marcelofabri: I'm wondering if it makes sense to make those `[String]` so you could easily support multiple patterns without having to combine regexes manually. Probably overengineering though.

Source/SwiftLintFramework/Rules/RuleConfigurations/FileNameConfiguration.swift
```diff
@@ -6,10 +6,15 @@ public struct FileNameConfiguration: RuleConfiguration, Equatable {
...
+    private(set) public var suffixPattern: String
 
-    public init(severity: ViolationSeverity, excluded: [String] = []) {
+    public init(severity: ViolationSeverity, excluded: [String] = [],
+                prefixPattern: String = "", suffixPattern: String = "") {
```
marcelofabri: I'd keep the `suffixPattern` default value as `\\+.*` here just in case someone is using SwiftLintFramework directly.

Tests/SwiftLintFrameworkTests/FileNameRuleTests.swift
```diff
@@ -50,4 +54,14 @@ class FileNameRuleTests: XCTestCase {
     func testMainDoesTriggerWithoutOverride() {
         XCTAssertEqual(try validate(fileName: "main.swift", excludedOverride: []).count, 1)
     }
+
+    func testCustomSuffixPattern() {
+        XCTAssert(try validate(fileName: "BoolExtension.swift", suffixPattern: "Extensions?").isEmpty)
+        XCTAssert(try validate(fileName: "BoolExtensions.swift", suffixPattern: "Extensions?").isEmpty)
+    }
+
+    func testCustomPrefixPattern() {
```
marcelofabri: can we have a test that validates a custom prefix and a custom suffix? I don't expect it to be a common case, but it'd be good to have something to catch regressions.

Source/SwiftLintFramework/Rules/FileNameRule.swift
```diff
@@ -33,11 +38,21 @@ public struct FileNameRule: ConfigurationProviderRule, OptInRule {
...
+        let prefixRegex = regex("\\A\(configuration.prefixPattern)")
+        let suffixRegex = regex("\(configuration.suffixPattern)\\z")
+
         var typeInFileName = fileName.components(separatedBy: CharacterSet(charactersIn: "+.")).first ?? fileName
 
-        let extensionSuffix = "Extension"
-        if typeInFileName.hasSuffix(extensionSuffix) {
-            typeInFileName = String(typeInFileName.dropLast(extensionSuffix.count))
+        if let prefixMatch = prefixRegex.firstMatch(in: typeInFileName, options: [], range: typeInFileName.fullNSRange) {
+            if let range = typeInFileName.nsrangeToIndexRange(prefixMatch.range) {
```
marcelofabri: small nit: I'd combine the two `if let`s

Source/SwiftLintFramework/Rules/FileNameRule.swift
```diff
@@ -33,11 +38,21 @@ public struct FileNameRule: ConfigurationProviderRule, OptInRule {
...
+                typeInFileName.removeSubrange(range)
+            }
+        }
+
+        if let suffixMatch = suffixRegex.firstMatch(in: typeInFileName, options: [], range: typeInFileName.fullNSRange) {
+            if let range = typeInFileName.nsrangeToIndexRange(suffixMatch.range) {
```
marcelofabri: small nit: I'd combine the two `if let`s

marcelofabri:
This looks good! Just added a few small comments. Thanks!

Source/SwiftLintFramework/Rules/RuleConfigurations/FileNameConfiguration.swift
Dschee: I thought about that too, but I thought exactly the same. I opted to wait until there's somebody who has an actual use case to prevent overengineering.

marcelofabri: 👍

Tests/SwiftLintFrameworkTests/FileNameRuleTests.swift
Dschee: What do you mean by "validate a custom prefix" etc.? I mean, we already have a test that validates that custom prefixes and suffixes work, that starts exactly here in this line, so I'm not sure what you mean...

Source/SwiftLintFramework/Rules/RuleConfigurations/FileNameConfiguration.swift
Dschee: Sure, just added it back.

Tests/SwiftLintFrameworkTests/FileNameRuleTests.swift
marcelofabri: I meant a configuration that sets both at the same time.

Source/SwiftLintFramework/Rules/FileNameRule.swift
Dschee: Fixed.

Tests/SwiftLintFrameworkTests/FileNameRuleTests.swift
marcelofabri: I meant a configuration that sets both at the same time.

Source/SwiftLintFramework/Rules/FileNameRule.swift
Dschee: Fixed.

Tests/SwiftLintFrameworkTests/FileNameRuleTests.swift
Dschee: Ahh, I see, sure. Let me think of something ...

Dschee: Done.

Dschee:
I worked through your comments, thanks for the quick feedback!

PR #2313, opened 2018-07-23
Add option allow_redundancy to rule explicit_type_interface

### LINKED ISSUE #2312: Make explicit_type_interface compatible with redundant_type_annotation
Currently when the two rules `explicit_type_interface` and `redundant_type_annotation` are active, the most common way of writing a Singleton doesn't work any more. The following violates the rule `explicit_type_interface`:

```swift
final class Logger {
    // MARK: - Stored Type Properties
    static let shared = Logger() // <-- Explicit Type Interface Violation: Properties should have a type interface (explicit_type_interface)

    // MARK: - Instance Properties
    func setup() {
         // some code
    }
}
```

When fixing it by explicitly stating the type, the code violates the rule `redundant_type_annotation`:

```swift
final class Logger {
    // MARK: - Stored Type Properties
    static let shared = Logger() // <-- Redundant Type Annotation Violation: Variables should not have redundant type annotation (redundant_type_annotation)

    // MARK: - Instance Properties
    func setup() {
         // some code
    }
}
```

While this could be fixed both ways, I think the rule `explicit_type_interface` is to be fixed here. Since the goal of it is to make clear what the type is and the type is the next thing right after the `=` sign, I think this can be seen as a false positive. One could also make it an option, but since Singletons are very common and there's no real added value (correct me if I'm wrong) in stating the type explicitly again if the next thing to be seen is the exact same thing (that's exactly the point of the rule `redundant_type_annotation`),  I think we should make it non-violating by default.

Dschee:
Fixes #2312.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@
   [#2315](https://github.com/realm/SwiftLint/pull/2315)
   
 * Make rule `explicit_type_interface` compatible with rule
-  `redundant_type_annotation`.  
+  `redundant_type_annotation` via new option `allow_redundancy`.  
   [Cihat Gündüz](https://github.com/Dschee)
   [#2312](https://github.com/realm/SwiftLint/issues/2312)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,10 @@
 * Fix autocorrection for `redundant_type_annotation` rule.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2314](https://github.com/realm/SwiftLint/issues/2314)
+* Make rule `explicit_type_interface` compatible with rule
+  `redundant_type_annotation`.  
+  [Cihat Gündüz](https://github.com/Dschee)
+  [#2312](https://github.com/realm/SwiftLint/issues/2312)
 
 ## 0.26.0: Maytagged Pointers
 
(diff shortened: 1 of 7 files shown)


Dschee:
I marked this PR [WIP] for now, as I need to update this to use a custom option following the discussions in #2312.

Dschee:
Just added an option like discussed in #2312 – this is now ready to merge from my point of view. Looking forward to for reviews!

CHANGELOG.md
```diff
@@ -67,6 +67,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2233](https://github.com/realm/SwiftLint/issues/2233)
 
+* Make rule `explicit_type_interface` compatible with rule
```
marcelofabri: can you move this to the "Enhancements" section and mention the configuration option?

Dschee: Sure, just did.

Dschee:
I just rebased and moved the changelog entry accordingly.

CHANGELOG.md
```diff
@@ -35,6 +35,11 @@
 * Add `private_only` configuration to `prefixed_toplevel_constant` rule  
   [Keith Smiley](https://github.com/keith)
   [#2315](https://github.com/realm/SwiftLint/pull/2315)
+  
+* Make rule `explicit_type_interface` compatible with rule
```
marcelofabri: can you please mention that this is possible with the `allow_redundancy` configuration key?

Dschee: Done.

PR #2315, opened 2018-07-24
Add private_only to prefixed_toplevel_constant

keith:
This allows users to opt in to only validate top level constants have
the given prefix if the constant is private or fileprivate.

I'm not sure how to add a test with a configuration difference, but I'm happy to do that if it's possible.

### DIFF
--- a/Source/SwiftLintFramework/Rules/PrefixedTopLevelConstantRule.swift
+++ b/Source/SwiftLintFramework/Rules/PrefixedTopLevelConstantRule.swift
@@ -2,7 +2,7 @@ import Foundation
 import SourceKittenFramework
 
 public struct PrefixedTopLevelConstantRule: ASTRule, OptInRule, ConfigurationProviderRule, AutomaticTestableRule {
-    public var configuration = SeverityConfiguration(.warning)
+    public var configuration = PrefixedConstantRuleConfiguration(onlyPrivateMembers: false)
 
     private let topLevelPrefix = "k"
 
@@ -54,6 +54,11 @@ public struct PrefixedTopLevelConstantRule: ASTRule, OptInRule, ConfigurationPro
     public func validate(file: File,
                          kind: SwiftDeclarationKind,
                          dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        if self.configuration.onlyPrivateMembers,
+            let acl = dictionary.accessibility.flatMap(AccessControlLevel.init(identifier:)), !acl.isPrivate {
+			return []
+        }
+
         guard
             kind == .varGlobal,
             dictionary.setterAccessibility == nil,
@@ -66,7 +71,7 @@ public struct PrefixedTopLevelConstantRule: ASTRule, OptInRule, ConfigurationPro
 
         return [
             StyleViolation(ruleDescription: type(of: self).description,
-                           severity: configuration.severity,
+                           severity: configuration.severityConfiguration.severity,
                            location: Location(file: file, byteOffset: nameOffset))
         ]
     }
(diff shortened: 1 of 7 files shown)


Source/SwiftLintFramework/Rules/PrefixedTopLevelConstantRule.swift
```diff
@@ -54,6 +54,11 @@ public struct PrefixedTopLevelConstantRule: ASTRule, OptInRule, ConfigurationPro
     public func validate(file: File,
                          kind: SwiftDeclarationKind,
                          dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        if self.configuration.onlyPrivateMembers,
```
ornithocoder: Maybe this could be combined with the other `guard` below, since both are used for early exiting.

ornithocoder: You can safely drop the `self.` here.

Source/SwiftLintFramework/Rules/PrefixedTopLevelConstantRule.swift
```diff
@@ -54,6 +54,11 @@ public struct PrefixedTopLevelConstantRule: ASTRule, OptInRule, ConfigurationPro
...
+            let acl = dictionary.accessibility.flatMap(AccessControlLevel.init(identifier:)), !acl.isPrivate {
+			return []
```
ornithocoder: The indentation here looks a little bit odd.

Source/SwiftLintFramework/Rules/PrefixedTopLevelConstantRule.swift
keith: fixed

Source/SwiftLintFramework/Rules/PrefixedTopLevelConstantRule.swift
keith: huh yea, tab got in here some how

Source/SwiftLintFramework/Rules/PrefixedTopLevelConstantRule.swift
keith: I thought about that but since we only want to take into account if the ACL is private if this configuration is set, it felt like it was getting too complicated to be in that one as well.

marcelofabri:
> I'm not sure how to add a test with a configuration difference, but I'm happy to do that if it's possible.

You can do it in a similar way as this commit: https://github.com/realm/SwiftLint/pull/2308/files/73fafd5c071f59b7ed8c09a2fe493f5e202215ea..dc44fbcac37900283ea54072f2fa75c881dc950c.

You can I remove the `AutomaticTestableRule` conformance and create a new `PrefixedTopLevelConstantRuleTests` that has a test for the default configuration and one for the new configuration.

keith:
Awesome, thanks for the example, added tests for this specific configuration

Tests/SwiftLintFrameworkTests/PrefixedTopLevelConstantRuleTests.swift
```diff
@@ -0,0 +1,27 @@
...
+            "private let ↓Foo = 20.0",
+            "fileprivate let ↓foo = 20.0"
+        ]
+        let nonTriggeringExamples = [
+            "let Foo = 20.0",
+            "internal let Foo = \"Foo\"",
+            "public let Foo = 20.0"
+        ]
+
+        let alwaysOnSameLineDescription = PrefixedTopLevelConstantRule.description
```
marcelofabri: can you rename this variable? (I just noticed it's also wrong on the linked PR 😅 )

keith: Ah yea, fixed

PR #2316, opened 2018-07-24
Add support for globs in excluded paths

keith:
This does not support recursive globs, but it seems like for many cases
plain globs would be enough. This also means we don't need any libraries
to support this case.


keith:
Looking at how to fix the test length

keith:
I moved the alias related configuration tests to a new file here to help with the length issue

keith:
Happy to move something else, or do that in a separate PR if needed

Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
```diff
@@ -13,14 +23,40 @@ extension Configuration {
...
+            }
         let includedPaths = included.flatMap {
             fileManager.filesToLint(inPath: $0, rootDirectory: rootPath)
         }
         return (pathsForPath + includedPaths).filter {
             !excludedPaths.contains($0)
         }
     }
+
+    private func resolveGlobs(in pattern: String) -> [String] {
```
marcelofabri: Do you think it'd be worth it to extract this to a new class and add some tests directly?

Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
```diff
@@ -13,14 +23,40 @@ extension Configuration {
...
+        }
+
+#if os(Linux)
+        let matchCount = globResult.gl_pathc
+#else
+        let matchCount = globResult.gl_matchc
+#endif
+
+        return (0..<Int(matchCount)).compactMap { index in
+            return String(validatingUTF8: globResult.gl_pathv[index]!)
```
marcelofabri: what do you think about `return globResult.gl_pathv[index].flatMap(String.init(validatingUTF8:))` just to avoid any potencial issues?

keith: Done, but FWIW in this case IMO it's better to crash since this would be breaking the API contract.

Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
keith: Done

keith: I left the first test since it's about Configuration calling glob, and added some new glob ones

Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
keith: Had to use it in a closure since it couldn't infer the type otherwise

marcelofabri:
This looks good to me from an implementation point of view. We probably should also update the README mentioning this and how to use it.

keith:
Added

keith:
Any other thoughts here @marcelofabri ?

marcelofabri:
Thanks @keith!

keith:
Thanks!


--
Keith Smiley

PR #2339, opened 2018-08-05
Improve performance of multiple_closures_with_trailing_closure

marcelofabri:
This was taking ~1.3s on a real project. Now it takes ~0.6s when using Swift 4.1 and ~0.4 when using Swift 4.2 (as it avoids regexes completely) 🚀


marcelofabri:
Running oss-check locally got me better results so I'm merging this 🚢

PR #2341, opened 2018-08-05
Organize rules by kind in Xcode project

jpsim:
We've had them in a big flat list in the `Rules/` directory since the beginning but that stopped scaling well a long time ago. Organize them by kind.


jpsim:
Are you ok if we do this @marcelofabri? Any open PRs with rules will likely need rebasing, but I feel like this is past due to do.

PR #2343, opened 2018-08-06
Add ability for SwiftLint to lint files with full type-checked AST awareness

jpsim:
**Note: very experimental and still a work in progress**

This gives SwiftLint the ability to lint files with the assistance of the fully type-checked AST. This can enable much more powerful rules, such as linting for:

* dead code
* unused/duplicate imports
* over-exposed declarations (e.g. `internal` that could be `private`)
* superfluous type annotations (e.g. `view.tintColor = UIColor.black` vs `view.tintColor = .black`)
* use of explicit _or_ implicit `self.`

It requires passing a file with the full output from a clean & successful `xcodebuild` invocation (no incremental builds):

```shell
$ swiftlint [lint|autocorrect] --no-cache --compiler-log-path /path/to/xcodebuild.log
```

I've written a first rule to get started, which enforces the use of explicit references to `self.` when accessing instance members.

You can test this out on SwiftLint itself (which doesn't abide by this rule) by checking out this branch and running the following:

```shell
$ # Ensure derived data for SwiftLint is clean
$ xcodebuild -workspace SwiftLint.xcworkspace -scheme swiftlint > xcodebuild.log
$ echo "included:
  - Source
whitelist_rules:
  - explicit_self" > .swiftlint.yml
$ swift build -c release
$ .build/release/swiftlint autocorrect --no-cache --compiler-log-path xcodebuild.log
```

This will add explicit references to `self.` to all files linted.

The rules are expensive and awkward to run, which is why I've punted on building this for a while, but it's clear to me that it's very useful nonetheless, so hopefully we can improve this on all fronts from this starting point.

I'd like to get some basic tests running for this and clean up a few things before merging, which is why this is marked as WIP.

### DIFF
--- a/Source/SwiftLintFramework/Protocols/Rule.swift
+++ b/Source/SwiftLintFramework/Protocols/Rule.swift
@@ -7,11 +7,16 @@ public protocol Rule {
     init() // Rules need to be able to be initialized with default values
     init(configuration: Any) throws
 
+    func validate(file: File, compilerArguments: [String]) -> [StyleViolation]
     func validate(file: File) -> [StyleViolation]
     func isEqualTo(_ rule: Rule) -> Bool
 }
 
 extension Rule {
+    public func validate(file: File, compilerArguments: [String]) -> [StyleViolation] {
+        return validate(file: file)
+    }
+
     public func isEqualTo(_ rule: Rule) -> Bool {
         return type(of: self).description == type(of: rule).description
     }
@@ -32,10 +37,31 @@ public protocol ConfigurationProviderRule: Rule {
 }
 
 public protocol CorrectableRule: Rule {
+    func correct(file: File, compilerArguments: [String]) -> [Correction]
     func correct(file: File) -> [Correction]
 }
 
+public extension CorrectableRule {
+    func correct(file: File, compilerArguments: [String]) -> [Correction] {
+        return correct(file: file)
+    }
+}
+
 public protocol SourceKitFreeRule: Rule {}
+
+public protocol CompilerArgumentRule: Rule {}
+
+public extension CompilerArgumentRule {
+    func validate(file: File) -> [StyleViolation] {
+        return validate(file: file, compilerArguments: [])
+    }
+}
+
+public extension CompilerArgumentRule where Self: CorrectableRule {
+    func correct(file: File) -> [Correction] {
+        return correct(file: file, compilerArguments: [])
+    }
+}
 
 // MARK: - ConfigurationProviderRule conformance to Configurable
 
--- a/Source/SwiftLintFramework/Rules/Style/ExplicitSelfRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/ExplicitSelfRule.swift
@@ -1,0 +1,186 @@
+import Foundation
+import SourceKittenFramework
+
+public struct ExplicitSelfRule: CorrectableRule, ConfigurationProviderRule, CompilerArgumentRule, OptInRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "explicit_self",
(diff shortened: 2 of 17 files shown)


jpsim:
I've added support for testing compiler argument rules and added changelog entries, so this is no longer WIP if anyone's interested in giving this a proper review.

Source/SwiftLintFramework/Protocols/Rule.swift
```diff
@@ -32,11 +37,32 @@ public protocol ConfigurationProviderRule: Rule {
...
+        return correct(file: file)
+    }
+}
+
 public protocol SourceKitFreeRule: Rule {}
 
+public protocol CompilerArgumentRule: Rule {}
+
+public extension CompilerArgumentRule {
+    func validate(file: File) -> [StyleViolation] {
```
keith: These default extensions don't seem like they would ever work for this type of rule? Are they required or you could omit them and require implementors to add them?

Source/SwiftLintFramework/Rules/Style/ExplicitSelfRule.swift
```diff
@@ -0,0 +1,199 @@
...
+            """
+            struct A {
+                let p1: Int
+                func f1() {
+                    _ = self.p1
+                }
+            }
+            """
+        ],
+        requiresFileOnDisk: true
```
keith: This property isn't very clear to me, which files does it require?

Source/SwiftLintFramework/Rules/Style/ExplicitSelfRule.swift
```diff
@@ -0,0 +1,199 @@
...
+            let location = Location(file: file, characterOffset: range.location)
+            corrections.append(Correction(ruleDescription: description, location: location))
+        }
+        file.write(contents.bridge())
+        return corrections
+    }
+
+    private func violationRanges(in file: File, compilerArguments: [String]) -> [NSRange] {
+        guard !compilerArguments.isEmpty else {
+            return []
```
keith: Should this be an error of some kind instead?

Source/swiftlint/Extensions/File+CompilerArguments.swift
```diff
@@ -0,0 +1,136 @@
...
+#endif
+
+private func compileCommand(compilerLogs: String, sourceFile: String) -> [String]? {
+    let escapedSourceFile = sourceFile.replacingOccurrences(of: " ", with: "\\ ")
+    guard compilerLogs.contains(escapedSourceFile) else {
+        return nil
+    }
+
+    var compileCommand: [String]?
+    compilerLogs.enumerateLines { line, stop in
```
keith: Maybe a future optimization, but I wonder if it would be worth creating a dictionary of file path to commands up front that you could more easily access. I have to assume that doing this over and over is expensive with how large these files can be?

Source/SwiftLintFramework/Rules/Style/ExplicitSelfRule.swift
jpsim: Requires that `File`s that are passed in have a valid `path` on disk, unlike other uses of File, some of which are initialized with `File(contents:)` and have no path on disk.

jpsim: Definitely a large change.

jpsim:
I spoke with @keith and @soffes about this earlier today and decided to take the following next steps:

1. Add an `analyze --xcode-log-path ... [--autocorrect]` command similar to `lint` and `autocorrect`
2. Ignore `CompilerArgumentRule`s when running normal lint or autocorrect
3. Add a `analyze_rules` configuration key where `CompilerArgumentRule`s can be opted into on a per-configuration basis

jpsim:
Continued in #2379. I opted to open a new PR to keep this one for historical purposes, given I rewrote large parts of it.

Issue #2350, opened 2018-08-09
Support SonarQube issue format


junkpiano:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Proposal

This is a new feature proposal.

Looking at #175 , SwiftLint was integrated by SonarQube plugin.
Now SonarQube 7.2 provides the way to import issues from the 3rd party analyzers.

- https://www.sonarqube.org/sonarqube-7-2/
- https://docs.sonarqube.org/display/SONAR/Generic+Issue+Data

So I added another reporter for SonarQube.
I conducted tests on my personal applications, and found [my patch](https://github.com/junkpiano/SwiftLint/commit/eaff52c38884c729693378292ea68a8ea6f6b87f) works fine and sents collect data.

 It looks like this.

![imageedit_2_9847954649](https://user-images.githubusercontent.com/25413198/43886138-d4055802-9bf5-11e8-865b-687f2b381991.png)

The difference is that the issue has `SwiftLint` tag besides the issue description.
We can continuously check our code violations on SonarQube.

On the other hand, there is a limitation. We cannot use `character` in the JSON we send.
I took a look at SonarQube code, where I got a java exception error.

https://github.com/SonarSource/sonarqube/blob/4bccde70aab57079a0c7b6055528f776fac315fe/sonar-scanner-engine/src/main/java/org/sonar/scanner/externalissue/ExternalIssueImporter.java#L120

If startColumn(= `character` in SwiftLint) is specified, it throws null pointer exception, since SwiftLint `StyleViolation` doesn't have endColumn, and endLine. So I skipped `startColumn` intentionally to avoid the exception.

That said, it's appreciated if you consider adding this feature to SwiftLint.

My code is https://github.com/junkpiano/SwiftLint/commit/eaff52c38884c729693378292ea68a8ea6f6b87f.
If you are OK with it, I can send PR.

marcelofabri:
A PR would be welcome!

marcelofabri:
Implemented on #2352.

PR #2352, opened 2018-08-10
added SonarQube reporter.

junkpiano:
code for #2350 .

### DIFF
--- a/Source/SwiftLintFramework/Reporters/SonarQubeReporter.swift
+++ b/Source/SwiftLintFramework/Reporters/SonarQubeReporter.swift
@@ -1,0 +1,35 @@
+import Foundation
+import SourceKittenFramework
+
+public struct SonarQubeReporter: Reporter {
+    public static let identifier = "sonarqube"
+    public static let isRealtime = false
+
+    public var description: String {
+        return "Reports violations in SonarQube import format."
+    }
+
+    public static func generateReport(_ violations: [StyleViolation]) -> String {
+        return toJSON(["issues": violations
+            .filter { $0.location.relativeFile != nil && $0.location.line != nil }
+            .map(dictionary(for:))
+            ])
+    }
+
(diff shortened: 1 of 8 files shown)


Source/SwiftLintFramework/Reporters/SonarQubeReporter.swift
```diff
@@ -0,0 +1,35 @@
...
+    public static let identifier = "sonarqube"
+    public static let isRealtime = false
+
+    public var description: String {
+        return "Reports violations in SonarQube import format."
+    }
+
+    public static func generateReport(_ violations: [StyleViolation]) -> String {
+        return toJSON(["issues": violations
+            .filter { $0.location.relativeFile != nil && $0.location.line != nil }
```
marcelofabri: Is there a reason to filter? We already have to use optional chaining below when creating the dictionary

Source/SwiftLintFramework/Reporters/SonarQubeReporter.swift
```diff
@@ -0,0 +1,35 @@
...
+            .map(dictionary(for:))
+            ])
+    }
+
+    // refer to https://docs.sonarqube.org/display/SONAR/Generic+Issue+Data
+    static func dictionary(for violation: StyleViolation) -> [String: Any] {
```
marcelofabri: `private`?

junkpiano: yes. I modified.

Source/SwiftLintFramework/Reporters/SonarQubeReporter.swift
junkpiano: Exactly. It was kind of error handling, but I tried again without filtering, and I found it works. so I removed this filtering.

junkpiano:
@marcelofabri updated. and rebased on the lastest master to resolve conflicts.

PR #2379, opened 2018-09-01
Add ability for SwiftLint to lint files with full type-checked AST awareness

jpsim:
**Note: very experimental, continued from #2343**

This gives SwiftLint the ability to lint files with the assistance of the fully type-checked AST. This can enable much more powerful rules, such as linting for:

* dead code
* unused/duplicate imports
* over-exposed declarations (e.g. `internal` that could be `private`)
* superfluous type annotations (e.g. `view.tintColor = UIColor.black` vs `view.tintColor = .black`)
* use of explicit _or_ implicit `self.`

It requires passing a file with the full output from a clean & successful `xcodebuild` invocation (no incremental builds):

```shell
$ swiftlint analyze [--autocorrect] --compiler-log-path /path/to/xcodebuild.log
```

I've written a first rule to get started, which enforces the use of explicit references to `self.` when accessing instance members.

You can test this out on SwiftLint itself (which doesn't abide by this rule) by checking out this branch and running the following:

```shell
$ # Ensure derived data for SwiftLint is clean
$ xcodebuild -workspace SwiftLint.xcworkspace -scheme swiftlint > xcodebuild.log
$ echo "included:
  - Source
analyzer_rules:
  - explicit_self" > .swiftlint.yml
$ swift build -c release
$ .build/release/swiftlint analyze --autocorrect --compiler-log-path xcodebuild.log
```

This will add explicit references to `self.` to all files linted.

The rules are expensive and awkward to run, which is why I've punted on building this for a while, but it's clear to me that it's very useful nonetheless, so hopefully we can improve this on all fronts from this starting point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,7 @@ swiftlint(
 $ swiftlint help
 Available commands:
 
+   analyze         [Experimental] Run analysis rules
    autocorrect     Automatically correct warnings and errors
    generate-docs   Generates markdown documentation for all rules
    help            Display general or command-specific help
@@ -153,8 +154,8 @@ Available commands:
 Run `swiftlint` in the directory containing the Swift files to lint. Directories
 will be searched recursively.
 
-To specify a list of files when using `lint` or `autocorrect` (like the list of
-files modified by Xcode specified by the
+To specify a list of files when using `lint`, `autocorrect` or `analyze`
+(like the list of files modified by Xcode specified by the
 [`ExtraBuildPhase`](https://github.com/norio-nomura/ExtraBuildPhase) Xcode
 plugin, or modified files in the working tree based on `git ls-files -m`), you
 can do so by passing the option `--use-script-input-files` and setting the
@@ -322,6 +323,8 @@ excluded: # paths to ignore during linting. Takes precedence over `included`.
   - Source/ExcludedFolder
   - Source/ExcludedFile.swift
   - Source/*/ExcludedFile.swift # Exclude files with a wildcard
+analyzer_rules: # Rules run by `swiftlint analyze` (experimental)
+  - explicit_self
 
 # configurable rules can be customized from this configuration file
 # binary rules can set their severity level
@@ -436,6 +439,17 @@ Please make sure to have backups of these files before running
 Standard linting is disabled while correcting because of the high likelihood of
 violations (or their offsets) being incorrect after modifying a file while
 applying corrections.
+
+### Analyze (experimental)
+
+The _experimental_ `swiftlint analyze` command can lint Swift files using the
+full type-checked AST. The compiler log path containing the clean `swiftc` build
+command invocation (incremental builds will fail) must be passed to `analyze`
+via the `--compiler-log-path` flag.
+e.g. `--compiler-log-path /path/to/xcodebuild.log`
+
+This command and related code in SwiftLint is subject to substantial changes at
+any time while this feature is marked as experimental.
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -305,6 +305,9 @@ Rule inclusion:
 * `whitelist_rules`: Acts as a whitelist, only the rules specified in this list
   will be enabled. Can not be specified alongside `disabled_rules` or
   `opt_in_rules`.
+* `analyzer_rules`: This is an entirely separate list of rules that are only
+  run by the `analyze` command. All analyzer rules are opt-in, so this is the
+  only configurable rule list (there is no disabled/whitelist equivalent).
 
 ```yaml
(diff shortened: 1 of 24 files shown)


Rules.md
```diff
@@ -5344,6 +5345,63 @@ func foo() -> [String] {
 
 
 
+## Explicit Self
+
+Identifier | Enabled by default | Supports autocorrection | Kind | Minimum Swift Compiler Version
+--- | --- | --- | --- | ---
+`explicit_self` | Disabled | Yes | style | 3.0.0
```
marcelofabri: we probably should list if this is an `analyzer` rule here (but we can do it later in another PR)

README.md
```diff
@@ -322,6 +323,8 @@ excluded: # paths to ignore during linting. Takes precedence over `included`.
   - Source/ExcludedFolder
   - Source/ExcludedFile.swift
   - Source/*/ExcludedFile.swift # Exclude files with a wildcard
+analyzer_rules: # Rules run by `swiftlint analyze` (experimental)
+  - explicit_self
```
marcelofabri: how does this play together with the other lists?

marcelofabri: also this probably should also be mentioned on the list before this snippet

Source/SwiftLintFramework/Protocols/Rule.swift
```diff
@@ -32,11 +37,32 @@ public protocol ConfigurationProviderRule: Rule {
...
 
+public extension CorrectableRule {
+    func correct(file: File, compilerArguments: [String]) -> [Correction] {
+        return correct(file: file)
+    }
+}
+
 public protocol SourceKitFreeRule: Rule {}
 
+public protocol AnalyzerRule: Rule {}
```
marcelofabri: should it inherit from `OptInRule` for now?

Source/SwiftLintFramework/Rules/Style/ExplicitSelfRule.swift
```diff
@@ -0,0 +1,203 @@
...
+                queuedPrintError("couldn't convert offsets")
+                return nil
+            }
+
+            return contents.byteRangeToNSRange(start: Int(byteOffset), length: 0)
+        }
+    }
+}
+
+private let kindsToFind = Set([
```
marcelofabri: you could avoid the runtime creation from an array: 

```swift
private let kindsToFind: Set = [
// ...
]
```

Rules.md
jpsim: Done

README.md
jpsim: Added an explanation

Source/SwiftLintFramework/Protocols/Rule.swift
jpsim: Good call, done.

Source/SwiftLintFramework/Rules/Style/ExplicitSelfRule.swift
jpsim: Done

README.md
```diff
@@ -437,6 +440,17 @@ Standard linting is disabled while correcting because of the high likelihood of
...
 
+### Analyze (experimental)
+
+The _experimental_ `swiftlint analyze` command can lint Swift files using the
+full type-checked AST. The compiler log path containing the clean `swiftc` build
+command invocation (incremental builds will fail) must be passed to `analyze`
+via the `--compiler-log-path` flag.
+e.g. `--compiler-log-path /path/to/xcodebuild.log`
+
+This command and related code in SwiftLint is subject to substantial changes at
```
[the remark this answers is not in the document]
jpsim: Added.

norio-nomura:
I opened https://github.com/jpsim/SourceKitten/pull/555 that parsing compiler arguments from `llbuild` manifest used in New Build System by Xcode. That can avoid rebuilding to ensure compiler arguments from `xcodebuild.log`.
I think the same method can be used with SwiftLint's type-checked AST linting.

jpsim:
Nice, thanks! I was going to do something like that next. We should expose this kind of stuff in SourceKitten IMO so that downstream tools like SwiftLint don't have to worry about it.

Although I'm not 100% sure that the compiler arguments without building yields the same result as when the targets are fully built. I think some of the SourceKit requests might return different data, like USRs not being fully resolved, among other things. This will need further investigation.

norio-nomura:
Oh, if the correct compiler arguments were passed to sourcekit, I did not think that rebuilding are unnecessary. And it seems to work without rebuilding.

jpsim:
This feature was released in 0.28.0.

PR #2381, opened 2018-09-02
Add Unused Import analyzer rule

jpsim:


Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
```diff
@@ -0,0 +1,157 @@
...
+        // Always disallow 'import Swift' because it's available without importing.
+        usrFragments = usrFragments.filter { $0 != "Swift" }
+        let unusedImports = imports.filter { !usrFragments.contains($0) }
+        return unusedImports.map { module in
+            return (module, contents.bridge().range(of: "import \(module)\n"))
+        }
+    }
+}
+
+private let syntaxKindsToSkip: [SyntaxKind] = [
```
marcelofabri: what do you think about making this a `Set`?

marcelofabri:
@jpsim Have you tried running this rule against Lyft's iOS app? I've run it against one app and gave some false positives - I think they're all related to importing a framework that imports other ones.

For example, one class was importing `Photos` to use `AVCaptureDevice`. That import was removed and the file doesn't compile anymore. Other examples are importing `Firebase` (instead of the specific Firebase framework).

jpsim:
Thats expected. If you import UIKit but only use `CGPoint`, a more appropriate import would be CoreGraphics for example.

However, there are times when autocorrect will cause the file to no longer compile, because the wider import wasn’t replaced by a more specific one.

PR #2383, opened 2018-09-02
Run format before autocorrect

### LINKED ISSUE #2374: Undesired behavior with --format
### New Issue Checklist

### Bug Report

As `closure_end_indentation`'s description states, a closure end should have the same indentation as the line that started it, but this behavior conflicts with Xcode's indentation (^i) and SwiftLint's `--format`.

Below, a sample `.swiftlint.yml` file with the rule enabled, a swift file - `file.swift` - with one of the examples from `swiftlint rules closure_end_indentation` and the output, showing that the code is correct (closure ending with correct indentation, as imposed by the rule).

```bash
$ cat .swiftlint.yml
opt_in_rules:
  - closure_end_indentation

$ cat file.swift
SignalProducer(values: [1, 2, 3])
    .startWithNext { number in
        print(number)
    }

$ swiftlint lint
Linting Swift files at paths 
Linting 'file.swift' (1/1)
```

However, when the command `swiftlint autocorrect --format` runs, it autocorrects the file and reformats it, reverting the autocorrection applied by the rule.

```bash
$ swiftlint autocorrect --format
Correcting Swift files at paths 
Correcting 'file.swift' (1/1)
Done correcting 1 file!

$ cat file.swift
SignalProducer(values: [1, 2, 3])
    .startWithNext { number in
        print(number)
}
```

As result, `autocorrect` has to run twice - first to format and then to correct the files -, adding a undesired delay to the build phase:

```bash
$ swiftlint autocorrect --format
$ swiftlint autocorrect
```

Ideally `swiftlint autocorrect --format` should first format and then autocorrect the files.

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? `master`
* Installation method used (Homebrew, CocoaPods, building from source, etc)? `swift build`

```yml
opt_in_rules:
  - closure_end_indentation
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? no.
* Which Xcode version are you using (check `xcode-select -p`)? 9.4.1

ornithocoder:
Fixes #2374.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,7 +66,7 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2360](https://github.com/realm/SwiftLint/issues/2360)
 
-* Changes `autocorrect --format` to run format before autocorrect, fixing
+* Change `autocorrect --format` to run format before autocorrect, fixing
   conflicts between default indentation and rules which modify indentation
   (i.e. `closure_end_indentation`).  
   [Ornithologist Coder](https://github.com/ornithocoder)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,12 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2360](https://github.com/realm/SwiftLint/issues/2360)
 
+* Changes `autocorrect --format` to run format before autocorrect, fixing
+  conflicts between default indentation and rules which modify indentation
+  (i.e. `closure_end_indentation`).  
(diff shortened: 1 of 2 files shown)


jpsim:
In either order, there can be problems, since there are times when a SwiftLint rule correction will also need reformatting. However, after looking through our rules that support autocorrection, I think you're right that format then autocorrect is the better order, because SwiftLint's modifications are more specific than SourceKit's.

👍 

Thanks for the PR!

ornithocoder:
Thanks @jpsim! :-)

PR #2385, opened 2018-09-03
Add UnusedPrivateDeclarationRule

jpsim:


jpsim:
Seems like this rule is crashing SourceKit on the (fairly old) Xcode 10 beta version that CircleCI has. Gonna keep this open until a new Xcode 10 image is available.

marcelofabri:
Does this catch using a private function in `#selector`? If so, what do you think about adding a test to document this?

jpsim:
Yes, it would catch that. There are so many potential test cases to include for this rule that I just haven’t added basically any. I should.

PR #2392, opened 2018-09-06
New Rule - Collection Element Alignment

### LINKED ISSUE #2326: Rule request: Vertical alignment of collection types
### New Issue Checklist

### Rule Request

This issue is to propose the addition of a new style rule that would check for vertical alignment of collection types (such as with arrays, dictionaries, and sets).

The proposed rule would ensure that collections that span multiple lines would be vertically aligned on the first character, or on colons for dictionaries.

**Triggering examples**

```swift
// ❌ Coordinates not aligned on first character
let coordinates = [
    CLLocationCoordinate2D(latitude: 0, longitude: 33),
        CLLocationCoordinate2D(latitude: 0, longitude: 66),
    CLLocationCoordinate2D(latitude: 0, longitude: 99)
]
```

```swift
// ❌ Strings not aligned on first character or on colons
let meals = [
            "breakfast": "oatmeal",
            "lunch": "sandwich",
"dinner": "burger"
]
```

```swift
// ❌ Integers not aligned on first character
var evenNumbers: Set<Int> = [
    2,
  4,
    6
]
```

**Non-triggering examples**

```swift
// ✅ Dictionary keys aligned on first character
let colorDictionary: [NSNumber: UIColor] = [
    0: .clear,
    5: .white,
    10: .green,
    15: .yellow,
    20: .red
]
```

```swift
// ✅ Dictionary keys aligned on colon
let colorDictionary: [String: UIColor] = [
          "mild": .green,
      "moderate": .yellow,
        "severe": .red
]
```

```swift
✅ Array of strings aligned on first character
let pets = [
    "cat",
    "dog",
    "fish"
]
```

No, I don't believe the rule should be configurable.

The proposed rule should be enabled by default to promote consistency in alignment, which is enforced through other rules. I'm open to this being different though.

daltonclaybrook:
This PR resolves #2326 by adding a new rule to validate that all elements in a collection literal are aligned vertically. By default, dictionary keys are validated for left-alignment, but can be configured for colon alignment instead using `align_colons: true`.


daltonclaybrook:
Looks like the job timed out. Is this something I should be able to repro locally? Running `make docker_test` locally seems to work fine for me. Any advice @marcelofabri ?

Source/SwiftLintFramework/Rules/Style/CollectionAlignmentRule.swift
```diff
@@ -0,0 +1,295 @@
...
+            }
+            return Location(file: file.path, line: line, character: character)
+        }
+    }
+
+    private func getDictionaryKeyLocations(with file: File,
+                                           dictionary: [String: SourceKitRepresentable]) -> [Location] {
+        var keys: [[String: SourceKitRepresentable]] = []
+        var values: [[String: SourceKitRepresentable]] = []
+        zip(dictionary.elements, 0...).forEach { element, index in
```
marcelofabri: what do you think about using `dictionary.elements.enumerated()` instead?

Source/SwiftLintFramework/Rules/Style/CollectionAlignmentRule.swift
```diff
@@ -0,0 +1,295 @@
...
+        let contents = file.contents.bridge()
+        let matchStart = keyOffset + keyLength
+        let matchLength = valueOffset - matchStart
+        let range = contents.byteRangeToNSRange(start: matchStart, length: matchLength)
+
+        guard let colonRange = file.match(pattern: ":", range: range).first?.0,
+            let (line, character) = contents.lineAndCharacter(forCharacterOffset: colonRange.location)
+            else { return nil }
+
+        return Location(file: file.path, line: line, character: character)
```
marcelofabri: you can init `Location` with a character offset:

```swift
return Location(file: file, characterOffset: colonRange.location)
```

Source/SwiftLintFramework/Rules/Style/CollectionAlignmentRule.swift
```diff
@@ -0,0 +1,295 @@
...
+        }
+    }
+
+    private func getColonLocation(with file: File, keyOffset: Int, keyLength: Int, valueOffset: Int) -> Location? {
+        let contents = file.contents.bridge()
+        let matchStart = keyOffset + keyLength
+        let matchLength = valueOffset - matchStart
+        let range = contents.byteRangeToNSRange(start: matchStart, length: matchLength)
+
+        guard let colonRange = file.match(pattern: ":", range: range).first?.0,
```
marcelofabri: what do you think about excluding comments just to avoid the (super unusual) case when there's a comment between the key and the value?

Source/SwiftLintFramework/Rules/Style/CollectionAlignmentRule.swift
```diff
@@ -0,0 +1,295 @@
...
+            let (line, character) = contents.lineAndCharacter(forCharacterOffset: colonRange.location)
+            else { return nil }
+
+        return Location(file: file.path, line: line, character: character)
+    }
+
+    private func getKeyLocation(with file: File, keyOffset: Int) -> Location? {
+        let contents = file.contents.bridge()
+        guard let (line, character) = contents.lineAndCharacter(forByteOffset: keyOffset) else { return nil }
+        return Location(file: file.path, line: line, character: character)
```
marcelofabri: you can use `Location(file:byteOffset)`

Source/SwiftLintFramework/Rules/Style/CollectionAlignmentRule.swift
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    private func getArrayElementLocations(with file: File, dictionary: [String: SourceKitRepresentable]) -> [Location] {
+        let contents = file.contents.bridge()
+        return dictionary.elements.compactMap { element -> Location? in
+            guard let byteOffset = element.offset,
+                let (line, character) = contents.lineAndCharacter(forByteOffset: byteOffset) else {
+                return nil
+            }
+            return Location(file: file.path, line: line, character: character)
```
marcelofabri: you can use `Location(file:byteOffset)`

Source/SwiftLintFramework/Rules/Style/CollectionAlignmentRule.swift
```diff
@@ -0,0 +1,295 @@
...
+        } else {
+            keyLocations = getDictionaryKeyLocations(with: file, dictionary: dictionary)
+        }
+
+        guard keyLocations.count >= 2 else {
+            return []
+        }
+
+        let firstKeyLocation = keyLocations[0]
+        let violationLocations = zip(keyLocations[1...], 1...)
```
marcelofabri: what do you think about using `enumerated` here?

Source/SwiftLintFramework/Rules/Style/CollectionAlignmentRule.swift
```diff
@@ -0,0 +1,295 @@
...
+            .compactMap { location, index -> Location? in
+                let previousLocation = keyLocations[index - 1]
+                if previousLocation.line! < location.line! && firstKeyLocation.character! != location.character! {
```
marcelofabri: just to be safe, we should avoid those force unwraps in case there's a bug in converting offsets to line and characters

Source/SwiftLintFramework/Rules/Style/CollectionAlignmentRule.swift
```diff
@@ -0,0 +1,295 @@
...
+                                        keyOffset: keyOffset,
+                                        keyLength: keyLength,
+                                        valueOffset: valueOffset)
+            } else {
+                return getKeyLocation(with: file, keyOffset: keyOffset)
+            }
+        }
+    }
+
+    private func getColonLocation(with file: File, keyOffset: Int, keyLength: Int, valueOffset: Int) -> Location? {
```
marcelofabri: what do you think about removing the `get` prefix from this method and the other ones?

marcelofabri:
@daltonclaybrook I've just made a few small comments, but this is looking great!

Source/SwiftLintFramework/Rules/Style/CollectionAlignmentRule.swift
daltonclaybrook: Good idea.

Source/SwiftLintFramework/Rules/Style/CollectionAlignmentRule.swift
daltonclaybrook: `enumerated` would not work out of the box with the `ArraySlice` being created here because `enumerated` is always zero-based. However, I've updated this to use the `indices` property, which I think makes more sense than using `1...`. Thoughts?

Source/SwiftLintFramework/Rules/Style/CollectionAlignmentRule.swift
daltonclaybrook: I completely agree. Updated.

Source/SwiftLintFramework/Rules/Style/CollectionAlignmentRule.swift
daltonclaybrook: That works for me. 👍

Issue #2403, opened 2018-09-14
Setting warning_threshold to 0 appears to incorrectly show a violation


mrkd:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

Setting my warning_threshold to 0 appears to incorrectly show a violation

`warning_threshold: 0`

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
~/working/SwiftLintTestDeleteMe (~4h|master) • $ ./Pods/SwiftLint/swiftlint 
Loading configuration from '.swiftlint.yml'
Linting Swift files at paths 
Linting 'ViewController.swift' (1/2)
Linting 'AppDelegate.swift' (2/2)
:0:0: error: Warning Threshold Violation: Number of warnings exceeded threshold of 0. (warning_threshold)
Done linting! Found 1 violation, 1 serious in 2 files.
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)?
```
$ ./Pods/SwiftLint/swiftlint version
0.27.0
```

* Installation method used (Homebrew, CocoaPods, building from source, etc)?
CocoaPods
* Paste your configuration file:

```yml
included:
  - SwiftLintTestDeleteMe
disabled_rules:

warning_threshold: 0
line_length:
  warning: 130 # default 120
  error: 300 # default 200
trailing_whitespace: 
  ignores_empty_lines: true
cyclomatic_complexity:
  warning: 10 # default 10
  error: 20 # default 20
file_length:
  warning: 400 # default 400
  error: 3400 # default 1000
function_body_length:
  warning: 40 # default 40
  error: 100 # default 100
function_parameter_count:
  warning: 5 # default 5
  error: 8 # default 8
type_body_length:
  warning: 200 # default 200
  error: 1750 # default 350
number_separator:
  minimum_length: 5 # default 0
identifier_name:
  excluded:
    - id
    - or
    - at
    - to

opt_in_rules: # some rules are only opt-in

  - array_init # Prefer using Array(seq) than seq.map { $0 } to convert a sequence into an Array.
  - attributes # Attributes should be on their own lines in functions and types, but on the same line as variables and imports.
  - closure_end_indentation # Closure end should have the same indentation as the line that started it.
  - closure_spacing # Closure expressions should have a single space inside each brace.
  - conditional_returns_on_newline # Conditional statements should always return on the next line
[...]

PR #2405, opened 2018-09-14
 Fix for isWarningThresholdBroken failing when threshold set to 0

mrkd:
* Fix for isWarningThresholdBroken failing when threshold set to 0.  
  [mrkd](https://github.com/mrkd)
  [#issue_number](https://github.com/realm/SwiftLint/issues/2403)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,10 @@
 
 #### Bug Fixes
 
+* Fix for isWarningThresholdBroken failing when threshold set to 0.  
+  [mrkd](https://github.com/mrkd)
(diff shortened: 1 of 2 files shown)


jpsim:
The idea behind this setting was that if it was reached, the exit code would report an error. So if you set the threshold to 5, if you get 5 warnings or more SwiftLint would exit with an error code.

So this patch changes that behavior in a way that's not backwards compatible.

I think what you want to use is `--strict`, which reports an error code if any warnings or errors are found. Or perhaps you want to set the warning threshold to 1.

PR #2411, opened 2018-09-18
Only mark custom rules as enabled in your config when custom rules are configured

jhildensperger:
Currently, custom rules always appear as "enabled in your config" when you run the `rules` command. This change will check if there are custom rules in your configuration file.


jpsim:
Pushed to master in c3d72c24 and b66446c0. Thanks @jhildensperger!

PR #2418, opened 2018-09-26
Fix false positive on file_name rule with specific patterns

### LINKED ISSUE #2417: False positive specific patterns in rule file_name
When I configure the rule `file_name` with a `suffix_pattern` which looks like this:

```
file_name:
  suffix_pattern: "Extensions?|\\+.*"
```

then my file `BoolExtension` is correctly handled. But when I add a test for that extension file called `BoolExtensionTests` then I get a violation warning there, although there shouldn't be one.

Dschee:
Fixes #2417.


jpsim:
Thanks for the PR!

PR #2419, opened 2018-09-26
New Rule - Legacy Random

yhkaplan:
This PR adds a new rule for preferring the use of `Int/Double/Float.random(in:)` in Swift 4.2.


yhkaplan:
`random` is a very short name, but it might also cover `Bool.random`, so I'm a little unsure of what else to call it....

Source/SwiftLintFramework/Rules/Idiomatic/RandomRule.swift
```diff
@@ -0,0 +1,38 @@
...
+import SourceKittenFramework
+
+public struct RandomRule: OptInRule, ConfigurationProviderRule, AutomaticTestableRule {
+
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static var description = RuleDescription(
+        identifier: "random",
```
marcelofabri: maybe `legacy_random` would be a better name? We already have some `legacy_*` rules

Source/SwiftLintFramework/Rules/Idiomatic/RandomRule.swift
```diff
@@ -0,0 +1,38 @@
...
+        name: "Random",
+        description: "Prefer using `type.random(in:)` over C-based functions.",
+        kind: .lint,
```
marcelofabri: this should be `idiomatic`

Source/SwiftLintFramework/Rules/Idiomatic/RandomRule.swift
```diff
@@ -0,0 +1,38 @@
...
+        ],
+        triggeringExamples: [
+            "↓arc4random(10)\n",
+            "↓arc4random_uniform(83)\n",
+            "↓drand48(52)\n"
+        ]
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        let pattern = "(arc4random|drand48)"
```
marcelofabri: this would be more reliably implemented by using an `ASTRule` checking for function calls. `LegacyConstructorRule` would be an example (although it's probably more complicated than what you need)

Source/SwiftLintFramework/Rules/Idiomatic/RandomRule.swift
yhkaplan: Sounds good, thanks! I’ll correct everything this weekend

yhkaplan:
@marcelofabri I made the corrections as above.

yhkaplan:
@marcelofabri I apologize if you're on holiday, but when do you think you'll have time to review this?

CHANGELOG.md
```diff
@@ -32,6 +32,10 @@
 
 #### Enhancements
 
+* Add `random` opt-in rule to encourage the use of `.random(in:)`
```
marcelofabri: `legacy_random` instead of `random`

Source/SwiftLintFramework/Rules/Idiomatic/LegacyRandomRule.swift
```diff
@@ -0,0 +1,63 @@
...
+            "Float.random(in: 0 ..< 1)\n"
+        ],
+        triggeringExamples: [
+            "↓arc4random(10)\n",
+            "↓arc4random_uniform(83)\n",
+            "↓drand48(52)\n"
+        ]
+    )
+
+    private let legacyRandomFunctions: [String] = [
```
marcelofabri: This could be a `Set`:

```swift
private let legacyRandomFunctions: Set = [
 /// ...
]
```

Source/SwiftLintFramework/Rules/Idiomatic/LegacyRandomRule.swift
```diff
@@ -0,0 +1,63 @@
...
+        return [
+            StyleViolation(ruleDescription: type(of: self).description,
+                           severity: configuration.severity,
+                           location: location)
+        ]
+    }
+
+    private func containsViolation(kind: SwiftExpressionKind, dictionary: [String: SourceKitRepresentable]) -> Bool {
+        guard kind == .call,
+            dictionary.offset != nil,
```
marcelofabri: you don't need to check this here because it's part of the `guard` on `validate`

PR #2426, opened 2018-09-28
Fix false positives on explicit_type_interface rule

### LINKED ISSUE #2425: False positives when using explicit_type_interface with allowRedundancy
Using the lastest SwiftLint version, when I use the rule `explicit_type_interface` with this configuration

```yaml
explicit_type_interface:
  allow_redundancy: true
```

most of the cases are handled correctly. But for some initializers Swift has some sugar syntax integrated and some initializers might throw. In both cases this option doesn't work correctly.

Specifically, the following three lines are shown as warnings but shouldn't:

```swift
let regex = try! NSRegularExpression(pattern: ".*")
let array = [String]()
let dict = [String: String]()
```

Dschee:
Fixes #2425.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -238,6 +238,10 @@ This is the last release to support building with Swift 4.0 and Swift 4.1.
   hexadecimal integer literals.  
   [Timofey Solonin](https://github.com/biboran)
   [#2423](https://github.com/realm/SwiftLint/issues/2423)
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks for the PR. Could you please rebase to address the merge conflict?

jpsim:
> Thanks for the PR. Could you please rebase to address the merge conflict?

I took care of this, rebased and force-pushed to your branch. Will merge when CI approves.

PR #2427, opened 2018-09-28
Added possibility to make lint fail if using wrong version

### LINKED ISSUE #2074: [Feature request] Validate version when linting
### New Issue Checklist

### Rule Request
I have been thinking about some minimum version validating when linting, something like fastlane have in their fastfiles. When the version is less than the required version there will be an warning or error. 

The reason behind this, can be new rules in newer versions and now all developers are aware of new versions.

kimdv:
Fixes #2074 

I don't know if this is the correct way solving it or the correct name of key in configuration file.


jpsim:
I wonder if this should simply be the behavior when using `swiftlint_version`. I suspect that people who use this don't want this error lost in SwiftLint's logs.

kimdv:
@jpsim i agree ☝️ 
Will change it when I’m at my Mac again 😎

kimdv:
Opened new PR as i screwed up rebase

#2491

Issue #2423, opened 2018-09-28
empty_count false positive on hex values


alejandro-isaza:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

The following code causes a `empty_count` false positive:

```
if result.count <= 0xff {
    ...
}
```

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint lint
abc.swift:115:30: error: Empty Count Violation: Prefer checking `isEmpty` over comparing `count` to zero. (empty_count)
```

### Environment

* SwiftLint version: 0.27.0
* Installation method used: Homebrew

* Which Xcode version: Xcode 10
* Sample: `echo "if result.count <= 0xff {}" | swiftlint lint --no-cache --use-stdin --enable-all-rules`

```swift
// This triggers a violation:
if result.count <= 0xff {
    ...
}
```

PR #2430, opened 2018-10-02
#2423 - Fix EmptyCountRule for binary, octal and hexadecimal integer literals

biboran:
As per #2423 
* Added condition to `empty_count` regex to check whether 0 ends the word or represents a 0 as a binary, decimal or hex

### DIFF
--- a/Source/SwiftLintFramework/Rules/Performance/EmptyCountRule.swift
+++ b/Source/SwiftLintFramework/Rules/Performance/EmptyCountRule.swift
@@ -15,6 +15,9 @@ public struct EmptyCountRule: ConfigurationProviderRule, OptInRule, AutomaticTes
             "[Int]().isEmpty\n",
             "[Int]().count > 1\n",
             "[Int]().count == 1\n",
+            "[Int]().count == 0xff\n",
+            "[Int]().count == 0b01\n",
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Rules/Performance/EmptyCountRule.swift
```diff
@@ -15,19 +15,26 @@ public struct EmptyCountRule: ConfigurationProviderRule, OptInRule, AutomaticTes
...
+            "[Int]().↓count == 0x00_00\n",
+            "[Int]().↓count == 0b00\n",
+            "[Int]().↓count == 0o00\n",
             "↓count == 0\n"
         ]
     )
 
     public func validate(file: File) -> [StyleViolation] {
-        let pattern = "\\bcount\\s*(==|!=|<|<=|>|>=)\\s*0"
+        let pattern = "\\bcount\\s*(==|!=|<|<=|>|>=)\\s*0(\\b|([box][0_]+\\b){1})"
```
marcelofabri: I thought `box` was a special keyword or something, took me a while to realize it's just the three letters that are allowed 😅

jpsim: Hahaha, same

PR #2434, opened 2018-10-06
New Rule - Redundant XCTAssert Parameter

yhkaplan:
This PR adds a new rule for preferring the use of `XCTAssertTrue/False/Nil()` when usable. I'm also considering adding auto-correct functionality.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Idiomatic/RedundantXCTAssertParameterRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/RedundantXCTAssertParameterRule.swift
@@ -36,6 +36,12 @@ public struct RedundantXCTAssertParameterRule: ASTRule, OptInRule, Configuration
         ]
     )
 
+    private let violatingArguments: Set<String> = [
+        "true",
+        "false",
+        "nil"
+    ]
+
     public func validate(
         file: File,
         kind: SwiftExpressionKind,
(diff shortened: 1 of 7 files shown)


Source/SwiftLintFramework/Rules/Idiomatic/RedundantXCTAssertParameterRule.swift
```diff
@@ -0,0 +1,92 @@
...
+    private func containsViolatingArgument(_ argument: [String: SourceKitRepresentable], for string: String) -> Bool {
+
+        guard let offset = argument.offset,
+            let length = argument.length,
+            let range = string.bridge().byteRangeToNSRange(start: offset, length: length) else {
+
+            return false
+        }
+
+        return regex("^(true|false|nil)$").firstMatch(in: string, options: [], range: range) != nil
```
yhkaplan: TODO: `substring(for: range)` is probably much faster...

yhkaplan: Hmm, didn't change much...

ornithocoder:
This rule is a little bit more complex than it looks like (I have a [WIP here](https://github.com/realm/SwiftLint/pull/1995/files)).

You also have to handle all the false positives when the parameter is optional (XCTAssertTrue/False doesn't work with optionals. e.g. `XCTAssertEqual(response?.result.isSuccess, true)` should not trigger it).

yhkaplan:
@ornithocoder It looks like you've put a lot more thought about how this rule should work than I have! Also, as your PR is already in review again, it looks like I should go ahead and close this one. Thanks for pointing this out!

Issue #131, opened 2015-09-14
No defaults in enum switch


matrinox:
I find the exhaustive `enum` warnings from the compiler super helpful. It ensures that whenever a new `enum` case is added, every `switch` statement is updated. But this won't work well with `default`. Often times the `default` is there as an easy way to group behaviour for the other cases at the time. I'd like to see this added as a warning

kevindelord:
+1 we also make sure to never use the `default` case in our enums :)

Noobish1:
I'd love this feature also.

jpsim:
I think this is a good rule to have, but it would be challenging to detect when the variable being switched on is an enum.

acecilia:
I would love to see this too :)

valentinradu:
This would be great!

sammy-SC:
@jpsim 

you mentioned that it would be challenging to detect when the variable being switched on is an enum.
In simple case like the one below, shouldn't it be sufficient to see whether the condition next to case is prepended with a dot? In this case `.one`? I think it should be safe to assume that variable being switched on is an enum if all conditions are prepended with a dot.

```swift
enum MyEnum {
  case one, two
}

let myEnum = MyEnum.two

switch myEnum {
  case .one: break
  ↓default: break
}
```

Please let me know whether you think this could work, I would be happy to implement this rule.

PR #2445, opened 2018-10-26
Add default case rule

rakaramos:
Implements: https://github.com/realm/SwiftLint/issues/131

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 #### Enhancements
 
-* Add `reduce_boolean` rule to prefer simpler constructs over `reduce(Boolean)`.
+* Add `reduce_boolean` rule to prefer simpler constructs over `reduce(Boolean)`.  
   [Xavier Lowmiller](https://github.com/xavierLowmiller)
   [#2675](https://github.com/realm/SwiftLint/issues/2675)
 
@@ -42,6 +42,7 @@
 * None.
 
 #### Enhancements
+
 * Add `deployment_target` rule to validate that `@availability` attributes and
   `#available` conditions are not using a version that is satisfied by the
   deployment target. Since SwiftLint can't read an Xcode project, you need to


Source/SwiftLintFramework/Rules/Idiomatic/DefaultCaseRule.swift
```diff
@@ -0,0 +1,59 @@
...
+public struct DefaultCaseRule: ASTRule, ConfigurationProviderRule, OptInRule, AutomaticTestableRule {
+
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "default_case",
+        name: "Default case",
+        description: "`default` statement on `switch` statement should be avoided.",
```
gobetti: perhaps you meant "`default` *case* on `switch` statement should be avoided"?

marcelofabri:
This is not so useful right now because it'll trigger even when you need a `default` (switching on a non-enum value).

Maybe we could use the new analyze mode to get more information about the expression being switched on and check if it's an enum?

rakaramos:
Since [0192-non-exhaustive-enums](https://github.com/apple/swift-evolution/blob/master/proposals/0192-non-exhaustive-enums.md) was implemented on Swift 5, I think  this rule will be useless (unless for people that don't adopt Swift 5 right away).

@marcelofabri do you think it still worth it?

gobetti:
@rakaramos Swift 5 is going to introduce a distinction between frozen and non-frozen enums, which is not going to change anything for the currently already exhaustive enums, which are what makes your rule useful.

In other words, Swift 5 will not require you to _avoid the default case_, which is what your rule does. They will oblige you to *use* it, _when_ the enum is non-frozen. And for that use specifically, it will be `@unknown default`, which again is different from your rule (as long as your regex matches `default` but doesn't match the @unknown version).

I agree with the other comments: for this rule to be more useful than a pain, the type being switched on must be checked, it only makes sense where exhaustiveness exists, i.e. enums, booleans, perhaps something else. I don't think it will be an easy job, though, when tuples come to mind...

Dschee:
Can we perhaps use [SwiftSyntax](https://github.com/apple/swift-syntax) to detect if the variable is of `enum` type? Does SwiftSyntax provide this info? I'm not really sure but kind of doubt it.

I also think this rule only should be merged if it can detect many common cases where actually an enum is used without producing any false positives. Do you still think it's useless after Swift 5 is released @rakaramos?

rakaramos:
@dschee that's a good question! I think it will depend on how much checks the compiler is doing...I haven't played around with Swift 5, but would be nice to dig into the new `@unknown` keyword.

marcelofabri:
> Does SwiftSyntax provide this info?

I don't think it does. We maybe can do it with our analyzer rules, but I don't know if it can handle variables with types that are declared in other files.

JulianKahnert:
Hi @rakaramos @marcelofabri,
I would love to work on this rule.
What else needs to be done here?

PR #2465, opened 2018-11-18
Improve performance of collecting files to lint and lint cache lookups

jpsim:
Performance has gotten pretty bad for complex SwiftLint configurations like the one used for Lyft's iOS code base involving lots of files in the directories being linted, large configuration files and many nested configuration files.

Two main areas were particularly ripe for improvement were:

1. Collecting which files to lint
2. Lint cache lookups

### Collecting which files to lint

Improve this by:

* using an NSOrderedSet to remove excluded paths instead of `Array.filter`
* parallelizing calls to `filesToLint` for all paths to lint and exclude
* using `FileManager.subpaths(atPath:)` instead of `enumerator(atPath:)`

|Change|Before|After|Speed up|
|-|-|-|-|
|NSOrderedSet|2.438s|0.917s|2.659x|
|Parallel Flat Map|2.438s|2.248s|1.085x|
|Subpaths|0.939s|0.867s|1.083x|
|**Total**|**2.438s**|**0.720s**|**3.386x**|

### Lint cache lookups

By using an MD5 hash of the Configuration description from CryptoSwift as the cache key instead of instead the full description, we can drastically speed up cache lookups for projects with complex SwiftLint configurations. I think the dictionary lookup for very large string keys doesn't perform very well.


jpsim:
This release will have warnings when building with Swift 4.2.

SwiftLint supports building with Swift 4.0 to 4.2.

There is no version of CryptoSwift to support both Swift 4.0 and Swift 4.2.

So allow warnings for now. We'll make one more Swift 4.0 compatible release, then we'll bump the build requirements to Swift 4.2 and remove the `--allow-warnings` flag from cocoapods commands.

This should be short-lived.

PR #2469, opened 2018-11-19
Use default Equatable and Hashable implementations

marcelofabri:
Since [we now require Swift 4.2 to build](https://github.com/realm/SwiftLint/pull/2466), we can use the compiler-generated implementations for `Equatable` and `Hashable` 🎉

### DIFF
--- a/Source/SwiftLintFramework/Extensions/NSRegularExpression+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/NSRegularExpression+SwiftLint.swift
@@ -6,13 +6,11 @@ private let regexCacheLock = NSLock()
 private struct RegexCacheKey: Hashable {
     let pattern: String
     let options: NSRegularExpression.Options
+}
 
-    var hashValue: Int {
-        return pattern.hashValue ^ options.rawValue.hashValue
-    }
-
-    static func == (lhs: RegexCacheKey, rhs: RegexCacheKey) -> Bool {
-        return lhs.options == rhs.options && lhs.pattern == rhs.pattern
+extension NSRegularExpression.Options: Hashable {
+    public var hashValue: Int {
+        return rawValue.hashValue
     }
 }
 
--- a/Source/SwiftLintFramework/Models/Location.swift
+++ b/Source/SwiftLintFramework/Models/Location.swift
@@ -45,14 +45,6 @@ public struct Location: CustomStringConvertible, Comparable {
         }
     }
 
-    // MARK: Comparable
-
-    public static func == (lhs: Location, rhs: Location) -> Bool {
-        return lhs.file == rhs.file &&
-            lhs.line == rhs.line &&
-            lhs.character == rhs.character
-    }
-
     public static func < (lhs: Location, rhs: Location) -> Bool {
         if lhs.file != rhs.file {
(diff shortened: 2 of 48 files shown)


Tests/SwiftLintFrameworkTests/RequiredEnumCaseRuleTestCase.swift
```diff
@@ -24,7 +24,7 @@ class RequiredEnumCaseRuleTestCase: XCTestCase {
 
     func testRequiredCaseHashValue() {
         let requiredCase = RequiredCase(name: "success")
-        XCTAssertEqual(requiredCase.hashValue, "success".hashValue)
+        XCTAssertEqual(requiredCase.hashValue, RequiredCase(name: "success").hashValue)
```
biboran: What's the point in testing `XCTAssertEqual(RequiredCase(name: "success").hashValue, RequiredCase(name: "success").hashValue)`?

marcelofabri: The ideia is to make sure that `hashCode` is the same for enums containing associated values. 

This is now basically testing the compiler, but the fact that we're generating `hashValue` automatically is an implementation detail and I don't think it hurts to test it.

Note that we can't compare to a value directly because now the `hashValue` can (and will) return different values in different program runs.

Source/SwiftLintFramework/Extensions/NSRegularExpression+SwiftLint.swift
```diff
@@ -6,13 +6,11 @@ private let regexCacheLock = NSLock()
...
     let options: NSRegularExpression.Options
+}
 
-    var hashValue: Int {
-        return pattern.hashValue ^ options.rawValue.hashValue
-    }
-
-    static func == (lhs: RegexCacheKey, rhs: RegexCacheKey) -> Bool {
-        return lhs.options == rhs.options && lhs.pattern == rhs.pattern
+extension NSRegularExpression.Options: Hashable {
```
jpsim: Is this not hashable already?

Source/SwiftLintFramework/Models/Location.swift
```diff
@@ -45,14 +45,6 @@ public struct Location: CustomStringConvertible, Comparable {
         }
     }
 
-    // MARK: Comparable
```
jpsim: Should keep this MARK

Tests/SwiftLintFrameworkTests/RequiredEnumCaseRuleTestCase.swift
jpsim: I'm not convinced these tests give much value, but fine to keep if you think they do.

Tests being `testRequiredCaseHashValue`, `testRequiredCaseEquatableReturnsTrue` and `testRequiredCaseEquatableReturnsFalseBecauseOfDifferentName`.

Source/SwiftLintFramework/Extensions/NSRegularExpression+SwiftLint.swift
marcelofabri: Surprisingly, it's not 😱

PR #2477, opened 2018-11-23
Add `testSimulateHomebrewTest()` to `IntegrationTests`

norio-nomura:
Motivation:

Sometimes, Homebrew bottling CI fails because it runs in the sandbox.

Modifications:

Add `testSimulateHomebrewTest()` to `IntegrationTests` that simulates test in `homebrew-core/Formula/swiftlint.rb` within sandbox.

Result:

Confirm that a test equivalent to `brew test swiftlint` will pass in the sandbox before publishing to homebrew.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@
   with CommonCrypto.  
   [JP Simard](https://github.com/jpsim)
 
+* Add `testSimulateHomebrewTest()` to `IntegrationTests` that simulates test in
+  `homebrew-core/Formula/swiftlint.rb` within sandbox.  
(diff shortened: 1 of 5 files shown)


norio-nomura:
> Sometimes, Homebrew bottling CI fails because it runs in the sandbox.

e.g. https://github.com/Homebrew/homebrew-core/pull/34268

> Confirm that a test equivalent to brew test swiftlint will pass in the sandbox before publishing to homebrew.

Since CircleCI does not use macOS 10.14.1+, it is necessary to check locally on macOS 10.14.1+ when `make publish`.

jpsim:
Thanks! I just made a few small stylistic edits but this should be good to go. This will be very useful 😅

PR #2478, opened 2018-11-23
Fix false positives for `explicit_init` when passing init as closure in multiline call

dominicfreeston:
First time (attempting) contributor here, hope this is helpful 😄 

As it stands, the following two syntactically equivalent lines give two different results - the former successfully passes validation whilst the latter does not.

``` swift
Observable.zip(obs1, obs2, resultSelector: MyType.init).asMaybe()

Observable.zip(
    obs1,
    obs2,
    resultSelector: MyType.init
).asMaybe()
```

The proposed changes relies on the fact that a class name can't have `(` in it and so any expression passed to the validator that has `(` between the initial capital letter and `.init` can't be an explicit init. An alternative could be to strip out the newlines. Any thoughts?

I'm also aware I've expanded on the documentation (and made an implicit reference to RxSwift) just to expand the number of tests, is there a cleaner way to do this, or is this valuable?

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@
 * Fix compiler warnings when building with Swift 4.2 introduced in the last
   release.  
   [JP Simard](https://github.com/jpsim)
+* Fix false positive in explicit_init rule  
+  [Dominic Freeston](https://github.com/dominicfreeston)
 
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks for the excellent PR!

Issue #2471, opened 2018-11-19
False positive with toggle_bool


rounak:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug
`object.property = !property` This triggers a toggle_bool warning even though `property` is a local variable, and calling `.toggle()` on `object.property` might produce incorrect results.

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.28.1
* Installation method used (Homebrew, CocoaPods, building from source, etc)? cocoapods 
* Which Xcode version are you using (check `xcode-select -p`)? 10.1
* Do you have a sample that shows the issue? Run `echo "object.property = !property" | swiftlint lint --no-cache --use-stdin --enable-all-rules`
  to quickly test if your example is really demonstrating the issue.

```swift
// This triggers a violation:
object.property = !property
```

jpsim:
Fixed in #2479

PR #2479, opened 2018-11-23
#2471 - Fix false positive toggle_bool

biboran:
As per #2471 
* Replaced word boundary with a negative lookbehind on a word character or a dot
* Added 2 relevant test cases


jpsim:
Amazing, thank you!

PR #2480, opened 2018-11-24
[Don't Merge] Migrate FallthroughRule to SwiftSyntax

jpsim:
This is a follow-up to #2476 but using [SourceKitten's syntaxTree request](https://github.com/jpsim/SourceKitten/pull/566) instead of invoking the swiftc process for each file.

Early results are promising but it's still significantly slower than the current non-SwiftSyntax implementation:

image

Lots of time is being spent in JSON decoding, so SwiftSyntax's upcoming [byte tree deserializer](https://github.com/apple/swift-syntax/blob/aa2489d6359d4ac971e815bd91b7bff875183436/Sources/SwiftSyntax/ByteTreeDeserialization.swift) should help bridge the remaining performance gap.

image


Dschee:
@jpsim What do you think, should we consider adding a SwiftSyntax based base rule type to the project now that Swift 5 will be released soon (I'm assuming that the byte tree deserialization will be part of the new release of SwiftSyntax) and start refactoring the first rules to it? Or do you have other ideas on how we can improve performance and correctness long term? (I know it's not the fastest yet, but I'm assuming performance will improve over time.)

marcelofabri:
> I'm assuming that the byte tree deserialization will be part of the new release of SwiftSyntax

IIRC this won't be a part of 5.0

PR #2491, opened 2018-11-29
Swiftlint now failed when using wrong version

kimdv:
Opened new PR as i screwed up the other one 🤦‍♂️🤦‍♂️

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,8 @@
   [Timofey Solonin](https://github.com/biboran)
   [#2435](https://github.com/realm/SwiftLint/issues/2435)
 
-* Swiftlint now fails when not using pinned version.
+* The `lint` command now exits with a code of 2 when not using pinned
+  version defined as `swiftlint_version` in the configuration file.  
   [Timofey Solonin](https://github.com/kimdv)
   [#2074](https://github.com/realm/SwiftLint/issues/2074)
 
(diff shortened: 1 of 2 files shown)


kimdv:
@jpsim open new PR here. Screwed the other one. 

Now using `queuedFatalError` when printing message

marcelofabri:
Can you please add a changelog entry?

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -56,7 +56,7 @@ public struct Configuration: Hashable {
                  indentation: IndentationStyle = .default) {
 
         if let pinnedVersion = swiftlintVersion, pinnedVersion != Version.current.value {
-            queuedPrintError("Currently running SwiftLint \(Version.current.value) but " +
+            queuedFatalError("Currently running SwiftLint \(Version.current.value) but " +
```
jpsim: This won't do what you want. This will crash SwiftLint and not print the message to the console in release builds.

You want to keep `queuedPrintError()` but `exit(2)` afterwards.

CHANGELOG.md
```diff
@@ -39,6 +39,10 @@
   [Timofey Solonin](https://github.com/biboran)
   [#2435](https://github.com/realm/SwiftLint/issues/2435)
 
+  * Swiftlint now fails when not using pinned version.
```
jpsim: Please match the indentation and formatting of other entries. Also it's "SwiftLint" not "Swiftlint". Also needs two trailing spaces as defined in CONTRIBUTING.md

kimdv:
@jpsim thanks for feedback! 

I have applied it. 

Is there a way we could test this?

PR #2492, opened 2018-11-29
Fix false positives and invalid corrections with unused import rule

jpsim:
Previously, removing an unused testable import would leave the `@testable` attribute, leading to changes like these:

```diff
-@testable import ModuleA
-import ModuleB
+@testable import ModuleB
```

This PR fixes this by also removing the testable attribute.

---

Also, if no Foundation declarations were used in a file that used the `@objc` family of attributes, those still require Foundation to be imported. This PR special cases Foundation to check for these keywords.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
@@ -19,6 +19,11 @@ public struct UnusedImportRule: CorrectableRule, ConfigurationProviderRule, Anal
             """
             @testable import Dispatch
             dispatchMain()
+            """,
+            """
+            import Foundation
+            @objc
+            class A {}
             """
         ],
         triggeringExamples: [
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
```diff
@@ -129,11 +161,36 @@ private extension File {
...
+            if testableImportRange.location != NSNotFound {
+                return (module, testableImportRange)
+            }
+
+            return (module, contentsNSString.range(of: "import \(module)\n"))
         }
     }
+
+    private func containsAttributesRequiringFoundation() -> Bool {
+        guard contents.contains("@objc") else {
```
keith: Do `IBAction` and `IBOutlet` also require this logic? I assume so but maybe it's less likely to have a file with one of those where it doesn't also require the import for other reasons.

jpsim: My thinking exactly. Let's cross that bridge when we get there.

Issue #2413, opened 2018-09-20
Add sorting to HTML reporter


kpiascik:
### Describe the bug

When viewing the HTML reporter output having the ability to sort by table header would be super useful.  It would help diagnose issues in groups by type and severity as well as file.

### Environment
- OSX installed via homebrew running 0.27.0
- The .yml I used is not relevant

#### Command line:
```
swiftlint lint --quiet --reporter html > swiftlint.html
```

PR #2494, opened 2018-12-01
Add sorting to HTML reporter

prathmeshranaut:
Add sorting to HTML reporter (https://github.com/realm/SwiftLint/issues/2413)


jpsim:
Thanks for the PR! However, I don't think we should be modifying the html reporter with this functionality. Loading online javascript from 3rd party sources shouldn't be done lightly, plus if anyone's built anything that relies on the table format behaving as it does today, this could break it.

Instead of changing SwiftLint's `html` reporter, I'd recommend you use the `json` reporter and pipe the output to a script that can generate the custom table format you'd like from stdin.

prathmeshranaut:
So should we add a new reporter in that case? Maybe a sortable-html reporter?

jpsim:
My recommendation is not a new reporter but rather a script approach I recommended in my last comment.

PR #2496, opened 2018-12-03
Added hash_value_overflow rule

### LINKED ISSUE #2108: Rule Request: [HashValue overflow]
### New Issue Checklist

### Rule Request

I don't know 100% if this should be a rule, but it is possible in swift to overflow hash value by doing `"test-1".hashValue + "test-2".hashValue`

And for avoiding it it should be done like this `"test-1".hashValue &+ "test-2".hashValue`

It should be a optional rule. 

It could prevent some crashes that people don't think about :)

kimdv:
This fixes #2108 

I'm not good at all to reggex. So I got a a little help from a friend 😅


jpsim:
Thanks for the PR @kimdv! Do you think we should be encouraging the use of Swift 4.2's modernized hashing instead?

This rule seems especially well suited to avoid using a regular expression to detect the use of `+`/`*` and instead using string matching.

For context, the heavy use of regular expressions in SwiftLint have been contributing to performance cliffs and false positives and I would like us to start weaning off of them.

kimdv:
Thanks, @jpsim! 

We could do that, but I have just finished a project where we needed custom hashes. 

I have seen your comments lately on avoiding regular expression, so I really would like to remove the hash regular expression, so can you point where I could look for an example?

marcelofabri:
@kimdv I think @jpsim's suggestion was to use the new `func hash(into hasher: inout Hasher)` instead of manually combining hashes:

```swift
extension GridPoint: Hashable {
  func hash(into hasher: inout Hasher) {
    hasher.combine(x)
    hasher.combine(y)
  }
}
```

(from https://github.com/apple/swift-evolution/blob/master/proposals/0206-hashable-enhancements.md#the-hashinto-requirement)

kimdv:
Sure we can do that! 
Will make it min requirement Swift 4.2, what do you think? 

@marcelofabri can you maybe point out where to look for inspiration to avoid regular expression?

marcelofabri:
@kimdv sounds good for min requirement. 

If the rule is updated to validate that `hash(into:)` should be used (instead of checking for overflow values), it becomes easier. You can implement `ASTRule` and check for `hashValue` computed properties declarations instead of using regexes. `ProhibitedInterfaceBuilderRule` seems a good place to start. `WeakDelegateRule` has a check for computed variables, which will be needed as well.

Source/SwiftLintFramework/Rules/Lint/HashValueOverflowRule.swift
```diff
@@ -0,0 +1,65 @@
...
+        ]
+    )
+
+    // MARK: - ASTRule
+
+    public func validate(file: File,
+                         kind: SwiftDeclarationKind,
+                         dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard kind == .varInstance,
+            dictionary.name == "hashValue",
```
marcelofabri: you probably should check also the property type with `dictionary.typeName == "Int"`

Source/SwiftLintFramework/Rules/Lint/HashValueOverflowRule.swift
```diff
@@ -0,0 +1,65 @@
...
+            let offset = dictionary.offset else {
+            return []
+        }
+
+        func makeViolation() -> StyleViolation {
```
marcelofabri: nitpick: no need for the function

Source/SwiftLintFramework/Rules/Lint/HashValueOverflowRule.swift
```diff
@@ -0,0 +1,65 @@
+import Foundation
+import SourceKittenFramework
+
+public struct HashValueOverflowRule: ASTRule, OptInRule, ConfigurationProviderRule, AutomaticTestableRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "hash_value_overflow",
```
marcelofabri: I'd rename this function and update the description, because IMO the rule is now more general than just guarding against overflows: the new hash function should be preferred in general.

Source/SwiftLintFramework/Rules/Lint/HashValueOverflowRule.swift
```diff
@@ -0,0 +1,65 @@
+import Foundation
+import SourceKittenFramework
+
+public struct HashValueOverflowRule: ASTRule, OptInRule, ConfigurationProviderRule, AutomaticTestableRule {
```
marcelofabri: IMO this doesn't need to be an opt-in rule

kimdv: Sounds good to me! 
Will remove it

Source/SwiftLintFramework/Rules/Lint/HashValueOverflowRule.swift
kimdv: Would `HashFunctionRule` be something that is OK?

marcelofabri: sounds good to me 👍

marcelofabri:
@kimdv Please do it!

kimdv:
Pushed changes and stuff starts to pass now! 🚀

Source/SwiftLintFramework/Rules/Lint/HashFunctionRule.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    // MARK: - ASTRule
+
+    public func validate(file: File,
+                         kind: SwiftDeclarationKind,
+                         dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard kind == .varInstance,
+            dictionary.typeName == "Int",
+            dictionary.name == "hashValue",
+            let offset = dictionary.offset else {
```
marcelofabri: you probably should check if it's a computed property too

Source/SwiftLintFramework/Rules/Lint/HashFunctionRule.swift
```diff
@@ -0,0 +1,62 @@
...
+
+public struct HashFunctionRule: ASTRule, ConfigurationProviderRule, AutomaticTestableRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "hash_function",
+        name: "Hash Function",
+        description: "The new hash function should be preferred in general. " +
```
marcelofabri: Maybe "Prefer using implementing `hash(into:)` instead of `hashValue.`?

Source/SwiftLintFramework/Rules/Lint/HashFunctionRule.swift
```diff
@@ -0,0 +1,62 @@
...
+        "Consider using `func hash(into hasher: inout Hasher)` instead.",
+        kind: .lint,
+        minSwiftVersion: .fourDotTwo,
+        nonTriggeringExamples: [
+            """
+            struct Foo: Hashable {
```
marcelofabri: It'd be great to have some more examples just to be sure (e.g. a global var called `hashValue`, `hashValue` with another type, etc)

Also, a small nitpick: maybe use a `class` as `struct`s now automagically get `Hashable` implementation for free?

marcelofabri:
@kimdv Thanks for this - just had some minor comments but overall it looks great!

We also need a changelog entry for this change ✨

Source/SwiftLintFramework/Rules/Lint/HashFunctionRule.swift
kimdv: My English is not that good, but should it not be "Prefer using implementation `hash(into:)` instead of `hashValue`"?

"using implementing" sounds a little strange to me. But I could also be wrong 😄

Source/SwiftLintFramework/Rules/Lint/HashFunctionRule.swift
kimdv: I think we should keep the struct as some people override it and will implement a custom hash function.

Source/SwiftLintFramework/Rules/Lint/HashFunctionRule.swift
marcelofabri: Yeah, you’re right. Probably changed my mind when editing and forgot it there

kimdv:
@marcelofabri I rebased with master (Don't know if I did it right) and added your feedback and added changelog entry

CHANGELOG.md
```diff
@@ -24,6 +24,11 @@
   for large files.  
   [Niil Öhlin](https://github.com/niilohlin)
 
+* Add new opt-in rule `hash_function` to warn against empty wrong hashValue
+  implementation.
```
jpsim: Nit: we ask for two trailing spaces here to cause markdown to render this as a line break. See https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes

Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
```diff
@@ -61,8 +61,8 @@ extension Configuration {
             return type(of: lhs.rule).description.identifier == type(of: rhs.rule).description.identifier
         }
 
-        fileprivate var hashValue: Int {
-            return type(of: rule).description.identifier.hashValue
+        public func hash(into hasher: inout Hasher) {
```
jpsim: This needs to be `private`.

Source/SwiftLintFramework/Rules/Lint/HashFunctionRule.swift
```diff
@@ -0,0 +1,97 @@
...
+
+public struct HashFunctionRule: ASTRule, ConfigurationProviderRule, AutomaticTestableRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "hash_function",
+        name: "Hash Function",
+        description: "Prefer using implementation `hash(into:)` instead of `hashValue`",
```
jpsim:
```suggestion
        description: "Prefer using the `hash(into:)` function instead of overriding `hashValue`",
```

Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
kimdv: Will set to `fileprivate` as before, ok?

CHANGELOG.md
kimdv: Thanks!

Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
jpsim: 👍

kimdv:
@jpsim should be resolved now 🚀

jpsim:
Should we name this rule "Legacy Hashing" to align with the other legacy family of rules?

* Legacy CGGeometry Functions
* Legacy Constant
* Legacy Constructor
* Legacy NSGeometry Functions
* Legacy Random

CHANGELOG.md
```diff
@@ -24,6 +24,11 @@
   for large files.  
   [Niil Öhlin](https://github.com/niilohlin)
 
+* Add new opt-in rule `hash_function` to warn against empty wrong hashValue
```
jpsim: This is incorrect and poorly worded. How about

> Add new `legacy_hashing` rule to encourage the use of Swift 4.2's new hashing interface.

kimdv: Changed the name. 

Also changed kind to `.idiomatic` to math the other legacy rules

jpsim:
Merged! 🎉 

Thanks @kimdv !

PR #2501, opened 2018-12-04
Fix non accessible declarations under @objcMembers

dirtydanee:
Issue reported as comment under https://github.com/realm/SwiftLint/pull/2270 PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,16 +10,13 @@
 
 #### Enhancements
 
-* Fix warnings for non accessible declarations under `@objcMembers`.  
-  [Daniel Metzing](https://github.com/dirtydanee)
-
-* Add new opt-in rule `vertical_whitespace_opening_braces` to warn against empty lines
-  after opening braces.  
+* Add new opt-in rule `vertical_whitespace_opening_braces` to warn against empty
+  lines after opening braces.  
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks for the fix! I made some small changes and pushed to your branch:

* Refactored to avoid false optionality. Because a dictionary was used with two guaranteed keys, this required treating their values as optional even though we knew they weren't. Fixed this by using a struct with two named members instead.
* Moved the changelog entry to the "Bug Fixes" section like we usually do when fixing false positives for rules.
* Linked the PR where the issue was reported in the changelog.
* Hard-wrapped other changelog entries in master that were over the 80 character limit.

---

I have one last question before we can merge this: why did you bump the minimum Swift version to 4.1? Whatever the reason, it'd be good to add a comment explaining why this rule wouldn't be available to older Swift versions.

dirtydanee:
thanks for cleaning the code up, the PR was originally a bit sloppy.

`I have one last question before we can merge this: why did you bump the minimum Swift version to 4.1?`

The ACL attribute modifiers (`.private` and `.fileprivate`) are only available from Swift 4.1 as part of the `SwiftDeclarationAttributeKind`.

jpsim:
Sounds good, thanks!

PR #2505, opened 2018-12-06
@objc to private unit test

### LINKED ISSUE #2282: [Rule Request Modification] - PrivateUnitTestRule.swift Unit tests marked 'private' warning incorrect
### New Issue Checklist

- [X ] Updated SwiftLint to the latest version
- [X ] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Rule Request Modification

On an XCTestCase if a method is marked private the following warning triggers: 
Private Unit Test Violation: Unit test marked "private" will not be run by XCTest. (private_unit_test)

However, if the @objc keyword is used it will run.

kimdv:
Fixes #2282 🚀

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/PrivateUnitTestRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/PrivateUnitTestRule.swift
@@ -28,16 +28,6 @@ private extension Dictionary where Key: ExpressibleByStringLiteral {
     }
 }
 
-private extension Array where Element == [String: SourceKitRepresentable] {
-    func enclosedSwiftAttributes() -> Set<SwiftDeclarationAttributeKind> {
-        let swiftAttributes = flatMap { $0.swiftAttributes }
-            .compactMap { $0.attribute }
-            .compactMap(SwiftDeclarationAttributeKind.init(rawValue: ))
-
-        return Set(swiftAttributes)
-    }
-}
-
 public struct PrivateUnitTestRule: ASTRule, ConfigurationProviderRule, CacheDescriptionProvider, AutomaticTestableRule {
     public var configuration: PrivateUnitTestConfiguration = {
         var configuration = PrivateUnitTestConfiguration(identifier: "private_unit_test")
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Rules/Lint/PrivateUnitTestRule.swift
```diff
@@ -63,6 +73,12 @@ public struct PrivateUnitTestRule: ASTRule, ConfigurationProviderRule, CacheDesc
                 "internal func test2() {}\n " +
                 "public func test3() {}\n " +
             "}",
+            "@objc" +
+            "private class FooTest: XCTestCase { " +
```
ornithocoder: This will generate "@objcprivate". How about keeping it in one line as "@objc private class ..."

ornithocoder:
Don't forget the add you work to the change log :-)

Source/SwiftLintFramework/Rules/Lint/PrivateUnitTestRule.swift
```diff
@@ -63,6 +63,11 @@ public struct PrivateUnitTestRule: ASTRule, ConfigurationProviderRule, CacheDesc
                 "internal func test2() {}\n " +
                 "public func test3() {}\n " +
             "}",
+            "@objc private class FooTest: XCTestCase { " +
```
jpsim: In the future, you can use multiline string literals and format this as you would normal Swift code. The code layout here is a bit odd. No big deal, we can still merge with this.

jpsim:
Yeah it could be tedious but nice to clean up our triggering & non-triggering examples. I see now that you followed the style used for the other examples in that rule, but here's what it looks like:

```swift
@objc private class FooTest: XCTestCase { @objc private func test1() {}
 internal func test2() {}
 public func test3() {}
 }
```

Whereas I think it could be formatted like this for improved legibility:

```swift
@objc private class FooTest: XCTestCase {
  @objc private func test1() {}
  internal func test2() {}
  public func test3() {}
}
```

I do think generally aligning with the style used in the SwiftLint codebase itself is a good start. I prefer two-space indentation in these examples as opposed to our usual 4-space indentation because the snippets are usually very short, but I could be convinced otherwise.

PR #2506, opened 2018-12-06
Converted idiomatic triggers and nonTriggers to multiline string literals

kimdv:
I still need to convert some rules triggers
But here is the first approach as you described [here](https://github.com/realm/SwiftLint/pull/2505#issuecomment-444961761) @jpsim


jpsim:
Big thanks for this cleanup! I made a few small changes and will be merging once CI is happy.

PR #2508, opened 2018-12-12
Support glob patterns without the star

mgrebenets:
Enables use of '[abc]', '?' and '[a-z]' glob syntax without the need to use '*' in the pattern.


jpsim:
Awesome, thank you!

PR #2514, opened 2018-12-17
Update Mint installation instructions

yonaskolb:
Now that Mint links a global version on install, a better example of installation is to use `mint install` rather than `mint run`


jpsim:
Thanks @yonaskolb! I also updated the Korean readme.

PR #2517, opened 2018-12-19
Make WeakDelegateRule not trigger on immediately initialized properties

skagedal:
> Any such fooDelegate would be something the object holds on to and is a delegate for something else.

It can sometimes be convenient to use a helper class that only has the job of being a delegate for another class. It may make sense to call that class `FooDelegate` and holding on to it in a private property called `fooDelegate`. But this isn't something that acts as a delegate for something, thus should not be weak and should not be affected by this rule.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2433](https://github.com/realm/SwiftLint/issues/2433)
 
+* Make `weak_delegate` not trigger on properties that are immediately assigned.  
+  [skagedal](https://github.com/skagedal)
+  
 #### Bug Fixes
 
 * Fix inaccessible custom rules in nested configurations.  
--- a/Rules.md
+++ b/Rules.md
@@ -23017,6 +23017,13 @@ class Foo {
 }
 ```
 
+```swift
+class Foo {
+ var immediatelyInitializedDelegate = SomeClass()
+ }
+
+```
+
 </details>
 <details>
 <summary>Triggering Examples</summary>
--- a/Source/SwiftLintFramework/Rules/Lint/WeakDelegateRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/WeakDelegateRule.swift
@@ -24,7 +24,8 @@ public struct WeakDelegateRule: ASTRule, ConfigurationProviderRule, AutomaticTes
             // There's no way to declare a property weak in a protocol
             "protocol P {\n var delegate: AnyObject? { get set }\n}\n",
             "class Foo {\n protocol P {\n var delegate: AnyObject? { get set }\n}\n}\n",
-            "class Foo {\n var computedDelegate: ComputedDelegate {\n return bar() \n} \n}"
+            "class Foo {\n var computedDelegate: ComputedDelegate {\n return bar() \n} \n}",
+            "class Foo {\n var immediatelyInitializedDelegate = SomeClass()\n }\n"
         ],
         triggeringExamples: [
             "class Foo {\n  ↓var delegate: SomeProtocol?\n}\n",
@@ -57,6 +58,9 @@ public struct WeakDelegateRule: ASTRule, ConfigurationProviderRule, AutomaticTes
         // Check if non-computed
         let isComputed = (dictionary.bodyLength ?? 0) > 0
         guard !isComputed else { return [] }
+
+        // Check if the declaration is followed by a = sign
+        guard !isImmediatelyAssigned(file: file, dictionary: dictionary) else { return [] }
 
         // Violation found!
         let location: Location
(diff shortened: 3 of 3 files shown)


jpsim:
Thanks for the PR! I'd like to confirm with the original author of the rule @AliSoftware to see if he thinks this should be the default behavior or a configuration option.

AliSoftware:
I personally used private weak delegates, especially in cases where I have an object on which I don't want the delegate to be changed during the lifetime of the object, so I make the delegate private (or at least `private(set)`) and pass the delegate to be used via the `init(delegate:…)`. That delegate, even if it's private, still has to be weak to avoid retain cycles (`let foo = Foo(); foo.bar = Bar(delegate: foo)`) though.
I think I've used that pattern more often than the proxy pattern described by @skagedal here, so I'd personally prefer weak to be the default even for private delegates, and have this rule opt-in for this reason… but of course YMMV 😉

AliSoftware:
Although… in fact I did also use the pattern described by @skagedal more recently and even [made a blog post about it and about the need to retain those data source/delegates in such cases indeed](http://alisoftware.github.io/architecture/2018/05/20/dedicated-datasources/#retain-your-datasources) now that I think about it… so… 🤷‍♂️😅

jpsim:
Thanks for the quick reply, Olivier! I'm also thinking a configuration option would be best here.

skagedal:
Ah, thanks – I did not consider the case where the delegate is injected in `init`, even though I have used such a pattern myself before. The best solution would of course support both patterns in the same code base... One possible heuristic would be to assume that a fully `private` delegate variable is likely the "proxy" pattern, so should not warn, while a `private(set)` one would be the injected kind (that's how I would do it...), so it should warn. But perhaps that is a bit too ad-hoc...? Also, you seem to say that you sometimes have injected delegates as fully private, @AliSoftware. 

If we would make it an option, now when I think more about it, I think the current behavior should be the default – false negatives should be considered worse than false positives. If I opted in to this rule, I would be very annoyed to find that it had missed a missing `weak`, but only slightly annoyed if it found one where it shouldn't have. :)  

Here's yet another suggestion. The cases I had in my codebase that were false positives looked like this:

```
private let navigationDelegate = MyNavigationDelegate()
```

So how about looking for properties that are `let` properties and are initialized at the point of declaration, and make those exceptions to the warning? That should guarantee that the delegate is not in fact something from an external source. I have no idea how easy this would be to implement...

skagedal:
I have now implemented this logic: whenever a property is immediately initialized, i.e. the name of the var is followed by an equals sign, then we should not warn about weak delegate. 

In such instances, it would not make sense to make the property weak – Xcode would warn that the object would be immediately deallocated, since nothing is holding on to it. We thus assume that the user intended this to be a "helper delegate" of the above described kind.

I think this logic makes sense but it's a good idea of course if someone would take a look at my implementation.

AliSoftware:
So `var delegate: MyDelegate? = nil` would not warn?

skagedal:
Sigh. No, it wouldn't. Did not consider that. Back to the drawing board...

skagedal:
People shouldn't explicitly write `= nil`. :)

skagedal:
If I update it so that ` = nil` is an exception to the exception, do you think that would work?

AliSoftware:
Might be. But tbh at that point I'm wondering if we're not engaging in a dangerous route, trying to think about all the special cases we'll likely forget one… and have cases where it should warn end up not warning (which would be more problematic imho).

Does this case of helper delegate happens _that_ often in codebases — compared to real weak delegates — so that it really is worth the risk of having false negatives in that rule? As opposed to letting people who use helper delegates in some isolated places just use `// swiftlint:disable:next weak_delegate` on the previous line on a case by case basis?

AliSoftware:
Also personally I prefer a codebase where I have `// swiftlint:disable:next weak_delegate` explicitly, because then when I read the code, it will signal me and any coworker reviewing the code that I do want that delegate to be strong, that this is not a mistake and that it's intentionally retained, and that if you mark it `weak` later thinking it was a mistake, it will not work as expected.

skagedal:
Yeah, I guess you're right... not sure I agree on the last point – I'd prefer the linter to not warn in cases where it makes no sense to warn – but regarding the rule getting too complex. I'm willing to close the PR.

AliSoftware:
Yeah, the problem here is "cases where it makes no sense to warn" might be too difficult to detect, because those cases happen to also risk being "cases where the user wrote bad code and we should warn, or have a twisted use case we didn't think about".

* Avoiding false positives is easy for the end user: they just can `swiftlint:disable` to acknowledge that this is a case they thought about and know what they are doing by not making it `weak`.
* Avoiding false negatives is dangerous, as the whole point of SwiftLint is to warn the end user about cases they might have overlooked, and the user won't be the one telling "hey you didn't warn here, but you should have", as if we're at that point, that would mean the user already knew something was wrong and would then fix it directly.

If there were a definitive and clear way to distinguish between "the user forgot to mark it weak" and "this is normal that this is weak", I'd be all for such an adjustment to the rule of course, but given how many trials and errors and uncertainty and cases we didn't think about while we had that conversation in this PR, we can clearly see that it's far from black & white and imho the risk is too high. That's too bad, as I agree that it would be cool to avoid false positives if it were possible without doubt, and I'd still love if someone came up with a definitive way to determine that, but doesn't seem to be the case 😞

Issue #2518, opened 2018-12-20
swiftlint_version silently ignores incorrectly typed values


skagedal:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

SwiftLint has a mechanism for pinning the tool to a specific version in the config file. One may write like this:

```
swiftlint_version: 0.29.1
```

And it will only run with that version of SwiftLint. If you write like this instead:

```
swiftlint_version: 0.29
```

It will also run. The user may think, as I did, that it has accepted 0.29.1 as being "at least" 0.29. That is however not what's happening, instead it is parsing `0.29` as a float and silently ignoring the directive. See [Configuration+Parsing.swift, line 99](https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Extensions/Configuration%2BParsing.swift#L99).
##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint version
0.29.1
$ echo 'swiftlint_version: 0.29.1' > .swiftlint.yml
$ swiftlint                                        
Loading configuration from '.swiftlint.yml'
Linting Swift files at paths 
No lintable files found at paths: ''
$ echo 'swiftlint_version: 0.30.1' > .swiftlint.yml
$ swiftlint                                        
Loading configuration from '.swiftlint.yml'
Currently running SwiftLint 0.29.1 but configuration specified version 0.30.1.
$ echo 'swiftlint_version: 0.30' > .swiftlint.yml 
$ swiftlint                                      
Loading configuration from '.swiftlint.yml'
Linting Swift files at paths 
No lintable files found at paths: ''
```

Expected: that the last invocation of SwiftLint either printed "Currently running SwiftLint 0.29.1 but configuration specified version 0.30" or failed with "incorrect format for .swiftlint.yml"

### Environment

* SwiftLint version (run `swiftlint version` to be sure)?
0.29.1
* Installation method used (Homebrew, CocoaPods, building from source, etc)?
Homebrew
* Paste your configuration file:
As above

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)?
  If so, paste their relative paths and respective contents.
No
* Which Xcode version are you using (check `xcode-select -p`)?
10.1

jpsim:
Fix incoming in #2523.

PR #2519, opened 2018-12-20
Check the type of swiftlint_version input.

skagedal:
See issue #2518. This fixes the problem by checking the type of input.

If the wrong type is found, it prints an error message and continues. This is consistent with some other similar parsing in the file, like the check for invalid keys. I personally think it would make sense to exit with error. I tried to instead return nil from the initializer, but then SwiftLint crashed. 

Not the prettiest code, I could try clean it up if you think the general approach is right.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,10 @@
   [Daniel Metzing](https://github.com/dirtydanee)
   [#2499](https://github.com/realm/SwiftLint/issues/2499)
 
+* Verify correct type of the `swiftlint_version` configuration value.  
+  [skagedal](https://github.com/skagedal)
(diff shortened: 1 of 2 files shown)


jpsim:
Thanks for the PR @skagedal! Sorry I didn't see this before implementing #2523. Going with that one instead.

skagedal:
No worries, thanks for looking at it. I considered the stringification solution first but then thought that the diagnostics would be weird if you had `swiftlint_version = 0.30`. It would then say something like:

> Currently running SwiftLint 0.30.1 but configuration specified version 0.3.

jpsim:
Yeah, you're right that's not ideal. I think given the way in which this will be (mis-)used though, a misconfiguration would be caught just as quickly and still avoids the risk of not knowing it was misconfigured like was previously the case.

PR #2527, opened 2018-12-24
Add configuration to discouraged_object_literal

### LINKED ISSUE #2439: Allow configuration of 'discouraged_object_literal' rule
### New Issue Checklist

### New rule request

Similar to the `object_literal` rule, I'd like to have configurability for the inverse `discouraged_object_literal` rule, to enforce literals for some objects, and enforce the opposite for others.

Currently, only enforcing literals can be configured, but you cannot selectively discourage the others.

The main motivation comes from Xcode 10's terrible image literal support (you can basically only use them by dragging from the media library, and it's very hard to see or edit the raw image names in Xcode, nor do you see a preview).

Our goal is to enforce literals for all objects (currently that means colors), except for images, where we want to enforce non-literal syntax.

The relevant part of our config file would look like this:

```
opt_in_rules:
  - discouraged_object_literal
  - object_literal

# this part doesn't work yet
discouraged_object_literal:
  color_literal: false

# this already works
object_literal:
  image_literal: false
```

I would assume the existing `ObjectLiteralConfiguration` can be used for both rules.

Unless I'm missing something and this option shouldn't be implemented after all, I would volunteer to look into the code myself.

marcelofabri:
Fixes #2439

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,10 @@
 
 #### Enhancements
 
-* None.
+* Allow configuring `discouraged_object_literal` rule to only discourage one kind
+  of object literal.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#2439](https://github.com/realm/SwiftLint/issues/2439)
 
 #### Bug Fixes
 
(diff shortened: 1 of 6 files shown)


jpsim:
Thanks!

Issue #2441, opened 2018-10-16
Can't disable custom rule in nested config file


jpsim:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

When attempting to disable a custom rule defined in a parent configuration from a nested configuration, SwiftLint can't find that rule and prints the message `configuration error: 'my_custom_rule' is not a valid rule identifier`.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ tree -a
.
├── .swiftlint.yml
├── dir1
│   ├── .swiftlint.yml
│   └── file2.swift
└── file1.swift

1 directory, 4 files
$ cat .swiftlint.yml
custom_rules:
  no_abc:
    name: "Don't use abc"
    regex: 'abc'

$ cat file1.swift
// abc
$ cat dir1/.swiftlint.yml
disabled_rules:
  - no_abc
$ cat dir1/file2.swift
// abc
$ swiftlint
Loading configuration from '.swiftlint.yml'
Linting Swift files at paths ...
configuration error: 'no_abc' is not a valid rule identifier
Valid rule identifiers:
anyobject_protocol
...
yoda_condition
Linting 'file1.swift' (1/2)
Linting 'file2.swift' (2/2)
/path/to/dir/file1.swift:1:4: warning: Don't use abc Violation: Regex matched. (no_abc)
/path/to/dir/dir1/file2.swift:1:4: warning: Don't use abc Violation: Regex matched. (no_abc)
Done linting! Found 2 violations, 0 serious in 2 files.
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.27.0
* Installation method used (Homebrew, CocoaPods, building from source, etc)? Homebrew
* Which Xcode version are you using (check `xcode-select -p`)? `/Applications/Xcode-10.app/Contents/Developer`

jpsim:
Fixed in #2556.

Issue #1815, opened 2017-09-03
Custom Rules in nested configuration


ornithocoder:
### New Issue Checklist

- [x] I updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Bug Report

I'm using a custom rule to validate the name of my Quick Specs (I want them to end with `Tests`, e.g. `AppCordinatorTests`). For that, I implemented a simple custom rule and placed it in the `.swiftlint.yml` that I have inside the Tests directory. It didn't work. So I moved the custom rule to the top level `.swiftlint.yml` and it worked.

Apparently custom rules don't work if placed in nested configurations.

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.22.0
* Installation method used (Homebrew, CocoaPods, building from source, etc)? Homebrew
* Paste your configuration file:

```yml
# ...

custom_rules:
  quickspec_class_name:
    name: "Invalid Test Class Name"
    regex: "class\\s+(.*)[^Tests]:\\s+QuickSpec"
    message: "A test class name must end with 'Tests'"
    severity: error
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? Yes, I am.

* Which Xcode version are you using (check `xcode-select -p`)? Xcode 9b6 (App and Command Line)

```swift
// This triggers a violation:
class Foo: QuickSpec { }

// Doesn't trigger a violation:
class FooTests: QuickSpec { }
```

ornithocoder:
To reproduce, create a file called `issue1815.sh` and paste the following:

```bash
#!/usr/bin/env sh

mkdir Issue1815
cd Issue1815

touch .swiftlint.yml
mkdir Tests
cd Tests

cat > .swiftlint.yaml << EOF
custom_rules:
  quickspec_class_name:
    name: "Invalid Test Class Name"
    regex: "class\\\\s+(.*)[^Tests]:\\\\s+QuickSpec"
    message: "A test class name must end with 'Tests'"
    severity: error
EOF

cat > FooTests.swift << EOF
// This triggers a violation:
class Foo: QuickSpec { }

// Doesn't trigger a violation:
class FooTests: QuickSpec { }
EOF
```

[...]

PR #2532, opened 2018-12-25
#2441 - Fix unaccessible custom rules in nested configurations

biboran:
This PR addresses 2 issues:
* As per #2441 a nested configuration wasn't able properly initialize due to temporal coupling with `enabledRules` routine that was called before the nested configuration merge. To remedy this issue I just passed the `customRulesIdentifiers` to the `enabledRules` thus making it aware of the parent's custom rules. I guess it is not an elegant solution but it works.
* Another issue I found was that nested configuration wasn't able to define its own custom rules because they wouldn't be merged (#1815). The parent's `CustomRules` instance would always be the one left by the `union` in `Configuration+Merging.swift`. This was also the reason why disabling custom rules in the nested configuration was impossible. I just added a special method that merges custom rules separately.

### DIFF
--- a/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
@@ -47,7 +47,7 @@ extension Configuration {
     }
 
     public init?(dict: [String: Any], ruleList: RuleList = masterRuleList, enableAllRules: Bool = false,
-                 cachePath: String? = nil) {
+                 cachePath: String? = nil, customRulesIdentifiers: [String] = []) {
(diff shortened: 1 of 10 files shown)


Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
```diff
@@ -46,8 +46,9 @@ extension Configuration {
         return .default
     }
 
+    // swiftlint:disable:next function_body_length
```
biboran: Not sure what to refactor here to prevent this

jpsim:
Very good work here @biboran! I just had a few edits and decided to make them in a new PR instead of overriding your branch here. Let's continue in #2556.

Issue #2529, opened 2018-12-24
[RFC] Plugin system for custom rules


marcelofabri:
## Motivation

We currently support custom rules as regexes, but often that's not the right tool for the job. Rules implemented with regexes on SwiftLint are usually the slower ones and also the ones with more false positives. Having first-class support to a plugin system would be essential for empowering users to automatically enforce specific patterns that don't necessarily belong to SwiftLint itself.

## Proposal

This proposal describes a flexible plugin system that doesn't require new rules to be written using Swift. We should provide tools to make it possible though, but that will be left as a future enhancement.

A plugin is an executable that will be run by SwiftLint and should respond to certain commands. Communication will happen via stdin and stdout using JSON unless specified.

A plugin maps 1:1 to a rule, so you can't implement more than one rule in the same plugin.

**Note**: so far this is just a proposal, no concrete implementation has started as far as I know.

### Commands

`plugin` represents the path to a generic custom plugin executable in the commands below.

#### `plugin rule_description`

This will return a JSON containing all the required data to create a `RuleDescription`. Will be run by SwiftLint when parsing the configuration and resolving the rules.

**Output:**

```json
{
    "identifier": "my_cool_rule",
    "name": "Cool Rule",
    "kind": "style",
    "non_triggering_examples": [
        "let x = 10"
    ],
    "triggering_examples": [
        "let x = 20"
    ],
    "min_swift_version": "4.1.0"
}
```

#### `plugin lint`

This will receive a JSON as input and should return another JSON containing the violations.

**Input**:
```json
{
    "structure": ... ,
    "syntax_map": ...,
    "path": "file.swift",
    "contents": "",
    "configuration": {
        "cool_number": 10
    }
}
```

Note: `path` is optional. I have mixed feeling whether `contents` should be populated if `path` is present.

`configuration` contains the raw configuration loaded from the configuration YAML file as described below in "Configuring a plugin".

**Output**:
```json
[
    {
        "severity": "warning",
        "reason": "Failed",
        "location": {
            "line": 1,
            "character: 10
        },
        "byte_offset": 203,
        "character_offset": 300
    }
]
```

Note: `location`, `byte_offset` and `character_offset` are exclusive (only one should be provided). `reason` is optional.

#### Alternatives considered

Another option is to have the `lint` command run only once and keep feeding it JSONs. This would have the benefit of being faster (as fixed costs such as loading dynamic libraries would be paid only once) but adds more complexity.

[...]

PR #2536, opened 2018-12-31
Remove OptInRule protocol

marcelofabri:
This PR removes the `OptInRule` protocol in favor of a property in `RuleDescription`.

It allows [plugins to expose opt-in rules](https://github.com/realm/SwiftLint/issues/2529#issuecomment-450695614) without much effort.

The downside IMO is that for we don't get "being opt-in" for free in `AnalyzerRule` anymore.

### DIFF
--- a/Source/SwiftLintFramework/Models/Configuration.swift
+++ b/Source/SwiftLintFramework/Models/Configuration.swift
@@ -245,7 +245,7 @@ private func enabledRules(from configuredRules: [Rule],
         return configuredRules.filter { rule in
             let id = type(of: rule).description.identifier
             if validDisabledRuleIdentifiers.contains(id) { return false }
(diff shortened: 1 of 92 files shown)


Dschee:
This looks like a nice simplification, but I don't really understand what `AnalyzerRule` was for and therefore don't understand the impact of this being merged. In what situation do the errors arise, is it part of the tests?

marcelofabri:
This will likely not be merged if we don't go with socket-based plugins (which is probably a good idea)

Dschee:
So we can close this if the evaluation of https://github.com/realm/SwiftLint/issues/2529 leads to a non-socket-based plugin system?

marcelofabri:
> So we can close this if the evaluation of #2529 leads to a non-socket-based plugin system?

That's right.

PR #2540, opened 2019-01-03
Fix false positives on `redundant_objc_attribute` rule when using nested types

### LINKED ISSUE #2539: redundant_objc_attribute incorrectly flagging inner @objc enum
The following code triggers `redundant_objc_attribute`:

```swift
@objcMembers 
class Foo {
  @objc enum Bar { }
}
```

The code to flag a redundant `@objc` should only trigger if the declaration is a member, not an inner type.

I'm gonna take a stab at fixing this myself.

marcelofabri:
Fixes #2539

### DIFF
--- a/Rules.md
+++ b/Rules.md
@@ -15320,16 +15320,6 @@ class Foo: NSObject {
 ```
 
 ```swift
-@objc
-extension Foo {
-  @objc
-  private var bar: Int {
-    return 0
-  }
-}
-```
-
-```swift
 @objcMembers
 class Foo {
     class Bar: NSObject {
@@ -15431,6 +15421,16 @@ class Foo {
     @objcMembers
     class Bar: NSObject {
         @objc ↓var foo: Any
+    }
+}
+```
+
+```swift
+@objc
+extension Foo {
+    @objc
+    private var bar: Int {
+        return 0
     }
 }
 ```
--- a/Rules.md
+++ b/Rules.md
@@ -15429,7 +15429,7 @@ class Foo {
 @objc
 extension Foo {
     @objc
-    private var bar: Int {
+    private ↓var bar: Int {
         return 0
     }
 }
--- a/Rules.md
+++ b/Rules.md
@@ -15319,6 +15319,32 @@ class Foo: NSObject {
 }
 ```
 
+```swift
+@objc
+extension Foo {
+  @objc
+  private var bar: Int {
+    return 0
+  }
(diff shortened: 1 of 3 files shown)


Rules.md
```diff
@@ -15400,12 +15426,12 @@ extension Foo {
 ```
 
 ```swift
-@objc
```
dirtydanee: It seems from the diff that this triggering example was removed. Could you add it back to make sure it is triggered?

marcelofabri: I moved it to be a non-triggering example because the declaration is private (similar to https://github.com/realm/SwiftLint/pull/2501)

dirtydanee: Ok, i see, i agree, its a good idea to have it in the non triggering too.
However it is redundant to mark any attribute in an already `@objc` marked `extension`, regardless what the `ACL` is.

marcelofabri: Added it back in 0c92b948a9ebe52e30b44371ebf9cff9c4de8832

Source/SwiftLintFramework/Rules/Idiomatic/RedundantObjcAttributeRule.swift
```diff
@@ -1,12 +1,11 @@
-import Foundation
 import SourceKittenFramework
 
 private let kindsImplyingObjc: Set<SwiftDeclarationAttributeKind> =
     [.ibaction, .iboutlet, .ibinspectable, .gkinspectable, .ibdesignable, .nsManaged]
 
 private let privateACL: Set<SwiftDeclarationAttributeKind> = [.private, .fileprivate]
```
dirtydanee: Due to your last change (`!acl.isPrivate`), this attribute is not needed any more.

dirtydanee:
> @dirtydanee @JaviSoto could you take a look at this and see if it makes sense? I'm not super familiar with `@objcMembers` to be 100% sure this is right.

I think the updated implementation is pretty neat, and its a much better approach to only check the parent structure instead of iterating on the whole file on each declaration kind.

JaviSoto:
Hey! Sorry it took me so long to reply. Yeah this code looks so much better! It also solves the scenario I posted on #2539:

```swift
@objcMembers
class Foo {
  @objc enum Bar { }
}
```

But if I add it to the "non triggering examples", it's passing now 👍

jpsim:
Nice work!

PR #2541, opened 2019-01-04
[WIP] Plugins

### LINKED ISSUE #2529: [RFC] Plugin system for custom rules

marcelofabri:
Will eventually fix #2529.

Posting it here to gather feedback and highlight some of the changes/decisions.

Missing:
- [ ] Handle nested configurations
- [ ] Add tests for caching
- [ ] Show remote rules in `swiftlint rules` (although I don't think this is *necessarily* a blocker)
- [ ] Allow plugins to expose more than one rule (depends on https://github.com/realm/SwiftLint/pull/2536)
- [ ] Documentation

Will be done later:
- [ ] Create APIs on `SwiftLintFramework` that a plugin can use to register a rule. My idea is to allow users to create a rule using the existing types (`Rule`, `ASTRule`, etc) and just call a register method. Something like this:

```swift

struct MyRule: Rule { 
 // ...
}

let plugin = Plugin()
plugin.register(rules: [MyRule.self])
plugin.run()
```

- [ ] Revise public APIs on `SwiftLintFramework`. In order to be productive, we should consider exposing (a lot) of the convenience functions/types/extensions that are currently `internal`. However, this increases the public API surface. I don't think that too many people use `SwiftLintFramework` directly, so I'm not sure how important it's to not break things.


Cartfile.private
```diff
@@ -1,3 +1,4 @@
 github "Carthage/Commandant" ~> 0.15.0
+github "IBM-Swift/BlueSocket" ~> 1.0.0
```
marcelofabri: new dependency: are we ok with it?

Makefile
```diff
@@ -119,7 +119,7 @@ archive:
 release: package archive portable_zip
 
 docker_test:
-	docker run -v `pwd`:`pwd` -w `pwd` --name swiftlint --rm norionomura/swift:42 swift test --parallel
+	docker run -v `pwd`:`pwd` -w `pwd` --name swiftlint --rm norionomura/swift:421 swift test --parallel
```
marcelofabri: there's an issue on Linux when compiling with 4.2 which is fixed on 4.2.1 (should we highlight this bump somewhere)?

Source/SwiftLintFramework/Models/Linter.swift
```diff
@@ -7,16 +7,83 @@ private struct LintResult {
     let deprecatedToValidIDPairs: [(String, String)]
 }
 
-private extension Rule {
-    static func superfluousDisableCommandViolations(regions: [Region],
-                                                    superfluousDisableCommandRule: SuperfluousDisableCommandRule?,
-                                                    allViolations: [StyleViolation]) -> [StyleViolation] {
+private enum LintableBox {
+    case rule(Rule)
+    case remoteRule(RemoteRule)
```
marcelofabri: I'm not a big fan of having 2 totally different models, but it seemed a better idea than having something like `CustomRules` which gives us surprising results sometimes.

Source/SwiftLintFramework/Models/RemoteRule.swift
```diff
@@ -0,0 +1,79 @@
...
+                                        path: file.path,
+                                        contents: Lazy(file.contents),
+                                        configuration: configuration)
+        return validate(payload: payload, file: file)
+    }
+
+    private func validate(payload: RemoteRulePayload, file: File) -> [StyleViolation] {
+        do {
+            let socket = try Socket.create(family: .unix, type: .stream, proto: .unix)
+            try socket.connect(to: "/tmp/\(ruleDescription.identifier).socket")
```
marcelofabri: are we okay with hardcoding the path like this?

Source/SwiftLintFramework/Models/RemoteRulePayload.swift
```diff
@@ -0,0 +1,81 @@
...
+        if path == nil && contents == nil {
+            return nil
+        }
+
+        self.init(structure: Lazy(structure), syntaxMap: Lazy(syntaxMap),
+                  path: path, contents: Lazy(contents), configuration: configuration)
+    }
+}
+
+private func convertingIntsToInt64(value: Any) -> Any {
```
marcelofabri: not a big fan, but couldn't think about anything better

SwiftLint.xcodeproj/project.pbxproj
```diff
@@ -2195,7 +2247,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = D0D1212819E878CC005E4BAA /* Release.xcconfig */;
 			buildSettings = {
-				MACOSX_DEPLOYMENT_TARGET = 10.10;
+				MACOSX_DEPLOYMENT_TARGET = 10.11;
```
marcelofabri: had to bump this because of `BlueSocket` - hopefully, it's ok since 10.11 is already pretty old by now

therealbnut:
This looks fantastic, an epic undertaking. It should be really powerful and useful when it’s done. 💯

While we wait there’s a few neglected PRs that improve custom rules, which should allow people to start using them properly. IMO more custom rules will seed a more vibrant plugin community. The oldest open PR on this repo is such a PR, it was blocked 2 years ago with the only real comment saying to wait for plugins. I’ve made one more recently: 
https://github.com/realm/SwiftLint/pull/2586

I don’t think it’s productive to block custom rule improvements until plugins are ready, even if plugins will be ready in as little as a month. Custom rule improvements allow people to explore the possibilities plugins will allow.

Awesome work, I look forward to this landing! Please say if there’s any plugin tasks the community can help with. This weekend if I have time I might experiment with moving the existing rules into a separate module, so we can determine what APIs need to be public for plugins.

marcelofabri:
> I don’t think it’s productive to block custom rule improvements until plugins are read

I don't think that's why these PRs weren't reviewed. It's probably just that no maintainer was able to dedicate the time to review them, as all maintainers work on SwiftLint on their free time.

> Awesome work, I look forward to this landing! Please say if there are any plugin tasks the community can help with. This weekend if I have time I might experiment with moving the existing rules into a separate module, so we can determine what APIs need to be public for plugins.

Thanks! I think that'd be an interesting exercise. 

Another update on plugins: I've talked with @jpsim a few weeks ago and we decided that we should explore having plugins as Swift files in a `.swiftlint` folder and then SwiftLint would compile them as one (or multiple - TBD) dynamic library. This has some benefits over my initial suggestion:

- Plugins can run in process and much faster
- Less complexity (we don't need to change much code after a `Rule` is registered from a plugin, no sockets, etc)
- Better experience for users

However, I wasn't able to get a proof of concept working. I've tried using SPM, but run into https://bugs.swift.org/browse/SR-6091. The problem is:

1. Plugin will need to depend on SwiftLintFramework dynamically 
2. We link SwiftLintFramework statically on the `swiftlint` binary
3. We'll open the plugin dylib on runtime with `dlopen`

Help with getting the right incantation would be super valuable! We don't necessarily need to use SPM,  but I thought it would be helpful for `(1)`.

PR #2543, opened 2019-01-05
Issue 2538: vertical_whitespace_between_cases gives false warning if blank line has whitespace

BenStaveleyTaylor:
Here's a go at fixing issue 2538 in vertical_whitespace_between_cases.

It would be simpler if there was a regex pattern change that would correctly skip lines containing whitespace, but the only way I could think of was to use a negative look-behind assertion and NSRegularExpression does not support use of `*` or `+` in those (to look-behind for `\n\s+\n`). Maybe someone with greater skill in regexes has way?

So instead I introduced a filter to remove false positives in the violating range set, which I can see some other rules use too.

I tested the behaviour both when linting and autocorrecting.

I can't quite see what to do about adding a unit test for this case. Trailing white space in the `nonTriggeringExamples` literal array will be invisible to users and also causes a trailing_whitespace violation itself. Please advise me on what to do if I need to add a test somehow.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Style/VerticalWhitespaceBetweenCasesRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/VerticalWhitespaceBetweenCasesRule.swift
@@ -104,6 +104,28 @@ public struct VerticalWhitespaceBetweenCasesRule: ConfigurationProviderRule {
     ]
 
     private let pattern = "([^\\n{][ \\t]*\\n)([ \\t]*(?:case[^\\n]+|default):[ \\t]*\\n)"
+
+    private func violationRanges(in file: File) -> [NSRange] {
+        return file.violatingRanges(for: pattern).filter {
+            !isFalsePositive(in: file, range: $0)
+        }
+    }
+
+    private func isFalsePositive(in file: File, range: NSRange) -> Bool {
+        // Regex incorrectly flags blank lines that contain trailing whitespace (#2538)
+        let patternRegex = regex(pattern)
+        let substring = file.contents.substring(from: range.location, length: range.length)
+        let substringRange = NSRange(location: 0, length: substring.count)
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Rules/Style/VerticalWhitespaceBetweenCasesRule.swift
```diff
@@ -104,6 +104,28 @@ public struct VerticalWhitespaceBetweenCasesRule: ConfigurationProviderRule {
...
+        return file.violatingRanges(for: pattern).filter {
+            !isFalsePositive(in: file, range: $0)
+        }
+    }
+
+    private func isFalsePositive(in file: File, range: NSRange) -> Bool {
+        // Regex incorrectly flags blank lines that contain trailing whitespace (#2538)
+        let patternRegex = regex(pattern)
+        let substring = file.contents.substring(from: range.location, length: range.length)
+        let substringRange = NSRange(location: 0, length: substring.count)
```
marcelofabri: this should be `substring.utf16.count` (or even better, you can use `substring.fullNSRange`)

marcelofabri:
@BenStaveleyTaylor this looks great! I've added one comment about the implementation.

About the tests, it's fine if it looks the same IMO. You can use `swiftlint:disable trailing_whitespace` and `swiftlint:enable trailing_whitespace` to only disable the rule in the test region.

BenStaveleyTaylor:
@marcelofabri Changed as you suggested. (That range code was in the rule before, I only moved the line around actually).

I added a new validation test case and avoided the need for `swiftlint:disable trailing_whitespace` by using single line string syntax with concatenation rather than multiline. It is not consistent with the other tests in the rule, but I think it is safer as Xcode or other tool might trim the whitespace off if it is not inside explicit quotes.

PR #2545, opened 2019-01-05
Add unused_control_flow_label rule

### LINKED ISSUE #2227: Rule Request: unused control flow labels

Typescript just added a way to identify labels which are not being used. I think this would be really nice to have in swift as lint rule. 

https://github.com/Microsoft/TypeScript/pull/24037/files

>2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

```c
label2:  /// label2 not being used. Consider removing
    while (true) {}
```

do, while, repeat should be supported

non triggering examples:

```c
label2: 
    while (true) { break label2}
```
>3. Should the rule be configurable, if so what parameters should be configurable?

none

>4. Should the rule be opt-in or enabled by default? Why?
   rule as opt-in.

probably opt in

marcelofabri:
Fixes #2227 

Adding as a default rule because I *think* there shouldn't be false positives. Let's wait for oss-check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,11 @@
   `.last(where: { /* ... */ })` is more efficient.  
   [Marcelo Fabri](https://github.com/marcelofabri)
 
+* Add `unused_control_flow_label` rule to validate that control flow labels are
+  used.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#2227](https://github.com/realm/SwiftLint/issues/2227)
+
 #### Bug Fixes
 
 * Fix false positives on `first_where` rule when calling `filter` without a
--- a/Rules.md
+++ b/Rules.md
@@ -21544,7 +21544,7 @@ func foo () {
 
 Identifier | Enabled by default | Supports autocorrection | Kind | Analyzer | Minimum Swift Compiler Version
 --- | --- | --- | --- | --- | ---
-`unused_control_flow_label` | Enabled | No | lint | No | 3.0.0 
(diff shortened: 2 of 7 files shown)


jpsim:
Thanks for doing this! I haven't reviewed yet, but here are some edge cases we should make sure to support:

```swift
if #available(iOS 11, *) {} else {
  // Impossible to negate an `#available` condition
}

if case .loading = state {} else {
  // Impossible to negate a case binding
}
```

jpsim:
Oops, disregard my comment, I _completely_ misunderstood what this rule was validating 😅

CHANGELOG.md
```diff
@@ -27,6 +27,11 @@
   `.last(where: { /* ... */ })` is more efficient.  
   [Marcelo Fabri](https://github.com/marcelofabri)
 
+* Add `unused_control_flow_label` rule to validate that control flow labels are
+  being used.
```
jpsim: nit: s/being used/used/

marcelofabri:
I thought about adding autocorrection but given how rare it’s to use control labels, I thought that maybe it would be better to let a human evaluate if the right fix is to remove the label (vs using it)

jpsim:
IMO it’s easier to review corrections & revert if necessary than to go through the warnings and manually fix them.

marcelofabri:
Fair enough, will do that later 👍

jpsim:
Thanks! Gonna do another quick review pass.

jpsim:
Ran OSSCheck locally since performance numbers from CI were concerning, but got -3% to +3%.

marcelofabri:
Thanks @jpsim 🎉

PR #2556, opened 2019-01-13
#2441 - Fix inaccessible custom rules in nested configurations

jpsim:
Continued from #2532, all credit to @biboran I just made a few small modifications

> This PR addresses 2 issues:
> * As per #2441 a nested configuration wasn't able properly initialize due to temporal coupling with `enabledRules` routine that was called before the nested configuration merge. To remedy this issue I just passed the `customRulesIdentifiers` to the `enabledRules` thus making it aware of the parent's custom rules. I guess it is not an elegant solution but it works.
> * Another issue I found was that nested configuration wasn't able to define its own custom rules because they wouldn't be merged (#1815). The parent's `CustomRules` instance would always be the one left by the `union` in `Configuration+Merging.swift`. This was also the reason why disabling custom rules in the nested configuration was impossible. I just added a special method that merges custom rules separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,10 @@
   [Ben Staveley-Taylor](https://github.com/BenStaveleyTaylor)
   [#2538](https://github.com/realm/SwiftLint/issues/2538)
 
+* Fix unaccessible custom rules in nested configurations.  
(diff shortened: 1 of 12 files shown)


jpsim:
No performance impact running OSSCheck locally.

PR #2557, opened 2019-01-14
Add strong_iboutlet opt-in rule

### LINKED ISSUE #2433: Rule Request: Strong IBOutlet
### New Issue Checklist

### New rule request

Enforce not using `weak` for `@IBOutlet`. Even though Xcode defaults to `weak`, [Apple has recommended using strong outlets since WWDC 2015](https://stackoverflow.com/a/31395938/64949).

Would trigger:

```swift
final class ViewController: NSViewController {
	@IBOutlet weak var label: NSTextField!
}
```

Would not trigger:

```swift
final class ViewController: NSViewController {
	@IBOutlet var label: NSTextField!
}
```

```swift
final class ViewController: NSViewController {
	weak var label: NSTextField = …
}
```

No

I think it should be enabled by default as it adheres to the guidelines and it's what Apple recommends. In the few cases you would actually want it to be weak you can just use a `// swiftlint:disable` comment.

marcelofabri:
Fixes #2433

Added this as an opt-in rule because even though Apple recommends using strong variables on WWDC videos, the Xcode default is to create a `weak` variable so I can see teams preferring sticking to it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,10 @@
 
 #### Enhancements
 
-* None.
+* Add `strong_iboutlet` opt-in rule to enforce that `@IBOutlet`s are not
+  declared as `weak`.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#2433](https://github.com/realm/SwiftLint/issues/2433)
 
 #### Bug Fixes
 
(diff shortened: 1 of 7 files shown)


marcelofabri:
Hmm, not sure why `testSimulateHomebrewTest` is failing since this is not touching anything related to that. Also, that test it passing locally even when running with `swift test` 🤔

norio-nomura:
Tried to investigate with "Rerun job with SSH", but [it passed](https://circleci.com/gh/realm/SwiftLint/14347). 🤔

marcelofabri:
Thanks for looking into this, @norio-nomura! I’ll merge this for now and we can investigate the failure later if we see it again.

PR #2558, opened 2019-01-14
Remove workaround

kimdv:


jpsim:
The fix https://github.com/apple/swift-corelibs-foundation/pull/1471 will ship with Swift 5, so if you'd like you could add that compiler check for the workaround.

kimdv:
Hmm I think the best solution would be to wait and remove it when the project is move to Swift5.
Otherwise we will just add another `if` statement 😀

jpsim:
Historically we've supported building SwiftLint with multiple released Swift versions, so we'll need that `if` statement to build with both Swift 4.2 and Swift 5 at the same time 😄

PR #2004, opened 2018-01-13
Add `DuplicateImportsRule`

sammy-SC:
implements #1881 

Any recommendations for improvement are very welcome.

if you are happy with implementation, I will be happy to get started on autocorrection.

PR #2566, opened 2019-01-17
Add Duplicate Imports Rule

jpsim:
This is \#2004, but from a branch on this repo to work around https://github.com/danger/danger/issues/1034


jpsim:
Continuing in #2004.

PR #2567, opened 2019-01-18
Add unused_setter_argument rule

### LINKED ISSUE #1136: Rule request: newValue ignored on var set
I've seen a lot of times people using the variable name instead of `newValue` which leads to bugs:

```swift
var aValue: String {
    get {
        return Persister.shared.aValue
    }
    set {
        Perister.shared.aValue = aValue //should be newValue
    }
}
```

marcelofabri:
Fixes #1136

~Still WIP because it's not handling explicitly named arguments. Just wanted to see oss-check.~

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/UnusedSetterArgumentRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/UnusedSetterArgumentRule.swift
@@ -1,0 +1,182 @@
+import Foundation
+import SourceKittenFramework
+
+public struct UnusedSetterArgumentRule: ConfigurationProviderRule, AutomaticTestableRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "unused_setter_argument",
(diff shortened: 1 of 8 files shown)


jpsim:
How about `unused_setter_value` instead?

marcelofabri:
I don't have a strong preference, will rename it 👍

Source/SwiftLintFramework/Rules/Lint/UnusedSetterArgumentRule.swift
```diff
@@ -0,0 +1,234 @@
...
+        description: "Setter argument is not used.",
+        kind: .lint,
+        nonTriggeringExamples: [
+            """
+            var aValue: String {
+                get {
+                    return Persister.shared.aValue
+                }
+                set {
+                    Perister.shared.aValue = newValue
```
jpsim: Nit: `Perister` here and a few other places

marcelofabri: Copied from #1136 😅

PR #2568, opened 2019-01-18
Add `only_single_muted_parameter` configuration on `trailing_closure` rule

marcelofabri:
This configuration allows using `trailing_closure` without losing call site semantics, e.g. `filter(where:)`


jpsim:
Will this still trigger for functions with multiple arguments? e.g. `foo(bar: true, baz: {})`

Should we validate this in tests?

marcelofabri:
It won't, it should trigger only if the closure is the single argument to the function.

I believe the tests cover this, don't they? https://github.com/realm/SwiftLint/pull/2568/files#diff-a375b201cc727506363c295b93a0e54eR13 for example.

PR #2575, opened 2019-01-21
Unused private coding keys

### LINKED ISSUE #2573: [unused_private_declaration] False positives on CodingKeys
### New Issue Checklist

```bash
$ swiftlint lint
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.30.0

```yml
disabled_rules: # rule identifiers to exclude from running
  - trailing_whitespace
  - comma
  - line_length
  - todo
  - identifier_name

opt_in_rules: # some rules are only opt-in
#  - empty_count
  # Find all the available rules by running:
  # swiftlint rules
included: # paths to include during linting. `--path` is ignored if present.
  - AppName

  - Carthage

analyzer_rules:
  - unused_private_declaration
  - unused_import

type_body_length:
  - 300 # warning
  - 500 # error

file_length:
  warning: 500
  error: 1200

type_name:
  max_length: # warning and error
    warning: 50
    error: 60

identifier_name:
  max_length: # warning and error
    warning: 60
    error: 70
  excluded: # excluded via string array
    - id
    - db

```

```swift
// This triggers a violation:
struct ResponseModel: Codable {
    let items: [Item]

    private enum CodingKeys: String, CodingKey { // <-- False 'unused_private_declaration' 
        case items = "ResponseItems"
    }
}
```

kimdv:
Fixes #2573

@jpsim i gave it a shot! 
Will add Changelog when it is good to avoid merge conflicts 😄

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/UnusedPrivateDeclarationRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/UnusedPrivateDeclarationRule.swift
@@ -15,11 +15,35 @@ public struct UnusedPrivateDeclarationRule: AutomaticTestableRule, Configuration
             """
             private let kConstant = 0
             _ = kConstant
+            """,
+            """
+            struct ResponseModel: Codable {
+                let items: [Item]
+
+                private enum CodingKeys: String, CodingKey {
+                    case items = "ResponseItems"
+                }
+            }
+            """,
+            """
+            class ResponseModel {
+                @objc private func foo() {
+                }
+            }
             """
         ],
         triggeringExamples: [
             """
             private let ↓kConstant = 0
+            """,
+            """
+            struct ResponseModel: Codable {
+                let items: [Item]
+
+                private enum ↓CodingKeys: String {
+                    case items = "ResponseItems"
+                }
+            }
             """
         ],
         requiresFileOnDisk: true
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Rules/Lint/UnusedPrivateDeclarationRule.swift
```diff
@@ -98,6 +122,14 @@ private extension File {
             if cursorInfo["key.overrides"] != nil {
                 return nil
             }
+
+            // Skip CodingKeys as they are used
+            if kind == .enum,
+                cursorInfo.name == "CodingKeys",
+                let annotatedDecl = cursorInfo["key.annotated_decl"] as? String,
+                annotatedDecl.contains(">CodingKey<") {
```
kimdv: It was the "best" way to find if it did conform to a protocol.

jpsim: On my phone right now so not 100% sure but you should be able to use `key.inheritedtypes` or `key.inherited_types` for this. If you search the SwiftLint or SourceKitten codebases you’ll find usage examples.

jpsim: Ah so that won't work because we just have `cursorInfo` here. However, we can strengthen this check by making sure it's the right `CodingKey` protocol.

```swift
annotatedDecl.contains("usr=\"s:s9CodingKeyP\">CodingKey<") {
```

kimdv:
Thanks you very much for the last small fixes and changelog!

jpsim:
Thanks for not minding 😊

I also think I fixed Danger on CI when running on forked branches.

kimdv:
The analyzer is a powerful feature, and more power is in it! I'm looking forward to see what it can do! 🚀

PR #2581, opened 2019-01-23
Add `nslocalizedstring_key` opt-in rule

marcelofabri:


jpsim:
Can you link to docs that describe this behavior? The violations reported by OSSCheck have static string concatenation. Is that supported?

marcelofabri:
I was actually just double checking that. Concatenating strings seems to work, so I'll remove that case.

The main motivation for that rule was to avoid things like `"Hello, \(name)"`

Source/SwiftLintFramework/Rules/Lint/NSLocalizedStringKeyRule.swift
```diff
@@ -0,0 +1,44 @@
...
+                         kind: SwiftExpressionKind,
+                         dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
+        guard kind == .call,
+            dictionary.name == "NSLocalizedString",
+            let firstArgument = dictionary.enclosedArguments.first,
+            firstArgument.name == nil,
+            let offset = firstArgument.offset,
+            let length = firstArgument.length,
+            case let kinds = file.syntaxMap.kinds(inByteRange: NSRange(location: offset, length: length)),
+            Set(kinds) != [.string] else {
```
jpsim: Just curious what you think of `!kinds.allSatisfy({ $0 == .string })` instead? More efficient since it can return early instead of iterating through the whole sequence and doesn't have to compute the hash for each member of `kinds`.

jpsim: Agreed it shouldn't make much of a difference in this case.

PR #2586, opened 2019-01-24
Adds more fields to `custom_rules` including autocorrection capabilities

therealbnut:
I wanted to be able to tweak something like "OpeningBraceRule" without forking SwiftLint, this lets me do that in many more cases:

```yaml
  rule_test_brace:
    name: Opening Brace Rule
    message: "[Custom] Opening braces should be preceded by a single space and on the same line as the declaration."
    regex: "(?:[^( ]|[\\s(][\\s]+)\\{"
    exclude_kinds: comment_and_string_kinds
    exclude_regex: "(?:if|guard|while)\\n[^\\{]+?[\\s\\t\\n]\\{"
    severity: error
    correction: "$1 {"
```

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -59,6 +59,9 @@ opt_in_rules:
   - xct_specific_matcher
   - yoda_condition
 
+disabled_rules:
+  - opening_brace
+
 identifier_name:
   excluded:
     - id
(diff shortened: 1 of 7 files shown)


Dschee:
Note that this looks similar to https://github.com/realm/SwiftLint/pull/944. Not sure if it's a duplicate, but this maybe supersedes https://github.com/realm/SwiftLint/pull/944, so it should be closed once this is merged.

Having that said, this sounds interesting. @therealbnut Do you think watiting for the plugin system to complete will benefit this or could this be merged independently? Also we should make sure to check if this has high impact on custom rule performance as I'm not sure if the current tests test that.

PR #2591, opened 2019-01-25
Add `deployment_target` rule

### LINKED ISSUE #2589: [Rule Request] Deployment Target rule
### New Issue Checklist

### New rule request

This rule will require a configuration to work.

When bumping the deployment target of an app, it's common to leave unneeded code behind.

```swift
// would trigger
if #available(iOS 8.0, *) {            
}

@available(iOS 8.0, *) 
class A {}

// would not trigger
if #available(iOS 12.0, *) {            
}

@available(iOS 12.0, *) 
class A {}
```

Yes! `ios_deployment_target` and `macos_deployment_target`.

This could go either way, because the rule shouldn't trigger if `ios_deployment_target` and `macos_deployment_target` are not set.

marcelofabri:
Fixes #2589 

This is a pretty useless rule unless people configure it properly. Wonder how much that will hurt its adoption.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,15 @@
 
 #### Enhancements
 
-* None.
+* Add `deployment_target` rule to validate that `@availability` attributes and
+  `#available` conditions are not using a version that is satisfied by the
+  deployment target. Since SwiftLint can't read an Xcode project, you need to
+  configure this rule with these keys: `iOS_deployment_target`,
+  `macOS_deployment_target`, `watchOS_deployment_target` and
+  `tvOS_deployment_target`. By default, these values are configured with the
+  minimum versions supported by Swift.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#2589](https://github.com/realm/SwiftLint/issues/2589)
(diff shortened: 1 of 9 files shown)


marcelofabri:
> Would be nice if it could adhere to the `MACOSX_DEPLOYMENT_TARGET` and `IPHONEOS_DEPLOYMENT_TARGET` environment variables. That way I wouldn't have to manually duplicate that information into the `.swiftlint.yml`.

This is a good idea! However, that implies that `swiftlint` will only have the correct version if it's run during a build. Since we support using env variables on `.swiftlint.yml`, I'd rather have people to opt-in into this behavior than having the confusing behavior of getting different results when running as a build script x running from the command line (or from other tools like `danger`, `fastlane` and other IDEs/editors plugins)

jpsim:
Just a heads up I probably won't have time to review this. Might want to ask other contributors or forego code review on this one.

marcelofabri:
No worries, @jpsim!

Merging this now as it's basically an opt-in rule at this point. We can always revisit this later if you or other contributors have any comments or suggestions.

PR #2597, opened 2019-01-26
Add `weak_computed_property` rule

### LINKED ISSUE #2596: [Rule Request] Weak Computed Property
### New Issue Checklist

### New rule request

`weak` on a computed property doesn't do anything.

There [are](https://forums.swift.org/t/can-we-forbid-weak-keyword-for-computed-properties/19811) [discussions](https://github.com/apple/swift-evolution/pull/528) about adding a warning on the compiler directly.

```swift
class A {
    var b1: B { return B() } // ok
    weak var b2: B? // ok

    weak var b3: B? { return nil } // warning
    weak var b4: B? {  // warning
        get { return b3 }
        set { b3 = newValue }
    }
}
```

Just severity

I think it should be enabled by default. The only reasonable case for allowing this is to document the behavior, but since it's really not enforced by the compiler I'm not sure if it's worth the possible confusion.

marcelofabri:
Fixes #2596

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2589](https://github.com/realm/SwiftLint/issues/2589)
 
+* Add `weak_computed_property` rule to warn against using `weak` in a computed
+  property as it has no effect.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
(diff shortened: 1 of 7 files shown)


marcelofabri:
> We really ought to simplify the boilerplate for corrections since the implementation is nearly always the same or very close to this pattern.

Totally agreed. I will try to come up with something when I have some free time.

PR #2593, opened 2019-01-25
Make `redundant_objc_attribute` rule autocorrectable. 

dirtydanee:
Hey all! Thanks for the great efforts on keeping the project nice and sound! With this PR I am aiming to add autocorrection to an already existing rule.

Main changes:
* Moved the violation point from the variables offset to the `@objc` attribute startIndex. I think it kinda makes sense, since the violations warns against the redundant attribute, there is nothing to fix on the variable itself.
* Searching for ranges of violations initially, and converting later only to `StyleViolation`-s
* Making the rule conforming to `CorrectableRule`. Kinda straightforward, the only extra is that I am counting the attribute's trailing whitespace and newline characters and removing with the attribute. I am doing so since I would like to keep the format of the code.

PR #2598, opened 2019-01-27
Add SubstitutionCorrectableRule and SubstitutionCorrectableASTRule

marcelofabri:
This adds two new protocols that rules can adopt `SubstitutionCorrectableRule` and `SubstitutionCorrectableASTRule`.

This should make implementing corrections easier in most cases while keeping the flexibility of implementing `CorrectableRule` directly if needed.

I've updated 3 rules that I think are representative of the typical use cases to implement the new protocols.

### DIFF
--- a/Source/SwiftLintFramework/Protocols/Rule.swift
+++ b/Source/SwiftLintFramework/Protocols/Rule.swift
@@ -47,6 +47,61 @@ public extension CorrectableRule {
     }
 }
 
+public protocol SubstitutionCorrectableRule: CorrectableRule {
+    func violationRanges(in file: File) -> [NSRange]
+    func substitution(for violationRange: NSRange, in file: File) -> (NSRange, String)
+}
+
+public extension SubstitutionCorrectableRule {
+    func correct(file: File) -> [Correction] {
+        let violatingRanges = file.ruleEnabled(violatingRanges: violationRanges(in: file), for: self)
+        guard !violatingRanges.isEmpty else { return [] }
+
+        let description = type(of: self).description
+        var corrections = [Correction]()
+        var contents = file.contents
+        for range in violatingRanges {
+            let contentsNSString = contents.bridge()
+
+            let (rangeToRemove, substitution) = self.substitution(for: range, in: file)
+            contents = contentsNSString.replacingCharacters(in: rangeToRemove, with: substitution)
+            let location = Location(file: file, characterOffset: range.location)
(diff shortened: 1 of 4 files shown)


Source/SwiftLintFramework/Protocols/Rule.swift
```diff
@@ -47,6 +47,61 @@ public extension CorrectableRule {
...
+
+public extension SubstitutionCorrectableRule {
+    func correct(file: File) -> [Correction] {
+        let violatingRanges = file.ruleEnabled(violatingRanges: violationRanges(in: file), for: self)
+        guard !violatingRanges.isEmpty else { return [] }
+
+        let description = type(of: self).description
+        var corrections = [Correction]()
+        var contents = file.contents
+        for range in violatingRanges {
```
marcelofabri: I wonder if we should sort here just to be sure it's always correct

marcelofabri:
@dirtydanee would love to know your thoughts on this and if this would've made your life easier when implementing #2593

Source/SwiftLintFramework/Protocols/Rule.swift
jpsim: Yes we should always sort. Applying corrections, especially multiple ones in a file, is so rare in practice compared to the amount of code being linted that a slight (probably immeasurable) perf hit is worth the wins in correctness, maintainability and readability.

jpsim:
Should we migrate other correctable rules? Seems like `ExplicitInitRule` is a good candidate. There are probably a lot more too.

Source/SwiftLintFramework/Protocols/Rule.swift
dirtydanee: Would be a good move to move the sorting here, since you could be overwritting `violationRanges` function, and forgot the sorting. Place for great frustration upon trying to figure out why your corrections are all messed up 🤯

marcelofabri:
> Should we migrate other correctable rules? Seems like `ExplicitInitRule` is a good candidate. There are probably a lot more too.

Definetely! I wanted to first just introduce/validate the pattern and then migrate others separately.

dirtydanee:
@marcelofabri Its a great addition! All the boiler plate code one must write just to make the rule correctable would be all removed. Once merged, I can rewrite my PR using this protocol.

I would also offer a hand in migrating the already existing rules, if needed 🤓

PR #2615, opened 2019-01-31
Add SPM instructions to README

twof:
Introduced the instructions needed to use SwiftLint with SPM. Note that this doesn't cover setting up the build phase in generated Xcode projects, so users will need to set it up manually. This is a problem that will be common in projects that use SPM, because it's common for projects that use SPM to not ship with an Xcode project due to how easy xcprojes are to generate.

Hopefully this problem will be resolved in a future proposal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,17 @@ Note that this will add the SwiftLint binaries, its dependencies' binaries and t
 library distribution to the `Pods/` directory, so checking in this directory to SCM such as
 git is discouraged.
 
+### Using Swift Package Manager:
+
+Add the following line to your `Package.swift`:
+
+```swift
+dependencies: [
+    // ...
+    .package(url: "https://github.com/realm/SwiftLint.git"),
+],
+```
+
 ### Using [Mint](https://github.com/yonaskolb/mint):
 ```
 $ mint install realm/SwiftLint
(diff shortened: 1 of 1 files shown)


Dschee:
@twof I'm not sure if I understand this correctly. Are you aware that [Mint](https://github.com/yonaskolb/Mint) basically is SwiftPM with extra love for executables and that SwiftLint already supports installation via Mint? Or to put it differently: Who might profit from your additional instructions?

CypherPoet:
@Dschee I think this would benefit people who A) didn't know about `Mint` and B) are using SPM regardless.

Anecdotally, I stumbled upon this PR because I was wondering how to use `SwiftLint` with SPM.

Dschee:
@twof Is this still up-to-date now that SPM is integrated into Xcode 11?

twof:
It's not, but I'll try and updated later today.

maxchuquimia:
The example of the Run Script phase in this PR isn't working for me (Xcode 11.4).
What I did was add the SwiftLint packages to My Project > Swift Packages as per usual but didn't add it to any targets.

To work around the executable not appearing under `DWARF_DSYM_FOLDER_PATH` I added `swiftlint (SwiftLint)` to the Dependencies list at the top of Build Phases and then updated my Run Scripts to be:
```
"${SYMROOT}/${CONFIGURATION}/swiftlint"
```

This results in `swiftlint` being run correctly - however now when I archive I am seeing `swiftlint.dSYM` in my dSYMS directory... which makes me nervous, so I think I'm going to switch to a manual installation now.

Has anyone figured out updated steps on how to get it working and ensure the final project builds are untouched?

twof:
I don't have the time to update this. Someone else can feel free to steal it.

Issue #2620, opened 2019-01-31
Rule Request: Deinit Required


BenStaveleyTaylor:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

I would like to have the ability to require classes to have a `deinit` method.

1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.

When investigating memory leaks it is useful to be able to set a breakpoint on the `deinit` method. Our house style is that all classes should have a deinit even if it is empty, to facilitate debugging. 
Frequently the class does not have a method, so you have to quit, add the method and rebuild in order to proceed. 

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.
```swift
// Would trigger
class Apple {
}

class Banana
{
   func peel() {
   }
}

class Cherry: NSObject {
}

class Damson: Protocol1, Protocol2 {
}

// Would not trigger
class Apple {
   deinit {
   }
}

class Banana
{
   func peel() {
   }  
   deinit {
      print("Deinit of Banana: \(self)")
   }
}

struct Cherry {
}

enum Damson {
}

protocol Elderberry {
}
```
3. Should the rule be configurable, if so what parameters should be configurable?
No, I don't think anything beyond the standard error/warning, enable/disable is needed.

4. Should the rule be opt-in or enabled by default? Why?
opt-in. Clearly not everybody wants this all the time.

BenStaveleyTaylor:
@marcelofabri I'd be prepared to have a go at this. Before I invest time in it, is this a rule that would be accepted if it worked as described? If the feedback is that people don't want it, that's OK.

marcelofabri:
Implemented in #2624.

PR #2624, opened 2019-02-02
Add Deinit Required rule

BenStaveleyTaylor:
Classes are required to have a deinit method.

As proposed in https://github.com/realm/SwiftLint/issues/2620

This is a style to help debugging memory issues, when it is common to want to set a breakpoint at the point of deallocation. Often classes don't have a deinit, so the developer ends up having to quit, add a deinit and rebuild to proceed. If all classes have a deinit, debugging is much smoother.


Source/SwiftLintFramework/Rules/Style/DeinitRequiredRule.swift
```diff
@@ -0,0 +1,76 @@
+import SourceKittenFramework
+
+public struct DeinitRequiredRule: OptInRule, ConfigurationProviderRule, AutomaticTestableRule {
```
marcelofabri: what do you think about renaming it to `RequiredDeinitRule`? It'd be more in line with the existing `RequiredEnumCaseRule`.

Source/SwiftLintFramework/Rules/Style/DeinitRequiredRule.swift
```diff
@@ -0,0 +1,76 @@
...
+            }
+            """
+        ]
+    )
+
+    public init() {}
+
+    public func validate(file: File) -> [StyleViolation] {
+        let classCollector = NamespaceCollector(dictionary: file.structure.dictionary)
+        let classes = classCollector.findAllElements(of: [.class])
```
marcelofabri: nitpick: you can implement `ASTRule` instead of using `NamespaceCollector`

marcelofabri:
> This cuts out the classes NamespaceCollector but I still use one for the method instances unless there's a better way of listing all methods in a class?

It's great now, thanks!

> I confess I was unsure whether this was an Style, Idiomatic or Lint rule. I hope Style is OK.

Good point. IMO `style` is more for just code formatting. I'd use `lint` in this case. What do you think?

BenStaveleyTaylor:
> Good point. IMO `style` is more for just code formatting. I'd use `lint` in this case. What do you think?

Sure, I've made the change. I couldn't find any guidelines for when to use the different kinds -- is this documented somewhere? It might be nice to cover it in Contributing.md? To be brutally honest, I don't think the kind adds much value to a rule definition and I'd rather it went away :-) ...

marcelofabri:
> I couldn't find any guidelines for when to use the different kinds -- is this documented somewhere? 

Unfortunately, no. I'll try to write something.

>  To be brutally honest, I don't think the kind adds much value to a rule definition and I'd rather it went away

The hope is that it'd be more useful in the future. It could be used to group documentation (as Rules.md is already quite long) or even enable/disable a group of rules on a configuration.

Right now it serves to group rules in Xcode and it's also shown in Rules.md.

PR #2627, opened 2019-02-03
Throw error when YAML contains duplicated keys

### LINKED ISSUE #2191: SwiftLint doesn't complain when rule is configured twice
### New Issue Checklist

### Bug Report
Steps to reproduce:
1. Create a new project (e.g. `swift package init`)
2. Add a `.swiftlint.yml` with the following contents:
```yaml
identifier_name:
  severity: error
identifier_name:
  severity: warning
```
3. Run `swiftlint`

Expected outcome: SwiftLint complains that the rule is configured twice (in this case, even in an incompatible way)
Effective outcome: No warning is raised and apparently the first configuration is used (not sure if this is always the case).
Why this could be a problem: I encountered a duplicate rule configuration in one of our `.swiftlint.yml`. This had gone unnoticed for quite a while, so the rule didn't do what it was supposed to.

Interestingly, if either of the rule configurations are invalid (e.g. `identifier_name: foo`), SwiftLint will complain, suggesting that both configurations are actually being parsed.

### Environment

* SwiftLint version

I tested this both on the released v0.25.1 and the current master

Homebrew for the released version and from source for the current master

see above

no

This just returns `/Applications/Xcode.app/Contents/Developer` for me.

marcelofabri:
Fixes #2191

### DIFF
--- a/Source/SwiftLintFramework/Models/YamlParser.swift
+++ b/Source/SwiftLintFramework/Models/YamlParser.swift
@@ -13,10 +13,41 @@ public struct YamlParser {
     public static func parse(_ yaml: String,
                              env: [String: String] = ProcessInfo.processInfo.environment) throws -> [String: Any] {
         do {
-            return try Yams.load(yaml: yaml, .default,
-                                 .swiftlintContructor(env: env)) as? [String: Any] ?? [:]
+            let parser = try Parser(yaml: yaml, resolver: .default, constructor: .swiftlintContructor(env: env))
+            guard let node = try parser.singleRoot() else {
+                return [:]
+            }
+
+            try node.validateDuplicates(keyPath: nil)
+
+            return node.any as? [String: Any] ?? [:]
+        } catch let error as YamlParserError {
+            throw error
         } catch {
             throw YamlParserError.yamlParsing("\(error)")
+        }
+    }
+}
+
+private extension Node {
+    func validateDuplicates(keyPath: String?) throws {
+        var visitedKeys: Set<String> = []
+        for (key, value) in mapping ?? [:] {
+            guard let key = key.string else {
+                continue
+            }
+
+            let (inserted, _) = visitedKeys.insert(key)
+            if !inserted {
+                var message = "Duplicated keys found: '\(key)'."
+                if let keyPath = keyPath {
+                    message += " Found inside keypath '\(keyPath)'."
+                }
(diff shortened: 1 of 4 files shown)


Source/SwiftLintFramework/Models/YamlParser.swift
```diff
@@ -13,14 +13,45 @@ public struct YamlParser {
...
+                continue
+            }
+
+            let (inserted, _) = visitedKeys.insert(key)
+            if !inserted {
+                var message = "Duplicated keys found: '\(key)'."
+                if let keyPath = keyPath {
+                    message += " Found inside keypath '\(keyPath)'."
+                }
+                throw YamlParserError.yamlParsing(message)
```
marcelofabri: this could be a `queuedPrintError` but given that 99% of people run `swiftlint` from Xcode, they'd never see it.

Source/SwiftLintFramework/Models/YamlParser.swift
```diff
@@ -13,14 +13,45 @@ public struct YamlParser {
...
+            throw error
         } catch {
             throw YamlParserError.yamlParsing("\(error)")
         }
     }
 }
 
+private extension Node {
+    func validateDuplicates(keyPath: String?) throws {
+        var visitedKeys: Set<String> = []
```
norio-nomura: Since `Node` is `Hashable`, it could be `Set<Node>`.

Source/SwiftLintFramework/Models/YamlParser.swift
```diff
@@ -13,14 +13,45 @@ public struct YamlParser {
...
+        for (key, value) in mapping ?? [:] {
+            guard let key = key.string else {
+                continue
+            }
+
+            let (inserted, _) = visitedKeys.insert(key)
+            if !inserted {
+                var message = "Duplicated keys found: '\(key)'."
```
norio-nomura: `Node` has location information as `marker` property.
So, it can generate error message with location and ` error:` keyword  that is recognized by Xcode:
```
                var message = "\(key.mark?.line ?? 0):\(key.mark?.column ?? 0): error: Duplicated keys found: '\(key.string!)'."
```


Yams is created to generate error messages that Xcode will recognize by obeying the above manners.

jpsim:
It'd be great to push this to the finish line by addressing @norio-nomura's feedback.

PR #2633, opened 2019-02-11
Colon rule now catches violations on generic type declarations

### LINKED ISSUE #2628: Colon rule doesn't show warning for generic types
### New Issue Checklist

### Describe the bug

'Colon' rule does not trigger for class and struct definitions if the type is generic.

Input:

```swift
import Foundation

print("Hello, World!")

protocol MyProtocol { }

class Baz<String> : MyProtocol { }
struct Qux<String> : MyProtocol { }
```

Command Line:

```bash
$ swiftlint lint
Linting Swift files at paths
Linting 'main.swift' (1/1)
```

### Environment

    * 0.30.1
    * Homebrew
    * N/A

    * No
    * Xcode 10.1
* Do you have a sample that shows the issue? Run `echo "[string here]" | swiftlint lint --no-cache --use-stdin --enable-all-rules` to quickly test if your example is really demonstrating the issue. If your example is more complex, you can use `swiftlint lint --path [file here] --no-cache --enable-all-rules`.

```swift
// These trigger a warning as expected:
class Foo : MyProtocol { }
struct Bar : MyProtocol { }

// These do not:
class Baz<String> : MyProtocol { }
struct Qux<String> : MyProtocol { }
```

marcelofabri:
Fixes #2628


marcelofabri:
Not merging right now because it now doesn't warn for both violations if you have `class A<T : B> : C`.

Not such a big deal, but it'd be nice to fix it.

PR #2639, opened 2019-02-12
Fix false positives on rule `unused_closure_parameter` with `self` arguments

### LINKED ISSUE #2437: False positive on unused_closure_parameter rule with self
Using SwiftLint 0.27.0 I get the following false positive:

```swift
// setup view model bindings
viewModel?.profileImage.didSet(weak: self) { (self, profileImage) in // << `self` in a closure should be replaced with _
    if let profileImage = profileImage {
        self.profileImageView.image = profileImage
    }
    else {
        self.profileImageView.image = UIImage(named: "avatar_100pt")
    }
}
```

The underlying implementation looks like this:
```
import Foundation

/// A binding wrapper to track changes on constant properties.
class ObservableProperty<ValueType> {
    // MARK: - Stored Instance Properties
    private var didSetClosures: [(ValueType) -> Void] = []

    private(set) var value: ValueType {
        didSet {
            didSetClosures.forEach { $0(value) }
        }
    }

    // MARK: - Initializers
    init(_ value: ValueType) {
        self.value = value
    }

    // MARK: - Instance Methods
    /// Will be called after the wrapped value is changed.
    /// `$0` is the passed parameter (usually `self`), `$1` is the new value.
    func didSet<WeakObject: AnyObject>(weak object: WeakObject, _ closure: @escaping (WeakObject, ValueType) -> Void) {
        let weakClosure: (ValueType) -> Void = { [weak object] value in
            guard let object = object else { return }
            closure(object, value)
        }

        didSetClosures.append(weakClosure)
        weakClosure(value)
    }

    /// Use to change the value wrapped as a variable.
    func set(_ newValue: ValueType) {
        value = newValue
    }
}
```

This is an implementation of the idea also implemented in https://github.com/dreymonde/Delegated and should therefore be warning-free. My best guess is that there's some kind of exception logic going on in the rule logic for `self` references.

Dschee:
Fixes #2437.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/UnusedClosureParameterRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/UnusedClosureParameterRule.swift
@@ -44,6 +44,11 @@ public struct UnusedClosureParameterRule: SubstitutionCorrectableASTRule, Config
             withPostSideEffect { input in
                 if true { print("\\(input)") }
             }
+            """,
+            """
+            viewModel?.profileImage.didSet(weak: self) { (self, profileImage) in
+                self.profileImageView.image = profileImage
+            }
             """
         ],
         triggeringExamples: [
--- a/Source/SwiftLintFramework/Rules/Lint/UnusedClosureParameterRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/UnusedClosureParameterRule.swift
@@ -160,7 +160,8 @@ public struct UnusedClosureParameterRule: SubstitutionCorrectableASTRule, Config
                 }
 
                 let token = tokens.first(where: { token -> Bool in
-                    return (SyntaxKind(rawValue: token.type) == .identifier || name == "self") &&
+                    return (SyntaxKind(rawValue: token.type) == .identifier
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Rules/Lint/UnusedClosureParameterRule.swift
```diff
@@ -150,7 +160,7 @@ public struct UnusedClosureParameterRule: SubstitutionCorrectableASTRule, Config
                 }
 
                 let token = tokens.first(where: { token -> Bool in
-                    return SyntaxKind(rawValue: token.type) == .identifier &&
+                    return (SyntaxKind(rawValue: token.type) == .identifier || name == "self") &&
```
marcelofabri: you should check if `SyntaxKind(rawValue: token.type) == .keyword` in this case to avoid any potential issues

Dschee: Done.

marcelofabri:
@Dschee feel free to merge this after the conflict is resolved

PR #2641, opened 2019-02-12
Fixes false positives on `explicit_type_interface` with type reference assignments

### LINKED ISSUE #2636: False positive in rule explicit_type_interface with allowRedundancy
For this code

```swift
private let l10n = L10n.Communication.self
```

using this configuration

```yaml
explicit_type_interface:
  allow_redundancy: true
  excluded:
    - local
```

there shouldn't be any warnings, but there is.

Rationale: The usage of this rule here is mostly for improving readability, not performance of the compiler. We also try to prevent redundant type annotations, why explicitly writing the type here wouldn't be right. Thus this rules behavior must be changed.

Dschee:
Fixes #2636.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Idiomatic/ExplicitTypeInterfaceRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/ExplicitTypeInterfaceRule.swift
@@ -92,7 +92,9 @@ public struct ExplicitTypeInterfaceRule: OptInRule, ConfigurationProviderRule {
         guard configuration.allowedKinds.contains(kind),
             let offset = dictionary.offset,
             !dictionary.containsType,
-            (!configuration.allowRedundancy || !dictionary.isInitCall(file: file)),
+            (!configuration.allowRedundancy ||
+                (!dictionary.isInitCall(file: file) && !dictionary.isTypeReferenceAssignment(file: file))
+            ),
             !parentStructure.contains([.forEach, .guard]),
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Rules/Idiomatic/ExplicitTypeInterfaceRule.swift
```diff
@@ -130,6 +132,23 @@ private extension Dictionary where Key == String, Value == SourceKitRepresentabl
...
+            let nameLength = nameLength,
+            case let contents = file.contents.bridge(),
+            let afterNameRange = contents.byteRangeToNSRange(start: nameOffset + nameLength, length: 0)
+        else {
+            return false
+        }
+
+        let contentAfterName = contents.substring(from: afterNameRange.location)
+        let typeAssignment =
+            regex("^\\s*=\\s*(?:try[!?]?\\s+)?\\[?\\p{Lu}[^\\(\\s<]*(?:<[^\\>]*>)?(?::\\s*[^\\(\\n]+)?\\]?self")
```
marcelofabri: Can this be simplified? 😲
For example, we don't need to worry about `try` in this case.

Also, shouldn't it enforce a `.` before `self`?

Dschee: I simplified it and made sure there's a `.` before the `self`.

PR #2642, opened 2019-02-12
Fix false positive on `number_separator` with year integer literals

### LINKED ISSUE #2637: False positive on rule `number_separator`
Having the rule `number_separator` turned on for large numbers, years should still be possible to be written without those separators. For example, the following code should not trigger the rule:

```Swift
HolidayCalendar.Day(year: 2019, month: 12, day: 25)
```

Rationale: Writing year literals with a separator like `2_019` actually makes them less readable and should be made an exception for. Of course, it could also be the number `2_019` as a random case but developers should be allowed to give the number the look of a year if it is a year.

Dschee:
Fixes #2637.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,8 +45,8 @@
 * Warn if a configured rule is not enabled.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1350](https://github.com/realm/SwiftLint/issues/1350)
-  
-* Add `valid_ranges` option to `number_separator` for exclusion.  
+
+* Add `exclude_ranges` option to `number_separator` for exclusion.  
   [Cihat Gündüz](https://github.com/Dschee)
   [#2637](https://github.com/realm/SwiftLint/issues/2637)
 
@@ -81,9 +81,6 @@
 * Fix `unused_closure_parameter` when argument is named `self`.  
   [Cihat Gündüz](https://github.com/Dschee)
   [#2437](https://github.com/realm/SwiftLint/issues/2437)
-* Fix `number_separator` with year integer literals.  
-  [Cihat Gündüz](https://github.com/Dschee)
-  [#2637](https://github.com/realm/SwiftLint/issues/2637)
 
 ## 0.30.1: Localized Stain Remover
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,9 @@
 * Fix `unused_closure_parameter` when argument is named `self`.  
   [Cihat Gündüz](https://github.com/Dschee)
   [#2437](https://github.com/realm/SwiftLint/issues/2437)
+* Fix `number_separator` with year integer literals.  
+  [Cihat Gündüz](https://github.com/Dschee)
+  [#2637](https://github.com/realm/SwiftLint/issues/2637)
 
 ## 0.30.1: Localized Stain Remover
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,10 @@
 * Warn if a configured rule is not enabled.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1350](https://github.com/realm/SwiftLint/issues/1350)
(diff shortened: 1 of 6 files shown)


Source/SwiftLintFramework/Rules/Style/NumberSeparatorRule.swift
```diff
@@ -108,6 +115,14 @@ public struct NumberSeparatorRule: OptInRule, CorrectableRule, ConfigurationProv
         return prefixes.filter { lowercased.hasPrefix($0) }.isEmpty
     }
 
+    private func isInValidRanges(number: String) -> Bool {
+        if let double = Double(number), configuration.validRanges.contains(where: { $0.contains(double) }) {
```
marcelofabri: I don't think `Double(number)` works if there's a `_`.

Tests/SwiftLintFrameworkTests/NumberSeparatorRuleTests.swift
```diff
@@ -58,4 +58,41 @@ class NumberSeparatorRuleTests: XCTestCase {
...
+        let nonTriggeringExamples = [
+            "let foo = 1950",
+            "let foo = 1985",
+            "let foo = 2020",
+            "let foo = 2.10042",
+            "let foo = 2.833333"
+        ]
+        let triggeringExamples = [
+            "let foo = ↓1000",
+            "let foo = ↓2100",
```
marcelofabri: it'd be good to add an example here with a number in the valid range formatted with `_`

Source/SwiftLintFramework/Rules/RuleConfigurations/NumberSeparatorConfiguration.swift
```diff
@@ -33,6 +35,13 @@ public struct NumberSeparatorConfiguration: RuleConfiguration, Equatable {
             self.minimumFractionLength = minimumFractionLength
         }
 
+        if let validRanges = configuration["valid_ranges"] as? [[String: Any]] {
```
marcelofabri: I know that I kind of suggested the name, but maybe we can think of about a better name for this? `no_separator_ranges` maybe?

Source/SwiftLintFramework/Rules/Style/NumberSeparatorRule.swift
Dschee: Fixed in a8d06ef162f79bb444e473b9720f3f9708bd91d7.

Tests/SwiftLintFrameworkTests/NumberSeparatorRuleTests.swift
Dschee: Fixed.

Source/SwiftLintFramework/Rules/RuleConfigurations/NumberSeparatorConfiguration.swift
Dschee: I don't think `no_separator_ranges` explains it well, as it suggests that no separators can be used for the values in the specified ranges. The option was meant to allow users both, separators or no separators depending on the semantic meaning of the number. For examples `2_048` could be used for 2^11 but `2048` could be used for the year. Both should be possible and what the range basically does is turning off the rule for the given ranges. So I'm proposing `exclude_ranges` instead. I've changed the code accordingly.

marcelofabri: ok, I had misunderstood the change before. This makes sense.

CHANGELOG.md
```diff
@@ -45,6 +45,10 @@
 * Warn if a configured rule is not enabled.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#1350](https://github.com/realm/SwiftLint/issues/1350)
+  
+* Add `valid_ranges` option to `number_separator` for exclusion.
```
marcelofabri: can you update the configuration name here?

PR #2644, opened 2019-02-13
Add more configuration options to OpeningBrace

### LINKED ISSUE #2635: Add more configuration options to OpeningBraceRule
`OpeningBraceRule` currently excludes the cases where `if` `guard` and `while` statements are spread over multiple lines. Example:

```swift
if let x = someX { // This is OK
}

if let x = someX 
{ // This violates the rule
}

if 
   let x = someX,
   let y = someY
{ // This is also OK because multiline `if` is excluded
}
```
**I think it would be worth it to make the excluded keywords configurable.**

In my case, for example, I'd like to exclude multiline `func` from this rule, too, and there's currently no way to do that, without modifying the source code.

 The default configuration will be the same as the current hardcoded one, so the behavior of the rule won't change.

I'm OK with implementing this change myself. I just want to test the waters and see whether the community would be OK with this change.

VojtaStavik:
### This PR adds more configuration options to `opening_brace`

Changes:

- `OpeningBraceRule` now uses `OpeningBraceRuleConfiguration`. Using `first_line_excluding_regex `, we can configure the keywords which should be excluded when in a multiline statement. 

#### Example 1

In the default configuration, the following code triggers a violation:
```swift
func abc(
   a: Int,
   b: Int
) -> Int
↓{ … }
```

When we add `func` into the `first_line_excluding_regex` configuration, the violation is gone:
```swift
["first_line_excluding_regex": "if|guard|while|func \\w+\\("]
/ ... /

func abc(
   a: Int,
   b: Int
) -> Int
{ … }
```

#### Example 2

The default configuration allows multiline `if` statements:
```swift
if
   let a = b
   let c = d
   where a == c
{ … }
```

If we remove `if` from `first_line_excluding_regex`, the same code now triggers a violation:
```swift
["first_line_excluding_regex": "guard|while"]
/ ... /

if
   let a = b
   let c = d
   where a == c
↓{ … }
```

---

[Closes #2635]

### DIFF
--- a/Source/SwiftLintFramework/Rules/RuleConfigurations/OpeningBraceRuleConfiguration.swift
+++ b/Source/SwiftLintFramework/Rules/RuleConfigurations/OpeningBraceRuleConfiguration.swift
@@ -1,0 +1,32 @@
+public struct OpeningBraceRuleConfiguration: RuleConfiguration, Equatable {
+    private(set) var severityConfiguration: SeverityConfiguration
+
+    /// If this regex matches the first line of a statement, the potential rule
+    /// violations are ignored for the given statement.
+    private(set) var firstLineExcludingRegex: String
+
+    init(severity: ViolationSeverity, firstLineExcludingRegex: String) {
+        self.severityConfiguration = SeverityConfiguration(severity)
+        self.firstLineExcludingRegex = firstLineExcludingRegex
+    }
+
+    public var consoleDescription: String {
+        return severityConfiguration.consoleDescription +
+        ", first_line_excluding_regex: \(firstLineExcludingRegex)"
+    }
+
+    public mutating func apply(configuration: Any) throws {
+        let configurationDict = configuration as? [String: Any]
+
+        let severityString = configuration as? String
+            ?? configurationDict?["severity"] as? String
+
+        if let severityString = severityString {
+            try severityConfiguration.apply(configuration: severityString)
+        }
+
+        if let excludingRegex = configurationDict?["first_line_excluding_regex"] as? String {
+            firstLineExcludingRegex = excludingRegex
+        }
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@
 
 #### Enhancements
 
+* Add more configuration options to `opening_brace`. It's now possible to
+  configure the excluded keywords for multiline statements, previously
+  hardcoded to `if` `guard` and `while`.
+  [Vojta Stavik](https://github.com/VojtaStavik)
+  [#2635](https://github.com/realm/SwiftLint/issues/2635)
+
 * Add `deployment_target` rule to validate that `@availability` attributes and
   `#available` conditions are not using a version that is satisfied by the
   deployment target. Since SwiftLint can't read an Xcode project, you need to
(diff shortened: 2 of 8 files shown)


VojtaStavik:
I see the bug that causes the warnings. Will fix that soon.

VojtaStavik:
I think I will need some help with this 😅 

Our `match()` function ignores `excludingSyntaxKinds` when searching for `excludingPattern`. In the edge case below, the excluded pattern partly matches a string from a comment, which causes the violation to not be triggered.

Is this a bug or was it a deliberate decision to match `excludingPatter` over excluded syntax kinds?

screen shot 2019-02-14 at 14 38 05

marcelofabri:
@VojtaStavik I currently don't have so much time to look into your specific case, but there're methods on `File+SwiftLint` that return all syntax kinds, you can implement the filter manually if you need a different behavior.

VojtaStavik:
@marcelofabri Cool, will do 👍 Thank you!

VojtaStavik:
I had to change it a little bit from the original direction I went but I think I'm done. It should be ready for a review @marcelofabri.

Source/SwiftLintFramework/Rules/RuleConfigurations/OpeningBraceRuleConfiguration.swift
```diff
@@ -0,0 +1,32 @@
...
+    private(set) var firstLineExcludingRegex: String
+
+    init(severity: ViolationSeverity, firstLineExcludingRegex: String) {
+        self.severityConfiguration = SeverityConfiguration(severity)
+        self.firstLineExcludingRegex = firstLineExcludingRegex
+    }
+
+    public var consoleDescription: String {
+        return severityConfiguration.consoleDescription +
+        ", first_line_excluding_regex: \(firstLineExcludingRegex)"
```
Dschee: This should probably be indented.

marcelofabri:
@VojtaStavik while I agree with having more customization, I don't know if exposing this configuration with a regex is the right call. Can we try to abstract this to a more user-friendly configuration?

VojtaStavik:
@marcelofabri That was actually my original approach. It didn't really work out because the existing excluding pattern is very strict. It requires `if` `guard` `while` to be immediately followed by a new line:
```swift
if
  let a = b 
{ }
```

However, this simple rule doesn't work for every keyword. For example `func`:
```swift
func
  doSomething()
{ }
```
It means that a simple configurable list of keywords if not enough to exactly specify the behavior you want. You need to provide more context. 

That's why I ended up exposing the whole first line regex pattern. It actually makes the configuration options much more powerful. You can configure it to exactly match your needs:
```swift
// you might want to have the first line without any parameter
func abc(
  a: Int,
  b: Int
) { }

// or you can require the first parameter to be on the same line
func abc(a: Int,
         b: Int) 
{ }

// maybe, instead of this...
if
  let a = a(),
  let b = b()
{ }

// ... you like this better
if let a = a(),
       b = b()
{ }
```

I wasn't sure about exposing the regex either but it was the best solution I had found. Unfortunately, it makes this configuration a power-user feature, that's true.

VojtaStavik:
@marcelofabri So I guess we can close this PR then?

Issue #2444, opened 2018-10-22
Rule Request: Deprecation Count Rule


devandanger:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request
Deprecation Count rule would run like normal accept it would need to globally keep track of the number of violating deprecations.  This would be useful for general curation of the application.  If you have several deprecated custom class/struct you could prevent further re-usage of them.  

1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.
3. Should the rule be configurable, if so what parameters should be configurable? Configurable with counts allowed for certain usages of a deprecated class/struct.
4. Should the rule be opt-in or enabled by default? Why?
opt-in.  Its usage it more similar to the custom_rule.
   See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a rule as opt-in.

marcelofabri:
Closing this because the template is not filled completely. Feel free to reopen this once we have more details.

You can also [mark declarations as deprecated with Swift attributes](https://www.hackingwithswift.com/example-code/language/how-to-use-available-to-deprecate-old-apis) which is probably a better tool for the job.

devandanger:
@marcelofabri can you reopen this issue?... I don't have access to.

devandanger:
Also regarding.

> You can also [mark declarations as deprecated with Swift attributes](https://www.hackingwithswift.com/example-code/language/how-to-use-available-to-deprecate-old-apis) which is probably a better tool for the job.

Looking into that approach, only problem is we have warnings/deprecations turned on 'as Errors'.

marcelofabri:
It still doesn't make much sense to me to add this rule. One thing that bothers me is that which violations will be triggered is not deterministic: it depends on the order of the files that were checked.

It's also problematic because it introduce shared state (a singleton) between different files being checked - and we do it in parallel, so it can also create multithread issues.

If using `@availability` is not an option, I think it'd make more sense to use a custom rule and mark existing code with `// swiftlint:disable`.

devandanger:
> It still doesn't make much sense to me to add this rule. One thing that bothers me is that which violations will be triggered is not deterministic: it depends on the order of the files that were checked.

[...]

PR #2645, opened 2019-02-13
WIP - Custom threshold rule

devandanger:
This idea is related to the following rule request (2444)[https://github.com/realm/SwiftLint/issues/2444]
I don't think I'm able to re-open this as instructed by repo maintainer.

Couple caveats to this PR...
I wasn't familiar with how the custom_rule was really implemented prior to. I just thought I needed to base this rule feature on how the custom_rule was implemented.  Now I see with all the boilerplate code copied it might be a better consideration to just implement this feature around an optional threshold attribute into the custom_rule. 

As for why I think this idea is valuable I'm open to discussing in the previous issue or here in this PR. 

**Obviously this PR is not ready to be merged in fact I'm still testing (I don't think it actually works), but I wanted it at least for a vehicle of discussion though before I go too far with it.**

### DIFF
--- a/Rules.md
+++ b/Rules.md
@@ -20,6 +20,7 @@
 * [Control Statement](#control-statement)
 * [Convenience Type](#convenience-type)
 * [Custom Rules](#custom-rules)
+* [Custom Threshold Rules](#custom-threshold-rules)
 * [Cyclomatic Complexity](#cyclomatic-complexity)
 * [Deployment Target](#deployment-target)
 * [Discarded Notification Center Observer](#discarded-notification-center-observer)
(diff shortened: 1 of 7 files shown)


Dschee:
Apart from the fact that I seem not to understand the point of this feature, in case you want to reimplement it, feel free to have a look at https://github.com/realm/SwiftLint/pull/2586/files where you can find an example of how you could add this to the already existing `custom_rule` as a new option.

devandanger:
Closing this see  [2444](https://github.com/realm/SwiftLint/issues/2444) for more information.

PR #2647, opened 2019-02-14
Fix false positives on no_grouping_extension rule when using where clause

almazrafi:
The `no_grouping_extension` rule diagnoses extensions with `where` clause as a violation. For example:

```
	class Box<T> {}

	extension Box where T: FloatingPoint {}
```

Since the rule prohibits the use of extensions to group code ([No Grouping Extension](https://github.com/realm/SwiftLint/blob/master/Rules.md#no-grouping-extension)), it must ignore the extensions with `where` cause.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,9 @@
 
 #### Bug Fixes
 
+* Fix false positives on `no_grouping_extension` rule when using `where` clause.
+  [Almaz Ibragimov](https://github.com/almazrafi)
+
 * Fix `explicit_type_interface` when used in statements.  
   [Daniel Metzing](https://github.com/dirtydanee)
   [#2154](https://github.com/realm/SwiftLint/issues/2154)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,8 @@
 
 #### Bug Fixes
 
-* Fix false positives on `no_grouping_extension` rule when using `where` clause.
+* Fix false positives on `no_grouping_extension` rule when using `where`
+  clause.  
   [Almaz Ibragimov](https://github.com/almazrafi)
 
 * Fix `explicit_type_interface` when used in statements.  
--- a/Source/SwiftLintFramework/Rules/Idiomatic/NoGroupingExtensionRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/NoGroupingExtensionRule.swift
@@ -34,7 +34,7 @@ public struct NoGroupingExtensionRule: OptInRule, ConfigurationProviderRule, Aut
                 return nil
             }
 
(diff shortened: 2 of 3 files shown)


Source/SwiftLintFramework/Rules/Idiomatic/NoGroupingExtensionRule.swift
```diff
@@ -28,12 +29,37 @@ public struct NoGroupingExtensionRule: OptInRule, ConfigurationProviderRule, Aut
...
+        }
+
+        let rangeStart = nameOffset + nameLength
+        let rangeLength = bodyOffset - rangeStart
+
+        guard let range = contents.byteRangeToNSRange(start: rangeStart, length: rangeLength) else {
+            return false
+        }
+
+        return (regex(" where ").firstMatch(in: file.contents, options: [], range: range) != nil)
```
marcelofabri: just to be safe, you should check for the syntax kind:

```swift
return !file.match(pattern: "\\bwhere\\b", with: [.keyword], range: range).isEmpty
```

Source/SwiftLintFramework/Rules/Idiomatic/NoGroupingExtensionRule.swift
```diff
@@ -28,12 +29,37 @@ public struct NoGroupingExtensionRule: OptInRule, ConfigurationProviderRule, Aut
...
+                return nil
+            }
+
+            return StyleViolation(ruleDescription: type(of: self).description,
+                                  severity: configuration.severity,
+                                  location: Location(file: file, byteOffset: element.offset))
+        }
+    }
+
+    private func hasWhereClause(element: NamespaceCollector.Element, file: File) -> Bool {
```
marcelofabri: nitpick: I'd pass `[String: SourceKitRepresentable]` instead of `NamespaceCollector.Element` to avoid all the `.dictionary` calls

CHANGELOG.md
```diff
@@ -48,6 +48,9 @@
 
 #### Bug Fixes
 
+* Fix false positives on `no_grouping_extension` rule when using `where` clause.
```

Source/SwiftLintFramework/Rules/Idiomatic/NoGroupingExtensionRule.swift
almazrafi: Done

Source/SwiftLintFramework/Rules/Idiomatic/NoGroupingExtensionRule.swift
almazrafi: Done

CHANGELOG.md
almazrafi: Done. Thanks!

marcelofabri:
Thanks for your PR!

PR #2648, opened 2019-02-14
Fix bug where nested configuration excluded files are ignored

### LINKED ISSUE #2447: Excluded paths in nested configurations are ignored
### New Issue Checklist

### Describe the bug

My example project has this structure:

```
.
├── .swiftlint.yml
└── A
    ├── .swiftlint.yml
    ├── B
    │   └── Longcat.swift
    └── Hello.swift
```

`./swiftlint.yml` is empty, `./A/swiftlint.yml` has the following contents:

```yaml
excluded:
  - B
```

When I run `swiftlint` from the project's root directory, `Longcat.swift` gets linted although it should be excluded.

An explanation as to why I'm trying to do this: We have a common base `.swiftlint.yml` for our apps, which we symlink from a CocoaPod. Since we need to add project-specific excludes as well (and SwiftLint unfortunately doesn't seem to support shared base configurations at all), I need to mark individual folders as excluded from nested configurations.

```bash
$ swiftlint lint
Linting Swift files at paths
Linting 'Longcat.swift' (1/2)
Linting 'Hello.swift' (2/2)
.../swiftlint-example/A/B/Longcat.swift:1: warning: Line Length Violation: Line should be 120 characters or less: currently 189 characters (line_length)
Done linting! Found 1 violation, 0 serious in 2 files.
```

### Environment

`0.27.0`

In this instance, I was using the `swiftlint` from `brew install swiftlint`

See above.

Yes. Toplevel is empty, nested config is the one shown above.

`Version 10.0 (10A255)`

I've attached my minimal example project: [swiftlint-nested-exclude-example.zip](https://github.com/realm/SwiftLint/files/2526054/swiftlint-nested-exclude-example.zip)

Bruschidy54:
Fixes #2447

When SwiftLint is linting files in `visitLintableFiles` in `Configuration+CommandLine`, it:

1. Gathers all lintable files in `getFiles`. This is where the exclusion of files occurs based on the parent configuration's exclusion list.
2. These files are grouped by their specific configuration in `groupFiles`. This is where configurations for each available file are determined (and if nested configurations exist, merged). After these configurations are determined and the files are grouped accordingly, no more files are excluded from the lintable files list. Even though a file's configuration thinks it should be excluded, these files are not removed from the list of lintable files, generating the bug.
3. Finally, each file is visited by the linter.

My solution is to skip files whose merged configurations specify they should be excluded in step 2 or `groupFiles`. Therefore, they will not be visited in step 3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,8 +49,8 @@
 #### Bug Fixes
 
 * Fix bug where SwiftLint ignores excluded files list in a nested configuration file.  
-[Dylan Bruschi](https://github.com/Bruschidy54)
-[#2447](https://github.com/realm/SwiftLint/issues/2447)
+  [Dylan Bruschi](https://github.com/Bruschidy54)
+  [#2447](https://github.com/realm/SwiftLint/issues/2447)
 
 * Fix false positives on `no_grouping_extension` rule when using `where`
   clause.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,10 @@
 
 #### Bug Fixes
 
+* Fix bug where SwiftLint ignores excluded files list in a nested configuration file.  
+[Dylan Bruschi](https://github.com/Bruschidy54)
+[#2447](https://github.com/realm/SwiftLint/issues/2447)
+
 * Fix false positives on `no_grouping_extension` rule when using `where`
   clause.  
   [Almaz Ibragimov](https://github.com/almazrafi)
--- a/Source/swiftlint/Extensions/Configuration+CommandLine.swift
+++ b/Source/swiftlint/Extensions/Configuration+CommandLine.swift
@@ -61,7 +61,34 @@ extension Configuration {
             let errorMessage = "No lintable files found at paths: '\(visitor.paths.joined(separator: ", "))'"
             return .failure(.usageError(description: errorMessage))
         }
-        return .success(Dictionary(grouping: files, by: configuration(for:)))
+
+        var groupedFiles = [Configuration: [File]]()
+        for file in files {
+            // Files whose configuration specifies they should be excluded will be skipped
+            let fileConfiguration = configuration(for: file)
+            let excludedPaths = fileConfiguration.excluded
+                .compactMap { fileConfiguration.rootPath?.bridge().appendingPathComponent($0) }
+
+
+            var shouldSkip = false
+            for excludedPath in excludedPaths {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -48,6 +48,10 @@
 
 #### Bug Fixes
 
+* Fix bug where SwiftLint ignores excluded files list in a nested configuration file.  
+[Dylan Bruschi](https://github.com/Bruschidy54)
```
Dschee: Pleae make sure indentation is added here as in the other entries.

Source/swiftlint/Extensions/Configuration+CommandLine.swift
```diff
@@ -61,7 +61,33 @@ extension Configuration {
...
-        return .success(Dictionary(grouping: files, by: configuration(for:)))
+
+        var groupedFiles = [Configuration: [File]]()
+        for file in files {
+            // Files whose configuration specifies they should be excluded will be skipped
+            let fileConfiguration = configuration(for: file)
+            let excludedPaths = fileConfiguration.excluded
+                .compactMap { fileConfiguration.rootPath?.bridge().appendingPathComponent($0) }
+
+            var shouldSkip = false
```
Dschee: Can't the entire for loop be prevented with a much more readable `contains` or `reduce` here?

I'm thinking something like:

```swift
let shouldSkip: Bool = excludedPaths.contains { file.path?.bridge().pathComponents.starts(with: $0.bridge().pathComponents) }
```

Dschee:
Thanks for updating the code. I have one more question: Do you think you could add tests to prevent this change from breaking again in the future?

jpsim:
Thanks for the PR @Bruschidy54!

> Do you think you could add tests to prevent this change from breaking again in the future?

Tests would be great, although we don't currently have the ability to run unit tests against the `swiftlint` target, only `SwiftLintFramework`. I don't 100% recall why I set things up this way, but my initial intention was for the `swiftlint` target to be as small as possible, really just a shim around `SwiftLintFramework`. Evidently it grew over time to be much more than a shim.

Bruschidy54:
@Dschee @jpsim Is it possible to accept the PR without ```SwiftLintFramework``` tests for now? Work is really busy at the moment, so I won't have time to look at this in the near future.

jpsim:
Yes, should be fine. I have some small followups that I'll do in a separate PR to avoid force-pushing to your fork's `master` branch.

Thanks for the fix @Bruschidy54 and for your code review @Dschee!

jpsim:
Small followups here: https://github.com/realm/SwiftLint/pull/2707

Issue #2655, opened 2019-02-21
Rule Request: [disable_rationale]


acecilia:
### New Issue Checklist

- [ *] Updated SwiftLint to the latest version
- [* ] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

Please describe the rule idea, format
this issue's title as `Rule Request: [Rule Name]` and describe:

1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.

I would like to have an explanation of why a rule is disabled next to it

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

* would:

```
// swiftlint:disable missing_docs
```

* would not:
```
// This rule is disabled because of blablabla
// swiftlint:disable missing_docs
```

3. Should the rule be configurable, if so what parameters should be configurable?
No configuration required.

4. Should the rule be opt-in or enabled by default? Why?
   See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a rule as opt-in.
Opt-in.

### Custom rule implementation
At the same time I am proposing this rule, I tried to implement it as a custom rule. I did not managed, and I would appreciate some help. My main idea is:

(1) Find a line containing 'swiftlint:disable' ==> (2) Check that the previous line is a comment

I thought I could do this with a lookbehind regex. The closest I got was the following regex:

```
custom_rules:
  disabled_rules_rationale:
    regex: "(?<!\/\/[^\n]*$\n)^[^\n]*swiftlint:disable"
    match_kinds:
      - comment
    message: "All swiftlint disable comments should include an explanation on top"
```

In [https://regexr.com](regexr) the outcome is the desired one:

![screenshot 2019-02-21 at 16 25 08](https://user-images.githubusercontent.com/7938671/53180118-49a01b80-35f5-11e9-9b25-652b6cf12701.png)

But using the same regex in swiftlint does not result the same (note there are no errors):
![screenshot 2019-02-21 at 16 26 38](https://user-images.githubusercontent.com/7938671/53180244-7f450480-35f5-11e9-94f6-ae817347cd96.png)

Some final notes:
* I am using `[^\n]` instead of `.` in the regex because I did not manage to make `.` work in swiftlint.
[...]

PR #2660, opened 2019-02-25
Add DisableRationaleRule

acecilia:
Solves https://github.com/realm/SwiftLint/issues/2655

### DIFF
--- a/Source/SwiftLintFramework/Rules/Idiomatic/DisableRationaleRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/DisableRationaleRule.swift
@@ -1,7 +1,7 @@
 import Foundation
 import SourceKittenFramework
 
-public struct DisableCommentsRationaleRule: OptInRule, ConfigurationProviderRule {
+public struct DisableRationaleRule: OptInRule, ConfigurationProviderRule {
     public var configuration = SeverityConfiguration(.warning)
 
     public init() { }
@@ -42,10 +42,6 @@ public struct DisableCommentsRationaleRule: OptInRule, ConfigurationProviderRule
         let url = URL(string: "https://github.com")!
         // swiftlint:disable:previous force_unwrapping
         // Force unwrapping is used here, because initializing it with the github domain can not fail
-        """,
-        // For now, this is failing
-        """
-        let string = "let url = URL(string: "https://github.com")! // swiftlint:disable:this force_unwrapping"
         """
     ]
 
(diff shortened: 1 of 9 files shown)


Tests/SwiftLintFrameworkTests/RulesTests.swift
```diff
@@ -2,6 +2,15 @@ import SwiftLintFramework
 import XCTest
 
 class RulesTests: XCTestCase {
+    func testDisableRationale() {
+        verifyRule(
+            DisableRationaleRule.description,
+            commentDoesntViolate: false,
+            skipStringTests: true,
```
acecilia: I disabled the string tests because, even though I used `let disableMatches = file.match(pattern: "swiftlint:disable", with: [.comment])` (note the `.comment`), still violations are raised when executing the string tests. 
I have been looking into it but could not find the reason behind it, could it be a bug inside `let disableMatches = file.match(pattern: "swiftlint:disable", with: [.comment])`?

acecilia:
Also, I tried to implement the rule using `ASTRule` (this is my first time doing a Swiftlint rule), but seems to me that it is not possible to get comment information from the AST at the moment, right? Only detecting `SwiftDeclarationKind` and `SwiftExpressionKind` work.

Tests/SwiftLintFrameworkTests/RulesTests.swift
acecilia: Okey, I think the problem was that my code has double quote characters in it that where not being escaped when converting them to string literal. I think I fixed that, now string tests are also passing.

Source/SwiftLintFramework/Rules/Idiomatic/DisableRationaleRule.swift
```diff
@@ -0,0 +1,144 @@
...
+        name: "Disable Rationale",
+        description: "When disabling a rule, the rationale should be added as a comment in the previous or next line.",
+        kind: .idiomatic,
+        nonTriggeringExamples: nonTriggeringExamples,
+        triggeringExamples: triggeringExamples
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        let disableMatches = file.match(pattern: "swiftlint:disable", with: [.comment])
+        guard disableMatches.isEmpty == false else {
```
Dschee: The ` == false` should be removed here, use `!` instead.

acecilia:
@Dschee Thanks! About the (non-)triggering examples:

1. multiple rules are disabled in a single line: I will be adding them.
2. the disable comment is not done on it's own line but on a code line: this is already done. See [here](https://github.com/realm/SwiftLint/pull/2660/files#diff-b5d5ebb28e58c901be755c56894b349bR3080) and [here](https://github.com/realm/SwiftLint/pull/2660/files#diff-b5d5ebb28e58c901be755c56894b349bR3084) for non-triggering, and [here](https://github.com/realm/SwiftLint/pull/2660/files#diff-b5d5ebb28e58c901be755c56894b349bR3129) for triggering.

I will be addressing your comments about rule direction in the other issue.

PR #2663, opened 2019-03-01
Add NSObjectPreferIsEqualRule

matthew-healy:
After accidentally mis-implementing `==` on an `NSObject` subclass one too many times, I decided to implement this rule. I also filled out the Rule Request info below, to explain why I think it's necessary.

When subclassing `NSObject` in Swift, properly implementing `Equatable` can be a little tricky. Whereas with most Swift types the correct thing to do is implement `==`, `NSObject` already exposes an implementation of `==` that calls out to `isEqual`. Adding a new implementation of `==` to an `NSObject` subclass can lead to odd situations where `a == b` is `true`, but `(a as NSObject) == b` (or, equivalently `a.isEqual(b)`) is `false`.

This rule is triggered whenever an `@objc` class or a direct `NSObject` subclass implements a `static` `==` function with exactly 2 arguments, both of which are of the same type as the class itself.

See the triggering & non-triggering examples in `NSObjectPreferIsEqualRuleExamples`.

I included a `SeverityConfiguration`, because it seemed reasonable that folk might want to be aware of this violation without fixing it immediately.

I'm not entirely sure. I implemented it as an `OptInRule` because it seemed safer, but I also can't think of a particularly good reason why you wouldn't want this to run, given the relative risks of differing notions of equality existing on the same type depending on the context. Happy to either leave as-is or enable by default depending on what others think.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,11 @@
 
 #### Enhancements
 
-* None.
+* Add `nsobject_prefer_isequal` rule to warn against implementing `==` on an
+  `NSObject` subclass as calling `isEqual` (i.e. when using the class from
+  Objective-C) will will not use the defined `==` method.  
+  [Matthew Healy](https://github.com/matthew-healy)
+  [#2663](https://github.com/realm/SwiftLint/pull/2663)
 
 #### Bug Fixes
 
@@ -27,7 +31,6 @@
 * None.
 
 #### Enhancements
-
 * Add `deployment_target` rule to validate that `@availability` attributes and
   `#available` conditions are not using a version that is satisfied by the
   deployment target. Since SwiftLint can't read an Xcode project, you need to
--- a/Source/SwiftLintFramework/Rules/Lint/NSObjectPreferIsEqualRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/NSObjectPreferIsEqualRule.swift
@@ -1,0 +1,20 @@
+import SourceKittenFramework
+
+public struct NSObjectPreferIsEqualRule: Rule, ConfigurationProviderRule, AutomaticTestableRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "nsobject_prefer_isequal",
+        name: "NSObject Prefer isEqual",
+        description: "NSObject subclasses should implement isEqual instead of ==.",
+        kind: .lint,
+        nonTriggeringExamples: NSObjectPreferIsEqualRuleExamples.nonTriggeringExamples,
+        triggeringExamples: NSObjectPreferIsEqualRuleExamples.triggeringExamples
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        return []
+    }
+}
--- a/Source/SwiftLintFramework/Rules/Lint/NSObjectPreferIsEqualRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/NSObjectPreferIsEqualRule.swift
@@ -15,6 +15,53 @@ public struct NSObjectPreferIsEqualRule: Rule, ConfigurationProviderRule, Automa
     )
 
     public func validate(file: File) -> [StyleViolation] {
-        return []
+        return objcVisibleClasses(in: file).flatMap { violations(in: file, for: $0) }
+    }
+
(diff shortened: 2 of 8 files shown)


Source/SwiftLintFramework/Rules/Lint/NSObjectPreferIsEqualRuleExamples.swift
```diff
@@ -0,0 +1,111 @@
...
+            static func ==(lhs: AnEnum, rhs: AnEnum) -> Bool {
+                return true
+            }
+        }
+        """
+    ]
+
+    static let triggeringExamples: [String] = [
+        // NSObject subclass implementing ==
+        """
```
marcelofabri: you should use the violation marker (`↓`) to test where the violation location

CHANGELOG.md
```diff
@@ -10,7 +10,11 @@
 
 #### Enhancements
 
-* None.
+* Add `nsobject_prefer_isequal` rule to warn against implementing `==` on an
+  `NSObject` subclass as calling `isEqual` (i.e. when using the class from
+  Objective-C) will will not use the defined `==` method.
```

Source/SwiftLintFramework/Rules/Lint/NSObjectPreferIsEqualRule.swift
```diff
@@ -0,0 +1,68 @@
...
+    // MARK: - Private
+
+    private func objcVisibleClasses(in file: File) -> [[String: SourceKitRepresentable]] {
+        return file.structure.dictionary.substructure.filter { dictionary in
+            guard
+                let kind = dictionary.kind,
+                SwiftDeclarationKind(rawValue: kind) == .class
+            else { return false }
+            let isDirectNSObjectSubclass = dictionary.inheritedTypes.contains("NSObject")
+            let isMarkedObjc = dictionary.enclosedSwiftAttributes.contains(where: { $0 == .objc })
```
marcelofabri: nitpick: you can use `.contains(.objc)`

Source/SwiftLintFramework/Rules/Lint/NSObjectPreferIsEqualRule.swift
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+
+    private func isDoubleEqualsMethod(_ method: [String: SourceKitRepresentable],
+                                      onType typeName: String) -> Bool {
+        guard
+            let kind = method.kind.flatMap(SwiftDeclarationKind.init),
+            let name = method.name,
+            kind == .functionMethodStatic,
+            name.hasPrefix("=="),
```
marcelofabri: maybe check if it's equal to `==(_:_:)`? I'm a bit worried that using just `==` might cause some (very rare) false positives on custom operators.

Source/SwiftLintFramework/Rules/Lint/NSObjectPreferIsEqualRule.swift
```diff
@@ -0,0 +1,68 @@
...
+            method.enclosedVarParameters.count == 2,
+            areAllArguments(toMethod: method, ofType: typeName)
+        else { return false }
+        return true
+    }
+
+    private func areAllArguments(toMethod method: [String: SourceKitRepresentable],
+                                 ofType typeName: String) -> Bool {
+        return method.enclosedVarParameters.reduce(true) { soFar, param -> Bool in
+            soFar && (param.typeName == typeName)
```
marcelofabri: I was going to say that this might not work for generic types, but `@objc` classes can't be generic, so no problem here!

Source/SwiftLintFramework/Rules/Lint/NSObjectPreferIsEqualRule.swift
```diff
@@ -0,0 +1,68 @@
+import SourceKittenFramework
+
+public struct NSObjectPreferIsEqualRule: Rule, OptInRule, ConfigurationProviderRule, AutomaticTestableRule {
```
marcelofabri: I think we could promote this to a default rule, given how easy it's to make a mistake (I've definitely made this mistake several times)

marcelofabri:
@matthew-healy Thanks so much for this! This is a great first contribution! 💯

I've added a few minor comments, but otherwise, it looks good to me 🚀

matthew-healy:
@marcelofabri Thanks! I think I've addressed all of your comments. 😄

Issue #2658, opened 2019-02-22
[Rule Request] Prefer reduce(into:_:) over reduce(_:_:)


marcelofabri:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.

[The docs](https://developer.apple.com/documentation/swift/array/3126956-reduce) say:

> This method is preferred over reduce(_:_:) for efficiency when the result is a copy-on-write type, for example an Array or a Dictionary.

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

See https://github.com/realm/SwiftLint/pull/2657 for some examples.

3. Should the rule be configurable, if so what parameters should be configurable?

Just severity

4. Should the rule be opt-in or enabled by default? Why?
   See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a rule as opt-in.

Enabled by default if we can avoid false positives

jpsim:
> Enabled by default if we can avoid false positives

I’d love to see a rule like this but I don’t know if I’d enable it by default. If the code isn’t performance critical, there are times when the legibility of the other reduce function makes it worth using it.

PR #2666, opened 2019-03-05
Add reduce_into opt-in rule

daltonclaybrook:
Fixes: #2658 
Prefer `reduce(into:_:)` over `reduce(_:_:)` for performance reasons


daltonclaybrook:
I'm guessing this last queued check is broken. Thoughts @marcelofabri ?

CHANGELOG.md
```diff
@@ -15,6 +15,10 @@
   Objective-C) will will not use the defined `==` method.  
   [Matthew Healy](https://github.com/matthew-healy)
   [#2663](https://github.com/realm/SwiftLint/pull/2663)
+* Add `reduce_into` opt-in rule to encourage the use of `reduce(into:_:)`
```
marcelofabri: missing empty line before this entry

Package.resolved
```diff
@@ -10,6 +10,15 @@
           "version": "0.15.0"
         }
       },
+      {
```
marcelofabri: this shouldn't be commited (I think you run on linux)

Source/SwiftLintFramework/Models/MasterRuleList.swift
```diff
@@ -1,4 +1,4 @@
-// Generated using Sourcery 0.15.0 — https://github.com/krzysztofzablocki/Sourcery
+// Generated using Sourcery 0.14.0 — https://github.com/krzysztofzablocki/Sourcery
```
marcelofabri: 👀

Source/SwiftLintFramework/Rules/Idiomatic/ReduceIntoRule.swift
```diff
@@ -0,0 +1,35 @@
...
+        ],
+        triggeringExamples: [
+            "let matches = lines.↓reduce([]) { matches, line in",
+            "zip(group, group.dropFirst()).↓reduce([]) { result, pair in",
+            "values.↓reduce(\"\") { $0 + \"\\($1)\" }"
+        ]
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        let pattern = "reduce\\((?!into:)"
```
marcelofabri: what do you think about implementing this as an `ASTRule` instead?

You can check the method name and check that the first parameter is a COW structure (a collection or string literal).

Source/SwiftLintFramework/Rules/Idiomatic/ReduceIntoRule.swift
```diff
@@ -0,0 +1,35 @@
...
+    public init() {}
+
+    public static var description = RuleDescription(
+        identifier: "reduce_into",
+        name: "Reduce Into",
+        description: "Prefer `reduce(into:_:)` over `reduce(_:_:)`",
+        kind: .performance,
+        minSwiftVersion: .three,
+        nonTriggeringExamples: [
+            "let matches = lines.reduce(into: []) { matches, line in",
```
marcelofabri: can you make sure that the examples are valid Swift code (e.g. the closures are also valid)?

Tests/LinuxMain.swift
```diff
@@ -1,4 +1,4 @@
-// Generated using Sourcery 0.15.0 — https://github.com/krzysztofzablocki/Sourcery
+// Generated using Sourcery 0.14.0 — https://github.com/krzysztofzablocki/Sourcery
```
marcelofabri: 👀

Tests/SwiftLintFrameworkTests/AutomaticRuleTests.generated.swift
```diff
@@ -1,4 +1,4 @@
-// Generated using Sourcery 0.15.0 — https://github.com/krzysztofzablocki/Sourcery
+// Generated using Sourcery 0.14.0 — https://github.com/krzysztofzablocki/Sourcery
```
marcelofabri: 👀

Source/SwiftLintFramework/Rules/Idiomatic/ReduceIntoRule.swift
daltonclaybrook: Oh interesting. I'll look into this.

marcelofabri: Let me know if you need help or have any questions!

daltonclaybrook: Just updated. I'm sure my implementation isn't exactly what you meant, but I'll need a bit more direction on how to proceed. I'm happy to keep tweaking it if you'll clarify a bit more, though.

marcelofabri: I think this is a good start! 

I'd also check if the first parameter is one of the following:

- array literal: `kind == source.lang.swift.expr.array` (we have this as an enum in `SwiftExpressionKind`)
- dictionary literal: `kind == source.lang.swift.expr.dictionary`
- string literal: this one is trickier because you'll need to use the `syntaxMap` and look for a string syntax type inside the first argument range
- an array or dictionary init, e.g `[Int]()`, `Array<Int>()`, `[Int: String]()`, `Dictionary<Int, String>()`

You can check https://github.com/realm/SwiftLint/pull/2657/files for some real-world examples.

daltonclaybrook: Thanks for the feedback! I'll keep iterating.

daltonclaybrook:
@marcelofabri You should be good to take another look. Sorry for the delay! As always, I'm happy to keep tweaking if necessary. 😄

marcelofabri:
@daltonclaybrook Thanks! I was on vacation, I'll try to take a look in the next few days.

daltonclaybrook:
@marcelofabri Anything I can do to help push this along? Happy to make additional updates if necessary.

marcelofabri:
Sorry for the delay, @daltonclaybrook. Thanks again for your recent contributions!

daltonclaybrook:
Happy to help. We ❤️ SwiftLint at Peloton!

Source/SwiftLintFramework/Rules/Performance/ReduceIntoRule.swift
```diff
@@ -0,0 +1,155 @@
...
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static var description = RuleDescription(
+        identifier: "reduce_into",
+        name: "Reduce Into",
+        description: "Prefer `reduce(into:_:)` over `reduce(_:_:)` for copy-on-write types",
+        kind: .performance,
+        minSwiftVersion: .three,
```
jpsim: Thanks for the great contribution, @daltonclaybrook!

`reduce(into:)` was added in Swift 4 however, not Swift 3: https://github.com/apple/swift-evolution/blob/master/proposals/0171-reduce-with-inout.md

jpsim: Bumped it to 4 in 230bedcd.

jpsim:
Thanks again for the PR. It'd be nice to expand this to handle more cases, like `reduce(into: String(), ...)`. I'm sure there are others too.

PR #2671, opened 2019-03-06
Show test failures on correct line

matthew-healy:
I noticed while running tests in Xcode that test failures were showing up in the `TestHelpers.swift` file, rather than inside the failing test method. (This was also true of tests run via `swift test`, though perhaps less noticeable).

This is easy enough to fix by passing the `#file` and `#line` values from the custom assertion's callsite down to the `XCT` assertions that are eventually called.

(The standard names for these arguments are `file` and `line`, but in some cases there was already a variable called `file` created in the body of the test, so I renamed the `file` argument to `testFile`).

### DIFF
--- a/Tests/SwiftLintFrameworkTests/TestHelpers.swift
+++ b/Tests/SwiftLintFrameworkTests/TestHelpers.swift
@@ -91,7 +91,8 @@ private func render(locations: [Location], in contents: String) -> String {
 }
 
 private extension Configuration {
-    func assertCorrection(_ before: String, expected: String) {
+    func assertCorrection(_ before: String, expected: String,
+                          testFile: StaticString = #file, line: UInt = #line) {
         let (cleanedBefore, markerOffsets) = cleanedContentsAndMarkerOffsets(from: before)
         let file = File.temporary(withContents: cleanedBefore)
         // expectedLocations are needed to create before call `correct()`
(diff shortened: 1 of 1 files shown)


marcelofabri:
@matthew-healy, thanks for the PR! 

I have mixed feelings though - I feel like  `verifyRule` tests a lot of things and sometimes it's useful to know exactly which check failed, using a test failure breakpoint.

Let me take some time to think about this. If other contributors want to chime in, that'd be very helpful too 💯

matthew-healy:
@marcelofabri Sure thing! I assumed there might be some discussion around this. 👍 

Totally agree that `verifyRule` is checking a lot of things, and that might impact how useful a change this is. 

My approach on other projects would be to break tests like `verifyRule` into a few more granular tests. With this approach passing in the `file` and `line` tells you not only what rule a test failed for, but also exactly which check failed. (I haven't looked in enough detail to know if that's desirable for SwiftLint tests, but it might provide some context on this change.)

When I added a new `Rule` to the project, knowing what check failed was enough because I could safely assume that only the tests for the new `Rule` were failing. I suspect that if you're modifying some of the underlying machinery of SwiftLint in such a way that multiple `Rule`s start breaking, it might be useful to know quickly exactly which ones they were. (Though admittedly I haven't done that, so it's just a guess. 😁) 

Either way, I'm happy to merge or close this PR depending on what folk think. 😄

matthew-healy:
Hey folks, I'm going to close this PR since it doesn't seem like there was much interest in discussing it (and I don't think it's fair to leave open PRs around for too long).

PR #2672, opened 2019-03-06
Add NSLocalizedStringRequireBundleRule

### LINKED ISSUE #2595: Rule Request: Enforce bundle parameter to NSLocalizedString
### New Issue Checklist

### New rule request

When writing frameworks that have localized strings, it's really easy to forget the `bundle:` parameter to `NSLocalizedString`. Unfortunately this isn't something that can be solved in code; adding my own override of `NSLocalizedString` for the bundle-less case trips up `genstrings`, and adding a custom function with a different name isn't compatible with `xcodebuild -exportLocalizations`.

The goal of the rule here is to catch all calls to `NSLocalizedString` and enforce that the `bundle:` parameter is present. It doesn't matter what the value is, it should just be specified.

This would be opt-in. I don't think there's any configuration for it.

matthew-healy:
Fixes #2595, by adding `NSLocalizedStringRequireBundleRule`. This is an opt-in rule that triggers when calls to `NSLocalizedString` do not contain a `bundle` argument.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,11 @@
   [Cihat Gündüz](https://github.com/Dschee)
   [#2294](https://github.com/realm/SwiftLint/issues/2294)
 
+* Add `nslocalizedstring_require_bundle` rule to ensure calls to
+  `NSLocalizedString` specify the bundle where the strings file is located.  
+  [Matthew Healy](https://github.com/matthew-healy)
+  [#2595](https://github.com/realm/SwiftLint/2595)
+
 #### Bug Fixes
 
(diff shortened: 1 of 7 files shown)


matthew-healy:
@marcelofabri Conflicts fixed. Sorry for the delay. 😄

matthew-healy:
@marcelofabri Conflicts fixed again. Can you let me know if you want to merge this? If not I'm happy to close it. 😄

marcelofabri:
Sorry for the delay, @matthew-healy, I was on vacation 🌴

Merging!

Issue #2583, opened 2019-01-23
Rule Request: Reduce Boolean


jpsim:
Now that we have `Sequence.allSatisfy` in Swift 4.2+, it's quite likely that any time we see `reduce(true)` or `reduce(false)` it could be simplified to either `.contains(...)` or `.allSatisfy(...)` to allow for bailing out early and better readability.

> 1. Why should this rule be added? Share links to existing discussion about what
>    the community thinks about this.

Should be added to improve readability and performance. See examples in https://github.com/apple/swift-evolution/blob/master/proposals/0207-containsOnly.md

> 2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

```swift
// Trigger
let allNines = nums.reduce(true) { $0.0 && $0.1 == 9 } // should use allSatisfy instead
let anyNines = nums.reduce(false) { $0.0 || $0.1 == 9 } // should use contains instead
let allValid = validators.reduce(true) { $0 && $1(input) } // should use allSatisfy instead
let anyValid = validators.reduce(false) { $0 || $1(input) } // should use contains instead

// Wouldn't trigger
nums.reduce(0) { $0.0 + $0.1 }
nums.reduce(0.0) { $0.0 + $0.1 }
```

> 3. Should the rule be configurable, if so what parameters should be configurable?

We should only lint this rule when we detect a Swift version of 4.2 or later.

> 4. Should the rule be opt-in or enabled by default? Why?
>    See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a rule as opt-in.

I can't think of any case where you'd want to reduce over a boolean rather than use contains or allSatisfy, so I think this should be enabled by default. I'd love to know if I'm missing some use cases though.

marcelofabri:
This was implemented by #2675 🎉

cjamie:
you would want to reduce boolean over a contains when you want to execute absolutely every block passed in. 

example: 


```
protocol Cond {
    var condition: Bool { get }
}

struct Hello: Cond {
    var condition: Bool {
        // some code I must execute
        return false
    }
}

struct Hello1: Cond {
    var condition: Bool {
        // some code I must execute
        return false
    }
}
struct HelloTrue: Cond {
    var condition: Bool {
        // some code I must execute
        return false
    }
}


let temp: [Cond] = [Hello(), HelloTrue(), Hello1()]
    
let result = temp.contains(where: {$0.condition})

print(result)
```

[...]

PR #2675, opened 2019-03-09
Add ReduceBooleanRule

xavierLowmiller:
Implements rule request #2583.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Performance/ReduceBooleanRule.swift
+++ b/Source/SwiftLintFramework/Rules/Performance/ReduceBooleanRule.swift
@@ -7,7 +7,7 @@ public struct ReduceBooleanRule: Rule, ConfigurationProviderRule, AutomaticTesta
 
     public static let description = RuleDescription(
         identifier: "reduce_boolean",
-        name: "Prefer allSatisfy or contains over reduce(true) or reduce(false)",
+        name: "Reduce Boolean",
         description: "Prefer using `.allSatisfy()` or `.contains()` over `reduce(true)` or `reduce(false)`",
         kind: .performance,
         minSwiftVersion: .fourDotTwo,
--- a/Source/SwiftLintFramework/Rules/Performance/ReduceBooleanRule.swift
+++ b/Source/SwiftLintFramework/Rules/Performance/ReduceBooleanRule.swift
@@ -32,7 +32,7 @@ public struct ReduceBooleanRule: Rule, ConfigurationProviderRule, AutomaticTesta
         return file
             .match(pattern: pattern,
                    excludingSyntaxKinds: SyntaxKind.commentAndStringKinds)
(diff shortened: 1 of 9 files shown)


Source/SwiftLintFramework/Rules/Performance/ReduceBooleanRule.swift
```diff
@@ -0,0 +1,51 @@
...
+            "let allValid = validators.↓reduce(true, { $0 && $1(input) })",
+            "let anyValid = validators.↓reduce(false, { $0 || $1(input) })"
+        ]
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        let pattern = "\\breduce\\((true|false)"
+        return file
+            .match(pattern: pattern,
+                   excludingSyntaxKinds: SyntaxKind.commentAndStringKinds)
```
jpsim: This pattern should always have the same two syntax kinds. We should match for those instead of excluding comments/strings.

Source/SwiftLintFramework/Rules/Performance/ReduceBooleanRule.swift
```diff
@@ -0,0 +1,51 @@
+import SourceKittenFramework
+
+public struct ReduceBooleanRule: Rule, ConfigurationProviderRule, AutomaticTestableRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "reduce_boolean",
+        name: "Prefer allSatisfy or contains over reduce(true) or reduce(false)",
```
jpsim: This is a bit verbose compared to our usual rule names. Can we change this to “Reduce Boolean”?

Source/SwiftLintFramework/Rules/Performance/ReduceBooleanRule.swift
```diff
@@ -0,0 +1,51 @@
...
+            "let anyValid = validators.↓reduce(false, { $0 || $1(input) })"
+        ]
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        let pattern = "\\breduce\\((true|false)"
+        return file
+            .match(pattern: pattern,
+                   excludingSyntaxKinds: SyntaxKind.commentAndStringKinds)
+            .compactMap { range in
```
jpsim: I think this is just `map`.

xavierLowmiller: Yes, you are right.

Source/SwiftLintFramework/Rules/Performance/ReduceBooleanRule.swift
xavierLowmiller: Sure thing

Issue #2578, opened 2019-01-22
Rule Request: delete unused auto-generated code


yonat:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

> 1. Why should this rule be added?

Xcode auto-generates function like `didReceiveMemoryWarning()` that are often unused, and just left in code to rot.

> 2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

Triggering:

```swift
    override func didReceiveMemoryWarning() {
        super.didReceiveMemoryWarning()
        // Dispose of any resources that can be recreated.
    }

    func applicationWillResignActive(_ application: UIApplication) {
        // Sent when the application is about to move from active to inactive state. This can occur for certain types of temporary interruptions (such as an incoming phone call or SMS message) or when the user quits the application and it begins the transition to the background state.
        // Use this method to pause ongoing tasks, disable timers, and invalidate graphics rendering callbacks. Games should use this method to pause the game.
    }

    override func viewDidDisappear(_ animated: Bool) {
        super.viewDidDisappear(animated)
    }

    override func setSelected(_ selected: Bool, animated: Bool) {
        super.setSelected(selected, animated: animated)

        // Configure the view for the selected state
    }

    override func viewDidLoad() {
        super.viewDidLoad()

        // Do any additional setup after loading the view.
    }
```

Non-triggering:

```swift
func emptyImplementationForRquiredProtocolFunction() {
    }
```

> 3. Should the rule be configurable, if so what parameters should be configurable?

No params, I think.
> 4. Should the rule be opt-in or enabled by default? Why?

Enabled by default, unused code should be deleted.

jpsim:
I would welcome a rule that validated that overridden methods do more than simply call `super`.

PR #2680, opened 2019-03-13
Implement Empty Method Override

xavierLowmiller:
Implements the rule mentioned in #2578:
This will match code that doesn't do anything but call super.

I'm very open for better names for this rule.

I'd love to have some feedback on the implementation of this before I tackle implementing an autocorrection for this.


Source/SwiftLintFramework/Rules/Lint/CallSuperOnlyRule.swift
```diff
@@ -0,0 +1,74 @@
...
+
+        let pattern = signature + body
+
+        return file
+            .match(pattern: pattern)
+            .compactMap { range, syntaxKinds in
+                // Skip matches that occur in strings and comments
+                guard syntaxKinds != [.string],
+                    syntaxKinds != [.comment]
+                    else { return nil }
```
xavierLowmiller: I'd really like a better solution for this.

If I use `.match(pattern: pattern, excludingSyntaxKinds: SyntaxKind.commentAndStringKinds)`, the rule will stop working, because comments are part of this rule.

marcelofabri: can you try implementing this as an `ASTRule`? Using regexes is very error prone and one of the major causes of bugs and slowness in SwiftLint.

You can check for method declarations with the `override` attribute and then search for method calls inside that declaration's body.

xavierLowmiller: That sounds like a great idea 👍 

I also noticed I get a lot of false positives from the open source repos with the regex implementation.

I'll figure it out and update the PR.

marcelofabri: Awesome! Feel free to ask any questions 💯

xavierLowmiller: So, `ASTRule`s are awesome and the implementation is much simpler now. Thanks 🙌!

I have a problem with the offset of the expected failure location markers (`↓`).
I'm pretty sure the offset that is calculated is at the right spot, but the test fails anyway.

Can you take a look at my latest commit?

xavierLowmiller: Also, I'm getting a lot of false positives when the body contains a variable assignment (which doesn't show up in the `[String: SourceKitRepresentable]` dictionary).

Do you think it's sufficient to check for an `=` in the method body?

Rules.md
```diff
@@ -5,6 +5,7 @@
 * [Array Init](#array-init)
 * [Attributes](#attributes)
 * [Block Based KVO](#block-based-kvo)
+* [Call Super Only](#call-super-only)
```
Dschee: What about naming the rule `empty_method_override` similar to the already existing empty prefixed rules? Since the call to super doesn't actually do anything, the method can be considered empty.

Alternatively, `unused_method_override`, `unused_method_declaration` or `unused_method_body` might also work. I don't like th ename `call_super_only` because it reads like an invitation to call super and might be confused to be a rule, that ensures you don't forget to call super in methods decared with `override`.

xavierLowmiller: Yes, these are great suggestions, thank you! 👍

I'm going to go for `empty_method_override`.

xavierLowmiller:
@Dschee could you elaborate on what indentation you're not a fan of?

I'm willing to change some stuff if it looks weird to you and the result will better match the overall code style of SwiftLint.

Dschee:
My preference doesn't necessarily represent the SwiftLint preferred style, but I'll give you some comments what I mean via another review.

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
```diff
@@ -0,0 +1,158 @@
+import Foundation
+import SourceKittenFramework
+
+public struct EmptyMethodOverrideRule: SubstitutionCorrectableASTRule,
+ConfigurationProviderRule, AutomaticTestableRule {
```
Dschee: This looks ugly to me. If it doesn't fit into a single line, one solution could be to extract some conformances to extensions if possible. Something like this:

```swift
public struct EmptyMethodOverrideRule: SubstitutionCorrectableASTRule {
...
}

extension EmptyMethodOverrideRule: ConfigurationProviderRule, AutomaticTestableRule {}
```

If that's not possible or doesn't fit well, I'd rather indent it like this:

```swift
public struct EmptyMethodOverrideRule: SubstitutionCorrectableASTRule, 
        ConfigurationProviderRule, AutomaticTestableRule { // indent by 8 spaces
    public var configuration = SeverityConfiguration(.warning) // normal indent
```
or even like this:
```swift
public struct EmptyMethodOverrideRule: 
    SubstitutionCorrectableASTRule, 
    ConfigurationProviderRule, 
    AutomaticTestableRule 
{
    public var configuration = SeverityConfiguration(.warning) // normal indent
```
which I agree also looks weird though. 😅

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
```diff
@@ -0,0 +1,158 @@
...
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    static let nonTriggeringExamples = [
+        """
+        override func viewDidDisappear(_ animated: Bool) {
```
Dschee: I like to think of `"""` like of opening braces. The only problem is the closing `"""` which defines if there will be spaces added to the String. Thus I use this:

```swift
    """
        override func viewDidDisappear(_ animated: Bool) {
            super.viewDidDisappear(animated)
            print("View controller did disappear")
        }
        """,
    """
        ...
        """,
```

But this doesn't bother me too much as it also looks quite clean and readable.

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
```diff
@@ -0,0 +1,158 @@
...
+        class
+        override
+        func setUp() {
+            super.setUp()
+        }
+        """
+    ].map(wrapInClass)
+
+    static let corrections = Dictionary(uniqueKeysWithValues:
+        EmptyMethodOverrideRule.triggeringExamples.map { ($0, wrapInClass("")) })
```
Dschee: Clearly preferred:

```swift
    static let corrections = Dictionary(
        uniqueKeysWithValues: EmptyMethodOverrideRule.triggeringExamples.map { ($0, wrapInClass("")) }
    )
```
or by using a variable like this:

```swift
    static let corrections: [String: String] = {
        let keysWithValues = EmptyMethodOverrideRule.triggeringExamples.map { ($0, wrapInClass("")) }
        return Dictionary(uniqueKeysWithValues: keysWithValues)
    }()
```

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
```diff
@@ -0,0 +1,158 @@
...
+        name: "Empty Method Override",
+        description: "Methods that don't do anything but call `super` can be removed",
+        kind: .lint,
+        nonTriggeringExamples: EmptyMethodOverrideRule.nonTriggeringExamples,
+        triggeringExamples: EmptyMethodOverrideRule.triggeringExamples,
+        corrections: EmptyMethodOverrideRule.corrections
+    )
+
+    public func violationRanges(in file: File, kind: SwiftDeclarationKind,
+                                dictionary: [String: SourceKitRepresentable]) -> [NSRange] {
```
Dschee: Clearly preferred:

```swift
public func violationRanges(
    in file: File, 
    kind: SwiftDeclarationKind,
    dictionary: [String: SourceKitRepresentable]
) -> [NSRange] {
```

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
```diff
@@ -0,0 +1,158 @@
...
+            .functionMethodClass,
+            .functionMethodStatic
+        ]
+        guard overridingKinds.contains(kind),
+            dictionary.enclosedSwiftAttributes.contains(.override),
+            !dictionary.enclosedSwiftAttributes.contains(.public),
+            file.onlyCallsSuper(dictionary),
+            let offset = dictionary.offset,
+            let length = dictionary.length
+            else { return [] }
```
Dschee: The else has to match the level of the `guard` here:

```swift
        guard 
            overridingKinds.contains(kind),
            dictionary.enclosedSwiftAttributes.contains(.override),
            !dictionary.enclosedSwiftAttributes.contains(.public),
            file.onlyCallsSuper(dictionary),
            let offset = dictionary.offset,
            let length = dictionary.length
        else { return [] }
```

or with the body not on a single line:

```swift
        guard 
            overridingKinds.contains(kind),
            dictionary.enclosedSwiftAttributes.contains(.override),
            !dictionary.enclosedSwiftAttributes.contains(.public),
            file.onlyCallsSuper(dictionary),
            let offset = dictionary.offset,
            let length = dictionary.length
        else { 
            return [] 
        }
```

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
```diff
@@ -0,0 +1,158 @@
...
+        let startIndex = dictionary.swiftAttributes
+            .compactMap { $0.offset }
+            .min() ?? offset
+        let endIndex = offset + length
+
+        return [NSRange(startIndex..<endIndex)]
+    }
+
+    public func validate(file: File, kind: SwiftDeclarationKind,
+                         dictionary: [String: SourceKitRepresentable]) -> [StyleViolation] {
```
Dschee: See `violationRanges` comment above.

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
```diff
@@ -0,0 +1,158 @@
...
+private func wrapInClass(_ string: String) -> String {
+    return """
+    class ViewController: UIViewController {
+    \(string.isEmpty ? "    " : string
+        .split(separator: "\n")
+        .map { "    " + $0 }
+        .joined(separator: "\n")
+    )
+    }
+    """
```
Dschee: Here, clearly my style pays off as the `class ViewController ...` doesn't look good here, but with my style it would:

```swift
return """
    class ViewController: UIViewController {
    \(string.isEmpty ? "    " : string
        .split(separator: "\n")
        .map { "    " + $0 }
        .joined(separator: "\n")
    )
    }
    """
```

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
```diff
@@ -0,0 +1,158 @@
...
+    func onlyCallsSuper(_ dictionary: [String: SourceKitRepresentable]) -> Bool {
+        if let name = dictionary.name?.split(separator: "(").first,
+            dictionary.substructure.count == 1,
+            let methodCall = dictionary.substructure.first,
+            methodCall.name == "super.\(name)",
+            !hasAssignmentInBody(dictionary) {
+            return true
+        } else {
+            return false
+        }
```
Dschee: Here, you're doing it nearly as I suggest, but having the beginning of the condition on the same line like `if` doesn't give a nice balance, also it's not clear where the body of the `if` begins, so I'd rather do this:

```swift
        if 
            let name = dictionary.name?.split(separator: "(").first,
            dictionary.substructure.count == 1,
            let methodCall = dictionary.substructure.first,
            methodCall.name == "super.\(name)",
            !hasAssignmentInBody(dictionary)
        {
            return true
        } else {
            return false
        }
```

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
```diff
@@ -0,0 +1,158 @@
...
+
+    private func hasAssignmentInBody(_ dictionary: [String: SourceKitRepresentable]) -> Bool {
+        guard let bodyOffset = dictionary.bodyOffset,
+            let bodyLength = dictionary.bodyLength,
+            let range = contents.bridge().byteRangeToNSRange(start: bodyOffset, length: bodyLength)
+            else { return false }
+
+        let body = contents.substring(from: range.location, length: range.length)
+        let assignmentOperators = ["=", "*=", "/=", "%=", "+=", "-=",
+                                   "<<=", ">>=", "&=", "|=", "^="]
```
Dschee: This is Objective-C style indentation, I'd rather do this:
``` swift
        let assignmentOperators = [
            "=", "*=", "/=", "%=", "+=", "-=", "<<=", ">>=", "&=", "|=", "^="
        ]
```

or even this:

```swift
        let assignmentOperators = [
            "=", 
            "*=", 
            "/=", 
            "%=", 
            "+=", 
            "-=", 
            "<<=", 
            ">>=", 
            "&=", 
            "|=", 
            "^="
        ]
```

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
```diff
@@ -0,0 +1,158 @@
...
+                ruleDescription: type(of: self).description,
+                severity: configuration.severity,
+                location: Location(file: file, byteOffset: $0.location)
+            )
+        }
+    }
+
+    public func substitution(for violationRange: NSRange, in file: File) -> (NSRange, String) {
+        let range = file.contents.bridge()
+            .byteRangeToNSRange(start: violationRange.location, length: violationRange.length)!
```
Dschee: This is actually fine, but I slightly prefer this:

```swift
        let range = file.contents.bridge().byteRangeToNSRange(
            start: violationRange.location, 
            length: violationRange.length
        )!
```

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
xavierLowmiller: I agree

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
xavierLowmiller: Yes

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
xavierLowmiller: I think this is much better :+1:

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
xavierLowmiller: I like the balance of the indentation...

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
xavierLowmiller: I agree about the balance, but wasn't sure about Allman in Swift, so I didn't go for it at first. Looks better this way, so let's do it like this.

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
xavierLowmiller: I really like the `extension X: Protocol` approach, but the SwiftLint default seems to be to have all protocols on line 1, so I went with that.

Dschee:
@xavierLowmiller Thank you for the indentation changes, looks much better to me now!

Note that you forgot to change the name in the CHANGELOG entry.
Apart from that, I think this looks good.

xavierLowmiller:
Is there anything I can do to help get this merged?

Dschee:
Nothing from my side. I just waited for someone else to give their 👍 because this is enabled by default and I'd like to have at least one other opinion for enabled rules. But according to the introduced warnings in the test projects, this looks good.

The only thing that might hold this off from my point of view is performance, for example:

`Linting Realm with this PR took 7.06s vs 4.94s on master (42% slower)`

Maybe we should optimize performance if this is possible in an easy way. Would be good to know why the change is so big in the Realm project ... 42% is really not to be overlooked.

xavierLowmiller:
Ah yes 42% is a pretty bad regression. I'll see if I can reproduce and fix it later today.

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
```diff
@@ -0,0 +1,172 @@
...
+        let overridingKinds: [SwiftDeclarationKind] = [
+            .functionMethodInstance,
+            .functionMethodClass,
+            .functionMethodStatic
+        ]
+
+        guard
+            overridingKinds.contains(kind),
+            dictionary.enclosedSwiftAttributes.contains(.override),
+            !dictionary.enclosedSwiftAttributes.contains(.public),
```
marcelofabri: why is there this condition?

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
```diff
@@ -0,0 +1,172 @@
...
+            let offset = dictionary.offset,
+            let length = dictionary.length
+        else { return [] }
+
+        let startIndex = dictionary.swiftAttributes
+            .compactMap { $0.offset }
+            .min() ?? offset
+        let endIndex = offset + length
+
+        return [NSRange(startIndex..<endIndex)]
```
marcelofabri: nit: can you just return the location instead? I don't think this length is always correct because the start index depends on the attributes

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
```diff
@@ -0,0 +1,172 @@
...
+        guard
+            overridingKinds.contains(kind),
+            dictionary.enclosedSwiftAttributes.contains(.override),
+            !dictionary.enclosedSwiftAttributes.contains(.public),
+            file.onlyCallsSuper(dictionary),
+            let offset = dictionary.offset,
+            let length = dictionary.length
+        else { return [] }
+
+        let startIndex = dictionary.swiftAttributes
```
marcelofabri: can you use `offset` instead?

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
marcelofabri: nit: you can use `NSRange(location:length)` to avoid the need for `endIndex`

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
```diff
@@ -0,0 +1,172 @@
...
+                location: Location(file: file, byteOffset: $0.location)
+            )
+        }
+    }
+
+    public func substitution(for violationRange: NSRange, in file: File) -> (NSRange, String) {
+        let range = file.contents.bridge().byteRangeToNSRange(
+            start: violationRange.location,
+            length: violationRange.length
+        )!
```
marcelofabri: can we avoid this force unwrap here just to be safe?

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
```diff
@@ -0,0 +1,172 @@
...
+            let bodyLength = dictionary.bodyLength,
+            let range = contents.bridge().byteRangeToNSRange(start: bodyOffset, length: bodyLength)
+            else { return false }
+
+        let body = contents.substring(from: range.location, length: range.length)
+        let assignmentOperators = [
+            "=", "*=", "/=", "%=", "+=", "-=", "<<=", ">>=", "&=", "|=", "^="
+        ]
+
+        return assignmentOperators.contains(where: body.contains)
```
marcelofabri: should we check that the operator is not in a comment or string?

CHANGELOG.md
```diff
@@ -90,6 +90,10 @@
   [Cihat Gündüz](https://github.com/Dschee)
   [#2294](https://github.com/realm/SwiftLint/issues/2294)
 
+* Add `empty_method_override` rule to warn against methods that only call `super`.
```
marcelofabri: this is on the wrong section now

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
xavierLowmiller: Consider the following example:

```swift
class A {
    internal init() {}
}

class B: A {
    public override init() {
        super.init()
    }
}
```

In this case, the empty override is used to expose a method that has a more restrictive ACL in the superclass.

I'm not sure if we should address this, but I realized the same should apply for `open` declarations.

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
xavierLowmiller: The `offset` of `functionMethodInstance` gives me the offset of the `func` keyword, so I have to find the offset of any attributes like `internal` and `override` manually.

Maybe there's a better way to do this I don't know of...

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
xavierLowmiller: Same as [here](https://github.com/realm/SwiftLint/pull/2680#discussion_r288588180), the `offset` variable isn't really the start of the violation, so I think it's more convenient to use the `endIndex` here for ensuring the correct violation range.

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
xavierLowmiller: I'm struggling with this... Especially multiline comments are a problem detecting this.

I'll look into this some more, and maybe find a better solution that doesn't rely on listing out all assignment operators.

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
marcelofabri: but does this actually work? won't this be missing the length of the attributes? because `length` should be the length from `func` to the last `}`.

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
marcelofabri: You can combine the operators into a regex and use `file.match(pattern:excludingSyntaxKinds:)`

Source/SwiftLintFramework/Rules/Lint/EmptyMethodOverrideRule.swift
xavierLowmiller: Yes, the length of the attributes is why I'm using this NSRange initializer:

```swift
internal override func a() { super.a() }
^ 1               ^ 2                  ^ 3

1 startIndex
2 offset
3 endIndex (offset + length)
```

So the range from `startIndex` to `endIndex` is the one that should be removed. `NSRange(location:length)` would require the knowledge of how long the attributes (`internal` + `override`) are, and that seems more complicated to me.

marcelofabri:
@xavierLowmiller I run this on the project that I work on and found a false positive:

```swift
class A: B {
    var forceFoo = false

    override func foo() -> Bool {
        return forceFoo || super.foo()
    }
}
```

It'd be good to fix this before merging or making the rule opt-in.

PR #2681, opened 2019-03-18
Update bundler version to v2.0.1

pixyzehn:
Updated the bundler version to v2.0.1 which is the latest release: https://github.com/bundler/bundler/releases

Also, on Azure Pipelines, it'll use the version in `Gemfile.lock`.


jpsim:
Thanks for the PR!

PR #2682, opened 2019-03-20
[Makefile] Avoid using `-Xswiftc -static-stdlib` if toolchain does not have `usr/lib/swift_static` directory

norio-nomura:
This is required on building with Xcode 10.2, because Xcode 10.2’s toolchain does not have `usr/lib/swift_static` directory.


jpsim:
Thanks for the fix!

Issue #2684, opened 2019-03-23
Discarded Notification Center Observer Violation false trigger


jsloop42:
New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for existing GitHub issues

### Describe the bug

Discarded Notification Center Observer Violation is triggered when I use the below code pattern where the notification observer result is appended to an array instead of assigning it to a variable.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint lint App/NotificationController.swift 
Loading configuration from '.swiftlint.yml'
Linting Swift files at paths App/NotificationController.swift
Linting 'NotificationController.swift' (1/1)
/Users/jsloop/dev/App/App/NotificationController.swift:34:27: warning: Discarded Notification Center Observer Violation: When registering for a notification using a block, the opaque observer that is returned should be stored so it can be removed later. (discarded_notification_center_observer)
/Users/jsloop/dev/App/App/NotificationController.swift:36:27: warning: Discarded Notification Center Observer Violation: When registering for a notification using a block, the opaque observer that is returned should be stored so it can be removed later. (discarded_notification_center_observer)
Done linting! Found 2 violations, 0 serious in 1 file.
```

### Environment

* SwiftLint version: 0.31.0
* Installation method used: Hombrew
* Paste your configuration file:

```yml
excluded:
- Carthage
- Pods
- SwiftLint/Common/3rdPartyLib

disabled_rules:
- trailing_whitespace
- identifier_name
- type_body_length
- large_tuple
- cyclomatic_complexity

opt_in_rules:
- unneeded_parentheses_in_closure_argument

force_cast: warning

line_length:
  warning: 160

file_length:
  warning: 1000
  
function_body_length:
  warning: 100
  error: 200

reporter: "xcode"

```

* Which Xcode version are you using: 10.1
* Do you have a sample that shows the issue? 

```swift
// This triggers a violation:
class NotificationController {
    private var notifs: [Any?] = []

    init() {
        notifs.append(NotificationCenter.default.addObserver(forName: Notification.Name("n1"),
                                                             object: nil, queue: nil, using: {_ in}))
        notifs.append(NotificationCenter.default.addObserver(forName: Notification.Name("n2"), 
                                                            object: nil, queue: nil, using: {_ in}))
    }
}
```

If I assign the return to a variable, the violation is not triggered.

```swift
// This *does not* trigger a violation:
class NotificationController {
    private var notifs: [Any?] = []

[...]

PR #2685, opened 2019-03-23
Update Discarded Notification Center Observer Violation false positive trigger on array append

jsloop42:
- This fixes an issue where the notification center observer is appended to an array, which triggers the violation. Fixes https://github.com/realm/SwiftLint/issues/2684.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@
   [#2628](https://github.com/realm/SwiftLint/issues/2628)
 
 * `discarded_notification_center_observer` rule now checks if the observer is
-  appended to an array before triggering the violation.
+  appended to an array before triggering the violation.  
   [jsloop42](https://github.com/jsloop42)
   [#2684](https://github.com/realm/SwiftLint/issues/2684)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@
   [#2628](https://github.com/realm/SwiftLint/issues/2628)
 
 * `discarded_notification_center_observer` rule now checks if the observer is
-  appended to an array before triggering the violation.  
+  added to any collection or passed to a function before triggering the violation.  
   [jsloop42](https://github.com/jsloop42)
   [#2684](https://github.com/realm/SwiftLint/issues/2684)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2628](https://github.com/realm/SwiftLint/issues/2628)
 
+* `discarded_notification_center_observer` rule now checks if the observer is
+  appended to an array before triggering the violation.
+  [jsloop42](https://github.com/jsloop42)
+  [#2684](https://github.com/realm/SwiftLint/issues/2684)
+
 ## 0.31.0: Busy Laundromat
 
 #### Breaking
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -31,6 +31,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2628](https://github.com/realm/SwiftLint/issues/2628)
 
+* `discarded_notification_center_observer` rule now checks if the observer is
+  appended to an array before triggering the violation.
```
Dschee: Please add two spaces after this line. (And don't ask me why, it's just the style here, probably becaus of Markdown restrictions on GitHub).

Source/SwiftLintFramework/Rules/Lint/DiscardedNotificationCenterObserverRule.swift
```diff
@@ -51,6 +53,11 @@ public struct DiscardedNotificationCenterObserverRule: ASTRule, ConfigurationPro
                 return []
         }
 
+        if let lastMatch = regex("append\\(").matches(in: file.contents, options: [], range: range).last?.range,
```
Dschee: This only prevents issues with `append` calls, but there are multiple ways of adding objects to a collection. Including these:

```swift
set.insert(notification)
array += [notification]
let array = [notification1, notification2, notification3]
array.insert(notification, at: 0)
```

A much easier way to prevent all of these issues is to simply check if the notification is passed to a function as a parameter. Such cases should not trigger warnings. The examples should be also extended with some additional ones like I listed above.

jsloop42:
Okay, I will update the PR to make it more generic.

Source/SwiftLintFramework/Rules/Lint/DiscardedNotificationCenterObserverRule.swift
jsloop42: I have updated to check if the notification is passed to any function.

CHANGELOG.md
```diff
@@ -31,6 +31,11 @@
...
```
Dschee: `to a collection` would be more correct imho, but nothing too important.

jsloop42:
I have added non-triggering example for [array append](https://github.com/realm/SwiftLint/pull/2685/files#diff-b5d5ebb28e58c901be755c56894b349bR3088) and [passing as function call argument](https://github.com/realm/SwiftLint/pull/2685/files#diff-b5d5ebb28e58c901be755c56894b349bR3094). I will add an example on dictionary.

Dschee:
Looks good to me now. Thank you very much! Merging. 🎉

jsloop42:
Welcome :)

Source/SwiftLintFramework/Rules/Lint/DiscardedNotificationCenterObserverRule.swift
```diff
@@ -51,6 +61,11 @@ public struct DiscardedNotificationCenterObserverRule: ASTRule, ConfigurationPro
                 return []
         }
 
+        if let lastMatch = regex("\\b[^\\(]+").matches(in: file.contents, options: [], range: range).last?.range,
```
marcelofabri: won't this cause false negatives? e.g.

`print(nc.addObserver(forName: .NSSystemTimeZoneDidChange, object: nil, queue: nil, using: { }))`

Issue #2683, opened 2019-03-20
False positives with negating parantheses on number_separator rule


Dschee:
For this code
```swift
let x: Double = -(200 / 447.214)
```
I'm getting a `number_separator` rule violation which I don't expect. 

I'm using the default configuration for `number_separator` with SwiftLint at version 0.31.0.

daltonclaybrook:
Looking into this.

PR #2687, opened 2019-03-25
Fix false positives on `number_separator` when the number is wrapped in parentheses.

daltonclaybrook:
Resolves: #2683

### DIFF
--- a/Source/SwiftLintFramework/Rules/Style/NumberSeparatorRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/NumberSeparatorRule.swift
@@ -137,19 +137,15 @@ public struct NumberSeparatorRule: OptInRule, CorrectableRule, ConfigurationProv
 
         let shouldAddSeparators = clean.count >= minimumLength
 
-        // number may contain characters that are not digits, such as when it is
-        // wrapped in parentheses. We only want to insert a "_" between each third
-        // _digit_, not character, so idx is now only incremented if the character is
-        // a digit.
-        var idx = 0
+        var numerals = 0
         for char in reversedIfNeeded(clean, reversed: !isFraction) {
             defer { correctComponents.append(String(char)) }
             guard char.unicodeScalars.allSatisfy(CharacterSet.decimalDigits.contains) else { continue }
 
-            if idx % 3 == 0 && idx > 0 && shouldAddSeparators {
+            if numerals % 3 == 0 && numerals > 0 && shouldAddSeparators {
                 correctComponents.append("_")
             }
-            idx += 1
+            numerals += 1
         }
 
         let expected = reversedIfNeeded(correctComponents, reversed: !isFraction).joined()
--- a/Source/SwiftLintFramework/Rules/Style/NumberSeparatorRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/NumberSeparatorRule.swift
@@ -137,24 +137,27 @@ public struct NumberSeparatorRule: OptInRule, CorrectableRule, ConfigurationProv
 
         let shouldAddSeparators = clean.count >= minimumLength
 
-        for (idx, char) in reversedIfNeeded(Array(clean), reversed: !isFraction).enumerated() {
+        var idx = 0
+        for char in reversedIfNeeded(clean, reversed: !isFraction) {
+            defer { correctComponents.append(String(char)) }
+            guard char.unicodeScalars.allSatisfy(CharacterSet.decimalDigits.contains) else { continue }
+
             if idx % 3 == 0 && idx > 0 && shouldAddSeparators {
                 correctComponents.append("_")
             }
-
-            correctComponents.append(String(char))
+            idx += 1
         }
 
         let expected = reversedIfNeeded(correctComponents, reversed: !isFraction).joined()
         return (expected == number, expected)
     }
 
-    private func reversedIfNeeded<T>(_ array: [T], reversed: Bool) -> [T] {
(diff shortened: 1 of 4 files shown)


Source/SwiftLintFramework/Rules/Style/NumberSeparatorRule.swift
```diff
@@ -137,24 +137,27 @@ public struct NumberSeparatorRule: OptInRule, CorrectableRule, ConfigurationProv
 
         let shouldAddSeparators = clean.count >= minimumLength
 
-        for (idx, char) in reversedIfNeeded(Array(clean), reversed: !isFraction).enumerated() {
+        var idx = 0
```
Dschee: Why is the `idx` parameter extracted into a manually incremented variable? As far as I can see, when the `guard` statements condition is `false`, then the `continue` will skip incrementing the `idx` variable. Is this intentional? And if so, why is this correct? Please either revert to the `enumerated()` variant or properly document the reasoning behind the manuel `idx` handling including why skipping the incrementation step prevents issues here.

Source/SwiftLintFramework/Rules/Style/NumberSeparatorRule.swift
```diff
@@ -137,24 +137,27 @@ public struct NumberSeparatorRule: OptInRule, CorrectableRule, ConfigurationProv
...
-            correctComponents.append(String(char))
+            idx += 1
         }
 
         let expected = reversedIfNeeded(correctComponents, reversed: !isFraction).joined()
         return (expected == number, expected)
     }
 
-    private func reversedIfNeeded<T>(_ array: [T], reversed: Bool) -> [T] {
+    private func reversedIfNeeded<T>(_ collection: T, reversed: Bool) -> [T.Element] where T: Collection {
```
Dschee: Is there a specific reason for changing this to a collection? I am by no means against it, I just wonder why you changed this when the change doesn't affect the code during execution time, or does it? Maybe the previous solution might even be faster since there's no additional array initialization step like here in the return statement, but I'm not sure about that ...

daltonclaybrook: This is purely a style change and should not affect the correctness of the program. I noticed that the call site was converting the `String` to an `Array` before passing to this function, but this shouldn't be necessary since `String` is a `Collection` and has its own implementation of `.reversed()`. Though, I would be happy to revert this change if necessary since it's not related to the fix.

Source/SwiftLintFramework/Rules/Style/NumberSeparatorRule.swift
daltonclaybrook: I added a comment. Let me know if this makes sense, or if you have a way which is more clear.

Source/SwiftLintFramework/Rules/Style/NumberSeparatorRule.swift
Dschee: No, it's fine. I just wanted to know the rationale behind the change but it's fine for me.

Source/SwiftLintFramework/Rules/Style/NumberSeparatorRule.swift
Dschee: Ahh, now I understand. I think the easiest way to make this clear would be simply to rename the genericly named var `idx` to something like `numerals`. Then I would have immediately understood even without the comment. I think, having a proper name makes the comment unecessary, too.

daltonclaybrook: Ah, okay. That makes sense and I completely agree. I'll update the PR.

Dschee:
Ah, one last thing before I merge this: The SwiftLintBot shows that this change makes many projects considerably slower, which makes sense given the fact, that we run the code for every numeral used in code. We should think about ways to make this faster again, but before doing that we should check that those numbers are actually reliable given that there was an Xcode update recently and that could be behind those changes, too.

@marcelofabri Is there a way we can re-run the SwiftLintBot manually on this PR to evaluate performance of changes? Also what is a significant perfmormance change in you opinion that would hold off from merging a bug fix?

daltonclaybrook:
@Dschee Looks like the performance warnings may have been a false-positive.

Dschee:
Nice, just took a while to update. Merging. 👏

marcelofabri:
> @marcelofabri Is there a way we can re-run the SwiftLintBot manually on this PR to evaluate performance of changes? Also what is a significant perfmormance change in you opinion that would hold off from merging a bug fix?

You should be able to retrigger it from Azure Pipelines. However, it's usually better to run locally to be sure and avoid any infrastructure issues (CI is shared, etc).

Source/SwiftLintFramework/Rules/Style/NumberSeparatorRuleExamples.swift
```diff
@@ -13,7 +13,9 @@ internal struct NumberSeparatorRuleExamples {
                 "let hex = \(sign)0xAA_BB",
                 "let octal = \(sign)0o21",
                 "let octal = \(sign)0o21_1",
-                "let exp = \(sign)1_000_000.000_000e2"
+                "let exp = \(sign)1_000_000.000_000e2",
+                "let foo: Double = \(sign)(200)",
+                "let foo: Double = \(sign)(200 / 447.214)"
```
marcelofabri: can we add a triggering example with parenthesis to document the behavior?

daltonclaybrook: Definitely. I'll make a separate PR.

PR #2704, opened 2019-04-02
Add triggering examples with parentheses to the number separator rule tests

daltonclaybrook:
Further improves #2683 

I did not create a new changelog entry for this PR because it should have been implemented as part of #2687, but I'll be happy to add one anyway if necessary.

Due to how the number is parsed, the violation can occur either inside or outside the opening parenthesis depending on the sign, so I could not simply append new values inside the existing `makeTriggeringExamples` function. Let me know if you think this should be handled differently.

### DIFF
--- a/Rules.md
+++ b/Rules.md
@@ -14010,6 +14010,42 @@ let foo = ↓1_000_000.000000_1
 let foo = ↓1000000.000000_1
 ```
 
+```swift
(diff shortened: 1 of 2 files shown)


daltonclaybrook:
Perhaps the rule itself should be improved such that the violation always occurs outside the parentheses. If so, a new issue should be created for that.

PR #2706, opened 2019-04-03
Return false for isFile for empty strings

keith:
As of Swift 5, corelibs-foundation crashes if you pass an empty string
to `fileExists` https://github.com/apple/swift-corelibs-foundation/blob/cae5eaca63cd57f64a93b4d5de87a4ffc04466d5/Foundation/FileManager.swift#L2033

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,11 @@
   [Taiki Komaba](https://github.com/r-plus)
   [#2700](https://github.com/realm/SwiftLint/issues/2700)
 
+* Fix crash when running on Linux with Swift 5 without specifying a `--path`
+  value or specifying an empty string.  
+  [Keith Smiley](https://github.com/keith)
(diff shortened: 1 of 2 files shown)


marcelofabri:
Why are we passing `""` in the first place? 🤔

keith:
It looks like that's the default if no path is passed https://github.com/realm/SwiftLint/blob/dd39c438e1f0d5fa1db84c74c83b528195e7f731/Source/swiftlint/Helpers/CommonOptions.swift#L4-L8

keith:
Maybe that default should be a `.` though?

marcelofabri:
> Maybe that default should be a `.` though?

We can investigate and change, but I'm a bit afraid because we don't have unit tests for the `swiftlint` target. IMO we can merge this as is and then follow up.

jpsim:
Fix looks good to me. Added a changelog entry in 022a217a. Will merge when CI passes.

PR #2707, opened 2019-04-07
Refactor `Configuration.groupFiles(...)` for style and performance

jpsim:
This is a small followup from #2648. Also fixes the changelog entry's location to put it under the right version section.

### DIFF
--- a/Source/swiftlint/Extensions/Configuration+CommandLine.swift
+++ b/Source/swiftlint/Extensions/Configuration+CommandLine.swift
@@ -66,20 +66,16 @@ extension Configuration {
         for file in files {
             // Files whose configuration specifies they should be excluded will be skipped
             let fileConfiguration = configuration(for: file)
-            let excludedPaths = fileConfiguration.excluded
-                .map { (fileConfiguration.rootPath ?? "").bridge().appendingPathComponent($0) }
-
-            let shouldSkip: Bool = excludedPaths.contains {
-                file.path?.bridge().pathComponents.starts(with: $0.bridge().pathComponents) ?? false
+            let fileConfigurationRootPath = (fileConfiguration.rootPath ?? "").bridge()
+            let shouldSkip = fileConfiguration.excluded.contains { excludedRelativePath in
+                let excludedPath = fileConfigurationRootPath.appendingPathComponent(excludedRelativePath)
+                let filePathComponents = file.path?.bridge().pathComponents ?? []
+                let excludedPathComponents = excludedPath.bridge().pathComponents
+                return filePathComponents.starts(with: excludedPathComponents)
             }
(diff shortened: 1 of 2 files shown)


jpsim:
@Bruschidy54 I'd appreciate it if you could sanity check this refactor to ensure it still fixes #2447 for you.

Source/swiftlint/Extensions/Configuration+CommandLine.swift
```diff
@@ -66,20 +66,16 @@ extension Configuration {
         for file in files {
             // Files whose configuration specifies they should be excluded will be skipped
             let fileConfiguration = configuration(for: file)
-            let excludedPaths = fileConfiguration.excluded
-                .map { (fileConfiguration.rootPath ?? "").bridge().appendingPathComponent($0) }
-
-            let shouldSkip: Bool = excludedPaths.contains {
-                file.path?.bridge().pathComponents.starts(with: $0.bridge().pathComponents) ?? false
+            let fileConfigurationRootPath = (fileConfiguration.rootPath ?? "").bridge()
```
jpsim: This will now only be bridged once instead of once per excluded entry.

Source/swiftlint/Extensions/Configuration+CommandLine.swift
```diff
@@ -66,20 +66,16 @@ extension Configuration {
         for file in files {
             // Files whose configuration specifies they should be excluded will be skipped
             let fileConfiguration = configuration(for: file)
-            let excludedPaths = fileConfiguration.excluded
-                .map { (fileConfiguration.rootPath ?? "").bridge().appendingPathComponent($0) }
```
jpsim: By folding this `map` into the `contains` check below, we avoid iterating and appending paths as soon as one would have been "skippable".

Source/swiftlint/Extensions/Configuration+CommandLine.swift
```diff
@@ -66,20 +66,16 @@ extension Configuration {
...
             }
 
             if !shouldSkip {
-                if var configuredFiles = groupedFiles[fileConfiguration] {
-                    configuredFiles.append(file)
-                    groupedFiles[fileConfiguration] = configuredFiles
-                } else {
-                    groupedFiles[fileConfiguration] = [file]
-                }
+                groupedFiles[fileConfiguration, default: []].append(file)
```
jpsim: Default subscripts available since Swift 4.0 makes this a bit nicer: https://swift.org/blog/dictionary-and-set-improvements/

Bruschidy54: oh cool didn't know this was a thing

Issue #2431, opened 2018-10-02
Designing a dead code detection rule that needs to analyze across files


elliottwilliams:
I'm looking into the whether it would be possible to create an analyzer rule to detect dead code in a project. It seems like this is totally doable with `CursorInfo` requests, but I'm at a loss for how a rule like this would integrate into SwiftLint.

For dead code detection to be useful, it has to work between files: it should build up a set of referenced declarations across a project, and then look at the declarations in a file to determine whether that declaration is ever referenced. This means that to lint any single file, the whole project (i.e. the other files given to SwiftLint) needs to be analyzed first.

Since SwiftLint rules are designed to run per-file, in isolation, I'm not sure how I would build this rule. Could it make a pass over every file to build a `references` set, and then run again on each file to perform actual linting? Should I write a separate script to build a `references` set, and pass it to SwiftLint via a command line argument? Are multi-file rules something SwiftLint wants to support in the future?

### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

kimdv:
There is one for private declarations. 
Maybe that can be extended ?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jpsim:
> I'm looking into the whether it would be possible to create an analyzer rule to detect dead code in a project.

Great! This has been next up on my list of analyzer rules I wanted to build. I'd love to support you in doing this in whatever way I can. But it's predicated on this next part:

> Are multi-file rules something SwiftLint wants to support in the future?

Yes, but this will need significant changes. I was thinking of having a "analyzer pass" model where the first part of a rule would collect information for later processing, then once all files had been processed, a final phase would generate violations from all the collected info.

Rough idea:

```swift
protocol AnalyzerPass {
    associatedtype FileInfo
    func collect(file: File) -> FileInfo
    func validate(fileInfo: [FileInfo]) -> [StyleViolation]
    func correct(fileInfo: [FileInfo]) -> [Correction]
}
```

For a dead code detection rule, its FileInfo could look like this:

```swift
struct USR {
    let identifier: String
}

struct DeadCodeFileInfo {
    let declaredUSRs: [String]
}
```

mikemee:
Side note, https://peripheryapp.com/, seems to provide this.

[...]

PR #2714, opened 2019-04-11
Support rules that collect data from all files before validating

elliottwilliams:
TODOs:
- [x] Write test cases for `CorrectingRule & CollectingRule`
- [x] Test `Configuration`'s ability to invoke the linter correctly
- [x] ~~Add documentation on how to write CollectingRules (should it be marked "experimental" like AnalyzerRule is?)~~  
    Deferring this until the first `CollectingRule` is written.
- [x] ~~Figure out a way to have non/triggering examples in a CollectingRule~~
    While it seems like we need to do this before we can ship a CollectingRule (so that SwiftLint can generate examples and easily verify the rule), I think it makes sense to do it separately to keep this PR from getting any larger.
---

Hi! This is a first pass at adding a `CollectingRule` type, which lays the groundwork for supporting new rules like dead code detection (#2431). **`CollectingRule` allows rules to perform a "collection step" across all files before any file is validated**. It's a substantial change to some core infrastructure of SwiftLint, but it centers around this protocol:

```swift
public protocol CollectingRule: Rule {
    // the extra stuff your rule keeps track of
    associatedtype FileInfo

    // returns info to store about the given file. called for all files before any file is validated.
    func collect(infoFor file: File, compilerArguments: [String]) -> FileInfo
    func collect(infoFor file: File) -> FileInfo

    // computes violations for a file using a dictionary of captured info for all files
    func validate(file: File, collectedInfo: [File: FileInfo], compilerArguments: [String]) -> [StyleViolation]
    func validate(file: File, collectedInfo: [File: FileInfo]) -> [StyleViolation]
}
```

`CollectingRule` pairs well with `AnalyzerRule` because access to USRs makes it possible to identify references to the same code in different files, but there's nothing preventing a `CollectingRule` from being written that doesn't take compiler arguments.

## Storage

A `CollectingRule` shares arbitrary data between instances of the rule, so we need a container to pass around the info that's collected. I've added a `RuleStorage` type which can store and retreive collected info:

```swift
public struct RuleStorage {
    mutating func collect<R: CollectingRule>(info: R.FileInfo, for file: File, in rule: R)
    func collectedInfo<R: CollectingRule>(for rule: R) -> [File: R.FileInfo]
}
```

All rules have protocol methods which take a `RuleStorage` and invoke the appropriate method on the rule:
```swift
public protocol Rule {
    // These are called by the linter and are always implemented in extensions.
    func collect(infoFor file: File, into storage: inout RuleStorage, compilerArguments: [String])
    func validate(file: File, using storage: RuleStorage, compilerArguments: [String]) -> [StyleViolation]
}
```

## How collection is invoked

SwiftLint checks a file for violations by creating a `Linter` for that file. I moved most of `Linter`'s functionality into a new type, `CollectedLinter`. In its place, `Linter` now exposes a `collect` method which collects a file and returns a `CollectedLinter` for further analysis.

```swift
public struct Linter {
    public func collect(into storage: inout RuleStorage) -> CollectedLinter
}
public struct CollectedLinter {
    public func styleViolations(using storage: RuleStorage) -> [StyleViolation]
}
```

When `Configuration.visit(filesPerConfiguration:visitor:storage:)` is invoked to process each file and compute violations, the visitor now makes two passes: once to collect info and transform each `Linter` into a `CollectedLinter`, and again to get violations from the `CollectedLinter`s.

## Meta

I've been in discussions how SwiftLint can support behavior like `CollectingRule` before, but I know this is an awfully large diff for my first PR 😬. If there's anything I can do to break this up, revisit some of the design, or handle new edge cases, I'd be happy to!


jpsim:
Thanks so much for doing this! I haven't looked at it yet, but the ability to collect information from all lintable files will pave the way to extend the `unused_private_declaration` rule to also catch `internal`, `public` and `open` declarations (for projects with a finite set of consumers at least, like iOS apps).

It looks like it's crashing, however. @elliottwilliams do you expect to be able to get this passing on CI soon?

elliottwilliams:
@jpsim sure thing — looks like a simultaneous access conflict and a typo in tests. I'll try to get it passing over the next few days and resolve some of the outstanding things I mentioned above 🙂

jpsim:
Thanks for pushing this forward, @elliottwilliams!

jpsim:
Just going to close and reopen to see if that kicks off CI for this again (probably won't).

jpsim:
Hmm, can't seem to get CI to run on this.

jpsim:
Opened #2743 to check CI.

jpsim:
Looks like there are some TSan races detected and Linux compilation failures: https://dev.azure.com/jpsim/SwiftLint/_build/results?buildId=598

elliottwilliams:
@jpsim 
> Looks like there are some TSan races detected and Linux compilation failures

Linux compiler errors look like me forgetting to put a return in the `autoreleasepool` shim 🙄 and analyzer failure looks like a style violation, so those should be easy to fix. I'll dive in to the TSan failure, at first glance it looks similar to the exclusivity enforcement error I addressed in [`
cf9f49d`](https://github.com/realm/SwiftLint/pull/2714/commits/cf9f49d16c3aafef3559143c6d9159ff7102db87) so perhaps there's more work to be done there.

CHANGELOG.md
```diff
@@ -10,6 +10,14 @@
 
 #### Enhancements
 
+* Add a two-stage `CollectingRule` protocol to support rules that collect data
```
jpsim: Maybe we put this in the `Experimental` section above?

Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
```diff
@@ -24,7 +24,7 @@ extension Array where Element: Equatable {
     }
 }
 
-extension Array {
+public extension Array {
```
jpsim: I'm not thrilled about exposing these utilities as part of SwiftLintFramework's public API. Can we instead create a simlink to this file in `Source/swiftlint/...` so we can use it from both targets without polluting our public API?

Source/SwiftLintFramework/Models/RuleStorage.swift
```diff
@@ -0,0 +1,42 @@
...
+
+    func collectedInfo<R: CollectingRule>(for rule: R) -> [File: R.FileInfo] {
+        return access.sync {
+            // swiftlint:disable:next force_cast
+            storage[ObjectIdentifier(R.self)] as! [File: R.FileInfo]
+        }
+    }
+}
+
+extension File: Hashable {
```
jpsim: I'd rather we did this at File's implementation instead, in SourceKittenFramework.

Source/swiftlint/Extensions/Configuration+CommandLine.swift
```diff
@@ -45,14 +45,15 @@ private func scriptInputFiles() -> Result<[File], CommandantError<()>> {
 }
 
 #if os(Linux)
-private func autoreleasepool(block: () -> Void) { block() }
+private func autoreleasepool<V>(block: () -> V) -> V { return block() }
```
jpsim: Nit: `T` is the standard generic type parameter name, not `V`. Especially since now this no longer necessarily means "Void".

Source/swiftlint/Extensions/Configuration+CommandLine.swift
```diff
@@ -82,22 +83,27 @@ extension Configuration {
         return .success(groupedFiles)
     }
 
+    // swiftlint:disable:next function_body_length
```
jpsim: This function was already pretty hard to follow, now it's nearly impossible. Let's find a better way to refactor this.

Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
elliottwilliams: I didn't think of symlinking here—sounds like a much better alternative.

elliottwilliams:
OSSCheck timeout looks weird. I'm running danger locally and will see if that gives any insight.

elliottwilliams:
Okay, so I found that oss-check was causing SwiftLint to [deadlock in a call to `write`][1], but that I couldn't reproduce the deadlock if I manually ran it on the repos oss-check clones. I traced this back to oss-check using `popen3` without flushing stderr. [Ruby's docs mention][2] that popen uses fixed-size buffers, and that not flushing stderr can deadlock if the program fills the buffer:

> You should be careful to avoid deadlocks. Since pipes are fixed length buffers, ::popen3(“prog”) {|i, o, e, t| o.read } deadlocks if the program generates too much output on stderr. You should read stdout and stderr simultaneously (using threads or IO.select). However, if you don't need stderr output, you can use ::popen2. 

My guess is that because the collection stage prints one line per file, this PR was causing SwiftLint to fill up the buffer and break oss-check.

I committed the fix to oss-check, but if we don't want to do that right now, perhaps we could prevent SwiftLint from printing collection messages.

[1]: https://gist.github.com/elliottwilliams/240d9b66a15fc21fb48380009741e156
[2]: http://ruby-doc.org/stdlib-2.6.3/libdoc/open3/rdoc/Open3.html#method-c-popen3

jpsim:
Whoa, thanks so much for digging deeply into that and fixing it!

I’m looking forward to reviewing this when I get some computer time. Do you have a work-in-progress branch somewhere with a rule making use of this change? It would be helpful in validating that it works as expected, even if the rule itself isn’t ready for review.

elliottwilliams:
> Do you have a work-in-progress branch somewhere with a rule making use of this change? It would be helpful in validating that it works as expected, even if the rule itself isn’t ready for review.

@jpsim not immediately, but I’ll see if I can put one together in the next few days. It might be cool to start working on an unused declaration rule.

The closest thing I have is `CollectingRuleTests`, which has some bare-bones rules that ensure the protocol methods are implemented correctly, the storage is working, etc.

CHANGELOG.md
```diff
@@ -6,7 +6,13 @@
 
 #### Experimental
 
-* None.
+* Add a two-stage `CollectingRule` protocol to support rules that collect data
+  from all files before validating. Collecting rules implement a `collect`
+  method which is called once for every file, before _any_ file is checked for
+  violations. By collecting, rules can be written which validate across
+  multiple files for things like unused declarations.
```
jpsim: Needs two trailing spaces here: https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes

```suggestion
  multiple files for things like unused declarations.  
```

Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
```diff
@@ -45,11 +45,11 @@ extension Array {
             return (copy[0..<pivot], copy[pivot..<count])
     }
 
-    func parallelFlatMap<T>(transform: @escaping ((Element) -> [T])) -> [T] {
+    func parallelFlatMap<T>(transform: (Element) -> [T]) -> [T] {
```
jpsim: I was about to say this is wrong, but then after thinking about it some more realized that it was _previously_ wrong! Thanks! 😅

Source/SwiftLintFramework/Extensions/File+Hashable.swift
```diff
@@ -0,0 +1,18 @@
+import SourceKittenFramework
+
+extension File: Hashable {
```
jpsim: We should do this in SourceKittenFramework itself.

Source/SwiftLintFramework/Models/Linter.swift
```diff
@@ -108,59 +109,101 @@ private extension Rule {
...
+
+    public func collect(into storage: RuleStorage) -> CollectedLinter {
+        DispatchQueue.concurrentPerform(iterations: rules.count) { idx in
+            rules[idx].collect(infoFor: file, into: storage, compilerArguments: compilerArguments)
+        }
+        return CollectedLinter(from: self)
+    }
+}
+
+public struct CollectedLinter {
```
jpsim: What do you think of calling this a "CollectingLinter" instead?

Source/SwiftLintFramework/Models/Linter.swift
```diff
@@ -108,59 +109,101 @@ private extension Rule {
...
-                    configuration: self.configuration,
-                    superfluousDisableCommandRule: superfluousDisableCommandRule,
-                    compilerArguments: self.compilerArguments)
-        }
-        let undefinedSuperfluousCommandViolations = self.undefinedSuperfluousCommandViolations(
-            regions: regions, configuration: configuration,
-            superfluousDisableCommandRule: superfluousDisableCommandRule)
+    private func getStyleViolations(using storage: RuleStorage, benchmark: Bool = false)
+        -> ([StyleViolation], [(id: String, time: Double)]) {
+            if let cached = cachedStyleViolations(benchmark: benchmark) {
```
jpsim: The entire body of this function seems over-indented by one level now.

Source/SwiftLintFramework/Models/RuleStorage.swift
```diff
@@ -0,0 +1,25 @@
...
+        let key = ObjectIdentifier(R.self)
+        access.sync(flags: .barrier) {
+            storage[key, default: [:]][file] = info
+        }
+    }
+
+    func collectedInfo<R: CollectingRule>(for rule: R) -> [File: R.FileInfo] {
+        return access.sync {
+            // swiftlint:disable:next force_cast
+            storage[ObjectIdentifier(R.self)] as! [File: R.FileInfo]
```
jpsim: Can you explain to me why this is a safe use of force casting? If we fail to run `collect` with a rule, then call `collectedInfo` for that rule, this will crash, without printing a nice error message to the user. They'll just see a segfault.

Why not default to an empty dictionary?

```suggestion
            storage[ObjectIdentifier(R.self)] as? [File: R.FileInfo] ?? [:]
```

Source/SwiftLintFramework/Protocols/Rule.swift
```diff
@@ -11,9 +11,17 @@ public protocol Rule {
     func validate(file: File, compilerArguments: [String]) -> [StyleViolation]
     func validate(file: File) -> [StyleViolation]
     func isEqualTo(_ rule: Rule) -> Bool
+
+    // These are called by the linter and are always implemented in extensions.
+    func collect(infoFor file: File, into storage: RuleStorage, compilerArguments: [String])
```
jpsim: Nit: I think this should be called `collectInfo(for:into:compilerArguments:)`

```suggestion
    func collectInfo(for file: File, into storage: RuleStorage, compilerArguments: [String])
```

Source/SwiftLintFramework/Extensions/File+Hashable.swift
jpsim: I liked your implementation so pasted it verbatim in a PR towards SourceKitten and credited you in the changelog: https://github.com/jpsim/SourceKitten/pull/594

Thanks!

jpsim: And a new release has been cut: https://github.com/jpsim/SourceKitten/releases/tag/0.23.2

Source/SwiftLintFramework/Models/Linter.swift
elliottwilliams: Hmm, I'm not married to this naming convention, but I think that calling this type a `CollectingLinter` would imply that _it_ does the collecting, when in actually `Linter` is what runs collection. Linter and CollectedLinter are related by the `collect` function, i.e.

```swift
Linter.collect(into storage: RuleStorage) -> CollectedLinter
```

What if we renamed Linter to `CollectingLinter` and CollectedLinter to `Linter`? Then the relation would be
```swift
CollectingLinter.collect(into storage: RuleStorage) -> Linter
```

Or we could stop insisting that both types are a "linter", and name them something like
```swift
Collecter.collect(into storage: RuleStorage) -> Linter
```

jpsim: Oh I see. I think "Collected Linter" makes sense, but this could use a docstring to explain that.

Source/SwiftLintFramework/Models/RuleStorage.swift
elliottwilliams: IIRC, I wanted to crash if the collected info for a file was being read from before it was written to. This should never happen and indicates that something's gone wrong internally.

But looking at it now, the force-cast doesn't make that assertion clear. If you agree that crashing is here okay, I can make this return value optional and put a precondition in `CollectingRule.validate(file:using:compilerArguments:)`. If you want to avoid a crash, defaulting to an empty dictionary makes sense to me 🙂

jpsim: A precondition or assertion failure makes sense to me. I'd just rather have informative crashes 😄

Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
elliottwilliams: Hah, no worries 😁 Yeah if I understand closure allocation correctly this should improve performance / reduce copying of the linters that get passed around.

Source/SwiftLintFramework/Models/RuleStorage.swift
elliottwilliams: Sounds good! Done in https://github.com/realm/SwiftLint/pull/2714/commits/7bd359e69f8b969921123590c0cd1cb9e84ba567

elliottwilliams:
Thanks for your help with this! I don't know if it's useful, but I put together a table when I was working on the new `Rule` methods to help me visualize which methods needed to be implemented under which circumstances: https://gist.github.com/elliottwilliams/e68c2e2c6c8ced58d19f617bb10a3dd2

Would be happy to drop this into a comment in Rule.swift.

jpsim:
Hey @elliottwilliams I hope I didn't scare you off with my last comment. You don't need to answer all those questions, I can run with some of them if you say the word.

elliottwilliams:
Heh, no worries, I've been busy the past two weekends but will have some time this week to push things forward.

> Right now, collecting rules basically need to be analyzer rules. Files are skipped from being collected if there are no compiler arguments found for it for example. I could see non-analyzer collecting rules as being useful, so ideally we remove this restriction.

Unless I'm missing something, I intended for `CollectingRule` to be distinct from `CollectingRule & AnalyzerRule` to support this. For example, [this rule](https://github.com/realm/SwiftLint/pull/2714/files#diff-120d19e37bc577721ba47d25db67cd25R9) in CollectingRuleTests is non-analyzing.

> New logs are being added "Collecting 'file...'" even if no collecting rules are being run. Maybe we should avoid the collection altogether if there are no collecting rules to run?

Yeah, I was thinking about this as well. Would you prefer it in a followup PR, or should I come up with a way to do it now?

> Rather than make a symlink to share Array+SwiftLint.swift across both the SwiftLintFramework and swiftlint targets, we could make a new Utility library target. Slightly less "hacky" but it doesn't particularly bother me.

I'd be happy to do this. Do you think it makes sense to move every Swift/SourceKitten extension file into a `Utility` target?

> The public API for Linter is changing with this PR, which is technically a breaking change, although I'm still unaware of any external SwiftLintFramework consumers to this day, so it may not matter much in practice.
> ...
> I wonder if we should pause on merging this until we have a first CollectingRule built. What if we realize that we need some changes to the APIs as we're writing the first rule to make use of this? Not a big deal given that this is marked as experimental, but more for the breaking public API changes that are being introduced here.

That makes sense to me—I'm fine basing a future CollectingRule PR off of this one until it's time to merge. It seems like whatever API changes we make, we should try to finalize them by the next release.

jpsim:
> Unless I'm missing something, I intended for `CollectingRule` to be distinct from `CollectingRule & AnalyzerRule` to support this. For example, [this rule](https://github.com/realm/SwiftLint/pull/2714/files#diff-120d19e37bc577721ba47d25db67cd25R9) in CollectingRuleTests is non-analyzing.

I may be misreading the code, but it looks like this part here requires compiler arguments:

image

> Yeah, I was thinking about this as well. Would you prefer it in a followup PR, or should I come up with a way to do it now?

The issue comes if we merge this as-is, and you get busy so we end up shipping a release with different logging that isn't actually representative of what's happening, etc. So yes, I do think we'll need to tackle that before merging this PR, but again I offer my help because I don't want to unnecessarily burden you considering how much work you've already put into this.

> I'd be happy to do this. Do you think it makes sense to move every Swift/SourceKitten extension file into a `Utility` target?

No, I think just things that are `public` but only meant to be shared from the `swiftlint` target rather than any consumer of SwiftLintFramework. This could start really small by just having the current symlink. I don't see this as a blocker for merging this PR though.

elliottwilliams:
> I may be misreading the code, but it looks like this part here requires compiler arguments

Oh, I think I see the confusion: `LintableFilesVisitor.shouldSkipFile(atPath:)` only returns true if it's in analyzer mode and the path given can't be found in the compiler args. It always returns false for non-analyzer rules. I moved this check around but didn't change the actual logic in it; here's the same check [in master](https://github.com/realm/SwiftLint/blob/a4e62d569237019af8b60ef393418c62a0920abb/Source/swiftlint/Extensions/Configuration%2BCommandLine.swift#L90-L100). It might be worth renaming this method to `shouldSkipAnalyzingFile` to clarify.

> So yes, I do think we'll need to tackle that before merging this PR, but again I offer my help because I don't want to unnecessarily burden you considering how much work you've already put into this.

Totally understand (and thank you!). I had an idea on how to approach this, so I figured I'd keep the ball rolling and pushed https://github.com/realm/SwiftLint/pull/2714/commits/e8eee47f66c6e25915a5583594091a192ccef2fa.

jpsim:
Ran OSSCheck locally and saw a 2%-15% speed improvement compared to `master`. Not sure why this would be any faster, but I'm not concerned about this change being _slower_ at least.

elliottwilliams:
Nice! 

> I wonder if we should pause on merging this until we have a first CollectingRule built. What if we realize that we need some changes to the APIs as we're writing the first rule to make use of this? Not a big deal given that this is marked as experimental, but more for the breaking public API changes that are being introduced here.

@jpsim anything besides this blocking this PR? I'm not sure I have the bandwidth to draft up a CollectingRule right now (as evidenced by this PR sitting idle 😬). Maybe it'd be good to try to come up with a really simple CollectingRule to start with?

jpsim:
> anything besides this blocking this PR?

Nope, merging now 😄 

> Maybe it'd be good to try to come up with a really simple CollectingRule to start with?

I'd like to introduce you to my good friend Unused Declaration Rule: #2814

PR #2718, opened 2019-04-12
Add nested type handling specification to file_name rule

### LINKED ISSUE #2717: Add nested type handling specification to file_name rule
Assuming I have a file that extends `MyType.MySubtype` like this:

```
extension MyType.MySubtype {
    // ...
}
```

I may call this file `MyType.MySubtype+Extension.swift` without triggering the `file_name` rule. I may also specify a custom file name suffix, so that I could name it `MyType.MySubtypeExtension.swift`.

In contrast, I cannot call my file `MyTypeMySubtypeExtension.swift` without triggering a warning. As such a naming (without a second `.` character in the file naming) may be preferred by some, I propose to extend the `file_name` configurability to allow for custom nested type separators (with `.` remaining the default).

fredpi:
Closes #2717.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@
     swift test --filter testSimulateHomebrewTest
     ```  
   [Norio Nomura](https://github.com/norio-nomura)
+  
+* Add option to configure how nested types should be separated in file names by
+  introducting `nested_type_separator` configuration for the `file_name` rule.  
+  [Frederick Pietschmann](https://github.com/fredpi)
+  [#2717](https://github.com/realm/SwiftLint/issues/2717)
 
 #### Bug Fixes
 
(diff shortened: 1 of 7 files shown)


fredpi:
Any thoughts on this? ;)

jpsim:
Thanks for the PR! What are your thoughts on just allowing both dots and no dots without necessarily adding a configuration option for it?

IMHO that would follow the mantra of "don't let your tools get in your way" a lot better than requiring a strict consistent format for nested type extensions.

fredpi:
@jpsim This was my first thought as well, but then I figured that one could already exactly specify prefix and suffix patterns, so I thought making the nested type separator just as configurable would make sense (considering that some may also prefer other separators than ` ` or `.` (e. g. `_`)).

However, I don't care very much about this myself, I just wanted to follow the existing pattern of configurability. If you prefer just allowing certain characters (dot / underscore / no char), I'd happily adjust the PR accordingly. 😉

jpsim:
Good point on aligning with the other configurability options. Thanks for the PR @fredpi!

PR #2719, opened 2019-04-14
Add unused_capture_list rule

### LINKED ISSUE #2715: Rule Request: unused value in capture group
### New Issue Checklist

### New rule request

Similar to unused parameter in closure.

```swift
// Triggering
{ [foo] in _ }()

// Non- Triggering
{ [foo] in foo.bar() }()
```

Only severity.

Enabled by default, same as `unused_closure_parameter`.

daltonclaybrook:
Resolves #2715 

I'm using an `ASTRule` in order to find closures, but I was unable to find any first-class parsing support within SourceKitten for capture lists, so I am doing this with regex and string manipulation. Let me know if I missed something.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/UnusedCaptureListRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/UnusedCaptureListRule.swift
@@ -71,26 +71,8 @@ public struct UnusedCaptureListRule: ASTRule, ConfigurationProviderRule, Automat
             .NSRangeToByteRange(start: restOfClosureRange.location, length: restOfClosureRange.length)
             else { return [] }
 
-        let tokens = file.syntaxMap.tokens(inByteRange: restOfClosureByteRange)
-        let identifiers = tokens
-            .compactMap { token -> String? in
-                guard token.type == SyntaxKind.identifier.rawValue || token.type == SyntaxKind.keyword.rawValue,
-                    let range = contents.byteRangeToNSRange(start: token.offset, length: token.length)
-                    else { return nil }
-                return contents.substring(with: range)
-            }
-
-        return references.compactMap { reference, location -> StyleViolation? in
-            guard !identifiers.contains(reference) else { return nil }
-            let offset = captureListRange.location + location
-            let reason = "Unused reference \(reference) in a capture list should be removed."
-            return StyleViolation(
-                ruleDescription: type(of: self).description,
-                severity: configuration.severity,
-                location: Location(file: file, characterOffset: offset),
-                reason: reason
-            )
-        }
+        let identifiers = identifierStrings(in: file, byteRange: restOfClosureByteRange)
+        return violations(in: file, references: references, identifiers: identifiers, captureListRange: captureListRange)
     }
 
     // MARK: - Private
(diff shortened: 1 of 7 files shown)


Source/SwiftLintFramework/Rules/Lint/UnusedCaptureListRule.swift
```diff
@@ -0,0 +1,128 @@
...
+            """
+            let failure: Failure = { [↓weak self, ↓unowned delegate = self.delegate!] foo in
+                print(foo)
+            }
+            """,
+            "{ [↓foo] in _ }()"
+        ]
+    )
+
+    private let captureListRegex = regex("^\\{\\h*\\[([^\\]]+)\\].*\\bin\\b")
```
marcelofabri: any particular reason to use `\h` here? although not common, the capture list could be in a new line

Source/SwiftLintFramework/Rules/Lint/UnusedCaptureListRule.swift
```diff
@@ -0,0 +1,128 @@
...
+public struct UnusedCaptureListRule: ASTRule, ConfigurationProviderRule, AutomaticTestableRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static var description = RuleDescription(
+        identifier: "unused_capture_list",
+        name: "Unused Capture List",
+        description: "Unused reference in a capture list should be removed.",
+        kind: .lint,
```
marcelofabri: can you add `swiftVersion: .fourDotTwo`? `SwiftSyntaxKind.closure` is only returned by SourceKit in 4.2+

Source/SwiftLintFramework/Rules/Lint/UnusedCaptureListRule.swift
```diff
@@ -0,0 +1,128 @@
...
+                    .trimmingCharacters(in: .whitespacesAndNewlines)
+                    .components(separatedBy: .whitespaces)
+                    .last
+                if let reference = reference {
+                    referencesAndLocations.append((reference, location))
+                }
+            }
+    }
+
+    private func identifierStrings(in file: File, byteRange: NSRange) -> [String] {
```
marcelofabri: nitpick:: this could return a `Set<String>`

marcelofabri:
> I'm using an ASTRule in order to find closures, but I was unable to find any first-class parsing support within SourceKitten for capture lists

They're reported as local variables inside the closure, but you'd need to be creative to figure out if it's a captured variable (i.e. check if the location is before the `in` token location).

Source/SwiftLintFramework/Rules/Lint/UnusedCaptureListRule.swift
daltonclaybrook: ah, good point.

jpsim:
Thanks for the contribution @daltonclaybrook, and this is indeed a useful rule, although there are some false positives (3/4 of the violations reported by OSSCheck are false positives). If we can't avoid those false positives, I think we should make this rule opt-in (disabled by default).

daltonclaybrook:
@jpsim oh whoops, I guess I missed those warnings. I may have some time this afternoon to take a look at this.

PR #2723, opened 2019-04-19
Add No Return Void Style Rule

johnnysparks:
Adding a new `No Return Void` rule. This rule issues a warning when adding functional expressions after a function's return statement for Void functions.


Rules.md
```diff
@@ -13515,6 +13516,104 @@ extension External { struct Gotcha {}}
 
 
 
+## No Guard Return Void
+
+Identifier | Enabled by default | Supports autocorrection | Kind | Analyzer | Minimum Swift Compiler Version
+--- | --- | --- | --- | --- | ---
+`no_guard_return_void` | Enabled | No | style | No | 3.0.0
```
johnnysparks: Not sure how to disable by default.

Source/SwiftLintFramework/Rules/Style/NoGuardReturnVoidRule.swift
```diff
@@ -0,0 +1,151 @@
...
+
+public struct NoGuardReturnVoidRule: ASTRule, ConfigurationProviderRule, AutomaticTestableRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "no_guard_return_void",
+        name: "No Guard Return Void",
+        description: "Guard statements in void functions should not have a statement after the return.",
```
johnnysparks:
```suggestion
        description: "Guard statements in void functions should not have a statement after the return, add them to the previous line.",
```

Rules.md
johnnysparks: Got it, `OptInRule`

Source/SwiftLintFramework/Rules/Style/NoGuardReturnVoidRule.swift
```diff
@@ -0,0 +1,196 @@
...
+
+    var isKeyword: Bool {
+        return self.type == "source.lang.swift.syntaxtype.keyword"
+    }
+
+    func isReturnKeyword(in file: File) -> Bool {
+        return self.isKeyword && (file.contents(for: self) == "return")
+    }
+}
+
```
jpsim: Oops, you now have _two_ trailing newlines here. Unix standard is a single one: https://unix.stackexchange.com/a/18789/243788

jpsim:
This is off to a great start @johnnysparks! My main concern is if we want to make this more generic in the near future, like preventing _all_ returns of void expressions, no matter where the context, that the rule name will no longer apply.

I'd rather have a generic rule name that's only applied _some_ of the time than an accurate rule name that will need to be deprecated after a release or two.

Source/SwiftLintFramework/Rules/Style/NoGuardReturnVoidRule.swift
johnnysparks: Ah, I was overshooting instead of undershooting. 🙄Thanks!

johnnysparks:
Ah, that makes sense. We'd also want it to fire in breaks/if statements etc. Like:

```swift
func foo() {}
func bar() {}
func baz() {}

func test() {
  if x {
    return↓ foo()
  }
  bar()
}

func test() {
  switch x {
    case .a:
      return↓ foo() // return to skip baz()
    case .b:
      bar()
  }
  baz()
}

func test() {
  guard x else {
    return↓ foo()
  }
  bar()
}
```

jpsim:
Or at the complete end of the scope too:

```swift
func foo() {}

func test() {
  if x {
    return
  }

  ↓return foo()
}
```

jpsim:
To be clear, if you don't have time to extend this to catch the other use cases, that's fine, as long as the rule is named generically enough to allow catching those in the future.

jpsim:
FYI we disallow merging in PRs on this project, so you’ll need to rebase to avoid that error.

CHANGELOG.md
```diff
@@ -51,6 +56,7 @@
 
 #### Enhancements
 
+
```
jpsim: Nit: random newline

Source/SwiftLintFramework/Rules/Style/NoReturnVoidRule.swift
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        let statementRanges = dictionary.statementRanges()
+        var bodyRangeAfterLastStatement = bodyRange
+
+        if let lastStatement = statementRanges?.last {
+            bodyRangeAfterLastStatement.location = lastStatement.upperBound
+        }
+
+        let ranges: [NSRange] = [bodyRangeAfterLastStatement] + (statementRanges ?? [])
```
jpsim: Suggested changes:

```swift
let ranges = [bodyRangeAfterLastStatement] + (statementRanges ?? [])
return ranges.compactMap { range in
    let nonCommentokens = file.syntaxMap.tokens.lazy
        .filter { range.contains($0.offset) && !$0.isComment }
    guard let last = nonCommentokens.last,
        let lastReturn = nonCommentokens.last(where: { $0.isReturnKeyword(in: file) }),
        last != lastReturn
    else {
        return nil
    }

    let location = Location(file: file, byteOffset: lastReturn.offset)
    return StyleViolation(ruleDescription: type(of: self).description,
                          location: location)
}
```

This should be more efficient by virtue of not mapping then compact mapping, using `lazy` to avoid computing all filtered tokens that aren't comments since we just care about the last one(s).

It also changes the violation location to be at the beginning of the `return` keyword, which looks a bit nicer IMHO.

jpsim:
Looks like there are quite a few false positives reported by OSSCheck.

Source/SwiftLintFramework/Rules/Style/NoReturnVoidRule.swift
johnnysparks: Love it! Yeah looks like the rule was getting a little slow. 👍

marcelofabri:
Closing this in favor of https://github.com/realm/SwiftLint/pull/3054. Thanks @johnnysparks for working on this, the examples were really helpful!

PR #2726, opened 2019-04-23
Fix false positive in UnusedCaptureListRule

daltonclaybrook:
Fixes concern raised by @jpsim [here](https://github.com/realm/SwiftLint/pull/2719#issuecomment-485172561).

An array literal was being mistaken for a capture list in some cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,11 @@
   [Frederick Pietschmann](https://github.com/fredpi)
   [#2670](https://github.com/realm/SwiftLint/issues/2670)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -95,6 +95,11 @@
   [Frederick Pietschmann](https://github.com/fredpi)
   [#2670](https://github.com/realm/SwiftLint/issues/2670)
 
+* Fix false positives on `unused_capture_list` which resulted in an array
```
jpsim: Changelog entry isn't necessary because this rule hasn't shipped in a release yet.

PR #2730, opened 2019-04-29
Mark Swift 5 as supported

jpsim:


jpsim:
Needs https://github.com/jpsim/SourceKitten/pull/588

jpsim:
`testSimulateHomebrewTest` is flaky, going to ignore.

norio-nomura:
> `testSimulateHomebrewTest` is flaky, going to ignore.

Apparently there seems to be an unknown condition that SourceKit will not crash even with sandbox. :thinking:
It seems better to remove `testSimulateHomebrewTest`. To confirm that sandbox crash avoidance works, leaving `testSimulateHomebrewTestWithDisableSourceKit` is sufficient.

norio-nomura:
I'll open a PR that makes `testSimulateHomebrewTest` as opt-in test by environment variable.

Issue #2608, opened 2019-01-28
Ability to add SwiftLint Preemptively before .Swift files have been added 


christoff-1992:
### New Issue Checklist

- [X] Updated SwiftLint to the latest version
- [X] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug
If no .Swift files have been added to the directory then the build will fail with 'No lintable files found at paths'.

This is the correct behaviour currently but it would be nice to add the configuration to a lib/framework before any swift files have been added to the project.

At the moment I am looking at creating a shared framework that builds across iOS, OSX and TVOS. The workspace will use a shared structure for the source files and tests but I would also like to prepare the individual targets for each iOS, OSX and TVOS for SwiftLint without the build failing.

I can edit the run script to check for this manually but wondered if there is a better way around this.

```
swiftFileCount=`ls -1 *.swift 2>/dev/null | wc -l`
if [ $swiftFileCount != 0 ]
then
    ${PODS_ROOT}/SwiftLint/swiftlint
else
    echo "No lintable files to check, skipping Swiftlint"
fi
```

jpsim:
I think there's a broader choice to discuss which is wether or not we should treat linting no Swift files as an error or not. See #2619 where this is also being discussed but for other reasons. /cc @fabb 

The argument for exiting with an error code is that if you're expecting to lint a bunch of files and SwiftLint is misconfigured, it's a bit misleading to say "all good". Open to discussing this though since I think the points here and in #2619 are valid.

fabb:
I would be happy if there just was a separate exit code for the „no files“ case.

christoff-1992:
A specific config entry for `allow_zero_lintable_files: true` would be a conscious decision by the user implementing the config. You could always provide an exit code with a message of `Lintable files found at paths: '', when zero lintable files are expected by the current configuration` as an additional safety check to ensure that Swiftlint has not been misconfigured.

thedavidharris:
This is popping up as an issue when using a CI check like Danger, where editing only files that are covered by the ignore triggers a failure (https://github.com/danger/swift/issues/231).

`allow_zero_lintable_files: true` would make sense to me.

thedavidharris:
I created a draft PR at https://github.com/realm/SwiftLint/pull/2732 to see if it's satisfactory to get around this

teameh:
[...]

PR #2732, opened 2019-04-29
Add option to pass successfully if no files passed to SwiftLint are lintable

thedavidharris:
Addresses https://github.com/realm/SwiftLint/issues/2608.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@
 * None.
 
 #### Enhancements
+* Add option to pass successfully if no files passed to SwiftLint are lintable.  
+  [thedavidharris](https://github.com/thedavidharris)
+  [#2608](https://github.com/realm/SwiftLint/issues/2608)i
 
 * Add `deinitializer` type content to `type_contents_order` rule instead of
   grouping it with initializers.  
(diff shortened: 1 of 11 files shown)


dharr419-ford:
Opening this up for review; if this is worth an integration test can add, but that seems to be the only way to test this the way things are structured

thedavidharris:
Not totally sure the best way to add tests to this, since it appears that most config-level things are integration tested and I'm not sure I want to mess with those just to test this early-exit case.

jpsim:
Sorry for the delay in reviewing this @thedavidharris. Instead of compiler arguments, could this be done as a Configuration file setting? My reasoning is that I believe well-configured projects with no violations should pass with a plain `swiftlint` invocation.

You can look at several of the other top-level keys in [Configuration+Parsing.swift](https://github.com/realm/SwiftLint/blob/0.38.1/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift#L9) for examples of how to wire them in.

thedavidharris:
@jpsim yep makes sense, will have a go at it this week

thedavidharris:
Should be all updated now with the right docs and as a config param

Fab1n:
Is there a reason why this flag is documented nowhere and you need to really dig in to find out that this option exists?
@thedavidharris

vakhidbetrakhmadov:
Hi, `allow_zero_lintable_files: true` doesn't work when added to a parent config that's loaded by a child config.
It only work if `allow_zero_lintable_files: true` is specified in the child config.
Is this is a know issue ?

https://github.com/realm/SwiftLint/issues/6240

PR #2733, opened 2019-04-29
Don’t trigger redundant_void_return when using subscripts

marcelofabri:


jpsim:
I think `testSimulateHomebrewTest` is flaky 🤔

PR #2734, opened 2019-04-29
Make `testSimulateHomebrewTest()` test opt-in

norio-nomura:
because it may fail on unknown condition.
Set `SWIFTLINT_FRAMEWORK_TEST_ENABLE_SIMULATE_HOMEBREW_TEST` environment variable to test like following:
```terminal.sh-session
$ SWIFTLINT_FRAMEWORK_TEST_ENABLE_SIMULATE_HOMEBREW_TEST=1 \
swift test --filter testSimulateHomebrewTest
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,14 @@
 
 #### Enhancements
 
-* None.
+* Make `testSimulateHomebrewTest()` test opt-in because it may fail on unknown
+  condition. Set `SWIFTLINT_FRAMEWORK_TEST_ENABLE_SIMULATE_HOMEBREW_TEST` 
+  environment variable to test like:
+    ```terminal.sh-session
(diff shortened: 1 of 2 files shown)


norio-nomura:
ref: https://github.com/realm/SwiftLint/pull/2730#issuecomment-487772518

norio-nomura:
> 👍 Should we add this as a step in `Releasing.md`?

I think that confirmation necessary for release is done with `testSimulateHomebrewTestWithDisableSourceKit()`, so I don't think that adding to `Releasing.md` is necessary.

norio-nomura:
The reason for leaving `testSimulateHomebrewTest()` as opt-in instead of deleting it is because I wanted to leave a way to check the behavior on sandbox other than `brew install swiftlint` with editing formula or `brew install swiftlint --HEAD`.

PR #2746, opened 2019-05-06
Skip module import if cursor info is missing module info

alvarhansen:
I believe setting `nextIsModuleImport = false` in current code is accidental bug, as it makes no sense to set it false if it already is false. It makes more sense if it is in inner if statements else block.

At the moment it will lead to crash if you import module that SourceKit is unable to find info for.

With following code:

```swift
import UnknownModule
func foo(error: Swift.Error) {}
```

The execution goes like this:
1. `import` token is found and `nextIsModuleImport` is set to `true`
2. `identifier` token is found but if SourceKit returns empty dictionary the loop just continues
3. `Swift` token in `foo(error: Swift.Error) ` is visited and if it has `key.modulename` `source.lang.swift.ref.module` it is falsely considered as an import statement.
4. Its name is added to `unusedImports`
5. `contentsNSString.range(of: "import \(module)\n")` returns NSRange with location NSNotFound
...
6. `func correct(file: File, compilerArguments: [String]) -> [Correction]` will throw NSException when attempting to replace this range.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,10 @@
   return type is required.  
   [Marcelo Fabri](https://github.com/marcelofabri)
 
+* Skip module import if cursor info is missing module info.  
+  [alvarhansen](https://github.com/alvarhansen)
+  [#2746](https://github.com/realm/SwiftLint/issues/2746)
+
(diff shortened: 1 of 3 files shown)


jpsim:
Makes sense. Thanks!

PR #2747, opened 2019-05-07
Always read root directory configuration

alvarhansen:
Removes configuration “working directory” check as the configuration
itself can reside in different directory. This check was introduced
with PR #1604 for issue #1531. At the time it was necessary as
SwiftLint did not support Configuration merging and this stopped
overwriting the configuration. Replaced it with configuration file
path check.

At the moment, if you load configuration from different path than root
path **and** you also have configuration file in root path then it is
skipped. Now that configuration merging is supported, it is more
correct to merge them.

Example where `Level3/.swiftlint.yml` is skipped:
`swiftlint lint --config ../.swiftlint.yml --path Level3`

```
Level2
│   .swiftlint.yml
│
└───Level3
│   │   .swiftlint.yml
│   │   Level3.swift
```

Adds test for making sure that configuration that is out of working
directory is merged with configuration from working directory.

Adds test for #1531 to make sure regression won’t happen in the future.


alvarhansen:
Filed this PR under "Breaking" as this may break linting for some projects.

Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
```diff
@@ -10,27 +10,26 @@ extension Configuration {
...
+            return self
+        }
+
         // If a configuration exists and it isn't us, load and merge the configurations
         if configurationSearchPath != configurationPath &&
             FileManager.default.fileExists(atPath: configurationSearchPath) {
             let fullPath = pathNSString.absolutePathRepresentation()
             let customRuleIdentifiers = (rules.first(where: { $0 is CustomRules }) as? CustomRules)?
                 .configuration.customRuleConfigurations.map { $0.identifier }
-            let config = Configuration.getCached(atPath: fullPath) ??
```
alvarhansen: I removed `Configuration.getCached(atPath: fullPath)` because it is already done in init that we are calling from here. Difference is that `getCached` in that init has extra functionality that is missing from here.

https://github.com/realm/SwiftLint/blob/3a36212/Source/SwiftLintFramework/Models/Configuration.swift#L141

alvarhansen:
@jpsim, @marcelofabri. Is there anything I can do to get this approved and merged?

Issue #2749, opened 2019-05-08
file_types_order false positive for file only containing extensions


samrayner:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

A file containing only extensions is flagged as placed amongst the file type(s) 'main_type'.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
swiftlint lint --path Test.swift --no-cache --enable-all-rules
Linting Swift files at paths Test.swift
Linting 'Test.swift' (1/1)
/Users/samrayner/Desktop/Test.swift:1:1: warning: File Name Violation: File name should match a type or extension declared in the file (if any). (file_name)
/Users/samrayner/Desktop/Test.swift:1:1: warning: File Types Order Violation: An 'extension' should not be placed amongst the file type(s) 'main_type'. (file_types_order)
Done linting! Found 2 violations, 0 serious in 1 file.
```

```swift
//Test.swift
extension Foo {}
extension Bar {
}

```

**NOTE: This only occurs with the above simplified example if the file name is different to the first extended type. Renaming Test.swift to Foo.swift fixes both warnings.**

### Environment

* SwiftLint version (run `swiftlint version` to be sure)?
0.32.0
* Installation method used (Homebrew, CocoaPods, building from source, etc)?
Homebrew
* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)?
No
* Which Xcode version are you using (check `xcode-select -p`)?
10.2.1

Non-trivial example that triggers the warning twice (first two extensions):

```swift
//RequestHeadersStubs.swift
import Foundation
@testable import API

extension UniversalHeaders: Stubbable {
    internal static func stubbed() -> UniversalHeaders {
        return .init(aPIKey: "")
    }
}

extension StandardHeaders: Stubbable {
    internal static func stubbed() -> StandardHeaders {
        return .init(environment: .design)
    }
}

extension AuthorizationHeaders: Stubbable {
    internal static func stubbed() -> AuthorizationHeaders {
        return .init(environment: .design, authorization: "")
    }
}
```

Dschee:
@samrayner Actually this was the intended behavior as file_type_order was designed with expecting "file_name" being turned on as well. But that is not documented and is probably not a good idea given your code example anyways. So making an exception for cases when the file only contains extensions might be a good idea.

[...]

PR #2754, opened 2019-05-10
#2749 Fix file_types_order in extension-only files

samrayner:
Prevents extensions being flagged as in the wrong order when a file only contains extensions and nothing else (enums, classes, structs). In this case it's not intuitive that the longest extension be considered the "main" one so the file can be considered as having no "main" type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,11 @@
   [alvarhansen](https://github.com/alvarhansen)
   [#2746](https://github.com/realm/SwiftLint/issues/2746)
 
+* Don't trigger `file_types_order` violations in files only containing 
+  extensions.  
+  [Sam Rayner](https://github.com/samrayner)
+  [#2749](https://github.com/realm/SwiftLint/issues/2749)
+
 ## 0.32.0: Wash-N-Fold-N-Reduce
 
 #### Breaking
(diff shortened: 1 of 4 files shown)


Source/SwiftLintFramework/Rules/Style/FileTypesOrderRule.swift
```diff
@@ -137,19 +137,10 @@ public struct FileTypesOrderRule: ConfigurationProviderRule, OptInRule {
...
-        }
-
-        guard let mainTypeSubstructure = substructuresSortedByBodyLength.first else {
-            let substructuresSortedByBodyLength = dict.substructure.sorted { lhs, rhs in
-                return (lhs.bodyLength ?? 0) > (rhs.bodyLength ?? 0)
-            }
-
-            // specify substructure with longest body as main type if existent
-            return substructuresSortedByBodyLength.first
+            return (lhs.bodyLength ?? 0) >= (rhs.bodyLength ?? 0)
```
samrayner: This subtle change (> to >=) ensures that the "main" substructure (struct, enum or class) is assumed to be the first of the longest substructures if more than one is tied for longest. Previously, the last of equal length longest substructures was declared the main type, making fixing the warning by rearranging impossible.

CHANGELOG.md
```diff
@@ -39,6 +39,11 @@
   [alvarhansen](https://github.com/alvarhansen)
   [#2746](https://github.com/realm/SwiftLint/issues/2746)
 
+* Don't trigger `file_types_order` violations in files only containing 
+  extensions.
```
Dschee: Please add two whitespaces at the end of this line for correct GitHub line breaks.

samrayner: Done, thanks!

Dschee:
Great, now that the comment is fixed, I'm merging this. Thanks again! 🚀

PR #2764, opened 2019-05-18
Fix issue where force-unwrapping self does not trigger a violation of the `force_unwrapping` rule

### LINKED ISSUE #2759: Force Unwrapping rule doesn't apply to self
### New Issue Checklist

### Describe the bug

Linting the following snippet with `force_unwrapping` enabled in `.swiftlint.yml`: 

```swift
class Test {
    func foo() {
        DispatchQueue.main.async { [weak self] in
            var string: String?
            _ = string!
            _ = self!
        }
    }
}
```

I would expect SwiftLint to warn on both force unwrappings. However, it does not catch `self!`.

```bash
$ swiftlint        
Linting Swift files at paths 
Linting 'foo.swift' (1/1)
/Users/simon/tmp/testswiftlint/foo.swift:5:23: warning: Force Unwrapping Violation: Force unwrapping should be avoided. (force_unwrapping)
```

### Environment

0.32.0 - installed with Homebrew

```yml
opt_in_rules:
  - force_unwrapping
```

Running Xcode 10.2.1 (confirmed with `xcode-select -p`), not using nested configurations.

daltonclaybrook:
Resolves #2759

### DIFF
--- a/Source/SwiftLintFramework/Rules/Idiomatic/ForceUnwrappingRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/ForceUnwrappingRule.swift
@@ -30,7 +30,8 @@ public struct ForceUnwrappingRule: OptInRule, ConfigurationProviderRule, Automat
             "func foo(_ options: [AnyHashable: Any]!) {",
             "func foo() -> [Int]!",
             "func foo() -> [AnyHashable: Any]!",
-            "func foo() -> [Int]! { return [] }"
+            "func foo() -> [Int]! { return [] }",
+            "return self"
         ],
         triggeringExamples: [
             "let url = NSURL(string: query)↓!",
@@ -49,7 +50,8 @@ public struct ForceUnwrappingRule: OptInRule, ConfigurationProviderRule, Automat
               var foo: [Bool]! = dict[\"abc\"]↓!
             }
             """,
-            "open var computed: String { return foo.bar↓! }"
+            "open var computed: String { return foo.bar↓! }",
+            "return self↓!"
         ]
     )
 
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks for the PR, changes look good, but I'll run OSSCheck locally to see if the performance hits reported by CI are legit.

jpsim:
If anyone's curious, you can run OSSCheck locally by just running `./script/oss-check` in your terminal. It takes a few minutes to run and your machine shouldn't be under other heavy load in order for the benchmarks to be trusted.

jpsim:
Running OSSCheck locally, it seems there might be a performance hit, but if so it's small enough to live with, in the 1%-6% range, closer to 3-4%.

Source/SwiftLintFramework/Rules/Idiomatic/ForceUnwrappingRule.swift
```diff
@@ -119,20 +121,18 @@ public struct ForceUnwrappingRule: OptInRule, ConfigurationProviderRule, Automat
...
+            return nil
         }
 
         let violationRange = NSRange(location: NSMaxRange(firstRange), length: 0)
+        let kindsInFirstRange = syntaxMap.kinds(inByteRange: matchByteFirstRange)
 
-        // if first captured range is identifier, generate violation
-        if kindsInFirstRange.contains(.identifier) {
+        // if first captured range is identifier or keyword (self), generate violation
+        if !Set(kindsInFirstRange).intersection([.identifier, .keyword]).isEmpty {
```
jpsim: Made a small improvement to this in a55178e68103fb51722f65bc5f921a8f99f08921

daltonclaybrook: Very cool, thanks!

PR #2765, opened 2019-05-21
Add new indentation_width rule

fredpi:
It's been a while since https://github.com/realm/SwiftLint/issues/227 has last been discussed and it seems that an advanced indentation rule is somehow also pending some clarification on Xcode's side. However, since this issue has been open for such a long time, I thought about adding a basic additive indentation rule that only does the following things:

- Check whether the next line is on the same level as the previous line or one tab or `n` spaces indented (where `n` can be configured via `indentation_width`) or unindented keeping the grid.
- Check whether the first line is not indented.
- Check whether spaces and tabs aren't mixed up in one line's indentation.

This PR implements such a basic rule. I tested it with a large project we have in our company and found out that it matches the indentation style we already pursued before:

```
let testStruct = TestStruct(
    name: "Test",
    indentation: "Cool"
)
```

However, this rule won't be compatible with other indentation styles like the one Xcode does by default and that is enforced by the `vertical_parameter_alignment_on_call` opt-in-rule:

```
let testStruct = TestStruct(name: "Test",
                            indentation: "Cool")
```

That's why this is an opt-in rule. As such, it won't do harm to those following another indentation style, yet enable built-in indentation linting for those who follow this basic additive pattern. Previously, we had implemented such a rule using multiple regexes that checked for indentation but didn't always work properly, so having this built-in is definitely a plus.

I want to mention that, according to this rule, unindentation can not only be done by one step (~ `indentation_width` spaces), but also by multiples of one step. This is to allow for the following indentation which I consider valid:

```
if
    let myLet = someBool ?
        myLet1 :
        myLet2
{
    // Do something with myLet
} 
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -172,6 +172,11 @@
   expressions.  
   [PaulTaykalo](https://github.com/PaulTaykalo)
   [#2918](https://github.com/realm/SwiftLint/issues/2918)
+  
+* Add new indentation opt-in-rule checking for
+  basic additive indentation pattern.  
+  [Frederick Pietschmann](https://github.com/fredpi)
+  [#227](https://github.com/realm/SwiftLint/issues/227)
 
 * Speed up syntax token lookups, which can improve performance when
   linting large files.  
@@ -215,11 +220,6 @@
 
 * Make `contains_over_first_not_nil` rule also match `first(where:) == nil`.  
   [Colton Schlosser](https://github.com/cltnschlosser)
-  
-* Add new indentation opt-in-rule checking for
-  super-basic additive indentation pattern.  
-  [Frederick Pietschmann](https://github.com/fredpi)
-  [#227](https://github.com/realm/SwiftLint/issues/227)
 
 * Add two new cases to the Mark rule to detect a Mark using three slashes.  
   [nvanfleet](https://github.com/nvanfleet)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -172,11 +172,6 @@
   expressions.  
   [PaulTaykalo](https://github.com/PaulTaykalo)
   [#2918](https://github.com/realm/SwiftLint/issues/2918)
-  
-* Add new indentation opt-in-rule checking for
-  basic additive indentation pattern.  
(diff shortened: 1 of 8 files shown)


fredpi:
Any thoughts on this? ;)

fredpi:
As I'm personally not that active in reviewing other people's PRs, I'm certainly not in the position to complain that, seemingly, nobody wants to review this.

However, I'd still like to see this merged at some point. Don't be scared by the `41903 Warnings` message, it's because not all projects follow the indentation pattern this rule enforces.

fredpi:
@jpsim Would you mind having a look here? ;)

fredpi:
Anyone?

fredpi:
Rebasing this every 3 weeks or so is getting frustrating...

If nobody has time to review this, maybe one could give me push access so that I can self-review and merge thereafter...
As this is just an opt-in-rule, it doesn't do any harm...

fredpi:
`/__w/1/s/Tests/SwiftLintFrameworkTests/LinterCacheTests.swift:351: error: LinterCacheTests.testDetectSwiftVersion : XCTAssertEqual failed: ("5.1.3") is not equal to ("5.1.2")`

This test failure does not seem to be related to this PR specifically...

fredpi:
@marcelofabri Would you mind having a look here?

jpsim:
> Rebasing this every 3 weeks or so is getting frustrating...
> 
> If nobody has time to review this, maybe one could give me push access so that I can self-review and merge thereafter...
> As this is just an opt-in-rule, it doesn't do any harm...

Sorry for the lack of communication here @fredpi. I don't have a whole lot of time for OSS work lately 😞

I've given you commit access to the repo. In general, you should feel empowered to merge PRs to SwiftLint if other maintainers aren't being responsive.

CHANGELOG.md
```diff
@@ -9,8 +9,11 @@
 * None.
 
 #### Enhancements
-
-* None.
+  
+* Add new indentation opt-in-rule checking for
```
jpsim: Nit:
```suggestion
* Add new indentation opt-in rule checking for
```

PR #2768, opened 2019-05-23
Print relative path for duplicate filenames

keith:
In projects with multiple Swift targets, it's possible to have multiple
Swift files with the same name. In this case you can't differentiate
from the log, which one SwiftLint is printing about. Now in this case it
prints the relative (to your pwd) path of the file.

### DIFF
--- a/Source/swiftlint/Extensions/Configuration+CommandLine.swift
+++ b/Source/swiftlint/Extensions/Configuration+CommandLine.swift
@@ -98,40 +98,9 @@ extension Configuration {
 
     private func visit(filesPerConfiguration: [Configuration: [File]],
                        visitor: LintableFilesVisitor) -> Result<[File], CommandantError<()>> {
-        var index = 0
+        var fileIndex = 0
         let fileCount = filesPerConfiguration.reduce(0) { $0 + $1.value.count }
         let root = FileManager.default.currentDirectoryPath.bridge().standardizingPath.bridge()
-        let visit = { (file: File, config: Configuration, duplicateFileNames: Set<String>) -> Void in
-            let skipFile = visitor.shouldSkipFile(atPath: file.path)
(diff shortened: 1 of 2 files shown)


Source/swiftlint/Extensions/Configuration+CommandLine.swift
```diff
@@ -82,22 +82,38 @@ extension Configuration {
...
     }
 
+    private func outputFilename(for path: NSString, in root: NSString, duplicateFileNames: Set<String>) -> String {
+        let basename = path.lastPathComponent
+        if !duplicateFileNames.contains(basename) {
+            return basename
+        }
+
+        var pathComponents = path.pathComponents
+        for component in root.pathComponents where !pathComponents.isEmpty && pathComponents[0] == component {
```
jpsim:
```suggestion
        for component in root.pathComponents where pathComponents.first == component {
```

marcelofabri:
> It may also report some removed and introduced violations due to the difference in how before/after violations will be printed.

Isn't this only touching the verbose logs? My understanding is that all the reporters output should still be the same.

jpsim:
> Isn't this only touching the verbose logs? My understanding is that all the reporters output should still be the same.

Correct, I didn't think this one through.

jpsim:
Looks like performance is unaffected with this change, only issues remaining is Linux support and fixing the two SwiftLint violations.

keith:
I've updated for the Linux fix just to make sure CI is happy, I'm thinking about the best way to break this function up

jpsim:
I’m ok if you just want to slap a disable comment to avoid needing to refactor this just because you happen to contribute something at a point where the function was already nearing the limit. It’s not exactly fair.

keith:
Thanks, I had been meaning to get back to this

PR #2770, opened 2019-05-26
Don't trigger no_fallthrough_only if next case is @unknown

### LINKED ISSUE #2696: Update no_fallthrough_only rule condition to handle @unknown default
In swift 4.2 and later we can use [`@unknown default`](https://github.com/apple/swift/pull/14382) in a switch block case to handle _future_ enum cases.
This is ok however, as far as I know, this case has to be handled separately from other cases in a switch.

Let's say that we have the following script:

```swift
#!/usr/bin/swift
import Foundation

enum EnumExample {
    case caseA, caseB
}

let enumInstance: EnumExample = .a

switch enumInstance {
case .caseA:
    print("it's a")
case .caseB:
    fallthrough
@unknown default:
    print("it's not a")
}
```

This works as expected, however, again, as far as I know, there's no way to merge the cases `case .caseB:` and `@unknown default:`.

As the latest SwiftLint version (0.31.0), this code triggers the `no_fallthrough_only` warning:
```
warning: No Fallthrough Only Violation: Fallthroughs can only be used if the `case` contains at least one other statement. (no_fallthrough_only)
```

### Steps to Reproduce

1. copy the code above in a empty `.swift` file
2. make said file executable (if you want to run it) by running `$ chmod +x YOURFILENAME.swift`
3. run `$ swiftlint`

### Expected Results
no `no_fallthrough_only` warning

### Actual Results
`no_fallthrough_only` warning

I understand why the warning triggers, however I believe this case where we use `@unknown default` should be an exception to this rule.

Thank you in advance for all the support! 🤗

marcelofabri:
Fixes #2696

### DIFF
--- a/Source/SwiftLintFramework/Rules/Idiomatic/NoFallthroughOnlyRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/NoFallthroughOnlyRule.swift
@@ -50,7 +50,7 @@ public struct NoFallthroughOnlyRule: ASTRule, ConfigurationProviderRule, Automat
     }
 
     private func isNextTokenUnknownAttribute(afterOffset offset: Int, file: File) -> Bool {
-        let nextNonCommentToken = file.syntaxMap.tokens.lazy
+        let nextNonCommentToken = file.syntaxMap.tokens
             .first { token in
                 guard let kind = SyntaxKind(rawValue: token.type), !kind.isCommentLike else {
                     return false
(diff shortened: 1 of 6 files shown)


Source/SwiftLintFramework/Rules/Idiomatic/NoFallthroughOnlyRule.swift
```diff
@@ -192,6 +49,20 @@ public struct NoFallthroughOnlyRule: ASTRule, ConfigurationProviderRule, Automat
         return []
     }
 
+    private func isNextTokenUnknownAttribute(afterOffset offset: Int, file: File) -> Bool {
+        let nextNonCommentToken = file.syntaxMap.tokens.lazy
```
jpsim: I don’t think lazy does anything here.

PR #2771, opened 2019-05-26
Add legacy_multiple opt-in rule

### LINKED ISSUE #2612: Rule Request: Legacy Multiple
### New Issue Checklist

### New rule request

Prefer using `x.isMultiple(of: y)` than `x % y == 0`.

https://github.com/apple/swift-evolution/blob/master/proposals/0225-binaryinteger-iseven-isodd-ismultiple.md

**Would trigger**
```swift
// UITableView alternating row colour
cell.contentView.backgroundColor = indexPath.row % 2 == 0 ? .gray : .white

// Codable.swift.gyb in apple/swift
guard count % 2 == 0 else { throw DecodingError.dataCorrupted(...) }

// Bool.swift in apple/swift
public static func random<T: RandomNumberGenerator>(using generator: inout T) -> Bool {
    return (generator.next() >> 17) & 1 == 0
}

// KeyPath.swift in apple/swift
_sanityCheck(bytes > 0 && bytes % 4 == 0, "capacity must be multiple of 4 bytes")

// ReversedCollection Index.base documentation https://developer.apple.com/documentation/swift/reversedcollection/index/2965437-base
guard let i = reversedNumbers.firstIndex(where: { $0 % 2 == 0 })
```

**Wouldn't trigger**

```swift
// UITableView alternating row colour
cell.contentView.backgroundColor = indexPath.row.isMultiple(of: 2) ? .gray : .white

// Codable.swift.gyb in apple/swift
guard count.isMultiple(of: 2) else { throw DecodingError.dataCorrupted(...) }

// Bool.swift in apple/swift
public static func random<T: RandomNumberGenerator>(using generator: inout T) -> Bool {
    return (generator.next() >> 17).isMultiple(of: 2)
}

// KeyPath.swift in apple/swift
_sanityCheck(bytes > 0 && bytes.isMultiple(of: 4), "capacity must be multiple of 4 bytes")

// not used in a boolean context
let secretValue = (value % 3) + 2
```

Just severity

Opt-in as `isMultiple(of:)` is only available on Swift 5+.

marcelofabri:
Fixes #2612

### DIFF
--- a/Source/SwiftLintFramework/Rules/Idiomatic/LegacyMultipleRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/LegacyMultipleRule.swift
@@ -1,0 +1,52 @@
+import SourceKittenFramework
+
+public struct LegacyMultipleRule: OptInRule, ConfigurationProviderRule, AutomaticTestableRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "legacy_multiple",
+        name: "Legacy Multiple",
+        description: "Prefer using the `isMultiple(of:)` function instead of using the remainder operator (`%`).",
+        kind: .idiomatic,
+        minSwiftVersion: .five,
+        nonTriggeringExamples: [
+            "cell.contentView.backgroundColor = indexPath.row.isMultiple(of: 2) ? .gray : .white",
+            "guard count.isMultiple(of: 2) else { throw DecodingError.dataCorrupted(...) }",
+            "sanityCheck(bytes > 0 && bytes.isMultiple(of: 4), \"capacity must be multiple of 4 bytes\")",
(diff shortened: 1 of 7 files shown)


Source/SwiftLintFramework/Rules/Idiomatic/LegacyMultipleRule.swift
```diff
@@ -0,0 +1,52 @@
...
+            let constant = 56
+            let isMultiple = value ↓% constant == 0
+            """
+        ]
+    )
+
+    // MARK: - Rule
+
+    public func validate(file: File) -> [StyleViolation] {
+        let pattern = "(?!\\b\\s*)%\(RegexHelpers.variableOrNumber)==\\s*0\\b"
```
marcelofabri: I wonder if this should also catch `!= 0` 🤔

jpsim: Probably could, yeah.

PR #2777, opened 2019-06-02
Add `duplicate_enum_cases` rule

### LINKED ISSUE #2676: [Rule Request]: Duplicate Enum Cases
### New Issue Checklist

### New rule request

In Swift, it's currently legal to define two identical enum cases as long as the label of an associated value is different:

```swift
// This compiles
enum PictureImport {
    case add(image: UIImage)
    case add(data: Data)
}
```

This, however, leads to compiler errors when switching:

```swift
// This doesn't compile
func importImage(_ importedImage: PictureImport) {
    switch importedImage {
    case .add(image: let image): // Error: Tuple pattern element label 'image' must be 'data'
        // ...
    case .add(data: let data):
        // ...
    }
}
```

I'd like to propose a rule that warns against identical enum case names.

#### Triggers:

```swift
enum PictureImport {
    case add(image: UIImage)
    case add(data: Data)
}
```

#### Does Not Trigger:

```swift
enum PictureImport {
    case addImage(UIImage)
    case addData(Data)
}
```

In my opinion, this rule should be enabled by default and be marked as an error.

marcelofabri:
Fixes #2676

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,12 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2696](https://github.com/realm/SwiftLint/issues/2696)
 
+* Add `duplicate_enum_cases` rule to validate that an enum doesn't contain
+  duplicated cases, as it's impossible to switch on it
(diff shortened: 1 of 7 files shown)


jpsim:
Nice!

PR #2796, opened 2019-07-06
Split cache into one file per configuration

cltnschlosser:
This improves cache deserialization performance. I discovered this potential bottleneck after accidentally running swiftlint over a large amount of generated files. This caused my cache to balloon in size.

Here are some rough performance numbers from a large app, with many private pods. Times collected using `time` utility averaged over several runs and rounded. This isn't the most accurate way to benchmark, but I was looking for big noticeable changes.
Also I found I could improve the script we were using for running swiftlint over the whole project. You'll see this referenced as `old script` and `new script`.
`old script` was implemented by running `swiftlint` once for each private pod path.
`new script` batches all the paths together and runs `swiftlint` once by providing it multiple paths.

```
0.33.0:
Old script:
- Without cache: 37
- With bloated cache: 3:08
- After minimal cache: 32

New script:
- Without cache: 35
- With bloated cache: 5.6
- With minimal cache: 2.8

cs_splitCache:
Old script:
- Without cache: 38
- With empty cache: 39
- With cache: 5.9

New script:
- Without cache: 34
- With empty cache: 34
- With cache: 2.8
```

`Without cache` tests were run with `--no-cache`.
`With empty cache` tests were run after deleting the caches directory. This was to confirm that saving to the cache didn't have large impact to performance.

Key take aways here:
- Old script was harshly punished by bloated and long cache load times. This is because the cache had to be loaded multiple times.
- Splitting the cache helps reduce cache bloat. This is especially useful for people who use swiftlint on multiple large projects.

Other things tested:
- Using read/write locks or DispatchQueues instead of NSLock. This didn't have a measurable effect in my testing, so I left the NSLocks because they are more simple.

Concerns:
- Possible regression for projects with lots of `.swiftlint.yml` configuration files. If anyone has a project I can test, or wants to help make an example worst case project, let me know. Considering all the fileIO swiftlint already does, I wouldn't expect this to be a bottleneck, but I consider it an area where there could be a performance regression.

### DIFF
--- a/Source/SwiftLintFramework/Models/LinterCache.swift
+++ b/Source/SwiftLintFramework/Models/LinterCache.swift
@@ -118,11 +118,7 @@ public final class LinterCache {
         let readCache = lazyReadCache
         readCacheLock.unlock()
 
-        for (description, writeFileCache) in writeCache {
-            guard !writeFileCache.isEmpty else {
-                continue
-            }
-
+        for (description, writeFileCache) in writeCache where !writeFileCache.isEmpty {
             let fileCache = readCache[description]?.merging(writeFileCache) { _, write in write } ?? writeFileCache
             let json = try fileCache.toJSON()
             let file = url.appendingPathComponent(description).appendingPathExtension("json")
--- a/Source/SwiftLintFramework/Models/LinterCache.swift
+++ b/Source/SwiftLintFramework/Models/LinterCache.swift
@@ -126,7 +126,7 @@ public final class LinterCache {
             let fileCache = readCache[description]?.merging(writeFileCache) { _, write in write } ?? writeFileCache
             let json = try fileCache.toJSON()
             let file = url.appendingPathComponent(description).appendingPathExtension("json")
-            try json.write(to: file, atomically: true, encoding: .utf8)
+            try json.write(to: file, options: .atomic)
         }
     }
 
@@ -233,13 +233,8 @@ private extension StyleViolation {
 }
 
 internal extension Dictionary where Key == String {
-    func toJSON() throws -> String {
+    func toJSON() throws -> Data {
         // not using .sortedKeys to avoid crash
-        let prettyJSONData = try JSONSerialization.data(withJSONObject: self, options: .prettyPrinted)
-        if let jsonString = String(data: prettyJSONData, encoding: .utf8) {
-            return jsonString
-        } else {
-            throw LinterCacheError.invalidFormat
-        }
+        return try JSONSerialization.data(withJSONObject: self, options: .prettyPrinted)
     }
 }
--- a/Source/SwiftLintFramework/Models/LinterCache.swift
+++ b/Source/SwiftLintFramework/Models/LinterCache.swift
@@ -9,11 +9,10 @@ public final class LinterCache {
     private typealias FileCache = [String: [String: Any]]
     private typealias Cache = [String: FileCache]
 
-    private let readCacheLock = NSLock()
-
     private var lazyReadCache: Cache
+    private let readCacheLock = NSLock()
     private var writeCache = Cache()
-    private let lock = NSLock()
+    private let writeCacheLock = NSLock()
     internal let fileManager: LintableFileManager
     private let location: URL?
     private let swiftVersion: SwiftVersion
@@ -61,7 +60,7 @@ public final class LinterCache {
 
         let configurationDescription = configuration.cacheDescription
 
-        lock.lock()
+        writeCacheLock.lock()
         var filesCache = writeCache[configurationDescription] ?? [:]
         filesCache[file] = [
             Key.violations.rawValue: violations.map(dictionary(for:)),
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks for doing this! I think this change makes sense and I really appreciate your descriptions of you rationale, testing methodology and even potential concerns.

I hope to review this shortly.

Package.resolved
```diff
@@ -10,6 +10,15 @@
           "version": "0.16.0"
         }
       },
+      {
+        "package": "CryptoSwift",
```
jpsim: Can you please revert this change?

We only add CryptoSwift as a dependency on Linux (or anywhere else `CommonCrypto` can't be imported).

Given that most SwiftLint developers use macOS, it's less noisy to leave this out.

Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -110,31 +106,73 @@ public final class LinterCache {
...
         }
 
-        let cache = mergeCaches()
-        let json = try cache.toJSON()
-        try json.write(to: url, atomically: true, encoding: .utf8)
+        readCacheLock.lock()
+        let readCache = lazyReadCache
+        readCacheLock.unlock()
+
+        for (description, writeFileCache) in writeCache {
```
jpsim: Nit:

```suggestion
        for (description, writeFileCache) in writeCache where !writeFileCache.isEmpty {
```

Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -195,13 +233,8 @@ private extension StyleViolation {
 }
 
 internal extension Dictionary where Key == String {
-    func toJSON() throws -> String {
+    func toJSON() throws -> Data {
```
jpsim: Nice, I didn't realize we were converting the JSON data to a string then back to data again without ever needing the intermediate string representation. 👌

Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -110,31 +106,73 @@ public final class LinterCache {
...
+        }
+        if let fileCache = lazyReadCache[cacheDescription] {
+            return fileCache
+        }
+
+        guard let location = location else {
+            return [:]
+        }
+
+        let file = location.appendingPathComponent(cacheDescription).appendingPathExtension("json")
```
jpsim: Will `appendingPathComponent(cacheDescription)` always produce a safe path for writing? I don't remember how we compute it these days (we've changed it over time) so I'll need to trace through the code to confirm.

Package.resolved
cltnschlosser: Okay, it showed up after running test, so I assumed it needed to be there.

jpsim:
> I'm not convinced yet that all the lock-related changes are safe.

After looking more closely, I think these changes are safe. The TSan CI job would have likely caught unsafe behavior too.

Package.resolved
jpsim: Yeah I think that's a bug in SwiftPM somewhere since CommonCrypto should always be importable on macOS.

Source/SwiftLintFramework/Models/LinterCache.swift
cltnschlosser: It's an `MD5`, that should be safe.

cltnschlosser:
> After looking more closely, I think these changes are safe. The TSan CI job would have likely caught unsafe behavior too.

Yeah, I've run the tests and swiftlint itself several times now, I've had no issues. I was pretty aggressive with the locking. Doing something like this, most of the time:
```
writeCacheLock.lock()
defer {
	writeCacheLock.unlock()
}
```

jpsim:
Fantastic work! Here are some of my benchmarks after running on a project with lots of nested `.swiftlint.yml` configuration files:

Running on the large project:

```
$ hyperfine -w 1 './swiftlint-master lint --quiet' './swiftlint-cs_splitCache lint --quiet'
Benchmark #1: ./swiftlint-master lint --quiet
  Time (mean ± σ):      4.010 s ±  0.264 s    [User: 5.709 s, System: 1.244 s]
  Range (min … max):    3.592 s …  4.438 s    10 runs
 
Benchmark #2: ./swiftlint-cs_splitCache lint --quiet
  Time (mean ± σ):      3.662 s ±  0.077 s    [User: 5.463 s, System: 1.155 s]
  Range (min … max):    3.516 s …  3.767 s    10 runs
 
Summary
  './swiftlint-cs_splitCache lint --quiet' ran
    1.10 ± 0.08 times faster than './swiftlint-master lint --quiet'
```

Running on SwiftLint itself after:

```
$ hyperfine './swiftlint-master lint --quiet' './swiftlint-cs_splitCache lint --quiet'
Benchmark #1: ./swiftlint-master lint --quiet
  Time (mean ± σ):     811.4 ms ±  20.6 ms    [User: 915.1 ms, System: 56.4 ms]
  Range (min … max):   785.0 ms … 856.6 ms    10 runs
 
Benchmark #2: ./swiftlint-cs_splitCache lint --quiet
  Time (mean ± σ):     213.4 ms ±   9.6 ms    [User: 317.6 ms, System: 53.5 ms]
  Range (min … max):   199.3 ms … 231.9 ms    12 runs
 
Summary
  './swiftlint-cs_splitCache lint --quiet' ran
    3.80 ± 0.20 times faster than './swiftlint-master lint --quiet'
```

PR #2797, opened 2019-07-06
Update Rules.md

jpsim:


jpsim:
Oops, can't do this until we update CI.

marcelofabri:
The tests will fail in CI because that example only runs when using Swift 5.1 😬

PR #2799, opened 2019-07-06
Migrate LinterCache to use Codable models

jpsim:
Improving performance and type safety. This changes the cache file format from JSON to binary plist when available, which also happens to reduce disk usage by around ~10% in my tests.

For a project with about 7k Swift files amounting to about ~500k lines of Swift code, here are the benchmarks for before and after:

```
$ hyperfine 'swiftlint-master lint --quiet' 'swiftlint-codable-cache lint --quiet'                    
Benchmark #1: swiftlint-master lint --quiet
  Time (mean ± σ):      3.500 s ±  0.037 s    [User: 5.208 s, System: 1.115 s]
  Range (min … max):    3.439 s …  3.573 s    10 runs
 
Benchmark #2: swiftlint-codable-cache lint --quiet
  Time (mean ± σ):      3.013 s ±  0.026 s    [User: 4.807 s, System: 1.117 s]
  Range (min … max):    2.966 s …  3.041 s    10 runs
 
Summary
  'swiftlint-codable-cache lint --quiet' ran
    1.16 ± 0.02 times faster than 'swiftlint-master lint --quiet'
```

### DIFF
--- a/Source/SwiftLintFramework/Models/LinterCache.swift
+++ b/Source/SwiftLintFramework/Models/LinterCache.swift
@@ -18,6 +18,16 @@ private struct FileCache: Codable {
 }
 
 public final class LinterCache {
+#if canImport(Darwin)
+    private typealias Encoder = PropertyListEncoder
+    private typealias Decoder = PropertyListDecoder
+    private static let fileExtension = "plist"
+#else
+    private typealias Encoder = JSONEncoder
+    private typealias Decoder = JSONDecoder
+    private static let fileExtension = "json"
+#endif
+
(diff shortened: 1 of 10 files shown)


Source/SwiftLintFramework/Models/LinterCache.swift
```diff
@@ -1,12 +1,33 @@
 import Foundation
 
-internal enum LinterCacheError: Error {
+private enum LinterCacheError: Error {
     case invalidFormat
```
marcelofabri: this seems to be unused now

PR #2810, opened 2019-07-14
Add `empty_collection_literal` rule

### LINKED ISSUE #2807: Rule Request: Empty Array Literal
### New Issue Checklist

### New rule request

Prefer checking `isEmpty` over comparing string to an empty array literal.

This reflects the behaviour already implemented by `empty_string`.

#### What would trigger ####
```swift
if myArray == [] { … }

guard list != [] else { return }

if [] == anArray { … }

theArray==[] ? return : f(theArray)
```

#### What would not trigger ####
```swift
let myArray = []

if myArray.isEmpty { … }

guard list.isEmpty else { return }

if myArray == [1, 2, 3] { … }
```

#### 3. Should the rule be configurable, if so what parameters should be configurable? ####
No.

#### 4. Should the rule be opt-in or enabled by default? Why? ####
Same as for `empty_string`.

cltnschlosser:
Resolves #2807

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@
   expressions like `filter(where:).isEmpty` instead of `contains(where:)`.  
   [Marcelo Fabri](https://github.com/marcelofabri)
 
+* Add `empty_collection_literal` opt-in rule to prefer using `isEmpty` to comparison to `[]` or `[:]`.  
+  [Colton Schlosser](https://github.com/cltnschlosser)
+  [#2807](https://github.com/realm/SwiftLint/issues/2807)
+
 #### Bug Fixes
 
 * Fixed false positive in `colon` rule inside guard and ternary operator.  
--- a/Rules.md
+++ b/Rules.md
@@ -32,6 +32,7 @@
 * [Duplicate Enum Cases](#duplicate-enum-cases)
 * [Duplicate Imports](#duplicate-imports)
 * [Dynamic Inline](#dynamic-inline)
+* [Empty Collection Literal](#empty-collection-literal)
 * [Empty Count](#empty-count)
 * [Empty Enum Arguments](#empty-enum-arguments)
 * [Empty Parameters](#empty-parameters)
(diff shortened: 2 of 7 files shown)


Source/SwiftLintFramework/Rules/Performance/EmptyArrayRule.swift
```diff
@@ -0,0 +1,33 @@
...
+            "!myArray.isEmpy"
+        ],
+        triggeringExamples: [
+            "myArray↓ == []",
+            "myArray↓ != []"
+        ]
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        let pattern = "\\b\\s*(==|!=)\\s*\\[\\]"
```
niil-ohlin: What about the case `myArray == [ ]`? Is that covered by this pattern?

cltnschlosser: Good catch, updating now!

Source/SwiftLintFramework/Rules/Performance/EmptyArrayRule.swift
```diff
@@ -0,0 +1,34 @@
+import SourceKittenFramework
+
+public struct EmptyArrayRule: ConfigurationProviderRule, OptInRule, AutomaticTestableRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "empty_array",
```
marcelofabri: should we change this to `empty_collection_literal` and match empty dictionaries too?

cltnschlosser: I like this, should be easy to add.

CHANGELOG.md
```diff
@@ -10,7 +10,9 @@
 
 #### Enhancements
 
-* None.
+* Add `empty_collection_literal` rule to prefer using `isEmpty` to comparison to `[]` or `[:]`.  
+  [Colton Schlosser](https://github.com/cltnschlosser)
+  [2807](https://github.com/realm/SwiftLint/issues/2807)
```
marcelofabri: since you'll have to fix a conflict here, can you also add `#` before the issue number?

marcelofabri: can you also add that this is an opt-in rule?

PR #2811, opened 2019-07-14
Add `contains_over_range_not_nil` rule

### LINKED ISSUE #2776: Rule request: use contains() instead of range(of:) != nil
### New Issue Checklist

### New rule request

Similar to `contains_over_first_not_nil` (https://github.com/realm/SwiftLint/pull/1846) but for strings

1. **Why should this rule be added?** It’s more idiomatic to ask whether the collection contains the element. Checking whether the range of the substring is nil is a more Obj-C way of doing it (e.g. with `range.location == NSNotFound`)

Triggering:

```swift
if items.range(of: item) != nil {
    // items contains item
}

if items.range(of: item) == nil { // notice the negation of the previous example
    // items does not contain item
}
```

Not triggering:

```swift
if item.contains(item) {
    // items contains item
}

if !items.contains(item) { // notice the negation of the previous example
    // items does not contain item
}
```
3. **Should the rule be configurable, if so what parameters should be configurable?** I don’t think it needs to be configurable.
4. **Should the rule be opt-in or enabled by default? Why?** I think it should be on by default. It’s always clearer, in my opinion, to express your intent in checking containment, rather than grabbing the index and checking if it’s nil. Plus, it saves on the calculation and storage allocation for the range, which isn’t much, but is probably more than a `Bool`.

cltnschlosser:
Also make `contains_over_first_not_nil` also match == nil

Resolves #2776

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,14 @@
   [Colton Schlosser](https://github.com/cltnschlosser)
   [#2807](https://github.com/realm/SwiftLint/issues/2807)
 
+* Add `contains_over_range_nil_comparison` opt-in rule to prefer
+  using `contains` over comparison of `range(of:)` to `nil`.  
+  [Colton Schlosser](https://github.com/cltnschlosser)
+  [#2776](https://github.com/realm/SwiftLint/issues/2776)
+
+* Make `contains_over_first_not_nil` rule also match `first(where:) == nil`.  
+  [Colton Schlosser](https://github.com/cltnschlosser)
+
 #### Bug Fixes
 
 * Fixed false positive in `colon` rule inside guard and ternary operator.  
--- a/Tests/LinuxMain.swift
+++ b/Tests/LinuxMain.swift
@@ -214,6 +214,12 @@ extension ContainsOverFirstNotNilRuleTests {
     ]
 }
 
+extension ContainsOverRangeNilComparisonRuleTests {
+    static var allTests: [(String, (ContainsOverRangeNilComparisonRuleTests) -> () throws -> Void)] = [
+        ("testWithDefaultConfiguration", testWithDefaultConfiguration)
+    ]
+}
+
 extension ControlStatementRuleTests {
     static var allTests: [(String, (ControlStatementRuleTests) -> () throws -> Void)] = [
         ("testWithDefaultConfiguration", testWithDefaultConfiguration)
@@ -1571,6 +1577,7 @@ XCTMain([
     testCase(ContainsOverFilterCountRuleTests.allTests),
     testCase(ContainsOverFilterIsEmptyRuleTests.allTests),
     testCase(ContainsOverFirstNotNilRuleTests.allTests),
+    testCase(ContainsOverRangeNilComparisonRuleTests.allTests),
     testCase(ControlStatementRuleTests.allTests),
     testCase(ConvenienceTypeRuleTests.allTests),
     testCase(CustomRulesTests.allTests),
--- a/Rules.md
+++ b/Rules.md
@@ -19,6 +19,7 @@
 * [Contains Over Filter Count](#contains-over-filter-count)
 * [Contains Over Filter Is Empty](#contains-over-filter-is-empty)
 * [Contains over first not nil](#contains-over-first-not-nil)
+* [Contains over range(of:) comparison to nil](#contains-over-rangeof-comparison-to-nil)
 * [Control Statement](#control-statement)
 * [Convenience Type](#convenience-type)
 * [Custom Rules](#custom-rules)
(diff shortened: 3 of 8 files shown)


Tests/LinuxMain.swift
```diff
@@ -1,4 +1,4 @@
-// Generated using Sourcery 0.16.0 — https://github.com/krzysztofzablocki/Sourcery
+// Generated using Sourcery 0.16.1 — https://github.com/krzysztofzablocki/Sourcery
```
cltnschlosser: I ran sourcery, but tests weren't generated, suggestions?

marcelofabri:
This should only match `contains` with a single param (https://github.com/mozilla-mobile/firefox-ios/blob/4d52a5297ee08ee69f0c366e0aa8226093bfa086/content-blocker-lib-ios/src/TrackingProtectionPageStats.swift#L192:16 is a false positive)

Tests/LinuxMain.swift
cltnschlosser: I was missing `AutomaticTestableRule`

cltnschlosser:
@marcelofabri Added a check for argument count

CHANGELOG.md
```diff
@@ -10,7 +10,12 @@
 
 #### Enhancements
 
-* None.
+* Add `contains_over_range_not_nil` rule to prefer using `contains` over comparison of `range(of:)` to `nil`.
```
marcelofabri: same as the other PR, can you please mention that this is an opt-in rule and add `#` before the issue number?

Source/SwiftLintFramework/Rules/Performance/ContainsOverRangeNotNilRule.swift
```diff
@@ -0,0 +1,39 @@
...
+            ]
+        }
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        let pattern = "\\)\\s*(==|!=)\\s*nil"
+        return validate(file: file, pattern: pattern, patternSyntaxKinds: [.keyword],
+                        callNameSuffix: ".range", severity: configuration.severity,
+                        reason: "Prefer `contains` over `range(of:) != nil`") { expression in
+                            guard let substructure = expression["key.substructure"]
```
marcelofabri: you should the accessors defined in `Dictionary+SwiftLint` instead:

```
return expression.enclosedArguments.count == 1
```

CHANGELOG.md
```diff
@@ -23,6 +23,13 @@
   expressions like `filter(where:).isEmpty` instead of `contains(where:)`.  
   [Marcelo Fabri](https://github.com/marcelofabri)
 
+* Add `contains_over_range_not_nil` opt-in rule to prefer using `contains` over comparison of `range(of:)` to `nil`.  
+  [Colton Schlosser](https://github.com/cltnschlosser)
+  [#2776](https://github.com/realm/SwiftLint/issues/2776)
+
+* Make `contains_over_first_not_nil` rule also match `first(where:) == nil`.
```
marcelofabri: Maybe we should split this into a different PR and discuss there? I'm a bit conflicted because while I agree that matching `== nil` is useful, the rule is named `contains_over_first_not_nil`.

CHANGELOG.md
```diff
@@ -23,6 +23,13 @@
   expressions like `filter(where:).isEmpty` instead of `contains(where:)`.  
   [Marcelo Fabri](https://github.com/marcelofabri)
 
+* Add `contains_over_range_not_nil` opt-in rule to prefer using `contains` over comparison of `range(of:)` to `nil`.
```
marcelofabri: Can we rename the rule to `contains_over_range_nil_comparison` or something like that?

Source/SwiftLintFramework/Rules/Performance/ContainsOverRangeNotNilRule.swift
```diff
@@ -0,0 +1,34 @@
...
+
+public struct ContainsOverRangeNotNilRule: CallPairRule, OptInRule, ConfigurationProviderRule, AutomaticTestableRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "contains_over_range_not_nil",
+        name: "Contains over range not nil",
+        description: "Prefer `contains` over `range(of:) != nil`.",
```
marcelofabri: if we rename the rule, we should also update the description

Source/SwiftLintFramework/Rules/Performance/ContainsOverRangeNotNilRule.swift
```diff
@@ -0,0 +1,34 @@
...
+            ]
+        }
+    )
+
+    public func validate(file: File) -> [StyleViolation] {
+        let pattern = "\\)\\s*(==|!=)\\s*nil"
+        return validate(file: file, pattern: pattern, patternSyntaxKinds: [.keyword],
+                        callNameSuffix: ".range", severity: configuration.severity,
+                        reason: "Prefer `contains` over `range(of:) != nil`") { expression in
+                            return expression.enclosedArguments.count == 1
```
marcelofabri: on a second thought, can we compare the param name too to be extra careful?

This should do the trick: 

```swift
return expression.enclosedArguments.map { $0.name } == ["of"]
```

cltnschlosser: I'll update with the rename and the parameter check improvement.

cltnschlosser: Done

jpsim:
`String.contains(_: String)` only works if you import Foundation.

```
$ echo "import Foundation\nprint(\"asdf\".contains(\"as\"))" > test.swift
$ swift test.swift 
true
$ echo "print(\"asdf\".contains(\"as\"))" > test.swift 
$ swift test.swift                                    
test.swift:1:23: error: cannot convert value of type 'String' to expected argument type 'String.Element' (aka 'Character')
print("asdf".contains("as"))
                      ^~~~
```

cltnschlosser:
@jpsim `range(of:)` requires Foundation as well.

PR #2814, opened 2019-07-18
Add UnusedDeclarationRule

jpsim:
This PR adds a new `unused_declaration` analyzer rule to lint for unused declarations.
By default, detects unused `fileprivate`, `private` and `internal` declarations.
Configure the rule with `include_public_and_open: true` to also detect unused `public` and `open` declarations.

Completely remove the `unused_private_declaration` rule.

This is built on the work enabling collecting rule infrastructure in https://github.com/realm/SwiftLint/pull/2714.


Source/SwiftLintFramework/Rules/Lint/UnusedDeclarationRule.swift
```diff
@@ -0,0 +1,277 @@
...
+            // Skip declarations marked as @IBOutlet, @IBAction or @objc
+            // since those might not be referenced in code, but only dynamically (e.g. Interface Builder)
+            if let annotatedDecl = cursorInfo["key.annotated_decl"] as? String,
+                ["@IBOutlet", "@IBAction", "@objc", "@IBInspectable"].contains(where: annotatedDecl.contains) {
+                return nil
+            }
+
+            // Classes marked as @UIApplicationMain are used by the operating system as the entry point into the app.
+            if let annotatedDecl = cursorInfo["key.annotated_decl"] as? String,
+                annotatedDecl.contains("@UIApplicationMain") {
```
keith: should we add `@objcMembers` in this same spot?

jpsim: Good question. You could still have unused `@objcMembers` classes though.

I'm inclined to still treat those as violations until someone requests otherwise.

PR #2822, opened 2019-07-23
Add parameter max_declarations to pattern_matching_keywords

mpodeszwa:
This change adds additional configuration for `pattern_matching_keyword` rule (purely for styling).

Consider following enum:

```
enum Foo {
    case single(Int)
    case double(Int, Int)
    case triple(Int, Int, Int)

    var sum: Int {
        switch self {
        case .single(let x):
            return x
        case .double(let x, let y):
            return x + y
        case .triple(let x, let y, let z):
            return x + y + z
        }
    }
}
```

Enabling `pattern_matching_keyword` would enforce following code:

```
enum Foo {
    case single(Int)
    case double(Int, Int)
    case triple(Int, Int, Int)

    var sum: Int {
        switch self {
        case .single(let x):
            return x
        case let .double(x, y):
            return x + y
        case let .triple(x, y, z):
            return x + y + z
        }
    }
}
```

However, in this case `let`s are not in the same place. `let` is either before or after identifier depending on number of parameters. 

This PR allows you to specify in `.swiftlint.yml`:

```
pattern_matching_keyword:
  max_declarations: 0
```

and then example code will look like this:

```
enum Foo {
    case single(Int)
    case double(Int, Int)
    case triple(Int, Int, Int)

    var sum: Int {
        switch self {
        case let .single(x):
            return x
        case let .double(x, y):
            return x + y
        case let .triple(x, y, z):
            return x + y + z
        }
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,10 @@
 
 #### Enhancements
 
-* None.
+* Add `max_declarations` configuration on `pattern_matching_keywords` rule
+  to specify number of declarations allowed inside tuples in a `switch`.  
+  [Marcin Podeszwa](https://github.com/mpodeszwa)
+  [#2822](https://github.com/realm/SwiftLint/pull/2822)
 
 #### Bug Fixes
 
--- a/Rules.md
+++ b/Rules.md
@@ -15095,6 +15095,18 @@ switch foo {
 }
 ```
 
+```swift
+switch foo {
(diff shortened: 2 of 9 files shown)


jpsim:
Sorry for the delay, but I think this is a good idea, and would be willing to review a new PR with these changes if you're still interested.

mpodeszwa:
@jpsim No worries, I rebased this branch over master and opened PR #3566

Issue #2729, opened 2019-04-25
Feature Request: Enable .swiftlint.yml to include a remote rules configuration


72A12F4E:
It would be really useful in our organization to allow a `.swiftlint.yml` to be remotely included from another `.swiftlint.yml` config. We maintain several iOS / Swift projects, and keeping coding standards synchronized across all of them is difficult. This feature would enable us to create a company-wide config and have each project import that configuration, and we could easily update the config and apply new rules to the portfolio of projects we maintain.

Here is an example of how this could work.

**.swiftlint.yml**
```yaml
parent_config: https://foo.com/path/to/another/.swiftlint.yml

#...
```

What do y'all think?

guzhenhuaGitHub:
So we can use one .swiftlint.yml in several projects. expecting...

jpsim:
I like this, but a few details would need to be figured out:

* What do you do when there's poor to no network connectivity?
* Should SwiftLint make network calls for every command invocation? If not, how should these remote configs be cached?
* Would these external configs be versioned, if so how?
* Would it be possible to import more than one parent configuration?
* Would this feature apply equally to local configurations as well as remote ones?

72A12F4E:
What do you do when there's poor to no network connectivity?

* I think the job should surface an error / warning. This could be mitigated by caching the config on repeated invocations.

Should SwiftLint make network calls for every command invocation? If not, how should these remote configs be cached?

* I think it should attempt to get a fresh config every time, maybe there are config values to error/warn if a config is unavailable. The configs could be cached in the project root or somewhere like that.

Would these external configs be versioned, if so how?

* Thats a good question. I assume we could hash a config and use that to figure out if they have updated.

Would it be possible to import more than one parent configuration?

* I think that would make this more complicated than it needs to be.

Would this feature apply equally to local configurations as well as remote ones?

* There would have to be some sort of conflict resolution strategy for favoring the remote config or the local config... Favoring one or the other could be useful in different circumstances.

jpsim:
Your suggestions sound reasonable. Do you need any help to get started on this @72A12F4E?

bielikb:
@72A12F4E I would like to follow up on these efforts, would you like to collab in providing a solution? 
Have you already kicked-off some work in these regards?

fredpi:
There's still an active PR addressing this: https://github.com/realm/SwiftLint/pull/3058

fredpi:
This is now implemented in #3058 and will be available in the next release!

PR #2824, opened 2019-07-25
Add child and parent config configuration features

### LINKED ISSUE #1352: Enhancement: Allow inclusion of multiple configuration files

fredpi:
This PR closes #1352 by introducing ~a new `sub_config` option that can be used in any configuration file. The `sub_config` must reference a file on the same level so that the option isn't confused with nested configurations. The `sub_config` feature also  allows a `sub_config` to itself specify a `sub_config`. There's also a reference cycle detection.~

--------------------------

**Scroll down for an updated version of what this PR implements.**

--------------------------

I'd love to see this merged as soon as possible, as I'm eagerly waiting to use this feature.


fredpi:
Just a note regarding the changes / fixes to rule configuration: As _SwiftLintBot_ noted, with this new implementation, 3 warnings in Moya are now fixed. In fact, previously those have been false positives, as in the nested `.swiftlint.yml` covering these files there's the following configuration for the `file_length` rule:

```
file_length:
  warning: 1000
```

As the parent rule configuration was used for merging and there was no smart mechanism to decide which configuration to choose, the configuration originating from the nested configuration was ignored. With this PR, that's now fixed.

jpsim:
Thanks for kicking this off @fredpi! This is a very commonly requested feature. I haven't taken a close look just yet, but I can comment on the two failing CI jobs.

The first is because the `Foundation` import in `RulesStorage.swift` is unused. You can just remove that import.

The second failing job is TSan detecting a data race because `RulesStorage.resultingRules` is being mutated and read from different threads without any coordination:

```
WARNING: ThreadSanitizer: data race (pid=3478)
  Write of size 8 at 0x7b18000582a8 by thread T6:
    #0 RulesStorage.resultingRules.getter <compiler-generated> (SwiftLintPackageTests:x86_64+0x23437a)
    #1 Configuration.rules.getter Configuration.swift:26 (SwiftLintPackageTests:x86_64+0x1d65b2)
    #2 Linter.init(file:configuration:cache:compilerArguments:) Linter.swift:126 (SwiftLintPackageTests:x86_64+0x1f0490)
    #3 closure #1 in IntegrationTests.testSwiftLintAutoCorrects() IntegrationTests.swift:36 (SwiftLintPackageTests:x86_64+0x80e41f)
    #4 partial apply for closure #1 in IntegrationTests.testSwiftLintAutoCorrects() <compiler-generated> (SwiftLintPackageTests:x86_64+0x80e4e8)
    #5 thunk for @callee_guaranteed (@guaranteed File) -> (@owned [Correction]) <compiler-generated> (SwiftLintPackageTests:x86_64+0x80e55b)
    #6 partial apply for thunk for @callee_guaranteed (@guaranteed File) -> (@owned [Correction]) <compiler-generated> (SwiftLintPackageTests:x86_64+0x80e600)
    #7 thunk for @callee_guaranteed (@in_guaranteed A) -> (@owned [A1]) <compiler-generated> (SwiftLintPackageTests:x86_64+0x16d563)
    #8 partial apply for thunk for @callee_guaranteed (@in_guaranteed A) -> (@owned [A1]) <compiler-generated> (SwiftLintPackageTests:x86_64+0x16d655)
    #9 closure #1 in closure #1 in Array.parallelMap<A>(transform:) Array+SwiftLint.swift:60 (SwiftLintPackageTests:x86_64+0x16ea1d)
    #10 partial apply for closure #1 in closure #1 in Array.parallelMap<A>(transform:) <compiler-generated> (SwiftLintPackageTests:x86_64+0x16f2ab)
    #11 partial apply for thunk for @callee_guaranteed (@unowned Int) -> () <null>:2670432 (libswiftDispatch.dylib:x86_64+0xeaf0)
    #12 _dispatch_client_callout2 <null>:2670432 (libdispatch.dylib:x86_64+0x3671)
```

You can run this yourself locally with `make test_tsan`.

fredpi:
@jpsim Thanks for giving me a direction how to make the tests work - they are all passing now 👍

jpsim:
Thinking through this feature a bit, I wonder if we should consider how this could live alongside #2729.

Can we align these to use the same mechanism for pulling in external (sub)configurations?

Seems like we should be able to align them, and maybe even come up with a single unified solution.

acecilia:
Would it be attainable in this PR to pass multiple swiftlint configuration files through the command line?

fredpi:
@jpsim @acecilia Sorry for my late reply ;)

Your suggestions to combine this feature with a `parent_config` configuration and multiple configuration files via the command line sound reasonable and I will try to implement them.

### Graph

However, this may require more refactoring, as all configuration files used to build the final configuration and their relationship must be stored, resulting in a graph that has some requirements to represent a valid configuration:

* The graph must be cycle-free. This is what this PR already implemented: It's impossible to create cycles in `sub_config` references. This behavior must now be extended to the new configuration options. 
* Each vertex must be the origin of either zero edges or one edge, and the target of either zero edges or one edge, meaning it's not possible to have multiple `sub_config` references. Also, a file that specifies a `parent_config` shall not be referenced as a `sub_config` from somewhere else, as the vertex representing the config would then be the target of two edges. One could circumvent this restriction by defining a precedence (e. g. considering `parent_config` earlier than `sub_config`), but I guess that's misleading.

### Configuration Interferences

How would you handle interferences between the `sub_config` / `parent_config` / command line configurations: If a user passes `config1.yml config2.yml config3.yml` via the command line, should we ignore `sub_config` / `parent_config` references within these files? I'm rather in favor of combining all configurations, checking for validity and breaking without any recovering mechanism if the resulting graph is invalid. At least, this is what the current implementation does when it detects cycles...

### Nested Configurations

Another issue comes with nested configurations: Consider a setup with the two configurations A, B on the root file level and configuration C one level deeper. A has B as its `sub_config`, C has B as its `parent_config`. Now, with the current implementation, a nested sub config is different to a `sub_config`: It's merged after the top-level-config has been fully merged already (instead of collecting everything and merging thereafter). Still, the original configs used to merge the top-level-config must be referenced, so that we can check a nested config doesn't reference a config used to build the top-level-config.

Previously, this could be ensured by only allowing `sub_config` references in the same directory, but I guess it's also reasonable to allow references in different directories and then check that the nested configuration has not been used to build the top-level config already.

fredpi:
Unfortunately, I'm on holiday for the next few weeks, but I'll continue with the implementation thereafter

jpsim:
I'm assuming you're still working through the CI failures, let me know if you'd like some help.

fredpi:
@jpsim Thanks for offering help – I'm still in the process of completely rearranging this PR to also allow for parent configs, multiple configs via the command line and remote configs. This will still take some time and I will drop a notice when I think I'm done ;)

fredpi:
@jpsim @acecilia

This PR is finally approaching its completion!

Only tests (+ bugfixes?) and documentation are missing as of now. Both of them need to be taken seriously and I will look and try if I can take care of them this Saturday.

Until then, someone could finally have look at [this PR](https://github.com/realm/SwiftLint/pull/2765) (😜) or get familiar with the features introduced here:
- Specify `child_config` / `parent_config`. This even works recursively, as long as there are no cycles and no ambiguities.
- Nested configurations got a speed bump and of course now ignore configs that were already used to build the primary config (using `child_config` / `parent_config` references)
- You can pass multiple configurations via the command line that are interpreted as a hierarchy.
- Rule Configurations now get merged properly (meaning that child rule configurations are only chosen if they are non-default (allowing parent rule configurations to persist if a child only uses the default configuration))
- You can specify a remote `child_config` / `parent_config`! This is just easy as dropping in a url that starts with `http://`or `https://` where normally there would be the path to a local file. Remote configs get cached (in a manner that git doesn't see it). There are timeouts to loading remote configs, that can also be customized using `remote_timeout` / `remote_timeout_if_cached`
- The whole Configuration clutter has been cleaned up a lot internally in a way that I hope everyone is happy with.

fredpi:
`SwiftLintFrameworkTests.IntegrationTests testSimulateHomebrewTestWithDisableSourceKit` seems to fail with `swift test`, but is successful when run in Xcode. Any idea what that could mean?

Source/SwiftLintFramework/Helpers/Reachability.swift
```diff
@@ -0,0 +1,40 @@
...
+#if !os(Linux)
+import SystemConfiguration
+#endif
+
+public class Reachability {
+    /// Returns whether the device is connected to a network.
+    /// On Linux, this always evaluates to true.
+    public static func isConnectedToNetwork() -> Bool {
+        #if os(Linux)
+        return true
```
fredpi: @jpsim Do you have a better idea how to do this considering `SystemConfiguration` isn't available on Linux...? In the end, it doesn't matter too much, as there's also still the timeout – this just makes it faster to recognize there's no network...

Source/SwiftLintFramework/Extensions/Configuration+Remote.swift
```diff
@@ -0,0 +1,240 @@
...
+
+    private mutating func handleWrongData(
+        urlString: String,
+        cachedFilePath: String?,
+        taskDone: Bool,
+        timeout: TimeInterval
+    ) throws -> String {
+        if let cachedFilePath = cachedFilePath {
+            if taskDone {
+                queuedPrint("Unable to load remote config from \"\(urlString)\". Using cached version as a fallback.")
```
fredpi: @jpsim Should we drop a warning in cases where a new version from the remote could not be loaded or just print it?

I'm in favor of printing so that you're not spammed with warnings if running SwiftLint in an offline environment. In the end, that's the whole thing about caching, that it's there if needed...

fredpi:
`/__w/1/s/Tests/SwiftLintFrameworkTests/LinterCacheTests.swift:351: error: LinterCacheTests.testDetectSwiftVersion : XCTAssertEqual failed: ("5.1.3") is not equal to ("5.1.2")`

This test failure does not seem to be related to this PR specifically...

Source/SwiftLintFramework/Extensions/Configuration+Remote.swift
jpsim: I'm leaning a bit more towards a warning than just a log. Otherwise the output of SwiftLint will be incorrect even for a valid configuration based on the host machine's conditions.

jpsim:
Hi @fredpi 👋 

Thanks for your continued work on this, even with my absence. I don't have a whole lot of time for OSS work lately 😞

Where is this at now? Are there top-level action items left to tackle or is it ready for review?

Source/SwiftLintFramework/Helpers/Reachability.swift
jpsim: Fine to assume always reachable on Linux. 👍

jpsim:
Are the violations that were reported as fixed by OSSCheck valid (https://github.com/realm/SwiftLint/pull/2824#issuecomment-515234516)? As far as I can follow, this PR shouldn't have any impact on configurations that don't use `sub_config` or `parent_config`:

* This PR fixed a violation in Moya: <a href="https://github.com/Moya/Moya/blob/939b76fdd1483484f50d37172a930400475fa432/Tests/SignalProducer+MoyaSpec.swift#L575:1">/usr/local/var/buildkite-agent/builds/coastalelectricgaserver-com-1/swiftlint/swiftlint/osscheck/Moya/Tests/SignalProducer+MoyaSpec.swift:575:1: warning: File Line Length Violation: File should contain 400 lines or less: currently contains 575 (file_length)</a>
* This PR fixed a violation in Moya: <a href="https://github.com/Moya/Moya/blob/939b76fdd1483484f50d37172a930400475fa432/Tests/Observable+MoyaSpec.swift#L598:1">/usr/local/var/buildkite-agent/builds/coastalelectricgaserver-com-1/swiftlint/swiftlint/osscheck/Moya/Tests/Observable+MoyaSpec.swift:598:1: warning: File Line Length Violation: File should contain 400 lines or less: currently contains 598 (file_length)</a>
* This PR fixed a violation in Moya: <a href="https://github.com/Moya/Moya/blob/939b76fdd1483484f50d37172a930400475fa432/Tests/Single+MoyaSpec.swift#L585:1">/usr/local/var/buildkite-agent/builds/coastalelectricgaserver-com-1/swiftlint/swiftlint/osscheck/Moya/Tests/Single+MoyaSpec.swift:585:1: warning: File Line Length Violation: File should contain 400 lines or less: currently contains 585 (file_length)</a>

fredpi:
@jpsim It's almost finished, meaning there are only some tests left to be added. So I think it's already ready for a review!

Regarding the resolved violations in Moya that you asked about: I wrote about them in the second comment in this thread:

> Just a note regarding the changes / fixes to rule configuration: As SwiftLintBot noted, with this new implementation, 3 warnings in Moya are now fixed. In fact, previously those have been false positives, as in the nested `.swiftlint.yml` covering these files there's the following configuration for the file_length rule:
> 
> ```yaml
> file_length:
>   warning: 1000
> ```
> 
> As the parent rule configuration was used for merging and there was no smart mechanism to decide which configuration to choose, the configuration originating from the nested configuration was ignored. With this PR, that's now fixed.

So I think it's good that these warnings are now fixed with this PR 😀

fredpi:
Closing this, as I created a new PR for these changes: #3058

Issue #2737, opened 2019-04-30
unused_import produces false positive when only operators from a module are used


biboran:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

`UnusedImportRule` produces a false positive result when a file that imports a certain module uses only the operators defined by the imported module. 

It seems that the `File.synthaxMap` property doesn't contain the operators and thus the `cursorInfo` for the operator is never queried.   

### Environment

I made an [example repo](https://github.com/biboran/synthax-bug-example) which reproduces the bug. Here is how to run it. 
Build the library and invoke swiftlint:
```bash
xcodebuild -project ./synthax-bug.xcodeproj -scheme synthax-bug-Package > xcodebuild.log
swiftlint analyze ./Sources/MainLib/MainLib.swift --compiler-log-path ./xcodebuild.log
```
A warning of unused import should appear even though the `MainLib` uses an overload of a + operator. Using any other function from the `Extensions` module dismisses the violation.

jpsim:
Yeah, we're aware of this but haven't been able to come up with any workarounds. Can you think of a way to identify when operators are used from a given module?

biboran:
What's interesting is that by querying  a `source.request.cursorinfo` with an offset of 160 in MainLib.swift (where the + operator is) I am able to get the following structure:
```json
{
  "key.annotated_decl": "<Declaration>static func + (lhs: <Type usr=\"s:Sq\">Optional</Type>&lt;<Type usr=\"s:s4Voida\">Void</Type>&gt;, rhs: <Type usr=\"s:Sq\">Optional</Type>&lt;<Type usr=\"s:s4Voida\">Void</Type>&gt;)</Declaration>",
  "key.containertypeusr": "$Ss4VoidaSgmD",
[...]

PR #2840, opened 2019-08-19
#2737 - Fix unused_imports false positive when only operators from the module are used

biboran:
As was discussed #2737, I utilized the `indexsource` request to look up the operators and fetch the cursorInfo per operator. In the implementation I refrained from using `usr` to look up the operator because SourceKit [doesn't support](https://github.com/apple/swift/blob/5add16804272b4df917da15c46eb6f28d826d656/tools/SourceKit/lib/SwiftLang/SwiftSourceDocInfo.cpp#L1799) fetching cursorInfo by `usr` when it comes from C (but it can still be fetched by offset).

I also made [an alternative implementation](https://github.com/realm/SwiftLint/compare/master...biboran:fix-unused-imports-for-operators-alternative) which uses the `indexsource` request instead of the `editor.open`. It seems to work with the unit tests but I am not 100% sure it doesn't introduce a regression since there is no oss-check for analyzer rules.

I also updated [the example](https://github.com/biboran/synthax-bug-example) to better reflect the original issue in case you want to test the changes manually

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
@@ -28,6 +28,12 @@ public struct UnusedImportRule: CorrectableRule, ConfigurationProviderRule, Anal
             """
             import UnknownModule
             func foo(error: Swift.Error) {}
+            """,
+            """
+            import Foundation
+            import ObjectiveC
+            let 👨‍👩‍👧‍👦 = #selector(NSArray.contains(_:))
+            👨‍👩‍👧‍👦 == 👨‍👩‍👧‍👦
             """
         ],
         triggeringExamples: [
@@ -165,7 +171,8 @@ private extension File {
         var imports = Set<String>()
         var usrFragments = Set<String>()
         var nextIsModuleImport = false
-        for token in syntaxMap.tokens {
+        let tokens = syntaxMap.tokens
+        for token in tokens {
             guard let tokenKind = SyntaxKind(rawValue: token.type) else {
                 continue
             }
@@ -195,10 +202,9 @@ private extension File {
                 }
             }
 
-            if let usr = cursorInfo["key.modulename"] as? String {
-                usrFragments.formUnion(usr.split(separator: ".").map(String.init))
-            }
+            appendUsedImports(cursorInfo: cursorInfo, usrFragments: &usrFragments)
         }
+
         // Always disallow 'import Swift' because it's available without importing.
         usrFragments.remove("Swift")
         var unusedImports = imports.subtracting(usrFragments)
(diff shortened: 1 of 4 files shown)


Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
```diff
@@ -28,6 +28,12 @@ public struct UnusedImportRule: CorrectableRule, ConfigurationProviderRule, Anal
             """
             import UnknownModule
             func foo(error: Swift.Error) {}
+            """,
+            """
+            import Foundation
+            import ObjectiveC
+            let 👨‍👩‍👧‍👦 = #selector(NSArray.contains(_:))
+            👨‍👩‍👧‍👦 == 👨‍👩‍👧‍👦
```
biboran: `ObjectiveC` module introduces an overload of the `==` operator:
```swift
public static func == (lhs: Selector, rhs: Selector) -> Bool
```
and even though `Selector` is introduced by the `ObjectiveC` module it can be indirectly created with `#selector`. This operator overload is the only way I found to reproduce the original issue in the unit tests.

jpsim:
Great to see this 😍

Give me some time to take it for a spin and review.

biboran:
I tested the alternative implementation which uses `indexsource` exclusively and found some false positives. I guess lets leave it out of the scope of this issue.

Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
```diff
@@ -195,30 +202,119 @@ private extension File {
...
 
-                return (module, contentsNSString.range(of: "import \(module)\n"))
+                return (module, contents.range(of: "import \(module)\n"))
             }
             .sorted(by: { $0.1.location < $1.1.location })
     }
 
-    private func containsAttributesRequiringFoundation() -> Bool {
+    // Operators are omitted in the editor.open request and thus have to be looked up by the indexsource request
+    func operatorImports(arguments: [String], processedTokenOffsets: Set<Int>) -> Set<String> {
```
jpsim: Can you mark these as `private` where possible?

jpsim:
Code looks good, I'm validating this now on a large closed source project with a number of files where only operator overloads are used.

Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
biboran: Since these methods are a part of a private extension, the compiler warns that using private is redundant in this case. Do you still want me to mark them private?
I could also make the extension non-private and the methods private for the sake of consistency.

jpsim: Oh I didn’t realize these were in a private extension. All good!

jpsim:
One thing that running on a large codebase that this identified was that subscript extensions still aren't detected. Not a blocker to merging this but I'm curious to see if you think this approach can be extended to catch those? For example, if this is defined in ModuleA:

```swift
extension Collection where Index == Int {
    public subscript(safe index: Int) -> Iterator.Element? {
        return index < self.count && index >= 0 ? self[index] : nil
    }
}
```

and ModuleB has this:

```swift
import ModuleA

let array = ["a", "b", "c"]
_ = array[safe: 0] // equivalent to array.first
```

Then ModuleA will still be listed as an unused import.

jpsim:
ok, finally validated this. Looks great, thank you so much for owning this @biboran ! ❤️

PR #2841, opened 2019-08-25
Add contains_over_filter_count opt-in rule

### LINKED ISSUE #2803: Contains over `filter{}.count > 0`
### New Issue Checklist

### New rule request

On collections we can filter the elements based on a predicate, count them and compare against zero to check if there is at least one element satisfying that predicate. However, we can use the `contain` API to achieve the same result.

What should trigger?
```swift
struct Foo {
    let bar: Bool
}

let foos = [Foo(bar: false), Foo(bar: true)]
let canBar: Bool = foos.filter { $0.bar }.count > 0
```

```swift
struct Foo {
    let bar: Bool
}

let foos = ["1": Foo(bar: false),"2": Foo(bar: true)]
let canBar = foos.filter { $1.bar }.count > 0
```

```swift
let canNumber2 = [1, 2, 3].filter { $0 == 2 }.count > 0
```

```swift
let numbers: Set<Int> = [1, 2, 3]
numbers.filter { $0 == 2 }.count > 0
```

What should not trigger?

```swift
struct Foo {
    let bar: Bool
}

let foos = [Foo(bar: false), Foo(bar: true)]
let canBar: Bool = foos.contains(where: { $0.bar })
```

```swift
struct Foo {
    let bar: Bool
}

let foos = ["1": Foo(bar: false),"2": Foo(bar: true)]
let canBar = foos.contains { $1.bar }
```

```swift
let canNumber2 = [1, 2, 3].contains(2)
```

```swift
let numbers: Set<Int> = [1, 2, 3]
numbers.contains(2)
```

The rules should not be configurable and could be enabled by default.

marcelofabri:
Fixes #2803

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -13,6 +13,7 @@ opt_in_rules:
   - closure_end_indentation
   - closure_spacing
   - collection_alignment
+  - contains_over_filter_count
   - contains_over_first_not_nil
   - empty_count
   - empty_string
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@
   [atfelix](https://github.com/atfelix)
   [#2791](https://github.com/realm/SwiftLint/issues/2791)
 
+* Add `contains_over_filter_count` opt-in rule to warn against using
+  expressions like `filter(where:).count > 0` instead of `contains(where:)`.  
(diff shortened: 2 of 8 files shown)


jpsim:
We should do the same for `filter(...).isEmpty`

Issue #2820, opened 2019-07-22
`discouraged_optional_collection` false positive in property observer


arielelkin:
### New Issue Checklist

- [x ] Updated SwiftLint to the latest version
- [ x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

This triggers a `discouraged_optional_collection` violation:

```swift
class MyClass {
    var items: [Int]? { 
        didSet {
            print("Hello")
        }
    }
}
```

It shouldn't in cases where there are a property observers.

marcelofabri:
Why having a property observer should change the rule behavior?

arielelkin:
on second thoughts i can't see why the rule's behaviour should change in that case. I think the rule itself was odd because it can be useful for a collection to be `nil`. This rule makes sense if you replace nil checks with `isEmpty` checks. Would it be useful to send a PR updating rules.md so that it adds a line doing an `isEmpty` check?

marcelofabri:
The problem with `nil` collections is that you have two empty states: `nil` and `.isEmpty`. `Rules.md` is autogenerated, so you would need to update the respective Rule file.

marcelofabri:
Closing this, but feel free to submit a PR improving the docs!

PR #2849, opened 2019-08-28
Update DiscouragedOptionalCollectionRule.swift

arielelkin:
addressing https://github.com/realm/SwiftLint/issues/2820


arielelkin:
507 warnings is daunting.. is it just because I forgot to update the changelog?

jpsim:
> 507 warnings is daunting.. is it just because I forgot to update the changelog?

Ah sorry for the noise, this is because we run a lint with master and the PR branch and compare the output to know if we've introduced or fixed violations in a variety of open source projects.

So changing the wording of a rule description is treated as removing all previous violations and re-introducing them as "new" ones.

It's a limitation in our CI tool basically.

jpsim:
I'm not sure this is what I'd recommend in the rule description though.

The fundamental challenge when dealing with optional collections is that the semantic meaning of the member being `nil` isn't immediately obvious, especially when compared to the collection being empty.

---

Finally, please run the tests locally in Xcode before pushing to this PR again, since this is failing due to `Rules.md` not being up to date (which is done automatically for you when running unit tests in Xcode) and also due to the line length being too long.

PR #2853, opened 2019-08-31
Don't trigger `explicit_type_interface` on if case statements

cmillani:
The `explicit_type_interface` is being triggered on `if case let` statements with enum with types.

For exemple, this code will trigger a warning:

```swift
enum Foo {
  case caseOne(propOne: Int)
}

let testA: Foo = .caseOne(propOne: 2)

if case let Foo.caseOne(propOne) = testA { }
```

while this one will not:

```swift
enum Foo {
  case caseOne(propOne: Int)
}

let testA: Foo = .caseOne(propOne: 2)

switch testA {
  case let .caseOne(propOne): break
}
```

On `if case Enum.case(variable) = enum` the parent of the variable structure is of `call` type, because of this the current validation for `switch` and `cases` are not working

I was in doubt if a better solution would be:
* Verify if the parent of the variable declaration is a call, and this caracterizes a `if case let`
* Verify if the words before call are `if case`

I think the first one seems to follow better the pattern I saw on the code, but I am not comfortable assuming every combination of "Parent: Call, child: Var" is indeed a `if case let` situation. With this one, checking if the grandparent of the var is an if statement should also be good, what makes it even more complicated.

The second one seems a little bit inefficient, but is clear when reading and seems more assertive to me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,9 @@
 
 #### Bug Fixes
 
-* None.
+
+* Don't trigger `explicit_type_interface` on `if case let` statements.  
+  [Carlos Millani](https://github.com/cmillani)
 
 ## 0.35.0: Secondary Lint Trap
 
--- a/Source/SwiftLintFramework/Models/MasterRuleList.swift
+++ b/Source/SwiftLintFramework/Models/MasterRuleList.swift
@@ -1,4 +1,4 @@
-// Generated using Sourcery 0.17.0 — https://github.com/krzysztofzablocki/Sourcery
+// Generated using Sourcery 0.16.2 — https://github.com/krzysztofzablocki/Sourcery
 // DO NOT EDIT
 
 public let masterRuleList = RuleList(rules: [
--- a/Source/SwiftLintFramework/Rules/Idiomatic/ExplicitTypeInterfaceRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/ExplicitTypeInterfaceRule.swift
@@ -95,8 +95,8 @@ public struct ExplicitTypeInterfaceRule: OptInRule, ConfigurationProviderRule {
             (!configuration.allowRedundancy ||
                 (!dictionary.isInitCall(file: file) && !dictionary.isTypeReferenceAssignment(file: file))
             ),
-            !dictionary.isIfCaseCall(file: file, parentStructure: parentStructure),
             !parentStructure.contains([.forEach, .guard]),
+            !parentStructure.isIfCaseCall(),
             !parentStructure.caseStatementPatternRanges.contains(offset),
             !parentStructure.caseExpressionRanges.contains(offset),
             !file.captureGroupByteRanges.contains(offset) else {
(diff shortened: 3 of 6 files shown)


cmillani:
Although the fix worked for the sample inserted in the `nonTriggeringExamples`,  running this build on my project (where I am having this kind of warning) apparently did not fix the issue. 

I am looking into it to check if I am building correctly and will push another commit as soon as I have an update :)

marcelofabri:
Thanks for the contribution! I'd look into fixing the existing logic (`caseStatementPatternRanges` and `caseExpressionRanges`) instead of adding another condition. This also has the benefit of using the AST instead of regexes, which is usually faster and more reliable (less false positives).

cmillani:
Thanks, I will look into it! 

I had some trouble using the AST because in these scenarios we do not have a `case`, but a `source.lang.swift.expr.call` with a `source.lang.swift.decl.var.local` inside it.

I do believe the "if case" statement is the only place this will happen, but since I know very little of the AST, I thought it was safer to use regex.

I would love to get an opinion about this, do you think it is safe to use this pattern in the AST (`var` inside a `call`) to detect an `if case` statement?

cmillani:
I found another scenario that my first commit did not caught, if the if case has more than one property, for example:

```swift
enum Foo {
  case caseOne(propOne: Int, propTwo: Int)
}

let testA: Foo = .caseOne(propOne: 2, propTwo: 4)

if case let Foo.caseOne(propOne, _) = testA { }
```

The AST returns a `source.lang.swift.expr.argument` as the parent of the variable in this case.

Running `oss-check` gave me:

Message: Linting Aerial with this PR took 3.98s vs 4.08s on master (2% faster)
Message: Linting Alamofire with this PR took 9.03s vs 9.63s on master (6% faster)
Message: Linting Firefox with this PR took 24.28s vs 27.97s on master (13% faster)
Message: Linting Kickstarter with this PR took 51.33s vs 72.97s on master (29% faster)
Message: Linting Moya with this PR took 4.63s vs 8.5s on master (45% faster)
Message: Linting Nimble with this PR took 4.38s vs 5.65s on master (22% faster)
Message: Linting Quick with this PR took 1.79s vs 2.41s on master (25% faster)
Message: Linting Realm with this PR took 7.85s vs 10.94s on master (28% faster)
Message: Linting SourceKitten with this PR took 2.92s vs 4.62s on master (36% faster)
Message: Linting Sourcery with this PR took 8.19s vs 11.34s on master (27% faster)
Message: Linting Swift with this PR took 113.46s vs 84.8s on master (33% slower)
Message: Linting WordPress with this PR took 93.44s vs 88.15s on master (6% slower)

And handling the output of it with `grep explicit_type | cut -d ":" -f 2 | uniq -c` gives:

```
   1  This PR fixed a violation in Firefox
  21  This PR fixed a violation in Kickstarter
  11  This PR fixed a violation in Moya
   1  This PR fixed a violation in Nimble
   1  This PR fixed a violation in Sourcery
  51  This PR fixed a violation in WordPress
```

I could not run only the script at the moment, so the performance data may not be that accurate, if necessary I can run it again later and post the full output here :)

cmillani:
Wow, didn't know the comment from Danger would auto update :)

But just noticed that this also fixes warnings triggered in capture lists, where apparently it is not possible to explicitly define the type (this is the case of line 113 of NotificationViewController.swift in WordPress, for example)

danilobecke:
let's get this in 🙏

marcelofabri:
Can you please rebase it against master and add some tests for the capture list in closures? Thanks!

cmillani:
Yes, sure, I'll rebase and add the tests! :)

cmillani:
Sorry it took so long! 

I just rebased the branch and when I was to implement the tests for the Capture groups, I saw there is already a `testCaptureGroup` function with some tests.

I still implemented locally some tests, following the same patterns of some fixed violations according to the Bot (capture groups with one word or with equal symbol), but those did not pass.

I think it would make more sense to work on this on a separate issue, what do you think @marcelofabri ? I could start working on this tomorrow morning :)

One more thing, I also included a test to catch definitions, such as this one [here](https://github.com/Moya/Moya/blob/cdef4ea7ecb7818491c5b710a718077d9513dcbd/Sources/Moya/MoyaProvider+Defaults.swift#L19:46), that is one of the corrections of this PR. But I had to disable type_body_length on ExplicitTypeInterfaceRuleTests, is this ok?

Issue #1188, opened 2017-01-13
Add a rule for NPath Complexity


nhgrif:
Currently Swiftlint has a rule for Cyclomatic complexity, but no rule for NPath Complexity.  A rule should be added for NPath Complexity configurable in the same sorts of ways that Cyclomatic Complexity is configured.

PR #2854, opened 2019-08-31
Add `npath_complexity` rule.

niilohlin:
Solves https://github.com/realm/SwiftLint/issues/1188. 

NPath Complexity is related to Cyclomatic Complexity but instead of adding up the different possible branching statements it tries to calculate the possible number of code paths in a given function.

**Known limits:** This implementation does not take `try/catch`, `return` or multi-clause conditions (`||`, `&&`, and `,`) statements into consideration when calculating NPath complexity. Since it is not available in the AST.

The default values are inspired by [phpmd](https://phpmd.org/rules/codesize.html) and the test values are verified against phpmd.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Metrics/NPathComplexityRule.swift
+++ b/Source/SwiftLintFramework/Rules/Metrics/NPathComplexityRule.swift
@@ -11,20 +11,68 @@ public struct NPathComplexityRule: ASTRule, ConfigurationProviderRule, OptInRule
         description: "The number of paths of function bodies should be limited.",
         kind: .metrics,
         nonTriggeringExamples: [
-            "func f1() {\nif true {\nfor _ in 1..5 { } }\nif false { }\n}",
-            "func f(code: Int) -> Int {\n" +
-                "switch code {\n case 0: fallthrough\n case 0: return 1\n case 0: return 1\n" +
-                " case 0: return 1\n case 0: return 1\n case 0: return 1\n case 0: return 1\n" +
-                " case 0: return 1\n case 0: return 1\n default: return 1\n }\n}",
-            "func f1() {" +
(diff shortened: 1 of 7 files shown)


marcelofabri:
My (limited) understanding about NPath complexity is that it should also count multi-clause conditions (`&&`, `||` and `,` in case of Swift). 

Since it's an opt-in rule we can choose to handle that as a current limitation, but I just wanted to raise this.

niilohlin:
Yes, this is a limitation of the AST. The only information we get in the dictionary is `"source.lang.swift.structure.elem.condition_expr"` and not what kind of expression it is. The same is true for the `cyclomatic_complexity` rule. It should also include multi-clause conditions.

Source/SwiftLintFramework/Rules/Metrics/NPathComplexityRule.swift
```diff
@@ -0,0 +1,98 @@
...
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "npath_complexity",
+        name: "NPath Complexity",
+        description: "The number of paths of function bodies should be limited.",
+        kind: .metrics,
+        nonTriggeringExamples: [
+            "func f1() {\nif true {\nfor _ in 1..5 { } }\nif false { }\n}",
```
marcelofabri: nitpick: please use multiline string literals and properly format the examples. I'd be nice to also avoid `;` and use different values inside a switch.

marcelofabri: maybe using the same examples from the tests would be nice as they look more like "real code". I'd also keep the comment about the expected complexity inside the example as it's a good way to help people understand what NPath complexity means.

PR #2858, opened 2019-09-03
Cleanup examples for cyclomatic complexity

niilohlin:
Cleanup example code in the `cyclomatic_comlexity` rule, based on feedback from https://github.com/realm/SwiftLint/pull/2854


Source/SwiftLintFramework/Rules/Metrics/CyclomaticComplexityRule.swift
```diff
@@ -12,22 +12,75 @@ public struct CyclomaticComplexityRule: ASTRule, ConfigurationProviderRule {
...
+                    }
+                }
+                if false {
+                }
+            }
+            """,
+            """
+            func switchFunction(code: Int) -> Int {
+                switch code {
+                case 0: fallthrough
```
marcelofabri: can you please change the `case`s to avoid repetition as we're changing them anyway?

marcelofabri:
@niilohlin Thanks for the PR! Can you fix the conflict and address my comment? Then we should be good to merge it.

PR #2859, opened 2019-09-04
Add optional support for Missing Import Result

ghv:
Swift 5 and 5.1 compiled code (including Swift 4, 4.1, 4.2 code compiled by Swift 5) all have access to the intrinsic Result type. If your code still depends on the `Result` (v4.0.1 and earlier) framework, then you must import it or your code will either not compile or silently compile with the intrinsic `Result` type. Turn this rule on in a project that depends on this framework to avoid missing this crucial import. NB: Version 5.0 of `Result` makes it compatible with Swift 5's `Result` type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -438,7 +438,7 @@ This is the last release to support building with Swift 5.0.x.
   using `flatMap` over `map { ... }.reduce([], +)`.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2883](https://github.com/realm/SwiftLint/issues/2883)
-  
+
 * Add autocorrection to `syntactic_sugar`.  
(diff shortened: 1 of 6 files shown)


ghv:
This PR is ready for review. My force-pushes have been just to rebase with the current `master` to resolve merge conflicts. Thanks!

ghv:
Apologies. I was rebasing+force pushing a number of times to keep up with conflicts. Unfortunately, I skipped some steps: running unit tests and checking if I broke the build above. Had I done that, I would have realized there were some breaking refactoring that needed to be addressed.

This time I did all the steps and will keep an eye on the above build. Please review, I would appreciate some comments.

Thanks!

ghv:
Hi @marcelofabri and @jpsim can you folks look at this PR? I would love to get this in there. Thanks!

Issue #2760, opened 2019-05-15
vertical_whitespace_between_cases rule is not handling 1 edge case properly


LeonidKokhnovich:
### New Issue Checklist

- [+] Updated SwiftLint to the latest version
- [+] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug
When vertical_whitespace_between_cases is enabled and I run "swiftlint autocorrect", the following code is not updated:
```swift
public enum ConnectionState: String {
    case idle
    case connecting

    init?(from string: String) {
        switch string {
        case "Idle":
            self = .idle
        case "Connecting":
            self = .connecting

        default:
            return nil
        }
    }
}
```
The expected code would be:
```swift
public enum ConnectionState: String {
    case idle
    case connecting

    init?(from string: String) {
        switch string {
        case "Idle":
            self = .idle

        case "Connecting":
            self = .connecting

        default:
            return nil
        }
    }
}
```

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint autocorrect
Loading configuration from '.swiftlint.yml'
Correcting Swift files at paths 
...
Correcting 'FortiExtensionDevice.swift' (26/368)
...
Done correcting 368 files!
```
FortiExtensionDevice.swift - file where the enum is declared.

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 
0.32.0
* Installation method used (Homebrew, CocoaPods, building from source, etc)?
homebrew, pods
* Paste your configuration file:
```yml
opt_in_rules:
  - vertical_whitespace_between_cases
```

* Which Xcode version are you using (check `xcode-select -p`)?
Xcode 10.2

PR #2865, opened 2019-09-05
fixed `vertical_whitespace_between_cases` rule containing strings

a-25:
Fixes https://github.com/realm/SwiftLint/issues/2760

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,10 @@
 
 #### Bug Fixes
 
-* None.
+* Fixed `vertical_whitespace_between_cases` rule containing strings.
+  [Andrey Uryadov](https://github.com/a-25)
+  [#2760](https://github.com/realm/SwiftLint/issues/2760)
+
 
 ## 0.35.0: Secondary Lint Trap
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@
 
 #### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -21,7 +21,10 @@
 
 #### Bug Fixes
 
-* None.
+* Fixed `vertical_whitespace_between_cases` rule containing strings.
```

a-25: @marcelofabri: fixed.
As far as I understand, it is OK that some checks failed at https://github.com/realm/SwiftLint/pull/2865#issuecomment-528404066 ? The check assumes that this rule is turned on while the code is written without such assumption.

marcelofabri: It's expected - they serve as a sanity check so we can manually inspect them and see if we're introducing any false positives/negatives

a-25: @marcelofabri: fixed trailing spaces, is there any other problems with the code?

marcelofabri: I think this would introduce false positives on strings (which we might be ok with since this is opt-in). Can you add a test that contains a string a switch as a string?

a-25: @marcelofabri: Added a test example here: https://github.com/realm/SwiftLint/pull/2865/commits/90a235971393b2d06f603be47780ac05e5a7446a. Tests are running correctly. Is this test ok?

a-25: @marcelofabri: any updates about this issue?

a-25:
@marcelofabri: are you going to merge this?

PR #2871, opened 2019-09-22
Add `todo_attribution` opt-in rule

hectr:
This rule enforces TODO and FIXME comments to be attributed to their owner or to a related issue from the issue tracker.

As I saw in another pull request, I have filled out the *New rule request* info below.

# New rule request

> 1. Why should this rule be added? Share links to existing discussion about what the community thinks about this.

TODO and FIXME comments are prone to overuse. That made them very unpopular (e.g. [Todo Comments Considered Harmful](http://wiki.c2.com/?TodoCommentsConsideredHarmful), [The case against TODO](http://wordaligned.org/articles/todo)).

Even SwiftLint complains about them: `"TODOs and FIXMEs should be resolved"`.

But it should be noted that, in quite a few development teams it is a common practice to split a feature into several small pull requests that are easier to code review. This technique may lead to a situation where you need to leave a TODO comment in the code of one PR, that will be resolved in the next.

There may also be situations where you only want to mark (and possibly explain) a non-fatal defect in the code (e.g. a performance issue). This could be the case if the fix required is beyond the scope of the task you are committed to at that particular time.

So, given that a handful of developers and teams will continue to use them, IMO `todo_attribution` rule should be added to ensure that it is easy to keep track of TODOs and FIXMEs. The alternative is risking for them to be lost and forgotten and never resolved.

Best `todo_attribution` selling points are that:

- as it enforces the format, it becomes easier to find TODOs and FIXMEs with a simple text search;
- as it makes mandatory to relate the comment with a developer or a tracked issue, it points you to a source of information about the signaled problem.

> 2. Provide several examples of what would and wouldn't trigger violations.

**Non-triggering examples:**

- `TODO: @hectr do something`
- `TODO: #2871`
- `TODO: @hectr #2871 implement`
- `FIXME: @hectr`
- `FIXME: #2871 fix`

**Triggering examples:**

- `TODO: #`
- `ToDo: @hectr`
- `To-Do`
- `TO-DO: #2871`
- `todo`
- `to-do do something`
- `FIXME: ` ` @hectr assumes valid input`
- `FixMe ...`
- `Fix-Me`
- `FIX-ME`
- `fixme#2871`
- `fix-me`

It seems reasonable to use `SeverityConfiguration` because you may want to be aware of this violation without fixing it immediately.

> 4. Should the rule be opt-in or enabled by default? Why? See README.md for guidelines on when to mark a rule as opt-in.

This rule fits well in the "*A rule that is not general consensus or is only useful in some cases*" category for 2 reasons:

- it enforces a very specific format for TODO and FIXME comments;
- TODOs and FIXMEs are not allowed by `todo` rule, which is enabled by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,10 @@
   [PaulTaykalo](https://github.com/PaulTaykalo)
   [#2901](https://github.com/realm/SwiftLint/issues/2901)  
 
+* Add `todo_attribution` opt-in rule to enforce TODOs and FIXMEs attribution.
+  [Hèctor Marquès](https://github.com/hectr)
+  [#2871](https://github.com/realm/SwiftLint/issues/2871)
+
 #### Bug Fixes
 
 * Fix running analyzer rules on the output of builds performed with
--- a/Rules.md
+++ b/Rules.md
@@ -151,6 +151,7 @@
 * [Switch Case on Newline](#switch-case-on-newline)
 * [Syntactic Sugar](#syntactic-sugar)
 * [Todo](#todo)
+* [Todo Attribution](#todo-attribution)
 * [Toggle Bool](#toggle-bool)
 * [Trailing Closure](#trailing-closure)
 * [Trailing Comma](#trailing-comma)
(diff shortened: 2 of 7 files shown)


christiankm:
I like this, and had created a custom rule for this for our projects, in a specific format. Essentially for use with for example JIRA, where `XXX` should be replaced with the project ID.

```
todo_jira:
    name: "TODO without JIRA reference"
    regex: "(?i)(?=[^\\n]\\/\\/\\s+(TODO|TO DO|FIX|FIXME|FIX ME):)(?!.*XXX\\-\\d+)[^\\n]*"
    message: "Do not add TODOs without a JIRA reference. Create new JIRA Task and add Issue Id to comment."
    severity: warning
```

I'll try to review and give comments when I get a chance.

hectr:
> ```
> todo_jira:
>     name: "TODO without JIRA reference"
>     regex: "(?i)(?=[^\\n]\\/\\/\\s+(TODO|TO DO|FIX|FIXME|FIX ME):)(?!.*XXX\\-\\d+)[^\\n]*"
>     message: "Do not add TODOs without a JIRA reference. Create new JIRA Task and add Issue Id to comment."
>     severity: warning
> ```

Thanks for sharing the snippet @christiankm!

I have been using a similar custom rule for some time now. I needed it to be more permissive, allowing not only an issue/task identifier (`#`), but also a user handle (`@`):

```
todo_attribution:
    name: "Todo Attribution"
    regex: '(?!\b(?:TODO|FIXME): (?:@[:alnum:]+|#[:alnum:]+)\b)\b(?:(?:[tT][oO][-_]{0,1}[dD][oO])|(?:[fF][iI][xX][-_]{0,1}[mM][eE]))\b'
    message: "TODOs and FIXMEs should be attributed to an owner or related issue"
    severity: error
```

Only issue I had with it so far is that, a few times, I also wanted to use the `ExpiringTodo` rule. And I ended up doing this ugly thing:

```
// swiftlint:disable:next todo_attribution
// TODO: [2021-09-01] #1234 whatever I want to be reminded of
```

Nothing that cannot be fixed changing the regex of the custom rule.

PR #2878, opened 2019-09-25
Trigger control_statement when using the not operator

marcelofabri:


jpsim:
FYI there were some false positives even with the latest version you had: https://github.com/apple/swift/blob/56edd08fa29865db6fa4dd86967afd351c527fc9/stdlib/public/Darwin/XCTest/XCTest.swift#L514:5

PR #2885, opened 2019-09-26
Add `flatmap_over_map_reduce` opt-in rule

### LINKED ISSUE #2883: [Rule Request] Prefer flatMap instead of map { }.reduce([], +)
### New Issue Checklist

### New rule request

`flatMap` can be used instead of `map { ... }.reduce([], +)` to avoid iterating the collection twice.

```swift
let foo = bar.map { $0.array }.reduce([], +) // would trigger

let foo = bar.map { $0.int }.reduce(0, +) // ok
let foo = bar.flatMap { $0.array } // ok
``` 

Just the severity.

Opt-in as this can have false positives

marcelofabri:
Fixes #2883

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -27,6 +27,7 @@ opt_in_rules:
   - file_header
   - file_name
   - first_where
+  - flatmap_over_map_reduce
   - identical_operands
   - joined_default_parameter
   - legacy_random
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,11 @@
   [nvanfleet](https://github.com/nvanfleet)
   [#2866](https://github.com/realm/SwiftLint/issues/2866)
 
+* Add `flatmap_over_map_reduce` opt-in rule to prefer
+  using `flatMap` over `map { ... }.reduce([], +)`.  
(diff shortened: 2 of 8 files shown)


jpsim:
Why opt-in?

jpsim:
Also, isn't curious that _nothing_ in OSSCheck triggered this?

marcelofabri:
> Why opt-in?

The other similar rules are opt-in. In theory there could be a type that implements `map` but not `flatMap`

jpsim:
Can this support the `reduce(into:)` variant?

marcelofabri:
> Can this support the `reduce(into:)` variant?

it could, but it's more complicated because we need to analyze the closure (although we probably should also do it for the tradicional `reduce`). I just wanted something quick that would catch the more obvious (and what I thought to be more common) case as this showed up in a PR that I was reviewing today.

marcelofabri:
> Also, isn't curious that _nothing_ in OSSCheck triggered this?

I was able to get a warning on SwiftLint itself by changing

```swift
let allPaths = paths.flatMap { $0 }
```

to 

```swift
 let allPaths = paths.map { $0 }.reduce([], +)
```

So I guess this is just not a common issue? Or at least in this form, using `+` as the parameter.

Issue #2867, opened 2019-09-13
Xcode 11 install via SPM won't compile because of SourceKitten


RamblinWreck77:
Installing SwiftLint via SPM installs dependencies that prevent the project from compiling. 

SourceKitten doesn't seem to consider this their problem:

https://github.com/jpsim/SourceKitten/issues/612

Any proposed solution?

Also, since I'm trying to move from Cocoapods > SPM, I'm running into another issue:

In cocoapods you could do:

```
pod 'SwiftLint', :configurations => ['Debug']
```

And SwiftLint wouldn't get included in your release builds. Is this possible with SPM?

crsantos:
But does bumping to jpsim/SourceKitten#612 's `0.26` fixes this?

MaxHaertwig:
No, still two issues:
<img width="269" alt="Screen Shot 2019-10-08 at 14 19 15" src="https://user-images.githubusercontent.com/9899051/66395037-c80bcb00-e9d6-11e9-9001-8944f66d7600.png">

marcelofabri:
> No, still two issues:
> <img alt="Screen Shot 2019-10-08 at 14 19 15" width="269" src="https://user-images.githubusercontent.com/9899051/66395037-c80bcb00-e9d6-11e9-9001-8944f66d7600.png">

You can't use SwiftLint via SPM on an iOS project like that - see the comment on https://github.com/jpsim/SourceKitten/issues/612#issuecomment-529001750

crsantos:
**Question**: (if someone knows)

Swiftlint used via SPM wasn't even compiling for macOS, but right now, in [`master`](https://github.com/realm/SwiftLint/commit/495407b452e964486027157ca96d7026ec99b733) it compiles perfectly when you add `swiftlint` as an SPM dependency

But we don't have a tag 😬 

When should be the next release?

fruitcoder:
@crsantos after 2h of searching I finally found this issue. It seems that the newest release (0.38.2) still doesn‘t work with SPM. After using `master` Swiftlint finally builds 🎉 thanks for pointing me into the right direction

crsantos:
@fruitcoder Previous tags were working fine, guess it was fine on `0.35.0`+

but if `master` is working and `0.38.2` isn't, was fixed after that tag was created. Same problem as before.

szuniverse:
I tried with the latest master and I got the same errors as here: https://github.com/realm/SwiftLint/issues/2867#issuecomment-539486303
[...]

PR #2891, opened 2019-10-05
Update build instructions for Xcode 11

gobetti:
It took me a couple minutes to figure out how to build it with Xcode 11 so I figured it could be helpful to have specific instructions for it.

**However this should not be merged until https://github.com/realm/SwiftLint/issues/2867 is closed** (might want to explicitly mention in the README for now that building with SPM is not yet supported?)

Note: this PR is intended for SwiftLint contributors or anyone wanting to build it from source. There's https://github.com/realm/SwiftLint/pull/2615 with instructions for end users that want to integrate SwiftLint to their projects using SPM.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,18 @@ running it.
 
 ### Compiling from source:
 
-You can also build from source by cloning this project and running
-`git submodule update --init --recursive; make install` (Xcode 10.2 or later).
+You can also build from source by cloning this project and following the steps according to the Xcode version (10.2 or later)
+
+#### Xcode 10.2.x or 10.3.x
+
+- Run `git submodule update --init --recursive; make install`
+- Build SwiftLint to a Mac device from SwiftLint.xcworkspace
+
+#### Xcode 11
+
+- Run `swift package update; swift package generate-xcodeproj`
+	- Alternatively, make sure to open with Xcode the SwiftLint _folder_ (**not** the xcworkspace) and wait for the Swift Package Dependencies to finish loading.
+- Build SwiftLint to a Mac device (may need to manually select "My Mac" as the target device)
 
 ### Known Installation Issues On MacOS Before 10.14.4
 


README.md
```diff
@@ -60,8 +60,18 @@ running it.
 
 ### Compiling from source:
 
-You can also build from source by cloning this project and running
-`git submodule update --init --recursive; make install` (Xcode 10.2 or later).
+You can also build from source by cloning this project and following the steps according to the Xcode version (10.2 or later)
+
+#### Xcode 10.2.x or 10.3.x
+
+- Run `git submodule update --init --recursive; make install`
```
crsantos: what about the other code generated stuff that `make install` was generating? 🤔 

Things like: https://github.com/realm/SwiftLint/blob/master/Makefile#L41

gobetti: thank you @crsantos, great point! I'll make sure to double-check that once the issue with SourceKitten is resolved, I didn't have any problems without `make install` on Xcode 11 except for SourceKitten but it may be that I would start to have some after SourceKitten succeeds 😅

marcelofabri:
I'm not sure why we need this - I'm using Xcode 11 without any issues 🤔

crsantos:
@marcelofabri Do you use swiftlint as a dependency?
Via SPM? Or Cocoapods?
This is the error the pipeline is triggering: https://dev.azure.com/jpsim/SwiftLint/_build/results?buildId=1080

And that's the same error when I add Swiftlint as a dependency, either via CP, or SPM.

gobetti:
hi @marcelofabri! Sorry I should have been clearer indeed. Yes, you can use Xcode 11 using the same steps as Xcode 10, it's just not the natural way.
I mean, if Xcode 10 is your default Xcode, using `xed .` will open the workspace with clear errors and missing folders. The natural next step for a dev is going to be to look at the README, in case a lazy one like me hasn't done it in the first place.
Now, if Xcode 11 is your default one, it will load the SwiftLint package and start loading the Package.swift dependencies beautifully, which probably replaces most if not all of the Carthage dependencies. I thought it would be nice then to have our suggested build instructions to differ between Xcode 10 and 11, especially if that means one may not need to install third-party tools in order to build the project.

README.md
gobetti: the build succeeded without `make install` when using the branch from https://github.com/realm/SwiftLint/pull/2894 but I suppose `make install` is for actually installing SwiftLint to the machine. I should definitely add it back then, but I'm curious (sorry I'm not very acquainted to SPM), why does it fetch, clone and resolve all of the dependencies again if I have already built the project successfully from inside Xcode 11? It felt to me like the script was doing redundant work that SPM already does when loading the project. If this is all my misunderstanding, I apologize and close this PR 👍

marcelofabri:
>  especially if that means one may not need to install third-party tools in order to build the project

You don't need to install Carthage, because we use it only to fetch dependencies as submodules. You can simply run `script/bootstrap` which in the end will just fetch the submodules.

I don't think we should encourage another build method for now. We probably should fully migrate to SPM at some point instead. Thoughts?

gobetti:
@marcelofabri yeah, I thought we could instruct people to use what's embedded in Xcode 11 instead of updating submodules (given Xcode 11 will do that anyway when you load the package and it's confusing I still need to run other scripts according to the README). A full migration with updated instructions will be even better, but could we have a baby step (updated instructions only) with what we have so far?

PR #2892, opened 2019-10-06
Add autocorrection to the syntactic sugar rule

vani2:
I've added autocorrection to the rule:

```swift
let x: Array<String> /* will be */ let x: [String]
let x: Dictionary<Int, String> /* will be */ let x: [Int: String]
let x: Optional<Int> /* will be */ let x: Int?
let x: ImplicitlyUnwrappedOptional<Int> /* will be */ let x: Int!
func x(a: Array<Int>, b: Int) -> [Int: Any] /* will be */ func x(a: [Int], b: Int) -> [Int: Any]
func x(a: [Int], b: Int) -> ↓Dictionary<Int, String> /* will be */ func x(a: [Int], b: Int) -> [Int: String]
let x = Array<String>.array(of: object) /* will be */ let x = [String].array(of: object)
let x: Swift.Optional<String> /* will be */ let x: String?
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
   using `flatMap` over `map { ... }.reduce([], +)`.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2883](https://github.com/realm/SwiftLint/issues/2883)
+  
 * Add autocorrection to `syntactic_sugar`.  
   [Ivan Vavilov](https://github.com/vani2)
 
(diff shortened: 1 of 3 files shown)


marcelofabri:
Can you run the unit tests locally to update the docs? Thanks!

CHANGELOG.md
```diff
@@ -27,6 +27,8 @@
   using `flatMap` over `map { ... }.reduce([], +)`.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#2883](https://github.com/realm/SwiftLint/issues/2883)
+* Add autocorrection to `syntactic_sugar`.
```
marcelofabri: please add an empty line before this entry

vani2:
> Can you run the unit tests locally to update the docs? Thanks!

I did that, that has no effect. So I've updated `Rules.md` manually.

marcelofabri:
@vani2 I think this was failing because of some other issues in CI that are fixed in `master`. I rebased and force pushed to your branch, hope that's ok.

marcelofabri:
Thanks @vani2 🎉

PR #2894, opened 2019-10-08
Update SourceKitten to 0.26.0

### LINKED ISSUE #2874: Doesn't compile on Linux with Swift 5.1
### New Issue Checklist

- [ x ] Updated SwiftLint to the latest version
- [ x ] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

The reason is that [the XML support has been moved to a separate package on Linux](https://github.com/apple/swift-corelibs-foundation/blob/master/Docs/ReleaseNotes_Swift5.md), which breaks something in the transitive dependency SWXMLHash, where [the issue has apparently been fixed](https://github.com/drmohundro/SWXMLHash/pull/216), but of course this package is still using an old version.

Since the package is not a direct dependency of SwiftLint, it probably needs to be fixed in some other package first, but I have no idea how to find out which dependency pulls this in, so apologies for filing this here.

marcelofabri:
Fixes #2874?

Also should fix the CI failures on the `pod lib lint`.


Fryie:
Now it fails with a different error (repeated a couple of times):
```
/SwiftLint/Source/SwiftLintFramework/Models/Command.swift:83:59: error: 'scanUpToString' is unavailable
        guard let actionString = actionAndModifierScanner.scanUpToString(":"),
                                                          ^~~~~~~~~~~~~~
Foundation.Scanner:19:17: note: 'scanUpToString' was introduced in Swift 5.0
    public func scanUpToString(_ substring: String) -> String?
```

script/oss-check
```diff
@@ -195,15 +195,15 @@ def build(branch)
     perform("git fetch && git worktree add --detach #{dir} #{target}")
   end
 
+  Dir.chdir(dir) do
```
jpsim: Why was this moved? This will slow down OSSCheck by preventing incremental builds.

marcelofabri: I was getting build failures all the time IIRC

marcelofabri: unfortunately CI has deleted the build, but you can see that OSSCheck had failed before https://github.com/realm/SwiftLint/pull/2894/commits/956ea6f949d15f0fde264a0b92bbccb9d1c6ca20:  https://github.com/realm/SwiftLint/runs/251739677

PR #2898, opened 2019-10-11
Speedup Swiftlint a little

PaulTaykalo:
## What
A little speedup of a Swiftlint by

- Using lazy collections which prevents temporary structures creation
- changing insert at 0 -> append && reverse after cycle

## Why
Hacktoberfest

image


PaulTaykalo:
The graphs showing that there's a little speedup. There's actually a slowdown when using lazy in real-world situations.

PR #2899, opened 2019-10-12
Use the best run when deciding which branch is performing better

PaulTaykalo:
If we're about to compare two solutions on which one is faster we need to take the fastest run of all
This can be due a lot of levels of caching on file level etc

i.e. this is how it's done in Python
https://docs.python.org/2/library/timeit.html
> It’s tempting to calculate mean and standard deviation from the result vector and report these. However, this is not very useful. In a typical case, the lowest value gives a lower bound for how fast your machine can run the given code snippet; higher values in the result vector are typically not caused by variability in Python’s speed, but by other processes interfering with your timing accuracy. So the min() of the result is probably the only number you should be interested in. After that, you should look at the entire vector and apply common sense rather than statistics.

### DIFF
--- a/script/oss-check
+++ b/script/oss-check
@@ -15,7 +15,7 @@ require 'erb'
 
 @options = {
   branch: 'HEAD',
-  iterations: 5,
+  iterations: 8,
   skip_clean: false,
   verbose: false
 }
--- a/script/oss-check
+++ b/script/oss-check
@@ -174,11 +174,11 @@ def generate_reports(branch)
         durations << Time.now - start
       end
       puts ''
-      average_duration = (durations.reduce(:+) / iterations).round(2)
+      best_duration = durations.min.round(2)
       if branch == 'branch'
-        repo.branch_duration = average_duration
+        repo.branch_duration = best_duration
       else
-        repo.master_duration = average_duration
+        repo.master_duration = best_duration
       end
     end
   end


jpsim:
I can appreciate the rationale here, but in practice this seems to be introducing more jitter for a no-op change (OSSCheck comment above shows this branch runs anywhere between 0% and 17% slower than `master`).

PaulTaykalo:
@jpsim True.
But since we're working with a lot of caches. I theory, on first run most of the files will be cached by the system. So whichever branch will be first (**PR branch**), it will be slower than next one (because of disk caching, mostly). And since that first run will be counted in average, it will be hard to achieve at least the same speed the **master** branch will have.

Was thinking about this. the most time fluctuations are caused by the file system in general.
What do you think if benchmark from the `swiftlint` itself will be used?.
Just simple sum of all benchmark values for each file.

This, probably will be a bit closer in terms of speed, unless there'll be heavy file system work.

At least it'll be good to check how big distribution will be if the speed will be measure by inner benchmarks only.

This one was mostly to check how big is the distribution of times will be.
In this PR there are 7 rounds of check and still, instead of 5 in master branch and still, there's a big time distribution.

Will check what can be done with switlint benchmarks instead.

jpsim:
Part of the large variation in run times comes from the CI machines themselves being a bit unpredictable. It's likely that these are VMs running on shared machines, so we really can't control their load to be uniform.

So it might be worth investing the time to integrate a dedicated mac mini just for OSSCheck. MacStadium offers a free hosted mac mini for open source projects, so we could run on bare metal: https://www.macstadium.com/opensource

I'm not totally sure how we could integrate that alongside our existing CI though.

PaulTaykalo:
@jpsim 
What do you think of performing runs in a way that will look something like:
```
for each file in swift files perform next:
  for 1..iterations do
      durations_mater << measure(master)
      durations_branch << measure(branch)
```

And then sum all durations... or do max

The idea is to perform relatively fast checks and switch every time between branches.
By this  if there'll be a fluctuation it will appear in both `branch` and `master` branches, somewhat equally distributed

jpsim:
Yes I think it might be worth experimenting with interleaving master & branch runs to help counter CPU prediction or variable machine load over time.

Also experimenting with comparing min/max/mean/sum might also lead to more accurate results.

PR #2900, opened 2019-10-12
Speed up Swift Lint by using Lazy collections and prevent recalculations

PaulTaykalo:
I would highly recommend using https://github.com/realm/SwiftLint/pull/2899 in order to measure speed of linting bit closer to real values


jpsim:
What numbers do you get when you run locally (or on CI) with #2899?

PaulTaykalo:
Sadly speedup is negligeble or missing

Issue #727, opened 2016-07-22
Expiring TODO/FIXMEs


msanders:
It would be cool if there were a way to add expiring TODOs with SwiftLint.

I.e. `// TODO [2016-06-01]`, which would only warn after the given date.

Right now there isn't a great way of making use of TODOs with SwiftLint without either turning the warning off or adding noise to the project.

yonaskolb:
This is a great idea! You could remind yourself to remove some outdated API when a new version of iOS comes out for example

jpsim:
Great idea!

jeremy-w:
Unless things have changed, Xcode only picks up `TODO:` as a group, so the proposed syntax would stop it from surfacing it in the method list popup. I recommend `[before: 2016-06-01] TODO: Something` as the format.

I use a variation of [Google's TODO comment recommendation](https://google.github.io/styleguide/cppguide.html#TODO_Comments) that works around this by prefixing the author info, like so:

```
(jeremy-w/2016-08-03)TODO: Maybe this should be FIXME?
```

(I stamp it with the current date so I get an implicit "oh that must not have been terribly urgent" read on a comment without having to hit up version control history to track when the TODO landed in the source.)

I handle this simple formatting with a custom rule:

```yaml
    todo_format:
        match_kinds: comment
        message: >
            Prefix TODO/FIXME/etc with email prefix and date,
            so if you're <name@example.com>, then you'd write:
            // (name/2016-07-28)TODO: Document custom rule
        # This isn't a perfect check, but it should guide folks the right
        # direction for common cases.
        regex: "[^)](TODO|FIXME|XXX|HACK|!{3}|[?]{3}):"
```

(That should probably be a non-capturing group, come to think.)

AliSoftware:
That's not a solution to the problem, but for TODOs, we don't use SwiftLint and instead use a separate Build Phase which `grep` the files for `(TODO|FIXME|\!\!\!|\?\?\?):` and generate warnings from that.

The main benefit in having a separate Build Phase for that is that in fact we can execute that `grep` only in Analyze builds, not in standard builds. This means that we don't see any TODO warnings when we "Build" everyday, and only see them when we do an "Analyze". This avoids warning pollution.

Here's our the build phase we use, for reference:
```
if [ "${RUN_CLANG_STATIC_ANALYZER}" == "YES" ]; then
  echo "Printing TODOs as warnings..."
[...]

PR #2911, opened 2019-10-19
Add ExpiringTodoRule

namolnad:
This PR adds a new `expiring_todo` linter rule which enables developers to add an expiry date to their TODO comments. Example: `// TODO: [10/19/2019] Fix that thing`

By default the rule emits a warning when the todo expiry is within the next 15 days and an error if the expiry is at or past the current date.

Configurable parameters:
- `approaching_expiry_severity`
- `expired_severity`
- `approaching_expiry_threshold`
- `date_format`
- `date_delimiters` (`opening` and `closing`)
- `date_separator` (used to adjust the regex of the rule)

This should solve issue #727

### DIFF
--- a/Source/SwiftLintFramework/Rules/RuleConfigurations/ExpiringTodoConfiguration.swift
+++ b/Source/SwiftLintFramework/Rules/RuleConfigurations/ExpiringTodoConfiguration.swift
@@ -15,9 +15,8 @@ public struct ExpiringTodoConfiguration: RuleConfiguration, Equatable {
 
     private(set) var expiredSeverity: SeverityConfiguration
 
-    // swiftlint:disable todo
+    // swiftlint:disable:next todo
     /// The number of days prior to expiry before the TODO emits a violation
-    // swiftlint:enable todo
     private(set) var approachingExpiryThreshold: Int
     /// The opening/closing characters used to surround the expiry-date string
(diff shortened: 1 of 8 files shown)


namolnad:
@marcelofabri thanks for the help the other day to kickstart this. I think it should be ready for review now—excited to hear what the group thinks

Source/SwiftLintFramework/Rules/RuleConfigurations/ExpiringTodoConfiguration.swift
```diff
@@ -0,0 +1,91 @@
...
+    public var consoleDescription: String {
+        return "(approaching_expiry_severity) \(approachingExpirySeverity.consoleDescription), " +
+        "(reached_or_passed_expiry_severity) \(expiredSeverity.consoleDescription)"
+    }
+
+    private(set) var approachingExpirySeverity: SeverityConfiguration
+
+    private(set) var expiredSeverity: SeverityConfiguration
+
+    // swiftlint:disable todo
```
marcelofabri: nitpick: you can use `disable:next` to avoid line 20

namolnad: thanks—didn't know about this!

PR #2913, opened 2019-10-20
Optimize `redundant_void_return`

marcelofabri:
This speed ups `redundant_void_return` from ~1.5s to 300-400ms on a mid size codebase.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Idiomatic/RedundantVoidReturnRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/RedundantVoidReturnRule.swift
@@ -40,6 +40,7 @@ public struct RedundantVoidReturnRule: ConfigurationProviderRule, SubstitutionCo
             }
             """,
             "func foo()↓ -> () {}\n",
+            "func foo()↓ -> ( ) {}",
             """
             protocol Foo {
               func foo()↓ -> ()
@@ -72,6 +73,7 @@ public struct RedundantVoidReturnRule: ConfigurationProviderRule, SubstitutionCo
     public func violationRanges(in file: File, kind: SwiftDeclarationKind,
                                 dictionary: [String: SourceKitRepresentable]) -> [NSRange] {
         guard functionKinds.contains(kind),
+            containsVoidReturnTypeBasedOnTypeName(dictionary: dictionary),
             let nameOffset = dictionary.nameOffset,
             let nameLength = dictionary.nameLength,
             let length = dictionary.length,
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Rules/Idiomatic/RedundantVoidReturnRule.swift
```diff
@@ -91,4 +92,16 @@ public struct RedundantVoidReturnRule: ConfigurationProviderRule, SubstitutionCo
     public func substitution(for violationRange: NSRange, in file: File) -> (NSRange, String) {
         return (violationRange, "")
     }
+
+    private func containsVoidReturnTypeBasedOnTypeName(dictionary: [String: SourceKitRepresentable]) -> Bool {
+        guard SwiftVersion.current >= .fourDotOne else {
+            return true
```
jpsim: This doesn't seem to match what the function name says.

marcelofabri: Are you referring to this first `guard/else/return`? This is because this information is only available in SourceKit after Swift 4.1 and I wanted to keep the rule working as-is on older Swift versions.

jpsim: We should short circuit this earlier. This doesn’t belong in this function.

marcelofabri: Sure, I can change the call site but I thought it'd add more complexity:

```swift
guard functionKinds.contains(kind),
   SwiftVersion.current < .fourDotOne || containsVoidReturnTypeBasedOnTypeName(dictionary: dictionary),
// ... else {

}
```

jpsim: Why would you run this check `functionKinds.contains(kind)` when `SwiftVersion.current < .fourDotOne` will either always fail or always pass?

marcelofabri: Because `containsVoidReturnTypeBasedOnTypeName` is really just a shortcut for catching negatives faster on Swift 4.1+. We still want to support all Swift versions in this rule.

jpsim: I see. My main concern for this change then is that we return a value that doesn't answer the question asked by the function name:

Q: contains void return type based on type name?
A: Always yes for swift < 4.1, regardless of if the structure contains void return type based on type name

Maybe rename to something that better reflects what the function does?

marcelofabri: That's a good point. I wasn't too concerned about this because it's a private function, but I've  added an auxiliar function, hopeful that will make things more explicit 🙌

jpsim:
Mind if I cut a release in the coming day or so?

marcelofabri:
>Mind if I cut a release in the coming day or so?

@jpsim go for it 🥳

PR #2229, opened 2018-06-03
Fix a false positive in `unused_closure_parameter` rule

marcelofabri:
Fixes #2062

This will probably have some performance impact, but correctness should be a priority.

PR #2916, opened 2019-10-23
Faster tokens resolving in syntaxmap

PaulTaykalo:
When the request is asked which tokens are have in an intersection, the previous solution was searching for first Index (linearly) and then filtered everything that was coming after that index using `intersect function`

The updated solution  will search for the first token index using`binary search`

# Speedup

While this is only one function was updated, the next options were considered:

- [**lin+filter**] old solution 
- [**lin+prefix**] old solution with `prefix:wihle` instead of filtering
- [**bin+filter**] binary search with filter 
- [**bin+prefix**] binary search with `prefix:wihle` instead of filtering

The speedup highly depends on the file sizes. The bigger/longer files the bigger win is

# Benchmark

## Kickstarter

|lin+filter|lin+prefix|bin+filter|bin+prefix|speedup|
|-|-|-|-|-|
|0.494|0.243|0.390|\***0.117\***|  ~4x |

## Swift

|lin+filter|lin+prefix|bin+filter|bin+prefix|speedup|
|-|-|-|-|-|
|1.739|0.740|1.273|\***0.103**\*| ~16x |

## WordPress
|lin+filter|lin+prefix|bin+filter|bin+prefix|speedup|
|-|-|-|-|-|
|1.270|0.526|0.918|0.148| ~8x |

# Testing code

This code was tested with these parts of code (in Release build)
```
fileprivate var counter = 0
fileprivate var times: [String: Double] = [:]
fileprivate let timesQueue = DispatchQueue.init(label: "benchmarks")

fileprivate func timeLog<T>(_ name: String, block: () -> T) -> T {
    let start = DispatchTime.now()
    let result = block()
    let end = DispatchTime.now()
    timesQueue.async {
        let oldValue = times[name, default:0.0]
        let diff = TimeInterval(end.uptimeNanoseconds - start.uptimeNanoseconds) / 1_000_000_000
        let newValue = oldValue + diff
        times[name] = newValue
        counter += 1
        if counter % 1000 * times.count == 0 {
            print("!!!!: \(times)")
        }
    }
    return result
}

    internal func tokens(inByteRange byteRange: NSRange) -> [SyntaxToken] {
        let new = timeLog("new") { tokensFast(inByteRange: byteRange) }
        let new2 = timeLog("old") { tokensOld(inByteRange: byteRange) }
        return arc4random() % 2 == 1 ? new : new2
    }

```

### DIFF
--- a/Source/SwiftLintFramework/Extensions/SyntaxMap+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/SyntaxMap+SwiftLint.swift
@@ -11,57 +11,21 @@ extension SyntaxMap {
                 .intersects(byteRange)
         }
 
-        guard let startIndex = firstIntersectingTokenIndex(inByteRange: byteRange) else {
-            return []
+        func intersectsOrAfter(_ token: SyntaxToken) -> Bool {
+            return token.offset + token.length > byteRange.location
         }
 
-        func isAfterByteRange(_ token: SyntaxToken) -> Bool {
-            return token.offset < byteRange.upperBound
+        guard let startIndex = tokens.firstIndexAssumingSorted(where: intersectsOrAfter) else {
+            return []
         }
 
         let tokensAfterFirstIntersection = tokens
             .lazy
             .suffix(from: startIndex)
-            .prefix(while: isAfterByteRange)
+            .prefix(while: { $0.offset < byteRange.upperBound })
             .filter(intersect)
 
         return Array(tokensAfterFirstIntersection)
-    }
-
-    // Index of first token which intersects byterange
-    // Using binary search
-    func firstIntersectingTokenIndex(inByteRange byteRange: NSRange) -> Int? {
-        let lastIndex = tokens.count
-
-        // The idx, which definitely beyound byteOffset
-        var bad = -1
-
-        // The index which is definitely after byteOffset
-        var good = lastIndex
-        var mid = (bad + good) / 2
-
-        // 0 0 0 0 0 1 1 1 1 1
-        //           ^
-        // The idea is to get _first_ token index which intesects the byteRange
-        func intersectsOrAfter(at index: Int) -> Bool {
-            let token = tokens[index]
-            return token.offset + token.length > byteRange.location
-        }
-
-        while bad + 1 < good {
-            if intersectsOrAfter(at: mid) {
-                good = mid
-            } else {
-                bad = mid
-            }
-            mid = (bad + good) / 2
-        }
-
-        // Corner case, we' re out of bound, no good items in array
-        if mid == lastIndex {
-            return nil
-        }
-        return good
     }
 
     internal func kinds(inByteRange byteRange: NSRange) -> [SyntaxKind] {
--- a/Source/SwiftLintFramework/Extensions/SyntaxMap+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/SyntaxMap+SwiftLint.swift
@@ -15,11 +15,17 @@ extension SyntaxMap {
             return []
         }
 
-        let intersectingTokens = tokens
+        func isAfterByteRange(_ token: SyntaxToken) -> Bool {
+            return token.offset < byteRange.upperBound
+        }
+
+        let tokensAfterFirstIntersection = tokens
             .lazy
             .suffix(from: startIndex)
-            .prefix(while: intersect)
-        return Array(intersectingTokens)
+            .prefix(while: isAfterByteRange)
+            .filter(intersect)
+
+        return Array(tokensAfterFirstIntersection)
     }
 
     // Index of first token which intersects byterange
--- a/Source/SwiftLintFramework/Extensions/SyntaxMap+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/SyntaxMap+SwiftLint.swift
@@ -11,11 +11,51 @@ extension SyntaxMap {
                 .intersects(byteRange)
         }
 
-        guard let startIndex = tokens.firstIndex(where: intersect) else {
+        guard let startIndex = firstIntersectingTokenIndex(inByteRange: byteRange) else {
             return []
         }
-        let tokensBeginningIntersect = tokens.lazy.suffix(from: startIndex)
-        return Array(tokensBeginningIntersect.filter(intersect))
+
+        let intersectingTokens = tokens
+            .lazy
+            .suffix(from: startIndex)
+            .prefix(while: intersect)
+        return Array(intersectingTokens)
+    }
+
+    // Index of first token which intersects byterange
+    // Using binary search
+    func firstIntersectingTokenIndex(inByteRange byteRange: NSRange) -> Int? {
+        let lastIndex = tokens.count
+
+        // The idx, which definitely beyound byteOffset
+        var bad = -1
+
+        // The index which is definitely after byteOffset
+        var good = lastIndex
+        var mid = (bad + good) / 2
+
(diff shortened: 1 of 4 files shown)


PaulTaykalo:
Wow. It seems that lazy prefix suffix combination behaves strangely in Swift 5.0/Xcode 10
Will drill it down tomorrow.

PaulTaykalo:
Funny thing...
Bin prefix without lazyness works faster that with lazy

|bin+prefix+nolazy|bin+prefix|lin+filter|bin+filter|lin+prefix|  
|-|-|-|-|-|  
|0.0688|0.147|1.382|1.039|0.543|

PaulTaykalo:
@jpsim any ideas why this could fail on Swift 5.0? :(

Source/SwiftLintFramework/Extensions/SyntaxMap+SwiftLint.swift
```diff
@@ -11,11 +11,51 @@ extension SyntaxMap {
...
+        guard let startIndex = firstIntersectingTokenIndex(inByteRange: byteRange) else {
             return []
         }
-        let tokensBeginningIntersect = tokens.lazy.suffix(from: startIndex)
-        return Array(tokensBeginningIntersect.filter(intersect))
+
+        let intersectingTokens = tokens
+            .lazy
+            .suffix(from: startIndex)
+            .prefix(while: intersect)
```
marcelofabri: one thing to keep in mind is that we changed this to fix a false positive: https://github.com/realm/SwiftLint/pull/2229

PaulTaykalo: It seems that this one changed in Xcode11 while still true fro Xcode 10.

Can we know what xcode is used? I think it's SourceKit (version) which matters.

I still don't want to filter it, but can be that adding additional N tokens should help? If this happens for string interpolation, can we assume that there are no more than N string interpolations in the String?

PaulTaykalo: Or, just check on Xcode 10 which Tokens are returned within that order (non-ascending)

marcelofabri: > It seems that this one changed in Xcode11 while still true fro Xcode 10.
> 
> Can we know what xcode is used? I think it's SourceKit (version) which matters.
> 
You can check `SwiftVersion.current`. The compiler version is what matters

> I still don't want to filter it, but can be that adding additional N tokens should help? If this happens for string interpolation, can we assume that there are no more than N string interpolations in the String?

I'd vote against it. Correctness is more important than speed in this case.

marcelofabri: And in fact, this change didn't seem to cause too severe regressions according to my past self: https://github.com/realm/SwiftLint/pull/2229#issuecomment-394137732

PaulTaykalo: It is hard to know what time spent on the shole linting process. Most of the time will be spent on File I/O operations. For this only I checked this solution by measuring function running time on OSS products.
If we look at the winning times, those are within 1-3% or less of total project liting time.
In the measurings above `map` was used instead of `parallelMap` in order to prevent race condition in compilation time. This technically means that  the total winning time on the project should be divided by the number of processes.
By that - we definitely  can say that there's a win in time, but it is not that big as we want it to be

In short. the win is not that big. There should be few more PRs like this one in order to see the results )

PaulTaykalo: Please note that the timings above are in total CPU time. So for the full project run this time will be divided between cores

PaulTaykalo:
@marcelofabri updated. Works without even checking for Swift version.

jpsim:
Thanks again for your continued investments in SwiftLint performance @PaulTaykalo.

However, I want to caution against over-optimizing parts of SwiftLint's runtime that already have a negligible impact on total lint time.

In this case we're significantly increasing the complexity of this code, but because finding token indices is such a small part of the total work spent during linting, there ends up being no measurable improvement on linting overall.

Here's what I measured before & after this change on a large project (~450k lines of Swift):

```
$ hyperfine -w 1 'swiftlint-faster-tokens-resolving lint --no-cache --quiet' 'swiftlint-master lint --no-cache --quiet'
Benchmark #1: swiftlint-faster-tokens-resolving lint --no-cache --quiet
  Time (mean ± σ):     99.046 s ±  3.707 s    [User: 361.074 s, System: 17.021 s]
  Range (min … max):   93.193 s … 103.159 s    10 runs
 
Benchmark #2: swiftlint-master lint --no-cache --quiet
  Time (mean ± σ):     93.638 s ±  3.545 s    [User: 342.293 s, System: 15.224 s]
  Range (min … max):   90.103 s … 101.004 s    10 runs
 
Summary
  'swiftlint-master lint --no-cache --quiet' ran
    1.06 ± 0.06 times faster than 'swiftlint-faster-tokens-resolving lint --no-cache --quiet'
```

Maybe if we extracted the binary search logic into an extension on `Collection` somewhere else, it could simplify the resulting code while also making it more performant? I'd be more open to accepting the change then, even if the impact on performance would be the same.

---

I don't want this to discourage your interest in making SwiftLint faster though! I'd recommend profiling complete lint runs to get a better sense of performance hotspots and optimize _those_ for the best value.

PaulTaykalo:
It's not that I'm not doing that :)
I'll check where it can be used in the project.

Technically, this part above looks the same as the previous solution, but instead `firstIndex` is using `firstIndexUsingBinarySearch`. Almost the same as `func indexAssumingSorted`.

Will it work if I put the binary search part in the extension with comments and make it more abstract?

---
Most of the time is spent on the regexes
The second-most part of the time is spent on the Dictionaries access
I'm fully aware of the bottlenecks. While this speedup is not visible on the whole linting, it's there.
The deviation of the file accesses and sourcekit requests is huge. 
P.S. I would suggest to measure this function running time as in PR comment in both versions and check if running times of both versions. (like it did :) I'm just curious what are the timings in your case.

---

There's `IdenticalOperandsRule` on the way. without too many regexes. Not sure if will faster though

jpsim:
Thanks for the reply.

> Will it work if I put the binary search part in the extension with comments and make it more abstract?

Yes, specifically I think a reusable binary search function on `Collection` would be pretty helpful.

PaulTaykalo:
@jpsim So, how about this one?

PaulTaykalo:
This is how I measured that this solution is a bit faster than previous one

---
## On Swift Repository
**master** branch  3.1% of total time
image

**this** branch 0.0%
image
(scroll scroll)
image

---
## On Kickstarter Repository

**master** branch 0.7% of total time
image

**this** branch 0.0
image
(scroll scroll)
image

---

## On Wordpress Repository

**master** branch 1.4 % of total time
image

**this branch** 0.0%
image
(scroll scroll)
image


----
In general - the bigger the files in the project - the more speedup can be achieved.
At least this solution is not slower than the previous one. 

You, probably don't see any improvements mostly because of the average sizes of files in the project. This is totally explainable. There technically no improvement in the project I'm working on as well.

jpsim:
Thanks @PaulTaykalo, I've made a few small edits and will merge when CI passes.

jpsim:
There's an issue with OSScheck it didn't run properly, I think it's fall out from running on MacStadium. I'm fixing it now,.

jpsim:
Validating OSSCheck fixes here: https://github.com/realm/SwiftLint/pull/2945

PR #2917, opened 2019-10-23
Fix isEmpty typos

simonbilskyrollins:


marcelofabri:
Thanks! Can you please run the tests locally? Doing it should update the documentation.

simonbilskyrollins:
@marcelofabri done 😄

jpsim:
We've changed how we generate our docs so the manual update steps are no longer required. Thanks for the fixes! Landed in master in 0b1140d8.

PR #2918, opened 2019-10-25
Identical operands rule using syntaxmap

PaulTaykalo:
Rewrite identical operands rule which using syntax map and tokens access instead of full Regex solution.

This solutions search for the operator first and then tries to find operands to the right and to the left of the operator. This is done by searching tokens which has dots between them.

````
|   asd    ?? |      self   .    a      ==    self   .    a
| [token]     |    [token]     [token]       [token]     [token]
| identifier  |    keyword    identifier     keyword    identifier
````

Operands are matching if the next is true:

- they have same tokens count
- they have tokens of the same type
- they have identical tokens strings
- if there's a token to the left of the left token, string between them doesn't have the nil-coalescing operator

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/IdenticalOperandsRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/IdenticalOperandsRule.swift
@@ -35,7 +35,8 @@ public struct IdenticalOperandsRule: ConfigurationProviderRule, OptInRule, Autom
             ]
         } + [
             "func evaluate(_ mode: CommandMode) -> Result<AutoCorrectOptions, CommandantError<CommandantError<()>>>",
-            "let array = Array<Array<Int>>()"
+            "let array = Array<Array<Int>>()",
+            "guard Set(identifiers).count != identifiers.count else { return }"
         ],
         triggeringExamples: operators.flatMap { operation in
             [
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by PaulTaykalo** [comment]
That's weird. Will fix in evening 😊

**Comment by PaulTaykalo** [comment]
Well, at least it's faster

**Comment by PaulTaykalo** [comment]
```
Message: Linting Aerial with this PR took 2.11s vs 2.12s on master (0% faster)
Message: Linting Alamofire with this PR took 3.76s vs 3.75s on master (0% slower)
Message: Linting Firefox with this PR took 11.88s vs 12.12s on master (1% faster)
Message: Linting Kickstarter with this PR took 27.16s vs 27.91s on master (2% faster)
Message: Linting Moya with this PR took 2.38s vs 2.3s on master (3% slower)
Message: Linting Nimble with this PR took 2.29s vs 2.41s on master (4% faster)
Message: Linting Quick with this PR took 0.95s vs 0.97s on master (2% faster)
Message: Linting Realm with this PR took 3.52s vs 3.67s on master (4% faster)
Message: Linting SourceKitten with this PR took 1.51s vs 1.62s on master (6% faster)
Message: Linting Sourcery with this PR took 4.0s vs 4.14s on master (3% faster)
Message: Linting Swift with this PR took 18.9s vs 19.78s on master (4% faster)
Message: Linting WordPress with this PR took 30.3s vs 31.38s on master (3% faster)
```

File: Source/SwiftLintFramework/Rules/Lint/IdenticalOperandsRule.swift [hunk 1]
```diff
@@ -48,57 +50,148 @@ public struct IdenticalOperandsRule: ConfigurationProviderRule, OptInRule, Autom
...
+        let operatorString = contents.substring(with: operatorRange)
+        guard let leftToken = leftOperand.tokens.last, let rightToken = rightOperand.tokens.first else {
+            return nil
+        }
+        guard contents.isRegexBetweenTokens(leftToken, operatorString, rightToken) else {
+            return nil
+        }
+
+        // Make sure both operands have same token types
+        guard zip(leftOperand.tokens, rightOperand.tokens).allSatisfy({ $0.0.type == $0.1.type }) else {
```
jpsim: Nit:

```suggestion
        guard leftOperand.tokens.map({ $0.type }) == rightOperand.tokens.map({ $0.type }) else {
```

Source/SwiftLintFramework/Rules/Lint/IdenticalOperandsRule.swift
```diff
@@ -48,57 +50,148 @@ public struct IdenticalOperandsRule: ConfigurationProviderRule, OptInRule, Autom
...
+            return nil
+        }
+
+        // Make sure that every part of the operand part is equal to previous on
```
jpsim:
```suggestion
        // Make sure that every part of the operand part is equal to previous one
```

Source/SwiftLintFramework/Rules/Lint/IdenticalOperandsRule.swift
```diff
@@ -48,57 +50,148 @@ public struct IdenticalOperandsRule: ConfigurationProviderRule, OptInRule, Autom
...
+        guard zip(leftOperand.tokens, rightOperand.tokens).allSatisfy({
```
jpsim: Why not use `==` instead of zip+allSatisfy?

```swift
guard leftOperand.tokens.map(contents.subStringWithSyntaxToken) ==
    rightOperand.tokens.map(contents.subStringWithSyntaxToken) else {
    return nil
}
```

PaulTaykalo: ¯\\_(ツ)\_/¯

PaulTaykalo:
@jpsim yeah. also, it seems that this solution also can handle bit complex examples like
```
a?.b == a?.b
```

PaulTaykalo:
Let's call it a day.

jpsim:
> Let's call it a day.

Let's fix the remaining false positives first please. They're missing from the latest OSSCheck comment, but here's one that I copied from before: https://github.com/mozilla-mobile/firefox-ios/blob/48798608a791bbaea7ba3e22d58694318a40da0b/Shared/Extensions/URLExtensions.swift#L363:108

PaulTaykalo:
> > Let's call it a day.
> 
> Let's fix the remaining false positives first please. They're missing from the latest OSSCheck comment, but here's one that I copied from before: https://github.com/mozilla-mobile/firefox-ios/blob/48798608a791bbaea7ba3e22d58694318a40da0b/Shared/Extensions/URLExtensions.swift#L363:108

it's done in the latest commits. The issue was with Optional chainings like `a?.b == b`

jpsim:
@PaulTaykalo you should now have push access, feel free to merge when you're ready.

PR #2922, opened 2019-10-26
Cache Structure Dictionaries

PaulTaykalo:
## What
- Add wrapper over `[String: SourceKitRepresentable]` -> `SourceKittenDictionary`
- Cache `File.structure.dictionary`

## Why
- This is one of the most often called and time-consuming function after the regex matching
image
- Speed up mostly
- Swiftlint using readonly dictionaries from SourceKitten this allows to pre-cache all information used in Swiftlint
- Swiftlint accesses `substructure` key too often, and since it is a computed property, too many objects will be created and removed upon accessing it 
- Having own type instead of extensions over Dictionary will also speed up indexer and compiler a bit

# Note
This is more of a suggestion. I totally understand that this is a big change. 
Some minor stylistic improvements can be done( like namings and moving extensions in and out)
i.e. this one probably should be moved to `File.structurekinds`
https://github.com/realm/SwiftLint/pull/2922/files#diff-74fa1a3ab98dac161c19949597728f07R9

This solution calculates all substructures of the original dictionary. Technically, this can lead to regression in time in some cases, if `substructure` is not used in rules. 

This is also can lead to some code duplication, in places, where it is still easier to work with `[Strings: SourekitRepresentable]`

----

BTW in order to have backward compatible API, we should probably use duplicated methods which still works with `[String: SourceKitRepresentable]` and will wrap those under the hood.

----

Please see https://github.com/realm/SwiftLint/pull/2924 and  https://github.com/realm/SwiftLint/pull/2929 for the more improvements, based on this solution

### DIFF
--- a/Source/SwiftLintFramework/Extensions/Dictionary+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/Dictionary+SwiftLint.swift
@@ -1,87 +1,91 @@
 import SourceKittenFramework
 
-extension Dictionary where Key: ExpressibleByStringLiteral {
+
+private enum LazyValue<T> {
+    case notYetComputed(() -> T)
+    case computed(T)
+}
+
+final class LazyBox<T> {
+    init(computation: @escaping () -> T) {
+        _value = .notYetComputed(computation)
+    }
+
+    private var _value: LazyValue<T>
+
+    var value: T {
+        switch self._value {
+        case .notYetComputed(let computation):
+            let result = computation()
+            self._value = .computed(result)
+            return result
+        case .computed(let result):
+            return result
+        }
+    }
+}
+
+public struct SourceKittenDictionary {
+    public let value: [String: SourceKitRepresentable]
+    init(value: [String: SourceKitRepresentable]) {
+        self.value = value
+    }
+
     /// Accessibility.
     var accessibility: String? {
-        return self["key.accessibility"] as? String
+        return value["key.accessibility"] as? String
     }
-    /// Body length.
+
+    /// Body length
     var bodyLength: Int? {
-        return (self["key.bodylength"] as? Int64).flatMap({ Int($0) })
+        return (value["key.bodylength"] as? Int64).flatMap({ Int($0) })
     }
-    /// Body offset.
-    var bodyOffset: Int? {
-        return (self["key.bodyoffset"] as? Int64).flatMap({ Int($0) })
-    }
+    
     /// Kind.
     var kind: String? {
-        return self["key.kind"] as? String
+        return value["key.kind"] as? String
     }
+    
     /// Length.
     var length: Int? {
-        return (self["key.length"] as? Int64).flatMap({ Int($0) })
+        return (value["key.length"] as? Int64).flatMap({ Int($0) })
     }
     /// Name.
     var name: String? {
-        return self["key.name"] as? String
+        return value["key.name"] as? String
     }
+
     /// Name length.
-    var nameLength: Int? {
-        return (self["key.namelength"] as? Int64).flatMap({ Int($0) })
+    var nameLength : Int?{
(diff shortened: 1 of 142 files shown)


Source/SwiftLintFramework/Extensions/Dictionary+SwiftLint.swift
```diff
@@ -1,87 +1,106 @@
...
 
-extension Dictionary where Key: ExpressibleByStringLiteral {
+public struct SourceKittenDictionary {
+    public let value: [String: SourceKitRepresentable]
+    public let substructure: [SourceKittenDictionary]
+
+    init(_ value: [String: SourceKitRepresentable]) {
+        self.value = value
+
+        let substructure = value["key.substructure"] as? [SourceKitRepresentable] ?? []
```
PaulTaykalo: This caching is mostly what this PR is about

Source/SwiftLintFramework/Extensions/File+Cache.swift
```diff
@@ -18,6 +18,11 @@ private var structureCache = Cache({ file -> Structure? in
     }
     return nil
 })
+
+private var structureDictionaryCache = Cache({ file in
+    return structureCache.get(file).map { SourceKittenDictionary($0.dictionary) }
```
jpsim: Nit:

```suggestion
    return structureCache.get(file).map(SourceKittenDictionary.init)
```

Source/SwiftLintFramework/Rules/Lint/DiscardedNotificationCenterObserverRule.swift
```diff
@@ -82,24 +82,23 @@ public struct DiscardedNotificationCenterObserverRule: ASTRule, ConfigurationPro
     }
 }
 
-private extension Structure {
-    func functions(forByteOffset byteOffset: Int) -> [[String: SourceKitRepresentable]] {
-        var results = [[String: SourceKitRepresentable]]()
+private func functions(forByteOffset byteOffset: Int, in dictionary: SourceKittenDictionary)
```
jpsim: Nit: This could be an extension on `SourceKittenDictionary`.

Source/SwiftLintFramework/Rules/Lint/UnusedDeclarationRule.swift
```diff
@@ -252,6 +252,12 @@ private extension File {
     }
 }
 
+private extension Dictionary where Value == SourceKitRepresentable, Key == String {
```
jpsim: Why can't this use SourceKittenDictionary?

Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
```diff
@@ -317,15 +317,22 @@ private extension File {
...
                 return dict.substructure.contains(where: containsAttributesRequiringFoundation)
             }
         }
 
-        return containsAttributesRequiringFoundation(dict: self.structure.dictionary)
+        return containsAttributesRequiringFoundation(dict: self.structureDictionary)
+    }
+}
+
+private extension Dictionary where Value == SourceKitRepresentable, Key == String {
```
jpsim: Same thing here. Why can't this use SourceKittenDictionary?

jpsim:
Please add an entry in the "breaking changes" section of the changelog since this breaks backwards compatiblity with the API.

Source/SwiftLintFramework/Extensions/File+Cache.swift
PaulTaykalo: Not really, since it uses `$0.dictionary`, not the `$0`

jpsim: Oops, missed that.

jpsim:
Thanks for the updates and merging! Can you delete branches after you merge PRs if you don't need the branch any more?

PR #2926, opened 2019-10-27
Optimize mark rule

PaulTaykalo:
Instead of checking for all regexes and then try to search whether they're comments, we can reuse information about comments (via Syntax.tokens) and then try to match whether they're violating rule


Source/SwiftLintFramework/Rules/Lint/MarkRule.swift
```diff
@@ -183,17 +183,25 @@ public struct MarkRule: CorrectableRule, ConfigurationProviderRule {
     }
 
     private func violationRanges(in file: File, matching pattern: String) -> [NSRange] {
+        // find all comment Kinds first and then run matching patter over each of those
```
MaxHaertwig: Just a small typo: `patter` should be `pattern`

jpsim: 👍 fixed

PR #2927, opened 2019-10-28
Fixed false positive in opening_brace rule on anonymous closure

a-25:


jpsim:
Looks like this is introducing new false positives though.

a-25:
@jpsim: Do you mean false positives like this? 
https://github.com/apple/swift/blob/fd1a2a42ae7c8be2095c13339c21d27cf94cec92/stdlib/public/Darwin/CoreGraphics/Private.swift#L139:3

a-25:
@jpsim: Hello.
I tried to lint Private.swift (from Danger message https://github.com/apple/swift/blob/fd1a2a42ae7c8be2095c13339c21d27cf94cec92/stdlib/public/Darwin/CoreGraphics/Private.swift#L139:3) and it shows many violations of opening_brace (and no false positives). So I cannot understand why do our results differ. 
I tried other files from Danger message with same result: all opening_brace violations were found.
Could you explain what is the difference between our tests? And what false positives did you mean in https://github.com/realm/SwiftLint/pull/2927#issuecomment-547059756 ?

jpsim:
Here's one where the violation should still be triggered: https://github.com/apple/swift/blob/fd1a2a42ae7c8be2095c13339c21d27cf94cec92/stdlib/public/core/ThreadLocalStorage.swift#L93:3

```
  // Get the current thread's TLS pointer. On first call for a given thread,
  // creates and initializes a new one.
  internal static func getPointer()
    -> UnsafeMutablePointer<_ThreadLocalStorage>
  { // <- here
    return _swift_stdlib_threadLocalStorageGet().assumingMemoryBound(
      to: _ThreadLocalStorage.self)
  }
```

a-25:
@jpsim: checked this violation on my version, it triggers. Even added it to triggering expressions and committed.
And now I see that all the checks surprisingly passed! But I did not change the code since the last commit: only rebased by master and added triggering example :)
Is this fix OK now?

PaulTaykalo:
@a-25 should these be marked as non-triggered?
https://github.com/apple/swift/blob/0b6ef0d4d15a2f9a75009138032db2b0fe0472be/benchmark/single-source/ExistentialPerformance.swift#L535:23
https://github.com/apple/swift/blob/0b6ef0d4d15a2f9a75009138032db2b0fe0472be/benchmark/single-source/RGBHistogram.swift#L151:1

nice PR btw!

a-25:
@PaulTaykalo: Hello, Paul, thank you.
I think, it should violate, found out that it does not trigger in the whole file (if I place the method individually, it triggers). Going to debug that.

a-25:
@PaulTaykalo: Hello. Fixed the code, so the only case https://github.com/apple/swift/blob/be1d896c9c082c3afc38a6e332027c55da50a567/stdlib/public/core/KeyPath.swift#L2815:7 remained.
I doubt if this case is a problem because it is an anonymous multiline closure. And if it was short, it should be written in one line (like `map( {$0} )`). Do you think it should be fixed or this is ok now?

a-25:
@jpsim: Hello.
I analyzed the remaining skipped case (https://github.com/apple/swift/blob/be1d896c9c082c3afc38a6e332027c55da50a567/stdlib/public/core/KeyPath.swift#L2815:7) and think that it was skipped correctly. So the PR is ready to be merged.

I think it is OK because I compared two cases: this one that was skipped and the other one, that was in the original bug. I suppose they are the same.

Skipped case.
```
_internalInvariant(
      {
        // Round the amount of data we read up to alignment.
        let popped = MemoryLayout<Int32>._roundingUpToAlignment(
           bufferSizeBefore - buffer.count)
        return expectedPop == popped
      }(),
      """
      component size consumed during pattern walk does not match \
      component size returned by patternComponentBodySize
      """)
```

Original bug case.
```
func f(rect: CGRect) {
   {
      let centre = CGPoint(x: rect.midX, y: rect.midY)
      print(centre)
   }()
}
```

Both of them have anonymous closure that should begin from the new line because it is either a function parameter or an operator.

PR #2928, opened 2019-10-29
[Speedup] Use object identifier as a cache key for files

PaulTaykalo:
## What
Change the way how files information is cached in Swiftlint

## Why
Speedup mostly

---

## Note

In general, swiftlint is using `contents` or `path` as cacheKey. Unfortunately using keys so big is slow.
Using simple, fast, unique cacheKey, allows speed up running time.

On the Image. Most of the time spent in `Dictionary.getter` is actually calling `Cache.get(file.cacheKey)`

---
 The worst-case scenario for this one is to have multiple files with identical contents. which  is not-really li

### DIFF
--- a/Source/SwiftLintFramework/Extensions/File+Cache.swift
+++ b/Source/SwiftLintFramework/Extensions/File+Cache.swift
@@ -1,6 +1,8 @@
 import Foundation
 import SourceKittenFramework
 
+private typealias CacheKey = Int
+
 private var responseCache = Cache({ file -> [String: SourceKitRepresentable]? in
     do {
         return try Request.editorOpen(file: file).sendIfNotDisabled()
@@ -24,7 +26,7 @@ private var syntaxTokensByLinesCache = Cache({ file in file.syntaxTokensByLine()
 
 internal typealias AssertHandler = () -> Void
 
-private var assertHandlers = [String: AssertHandler]()
+private var assertHandlers = [CacheKey: AssertHandler]()
 private var assertHandlerCache = Cache({ file in assertHandlers[file.cacheKey] })
 
 private struct RebuildQueue {
@@ -47,7 +49,7 @@ private struct RebuildQueue {
 private var queueForRebuild = RebuildQueue()
 
 private class Cache<T> {
-    private var values = [String: T]()
+    private var values = [CacheKey: T]()
     private let factory: (File) -> T
     private let lock = NSLock()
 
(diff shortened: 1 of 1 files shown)


Source/SwiftLintFramework/Extensions/File+Cache.swift
```diff
@@ -91,8 +93,8 @@ private class Cache<T> {
 }
 
 extension File {
-    fileprivate var cacheKey: String {
-        return path ?? contents
+    fileprivate var cacheKey: CacheKey {
+        return ObjectIdentifier(self).hashValue
```
norio-nomura: This has been changed from `ObjectIdentifier` in https://github.com/realm/SwiftLint/pull/837 because there was a problem using `ObjectIdentifier` as Key.

PaulTaykalo: Thanks.

Okay. there's one not-that small solution :) like this.

https://github.com/realm/SwiftLint/blob/speedup/test-use-cache-as-int/Source/SwiftLintFramework/Models/SwiftLintFile.swift#L10

PaulTaykalo: @norio-nomura so is it only because of tests?

PaulTaykalo: Oh they're crashing... then we'll go with the complex solution

norio-nomura: If an operation that releases a `File` occurs while the` Process` is alive, it may also occur when `swiftlint` is executed. I don't know how long the `File` instances have survived in recent SwiftLint, so I don't know if it will happen in` master` either.

PaulTaykalo: Got it. My idea is to create `SwiftLintFile` wrapper over `File` which will have unique id se up on creation. Time spent on simple `Cache.get(key: contents)` is enormous

norio-nomura: > https://github.com/realm/SwiftLint/blob/speedup/test-use-cache-as-int/Source/SwiftLintFramework/Models/SwiftLintFile.swift#L10

Since `File` instances created from the same file on the file system are required to return the same `cacheKey`, I think using `id` is not suitable for the purpose.

PaulTaykalo:
Already tried in #837

PR #2929, opened 2019-10-29
[Speedup] [Proof Of Concept] Use unique file Id for each File

PaulTaykalo:
## In short
How about having unique key per file?
This can be done in a lot of ways, one of these. is below.

## What
Change the way how files information is cached in Swiftlint

## Why
Speedup mostly

---

## Note

In general, swiftlint is using `contents` or `path` as cacheKey. Unfortunately using keys so big is slow.
Using simple, fast, unique cacheKey, allows speed up running time.

On the Image. Most of the time spent in `Dictionary.getter` is actually calling `Cache.get(file.cacheKey)`

---
 The worst-case scenario for this one is to have multiple files with identical contents. which  is not-really li

---

# Alternatives

- Having unique key per file in SourceKittent framework
- making File overrideable. (this actually can decrease speedup)


Source/SwiftLintFramework/Models/SwiftLintFile.swift
```diff
@@ -0,0 +1,56 @@
+import Foundation
```
PaulTaykalo: This one is mostly what this suggestion about

Source/SwiftLintFramework/Extensions/File+Cache.swift
```diff
@@ -1,9 +1,10 @@
 import Foundation
 import SourceKittenFramework
 
+private typealias FileCacheKey = Int
```
PaulTaykalo: And this one

PaulTaykalo:
The main question here whether it is appropriate or allowable to make changes like this one :)
If YES - in general, but some additional things need to be done - then let's discuss this. If NO - then consider using alternative as suggested.

PaulTaykalo:
This one is probably, the most we can take out in terms of speed from SwiftLint not touching regular expressions.

@jpsim @marcelofabri Please, review

PaulTaykalo:
@jpsim ping?

PR #2936, opened 2019-10-31
Add allowed_symbols to type_name example

Blackjacx:
This extends the example by `allowed_symbols` for the `type_name` rule. There was an issue which is hard to find though. This way it is more discoverable.


marcelofabri:
Thanks!

PR #2941, opened 2019-11-05
Run Danger and OSSCheck on MacStadium instead of on Azure Pipelines

jpsim:
This will allow us to run OSSCheck on a bare metal Mac, which should make the OSSCheck performance results more reliable than running on virtual machines from Azure Pipelines.


jpsim:
Promising so far! A no-op change had an average performance impact of less than 0.4% across all repos linted.

image

norio-nomura:
How can we see the results of OSSCheck on MacStadium?

jpsim:
It's a public pipeline, anyone can see it: https://buildkite.com/swiftlint/swiftlint/builds/5

norio-nomura:
I didn't know that buildkite doesn't support GitHub Checks.
It can be reached from the link in the commit status to the log.
スクリーンショット 2019-11-06 9 19 30

jpsim:
Ah yes, sorry that’s how I got to see the results.

norio-nomura:
Is it possible to run [GitHub Actions self-hosted runner](https://github.blog/2019-11-05-self-hosted-runners-for-github-actions-is-now-in-beta/) on MacStadium instead of buildkite?

jpsim:
Probably, but I saw that announcement immediately after finishing setting up buildkite 😆

PR #2943, opened 2019-11-06
Run Danger on MacStadium using GitHub Actions instead of Buildkite

jpsim:
So the output can be displayed in Checks.


jpsim:
I strongly prefer Buildkite's UI, the UI for GitHub Actions is unresponsive at times, not showing job output in realtime unless I click several times. Closing.

PR #2948, opened 2019-11-07
Change superset calculation to inverted set in IdentifierNameRule

PaulTaykalo:


PaulTaykalo:
@jpsim Doesn't seem so. Or it's hard to identify whether there's an improvement.

PR #2950, opened 2019-11-08
[OSSCheck] Only run on changed rules if only rules have changed

jpsim:
Which should hopefully give a better idea of performance impact when only modifying rules by only running those rules, instead of all rules.

### DIFF
--- a/script/oss-check
+++ b/script/oss-check
@@ -133,9 +133,17 @@ def setup_repos
     dir = "#{@working_dir}/#{repo.name}"
     puts "Cloning #{repo}"
     perform("git clone #{repo.git_url} --depth 1 #{dir} 2> /dev/null")
+    swiftlint_config = "#{dir}/.swiftlint.yml"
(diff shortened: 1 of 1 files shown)


marcelofabri:
Nice! This was in my mind for a while, but never got to implement it

PaulTaykalo:
@jpsim Just thinking that in case if only rules were changed. it is, probably better to use inner benchmark times.
Currently, most of the rules are taking relatively small time in comparison with the whole cycle of searching for files + linting.
So, I probably would suggest having both checks: the whole script + changed rules using inner benchmark times. what do you think?

jpsim:
> Just thinking that in case if only rules were changed. it is, probably better to use inner benchmark times.

Agreed, we could do that. However, I think this PR's approach is already an improvement because it measures less unrelated work. It's also going to be a bit tricky to get those internal measurements, especially in the case that more than one rule was changed. You're welcome to do that later, but I'll be merging this as-is for now.

Issue #2952, opened 2019-11-09
False positive in identical_operands rule


marcelofabri:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

False positive in identical_operands rule. This was introduced in 0.37.0.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint --quiet --path file.swift
file.swift:1:17: warning: Identical Operands Violation: Comparing two identical operands is likely a mistake. (identical_operands)
$ cat file.swift 
type(of: model).cachePrefix == cachePrefix
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.37.0
* Installation method used (Homebrew, CocoaPods, building from source, etc)? portable_zip
* Paste your configuration file:

Default configuration

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)?
  If so, paste their relative paths and respective contents. No
* Which Xcode version are you using (check `xcode-select -p`)? 11.1
* Do you have a sample that shows the issue? Run `echo "[string here]" | swiftlint lint --no-cache --use-stdin --enable-all-rules`
  to quickly test if your example is really demonstrating the issue. If your example is more
  complex, you can use `swiftlint lint --path [file here] --no-cache --enable-all-rules`.

```swift
// This triggers a violation:
type(of: model).cachePrefix == cachePrefix
```

jpsim:
Surprisingly wasn't caught by OSSCheck in #2918. Thanks for reporting!

PaulTaykalo:
Fixed in #2953

PR #2953, opened 2019-11-10
Fix for false-positive identical operands rule

PaulTaykalo:


PaulTaykalo:
Fixes #2952

jpsim:
Thanks for the fix! Could you please add a changelog entry?

PR #2955, opened 2019-11-10
Add own wrappers for SyntaxTokens and Syntax Map

PaulTaykalo:
## What
This PR adds `SwiftLintSyntaxToken` and `SwiftLintSyntaxMap` wrappers over the existing `SyntaxToken` and `SyntaxMap` structures.  

### SwiftLintSyntaxToken
- Now has precomputed `kind: SyntaxKind` property

### SwiftLintSyntaxMap
- Now have precached `tokens: [SwiftLintSyntaxToken]`

## Why
- Wrappers are used mainly for storing additional information about the structures
- Precomputed values with types allowing to write simpler code since no unneeded conversation between String and enums are needed

### DIFF
--- a/Source/SwiftLintFramework/Extensions/Dictionary+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/Dictionary+SwiftLint.swift
@@ -1,3 +1,4 @@
+import Foundation
 import SourceKittenFramework
 
 public struct SourceKittenDictionary {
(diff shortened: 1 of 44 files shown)


jpsim:
Were you hoping that this would improve performance or is this for some other reason?

PaulTaykalo:
I hoped that it would improve performance. But there's nothing apart from readability :)

PaulTaykalo:
I, personally don't like the long names `SwiftLintSyntaxToken` etc.
We could probably use SyntaxToken in the Swiftlint itself, but this can lead to confusion about which one is where.

Still, since all of them were renamed, it can be a viable solution :)

Source/SwiftLintFramework/Extensions/Dictionary+SwiftLint.swift
```diff
@@ -64,6 +65,12 @@ public struct SourceKittenDictionary {
         return (value["key.offset"] as? Int64).flatMap({ Int($0) })
     }
 
+    /// Returns byte range startif from `offset` with `length` bytes
```
jpsim:
```suggestion
    /// Returns byte range starting from `offset` with `length` bytes
```

PR #2956, opened 2019-11-12
Speedup LetVarWhitespaceRule

PaulTaykalo:


Source/SwiftLintFramework/Rules/Style/LetVarWhitespaceRule.swift
```diff
@@ -231,7 +232,9 @@ private extension SwiftDeclarationKind {
 private extension SwiftLintFile {
     // Zero based line number for specified byte offset
     func line(byteOffset: Int) -> Int {
-        guard let line = contents.bridge().lineAndCharacter(forByteOffset: byteOffset)?.line else { return -1 }
-        return line - 1
+        let lineIndex = contents.bridge().lines().firstIndexAssumingSorted { line in
```
jpsim: Could probably make this even faster by reusing `lines()` instead of computing every time.

PaulTaykalo: Yeah. just found this one.

Source/SwiftLintFramework/Rules/Style/LetVarWhitespaceRule.swift
```diff
@@ -195,10 +196,9 @@ public struct LetVarWhitespaceRule: ConfigurationProviderRule, OptInRule, Automa
             }
         }
 
-        let directives: Set = ["#if", "#elseif", "#else", "#endif", "#!", "#warning", "#error"]
         let directiveLines = file.lines.filter {
-            let trimmed = $0.content.trimmingCharacters(in: .whitespaces)
-            return directives.contains(where: trimmed.hasPrefix)
+            return regex(#"^\s*#(if|elseif|else|endif|\!|warning|error)"#)
+                .firstMatch(in: $0.content, options: [], range: $0.content.fullNSRange) != nil
```
PaulTaykalo: Funny, but here, regex is faster than trying trimming files and checking for prefix

CHANGELOG.md
```diff
@@ -10,7 +10,9 @@
 
 #### Enhancements
 
-* None.
+* Speedup LetVarWhitespaceRule.
```
jpsim: Rather than claim a speed up here, can we instead just mention that you fixed a false positive? OSSCheck doesn’t demonstrate a genuine performant improvement, which may be a factor of its overhead but still.

PR #2958, opened 2019-11-13
SubstitutionCorrectableRule return optional correction

MaxHaertwig:
This PR allows `SubstitutionCorrectableRule` to return `nil` instead of a correction to indicate that a suitable correction couldn't be found for a specific case.

This could be useful for providing corrections for certain rules that cannot offer corrections in all cases. E.g. `discouraged_direct_init` could provide corrections for the default cases (`UIDevice()` -> `UIDevice.current`), but wouldn't require users to provide corrections for their custom rule configuration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@
 
 #### Enhancements
 
-* None.
+* Allow `SubstitutionCorrectableRule` to return `nil` instead of a correction
+  to indicate that a suitable correction couldn't be found for a specific case.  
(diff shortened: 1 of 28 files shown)


marcelofabri:
@MaxHaertwig FYI we enforce using rebases instead of merge commits, that's why this PR isn't passing anymore

MaxHaertwig:
Rebase done 👍

marcelofabri:
Thanks!

PR #2962, opened 2019-11-14
Add support for compilation databases

kastiglione:
This is to add support for [compilation databases](https://clang.llvm.org/docs/JSONCompilationDatabase.html) (aka `compile_commands.json`) as an alternative to parsing compiler invocations from a build log.

A compilation database can be passed using the newly added `--compile-commands` flag. This new flag is mutually exclusive to `--compiler-log-path`.

TODO

- [ ] ~~Add tests~~

### DIFF
--- a/Source/SwiftLintFramework/Models/YamlParser.swift
+++ b/Source/SwiftLintFramework/Models/YamlParser.swift
@@ -19,6 +19,16 @@ public struct YamlParser {
             throw YamlParserError.yamlParsing("\(error)")
         }
     }
+
+    public static func parseArray(_ yaml: String,
+                             env: [String: String] = ProcessInfo.processInfo.environment) throws -> [[String: Any]] {
+        do {
+            return try Yams.load(yaml: yaml, .default,
+                                 .swiftlintContructor(env: env)) as? [[String: Any]] ?? []
+        } catch {
+            throw YamlParserError.yamlParsing("\(error)")
+        }
+    }
 }
 
 private extension Constructor {
--- a/Source/SwiftLintFramework/Models/YamlParser.swift
+++ b/Source/SwiftLintFramework/Models/YamlParser.swift
@@ -19,16 +19,6 @@ public struct YamlParser {
             throw YamlParserError.yamlParsing("\(error)")
         }
     }
-
-    public static func parseArray(_ yaml: String,
-                             env: [String: String] = ProcessInfo.processInfo.environment) throws -> [[String: Any]] {
-        do {
-            return try Yams.load(yaml: yaml, .default,
-                                 .swiftlintContructor(env: env)) as? [[String: Any]] ?? []
-        } catch {
-            throw YamlParserError.yamlParsing("\(error)")
-        }
-    }
 }
 
 private extension Constructor {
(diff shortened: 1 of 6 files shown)


kastiglione:
This is an early draft, I haven't added tests and some changes should be refactored.

Source/SwiftLintFramework/Models/YamlParser.swift
```diff
@@ -19,6 +19,16 @@ public struct YamlParser {
             throw YamlParserError.yamlParsing("\(error)")
         }
     }
+
+    public static func parseArray(_ yaml: String,
+                             env: [String: String] = ProcessInfo.processInfo.environment) throws -> [[String: Any]] {
```
kastiglione: I'm an idiot, the input is json not yaml.

kastiglione:
I don't see any tests related to `--compiler-log-path`. Is there any precedent for how to test this, or should I come up with something?

PaulTaykalo:
@kastiglione 
Well, In general, it is expected to have some fixtures with files we're about to parse in this type of functionality. Some small example will be enough. as well as the results we're about to expect after parsing

jpsim:
This looks great, Dave! I wanted to take a stab at adding a test target for the `swiftlint` CLI executable, but it seems like SwiftPM doesn't support that 😞. I filed [SR-11866](https://bugs.swift.org/browse/SR-11866) to track.

With this in mind, I'm ok merging this as-is without tests since that's (unfortunately) in line with the rest of what we have in `Source/swiftlint`.

Let me know when you're done with this and I can merge.

kastiglione:
Thanks for reviewing and checking into the tests. I'll convert this out of Draft and ping you when it's ready.

kastiglione:
Ready, thanks!

Source/swiftlint/Helpers/LintableFilesVisitor.swift
```diff
@@ -93,25 +109,69 @@ struct LintableFilesVisitor {
...
     }
+
+    private static func loadCompileCommands(_ path: String) -> [String: [String]]? {
+        guard let jsonContents = FileManager.default.contents(atPath: path),
+            let object = try? JSONSerialization.jsonObject(with: jsonContents),
+            let compileDB = object as? [[String: Any]] else {
+            return nil
+        }
+
+        // Convert the compilation database to a dictionary, with source files as keys and compiler arguments as values.
```
kastiglione: A note about this dictionary. It could use quite a lot of redundant memory for large modules. For each file in a module, there will be a single compiler invocation (listing each file in the module). For example:

```
swiftc --yada --yada File1.swift File2.swift ... FileN.swift 
```

This command line will be copied N times, one for each path in the compilation database. In theory, the same array could be shared for each file in the module. Array deduping (or string interning) would reduce the memory here.

Just wanted to make a note of this.

PaulTaykalo:
Just curious, @kastiglione How are you going to receive compilation database files?
I was wondering if there's an option to generate those for testing this functionality

kastiglione:
We have a tool to generate compilation databases for Bazel projects. I would like to open source it, but even then it's use is limited to Bazel projects.

It's doable to generate a compilation database from an xcodebuild log.

jpsim:
Talking to Dave, it occurred to us that if we standardized SourceKitten/SwiftLint to use the compilation database format as input whenever we need to deal with SourceKit compiler arguments, that could simplify and standardize working with a variety of build tools, such as Xcode, bazel, SwiftPM and CMake.

acecilia:
Thanks. I end up adapting my tool to the existing format

PR #2964, opened 2019-11-14
[Proof of Concept] Validation with the new SourceKitten Structures

PaulTaykalo:


PaulTaykalo:
@jpsim Draft is done.
Namings are questionable, but in general, the idea is to have some structure that allows performing fast byte<->range conversion. + line caching.

jpsim:
I think the names you picked are great. I'm continuing this in https://github.com/realm/SwiftLint/pull/3011

jpsim:
Merged https://github.com/realm/SwiftLint/pull/3011

PR #2976, opened 2019-11-25
Include functions and getters in rule `implicit_return` and make it configurable

### LINKED ISSUE #2870: Rule Request: [implicit-return] Support SE-0255: implicit returns in single-expression functions
### New Issue Checklist

### New rule request

I'm not sure if this should be an extension (and thus be a breaking change) to the currently existing `implicit-return`; [that rule](https://github.com/realm/SwiftLint/blob/406a8f20f8c55d269d52844c41382bccda868aac/Rules.md#L10067-L10073) is currently only referring to closures. This could be a separate rule (`implicit-return-single-expression-functions`?) but that could still be confusable with `implicit-return` (renaming that to `implicit-return-closures` also feels like a tough sell.

Swift introduced [implicit returns in single-expression functions](https://github.com/apple/swift-evolution/blob/master/proposals/0255-omit-return.md) in SE-0255. From there:

> Swift provides a pleasant shorthand for short closures: if a closure contains just a single expression, that expression is implicitly returned--the `return` keyword can be omitted.  We should provide this shorthand for functions as well.

Triggers:

```swift
var x: String { return "" }
func y() -> String { return "" }
func z() -> ((String) -> Void) { return { _ in }}
```

No triggers:

```swift
var x: String { print("hi!"); return "" }
func y() -> String {
  print("hi!")
  return "bye"
}
```

This should be considered as an alternative to what's mentioned above (above point 1.).

Yes, it could be, assuming we can add this to the existing `implicit-return` rule:

- `single_expression_functions` (set to `false` by default)
- `closures` (set to `true` by default)

No; just like `implicit-return` is already disabled by default.

svenmuennich:
Currently the rule `implicit_return` only lints closures, which was the only context in which implicit returns could occur up until Swift 5.0. However, since Swift 5.1 it is possible to drop the `return` keyword from all closures, functions and getters if they contain only a single expression ([SE-0255](https://github.com/apple/swift-evolution/blob/cbb73367484a24a15e59f1d96885f637f4207b5d/proposals/0255-omit-return.md)).

This PR updates the rule to lint all `return`s that can be omitted in closures, functions and getters. It also adds a custom configuration that allows to finely tune which of these context should use implicit `return`s:
```yaml
implicit_return:
  included:
    - closure
    - function
    - getter
```

By default all kinds are linted, which is a **breaking change for existing configurations**.

---
Fixes #2870

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,12 @@
   [Dan Loman](https://github.com/namolnad)
   [#727](https://github.com/realm/SwiftLint/issues/727)
 
+* Add option to configure which kinds of `return` statements should omit their
+  `return` keyword by introducting an `included` configuration for the
+  `implicit_return` rule.  
+  [Sven Münnich](https://github.com/svenmuennich)
+  [#2870](https://github.com/realm/SwiftLint/issues/2870)
+
 #### Bug Fixes
 
 * Fix false positive for LetVarWhitespaceRule.  
--- a/Rules.md
+++ b/Rules.md
@@ -10358,12 +10358,18 @@ Identifier | Enabled by default | Supports autocorrection | Kind | Analyzer | Mi
 --- | --- | --- | --- | --- | ---
 `implicit_return` | Disabled | Yes | style | No | 3.0.0 
 
-Prefer implicit returns in closures.
+Prefer implicit returns in closures, functions and getters.
 
 ### Examples
 
 <details>
 <summary>Non Triggering Examples</summary>
+
+```swift
+if foo {
+  return 0
+}
+```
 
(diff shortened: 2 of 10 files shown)


jpsim:
Awesome rule @svenmuennich! Thank you so much for working on this and apologies for the delay in reviewing, I don't have much time for OSS work lately 😞.

I'll be merging this in a few minutes, applying minor tweaks.

I've also granted you commit access to SwiftLint. You and other contributors should feel free to merge PRs if other maintainers aren't being responsive.

jpsim:
Pushed to master in 72e20635. Thank you so much again for your contribution!

PR #2985, opened 2019-12-05
Fix parsing of rule disable command comments containing a URL

svenmuennich:
PR #2721 introduced support for trailing comments in `swiftlint:disable` commands. However, if such a trailing comment contains a URL, the command is currently not parsed correctly and as a result is not recognized at all.

This PR fixes parsing `swiftlint:disable` commands having a URL in their trailing comment, e.g.:
```swift
// swiftlint:disable rule_id - See https://github.com/realm/SwiftLint/pull/2985
```


Source/SwiftLintFramework/Extensions/SwiftLintFile+Regex.swift
```diff
@@ -61,8 +61,10 @@ extension SwiftLintFile {
         let contents = self.contents.bridge()
         let range = range ?? NSRange(location: 0, length: contents.length)
         let pattern = "swiftlint:(enable|disable)(:previous|:this|:next)?\\ [^\\n]+"
-        return match(pattern: pattern, with: [.comment], range: range).compactMap { range in
-            return Command(string: contents, range: range)
+        return match(pattern: pattern, range: range).filter { match in
+            return Set(match.1).isSubset(of: [.comment, .commentURL])
```
marcelofabri: maybe we should use `SyntaxKind.commentKinds` just to be safe?

CHANGELOG.md
```diff
@@ -14,7 +14,8 @@
 
 #### Bug Fixes
 
-* None.
+* Fix parsing of SwiftLint commands containing a URL in their trailing comment.
```
marcelofabri: can you please rebase and fix this conflict? Thanks!

svenmuennich: I merged `master` and resolved this conflict.

Source/SwiftLintFramework/Extensions/SwiftLintFile+Regex.swift
svenmuennich: `SyntaxKind.commentKinds` contains
* `.comment`,
* `.commentUrl`,
* `.docComment`,
* `.docCommentField`, and
*  `.commentMark`.

Changing this as proposed would allow adding SwiftLint commands to doc comments as well as `MARK:` comments, which is currently not supported. I'm not sure whether such a change should be part of this PR.

jpsim:
Lovely, thank you for the fix!

PR #2987, opened 2019-12-10
Return early if all imports are used.

niilohlin:
This fix speeds up the `unused_import` rule. However it assumes that all imports are clustered and will probably cause false negatives if modules are imported later in the file. However this is a very rare case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@
 
 #### Enhancements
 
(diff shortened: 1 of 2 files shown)


marcelofabri:
Thanks for the PR! However, I think that correctness is more important than speed, especially because analyzer rules are already much slower than tradicional rules (and do require a clean build).

Issue #2877, opened 2019-09-24
`unused_import` not taking into account `@_exported` declarations.


rz-robsn:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

Running the analyzer with the `unused_import` rule enabled sometimes wrongfully deletes an import if the imported Module contains a `@_exported` declaration.

Consider the following example :

```
/// Module A
@_exported import B

/// Module B
public class MyBClass {}

/// Module C
import A

let bClass = MyBClass()
```

Running the analyzer would currently remove the `import A` declaration from Module C, leading to a compile error.

### Environment

* SwiftLint version (run `swiftlint version` to be sure)?
0.35.0

* Installation method used (Homebrew, CocoaPods, building from source, etc)?
Downloaded binary from the github release page

* Paste your configuration file:

```
whitelist_rules:
  - closure_spacing
  - colon
  - empty_enum_arguments
  - fatal_error_message
  - force_cast
  - force_try
  - force_unwrapping
  - implicitly_unwrapped_optional
  - legacy_cggeometry_functions
  - legacy_constant
  - legacy_constructor
  - legacy_nsgeometry_functions
  - operator_usage_whitespace
  - redundant_string_enum_value
  - redundant_void_return
  - return_arrow_whitespace
  - trailing_newline
  - type_name
  - unused_closure_parameter
  - unused_optional_binding
  - vertical_whitespace
  - void_return
  - custom_rules

analyzer_rules:
  - unused_import

excluded:
  - Carthage
  - Pods

colon:
  apply_to_dictionaries: false

indentation: 2

custom_rules:
  no_objcMembers:
    name: "@objcMembers"
    regex: "@objcMembers"
    message: "Explicitly use @objc on each member you want to expose to Objective-C"
    severity: error
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)?
No

* Which Xcode version are you using (check `xcode-select -p`)?
10.3

jpsim:
Fixed in #2993.

PR #2993, opened 2019-12-17
[UnusedImportRule] Handle comments after the import statement

### LINKED ISSUE #2990: Unused imports crashes on import statements with trailing comment
### New Issue Checklist

### Describe the bug

The `unused_import` rule does not support import statements with a trailing comment.

https://github.com/realm/SwiftLint/blob/5bcbe36f4b13e946d1086c926c3fb1d43e679094/Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift#L234

Currently, this causes a crash.

```
2019-12-16 18:08:48.968 swiftlint[61765:2014180] *** Terminating app due to uncaught exception 'NSRangeException', reason: '-[__NSCFString replaceCharactersInRange:withString:]: Range or index out of bounds'
  | *** First throw call stack:
  | (
  | 0   CoreFoundation                      0x00007fff52efccf9 __exceptionPreprocess + 256
  | 1   libobjc.A.dylib                     0x00007fff7daafa17 objc_exception_throw + 48
  | 2   CoreFoundation                      0x00007fff52efcb2b +[NSException raise:format:] + 201
  | 3   CoreFoundation                      0x00007fff52f3ab2a mutateError + 261
  | 4   Foundation                          0x00007fff550bdeba -[NSString stringByReplacingCharactersInRange:withString:] + 142
  | 5   swiftlint                           0x000000010ee9ed8c $s18SwiftLintFramework16UnusedImportRuleV7correct4file17compilerArgumentsSayAA10CorrectionVGAA0aB4FileC_SaySSGtF + 684
  | 6   swiftlint                           0x000000010ed45f03 $s18SwiftLintFramework15CollectedLinterV7correct5usingSayAA10CorrectionVGAA11RuleStorageC_tF + 1219
  | 7   swiftlint                           0x000000010f02f692 $s9swiftlint14AnalyzeCommandV11autocorrect33_D106717C80DA5908B379D5A43478964DLLys6ResultOyyt10Commandant0L5ErrorOyytGGAA06LintOrB7OptionsVFy05SwiftN9Framework15CollectedLinterVcfU_ + 34
  | 8   swiftlint                           0x000000010f0472a8 $s18SwiftLintFramework13ConfigurationV9swiftlintE5visit33_30E847BB8C9F32DCEAC767137A79F9B7LL7linters7visitor7storage18duplicateFileNamesSayAA0abS0CGSayAA15CollectedLinterVG_AD20LintableFilesVisitorVAA11RuleStorageCShySSGtFAlOcfU_yyXEfU0_TA + 40
  | 9   swiftlint                           0x000000010f0472d1 $ss5Error_pIgzo_ytsAA_pIegrzo_TRTA + 17
  | 10  libswiftObjectiveC.dylib            0x00007fff7eff3f4f $s10ObjectiveC15autoreleasepool8invokingxxyKXE_tKlF + 47
  | 11  swiftlint                           0x000000010f0462a1 $s18SwiftLintFramework13ConfigurationV9swiftlintE5visit33_30E847BB8C9F32DCEAC767137A79F9B7LL7linters7visitor7storage18duplicateFileNamesSayAA0abS0CGSayAA15CollectedLinterVG_AD20LintableFilesVisitorVAA11RuleStorageCShySSGtFAlOcfU_ + 721
  | 12  swiftlint                           0x000000010f047269 $s18SwiftLintFramework15CollectedLinterVAA0aB4FileCIggo_AcEIegnr_TRTA + 57
  | 13  swiftlint                           0x000000010f0409e4 $sSa9swiftlintE11parallelMap9transformSayqd__Gqd__xXE_tlFADSryqd__SgGzXEfU_ySiXEfU_18SwiftLintFramework15CollectedLinterV_AG0eF4FileCTg5 + 740
  | 14  libswiftDispatch.dylib              0x00007fff7edefaf1 $sSiIgy_SiIegy_TRTA + 17
  | 15  libswiftDispatch.dylib              0x00007fff7edeab34 $sSiIegy_SiIyBy_TR + 20
  | 16  libdispatch.dylib                   0x00007fff7f290672 _dispatch_client_callout2 + 8
  | 17  libdispatch.dylib                   0x00007fff7f29ff95 _dispatch_apply_invoke + 157
  | 18  libdispatch.dylib                   0x00007fff7f29063d _dispatch_client_callout + 8
  | 19  libdispatch.dylib                   0x00007fff7f29e509 _dispatch_root_queue_drain + 657
  | 20  libdispatch.dylib                   0x00007fff7f29eb46 _dispatch_worker_thread2 + 90
  | 21  libsystem_pthread.dylib             0x00007fff7f4ca6b3 _pthread_wqthread + 583
  | 22  libsystem_pthread.dylib             0x00007fff7f4ca3fd start_wqthread + 13
  | )
  | libc++abi.dylib: terminating with uncaught exception of type NSException
```

### Example

```swift
// This triggers a violation:
import Foo // Some comment about this
```

jpsim:
Fixes #2990

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
@@ -13,7 +13,7 @@ public struct UnusedImportRule: CorrectableRule, ConfigurationProviderRule, Anal
         kind: .lint,
         nonTriggeringExamples: [
             """
-            import Dispatch
+            import Dispatch // This is used
             dispatchMain()
             """,
             """
@@ -45,7 +45,7 @@ public struct UnusedImportRule: CorrectableRule, ConfigurationProviderRule, Anal
             A.dispatchMain()
             """,
             """
-            ↓import Foundation
+            ↓import Foundation // This is unused
             struct A {
               static func dispatchMain() {}
             }
@@ -83,7 +83,7 @@ public struct UnusedImportRule: CorrectableRule, ConfigurationProviderRule, Anal
             A.dispatchMain()
             """,
             """
-            ↓import Foundation
+            ↓import Foundation // This is unused
             struct A {
               static func dispatchMain() {}
             }
@@ -225,13 +225,10 @@ private extension SwiftLintFile {
 
     func rangedAndSortedUnusedImports(of unusedImports: [String], contents: NSString) -> [(String, NSRange)] {
         return unusedImports
-            .map { module in
-                let testableImportRange = contents.range(of: "@testable import \(module)\n")
-                if testableImportRange.location != NSNotFound {
-                    return (module, testableImportRange)
+            .compactMap { module in
+                return self.match(pattern: #"^(@testable )?import \#(module)\b.*?\n"#).first.map { match in
+                    return (module, match.0)
                 }
-
-                return (module, contents.range(of: "import \(module)\n"))
             }
             .sorted(by: { $0.1.location < $1.1.location })
     }
--- a/Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
@@ -235,7 +235,12 @@ private extension SwiftLintFile {
     func rangedAndSortedUnusedImports(of unusedImports: [String], contents: NSString) -> [(String, NSRange)] {
         return unusedImports
             .compactMap { module in
-                return self.match(pattern: #"^(@.+\s+)?import\s+\#(module)\b.*?\n"#).first.map { match in
+                // We can't use raw string literals because it breaks SourceKit
+                // https://bugs.swift.org/browse/SR-11099
+                // When we require Swift 5.2 or later to build SwiftLint we can switch this back.
+                // let pattern = #"^(@.+\s+)?import\s+\#(module)\b.*?\n"#
+                let pattern = "^(@.+\\s+)?import\\s+\(module)\\b.*?\\n"
+                return self.match(pattern: pattern).first.map { match in
                     return (module, match.0)
                 }
             }
--- a/Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
@@ -226,7 +226,7 @@ private extension SwiftLintFile {
     func rangedAndSortedUnusedImports(of unusedImports: [String], contents: NSString) -> [(String, NSRange)] {
         return unusedImports
             .compactMap { module in
-                return self.match(pattern: #"^(@testable )?import \#(module)\b.*?\n"#).first.map { match in
+                return self.match(pattern: #"^(@testable\s+)?import\s+\#(module)\b.*?\n"#).first.map { match in
                     return (module, match.0)
                 }
             }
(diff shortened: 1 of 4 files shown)


Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
```diff
@@ -225,13 +225,10 @@ private extension SwiftLintFile {
 
     func rangedAndSortedUnusedImports(of unusedImports: [String], contents: NSString) -> [(String, NSRange)] {
         return unusedImports
-            .map { module in
-                let testableImportRange = contents.range(of: "@testable import \(module)\n")
-                if testableImportRange.location != NSNotFound {
-                    return (module, testableImportRange)
+            .compactMap { module in
+                return self.match(pattern: #"^(@testable )?import \#(module)\b.*?\n"#).first.map { match in
```
kastiglione: This will silently swallow modules in `unusedImports` that fail to match the regex. Should this be a `guard` with a warning in the unmatched case?

kastiglione: There's an unlikely case that could cause problems:

```swift
import Foo /* I hope no-one does this
because it would be bad */
```

I wonder if SwiftLint should be conservative and remove just the import, and leave anything that follows in place.

kastiglione:
```suggestion
                return self.match(pattern: #"^(@testable\s+)?import\s+\#(module)\b.*?\n"#).first.map { match in
```

May want to change the literal spaces to `\s+` to ensure that multiple spaces and tabs are matched too.

jpsim: > This will silently swallow modules in `unusedImports` that fail to match the regex. Should this be a `guard` with a warning in the unmatched case?

There shouldn't be any inputs here that fail the regex. If there are it's a bug in SwiftLint, not user error, so I'd rather fail loudly.

---

The multiline comment case isn't something I care to handle.

---

I'll apply the `\s+` changes, good idea!

kastiglione: > There shouldn't be any inputs here that fail the regex. If there are it's a bug in SwiftLint, not user error, so I'd rather fail loudly.

I think we're agreeing. But right now, this code will be silent in the case the pattern doesn't match. For example `@_exported import Foo` will silently be dropped https://github.com/realm/SwiftLint/issues/2877.

kastiglione:
I replied here https://github.com/realm/SwiftLint/pull/2993#discussion_r359030662 but that thread is marked resolved so may not be visible @jpsim.

Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
jpsim: Ah thanks, I didn't know about that issue. I've added support for that now in 45fa67c.

Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
```diff
@@ -225,13 +234,15 @@ private extension SwiftLintFile {
...
-            .map { module in
-                let testableImportRange = contents.range(of: "@testable import \(module)\n")
-                if testableImportRange.location != NSNotFound {
-                    return (module, testableImportRange)
+            .compactMap { module in
+                // We can't use raw string literals because it breaks SourceKit
+                // https://bugs.swift.org/browse/SR-11099
+                // When we require Swift 5.2 or later to build SwiftLint we can switch this back.
+                // let pattern = #"^(@.+\s+)?import\s+\#(module)\b.*?\n"#
+                let pattern = "^(@.+\\s+)?import\\s+\(module)\\b.*?\\n"
```
kastiglione: In surprised that `\n` needs to escaping

jpsim: You're right. Unescaped in efc75042.

PR #2995, opened 2019-12-20
Rules command options for correctable rules only and ignore terminal width

optionalendeavors:
Add support to the `rules` command for only showing correctable rules, and a switch for ignoring the width of the terminal so full configuration detail is always shown in the table.


jpsim:
Great PR, thanks! Merged to master in 8fb90065.

PR #3002, opened 2020-01-02
Add optional_enum_case_matching rule

marcelofabri:
Made it opt-in because this introduces breaking changes (if violations are fixed) for projects that support compiling with earlier Swift versions - e.g. SwiftLint itself.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Style/OptionalEnumCaseMatchingRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/OptionalEnumCaseMatchingRule.swift
@@ -1,0 +1,158 @@
+import Foundation
+import SourceKittenFramework
+
+public struct OptionalEnumCaseMatchingRule: SubstitutionCorrectableASTRule, ConfigurationProviderRule,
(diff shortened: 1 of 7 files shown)


Source/SwiftLintFramework/Rules/Style/OptionalEnumCaseMatchingRule.swift
```diff
@@ -0,0 +1,158 @@
...
+    // MARK: - SubstitutionCorrectableASTRule
+
+    public func substitution(for violationRange: NSRange, in file: SwiftLintFile) -> (NSRange, String)? {
+        return (violationRange, "")
+    }
+
+    public func violationRanges(in file: SwiftLintFile,
+                                kind: StatementKind,
+                                dictionary: SourceKittenDictionary) -> [NSRange] {
+        guard SwiftVersion.current >= type(of: self).description.minSwiftVersion, kind == .case else {
```
jpsim: Nit: could've put the cheaper check before the more expensive check

jpsim:
💯

PR #3006, opened 2020-01-03
Add opt-in `prefer_self_type_over_type_of_self` rule

### LINKED ISSUE #3003: Rule Request: Prefer `Self` over `type(of: self)`
### New Issue Checklist

https://github.com/realm/SwiftLint/issues/2944 and https://github.com/realm/SwiftLint/issues/2921 are related, but I think this should be a separate rule.

### New rule request

[Swift 5.1](https://mjtsai.com/blog/2019/09/23/swift-5-1-released/#comment-3071047) introduces the [`Self` type](https://docs.swift.org/swift-book/ReferenceManual/Types.html#grammar_self-type), which can be used in place of `type(of: self)`.

```swift
// ok
class A {
  func foo() {
    Self.bar() 
   }
}

class A {
  func foo(param: B) {
    type(of: param).bar() 
   }
}

// triggers
class A {
  func foo() {
    type(of: self).bar() 
   }
}
```

Just severity.

Opt-in.

marcelofabri:
Fixes #3003

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,11 @@
   details.  
   [Marcelo Fabri](https://github.com/marcelofabri)
 
+* Add `prefer_self_type_over_type_of_self` opt-in rule to enforce using
+  `Self` instead of `type(of: self)` when using Swift 5.1 or above.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#3003](https://github.com/realm/SwiftLint/issues/3003)
+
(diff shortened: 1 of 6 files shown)


jpsim:
It'd be interesting to support something like a `min_swift_version` key in `.swiftlint.yml` so that some rules that require 5.1+ Swift features skip themselves when `min_swift_version` is set to a lower value. That could allow us to make some opinionated rules enabled by default like this one.

Issue #3007, opened 2020-01-03
Rule Request: file_name_no_space prevent whitespace characters in file names


timcmiller:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

Prevent any whitespace characters in file names. Xcode allows you to name files with spaces, but it causes issues (or annoyances) when trying to get a path to those files since you have to escape the space.

1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.

While there is a lot of disagreement on weather file names should be camelCase(Xcode default), kabab-case, or snake_case. There is a general consensus that there should not be spaces in file names. [Apple](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingBasics.html), [Google](https://developers.google.com/style/filenames), [Smithsonian](https://library.si.edu/sites/default/files/tutorial/pdf/filenamingorganizing20180227.pdf)

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

Does not trigger
```
File.swift
File+Extension.swift
NSFile.swift
File_Name.swift
File-Name.swift
```

Does trigger
```
File Name.swift
```

3. Should the rule be configurable, if so what parameters should be configurable?

You would want to be able to exclude specific files, and I could see a use exclude certain suffixes as well. For example there may be files named `File+Test Extension.swift`.

4. Should the rule be opt-in or enabled by default? Why?
   See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a rule as opt-in.

Because Apple doesn't specifically call out no spaces in file names, it should be opt-in.

PR #3008, opened 2020-01-03
Added 'file_name_no_space' rule

timcmiller:
https://github.com/realm/SwiftLint/issues/3007

### DIFF
--- a/Source/SwiftLintFramework/Rules/Idiomatic/FileNameNoSpaceRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/FileNameNoSpaceRule.swift
@@ -1,0 +1,48 @@
+import Foundation
+import SourceKittenFramework
+
+private let typeAndExtensionKinds = SwiftDeclarationKind.typeKinds + [.extension, .protocol]
+
+public struct FileNameNoSpaceRule: ConfigurationProviderRule, OptInRule {
+    public var configuration = FileNameNoSpaceConfiguration(
+        severity: .warning,
+        excluded: [],
+        suffixPattern: "\\.*"
+    )
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "file_name_no_space",
+        name: "File Name No Space",
+        description: "File name should not contain any whitespace.",
+        kind: .idiomatic
+    )
+
+    public func validate(file: SwiftLintFile) -> [StyleViolation] {
+        guard let filePath = file.path,
+            case let fileName = filePath.bridge().lastPathComponent,
+            !configuration.excluded.contains(fileName) else {
+            return []
+        }
+
+        let suffixRegex = regex("(?:\(configuration.suffixPattern))\\z")
+        let whitespaceRegex = regex("(?:[\\s])")
+
+        var typeInFileName = fileName.bridge().deletingPathExtension
+
+        // Process suffix
+        if let match = suffixRegex.firstMatch(in: typeInFileName, options: [], range: typeInFileName.fullNSRange),
+            let range = typeInFileName.nsrangeToIndexRange(match.range) {
+            typeInFileName.removeSubrange(range)
+        }
+
+        if whitespaceRegex.firstMatch(in: typeInFileName, options: [], range: typeInFileName.fullNSRange) == nil {
+            return []
+        }
+
+        return [StyleViolation(ruleDescription: type(of: self).description,
+                               severity: configuration.severity.severity,
+                               location: Location(file: filePath, line: 1))]
+    }
+}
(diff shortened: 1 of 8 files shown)


Source/SwiftLintFramework/Rules/Idiomatic/FileNameNoSpaceRule.swift
```diff
@@ -0,0 +1,48 @@
+import Foundation
+import SourceKittenFramework
+
+private let typeAndExtensionKinds = SwiftDeclarationKind.typeKinds + [.extension, .protocol]
```
jpsim: This is unused

Source/SwiftLintFramework/Rules/RuleConfigurations/FileNameNoSpaceConfiguration.swift
```diff
@@ -0,0 +1,33 @@
+public struct FileNameNoSpaceConfiguration: RuleConfiguration, Equatable {
+    public var consoleDescription: String {
+        return "(severity) \(severity.consoleDescription), " +
+            "excluded: \(excluded.sorted())"
```
jpsim: This should also include the suffix pattern.

Source/SwiftLintFramework/Rules/Idiomatic/FileNameNoSpaceRule.swift
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+    public func validate(file: SwiftLintFile) -> [StyleViolation] {
+        guard let filePath = file.path,
+            case let fileName = filePath.bridge().lastPathComponent,
+            !configuration.excluded.contains(fileName) else {
+            return []
+        }
+
+        let suffixRegex = regex("(?:\(configuration.suffixPattern))\\z")
```
jpsim: This rule is doing more than just checking for whitespace in file names. Why is that?

Source/SwiftLintFramework/Rules/RuleConfigurations/FileNameNoSpaceConfiguration.swift
timcmiller: I used `file_name` rule as an example and didn't see suffix or prefix in the [consoleDescription](https://github.com/realm/SwiftLint/blob/20ba18c820602c29fe0e3c335adcc4f78bf6ca84/Source/SwiftLintFramework/Rules/RuleConfigurations/FileNameConfiguration.swift#L4). Is that a mistake?

Source/SwiftLintFramework/Rules/Idiomatic/FileNameNoSpaceRule.swift
timcmiller: This is intended to just remove the suffix of the filename. By default it would be the extension. But it's configureable so if a user wanted to all a suffix in file names to include whitespace characters, they could. See `testCustomSuffixPattern` test in the PR.

Source/SwiftLintFramework/Rules/RuleConfigurations/FileNameNoSpaceConfiguration.swift
jpsim: Yes that's a mistake. Would you mind adding that in a separate PR please?

Source/SwiftLintFramework/Rules/Idiomatic/FileNameNoSpaceRule.swift
jpsim: Is that something worth supporting? It seems to me like either you allow whitespace in file names or not at all. Seems odd to allow it in some parts of the name but not others, no?

timcmiller: Seems fair, I went back and forth on included it or not. I'll remove it.

Source/SwiftLintFramework/Rules/Idiomatic/FileNameNoSpaceRule.swift
```diff
@@ -0,0 +1,37 @@
...
+    )
+
+    public func validate(file: SwiftLintFile) -> [StyleViolation] {
+        guard let filePath = file.path,
+            case let fileName = filePath.bridge().lastPathComponent,
+            !configuration.excluded.contains(fileName) else {
+            return []
+        }
+
+        let whitespaceRegex = regex("(?:[\\s])")
```
jpsim: Rather than use a regex, why not use `CharacterSet`? Should be more performant and simpler.

timcmiller: Sounds good. 👍. Have never used a CharacterSet, so just defaulted to what I know, regex. Looks like it already has a [whitespace](https://developer.apple.com/documentation/foundation/characterset/1779930-whitespaces) static property so it makes it easy.

PR #3014, opened 2020-01-05
Add orphaned_doc_comment rule

### LINKED ISSUE #2989: Rule Request: Orphaned docstrings
At Lyft, we've seen several places where someone adds a disable between a docstring and whatever it's describing like this:

```swift
/// My great property
// swiftlint:disable:next force_unwrap
var myGreatProperty: String!
```

This prevents Xcode from properly rendering the docstring.

I created a very simple rule for this that doesn't catch several things:

```yaml
orphan_docstring:
  name: For a docstring to work correctly, it must be next to what it is describing.
  regex: '///\w*\n\s*//(?!/)'
```

Ideally, it would ensure the following line is some value property, type, function, etc. Doing this would also catch `///` instead of `//` for regular comments inside a method body or something.

It would be great if there was a more sophisticated official rule to properly figure out when a docstring is orphaned and warn the user.

marcelofabri:
Fixes #2989

Making this enabled by default for now, I'll re-evaluate after OSSCheck is run.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/OrphanedDocCommentRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/OrphanedDocCommentRule.swift
@@ -1,0 +1,56 @@
+import Foundation
+import SourceKittenFramework
+
+public struct OrphanedDocCommentRule: ConfigurationProviderRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "orphaned_doc_comment",
+        name: "Orphaned Doc Comment",
+        description: "A doc comment should be attached to a declaration.",
+        kind: .lint,
+        minSwiftVersion: .fourDotOne,
+        nonTriggeringExamples: [
+            """
+            /// My great property
+            var myGreatProperty: String!
+            """
+        ],
+        triggeringExamples: [
+            """
+            ↓/// My great property
+            // Not a doc string
+            var myGreatProperty: String!
+            """
+        ]
+    )
+
+    public func validate(file: SwiftLintFile) -> [StyleViolation] {
+        let docStringsTokens = file.syntaxMap.tokens.compactMap { token -> NSRange? in
+            guard token.kind == .docComment || token.kind == .docCommentField else {
+                return nil
+            }
+
+            return token.range
+        }
+
+        let docummentedDeclsRanges = file.structureDictionary.traverseDepthFirst { dictionary -> [NSRange]? in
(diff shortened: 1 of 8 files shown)


Source/SwiftLintFramework/Rules/Lint/OrphanedDocCommentRule.swift
```diff
@@ -0,0 +1,66 @@
...
+            var myGreatProperty: String!
+            """
+        ]
+    )
+
+    private static let characterSet = CharacterSet.whitespacesAndNewlines.union(CharacterSet(charactersIn: "/"))
+
+    public func validate(file: SwiftLintFile) -> [StyleViolation] {
+        let docStringsTokens = file.syntaxMap.tokens.filter { token in
+            return token.kind == .docComment || token.kind == .docCommentField
```
jpsim: What kind is a URL in a doc comment?

marcelofabri: It's `source.lang.swift.syntaxtype.comment.url`:

```swift
/// See
/// https://apple.com
func foo() {}
```

```js
[
  {
    "length" : 8,
    "offset" : 0,
    "type" : "source.lang.swift.syntaxtype.doccomment"
  },
  {
    "length" : 4,
    "offset" : 8,
    "type" : "source.lang.swift.syntaxtype.doccomment"
  },
  {
    "length" : 17,
    "offset" : 12,
    "type" : "source.lang.swift.syntaxtype.comment.url"
  },
  {
    "length" : 1,
    "offset" : 29,
    "type" : "source.lang.swift.syntaxtype.doccomment"
  },
  // ...
]
```

jpsim: Right so shouldn't this check include comment URLs too then?

marcelofabri: Not really because it's the same syntax kind of a regular comment URL. Since there should be a `source.lang.swift.syntaxtype.doccomment` together, I think it's fine. 

The case where this could lead to false negatives is `/// URL` as the `doccomment` piece would just be `/// ` and then the exception for `/` only lines would avoid the violation to be triggered.

jpsim: Sounds good. Probably a good idea to add a test case to be sure though.

marcelofabri: Done: https://github.com/realm/SwiftLint/pull/3017 ✅

soffes:
Thanks for doing this!

PR #3015, opened 2020-01-05
[DO NOT MERGE] Add return_value_from_void_function using SwiftSyntax

marcelofabri:
Used https://github.com/realm/SwiftLint/pull/2480 as a starting point

This is an example of a rule that SwiftSyntax enables us to write with not too much code and in a reliable way.

Added "DO NOT MERGE" because there are a few drawbacks of adding SwiftSyntax and we should address those separately first:

- We can't import `SwiftSyntax` via CocoaPods, so we'd have to deprecate `SwiftLintFramework.podspec`
- Apparently [there's an issue](https://forums.swift.org/t/swiftsyntax-with-swift-5-1/29051/15) with SPM + Xcode, so we can't use the regular Xcode 11 SPM support
- We'd need to move away from Carthage in favor of building SwiftLint in Xcode with SPM 
- Running `swift package generate-xcodeproj` might have changed important settings (e.g. deployment target)

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -56,6 +56,15 @@
         }
       },
       {
+        "package": "SwiftSyntax",
+        "repositoryURL": "https://github.com/apple/swift-syntax.git",
+        "state": {
+          "branch": null,
+          "revision": "3e3eb191fcdbecc6031522660c4ed6ce25282c25",
+          "version": "0.50100.0"
+        }
+      },
+      {
         "package": "SwiftyTextTable",
         "repositoryURL": "https://github.com/scottrhoyt/SwiftyTextTable.git",
         "state": {
--- a/Package.swift
+++ b/Package.swift
@@ -18,6 +18,7 @@ let package = Package(
         .package(url: "https://github.com/jpsim/SourceKitten.git", from: "0.28.0"),
         .package(url: "https://github.com/jpsim/Yams.git", from: "2.0.0"),
         .package(url: "https://github.com/scottrhoyt/SwiftyTextTable.git", from: "0.9.0"),
+        .package(url: "https://github.com/apple/swift-syntax.git", .exact("0.50100.0")),
     ] + (addCryptoSwift ? [.package(url: "https://github.com/krzyzanowskim/CryptoSwift.git", .upToNextMinor(from: "1.0.0"))] : []),
     targets: [
         .target(
@@ -33,6 +34,7 @@ let package = Package(
             dependencies: [
                 "SourceKittenFramework",
                 "Yams",
+                "SwiftSyntax",
             ] + (addCryptoSwift ? ["CryptoSwift"] : [])
         ),
         .testTarget(
--- a/Rules.md
+++ b/Rules.md
@@ -143,6 +143,7 @@
 * [Required Deinit](#required-deinit)
 * [Required Enum Case](#required-enum-case)
 * [Returning Whitespace](#returning-whitespace)
+* [Return Value from Void Function](#return-value-from-void-function)
 * [Shorthand Operator](#shorthand-operator)
 * [Single Test Class](#single-test-class)
 * [Min or Max over Sorted First or Last](#min-or-max-over-sorted-first-or-last)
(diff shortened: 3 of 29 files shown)


jpsim:
> * We can't import `SwiftSyntax` via CocoaPods, so we'd have to deprecate `SwiftLintFramework.podspec`

Couldn't we instead either 1) add a podspec for SwiftSyntax or 2) conditionally use SwiftSyntax when available by using `canImport(SwiftSyntax)`?

marcelofabri:
> > * We can't import `SwiftSyntax` via CocoaPods, so we'd have to deprecate `SwiftLintFramework.podspec`
> 
> Couldn't we instead either 1) add a podspec for SwiftSyntax or 2) conditionally use SwiftSyntax when available by using `canImport(SwiftSyntax)`?

We could, but not sure if it's worth the complexity/extra work in the first option and in the long term I imagine that we'd have more and more rules using `SwiftSyntax`, so it wouldn't be ideal to use `canImport(SwiftSyntax)`

jpsim:
Neither approach seem particularly time-consuming to me.

marcelofabri:
- What about supporting old Swift compiler versions? Is there a way to change a dependency's version with SPM based on the Swift version? SwiftSyntax requires a different tag for different Swift versions. Is supporting older compiler versions worth it?

- With SPM, it looks like we'd either have to regenerate the project very often (as we can't use Xcode 11's SPM support because of a [bug](https://forums.swift.org/t/swiftsyntax-with-swift-5-1/29051/15)). Should we just not commit the project at all then?

jpsim:
> * What about supporting old Swift compiler versions? Is there a way to change a dependency's version with SPM based on the Swift version? SwiftSyntax requires a different tag for different Swift versions. Is supporting older compiler versions worth it?

Swift is pretty good at handling old syntax, so using the latest seems fine for linting all Swift versions going back to 3.0.

> * With SPM, it looks like we'd either have to regenerate the project very often (as we can't use Xcode 11's SPM support because of a [bug](https://forums.swift.org/t/swiftsyntax-with-swift-5-1/29051/15)). Should we just not commit the project at all then?

It'd be awesome to get rid of our `xcodeproj` and `xcworkspace` altogether. The tricky thing to figure out is how we can do that without dropping support for integrating SwiftLintFramework via Carthage. Maybe we could integrate something like [XcodeGen](https://github.com/yonaskolb/XcodeGen)?

marcelofabri:
> Swift is pretty good at handling old syntax, so using the latest seems fine for linting all Swift versions going back to 3.0.

The issue is not linting old Swift versions, but compiling SwiftLint itself (see CI for this PR, e.g. the "SwiftPM xcode103" job).

> Maybe we could integrate something like XcodeGen?

That's a good idea, as it (apparently) supports using dependencies from SPM!

marcelofabri:
This is my suggestion to move forward:

- Bump the required Xcode version to Xcode 11 (thus bumping the required Swift version to compile SwiftLint to 5.1). This is need to use Xcode's SPM support as I couldn't find a way to use SPM with XcodeGen that doesn't create Xcode 11's package references.
- Use XcodeGen to generate a project, importing dependencies via Xcode's SPM support. I have a WIP branch with it: https://github.com/realm/SwiftLint/tree/mf-xcodegen
- Stop using the xcconfigs that we import with Carthage. My understanding is that we actually don't use them anyway when cutting releases as we rely on `swift build`.
- Remove the project and workspace from source control, as well  as the dependencies submodules
- Add the workaround mentioned [here](https://forums.swift.org/t/swiftsyntax-with-swift-5-1/29051) as a build step so we can import SwiftSyntax when building from Xcode
- Use `canImport` to still allow using CocoaPods to import `SwiftLintFramework`. We could later create a proper podspec for SwiftSyntax.
- Generate the project when creating a release, commit it, create a tag and then removing it from source control again in order to support Carthage. I don't know if this would work with Carthage giving that the project would have SPM references. Alternatively, we could drop Carthage support as we don't mention it in the README anyway and I think it's very unlikely that someone is using it to import `SwiftLintFramework`. Worst case scenario we could work to add it back if someone is actually using it.

marcelofabri:
Closing in favor of #3054

PR #3024, opened 2020-01-07
Improve tuple handling in optional_enum_case_matching rule

jpsim:
Catch previously missed violations in the `optional_enum_case_matching` rule when case expressions involved tuples. For example:

```swift
switch foo {
  case (.bar↓?, .baz↓?): break
  case (.bar↓?, _): break
  case (_, .bar↓?): break
  default: break
}
```

### DIFF
--- a/Source/SwiftLintFramework/Rules/Style/OptionalEnumCaseMatchingRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/OptionalEnumCaseMatchingRule.swift
@@ -125,7 +125,7 @@ public struct OptionalEnumCaseMatchingRule: SubstitutionCorrectableASTRule, Conf
              case (_, .bar↓?): break
              default: break
             }
-            """:"""
+            """: """
             switch foo {
              case (.bar, .baz): break
              case (.bar, _): break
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Rules/Style/OptionalEnumCaseMatchingRule.swift
```diff
@@ -133,26 +164,61 @@ public struct OptionalEnumCaseMatchingRule: SubstitutionCorrectableASTRule, Conf
...
+            return stringView.byteOffset(fromLocation: distance(from: startIndex, to: range.lowerBound))
+        }
+    }
+}
+
+private extension SwiftLintFile {
+    func isTokenUnderscoreKeyword(_ token: SwiftLintSyntaxToken) -> Bool {
+        return token.kind == .keyword &&
+            token.length == 1 &&
+            stringView.substringWithByteRange(start: token.offset, length: token.length) == "_"
```
marcelofabri: IIRC we have an utility method somewhere that returns a string for `SwiftLintSyntaxToken`

marcelofabri:
Alternatively, we can find the `source.lang.swift.expr.tuple` entries in the structure dictionary

Source/SwiftLintFramework/Rules/Style/OptionalEnumCaseMatchingRule.swift
jpsim: Found it.

Issue #3026, opened 2020-01-08
Improve API doc coverage to 100%


jpsim:
We recently added API doc generation in #3016. We're currently at 11% according to jazzy.

There's no need to document any declarations outside SwiftLintFramework, or anything that's not public/open, or anything under `Source/SwiftLintFramework/Rules` since those aren't as much part of the framework's public API.

Other than those exceptions, there's no reason why we shouldn't strive for 100% API documentation coverage. In cases where it's not ideal to document a public/open declaration, we can exclude it from API docs by adding `:nodoc:` to the documentation comment.

PR #3027, opened 2020-01-08
Add documentation comments to all public declarations

jpsim:
Addresses https://github.com/realm/SwiftLint/issues/3026


marcelofabri:
👏

PR #3028, opened 2020-01-09
Make strong-iboutlet rule correctable

MaxHaertwig:


marcelofabri:
shouldn't the rule also change the type to be non-optional (and non-implicitly unwrapped)?

MaxHaertwig:
IBOutlets can't be non-optionals. This would require views or view controllers to set the property in their initializer, but in case of IBOutlets, they're set up in the storyboard.

Source/SwiftLintFramework/Rules/Lint/StrongIBOutletRule.swift
```diff
@@ -18,24 +20,44 @@ public struct StrongIBOutletRule: ConfigurationProviderRule, ASTRule, OptInRule,
...
+        return [range]
+    }
 
-        return [
-            StyleViolation(ruleDescription: type(of: self).description,
-                           severity: configuration.severity,
-                           location: Location(file: file, byteOffset: offset))
-        ]
+    public func substitution(for violationRange: NSRange, in file: SwiftLintFile) -> (NSRange, String)? {
+        let prefixString = file.contents.bridge().substring(to: violationRange.location)
```
marcelofabri: we should use the range of the `weak` attribute that SourceKitten provides instead:

```
{
  "key.length" : 35,
  "key.offset" : 0,
  "key.substructure" : [
    {
      "key.attributes" : [
        {
          "key.attribute" : "source.decl.attribute.weak",
          "key.length" : 4,
          "key.offset" : 10
        },
        {
          "key.attribute" : "source.decl.attribute.iboutlet",
          "key.length" : 9,
          "key.offset" : 0
        }
      ],
      "key.kind" : "source.lang.swift.decl.var.global",
      "key.length" : 19,
      "key.name" : "label",
      "key.namelength" : 5,
      "key.nameoffset" : 19,
      "key.offset" : 15,
      "key.setter_accessibility" : "source.lang.swift.accessibility.internal",
      "key.typename" : "UILabel?"
    }
  ]
}
```

MaxHaertwig: Great idea! How can I generate this dictionary for the string in the violation range? I know it's possible to get it for the entire file, but it's not very efficient to go through it and filter for the violation range.

marcelofabri: You can use `dictionary.swiftAttributes` in `violationRanges` (instead of `enclosedSwiftAttributes`)

MaxHaertwig: I was referring to `substitution(for:in:)`. The only inputs are NSRange and SwiftLintFile, but no dictionary. I can call `file.structureDictionary`, but this gives me the dictionary for the entire file and not just for the violation range. I figured there must be a better way than to traverse this dictionary to find a declaration that falls within the violation range.

marcelofabri: To make things easier, I'd change the violation location to be in the `weak`/`unowned` keyword instead, this way you can keep most of the logic in `violationRanges`. That's the main use case that `SubstitutionCorrectableASTRule` was created for.

MaxHaertwig:
@marcelofabri @jpsim I rebased my code on top of master. Can we get this merged?

jpsim:
Sorry for the super long delay, but I've revived this and will merge when CI passes: https://github.com/realm/SwiftLint/pull/3530

PR #3037, opened 2020-01-14
Migrate to use SourceKitten's new ByteCount/ByteRange types

jpsim:
New APIs were introduced in SourceKitten to allow for a more typesafe distinction between integers meaning NSString-based distances and byte-based distances.

* https://github.com/jpsim/SourceKitten/pull/639
* https://github.com/jpsim/SourceKitten/pull/642

This PR migrates SwiftLint's use of those APIs.


jpsim:
According to OSSCheck, it looks like we're now counting line length differently somehow? It doesn't look like any other rules behave differently.

PR #3042, opened 2020-01-16
Add deinitializer to type_contents_order

StevenMagdy:
Add `deinitializer` type content to `type_contents_order` rule instead of grouping it with initializers.
This allows the common use case of putting the deinitializer at the end of the class.


jpsim:
Thanks!

PR #3016, opened 2020-01-06
[Docs] Build docs using jazzy

jpsim:
* Add `.jazzy.yaml` configuration file
* Update `swiftlint generate-docs` to write docs to a directory rather than a single file
* Add jazzy to the Gemfile
* Run `bundle update`
* Add GitHub Actions workflow to run jazzy automatically and publish to GitHub Pages

swiftlint-jazzy

PR #3047, opened 2020-01-17
Add `swiftlint show-docs` command to easily open online docs.

417-72KI:
GitHub Pages is made in #3016.
It would be nice to be able to open in browser easily.

In this PR, `path/to/swiftlint show-docs` command will open https://realm.github.io/SwiftLint in default browser.

### DIFF
--- a/Source/swiftlint/Commands/ShowDocsCommand.swift
+++ b/Source/swiftlint/Commands/ShowDocsCommand.swift
@@ -6,7 +6,7 @@ struct ShowDocsCommand: CommandProtocol {
     let function = "Open `SwiftLint Framework Docs` on web browser"
 
     func run(_ options: NoOptions<CommandantError<()>>) -> Result<(), CommandantError<()>> {
-        let url = URL(string: "https://realm.github.io/SwiftLint/rule-directory.html")!
+        let url = URL(string: "https://realm.github.io/SwiftLint")!
         open(url)
         return .success(())
     }
--- a/Source/swiftlint/Commands/ShowDocsCommand.swift
+++ b/Source/swiftlint/Commands/ShowDocsCommand.swift
@@ -3,7 +3,7 @@ import Foundation
 
 struct ShowDocsCommand: CommandProtocol {
     let verb = "show-docs"
-    let function = "Open `SwiftLint Framework Docs` on web browser"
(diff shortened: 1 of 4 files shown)


Source/swiftlint/Commands/ShowDocsCommand.swift
```diff
@@ -0,0 +1,29 @@
...
+    }
+}
+
+private extension ShowDocsCommand {
+    func open(_ url: URL) {
+        let process = Process()
+        process.launchPath = "/usr/bin/env"
+        let command: String = {
+            #if os(Linux)
+            return "xdg-open"
```
marcelofabri: I don't think this works on every Linux distribution 🤔

Source/swiftlint/Commands/ShowDocsCommand.swift
```diff
@@ -0,0 +1,29 @@
+import Commandant
+import Foundation
+
+struct ShowDocsCommand: CommandProtocol {
+    let verb = "show-docs"
```
marcelofabri: maybe just `docs`?

Source/swiftlint/Commands/ShowDocsCommand.swift
```diff
@@ -0,0 +1,29 @@
...
+    let function = "Open `SwiftLint Framework Docs` on web browser"
```
marcelofabri: no need for the backquotes (and I would remove the "Framework" from the sentence)

Source/swiftlint/Commands/ShowDocsCommand.swift
417-72KI: It seems `generate-docs` command is still available, so I named it to prevent confusion.
If not needed, I will fix it.

Source/swiftlint/Commands/ShowDocsCommand.swift
417-72KI: Fixed it!

Source/swiftlint/Commands/ShowDocsCommand.swift
417-72KI: I choose the most universal command as far as I know.
I'm not expert with Linux, and don't know if there is more better solution😔

PR #3052, opened 2020-01-22
Empty count configuration

lordzsolt:
This PR provides an improvement for #827 

It adds a a configuration option, to ignore variables named `count`. This option is disabled by default.

### DIFF
--- a/Source/SwiftLintFramework/Rules/RuleConfigurations/EmptyCountConfiguration.swift
+++ b/Source/SwiftLintFramework/Rules/RuleConfigurations/EmptyCountConfiguration.swift
@@ -4,7 +4,7 @@ private enum ConfigurationKey: String {
 }
 
 public struct EmptyCountConfiguration: RuleConfiguration, Equatable {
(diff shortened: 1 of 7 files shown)


Source/SwiftLintFramework/Rules/RuleConfigurations/EmptyCountConfiguration.swift
```diff
@@ -0,0 +1,26 @@
+private enum ConfigurationKey: String {
+    case severity = "severity"
+    case onlyAfterDot = "only_after_dot"
+}
+
+public struct EmptyCountConfiguration: RuleConfiguration, Equatable {
+    private(set) var severityConfiguration = SeverityConfiguration(.warning)
```
marcelofabri: the existing default is `error`, can you change it to keep the same behavior?

lordzsolt: Updated.

Sorry, I didn't notice it was `.error before

marcelofabri:
Thanks! Great first contribution! 🎉

Issue #3060, opened 2020-01-31
False positive when line ends with carriage return ("\r\n")


john-mueller:
### New Issue Checklist

- [X] Updated SwiftLint to the latest version
- [X] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

When a line ends with a carriage return + newline ("\r\n"), false errors are reported. This can occur if the file or contained text originated from or was touched by a text editor using the Windows line ending standard.

Since Swift compiles without issue when lines are separated by "\r\n", SwiftLint should also consider it acceptable.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ echo -e "//swiftlint:disable all\nprint(123)\n" | swiftlint lint --use-stdin
Done linting! Found 0 violations, 0 serious in 1 file.
$ echo -e "//swiftlint:disable all\r\nprint(123)\r\n" | swiftlint lint --use-stdin
<nopath>:5:1: warning: Trailing Newline Violation: Files should have a single trailing newline. (trailing_newline)
<nopath>:5:1: warning: Vertical Whitespace Violation: Limit vertical whitespace to a single empty line. Currently 2. (vertical_whitespace)
' is not a valid SwiftLint rule. Please remove it from the disable command. (superfluous_disable_command)
Done linting! Found 3 violations, 0 serious in 1 file.
```

The last violation is
```
<nopath>:1:24: warning: Superfluous Disable Command Violation: 'all' is not a valid SwiftLint rule. Please remove it from the disable command.  (superfluous_disable_command)
```
 but the \r after all causes the first part of the warning to be overwritten.

### Environment

* SwiftLint version (run `swiftlint version` to be sure)?
0.38.2
* Installation method used (Homebrew, CocoaPods, building from source, etc)?
Building from source.
* Paste your configuration file:
None.
* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)?
  If so, paste their relative paths and respective contents.
No.
* Which Xcode version are you using (check `xcodebuild -version`)?
Xcode 11.3
Build version 11C29
* Do you have a sample that shows the issue? Run `echo "[string here]" | swiftlint lint --no-cache --use-stdin --enable-all-rules`
  to quickly test if your example is really demonstrating the issue. If your example is more
  complex, you can use `swiftlint lint --path [file here] --no-cache --enable-all-rules`.
See example above.

PR #3061, opened 2020-01-31
Fix false positives when line ends with carriage return + line feed

john-mueller:
Fixes: #3060

I wasn't exactly sure where the test should go, so I added it in its own file. If there's a better place for it, I'm happy to move it.


marcelofabri:
Thanks!

PR #3040, opened 2020-01-15
Inline test failure messages

ZevEisenberg:
Fixes #3035 

- [x] Make it work.
- [x] Finish adding `Example` literal init to all strings and remove `ExpressibleByStringLiteral` conformance from `Example` type.
- [x] Find out whether we can remove trailing `\n` from the many test cases that have them.
- [x] ~Remove trailing newlines from all examples where possible.~ rolled back because it was too invasive
- [x] Resolve file and line length warnings.
- [x] fix broken tests
- [x] ~merge https://github.com/ZevEisenberg/SwiftLint/pull/1~ this PR is no longer dependent on that one

@jpsim here's a start! I took a pretty lazy approach to converting files over to the new format, opting for an `ExpressibleByStringLiteral` band-aid (that [doesn't even partially work](https://bugs.swift.org/browse/SR-12034)) just so I can get it up and running. Before merging, we'll need to go through and add an explicit `Example(...)` initializer to literally every example.

I've also been pretty liberal with smashing things in order to get it all compiling, so I'm very open to ideas if you have suggestions for how to make the PR more conservative.

I have two open questions that I'd love some feedback on:

1. A ton of examples have a trailing `\n`, but not all of them do. Do you know why it's there? Can I delete those newlines? Here's an example: `"public func a() {}\n"`.
1. Some of the file and line length warnings don't seem that useful on a project like this, where file length isn't a good indicator of code complexity. it's just a bunch of "data"-like string literals, and sometimes the file gets longer just because we wrap lines to make them clearer. However, I'm not the project owner, so I'd appreciate any guidance you have on how to get around those warnings.

PR #3065, opened 2020-02-02
Restructure test helpers

ZevEisenberg:
The test helpers were getting ungainly, and I had to suppress some length warnings in #3040. This re-enables those warnings, and breaks up the test helpers into separate files.


CHANGELOG.md
```diff
@@ -24,6 +24,9 @@
   and which are not.  
   [ZevEisenberg](https://github.com/ZevEisenberg)
   [#3040](https://github.com/realm/SwiftLint/pull/3040)
+* Restructure test helpers.
```
ZevEisenberg: I wasn't sure if this belonged under Enhancements, since it's not user-facing.

Tests/SwiftLintFrameworkTests/CommandTests.swift
```diff
@@ -280,93 +285,99 @@ class CommandTests: XCTestCase {
...
         XCTAssertEqual(
-            violations(Example("print(123)\n// swiftlint:disable:previous nesting_foo\n"))[0]
+            TestHelpers.violations(Example("print(123)\n// swiftlint:disable:previous nesting_foo\n"))[0]
                 .ruleIdentifier,
             "superfluous_disable_command"
         )
 
         XCTAssertEqual(
-            violations(Example("print(123)\n// swiftlint:disable:previous nesting_foo \n")).count,
+            TestHelpers.violations(Example("print(123)\n// swiftlint:disable:previous nesting_foo \n")).count,
```
PaulTaykalo: Is there a need to put everything under a static function?
It's not that this should be included in PR

ZevEisenberg: Not needed, but I thought it made it clearer where things were coming from. But I'm happy to move it to a global, since it's still nicely scoped inside the test helpers, so we can use `SwiftLintFrameworkTests.violations` where we need to disambiguate.

Side note: I'd prefer `violations(in:...)` instead of `violations(...)`, but don't want to change too much at once.

CHANGELOG.md
jpsim: It's fine to list internal-only changes in the changelog, I just typically don't do that myself.

By the way, we typically list the _issue_ a PR is resolving in a changelog entry, not the PR number itself, but this also isn't a big deal.

ZevEisenberg: While we’re on the topic, why the two trailing spaces?

ZevEisenberg:
Rebased. Should be good to merge.

CHANGELOG.md
ZevEisenberg: LOL, I removed the changeling entry and Danger complained 😆

ZevEisenberg:
Bump - any interest in merging this? I totally get it if not, since it's more invasive than your typical pull request; I just want to know if I should keep it updated as `master` changes, and figured now, right after a release, was a good time to ask.

ZevEisenberg:
Writing is on the wall for this one ☠️

Issue #1921, opened 2017-10-23
Restrict opening_brace to single-line declarations


lilyball:
### New Issue Checklist

- [X] Updated SwiftLint to the latest version
- [X] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Rule Request

I'd like to be able to configure the opening_brace rule so it only applies to single-line declarations. If I'm splitting a method across multiple lines then it's usually a good idea to put the opening brace on a new line as well, but single-line declarations should still always have the brace on the same line.

kballard:
For example, the following looks weird:

```swift
func foo<T>(_ arg1: T)
    where T: Equatable {
    // bar
}
```

This reads much better when I can put the brace on a new line:

```swift
func foo<T>(_ arg1: T)
    where T: Equatable
{
    // bar
}
```

georgejlee:
I agree, I have some functions with many long arguments so I find it easier to read when I split the arguments across multiple lines and put the opening brace on its own line, like this:

```
    open class func request(
        _ URLRequest: URLRequestConvertible,
        success: (((NSDictionary) -> Void)?),
        failure: (failureCallback?),
        isSuccess: @escaping ((DataResponse<Any>) -> Bool) = WebCom.defaultIsSuccess)
    {
        ...
    }
```

exevil:
+1 here.

hartbit:
+1 here too

petereddy:
+1

I'm hoping to use SwiftLint to enforce just the thing kballard describes. I still want it to be possible to write:
> func something() -> String {
>     return "hi"
> }

But in the case of functions with multi-line parameters/declarations, I'd like the opening brace to be allowed on a new line for clarity.

marcelofabri:
PRs adding this as a configuration are welcome!

kubatruhlar:
This is a good idea. If we go vertically, it's usually better to put the `{` on its own line to optically separate the parameters from the body of a function.

kballard:
This is becoming more of an issue on my team. I'm not willing to re-enable the `opening_brace` rule without the ability to restrict to single-line declarations, but my coworkers are starting to grumble about PRs being submitted that incorrectly put the opening brace on a new line for single-line declarations.

[...]

PR #3068, opened 2020-02-03
Allow opening brace on newline in case of multiline function

lordzsolt:
Implements #1921 

I kept the option disabled, to keep the old behavior.

Note: Indentation is not validated, similar to the old `if`, `guard` and `while` behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,12 @@
   [Keith Smiley](https://github.com/keith)
   [#3184](https://github.com/realm/SwiftLint/issues/3184)
 
+* Add `allow_multiline_func` configuration option to `opening_brace`
+  rule, to allow placing `{` on new line in case of multiline function.
+  Defaults to `false`.  
+  [Zsolt Kovács](https://github.com/lordzsolt)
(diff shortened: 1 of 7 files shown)


julsh:
I would love to see this merged!!

weakfl:
This would be great, especially since SwiftFormat has introduced a `wrapMultilineStatementBraces` rule.

jpsim:
Thanks for this PR and for your patience, I'm slowly making my way through old PRs. I just rebased this and will be merging when CI passes.

weakfl:
Thanks for merging @jpsim. But imho the config option should be applied for `init(...)` as well. Or alternatively introduce a separate config option for initializers.

jpsim:
Agreed, we use this style at work but for all multi line declarations, not just functions.

lordzsolt:
@weakfl Yeah, it was my bad 😞 
I only realized about `init` after I made the PR. I can look into it to see if it's just as simple as adding a new keyword.

weakfl:
> I can look into it to see if it's just as simple as adding a new keyword.

That would be great, thanks @lordzsolt

PR #3070, opened 2020-02-03
Remove base path from reported message in oss-check

marcelofabri:


marcelofabri:
Screen Shot 2020-02-03 at 12 28 15 AM

jpsim:
Nice!

PR #3077, opened 2020-02-06
Fix skipped files when target names contain spaces

### LINKED ISSUE #3076: Files not linted if target name contains a space
### New Issue Checklist

### Describe the bug

If a target name contains a space, i.e. "Experiences - Account", SwiftLint will not lint any of the files associated with that target.

image

We like to "categorize" our targets like this, so if there's a space bug we'll find it!

The root cause is a mix of needed escaping/unescape handling in **CompilerArgumentsExtractor.swift**:  `allCompilerInvocations` (and potentially `expandingResponseFiles`)

(Abridged since problem is known)

Expected ~4000 files total, not 1356:

```bash
$ swiftlint analyze --reporter emoji --compiler-log-path build_logs/xcode_build_raw.log *

Collecting 'Foo.swift' (1/1356)
Collecting 'Bar.swift' (2/1356)
Collecting 'Fizz.swift' (3/1356)
```

### Environment

0.38.2

Source

```iml
opt_in_rules:
    - unused_declaration
    - unused_import
```

  No

11.3

N/A

erikkerber:
Resolves #3076 

SwiftLint determines the files to lint based on build output. Specifically, it peeks inside of a `.SwiftFileList` file:

```bash
export SWIFT_RESPONSE_FILE_PATH_normal_x86_64="/Users/erik/Library/Developer/Xcode/DerivedData/Flagship-cxvakyfeadqmyecwvirxgmrbkpoa/Build/Intermediates.noindex/Flagship.build/Debug-iphonesimulator/Experiences - CartCheckout.build/Objects-normal/x86_64/CartCheckout.SwiftFileList"
```

These build output lines are split on spaces. If the line that contains the file list has a split - then the path itself will be split:

```
Debug-iphonesimulator/Experiences\ 
-\ 
CartCheckout.build/Objects-normal/x86_64/CartCheckout.SwiftFileList
```

No file list will be found, and no Swift files will be added to the list of files to lint.

### Fix

Adding onto the space-escaping pattern used in [CompilerArgumentsExtractor.swift ](https://github.com/realm/SwiftLint/blob/master/Source/swiftlint/Helpers/CompilerArgumentsExtractor.swift#L86), convert escaped spaces immediately into Unicode "null", allowing code that requires splitting on spaces to be able to rely simply on `components(separatedBy: " ")`.

When needed, unescape these strings just-in-time.

> **Note:** I just pulled the code today so I'm not familiar (yet) with the layout of the tests, but at first glance I didn't recognize any test fixtures/fake build files that tested the parsing and loading of build output. If there's something there, happy to add to it.

### DIFF
--- a/Source/swiftlint/Helpers/CompilerArgumentsExtractor.swift
+++ b/Source/swiftlint/Helpers/CompilerArgumentsExtractor.swift
@@ -1,14 +1,20 @@
 import Foundation
 import SourceKittenFramework
 
+private let escapedSpacePlaceholder = "\u{0}"
+
 struct CompilerArgumentsExtractor {
     static func allCompilerInvocations(compilerLogs: String) -> [String] {
         var compilerInvocations = [String]()
         compilerLogs.enumerateLines { line, _ in
             if let swiftcIndex = line.range(of: "swiftc ")?.upperBound, line.contains(" -module-name ") {
-                let invocation = line[swiftcIndex...]
+                let invocation = line
+                    .replacingOccurrences(of: "\\ ", with: escapedSpacePlaceholder)[swiftcIndex...]
                     .components(separatedBy: " ")
+                    .map { $0.replacingOccurrences(of: escapedSpacePlaceholder, with: " ") }
                     .expandingResponseFiles
+                    .map { $0.replacingOccurrences(of: "\\ ", with: escapedSpacePlaceholder) }
+                    .map { $0.replacingOccurrences(of: " ", with: escapedSpacePlaceholder) }
                     .joined(separator: " ")
                 compilerInvocations.append(invocation)
             }
@@ -18,7 +24,7 @@ struct CompilerArgumentsExtractor {
     }
 
     static func compilerArgumentsForFile(_ sourceFile: String, compilerInvocations: [String]) -> [String]? {
-        let escapedSourceFile = sourceFile.replacingOccurrences(of: " ", with: "\\ ")
+        let escapedSourceFile = sourceFile.replacingOccurrences(of: " ", with: escapedSpacePlaceholder)
         guard let compilerInvocation = compilerInvocations.first(where: { $0.contains(escapedSourceFile) }) else {
             return nil
         }
@@ -83,7 +89,6 @@ private extension Scanner {
 #endif
 
 private func parseCLIArguments(_ string: String) -> [String] {
-    let escapedSpacePlaceholder = "\u{0}"
     let scanner = Scanner(string: string)
     var str = ""
     var didStart = false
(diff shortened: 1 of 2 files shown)


Source/swiftlint/Helpers/CompilerArgumentsExtractor.swift
```diff
@@ -1,14 +1,19 @@
...
         var compilerInvocations = [String]()
         compilerLogs.enumerateLines { line, _ in
             if let swiftcIndex = line.range(of: "swiftc ")?.upperBound, line.contains(" -module-name ") {
-                let invocation = line[swiftcIndex...]
+                let invocation = line
+                    .replacingOccurrences(of: "\\ ", with: escapedSpacePlaceholder)[swiftcIndex...]
                     .components(separatedBy: " ")
+                    .map { $0.replacingOccurrences(of: escapedSpacePlaceholder, with: " ") }
                     .expandingResponseFiles
+                    .map { $0.replacingOccurrences(of: " ", with: escapedSpacePlaceholder) }
```
eskerber: I'm almost positive there is a more elegant solution to the 3 transformations I added here, but will want to think about how it affects the escaping strategy used in `parseCLIArguments` and `expandingResponseFiles`.

Source/swiftlint/Helpers/CompilerArgumentsExtractor.swift
```diff
@@ -1,14 +1,20 @@
...
         compilerLogs.enumerateLines { line, _ in
             if let swiftcIndex = line.range(of: "swiftc ")?.upperBound, line.contains(" -module-name ") {
-                let invocation = line[swiftcIndex...]
+                let invocation = line
+                    .replacingOccurrences(of: "\\ ", with: escapedSpacePlaceholder)[swiftcIndex...]
                     .components(separatedBy: " ")
+                    .map { $0.replacingOccurrences(of: escapedSpacePlaceholder, with: " ") }
                     .expandingResponseFiles
+                    .map { $0.replacingOccurrences(of: "\\ ", with: escapedSpacePlaceholder) }
+                    .map { $0.replacingOccurrences(of: " ", with: escapedSpacePlaceholder) }
```
eskerber: To get this to work for us in the short term, this fix got a little out of control. 

Because the invocations get "expanded" by reading/appending the `.SwiftFileList` files, which also may contain spaces, another escaping sequence needs to happen. 

There is almost certainly a more elegant way of achieving this, which I'll hopefully flush out.

brentleyjones:
@jpsim / @marcelofabri: Could we kindly get a review on this? Thank you 🙇

marcelofabri:
I'm not very familiar with this part of SwiftLint, so I'd be way more comfortable merging this if we first moved this file to `SwiftLintFramework` and added some unit tests (preferably in another PR).

brentleyjones:
@marcelofabri We've moved on from having to use this Patch, instead using a compilation database (`compile_commands.json`). For `xcodebuild` I'm sure it's still needed, but it will be very low priority for us to now work on it, if we can get to it at all.

marcelofabri:
No worries! I'm glad you found an alternative solution 💯

sethfri:
It looks like from @brentleyjones's comment on 8/4 that this PR has been abandoned, so I'm going to go ahead and close it out. Please correct me if I'm wrong!

erikkerber:
Nope! Good to close.

PR #3085, opened 2020-02-08
Add `void_function_in_ternary` opt-in rule

### LINKED ISSUE #2358: Don't use ternary as an alternative for if else
### Rule Request

Ternary operator should be used just on right side of assignment or in return statement. And should not be used as an alternative for `if/else`. As it is very unusual and harder to read than usual `if/else`
https://softwareengineering.stackexchange.com/a/294560

```
// NOT TRIGGERED
let result = success ? "Success" : "Fail"

if success {
    askQuestion()
} else {
    exit()
}

var price: Double {
    return hasDiscount ? initialPrice - discount : initialPrice
}

// TRIGGERED
success ? askQuestion() : exit()

```

No parameters expected

It can be enabled by default

marcelofabri:
Fixes #2358

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,11 @@
   `return <expression>` in a function that is `Void`.  
   [Marcelo Fabri](https://github.com/marcelofabri)
 
+* Add `void_function_in_ternary` opt-in rule to warn against using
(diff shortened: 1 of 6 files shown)


marcelofabri:
I need to update the rule name

Issue #3105, opened 2020-02-11
Swiftlint 0.39 library not loaded


Oliver-Kirkland-Evoke:
### New Issue Checklist

- [X] Updated SwiftLint to the latest version
- [X] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

swiftlint always crashes after updating to version 0.39 via Homebrew.  We've tried this on 3 systems and seen the same issue.  

We have tried installing "Swift 5 Runtime Support for Command Line Tools" as per the known issues but the systems are too new to allow this to be installed.  

We have downgraded to 0.38.2 which is working fine.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint lint
dyld: Library not loaded: @rpath/lib_InternalSwiftSyntaxParser.dylib
  Referenced from: /usr/local/bin/swiftlint
  Reason: image not found
Abort trap: 6
```

### Environment

* SwiftLint version: 0.39
* macOS versions: 10.15.3 && 10.14.6
* Installation method used: Homebrew

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)?
  If so, paste their relative paths and respective contents.
* Which Xcode version are you using: 11.3
* Do you have a sample that shows the issue? Run `swiftlint"

marcelofabri:
@Oliver-Kirkland-Evoke does it work if you try `brew reinstall --build-from-source swiftlint`?

Doxxer:
Same issue, Installation method used: Cocoapods

marcelofabri:
> Same issue, Installation method used: Cocoapods

Can you post my details about your environment? I can image why using homebrew wouldn't work, but using CP should work fine 🤔

PatrickGaissert:
We have the same problem. Integration is also via Cocoapods, Xcode 11.3.1.

marcelofabri:
OK, so looks like the issue is that the binary is somehow hardcoding the path to the Xcode path from the machine that created the binary:

```
$ DYLD_PRINT_RPATHS=1 swiftlint help
RPATH failed expanding     @rpath/lib_InternalSwiftSyntaxParser.dylib to: /usr/lib/swift/lib_InternalSwiftSyntaxParser.dylib
[...]

PR #3107, opened 2020-02-11
Temporarily remove all SwiftSyntax rules and support

jpsim:
The new rules introduced in 0.39.0 that depend on SwiftSyntax have been temporarily removed as we work out release packaging issues.

* `prohibited_nan_comparison`
* `return_value_from_void_function`
* `tuple_pattern`
* `void_function_in_ternary`

See https://github.com/realm/SwiftLint/issues/3105 for details.


jpsim:
Sorry to revert all this stuff @marcelofabri. At least all this stuff is fairly self-contained and shouldn't lead to too many merge conflicts when we sort out the packaging details.

4np:
@jpsim : now that #3105 is closed, will these rules be re-enabled?

jpsim:
Once someone comes up with a suitable distribution plan yes.

Issue #2758, opened 2019-05-14
Rule Request: Warn on use of some NS types in Swift code


72A12F4E:
### New Issue Checklist

- [X] Updated SwiftLint to the latest version
- [X] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

Please describe the rule idea, format
this issue's title as `Rule Request: [Rule Name]` and describe:

1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.

This would prevent developers from using legacy NS types where swift versions of those types already exist.

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

**Bad**

All of the below types have a Swift analogue and should not be used in Swift code

```swift
let foo1: NSDictionary = bar1
let foo2 = NSArray()
let foo3: NSString = bar3
let foo4 = NSSet()
```

**Ok**

These types do not have a Swift analogue, so they should be allowed.

```swift
let vc = NSViewController()
let orderedSet = NSOrderedSet()
```

3. Should the rule be configurable, if so what parameters should be configurable?

Unsure

4. Should the rule be opt-in or enabled by default? Why?
   See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a rule as opt-in.

I think this rule should be opt-in by default. There is a lot of potential for flagging types that should be completely valid in normal Swift code, and in codebases where very particular Swift-ObjC is required.

72A12F4E:
Here is a regex based rule that I implemented to identify these issues

```iml
custom_rules:
  legacy_objc_type:
    included: ".*.swift"
    name: "Legacy Obj-C Type"
    regex: "NSString|NSSet|NSArray|NSDictionary"
    message: "Avoid using legacy ObjC types where Swift types will suffice."
    severity: warning
    match_kinds:
      - typeidentifier
```

PR #3110, opened 2020-02-13
Add `legacy_objc_type` Opt In Rule

72A12F4E:
This change implements #2758

I was getting some unexpected behavior when I tried using SwiftLintFile.match(pattern:with:range:) to match the expected areas of code that I wanted, so I just filtered the matches manually.

I expected SwiftLintFile.match(pattern:with:range:) to return if any of the SyntaxKinds in the array I passed in were a match, but in the implementation its checking to see if the parameter has a match on every SyntaxKind passed in. It's like an AND instead of an OR?

New to the codebase, so I'm not sure if there is a better method for doing what I wanted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@ This is the last release to support building with Swift 5.0.x.
   [JP Simard](https://github.com/jpsim)
   [#3116](https://github.com/realm/SwiftLint/issues/3116)
 
+* Add `legacy_objc_type` opt-in rule to warn against using
+  bridged Objective-C reference types instead of Swift value types.  
+  [Blake](https://github.com/72A12F4E)
+  [#2758](https://github.com/realm/SwiftLint/issues/2758)
+
 #### Bug Fixes
 
 * Fix more false positives in `implicit_getter` rule in extensions when using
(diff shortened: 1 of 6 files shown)


72A12F4E:
I was getting some unexpected behavior when I tried using `SwiftLintFile.match(pattern:with:range:)` to match the expected areas of code that I wanted, so I just filtered the matches manually. New to the codebase, so I'm not sure if there is a better method for doing what I wanted

72A12F4E:
Also, I seem to be getting a ton of CI build errors unrelated to my changes where SourceKittenFramework doesn't look like its being imported even though it is. 

Is this an existing problem?

72A12F4E:
@jpsim I think I have this to a place where I am comfortable with it. Can you take a quick pass and let me know if there is anything I could improve?

jpsim:
Sorry for the long delay, this looks good, I've revived in https://github.com/realm/SwiftLint/pull/3531 and will merge when CI passes.

72A12F4E:
Thanks so much JP! I'm really glad this finally made it! 😄 Keep up the great work on SwiftLint!

PR #3117, opened 2020-02-16
Add unused_import config option to require imports for each module used

jpsim:
*This is a proof of concept, see TODOs for what would need to be cleaned up*

For example, if `CGFloat` is used in a file where only `UIKit` is imported but not `CoreGraphics`, this will be a violation even if the file previously compiled.

This is because Swift allows referencing some declarations that are only transitively imported.

Enabling the `require_explicit_imports` configuration option will require that the module of every declaration referenced in a source file be explicitly imported.

This will add significant noise to the imports list, but has a few advantages:

1. It will be easier to understand all the dependencies explicitly referenced in a source file.
2. Correcting the `unused_import` rule will no longer introduce compilation errors in files that compiled prior to the correction.

Missing imports will always be added to the top of the file, which may cause `sorted_imports` violations. You may run SwiftLint autocorrect with the `sorted_imports` rule enabled to re-sort those import statements afterwards.


jpsim:
Continuing in #3123.

PR #3121, opened 2020-02-18
Reduce overhead when creating unique IDs for SwiftLintFile instances

jpsim:
Generating a globally unique integer ID for each class instance should be functionally equivalent to using that class's `ObjectIdentifier`, without the overhead of locking or storing that property in memory.


jpsim:
Not sure what's up here but I also won't have time to continue this, so closing.

PR #3123, opened 2020-02-18
Add unused_import config options to require imports for each module used

jpsim:
Continued from https://github.com/realm/SwiftLint/pull/3117

For example, if `CGFloat` is used in a file where only `UIKit` is imported but not `CoreGraphics`, this will be a violation even if the file previously compiled.

This is because Swift allows referencing some declarations that are only transitively imported.

Enabling the `require_explicit_imports` configuration option will require that the module of every declaration referenced in a source file be explicitly imported.

This will add significant noise to the imports list, but has a few advantages:

1. It will be easier to understand all the dependencies explicitly referenced in a source file.
2. Correcting the `unused_import` rule will no longer introduce compilation errors in files that compiled prior to the correction.

If missing imports are added to a file when correcting it, the `sorted_imports` rule will be automatically run on that file.

If you with to allow some imports to be implicitly importable transitively, you may specify the `allowed_transitive_imports` configuration:

```yaml
unused_import:
  require_explicit_imports: true
  allowed_transitive_imports:
    - module: Foundation
      allowed_transitive_imports:
        - CoreFoundation
        - Darwin
        - ObjectiveC
```


jpsim:
I want to add tests and a changelog entry before merging this.

jpsim:
Oh Swift on Linux 🤦‍♂ 

```
error: use of undeclared type 'CFData'
typealias Foo = CFData
                ^~~~~~
```

will fix tomorrow

jpsim:
@ZevEisenberg @marcelofabri @PaulTaykalo I thought you all might be interested in the enhancements to `Example` included in this PR.

We can now use the `AutomaticTestableRule` protocol & associated codegen with multiple configurations. Previously it was necessary to create new test cases manually when testing custom configurations for rules.

Source/SwiftLintFramework/Rules/Lint/UnusedImportRuleExamples.swift
```diff
@@ -0,0 +1,192 @@
...
+        ↓import Foundation
+        import UnknownModule
+        func foo(error: Swift.Error) {}
+        """),
+        Example("""
+        ↓import Foundation
+        typealias Foo = CFData
+        """, configuration: [
+            "require_explicit_imports": true
+        ])
```
PaulTaykalo: I already wrote it somewhere. there' s an idea to add a comment to the Example structure so that
it will explain why trigger should be triggered :) Totally out of the scope of this PR
```
Example("""
        ↓import Foundation
        typealias Foo = CFData
        """, configuration: [
            "require_explicit_imports": true
        ],comment: "Using transitive type 'CFData' from CoreFoundation should require CoreFoundation import")
```

This example is pretty obvious, but there are some examples that will help understanding why one or another rule should or should not be triggered.

also, this comment part can be shown in violation. I think additional PR will be needed :)

jpsim: Ah yeah, that would be great for documentation purposes, but also to have nicer descriptions when tests fail.

PaulTaykalo: Added. as an issue here. 
https://github.com/realm/SwiftLint/issues/3125

Source/SwiftLintFramework/Models/Example.swift
```diff
@@ -3,22 +3,36 @@
...
+    ///   - code:                 The contents of the example.
+    ///   - configuration:        The untyped configuration to apply to the rule, if deviating from the default
+    ///                           configuration.
+    ///   - testMultibyteOffsets: Whether the example should be tested by prepending multibyte grapheme clusters.
+    ///   - testOnLinux:          Whether the example should be tested on Linux.
+    ///   - file:                 The path to the file where the example is located.
+    ///                           Defaults to the file where this initializer is called.
+    ///   - line:                 The line in the file where the example is located.
+    ///                           Defaults to the line where this initializer is called.
+    init(_ code: String, configuration: Any? = nil, testMultiByteOffsets: Bool = true, testOnLinux: Bool = true,
```
ZevEisenberg: It would be great to add a test asserting that the two new parameters default to true unless otherwise specified.

Source/SwiftLintFramework/Rules/Lint/UnusedImportRuleExamples.swift
```diff
@@ -0,0 +1,192 @@
...
+        @testable import Foundation
+        @objc
+        class A {}
+        """),
+        Example("""
+        ↓↓import Foundation
+        typealias Foo = CFData
+        """, configuration: [
+            "require_explicit_imports": true
+        ], testMultiByteOffsets: false, testOnLinux: false):
```
ZevEisenberg: Is it just a coincidence that `testMultibyteOffsets` and `testOnLinux` are always paired in this PR? Is there a scenario where you would want one but not the other?

Source/SwiftLintFramework/Rules/Lint/UnusedImportRuleExamples.swift
```diff
@@ -0,0 +1,192 @@
...
+    ]
+
+    static let corrections = [
+        Example("""
+        ↓import Dispatch
+        struct A {
+          static func dispatchMain() {}
+        }
+        A.dispatchMain()
+        """): Example("""
```
ZevEisenberg: If you put the values (as opposed to the keys) in this dictionary on their own line, they get indented, which to my eye is easier to read.

Source/SwiftLintFramework/Rules/RuleConfigurations/UnusedImportConfiguration.swift
```diff
@@ -0,0 +1,53 @@
+public struct TransitiveModuleConfiguration: Equatable {
```
ZevEisenberg: Docs on this type would be nice

CHANGELOG.md
```diff
@@ -10,7 +10,14 @@
 
 #### Enhancements
 
-* None.
+* Add configuration options to the `unused_import` rule to require
+  explicit import statements for each module referenced in a source
+  file (`require_explicit_imports`). When this setting is enabled,
+  a `allowed_transitive_imports` setting may also be specified to allow
```
ZevEisenberg:
```suggestion
  an `allowed_transitive_imports` setting may also be specified to allow
```

Source/SwiftLintFramework/Extensions/Dictionary+SwiftLint.swift
```diff
@@ -118,6 +118,21 @@ public struct SourceKittenDictionary {
         return value["key.attribute"] as? String
     }
 
+    /// Module name in @import expressions.
```
ZevEisenberg:
```suggestion
    /// Module name in `@import` expressions.
```

Source/SwiftLintFramework/Extensions/SwiftDeclarationAttributeKind+Swiftlint.swift
```diff
@@ -1,6 +1,15 @@
 import SourceKittenFramework
 
 extension SwiftDeclarationAttributeKind {
+    static var attributesRequiringFoundation: Set<SwiftDeclarationAttributeKind> {
+        return [
+            .objc,
+            .objcName,
+            .objcMembers,
+            .objcNonLazyRealization
```
ZevEisenberg: Not in scope for this PR, but I wanted to call it out: the lack of trailing commas on the last line of collection literals made #3040 harder, because I had to special-case a bunch of search and replace. I strongly recommend enforcing trailing commas in any collection-heavy codebase. Just my 2¢

Source/SwiftLintFramework/Models/Example.swift
```diff
@@ -3,22 +3,36 @@
 public struct Example {
     /// The contents of the example
     public private(set) var code: String
+    /// The untyped configuration to apply to the rule, if deviating from the default configuration
+    public private(set) var configuration: Any?
```
ZevEisenberg: Why is the configuration untyped? Might be nice to include something in the comma, so others understand the intended usage.

Source/SwiftLintFramework/Models/Example.swift
```diff
@@ -3,22 +3,36 @@
...
+    /// Whether the example should be tested by prepending multibyte grapheme clusters
+    public private(set) var testMultiByteOffsets: Bool
```
ZevEisenberg: It's not very clear from the comment what is actually meant. After reading it, I have several questions that I'd prefer not to have to go digging through the source code to answer:

- why do multibyte grapheme clusters matter?
- why might I want to disable them?
- which multibyte clusters are tested? Are they specific or random?
- where and when are the multibyte characters prepended, and what effect does that have?

Source/SwiftLintFramework/Rules/Lint/UnusedImportRuleExamples.swift
```diff
@@ -0,0 +1,192 @@
...
+        }
+        A.dispatchMain()
+        ↓import Dispatch
+
+        """): Example("""
+        struct A {
+          static func dispatchMain() {}
+        }
+        A.dispatchMain()
+
```
ZevEisenberg: Intentional whitespace? Here and above.

Source/SwiftLintFramework/Models/SwiftLintFile.swift
```diff
@@ -67,6 +67,24 @@ public final class SwiftLintFile {
     public var lines: [Line] {
         return file.lines
     }
+
+    /// Returns whether or not the file contains any attributes that require the Foundation module.
+    func containsAttributesRequiringFoundation() -> Bool {
+        guard contents.contains("@objc") else {
```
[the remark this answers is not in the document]
jpsim: No, this is just a fast path that skips more expensive checks further down.

Source/SwiftLintFramework/Rules/Lint/UnusedImportRuleExamples.swift
jpsim: Just a coincidence.

Source/SwiftLintFramework/Rules/Lint/UnusedImportRuleExamples.swift
```diff
@@ -0,0 +1,192 @@
...
+        A.dispatchMain()
+        """),
+        Example("""
+        ↓import Foundation // This is unused
+        struct A {
+          static func dispatchMain() {}
+        }
+        A.dispatchMain()
+        ↓import Dispatch
+
```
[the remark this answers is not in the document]
jpsim: Yes, it's a limitation of the rule right now that the import statement can't be at the very end of the file. If the file has a trailing newline, that's fine.

Source/SwiftLintFramework/Rules/Lint/UnusedImportRuleExamples.swift
jpsim: Intentional, yes: https://github.com/realm/SwiftLint/pull/3123#discussion_r382943699

Source/SwiftLintFramework/Models/Example.swift
jpsim: > why do multibyte grapheme clusters matter?

In the past, some rules were written with some assumptions that each character was one byte.

Characters that extend beyond a single byte encoded in UTF-8 need to be handled correctly.

The `testMultiByteOffsets` setting prepends some multi-byte grapheme clusters to the code in the example (specifically `/* 👨‍👩‍👧‍👦👨‍👩‍👧‍👦👨‍👩‍👧‍👦 */`) to ensure that the locations where violations are reported are correct, and that corrections are properly applied.

> why might I want to disable them?

Some examples aren't resilient to prepending this content to its code and need to disable this part of the test. For example, this unused import code:

```swift
import Foundation
typealias Foo = CFData
```

Gets corrected by replacing `import Foundation` with `import CoreFoundation`:

```swift
import CoreFoundation
typealias Foo = CFData
```

However, when the violating example gets translated to this by the `testMultiByteOffsets` setting:

```swift
/* 👨‍👩‍👧‍👦👨‍👩‍👧‍👦👨‍👩‍👧‍👦 */
import Foundation
typealias Foo = CFData
```

then the correction becomes:

```swift
import CoreFoundation
/* 👨‍👩‍👧‍👦👨‍👩‍👧‍👦👨‍👩‍👧‍👦 */
typealias Foo = CFData
```

Which fails the test. There's nothing wrong with the rule's behavior, it just doesn't match what the simplistic auto-generated rule expects.

> which multibyte clusters are tested? Are they specific or random?

Specific: `/* 👨‍👩‍👧‍👦👨‍👩‍👧‍👦👨‍👩‍👧‍👦 */`

> where and when are the multibyte characters prepended, and what effect does that have?

See above.

AdrianBinDC:
Awesome addition! Thank you for adding this.

AdrianBinDC:
If I'm running `swiftlint analyze --compiler-log-path ./xcodebuild.log --autocorrect --reporter json > swiftlint.json`, can you clarify what the syntax should be in my `.swiftlint.yml` file?

In the example above, I'd like to require explicit imports with no transitive imports

```
unused_import:
  require_explicit_imports: true
```

What would the `analyzer_rules` section of the `yml` file look like?

```
analyzer_rules:
- unused_import:

opt_in_rules:
  -   require_explicit_imports
```

logs this error...

> Loading configuration from '.swiftlint.yml'
> Found a configuration for 'empty_count' rule, but it is not enabled on 'opt_in_rules'.
> configuration error: 'require_explicit_imports' is not a valid rule identifier

AdrianBinDC:
When will this be released? The last release was 15 days ago.

mthole:
(two years later)

I used this today on a large project. Thanks for this addition!

jpsim:
I’m glad you found it useful! We run it daily on Lyft’s 1M+ lines of Swift.

PR #3126, opened 2020-02-19
Add `prohibited_transparent` and `prohibited_inline` opt-in rules

### LINKED ISSUE #3138: Rule request: `prohibited_inline` and `prohibited_transparent`
See #3126 for PR.

### New Issue Checklist

### New rule request

Apple engineers have gone on record to write not to use inline and transparent decorators:
https://lists.swift.org/pipermail/swift-users/Week-of-Mon-20170227/004886.html

#### Inline
_Would_
`@inline(__always) func f() {}`
`@inline(never) dynamic func f() {}`
`@objc @inline(__always) func f() {}`
`@inline(__always) @objc func f() {}`
`@inline(never) dynamic func f() {}`
`@inline(__always)\nfunc f() {}`
`@inline(never)\ndynamic func f() {}`
`@objc\n@inline(__always) func f() {}`
`@inline(__always)\n@objc func f() {}`
`@inline(never)\ndynamic func f() {}`
`@objc\n@inline(__always)\nfunc f() {}`
`@inline(__always)\n@objc\nfunc f() {}`
`@inline(never)\ndynamic\nfunc f() {}`

_Wouldn't_
`func f() {}`
`dynamic func f() {}`
`@objc func f() {}`
`@_transparent func f()`

#### Transparent
_Would_
`@_transparent func f() {}`

_Wouldn't_
`func f() {}`
`@inline(__always) func f() {}`
`@inline(never) func f() {}`
`@objc func f() {}`

The configuration severity should be configurable.

These rules ought to be enabled by default.

iliaskarim:
Resolves #3138


PaulTaykalo:
@iliaskarim please, revert the changed from the README.md and leave only relevant ones

iliaskarim:
> * It's a bit unclear why would someone want to prohibit dynamic or inline attributes in files
- [Dynamic dispatch is at odds with Swift performance](https://developer.apple.com/swift/blog/?id=27). It's also a good way to introduce bugs.
- [Inline was discouraged](https://lists.swift.org/pipermail/swift-users/Week-of-Mon-20170227/004886.html)
Mainly the reasons are to regulate code style by discouraging poor practice which is what the linter helps with.

> * There are too many unrelated changes in CHANGELOG.md
It was trimming trailing whitespace for you. There's no problem changing these back if you like them.

> * These rules shouldn't be enabled by default
They witness `OptInRule`.
 
> Is there any specific need of having this kind of rules?
Yes there is, see above.

PaulTaykalo:
@iliaskarim Nice. It is a good idea to put why the rule needs to be added in the description.
Now it makes sense.

The tests are failing. 

P.S. As for README.md 
1) In general, it is a good idea to make PR only about functionality it describes. So the best way (if that reformatting was needed) would be to make another PR with reformatting

2) those spaces are required for forcing `newline` in .md files rendered by Github

## Before

image


## Update
image

## After

image

Source/SwiftLintFramework/Rules/Lint/ProhibitedInlineRule.swift
```diff
@@ -0,0 +1,65 @@
...
+import SourceKittenFramework
+
+public struct ProhibitedInlineRule: ASTRule, OptInRule, ConfigurationProviderRule, AutomaticTestableRule {
+    public var configuration = SeverityConfiguration(.error)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "prohibited_inline",
+        name: "Prohibted Inline",
```
esteluk: There's a typo in the spelling of Prohibited here :)

iliaskarim:
@PaulTaykalo I've updated the dynamic rule to be about transparent instead. These two rules seemed more complementary to me than dynamic, which would be nice to revisit later as an opt-in rule.

We have on record engineers from Apple discouraging the use of transparent and inline decorators. Does enabling the rules by default make sense to you? The severities can always be reconfigured. If it makes sense, the two rules could be made into one rule.

Issue #59, opened 2015-05-28
Validate that self is only used in escaping closures and conflicts


jpsim:
Use of `self` should be limited to escaping closures and scopes in which other declarations conflict.

scottrhoyt:
Counterpart of #321

eneko:
:+1:

diederich:
Love this. It would make searching for ref-cycles quiet a bit easier... :-)

charleyfromage:
I could find the "explicit_self" rule but I couldn't find the identifier for this one.

Anyone could help me with finding it ?

Thanks!

jpsim:
> Anyone could help me with finding it ?

Doesn't exist yet. If anyone would like to build this rule, please go ahead!

patrickbdev:
Any update on this?

sethfri:
Reopening, as there is still demand for this rule based on #3579 and prior comments on 2020.

pronebird:
Would love to see that implemented.

lizjakubowski:
bump - my team would find this rule to be super useful ❤️

mqln:
Bumptiybumpbump.

Any status updates on this issue? I think we all want it ❤️

amadeu01:
Any updates on this?

TripwireNL:
Has this been implemented already? I would love to have this rule in SwiftLint!

acicartagena:
hello, bump any updates on the concerns/challenges of implementing this?

mqln:
Bumpbumpbump part 2?

SimplyDanny:
I've implemented parts for this rule in #4911. As mentioned there, what's missing are two cases:

* Weakly captured `self` with explicit unwrapping according to [SE-0365](https://github.com/apple/swift-evolution/blob/main/proposals/0365-implicit-self-weak-capture.md)
* Implicit `self` in classes then the closure isn't `@escaping`

These can be implemented in a later step. I didn't want to make the first implementation overly complex and rather gather your feedback on other cases I might have missed.

Please have a look into the PR and let me know which other test cases you can think of, what I've missed or what doesn't work like the rule assumes!

SimplyDanny:
Closing this as completed. The remaining part "implicit `self` in classes when the closure isn't `@escaping`" is requested by #4964.

PR #3140, opened 2020-03-09
Add 'implicit_self' rule

msbm:
https://github.com/realm/SwiftLint/issues/59

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@
   [JP Simard](https://github.com/jpsim)
   [#3116](https://github.com/realm/SwiftLint/issues/3116)
 
+* Added 'implicit_self' opt-in rule.
+[msbm](https://github.com/msbm)
+[#59](https://github.com/realm/SwiftLint/issues/59)
+
 #### Bug Fixes
 
 * None.
--- a/Source/SwiftLintFramework/Rules/Style/ImplicitSelfRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/ImplicitSelfRule.swift
@@ -1,0 +1,231 @@
+import Foundation
+import SourceKittenFramework
+
+public struct ImplicitSelfRule: OptInRule, ConfigurationProviderRule {
+    public var configuration = ImplicitSelfConfiguration(severity: .warning, initSelfUsage: .never)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "implicit_self",
+        name: "Implicit Self",
+        description: "Instance variables and methods should not be accessed with 'self.' unless necessary.",
+        kind: .style
+    )
+
+    public func validate(file: SwiftLintFile) -> [StyleViolation] {
+        let identifiers = file.syntaxMap.tokens.compactMap { token -> Identifier? in
+            guard let kind = token.kind, kind == .identifier else { return nil }
+
+            return Identifier(token: token, file: file)
+        }
+
+        let structure = Structure(dict: file.structureDictionary.value)
+
+        for identifier in identifiers {
+            structure.add(identifier: identifier)
+        }
+
+        return violations(in: structure, declInScope: [], file: file)
+    }
+
+    private func violations(in structure: Structure,
+                            declInScope: Set<String>,
+                            file: SwiftLintFile) -> [StyleViolation] {
+        var existingDeclInScope: Set<String> = declInScope
+        guard !structure.isClosure else { return [] }
+        if structure.isInit && configuration.initSelfUsage == .always { return [] }
+        let violations = structure.identifiers.compactMap { identifier -> StyleViolation? in
+            guard identifier.name != "init" else { return nil }
+
+            guard !identifier.isParamDecl && !identifier.isLocalDecl else {
+                existingDeclInScope.insert(identifier.name)
+                return nil
+            }
+
+            if configuration.initSelfUsage == .beforeInitCall &&
+                structure.isInit &&
+                structure.hasInitBefore(identifier: identifier) {
+                return nil
+            }
+
+            let selfOffset = identifier.offset - 5
+            guard file.stringView.substringWithByteRange(ByteRange(location: selfOffset, length: 5)) == "self." else {
+                return nil
+            }
+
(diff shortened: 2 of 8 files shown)


svenmuennich:
I love how this is implemented without using `AnalyzerRule`, which this rule's counterpart `explicit_self` is build on, since those rules are quite slow in large code bases.

Having that said, I wonder if we can unify these two rules into a single one and just make it configurable with a `mode`:
* `always`: The original behavior of `explicit_self`. This would be the default to not break existing configurations.
* `never`: Equivalent to the proposed `implicit_self` rule configured with `initSelfUsage: never`.
* `never_except_initializer`: Equivalent to the proposed `implicit_self` rule configured with `initSelfUsage: always`.
* `never_except_initializer_before_init_call`: Equivalent to the proposed `implicit_self` rule configured with `initSelfUsage: before_init_call`.

The implementation would be based on your PR of course, to get the performance boost from dropping `AnalyzerRule`.

What do you think @msbm ?

msbm:
Yeah I think that makes total sense, given `explicit_self` and `implicit_self` are mutually exclusive. I will merge them into one rule as `explicit_self` so the new rule is backwards compatible

svenmuennich:
@msbm Did you get a chance to work on this? If not, I might find some time on the weekend to give it a shot.

msbm:
Yeah I've been (slowly, sorry) working on it. I think I'm pretty close so I'll set aside some time this weekend to finish it up and update the PR

msbm:
@svenmuennich After working on this more this weekend I'm realizing i don't think it'll actually be possible to simplify ExplicitSelf using this logic. For ImplicitSelf I'm looking at every identifier and asserting that it isn't preceded with `self.` unless it's being shadowed. Sadly, ExplicitSelf isn't the opposite of 'nothing preceded with self'. I thought I could work some other logic to do Explicit self in a similar way by looking at all the definitions of instance vars and methods and seeing if they are reused (and this worked for the existing ExplicitSelf test cases), however it won't work in the subclassing case as far as I can figure.

ricardorauber-poatek:
Hi, any news on this? Thanks!

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@
   [JP Simard](https://github.com/jpsim)
   [#3116](https://github.com/realm/SwiftLint/issues/3116)
 
+* Added 'implicit_self' opt-in rule.
+[msbm](https://github.com/msbm)
+[#59](https://github.com/realm/SwiftLint/issues/59)
```
svenmuennich, line 24:
```suggestion
* Added 'implicit_self' opt-in rule.
  [msbm](https://github.com/msbm)
  [#59](https://github.com/realm/SwiftLint/issues/59)
```

svenmuennich:
Sorry @msbm, I somehow missed the notification when you commented again 😞 

> Sadly, ExplicitSelf isn't the opposite of 'nothing preceded with self'.

This is indeed correct. I missed that when initially looking over this PR. So we probably have to stick to having two methods `explicit_self` and `implicit_self`.

Thank you for putting in the time trying to work this out though 👏

Source/SwiftLintFramework/Rules/Style/ImplicitSelfRule.swift
```diff
@@ -0,0 +1,231 @@
...
+                return nil
+            }
+
+            if configuration.initSelfUsage == .beforeInitCall &&
+                structure.isInit &&
+                structure.hasInitBefore(identifier: identifier) {
+                return nil
+            }
+
+            let selfOffset = identifier.offset - 5
```
svenmuennich, line 52: Does this work when `self` is placed on a different line than the `identifier`? E.g.:
```
self
    .someMethodCall()
```

We should probably add a test case for this either way.

Source/SwiftLintFramework/Rules/Style/ImplicitSelfRule.swift
```diff
@@ -0,0 +1,231 @@
...
+        let violations = structure.identifiers.compactMap { identifier -> StyleViolation? in
+            guard identifier.name != "init" else { return nil }
+
+            guard !identifier.isParamDecl && !identifier.isLocalDecl else {
+                existingDeclInScope.insert(identifier.name)
+                return nil
+            }
+
+            if configuration.initSelfUsage == .beforeInitCall &&
+                structure.isInit &&
```
svenmuennich, line 47: These two conditions are constant for all elements of `structure.identifiers` (i.e. `identifier`) and both are reused below  for determining `subViolations`. It might be worth extracting this into a variable.

Source/SwiftLintFramework/Rules/Style/ImplicitSelfRule.swift
```diff
@@ -0,0 +1,231 @@
...
+        return violations + subViolations
+    }
+}
+
+private struct Identifier {
+    let name: String
+    let kind: String
+    let offset: ByteCount
+
+    init?(token: SwiftLintSyntaxToken, file: SwiftLintFile) {
```
svenmuennich, line 87: Why is this an optional initializer? Should we maybe return `nil` if `name` and/or `kind` would otherwise be initialized with a fallback value of `""`?

Source/SwiftLintFramework/Rules/Style/ImplicitSelfRule.swift
```diff
@@ -0,0 +1,231 @@
...
+        return kind == "source.lang.swift.expr.tuple"
+    }
+
+    var isInit: Bool {
+        return kind == "source.lang.swift.decl.function.method.instance" &&
+            name?.starts(with: "init") ?? false
+    }
+
+    private var enclosedParameters: [Identifier] {
+        return substructure.flatMap { structure -> [Identifier] in
```
svenmuennich, line 149: A `compactMap` seems to fit better here.

jpsim:
I don't think it's possible to write this reliably without access to the post-typechecked AST (aka `AnalyzerRule`).

The SwiftLint codebase itself uses an implicit self style, so I tried turning this on for this project and got ~80 violations, most of which are false positives.

DanielKrofchick:
Are there any plans to implement this lint rule?

PR #3154, opened 2020-03-27
Fix attributes rule false positive with Swift 5.2

jpsim:
The following was triggering:

```swift
func printBoolOrTrue(_ expression: @autoclosure () throws -> Bool?) rethrows {
  try print(expression() ?? true)
}
```

Fix by adding the `rethrows` attribute kind to the rule's blacklist.


Source/SwiftLintFramework/Rules/Style/AttributesRule.swift
```diff
@@ -293,28 +293,29 @@ public struct AttributesRule: ASTRule, OptInRule, ConfigurationProviderRule {
...
-            .required,
-            .weak,
             .private,
-            .fileprivate,
-            .internal,
             .public,
-            .open,
-            .setterPrivate,
+            .required,
+            .rethrows,
```
jpsim: The addition of this kind is the only change in this file, but I resorted the collection literal to make it easier to update in the future.

PR #3155, opened 2020-03-27
Fix false positives in valid_ibinspectable rule when using Swift 5.2

jpsim:
when defining inspectable properties in class extensions with computed properties.

The following was triggering:

```swift
extension Foo {
  @IBInspectable var color: UIColor {
    set {
      self.bar.textColor = newValue
    }

    get {
      return self.bar.textColor
    }
  }
}
```

Fix by checking to see if an instance property has `set` keywords in its body when running with Swift 5.2 or later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@
   Swift 5.2.  
   [JP Simard](https://github.com/jpsim)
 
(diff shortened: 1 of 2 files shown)


ahmedsafadii:
same problem

jpsim:
Did the fix work for you?

PR #3159, opened 2020-03-29
Run jobs with Xcode 11.4

marcelofabri:


marcelofabri:
i thought there would be at least one Xcode 11.4 beta image, but i guess not

jpsim:
It's now available: https://github.com/microsoft/azure-pipelines-image-generation/pull/1511

Issue #3167, opened 2020-04-03
multiline_arguments_brackets reports false positives with calls whose one parameter spans multiple lines


noahsark769:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

The `multiline_arguments_brackets` rule produces what I would categorize as a false positive in calls which have a single argument with a multiline value. For example, `multiline_arguments_brackets` is triggered on the following examples:

```swift
public final class Logger {
    public static let shared = Logger(outputs: [
        OSLoggerOutput(),
        ErrorLoggerOutput()
    ])
}
```

```swift
let errors = try self.download([
    (description: description, priority: priority),
])
```

```swift
return SignalProducer({ observer, _ in
    observer.sendCompleted()
}).onMainQueue()
```

I'm not 100% sure that this is not intentional, but it seems to me that these cases should pass the "smell test" of multiline_arguments_brackets, since the idea with this rule is to prevent argument brackets from being mis-aligned with the actual passed arguments. I think that if `return SignalProducer({ observer, _ in observer.sendCompleted() })` passes the rule, then

```
return SignalProducer({ observer, _ in
    observer.sendCompleted()
}).onMainQueue()
```
should also pass the rule.

None of these cases are noted in the triggering or non-triggering examples at https://realm.github.io/SwiftLint/multiline_arguments_brackets.html, though a few non-triggering examples are similar:

```swift
foo { param1, param2 in
    print("hello world")
}
```

```swift
AlertViewModel.AlertAction(title: "some title", style: .default) {
    AlertManager.shared.presentNextDebugAlert()
}
```

I'd further argue that the currently triggering cases I listed above match the style of these non-triggering cases.

I'm looking for feedback here on whether this is intentional, and if it seems to the maintainers that the rule should cover these cases then I'm happy to take a crack at implementing it 😄

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? `0.39.1`
* Installation method used (Homebrew, CocoaPods, building from source, etc)? Cocoapods
* Paste your configuration file: I can provide a full configuration file if necessary, but this should be fairly easy to reproduce

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? No
* Which Xcode version are you using (check `xcodebuild -version`)? `Xcode 11.2.1 Build version 11B500`
[...]

PR #3171, opened 2020-04-05
#3167: Fixes false positives for multiline_parameters_brackets and multiline_arguments_brackets

noahsark769:
As documented in #3167, the `multiline_parameters_brackets` and `multiline_arguments_brackets` verification code assumes that the definition/call body is multiline if there's a newline in the body, but this creates false positives when the argument itself contains newlines (e.g. an array or closure) or when the parameter has a default value which contains newlines (e.g. array or dictionary).

This PR fixes the false positives by calculating whether a definition/call is multiline based on whether there are more newlines in the definition/call body than in the sum of the parameter bodies.

I added a couple of utils (+ tests) to enable this code and added a few triggering/nontriggering examples that would have caught this issue.

This is my first PR to SwiftLint, so happy to take advice from the maintainers on any part of this code. Thanks for maintaining SwiftLint, and can't wait to see these fixes applied so we can enable `multiline_parameters_brackets` in our app!

### DIFF
--- a/Tests/SwiftLintFrameworkTests/ExtendedStringTests.swift
+++ b/Tests/SwiftLintFrameworkTests/ExtendedStringTests.swift
@@ -1,10 +1,3 @@
-//
-//  ExtendedStringTests.swift
-//  SwiftLintFrameworkTests
-//
-//  Created by Noah Gilmore on 4/5/20.
-//
-
 import XCTest
 
 class ExtendedStringTests: XCTestCase {
--- a/Tests/SwiftLintFrameworkTests/ExtendedStringTests.swift
+++ b/Tests/SwiftLintFrameworkTests/ExtendedStringTests.swift
@@ -1,0 +1,16 @@
+//
+//  ExtendedStringTests.swift
+//  SwiftLintFrameworkTests
+//
+//  Created by Noah Gilmore on 4/5/20.
+//
+
+import XCTest
+
+class ExtendedStringTests: XCTestCase {
+    func testCountOccurrences() {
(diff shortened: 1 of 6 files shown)


noahsark769:
Seems that this PR (correctly) fixes quite a few violations in the evaluated open source projects. Note that there are a few instances which are tricky since they look like violations of multiline_arguments_brackets but are actually violations of multiline_literal_brackets, like:

```
someCall([Something()
          Someting])
```

I'll add myself to the changelog and try and see why danger is failing.

Tests/SwiftLintFrameworkTests/ExtendedStringTests.swift
```diff
@@ -0,0 +1,16 @@
+//
+//  ExtendedStringTests.swift
+//  SwiftLintFrameworkTests
+//
+//  Created by Noah Gilmore on 4/5/20.
+//
```
PaulTaykalo, line 6: There are no header comments in other tests. Can we remove these?

noahsark769: Thanks for the review! Sure, will take care of this soon.

noahsark769:
@PaulTaykalo done! Thanks again for taking a look 👍

noahsark769:
Ah, danger says:

> Please rebase to get rid of the merge commits in this PR

@PaulTaykalo not familiar with the git workflow we're using in SwiftLint, should I rebase this branch and force push? Didn't see anything about it in CONTRIBUTING.md

PaulTaykalo:
@noahsark769 merged in

noahsark769:
Awesome, thank you!

PR #3174, opened 2020-04-07
Add IBInspectableInExtensionRule

keith:
This rule validates that there are no `@IBInspectable` variables in
`extension`s. The reason for this is that Interface Builder doesn't take
into account what code is actually included in your target to determine
if an inspectable should be valid. So in this case:

```swift
// Module A
extension UIView {
  @IBInspectable public var foo: String? {
    // do something
  }
}
```

```swift
// Module B
// Uses `foo` in IB
```

If module B does not depend on module A interface builder will still
show the `foo` IBInspectable, but it will fail at runtime.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
   found issues.  
   [krin-san](https://github.com/krin-san)
   [#3177](https://github.com/realm/SwiftLint/pull/3177)
-* Add `ibinspectable_in_extension` rule to lint against `@IBInspectable`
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,9 @@
   found issues.  
   [krin-san](https://github.com/krin-san)
   [#3177](https://github.com/realm/SwiftLint/pull/3177)
+* Add `ibinspectable_in_extension` rule to lint against `@IBInspectable`
```
marcelofabri, line 18: maybe update it to mention that this is an opt-in rule?

PR #3184, opened 2020-04-23
Remove @IBOutlet and @IBInspectable from UnusedDeclarationRule

keith:
This list was intended to be things that could have side effects from
being in IB, or just being dynamic, and therefore we cannot tell if they
are used or not. In the case of the 2 attributes I'm removing, they
could be used in IB, but if they aren't used in code, they should still
be considered dead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@
   [mknabbe](https://github.com/mknabbe)
   [#1280](https://github.com/realm/SwiftLint/issues/1280)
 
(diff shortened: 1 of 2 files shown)


keith:
Note making this change for folks is not entirely safe. If you remove an IBInspectable definition, without removing it from IB, it just warns in the console, but if you remove an IBOutlet without removing it from IB, it crashes.

keith:
@jpsim what do you think about this?

jpsim (approved):
I think it makes sense to do this. Can you please add a change log entry and merge yourself when CI is green? Thanks!

PR #3188, opened 2020-04-24
Improve compile commands error handling

keith:
Previously when a compile commands file was invalid, it could be hard to
figure out what about it was wrong. This adds some more granular error
handling which hopefully helps provide valid files here.

### DIFF
--- a/Source/swiftlint/Helpers/LintableFilesVisitor.swift
+++ b/Source/swiftlint/Helpers/LintableFilesVisitor.swift
@@ -137,13 +137,14 @@ struct LintableFilesVisitor {
             return .success(.buildLog(compilerInvocations: compilerInvocations))
         } else if !options.compileCommands.isEmpty {
             let path = options.compileCommands
-            guard let compileCommands = self.loadCompileCommands(path) else {
+            switch self.loadCompileCommands(path) {
+            case .success(let compileCommands):
+                return .success(.compilationDatabase(compileCommands: compileCommands))
+            case .failure(let error):
                 return .failure(
-                    .usageError(description: "Could not read compilation database at path: '\(path)'")
+                    .usageError(description: "Could not read compilation database at path: '\(path)' \(error)")
                 )
             }
-
-            return .success(.compilationDatabase(compileCommands: compileCommands))
         }
 
         return .failure(.usageError(description: "Could not read compiler invocations"))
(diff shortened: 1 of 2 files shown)


Source/swiftlint/Helpers/LintableFilesVisitor.swift
```diff
@@ -162,25 +163,64 @@ struct LintableFilesVisitor {
...
+            }
+
+            // Compilation databases include the compiler, but it's left out when sending to SourceKit.
+            if arguments.first == "swiftc" {
+                arguments.removeFirst()
             }
+
+            if !arguments.contains(file) {
+                return .failure(.missingFileInArguments(path, index, arguments))
+            }
```
keith, line 196: This is a new check, I'm a bit worried about the performance implications of this, especially since it's a small sanity check, but I added it here for discussion

Source/swiftlint/Helpers/LintableFilesVisitor.swift
```diff
@@ -162,25 +163,64 @@ struct LintableFilesVisitor {
...
+        guard let object = try? JSONSerialization.jsonObject(with: jsonContents),
             let compileDB = object as? [[String: Any]] else {
-            return nil
+            return .failure(.malformedCommands(path))
         }
 
         // Convert the compilation database to a dictionary, with source files as keys and compiler arguments as values.
         //
         // Compilation databases are an array of dictionaries. Each dict has "file" and "arguments" keys.
-        return compileDB.reduce(into: [:]) { (commands: inout [File: Arguments], entry: [String: Any]) in
```
keith, line 175: I wasn't happy swapping this reduce to a for loop, but chose to do this over throwing since `Result` is more widely used in this codebase AFAICT

keith:
cc @kastiglione

Source/swiftlint/Helpers/LintableFilesVisitor.swift
```diff
@@ -137,13 +137,14 @@ struct LintableFilesVisitor {
...
         } else if !options.compileCommands.isEmpty {
             let path = options.compileCommands
-            guard let compileCommands = self.loadCompileCommands(path) else {
+            switch self.loadCompileCommands(path) {
+            case .success(let compileCommands):
+                return .success(.compilationDatabase(compileCommands: compileCommands))
+            case .failure(let error):
                 return .failure(
-                    .usageError(description: "Could not read compilation database at path: '\(path)'")
+                    .usageError(description: "Could not read compilation database at path: '\(path)' \(error.description)")
```
jpsim, line 145: This line is triggering a line length violation because it's 123 characters long, beyond the 120 character limit.

Source/swiftlint/Helpers/LintableFilesVisitor.swift
```diff
@@ -162,25 +163,64 @@ struct LintableFilesVisitor {
...
+
+            commands[file] = CompilerArgumentsExtractor.filterCompilerArguments(arguments)
+        }
+
+        return .success(commands)
+    }
+}
+
+private enum CompileCommandsLoadError: Error {
+    case nonExistantFile(String)
```
jpsim, line 206: Nit:

```suggestion
    case nonExistentFile(String)
```

Source/swiftlint/Helpers/LintableFilesVisitor.swift
```diff
@@ -162,25 +163,64 @@ struct LintableFilesVisitor {
...
+        case let .nonExistantFile(path):
+            return "Could not read compile commands file at '\(path)'"
+        case let .malformedCommands(path):
+            return "Compile commands file at '\(path)' isn't in the correct format"
+        case let .malformedFile(path, index):
+            return "Missing or invalid (must be a string) 'file' key in \(path) at index \(index)"
+        case let .malformedArguments(path, index):
+            return "Missing or invalid (must be an array of strings) 'arguments' key in \(path) at index \(index)"
+        case let .missingFileInArguments(path, index, arguments):
+            return "Entry in \(path) at index \(index) has 'arugments' which do not contain the 'file': \(arguments)"
```
jpsim, line 223:
```suggestion
            return "Entry in \(path) at index \(index) has 'arguments' which do not contain the 'file': \(arguments)"
```

Source/swiftlint/Helpers/LintableFilesVisitor.swift
jpsim: Pretty minor, but I'd prefer if this check were a `guard` above the `swiftc` removal for consistency with the rest of the error handling checks.

As for the performance overhead, I think this is trivial and completely dwarfed by the cost of running analyzer rules.

Source/swiftlint/Helpers/LintableFilesVisitor.swift
jpsim: I'm happy with the change.

keith:
> we should note this in the readme

Added to the changelog, should I also put it somewhere else?

jpsim:
Yes sorry by readme I meant changelog. We should also add an entry to the Breaking changes section based on my last comment.

keith:
I updated the sentence there to make it clear what is breaking, in the breaking changes section

jpsim:
Thanks Keith!

keith:
Thanks! <3

PR #3189, opened 2020-04-26
fix unused_import false positive

Agarunov:
Fix unused_import rule false positive in this example:
Configuration:
```yml
unused_import:
    require_explicit_imports: true
    allowed_transitive_imports:
        - module: Foundation
          allowed_transitive_imports:
            - CoreFoundation
```
```swift
import Foundation

typealias Baz = CFData
```

Current master build corrects this code which leads to compilation errors.
```swift

typealias Baz = CFData
```

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/UnusedImportRule.swift
@@ -1,7 +1,7 @@
 import Foundation
 import SourceKittenFramework
 
-public struct UnusedImportRule: CorrectableRule, ConfigurationProviderRule, AnalyzerRule, AutomaticTestableRule {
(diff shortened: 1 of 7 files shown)


Agarunov:
Hm, strange that new non triggering example invoked in tests with default `UnusedImportConfiguration`, but not with provided one

```
Example("""
import Foundation
typealias Foo = CFData
""", configuration: [
    "require_explicit_imports": true,
    "allowed_transitive_imports": [
        [
            "module": "Foundation",
            "allowed_transitive_imports": ["CoreFoundation"]
        ]
    ]
], testOnLinux: false)
```

keith:
I think this case was fixed in https://github.com/realm/SwiftLint/pull/3198, can you verify?

Agarunov:
Yes, this case is fixed. Thanks.

PR #3192, opened 2020-04-27
Add XCTestCase examples to unused declaration rule

keith:
This rule correctly handles subclasses of XCTestCase being used, this adds some tests validating this behavior.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/UnusedDeclarationRuleExamples.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/UnusedDeclarationRuleExamples.swift
@@ -110,7 +110,7 @@ struct UnusedDeclarationRuleExamples {
         class ↓MyTests: NSObject {
             func ↓testExample() {}
         }
-        """),
+        """)
     ] + platformSpecificTriggeringExamples
 
 #if os(macOS)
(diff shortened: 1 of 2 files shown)


jpsim:
I'd like to teach the tests to link XCTest so we can write tests for this and the direct `XCTestCase` subclass behavior.

keith:
Would that mean collecting the inheritance tree across the entire analyze and walking up it?

jpsim:
I had something much less exciting in mind, the ability to specify additional compiler arguments in `Example`s.

keith:
I'm not sure exactly what you mean here, but I'm open to going down that path if you'd like!

keith:
Still relevant

jpsim:
I think the only thing missing here is tests.

keith:
Happy to add some, since it requires custom configuration should I write them in a separate custom test function?

jpsim:
I'm sorry I wasn't communicating clearly what I had in mind. Sometimes the best way to communicate is with code so I pushed up what I meant:

* Teach the test helpers how to find the XCTest framework: aa8b4bf
* Teach the `violations(...)` test helper to apply the configuration from Examples: fb9b00b

keith:
Updated this PR to just add test cases for this behavior since it was fixed with the switch to indexing!

PR #3193, opened 2020-04-29
Fix severity level configuration for duplicate_imports

yusukegoto:
Couldn't overwrite the severity configuration of `duplicate_imports` .
There seems be no other similar case inside the Idiomatic rules at lease I see.


marcelofabri:
Thanks!

PR #3198, opened 2020-05-06
Fix UnusedImportRule breaking transitive imports

keith:
Previously in the case a unused import was allowed to transitively
pollute another module, when that unused import was removed, this rule
would not insert the previously transitively included module, leading
the code to fail to compile. Now unused imports, that will be removed,
are no longer considered for which imports can be implicit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,10 +15,13 @@
 
 #### Enhancements
 
+* Fix UnusedImportRule breaking transitive imports.  
+  [keith](https://github.com/keith)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -15,11 +15,14 @@
 
 #### Enhancements
 
+* Fix UnusedImportRule breaking transitive imports.
```
jpsim, line 18: Move to the bug fixes section?

keith: fixed!

keith:
Thanks!

Issue #2708, opened 2019-04-08
"validates_start_with_lowercase" only applied to variable's name, but not to function's name


ylorn:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

As the title states, the rule option "validates_start_with_lowercase" only applied to the variable's name, but not to the function's name.

### Environment

* SwiftLint version (run `swiftlint version` to be sure)?
0.31.0
* Installation method used (Homebrew, CocoaPods, building from source, etc)?
Homebrew
* Paste your configuration file:

```yml
identifier_name:
  allowed_symbols: 
    - _
  validates_start_with_lowercase: false
  min_length: 3
  max_length:
    warning: 50
    error: 60
  excluded: # excluded via string array
    - id
    - URL
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)?
  No
* Which Xcode version are you using (check `xcode-select -p`)?
  Xcode 10.2
* Do you have a sample that shows the issue? 

```swift
public class Test {
    internal let _version = "1.2.3"
    internal func _start() { }
}
```

output:

> Loading configuration from '.swiftlint.yml'
> Linting Swift files at paths testlint.swift
> Linting 'testlint.swift' (1/1)
> /Users/troyl/SDKLite/airwatchsdk/testlint.swift:3:14: error: Identifier Name Violation: Function name should start with a lowercase character: '_start()' (identifier_name)
> Done linting! Found 1 violation, 1 serious in 1 file.

According to the output, the rule `validates_start_with_lowercase` applies to the variable name `_version`, but not to the function name `_start`. 

I do understand that mark the `_start` function as `private` will resolve this violation. But I have to argue that the behavior of `validates_start_with_lowercase` should be consistent and apply to all.

jpsim:
I agree, I think this is a bug.

luis-:
I believe I have fixed this with #3203 (but its my first PR so not sure how linking PR to issue is done).

ylorn:
Any updates on this?

PR #3203, opened 2020-05-10
Make validates_start_with_lowercase apply to function names (#2708)

luis-:
When `validates_start_with_lowercase` is disabled, functions names will not longer be validated.


ragzy15:
@SwiftLintBot @luis- any idea when this PR will be able to merge?

PR #3223, opened 2020-06-05
Update README files to include complete list of reporters

liamnichols:
👋 

I didn't realise that Realm supported formatting the report to support GitHub Actions and it wasn't until I dug into the source code (and later on when I realised that the docs listed it [here](https://realm.github.io/SwiftLint/Structs/GitHubActionsLoggingReporter.html)) that I realised support was already there.

I thought this because I thought the list included in the README.md was complete but it looks like the English one had been missing `github-actions-logging` and the translations had also been missing a couple of the more recent additions. 

Thought I'd add them to help anybody else in the future 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,7 +376,7 @@ identifier_name:
     - id
     - URL
     - GlobalAPIKey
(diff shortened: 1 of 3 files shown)


marcelofabri:
Thanks!

PR #3058, opened 2020-01-28
Add remote, parent & child configuration features

fredpi:
This PR closes #1352 and #2729.

It's basically the same as #2824, but I squashed all commits to keep it more comprehensible.

## Change Overview

- Using multiple configurations is now documented extensively in the README
- You can now specify `child_config` / `parent_config` configurations. This even works recursively, as long as there are no cycles and no ambiguities.
- Nested configurations got a speed bump (via caching) and of course now ignore configs that were already used to build the primary config (using `child_config` / `parent_config` references)
- You can pass multiple configurations via the command line that are interpreted as a hierarchy
- You can specify a remote `child_config` / `parent_config`! This is just easy as dropping in a url that starts with `http://`or `https://` where normally there would be the path to a local file. Remote configs get cached (in a manner that git doesn't see it). There are timeouts to loading remote configs, that can also be customized using `remote_timeout` / `remote_timeout_if_cached`
- Rule Configurations now get merged properly (Previously, the rule configurations from the parent configuration were always used). This change results in the following behavior:
   - When a child config doesn't configure a rule, yet the parent config does, the parent config will be used
    - When a child config configures a rule, and the parent doesn't, the child config will be used
    - When a child config configures a rule, and the parent does, too, the child config will be used
- Warnings about duplicate and invalid identifiers now appear as actual warnings in Xcode
- Internally, the whole `Configuration` clutter has been cleaned up in a way that I hope everyone is happy with. I strongly felt like the current merging implementation that relied on a `rules` array and a `rulesMode` didn't express the very foundation of what a configuration is: A mode that is applied to an array of all available rules with configurations, resulting in an array of rules that should be linted. To express this very relation, I created the new `RulesWrapper` type. While offering a `resultingRules` variable that is lazily computed, it still retains what caused these rules and therefore eases understanding of the merging algorithm.

#### Sidenote: Fixed Warnings in Moya

SwiftLintBot will probably complain that, with this new implementation, 3 warnings in Moya are now fixed. In fact, previously those have been false positives, as in the nested `.swiftlint.yml` covering these files there's the following configuration for the `file_length` rule:

```
file_length:
  warning: 1000
```

As the parent rule configuration was used for merging and there was no smart mechanism to decide which configuration to choose, the configuration originating from the nested configuration was ignored. With this PR, that's now fixed.

PR #3231, opened 2020-06-10
Fix cache directory resolution when using remote config file

mokagio:
I'm spiking using a [centralized SwiftLint configuration](https://github.com/Automattic/swiftlint-config) across the Automattic's Swift repositories, as we have quite a few.

I used https://github.com/realm/SwiftLint/pull/3058 but it wasn't working. After a bit of digging around, I discovered the issue had to do with how the cache directory path was build, or to be precise resolved as an absolute path.

You can see the SwiftLint build failing [here](https://github.com/wordpress-mobile/WordPress-iOS-Shared/runs/756641590?check_suite_focus=true#step:7:7) (although the message is a bit misleading) and succeeding [here](https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/262/checks?check_run_id=756740142), with [the only change](https://github.com/wordpress-mobile/WordPress-iOS-Shared/compare/39b9e33...3c8dcc8) being that the former uses https://github.com/realm/SwiftLint/pull/3058 as the source for SwiftLint and the latter this PR.

I would have liked to add tests for this fix, but the only way I could see would have been to add an abstraction on `FileManager` in order to provide a test double for it in the test. That seemed like a lot of work as well as a substantial change in the design of this code, which is not on `master` yet. I opted for simply fixing the production code, instead. If you can think of a quick way to add tests which would fit in the context of this PR, I'd love to make that happen.

Thank you for working on this @fredpi. It's super useful, I really love this feature.

### DIFF
--- a/Source/SwiftLintFramework/Extensions/Configuration+Remote.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+Remote.swift
@@ -203,8 +203,8 @@ internal extension Configuration.FileGraph.FilePath {
         // Create directory if needed
         let directory = "\(Configuration.FileGraph.FilePath.remoteCachePath)/"
             + "\(Configuration.FileGraph.FilePath.remoteCacheVersionNumber)/"
-            .bridge().absolutePathRepresentation(rootDirectory: rootDirectory)
-        if !FileManager.default.fileExists(atPath: directory) {
+        let absoluteDirectory = directory.bridge().absolutePathRepresentation(rootDirectory: rootDirectory)
+        if !FileManager.default.fileExists(atPath: absoluteDirectory) {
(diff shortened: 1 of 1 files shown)


fredpi:
@mokagio Thanks for trying out the PR and fixing this! 🚀

Apart from the little restructuring you did, the only change seems to be the removal of `/` before building the path. I'm just wondering why it didn't fail when I tested it – maybe because the path was already there from a previous version? 🤔

Thanks again for fixing it – looks good to me and I'm going to merge it into my branch. If you want, you can create another PR, adding your name to the changelog section in my branch. 👍

mokagio:
Thank you for the quick review 🙌 

> maybe because the path was already there from a previous version? 🤔
I'd say that's likely. I experienced different error messages depending on whether the `.swiftlint` folder was there or not.

PR #3243, opened 2020-06-16
Add `inclusive_language` rule

daltonclaybrook:
Current events have renewed the conversation in our community about the roles of terminology with racist connotations in our software. Many companies and developers are now taking appropriate steps to remove this terminology from their codebases and products. (e.g. [GitHub](https://twitter.com/natfriedman/status/1271253144442253312)) This small rule prevents the use of declarations that contain any of the terms: whitelist, blacklist, master, and slave. It may be appropriate to add more terms to this list now or in the future. Of course, this can be a divisive issue for some, so I wanted to open this PR to start the conversation about whether a rule like this is appropriate for SwiftLint. Any feedback is welcome and greatly appreciated.


72A12F4E:
I think that this should be generalized to a disallowed terms rule, where SwiftLint would discourage using some words by default but also support configuring your own words/regex matches.

Source/SwiftLintFramework/Rules/Style/RacistDeclarationRule.swift
```diff
@@ -0,0 +1,72 @@
...
+            }
+            """)
+        ]
+    )
+
+    private let racistTerms: Set<String> = [
+        "whitelist",
+        "blacklist",
+        "master",
+        "slave"
```
superarts, line 49: Are you sure these are the terms you want to discourage? Are you OK to use the words listed [here](https://en.wikipedia.org/wiki/List_of_ethnic_slurs) in your source code, including the N-word?

superarts:
> I think that this should be generalized to a disallowed terms rule, where SwiftLint would discourage using some words by default but also support configuring your own words/regex matches.

I don't think SwiftLint should include any pre-defined words at all. The list should be configurable on project level; `.swiftlint.yml` would be a good place.

Source/SwiftLintFramework/Rules/Style/RacistDeclarationRule.swift
thuss: The point is to stop using words that up until now have been commonly used in source code. To provide a gentle nudge or reminder to developers that accidentally use the old terminology to use the new terminology.

superarts: Could you also try to justify your design with a hard-coded array in Swift code?

thuss: This isn't my design but the reason a hardcoded array makes sense for this case is that every tech company in the US is currently going through their source code to eliminate exactly those 4 words:

https://developer-tech.com/news/2020/jun/15/github-replace-slavery-terms-master-whitelist/
https://tools.ietf.org/id/draft-knodel-terminology-00.html#rfc.section.1.1

While making the words per-project configurable in swiftlint.yml would technically allow every company to solve the problem, every company is removing the same exact language right now, so having those 4 words seems like a very sensible default to me. I could see a case for allowing users to allow additional words via swiftlint.yml but I don't see any reason to make every project independently specify the same 4 words.

72A12F4E: I think there is room for both here. The term `racist_declaration` is combative where something like `disallowed_terminology` would be more likely to gain support from the individuals who don't have an opinion either way on the usage of the words outlined here. I do agree that the terms on this list cause conflict, and do have a place here as a set of default terms.

However, expanding this rule to `disallowed_terminologoy` would also allow more flexibility for projects to adopt their own particular rulesets. For a particular library use case, I might want to avoid certain terms or phrases in the source code that reference a particular company/technology stack/misnomer/etc. 

Expanding the scope of this rule enables that use case, which might help with the adoption.

thuss: Totally agree. `disallowed_terminology` reads much better. Then maybe the default list would be those 4 words as well as an option to expand upon that list in their own projects?

jpsim:
Thanks for contributing this rule. Our societies need **real change** to address systemic inequalities but symbolic changes also have a cumulative effect and shouldn't be ignored either.

I'd love to see this extend to promoting _inclusive language_ beyond race. The rule could in fact be called `inclusive_language`, be on by default and enable _extending and overwriting_ the default word set. Of course like all SwiftLint rules this would be able to be disabled locally or globally.

daltonclaybrook:
@jpsim Updated based on your feedback. Obviously turning this rule on by default resulted in a number of changes to the project as a whole, specifically around the use of "whitelist." Let me know if you have any more feedback.

mbartlett505:
Sir this is Michael, I have no idea what you are talking about . usually something like this would bother me. I’m not even gonna look in to it.  I have bigger problems with my morals. I haven’t the room to add racism also. My apologies sir. Please Reply with everything thing your company offers and I’ll remove myself from all projects. Goodnight sir

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

daltonclaybrook:
Is there anything stopping this PR from moving forward? I should have availability this week to make any necessary modifications.

Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
```diff
@@ -10,7 +10,7 @@ extension Configuration {
         case swiftlintVersion = "swiftlint_version"
         case useNestedConfigs = "use_nested_configs" // deprecated
         case warningThreshold = "warning_threshold"
-        case whitelistRules = "whitelist_rules"
+        case allowlistRules = "allowlist_rules"
```
marcelofabri, line 13: can we keep parsing `whitelist_rules` and print a warning saying it's deprecated? I don't think it's fair to break the setup for everyone using it (https://github.com/realm/SwiftLint/issues/3274)

Source/SwiftLintFramework/Rules/RuleConfigurations/InclusiveLanguageConfiguration.swift
```diff
@@ -0,0 +1,49 @@
...
+            throw ConfigurationError.unknownConfiguration
+        }
+
+        if let severityString = configuration[ConfigurationKey.severity.rawValue] {
+            try severityConfiguration.apply(configuration: severityString)
+        }
+
+        let configAllowList = lowercasedList(for: .allow, from: configuration)
+        let configDenyList = lowercasedList(for: .deny, from: configuration)
+        self.denyList = defaultDenyList.union(configDenyList).subtracting(configAllowList)
```
marcelofabri, line 42: is the idea that people should use `deny` to add more cases and `allow` to remove from the default ones?

I wonder if we should make the configuration like `additional_terms` and `terms` where you could use the latter to ignore the default list and setup yours completely

Source/SwiftLintFramework/Rules/RuleConfigurations/InclusiveLanguageConfiguration.swift
```diff
@@ -0,0 +1,49 @@
...
+    case deny
+}
+
+public struct InclusiveLanguageConfiguration: RuleConfiguration, Equatable {
+    public var severityConfiguration = SeverityConfiguration(.warning)
+    public var denyList: Set<String>
+
+    public var consoleDescription: String {
+        severityConfiguration.consoleDescription
+            + ", deny: \(denyList.sorted())"
```
marcelofabri, line 13: i think it's worth keeping the values from configuration in memory so we can print here the exact configuration from `.swiftlint.yml` instead of the "resolved" one

Source/SwiftLintFramework/Rules/Style/AttributesRule.swift
```diff
@@ -292,7 +292,7 @@ public struct AttributesRule: ASTRule, OptInRule, ConfigurationProviderRule {
 
     private func parseAttributes(dictionary: SourceKittenDictionary) -> [SwiftDeclarationAttributeKind] {
         let attributes = dictionary.enclosedSwiftAttributes
-        let blacklist: Set<SwiftDeclarationAttributeKind> = [
+        let denylist: Set<SwiftDeclarationAttributeKind> = [
```
marcelofabri, line 295: I updated this in another PR 😅

marcelofabri:
Awesome work as usual @daltonclaybrook ✨👏

Source/SwiftLintFramework/Rules/RuleConfigurations/InclusiveLanguageConfiguration.swift
daltonclaybrook: Yah, that's the idea. I'm a bit worried that `terms` could be more ambiguous than `allow`/`deny`. `allow` means you're allowed to use the term in your project, and `deny` means you're not allowed. So, if the default list contains terms you want to use, you'd need to add them to `allow`. In your example, would the mere presence of the `terms` key cause the default list to be bypassed? I think we'd want to make sure this was super clear to users. I think I could be convinced either way.

Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
marcelofabri: Also this should be mentioned in the CHANGELOG

Source/SwiftLintFramework/Rules/RuleConfigurations/InclusiveLanguageConfiguration.swift
marcelofabri: Maybe `override_terms`? 

> In your example, would the mere presence of the terms key cause the default list to be bypassed? 

Yeah, that's what I was thinking

daltonclaybrook:
@marcelofabri Do you know if this job is just hung, or is there a real issue here?

marcelofabri:
> @marcelofabri Do you know if this job is just hung, or is there a real issue here?

I think it's just hanging. Can you please push another commit just to retrigger CI? 🙏

daltonclaybrook:
@marcelofabri Any updates on this? It seemed like there was some traction for a while...

jpsim:
Thanks for pushing this through, your patience and persistence.

I rebased this and applied a few small edits. I only made one bigger change:

I wasn't totally thrilled with the "allowlist" terminology. Not only is "allowlist" not a word, but it's also not very clear what it does. It implies the existence of rules that's aren't allowed somehow.
    
So I changed it to "Only rules" makes it clearer than "allowlist" or even "whitelist" that these are the only rules that will be applied.

When CI is happy with this, I'll merge and cut a release at some point between tonight and the next few days.

DavidJRobertson:
You shouldn't impose your politics on everyone by making this on by default. It's rude.

Omara8:
this is what happens when you have nothing better to do with your life
sorry, not sorry.

borisyurkevich:
I have some payments logic which produced the new warning now.

```swift
let masterCard = ["5268 6133 8448 6284"] // ⚠️ Inclusive Language Violation: Declaration masterCard contains the term "master" which is not considered inclusive. (inclusive_language)
```

What's recommended way for fixing this?

DavidJRobertson:
@borisyurkevich  I went with turning off the `inclusive_language` rule entirely.

jpsim:
@borisyurkevich We've opted to explicitly allow the "mastercard" term in https://github.com/realm/SwiftLint/pull/3439, which will be included in the next release.

hajjD:
Yet how many jokers who support this have a MASTERS degree? Imagine having so little actual problems in your life that you have to create problems and make fake solutions like this as if it will in anyway shape or form end racism or solve actual issues.

And I say this as a person of color. This should never have been enabled by default, should only be used by companies that pretend to care (but not actually do anything)

borisyurkevich:
@hajjD I understand that you perceive issues like this as unimportant or even distracting, and I hope this change doesn't affect your productivity too much. As you can see by the reactions, this is not a unanimous decision, some people would agree with you. However, there are people for whom details like this are important. The whole idea of Swift Lint is taking care of something which many of us consider a minor detail, like a code style. I mean really, developers love discussions about use spaces versus tabs or where to put brackets, but it doesn't affect user experience. Yet we use Swift Lint and it helps reduce the amount of debates we have during PR review, because it just dictates one way of writing code. Language matters, it forms our thinking.

> In a [2011 study](https://www.researchgate.net/publication/257663669_The_Gendering_of_Language_A_Comparison_of_Gender_Equality_in_Countries_with_Gendered_Natural_Gender_and_Genderless_Languages), Jennifer Prewitt-Freilino, associate professor at the Rhode Island school of design, along with colleagues, found that countries with gendered languages experienced higher average gender inequality.

Source, [BBC.](https://www.bbc.com/culture/article/20201006-are-some-languages-more-sexist-than-others)

There are many things in Western culture you can categorise as "first world problems", I mean. there's even a phrase for this. But that's what makes it interesting. The English language constantly evolves, it gains new words, and sometimes those words are coming even from sitcoms TV shows, like Seinfeld. Old words are going out of fashion. It's just a pop-culture thing, you can't deny it.

When StarCraft 1 became popular in my school, I remember having a discussion with my friend about using the word "slave" when referencing workers, I mean, does it matter how we call a class of characters in the video-game? It mattered to me. There are things in life which I ignore because I find them distracting, like Facebook, or Instagram. But it doesn't mean we should cancel them. Similarly, some people would not care about the choice of fonts, for me, it's very important, but for others not. This doesn't mean we should ignore the choice of fonts though. The fact it doesn't matter to you doesn't mean we should avoid changes or work like this one.

I appreciate you finding the time and expressing your opinion, and it is a good argument about a "Masters degree" in education circles, there's [process](https://www.bbc.com/news/education-35659685) about it as well.

PR #3248, opened 2020-06-22
Add NonPrivateXCTestMembersRule

keith:
This rule lints against any members of XCTestCase subclasses that aren't
implementing XCTestCase methods, or prefixed with `test`. This can help
encourage reducing ACLs on these classes, and potentially find
mis-spelled test functions.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/NonPrivateXCTestMembersRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/NonPrivateXCTestMembersRule.swift
@@ -1,6 +1,8 @@
+import Foundation
 import SourceKittenFramework
 
-public struct NonPrivateXCTestMembersRule: Rule, OptInRule, ConfigurationProviderRule, AutomaticTestableRule {
+public struct NonPrivateXCTestMembersRule: Rule, OptInRule, ConfigurationProviderRule, AutomaticTestableRule,
+                                           SubstitutionCorrectableRule {
     public var configuration = SeverityConfiguration(.warning)
 
     public init() {}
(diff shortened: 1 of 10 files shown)


keith:
Open questions:

- [ ] Should we provide a way to identify custom test subclasses so this validation handles those as well (potentially similar to https://github.com/realm/SwiftLint/pull/3192)
- [ ] Should we explicitly ignore `override`s? Probably but probably only matters if we implement the above
- [ ] Instead of linting against ACL > private / fileprivate maybe instead we should only lint against not explicitly defined ACLs, since if you mark it as public and that's not validated somewhere else maybe that will always be by design (potentially this difference in behavior could be behind a setting)

keith:
Still relevant

jpsim:
Are you still working through the open questions? Would you like my take on them?

keith:
I'd love your take!

Source/SwiftLintFramework/Rules/Lint/NonPrivateXCTestMembersRule.swift
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    // MARK: - SubstitutionCorrectableRule
+
+    public func violationRanges(in file: SwiftLintFile) -> [NSRange] {
+        return testClasses(in: file).flatMap { violationRanges(in: file, for: $0) }
+    }
+
+    public func substitution(for violationRange: NSRange, in file: SwiftLintFile) -> (NSRange, String)? {
+        return (violationRange, "private ")
```
jpsim, line 35: This has the potential of making code that previously compiled stop compiling in several cases:

1. If there was an explicit ACL keyword used
2. If the type was used either outside the file (internal) or outside the type scope (fileprivate)

I'm comfortable with those tradeoffs but let me know if you're not.

jpsim:
> Should we provide a way to identify custom test subclasses so this validation handles those as well

I'm comfortable keeping things simple to start and saying no to this one. This is an opt-in rule and we can always add support for this if there's an appetite.

> Should we explicitly ignore overrides? Probably but probably only matters if we implement the above

Again, I'm ok keeping things simple and not handling this right now.

> Instead of linting against ACL > private / fileprivate maybe instead we should only lint against not explicitly defined ACLs, since if you mark it as public and that's not validated somewhere else maybe that will always be by design (potentially this difference in behavior could be behind a setting)

I think you see where I'm going with this. Let's start with keeping things simple.

In most cases it's probably better to use composition for test helpers rather than subclassing anyway IMHO.

jpsim (approved):
👍 This looks good to me. @keith I made some small changes (made the rule correctable, turned it on in SwiftLint) so I'll let you merge this if/when you're comfortable with its state.

PR #3249, opened 2020-06-24
Add SignificantDeclarationWhitespaceRule

igor-makarov:


igor-makarov:
@jpsim hi! First time PR I think for me.

What do I do about the bot adding 3000 warnings? The rule is opt-in, so just ignore?

Also, I'd love to have some comments about naming it.

The idea is that if your class/extension/func is of 'significant' size (default 3 lines), it needs to be preceded by a newline, like so:
```
class A {

}

class B {
   var a = 1
   var b = 1
}
```

PR #3250, opened 2020-06-25
Use Ray Wenderlich’s Swift Style guide

### LINKED ISSUE #3200: https://github.com/github/swift-style-guide
**Problem:** https://github.com/github/swift-style-guide has been archived and not active anymore.

**Question:** What swift style guide reference does SwiftLint do?

christiankm:
GitHub has archived their current style guide.
The style guide from Wenderlich closely follows the existing rules, and is still being maintained.
Closes #3200

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,6 @@
 # SwiftLint
 
-A tool to enforce Swift style and conventions, loosely based on
-[GitHub's Swift Style Guide](https://github.com/github/swift-style-guide).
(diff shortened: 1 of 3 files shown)


christiankm:
@marcelofabri Should be ready to merge.

christiankm:
@jpsim this should be quick to review and close, as well as the Issue it resolves.

jpsim:
I'm not opposed to pointing to Ray Wenderlich's style guide, but the wording here is wrong. SwiftLint was _not_ based on Ray Wenderlich's style guide, it was based on GitHub's.

Can you update the wording to say something like "SwiftLint enforces the Swift style guide that has emerged as unofficial community standard, and reflective of the spirit of popular style guides like Ray Wenderlich's and GitHub's guides". That's overly verbose and needs some copy editing, which is why I'm not editing your PR directly.

christiankm:
@jpsim I have updated my PR according to your comments. Feel free to change it however you like and merge, or just decline. It's not too important of a change.

jpsim:
Thanks for the PR! I agree it's good to link to an actively maintained style guide.

PR #3254, opened 2020-06-30
Add 'always_keep_imports' to UnusedImportRule

keith:
Sometimes we hit issues with SourceKit that cause this rule to produce
false positives of some imports. While we would prefer to fix all of
these cases whenever possible, this gives us a temporary escape hatch to
tell the rule never to remove imports of specific modules until the
fixes land in SwiftLint or Swift itself.

Currently this is for us to opt out of https://bugs.swift.org/browse/SR-13120

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@
   found issues.  
   [krin-san](https://github.com/krin-san)
   [#3177](https://github.com/realm/SwiftLint/pull/3177)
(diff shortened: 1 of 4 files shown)


keith:
Still relevant

jpsim (approved):
Awesome, I hadn't seen this one. Thanks!

keith:
Thanks!

PR #3271, opened 2020-08-04
Run tests on Xcode 12 as well

marcelofabri:


Tests/SwiftLintFrameworkTests/IndentationWidthRuleTests.swift
```diff
@@ -175,7 +175,7 @@ class IndentationWidthRuleTests: XCTestCase {
         if let includeComments = includeComments { configDict["include_comments"] = includeComments }
 
         guard let config = makeConfig(configDict, IndentationWidthRule.description.identifier) else {
-            XCTFail("Unable to create rule configuration.", file: file, line: line)
+            XCTFail("Unable to create rule configuration.", file: (file), line: line)
```
marcelofabri, line 178: this is super annoying due to `#filePath` vs `#path`

PR #3279, opened 2020-08-05
Fix false positives in closure_spacing

### LINKED ISSUE #3270: closure_spacing rule also warns about constructs that are not closures (guard statements, protocol fields)
### New Issue Checklist

### Describe the bug

The `closure_spacing` rule also warns about braces that aren't closures, such as:
```
protocol SomeProtocol {
   var field: Int {get} // Closure Spacing Violation: Closure expressions should have a single space inside each brace. (closure_spacing)
}

func something(field: Int) {
   guard field != 0 else {return} // Closure Spacing Violation: Closure expressions should have a single space inside each brace. (closure_spacing)
}
```

### Environment

* SwiftLint version: 0.39.2
* Installation method used: download from releases

```yml
# See:
# https://realm.github.io/SwiftLint/rule-directory.html

disabled_rules:
 - line_length # we don't limit
 - function_body_length # we don't limit
 - cyclomatic_complexity # broken and we don't limit
 - force_cast # we don't consider this a problem
 - file_length # we don't limit
 - type_body_length # we don't limit
 - multiline_arguments_brackets # we don't like this
 - no_fallthrough_only # we use fallthrough on some places
 - no_grouping_extension # we're fine with this
 - compiler_protocol_init # too many false positives on NSNumber and friends
 - unused_closure_parameter # we like this for clarity
 - sorted_imports # we have our own sorting that's incompatible with swiftlint
 - file_types_order # we don't like this ordering
 - multiline_function_chains # too inflexible for us
 - implicit_return # simple and complex closures are handled the same unfortunately

opt_in_rules:
 - closure_spacing
 - file_types_order
 - fatal_error_message
 - let_var_whitespace
 - modifier_order
 - multiline_arguments
 - multiline_function_chains
 - multiline_parameters
 - no_space_in_method_call
 - operator_usage_whitespace
 - overridden_super_call
 - private_action
 - prohibited_super_call
 - protocol_property_accessors_order
 - redundant_type_annotation
 - sorted_first_last
 - sorted_imports
 - toggle_bool
 - unavailable_function

#implicit_return:
# included: [closure]

trailing_comma:
 mandatory_comma: true

switch_case_alignment:
 indented_cases: true

identifier_name:
 min_length:
  warning: 2
  error: 2
 max_length:
  warning: 60
  error: 100

type_name:
 max_length:
  warning: 60
  error: 100

nesting:
 type_level: 4

trailing_whitespace:
 ignores_empty_lines: true

excluded:
 - External

```

No
Xcode 11.3.1
Build version 11C504

`echo "protocol SomeProtocol {\n\tpublic var field: Int {get}\n}\n" | swiftlint lint --no-cache --use-stdin --enable-all-rules`

`echo "public func something(field: Int) {\n\tguard field != 0 else {return}\n\tprint(field)\n}\n" | swiftlint lint --no-cache --use-stdin --enable-all-rules`

marcelofabri:
Fixes #3270

I'm curious to see oss-check's results for this one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#3186](https://github.com/realm/SwiftLint/issues/3186)
 
+* Fix false positives in `closure_spacing` rule reported in 
+  non-closures expressions.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#3270](https://github.com/realm/SwiftLint/issues/3270)
+
 ## 0.39.2: Stay Home
 
 This is the last release to support building with Swift 5.0.x.
--- a/Source/SwiftLintFramework/Rules/Style/ClosureSpacingRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/ClosureSpacingRule.swift
@@ -30,7 +30,17 @@ public struct ClosureSpacingRule: CorrectableRule, ConfigurationProviderRule, Op
             Example("[].map ({ $0.description })"),
             Example("[].filter { $0.contains(location) }"),
             Example("extension UITableViewCell: ReusableView { }"),
-            Example("extension UITableViewCell: ReusableView {}")
+            Example("extension UITableViewCell: ReusableView {}"),
+            Example("""
+            protocol SomeProtocol {
+               var field: Int {get}
+            }
+            """),
+            Example("""
+            func something(field: Int) {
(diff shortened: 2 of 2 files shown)


marcelofabri:
I have fixed feelings about this: these violations are not really closures, but people might be relying on the rule to report them, since I'm not sure if other rules would catch them.

jpsim:
Should we just rename the rule to `brace_spacing`?

PR #3290, opened 2020-08-08
Improve documentation around Xcode build phases

### LINKED ISSUE #3169: Unclear “Xcode scheme”/“Run script phase” Readme Instructions
### New Issue Checklist

### Describe the bug

The current [README.md](https://github.com/realm/SwiftLint/blob/master/README.md)'s setup instructions instruct the user to _”Integrate SwiftLint into an Xcode scheme to get warnings and errors displayed in the IDE. Just add a new "Run Script Phase" with:”_.  I'm not sure what I'm supposed to do here.

The **Scheme** editor in Xcode doesn't have an option for Run Script Phase, only **New Run Script Action**.  Also, the instruction don't specify whether this should be a Pre-action or a Post-action.

Screen Shot 2020-04-03 at 8 10 57 PM
 

Alternatively, the instructions may have meant  _“Integrate SwiftLint into an Xcode **target**”_, in which case the instructions should be updated to reflect that.  (Targets and Schemes are distinctly different concepts in Xcode.)

Screen Shot 2020-04-03 at 8 03 55 PM

I'm going to assume the latter for now and hope it works, simply based on the fact that Target scripts tend to be a more commonplace solution than Phase scripts.  However, I'd like some term-fixing for the sake of clarity in the readme— it should be either:

1. _”Integrate SwiftLint into an Xcode **target** to get warnings and errors displayed in the IDE. Just add a new "Run Script Phase" with:”_, or
2. _”Integrate SwiftLint into an Xcode scheme to get warnings and errors displayed in the IDE. Just add a new **Build Pre-action "Run Script Action"** with:”_

Not relevant— this is an issue about the Markdown-formatted documentation.

### Environment

Not relevant— this is an issue about the Markdown-formatted documentation.  Environment is the website at https://github.com/realm/SwiftLint/ (technically, https://github.com/realm/SwiftLint/blob/master/README.md).

christiankm:
Closes #3169

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,8 +86,10 @@ we encourage you to watch this presentation or read the transcript:
 
 ### Xcode
 
-Integrate SwiftLint into an Xcode scheme to get warnings and errors displayed
-in the IDE. Just add a new "Run Script Phase" with:
+Integrate SwiftLint into your Xcode project to get warnings and errors displayed
+in the issue navigator. 
+
+To do this click the Project in the file navigator, then click the primary app target, and go to Build Phases. Click the + and select "New Run Script Phase". Insert the following as the script:
 
 ```bash
 if which swiftlint >/dev/null; then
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -99,7 +101,11 @@ fi
 
 ![](assets/runscript.png)
 
-Alternatively, if you've installed SwiftLint via CocoaPods the script should look like this:
+You might want to move your SwiftLint phase directly before 'Compile Sources' step, to detect errors quickly before compiling.
```
marcelofabri, line 104: i'm not sure we should suggest this since SwiftLint usually assumes that you have valid Swift code, which means that if for some reason you have syntax errors, you might get non-sense warnings or, in the worst case, even crashes when running `swiftlint`

christiankm: Okay, I've never been aware of any best-practice in this area. We've placed it both before the compile step, and as the last step with equal results.

Would you prefer to have this paragraph but with a warning, or just remove it?

jpsim:
Thank you! I made some edits and pushed directly to master in 67165f7c

PR #3313, opened 2020-08-20
Add support for params files for file path arguments

keith:
This allows you to pass `@path/to/file` which will be read line by line
for the list of files to lint / analyze. This is useful if you want to
pass a massive list of files which will either exceed arg max, or
break most unix tools if you use script input files + env vars.

### DIFF
--- a/Source/swiftlint/Helpers/LintableFilesVisitor.swift
+++ b/Source/swiftlint/Helpers/LintableFilesVisitor.swift
@@ -28,6 +28,16 @@ enum LintOrAnalyzeModeWithCompilerArguments {
     case analyze(allCompilerInvocations: CompilerInvocations)
 }
 
+private func resolveParamsFiles(args: [String]) -> [String] {
+    return args.reduce(into: []) { allArgs, arg in
+        if arg.hasPrefix("@"), let contents = try? String(contentsOfFile: String(arg.dropFirst())) {
+            allArgs += resolveParamsFiles(args: contents.split(separator: "\n").map(String.init))
+        } else if !arg.isEmpty {
+            allArgs.append(arg)
+        }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


keith:
I didn't add tests here because this is part of the main target which doesn't have a corresponding tests target

Source/swiftlint/Helpers/LintableFilesVisitor.swift
```diff
@@ -28,6 +28,16 @@ enum LintOrAnalyzeModeWithCompilerArguments {
     case analyze(allCompilerInvocations: CompilerInvocations)
 }
 
+private func resolveParamsFiles(args: [String]) -> [String] {
+    return args.reduce(into: []) { allArgs, arg in
+        if arg.hasPrefix("@"), let contents = try? String(contentsOfFile: String(arg.dropFirst())) {
+            allArgs += resolveParamsFiles(args: contents.split(separator: "\n").map(String.init))
```
keith, line 34: Probably up to us whether or not this should actually handle these recursively. I believe compilers do.

Source/swiftlint/Helpers/LintableFilesVisitor.swift
```diff
@@ -28,6 +28,16 @@ enum LintOrAnalyzeModeWithCompilerArguments {
     case analyze(allCompilerInvocations: CompilerInvocations)
 }
 
+private func resolveParamsFiles(args: [String]) -> [String] {
```
keith, line 31: I named this more generically since theoretically this could be used for things besides paths

keith:
I'm happy to rename these if you want, I've seen both, bazel uses params file which is where I've seen it recently. Also if that logic was public I guess we could use that, but it makes sense that it's not

PR #3325, opened 2020-08-28
Improving excluding paths algorithm

### LINKED ISSUE #3304: Excluding takes too long
I use Swiftlint for one very big project for only changed files by specyifing `--use-script-input-files --force-exclude` and after some checks and measurements with Time Profiler I found out that excluding paths takes more then 3-4 seconds on my machine.

I investigated that problem and found next: while excluding it finds ALL lintable files in all excluded paths from config. And then the result set of excluded swift files is subtracted from the input paths set.

I find it quiet strange algorithm, because why do we need to collect all lintable files for example in excluded Pods directory. 

So, I tryed to changed it to the next exclusion algorithm:

```swift
let allPaths = paths.flatMap { $0 }
let excludedPaths = excluded.flatMap(Glob.resolveGlob).map { $0.absolutePathStandardized() }
return allPaths.filter { path in
    !excludedPaths.contains { path.hasPrefix($0) }
}
```

And it still excludes files, lints and works for those 3-4 seconds faster 

Are there any special reasons for current exclusion algorithm? or it can be changed to something like I used?

JohnReeze:
Resolves #3304

As I described in #3304 current excluding process is quiet inefficient. I changed it to more strait algorithm and got better results. 

In my case hardware: MacBook Pro 6-Core Intel Core i7, 16gb 
And for my quite big project for current algorithm which collects all lintable files in excluded directories (Pods, third party, etc)  the excluding method (`filterExcludedPaths`) takes  about 4.5-5.5 seconds, and for new algorithm for same conditions it takes about 1-1.6 seconds to do the same thing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@
 
 #### Enhancements
 
-* None.
+* Speed up excluding paths process.  
+  [JohnReeze](https://github.com/JohnReeze)
+  [#3304](https://github.com/realm/SwiftLint/issues/3304)
 
 #### Bug Fixes
 
--- a/Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
@@ -31,41 +31,20 @@ extension Configuration {
         let includedPaths = included.parallelFlatMap {
             fileManager.filesToLint(inPath: $0, rootDirectory: self.rootPath)
         }
-        return filterExcludedPaths(fileManager: fileManager, in: pathsForPath, includedPaths)
+        return filterExcludedPaths(in: pathsForPath, includedPaths)
     }
 
     /// Returns an array of file paths after removing the excluded paths as defined by this configuration.
     ///
-    /// - parameter fileManager: The lintable file manager to use to expand the excluded paths into all matching paths.
     /// - parameter paths:       The input paths to filter.
     ///
     /// - returns: The input paths after removing the excluded paths.
-    public func filterExcludedPaths(fileManager: LintableFileManager = FileManager.default,
-                                    in paths: [String]...) -> [String] {
+    public func filterExcludedPaths(in paths: [String]...) -> [String] {
         let allPaths = paths.flatMap { $0 }
-#if os(Linux)
-        let result = NSMutableOrderedSet(capacity: allPaths.count)
-        result.addObjects(from: allPaths)
-#else
-        let result = NSMutableOrderedSet(array: allPaths)
-#endif
-        let excludedPaths = self.excludedPaths(fileManager: fileManager)
-        result.minusSet(Set(excludedPaths))
-        // swiftlint:disable:next force_cast
-        return result.map { $0 as! String }
-    }
+        let excludedPaths = excluded.flatMap(Glob.resolveGlob).map { $0.absolutePathStandardized() }
 
-    /// Returns the file paths that are excluded by this configuration after expanding them using the specified file
-    /// manager.
-    ///
-    /// - parameter fileManager: The file manager to get child paths in a given parent location.
-    ///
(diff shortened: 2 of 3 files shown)


jpsim:
Thanks for the PR! Unfortunately this has worse performance in the projects I tested.

For a project with 14,382 lintable Swift files, this PR is 30% slower than `master`:

```
$ hyperfine ./swiftlint-master ./swiftlint-pr-3325
Benchmark #1: ./swiftlint-master
  Time (mean ± σ):      5.404 s ±  0.302 s    [User: 3.592 s, System: 1.961 s]
  Range (min … max):    5.036 s …  6.007 s    10 runs
 
Benchmark #2: ./swiftlint-pr-3325
  Time (mean ± σ):      7.052 s ±  0.312 s    [User: 5.225 s, System: 1.976 s]
  Range (min … max):    6.646 s …  7.569 s    10 runs
 
Summary
  './swiftlint-master' ran
    1.30 ± 0.09 times faster than './swiftlint-pr-3325'
```

I compiled SwiftLint in release mode (`swift build -c release`) and applied the following patch to just measure lintable files calculation time without measuring the actual linting:

```diff
--- a/Source/swiftlint/Extensions/Configuration+CommandLine.swift
+++ b/Source/swiftlint/Extensions/Configuration+CommandLine.swift
@@ -51,10 +51,11 @@ extension Configuration {
         -> Result<[SwiftLintFile], CommandantError<()>> {
             return getFiles(with: visitor)
                 .flatMap { groupFiles($0, visitor: visitor) }
-                .map { linters(for: $0, visitor: visitor) }
-                .map { ($0, $0.duplicateFileNames) }
-                .map { collect(linters: $0.0, visitor: visitor, storage: storage, duplicateFileNames: $0.1) }
-                .map { visit(linters: $0.0, visitor: visitor, storage: storage, duplicateFileNames: $0.1) }
+                .map { groupedFiles in
+                    let files = groupedFiles.flatMap({ $1 })
+                    queuedPrint(files.compactMap({ $0.path }).joined(separator: "\n"))
+                    return files
+                }
     }
 
     private func groupFiles(_ files: [SwiftLintFile],
```

JohnReeze:
@jpsim thanks for your response!

And are there any big excluding directories in these tests?) 

The problem I am trying to solve is that for running swiftlint for only changed files, current algorithm
it collects about 4,000 lintable files in excluded directories which takes about 3-5 seconds for my machine,  just in order to subtract it from set of 10 changed files. But without collecting, by filtering path prefix it works fast as it should)
I still think that performance in these cases more about the size of excluded directories

Maybe we can add something like excluding strategy which can be configured in .swiftlint.yml to provide more flexibility for described cases?

jpsim:
> And are there any big excluding directories in these tests?)

The shape of the included/excluded values in the configuration file look like this:

```yaml
included: [dir0, dir1]
excluded:
  - dir1/*/.ext1
  - dir1/dir2/dir3/dir4
  - dir1/dir5
```

`dir1` has around a thousand files, so it's not super big:

```
$ tree -a dir1 | wc -l
1249
```

---

> Maybe we can add something like excluding strategy which can be configured in .swiftlint.yml to provide more flexibility for described cases?

I'm definitely open to that but I'd rather find a way to automatically use the fastest way, maybe there's a heuristic we can use to know what will be fastest?

JohnReeze:
> I'm definitely open to that but I'd rather find a way to automatically use the fastest way, maybe there's a heuristic we can use to know what will be fastest?

Well, I think is quite hard to find that moment automatically when configuration of included/excluded paths or number of files in these directories or the complexity of excluded directories hierarchy  make collecting lintable files less efficient than just filtering by paths prefix. 

Maybe it would be better to provide an opportunity to use such option based on concrete cases

jpsim:
I’m open to it.

JohnReeze:
opened new PR for this https://github.com/realm/SwiftLint/pull/3345

PR #3331, opened 2020-09-03
Use correct term for associated values

lightsprint09:
See https://docs.swift.org/swift-book/LanguageGuide/Enumerations.html for reference


jpsim:
Thanks for the PR!

lightsprint09:
Thanks for accepting the PR so fast. Great project.

PR #3336, opened 2020-09-10
Add `excluded_match_kinds` custom rule config parameter

ryandemo:
This allows custom rules to define an `excluded_match_kinds` array instead of listing out all but one or a few of the `SyntaxKind`s in `match_kinds`. Rules that include both `match_kinds` and `excluded_match_kinds` will be invalid, since there's not an obvious way to resolve the two without an arbitrary priority between them.


jpsim (approved):
Thanks for the PR! Will merge once CI passes.

truongnhi08 (commented):
Yêu cầu kéo

PR #3342, opened 2020-09-15
Fix finding the nested config when a single file path is passed

### LINKED ISSUE #3341: SwiftLint doesn't find a nested configuration when one specific file is specified
### New Issue Checklist

  - #3038 is similar, but not the same

### Describe the bug

Assume a file exists called `Path/To/My/File.swift`, and assume `Path/To/My` is a directory that contains a `.swiftlint.yml` file. When you tell SwiftLint to lint _only_ the file, it uses the default configuration instead of the YAML file in `Path/To/My`.

I believe the root cause is due to [these lines](https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Extensions/Configuration+Merging.swift#L19-L21), which return the default configuration immediately instead of checking the directory for any configurations.

```bash
$ swiftlint lint -- Path/To/My/File.swift
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.40.2

```yml
disabled_rules:
    - cyclomatic_complexity
    - force_cast
    - function_body_length
```

  Yes, see above.
Xcode 11.7
Build version 11E801a
  See above.

```swift
// This triggers a violation:
let foo = "Test" as! Array
```

sethfri:
Fixes #3341

When SwiftLint searches for nested configurations and only one file has been passed in via the `paths` argument, SwiftLint returns early after inadvertently comparing the containing directory of the file-to-be-linted to itself. This happens because `rootDirectory` is calculated from `rootPath`, which is set to the file being linted in this scenario.

### DIFF
--- a/Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
@@ -16,7 +16,9 @@ extension Configuration {
     }
 
     private func configuration(forPath path: String) -> Configuration {
-        if path == rootDirectory {
+        let rootConfigurationDirectory = configurationPath?.bridge().deletingLastPathComponent
+        // We're linting a file in the same directory as the root configuration we've already loaded
+        if path == rootConfigurationDirectory {
             return self
         }
 
--- a/Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
@@ -51,23 +51,6 @@ extension Configuration {
         return self
     }
 
-    private var rootDirectory: String? {
-        guard let rootPath = rootPath else {
-            return nil
-        }
-
-        var isDirectoryObjC: ObjCBool = false
-        guard FileManager.default.fileExists(atPath: rootPath, isDirectory: &isDirectoryObjC) else {
-            return nil
-        }
-
-        if isDirectoryObjC.boolValue {
-            return rootPath
-        } else {
-            return rootPath.bridge().deletingLastPathComponent
-        }
-    }
-
     private struct HashableRule: Hashable {
         fileprivate let rule: Rule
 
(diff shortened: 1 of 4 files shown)


Package.resolved
```diff
@@ -10,6 +10,15 @@
...
       },
+      {
+        "package": "CryptoSwift",
+        "repositoryURL": "https://github.com/krzyzanowskim/CryptoSwift.git",
+        "state": {
+          "branch": null,
+          "revision": "3a2acbb32ab68215ee1596ee6004da8e90c3721b",
+          "version": "1.0.0"
+        }
+      },
```
sethfri, line 21: This was auto-generated when I ran one of the 3 commands in the [Contributing guide's Tests section](https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tests)

Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
```diff
@@ -16,7 +16,7 @@ extension Configuration {
     }
 
     private func configuration(forPath path: String) -> Configuration {
-        if path == rootDirectory {
+        if path == rootDirectory && configurationPath != nil {
```
sethfri, line 19: This is the fix

sethfri:
This approach turned out not to be a complete fix. I think I have this working a lot better in a branch by completely rewriting the merging algorithm. I'll get the PR updated in the next 24 hours

sethfri:
Rewriting the merging algorithm wasn't necessary after all. Ended up discovering a much simpler fix. Changes pushed, and I've updated the PR description to reflect the new change.

Part of the problem was `rootPath` is super confusingly named (it's more of a leaf path). I'll open up a new PR to propose changing the name.

sethfri:
@jpsim @marcelofabri @keith Sorry for the ping, not sure who to reach out to - just looked at recent PR reviewers. I'm not sure why the buildkite build failed. There doesn't seem to be any info in the log other than

```
Danger has failed this build.
--
  | Found 1 error.
  | 🚨 Error: The command exited with status 1
```

Does one of you mind taking a look? 🙏🏻

sethfri:
@jpsim @marcelofabri @keith Looks like the CI failure from before was just a flake. Should be good to go here

jpsim:
Thanks for the PR!

PR #3345, opened 2020-09-16
Add excluding by prefix option

JohnReeze:
As we discussed here https://github.com/realm/SwiftLint/pull/3325 sometimes current excluding algorithm maybe slower than    excluding paths by absolute prefix. So I added option for such cases.

Based on what I've checked it works faster for next scenarios:
- the number of input files is relatively small (e.g. when using `use-script-input-files`) and excluded directories contain relatively big number of lintable files
- the number of excluded directories relatively small (e.g. Pods + ThirdParty) and globs not used

But still it depends on many parameters which can be not be automatically evaluated so 

@jpsim looking forward to your review =)


jpsim:
@JohnReeze is this working as you intend it to? I wouldn't expect adding this option to impact existing projects, but OSSCheck reported hundreds of changes in [this comment](https://github.com/realm/SwiftLint/pull/3345#issuecomment-693723592).

JohnReeze:
@jpsim It should work (and it does as I use it for about a month). And it's strange that OSSCheck reported 430 warnings because I only added optional parameter which by default is not supposed to impact anything

JohnReeze:
I guess it's somehow connected to https://github.com/realm/SwiftLint/pull/3376 with same 430 warnings)
What am I supposed to do in that case? @jpsim

jpsim:
I see you merged master into your branch, let's see if the same thing happens.

jpsim:
Awesome, looks like OSSCheck got confused previously.

Overall, I'm supportive of this change, although I have a small concern of adding diverging behaviors in small ways like this, and the fact that now you'd get different results if you run `swiftlint` with or without this argument, but I understand that we weren't able to find a way to improve the performance of the current algorithm without changing its behavior.

I'll be pushing some very minor edits to your branch and then will merge when CI is green.

jpsim (approved):

Issue #3348, opened 2020-09-17
Fix all parsing warnings & errors in rule examples


jpsim:
Many rule examples are very succinct, focusing just on the syntax that is meant to be linted/corrected. For example: `if foo == nil`. However, this doesn't parse successfully:

```
$ echo "if foo == nil" > file.swift
$ swiftc -parse file.swift
file.swift:1:14: error: expected '{' after 'if' condition
if foo == nil
             ^
```

In an upcoming PR, I'll want to introduce a change to avoid correcting files if they have parser errors or warnings, since SwiftLint isn't designed to handle all possibilities of invalid Swift code. Generally, it's still useful for SwiftLint to _lint_ Swift files that don't parse 100% cleanly as a feedback mechanism, but we shouldn't attempt to _correct_ files with violations that don't parse cleanly, because that can mangle or delete code in unintended ways.

Because there are currently over 500 rule examples with parser diagnostics, I don't have time to fix all of them, but we probably should do that eventually.

This is a good starter task if someone's interested in contributing to SwiftLint.

lo1tuma:
> Generally, it's still useful for SwiftLint to lint Swift files that don't parse 100% cleanly

What is the reasoning behind this? And does this mean that just `SwiftLint` should be able to handle invalid code or does this apply for the examples as well. Maybe we could introduce a mechanism to parse all examples and check if there are any parse errors (maybe directly in the `Example` `init` method). That way we could make sure all examples are valid while still allowing invalid code to go through `SwiftLint`. 

I think there is not much value in fixing all those parse errors now, without a way to enforcing examples to be valid. If we don’t enforce it, then we would end up in the same situation again after a while, because examples are added or get modified and might become invalid again.

PR #3349, opened 2020-09-17
Skip correcting files with parser diagnostics

jpsim:
We should avoid correcting files if they have parser errors or warnings, since SwiftLint isn't designed to handle all possibilities of invalid Swift code.

Generally, it's still useful for SwiftLint to _lint_ Swift files that don't parse 100% cleanly as a feedback mechanism, but we shouldn't attempt to _correct_ files with violations that don't parse cleanly, because that can mangle or delete code in unintended ways.

This PR also fixes many rule examples with parser diagnostics, but there are still over 500 rule examples with parser diagnostics, I don't have time to fix all of them, but we probably should do that eventually. I filed https://github.com/realm/SwiftLint/issues/3348 to track that, which is a good starter task if someone's interested in contributing to SwiftLint.


keith (approved):
awesome!

marcelofabri:
I'm late to the party, but shouldn't we only skip files with errors? Warnings shouldn't affect SwiftLint's ability to format the code

jpsim:
I agree, I'm supportive if you want to make that change @marcelofabri.

marcelofabri:
> I agree, I'm supportive if you want to make that change @marcelofabri.

will do!

PR #3350, opened 2020-09-18
Include Linux zip in list of GitHub release binaries

sethfri:
Including the Linux zip to match what has been included in recent releases. 

I also took the liberty of renumbering the numbered list to all 1s to make it easier to maintain in the future. One can add/remove steps in the list without needing to renumber everything.


jpsim:
Thanks!

Issue #3267, opened 2020-07-31
unused_capture_list false positive with explicit capturing of strong self in Swift 5.3


holgerde:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

A new feature in Swift 5.3 allows to explicitly capture strong self and then drop the self. prefix in the closure (see [SE-0269](https://github.com/apple/swift-evolution/blob/master/proposals/0269-implicit-self-explicit-capture.md)). However, SwiftLint's unused_capture_list then treats the capture as unused, because self is not explicitly written out. It is still used implicitly though, so it should not produce an error.

See below for an example.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint lint Example.swift --no-cache                   
Linting Swift files at paths Example.swift
Linting 'Example.swift' (1/1)
Example.swift:14:50: warning: Unused Capture List Violation: Unused reference self in a capture list should be removed. (unused_capture_list)
Done linting! Found 1 violation, 0 serious in 1 file.
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)?
  * 0.39.2
* Installation method used (Homebrew, CocoaPods, building from source, etc)?
  * homebrew
* Paste your configuration file:
  * none, unused_capture_list is enabled by default

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)?
  If so, paste their relative paths and respective contents.
  * no
* Which Xcode version are you using (check `xcodebuild -version`)?
```
Xcode 12.0
Build version 12A8169g
```
* Do you have a sample that shows the issue?

```swift
URLSession.shared.dataTask(with: url) { [self] (data, _, _) in
    handle(data)
}
```

marcelofabri:
Hmm, looks like the Swift compiler actually warns on unused captures, so I wonder if we should just remove the rule instead:

```
$ cat file.swift 
final class A {
    func b() {
        let x = { [weak self] in print("") }
        x()
    }
}

$ swift file.swift 
file.swift:3:25: warning: variable 'self' was written to, but never read
        let x = { [weak self] in print("") }
```

jberkel:
Looks like this isn't handling all cases, if `self` is qualified with `unowned` it still triggers:

```swift
class A {
  func foo() {}

  func bar() {
    _ = { [unowned self] in
       foo()
    }
  }
[...]

PR #3352, opened 2020-09-18
[Fix] `UnusedCaptureListRule`: implicit self in @escaping closures

hank121314:
## Summary

This pr fixes: #3267 .

Since Swift 5.3 is released. It support implicit self in @escaping closures([SE-0269](https://github.com/apple/swift-evolution/blob/master/proposals/0269-implicit-self-explicit-capture.md))

I think we should exclude self keyword from capture list rule.

## Test Plan

I have already add some test in example.

Thanks you so much for your code review!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,9 @@
 
 #### Bug Fixes
 
-* None.
+* Rule `unused_capture_list` should not be triggered by self keyword.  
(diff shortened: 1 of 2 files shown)


jpsim:
Thank you so much for the PR!

shawnkoh:
This fix only catches [self]
[unowned self] still triggers.

PR #3353, opened 2020-09-18
Fix false positives for 'multiple_closures_with_trailing_closure'

### LINKED ISSUE #3295: "Multiple Closures With Trailing Closure" support for multiple trailing closures (Swift 5.3)
### New Issue Checklist

- [x] Updated SwiftLint to the latest version -- 0.40.0

### Describe the bug

SwiftLint's 'Multiple Closures With Trailing Closure' rule flags the use of Swift 5.3's new multiple trailing closure syntax ([SE-209](https://github.com/apple/swift-evolution/blob/master/proposals/0279-multiple-trailing-closures.md)) on the last closure.

Example:

```swift
Button {
    print("Button tapped")
} label: { // Multiple Closures with Trailing Closure Violation
    Text("Button")
}
```

### Proposed Fix/Behavior

The new syntax serves to alleviate the call site confusion that initially motivated creating the rule. Thus, the rule implementation could be adjust to detect multiple closures with trailing *closure(s)* -- flag when a non-trailing closure is used as an argument that has either a single trailing closure or multiple trailing closures.

This would mean this would get flagged:
```swift
Button(action: { print("Button tapped") }) { // Multiple Closures with Trailing Closure Violation
    Text("Button")
}
```

But this would _not_:
```swift
Button {
    print("Button tapped")
} label: {
    Text("Button")
}
```

And this would _not_:
```swift
UIView.animate(withDuration: 1.0) {
    self.view.alpha = 0
} completion: { _ in
    self.view.removeFromSuperview()
}
```

```bash
$ echo "Button {
    print(\"Button tapped\")
} label: {
    Text(\"Button\")
}" | swiftlint lint --no-cache --use-stdin --enable-all-rules

<nopath>:3:3: warning: Multiple Closures with Trailing Closure Violation: Trailing closure syntax should not be used when passing more than one closure argument. (multiple_closures_with_trailing_closure)
```

### Environment

* SwiftLint version: 0.40.0
* Installation method used (Homebrew, CocoaPods, building from source, etc): Homebrew
Empty file. Relevant rule: `multiple_closures_with_trailing_closure`.
  No.
Xcode 12.0 Beta 4 (12A8179i)

```swift
Button {} label: { Text("Label") }
```

```bash
echo "Button {} label: { Text(\"Label\") }" | swiftlint lint --no-cache --use-stdin --enable-all-rules
```

Jumhyn:
Resolves #3295

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,11 @@
   [hank121314](https://github.com/hank121314)
   [#2367](https://github.com/realm/SwiftLint/issues/3267)
 
+* Rule `multiple_closures_with_trailing_closure` no longer triggers when Swift
+  5.3's 'multiple trailing closures' feature is used.
+  [Jumhyn](https://github.com/jumhyn)
+  [#3295](https://github.com/realm/SwiftLint/issues/3295)
+
 ## 0.40.2: Demo Unit
 
 #### Breaking
--- a/Source/SwiftLintFramework/Rules/Style/MultipleClosuresWithTrailingClosureRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/MultipleClosuresWithTrailingClosureRule.swift
@@ -20,7 +20,9 @@ public struct MultipleClosuresWithTrailingClosureRule: ASTRule, ConfigurationPro
             UIView.animate(withDuration: 1.0) {
                 someView.alpha = 0.0
             }
-            """)
+            """),
+            Example("foo.method { print(0) } arg2: { print(1) }"),
+            Example("foo.methodWithParenArgs((0, 1), arg2: (0, 1, 2)) { $0 } arg4: { $0 }")
         ],
         triggeringExamples: [
             Example("foo.something(param1: { $0 }) ↓{ $0 + 1 }"),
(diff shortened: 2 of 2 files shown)


Jumhyn:
This is likely a good candidate to be written in SwiftSyntax—what’s the current support status for that? I see an old PR got merged, reverted, and then re-opened, but it looks like that last PR has gone stale...

Jumhyn:
@jpsim Is there further work this needs to be mergeable?

jpsim:
> This is likely a good candidate to be written in SwiftSyntax—what’s the current support status for that? I see an old PR got merged, reverted, and then re-opened, but it looks like that last PR has gone stale...

SwiftSyntax needs to be properly linked to handle the issues discussed in https://github.com/realm/SwiftLint/issues/3105 before we can use it in a release.

jpsim:
Thanks for the PR!

I suspect there are ways to improve this to avoid using regular expressions, but given this is giving false positives on Swift 5.3, I'll merge this now.

Jumhyn:
@jpsim Yeah I'm not thrilled with the implementation, but I couldn't come up with a great alternative off the top of my head given the info available in a `SourceKittenDictionary`. This was my first time actually digging into the source of SwiftLint, though, so it's very likely I've overlooked something. If you have a suggestion about any directions to pursue I'd be more than happy to investigate!

PR #3354, opened 2020-09-18
Use index for unused imports

jpsim:
This makes the rule run about 2 times faster for the projects I tested on.

image


jpsim:
I can't avoid false positives with this approach.

PR #3358, opened 2020-09-21
Fix false positives in unused_declaration

jpsim:
Avoid marking protocol definitions as unused when they have a default implementation.

Add more tests.


jpsim:
This doesn't fix false positives, but it does introduce false negatives.

PR #3362, opened 2020-09-22
Revert "Fix finding the nested config when a single file path is passed"

jpsim:
Reverts realm/SwiftLint#3342

This caused a regression in one of my projects. cc @sethfri 

It's a closed source project so it'll take some time for me to craft a repro case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,9 +29,6 @@
   [hank121314](https://github.com/hank121314)
   [#2367](https://github.com/realm/SwiftLint/issues/3267)
 
-* Fix finding a nested config when a single file path is passed.  
-  [Seth Friedman](https://github.com/sethfri)
-
 ## 0.40.2: Demo Unit
 
 #### Breaking
--- a/Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
@@ -16,9 +16,7 @@ extension Configuration {
     }
 
     private func configuration(forPath path: String) -> Configuration {
-        let rootConfigurationDirectory = configurationPath?.bridge().deletingLastPathComponent
-        // We're linting a file in the same directory as the root configuration we've already loaded
-        if path == rootConfigurationDirectory {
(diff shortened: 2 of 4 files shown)


sethfri:
Interesting! Thanks for the tag. Lmk once you're able to get a repro case together; I'd be more than happy to take another stab at a fix!

jpsim:
Yeah, sorry for the lack of details yet. I have to reduce what we're doing to something I can share here.

jpsim:
It has something to do with specifying an out-of-tree config file while linting a directory that already has a `.swiftlint.yml` file at its root.

sethfri:
@jpsim That's helpful, thanks! Is the behavior you're seeing that the passed in file is ignored in favor of the root .swiftlint.yml?

jpsim:
I think so.

sethfri:
What is the expected behavior in that scenario @jpsim? Should the root file be completely ignored in favor of the passed in file, or would we expect them to be merged?

jpsim:
Hmmm, good question. I can see arguments for either way. In almost every other case we merge configurations with parents.

sethfri:
@jpsim Personally I would expect the config argument to be treated as an override. "Ignore what I have, just test my project out with this config". You mentioned that there was a regression here though, so I wanted to check what the expected behavior was for your project

jpsim:
Yes, for an explicit config passed as an argument, I'd also expect that to take precedence.

sethfri:
Just had time to set up a repro for this. Using the sample project (out-of-tree config not included in the zip):

* **Before my change:** Out-of-tree config is used in place of the root config
* **After my change:** Out-of-tree config is _merged_ with root config

[SwiftLintBugTest.zip](https://github.com/realm/SwiftLint/files/5360699/SwiftLintBugTest.zip)

Will see if I can figure out how to fix this, but is this consistent with what you were seeing @jpsim?

jpsim:
That sounds like what I was seeing, yes. Thanks for taking the time to do this!

sethfri:
@jpsim Np! Opened a PR with a fix in #3379

PR #3368, opened 2020-09-25
Fix some false positives in rule `explicit_self`

svenmuennich:
Some accesses to instance properties attributed with a property wrapper resulted in false positives for rule `explicit_self`. This was caused by only checking the character directly preceding the member access to be a `.`. Hence accesses to a projected value of a wrapped property (e.g. `self.$foo`) or the property wrapper itself (e.g. `self._foo`) failed the test.


jpsim:
Awesome PR, Thanks!

Issue #3312, opened 2020-08-20
Add CLI option to treat warnings as errors


Jeehut:
This feature was requested several times, on a quick search I found these related links:
https://github.com/realm/SwiftLint/issues/160, https://github.com/realm/SwiftLint/issues/268, https://github.com/realm/SwiftLint/issues/834, https://github.com/realm/SwiftLint/issues/1249, [this SO thread](https://stackoverflow.com/q/42406844/3451975)

From what I've read in the discussions there, I observed these things:
1. There was **no reason mentioned not to add support for such an option**, most questions from contributors were just trying to understand the underlying problem to provide a good long-term solution.
2. The `--strict` option (which makes the whole `swiftlint` command fail if any warnings appears) clearly seems to be misleading to some who think that it will convert warnings to errors. Others were exactly looking for what `--strict` does.

I just ran into this problem in a different context: While setting up [danger-ruby-swiftlint](https://github.com/ashfurrow/danger-ruby-swiftlint/) for a project I noticed that they provide a `additional_swiftlint_args` option which will pass options like `--strict` to SwiftLint, but because they are not working directly with the status code but instead use the reported warnings and errors as the source of truth for deciding if they should fail or not, there's currently no way to make the CI fail on warnings if using SwiftLint via that [Danger](https://github.com/danger/danger) plugin.

While I have reported the issue [here](https://github.com/ashfurrow/danger-ruby-swiftlint/issues/151) and it could be solved without any changes on the side of SwiftLint, I ask myself the question: **Why should SwiftLint not just provide what people were asking for several times already**? @jpsim even already mentioned in [this comment](https://github.com/realm/SwiftLint/issues/268#issuecomment-164495814) that the solution requires only a single line to be changed (+ the option parsing, of course). Another option can't really hurt, right? And if some people prefer it over `--strict`, then we should support it. Especially since `--strict` doesn't work in some use cases as explained in the example above. Don't get me wrong, I'm not suggesting to replace `--strict` here, but just complement it by this slightly different approach.

I will actually look into this myself and provide a PR, but before doing that I thought I should probably bring that topic up again and maybe discuss what your thoughts are and also what such an option could be named. I think the clearest name is `--treat-warnings-as-errors`, or a little shorter `--warnings-as-errors`. But even shorter names like `--always-error` could also be used. What do you think? Any opinions or suggestions?

Jeehut:
[...]

PR #3371, opened 2020-10-02
Add option warnings-as-errors to lint & analyze

Jeehut:
Implementes #3312.

I tested that this works manually by building via `swift build -c release` and running `.build/release/swiftlint --warnings-as-errors` and it worked just as expected. Without that option, I had 1 warning & 1 error reported, with that options 2 errors were reported, so I can confirm it works. But I did not add any unit tests because there seem to be no unit tests for the CLI product, other options like `--lenient` seem also not to be unit tested.

Note that I did not mess with any tresholds (like `lenient` does), I think we could think about those things after seeing how people are using this new option, so I'd rather keep this simple for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@
 
 #### Enhancements
 
-* None.
+* Add new option `warnings-as-errors` to lint & analyze.  
+  [Jeehut](https://github.com/Jeehut)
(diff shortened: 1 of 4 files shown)


jpsim:
Closed as https://github.com/realm/SwiftLint/pull/3372 has been merged.

PR #3372, opened 2020-10-03
Adjust --strict to treat warnings as errors instead of only altering exit code

Jeehut:
Implements #3312 with @jpsim's [suggested solution](https://github.com/realm/SwiftLint/issues/3312#issuecomment-702955921).

I tested that this works manually by building via `swift build -c release` and running `.build/release/swiftlint --strict` and it worked just as expected. Without that option, I had 1 warning & 1 error reported, with that options 2 errors were reported, so I can confirm it works. But I did not add any unit tests because there seem to be no unit tests for the CLI product, other options like `--lenient` seem also not to be unit tested.

Note that I did not mess with any tresholds (like `lenient` does), I think we could think about those things after seeing how people are using this new option, so I'd rather keep this simple for now.

Also note that I have decided to put the Changelog entry into the "Breaking Changes" section as some people might have been relying on the fact that using `--strict` was still making a distinction between warnings & errors when using the output of reporters directly. As this is changing this documented behavior, I feel like it would be wrong to place the change into the "Bug Fixes" section. But I can also move it there if needed, I do not have a strong opinion here.

### DIFF
--- a/Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
+++ b/Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
@@ -109,15 +109,30 @@ struct LintOrAnalyzeCommand {
     }
 
     private static func applyLeniency(options: LintOrAnalyzeOptions, violations: [StyleViolation]) -> [StyleViolation] {
-        if !options.lenient {
+        switch (options.lenient, options.warningsAsErrors) {
+        case (false, false):
             return violations
-        }
-        return violations.map {
-            if $0.severity == .error {
-                return $0.with(severity: .warning)
-            } else {
-                return $0
+
+        case (true, false):
+            return violations.map {
+                if $0.severity == .error {
+                    return $0.with(severity: .warning)
+                } else {
+                    return $0
+                }
             }
+
+        case (false, true):
+            return violations.map {
+                if $0.severity == .warning {
+                    return $0.with(severity: .error)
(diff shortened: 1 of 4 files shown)


jpsim:
> But I did not add any unit tests because there seem to be no unit tests for the CLI product, other options like `--lenient` seem also not to be unit tested.

Correct, this is unfortunate, and ideally I'd like to get rid of our pbxproj Xcode project altogether, which would then make it easier to add a test target just using the SwiftPM manifest that can test the `swiftlint` executable product.

jpsim:
> Also note that I have decided to put the Changelog entry into the "Breaking Changes" section

Thanks, I think that's the right call too!

Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
```diff
@@ -109,15 +99,30 @@ struct LintOrAnalyzeCommand {
...
+            return violations.map {
+                if $0.severity == .warning {
+                    return $0.with(severity: .error)
+                } else {
+                    return $0
+                }
+            }
+
+        case (true, true):
+            queuedFatalError("Invalid command line options: 'lenient' and 'strict' are mutually exclusive.")
```
jpsim, line 125: 👌

jpsim (approved):
Great, I probably should have done it this way originally. Thanks!

PR #3375, opened 2020-10-07
Fix some missed cases in rule `unavailable_function`. (realm#3374)

quinntaylor:
Enhanced to detect `abort()` and `preconditionFailure()` calls, in addition to `fatalError()`.


jpsim:
Thanks for the PR! I'll be rebasing and merging this when CI passes.

PR #3376, opened 2020-10-07
Add TestCaseAccessibilityRule

keith:
This rule verifies that all members of XCTestCase subclasses are private
(or fileprivate) except for those defined by XCTestCase. This is useful
for a few things:

1. You can make sure code that isn't needed outside the test case isn't
   overly accessible
2. You can catch typos like `func tsetSomething` that won't actually be
   run

This rule has configuration for `method_prefixes` so that custom methods
can still be allowed. This is useful if you disable tests by renaming
them (so they continue to be compiled) like `func disabled_testFoo`

This change also extracts some useful shared logic around identifying
what members are provided by XCTest since they are now used in multiple
places.

### DIFF
--- a/Source/SwiftLintFramework/Helpers/XCTestHelpers.swift
+++ b/Source/SwiftLintFramework/Helpers/XCTestHelpers.swift
@@ -1,24 +1,14 @@
 import SourceKittenFramework
 
-private let testFunctionNames: Set = [
-    "setUp()",
-    "setUpWithError()",
-    "tearDown()",
-    "tearDownWithError()"
-]
-
 private let testVariableNames: Set = [
     "allTests"
 ]
 
 enum XCTestHelpers {
-    static func isXCTestMember(kind: SwiftDeclarationKind, name: String) -> Bool {
-        if SwiftDeclarationKind.functionKinds.contains(kind) {
-            return name.hasPrefix("test") || testFunctionNames.contains(name)
-        } else if SwiftDeclarationKind.variableKinds.contains(kind) {
-            return testVariableNames.contains(name)
-        }
-
-        return false
+    static func isXCTestMember(kind: SwiftDeclarationKind, name: String,
+                               attributes: [SwiftDeclarationAttributeKind]) -> Bool {
+        return attributes.contains(.override)
+            || (kind == .functionMethodInstance && name.hasPrefix("test"))
+            || ([.varStatic, .varClass].contains(kind) && testVariableNames.contains(name))
     }
 }
--- a/Source/SwiftLintFramework/Helpers/XCTestHelpers.swift
+++ b/Source/SwiftLintFramework/Helpers/XCTestHelpers.swift
@@ -1,0 +1,24 @@
+import SourceKittenFramework
+
+private let testFunctionNames: Set = [
+    "setUp()",
+    "setUpWithError()",
+    "tearDown()",
+    "tearDownWithError()"
+]
+
+private let testVariableNames: Set = [
+    "allTests"
+]
+
+enum XCTestHelpers {
+    static func isXCTestMember(kind: SwiftDeclarationKind, name: String) -> Bool {
+        if SwiftDeclarationKind.functionKinds.contains(kind) {
+            return name.hasPrefix("test") || testFunctionNames.contains(name)
+        } else if SwiftDeclarationKind.variableKinds.contains(kind) {
+            return testVariableNames.contains(name)
+        }
+
+        return false
+    }
+}
--- a/Source/SwiftLintFramework/Rules/Lint/TestCaseAccessibilityRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/TestCaseAccessibilityRule.swift
@@ -1,17 +1,17 @@
 import SourceKittenFramework
 
-public struct EmptyXCTestMethodRule: Rule, OptInRule, ConfigurationProviderRule, AutomaticTestableRule {
-    public var configuration = SeverityConfiguration(.warning)
+public struct TestCaseAccessibilityRule: Rule, OptInRule, ConfigurationProviderRule, AutomaticTestableRule {
+    public var configuration = TestCaseAccessibilityConfiguration()
 
     public init() {}
 
     public static let description = RuleDescription(
-        identifier: "empty_xctest_method",
-        name: "Empty XCTest Method",
-        description: "Empty XCTest method should be avoided.",
+        identifier: "test_case_accessibility",
+        name: "Test case accessibility",
+        description: "Test cases should only contain private non-test members.",
         kind: .lint,
-        nonTriggeringExamples: EmptyXCTestMethodRuleExamples.nonTriggeringExamples,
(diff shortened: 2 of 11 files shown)


Source/SwiftLintFramework/Helpers/XCTestHelpers.swift
```diff
@@ -0,0 +1,24 @@
+import SourceKittenFramework
+
+private let testFunctionNames: Set = [
+    "setUp()",
```
jpsim, line 4: Any reason not to include all the overridable members defined here? https://developer.apple.com/documentation/xctest/xctestcase & https://developer.apple.com/documentation/xctest/xctest

Source/SwiftLintFramework/Helpers/XCTestHelpers.swift
```diff
@@ -0,0 +1,24 @@
...
+]
+
+private let testVariableNames: Set = [
+    "allTests"
+]
+
+enum XCTestHelpers {
+    static func isXCTestMember(kind: SwiftDeclarationKind, name: String) -> Bool {
+        if SwiftDeclarationKind.functionKinds.contains(kind) {
+            return name.hasPrefix("test") || testFunctionNames.contains(name)
```
jpsim, line 17: Some of these are only test members if they're static or instance functions. For example, the following examples would be incorrectly counted as test members:

```swift
final class MyTests: XCTestCase {
  static func testHelper() -> TestHelper { ... }
}
```

Source/SwiftLintFramework/Helpers/XCTestHelpers.swift
```diff
@@ -0,0 +1,24 @@
...
+        } else if SwiftDeclarationKind.variableKinds.contains(kind) {
+            return testVariableNames.contains(name)
```
jpsim, line 19: Same here, seems like we should only be checking for a static var called `allTests`, otherwise this would be incorrectly counted as a test member:

```swift
final class MyTests: XCTestCase {
  let allTests: [TestInput] { ... }
}
```

Source/SwiftLintFramework/Rules/Lint/TestCaseAccessibilityRule.swift
```diff
@@ -0,0 +1,52 @@
...
+        return testClasses(in: file).flatMap { violations(in: file, for: $0) }
+    }
+
+    // MARK: - Private
+
+    private func testClasses(in file: SwiftLintFile) -> [SourceKittenDictionary] {
+        let dict = file.structureDictionary
+        return dict.substructure.filter { dictionary in
+            guard dictionary.declarationKind == .class else { return false }
+            return dictionary.inheritedTypes.contains("XCTestCase")
```
jpsim, line 27: Nit:

```suggestion
            dictionary.declarationKind == .class &&
                dictionary.inheritedTypes.contains("XCTestCase")
```

Source/SwiftLintFramework/Rules/Lint/TestCaseAccessibilityRule.swift
```diff
@@ -0,0 +1,52 @@
...
+
+            return StyleViolation(ruleDescription: Self.description,
+                                  severity: configuration.severity,
+                                  location: Location(file: file, byteOffset: offset))
+        }
+    }
+
+    private func isXCTestMember(kind: SwiftDeclarationKind, name: String) -> Bool {
+        return XCTestHelpers.isXCTestMember(kind: kind, name: name)
+            || configuration.methodPrefixes.contains { name.hasPrefix($0) }
```
jpsim, line 50: Nit:

```suggestion
            || configuration.methodPrefixes.contains(where: name.hasPrefix)
```

jpsim (commented):
Looks good, and useful! I just have a few questions/suggestions.

keith:
Updated to more accurately identify XCTestCase members

Source/SwiftLintFramework/Rules/RuleConfigurations/TestCaseAccessibilityConfiguration.swift
```diff
@@ -1,10 +1,10 @@
 public struct TestCaseAccessibilityConfiguration: RuleConfiguration, Equatable {
     public private(set) var severityConfiguration = SeverityConfiguration(.warning)
-    public private(set) var methodPrefixes: Set<String> = []
+    public private(set) var allowedPrefixes: Set<String> = []
 
     public var consoleDescription: String {
         return severityConfiguration.consoleDescription +
-        	", method_prefixes: [\(methodPrefixes)]"
+            ", allowed_prefixes: [\(allowedPrefixes)]"
```
jpsim, line 7: Examples can now specify configurations, so it'd be nice to test something like `disabled_test` as an allowed prefix. https://github.com/realm/SwiftLint/blob/da408b5901c0354a2930e9faf28dd678990fbca3/Source/SwiftLintFramework/Models/Example.swift#L6-L16

Source/SwiftLintFramework/Helpers/XCTestHelpers.swift
```diff
@@ -1,24 +1,14 @@
...
-        if SwiftDeclarationKind.functionKinds.contains(kind) {
-            return name.hasPrefix("test") || testFunctionNames.contains(name)
-        } else if SwiftDeclarationKind.variableKinds.contains(kind) {
-            return testVariableNames.contains(name)
-        }
-
-        return false
+    static func isXCTestMember(kind: SwiftDeclarationKind, name: String,
+                               attributes: [SwiftDeclarationAttributeKind]) -> Bool {
+        return attributes.contains(.override)
```
jpsim, line 10: Nice.

Source/SwiftLintFramework/Rules/RuleConfigurations/TestCaseAccessibilityConfiguration.swift
keith: Is there an example of that working? I tried that and I don't believe it's passed all the way through the tests

keith: AFAICT it only is for analyze tests

jpsim: Oh I think you're right. Let's hold off for now, but it should be easy to make normal lint tests apply this configuration value.

jpsim (approved):
Really useful rule, thanks!

davidelorenzitw:
Nice rule! 
There are cases where the test class is also implementing protocols (ie. for testing some delegate callback, rather than having to create a new mocked type that the test instantiates and checks), and these functions/properties can't be declared as private, meaning this new rule can't be used in those cases. (note: function prefix configuration can't be applied here)

So maybe a possible improvement to consider: would it be possible to ie. only execute this rule's check within the main class declaration, and skip any extension declared on the test class instead? This would give some room in how to organise the class definition while still having this rule enabled.

PR #3379, opened 2020-10-11
Fix finding the nested config when a single file path is passed 

### LINKED ISSUE #3341: SwiftLint doesn't find a nested configuration when one specific file is specified

sethfri:
This was previously attempted in #3342, but produced a bug in the case where `--config` is used to specify a config from outside of the source tree. The `--config` argument wasn't always being used as an override, and was being merged with the config in the source tree. This has now been addressed and reverts the revert done in #3362. 

Fixes #3341

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -141,6 +141,9 @@
   [Jumhyn](https://github.com/jumhyn)
   [#3295](https://github.com/realm/SwiftLint/issues/3295)
 
+* Fix finding a nested config when a single file path is passed.  
+  [Seth Friedman](https://github.com/sethfri)
+
 ## 0.40.2: Demo Unit
 
 #### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,9 @@
   [ZevEisenberg](https://github.com/ZevEisenberg)
   [#3033](https://github.com/realm/SwiftLint/issues/3033)
 
+* Fix finding a nested config when a single file path is passed.  
+  [Seth Friedman](https://github.com/sethfri)
+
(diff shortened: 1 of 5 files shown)


sethfri:
@jpsim This should be ready for your input! A couple things:

1. Are you able to test this against your closed source project to make sure it's the right fix?
2. Do you have any recs for an automated test for the edge case your project found? It's not immediately obvious to me where to test a passed-in command line argument

sethfri:
Hey @jpsim, I know it's a crazy time, but since it's been a couple weeks do you think you could take a look at the few PRs I have out? They're starting to encounter merge conflicts

jpsim (approved):
👍 Thanks for the PR! Will merge when CI is green.

PR #3384, opened 2020-10-19
Use working directory as root path instead of trying to calculate it

### LINKED ISSUE #3383: Bug when running SwiftLint on a non-root directory
### New Issue Checklist

### Describe the bug

Running SwiftLint directly on a module within my codebase (directory in the tree) only uses the configuration file in that directory instead of merging it with the configuration file at the root.

I would expect this to have the same behavior as running SwiftLint at the root does. In that case, the root configuration file is merged with the configuration file in my module, and everything works as expected.

**Sample project:** [DirectoryBug.zip](https://github.com/realm/SwiftLint/files/5398412/DirectoryBug.zip)

`swiftlint lint --no-cache`

```
Linting Swift files at paths 
Linting 'ViewController.swift' (1/4)
Linting 'AppDelegate.swift' (2/4)
Linting 'SceneDelegate.swift' (3/4)
Linting 'MyType.swift' (4/4)
Done linting! Found 0 violations, 0 serious in 4 files.
Program ended with exit code: 0
```

With working directory set to the root of the DirectoryBug project:
`swiftlint lint --no-cache -- DirectoryBug/Module1`

```
Linting Swift files at paths DirectoryBug/Module1
Linting 'MyType.swift' (1/1)
/Users/sfriedman/Desktop/DirectoryBug/DirectoryBug/Module1/MyType.swift:10:51: warning: Unneeded NotificationCenter Removal Violation: Observers are automatically unregistered on dealloc (iOS 9 / macOS 10.11) so you should't call `removeObserver(self)` in the deinit. (unneeded_notification_center_removal)
Program ended with exit code: 0
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? `master`
* Installation method used (Homebrew, CocoaPods, building from source, etc)? Building from source

See attached sample project

  Yes, `.swiftlint.yml` at the root of the project and `DirectoryBug/Module1/.swiftlint.yml`
* Which Xcode version are you using (check `xcodebuild -version`)? 11.7

Yes, see description

sethfri:
We currently try to derive the root path by using the path passed in. However, this doesn't allow us to actually get the root, just the containing directory of the path passed in. We also don't currently have a heuristic to use for deriving the root path for multiple paths passed in.

The fix (which fixes #3383) is to remove our root path calculation and just use the current working directory as the root instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -112,6 +112,11 @@
 * Fix finding a nested config when a single file path is passed.  
   [Seth Friedman](https://github.com/sethfri)
 
+* Fix incorrect calculation of the root path when a directory in the
+  tree is passed in as a path argument.  
+  [Seth Friedman](https://github.com/sethfri)
+  [#3383](https://github.com/realm/SwiftLint/issues/3383)
+
 ## 0.40.3: Greased Up Drum Bearings
 
 #### Breaking
--- a/Source/swiftlint/Extensions/Configuration+CommandLine.swift
+++ b/Source/swiftlint/Extensions/Configuration+CommandLine.swift
@@ -227,16 +227,12 @@ extension Configuration {
         })
     }
 
-    private static func rootPath(from paths: [String]) -> String? {
-        // We don't know the root when more than one path is passed (i.e. not useful if the root of 2 paths is ~)
-        return paths.count == 1 ? paths.first?.absolutePathStandardized() : nil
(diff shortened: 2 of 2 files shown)


sethfri:
@jpsim This one's ready for your input!

jpsim (approved):
Thanks for the PR! Will merge when CI is green.

PR #3387, opened 2020-10-19
Add `isNotEmpty` extension

PaulTaykalo:
From one point of view, this is a simple sugar-like extension.

**Pros**:
 - Some expressions that are using `isEmpty` calls will look like a natural language
 - No going back and forth when parsing some complex conditional expressions

**Cons**:
  - Some linting rules for the framework itself can break
  - There's no lining rule that will force this usage later.


jpsim:
Thanks!

PR #3388, opened 2020-10-20
Fix inconsistency in `operator_usage_whitespace` rule

PaulTaykalo:
This Fixes case when the right part of the expression is an array or a string
Previously, strings and comments tokens were ignored.
In the current implementation, matching done first and then those are filtered if the operator is within the string token

### DIFF
--- a/Source/SwiftLintFramework/Rules/Style/OperatorUsageWhitespaceRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/OperatorUsageWhitespaceRule.swift
@@ -32,7 +32,8 @@ public struct OperatorUsageWhitespaceRule: OptInRule, CorrectableRule, Configura
             Example("let doubleValue = -9e-11\n"),
             Example("let foo = GenericType<(UIViewController) -> Void>()\n"),
             Example("let foo = Foo<Bar<T>, Baz>()\n"),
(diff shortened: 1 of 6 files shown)


PaulTaykalo:
This should fix 3321 once done

Source/SwiftLintFramework/Rules/Style/OperatorUsageWhitespaceRule.swift
```diff
@@ -79,19 +85,18 @@ public struct OperatorUsageWhitespaceRule: OptInRule, CorrectableRule, Configura
     }
 
     private func violationRanges(file: SwiftLintFile) -> [(NSRange, String)] {
-        let escapedOperators = ["/", "=", "-", "+", "*", "|", "^", "~"].map({ "\\\($0)" }).joined()
```
PaulTaykalo, line 82: don't' need to be escaped 🤷

PaulTaykalo:
@jpsim Please review.

While this actually adds consistency in rule, table-formatted variants will no have a lot of "false-positives" from the user's perspective

jpsim (approved):
Looks great, thanks!

jpsim:
Will merge when CI is green.

PR #3391, opened 2020-10-22
Add option that allows skipping aligned constants

PaulTaykalo:
Skip aligned constants by default
Configuration for Whitespace rule was added for the ability to specify whether the aligned constants should be ignored or not

This PR will now fix cases like 
```
let a      = 1
let b      = 1
let c      = 1
```
for the cases, when the assignment operator is on the same column as in lines around


jpsim (approved):
Thanks for the PR, will merge when CI passes.

Issue #3389, opened 2020-10-20
unused_capture_list false positive with explicit capturing of unowned self in Swift 5.3


lordzsolt:
### New Issue Checklist

- [X] Updated SwiftLint to the latest version
- [X] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

Same issue as #3267 but with `unowned` instead of `strong`.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint lint Example.swift --no-cache                   
Linting Swift files at paths Example.swift
Linting 'Example.swift' (1/1)
Example.swift:14:50: warning: Unused Capture List Violation: Unused reference self in a capture list should be removed. (unused_capture_list)
Done linting! Found 1 violation, 0 serious in 1 file.
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.40.3
* Installation method used (Homebrew, CocoaPods, building from source, etc)? Cocoapods
* Paste your configuration file: none, `unused_capture_list` is enabled by default

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? No
* Which Xcode version are you using (check `xcodebuild -version`)? Xcode 12.2 Beta
* Do you have a sample that shows the issue? 

```swift
override func viewDidLoad() {
	super.viewDidLoad() {
		rx.onViewDidAppear.subscribe(onNext: { [unowned self] in 
			doSomething()
		}).disposed(by: disposeBag)
	}

	func doSomething() {
		print("Watch this")
	}
}
```

PR #3392, opened 2020-10-23
[Fix] `UnusedCaptureListRule`: unowned self in @escaping closures

hank121314:
## Summary

This pr fixes: #3389  .

It seems that we also need to prevent [unowned self] from triggering this rule.

## Test Plan

I have already add some test in example.

Thanks you so much for your code review! 😄


PaulTaykalo (approved):

PaulTaykalo:
Thanks!

PR #3410, opened 2020-11-08
Fix false positive in `toggle_bool` rule

### LINKED ISSUE #3398: [bug] toggle_bool false positive
### New Issue Checklist

- [ ] Updated SwiftLint to the latest version - I honestly didn't and I'm on v0.39.2 but I believe this still happens in latest version (can't upgrade to test right now 😿 ) - feel free to close if you're able to test it and it doesn't happen in the latest release 🙏 

### Describe the bug

`toggle_bool` false positive.

```bash
$ swiftlint lint
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.39.2

```yml
# The only relevant part is:

opt_in_rules:
  - toggle_bool
```

  If so, paste their relative paths and respective contents. 🤔 don't think so
* Which Xcode version are you using (check `xcodebuild -version`)? Xcode 12.0

This shows my issue:

```bash
echo "something = \!something.isEmpty ? executeA() : executeB()" | swiftlint lint --no-cache --use-stdin --enable-all-rules
```

There is a `\` before the `!` to escape the `!` because apparently that causes issues in the script. But the example in Swift would be:

```swift
// This triggers a violation:
something = !something.isEmpty ? executeA() : executeB()
```

And it shouldn't. In my project the line was much more complex but I could simplify and reproduce it using that statement. In my project I just switched the ternary to e.g. `something.isEmpty ? executeB() : executeA()` but not sure if that will be possible in all cases 🤔

marcelofabri:
Fixes #3398

### DIFF
--- a/Source/SwiftLintFramework/Rules/Idiomatic/ToggleBoolRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/ToggleBoolRule.swift
@@ -17,7 +17,8 @@ public struct ToggleBoolRule: SubstitutionCorrectableRule, ConfigurationProvider
             Example("view.clipsToBounds.toggle()\n"),
             Example("func foo() { abc.toggle() }"),
             Example("view.clipsToBounds = !clipsToBounds\n"),
-            Example("disconnected = !connected\n")
+            Example("disconnected = !connected\n"),
+            Example("result = !result.toggle()")
         ],
         triggeringExamples: [
             Example("↓isHidden = !isHidden\n"),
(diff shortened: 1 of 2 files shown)


jpsim (approved):

Source/SwiftLintFramework/Rules/Idiomatic/ToggleBoolRule.swift
```diff
@@ -41,7 +42,7 @@ public struct ToggleBoolRule: SubstitutionCorrectableRule, ConfigurationProvider
     }
 
     public func violationRanges(in file: SwiftLintFile) -> [NSRange] {
-        let pattern = "(?<![\\w.])([\\w.]+) = !\\1\\b"
+        let pattern = #"(?<![\w.])([\w.]+) = !\1\b(?!\.)"#
```
rogerluan, line 45: Simple and effective :) awesome!

PR #3411, opened 2020-11-08
Add `key_path_expression_as_function` opt-in rule

### LINKED ISSUE #3082: Favor Key Path Expressions as Functions
As of Swift 5.2 you can now use a key path expression as a function when calling map/filter/etc.

I am proposing a rule that would identify & flag classic function calls that could be expressed using the new key path style.

**Good**
```swift
users.map(\.email)
```

```swift
users.filter(\.isAdmin)
```

**Bad**
```swift
users.map { $0.email }
```

```swift
users.filter { $0.isAdmin }
```

```swift
users.filter { user in
    return user.isAdmin 
}
```

I am not sure if this should be opt-in or not, but I would suggest that this be enabled by default.

marcelofabri:
Fixes #3082

This only catches expressions using `$0`, but I think this is a great start and should cover a lot of cases in existing codebases.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Idiomatic/KeyPathExpressionAsFunctionRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/KeyPathExpressionAsFunctionRule.swift
@@ -1,0 +1,103 @@
+import Foundation
+import SourceKittenFramework
+
+public struct KeyPathExpressionAsFunctionRule: ASTRule, OptInRule, ConfigurationProviderRule, AutomaticTestableRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "key_path_expression_as_function",
+        name: "Key Path Expression as Function",
+        description: "Prefer using key paths instead of closures when possible.",
+        kind: .idiomatic,
+        minSwiftVersion: .fiveDotTwo,
+        nonTriggeringExamples: [
+            Example("let emails = users.map(\\.email)"),
+            Example("let admins = users.filter(\\.isAdmin)"),
+            Example("let all = users.filter { _ in true }"),
+            Example("let emails = users.map { $0.email() }"),
+            Example("""
+            let violatingRanges = violatingRanges.filter { range in
+                let region = fileRegions.first {
+                    $0.contains(Location(file: self, characterOffset: range.location))
+                }
+                return region?.isRuleEnabled(rule) ?? true
+            }
+            """),
+            Example("let ones = values.filter { $0 == 1 }")
+        ],
+        triggeringExamples: [
+            Example("let emails = users.map ↓{ $0.email }"),
+            Example("let emails = users.map(↓{ $0.email })"),
+            Example("let admins = users.filter(where: ↓{ $0.isAdmin })")
+        ]
+    )
+
+    // MARK: - ASTRule
+
+    public func validate(file: SwiftLintFile,
+                         kind: SwiftExpressionKind,
+                         dictionary: SourceKittenDictionary) -> [StyleViolation] {
+        guard SwiftVersion.current >= Self.description.minSwiftVersion else {
+            return []
+        }
+
+        return violationRanges(in: file, kind: kind, dictionary: dictionary).map {
+            StyleViolation(ruleDescription: Self.description,
+                           severity: configuration.severity,
+                           location: Location(file: file, byteOffset: $0))
+        }
+    }
+
+    private func violationRanges(in file: SwiftLintFile,
+                                 kind: SwiftExpressionKind,
+                                 dictionary: SourceKittenDictionary) -> [ByteCount] {
+        guard kind == .call else {
+            return []
+        }
+
+        let closures = dictionary.substructure.compactMap { dictionary -> SourceKittenDictionary? in
+            if dictionary.isClosure {
+                return dictionary
+            }
+
+            if dictionary.isClosureArgument {
+                return dictionary.substructure.first
+            }
+
+            return nil
+        }
+
+        return closures.compactMap { dictionary -> ByteCount? in
+            guard let offset = dictionary.offset,
+                  let bodyOffset = dictionary.bodyOffset,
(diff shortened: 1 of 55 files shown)


jpsim:
I have to look into this more, but if I recall correctly, there's a good number of cases where `abc.map { $0.xyz }` compiles but `abc.map(\.xyz)` does not.

marcelofabri:
> I have to look into this more, but if I recall correctly, there's a good number of cases where `abc.map { $0.xyz }` compiles but `abc.map(\.xyz)` does not.

I took an initial stab into fixing the violations in a larger private repository and so far the only case I've found is when we were using `abc.compactMap { $0.xyz }` but `xyz` didn't return `Optional` - so the fix was to change to `map(\.xyz)`

jpsim:
One that I've found inconsistently works is `compactMap { $0 }` compiles but `compactMap(\.self)` does not. I'm not sure what that's about, but it also looks like this rule doesn't report violations for these cases anyway.

jpsim:
I've bumped up the minimum required Swift version to build SwiftLint in https://github.com/realm/SwiftLint/pull/3413 so we should be able to enable this rule for SwiftLint itself now.

jpsim:
How hard would it be to make this correctable?

Source/SwiftLintFramework/Rules/Idiomatic/KeyPathExpressionAsFunctionRule.swift
```diff
@@ -0,0 +1,100 @@
...
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "key_path_expression_as_function",
+        name: "Key Path Expression as Function",
+        description: "Prefer using key paths instead of closures when possible.",
+        kind: .idiomatic,
+        minSwiftVersion: .fiveDotTwo,
+        nonTriggeringExamples: [
+            Example("let emails = users.map(\\.email)"),
```
jpsim, line 15: Do you think it would read better if we used raw string literals here?

```suggestion
            Example(#"let emails = users.map(\.email)"#),
```

jpsim (approved):
Looks great! I'd prefer if this was a correctable rule but that shouldn't block landing this PR if it's hard to do.

marcelofabri:
> How hard would it be to make this correctable?

I've started implementing it, but we can't correct the trailing closure case because the function might have an argument (eg `first(where:)`)

marcelofabri:
I've added autocorrection for the simple case in ffcbb7c - let me know what you think!

Source/SwiftLintFramework/Rules/Idiomatic/KeyPathExpressionAsFunctionRule.swift
```diff
@@ -0,0 +1,177 @@
...
+                  let byteRange = closureDictionary.byteRange,
+                  let searchRange = file.stringView.byteRangeToNSRange(bodyRange) else {
+                return nil
+            }
+
+            // Right now, this rule only catches cases where $0 is used (instead of named parameters) for simplicity
+            let pattern =  #"\A\s*\$0(\.\w+)\b\s*\z"#
+            return file
+                .matchesAndSyntaxKinds(matching: pattern, range: searchRange)
+                .compactMap { textCheckingResult, syntaxKinds in
```
jpsim, line 134: lazy?

```suggestion
                .lazy
                .compactMap { textCheckingResult, syntaxKinds in
```

Source/SwiftLintFramework/Rules/Idiomatic/KeyPathExpressionAsFunctionRule.swift
```diff
@@ -0,0 +1,177 @@
...
+
+private extension SourceKittenDictionary {
+    var isClosure: Bool {
+        return kind.flatMap(SwiftExpressionKind.init(rawValue:)) == .closure
+    }
+
+    var isClosureArgument: Bool {
+        return kind.flatMap(SwiftExpressionKind.init(rawValue:)) == .argument &&
+            substructure.count == 1 &&
+            substructure.allSatisfy(\.isClosure)
```
jpsim, line 175: Nit:

```suggestion
            substructure[0].isClosure
```

Source/SwiftLintFramework/Rules/Idiomatic/KeyPathExpressionAsFunctionRule.swift
```diff
@@ -0,0 +1,177 @@
...
+            guard let rangeToReplace = violation.rangeToReplace else {
+                continue
+            }
+
+            if let indexRange = correctedContents.nsrangeToIndexRange(rangeToReplace) {
+                let correction = "\\" + violation.keyPathContent
+                correctedContents = correctedContents.replacingCharacters(in: indexRange, with: correction)
+                adjustedLocations.insert(rangeToReplace.location, at: 0)
+            }
+        }
```
jpsim, line 89: Nit:

```suggestion
        for violation in violations.reversed() {
            guard let rangeToReplace = violation.rangeToReplace,
                  let indexRange = correctedContents.nsrangeToIndexRange(rangeToReplace)
            else {
                continue
            }

            let correction = "\\" + violation.keyPathContent
            correctedContents = correctedContents.replacingCharacters(in: indexRange, with: correction)
            adjustedLocations.insert(rangeToReplace.location, at: 0)
        }
```

jpsim (approved):
This is great, thanks for making some simple cases correctable too!

Just a few optional nitpicky feedback notes.

jpsim:
> I took an initial stab into fixing the violations in a larger private repository and so far the only case I've found is when we were using `abc.compactMap { $0.xyz }` but `xyz` didn't return `Optional` - so the fix was to change to `map(\.xyz)`

These are also the only issues I found in Lyft's codebase, although I only checked the ones that were corrected using autocorrect.

jpsim:
I spoke too soon, I have found some false positives in Lyft's codebase in cases where the closures are optional. I don't think we can detect that in lint rules though unfortunately.

We could constrain the functions that this rule looks at, like map/compactMap/filter, that we know are much more likely to have non-optional closures.

PR #3418, opened 2020-11-09
Rule Modification: Don't trigger `unavailable_function` for a function that returns `Never`

### LINKED ISSUE #3286: Rule Modification Request: Don't trigger `unavailable_function` for a function that returns `Never`
### New Issue Checklist

### Rule modification request

   Currently, the `unavailable_function` rule warns about every function that contains a `fatalError` and does not also contain a `return` statement. In most cases, it is correct to mark these functions as unavailable, but in a few small cases, functions which always crash are _intended_ to be called.

   Apple's documentation for [the `Never` type](https://developer.apple.com/documentation/swift/never) says:
   > Use `Never` as the return type when declaring a closure, function, or method that unconditionally throws an error, traps, or otherwise does not terminate.

   A return type of `Never` is an unambiguous signal to the caller that the function is not expected to return. I believe it should also be sufficient signal to SwiftLint that the function's author is not following the "use `fatalError` to avoid having to implement a method required by a protocol" pattern, and that the function is intended to be called.

   Wouldn't trigger:
   ```swift
   func resetOnboardingStateAndCrash() -> Never {
       resetUserDefaults()
       // Crash the app to re-start the onboarding flow.
       fatalError("Onboarding re-start crash.")
   }
   ```

   Would (still) trigger:
   ```swift
   func resetOnboardingStateAndCrash() {
       resetUserDefaults()
       // Crash the app to re-start the onboarding flow.
       fatalError("Onboarding re-start crash.")
   }
   ```

3. Should the rule be configurable? If so what parameters should be configurable?

   Maybe.

   I believe the use of `Never` is a strong enough signal of developer intention that this wouldn't need to be configurable, but if this change caused the rule to not warn in cases where others felt it should, an `include_functions_returning_never` configuration boolean could be added.

4. Should the rule (change) be opt-in or enabled by default? Why?

   Personally, I think the change should be enabled by default.

   Arguments for setting the `include_functions_returning_never` configuration to `false` by default, or not having a configuration option at all:
   - As noted above, I believe a function returning `Never` is clearly intended to be called.
   - Since the `unavailable_function` rule itself is opt-in, using an opinionated default once the rule has been voluntarily enabled seems less controversial than doing do for an enabled-by-default rule.

   Arguments for setting the `include_functions_returning_never` configuration to `true` by default:
   - This would maintain the current behavior by default (for users who opt in to the `unavailable_function` rule).
   - This would err on the side of _possibly-excessive_ warnings, rather than erring on the side of _not_ warning about some cases that the developer might have wanted to be warned about.

agarmash:
Implements the change suggested in #3286.
Doesn't add any additional configuration parameters.

Resolves #3286

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,10 @@
 
 #### Enhancements
 
-* None.
+* Rule Modification: Don't trigger `unavailable_function`for a function
+  that returns `Never`.
+  [Artem Garmash](https://github.com/agarmash)
+  [#3286](https://github.com/realm/SwiftLint/issues/3286)
 
 #### Bug Fixes
 
--- a/Source/SwiftLintFramework/Extensions/QueuedPrint.swift
+++ b/Source/SwiftLintFramework/Extensions/QueuedPrint.swift
@@ -49,7 +49,6 @@ public func queuedPrintError(_ string: String) {
  the source path from the compiled binary.
  */
 public func queuedFatalError(_ string: String, file: StaticString = #file, line: UInt = #line) -> Never {
-    // swiftlint:disable:previous unavailable_function
     outputQueue.sync {
         fflush(stdout)
         let file = "\(file)".bridge().lastPathComponent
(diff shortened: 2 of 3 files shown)


jpsim (approved):
Thanks for the PR!

PR #3421, opened 2020-11-09
Implement baseline functionality

PaulTaykalo:
This is the rebased version of #2580

### DIFF
--- a/Source/SwiftLintFramework/Baseline/Baseline.swift
+++ b/Source/SwiftLintFramework/Baseline/Baseline.swift
@@ -1,0 +1,58 @@
+import Foundation
+
+public class Baseline {
+    private let baselinePath: String
+    private var baselineViolations = [BaselineViolation]()
+
+    public init(baselinePath: String) {
+        self.baselinePath = baselinePath
+    }
+
+    public func isInBaseline(violation: StyleViolation) -> Bool {
+        let baselineViolation = BaselineViolation(
+                ruleIdentifier: violation.ruleIdentifier,
+                location: violation.location.description,
+                reason: violation.reason
+        )
+        let contains = baselineViolations.contains(baselineViolation)
+        return contains
+    }
+
+    public func saveBaseline(violations: [StyleViolation]) {
+        let fileContent = violations.map(generateForSingleViolation).joined(separator: "\n")
+        let fileManager = FileManager.default
+        if !fileManager.fileExists(atPath: baselinePath) {
+            fileManager.createFile(atPath: baselinePath, contents: fileContent.data(using: .utf8))
+        }
+    }
+
+    public func readBaseline() {
+        let fileManager = FileManager.default
+        guard let fileContent = fileManager.contents(atPath: baselinePath),
+              let stringContent = String(data: fileContent, encoding: .utf8) else {
+            return
+        }
+        stringContent.enumerateLines { [weak self] line, _ in
+            guard let self = self else { return }
+            let violation = self.parseLine(line: line)
+            self.baselineViolations.append(violation)
+        }
+    }
+
+    private func parseLine(line: String) -> BaselineViolation {
+        let components = line.components(separatedBy: ";")
+        let location = components[0]
+        let reason = components[1]
+        let ruleIdentifier = components[2]
+
+        return BaselineViolation(
+                ruleIdentifier: ruleIdentifier,
+                location: location,
+                reason: reason
+        )
+    }
+
+    private func generateForSingleViolation(_ violation: StyleViolation) -> String {
+        return "\(violation.location);\(violation.reason);\(violation.ruleIdentifier)"
+    }
+}
--- a/Source/SwiftLintFramework/Baseline/BaselineViolation.swift
+++ b/Source/SwiftLintFramework/Baseline/BaselineViolation.swift
@@ -1,0 +1,13 @@
+import Foundation
+
+struct BaselineViolation: Equatable {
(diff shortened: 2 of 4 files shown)


Source/SwiftLintFramework/Baseline/Baseline.swift
```diff
@@ -0,0 +1,58 @@
+import Foundation
```
jpsim, line 1: This file doesn’t appear to use Foundation

Source/SwiftLintFramework/Baseline/Baseline.swift
```diff
@@ -0,0 +1,58 @@
...
+
+public class Baseline {
```
jpsim, line 3: The jazzy CI job is failing because these public declarations are undocumented.

dhoskins:
> This issue has been automatically marked as stale because it has not had any recent activity. Please comment to prevent this issue from being closed. Thank you for your contributions!

I'd like to prevent this issue from being closed.

bpollman:
I have been hoping for this branch/feature to make it into the main release for a very long time. I'd be happy to help. What are the chances for this to get reviewed and have a chance at getting merged.

jpsim:
Looking at this again for the first time in a while, there's a few things I'd like the author(s) to think through before we can consider something like this, which I think overall the concept is a good idea.

1. Should users be able to specify the baseline file path?
2. Should baselines apply to analyzer rules, or just lint rules?
3. Right now the baseline implementation is fairly brittle in the sense that if the line numbers change, the baseline record becomes out of date. Do we want to consider a mechanism to minimize the thrash when say a violation is moved down by a few lines because new code was added above it?
4. Should there be an easy way to reset the baseline? Right now you have to find the baseline file and delete it.
5. Can we somehow leverage the lint results cache? We're already saving lint results for faster incremental linting operations, so perhaps snapshotting that would be a new command distinct from `swiftlint lint`?
6. Unit tests, documentation and changelog entries would be important to add before considering merging something like this.

jpsim:
As I was reviewing this, I was making some edits, which I always find helpful in understand code, they may or may not be helpful to anyone who's interested in continuing this work: https://github.com/realm/SwiftLint/commits/jp-baseline

igormatos:
+1 for that feature

iandreyshev:
+1 for that feature

Source/SwiftLintFramework/Baseline/Baseline.swift
```diff
@@ -0,0 +1,58 @@
...
+                ruleIdentifier: violation.ruleIdentifier,
+                location: violation.location.description,
+                reason: violation.reason
+        )
+        let contains = baselineViolations.contains(baselineViolation)
+        return contains
+    }
+
+    public func saveBaseline(violations: [StyleViolation]) {
+        let fileContent = violations.map(generateForSingleViolation).joined(separator: "\n")
```
abhimaandunzo, line 22: would \n work on all platform ? trying point out things . like when files gets modified in git or in other os

abhimaandunzo:
do we have this baseline implementation in pipeline . I dont see the any urgency in merging this pr. Let us know what is stopping may be we can help

AlfredAfutu:
This is obviously needed as mentioned by a good number of people including myself. Is there a plan on how to proceed with this @jpsim ?

aiKrice:
Please a lot of people are waiting for that.

mildm8nnered:
I'm interested in picking this up. Not sure when I'll have time, or if I'll start from here, or from scratch right now.

> Looking at this again for the first time in a while, there's a few things I'd like the author(s) to think through before we can consider something like this, which I think overall the concept is a good idea.
> 
> 1. Should users be able to specify the baseline file path?

I think so. My use case is that I will almost always want the baseline to be derived from my main branch.

So I think something like `--write-baseline <filename>` and `--read-baseline <filename>` perhaps.

> 2. Should baselines apply to analyzer rules, or just lint rules?

I think so, but I would implement this (in CI) with separate baseline files.

> 3. Right now the baseline implementation is fairly brittle in the sense that if the line numbers change, the baseline record becomes out of date. Do we want to consider a mechanism to minimize the thrash when say a violation is moved down by a few lines because new code was added above it?

So I was thinking something like, if the file has the same violations, in the same order, ignore the violations regardless of line number.

Not sure how to handle existing violations if say one new violation appearing in a file - whether to report all the violations in the file, or to try to detect which one is "new".

For example, if my baseline violations are 

`identifier_name`, line 20

and my new violations are

`identifier name`, line 25
`identifier name`, line 30

which is the new one

Possibly ignore them if the line numbers match exactly, but otherwise report them all.

> 4. Should there be an easy way to reset the baseline? Right now you have to find the baseline file and delete it.

I think if we supported `--write-baseline <filename>` and `--read-baseline <filename>` with the same `<filename>`, that would work.

> 5. Can we somehow leverage the lint results cache? We're already saving lint results for faster incremental linting operations, so perhaps snapshotting that would be a new command distinct from `swiftlint lint`?

I would definitely take a look at that.

> 6. Unit tests, documentation and changelog entries would be important to add before considering merging something like this.

aiKrice:
very eager to have this feature !!!

mildm8nnered:
So I have a very basic implementation at #5475

Seems to work for my basic test case, but will try to work on it and to add some more docs next week.

SimplyDanny:
Closing as #5475 has superseded this PR. With that being merged, we now have a baseline functionality experimentally available in SwiftLint.

PR #3422, opened 2020-11-10
Fix false positives in inclusive_language rule

### LINKED ISSUE #3415: inclusive_language false positives: Triggering terms as substrings of identifiers
### New Issue Checklist

### Describe the bug

We have a project that deals with payments provided by Mastercard and as such it is common to have code such as:

`let mastercard = newWalletItem(for: .mastercard)`

This triggers a violation of the `inclusive_language` rule as the identifier contains a triggering term as a substring.

Is it possible to the break triggering terms in identifiers using camel case so `let masterThread` is a violation but `let mastercard` is not?

N/A

### Environment

* SwiftLint version: 0.41.0
N/A - Default rule

  N/A
Xcode 12.1

```swift
// This triggers a violation:
`let mastercard = newWalletItem(for: .mastercard)`
```

```swift
public enum WalletItemType: String {
    case visa = "VISA"
    // This triggers a violation:
    case mastercard = "MASTERCARD"
}
```

daltonclaybrook:
This PR resolves #3415 

A noteworthy implication of this PR is that strings like `BlackList` and `whiteList` no longer trigger violations because these use camel case and are thus treated as separate terms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,7 +57,9 @@
 
 #### Bug Fixes
 
-* None.
+* Fix false positives in `inclusive_language` rule.  
+  [Dalton Claybrook](https://github.com/daltonclaybrook)
+  [#3415](https://github.com/realm/SwiftLint/issues/3415)
 
 ## 0.41.0: World’s Cleanest Voting Booth
 
--- a/Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/String+SwiftLint.swift
@@ -100,4 +100,29 @@ extension String {
             $1 == character ? $0 + 1 : $0
         })
     }
+
+    /// Return an array of strings by splitting the receiver into
+    /// its camel-case components.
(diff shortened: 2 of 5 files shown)


jpsim:
According to OSSCheck’s results posted above, it also seems like “Whitelisted” no longer trigger violations. Should that be added to the set of default terms?

daltonclaybrook:
@jpsim Should we add "Whitelisting" as well? Is there possibly another solution we haven't thought of?

jpsim:
What if you made the terms regex patterns instead, like the way custom rules work? That way you could do `[Ww]hitelist[\w]+` and cover everything you want, but just `[Mm]aster`. Still using the tokenization approach of splitting identifiers by camelCase though.

Regular expressions would also be more useful when configuring the rule.

Or we stick with adding a few permutations of the words ad-hoc.

daltonclaybrook:
I like the regex approach. I will tinker with this.

daltonclaybrook:
Closing this PR in favor of #3439

Issue #3424, opened 2020-11-10
Codeclimate report format for Gitlab MR Widget


jkroepke:
Gitlab make the Code Quality MR Widget available to the free edition.

https://docs.gitlab.com/ee/user/project/merge_requests/code_quality.html

At this time, gitlab only support the codeclimate format as source for the Code Quality MR Widget.

https://docs.gitlab.com/ee/user/project/merge_requests/code_quality.html#implementing-a-custom-tool

It's possible to provide such a report format? I guess, it could be done by copy the json reporter and adjust the fields.

jpsim:
> I guess, it could be done by copy the json reporter and adjust the fields.

That sounds like the right way forward to me. Let us know if you need any help.

jkroepke:
Thanks I opened https://github.com/realm/SwiftLint/issues/3424

PR #3425, opened 2020-11-10
Add codeclimate reporter

jkroepke:
Fixes: https://github.com/realm/SwiftLint/issues/3424

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,7 +61,7 @@
 * Add `codeclimate` reporter to generate JSON reports in codeclimate 
   format. Could be used for Gitlab Code Quality MR Widget.
 * Add `codeclimate` reporter to generate JSON reports in codeclimate
-  format. Could be used for Gitlab Code Quality MR Widget.  
+  format. Could be used for GitLab Code Quality MR Widget.  
   [jkroepke](https://github.com/jkroepke)
   [#3424](https://github.com/realm/SwiftLint/issues/3424)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,8 @@
   [#3233](https://github.com/realm/SwiftLint/issues/3233)
 * Add `codeclimate` reporter to generate JSON reports in codeclimate 
   format. Could be used for Gitlab Code Quality MR Widget.
+* Add `codeclimate` reporter to generate JSON reports in codeclimate
+  format. Could be used for Gitlab Code Quality MR Widget.  
   [jkroepke](https://github.com/jkroepke)
   [#3424](https://github.com/realm/SwiftLint/issues/3424)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,8 +58,7 @@
 * Add `comment_spacing` rule.  
   [Noah Gilmore](https://github.com/noahsark769)
   [#3233](https://github.com/realm/SwiftLint/issues/3233)
-* Add `codeclimate` reporter to generate JSON reports in codeclimate 
-  format. Could be used for Gitlab Code Quality MR Widget.
+
 * Add `codeclimate` reporter to generate JSON reports in codeclimate
   format. Could be used for GitLab Code Quality MR Widget.  
   [jkroepke](https://github.com/jkroepke)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,10 @@
 * Add `comment_spacing` rule.  
   [Noah Gilmore](https://github.com/noahsark769)
   [#3233](https://github.com/realm/SwiftLint/issues/3233)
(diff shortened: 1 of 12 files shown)


jkroepke:
Please Review.

One question: The path inside the report is escaped. This is expected?

```
  {
    "check_name" : "identifier_name",
    "description" : "Variable name should be between 3 and 40 characters long: 'to'",
    "engine_name" : "SwiftLint",
    "fingerprint" : "cc6071d559206b5f6487194ee6b3776c",
    "location" : {
      "lines" : {
        "begin" : 95
      },
      "path" : "\/Users\/jkr\/Documents\/app\/ios\/Pods\/lottie-ios\/lottie-swift\/src\/Private\/Utility\/Extensions\/MathKit.swift"
    },
    "severity" : "MINOR",
    "type" : "issue"
  },
```

jpsim:
> This is expected?

I think that's more of a question for you, what does the Codeclimate format expect?

jkroepke:
I couldn't find the spec yet.

Looking at the gitlab example, the path is not escaped:
https://docs.gitlab.com/ee/user/project/merge_requests/code_quality.html#implementing-a-custom-tool

I thought swiftlint escape the slashes by default. But it looks like a swift feature. (First time doing something with swift...)

CHANGELOG.md
```diff
@@ -55,6 +55,11 @@
   `unused_import` rule.  
   [Keith Smiley](https://github.com/keith)
 
+* Add `codeclimate` reporter to generate JSON reports in codeclimate 
+  format. Could be used for Gitlab Code Quality MR Widget.
+  [jkroepke](https://github.com/jkroepke)
+  [#3424](https://github.com/realm/SwiftLint/issues/3424)
```
jpsim, line 61: As per the format described in CONTRIBUTING.md

```suggestion
* Add `codeclimate` reporter to generate JSON reports in codeclimate
  format. Could be used for Gitlab Code Quality MR Widget.  
  [jkroepke](https://github.com/jkroepke)
  [#3424](https://github.com/realm/SwiftLint/issues/3424)
```

Source/SwiftLintFramework/Reporters/CodeClimateReporter.swift
```diff
@@ -0,0 +1,65 @@
...
+import CommonCrypto
+#else
+import CryptoSwift
+#endif
+import Foundation
+import SourceKittenFramework
+
+#if canImport(CommonCrypto)
+private extension String {
+    func md5() -> String {
```
jpsim, line 11: Rather than duplicate this, could you move this to a separate file and make it `internal`?

jpsim:
Looks good other than the md5 code duplication.

Source/SwiftLintFramework/Reporters/CodeClimateReporter.swift
jkroepke: Done. I hope I catch the Swift coding guidelines.

Source/SwiftLintFramework/Extensions/String+md5.swift
```diff
@@ -0,0 +1,20 @@
+#if canImport(CommonCrypto)
+import CommonCrypto
+#else
+import CryptoSwift
+#endif
+import Foundation
+
+#if canImport(CommonCrypto)
+extension String {
```
jpsim, line 9: Aren't all these other imports unused?

```suggestion
#if canImport(CommonCrypto)
import CommonCrypto

extension String {
```

jkroepke: I thought `CryptoSwift` is just fallback, if `CommonCrypto` is not available.

jpsim: Correct, but CryptoSwift already comes with `String.md5()`, so what this code does is add the same convenience method to CommonCrypto.

jkroepke: I just copied the code from `Source/SwiftLintFramework/Extensions/Configuration+Cache.swift`, I can't decide if the code could be throw away.

jpsim: That file uses CryptoSwift but this file doesn’t.

jkroepke: I suggest to split this in an own issue/pr.

jkroepke:
Rebased. 

Could we merge is feature soon? Looks like the CHANGELOG.md gets conflicted fast if any other PR is merged.

I still suggest to split the discussion about the crypto code in a separate PR.

jpsim:
Thanks for rebasing and keeping this up to date. I pushed up the import changes I had mentioned, fixed the changelog entry and will be merging once CI passes. Thanks for your contribution @jkroepke!

jkroepke:
Thanks!

PR #3426, opened 2020-11-10
Rename OverridenSuperCallConfiguration to fix typo

bricker:
Renames `OverridenSuperCallConfiguration` to
`OverriddenSuperCallConfiguration`

BREAKING CHANGE: As `OverridenSuperCallConfiguration` is a
publicly-exposed struct, this rename should be considered breaking.


jpsim (approved):
Thanks for fixing!

PR #3427, opened 2020-11-11
[WIP] New rule: Redundant Bool comparison

### LINKED ISSUE #3420: Rule Request: Discourage comparing Bool value with `true` or `false`
### New Issue Checklist

### New rule request

The usage of `==` or `!=` operators for checking if the Bool value is `true` or `false` should be discouraged in favour of using the Bool value itself or the one with negating operator.

#### 1. Why should this rule be added? Share links to existing discussion about what the community thinks about this.

Since Bool can only have 2 values, there's no need to compare it with constant values, like:
```swift
var isEnabled = true
if isEnabled == true { /* do stuff */ } 
```

The only exception is checking an optional Bool value, although this case is already covered in [Discouraged Optional Boolean](https://realm.github.io/SwiftLint/discouraged_optional_boolean.html):
```swift
var hasPermission: Bool? = true
if hasPermission == true { /* do stuff */ } 
```

Another point is that regarding [Swift code style](https://github.com/raywenderlich/swift-style-guide#naming), "boolean types should read like assertions".
Let's compare 2 examples:
* `if switch.isActive { updateAppearance() }` is naturally perceived as "If the switch is active, update the appearance";
* `if switch.isActive == true { updateAppearance() }` is a bit more cluttered and makes it harder to quickly grasp the intension of the code.

Also, [a recent forum thread shows](https://forums.swift.org/t/boolean-comparison-causes-extremely-slow-compilation/41609) that Bool comparison is quite slower than using the value itself during compilation.

#### 2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

**Triggering examples:**

```swift
var isEnabled = true
if ↓isEnabled == true { /* do stuff */ } 
```

Don't forget the [Yoda notation](https://en.wikipedia.org/wiki/Yoda_conditions):
```swift
var isEnabled = true
if ↓true == isEnabled { /* do stuff */ } 
```

**Non-triggering examples:**

```swift
var hasPermission: Bool? = true
if hasPermission == true { /* do stuff */ } 
```

```swift
var isEnabled = true
if isEnabled { /* do stuff */ } 
```

#### 3. Should the rule be configurable, if so what parameters should be configurable?

I don't think so, there's basically nothing to configure.

#### 4. Should the rule be opt-in or enabled by default? Why?

Due to perceivable performance drop, it worth enabling this rule by default

agarmash:
Implements the rule described in #3420.
I decided not to check if the value in the condition expression is optional Bool for the following reasons:
* the optional Bool case is already covered in `discouraged_optional_boolean`;
* it's not possible to easily check types of operands in the expression, so I decided that it doesn't worth additional time and effort.

Resolves #3420

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,10 @@
   [Noah Gilmore](https://github.com/noahsark769)
   [#3233](https://github.com/realm/SwiftLint/issues/3233)
 
+* Add `redundant_bool_comparison` rule.  
+  [Artem Garmash](https://github.com/agarmash)
+  [#3420](https://github.com/realm/SwiftLint/issues/3420)
+
 #### Bug Fixes
 
 * None.
--- a/Source/SwiftLintFramework/Models/PrimaryRuleList.swift
+++ b/Source/SwiftLintFramework/Models/PrimaryRuleList.swift
@@ -142,6 +142,7 @@ public let primaryRuleList = RuleList(rules: [
     RawValueForCamelCasedCodableEnumRule.self,
     ReduceBooleanRule.self,
     ReduceIntoRule.self,
+    RedundantBoolComparisonRule.self,
     RedundantDiscardableLetRule.self,
     RedundantNilCoalescingRule.self,
     RedundantObjcAttributeRule.self,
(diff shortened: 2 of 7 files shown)


jpsim:
Thanks for the PR! I'd rather this not be enabled by default nor be enabled in SwiftLint itself.

Personally, I don't think I'd turn this on in any of the projects I work on because of the risk of false positives affecting optionals.

agarmash:
Is there a way to check operand types inside the expression? SourceKitten can provide such information, but is it possible to get it in SwiftLint?
Also, speaking about optional bools, I'd rather check them differently. The code like `if flag == true {}` can be misleading, since it doesn't show explicitly whether the `flag` is optional or not and therefore hides the possible 3rd state, when the code like `if flag ?? false {}` or `if let flag = flag, flag {}` clearly shows the whole picture.

jpsim:
Yes you can get the post-typchecked AST in AnalyzerRules, but those run under a completely different SwiftLint mode.

agarmash:
Great, will check it and modify the rule to perform type-checking.  

And what about my point about dealing with optional bools? Don't you think their comparison should also be reported?

PR #3431, opened 2020-11-16
Add support for testing Examples with configuration

keith:
This was extracting from another change that no longer needed it. It
should be helpful for anyone testing custom configuration with their
non-analyzer rules.

I added a random example of something that wasn't previously tested as
an example.


keith:
You pushed this commit to my PR but we ended up taking that a different direction

jpsim (approved):
Thank you

PR #3432, opened 2020-11-16
Remove @IBOutlet and @IBInspectable from UnusedDeclarationRule

keith:
The original implementation wasn't tested so it regressed.

https://github.com/realm/SwiftLint/pull/3184

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,7 +74,9 @@
 
 #### Bug Fixes
 
-* None.
+* Remove `@IBOutlet` and `@IBInspectable` from UnusedDeclarationRule.  
+  [Keith Smiley](https://github.com/keith)
+  [#3184](https://github.com/realm/SwiftLint/issues/3184)
 
 ## 0.41.0: World’s Cleanest Voting Booth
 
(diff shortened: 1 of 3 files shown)


jpsim (approved):

jpsim:
Thinking this through some more after seeing in practice the kinds of declarations this found in a real iOS codebase, I understand the rationale for reporting unused `@IBOutlet` declarations since I think it's safe to remove those if they're never referenced in code, but it seems quite common for `@IBInspectable` declarations to only ever be referenced from Interface Builder, and removing those when they're not referenced from other Swift code just breaks them.

For example:

```swift
class MyView: UIView {
    @IBInspectable private var innerPaddingWidth: CGFloat {
        set { self.backgroundView.innerPaddingWidth = newValue }
        get { return self.backgroundView.innerPaddingWidth }
    }
}
```

If `MyView.innerPaddingWidth` is configured in Interface Builder but never referenced in code, it's still not safe to remove.

keith:
Ah yes that's a good use case where we need to keep them. The goal was to remove ones that were not using custom get + set since those could correctly be diagnosed as unused. Maybe we can use those having custom implementations as an indication to ignore them for this?

jpsim:
> Maybe we can use those having custom implementations as an indication to ignore them for this?

Yes I suppose that could work. Is that something you think you can take on?

keith:
Fixed here https://github.com/realm/SwiftLint/pull/3458

PR #3437, opened 2020-11-19
[Makefile] Build `swiftlint` as a universal binary on macOS

norio-nomura:
SwiftPM 5.3 uses the `XCBuild.framework` to generate universal binaries when it receives multiple `--arch` options.
In that case, `--show-bin-path` returns `.build/apple/Products/Release`.


jpsim:
Yay, thank you!!

PR #3439, opened 2020-11-20
Fix false positives in inclusive_language rule

### LINKED ISSUE #3415: inclusive_language false positives: Triggering terms as substrings of identifiers

daltonclaybrook:
This PR resolves #3415 by adding "allowed terms" to the rule configuration. If a disallowed term is found to overlap with an allowed term, the allowed term "wins." The only default allowed term at the moment is "mastercard"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,11 +72,14 @@
 * Add `non_private_xctest_member` rule.  
   [Keith Smiley](https://github.com/keith)
 
-#### Bug Fixes
-
-* Fix false positives in `inclusive_language` rule.  
+* Add an `override_allowed_terms` configuration parameter to the
+  `inclusive_language` rule, with a default value of `mastercard`.  
   [Dalton Claybrook](https://github.com/daltonclaybrook)
   [#3415](https://github.com/realm/SwiftLint/issues/3415)
+
+#### Bug Fixes
+
+* None.
 
 ## 0.41.0: World’s Cleanest Voting Booth
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,7 +74,9 @@
 
 #### Bug Fixes
 
-* None.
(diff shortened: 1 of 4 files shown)


daltonclaybrook:
@jpsim let me know what you think of this approach vs the old camel-case/regex strategy (#3422). I think this is a lot simpler.

jpsim:
Yes this is a lot simpler. Will take a closer look.

CHANGELOG.md
```diff
@@ -74,7 +74,9 @@
 
 #### Bug Fixes
 
-* None.
+* Fix false positives in `inclusive_language` rule.
```
jpsim, line 77: Technically, "mastercard" wasn't a false positive but rather just a commonly used name that we're special-casing it here. I'd also say this is less of a bug fix and more of an enhancement.

I'll push up some changes to the changelog to reflect this.

jpsim (approved):
Thanks for addressing this! 🙌

apps4everyone:
is there a way to also allow UPPERCASED version in the swiftlint.yml?
e.g: case masterCard = "MASTERCARD"

PR #3445, opened 2020-11-23
Fix type_contents_order rule treats methods starting with init like initializers

rabbitinspace:
When `type_contents_order` rule sees a method like `init.+(...)` it thinks that it's a `TypeContent.initializer` instead of `TypeContent.otherMethod` because of the `starts(with: "init")` check. Instead, we need to compare it with one of the following strings: `"init("`, `"init<"`, `"init?"` or `"init!"`, as per [language reference](https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#grammar_initializer-declaration) to make the correct decision. 

I think there's the same issue with the `deinit` check, but I didn't touch it in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,7 +86,8 @@ Improve documentation for multiple configuration files.
 
 #### Bug Fixes
 
-* None.
+* Fix methods starting `init` are treated like initializers.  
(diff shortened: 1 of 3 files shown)


rabbitinspace:
Any updates on this, anyone?

PR #3450, opened 2020-12-01
Add option to configure SortedImportsRule to sort @testable imports on the top or on the bottom

MaxHaertwig:
Hi :)

What do you think about this configuration option? I'm open to rename the options `top` and `bottom` to `first` and `last`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,9 @@
 
 #### Experimental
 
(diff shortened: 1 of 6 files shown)


agarmash:
I would rename `testable_imports` to `testable_imports_position`, so it makes a clear sense on the first glance.
Besides that - great job!

Issue #3456, opened 2020-12-04
Rule Request: Weak delegate rule should allow @UIApplicationDelegateAdaptor


jrturton:
### New Issue Checklist

- [X] Updated SwiftLint to the latest version
- [X] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

The weak delegate rule fires on the following declaration, which is not appropriate:

```
@UIApplicationDelegateAdaptor(AppDelegate.self) var appDelegate
```

I have begun making the changes to implement this but I want to check that I'm not missing something - there doesn't seem to be a way of detecting a property wrapper specifically from the SourceKittenDictionary, and I have to do it by reading the attributes out using the offsets? 

If this is a question better asked as part of a PR submission then that's https://github.com/realm/SwiftLint/pull/3457

PR #3457, opened 2020-12-04
Add exception for UIApplicationDelegateAdaptor

jrturton:
As discussed in https://github.com/realm/SwiftLint/issues/3456 

I'm unsure if I am detecting the property wrapper the correct way, this is my first time working with this code 😄.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,10 @@
 
 #### Enhancements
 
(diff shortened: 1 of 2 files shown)


jpsim:
Thanks for the contribution, looks good! I think there's also a way to check the attribute's name, but I'm not completely sure.

PR #3458, opened 2020-12-07
Fix @IBInspectable unused declarations with get/set

keith:
This fixes an issue where `@IBInspectables` that have both a custom
getter and setter were marked as dead if they weren't referenced in code
even though in that case they can still have behavior if they are used
from IB.


jpsim:
Fix looks good, thank you! Can you remove the changelog entry you added since the bug didn't ship in a production release? I'm still trying to come up with a good process for having pre-release versions, sorry for the confusion.

keith:
Np, done

jpsim:
CI is failing because the `declaredUSR(...)` spans too many lines, and in this case I think it's worth splitting up rather than silencing the violation.

keith:
Broke out the 2 statements about ignoring the current entity, happy to do something more if you have something in mind, it's a nice improvement IMO but not much

jpsim:
Perfect.

jpsim (approved):

PR #3460, opened 2020-12-09
Don't log a warning if no configuration files were explicitly specified

jpsim:
And none was found at the default location of `.swiftlint.yml`.

### DIFF
--- a/Source/SwiftLintFramework/Extensions/Configuration+FileGraphSubtypes.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+FileGraphSubtypes.swift
@@ -72,7 +72,7 @@ internal extension Configuration.FileGraph {
 
         private func read(at path: String) throws -> String {
             guard !path.isEmpty && FileManager.default.fileExists(atPath: path) else {
-                throw ConfigurationError.fileNotFound(path: path)
+                throw ConfigurationError.generic("File \(path) can't be found.")
             }
 
             return try String(contentsOfFile: path, encoding: .utf8)
--- a/Source/SwiftLintFramework/Extensions/Configuration+FileGraphSubtypes.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+FileGraphSubtypes.swift
@@ -22,11 +22,12 @@ internal extension Configuration.FileGraph {
         }
 
         private let originalRootDirectory: String
+        let isInitialVertix: Bool
         private(set) var filePath: FilePath
         private(set) var configurationString: String = ""
         private(set) var configurationDict: [String: Any] = [:]
 
-        init(string: String, rootDirectory: String) {
+        init(string: String, rootDirectory: String, isInitialVertix: Bool) {
             originalRootDirectory = rootDirectory
             if string.hasPrefix("http://") || string.hasPrefix("https://") {
                 originalRemoteString = string
@@ -37,19 +38,22 @@ internal extension Configuration.FileGraph {
                     path: string.bridge().absolutePathRepresentation(rootDirectory: rootDirectory)
                 )
             }
+            self.isInitialVertix = isInitialVertix
         }
 
-        init(originalRemoteString: String?, originalRootDirectory: String, filePath: FilePath) {
+        init(originalRemoteString: String?, originalRootDirectory: String, filePath: FilePath, isInitialVertix: Bool) {
             self.originalRemoteString = originalRemoteString
             self.originalRootDirectory = originalRootDirectory
             self.filePath = filePath
+            self.isInitialVertix = isInitialVertix
         }
 
         internal func copy(withNewRootDirectory rootDirectory: String) -> Vertix {
             let vertix = Vertix(
                 originalRemoteString: originalRemoteString,
                 originalRootDirectory: rootDirectory,
-                filePath: filePath
+                filePath: filePath,
+                isInitialVertix: isInitialVertix
             )
             vertix.configurationString = configurationString
             vertix.configurationDict = configurationDict
@@ -72,7 +76,9 @@ internal extension Configuration.FileGraph {
 
         private func read(at path: String) throws -> String {
             guard !path.isEmpty && FileManager.default.fileExists(atPath: path) else {
-                throw ConfigurationError.generic("File \(path) can't be found.")
+                throw isInitialVertix ?
+                    ConfigurationError.initialFileNotFound(path: path) :
+                    ConfigurationError.generic("File \(path) can't be found.")
             }
 
             return try String(contentsOfFile: path, encoding: .utf8)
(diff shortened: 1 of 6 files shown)


fredpi:
@jpsim I believe that your current implementation of this PR (d75b9c45df7ae2f58ca169e91950c086b40e9e6c) produces a new issue of "good" warnings not being generated.

Consider the case where there exists a `.swiftlint.yml`, but the user formatted the yml file badly or put a `child_config` reference that doesn't exist. If the user runs SwiftLint without explicitly specifying the `.swiftlint.yml` file, SwiftLint will now simply ignore the user's `.swiftlint.yml` without giving a warning or at least printing something.

It would be good to differentiate between a `file is not there` error and the other errors, as you did in a previous commit. Only when the error is a `.swiftlint.yml file is not there` error, SwiftLint shouldn't create a warning (but maybe still print about what it is doing, but that would require bringing the silent flag back to the Configuration initializer that I previously removed because it wasn't needed).

In all other cases, **including the case where a file other than the original `.swiftlint.yml` file can't be found**, SwiftLint should give a warning.

### Implementation

To implement this, I would add a new property to the `Vertix` class called `isInitialVertix` that is set to true for all vertices created in the the `FileGraph.init(commandLineChildConfigs:rootDirectory:ignoreParentAndChildConfig)` initializer. That way we know that a vertix is an initially referenced file in contrast to a file reference via another file. We can use this knowledge in the `Vertix.read(at:)` method where you previously already put your new error handling. Here we can handle a new error called `initialFileNotFound` or something like that, only if the vertix is an initial vertix.

jpsim:
Thanks for your feedback @fredpi, I agree with your assessment. I've pushed up ab700a3 with the approach you suggested and would appreciate another review if you have time 😄

Source/SwiftLintFramework/Extensions/Configuration+FileGraphSubtypes.swift
```diff
@@ -95,15 +101,6 @@ internal extension Configuration.FileGraph {
     struct Edge: Hashable {
         var parent: Vertix!
         var child: Vertix!
-
-        internal static func == (lhs: Edge, rhs: Edge) -> Bool {
```
jpsim, line 99: We can use the compiler-synthesized implementations for `==` and `hash(into:)` since they're identical to these hand-rolled versions.

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -238,6 +233,23 @@ public struct Configuration {
     }
 }
 
+func errorMessage(for error: Error, hasCustomConfigurationFiles: Bool) -> String? {
```
fredpi, line 236: It would be good to have this method marked as `private`, but that's an optional change.

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -238,6 +233,23 @@ public struct Configuration {
...
+    switch error {
+    case let ConfigurationError.initialFileNotFound(path):
+        if hasCustomConfigurationFiles {
+            return "SwiftLint Configuration Error: Could not read file at path: \(path)"
+        } else {
+            return nil
```
fredpi, line 242: As the semantics of this function are probably a bit unclear at first glance, it would be great to have a comment above the `return nil` line explaining why `nil` is returned, e. g. `// The initial configuration file wasn't found, but the user didn't explicitly specify one -> don't handle as error`

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -213,16 +213,11 @@ public struct Configuration {
...
-
-            case let YamlParserError.yamlParsing(message):
-                errorString = "YML Parsing Error: \(message)"
-
-            default:
-                errorString = "Unknown Error"
+            guard let errorString = errorMessage(for: error, hasCustomConfigurationFiles: hasCustomConfigurationFiles)
+            else {
+                // Fall back to default
+                self.init(rulesMode: rulesMode, cachePath: cachePath)
```
fredpi, line 219: I'd prefer if SwiftLint would at least print about what it is doing here (not as a warning, but as a regular print). That would require bringing back the `silent` flag as a parameter of this initializer.

However, this print is not so important and maybe it's better to not increase complexity by adding this (`silent`-conditional) print. Do what you think is the better solution ;)

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -213,16 +213,11 @@ public struct Configuration {
...
-                errorString = "SwiftLint Configuration Error: \(message)"
-
-            case let YamlParserError.yamlParsing(message):
-                errorString = "YML Parsing Error: \(message)"
-
-            default:
-                errorString = "Unknown Error"
+            guard let errorString = errorMessage(for: error, hasCustomConfigurationFiles: hasCustomConfigurationFiles)
+            else {
+                // Fall back to default
```
fredpi, line 218: Maybe `// Error isn't an actual error → silently fall back to default` or just `// Silently fall back to default`

Source/SwiftLintFramework/Models/Configuration.swift
fredpi: Also, maybe there is a better name of this function? In principle, it decides which error is an actual error and then only maps the actual errors to strings. Maybe `flatMapToErrorMessage(error:hasCustomConfigurationFiles:)`. That's also not that great but at least includes a hint to a potential `nil` return value.

fredpi (approved):
@jpsim Looks good to me! I just made some small suggestions, feel free to ignore them if you disagree or if you think time's not worth addressing them.

Source/SwiftLintFramework/Models/Configuration.swift
jpsim: I'd rather SwiftLint not print at all in the common case that you want to lint a file or files without a configuration.

jpsim:
Thanks again for your feedback @fredpi, you highlighted a few areas worth improving. I hope you like the direction I took based on your comments: cb3eed5

jpsim:
Not adding a changelog entry because the warning log hasn't yet shipped in an official release.

fredpi (approved):
@jpsim Looks good to me – the new `FileGraphInitializationResult` type certainly helps with understanding the semantics!

Apart from the little missing word in the comment, nothing is holding this back from getting merged IMHO. 👍

Source/SwiftLintFramework/Models/Configuration.swift
```diff
@@ -214,15 +214,16 @@ public struct Configuration {
...
-                errorString = "YML Parsing Error: \(message)"
-
-            default:
-                errorString = "Unknown Error"
+            let initializationResult = FileGraphInitializationResult(
+                error: error, hasCustomConfigurationFiles: hasCustomConfigurationFiles
+            )
+            switch initializationResult {
+            case .initialImplicitFileNotFound:
+                // Silently back to default
```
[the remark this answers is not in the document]
jpsim: Oops.

PR #3462, opened 2020-12-11
Fix false positives on unused_declaration for outlets with accessors

jpsim:


jpsim:
We forgot to do this in https://github.com/realm/SwiftLint/pull/3458

keith (approved):

PR #3465, opened 2020-12-11
Migrate from Commandant to swift-argument-parser

jpsim:
This is a breaking change because it changes the command line syntax slightly.

For the most part the breaking changes are all to make the syntax more unix compliant and intuitive to use. For example, commands such as `swiftlint --help` or `swiftlint -h` now work as expected.

The help output from various commands has greatly improved as well.

The most significant breaking change was that `swiftlint autocorrect` was removed in favor of `swiftlint --fix`.

Previous commands should continue to work temporarily to help with the transition. Please let us know if there's a command that no longer works and we'll attempt to add a bridge to help with its transition.


.swiftlint.yml
```diff
@@ -36,7 +36,6 @@ opt_in_rules:
   - last_where
   - legacy_multiple
   - legacy_random
-  - let_var_whitespace
```
jpsim, line 39: This rule doesn't appear to work well with property wrappers.

jpsim:
This PR is best reviewed commit-by-commit. The command files were renamed in 3a39d8b so that the next commit can show the changes within the file migrating to swift-argument-parser more clearly.

Source/swiftlint/Commands/Lint.swift
```diff
@@ -0,0 +1,54 @@
...
+        var noCache = false
+        @Flag(help: "Run all rules, even opt-in and disabled ones, ignoring `only_rules`.")
+        var enableAllRules = false
+        @Argument(help: pathsArgumentDescription(for: .lint))
+        var paths = [String]()
+
+        mutating func run() throws {
+            let options = LintOrAnalyzeOptions(
+                mode: .lint,
+                paths: paths + [path ?? ""],
```
keith, line 27: looks like this logic differs a bit from before?

```swift
            let allPaths: [String]
            if !path.isEmpty {
                allPaths = [path]
            } else {
                allPaths = paths
            }
```

wouldn't this mean if you passed explicit paths you would also get all files from PWD as well as them?

SwiftLintFramework.podspec
```diff
@@ -10,6 +10,6 @@ Pod::Spec.new do |s|
   s.source_files        = 'Source/SwiftLintFramework/**/*.swift'
   s.swift_versions      = ['5.2', '5.3']
   s.pod_target_xcconfig = { 'APPLICATION_EXTENSION_API_ONLY' => 'YES' }
-  s.dependency            'SourceKittenFramework', '~> 0.30.1'
+  s.dependency            'SourceKittenFramework', '~> 0.31.0'
```
keith, line 13: idk how this whole thing works with cp, but do you not need to specify swift-argument-parser here? I see we weren't doing commandant before either so maybe we're getting these transitively?

Source/swiftlint/Helpers/LintOrAnalyzeArguments.swift
```diff
@@ -0,0 +1,41 @@
...
+
+struct LintOrAnalyzeArguments: ParsableArguments {
+    @Option(help: "The path to one or more SwiftLint configuration files, evaluated as a parent-child hierarchy.")
+    var config = [String]()
+    @Flag(name: [.long, .customLong("autocorrect")], help: "Correct violations whenever possible.")
+    var fix = false
+    @Flag(help: "Use an alternative algorithm to exclude paths for `excluded`, which may be faster in some cases.")
+    var useAlternativeExcluding = false
+    @Flag(help: "Read SCRIPT_INPUT_FILE* environment variables as files.")
+    var useScriptInputFiles = false
```
keith, line 13: I wonder if it's worth trying to make any of these mutually exclusive in the future

keith (approved):

SwiftLintFramework.podspec
jpsim: This is a podspec for `SourceKittenFramework`, which is the framework that powers SwiftLint. It doesn't include the CLI frontend, which is in the `swiftlint` target. So we don't need SwiftyTextTable or swift-argument-parser.

Source/swiftlint/Commands/Lint.swift
jpsim: Good catch. Previously `path` defaulted to `""` which is treated as PWD. So without this, not specifying `--path` or `--paths` means that nothing gets linted. So I need to handle that case now, but I'll fix this bug.

Source/swiftlint/Helpers/LintOrAnalyzeArguments.swift
jpsim: Yes that'd be nice. Nothing stopping me from doing this now either. There's also `--reporter` that I'd like to make an enum instead of a string.

Source/swiftlint/Commands/Lint.swift
jpsim: Should be fixed in 9adb2dea.

keith (approved):

Source/swiftlint/Helpers/LintOrAnalyzeArguments.swift
jpsim: Unfortunately, I can't make `useScriptInputFiles` mutually exclusive with options outside this option group.

This is nice though:

```
swiftlint --strict --lenient
Error: Value to be set with flag '--lenient' had already been set with flag '--strict'
Usage: swiftlint lint <options>
  See 'swiftlint lint --help' for more information.
```

jpsim: It doesn't look like using an enum for `--reporter` gets me anything, but it makes the error reporting worse and adds more cruft, so I'm keeping it a string for now.

jpsim:
Pushed directly to master after releasing [0.42.0](https://github.com/realm/SwiftLint/releases/tag/0.42.0).

PR #3467, opened 2020-12-16
Fix typos in configuration options for file_name rule

advantis:
I was wondering why it doesn't work. It turns out that the configuration options in the documentation were listed in camel case, where they should be in snake case.


fredpi:
Thanks @advantis 👍

Issue #3472, opened 2020-12-18
Invalid rule identifier with nested configs and custom rules on SwiftLint 0.42.0


marcelofabri:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

When using custom rules + nested configs, SwiftLint is not recognizing the custom rule identifier if it's referenced in `disabled_rules`.

Interestingly, it's actually disabling the rule.

##### Complete output when running SwiftLint, including the stack trace and command used

**SwiftLint 0.42**

```bash
$ swiftlint 
Linting Swift files in current working directory
warning: 'no_hiding_in_strings' is not a valid rule identifier
Valid rule identifiers:
anyobject_protocol
array_init
attributes
block_based_kvo
class_delegate_protocol
closing_brace
closure_body_length
closure_end_indentation
closure_parameter_position
closure_spacing
collection_alignment
colon
comma
comment_spacing
compiler_protocol_init
computed_accessors_order
conditional_returns_on_newline
contains_over_filter_count
contains_over_filter_is_empty
contains_over_first_not_nil
contains_over_range_nil_comparison
control_statement
convenience_type
custom_rules
cyclomatic_complexity
deployment_target
discarded_notification_center_observer
discouraged_direct_init
discouraged_object_literal
discouraged_optional_boolean
discouraged_optional_collection
duplicate_enum_cases
duplicate_imports
dynamic_inline
empty_collection_literal
empty_count
empty_enum_arguments
empty_parameters
empty_parentheses_with_trailing_closure
empty_string
empty_xctest_method
enum_case_associated_values_count
expiring_todo
explicit_acl
explicit_enum_raw_value
explicit_init
explicit_self
explicit_top_level_acl
explicit_type_interface
extension_access_modifier
fallthrough
fatal_error_message
file_header
file_length
file_name
file_name_no_space
file_types_order
first_where
flatmap_over_map_reduce
for_where
force_cast
force_try
force_unwrapping
function_body_length
function_default_parameter_at_end
function_parameter_count
generic_type_name
ibinspectable_in_extension
identical_operands
identifier_name
implicit_getter
implicit_return
implicitly_unwrapped_optional
inclusive_language
indentation_width
inert_defer
is_disjoint
joined_default_parameter
large_tuple
last_where
leading_whitespace
legacy_cggeometry_functions
legacy_constant
legacy_constructor
[...]

Issue #3468, opened 2020-12-17
`parent_config` config does not include `custom_rules`


sindresorhus:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

I'm trying out the new `parent_config` property, but it seems it does not include custom rules.

I noticed since I have a lot of `// swiftlint:disable:next final_class` for [this rule](https://github.com/sindresorhus/swiftlint-config/blob/723a1fa4dd4af7dd0e0d87197997239dea9a0a1b/.swiftlint.yml#L187-L189) and it started complaining about "Superfluous Disable Command Violation: 'final_class' is not a valid SwiftLint rule. Please remove it from the disable command. (superfluous_disable_command)".

##### Complete output when running SwiftLint, including the stack trace and command used

N/A

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.42.0
* Installation method used (Homebrew, CocoaPods, building from source, etc)? GitHub Releases
* Paste your configuration file:

```yml
parent_config: https://github.com/sindresorhus/swiftlint-config/raw/723a1fa/.swiftlint.yml
```

---

// @fredpi

fredpi:
@sindresorhus Thanks for reporting! Will look into this tomorrow

fredpi:
@sindresorhus At first, I thought this issue is related to https://github.com/realm/SwiftLint/issues/3472, but now I believe it's a different issue. I will investigate further tomorrow.

fredpi:
I figured out the root cause of the issue (which turns out to be only one of multiple issues with `custom_rules`merging), but fixing it properly will require some time.

fredpi:
@sindresorhus Okay, it took less time than expected. See https://github.com/realm/SwiftLint/pull/3473. Would you mind checking if that PR fixes the issue for you?

mgray88:
Would that PR also fix overriding a custom rule of the same name from a parent config? If I just use the name of the custom rule to configure the severity it has no effect. If I redefine it under `custom_rules` it runs both checks and outputs a warning as well as an error

fredpi:
@mgray88 That's a separate issue, but I updated the PR to also address it. In addition, I opened #3477 to track that bug separately.

fredpi:
Closed via https://github.com/realm/SwiftLint/pull/3473.

jpsim:
[...]

PR #3473, opened 2020-12-18
Multiple Fixes for custom rules

fredpi:
Addresses #3472.

The reason why this issue occurred is that I didn't properly apply the new principle of _first defining all available rules and only then filtering those that are active depending on the mode_ to the custom rules. As you can see, I now moved the custom rules filtering logic to the same place where the normal rules get filtered so that the filtering isn't part of the custom rules merging algorithm.


marcelofabri:
Can you add a test that would fail without this change so we avoid regressions in the future? 🙏

In the mean time, I'll test it against my project to confirm the fix

fredpi:
@marcelofabri I thought about adding a test, but that would require me to leak some currently private interfaces of the `RulesWrapper` class across multiple layers. That may be worth it but I want to have a look at https://github.com/realm/SwiftLint/issues/3468 first to find out if an access level modification is needed anyway. I hope to find some time for that tomorrow.

fredpi:
Linux tests failing because of a new Swift version... I guess this shouldn't get fixed within this PR.

fredpi:
This PR also addresses #3468 now.

@marcelofabri I' rather not add the test you suggested as I can only think of a way of implementing it that would be depending largely on the implementation rather than on the result.

fredpi:
Now also addresses #3477.

fredpi:
As these are quite important fixes, I will test this again and probably merge it in some days' time, unless concerns emerge in the meantime.

PR #3482, opened 2020-12-28
adding platform specification to Swift package

leogdion:


ShadyGhalab (approved):

jpsim:
I'm surprised this doesn't break Linux support, thanks for the PR!

leogdion:
@jpsim that's probably where I ran into it actually. I was building a Swift AWS Lambda and using SwiftLint with [Komondor](https://github.com/shibapm/Komondor) (which I typically do).

PR #3491, opened 2021-01-15
Fix issue that prevented the inclusion of a configuration file from a parent folder

### LINKED ISSUE #3485: Since 0.41: False positive errors, seems to ignore my .swiftlint.yml
### New Issue Checklist

### Describe the bug

We've recently updated from 0.39.2 to the latest (0.42) and started getting thousands of errors and warnings in files we excluded with rules we excluded in our config file:

image

I believe this might be related to the fact we're using the `--path` flag to direct to our config file, like so:

```
Pods/SwiftLint/swiftlint --path ../
```

I've tracked this back to 0.41, since running 0.40.3 doesn't exhibit this issue. 

##### Complete output when running `swiftlint --path ../`

0.41:

```
Done linting! Found 7123 violations, 2267 serious in 886 files.
```

0.40.3:

```
Done linting! Found 3 violations, 0 serious in 321 files.
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.41
* Installation method used (Homebrew, CocoaPods, building from source, etc)? CocoaPods

Proprietary, but happy to paste specific portions you think might be an issue.

  No
* Which Xcode version are you using (check `xcodebuild -version`)? 12.3
* Do you have a sample that shows the issue? I don't - can probably try to figure something out with our structure.

fredpi:
Closes #3485.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,10 @@
 
 * Fix typos in configuration options for `file_name` rule.  
   [advantis](https://github.com/advantis)
+  
+* Fix issue that prevented the inclusion of a configuration file from a parent folder.  
+  [Frederick Pietschmann](https://github.com/fredpi)
+  [#3485](https://github.com/realm/SwiftLint/issues/3485)
 
 * Fix violation location and misplaced corrections for some function
   references in `explicit_self` rule.  
--- a/Source/SwiftLintFramework/Extensions/Configuration+FileGraph.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+FileGraph.swift
@@ -237,7 +237,15 @@ internal extension Configuration {
                 dict: firstConfigurationData.configurationDict,
                 enableAllRules: enableAllRules
             )
-            firstConfiguration.fileGraph = FileGraph(rootDirectory: firstConfigurationData.rootDirectory)
+
+            // Set the config's rootDirectory to rootDirectory (+ adjust included / excluded paths that relate to it).
+            // firstConfigurationData.rootDirectory may be different from rootDirectory,
+            // e. g. when ../file.yml is passed as the first config
+            firstConfiguration.fileGraph = FileGraph(rootDirectory: rootDirectory)
+            firstConfiguration.makeIncludedAndExcludedPaths(
+                relativeTo: rootDirectory,
+                previousBasePath: firstConfigurationData.rootDirectory
+            )
 
             // Build succeeding configurations
             return try configurationData.reduce(firstConfiguration) {
--- a/Source/SwiftLintFramework/Models/Configuration.swift
+++ b/Source/SwiftLintFramework/Models/Configuration.swift
@@ -15,10 +15,10 @@ public struct Configuration {
 
     // MARK: Public Instance
     /// The paths that should be included when linting
-    public let includedPaths: [String]
+    public private(set) var includedPaths: [String]
 
     /// The paths that should be excluded when linting
-    public let excludedPaths: [String]
+    public private(set) var excludedPaths: [String]
 
     /// The style to use when indenting Swift source code.
     public let indentation: IndentationStyle
@@ -46,7 +46,8 @@ public struct Configuration {
 
     /// The root directory is the directory that included & excluded paths relate to.
     /// By default, the root directory is the current working directory,
-    /// but in some merging algorithms it is used differently.
+    /// but during some merging algorithms it may be used differently.
+    /// The rootDirectory also serves as the stopping point when searching for nested configs along the file hierarchy.
     public var rootDirectory: String { fileGraph.rootDirectory }
 
     /// The rules mode used for this configuration.
@@ -190,11 +191,11 @@ public struct Configuration {
         let configurationFiles = configurationFiles.isEmpty ? [Configuration.defaultFileName] : configurationFiles
         defer { basedOnCustomConfigurationFiles = hasCustomConfigurationFiles }
 
-        let rootDirectory = FileManager.default.currentDirectoryPath.bridge().absolutePathStandardized()
+        let currentWorkingDirectory = FileManager.default.currentDirectoryPath.bridge().absolutePathStandardized()
         let rulesMode: RulesMode = enableAllRules ? .allEnabled : .default(disabled: [], optIn: [])
 
         // Try obtaining cached config
-        let cacheIdentifier = "\(rootDirectory) - \(configurationFiles)"
+        let cacheIdentifier = "\(currentWorkingDirectory) - \(configurationFiles)"
         if let cachedConfig = Configuration.getCached(forIdentifier: cacheIdentifier) {
             self.init(copying: cachedConfig)
             return
@@ -204,7 +205,7 @@ public struct Configuration {
         do {
             var fileGraph = FileGraph(
                 commandLineChildConfigs: configurationFiles,
-                rootDirectory: rootDirectory,
+                rootDirectory: currentWorkingDirectory,
                 ignoreParentAndChildConfigs: ignoreParentAndChildConfigs
             )
             let resultingConfiguration = try fileGraph.resultingConfiguration(enableAllRules: enableAllRules)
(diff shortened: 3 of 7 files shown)


fredpi:
The issue has been that, when only one single configuration file was provided, the relative path management for the included & excluded paths, as implemented [here](https://github.com/realm/SwiftLint/blob/537e53f6b395b7aad3aa9b3ee3588f820359d330/Source/SwiftLintFramework/Extensions/Configuration%2BMerging.swift#L32), was never called. That only happens for more than one configuration...

To fix this, I added relative path handling for the first configuration file, so that even without more configuration files, the included / excluded paths are put in proper relation to the root directory.

fredpi:
Seeing very weird SwiftLintBot reports here...

All new violations reported by the SwiftLintBot are from files that are excluded in Sourcery's `.swiftlint.yml`. However, if I understand the `osscheck` script and in particular [this line](https://github.com/realm/SwiftLint/blob/537e53f6b395b7aad3aa9b3ee3588f820359d330/script/oss-check#L141) correctly, the `.swiftlint.yml` configuration of a repo doesn't matter: it gets wiped and the repo is then linted using the `--enable-all-rules` option (unless only rules are changed but that's not the case here).

So to me it seems like the new violations are in fact intended behavior! But then, why are they new and why does **this** PR change the behavior? As a reference: [This test PR](https://github.com/realm/SwiftLint/pull/3493) that basically does nothing except for renaming a variable doesn't cause new warnings.

Primarily, I'd like to understand what the intended behavior is (meaning whether the new violations should not be there or should have always been there). Any ideas, @jpsim, @marcelofabri, @norio-nomura?

jpsim:
> Seeing very weird SwiftLintBot reports here...

Yeah it's not working correctly, see https://github.com/realm/SwiftLint/issues/3533

jpsim:
PR looks great! I hope you don't mind @fredpi I rebased and force-pushed to this branch and will merge when CI passes.

jpsim:
This is actually a breaking change, and in fact broke a SwiftLint integration with a closed source project of mine.

I believe this covers the scenario where the breakage is happening:

```
- /tmp/A/B/C/.swiftlint.yml <- config file generated by a script for `swiftlint analyze` purposes
- /src/MyProject/swiftlint.yml <- lint file checked in to the repo used for `swiftlint lint` purposes
```

The custom config looks like this:

```yaml
included:
  - Dir1/Dir2/Sources/MyFile.swift
analyzer_rules:
  - unused_import
```

And I'm running SwiftLint from the `/src/MyProject` working directory with the temporary config file specified: `swiftlint --config /tmp/A/B/C/.swiftlint.yml`.

Before this PR, it worked well:

```
$ swiftlint analyze --config \
  /tmp/A/B/C/.swiftlint.yml \
  --compile-commands \
  /tmp/A/B/C/compile_commands.json
Analyzing Swift files in current working directory
Analyzing 'MyFile.swift' (1/1)
/src/MyProject/Dir1/Dir2/Sources/MyFile.swift:1:2: warning: Unused Import Violation: Missing import for referenced module 'MyModule'. (unused_import)
Done analyzing! Found 1 violation, 0 serious in 1 file.
```

With this PR, it's failing to find files to lint:

```
$ swiftlint analyze --config \
  /tmp/A/B/C/.swiftlint.yml \
  --compile-commands \
  /tmp/A/B/C/compile_commands.json
Analyzing Swift files in current working directory
Error: No lintable files found at paths: ''
```

So unfortunately I'll be reverting this to avoid this breaking change until we can think through a possible solution.

jpsim:
Reverted in b66f7c7992ba347773e55658d68dab16a416e5f5

fredpi:
@jpsim I'm sorry for breaking things again – it seems like there were good reasons that nobody really touched the whole configuration system for some time... 🥲

If I understand your example correctly, we have different ideas of where the `included` / `excluded` paths should relate to: It seems like in previous SwiftLint versions (before 0.41.0) they related to the current working directory, right? In https://github.com/realm/SwiftLint/pull/3058 however I changed it so that the `included` / `excluded` paths are relative to the path their containing configuration file resides in.

I documented that behavior in the `README.md` (via https://github.com/realm/SwiftLint/pull/3058):

> You can specify a child_config and / or a parent_config reference within a configuration file. **These references should be local paths relative to the folder of the configuration file they are specified in.** This even works recursively, as long as there are no cycles and no ambiguities.

This PR only fixed an inconsistency with the design I chose, so if we decide that the `included` / `excluded` paths should instead be relative to the cwd, more changes than just reverting this PR are needed.

However, I still think that it is better if the paths are relative to the location of the config file:

- I don't know anymore how nested configurations worked previously, but with nested configurations it should definitely be the case that the paths specified in them relate to the location of the config file, not the cwd.
- It would be irritating if the behavior was different for other configuration files.
- Changing the location from which SwiftLint is run should only change the lintable paths IMHO. I don't think it should change the _absolute_ location of the `included` / `excluded` paths, because that should solely be up to the configuration and not depend on the cwd.

Of course, this change of concept is a breaking change and I should have better explained it in https://github.com/realm/SwiftLint/pull/3058 for us to be able to discuss its consequences before releasing it – also without the bug / inconsistency this PR fixed, we would have noticed the breaking character of this change earlier...

But now, that it is released, I think the best way to move forward is to leave it with the current state, remerge this PR and maybe drop a breaking notice in the next release notes (although parts of the breaking change already happened in 0.41.0).

But maybe it's just that I misunderstood the example you provided, so please tell me, if that's the case!

jpsim:
> it seems like there were good reasons that nobody really touched the whole configuration system for some time

This is why I really appreciate your work in this area, especially the added tests, documentation and perhaps most importantly _consideration_ about how this should all work, which I'll admit I still find pretty difficult to reason about.

After reading through your last comment and thinking it through, I completely agree that we should remerge this PR and that the behavior is what makes most sense.

However, I think there's still value in allowing out-of-tree configuration files. In my case, I don't want to dirty the git state in the repo being linted and the configuration files that I'm generating in `/tmp/` are exactly that, meant to be temporary.

I'd still like to find a way to support that. Do you agree we should find a way to support this?

Do you think it would make sense for configuration files that are loaded from locations that aren't descendant of the current working directory to have their in `included` / `excluded` paths be applied in relation to the current working directory?

jpsim:
> However, I think there's still value in allowing out-of-tree configuration files. In my case, I don't want to dirty the git state in the repo being linted and the configuration files that I'm generating in `/tmp/` are exactly that, meant to be temporary.

Actually I found a simple solution, at least for my use cases. Out-of-tree configuration files can use absolute paths, which works well today.

```swift
included:
  # Before
  - Dir1/Dir2/Sources/MyFile.swift
  # After
  - /src/MyProject/Dir1/Dir2/Sources/MyFile.swift
analyzer_rules:
  - unused_import
```

jpsim:
Re-merging in https://github.com/realm/SwiftLint/pull/3542

fredpi:
@jpsim

> which I'll admit I still find pretty difficult to reason about.

Yes, me too! There are so many use cases that need to be considered and when adding a fix for one use case, it's hard not to break another use case 😅
  
> I'd still like to find a way to support that. Do you agree we should find a way to support this?

Yes, definitely!

> Do you think it would make sense for configuration files that are loaded from locations that aren't descendant of the current working directory to have their in `included` / `excluded` paths be applied in relation to the current working directory?

I think such an approach would be rather irritating, e. g.: What behavior to expect for the use case where a configuration file from one directory above is used? Actually the issue this PR addressed (https://github.com/realm/SwiftLint/issues/3485) is an example for that: Their command is `swiftlint --config ../.swiftlint.yml`.

## The absolute path approach

But I think the solution you discovered for your use case (specifying an absolute path starting with `/`) is great! The reason why this works is because `"/foo".bridge().absolutePathRepresentation(rootDirectory: somePath)` always returns `"/foo"`.

Now, without having it tested yet, I'd claim that this does not only work for your use case (some distant path), but also when specifying an absolute path to a file that lives in a directory below the configuration file. Internally, the absolute path may get transformed to a relative path, but that wouldn't break the functionality.

So I suggest the following:

- Test whether specifying absolute paths always works with the current implementation. Add test for that to SwiftLint.
- If it works without a change, I'd still put an explanation into the [relevant method](https://github.com/realm/SwiftLint/blob/9d02cec17fdede94cd1250d2acb46c94790cb4ed/Source/SwiftLintFramework/Models/Configuration.swift#L243) where the `included` / `excluded` path management is done.
- If it doesn't work without a change, that change might be a simple one: not applying the path management in the method mentioned above for paths starting with `/`.
- Document the absolute path functionality in the `README`.

PR #3494, opened 2021-01-16
Allow multi-line parameters rule to always break lines

otaviocc:
When I originally implemented the `multiline_parameter` rule (as [@ornithocoder](https://github.com/realm/SwiftLint/graphs/contributors?from=2015-05-10&to=2021-01-17&type=a)) I made the design choice to allow either all the parameters in one line or one per line. But there are teams which prefer all the parameters to be always in different lines and never on the same line.

This PR adds the support to this use case. Setting `allows_single_line` to `false` will enforce that.

```swift
func foo(param1: Int,
         param2: Bool,
         param3: [String]) -> AnyPublisher<Response, Error>
```

And when used in combination with `multiline_parameters_brackets` it's possible to enforce the following style to all methods with multiple parameters:

```swift
func foo(
    param1: Int,
    param2: Bool,
    param3: [String]
) -> AnyPublisher<Response, Error>
```


otaviocc:
Hi @marcelofabri! Who's currently reviewing pull requests for the project? Thanks!

otaviocc:
@jpsim @marcelofabri @keith it seems there's a flaky rule (or test): `inclusive_language`, but it has nothing to do with this PR.

otaviocc:
Hi @keith @marcelofabri, do you think this one can make the next release? My team is looking forward to enabling this rule. I can leave without #3505, #3504, and #3495 for now.

Tests/SwiftLintFrameworkTests/MultilineParametersRuleTests.swift
```diff
@@ -0,0 +1,52 @@
+import SwiftLintFramework
+import XCTest
+
+final class MultilineParametersRuleTests: XCTestCase {
```
keith, line 4: Instead of having this custom test case you can actually now have examples that define their configuration https://github.com/realm/SwiftLint/blob/f66c51b9f277b8b73c814f851594c1b3b010f2e2/Source/SwiftLintFramework/Rules/Lint/TestCaseAccessibilityRuleExamples.swift#L54

otaviocc: Awesome! I'll update the PR later today.

otaviocc: Done :-)

keith:
Thanks!

Issue #3452, opened 2020-12-02
Rule request: [Test class having setup method, but no teardown]


kenji21:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

Please describe the rule idea, format
this issue's title as `Rule Request: [Rule Name]` and describe:

1. Why should this rule be added?

It's a good pratique to "rollback" in teardown what we do in setup on unit test
And XCtest create an instance of the test class for each method, if we don't set objects to nil in the teardown method, we use more memory than necessary, and attribute not setted back to nil can make side effects

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.
`setUp` (or `setUpWithError`) without `tearDown` (or `tearDownWithError()`) method in a file with Test/Tests in the end of its name (convention)

3. Should the rule be configurable, if so what parameters should be configurable?
No i don't see any possible configuration

4. Should the rule be opt-in or enabled by default? Why?
To lower memory consumption in tests, and prevent side effects

agarmash:
It's a common case for not having a `tearDown` method, consider this example:  

```swift
class StorageTests: XCTestCase {

    var storage: StorageProtocol!

    func setUp() {
        storage = InMemoryStorage()
    }

    func testSomething() {
        ...
    }
}
```

In this case, there wouldn't be any side effects since there's a brand new external dependency generated before each test. And the old one will be successfully deallocated by ARC, so memory consumption is also not an issue.  

Such rule will make more harm forcing developers to have redundant empty `tearDown` implementation just to silence the warning.

kenji21:
I found the original article pointing to me that teardown are necessary: https://qualitycoding.org/xctestcase-teardown/

```
protocol StorageProtocol {
    var i: Int { get }
}
class InMemoryStorage: NSObject, StorageProtocol {
    var i = 1
    
    deinit {
        print("\(self) gets deinit")
    }
}

class StorageWithTearDownTests: XCTestCase {
    var storage: StorageProtocol!
    
    deinit {
        print("\(self) gets deinit")
    }

    override func setUp() {
        storage = InMemoryStorage()
    }

    override func tearDown() {
        storage = nil
    }

    func testSomething() {
        XCTAssertTrue(storage.i == 1)
    }
}

class StorageTests: XCTestCase {
    var storage: StorageProtocol!

    deinit {
        print("\(self) gets deinit")
    }

[...]

PR #3495, opened 2021-01-17
Add opt-in rule balanced_xctest_lifecycle

otaviocc:
This PR implements `balanced_xctest_lifecycle`, requested in #3452. This rule enforces test classes to have balanced set-up and tear-down methods.

Triggering examples:

```swift
final class FooTests: XCTestCase {
    override func setUp() { /* ... */ }
}

final class FooTests: XCTestCase {
    override func setUpWithError() throws { /* ... */ }
}
```

Non-triggering examples:

```swift
final class FooTests: XCTestCase {
    override func setUp() { /* ... */ }
    override func tearDown() { /* ... */ }
}

final class FooTests: XCTestCase {
    override func setUpWithError() throws { /* ... */ }
    override func tearDown() { /* ... */ }
}

final class FooTests: XCTestCase {
    override func setUp() { /* ... */ }
    override func tearDownWithError() throws { /* ... */ }
}

final class FooTests: XCTestCase {
    override func setUpWithError() throws { /* ... */ }
    override func tearDownWithError() throws { /* ... */ }
}
```

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/RequiredXCTestTearDownRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/RequiredXCTestTearDownRule.swift
@@ -49,6 +49,12 @@ public struct RequiredXCTestTearDownRule: Rule, OptInRule, ConfigurationProvider
             final class FooTests: XCTestCase {
                 override func setUpAlLExamples() {}
             }
+            """#),
+            Example(#"""
+            final class FooTests: XCTestCase {
+                class func setUp() {}
+                class func tearDown() {}
+            }
             """#)
         ],
         triggeringExamples: [
@@ -69,6 +75,11 @@ public struct RequiredXCTestTearDownRule: Rule, OptInRule, ConfigurationProvider
             }
             final class ↓BarTests: XCTestCase {
                 override func setUpWithError() throws {}
+            }
+            """#),
+            Example(#"""
+            final class ↓FooTests: XCTestCase {
+                class func setUp() {}
             }
             """#)
         ]
--- a/Source/SwiftLintFramework/Rules/Lint/RequiredXCTestTearDownRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/RequiredXCTestTearDownRule.swift
@@ -1,6 +1,8 @@
 import SourceKittenFramework
 
 public struct RequiredXCTestTearDownRule: Rule, OptInRule, ConfigurationProviderRule, AutomaticTestableRule {
+    // MARK: - Properties
+
     public var configuration = SeverityConfiguration(.warning)
 
     public static let description = RuleDescription(
@@ -85,11 +87,17 @@ public struct RequiredXCTestTearDownRule: Rule, OptInRule, ConfigurationProvider
         ]
     )
 
+    // MARK: - Life cycle
+
     public init() {}
+
+    // MARK: - Public
 
     public func validate(file: SwiftLintFile) -> [StyleViolation] {
         testClasses(in: file).compactMap { violations(in: file, for: $0) }
     }
+
+    // MARK: - Private
 
     private func testClasses(in file: SwiftLintFile) -> [SourceKittenDictionary] {
         file.structureDictionary.substructure.filter { dictionary in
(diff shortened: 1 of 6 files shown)


otaviocc:
Rebased :-)

Source/SwiftLintFramework/Rules/Lint/RequiredXCTestTearDownRule.swift
```diff
@@ -0,0 +1,122 @@
...
+
+public struct RequiredXCTestTearDownRule: Rule, OptInRule, ConfigurationProviderRule, AutomaticTestableRule {
+    public var configuration = SeverityConfiguration(.warning)
+
+    public static let description = RuleDescription(
+        identifier: "required_xctest_teardown",
+        name: "Required XCTest Tear Down",
+        description: "Test classes must implement tearDown when setUp is provided.",
+        kind: .lint,
+        nonTriggeringExamples: [
```
marcelofabri, line 11: can you add some examples using the [`class func setUp`](https://developer.apple.com/documentation/xctest/xctestcase/1496262-setup) and [`class func tearDown`](https://developer.apple.com/documentation/xctest/xctestcase/1496280-teardown)?

I wonder if it makes sense to check that they're balanced (e.g. a `class setUp` requires a `class tearDown`)

marcelofabri:
FYI [`addTeardownBlock`](https://developer.apple.com/documentation/xctest/xctestcase/2887226-addteardownblock) is another way to solve this problem, but I don't know if it's worth detecting it

Source/SwiftLintFramework/Rules/Lint/RequiredXCTestTearDownRule.swift
otaviocc: Added :-) It does make sense to check if they're balanced. Since they are static methods they have potential to leak to other tests if misused.

otaviocc:
> FYI [`addTeardownBlock`](https://developer.apple.com/documentation/xctest/xctestcase/2887226-addteardownblock) is another way to solve this problem, but I don't know if it's worth detecting it

I'm on the fence about this one. Not sure it's worth it.

otaviocc:
@marcelofabri `danger` failed ˆˆ

marcelofabri:
> @marcelofabri `danger` failed ˆˆ

Can you try pushing an empty commit, something in the changelog, etc, to trigger a new build?

otaviocc:
> > @marcelofabri `danger` failed ˆˆ
> 
> Can you try pushing an empty commit, something in the changelog, etc, to trigger a new build?

@marcelofabri, done ;-)

otaviocc:
Rebased to resolve a conflict in CHANGELOG.md :-)

otaviocc:
Hi @marcelofabri, can you take a look on this one again? Thanks!

otaviocc:
Hi @marcelofabri, any chance to get this one in the next release? Thanks!

otaviocc:
Hi @jpsim, any chance to get this one merged before the release? Thanks!

Source/SwiftLintFramework/Rules/Lint/RequiredXCTestTearDownRule.swift
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private func violations(in file: SwiftLintFile,
+                            for dictionary: SourceKittenDictionary) -> StyleViolation? {
+        let methods = dictionary.substructure
+            .compactMap { XCTMethod($0.name) }
+
+        guard
+            methods.contains(.setUp) == true,
+            methods.contains(.tearDown) == false,
```
jpsim, line 116: Perhaps this rule could be called `BalancedXCTestLifecycleRule` and also enforce that if `tearDown()` is defined, then so should `setUp()`?

```suggestion
            methods.contains(.setUp) != methods.contains(.tearDown),
```

jpsim (approved):
Nice rule! I do think we should make it a bit more general to handle the inverse case too.

I've just granted you write access to the repo @otaviocc so please merge when you've made those changes (unless they don't make sense for some reason) and CI is green.

otaviocc:
> Nice rule! I do think we should make it a bit more general to handle the inverse case too.
> 
> I've just granted you write access to the repo @otaviocc so please merge when you've made those changes (unless they don't make sense for some reason) and CI is green.

Thank you for the access! I'll make the changes and will merge then.

Source/SwiftLintFramework/Rules/Lint/RequiredXCTestTearDownRule.swift
otaviocc: Exactly! I was going to say I will rename the rule, but just saw your comment :-) I'll rename it. Thanks!

otaviocc: All changes applied, @jpsim. I'm merging it.

PR #3501, opened 2021-01-22
[Analyze] Support compile commands with relative paths

jpsim:


keith (approved):

keith:
Maybe too aggressive but we could make this fail if it didn't find either

jpsim:
> Maybe too aggressive but we could make this fail if it didn't find either

Yeah that's too aggressive IMO. It's convenient to have files that you want to lint but don't have compilation commands when analyzing. For example, sample code, unit tests or sources that don't compile easily with the target you're analyzing.

Logging a warning might be appropriate though, but again depending on the configuration could be overly noisy.

PR #3504, opened 2021-01-24
Add opt-in rule private_passthrough_subject

otaviocc:
This PR adds a new opt-in rule, `private_subject`, which warns against public subjects (`PassthroughSubject` and `CurrentValueSubject`). Ideally subjects should be implemented as private properties to prevent external entities from calling `send(_:)/send()` on them.

```swift
final class Toto {
    // Doesn't trigger
    private let goodSubject: PassthroughSubject<Bool, Never> = .init()
    private let anotherGoodSubject = PassthroughSubject<Bool, Never>()

    // Triggers
    let badSubject: PassthroughSubject<Bool, Never> = .init()
    private(set) let anotherBadSubject = PassthroughSubject<Bool, Never>()
}
```

```swift
final class Toto {
    // Doesn't trigger
    private let goodSubject: CurrentValueSubject<Bool, Never> = .init(false)
    private let anotherGoodSubject = CurrentValueSubject<Bool, Never>(false)

    // Triggers
    let badSubject: CurrentValueSubject<Bool, Never> = .init(false)
    private(set) let anotherBadSubject = CurrentValueSubject<Bool, Never>(false)
}
```

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/PrivateSubjectRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/PrivateSubjectRule.swift
@@ -76,7 +76,7 @@ public struct PrivateSubjectRule: ASTRule, OptInRule, ConfigurationProviderRule,
     /// Looks for violations matching the format:
     ///
     /// * `let subject = PassthroughSubject<Bool, Never>()`
-    /// * `let subject = CurrentValueSubject<String, Never> = .ini("toto")`
+    /// * `let subject = CurrentValueSubject<String, Never>("toto")`
     ///
     /// - Returns: The violation offset.
     private func defaultValueViolationOffset(file: SwiftLintFile,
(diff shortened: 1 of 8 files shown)


otaviocc:
Rebased :-)

otaviocc:
Hi @keith, can you take a look at this one when you have a chance? Thanks!

Source/SwiftLintFramework/Rules/Lint/PrivateSubjectRule.swift
```diff
@@ -0,0 +1,98 @@
...
+
+    public init() {}
+
+    // MARK: - Public
+
+    public func validate(file: SwiftLintFile,
+                         kind: SwiftDeclarationKind,
+                         dictionary: SourceKittenDictionary) -> [StyleViolation] {
+        guard
+            kind == .varInstance,
```
jpsim, line 29: Do you also want to prevent global vars and static vars?

jpsim (commented):
Looks good, thanks for the rule and tests!

serges147:
@otaviocc Hi! Could you please give rationale why it's bad to "_external entities calling send(_:)/send() on them._"? I understand that sometime all you need is expose `Publisher` only - it that case, yes, rule is good. But if I do really want both publisher & `.send` functionality (aka Subject) to be exposed... can you suggest something (except swiftlint:disable) in that case?

PR #3505, opened 2021-01-25
Add opt-in rule discouraged_default_switch_case

otaviocc:
This PR adds the opt-in rule `discouraged_default_switch_case` to discourage `default` in a switch of enum cases.

Triggering example:

```swift
enum StateOfMatter {
    case liquid
    case solid
    case gas
    case plasm

    var description: String {
        switch self {
            case .liquid: return "Liquid"
            case .solid: return "Solid"
            default: return "Other stuff"
        }
    }
}
```

Non-triggering examples:

```swift
switch self {
    case .liquid: return "Liquid"
    case .solid: return "Solid"
    case .gas: return "Gas"
    case .plasm: return "Plasm"
}
```

```swift
switch result {
    case .success(let value): print(value)
    case .failure(let error): print(error)
}
```

```swift
switch number {
    case 1...4: return "Number is less than 5"
    case 5: return "Number is 5"
    default: return "Number is greater than 5"
}
```

```swift
switch (status, value) {
    case (.on, .green): return true
    default: return false
}
```

```swift
switch value {
    case .one: return "One"
    case .two: return "Two"
    case .default: return "Default"
}
```

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/DiscouragedDefaultSwitchCaseRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/DiscouragedDefaultSwitchCaseRule.swift
@@ -115,7 +115,7 @@ public struct DiscouragedDefaultSwitchCaseRule: ASTRule, OptInRule, Configuratio
 
     // MARK: - Nested type
 
-    private typealias EnumSwitchCase = (offset: ByteCount, content: String)
+    private typealias SwitchCase = (offset: ByteCount, content: String)
 
     // MARK: - Life cycle
 
@@ -123,6 +123,10 @@ public struct DiscouragedDefaultSwitchCaseRule: ASTRule, OptInRule, Configuratio
 
     // MARK: - Public
 
+    /// Checks if all the elements from cases in a `switch` are from an enum and
+    /// include `default`.
+    ///
+    /// - Returns: Violations in case `default` is found in a `switch` of enums.
     public func validate(file: SwiftLintFile,
                          kind: StatementKind,
                          dictionary: SourceKittenDictionary) -> [StyleViolation] {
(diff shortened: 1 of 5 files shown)


otaviocc:
Rebased :-)

otaviocc:
Rebased :-)

otaviocc:
@jpsim @keith do you see any other case I might be missing here?

otaviocc:
I found a case:

```swift
swift something {
    case Foo.one: return "one"
    case Foo.two return "two"
    case Foo.three return "three"
    // ... 
}
````

When the type - e.g., `Foo` above - is included in the case statement. I could check if everything before the last dot follows the same pattern. Should I include it in this first version or try it in a different PR to unblock this one?

File: Source/SwiftLintFramework/Rules/Lint/DiscouragedDefaultSwitchCaseRule.swift [hunk 1]
```diff
@@ -0,0 +1,190 @@
...
+                    }
+                }
+                """#
+            ),
+            Example(
+                #"""
+                switch CLLocationManager.authorizationStatus() {
+                    case .authorizedAlways, .authorizedWhenInUse, .notDetermined: return "authorized"
+                    case .denied, .restricted: return "denied"
+                    @unknown ↓default: fatalError("Unhandled case!")
```
keith, line 101: hrm, what's the resolution here though?

jpsim: Keith has a good point. I think these shouldn't trigger. What do you think @otaviocc ?

otaviocc: @keith, @jpsim Yep, it shouldn't. I'll work on it (as soon as I find a way to check if it's `unknown`. I'm gonna work on it tomorrow. Thanks!

jpsim:
It appears so, sorry about that, I thought GitHub was supposed to handle that automatically. 🤔

PR #3513, opened 2021-02-06
Fix false positives on indentation_width for lines commented at the start

### LINKED ISSUE #3430: indentation_width triggered when the previous line is commented at the beginning
### New Issue Checklist

### Describe the bug

Commenting a line at the beginning of it incorrectly triggers `indentation_width` for the next line.

Side note: This is how Xcode comments lines by default when using <kbd>cmd</kbd> + <kbd>/</kbd>.

```bash
$ ./swiftlint lint --path Test.swift
Linting Swift files at paths Test.swift
Linting 'Test.swift' (1/1)
/Users/revolt/Downloads/portable_swiftlint/Test.swift:4:1: warning: Indentation Width Violation: Code should be indented using one tab or 4 spaces. (indentation_width)
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? `0.41.0`
* Installation method used (Homebrew, CocoaPods, building from source, etc)? Using a pre-built package

```yml
opt_in_rules:
    - indentation_width
```

* Which Xcode version are you using (check `xcodebuild -version`)? Xcode 12.2 Build version 12B45b

```swift
class Test {
	func test() {
//		comment
		let string = "string"
	}
}
```

revolter:
Fixed #3430.


Source/SwiftLintFramework/Rules/Style/IndentationWidthRule.swift
```diff
@@ -46,14 +47,22 @@ public struct IndentationWidthRule: ConfigurationProviderRule, OptInRule {
     public init() {}
 
     // MARK: - Methods: Validation
-    public func validate(file: SwiftLintFile) -> [StyleViolation] { // swiftlint:disable:this function_body_length
+	// swiftlint:disable:next cyclomatic_complexity function_body_length
```
jpsim, line 50: Can you please use spaces for indentation here?

Source/SwiftLintFramework/Rules/Style/IndentationWidthRule.swift
```diff
@@ -46,14 +47,22 @@ public struct IndentationWidthRule: ConfigurationProviderRule, OptInRule {
...
+    public func validate(file: SwiftLintFile) -> [StyleViolation] {
         var violations: [StyleViolation] = []
         var previousLineIndentations: [Indentation] = []
 
         for line in file.lines {
             // Skip line if it's a whitespace-only line
-            let indentationCharacterCount = line.content.countOfLeadingCharacters(in: CharacterSet(charactersIn: " \t"))
-            if line.content.count == indentationCharacterCount { continue }
+            let comment = "//"
```
jpsim, line 57: Should this handle other comment styles (`/*`, `///`, `/**`)?

revolter:
I'm lost :( I temporarily reverted the changes from `IndentationWidthRule.swift`, but the test that I added in `IndentationWidthRuleTests.swift` still succeeds. And it's the same with an example that compiles:

```none
class Test {\n\tfunc test() {\n//\t\ttest\n\t\tprint("test")\n\t}\n}
```

```swift
class Test {
	func test() {
//		test
		print("test")
	}
}
```

Example which does trigger a warning when adding in my codebase.

Source/SwiftLintFramework/Rules/Style/IndentationWidthRule.swift
revolter: I added support for those too, but I don't quite like the implementation 😞 I thought that maybe `String.commentBody()` would help me, but it doesn't support `/*`: https://github.com/jpsim/SourceKitten/issues/695.

PR #3515, opened 2021-02-07
Improve parallelMap implementation by removing iteration for force unwrapping every element

cltnschlosser:
Utilizes `Array<T>(unsafeUninitializedCapacity:)` to get access to `UnsafeMutableBufferPoint<T>`. Instead of the previous approach which used `UnsafeMutableBufferPoint<T?>`, which required an unwrapping iteration.

### DIFF
--- a/Source/swiftlint/Extensions/Array+SwiftLint.swift
+++ b/Source/swiftlint/Extensions/Array+SwiftLint.swift
@@ -6,12 +6,11 @@ extension Array {
     }
 
     func parallelMap<T>(transform: (Element) -> T) -> [T] {
-        var result = ContiguousArray<T?>(repeating: nil, count: count)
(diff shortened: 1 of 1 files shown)


cltnschlosser:
Looks like I need to run this under TSAN, I'll reopen when I figure out what I'm doing wrong here.

cltnschlosser:
Okay, figured it out. Seems to either be a compiler bug with `buffer[x]` trying to release the assigned value, or just a failure in documentation that `buffer[x]` is unsafe. Adapted solution from the original swift evolution proposal. https://github.com/apple/swift-evolution/blob/main/proposals/0245-array-uninitialized-initializer.md#proposed-solution

I'll see if I can reproduce is a smaller context and create a SR bug.

PR #3517, opened 2021-02-07
Improve parallelMap implementation by removing iteration for force unwrapping every element

cltnschlosser:
#3515 Wouldn't reopen, so created new PR

### DIFF
--- a/Source/swiftlint/Extensions/Array+SwiftLint.swift
+++ b/Source/swiftlint/Extensions/Array+SwiftLint.swift
@@ -6,12 +6,14 @@ extension Array {
     }
 
     func parallelMap<T>(transform: (Element) -> T) -> [T] {
-        var result = ContiguousArray<T?>(repeating: nil, count: count)
-        return result.withUnsafeMutableBufferPointer { buffer in
-            DispatchQueue.concurrentPerform(iterations: buffer.count) { idx in
-                buffer[idx] = transform(self[idx])
+        return [T](unsafeUninitializedCapacity: count) { buffer, initializedCount in
+            let baseAddress = buffer.baseAddress!
+            DispatchQueue.concurrentPerform(iterations: count) { index in
+                // Using buffer[index] does assignWithTake which tries
+                // to read the uninitialized value (to release it) and crashes
+                (baseAddress + index).initialize(to: transform(self[index]))
             }
-            return buffer.map { $0! }
+            initializedCount = count
         }
     }
 }


Source/swiftlint/Extensions/Array+SwiftLint.swift
```diff
@@ -6,12 +6,14 @@ extension Array {
     }
 
     func parallelMap<T>(transform: (Element) -> T) -> [T] {
-        var result = ContiguousArray<T?>(repeating: nil, count: count)
-        return result.withUnsafeMutableBufferPointer { buffer in
-            DispatchQueue.concurrentPerform(iterations: buffer.count) { idx in
-                buffer[idx] = transform(self[idx])
+        return [T](unsafeUninitializedCapacity: count) { buffer, initializedCount in
+            let baseAddress = buffer.baseAddress!
```
jpsim, line 10: Is this always guaranteed to succeed? Is that behavior documented?

Source/swiftlint/Extensions/Array+SwiftLint.swift
```diff
@@ -6,12 +6,14 @@ extension Array {
...
+            DispatchQueue.concurrentPerform(iterations: count) { index in
+                // Using buffer[index] does assignWithTake which tries
+                // to read the uninitialized value (to release it) and crashes
+                (baseAddress + index).initialize(to: transform(self[index]))
```
jpsim, line 14: How does this work? I’d expect to have to multiply the index by the size of the transformed result type.

cltnschlosser: It used the type of the pointer to know how much to increase by. The `+` operator isn't documented, but it looks like this new function is and is documented to behave the same way: https://developer.apple.com/documentation/swift/unsafemutablepointer/3393143-advanced

Source/swiftlint/Extensions/Array+SwiftLint.swift
cltnschlosser: In this case yes. Documented, not really. That's how I ran into this: https://bugs.swift.org/browse/SR-14164

If you follow throw this, the buffer is initialized with a non-optional baseAddress: https://github.com/apple/swift/blob/06dc593cece2ded7457ba3d5a6c41857881da157/stdlib/public/core/Array.swift#L1468

jpsim (approved):
Thanks for answering my questions. Looks better than the previous implementation 👌

PR #3518, opened 2021-02-08
Capture non-constant variable

kustra:
Created a new analyzer rule that shows a warning if items listed in a closure's capture list are `var` variables, meaning that their value is captured at the time the closure is created, not at the time it's called.

This is useful when a former `let` constant is changed to `var`, so the developer doesn't have to remember to look for closures capturing it. Or the developer simply forgets this part of the capture semantics.

I think this rule should exclude `lazy var`s and implicitly unwrapped optionals from checking, but this is not implemented yet. I'll probably add it this Friday. Until then, I just posted the PR to get initial feedback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,10 +68,11 @@
 * Add Bool violation reporting in `redundant_type_annotation`.  
   [Artem Garmash](https://github.com/agarmash)
   [#3423](https://github.com/realm/SwiftLint/issues/3423)
-* Add opt-in rule `capture_variable` to warn about listing a non-constant
-  (`var`) variable in a closure's capture list. This captures the variable's
-  value at closure creation time instead of closure call time, which may be
-  unexpected.
+
+* Add a new `capture_variable` analyzer rule to warn about listing a
+  non-constant (`var`) variable in a closure's capture list. This
+  captures the variable's value at closure creation time instead of
+  closure call time, which may be unexpected.  
   [Laszlo Kustra](https://github.com/kustra)
 
 #### Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,11 @@
 * Add Bool violation reporting in `redundant_type_annotation`.  
   [Artem Garmash](https://github.com/agarmash)
   [#3423](https://github.com/realm/SwiftLint/issues/3423)
+* Add opt-in rule `capture_variable` to warn about listing a non-constant
+  (`var`) variable in a closure's capture list. This captures the variable's
+  value at closure creation time instead of closure call time, which may be
+  unexpected.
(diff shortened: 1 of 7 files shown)


jpsim:
Thanks for your patience, I hope to look at this soon!

jpsim:
This looks good in terms of adding this rule, you have tests, the rule is well structured, named, etc.

I rebased your branch to update it and made a few changes. I removed the conformance to the `OptInRule` protocol because that's only used for linter rules and does not apply to analyzer rules. I reworded the changelog entry to reflect this.

In terms of the practical utility of the rule, I can see some use cases for projects that follow very specific patterns. However, it's totally valid to want to capture the value of a mutable variable at closure creation time rather than invocation time.

That being said, just because I can't think of projects in which I'd turn this on myself, I fully support shipping this rule as part of SwiftLint, because we don't have a suitable alternative like a plugin system that would let you define your own analyzer rules without forking all of SwiftLint.

You said you wanted to make more changes to the rule, like also include lazy variables. I'll let you do that and you can ping me again for a review when this is ready to merge.

kustra:
> In terms of the practical utility of the rule, I can see some use cases for projects that follow very specific patterns. However, it's totally valid to want to capture the value of a mutable variable at closure creation time rather than invocation time.

For reference, the idea for this rule was born in an RxSwift-based iOS app project. I've observed (pun intended) that it was always a bug in our project to capture a `var` variable in one of the long-lived Rx chains of the service layer, or the ViewModel layer. These chains are usually created when the service starts, or the screen appears for the user, and are disposed when the service is terminated, or the user navigates away from the screen. 

For "one-shot" Rx chains, e.g. service operations, capturing a `var` variable at closure creation time is usually not a bug. Unless there's a multithreading issue. (In our project, a service operation may be called from any thread, and it returns an Rx Observable or similar that switches to the appropriate scheduler for the service with RxSwift's `subscribeOn` or `observeOn` operators. But there's no way to control the thread that _creating_ the Rx chain, along with its closures, runs on. That happens on the caller's thread. I'm not sure if this creates an issue when the captured variable is modified at the same time as when the capturing closure is created. This just occurred to me.)

> You said you wanted to make more changes to the rule, like also include lazy variables. I'll let you do that and you can ping me again for a review when this is ready to merge.

@jpsim done with the changes. I think the PR is ready to merge.

jpsim:
Thanks @kustra, I fixed the changelog conflict locally and pushed directly to `master`: 87d4accb39519fc9c2db084f75a3ce27b214469e

FYI you're the first person other than myself to contribute an analyzer rule, so thanks for the contribution 🙌 🚀

kustra:
> FYI you're the first person other than myself to contribute an analyzer rule, so thanks for the contribution 🙌 🚀

You're welcome. It was fun!

PR #3519, opened 2021-02-09
Feature/find incomplete docs

jortberends:
This PR adds the `mind_incomplete_docs` flag to the `missing_docs` rule to warn about incomplete documentation.

**Triggering examples:**
Return type not documented
`
/// docs
public func a() -> Bool {
    return true
}
`

Incorrect parameter documented
`
/// docs
/// - Parameters:
///   - b: docs
public func a(a: Bool) {}
`

Parameter not documented
`
/// docs
public func a(a: Bool) {}
`

**Non-triggering examples:**
`
/// docs
/// - Parameter a: docs
public func a(a: Bool) {}
`

`
/// docs
/// - Parameters:
///   - a: docs
public func a(a: Bool) {}
`

`
/// docs
/// - Returns: docs
public func a() -> Bool {
    return true
}
`


jortberends:
Rebased.

jpsim:
Since this rule was originally written, a new project specifically for linting and fixing Swift docstrings has appeared: https://github.com/dduan/DrString

I'm wondering if it's best to point people to use that tool instead of trying to match its feature set in SwiftLint.

jortberends:
Hey @jpsim thanks for reaching out. We were totally unaware of the DrString project. It looks way more extensive than the rule we wrote. Will try to use your project to see if this PR still makes any sense or should be closed.

TomVanDerSpek:
@jpsim, DrString helps you find and fix inconsistencies in your documentation. It doesn’t warn you if documentation is missing and you also cannot specify what kind of api access level you want to have documented (private / internal / public). 
So unfortunately DrString does not solve the same problem as this MR does at the moment. 

Would be awesome if someone could look at this MR and approve it.

bdfox325:
@jortberends What is the status of this MR. Would love to use this. It would also be great if the rule was configurable to catch missing docs on extensions as well.

jortberends:
@bdfox325 We're using it in our project since DrString didn't fulfil all our wishes as @TomVanDerSpek stated. 
It won't catch missing docs on extensions since this information is not easily provided by SourceKit afaik.

Not sure how to get more attention to this MR and get it merged, i hope it helps if people ask about the status like you did 👍

bdfox325:
@jortberends Please take a look at my related PR: https://github.com/realm/SwiftLint/pull/3701

Source/SwiftLintFramework/Rules/Lint/MissingDocsRule.swift
```diff
@@ -71,6 +139,26 @@ public struct MissingDocsRule: OptInRule, ConfigurationProviderRule, AutomaticTe
             """),
             Example("""
             public extension A {}
+            """),
+            // public, parameter documented
```
Harry-KNIGHT, line 143: Is it a relevant commentary or can you change it into documentation ?

Issue #3012, opened 2020-01-04
Rule Request: directory_name_no_space prevent whitespace characters in directory names


timcmiller:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

Prevent any whitespace characters in directory names. Having spaces in directories cause issues (or annoyances) when trying to get a path to files located inside the directory since you have to escape the space.

1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.

There is ambiguity on folder naming between different cases, however it it standard convention to not include spaces in folder names. See [Swift](https://github.com/apple/swift) which uses either kebab-case or camelCase for folder names. See this project which uses both camelCase and snake_case.

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

Does not trigger
```
DirectoryName
Directory_Name
Directory-Name
```

Does Trigger
```
Directory Name
```

3. Should the rule be configurable, if so what parameters should be configurable?

With the exception of an exclusion list for folders you might want to exclude (See Supporting Files for this project), no. 

4. Should the rule be opt-in or enabled by default? Why?

Yes, because Apple doesn't have a defined convention for naming folders.

jpsim:
Any reason not to combine this with the file name no space rule?

timcmiller:
> Any reason not to combine this with the file name no space rule?

While it's commonly accepted practice across projects, there is much less supporting documented standards that specifically call our folder names not having spaces. With a lack of supporting evidence I thought it made sense to have two separate rules to allow levels of opting in.

jpsim:
I think that's fine. The alternative would be to make this a configurable option on a combined `no_spaces_in_paths` rule, but this accomplishes the same effect.

PR #3523, opened 2021-02-16
Add Directory name no space opt-in rule

timcmiller:
#3012

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,10 @@
   public Combine subjects.  
   [Otavio Cordeiro](https://github.com/otaviocc)
 
+* Added 'directory_name_no_space' opt-in rule.
+  [timcmiller](https://github.com/timcmiller)
+  [#3012](https://github.com/realm/SwiftLint/issues/3012)
+
 #### Bug Fixes
 
 * Fix typos in configuration options for `file_name` rule.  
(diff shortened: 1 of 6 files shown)


jpsim:
What's the purpose of the parent directory configuration option? Note that SwiftLint already determines the root directory being linted, which might be want you want to access, but I'm not sure because I don't understand the purpose of the configuration option.

timcmiller:
> What's the purpose of the parent directory configuration option? Note that SwiftLint already determines the root directory being linted, which might be want you want to access, but I'm not sure because I don't understand the purpose of the configuration option.

Yes, this is precisely what I wanted. The parent directory that SwiftLint is linting. Is there an example you can point me towards where I could access the said parent directory that SwiftLint already has access to?

You can see demonstrated in this test I wrote that the user would be able to filter out based on subdirectory.

```swift
    func testCustomParentDirectory() {
        XCTAssert(try validate(fileName: "Directory Name/SubdirectoryName/File.swift",
                               parentDirectory: "SubdirectoryName").isEmpty)
    }
```

jpsim:
> Is there an example you can point me towards where I could access the said parent directory that SwiftLint already has access to?

https://github.com/realm/SwiftLint/blob/8c29a8988888e2376d5c20c0ed61a1b885394d4a/Source/SwiftLintFramework/Models/Configuration.swift#L47-L50

I don't think we have any rules accessing this at the moment.

timcmiller:
Thanks @jpsim, I'll get the rule updated to look at `rootDirectory` and drop the ` parentDirectory` configuration!

PR #3527, opened 2021-02-22
Update swiftlint Build phase code snippet to work on M1 Macs

ralfebert:
The code snippet in the README should also work out of the box on M1 Macs where homebrew is installed in `/opt/homebrew` by default; swiftlint will not be in the default $PATH then and the shell configuration files like '.bashrc' don't change the environment for Xcode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,9 @@ target, and go to Build Phases. Click the + and select "New Run Script Phase".
 Insert the following as the script:
 
 ```bash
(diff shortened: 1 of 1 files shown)


jpsim:
Can you move this suggestion to the Homebrew section of the installation instructions? This run script doesn't apply if you installed SwiftLint using any other method or at a location that _is_ in Xcode's path.

OlegEremenko991:
Any progress with this issue?

jpsim:
Merged #3778. Thanks for raising this.

PR #3536, opened 2021-02-24
Add "stable git revision" to configuration

jpsim:
This allows specifying a git revision or "commit-ish" that is considered stable. If specified, SwifLint will attempt to query the git repository index for files changed since that revision, using only those files as input as opposed to traversing the file system to collect lintable files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,13 @@
   performance.  
   [jpsim](https://github.com/jpsim)
 
+* Add "stable git revision" to configuration. This allows specifying a
+  git revision or "commit-ish" that is considered stable. If specified,
+  SwifLint will attempt to query the git repository index for files
+  changed since that revision, using only those files as input as
+  opposed to traversing the file system to collect lintable files.  
+  [jpsim](https://github.com/jpsim)
+
 #### Bug Fixes
 
 * Fix typos in configuration options for `file_name` rule.  
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,3 +1,4 @@
+stable_git_revision: origin/master
 included:
   - Source
   - Tests
--- a/Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
@@ -12,8 +12,10 @@ extension Configuration {
     ///
     /// - returns: Files to lint.
     public func lintableFiles(inPath path: String, forceExclude: Bool,
+                              fileManager: LintableFileManager,
                               excludeByPrefix: Bool = false) -> [SwiftLintFile] {
-        return lintablePaths(inPath: path, forceExclude: forceExclude, excludeByPrefix: excludeByPrefix)
+        return lintablePaths(inPath: path, forceExclude: forceExclude, excludeByPrefix: excludeByPrefix,
+                             fileManager: fileManager)
             .compactMap(SwiftLintFile.init(pathDeferringReading:))
     }
 
@@ -31,7 +33,7 @@ extension Configuration {
         inPath path: String,
         forceExclude: Bool,
         excludeByPrefix: Bool = false,
-        fileManager: LintableFileManager = FileManager.default
+        fileManager: LintableFileManager
     ) -> [String] {
         // If path is a file and we're not forcing excludes, skip filtering with excluded/included paths
         if path.isFile && !forceExclude { return [path] }
@@ -53,7 +55,7 @@ extension Configuration {
     ///
     /// - returns: The input paths after removing the excluded paths.
     public func filterExcludedPaths(
-        fileManager: LintableFileManager = FileManager.default,
+        fileManager: LintableFileManager,
         in paths: [String]...
     ) -> [String] {
         let allPaths = paths.flatMap { $0 }
--- a/Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+Merging.swift
@@ -25,7 +25,8 @@ extension Configuration {
             warningThreshold: mergedWarningTreshold(with: childConfiguration),
             reporter: reporter,
             cachePath: cachePath,
-            allowZeroLintableFiles: childConfiguration.allowZeroLintableFiles
+            allowZeroLintableFiles: childConfiguration.allowZeroLintableFiles,
+            stableGitRevision: stableGitRevision
         )
     }
 
--- a/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
@@ -22,6 +22,7 @@ extension Configuration {
         case parentConfig = "parent_config"
         case remoteConfigTimeout = "remote_timeout"
         case remoteConfigTimeoutIfCached = "remote_timeout_if_cached"
+        case stableGitRevision = "stable_git_revision"
     }
 
     // MARK: - Properties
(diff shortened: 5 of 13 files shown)


CHANGELOG.md
```diff
@@ -62,6 +62,13 @@
   performance.  
   [jpsim](https://github.com/jpsim)
 
+* Add "stable git revision" to configuration. This allows specifying a
+  git revision or "commit-ish" that is considered stable. If specified,
+  SwifLint will attempt to query the git repository index for files
```
keith, line 67:
```suggestion
  SwiftLint will attempt to query the git repository index for files
```

jpsim:
Good discussion offline with Keith, I think I'll remove this from the configuration and make it a CLI argument. This will make it easier to run against all files say on CI and use this faster mode locally without having to edit the config file. SwiftLint should also be plenty fasts for small projects even when linting everything, so if you can handle the complexity of a large Swift project, you can handle the complexity of running SwiftLint with a flag (`swiftlint --stable-git-revision=origin/main`).

jpsim:
There are a few correctness issues here:

1. If any of the configuration files used by SwiftLint are modified, that should force all files to be re-linted, but currently this is ignored
2. Same thing when the Swift version changes.

There might be a way to integrate into the existing `LinterCache` caching system to resolve this, but it's not clear to me that there's a way to do that without first traversing the file system, which in many ways defeats the purpose of this feature.

A naive way to solve the first issue could be to check if there are any files named `.swiftlint.yml` that were modified and then fall back to using the standard `FileManager.default` lintable file manager, but this doesn't handle the case where configuration files with different names are used, or if a configuration file that's not tracked in git is specified.

Maybe this feature is still valuable enough with these correctness to ship as-is, but I'd like to think about our options a bit more.

jpsim:
I'd be curious to hear some thoughts from contributors on this, especially the limitations involved and if anyone has ideas for resolving them. @keith @marcelofabri @fredpi @otaviocc

CHANGELOG.md
```diff
@@ -62,6 +62,13 @@
   performance.  
   [jpsim](https://github.com/jpsim)
 
+* Add `--stable-git-revision` option to the lint command. This allows
+  specifying a git revision or "commit-ish" that is considered stable.
+  If specified, SwiftLint will attempt to query the git repository index
+  for files changed since that revision, using only those files as input
+  as opposed to traversing the file system to collect lintable files.  
+  [jpsim](https://github.com/jpsim)
```
AvdLee, line 70: Would it be valuable to default to `HEAD`?

AvdLee (commented):
I really like this addition, it will definitely help us for https://github.com/WeTransfer/WeTransfer-iOS-CI/pull/67.

The thing I'm asking myself is how I would be able to dynamically specify the revision to compare to. I also realise our current implementation is far from ideal in this sense.

E.g. sometimes your local changes will be merged into develop while another time you might want to merge into another branch. I guess in our case, defaulting to HEAD would make the most sense 🤔

CHANGELOG.md
jpsim: I’d rather not have a default at all because of the aforementioned correctness pitfalls.

jpsim:
> The thing I'm asking myself is how I would be able to dynamically specify the revision to compare to. I also realise our current implementation is far from ideal in this sense.

I think it’s best to see this option as an optimization only. If you have a branch that you keep free of lint violations, then you can specify it as the stable git branch. Or if you run SwiftLint as a pre-commit check, specifying HEAD should work in that case. It’s “ok” if you lint more files than have actually changed, those will be skipped if they’re in the SwiftLint cache.

CHANGELOG.md
AvdLee: Makes sense! It's fairly simple as well to provide a default on the implementation level 👌

AvdLee:
> I think it’s best to see this option as an optimization only. If you have a branch that you keep free of lint violations, then you can specify it as the stable git branch. Or if you run SwiftLint as a pre-commit check, specifying HEAD should work in that case. It’s “ok” if you lint more files than have actually changed, those will be skipped if they’re in the SwiftLint cache.

Thanks for that explanation, it makes things more clear. I guess that makes sense and would even be a great fit for our custom implementation too, as we indeed have such "lint-violation-free branch".

fredpi:
> There are a few correctness issues here:
> 
> 1. If any of the configuration files used by SwiftLint are modified, that should force all files to be re-linted, but currently this is ignored
> 2. Same thing when the Swift version changes.
> 
> There might be a way to integrate into the existing `LinterCache` caching system to resolve this, but it's not clear to me that there's a way to do that without first traversing the file system, which in many ways defeats the purpose of this feature.
> 
> A naive way to solve the first issue could be to check if there are any files named `.swiftlint.yml` that were modified and then fall back to using the standard `FileManager.default` lintable file manager, but this doesn't handle the case where configuration files with different names are used, or if a configuration file that's not tracked in git is specified.
> 
> Maybe this feature is still valuable enough with these correctness to ship as-is, but I'd like to think about our options a bit more.

@jpsim I didn't think about this thoroughly, but want to mention that for remote configurations, the folder `.swiftlint/RemoteConfigCache` is created (and `.swiftlint` gets automatically added to the `.gitignore`). Maybe we could change it so that the that the `.swiftlint/RemoteConfigCache` gets added to the `.gitignore` while the `.swiftlint` folder itself doesn't get ignored. We could then use that folder to cache hashes for the configuration used **for each individual file**. If the hash changes for one file, that means that some configuration (maybe also a nested configuration) has changed which could then trigger a relint of all files. The **big** (and probably too big) disadvantage of such a behavior would be that SwiftLint generates files that get committed to git...

But it's the only solution I can think of which actually works – there's no other way to e. g. catch the change of a remote configuration...

The best option though would probably be to just don't add configuration change management...

jpsim:
> 1. If any of the configuration files used by SwiftLint are modified, that should force all files to be re-linted, but currently this is ignored
> 2. Same thing when the Swift version changes.

I think I struck a good balance between correctness and usefulness with my latest change.

If any files named `.swiftlint.yml`, or the configuration file(s) passed using `--config` have changed according to git, we'll fall back to linting all files as found by traversing the file system.

This won't work if remote configurations change, or if the root configuration is not tracked by git, whether because it's out of tree or in `.gitignore`.

I don't think it's very useful to fall back to linting all files when the Swift version changes because we don't store the Swift version used when the stable git revision was considered stable.

All in all, I think this is ready for usage as-is. My recommendation would be that this be used only as a local development optimization "fast-path" and that a continuous integration system always perform a full lint (although that full lint can also be accelerated by the linter cache, which isn't vulnerable to any known correctness issues).

I'll still need to write some documentation around this before we can merge though.

If anyone has objections or suggestions, please let me know!

stevelandeyasana:
Chiming in to say I'd definitely use this and it looks like it's in limbo. We'd like to run SwiftLint only on files changed vs our main branch as an Xcode build step, and linting the entire codebase takes too long.

jpsim:
@stevelandeyasana Can you try this out this branch and let me know if this works for your workflows? You might want to pull in recent changes from the `master` branch too.

ObjectiveCesar:
Hello everyone, this very useful and time saving addition is laying around almost a year new and still hasn't found it's way into a release. Is this even considered to get released?

getogrand:
Please merge this!

jpsim:
@ObjectiveCesar @getogrand have you used this in your workflows? Have you hit any issues, does this behave how you'd expect? I'm looking for feedback before merging.

getogrand:
> @ObjectiveCesar @getogrand have you used this in your workflows? Have you hit any issues, does this behave how you'd expect? I'm looking for feedback before merging.

Not yet, but I'm going to build this and try it out on our commercial source code, which has 6k swift cloc.
It may take a few weeks.

benjaminbreier:
I would definitely use this

PR #3503, opened 2021-01-22
Revert "Merge pull request #3473 from realm/feature/3472-fix-custom-rules-merging"

jpsim:
This reverts commit 537e53f6b395b7aad3aa9b3ee3588f820359d330, reversing changes made to ba49f7d3097d51b7f55b8df8b8a9f4c874fc5d79.

This caused custom rules in our private Lyft iOS repo to become unrecognized.

PR #3538, opened 2021-02-25
Multiple fixes for custom rules

fredpi:
This is the next attempt of https://github.com/realm/SwiftLint/pull/3473 which was reverted in https://github.com/realm/SwiftLint/pull/3503.

It builds upon https://github.com/realm/SwiftLint/pull/3473 and adds new tests for + fixes the issue that caused the need for temporary reverting as described in https://github.com/realm/SwiftLint/pull/3503.

The reason for the issue was that a mechanism that implicitly marks all custom rules as active in an `only_rules` mode scenario (iff the `custom_rules` rule is active) was only happening when merging configuration files, not for single configuration file scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,21 @@
 
 #### Bug Fixes
 
+* Fix `custom_rules` merging when the parent configuration is based on
+  `only_rules`.  
+  [Frederick Pietschmann](https://github.com/fredpi)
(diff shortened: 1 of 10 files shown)


jpsim (approved):
Amazing 😍 

Thanks for your patience and persistence despite my lack of details when originally reverting this.

I've confirmed this works in the project where I had originally detected the regression, so I'm merging this 🥳

PR #3545, opened 2021-03-01
Add integration with pre-commit tool

JesseCrocker:
[pre-commit](https://pre-commit.com/) is a popular tool used for managing git precommit hooks. This PR adds a .pre-commit-hooks.yaml file to allow swiftlint to be used with pre-commit.

Note: I am not proposing to use pre-commit in this project, this allows other projects that use pre-commit to more easily use swiftlint.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,7 +1,7 @@
--   id: swiftlint
-    name: SwiftLint
-    description: "Check swift files for issues with SwiftLint"
-    entry: swiftlint
-    language: swift
-    types: [swift]
-    pass_filenames: false
+- id: swiftlint
+  name: SwiftLint
+  description: "Check swift files for issues with SwiftLint"
+  entry: "swiftlint --quiet --"
+  language: swift
+  types: [swift]
+  pass_filenames: true
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,3 +4,4 @@
     entry: swiftlint
     language: swift
     types: [swift]
+    pass_filenames: false
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,6 @@
+-   id: swiftlint
+    name: SwiftLint
+    description: "Check swift files for issues with SwiftLint"
+    entry: swiftlint
+    language: swift
+    types: [swift]
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,6 +1,6 @@
 - id: swiftlint
   name: SwiftLint
-  description: "Check swift files for issues with SwiftLint"
+  description: "Check Swift files for issues with SwiftLint"
   entry: "swiftlint --quiet --"
   language: swift
   types: [swift]


jpsim:
Thanks for the PR!

What advantages do we get by checking this file in to source control as opposed to documenting this in the readme?

If we end up merging https://github.com/realm/SwiftLint/pull/3536, would it make sense to change the SwiftLint command to `swiftlint --stable-git-revision=HEAD`? What does precommit recommend?

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,6 @@
+-   id: swiftlint
```
jpsim, line 1: Minor nit: we indent `.swiftlint.yml` with two spaces, can you do the same here?

JesseCrocker:
> What advantages do we get by checking this file in to source control as opposed to documenting this in the readme?

Having this file checking into the repo allows projects to easily specify a specific version of swiftlint to use, ensuring that all developers on the project are using the same version. Otherwise it either requires wrapping swiftlint with a shell script that checks the version(what my company currently does), or enforcing version by social convention.

It also makes the install of swiftlint be fully isolated from the rest of the system, which is really helpful when working on multiple projects that are using different versions of swiftlint.

> If we end up merging #3536, would it make sense to change the SwiftLint command to swiftlint --stable-git-revision=HEAD? What does precommit recommend?

pre-commit has built in support to do this, but it works by passing file names as command line args, but making this work with swiftlint requires wrapping it with a shell script. This is the wrapper i use currently:
```
#!/usr/bin/env bash
if [[ -z `which swiftlint` ]]; then
	echo "SwiftLint not found. Run 'brew install swiftlint'"
	exit 1
fi

SWIFTLINT_VERSION="0.42.0"
if [ `swiftlint version` != $SWIFTLINT_VERSION ]; then
	echo "Incorrect swiftlint version. Install $SWIFTLINT_VERSION"
	exit 1
fi

for file in "$@"; do
    SCRIPT_INPUT_FILE_COUNT=1 SCRIPT_INPUT_FILE_0=$file swiftlint lint --quiet --use-script-input-files
done
``` 

The alternative to having pre-commit call swiftlint with command file names as command line args is to add `pass_filenames: false` to the hook definition, which i forgot at first but added in my last commit.  Combined that with `--stable-git-revision=HEAD` would be great as it would avoid processing unchanged files.

.pre-commit-hooks.yaml
JesseCrocker: Done

jpsim:
> Having this file checking into the repo allows projects to easily specify a specific version of swiftlint to use, ensuring that all developers on the project are using the same version.

Oh I see, so pre-commit actually handles downloading the git repository. Does it also handle building? Is that what the `language` key does? Tell it that it can build the tool with `swift build -c release swiftlint`? Sorry for all the questions, but I'd like to really understand this change before merging.

Since pre-commit already handles what files to check, seems best to use `pass_filenames: true` (or omit it if it's the default behavior), otherwise what SwiftLint checks could be inconsistent with other tools you integrate with pre-commit.

---

As a side note, your script will run a new instance of SwiftLint for every file that has changed, which could be very slow.

> for file in "$@"; do
>     SCRIPT_INPUT_FILE_COUNT=1 SCRIPT_INPUT_FILE_0=$file swiftlint lint --quiet --use-script-input-files
> done

Instead, you could pass the files directly:

```swift
swiftlint lint --quiet -- "$@"
```

This will only run the SwiftLint executable once, linting all the files from `"$@"`.

JesseCrocker:
> Oh I see, so pre-commit actually handles downloading the git repository. Does it also handle building? Is that what the language key does? Tell it that it can build the tool with swift build -c release swiftlint? Sorry for all the questions, but I'd like to really understand this change before merging.

Yeah, exactly.  Once this is merged, integrating into a project is as simple as adding the following to the projects .pre-commit-config.yaml file. pre-commit handles downloading and building the correct version.
```
  - repo: https://github.com/realm/SwiftLint
    rev: 0.43.0
    hooks:
      - id: swiftlint
```

I didn't realize swiftlint would accept a list of filenames like that, config updated.

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
+- id: swiftlint
+  name: SwiftLint
+  description: "Check swift files for issues with SwiftLint"
```
jpsim, line 3: Nit:

```suggestion
  description: "Check Swift files for issues with SwiftLint"
```

jpsim:
Thanks for answering all my questions @JesseCrocker!

Did you test this integration to make sure it works well with pre-commit?

Can you add a changelog entry according to the format specified in [CONTRIBUTING.md](https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#tracking-changes) and document how to use this in the readme?

elliottwilliams:
One thing that might be useful is to declare an additional hook that runs `swiftlint lint --fix`, so that users can declare a version of the hook which autocorrects violations. Pre-commit will block a commit if one of its hooks makes changes to a file, so this provides a decent UX for users: they try to commit, swiftlint runs and fixes violations for them, they `git add` the changes, and they commit again.

(Also, bumping this PR because I'd love to see this make it in 😄)

hannseman:
I can confirm that this PR works with pre-commit, I use this `.pre-commit-config.yaml`
```yml
repos:
  - repo: https://github.com/JesseCrocker/SwiftLint
    rev: f40fa78
    hooks:
      - id: swiftlint
```

@JesseCrocker do you plan on picking up this PR? If I'm reading this correctly the only missing piece here is docs. If you don't have the time I can try and finish up this in a new PR.

jpsim:
> If I'm reading this correctly the only missing piece here is docs. If you don't have the time I can try and finish up this in a new PR.

Correct, PR looks good, just needs a changelog entry & documentation in the README. @hannseman want to pick that up?

hannseman:
Sure thing! I've opened up https://github.com/realm/SwiftLint/pull/3714 to continue the work started here.

jpsim:
#3714 was merged

PR #3547, opened 2021-03-02
File false negatives in unused_declaration

jpsim:
When a declaration refers to an Objective-C type but doesn't have an explicit `@objc` attribute.

It would have the string `@objc` embedded in its USR. So add a space so we only trigger this condition when there's an explicit `@objc` attribute in the declaration source code.

The triggering example added here didn't trigger before this change.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/UnusedDeclarationRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/UnusedDeclarationRule.swift
@@ -171,8 +171,7 @@ private extension SwiftLintFile {
 
         let cursorInfo = self.cursorInfo(at: nameOffset, compilerArguments: compilerArguments)
 
-        if let annotatedDecl = cursorInfo?.annotatedDeclaration,
-            ["@IBAction", "@objc"].contains(where: annotatedDecl.contains) {
+        if cursorInfo?.annotatedDeclaration?.contains("@objc ") == true {
             return nil
         }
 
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Rules/Lint/UnusedDeclarationRule.swift
```diff
@@ -171,8 +171,7 @@ private extension SwiftLintFile {
 
         let cursorInfo = self.cursorInfo(at: nameOffset, compilerArguments: compilerArguments)
 
-        if let annotatedDecl = cursorInfo?.annotatedDeclaration,
-            ["@IBAction", "@objc"].contains(where: annotatedDecl.contains) {
```
keith, line 175: I'm surprised the IBAction isn't required here still, but the tests cover it so nbd if they pass!

keith (approved):

Source/SwiftLintFramework/Rules/Lint/UnusedDeclarationRule.swift
jpsim: It's already covered by `declarationAttributesToSkip`:

https://github.com/realm/SwiftLint/blob/fd97e2259c885aee7d8ec3120c35de57fde592f5/Source/SwiftLintFramework/Rules/Lint/UnusedDeclarationRule.swift#L324-L330

keith: ah sorry I was misremembering this change https://github.com/realm/SwiftLint/pull/3458

PR #3549, opened 2021-03-03
Fix parsing xcode logs for analyzer rules for target names with spaces

jpsim:


Source/swiftlint/Helpers/CompilerArgumentsExtractor.swift
```diff
@@ -2,62 +2,18 @@ import Foundation
...
 
 struct CompilerArgumentsExtractor {
-    static func allCompilerInvocations(compilerLogs: String) -> [String] {
-        var compilerInvocations = [String]()
+    static func allCompilerInvocations(compilerLogs: String) -> [[String]] {
+        var compilerInvocations = [[String]]()
         compilerLogs.enumerateLines { line, _ in
             if let swiftcIndex = line.range(of: "swiftc ")?.upperBound, line.contains(" -module-name ") {
-                let invocation = line[swiftcIndex...]
-                    .components(separatedBy: " ")
```
jpsim, line 10: This was the cause of the bug. The rest of the changes in this PR are just refactoring.

jpsim:
My buildkite machine is offline for maintenance, preventing OSSCheck from running, but that script doesn't run any analyzer rules, so I'll merge without it.

PR #3550, opened 2021-03-03
Fix OSSCheck noise

jpsim:
Try a few things:

1. Sort InclusiveLanguageRule terms since they're a Set and don't have deteministic ordering (will definitely help)
2. Fetch remote at the beginning of the script (likely to help)
3. Only fetch remote once (likely to help)
4. Sort branch & master output before diffing (unlikely to help)
5. Change diff filter from `d` to `AMRCU` (no idea if it will help)

### DIFF
--- a/script/oss-check
+++ b/script/oss-check
@@ -15,7 +15,7 @@ require 'erb'
 
 @options = {
   branch: 'HEAD',
-  iterations: 1,
+  iterations: 5,
   skip_clean: false,
   verbose: false
 }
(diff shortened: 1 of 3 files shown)


script/oss-check
```diff
@@ -15,7 +15,7 @@ require 'erb'
 
 @options = {
   branch: 'HEAD',
-  iterations: 5,
+  iterations: 1,
```
jpsim, line 18: Will revert this, just trying to validate these changes as quickly as possible.

script/oss-check
```diff
@@ -292,17 +292,7 @@ end
 @working_dir = 'osscheck'
 @repos = [
   Repo.new('Aerial', 'JohnCoates/Aerial'),
-  Repo.new('Alamofire', 'Alamofire/Alamofire'),
```
jpsim, line 295: Will revert this, just trying to validate these changes as quickly as possible.

jpsim:
My buildkite machine is offline for maintenance, preventing OSSCheck from running

PR #3551, opened 2021-03-03
Add an identifier to Reporters

walterg2:
This request adds an identifier to the location object so that consumers of the SwiftLint results who are not accessing via an IDE can quickly determine and report out about where the violation occurred.

This PR gets an identifier from the AST and adds it to the Location information when processing the `FunctionBodyLengthRule`. This identifier is reported out in the following reporters currently:
* `JSONReporter`
* `CSVReporter`
* `MarkdownReporter`
* `GitLabJUnitReporter`

### DIFF
--- a/Source/SwiftLintFramework/Rules/Metrics/CyclomaticComplexityRule.swift
+++ b/Source/SwiftLintFramework/Rules/Metrics/CyclomaticComplexityRule.swift
@@ -19,7 +19,7 @@ public struct CyclomaticComplexityRule: ASTRule, ConfigurationProviderRule {
                 }
                 if false { }
             }
-            """),
+            """, methodName: "f1()"),
             Example("""
             func f(code: Int) -> Int {
                 switch code {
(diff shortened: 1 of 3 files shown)


walterg2:
Just updated to take in other accepted changes and get the build passing. Any chance this could be reviewed for inclusion in the next release of SwiftLint?

jpsim:
@walterg2 looks like the identifiers are always null in the reporters?

walterg2:
@jpsim Apologies. It looks like a merge removed that addition to the ReporterTests. This has been added back in.

jpsim:
This is a generally useful feature for all rules, not just the function body length rule. Is there a way all rules that point to an identifier could benefit from this? Otherwise this will feel like an incomplete feature.

walterg2:
@jpsim I can take a look to see about implementing it for other rules as well. I shied away from this as an initial implementation due to the dynamic nature of how tests are currently built for other rules versus the Function Body rule.

jpsim:
If we do this it needs to work with all rules where this applies. Not all violations point to a specific identifier though, and not all identifiers are methods.

PR #3553, opened 2021-03-03
Require Swift 5.3 or higher to build

jpsim:
Swift 5.3 has been available for close to 6 months now, seems like a good time to require it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,19 +60,7 @@ running it.
 ### Installing from source:
 
 You can also build and install from source by cloning this project and running
-`make install` (Xcode 11.4 or later).
-
-### Known Installation Issues On MacOS Before 10.14.4
-
-Starting with [SwiftLint 0.32.0](https://github.com/realm/SwiftLint/releases/tag/0.32.0), if you get
-an error similar to `dyld: Symbol not found: _$s11SubSequenceSlTl` when running SwiftLint,
-you'll need to install the [Swift 5 Runtime Support for Command Line Tools](https://support.apple.com/kb/DL1998).
-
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -241,7 +229,8 @@ Here's a reference of which SwiftLint version to use for a given Swift version.
 | Swift 3.x       | SwiftLint 0.25.1                 |
 | Swift 4.0-4.1.x | SwiftLint 0.28.2                 |
 | Swift 4.2.x     | SwiftLint 0.35.0                 |
-| Swift 5.x       | Latest                           |
+| Swift 5.0-5.2.x | SwiftLint 0.43.0                 |
```
norio-nomura, line 232: The tag history of https://hub.docker.com/r/norionomura/swiftlint looks like the following.
https://hub.docker.com/r/norionomura/swiftlint/tags?name=swift-5.0
https://hub.docker.com/r/norionomura/swiftlint/tags?name=swift-5.1
```suggestion
| Swift 5.0.x     | SwiftLint 0.39.2                 |
| Swift 5.1.x     | SwiftLint 0.41.0                 |
| Swift 5.2.x     | SwiftLint 0.43.0                 |
```

jpsim: Thanks for flagging! I wonder if this table is actually useful to most people though. Maybe I should just remove it?

jpsim: I just removed it.

These weren't exactly accurate because SwiftLint should still work with older Swift versions long after we update the minimum version of Swift needed to _build_ the tool. So I doubt this table was very helpful to anyone.

PR #3568, opened 2021-03-11
Add vscode extension to README.md

vknabel:
**Full disclosure:** I maintain this extension.


sethfri:
[I don't see](https://marketplace.visualstudio.com/search?term=swiftlint&target=VSCode&category=All%20categories&sortBy=Relevance) any other SwiftLint extensions for VS Code, and this one was called out in an [NSHipster article](https://nshipster.com/vscode/), so this is good to merge IMO.

Thanks for the contribution!

vknabel:
Thanks for your review and work for SwiftLint!

PR #3597, opened 2021-04-13
[bug fix] expiring todo regex

namolnad:
Currently the regex for the expiring_todo rule requires zero-padded dates. (i.e. 01/01/2021 vs 1/1/2021). This loosens the regex digit searching to allow for dates which are not zero-padded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@
   with a static member (e.g. `if number == .zero`).  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#3562](https://github.com/realm/SwiftLint/issues/3562)
+* Fix the regex for expiring-todos  
+  [namolnad](https://github.com/namolnad)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -15,7 +15,9 @@
 
 #### Bug Fixes
 
-* None.
+* Fix the regex for expiring-todos  
+  [namolnad](https://github.com/namolnad)
+  [#3597](https://github.com/realm/SwiftLint/issues/3597)
```
marcelofabri, line 20: can you change the link here to `/pull/3597` (instead of _issue_)? Alternatively, it's fine to remove this line completely since most of times we haven't been linking to pull requests, only issues

marcelofabri (approved):

CHANGELOG.md
namolnad: Updated, thanks!

sethfri (approved):
Looks great, thanks for the contribution!

PR #3604, opened 2021-04-18
Add contributing screenshot with better example arguments

revolter:
Add a screenshot that shows the `--no-cache` argument, because this is
what should be used in most cases, since the cache is enabled by
default, so the debugger won't stop in the implementation of a rule on a
subsequent run.

Also update the second screenshot from the same macOS version (Big Sur)
for consistency.

[Preview](https://github.com/revolter/SwiftLint/blob/patch-1/CONTRIBUTING.md).


marcelofabri:
Thanks!

PR #3612, opened 2021-04-26
Add test functions with parameters to TestCaseAccessibilityRule

keith:
If a function starts with `test` but takes some parameters, it is not
actually a test.


jpsim (approved):
Nice

PR #3617, opened 2021-04-29
Run tests on Swift 5.4 in CI (Linux)

marcelofabri:


jpsim:
Continued in https://github.com/realm/SwiftLint/pull/3710

PR #3618, opened 2021-04-29
Fix some rules when using Swift 5.4

marcelofabri:


username0x0a:
👆 Would be good to expedite this in a release, too. 💡🙏

username0x0a:
Is there a way to expedite the release? 🤔

It's quite annoying, at this moment we have SwiftLint in the pipeline out of sync (sticking with an older Swift version so no warnings are ignored) but in case or major Xcode/bundled Swift updates, hooking a SwiftLint update should be done ASAP – even the testing cycle for this Xcode release was >3 months.

Will be looking for an update soon hopefully. 🙏

Issue #3590, opened 2021-04-05
expiring_todo severity is not configurable


cuddergambino:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

The expiring todo configuration does not take into account severity changes. For example, adding the following to `.swiftlint` does nothing.
```
expiring_todo:
  warning: reached_or_passed_expiry_severity
```
This is the line in a `.swift` file:
```
 // TODO: [04/01/2021] test
```
The compiler displays an error when a warning should be displayed.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint lint
```
Output:
```
 error: ExpiringTodo Violation: TODO/FIXME has expired and must be resolved. (expiring_todo)
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.43.1
* Installation method used (Homebrew, CocoaPods, building from source, etc)? Homebrew
* Paste your configuration file:

```yml
### Rules directory: https://realm.github.io/SwiftLint/rule-directory.html
opt_in_rules:
  - expiring_todo

disabled_rules:
  - todo
  - trailing_comma

expiring_todo:
  warning: reached_or_passed_expiry_severity
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? No
  If so, paste their relative paths and respective contents.
* Which Xcode version are you using (check `xcodebuild -version`)? Xcode 12.4 Build version 12D4e
* Do you have a sample that shows the issue? Run `echo "[string here]" | swiftlint lint --no-cache --use-stdin --enable-all-rules` `echo "// TODO: [04/01/2021]" | swiftlint line --no-cache --use-stdin --enable-all-rules`
  to quickly test if your example is really demonstrating the issue. If your example is more
  complex, you can use `swiftlint lint --path [file here] --no-cache --enable-all-rules`.

```swift
// This triggers a violation:
// TODO: [04/01/2021]
```

marcelofabri:
Two things here:

1. `reached_or_passed_expiry_severity` should be the key:

```yml
expiring_todo:
  reached_or_passed_expiry_severity: warning
```

2. There's a bug in the current implementation where the documented key is not the same as the key that's used. You can use  `expired_severity: warning` for now to workaround it. We should fix the rule to accept both keys.

cuddergambino:
[...]

PR #3620, opened 2021-04-30
Fix expiring_todo configuration key

namolnad:
Addresses expiring_todo severity configuration https://github.com/realm/SwiftLint/issues/3590:

- Updates documentation to reflect intended configuration key for `expiring_todo` rule
- Adds alternative config syntax handling
- Adds unit test for alternative config key

### DIFF
--- a/Source/SwiftLintFramework/Rules/RuleConfigurations/ExpiringTodoConfiguration.swift
+++ b/Source/SwiftLintFramework/Rules/RuleConfigurations/ExpiringTodoConfiguration.swift
@@ -13,7 +13,7 @@ public struct ExpiringTodoConfiguration: RuleConfiguration, Equatable {
 
     public var consoleDescription: String {
         return "(approaching_expiry_severity) \(approachingExpirySeverity.consoleDescription), " +
-        "(reached_or_passed_expiry_severity) \(expiredSeverity.consoleDescription)"
+        "(expired_severity) \(expiredSeverity.consoleDescription)"
     }
 
     private(set) var approachingExpirySeverity: SeverityConfiguration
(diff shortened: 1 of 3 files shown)


SimplyDanny (approved):
Sorry for the late feedback. The PR looks good to me! If you rebase the changes and resolve the conflicts, we can have it merged.

Source/SwiftLintFramework/Rules/RuleConfigurations/ExpiringTodoConfiguration.swift
```diff
@@ -53,7 +53,7 @@ public struct ExpiringTodoConfiguration: RuleConfiguration, Equatable {
         if let approachingExpiryConfiguration = configurationDict["approaching_expiry_severity"] {
             try approachingExpirySeverity.apply(configuration: approachingExpiryConfiguration)
         }
-        if let expiredConfiguration = configurationDict["expired_severity"] {
+        if let expiredConfiguration = configurationDict["expired_severity"] ?? configurationDict["reached_or_passed_expiry_severity"] {
```
SimplyDanny, line 56: We could report a warning here if the old key is still used.

SimplyDanny:
Now that #3626 fixed the option name in the configuration description and the rename of the option happened already more than one year ago I think we can be reassured to ignore the old option name completely. That being said, this PR may be declined. I still want to thank you again for the contribution!

SimplyDanny:
Now that #3626 fixed the option name in the configuration description and the rename of the option happened already more than one year ago I think we can be reassured to ignore the old option name completely. That being said, this PR may be declined. I still want to thank you again for the contribution!

PR #3626, opened 2021-05-05
Error on bad expiring todo date formats.

chrispomeroyhale:
## Motivation
For expiring todos, there is currently a lack of documentation on the default date format. This currently requires a deep dive into SwiftLint to figure out what it is. This is a detractor of this tool in a team setting. Unfortunately, the default is an American date format which is a little surprising. Although it may be too late to change the default to be in line with ISO 8601, we can reject badly formatted dates which otherwise never expire or unexpectedly expire.

## What Changed
1. A bad TODO date format throws an error by default. This can be configured
2. To alleviate the lack of documentation, `consoleDescription` has been updated to correct and include all the latest properties
3. Updated the regex to accept any 1-4 letter date format in any position. In the future we might consider matching anything within the delimiters to the `date_format` and remove `date_separator`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,10 @@
   [JP Simard](https://github.com/jpsim)
   [#3920](https://github.com/realm/SwiftLint/issues/3920)
 
+* Error by default on bad expiring todo date formatting.  
+  [Christopher Hale](https://github.com/chrispomeroyhale)
+  [#3636](https://github.com/realm/SwiftLint/pull/3626)
+
 ## 0.47.0: Smart Appliance
 
 #### Breaking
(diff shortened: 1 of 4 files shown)


SimplyDanny:
Thank you for the contribution and sorry for the late feedback!

The changes make sense to me. Would you like to rebase them and fix the conflicts?

> In the future we might consider matching anything within the delimiters to the `date_format` and remove `date_separator`.

This would indeed be a reasonable change in the future. The `date_format` can handle this. There is actually no need for the `data_separator` configuration.

chrispomeroyhale:
@SimplyDanny Rebased

SimplyDanny (approved):
Thank you, @chrispomeroyhale!

PR #3630, opened 2021-05-09
Update Number of Rules

rogerluan:
Simple and neat :) we're at 204 rules I think 🎉 
I couldn't find occurrences of "100" in the other non-English README files, but I wouldn't find them if they're not represented in using numbers 🤷 I searched for "100" and "150" literally

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,7 +226,7 @@ environment variable.
 
 ## Rules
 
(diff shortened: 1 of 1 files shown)


marcelofabri (approved):
thanks!

PR #3635, opened 2021-05-19
Disallow none case

### LINKED ISSUE #3624: Rule Request: Disallow enum & static/class properties named `none`
### New Issue Checklist

### New rule request

Disallow `enum` & `static`/`class` properties named `none`.

#### Why should this rule be added?

Naming enum cases (and static/class properties) `none` can cause unintended or unexpected behaviours that are not caught by the compiler. This is due to Swift's built-in `Optional<T>` enum that has two cases, one named `some`, and one named `none`, to represent the existence of a value and no value, respectively.

Before we go any further, let’s look at an example:

```swift
enum MyEnum {
  case a
  case none
}

// Somewhere else
let value: MyEnum = .a
print(value == .none)
```

In the above code, we have an enum that has a case named `a` and another case named `none`. Somewhere else in the code, someone is checking whether the provided enum value is equal to `none`.

What happens now, if someone refactors this code and changes the name of the `none` case, or just removes it entirely?

```swift
enum MyEnum {
  case a
}

// Somewhere else
let value: MyEnum = .a // Note that this constant is not optional
print(value == .none)
```

Unfortunately, this code will **still compile**, even though `none` no longer exists as a case name on `MyEnum`. 🥲

Remember again that Swift's built-in `Optional` enum has a case named `none`. This code still compiles without warning or error because Swift has a built-in `==` function on `Optional` which allows us to check the equality of two optional types. The signature looks like this:

```swift
static func == (lhs: MyEnum?, rhs: MyEnum?) -> Bool
```

When we write `MyEnum.a == .none`, after removing our enum case named `none`, the compiler now thinks we're writing the shorthand for `MyEnum.a == Optional<MyEnum>.none`, and so the code compiles, **even though the `value` constant is not optional at all**.

You can test this out for yourself in a Playground very easily. I'm not sure about the protocol for uploading attachments that have executable code in, so I won't upload one myself but it's trivial to create your own example.

It's also important to note this is also true for any `Equatable` type with `static` (or `class`) members:

```swift
struct MyStruct: Equatable {
  static let a = MyStruct()
  static let none = MyStruct()
}
let value: MyStruct = .a
print(value == .none) // true
```

The above code is very similar to the previous code example, except it's now an `Equatable` struct. The above code would print out `true` when checking `value == .none` because `a` and `none` are currently equal.

If we removed the `none` static property, the code will still compile, but our code is now broken because it will print out `false` since `a` is not equal to `Optional<MyStruct>.none`, which the compiler thinks you're checking:

```swift
struct MyStruct: Equtable {
  static let a = MyStruct()
}
let value: MyStruct = .a
print(value == .none) // false
```

In this simple example it would be easy to catch during development or during code review, but in a large codebase or in a distributed framework used by third parties, it could easily be missed without very high test coverage.

Apple themselves use an [enum case named `none`](https://developer.apple.com/documentation/localauthentication/labiometrytype/none) in the LocalAuthentication framework for describing the lack of biometric capabilities of the current device. If for some reason Apple decided to rename that case to something else or remove it entirely (since they could just use an actual optional value), many codebases would still continue to compile but have unexpected behaviours as a result.

In practice it's unlikely Apple will change it (they'd at least deprecate it first), but it's an example of how this could be a problem if `none` is allowed as a case name when writing our own internal code, or for distribution via open or closed source frameworks without strict deprecation procedures.

#### Triggers

Examples of what would trigger violations:

```swift
enum MyEnum {
  case none
}

enum MyEnum {
  case a, none  // All the permutations of this type of declaration
}

enum MyEnum {
  case none(AssociatedValue) 
}

enum MyEnum {
  case a, none(AssociatedValue)
}

struct MyStruct {
  static let none = MyStruct()
}

struct MyStruct {
  static let a = MyStruct(), none = MyStruct()
}

class MyClass {
  static let none = MyClass()
}

class MyClass {
  static let a = MyClass(), none = MyClass()
}

class MyClass {
  static var none = MyClass()
}

class MyClass {
  static var a = MyClass(), none = MyClass()
}

class MyClass {
  class var none: MyClass { MyClass() } 
}
```

Examples of what would not trigger violations:

```swift
enum MyEnum {
  case noneAtAll // Or anything that contains the name "none" but isn't just the word "none"
}

switch optionalValue {
  case .a: break
  case .none: break // Still allow explicitly checking `Optional<T>.none` for an optional value, which the compiler will catch if `optionalValue` isn't actually optional.
}

print(value == .none) // Still allow explicitly comparing for Swift's `Optional<T>.none` case
```

#### Should the rule be configurable?

I'll delegate this question to the community but I can't think off the top of my head what would be configurable about it.

#### Should the rule be opt-in or enabled by default?

I'd personally choose enabled by default, but potentially as a warning rather than an error.

KaneCheshire:
Fixes #3624


Source/SwiftLintFramework/Rules/Idiomatic/DiscouragedNoneName.swift
```diff
@@ -0,0 +1,217 @@
...
+            """),
+            Example("""
+            class MyClass {
+                var none = MyClass()
+            }
+            """)
+        ],
+        triggeringExamples: [
+            Example("""
+            enum MyEnum {
```
marcelofabri, line 87: can you please update the triggering examples with the violation marker (↓) so that the unit tests can also validate where the violation is triggered?

marcelofabri:
@KaneCheshire awesome first contribution! Just one minor comment, but this looks great! Thanks for implementing it🥇

Source/SwiftLintFramework/Rules/Idiomatic/DiscouragedNoneName.swift
KaneCheshire: Yep of course! I did wonder what the arrow was for, it makes sense now. Updated, thanks!

KaneCheshire:
@marcelofabri Thanks! 🙌 was really cool to be able to contribute

KaneCheshire:
Thanks for merging @marcelofabri. Really appreciate it! I have one final question, do you know when it's likely to be released? Only so I can add some sort of timeline to our internal workflow for updating and enabling the new rule. Thanks!

Issue #3633, opened 2021-05-17
Autocorrect on unneeded_parentheses_in_closure_argument rule does not insert whitespace


p-x9:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug
If a tuple is used in the argument of a closure, and there is no space between the tuple and "in", the code will be wrong when autocorrected.(See the 'Sample.swift' code below.)

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint lint --path Sample.swift --no-cache --enable-all-rules --fix
Correcting Swift files at paths Sample.swift
Correcting 'Sample.swift' (1/1)
/Users/p-x9/Desktop/Sample.swift:2:21 Corrected Unneeded Parentheses in Closure Argument
Done inspecting 1 file for auto-correction!
```

### Environment

* SwiftLint version : 0.43.1
* Installation method used :Homebrew
* Paste your configuration file: none
* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations): No
* Which Xcode version are you using (check `xcodebuild -version`): Xcode 12.5,Build version 12E262
* Do you have a sample that shows the issue? Run `echo "[string here]" | swiftlint lint --no-cache --use-stdin --enable-all-rules`
  to quickly test if your example is really demonstrating the issue. If your example is more
  complex, you can use `swiftlint lint --path [file here] --no-cache --enable-all-rules`.

**Sample.swift**
```swift
let dictionary = [String: Int]()
dictionary.forEach {(key, value)in
    print(key, value)
}
```

**Sample.swift(autocorrected)**
```swift
let dictionary = [String: Int]()
dictionary.forEach {key, valuein
    print(key, value)
}
```

marcelofabri:
@p-x9 This shouldn't be _too_ hard to fix, do you want to give it a shot?

p-x9:
@marcelofabri   please confirm the pullrequest.

PR #3638, opened 2021-05-20
Fix Autocorrect on no space between the tuple and "in" in unneeded_parentheses_in_closure_argument

p-x9:
Fix Autocorrect on no space between the tuple and "in" in unneeded_parentheses_in_closure_argument.
Related to #3633


marcelofabri:
Can you please add some tests? It should be as easy as adding a new dictionary entry in `corrections`, which is part of the rule description

p-x9:
Added a test and some fixs.

marcelofabri (approved):

marcelofabri:
Thanks!

PR #3653, opened 2021-06-08
Allow unused_setter_value for overrides

### LINKED ISSUE #2585: `unused_setter_value` incorrectly triggers for overridden properties.
### New Issue Checklist

### Describe the bug

`unused_setter_value` incorrectly triggers for overridden properties.

I use the following code that I gave as [a Stack Overflow answer](https://stackoverflow.com/questions/27988279/uitextview/49428307#49428307):
```swift
/// sub-class to disallow selection.
class UnselectableTappableTextView: UITextView {
    // required to prevent blue background selection from any situation
    override var selectedTextRange: UITextRange? {
        get { return nil }
        // This triggers a SwiftLint violation:
        set {}
    }
}
```

Having an empty setter is the desired behavior when overriding.

### Environment

* SwiftLint version 0.30.0
* Xcode 10.1
* Swift 4.2

grosem:
Fixes #2585.  Happy to make changes if necessary 😎

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/UnusedSetterValueRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/UnusedSetterValueRule.swift
@@ -156,7 +156,7 @@ public struct UnusedSetterValueRule: ConfigurationProviderRule, AutomaticTestabl
             }
 
             if dict.enclosedSwiftAttributes.contains(.override) &&
-                !file.syntaxMap.kinds(inByteRange: setterByteRange).contains(where: {!$0.isCommentLike}) {
+                !file.syntaxMap.kinds(inByteRange: setterByteRange).contains(where: { !$0.isCommentLike }) {
                 return nil
             }
 
--- a/Source/SwiftLintFramework/Rules/Lint/UnusedSetterValueRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/UnusedSetterValueRule.swift
@@ -102,6 +102,16 @@ public struct UnusedSetterValueRule: ConfigurationProviderRule, AutomaticTestabl
                     Persister.shared.aValue = aValue
                 }
             }
+            """),
+            Example("""
+            override var aValue: String {
+                get {
+                    return Persister.shared.aValue
+                }
+                ↓set {
+                    Persister.shared.aValue = aValue
+                }
+            }
             """)
         ]
     )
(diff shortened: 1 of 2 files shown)


marcelofabri:
@grosem what do you think about checking that the `set` body is empty? I feel like this change as is could end up introducing issues on overrides where the author would want to use `newValue` but made a mistake.

grosem:
Makes sense, I'll try to add that 👍

grosem:
@marcelofabri I tried to find the easiest way to accomplish that, what do you think?
The checks still run into an error here which I cannot reproduce locally - not really sure what the problem is.

zvonicek:
Hi, thanks for this PR! I'm seeing a similar problem in my codebase and would like to see the PR merged. Can you try to revive this PR @grosem?

Current problem I'm seeing in the checks on this PR – `Closure expressions should have a single space inside each brace.` – seem to be because you are missing a space characters inside of the `contains(where:)` closure in UnusedSetterValueRule.swift line 159. 

Instead of `contains(where: {!$0.isCommentLike})` try to write it as `contains(where: { !$0.isCommentLike })`.

grosem:
> Hi, thanks for this PR! I'm seeing a similar problem in my codebase and would like to see the PR merged. Can you try to revive this PR @grosem?
> 
> Current problem I'm seeing in the checks on this PR – `Closure expressions should have a single space inside each brace.` – seem to be because you are missing a space characters inside of the `contains(where:)` closure in UnusedSetterValueRule.swift line 159.
> 
> Instead of `contains(where: {!$0.isCommentLike})` try to write it as `contains(where: { !$0.isCommentLike })`.

Sure thing, thank you for your comment. In my local version I had that already fixed but didn't commit it for a reason I cannot remember... if there is anything else to adjust, please let me know!

zvonicek:
Thanks for the fix! The PR now seem to pass, hope we can get someone to review and merge this.

jpsim:
Thanks for the contribution!

Source/SwiftLintFramework/Rules/Lint/UnusedSetterValueRule.swift
```diff
@@ -41,6 +41,14 @@ public struct UnusedSetterValueRule: ConfigurationProviderRule, AutomaticTestabl
                     Persister.shared.aValue = value
                 }
             }
+            """),
+            Example("""
+            override var aValue: String {
+             get {
+                 return Persister.shared.aValue
+             }
+             set() { }
```
amomchilov, line 50: These examples get automatically run as tests, right?

This example fails for me, on v0.46.2

```shell
echo 'private var abc: Int { get { 123 } set() { } }' | swiftlint lint --no-cache --use-stdin --enable-all-rules
```

```
<nopath>:1:36: warning: Unused Setter Value Violation: Setter value is not used. (unused_setter_value)
```

Can you reproduce?

jpsim: Yes these are run as tests, please open a new issue with full steps to reproduce, what you're seeing and what you're expecting.

amomchilov: Opened https://github.com/realm/SwiftLint/issues/3863 :)

Issue #3621, opened 2021-04-30
redundant_optional_initialization does not find properties with observers (e.g. didSet) 


iressler:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

The redundant_optional_initialization rule does not trigger when a property has observers (e.g. didSet, willSet), but will without them.

For example this property will trigger redundant_optional_initialization:
```swift
var varName: String? = nil
```

but this property will not:
```swift
var varName: String? = nil {
  didSet {
    print("didSet")
  }
}
```

##### Complete output when running SwiftLint, including the stack trace and command used

```zsh
$ swiftlint lint
Loading configuration from '.swiftlint.yml'
Linting Swift files at paths
Linting 'SwiftLintTestApp.swift' (1/2)
Linting 'ContentView.swift' (2/2)
/Users/iressler/repos/SwiftLintTest/SwiftLintTest/ContentView.swift:11:23: warning: Redundant Optional Initialization Violation: Initializing an optional variable with nil is redundant. (redundant_optional_initialization)
Done linting! Found 1 violation, 0 serious in 2 files.
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)?
```zsh
swiftlint version
0.43.1
```
* Installation method used (Homebrew, CocoaPods, building from source, etc)?
CocoaPods, Homebrew, and install from source all have the same result.
* Paste your configuration file:
This occurs with an empty configuration file.

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)?
No.
* Which Xcode version are you using (check `xcodebuild -version`)?
```zsh
xcodebuild -version
Xcode 12.4
Build version 12D4e
```

* Do you have a sample that shows the issue?

This example properly triggers redundant_optional_initialization:
```zsh
echo "var varName: String? = nil" | swiftlint lint --no-cache --use-stdin
<nopath>:1:21: warning: Redundant Optional Initialization Violation: Initializing an optional variable with nil is redundant. (redundant_optional_initialization)
Done linting! Found 1 violation, 0 serious in 1 file.
```

This example does not trigger it:
```zsh
echo "var varName: String? = nil { didSet { print("something") } }" | swiftlint lint --no-cache --use-stdin
Done linting! Found 0 violations, 0 serious in 1 file.
```

iressler:
[...]

PR #3671, opened 2021-07-06
Address false positive in the redundant_optional_initialization rule when variable has observers

iressler:
This change addresses #3621 where the redundant_optional_initialization rule fails when the variable has observers.

I couldn't find any way to determine that the property had any observers, although I may have missed something. So I came up with a change to the regex used to (I assume) verify the match is at the end of the checked range. The change adds a check for any number of spaces and an optional `{`:

`"\\s*=\\s*nil\\b"` -> `"\\s*=\\s*nil\\b\\s*\\{?"`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@
 
 #### Enhancements
 
+* Fix false negative on `redundant_optional_initialization` rule when variable
+has observers.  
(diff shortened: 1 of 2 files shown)


iressler:
All of the new warnings introduced are exactly what I am trying to address with this change, I wasn't expecting that many though.

marcelofabri:
Thanks!

Issue #2156, opened 2018-04-13
Question: [type_name] wildcard `excluded` property


RishabhTayal:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

Is there a way to specify a wildcard in the `excluded` property of the rule `type_name`. For e.g: my class names are named as `ClassName_v2`. How can I disable the rule for this name? I tried using the following and it doesn't work

```
type_name:
  excluded: _v2
```

marcelofabri:
There's no way to specify a wildcard. You can use `allowed_symbols` to allow `_`.

Blackjacx:
Wow could you add this to the README examples? Many ppl search there first and in my case I was just lucky to discover this here in the ~300 issues.

marcelofabri:
@Blackjacx feel free to submit a PR!

arnauddorgans:
Is there still no way to do it?

PR #3676, opened 2021-07-13
NameConfiguration excluded_pattern

arnauddorgans:
## What is the purpose of this PR?
The goal of this PR is to answer this issue https://github.com/realm/SwiftLint/issues/2156.
It allows us to exclude regex pattern for `generic_type_name`, `type_name` and `identifier_name` rules.

## How?
I added a new property to `NameConfiguration` named `excluded_pattern`

## Testing?
Ran the three test commands stated in CONTRIBUTING.md file and all of them have passed locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
 
 #### Enhancements
 
+* Add `excluded_pattern` property to `NameConfiguration`.  
+  [Arnaud Dorgans](https://github.com/arnauddorgans)
+  [#3676](https://github.com/realm/SwiftLint/pull/3676) 
+  
 * Make `test_case_accessibility` rule identify invalid test functions
   with parameters.  
   [Keith Smiley](https://github.com/keith)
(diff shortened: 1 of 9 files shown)


SimplyDanny:
First of all: Thank you for the contribution and sorry for the late feedback!

I like the idea of supporting patterns for excluded names. However, the `NameConfiguration` already contains a few options and I would like to avoid adding a new one. The existing list of `excluded` names would only become a subset of the new `excluded_pattern` option.

What do you think about just interpreting the strings in the `excluded` list as regular expression that must match a whole name? If there is a list element `"MyName"` it would only match the name `MyName` but not `MyName_v1`. That makes sure that existing configurations work as before. With the possibility to add regular expressions to the list, one could add `"^MyName_v\d+$"` to match the other (versioned) name, too.

What do you (@arnauddorgans, @jpsim) think about that?

SimplyDanny:
Closing in favor of #4655.

PR #3695, opened 2021-08-17
Add `scope_depth` rule to validate that scopes aren't nested more than `N` levels deep

dalemyers:
`N` defaults to 10 for warnings and 7 for errors, but is configurable.

Why do we want/need this rule? Swift relies heavily on closures and it's very easy to end up in a [pyramid of doom](https://en.wikipedia.org/wiki/Pyramid_of_doom_(programming)). By enforcing a max depth, developers are forced to re-evaluate and refactor their code so that it does not end up in this state. 

Some non-triggering examples:

```swift
struct F0 { }
```

```swift
struct F0 { struct F1 { struct F2 { Struct F3 { struct F4 { struct F5 { struct F6 { struct F7 {}}}}}}}}
```

A couple of triggering examples (these can be quite large):

```swift
struct F0 { struct F1 { struct F2 { Struct F3 { struct F4 { struct F5 { struct F6 { struct F7 { ↓struct F8 { ↓struct F9 { ↓struct F10 { ↓struct F11 {}}}}}}}}}}}}"
```
Triggers warning on F8, F9 and F10, and errors on F11.

```swift
class Zero {
  class One {
    class Two {
      func three() {
        if 4 == 4 {
          repeat {
            if 6 == 6 {
              callFunc() {
                ↓print("Eight")
                ↓if 8 == 8 ↓{
                  ↓print("Nine")
                  ↓if 9 == 9 ↓{
                    ↓if 10 == 10 ↓{
                      ↓{
                        print("Bad")
                      }
                    }
                  }
                }
              }
            }
          } while 5 == 5
        }
      }
    }
  }
}
``` 

Once the error level is reached, anything deeper is ignored to avoid excessive noise. 

I'm not entirely happy with the naming, so definitely open to changes there. 

This is my first real PR for SwiftLint, so there's probably a lot I've missed or misunderstood.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Metrics/ScopeDepthRule.swift
+++ b/Source/SwiftLintFramework/Rules/Metrics/ScopeDepthRule.swift
@@ -1,0 +1,209 @@
+import SourceKittenFramework
+
+public struct ScopeDepthRule: ConfigurationProviderRule {
+    public var configuration = ScopeDepthConfiguration(warningDepth: 7, errorDepth: 10)
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "scope_depth",
+        name: "Scope Depth",
+        description: "Code should not be more than 10 scope levels deep",
+        kind: .metrics,
+        nonTriggeringExamples: [
+            Example("struct F0 { }"),
+            Example("struct F0 { struct F1 {}}"),
+            Example("struct F0 { struct F1 { struct F2 { Struct F3 { struct F4 {}}}}}"),
+            Example("struct F0 { struct F1 { struct F2 { Struct F3 { struct F4 { struct F5 { struct F6 { struct F7 {}}}}}}}}"), // swiftlint:disable:this line_length
+            Example("""
+                class One {
+                  class Two {
+                    class Three {
+                      func whatever() {}
+                      func four() {
+                        if 5 == 5 {
+                          print("Irrelevant")
+                          repeat {
+                            if 7 == 7 {
+                              print("Safe")
+                            }
+                          } while 6 == 6
+                        } else {
+                          print("Safe")
+                        }
+                      }
+                    }
+                  }
+                }
+                """)
+        ],
+        triggeringExamples: [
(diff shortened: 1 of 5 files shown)


Source/SwiftLintFramework/Rules/Metrics/ScopeDepthRule.swift
```diff
@@ -0,0 +1,209 @@
...
+                case .forEach, .`if`, .repeatWhile, .`guard`, .`while`, .`for`:
+                    return false
+                }
+            }
+
+            return false
+        }()
+
+        for substructure in structure.substructure {
+            let newDepth = shouldIncreaseScope ? depth + 1 : depth
```
aaroncrespo, line 204: Can you return depth from the transform block of `SourceKittenDictionary`'s depth first search method?

I'm not sure if it supports early return. But would avoid implementing a similar traversal algorithm.

dalemyers: I know basically nothing about SourceKitten, can you explain a little more about what you mean here?

aaroncrespo: `SourceKittenDictionary` has     
`func traverseDepthFirst<T>(traverseBlock: (SourceKittenDictionary) -> [T]?) -> [T] {`

I only wonder if you traverse the root nodes in a file with `T` = `Depth`. and return if any `Depth` in `[Depth]` >= the configured limits. Only wondering if anyone is going get hung up on multiple DFS implementations against SourceKittens AST data structure.

aaroncrespo:
Maybe you can explain what prompted this rule? Others might be seeing it.

Source/SwiftLintFramework/Rules/Metrics/ScopeDepthRule.swift
dalemyers: I wasn't aware of that, but it would be a little more complex to use anyway. As for returning early, we already do that where we can (but can't do it in all cases because we need _all_ violations, not just the first). Once we reach the error depth, we don't go any deeper.

aaroncrespo:
Do you think instead of a new rule we can complies this with some options on https://realm.github.io/SwiftLint/nesting.html ?

PR #3700, opened 2021-08-23
Named rule correctly for Sourcery to work properly

KaneCheshire:
When I implemented the discouraged_none_name rule I incorrectly named it without the word "Rule" at the end of the type name. At the time, I couldn't figure out that was needed for it to be included in the PrimaryRulesList.swift generated file from Sourcery, so I manually added it and then forgot to actually fix it properly before it was merged.

This PR addresses that so that the name is correct and running `make sourcery` doesn't remove DiscouragedNoneNameRule.self from the generated list of rules.


KaneCheshire:
Is a changelog update required for this minor change? If so I'll update

marcelofabri:
Thanks!

PR #3713, opened 2021-09-10
Fixes MissingDocRuleConfiguration

bdfox325:


bdfox325:
@jpsim Sorry about that. Created a bug in my previous PR when trying to clean up the code.

jpsim:
Can you cover this with a test?

jpsim:
Can you also add tests exercising the actual rule with these configuration options?

Package.resolved
```diff
@@ -1,6 +1,15 @@
 {
   "object": {
     "pins": [
+      {
```
bdfox325, line 4: This was generated when running `make docker_test`

jpsim: Please revert, as CyptoSwift is only used on Linux.

jpsim:
Thanks for the fix!

PR #3721, opened 2021-09-17
Add add_target_in_variable_declaration_closure rule

marcelofabri:


marcelofabri:
I wonder if this rule should just catch `self` instead

PR #3373, opened 2020-10-07
'nslocalizedstring_key' now validates the comment argument in addition to the key argument

daltonclaybrook:
This PR resolves #3334

PR #3729, opened 2021-10-02
Update `nslocalizedstring_key` to validate all NSLocalizedString parameters

theoriginalbit:
Also rename the rule to `nslocalizedstring_requires_staticstring_arguments` to better match what it is now doing.

This furthers the work done in #3373 and validates all parameters when provided in the call. After this change it will validate the `key`, `tableName`, `value`, and `comment` arguments and ensure they are static strings.


jpsim:
Sorry for the delay in reviewing this, can you please rebase, address any remaining CI failures and then ping me again for review?

PR #3730, opened 2021-10-04
Exclude `id` from `identifier_name` by default

### LINKED ISSUE #3651: Rule Request: identifier_name customization allow 'id' when type is conforming to Identifiable protocol
### New Issue Checklist

### New rule request

Swift has introduced a very commonly used protocol named Identifiable. It is used all over the place in SwiftUI. Out of the box, this `id` should be allowed when type conforms to the protocol. As a user of the library, I don't want to have to have to configure the exclusion list for this rule to include `id` because that could potentially allow real issues to slip through.

Shouldn't trigger

```
struct TaproomAnnotation: Identifiable {
    let id: UUID = UUID()
    let coordinate: CLLocationCoordinate2D
}
```

Should trigger
```
struct TaproomAnnotation {
    let id: UUID = UUID()
    let coordinate: CLLocationCoordinate2D
}
```

Yes, this is a proposal to allow a configuration point for the existing identifier_name rule. There would be benefits to catch names that are `id` outside of Identifiable types that should still flag as a warning.

Enabled by default, it is part of the SwiftUI framework. Unlikely that Apple will break their protocol at this point and rename `id` to `identifier`.

agarmash:
Closes #3651

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@
   the method.  
   [Marcelo Fabri](https://github.com/marcelofabri)
 
+* Exclude `id` from `identifier_name` by default.  
+  [Artem Garmash](https://github.com/agarmash)
+  [#3651](https://github.com/realm/SwiftLint/issues/3651)
+
 #### Bug Fixes
 
 * Fixes a bug with the `missing_docs` rule where `excludes_inherited_types` would not be set.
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks for the PR! I think this makes sense as a default value considering how prevalent `id` is as a variable name, especially with the advent of the `Identifiable` protocol.

jpsim (approved):

PR #3732, opened 2021-10-05
Add optional prefer_self_in_static_references rule

SimplyDanny:
Add optional `prefer_self_in_static_references` rule to warn if the class/struct/enum name is used to reference static variables/functions in the class/struct/enum. The advice is to use `Self` instead which is not effected by renamings.

One simple example is

    class C {
        static let i = 0
        static let j = C.i
    }

wherein `C.i` is used to reference the static `i` in this class. The rule would suggest to write `Self.i` instead.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -48,6 +48,7 @@ opt_in_rules:
   - overridden_super_call
   - override_in_extension
   - pattern_matching_keywords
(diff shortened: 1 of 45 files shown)


jpsim:
Rule looks good to me, thanks for the PR! Did you check all the violations reported by the OSSCheck comment above to see if there were any false positives?

PR #3746, opened 2021-10-15
Speedup compiler invocations

PaulTaykalo:


jpsim:
How much of a speed up are you seeing with this? OSSCheck only benchmarks normal linting not analyzer rules.

Issue #3555, opened 2021-03-04
False-positive for `legacy_objc_type` rule


sindresorhus:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

The rule seems to match anything that **starts** with the [hard-coded names](https://github.com/realm/SwiftLint/commit/eca9ea4bb8bfeb4db640af409bf31101184c5c24#diff-4a54730b34dd348d489825a0ee3fa743d7c4d38271aed57a85c3ae27f79c2acbR5-R29), causing false-positives. It should only match on the full identifier name.

For example, this causes a warning because `NSData` is in the list:

```swift
try? NSDataDetector(types: 0)
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)?

0.43.0

* Installation method used (Homebrew, CocoaPods, building from source, etc)?

Installer

* Paste your configuration file:

```yml
only_rules:
  - legacy_objc_type
```

PH9:
I have a problem with 2 of `NS` classes
- `NSNumber` that needed to be used with `NSFormatter`.
- `NSDictionary` needed for `XCTAssertEqual` in unit testing.

Hope it can exclude or custom. Should I open another issue or leave it here?

sindresorhus:
> Should I open another issue or leave it here?

It should be another issue. It's unrelated to this issue.

PR #3752, opened 2021-11-02
Fix false positive in legacy_objc_type when a non-legacy Objective-C type name begins with a legacy Objective-C type name.

iressler:
This change addresses #3555 where the `legacy_objc_type` rule triggers on non legacy Objective-C types because their name contains a legacy Objective-C type name.

I updated the `legacy_objc_type` regex to require a word boundary before and after the legacy type name. This allows it to still detect usage of classes like `NSData` without also detecting `NSDataDetector`.

<s>This isn't a perfect solution because it can allow some usage like `NSStringFromClass` (found in [WordPress-iOS/PageListViewController.swift](https://github.com/wordpress-mobile/WordPress-iOS/blob/7de64fb8c63eb556c5817ad4d3baf933ba1cb384/WordPress/Classes/ViewRelated/Pages/PageListViewController.swift#L844:27)). I'm not sure how we could detect types like `NSDataDetector` without also detecting usage like `NSStringFromClass`, unless there is a hard coded list of either the disallowed methods/types (`NSStringFromClass`) or the allowed types (`NSDataDetector`), neither of which seems reasonable to create. If anyone has any better suggestions I'd love to hear them.</s>

Edit: Turns out `NSStringFromClass()` actually creates a `String` despite the legacy name, so I'm less concerned about that than I was before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,11 @@
   [Jesse Crocker](https://github.com/JesseCrocker)
   [Hannes Ljungberg](https://github.com/hannseman)
 
+* Fix false positive on `legacy_objc_type` rule when using
+  types with names that start with a legacy type name.  
+  [Isaac Ressler](https://github.com/iressler)
+  [#3555](https://github.com/realm/SwiftLint/issues/3555)
+
 #### Bug Fixes
 
 * Fix `unused_import` rule incorrectly considering `SwiftShims` as a
--- a/Source/SwiftLintFramework/Rules/Idiomatic/LegacyObjcTypeRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/LegacyObjcTypeRule.swift
@@ -41,7 +41,9 @@ public struct LegacyObjcTypeRule: OptInRule, ConfigurationProviderRule, Automati
         kind: .idiomatic,
         nonTriggeringExamples: [
             Example("var array = Array<Int>()\n"),
-            Example("var calendar: Calendar? = nil")
+            Example("var calendar: Calendar? = nil"),
+            Example("var formatter: NSDataDetector"),
+            Example("var className: String = NSStringFromClass(MyClass.self)")
         ],
         triggeringExamples: [
             Example("var array = NSArray()"),
--- a/Source/SwiftLintFramework/Rules/Idiomatic/LegacyObjcTypeRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/LegacyObjcTypeRule.swift
@@ -49,7 +49,7 @@ public struct LegacyObjcTypeRule: OptInRule, ConfigurationProviderRule, Automati
         ]
     )
 
-    private let pattern = legacyObjcTypes.joined(separator: "|")
+    private let pattern = "(?:\(legacyObjcTypes.joined(separator: "|")))[^A-Za-z]"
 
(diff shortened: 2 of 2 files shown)


iressler:
I spot checked some of the fixed violations and it looks like they fall into 2 categories:

1. Types that being with legacy types, but aren't themselves legacy.
   - An example is [Aerial/VideoDownload.swift](https://github.com/JohnCoates/Aerial/blob/824ab901dc9e404e82b8792fe77fec777e01a0b4/Aerial/Source/Models/Cache/VideoDownload.swift#L102:44) which uses `NSURLConnection`, which is not a legacy type but contains `NSURL` which is.
2. Types with legacy types in their names.
   - An example is [realm-cocoa/MutableSetTests.swift](https://github.com/realm/realm-cocoa/blob/b67fa0043b7eb957ab4eeb9ac72433fe3f4dac02/RealmSwift/Tests/MutableSetTests.swift#L343:27) which has `optNSStringCol` that used to be a violation and is not anymore. 

1 is what I was trying to fix with this change, 2 happened to fall out of it. So from the dozen or so I checked every looked expected.

I did notice that some of the lines where violations were fixed (e.g. the [realm-cocoa/MutableSetTests.swift](https://github.com/realm/realm-cocoa/blob/b67fa0043b7eb957ab4eeb9ac72433fe3f4dac02/RealmSwift/Tests/MutableSetTests.swift#L343:27)  example from 2) also contain an initializer for legacy types (e.g. `NSString(...)`). I verified that those are still detected with the new regex. In that case the `optNSStringCol` violation was "fixed", but presumably the `NSString` violation still exists.

iressler:
Looking at the "fixed" violations made me realize that the new regex will also match types that don't start with legacy type name, e.g. `MyNSURL` or `My_NSURL`. Thinking about this made me realize I'm really looking for word boundaries, which regex support really well with `\b`. Some quick tests make it look like that would work both before and after the legacy type name, so I'm going to update the PR with leading and trailing `\b`s and see if that changes the test results.

iressler:
After that change it looks like it added ~100 more "fixed" violations that fall into the same categories as before.

I did, however, find an interesting example of category 2 (from first post above) in [WordPress-iOS/PageListViewController.swift](https://github.com/wordpress-mobile/WordPress-iOS/blob/7de64fb8c63eb556c5817ad4d3baf933ba1cb384/WordPress/Classes/ViewRelated/Pages/PageListViewController.swift#L844:27) where it uses `NSStringFromClass()`, which isn't a legacy type but it is creating an `NSString` which is.

I'm going to update the PR description to include this and some extra thoughts on how it affects the PR.

jpsim:
Can you add tests by way of non-triggering and triggering examples please?

jpsim:
Thanks for adding tests so quickly. Can you please rebase too so that OSSCheck results are more up to date?

iressler:
Added a couple non-triggering examples for tests, I didn't add any new triggering tests because this only removes triggers. I could add some examples like `var className: NSString = NSStringFromClass(MyClass.self) as NSString`, however they seem overly complicated and covered by the existing triggering examples.

iressler:
I can also address the conflicts in the CHANGELOG if there is a desire to merge this in, however I'm not going to do that until someone is ready to merge this in (to avoid having to do it again later).

jpsim:
Merged. Thanks for your contributions @iressler ❤️

PR #3754, opened 2021-11-03
Fix false positives with async closures for large_tuple rule

### LINKED ISSUE #3753: large_tuple violation false positive when async function is a method parameter
### New Issue Checklist

### Describe the bug
When a function accepts an `async` function of 3+ parameters as a parameter, it is erroneously considered a violation of the `large_tuple` rule.

```bash
$ echo "private func foo(bar: (Int, String, Float) async throws -> Void) {}" | swiftlint lint --no-cache --use-stdin --enable-all-rules
```

Output:
```
<nopath>:1:18: warning: Large Tuple Violation: Tuples should have at most 2 members. (large_tuple)
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.45.0
* Installation method used (Homebrew, CocoaPods, building from source, etc)? Built from source.
* Paste your configuration file: No config file, this reproduces with the default configuration options.

* Which Xcode version are you using (check `xcodebuild -version`)? I'm on Linux, so no Xcode version.

```swift
// This triggers a violation:
private func foo(bar: (Int, String, Float) async throws -> Void) {}
```

kmahar:
Fixes #3753. Will leave some comments inline

### DIFF
--- a/Source/SwiftLintFramework/Rules/Metrics/LargeTupleRuleExamples.swift
+++ b/Source/SwiftLintFramework/Rules/Metrics/LargeTupleRuleExamples.swift
@@ -1,0 +1,74 @@
+struct LargeTupleRuleExamples {
+    static var nonTriggeringExamples: [Example] {
+        let commonExamples = [
+            Example("let foo: (Int, Int)\n"),
+            Example("let foo: (start: Int, end: Int)\n"),
+            Example("let foo: (Int, (Int, String))\n"),
+            Example("func foo() -> (Int, Int)\n"),
+            Example("func foo() -> (Int, Int) {}\n"),
+            Example("func foo(bar: String) -> (Int, Int)\n"),
+            Example("func foo(bar: String) -> (Int, Int) {}\n"),
+            Example("func foo() throws -> (Int, Int)\n"),
+            Example("func foo() throws -> (Int, Int) {}\n"),
+            Example("let foo: (Int, Int, Int) -> Void\n"),
+            Example("let foo: (Int, Int, Int) throws -> Void\n"),
+            Example("func foo(bar: (Int, String, Float) -> Void)\n"),
+            Example("func foo(bar: (Int, String, Float) throws -> Void)\n"),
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Rules/Metrics/LargeTupleRuleExamples.swift
```diff
@@ -0,0 +1,74 @@
+struct LargeTupleRuleExamples {
```
kmahar, line 1: pulling out the examples into a separate file like this seemed to a strategy used elsewhere in the codebase to add Swift version-specific examples, as seen in `ImplicitGetterRuleExamples.swift`, but let me know if there some better way to handle this.

Source/SwiftLintFramework/Rules/Metrics/LargeTupleRuleExamples.swift
```diff
@@ -0,0 +1,74 @@
...
+            Example("var completionHandler: ((_ data: Data?, _ resp: URLResponse?, _ e: NSError?) -> Void)!\n"),
+            Example("func getDictionaryAndInt() -> (Dictionary<Int, String>, Int)?\n"),
+            Example("func getGenericTypeAndInt() -> (Type<Int, String, Float>, Int)?\n")
+        ]
+
+        guard SwiftVersion.current >= .fiveDotFive else {
+            return commonExamples
+        }
+
+        return commonExamples + [
```
kmahar, line 26: For both the triggering and non-triggering examples, I just copied the existing ones that involved functions and added `async`. Not all of these were actually false positives -- the issue is only with async closures -- but given that both `throws` and non-`throws` versions were tested for all of the functions already it seemed like getting coverage of the `async` versions would be a good idea too.

kmahar:
(happy to add additional tests if desired per the bot comment, but the only existing ones I found for the rule for those auto generated from the triggering/non-triggering examples.)

jpsim:
Thanks for the PR @kmahar! The Danger bot doesn't consider SwiftLint's automatically-generated tests, but the examples you added are the right way to test this change.

I'll merge shortly.

PR #3755, opened 2021-11-10
Expand README Rules url

BasThomas:
From https://realm.github.io/SwiftLint/index.html, this would route to `https://realm.github.io/SwiftLint/Source/SwiftLintFramework/Rules`, which does not exist. Explicitly link to GitHub.


jpsim:
Thanks!

PR #3758, opened 2021-11-18
Documentation: Separate out default rules vs opt in rules in rule list

bdotdub:
I had a tough time figuring out which rules were on by default and which ones were opt in. This PR separates them out on the "Rule Directory".

Screenshots from locally generated docs:

image
image

### DIFF
--- a/Source/SwiftLintFramework/Documentation/RuleDocumentation.swift
+++ b/Source/SwiftLintFramework/Documentation/RuleDocumentation.swift
@@ -1,6 +1,6 @@
 /// User-facing documentation for a SwiftLint rule.
 struct RuleDocumentation {
-    private let ruleType: Rule.Type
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Documentation/RuleDocumentation.swift
```diff
@@ -1,6 +1,6 @@
 /// User-facing documentation for a SwiftLint rule.
 struct RuleDocumentation {
-    private let ruleType: Rule.Type
+    public let ruleType: Rule.Type
```
bdotdub, line 3: I don't love having to expose the `ruleType` but seems harmless to do so.

jpsim:
Thanks for the PR! Merging once CI passes.

PR #3771, opened 2021-12-01
Ignore more locations of static references not using 'Self' in classes

### LINKED ISSUE #3768: Compilation errors after following new rule prefer_self_in_static_references
### New Issue Checklist

### Describe the bug

After implementing the new rule in SwiftLint 0.45.1 `prefer_self_in_static_references` and running `swiftlint --fix` our code no longer compiles. This rule is not ignoring some cases where it should not be followed. Please see screenshots:

Skærmbillede 2021-11-30 kl  15 15 19
Skærmbillede 2021-11-30 kl  15 15 30

For now, we can just solve this with a `// swiftlint:disable:this` but the rule should be fixed to ignore such cases, and prevent build failures.

### Environment

* SwiftLint version: 0.45.1

```yml
...

opt_in_rules:
...
```

* Which Xcode version are you using (check `xcodebuild -version`)? Xcode 13.2 Beta 2

SimplyDanny:
This fixes #3768. Static references with `Self` are not allowed to define

* default values of properties,
* annotation parameter values and
* default method parameter values

in classes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@
   [JP Simard](https://github.com/jpsim)
   [#3770](https://github.com/realm/SwiftLint/issues/3770)
 
+* Fix false positives in the `prefer_self_in_static_references` rule.  
+  [SimplyDanny](https://github.com/simplydanny)
+  [#3768](https://github.com/realm/SwiftLint/issues/3768)
+
 * Fix the regex for expiring TODO comments.  
   [Sergei Shirokov](https://github.com/serges147)
   [#3767](https://github.com/realm/SwiftLint/issues/3767)
--- a/Source/SwiftLintFramework/Rules/Style/PreferSelfInStaticReferencesRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/PreferSelfInStaticReferencesRule.swift
@@ -10,7 +10,7 @@ public struct PreferSelfInStaticReferencesRule: SubstitutionCorrectableASTRule, 
         nonTriggeringExamples: [
             Example("""
                 class C {
-                    static let i = 0, j = C.i
+                    static private(set) var i = 0, j = C.i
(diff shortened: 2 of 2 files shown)


SimplyDanny:
The changes cause an exception for declarations involving `private(set)` since this is seen as an attribute, too. The fix will take some more time.

SimplyDanny:
> The changes cause an exception for declarations involving `private(set)` since this is seen as an attribute, too. The fix will take some more time.

This issue and the compilation failures reported in #3768 should be fixed with the latest commit.

The failed run on the Swift repository reported by @SwiftLintBot above is caused by the file `validation-test/compiler_crashers_fixed/25533-bool.swift` which contains the only line `extension(Int[a,Int`. It is clearly invalid Swift code, but SourceKitten interprets it as a declaration. How to deal with that? I guess it is not okay if SwiftLint aborts with a fatal error on invalid input?

A possible solution is to check whether the `name` of the declaration is a [valid Swift identifier](https://docs.swift.org/swift-book/ReferenceManual/LexicalStructure.html#grammar_identifier). Such a check is not straightforward. Is there something similar somewhere in the code already?

jpsim:
> I guess it is not okay if SwiftLint aborts with a fatal error on invalid input?

I've taken the approach with SwiftLint that it only really makes sense to operate on valid Swift code, so this is fine.

jpsim:
Thanks for these fixes!

jpsim:
@SimplyDanny I'll merge this as-is so I can cut a release before I log off for the weekend. Any additional fixes can come in a follow-up release.

SimplyDanny:
> @SimplyDanny I'll merge this as-is so I can cut a release before I log off for the weekend. Any additional fixes can come in a follow-up release.

Alright. I just resolved a merge conflict in the `CHANGELOG.md` with the update a few minutes ago.

Issue #3777, opened 2021-12-07
`warning: SwiftLint not installed, download from https://github.com/realm/SwiftLint` on m1 Mac even though SwiftLint is installed through Homebrew


agisilaos:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

Hello there. I installed SwiftLint through Homebrew on my new m1 machine. As I'm trying to run the Xcode Project I get a warning that `SwiftLint not installed, download from https://github.com/realm/SwiftLint`SwiftLint is indeed installed though. 

 I know that I can modify the script in Xcode to 

```bash
if test -d "/opt/homebrew/bin/"; then
  PATH="/opt/homebrew/bin/:${PATH}"
fi

export PATH

if which swiftlint >/dev/null; then
  swiftlint
else
  echo "warning: SwiftLint not installed, download from https://github.com/realm/SwiftLint"
fi
```

But I want to fix it locally without the need of changing the script from within Xcode. Couldn't find any particular useful info on the topic. Can anyone point me to the right direction? How I need to modify my `.zshrc` file?

Thank you all in advance.

agisilaos:
Planning to close this issue because after discussing it internally the way forward is changing the script. Planning to make a PR and updating the docs.

b593771943:
good question, 3Q for your script, expecting author can be attention on this problem, provide the final solution which just modify my .zshrc file or .bash_profile file

PR #3778, opened 2021-12-08
Update build phase snippet for Apple Silicon

soffes:
Since Homebrew installs to a path that isn't in Xcode's path by default, we need to add to the path to support the default location Homebrew installs SwiftLint on Apple Silicon.

From https://stackoverflow.com/a/66003612/118631

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,8 @@ target, and go to Build Phases. Click the + and select "New Run Script Phase".
 Insert the following as the script:
 
 ```bash
+export PATH="$PATH:/opt/homebrew/bin"
+
 if which swiftlint >/dev/null; then
   swiftlint
 else


agisilaos:
@soffes What's the difference between `export PATH="$PATH:/opt/homebrew/bin"` and  `export PATH="/opt/homebrew/bin/:${PATH}"`

soffes:
@agisilaos everything in the `$PATH` takes priority over what we're adding in the snippet. I've been using this method for a year without issue for what it's worth.

agisilaos:
Is thing going to impact folks who use non m1 machines? I'm asking since many folks in my team aren't having m1 machines just yet. @soffes 

I opened an issue about it last night. #3777

soffes:
No affect if you’re not on M1. Used with a team with some on intel for months without issue.

jpsim:
Thanks Sam.

PR #3786, opened 2021-12-15
add rule to check for @Environment usage

hborawski:
There is no compiler warning about `@Environment` usage in SwiftUI, though it only works as intended if used in a `View` or `ViewModifier`, other usage results in receiving the default value only.

This rule checks for usage of the `@Environment` property wrapper usage in types not conforming to `View` or `ViewModifier` to warn of improper usage.


hborawski:
This only works correctly if a type directly implements `View` or `ViewModifier`, a type implementing `UIViewRepresentable` which implements `VIew` comes up as a false negative. I'm having trouble finding the entire inheritance chain through sourcekitten, any suggestions would be helpful!

PR #3787, opened 2021-12-20
add support excluded names for function_body_length rule

VldmrVstsk:


jpsim:
Let me know when you'd like this reviewed @VldmrVstsk !

PR #3792, opened 2021-12-26
Add rule identifiers in 'Rule Directory' documentation page.

GetToSet:
This PR adds  rule identifiers in 'Rule Directory' documentation page.

Screenshot 2021-12-26 20 17 55

This can help users to make sense how these rules sort and quickly reference them when editing `.swiftlint.yml` files.

Since jazzy automatically creating links for identifiers wrapped in backquotes, `RuleDocumentation.ruleIdentifier` is no longer needed anymore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@
 
 #### Enhancements
 
-* None.
+* Add rule identifiers in "Rule Directory" documentation page.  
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks for the useful addition! Merged in `master` in bb37a253e.

PR #3794, opened 2021-12-26
Fix malformed doc for rule "InclusiveLanguage".

GetToSet:
Fix malformed doc for rule "InclusiveLanguage".


jpsim:
Thanks, pushed in 5fd8ef38c.

PR #3796, opened 2021-12-28
Enable caching for collecting rules

acecilia:
Hi 👋 

This PR enables caching for collecting rules.

#### Main idea
The collection step required by collecting rules is usually a very time consuming process. Under many scenarios, the result of such computation can be cached

#### Implementation details:
* Each collecting rule can opt in by conforming their `FileInfo` type to `CollectingCachable` 
* The easiest way to conform to `CollectingCachable` is making `FileInfo` conform to `Codable`. If that is undesired, then creating a middleware Dto is also possible

#### Discussion
As an example, I enabled caching for the two collecting rules in the codebase: `UnusedDeclarationRule` and `CaptureVariableRule`. Both rules query `SourceKitService` during collection. The main assumption that allows caching is that **if a swift file does not change => the values returned by `SourceKitService` do not change either**, and can be cached and reused between SwiftLint executions.

This is specially useful for the `UnusedDeclarationRule`, as the process of removing all unused declarations is iterative: removing flagged unused declarations will uncover new unused declarations. In a big codebase where `UnusedDeclarationRule` takes long time to execute, caching the collection information dramatically speeds up the code removal process


acecilia:
@jpsim Would appreciate some feedback, so I can understand if this is something you would accept

jpsim:
Thanks for the PR and the discussion! I think this is an important optimization to make analyzer rules significantly more useful. Were you able to measure a significant performance improvement on a real world codebase? I think it's important to be able to measure if this is having the expected impact you want.

jpsim:
Also if you rebase, hopefully OSSCheck will produce more reasonable results than what's currently posted above.

acecilia:
> Were you able to measure a significant performance improvement on a real world codebase? I think it's important to be able to measure if this is having the expected impact you want.

I did not start using this on a real world codebase yet. Will do that and come back with an answer

jpsim:
Thanks! I wonder if caching the result of `cursor-info` requests is the best solution here, since that's really the expensive part that all analyzer rules need.

acecilia:
> Thanks! I wonder if caching the result of cursor-info requests is the best solution here, since that's really the expensive part that all analyzer rules need.

Mmm 🤔Seems like if file did not change (and `compile arguments` did not change - _I will update the PR to take them into account_), caching only `cursor-info`  or caching all the collection information should be the same. In such case I think makes sense to cache all the collection information for better performance

PR #3797, opened 2021-12-28
Fix specifying the cachePath from command line

acecilia:
Hi 👋 

Seems like passing `--cache-path` to swiftlint from command line was not working, because the value was not taken into account when calculating the final `Configuration`. This PR fixes that, treating `cachePath` parameter in a similar way as `enableAllRules`

### DIFF
--- a/Tests/SwiftLintFrameworkTests/ConfigurationTests.swift
+++ b/Tests/SwiftLintFrameworkTests/ConfigurationTests.swift
@@ -13,6 +13,7 @@ class ConfigurationTests: XCTestCase {
 
     override func setUp() {
         super.setUp()
+        Configuration.resetCache()
         previousWorkingDir = FileManager.default.currentDirectoryPath
         FileManager.default.changeCurrentDirectoryPath(Mock.Dir.level0)
     }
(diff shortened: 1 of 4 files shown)


Tests/SwiftLintFrameworkTests/ConfigurationTests.swift
```diff
@@ -449,4 +450,45 @@ extension ConfigurationTests {
...
+        )
+
+        let configuration2 = Configuration(
+            configurationFiles: [],
+            cachePath: "cache/path/2"
+        )
+
+        XCTAssertEqual(configuration1.cachePath, "cache/path/1")
+        XCTAssertEqual(configuration2.cachePath, "cache/path/1")
+    }
```
acecilia, line 493: I think this test demonstrates another bug: when the `Configuration` is obtained from the in-memory cache, the `cachePath` is not taken into account.

Could anybody confirm this is the case?

jpsim: Yeah should be fine in practice. Thanks for checking though.

jpsim (approved):
Thanks for the fix, and especially covering with tests.

PR #3799, opened 2022-01-02
Prevent false triggering of `unused_closure_parameter` when using list bindings

### LINKED ISSUE #3790: unused_closure_parameter false positive with SwiftUI Binding
### New Issue Checklist

### Describe the bug

The following code should not trigger a warning. It is legal to use `$binding` on a closure argument to get its `wrappedValue` in `binding`:
image

This code is equivalent to manually writing:
image

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.45.1
* Installation method used (Homebrew, CocoaPods, building from source, etc)? Portable binary

```yml
excluded:
  - Tuist

line_length: 200
```

squarefrog:
Fixes #3790.

In Xcode 13 / iOS 15, SwiftUI now lets us create a `List` or `ForEach` directly from a binding. However, this triggers a false positive `unused_closure_parameter` when running SwiftLint. Consider the following examples:

```swift
List($users) { $user in
  Text(user.name)
}
```

```swift
List($users { $user in
  Toggle("Is favorite", isOn: $user.isFavorite)
}
```

Both are valid cases, but both are triggering warnings.

This PR marks the binding character (`$`) as optional in the paramater name regex.

### Warning
The changes involved removing the `ignoreMetacharacters` option of the capture list regex. While the tests all still pass, I'm unaware of any history why this value was previously present, so it may be worth further scrutiny.

### Resources

- https://www.hackingwithswift.com/quick-start/swiftui/how-to-create-a-list-or-a-foreach-from-a-binding
- https://www.swiftbysundell.com/articles/bindable-swiftui-list-elements/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@
   [JP Simard](https://github.com/jpsim)
   [#3761](https://github.com/realm/SwiftLint/issues/3761)
 
+* Fix false trigger from `unused_closure_parameter` when using
+  list element bindings in SwiftUI.  
+  [Paul Williamson](https://github.com/squarefrog)
+  [#3790](https://github.com/realm/SwiftLint/issues/3790)
+
 ## 0.45.1: Clothes Drying Hooks
 
 #### Breaking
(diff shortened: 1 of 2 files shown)


jpsim (approved):
Thanks for the fix!

freak4pc:
Thank you, @squarefrog !

nsillik:
Hi, this is still getting "fixed" when doing `swiftlint lint --fix`, leading to it no longer compiling as it will change it to:

```
List($users) { _ in
  Text(user.name)
}
```

Issue #3801, opened 2022-01-09
Rule Request: structured_function_doc requiring a short summary and a parameter list


arangato:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

Please describe the rule idea, format
this issue's title as `Rule Request: [Rule Name]` and describe:

1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.

There is a structured [way of documenting using markup](https://github.com/apple/swift/blob/main/docs/DocumentationComments.md) and many use it in their code. It would be beneficial to enforce it and also check that the documentation matches the declaration.

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

Shouldn't trigger

```
/// This is a classic foo.
///  - Parameters:
///    - a: a parameter.
///    - b: b parameter.
///    - c: c parameter.
func foo(a: Type1, b: Type2, c: Type3)

/// This is a concise foo summary.
///
/// Here comes detailed discussion spanning lots and lots of lines.
///  - Parameters:
///    - a: a parameter.
///    - b: b parameter.
///    - c: c parameter.
func foo(a: Type1, b: Type2, c: Type3)
```

Should trigger

```
/// Foo doc is one parameter short.
///  - Parameters:
///    - a: a parameter.
///    - c: c parameter.
func foo(a: Type1, b: Type2, c: Type3)

/// Foo parameters are not in order.
///  - Parameters:
///    - c: c parameter.
///    - b: b parameter.
///    - a: a parameter.
func foo(a: Type1, b: Type2, c: Type3)

/// Too long summary taking
///  lots
///  and lots
///  and lots
///  and lots
///  and lots of lines.
///  - Parameters:
///    - a: a parameter.
///    - b: b parameter.
///    - c: c parameter.
func foo(a: Type1, b: Type2, c: Type3)

```
3. Should the rule be configurable, if so what parameters should be configurable?
It should be configurable to enable user fine tuning it to a specific need and override any unnecessary inconveniences.   
    * One should be able defining how long the short summary can be, including not limiting it at all.
    * minimal number of parameters should be configurable as one might treat it as unnecessary burden for functions with too few (1 for instance) parameters

4. Should the rule be opt-in or enabled by default? Why?
   Should be an opt-in, as it's a matter of personal style preference.

PR #3802, opened 2022-01-09
Add structured_function_doc rule

arangato:
This PR  adds basic implementation of the rule requested in https://github.com/realm/SwiftLint/issues/3801.
It covers the `parameters section` method to document function parameters described in [Documentation Comment Syntax](https://github.com/apple/swift/blob/main/docs/DocumentationComments.md).


Package.swift
```diff
@@ -19,6 +19,7 @@ let package = Package(
         .package(url: "https://github.com/jpsim/SourceKitten.git", from: "0.31.1"),
         .package(url: "https://github.com/jpsim/Yams.git", from: "4.0.2"),
         .package(url: "https://github.com/scottrhoyt/SwiftyTextTable.git", from: "0.9.0"),
+        .package(url: "https://github.com/apple/swift-markdown.git", .branch("main")),
```
jpsim, line 22: We can't depend on branches here for published packages. Is there a version we can pin to?

jpsim:
Do you plan on addressing the CI failures?

arangato:
It seems that [swift-markdown](https://github.com/apple/swift-markdown.git) package is problematic for 2 reasons:
- The only release version there has unstable dependency
- It fails linux builds with an unresolved dependency: ``` swift-markdown: error: product dependency 'cmark-gfm' in package 'swift-cmark' not found ```

I intend to replace swift-markdown with a healthier alternative. Suggestions are welcomed, by the way. (it was initially chosen because [swift-docc](https://github.com/apple/swift-docc) uses it)

I have to admit that I'm a complete novice to SPM. If there is a solution to this dependency issues I'd be super glad if somebody pointed it out for me.

SwiftLint.podspec
```diff
@@ -9,7 +9,7 @@ Pod::Spec.new do |s|
   s.preserve_paths            = '*'
   s.exclude_files             = '**/file.zip'
   s.ios.deployment_target     = '9.0'
-  s.macos.deployment_target   = '10.10'
+  s.macos.deployment_target   = '10.11'
```
arangato, line 12: This should probably be done in a separate PR.
By the way, package.swift targets 10.12, so shouldn't we use 10.12 in podspec as well ?
@jpsim WDYT?

jpsim: This is a bit confusing, because the SwiftLint podspec just vends the SwiftLint library. SwiftLint doesn't run on iOS, watchOS or tvOS either, yet this podspec claims to support those platforms.

That's done so iOS/nacOS/tvOS/watchOS Xcode projects using CocoaPods can all add SwiftLint as a dependency, not to embed it into their apps but to run as an Xcode build phase.

So the deployment target here doesn't have to match what's in Package.swift and in fact should just go as far back as possible.

jpsim:
Can this support the inline parameter list style?

```swift
/// Summary
///
/// - parameter a: First.
/// - parameter b: Second.
```

SwiftLint.podspec
arangato: I see. Thanks for detailed explanation. 
Change to 10.11 stems from using [Down](https://github.com/johnxnguyen/Down). It's far enough, isn't it ?

arangato:
> Can this support the inline parameter list style?
> 
> ```swift
> /// Summary
> ///
> /// - parameter a: First.
> /// - parameter b: Second.
> ```

I'll give it a go.

SwiftLint.podspec
jpsim: Yes.

arangato:
I've added support for inline parameters. The rule looks for either parameters section or inlined ones. An arbitrary intermix of the two, which is allowed by [documentation syntax](https://github.com/apple/swift/blob/main/docs/DocumentationComments.md), the rule doesn't cover.
I bet such style intermixes are unpopular anyway.

arangato:
@jpsim could you please take a look ?

arangato:
@marcelofabri could you please have a look ?

PR #3806, opened 2022-01-15
Add accessibility_label_for_image rule

rcole34:
Add `accessibility_label_for_image` rule to warn if a SwiftUI Image does not have an accessibility label and is not hidden from accessibility. When this is the case, the image's accessibility label defaults to the name of the image file.

First time adding a rule here — tried to follow all the guidelines but let me know if I should make any changes or need to take a slightly different approach to anything!


Source/SwiftLintFramework/Rules/Lint/AccessibilityLabelForImageRule.swift
```diff
@@ -0,0 +1,103 @@
...
+            Example("""
+            Image(decorative: "image1")
+            ↓Image("image2")
+            Image(uiImage: "image3")
+                .accessibility(label: Text("a pretty picture"))
+            """)
+        ]
+    )
+
+    private var pattern: String {
```
jpsim, line 82: Can you avoid the use of a regular expression for this rule? Regular expressions are great for custom rules when you don't have the fully programmatic environment and full access to the AST, but with native lint rules you do have access to that. It will be more performant and less likely to have false positives.

rcole34: Sure can definitely take a look at that @jpsim ! Any existing rules that I should look at for examples of how I can use the AST to check things like an `Image` literal with/without certain modifiers? My current level of understanding with linters is more on the regex level and I'm not super comfortable or confident working with the AST.

While there are some false positives with this, the majority of the violations this caught in the repos that were scanned are true positives — places where VoiceOver would read the image file name. This is an increasingly common accessibility issue as more apps are adopting SwiftUI, so would love to get some check for it into swiftlint, even if just as a warning level opt-in regex rule for now!

rcole34: Totally agree on wanting to keep things performant. I'll take a look at those and see what I can do 👍

jpsim: I'm here to help if you get stuck.

rcole34: Hey @jpsim , I finally got around to revisiting this and taking a stab at writing it as an ASTRule, but managed to mess up the rebase pretty badly (I think I may have rebase origin/master from my fork instead of upstream/master, or potentially rebased origin/master from my branch rather than the other way around... 🤦‍♂️) so I went ahead and created a new branch, cherry-picked the commits, and opened up #3862 to replace this.

PR #3813, opened 2022-01-23
Ignore `prefer_self_in_static_references` rule in extensions generally

### LINKED ISSUE #3775: prefer_self_in_static_references error with static property initializers in extensions
### New Issue Checklist

(Might be related to this one: https://github.com/realm/SwiftLint/issues/3768 )

### Describe the bug

Using `prefer_self_in_static_references` auto-fixes  code into a non-compilable state.

```
extension DateFormatter {
    /// Represents weekday as full name (e.g. Sunday, Monday, Tuesday, etc.)
    static let weekdayName: DateFormatter = {
        let result = DateFormatter()
        result.dateFormat = DateFormatter.dateFormat(fromTemplate: "EEEE", options: 0, locale: nil)
        return result
    }()
}
```
gets turned into:
```
extension DateFormatter {
    /// Represents weekday as full name (e.g. Sunday, Monday, Tuesday, etc.)
    static let weekdayName: DateFormatter = {
        let result = DateFormatter()
        result.dateFormat = Self.dateFormat(fromTemplate: "EEEE", options: 0, locale: nil)
        return result
    }()
}
```
…which causes the compilation error:
Covariant 'Self' type cannot be referenced from a stored property initializer

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.45.1

```yml
opt_in_rules:
```

* Which Xcode version are you using (check `xcodebuild -version`)? GM

(see above)

SimplyDanny:
Fixes #3775. The issue is that SourceKitten cannot tell whether an extension extends a class, a struct or an enum. Or can it somehow? 🤔 

To make it safe, it's better to ignore the rule in extensions completely for the time being.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 #### Breaking
 
-* None.
+* Ignore `prefer_self_in_static_references` rule in extensions generally.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+  [#3775](ihttps://github.com/realm/SwiftLint/issues/3775)
 
 #### Experimental
 
--- a/Source/SwiftLintFramework/Rules/Style/PreferSelfInStaticReferencesRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/PreferSelfInStaticReferencesRule.swift
@@ -60,6 +60,15 @@ public struct PreferSelfInStaticReferencesRule: SubstitutionCorrectableASTRule, 
                         return i + Self.s
                     }
                 }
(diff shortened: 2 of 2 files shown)


jpsim:
> The issue is that SourceKitten cannot tell whether an extension extends a class, a struct or an enum. Or can it somehow?

No it can't because that information is only available in the post-typechecked AST, and SwiftLint lint rules operate on the pre-typechecked AST. Only SwiftLint analyzer rules have access to the post-typechecked AST.

PR #3814, opened 2022-01-23
Use pre-defined set of extension kinds

SimplyDanny:


jpsim:
Thanks for the cleanup.

PR #3829, opened 2022-01-25
Fix `class_delegate_protocol` false positives when using `where` clause

StevenMagdy:
Fixes a rare use case where a class protocol is written as `protocol FooDelegate where Self: AnyObject`


jpsim (approved):
Thanks for the fix!

PR #3833, opened 2022-01-29
Rename values for "Enabled by default" property

SimplyDanny:
I think "Yes" or "No" fit better as answers to "Enabled by default?".


jpsim:
Sure

PR #3834, opened 2022-01-29
Add option to exclude examples from the rule documentation

SimplyDanny:
The option can be used if an example has mainly been added as another test case, but is not suitable as a user example. User examples should be easy to understand. They should clearly show where and why a rule is applied and where not. Complex examples with rarely used language constructs or pathological use cases which are indeed important to test but not helpful for understanding can be hidden from the documentation with this option.

I want to use this option especially to clean up the examples in the `prefer_self_in_static_references` rule. Some of them are not very user-friendly.


jpsim (approved):
This is great, thanks.

PR #3837, opened 2022-02-02
fix typo

gsbernstein:
anonimous -> anonymous


gsbernstein:
How do I request more detail on the allowMultilineFunc option? [The documentation](https://realm.github.io/SwiftLint/opening_brace.html) doesn't have any info of additional non-triggering examples and it doesn't seem to work where I expect it would.

PR #3842, opened 2022-02-04
`class_delegate_protocol`: Add line break separator

Tobisaninfo:
Hi,
recently I got some linter warnings for the rule `class_delegate_protocol` on a popper protocol like this (code formatting is important):
```swift
public protocol ImagePickerDelegate: AnyObject
{
    func didSelect(image: UIImage?)
}
``` 

If the open curly bracket of the protocol section is on the same line, the rule will not report any warning.  
To fix this issue, I added the line break symbol to the list of separators to detect the inherited reference type. At the moment, the substring and split method results a symbol like `AnyObject\n`. By adding `\n`to the separator list, the symbol is detected correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@
 * Fix crash in the `closure_end_indentation` rule when linting with Swift 5.6.  
   [JP Simard](https://github.com/jpsim)
   [#3830](https://github.com/realm/SwiftLint/issues/3830)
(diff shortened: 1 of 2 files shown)


StevenMagdy:
@Tobisaninfo  Sorry I missed that in #3829. Thanks for the fix.

jpsim:
Thanks! Pushed to master in 811ca32c6d43511633bc6c85a2fdee2aefd145b4

PR #3857, opened 2022-02-15
Fix default rules section in documentation

natanrolnik:
While browsing the rules documentation, I noticed **many** rules were both in the Default Rules and in the Opt In Rules section.

I ran the generate-docs command and found the following:

Xcode-2022-02-15-15 35 21

The numbers confirm they don't add up.

After looking into it, the docs of the `drop(while predicate: (Element) throws -> Bool` function explains why:

> A closure that takes an element of the sequence as its argument and returns true if the element should be skipped or false if it should be included. **Once the predicate returns false it will not be called again.**

This caused the `defaultRuleDocumentations` array to contain almost all `ruleDocumentation`.

This PR fixes it:

Xcode-2022-02-15-15 33 18


jpsim (approved):
Nice catch, thanks.

PR #3858, opened 2022-02-16
Extend `xct_specific_matcher` rule to check for boolean asserts on (un)equal comparisons

### LINKED ISSUE #3726: XCTSpecificMatcherRule should catch `XCTAssert(foo == bar)`
### New Issue Checklist

### Describe the bug

It'd be great if `XCTSpecificMatcherRule` could also catch instances of `XCTAssert(foo == bar)` or `XCTAssertTrue(foo == bar)` (and the `!=` versions).

Maybe this should be a different rule (or another configuration) as I personally don't see that much advantage of using `XCTAssertNil` but `XCTAssert(foo == bar)` is definitely more problematic.

SimplyDanny:
Closes #3726.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@
 * Add `sorted_enum_cases` rule which warns when enum cases are not sorted.  
   [kimdv](https://github.com/kimdv)
 
+* Extend `xct_specific_matcher` rule to check for boolean asserts on (un)equal
+  comparisons.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+  [#3726](https://github.com/realm/SwiftLint/issues/3726)
+
 #### Bug Fixes
 
 * None.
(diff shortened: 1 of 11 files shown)


jpsim:
Is this not problematic because there's no guarantees that types that implement `==` also conform to `Equatable`. Tuples come to mind, where you can write `XCTAssert(foo == bar)` but not `XCTAssertEqual(foo, bar)` if `foo` and `bar` are of types say `(Int, Int)`.

SimplyDanny:
> Is this not problematic because there's no guarantees that types that implement `==` also conform to `Equatable`. Tuples come to mind, where you can write `XCTAssert(foo == bar)` but not `XCTAssertEqual(foo, bar)` if `foo` and `bar` are of types say `(Int, Int)`.

Yes, that's right. I had to disable it at some places in the code because of that reason. However, this rule has always caused false-positives in examples like `XCTAssertEqual(b, true)` where `b` is an optional. In that case, the proposed `XCTAssertTrue(b)` does not compile.

PR #3866, opened 2022-03-02
Fix syntactic sugar autocorrect rule

PaulTaykalo:


jpsim:
Looks good to me, thanks for the fix. Is there a reason this PR is marked as a draft?

The Buildkite CI agent appears to be down, I’m looking into that now. I’d like to see the OSSCheck output before merging this.

jpsim:
@PaulTaykalo OSSCheck found some regressions, can you please audit and fix them?

jpsim:
> OSSCheck found some regressions, can you please audit and fix them?

Thanks, there's fewer regressions now but still 4 remaining.

jpsim:
I'm wondering if this would be easier to write using SwiftSyntax 🤔

PaulTaykalo:
@jpsim 🤔

jpsim:
I recently re-added support for SwiftSyntax-based rules (e.g. https://github.com/realm/SwiftLint/pull/3867) and I'm wondering if it would be easier to write this rule if we used a Swift Syntax visitor rather than these challenging regular expressions.

PaulTaykalo:
Well, in the general case, this kind of task is unresolvable using regular expressions. so this should fix most of the cases. but I'll check whether it is possible to use SwiftSyntax here

PaulTaykalo:
Oh, this is slow :) 
image

PaulTaykalo:
@jpsim after few hours of debugging, decided, that it is probably, will be a better idea to use `SwiftSyntax`

jpsim:
Great progress on this. Let me know if you'd like a hand at any point.

jpsim:
Great progress on this. Let me know if you'd like a hand at any point.

PaulTaykalo:
@jpsim committing between air alerts :)

jpsim:
> committing between air alerts :)

omg please stay safe ❤️

PaulTaykalo:
> omg please stay safe ❤️
👍🏼 It's relatively safe here.

@jpsim have a question regarding triggering/non-triggering examples.

I added these as "non-triggering".  are these examples fine?
```
            Example("var currentIndex: Array<OnboardingPage>.Index?"),
            Example("func x(a: [Int], b: Int) -> Array<Int>.Index"),
            Example("unsafeBitCast(nonOptionalT, to: Optional<T>.self)"),
            Example("unsafeBitCast(someType, to: Swift.Array<T>.self)"),
            Example("let a = Swift.Optional<String?>.none"),

// These were marked as "triggering" in previous variant
            Example("let x = Array<String>.array(of: object)"),
            Example("let x = Swift.Array<String>.array(of: object)"),
```

jpsim:
> I added these as "non-triggering". are these examples fine?

Those look perfectly fine to me.

PaulTaykalo:
Ok then, will move on to corrections :)

jpsim:
Sorry for the trouble, but you may need to rebase to get https://github.com/realm/SwiftLint/pull/3893 for Danger/OSSCheck to run on your next commit.

PaulTaykalo:
@jpsim can you look at it?

In general there are two cases which changed:
- `Optional<T>.none` will not be triggered
- Complex expression with `.self` suffix won't be triggered, like <Optional<Array<Optional>>>.self

jpsim:
Great job here! Sorry I didn't review faster, but I'd like to suggest a few minor changes in https://github.com/realm/SwiftLint/pull/3915. Please review if you get a chance.

PR #3869, opened 2022-03-07
Require Swift 5.5 to build SwiftLint

jpsim:
* Disable Swift 5.4 CI jobs
* Remove SwiftLintFramework podspec
  I couldn't find a single use in GitHub's search: https://github.com/search?q=SwiftLintFramework+filename%3APodfile+language%3ARuby+language%3ARuby&type=Code
  Just not worth maintaining anymore.
* Change package swift tools version to 5.5
* Update docker_test to use Swift 5.5 image
* Clean up old Swift versions in SwiftVersionTests
  Even though these older Swift versions should still be supported, these tests
  can't be built against those versions so they're useless.
* Run docker commands with `--platform linux/amd64`
* Remove LinuxMain
  It looks like tests are discovered dynamically on Linux as of recent
  SwiftPM versions.


jpsim:
This is needed to integrate SwiftSyntax: https://github.com/realm/SwiftLint/pull/3867

PR #3873, opened 2022-03-08
Add -I to _InternalSwiftSyntaxParser on Linux

keith:
Since `_InternalSwiftSyntaxParser` isn't in the search paths when building with the `-static-stdlib` flag, we have to manually pass an include path so the swift compiler discovers it. Unfortunately we cannot just pass `-I/usr/lib/swift` because that results in duplicate library errors since the libraries we're linking statically also live there as shared libraries. This first symlinks the module definition to the root of the checkout and uses `.` for the `-I`. Any temporary path could be used instead.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,7 +3,7 @@ name: docker
 on:
   push:
     branches:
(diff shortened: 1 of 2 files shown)


jpsim (approved):
Thanks for fixing this, Keith. 🙏

PR #3877, opened 2022-03-08
Fix Dockerfile

jpsim:
Image size goes from 336MB to 457MB.


jpsim:
Closing in favor of https://github.com/realm/SwiftLint/pull/3873

PR #3879, opened 2022-03-08
Fix regression in `trailing_closure` rule with Swift 5.6

### LINKED ISSUE #3848: [Swift 5.6] Regression in `TrailingClosureRule`
`TrailingClosureRuleTests.testWithOnlySingleMutedParameterEnabled()` test is failing with Xcode 13.3 beta 2.

marcelofabri:
Fixes #3848


jpsim (approved):
Nice, I was actually looking at this one and wasn't able to come up with a solution, but this makes total sense.

PR #3880, opened 2022-03-08
Exclude examples with complex language constructs from the rule documentation

SimplyDanny:
Also restructure, merge, split and simplify the examples.


jpsim (approved):
Thanks!

PR #3881, opened 2022-03-08
Automatically enforce min Swift version for all rules

marcelofabri:
- Automatically enforce min Swift version for all rules. This is something we've talked about but for some reason never done it. This would allow us to make rules about new Swift features enabled by default
- Clean up examples now that we require Swift 5.5 to compile and run our tests

My final goal is to bump the min supported version of Swift versions in runtime to 5.0.0. This would allow us to clean up some legacy implementations and rely on newer features of SourceKit, like the presence of closures and tuples in the structure.

### DIFF
--- a/Source/SwiftLintFramework/Models/Linter.swift
+++ b/Source/SwiftLintFramework/Models/Linter.swift
@@ -58,6 +58,10 @@ private extension Rule {
             return nil
         }
 
+        if SwiftVersion.current < Self.description.minSwiftVersion {
+            return nil
+        }
+
         let ruleID = Self.description.identifier
 
         let violations: [StyleViolation]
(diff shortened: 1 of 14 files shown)


jpsim:
This is a breaking change for folks who may be using recent SwiftLint releases for linting projects that are using pre-5.0 Swift versions. We don't collect any usage analytics to know how much this is happening. I think it makes sense to me to make this breaking change, but we should call it out in the changelog.

marcelofabri:
> This is a breaking change for folks who may be using recent SwiftLint releases for linting projects that are using pre-5.0 Swift versions. We don't collect any usage analytics to know how much this is happening. I think it makes sense to me to make this breaking change, but we should call it out in the changelog.

This PR itself shouldn’t be a breaking change, but I agree that requiring Swift 5 in runtime would be, that’s why I’m planning to do it in another PR

jpsim:
> This PR shouldn’t be a breaking change, but I agree that requiring Swift 5 in runtime would be, that’s why I’m planning to do it in another PR

👍

PR #3882, opened 2022-03-09
Add back return_value_from_void_function rule

marcelofabri:


jpsim:
Nice. Doesn't need much to fix up.

Source/SwiftLintFramework/Protocols/SyntaxRule.swift
```diff
@@ -0,0 +1,39 @@
+import SwiftSyntax
+
+/// A rule that leverages the SwiftSyntax library.
+public protocol SyntaxRule: Rule {
```
jpsim, line 4: Can you sell me on this protocol? If you look at the force cast rule for example, it seems a lot cleaner: https://github.com/realm/SwiftLint/blob/3273cef42b17a350aa993f5fb311ec03901ef89a/Source/SwiftLintFramework/Rules/Idiomatic/ForceCastRule.swift#L28-L51

marcelofabri: I think the main advantages would be:

- having a single place to host getting the tree + walking it. This was more important before where we had to use a workaround to navigate the tree without crashing, but could still be valuable to make sure rules are consistent in how they handle errors
- having a starting point for shortcuts in correctable rules, like we have some helper protocols for ASTRule

However I don't love that this makes the Visitors public.

Honestly I could go either way - I don't think this brings us major advantages but also don't agree that not having it is cleaner since we'll have to repeat the tree transversing code all over the place

jpsim: I think we should start without increasing the public API of SwiftLintFramework, because then we need to call out changes as breaking changes.

marcelofabri: Sounds good!

PR #3884, opened 2022-03-09
[Docs] Remove "experimental" label for analyzer rules

jpsim:
These have been fairly stable for over two years at this point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -391,7 +391,7 @@ excluded: # paths to ignore during linting. Takes precedence over `included`.
   - Source/ExcludedFolder
   - Source/ExcludedFile.swift
   - Source/*/ExcludedFile.swift # Exclude files with a wildcard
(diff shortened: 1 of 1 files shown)


marcelofabri (approved):

SimplyDanny:
Now that the analysis rules are "official", does it make sense to run them together with all other rules when `swiftlint lint` is executed? Currently, we always have to run two commands to check all rules.

jpsim:
The use cases are totally different.

Lint rules can be run file-by-file without any knowledge of the structure between them. A collection of many rules can run on the order of seconds on large projects.

analyzer rules require a fully built index and a complete and accurate compilation database, and running a single analyzer rule on a large project can take on the order of hours.

SimplyDanny:
I see. Thank you for clarification.

PR #3886, opened 2022-03-09
Require Swift 5.0 in runtime

marcelofabri:


jpsim:
We should take advantage of the fact that 0.47.0 will have breaking changes to add this one in there too.

Source/SwiftLintFramework/Rules/Idiomatic/ToggleBoolRule.swift
```diff
@@ -11,7 +11,6 @@ public struct ToggleBoolRule: SubstitutionCorrectableRule, ConfigurationProvider
         name: "Toggle Bool",
         description: "Prefer `someBool.toggle()` over `someBool = !someBool`.",
         kind: .idiomatic,
-        minSwiftVersion: .fourDotTwo,
```
marcelofabri, line 14: maybe this should be enabled by default now?

Source/SwiftLintFramework/Rules/Lint/AnyObjectProtocolRule.swift
```diff
@@ -12,7 +12,6 @@ public struct AnyObjectProtocolRule: SubstitutionCorrectableASTRule, OptInRule,
         name: "AnyObject Protocol",
         description: "Prefer using `AnyObject` over `class` for class-only protocols.",
         kind: .lint,
-        minSwiftVersion: .fourDotOne,
```
marcelofabri, line 15: this is a warning in the compiler now - depending when it was added (need to check) we should remove this rule

marcelofabri: looks like it was added in 5.4.1 https://github.com/apple/swift/commit/e3373a9efd9c9f01757f1d9878ec9981722e669c, so let's keep it for now

Source/SwiftLintFramework/Rules/Idiomatic/ToggleBoolRule.swift
marcelofabri: decided not to since it relies on regexes pretty heavily and I wouldn't be too surprised if there are false positives out there

jpsim:
Is there a reason this PR is marked as a draft?

jpsim (approved):
Epic clean up. Thanks!

marcelofabri:
> Is there a reason this PR is marked as a draft? Or is it ready to review?

I just forgot to mark it as ready after adding the changelog entry. Thanks for reviewing it, I'll rebase it in a sec

Issue #3891, opened 2022-03-11
Recursive glob still doesn't work


pomozoff:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

`excluded:` doesn't work recursively

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ tree .
.
├── another
│   └── 2.swift
└── classes
    ├── 3.swift
    └── tests
        ├── R.generated.swift
        └── unit
            └── Mocks.generated.swift

4 directories, 4 files

$ cat .swiftlint.yml
excluded:
  - "*.generated.swift"
  - "**/*.generated.swift"
  - "**/*/*.generated.swift"

$ swiftlint --config .swiftlint.yml
Linting Swift files in current working directory
Linting 'Mocks.generated.swift' (1/3)
Linting '3.swift' (2/3)
Linting '2.swift' (3/3)
Done linting! Found 0 violations, 0 serious in 3 files.
```

### Environment

* SwiftLint version - 0.46.5
* Installation method used - Homebrew
* Paste your configuration file:

```yml
excluded:
  - "*.generated.swift"
  - "**/*.generated.swift"
  - "**/*/*.generated.swift"
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? **NO**
  If so, paste their relative paths and respective contents.
* Which Xcode version are you using (check `xcodebuild -version`)? - **Xcode 13.2.1 Build version 13C100**
* Do you have a sample that shows the issue? **Look above**

jpsim:
Please try again with a build from the master branch.

pomozoff:
```bash
/usr/local/bin/swiftlint --config .swiftlint.yml
Linting Swift files in current working directory
[1]    36842 illegal hardware instruction  /usr/local/bin/swiftlint --config .swiftlint.yml
```

jpsim:
You found a crash! I'd really appreciate if you could share more information about it.

You can follow these steps to build and run in Xcode, which should pause the debugger when the crash happens: https://github.com/realm/SwiftLint/blob/master/CONTRIBUTING.md#building-and-running-locally

jpsim:
Actually, I was able to reproduce it thanks to your detailed report.

[...]

PR #3892, opened 2022-03-11
Use alternate implementation of glob with globstar support

jpsim:
Adapted from https://gist.github.com/efirestone/ce01ae109e08772647eb061b3bb387c3

The implementation from Pathos seems buggy.

Fixes https://github.com/realm/SwiftLint/issues/3891


jpsim:
FYI @funzin

PR #3894, opened 2022-03-12
Rewrite IdenticalOperandsRule with SwiftSyntax

marcelofabri:


jpsim:
I know this is a work in progress, but flagging so it’s not missed, this looks like a regression:

> This PR fixed a violation in Realm: [/RealmSwift/Tests/AnyRealmValueTests.swift:942:24: warning: Identical Operands Violation: Comparing two identical operands is likely a mistake. (identical_operands)](https://github.com/realm/realm-cocoa/blob/4a882a41c06212eb9296a2ed780e04f199b14035/RealmSwift/Tests/AnyRealmValueTests.swift#L942:24)

marcelofabri:
> I know this is a work in progress, but flagging so it’s not missed, this looks like a regression:
> 
> > This PR fixed a violation in Realm: [/RealmSwift/Tests/AnyRealmValueTests.swift:942:24: warning: Identical Operands Violation: Comparing two identical operands is likely a mistake. (identical_operands)](https://github.com/realm/realm-cocoa/blob/4a882a41c06212eb9296a2ed780e04f199b14035/RealmSwift/Tests/AnyRealmValueTests.swift#L942:24)

Yeah, I now remember trying to reimplement this a while ago and facing the same problem: https://forums.swift.org/t/detecting-binary-operator-precedence/33350/2

jpsim:
If porting this rule introduces some regressions but is overall more robust, I'm not opposed to that tradeoff, but it'd be nice to avoid regressions. Let me know if you'd like a hand to look into porting the expression folding algorithm like Tony mentioned in the forums.

jpsim:
If we add the swift syntax folding code, maybe that should go in a new `SwiftSyntaxFolding` module?

jpsim:
The new violations flagged by OSSCheck should probably not be reported: https://github.com/realm/realm-cocoa/blob/95ebae968d1183decef9ac5990a7f097d02b603e/RealmSwift/Tests/ModernObjectAccessorTests.swift#L341:27

I don't think it's in the spirit of this rule to include function call expressions.

marcelofabri:
Moving this to draft as we'd want to use https://github.com/apple/swift-syntax/pull/619 instead

marcelofabri:
Updated it to use the new `SwiftOperators` package. Still need to address this:

> I don't think it's in the spirit of this rule to include function call expressions.

Package.resolved
```diff
@@ -24,8 +24,8 @@
         "repositoryURL": "https://github.com/apple/swift-argument-parser.git",
         "state": {
           "branch": null,
-          "revision": "9f39744e025c7d377987f30b03770805dcb0bcd1",
-          "version": "1.1.4"
+          "revision": "df9ee6676cd5b3bf5b330ec7568a5644f547201b",
+          "version": "1.1.3"
```
jpsim, line 28: Oh, Xcode you silly thing

Source/SwiftLintFramework/Rules/Lint/IdenticalOperandsRule.swift
```diff
@@ -55,152 +56,54 @@ public struct IdenticalOperandsRule: ConfigurationProviderRule, OptInRule {
...
-        guard let betweenTokens = subStringBetweenTokens(startToken, endToken) else { return false }
+        override func visitPost(_ node: InfixOperatorExprSyntax) {
+            guard let operatorNode = node.operatorOperand.as(BinaryOperatorExprSyntax.self),
+                  IdenticalOperandsRule.operators.contains(operatorNode.operatorToken.withoutTrivia().text) else {
+                return
+            }
 
-        let range = betweenTokens.fullNSRange
-        return regex("^\\s*\(regexString)\\s*$").matches(in: betweenTokens, options: [], range: range).isNotEmpty
+            if node.leftOperand.withoutTrivia().description == node.rightOperand.withoutTrivia().description {
```
jpsim, line 104: I guess this should check that the operands are simple identifiers?

jpsim:
> Updated it to use the new `SwiftOperators` package.

Nice!

Source/SwiftLintFramework/Rules/Lint/IdenticalOperandsRule.swift
jpsim: I can only think of contrived examples, but if you did `random() == random()` it would be odd for the linter to tell you both sides are the same.

Maybe you’re right though that the value provided from catching legitimate violations outweighs the annoyance of contrived false positives.

jpsim (approved):

Source/SwiftLintFramework/Rules/Lint/IdenticalOperandsRule.swift
SimplyDanny: What about examples like `f(i: 3) == f(i:  3)` or `1+1 == 1 +  1`?

jpsim: @SimplyDanny I think the first one is currently covered by the rule. Not sure about the second one.

SimplyDanny: The code highlight ignores multiple spaces in a row. That's why both sides in the first example looked the same. I've updated it to make the whitespace differences visible.

My point is that when only the string representation of both sides are compared, whitespace has an influence even though they are semantically equal.

Is it possible to compare the ASTs of both sides?

PR #3901, opened 2022-03-16
Update SwiftSyntax to 0.50600.0

jpsim:
Uses SwiftSyntax 5.5 on Linux when building with Swift 5.5. We use the 5.6 version of SwiftSyntax when building with Swift 5.5 and 5.6 on macOS because we statically link `lib_InternalSwiftSyntaxParser` thanks to https://github.com/keith/StaticInternalSwiftSyntaxParser/releases/tag/5.6.

This keeps SwiftLint binaries portable across machines on macOS, regardless of _where_ or even _if_ `lib_InternalSwiftSyntaxParser` is installed.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,7 +3,7 @@ name: docker
 on:
   push:
     branches:
-      - update-swiftsyntax-to-0.50600.0
+      - master
     tags:
       - '*'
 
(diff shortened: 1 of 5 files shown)


jpsim:
Reported the TSan error here: https://bugs.swift.org/browse/SR-16004

PR #3902, opened 2022-03-16
Test with Swift 5.6 Linux in CI

marcelofabri:


marcelofabri:
Interesting, most (all?) the failures seem to be related to SwiftLint commands tests (using `swiftlint:disable` & friends). Maybe a regression from https://github.com/realm/SwiftLint/pull/3872? 

Or do we need to update SwiftSyntax before being able to compile/run the tests in a new Swift version now?

jpsim:
See https://github.com/realm/SwiftLint/pull/3901

PR #3903, opened 2022-03-17
Update applesilicon x brew fix instruction

michaelhenry:
Although this snippet(`export PATH="$PATH:/opt/homebrew/bin"`) has no bad effect on an intel machine, some developers might find this weird, especially those who are using intel as this particular directory doesn’t exist at all or if they had installed `SwiftLint` via source code, or mint, and would think why there is brew involved. So aside from giving them a quick workaround, it would be better to explain to them the actual problem and provide them with an alternative.


jpsim (approved):
Thanks

PR #3905, opened 2022-03-18
Add `explicit_return` opt-in rule

m-chojnacki:
Add `explicit_return` opt-in rule that warns against omitting the `return` keyword inside closures, functions and getters. This rule provides behavior opposite to `implicit_return` and complements it. It's disabled by default and provides configurable return kinds and severity.


Source/SwiftLintFramework/Rules/Style/ExplicitReturnRule.swift
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    public func substitution(for violationRange: NSRange, in file: SwiftLintFile) -> (NSRange, String)? {
+        return (violationRange, "return ")
+    }
+
+    private func violationOffsets(file: SwiftLintFile) -> [ByteCount] {
+        guard let tree = file.syntaxTree else { return [] }
+
+        let visitors: [ViolationReporting] = [
```
marcelofabri, line 43: what do you think about combining these visitors into one? 

my understanding is that this should make the rule faster, as SwiftSyntax still calls all the possible `visitNode` methods even if you don't override a specific one, so combining the visitors should make this faster.

Source/SwiftLintFramework/Rules/Style/ExplicitReturnRule.swift
```diff
@@ -0,0 +1,142 @@
...
+            configuration.isKindIncluded($0.kind)
+        }
+
+        for visitor in visitors {
+            visitor.walk(tree)
+        }
+
+        return visitors
+            .reduce(into: [], { $0.append(contentsOf: $1.positions) })
+            .map { ByteCount($0.utf8Offset) }
```
marcelofabri, line 57: if you don't combine the visitor into one, you should add a `sorted()` here since we prefer that the violations are sorted by offset (I think this is encoded in our unit tests, you probably just don't have an example with 2+ violations)

Source/SwiftLintFramework/Rules/Style/ExplicitReturnRule.swift
```diff
@@ -0,0 +1,142 @@
...
+        triggeringExamples: ExplicitReturnRuleExamples.triggeringExamples,
+        corrections: ExplicitReturnRuleExamples.corrections
+    )
+
+    public func validate(file: SwiftLintFile) -> [StyleViolation] {
+        return violationOffsets(file: file).map {
+            StyleViolation(
+                ruleDescription: Self.description,
+                severity: configuration.severityConfiguration.severity,
+                location: Location(file: file, byteOffset: $0)
```
marcelofabri, line 25: it'd be good to customize the `reason` based on `returnKind` so messages are more accurate like "Prefer explicit returns in functions" for a function

Source/SwiftLintFramework/Rules/Style/ExplicitReturnRuleExamples.swift
```diff
@@ -0,0 +1,351 @@
+// swiftlint:disable:next type_body_length
+internal struct ExplicitReturnRuleExamples {
```
marcelofabri, line 2: can you add some examples using `Never` and tuples that are not empty? 🙏

Source/SwiftLintFramework/Rules/Style/ExplicitReturnRule.swift
```diff
@@ -0,0 +1,142 @@
...
+
+public struct ExplicitReturnRule: ConfigurationProviderRule, SubstitutionCorrectableRule, OptInRule {
+    public var configuration = ExplicitReturnConfiguration()
+
+    public init() {}
+
+    public static let description = RuleDescription(
+        identifier: "explicit_return",
+        name: "Explicit Return",
+        description: "Prefer explicit returns in closures, functions and getters.",
```
marcelofabri, line 13: do you want to handle subscripts here too?

marcelofabri:
@m-chojnacki this is an amazing first contribution, congratulations! this looks great, made some small suggestions - let me know what you think!

Source/SwiftLintFramework/Rules/Style/ExplicitReturnRule.swift
m-chojnacki: Due to merging visitors into a single one, I consider this one done.

Source/SwiftLintFramework/Rules/Style/ExplicitReturnRuleExamples.swift
m-chojnacki: Done, added some examples with `Never` and non-empty tuples.

Source/SwiftLintFramework/Rules/Style/ExplicitReturnRule.swift
m-chojnacki: I'd like to, but I wonder if I should to this in this PR. The reason is, the opposite, upstreamed rule: `implicit_return` only works for closures, functions and getters as well. Preferably, we should add subscripts support there as well, and in my opinion - I'd rather do this separately. Let me know what do you think. If you think it should be done in this PR, I can handle it!

Source/SwiftLintFramework/Rules/Style/ExplicitReturnRule.swift
m-chojnacki: That's a brilliant idea! I did this that way to be consistent with the opposite, upstreamed rule: `implicit_return`. I believe if we want to add such differentiation here, we should also do it for the `implicit` rule. I'd rather do this separately, although if you think it should be done in this PR I can still do this.

m-chojnacki:
@marcelofabri Thank you so much for the review! I uploaded according fixes to the code, and answered all the comments. Will be happy to make any further changes as well.

otaviocc:
@m-chojnacki, hi there! :-) Could you update the `CHANGELOG.md` file? I'll take a look at the PR and merge it.

m-chojnacki:
@otaviocc Hi! I rebased this PR on `master`, corrected the changelog as you suggested! :)

otaviocc:
Thanks @m-chojnacki for rebasing it and fixing the `CHANGELOG` :-)

I was going thru the code and there are some cases it might raise a violation when it shouldn't. For instance:

```swift
DispatchQueue.main.async { [weak self] in
    self?.updateUI()
}
```

or

```swift
func removeSubviews() {
    subviews.forEach { $0.removeFromSuperview() }
}
```

In both cases the closure return `Void`.

It might not be easy to get the information about the closure/function signature at all times, so having it as an optional rule isn't that of a big issue. I'm just concerned by the number of `// swiftlint:disable*` one might have to add to their codebase to prevent the violation in such cases (which are not that uncommon).

SwiftLint has a rule called `explicit_self` which throws a violation in cases where `self` isn't used in the context of a method or closure. This rule runs after the project compiles; it's a "analyzer" rule. Maybe that would be the way to cover all the cases for the `explicit_return`. 

What's your take on this, @marcelofabri? If the current PR's OK for you, I can merge it.

SimplyDanny:
I'm willing to move this forward. You too, @m-chojnacki? If so, let's start with rebasing the changes onto the current `main` branch.

m-chojnacki:
Hey @SimplyDanny, I am willing to move this forward and rebase it, I'm just worried how to handle cases mentioned in https://github.com/realm/SwiftLint/pull/3905#issuecomment-1199085814

SimplyDanny:
> Hey @SimplyDanny, I am willing to move this forward and rebase it, I'm just worried how to handle cases mentioned in [#3905 (comment)](https://github.com/realm/SwiftLint/pull/3905#issuecomment-1199085814)

Well, I agree this rule will cause a lot of false positives for closures. An Analyzer rule would be able to know which type a closure has and so reason whether a `return` is actually valid. However, Analyzer rules are very slow an require to compile a project in order to get type information. Therefore, I consider them as something additional. Your rule is a good starting point being sufficiently accurate and very fast. Its configuration allows to exclude closures completely if the rate of false positives would drive the rule otherwise too annoying.

That said, I suggest 3 things:
* Don't enable analysis of closures by default.
* If closures specify a return type explicitly, the rule may have enough information to reason about adding `return` or not.
* Perhaps later add an Analyzer rule that is more accurate.

The `implicit_return` rule has meanwhile received support for subscripts and initializers. So these may be types you would like to consider in this rule as well.

PR #3911, opened 2022-03-20
Add ability to run only one(focused) example

PaulTaykalo:


SimplyDanny:
I like the idea of focusing on specific examples. Commenting out all but the relevant examples finds an end with this.

I wonder, though, whether it can be made sure that code with focused examples is not checked in. There could be a custom SwiftLint rule which checks for `.focused()` in rule code. Or the build breaks somehow if the function is called while running on CI.

CONTRIBUTING.md
```diff
@@ -55,7 +55,6 @@ $ xcodebuild -scheme swiftlint test
 $ swift test
 $ make docker_test
 ```
-
```
SimplyDanny, line 58: Line intentionally removed?

jpsim (approved):
Thanks, this will be helpful!

jpsim:
Ah yes Danny's suggestion would be good to consider. I'd be ok if we had a custom rule preventing this from being checked in.

PaulTaykalo:
@jpsim is there any additional env that is set on CI?

PaulTaykalo:
@jpsim how about allowing this functionality under `#if DEBUG` only ?

PaulTaykalo:
@jpsim @SimplyDanny 
added DISABLE_FOCUSED_EXAMPLES conditional, which will be enabled only on CI
If this conditional enabled, focused examples will fail
image

SimplyDanny (approved):

CONTRIBUTING.md
```diff
@@ -72,6 +72,19 @@ over time. This way adding a unit test for your new Rule is just a matter of
 adding a test case in `RulesTests.swift` which simply calls
 `verifyRule(YourNewRule.description)`.
 
+For debugging purposes examples can be marked as `focused`. If there are any
+focused examples found, then only those will be run when running tests.
```
jpsim, line 76:
```suggestion
focused examples found, then only those will be run when running tests for that rule.
```

CHANGELOG.md
```diff
@@ -48,6 +48,10 @@
   environments, such as in Swift Package Manager plugins.  
   [Juozas Valancius](https://github.com/juozasvalancius)
 
+* Add ability to run only one (focused) example
```
jpsim, line 51: Nit: two spaces at the end to force a newline, as per CONTRIBUTING.md

```suggestion
* Add ability to run only one (focused) example.  
```

jpsim (approved):

PR #3912, opened 2022-03-20
Accommodations for SPM Plugins

juozasvalancius:
With the release of Swift 5.6, it is now possible to use compiled build tools in Swift Package Manager.

These tools should be put in a specific structure, called artifactbundle, it is described in [SE-0305](https://github.com/apple/swift-evolution/blob/main/proposals/0305-swiftpm-binary-target-improvements.md).

So this PR adds the construction of this artifactbundle, which should be uploaded as a release asset together with other `zip` and `pkg` files.

----

Additionally, this PR adds a command line parameter to use the in-process version of sourcekit (https://github.com/jpsim/SourceKitten/pull/728). This is to work around the Xcode bug, which doesn't pass the specified environment variable in a `.buildCommand` in an SPM plugin.

----

With these changes, it is then possible to use swiftlint in a Swift package, and Xcode can display the lint warnings and errors, without the need to use cocoapods, or install swiftlint globally on the system. See this example project: https://github.com/juozasvalancius/ExampleSPMProjectWithSwiftLint

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,15 @@
   environments, such as in Swift Package Manager plugins.  
   [Juozas Valancius](https://github.com/juozasvalancius)
 
+* The above can also be achieved using `--in-process-sourcekit` command line
+  option.  
+  [Juozas Valancius](https://github.com/juozasvalancius)
+
+* Added macOS `artifactbundle` as a release asset that contains the built
+  `swiftlint` tool.  
+  [Juozas Valancius](https://github.com/juozasvalancius)
+  [#3840](https://github.com/realm/SwiftLint/issues/3840)
+
 #### Bug Fixes
 
 * Extend `class_delegate_protocol` to correctly identify cases with the protocol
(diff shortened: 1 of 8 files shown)


juozasvalancius:
@jpsim what do you think about this?

jpsim:
Thanks @juozasvalancius, will merge shortly after applying some minor edits.

jpsim (approved):
🙌

jpsim:
Pushed in 62cc0f1a & a28516d5. Thanks for the contributions.

juozasvalancius:
@jpsim the version in https://github.com/realm/SwiftLint/commit/62cc0f1a32e3bb2c52d5c9e1a123d1fa8d57f364 causes the command line option to override the environment variable, effectively making the environment variable non functional. That is, if you **don't** specify `--in-process-sourcekit`, but **do** specify `IN_PROCESS_SOURCEKIT=YES`, the in-process version will **not** be used, because the `false` from the command line options will overwrite the value in `SourceKittenConfiguration`.

It would probably be better to only override it if the command line option is `true`.

juozasvalancius:
Thanks for all the fixes and improvements @jpsim !

jpsim:
Yeah, sorry for the two (!!) regressions I introduced but hopefully all is resolved now.

Issue #3156, opened 2020-03-28
Critical error in autocorrect for `syntactic_sugar`


tsheaff:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

`swiftlint autocorrect` fails on the following code, in a way that fails to compile

```swift
func fooBarBaz() -> Dictionary<String, Dictionary<Int, Float>> {
    ...
}
```

It becomes:

```swift
func fooBarBaz() -> [String: Dictionary[Int: Float]> {
    ...
}
```

This does not compile.

It should become:

```swift
func fooBarBaz() -> [String: [Int: Float]] {
    ...
}
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)?
`0.39.1`

* Installation method used (Homebrew, CocoaPods, building from source, etc)?
CocoaPods

* Paste your configuration file:

```yml
disabled_rules: # rule identifiers to exclude from running
  - shorthand_operator
  - trailing_whitespace
  - line_length
  - cyclomatic_complexity
  - identifier_name
  - function_body_length
  - notification_center_detachment
  - multiple_closures_with_trailing_closure
  - file_length
  - type_body_length
  - switch_case_alignment
  - large_tuple
  - function_parameter_count
  - discouraged_direct_init
  - discarded_notification_center_observer
  - todo
  - no_fallthrough_only
  - nesting
  - operator_whitespace
  - type_name
  
opt_in_rules: # some rules are only opt-in
  - empty_count
  - empty_string
  - sorted_first_last
  - contains_over_first_not_nil
  - first_where
  - last_where
  - yoda_condition
  - prefer_self_type_over_type_of_self

included: # paths to include during linting. `--path` is ignored if present.
  - app/source
  - tests
  - watch/extension/source
  - widget

excluded: # paths to ignore during linting. Takes precedence over `included`.
  - Pods
  - fabric
  - fastlane
  - scripts
  - siri 

reporter: "xcode"
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)?
  If so, paste their relative paths and respective contents.

No

* Which Xcode version are you using (check `xcodebuild -version`)?

```
Xcode 11.0
Build version 11A419c
```

wjling:
Same problem. It's critical X(

mluisbrown:
The same problem occurs for optional closures:

```swift
[...]

PR #3915, opened 2022-03-23
Apply minor changes to the syntactic sugar rewrite

jpsim:
I was late in reviewing https://github.com/realm/SwiftLint/issues/3156, so I reviewed after it merged and am suggesting some minor changes here.

- Improve docstrings for `StringView+SwiftSyntax.swift`
- Move changelog entry to correct section & reword
- Change parameter type from `Int` to `ByteCount`
- Move AbsolutePosition / ByteCount conversion to internal API
- Only warn once if syntax tree cannot be parsed
- Move Syntactic Sugar examples to a dedicated file
- Change SyntacticSugarRuleVisitor from SyntaxAnyVisitor to SyntaxVisitor
- Add `SugaredType` enum to help with the implement `SyntacticSugarRule`


jpsim:
@PaulTaykalo looks like my changes unintentionally fixed some violations, but I think that's actually better, since the ones that were previously being flagged as violations were initializers like `Array<String>(repeating: ..., count: ...)` which shouldn't be violations in my opinion since `[String](repeating: ..., count: ...)` isn't in the spirit of the rule. Same for Optional: `Optional<SwiftStringObject>(SwiftStringObject(value: ["bye"])))` vs `SwiftStringObject?(SwiftStringObject(value: ["bye"])))` which I'm not even sure is legal Swift.

PaulTaykalo (approved):

PaulTaykalo:
In this case, I suggest adding these examples in non-triggering rules. So those will be documented

jpsim:
I still think this rule shouldn't apply to initializer cases, but doing that consistently will be more work, so I'll land these internal changes first.

I believe I've reverted the change and if OSSCheck confirms, I'll merge.

Issue #3628, opened 2021-05-05
False positive in `unused_closure_parameter` when using backticks


sberrevoets:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

SwiftLint reports a false positive for `unused_closure_parameter` when the parameter in question is wrapped in backticks. https://github.com/realm/SwiftLint/issues/2006 is a similar issue (and potentially a manifestation of the same bug) but was closed.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint lint --path Sample.swift --no-cache --enable-all-rules
Linting Swift files at paths Sample.swift
Linting 'Sample.swift' (1/1)
/Users/scott/Developer/Sample.swift:4:31: warning: Unused Closure Parameter Violation: Unused parameter "class" in a closure should be replaced with _. (unused_closure_parameter)
Done linting! Found 1 violation, 0 serious in 1 file.
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? **0.43.1**
* Installation method used (Homebrew, CocoaPods, building from source, etc)? **Homebrew**
* Paste your configuration file: **none**
* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? **No**
* Which Xcode version are you using (check `xcodebuild -version`)? **Xcode 12.4 12D4e**
* Do you have a sample that shows the issue? 

```swift
internal struct Sample {
    internal func test() -> [String] {
        let classes: [String] = ["a", "b", "c"]
        return array.filter { `class` in `class`.hasPrefix("a") }
    }
}
```

lhunath:
maybe we could add the `bug` label?

jpsim:
Fix: https://github.com/realm/SwiftLint/pull/3916

PR #3916, opened 2022-03-23
Fix false positives in `unused_closure_parameter`

jpsim:
When using parameters with backticks.

Fixes https://github.com/realm/SwiftLint/issues/3628


sberrevoets (approved):
Awesome 🙏🏻

PR #3917, opened 2022-03-23
Move StringView.byteOffset into generally accessible extension

SimplyDanny:
I will provide some unit tests tomorrow ...


SimplyDanny:
> You can if you want, but personally I wouldn't bother since this is indirectly tested via rule example verification.

I actually like to have some simple tests even for supposedly obvious basic functionality. So rules can just build on them without worrying about potential issues in their implementation. Also, tests sometimes serve as good examples for understanding.

But yeah, maybe it's too much and just a waste of time. 🤷‍♂️

jpsim (approved):
Nice clean up.

PR #3922, opened 2022-03-28
Simplify dictionary traversal code

jpsim:
By using flatMap.

I did not notice a performance impact when running analyzer rules, but I'm curious to see if OSSCheck reports any impact when linting.


jpsim:
OSSCheck reports 0-7% worse, so let's just keep as is.

PR #3929, opened 2022-04-01
Run custom rules in parallel

marcelofabri:


marcelofabri:
This didn't result in the improvement I was hoping for (even testing locally on a project with multiple custom rules)

jpsim:
There are generally two levels of parallelism when running `swiftlint lint`:

- files are processed in parallel
- rules are processed in parallel

So even if all custom rules are run serially, you still get concurrency from running them on all files concurrently.

To see an impact from running custom rules in parallel, you'd probably need to have a situation where SwiftLint is linting a single very large file with a lot of expensive custom rules and no other "native" SwiftLint rules. Not a common scenario and probably not worth the overhead incurred in more common situations.

marcelofabri:
Yeah, I imagined this could be the case given that we're already parallelizing so many things (and probably it didn't help that I was testing in a 6-core MacBook), but thought it was worth a shot.

PR #3931, opened 2022-04-03
Introduce a model for rule configuration description

SimplyDanny:
The idea is to not just build a description of a rule configuration by crafting a string representation manually but applying a defined set of elements to build a model of the description.

These elements are put together by a new result builder. The model enables a common but different formatting of the console output as well as the representation on the rule's web page. Some configurations currently look quite strange either on the website or the console, especially the more complex ones with regard to their configuration:

* [required_enum_case](https://realm.github.io/SwiftLint/required_enum_case.html)
* [type_name](https://realm.github.io/SwiftLint/type_name.html)
* [type_contents_order](https://realm.github.io/SwiftLint/type_contents_order.html)

The second one now looks like:

Screenshot 2022-04-04 at 22 09 42

Predefined types make sure that it is clear whether to put the values in quotes or brackets. Nested configurations are also considered which makes it easier to get the YAML structure right.

Even though the PR is quite huge, most of the changes deal with the conversion from the string representation to the new model syntax.

Edit: For an even easier setup of the model, simple configurations can use the `@ConfigurationElement` annotation to mark properties as relevant for documentation.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -56,6 +56,7 @@ disabled_rules:
 
 attributes:
   always_on_line_above:
+    - "@ConfigurationElement"
     - "@OptionGroup"
     - "@RuleConfigurationDescriptionBuilder"
 identifier_name:
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -57,6 +57,7 @@ disabled_rules:
 attributes:
   always_on_line_above:
     - "@OptionGroup"
+    - "@RuleConfigurationDescriptionBuilder"
 identifier_name:
   excluded:
     - id
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#5072](https://github.com/realm/SwiftLint/issues/5072)
 
+* Prettify the rule configuration presentation on the command line as well as
+  on the website.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+
 #### Bug Fixes
 
 * Fix false positives for the `unneeded_synthesized_initializer` rule, when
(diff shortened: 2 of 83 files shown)


marcelofabri:
I haven't fully reviewed this, but I love this direction! I wonder if there's something we can do to add that documentation closer to the variable it represents in the configuration with a custom property wrapper, like how ArgumentParser works.

SimplyDanny:
This is a great idea! I'd love to try this.

.swiftlint.yml
```diff
@@ -6,6 +6,8 @@ excluded:
 analyzer_rules:
   - unused_declaration
   - unused_import
+disabled_rules:
+  - redundant_optional_initialization
```
SimplyDanny, line 10: Temporarily disabled this rule. See #3998.

SimplyDanny:
> I wonder if there's something we can do to add that documentation closer to the variable it represents in the configuration with a custom property wrapper, like how ArgumentParser works.

I followed your advice using property wrappers now which looks even cleaner. 👍 However, the result builder variant is still required for a few configurations that are too complex to be expressed by the property wrapper approach.

.swiftlint.yml
SimplyDanny: Adding another constructor to the property wrapper allowed to avoid the explicit initialization with `nil`. Thus, the rule is now active again.

SimplyDanny:
The reported resolved/new issues (`untyped_error_in_catch`) will be resolved by #4001.

SimplyDanny:
I would highly appreciate if you could take a look on this again, @marcelofabri. Other changes are causing regular merge conflicts in this huge PR.

There is room for more improvements, of course. Some ideas are already on my todo list. I would like to use this as a basis, however, and not clutter it with more and more changes.

SimplyDanny:
Would you mind taking a look into this PR, @jpsim? I would like to continue with some more improvements to the documentation of rule configurations.

If we agree on the approach I will rebase the PR and fix the conflicts.

Issue #3672, opened 2021-07-09
Rule Request: ACL ordering within type contents


LucasVanDongen:
### New Issue Checklist

- [X] Updated SwiftLint to the latest version
- [X] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

The rule is meant to have the most visible (public) fields or functions on the top of your class, struct or enum. So for example `public var` goes above `internal var` which in turn goes above `private var`. However `public init` would come below all of the `var` declarations, including the `private var`s. This could be an extension of [type_contents_order](https://realm.github.io/SwiftLint/type_contents_order.html) where the contents are further sorted on access control, for example by adding an optional parameter to it.

This rule should be added to enforce consistency in an arbitrary way with as added bonus that the stuff you're most likely to interact with as consumer of the object / struct / enum is also the first you find when scrolling through the file. I have no strong opinion on this rule being mandatory or not, I think everybody's code would improve with better sorting of contents but [type_contents_order](https://realm.github.io/SwiftLint/type_contents_order.html) is optional as well so it would be strange (and even impossible without type_contents_order?) to enforce this rule by default.

If you think this is a good idea, I can try to implement the rule myself.

One thing that can make it slightly difficult to implement is the fact that often fields don't have an explicit acces control modifier.

### Would not trigger

```swift
class Car {
    public var horn = "Honk"
    internal var: owner: String
    private var tireType = "Whitewall"

    init(owner: String) {
        self.owner = owner
    }
    
    func replaceOwner(with newOwner: String) {
        owner = newOwner
        honk()
    }

    private func honk() {
        print(horn)
    }
}
```

### Will trigger

```swift
class Car {
    internal var: owner: String // Public fields should precede internal fields
    private var tireType = "Whitewall" // Public fields should precede private fields
    public var horn = "Honk" // Public fields should precede private fields

    private init(owner: String, tireType: String) { // Public initializers should precede private initializers
        self.owner = owner
        self. tireType = tireType
    }

    init(owner: String) { // Public initializers should precede private initializers
        self.owner = owner
    }

    static func owned(by owner: String, having tireType: String) -> Car {
        return Car(owner: owner, tireType: String)
    }

    private func honk() { // Public functions should precede private functions
        print(horn)
    }
    
    func replaceOwner(with newOwner: String) { // Public functions should precede private functions
        owner = newOwner
        honk()
    }
}
```

PR #3932, opened 2022-04-04
Add Type ACL Order Rule

### LINKED ISSUE #2410: [OrderOfProperties]
### New Issue Checklist

- [ ] Updated SwiftLint to the latest version

### New rule request

I need rule to control properties groups order to keep file neat
All properties should be divided on several groups in order:

```
public let
public var

internal let
internal var

private let
private var

IBOutlet private var
```
The rule should triggering on mixing up this groups and on incorrect order
The rule should be configurable, I need to set desired `order`, or maybe order written above will bw enough
The rule should be opt-in

jd014:
Overview:
This PR adds a new rule to enforce access control level (ACL) ordering within types. This rule expands upon the type contents order rule by using the same logic to separate a type into components. It then validates the ACL ordering of all instances for each component type. Closes #2410 and #3672.

Discussion:
1. Separate rule instead of an added configuration option on `type_contents_order`
A standalone rule allows for more configuration capabilities and no rule coupling. You could use this rule without even implementing `type_contents_order` and it would function properly, which seems to be a good enough argument to have a separate rule.
2. Configuration Options
I chose to keep it simple and just add an ACL order configuration option for now. If there is a need for further configuration such as content types to include/exclude/group then it can be extended to include this. 
3. Disabled by Default
This style rule is very opinionated and would trigger an obscene amount of warnings for most projects.
4. Autocorrect
Autocorrect is not implemented in this PR. I'm not confident there is a good way to automate this without impacting other stylistic elements like spacing and type ordering.
5. Rule Naming
`type_acl_order` is succinct and very meaningful if you know what ACL means, however I wouldn't qualify ACL as a common knowledge acronym for swift developers. However, other rules use ACL and I think consistency is important. I would also be open to including contents in the name such as `type_contents_acl_order`, but I'm not sure how much value it adds.


jd014:
Hey @jpsim, @SimplyDanny, @marcelofabri. Finally got around to updating this PR. Could I get some feedback when you get a chance? Thanks.

sethdeckard:
@jd014 Any plans to resolve the conflicts on this?

jd014:
@sethdeckard if any maintainers of this project have an interest in reviewing and potentially merging this PR, I can put aside some time to fix the conflicts. I had this PR up to date for awhile, but the core team was not prioritizing rule requests. I am not sure if that has changed.

mildm8nnered:
> @sethdeckard if any maintainers of this project have an interest in reviewing and potentially merging this PR, I can put aside some time to fix the conflicts. I had this PR up to date for awhile, but the core team was not prioritizing rule requests. I am not sure if that has changed.

I've had PRs merged fairly regularly (https://github.com/realm/SwiftLint/pulls?q=is%3Amerged+is%3Apr+author%3Amildm8nnered+) including new rules ...

PR #3936, opened 2022-04-05
[CI] Move TSan unit tests to Buildkite

jpsim:
This should even out the load a bit more and make CI complete faster overall.


jpsim:
We hit a stack overflow in SwiftSyntax with Xcode 13.3.

PR #3942, opened 2022-04-10
Enable `typesafe_array_init` rule in repository

SimplyDanny:


jpsim:
Is there a reason we'd run this considering we already run `array_init` and `typesafe_array_init` returns a strict subset of violations from the former?

jpsim:
I ask because running analyzer rules on CI takes a very long time, so adding another rule makes that even longer.

SimplyDanny:
I though we would enable as many rules as possible on the repository itself to have some kind of real-life tests, like "eat your own dog food".

But you are right, this rule cannot trigger as long as `array_init` is also enabled and only takes time. 🤦‍♂️

So I understand you would rather keep `array_init` enabled and disable `typesafe_array_init` again. And in general, we only enable analyzer rules that really have a benefit. In this case `array_init` is sufficient as long as it does not cause false-positives. Right?

Sorry for that!

jpsim:
No need to be sorry, you did nothing wrong, but based on this discussion I'd say we should revert. Can you take care of that when you get a chance?

Analyzer rules are already way longer with Xcode 13.3 (see https://github.com/realm/SwiftLint/pull/3921) that we should be weary of adding them to CI if they don't add clear value to the project.

SimplyDanny:
Sure. I'll take care of the revert.

PR #3948, opened 2022-04-12
Revert "Enable `typesafe_array_init` rule in repository (#3942)"

SimplyDanny:
See #3942 for the reasons behind the revert.


jpsim:
Thank you!

PR #3949, opened 2022-04-12
Don't trigger `attributes` violation when using `@Sendable`

### LINKED ISSUE #3945: attributes violation incorrectly triggered when using @Sendable
### New Issue Checklist

### Describe the bug

Attributes Violation: Attributes should be on their own lines in functions and types, but on the same line as variables and imports. (attributes)

The above rule is incorrectly triggered by the following code:
```swift
// This triggers a violation:
    @available(iOS, introduced: 14, obsoleted: 15)
    func refreshable(action: @escaping @Sendable () async -> Void) -> some View {
        modifier(RefreshableModifier(action: action))
    }
```

The warning goes away if `@Sendable` is removed.

marcelofabri:
Fixes #3945

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,10 @@
   [Christopher Hale](https://github.com/chrispomeroyhale)
   [#3636](https://github.com/realm/SwiftLint/pull/3626)
 
+* Don't trigger `attributes` violations when using `@Sendable`.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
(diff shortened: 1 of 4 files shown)


marcelofabri:
Looks like `@Sendable` can be used in function declarations as well, so need to tweak this

marcelofabri:
Looks like the new warnings are actually legit and fall into some categories who were previously unintentionally ignored:

- `class`/`static` functions
- Custom property wrappers annotations
- Protocol declarations

PR #3923, opened 2022-03-29
cannot lint multiple files when using --path command option

coffmark:
Hello, i'm @coffmark. thank you for all your maintenance & update.

### motivation
1. `swiftlint lint --fix --format --path {first-file-path} {second-file-path}` 
This command only executes {first-file-path}, i expect execute {first-file-path} and {second-file-path}

2. `swiftlint lint --fix --format {first-file-path} {second-file-path}`
This command executes {first-file-path} & {second-file-path}, this is as expected.

It only doesn't work if `--path` command option is present.

### solution
when execute `swiftlint lint --fix --format --path {first-file-path} {second-file-path}`
```swift
@Option(help: ...)
var path: String? // { is assigned to first-file-path }

@Argument(help: ...)
var paths = [String]() // { is assigned to second-file-path }
```
Now, `allpaths` only gets `path`, so `paths` are ignored.
The solution is to add `paths` and `path` to `allpaths`** 

Looking forward to hearing back from you!

PR #3952, opened 2022-04-15
Deprecate the `--path` option in favor of the positional `paths` argument

SimplyDanny:
There is no real benefit for this option. Files and folders to lint/analyze
can be passed to `swiftlint` as last arguments. There cannot be multiple
`--path` arguments. If there is more than one only the last path is considered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 #### Breaking
 
-* None.
+* Deprecate the `--path` options for `lint`/`analyze` commands. Prefer the
+  positional paths that can be added last to both commands.  
+  [SimplyDanny](https://github.com/SimplyDanny)
 
 #### Experimental
 
--- a/Source/swiftlint/Commands/Analyze.swift
+++ b/Source/swiftlint/Commands/Analyze.swift
@@ -1,4 +1,5 @@
 import ArgumentParser
+import SwiftLintFramework
 
 extension SwiftLint {
     struct Analyze: ParsableCommand {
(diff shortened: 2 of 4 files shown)


marcelofabri:
> There is no real benefit for this option. 

The only benefit is not requiring people to change their existing scripts. Before SwiftLint moved to ArgumentParser, using `--path` was the way to go, so changing this will cause some churn for users.

SimplyDanny:
> > There is no real benefit for this option.
> 
> The only benefit is not requiring people to change their existing scripts.

Is this a strong reason to avoid the change? There are the release notes mentioning it as well as the warning on the command line. 

Since there already has been the strange behavior described in #3923 I would rather prefer to remove it instead of fixing issues/confusions caused by it.

marcelofabri:
> > > There is no real benefit for this option.
> > 
> > 
> > The only benefit is not requiring people to change their existing scripts.
> 
> Is this a strong reason to avoid the change? There are the release notes mentioning it as well as the warning on the command line.
> 
> Since there already has been the strange behavior described in #3923 I would rather prefer to remove it instead of fixing issues/confusions caused by it.

IMO it doesn't cost us too much to keep it around. #3923 was the only issue in this area afaict and since it's been fixed, I don't see us having to touch this for a while.

You'd be surprised by how many people don't read changelogs, don't lock the version of SwiftLint (so each developer in the team could be running a different version) or don't ever check SwiftLint's output in Xcode, so I'm afraid removing it might cause even more headache than keeping this around.

That said, I'm not opposed to actually merging this since it keeps the functionality around.

SimplyDanny:
I've indeed no feeling for the headaches such a change may cause.

On the other hand, just deprecating it without ever actually removing the option at some point would also not make sense.

So I'm torn now as to what to do. 🙈

SimplyDanny:
> That said, I'm not opposed to actually merging this since it keeps the functionality around.

Let's just merge the deprecation. Thinking about potential headaches can happen at a later point in time. 😉

PR #3954, opened 2022-04-16
Add comma_inheritance_rule rule

### LINKED ISSUE #3950: [Rule Request] Prefer comma-separated inheritance
### New Issue Checklist

### New rule request

I don't think I've ever seen this mentioned in any style guide, but my gut feeling is that using commas is way more popular - especially since the `&` syntax came later in Swift's life.

```swift
// triggers
struct Size: Codable & Equatable {
  let width: Int
  let height: Int
}

// ok
struct Size: Codable, Equatable {
  let width: Int
  let height: Int
}
```

Just severity

I think this is popular enough that it can be enabled by default.

marcelofabri:
Fixes #3950

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#3897](https://github.com/realm/SwiftLint/issues/3897)
 
+* Add `comma_inheritance` rule to validate that inheritance clauses use commas
+  instead of `&`.    
+  [Marcelo Fabri](https://github.com/marcelofabri)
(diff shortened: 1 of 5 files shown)


SimplyDanny (approved):
LGTM!

CHANGELOG.md
```diff
@@ -41,6 +41,11 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#3897](https://github.com/realm/SwiftLint/issues/3897)
 
+* Add `comma_inheritance` rule to validate that inheritance clauses use commas
+  instead of `&`.
```
SimplyDanny, line 45: There are 4 spaces appending this line.
```suggestion
  instead of `&`.  
```

PR #3955, opened 2022-04-17
Add back `tuple_pattern` rule

### LINKED ISSUE #3466: Add back `tuple_pattern` rule?
The rule was removed in https://github.com/realm/SwiftLint/commit/ed37c284197ab4597b588b891655cdee6982938a, but never added back again.

marcelofabri:
Fixes #3466

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,12 @@
   `return <expression>` in a function that returns `Void`.  
   [Marcelo Fabri](https://github.com/marcelofabri)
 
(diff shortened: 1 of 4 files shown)


marcelofabri:
Actually, this is now a warning on the compiler, so maybe we don't want to add this rule after all:

```
$ swift file.swift       
file.swift:1:20: warning: expression shuffles the elements of this tuple; this behavior is deprecated
let (x: a, y: b) = (y: 4, x: 5.0)
```

PR #3956, opened 2022-04-17
Add back `void_function_in_ternary` rule

marcelofabri:


SimplyDanny (changes requested):
The following code causes the violation in the last function body while it shouldn't:
```swift
func get() -> Int { 1 }
func get(from elements: [Int]) -> Int { elements[0] }
func f(elements: [Int]) -> Int { elements.isEmpty ? get() : get(from: elements) }
```

marcelofabri:
> The following code causes a violation in the last function body while it shouldn't:
> 
> ```swift
> func get() -> Int { 1 }
> func get(from elements: [Int]) -> Int { elements[0] }
> func f(elements: [Int]) -> Int { elements.isEmpty ? get() : get(from: elements) }
> ```

nice catch!

Source/SwiftLintFramework/Rules/Idiomatic/VoidFunctionInTernaryConditionRule.swift
```diff
@@ -34,6 +34,40 @@ public struct VoidFunctionInTernaryConditionRule: ConfigurationProviderRule, Aut
...
+            func compute(data: [Int]) -> Int {
+                data.isEmpty ? 0 : expensiveComputation(data)
+            }
+            """),
+            Example("""
+            func get() -> Int { 1 }
+            func get(from elements: [Int]) -> Int { elements[0] }
+            func f(elements: [Int]) -> Int {
+                elements.isEmpty ? get() : get(from: elements)
+            }
```
SimplyDanny, line 48: Looks like the case tested in this example is already covered by the one above. Do you still want to keep it?

SimplyDanny (approved):
LGTM!

Source/SwiftLintFramework/Rules/Idiomatic/VoidFunctionInTernaryConditionRule.swift
```diff
@@ -0,0 +1,235 @@
...
+    var isClosureImplictReturn: Bool {
+        guard let parent = parent?.as(CodeBlockItemListSyntax.self),
+              let grandparent = parent.parent else {
+            return false
+        }
+
+        return parent.children.count == 1 && grandparent.is(ClosureExprSyntax.self)
+    }
+
+    var isFunctionImplicitReturn: Bool {
```
SimplyDanny, line 176: The blocks of the computed variables `isFunctionImplicitReturn`, `isVariableImplicitReturn`, `isSubscriptImplicitReturn` and `isAcessorImplicitReturn` are quite similar and could be replace by calls to the method
```swift
private func check<T: SyntaxProtocol>(for type: T.Type, condition: (T) -> Bool = { _ in true }) -> Bool {
    guard let parent = parent?.as(CodeBlockItemListSyntax.self),
          let grandparent = parent.parent?.parent?.as(type) else {
        return false
    }
    return parent.children.count == 1 && condition(grandparent)
}
```
to avoid some code duplication.

marcelofabri: That's clever! However I think that the code duplication here is small enough that using a generic function with a closure as condition would make it harder to follow. We can always refactor this in the future if we notice we're making changes to all these variables

PR #3943, opened 2022-04-11
Allow custom attributes on lines directly before let/var declarations

SimplyDanny:
Fixes #2980.

PR #3959, opened 2022-04-25
Make let_var_whitespace recognize more @ annotations

Uncommon:
Currently, `let_var_whitespace` flags properties that have something like `@MainActor` or a property wrapper on the previous line. This is a fix for that.

For example, the rule shouldn't expect a blank line in between these two lines:
```
@MainActor
var someProperty: Int { ··· }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,9 @@
 
 #### Bug Fixes
 
-* None.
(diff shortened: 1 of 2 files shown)


SimplyDanny:
There already is #3943 fixing the issue. Since you came up with basically the same solution I feel confident to merge it.

Sorry for the double work. It's always recommended to look for similar (open) issues/PRs first. 😉

Issue #3975, opened 2022-05-12
SwiftLint is reporting less violations with Xcode 13.3 than with Xcode 13.2


rd-david-wahlandt:
### New Issue Checklist

- [X] Updated SwiftLint to the latest version
- [X] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

Our development team recently moved to MacBooks with M1 chip. Although SwiftLint runs and reports most errors, the `multiline_arguments_brackets` rule doesn't seem to work the same way as on Intel MacBooks.

Fine on M1, reporting warning on Intel:
```
views.append(ViewModel(title: "MacBook", subtitle: "M1", action: { [weak self] in
    print("action tapped")
}))
```

Warning on both M1 and Intel:
```
views.append(ViewModel(
    title: "MacBook", subtitle: "M1", action: { [weak self] in
    print("action tapped")
}))
```

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
swiftlint
Linting Swift files in current working directory
[...]
Done linting! Found 0 violations, 0 serious in 1343 files.
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure): 0.47.1
* Installation method used (Homebrew, CocoaPods, building from source, etc): Homebrew
* Paste your configuration file:

```yml
disabled_rules: # rule identifiers turned on by default to exclude from running
  - cyclomatic_complexity
  - file_length
  - force_cast
  - function_body_length
  - function_parameter_count
  - identifier_name
  - inclusive_language
  - large_tuple
  - line_length
  - nesting
  - todo
  - type_body_length
  - type_name
  - xctfail_message

opt_in_rules:
  - anyobject_protocol
  - array_init
  - capture_variable
  - closure_end_indentation
  - closure_spacing
  - collection_alignment
  - contains_over_filter_count
  - contains_over_filter_is_empty
  - contains_over_first_not_nil
  - contains_over_range_nil_comparison
  - discouraged_assert
  - discouraged_object_literal
  - empty_collection_literal
  - empty_count
  - empty_string
  - empty_xctest_method
  - expiring_todo
  - explicit_init
  - fallthrough
  - fatal_error_message
  - file_name_no_space
  - first_where
  - flatmap_over_map_reduce
  - ibinspectable_in_extension
  - identical_operands
  - implicit_return
  - implicitly_unwrapped_optional
  - joined_default_parameter
  - last_where
  - legacy_multiple
  - legacy_random
[...]

PR #3979, opened 2022-05-15
Make `for_where` independent of order in structure dictionary

SimplyDanny:
The order of elements in the dictionaries changed from Xcode 13.2.1 to 13.3.1. The implementation should not rely on a fixed order.

This fixes one of the findings in #3975.


SimplyDanny:
The reported violations all seem to be valid.

CHANGELOG.md
```diff
@@ -29,6 +29,11 @@
 * Prevent crash for private types named `_` in `type_name` rules.
   [sinoru](https://github.com/sinoru)
   [#3971](https://github.com/realm/SwiftLint/issues/3971)
+  
+* Make `for_where` rule implementation independent of order in structure
```
marcelofabri, line 33: can you please add that it fixes this rule on Xcode 13.3?

marcelofabri, line 33: can you please add that it fixes this rule on Xcode 13.3?

marcelofabri (approved):

CHANGELOG.md
SimplyDanny: Good point. Done.

PR #3989, opened 2022-06-04
Add accessibility_trait_for_button rule

rcole34:
Similar to the [`accessibility_label_for_image`](https://realm.github.io/SwiftLint/accessibility_label_for_image.html) rule, this rule catches another common mistake developers make in SwiftUI that makes their apps less accessible.

The accessibility button and link traits are used to tell assistive technologies that an element is tappable. When an element has one of these traits, VoiceOver will automatically read "button" or "link" after the element's label to let the user know that they can activate it. When using a UIKit `UIButton` or SwiftUI `Button` or `Link`, the button trait is added by default, but when you manually add a tap gesture recognizer to an element, you need to explicitly add the button or link trait. In most cases the button trait should be used, but for buttons that open a URL in an external browser we use the link trait instead. This rule attempts to catch uses of the SwiftUI `.onTapGesture` modifier where the `.isButton` or `.isLink` trait is not explicitly applied.

Per @rwapp, Android will automatically add the button trait when a tap gesture is applied, but until iOS catches up with that I hope for this rule to help remind more people to add the proper accessibility traits!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@
 
 * Rewrite `large_tuple` rule using SwiftSyntax, fixing false negatives.  
   [Marcelo Fabri](https://github.com/marcelofabri)
+* Add `accessibility_trait_for_button` rule to warn if a SwiftUI
+  View has a tap gesture added to it without having the button or
+  link accessibility trait.  
+  [Ryan Cole](https://github.com/rcole34)
 
 * Add methods from SE-0348 to `UnusedDeclarationRule`.  
   [JP Simard](https://github.com/jpims)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -272,6 +272,7 @@ macOS < 12.
 * Support `UIEdgeInsets` type in `prefer_zero_over_explicit_init` rule.  
   [KokiHirokawa](https://github.com/KokiHirokawa)
   [#3986](https://github.com/realm/SwiftLint/issues/3986)
+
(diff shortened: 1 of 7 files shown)


rcole34:
@jpsim @SimplyDanny here's another proposed rule to help catch more accessibility issues in SwiftUI apps! There were some SourceKittenDictionary extensions for parsing SwiftUI modifiers that I copied over from the `accessibility_label_for_image` rule since I was trying to keep the extensions private, but let me know if you think it would be better to move some of these extensions to an internal extension file to centralize them. Appreciate your help reviewing this!

SimplyDanny (changes requested):
Well done!

> There were some SourceKittenDictionary extensions for parsing SwiftUI modifiers that I copied over from the accessibility_label_for_image rule since I was trying to keep the extensions private, but let me know if you think it would be better to move some of these extensions to an internal extension file to centralize them. Appreciate your help reviewing this!

I would definitely like to ask you to move them into a separate extension for reusability.

rcole34:
@SimplyDanny I tried to make a more generic way to check for modifiers with a given name and arguments that can hopefully scale for other SwiftUI linter rules. Happy to work through some revisions of it if you see room for improvement!

SimplyDanny:
Sorry for the late response! Could you please rebase your branch? Notice that the protocol `AutomaticTestableRule` has been removed. All rules are now automatically tested. The file `AutomaticRuleTests.generated.swift` has been renamed to `GeneratedTests.swift`.

SimplyDanny (approved):

PR #3999, opened 2022-06-18
Skip `redundant_optional_initialization` rule on properties being custom annotated

### LINKED ISSUE #3998: `redundant_optional_initialization` warns on wrapped nilable property leading to compile errors
### New Issue Checklist

### Describe the bug

In the example

```swift
@propertyWrapper
struct Wrapper<T> {
    let name: String
    var wrappedValue: T

    init(wrappedValue value: T, _ name: String) {
        self.name = name
        self.wrappedValue = value
    }
}

struct Wrapped {
    @Wrapper("name")
    var flag: Bool? = nil // redundant_optional_initialization warn here
}
```

SwiftLint warns on the wrapped property assigned to `nil`. The property wrapper requires this explicit initialization, however. Removing the explicit assignment leads to a compile error.

```bash
$ swiftlint lint main.swift
Linting Swift files at paths main.swift
Linting 'main.swift' (1/1)
main.swift:14:20: warning: Redundant Optional Initialization Violation: Initializing an optional variable with nil is redundant. (redundant_optional_initialization)
```

### Environment

* SwiftLint 0.47.1
* Homebrew
* Default configuration
* Xcode 13.4.1

SimplyDanny:
Fixes #3998.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,11 @@
   false-positives.  
   [SimplyDanny](https://github.com/SimplyDanny)
 
+* Skip `redundant_optional_initialization` rule on properties being custom
+  annotated.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+  [#3998](https://github.com/realm/SwiftLint/issues/3998)
+
 ## 0.47.1: Smarter Appliance
 
 #### Breaking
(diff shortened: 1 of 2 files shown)


SimplyDanny:
This fix now seems too strict to me. Skipping all custom attributes also disables the rule for all property wrappers (`@State`, `@Environment`, ...) which work fine with nilable values not being explicitly initialized. In fact, it's possible to implement property wrappers in a way that doesn't require the explicit `= nil`.

An alternative would be a configuration option for this rule allowing to list all attributes that let the rule ignore declarations annotated with them.

PR #4002, opened 2022-06-20
Do not trigger `unavailable_condition` rule if other `#(un)available` checks are involved

### LINKED ISSUE #3985: #available and #unavailable cannot be in the same statement
### New Issue Checklist

### Describe the bug

The following triggers the unavailable_condition warning but when using the recommended code, it gives compiler error

```
if #available(macOS 11.0, *) {
   // Do nothing
} else if #available(macOS 10.15, *) {
   print("do some stuff")
}
```

```
if #unavailable(macOS 11.0, *),  #available(macOS 10.15, *) { // #available and #unavailable cannot be in the same statement
   print("do some stuff")
}
```

-

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.47.1
* Paste your configuration file: -
* Which Xcode version are you using (check `xcodebuild -version`)? Xcode 13.4 Build version 13F17a
* Do you have a sample that shows the issue? In the `Describe the bug` section

SimplyDanny:
Fixes #3985.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Idiomatic/UnavailableConditionRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/UnavailableConditionRule.swift
@@ -9,7 +9,7 @@ public struct UnavailableConditionRule: ConfigurationProviderRule, AutomaticTest
     public static let description = RuleDescription(
         identifier: "unavailable_condition",
         name: "Unavailable Condition",
-        description: "Use #unavailable instead of #available with an empty body.",
+        description: "Use #unavailable/#available instead of #available/#unavailable with an empty body.",
         kind: .idiomatic,
         minSwiftVersion: .fiveDotSix,
         nonTriggeringExamples: [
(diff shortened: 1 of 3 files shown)


Source/SwiftLintFramework/Rules/Idiomatic/UnavailableConditionRule.swift
```diff
@@ -9,7 +9,7 @@ public struct UnavailableConditionRule: ConfigurationProviderRule, AutomaticTest
     public static let description = RuleDescription(
         identifier: "unavailable_condition",
         name: "Unavailable Condition",
-        description: "Use #unavailable instead of #available with an empty body.",
+        description: "Use #unavailable/#available instead of #available/#unavailable with an empty body.",
```
marcelofabri, line 12: The idea of this rule was to catch conditions that could be replaced with `#unavailable`.

Since now it's apparently catching empty `#unavailable` with an `#else`, we should use specific warning messages on the violations for clarity.

Source/SwiftLintFramework/Rules/Idiomatic/UnavailableConditionRule.swift
```diff
@@ -45,35 +61,83 @@ public struct UnavailableConditionRule: ConfigurationProviderRule, AutomaticTest
...
+                    byteOffset: ByteCount(check.positionAfterSkippingLeadingTrivia.utf8Offset)),
+                reason: provideViolationReason(for: check)
+            )
+        }
+    }
+
+    private func provideViolationReason(for check: SyntaxProtocol) -> String {
+        switch check {
+        case is AvailabilityConditionSyntax:
+            return "Use #unavailable instead of #unavailable with an empty body."
```
marcelofabri, line 92:
```suggestion
            return "Use #unavailable instead of #available with an empty body."
```

SimplyDanny: Oh dear! 🤦‍♂️

marcelofabri (approved):

PR #4003, opened 2022-06-23
Switch to SwiftPM conditional dependencies API

keith:
This API allows us to mostly remove `#if` conditionals which enables
cross compilation from macOS -> Linux. It also removes the thrashing of
Package.resolved if you compile in a docker image.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,26 +1,6 @@
 // swift-tools-version:5.7
 import PackageDescription
 
-#if os(macOS)
-private let addCryptoSwift = false
-private let binaryPlugin = true
-#else
-private let addCryptoSwift = true
-private let binaryPlugin = false
-#endif
-
-let frameworkDependencies: [Target.Dependency] = [
-    .product(name: "IDEUtils", package: "swift-syntax"),
-    .product(name: "SourceKittenFramework", package: "SourceKitten"),
-    .product(name: "SwiftSyntax", package: "swift-syntax"),
-    .product(name: "SwiftSyntaxBuilder", package: "swift-syntax"),
-    .product(name: "SwiftParser", package: "swift-syntax"),
-    .product(name: "SwiftOperators", package: "swift-syntax"),
-    "SwiftyTextTable",
-    "Yams",
-]
-+ (addCryptoSwift ? ["CryptoSwift"] : [])
-
 let package = Package(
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -65,9 +54,18 @@ let package = Package(
...
             ]
         ),
-    ] + (staticSwiftSyntax ? [.binaryTarget(
+        .target(
+            name: "SyntaxParserBinaryWrapper",
+            dependencies: [.target(name: "lib_InternalSwiftSyntaxParser")],
+            // Pass `-dead_strip_dylibs` to ignore the dynamic version of `lib_InternalSwiftSyntaxParser`
+            // that ships with SwiftSyntax because we want the static version from
+            // `StaticInternalSwiftSyntaxParser`.
+            linkerSettings: [.unsafeFlags(["-Xlinker", "-dead_strip_dylibs"])]
```
keith, line 63: Ideally we could just add this to the `.binaryTarget` but that isn't supported. https://github.com/apple/swift-package-manager/issues/5618

Package.swift
```diff
@@ -1,14 +1,6 @@
...
 
-#if os(macOS)
-private let addCryptoSwift = false
-private let staticSwiftSyntax = true
-#else
-private let addCryptoSwift = true
-private let staticSwiftSyntax = false
-#endif
-
 #if os(Linux) && compiler(<5.6)
```
keith, line 4: We could probably handle this one with a combined `#if` for the compiler which + platform, but I assume it'll go away eventually anyways so we probably don't care about cross compilation for older versions

keith:
CI doesn't have `unzip`

jpsim:
> CI doesn't have `unzip`

Needs https://github.com/apple/swift-docker/pull/300

jpsim:
Can you please rebase? There appear to be new docker images that should contain the fix: https://hub.docker.com/_/swift/tags

keith:
🤞🏻

SimplyDanny (approved):

SimplyDanny:
Rebased and updated the PR.

li-bei:
Could we revert `binaryPlugin` related changes? It seems Xcode will build `swiftlint` target regardless the platform condition.

Related: #4998

PR #4006, opened 2022-06-24
Support arrays for `included` & `excluded` in custom rules

marcelofabri:
In certain cases, it's easier to specify multiple patterns than combining everything in one.

### DIFF
--- a/Tests/SwiftLintFrameworkTests/CustomRulesTests.swift
+++ b/Tests/SwiftLintFrameworkTests/CustomRulesTests.swift
@@ -155,6 +155,17 @@ class CustomRulesTests: XCTestCase {
         XCTAssertEqual(violations.count, 0)
     }
 
+    func testCustomRulesExcludedArrayExcludesFile() {
+        var (regexConfig, customRules) = getCustomRules(["excluded": ["\\.pdf$", "\\.txt$"]])
+
+        var customRuleConfiguration = CustomRulesConfiguration()
+        customRuleConfiguration.customRuleConfigurations = [regexConfig]
+        customRules.configuration = customRuleConfiguration
+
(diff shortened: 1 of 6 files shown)


Tests/SwiftLintFrameworkTests/CustomRulesTests.swift
```diff
@@ -155,6 +155,17 @@ class CustomRulesTests: XCTestCase {
...
+    func testCustomRulesExcludedArrayExcludesFile() {
+        var (regexConfig, customRules) = getCustomRules(["excluded": ["\\.pdf$", "\\.txt$"]])
+
+        var customRuleConfiguration = CustomRulesConfiguration()
+        customRuleConfiguration.customRuleConfigurations = [regexConfig]
+        customRules.configuration = customRuleConfiguration
+
+        let violations = customRules.validate(file: getTestTextFile())
+        XCTAssertEqual(violations.count, 0)
+    }
```
SimplyDanny, line 167: My feeling is that these test deserve an extensions. One test cannot verify the proper application of multiple exclude/include patterns if only a single file is validated. Especially the logics "none of the exclude patterns must match" and "consider a file as soon as a single include pattern matches" could be tested more thoroughly.

SimplyDanny (commented):
Looks good. I have one remark regarding tests, though:

Tests/SwiftLintFrameworkTests/CustomRulesTests.swift
marcelofabri: good call! extracted the logic so it's easier to test it directly and added some tests for it in eb1ef78

SimplyDanny (approved):

PR #4014, opened 2022-07-02
Make `duplicate_imports` rule correctable

abdulowork:
This PR:

* Makes `duplicate_imports` rule correctable.
* Fixes an issue where more than one duplicate resulted in redundant violations being reported. For example in this file: 
```swift
import Foundation
import Foundation
import Foundation
```
3 violations were reported with a redundant one in the end:
```
file.swift:2:1: warning: Duplicate Imports Violation: Imports should be unique. (duplicate_imports)
file.swift:3:1: warning: Duplicate Imports Violation: Imports should be unique. (duplicate_imports)
file.swift:3:1: warning: Duplicate Imports Violation: Imports should be unique. (duplicate_imports)
```

* Makes the rule faster for large number of imports. I benchmarked it on a file with 1000 `import Foundation` and it now completes in 160 ms. compared to 9 s. prior to the change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,10 @@
 * None.
 
 #### Enhancements
+  
+* Make `duplicate_imports` rule correctable. Fix `duplicate_imports` rule 
+  reporting redundant violations when more than one duplicate is present.  
+  [Timofey Solonin](https://github.com/abdulowork)
 
 * Support for building SwiftLint with bazel.  
   [JP Simard](https://github.com/jpsim)
--- a/Source/SwiftLintFramework/Rules/Idiomatic/DuplicateImportsRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/DuplicateImportsRule.swift
@@ -143,27 +143,26 @@ public struct DuplicateImportsRule: ConfigurationProviderRule, CorrectableRule {
     }
 
     public func correct(file: SwiftLintFile) -> [Correction] {
-        var contents = file.stringView.nsString
-        var corrections = [Correction]()
-
-        duplicateImports(file: file).reversed().filter {
+        let duplicateImports = duplicateImports(file: file).reversed().filter {
             file.ruleEnabled(violatingRange: $0.range, for: self) != nil
-        }.forEach { duplicateImport in
-            contents = contents.replacingCharacters(
-                in: duplicateImport.range,
-                with: ""
-            ).bridge()
-            corrections.append(
-                Correction(
-                    ruleDescription: Self.description,
-                    location: duplicateImport.location
-                )
-            )
         }
 
-        file.write(contents.bridge())
+        let violatingRanges = duplicateImports.map(\.range)
+        let correctedFileContents = violatingRanges.reduce(file.stringView.nsString) { contents, range in
+            contents.replacingCharacters(
+                in: range,
+                with: ""
+            ).bridge()
+        }
 
-        return corrections
+        file.write(correctedFileContents.bridge())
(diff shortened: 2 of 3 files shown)


Source/SwiftLintFramework/Rules/Idiomatic/DuplicateImportsRule.swift
```diff
@@ -101,11 +193,23 @@ private extension Line {
...
+
+        let importedSubpathParts = importIdentifier.split(separator: ".").map { String($0) }
+        guard !importedSubpathParts.isEmpty else { return [] }
+
+        return [
+            ImportSlice(
+                subpath: importedSubpathParts[0..<importedSubpathParts.count],
+                type: .complete
+            )
+        ] + stride(from: 1, to: importedSubpathParts.count, by: 1).map {
```
SimplyDanny, line 208: `stride` might be a bit over-engineered here. What about a simple range expression to iterate over?
```suggestion
        ] + (1 ..< importedSubpathParts.count).map {
```

Source/SwiftLintFramework/Rules/Idiomatic/DuplicateImportsRule.swift
```diff
@@ -65,34 +120,71 @@ public struct DuplicateImportsRule: ConfigurationProviderRule, AutomaticTestable
...
+            contents = contents.replacingCharacters(
+                in: duplicateImport.range,
+                with: ""
+            ).bridge()
+            corrections.append(
+                Correction(
+                    ruleDescription: Self.description,
+                    location: duplicateImport.location
+                )
+            )
```
SimplyDanny, line 161: We could be a bit more explicit about what's happening here by separating the different steps from each other:
1. Collect all duplicated import for which the rule is enabled.
2. Map them to ranges and reduce the list by repeated execution of `contents.replacingCharacters(...)`.
3. Write the file.
4. Map the entries from the list created in the first step to locations, map them in turn to corrections and return the new list.

SimplyDanny (changes requested):
Thank you for the fix and for making the rule correctable!

There are only a view requests for changes from my side.

CHANGELOG.md
```diff
@@ -23,6 +23,10 @@
   [tahabebek](https://github.com/tahabebek)
   [#4004](https://github.com/realm/SwiftLint/issues/4004)
   
+* Make `duplicate_imports` rule correctable. Fix duplicate_imports rule
```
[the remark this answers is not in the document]
abdulowork: Fixed the formatting

Source/SwiftLintFramework/Rules/Idiomatic/DuplicateImportsRule.swift
abdulowork: I agree. Replaced `stride` with a range expression

Source/SwiftLintFramework/Rules/Idiomatic/DuplicateImportsRule.swift
abdulowork: Please see the revised implementation. I hope it is clearer now.

SimplyDanny: It is. Thank you!

SimplyDanny (approved):

jpsim:
Thanks for the PR @abdulowork and for the review @SimplyDanny !

jpsim:
I can also confirm that for a large codebase this rule is about ~2x faster with this change 🙌

PR #4020, opened 2022-07-11
[Update] Translate Additions (Homebrew Apple Silicon)

DCherish:
I added some words in README_KR.md because I omitted an important keyword in translation. 🙂


SimplyDanny (approved):
Thanks again!

PR #4035, opened 2022-07-24
Check in build that Sourcery generated files are up-to-date

SimplyDanny:
The idea is to check in the build that files generated by Sourcery are up-to-date, so that something like #4034 cannot happen anymore.

The job verifies that the generated files do not have changes other than the header comment containing the used Sourcery version. As long as there are no other changes in the files a different version should be acceptable.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,6 +2,17 @@ trigger:
 - master
 
 jobs:
+- job: Sourcery
+  pool:
+    vmImage: 'macOS-12'
+  steps:
+    - script: brew install sourcery
+      displayName: Install Sourcery
+    - script: make sourcery
+      displayName: Generate sources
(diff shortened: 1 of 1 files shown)


jpsim (approved):
Thanks for adding this! Should’ve done this a long time ago. I think you could check for all files since running `make sourcery` should only affect Sourcery outputs.

jpsim:
And more generally thanks for your help maintaining SwiftLint! 🙇‍♂️

SimplyDanny:
Seems like the Azure pipeline is not running at all in the build. Is something wrong in the pipeline description, @jpsim?

jpsim:
> Is something wrong in the pipeline description

Yes: https://dev.azure.com/jpsim/SwiftLint/_build?definitionId=4&_a=summary&view=triggerIssues

> /azure-pipelines.yml: (Line: 17, Col: 15, Idx: 403) - (Line: 17, Col: 16, Idx: 404): While parsing a tag, did not find expected tag URI.

image

PR #4037, opened 2022-07-25
Require macOS 12 & Swift 5.6

jpsim:
This will unblock using Swift Concurrency features and updating to the latest versions of Swift Argument Parser.

This won't drop support for linting projects with an older toolchain / Xcode selected, as long as SwiftLint was _built_ with 5.6+ and is _running_ on macOS 12+. So the main breaking change for end users here is requiring macOS 12 to run.

However, the upside to using Swift Concurrency features is worth the breaking change in my opinion. Also being able to stay on recent Swift Argument Parser releases.


jpsim:
There are a lot of accumulated bug fixes in `master` since the last tagged release, so I'll cut one before merging this in case someone can't immediately update to macOS 12.

jpsim:
@SimplyDanny I'm curious to hear your thoughts on this if you have time.

jpsim:
Also @marcelofabri (sorry didn't ask you earlier, I wasn't sure how active you still were, I've been out of the loop for a while)

marcelofabri:
> Also @marcelofabri (sorry didn't ask you earlier, I wasn't sure how active you still were, I've been out of the loop for a while)

not _that_ active tbh, but I think this makes sense. People can keep using older versions of SwiftLint if they need to.

SimplyDanny:
As I understand, with this change, SwiftLint itself must be built and run with the latest Swift toolchain and OS. Projects in Xcode can still reference an older toolchain as long as SwiftLint is run with a recent one, isn't it?

So the change would not exclude projects running on older toolchains completely, but it would require some more work on their side to get everything running. Am I right about that?

In general, I very much like being up-to-date to make use of the latest and greatest features. 👍

> There are a lot of accumulated bug fixes in `master` since the last tagged release, so I'll cut one before merging this in case someone can't immediately update to macOS 12.

Sounds good! There has really happened a lot since the last release.

jpsim:
Actually, I was wrong this won't drop support for linting projects with an older toolchain / Xcode selected, as long as SwiftLint was _built_ with 5.6+ and is _running_ on macOS 12+. So the main breaking change for end users here is requiring macOS 12 to run.

SimplyDanny (approved):

PR #4039, opened 2022-07-26
Add support for native custom rules

jpsim:
This change makes it possible to add native custom rules when building SwiftLint via Bazel (possible as of https://github.com/realm/SwiftLint/pull/4038).

First, add a local bazel repository where custom rules will be defined to your project's `WORKSPACE`:

```python
local_repository(
    name = "swiftlint_extra_rules",
    path = "swiftlint_extra_rules",
)
```

Then in the extra rules directory, add an empty `WORKSPACE` and a `BUILD` file with the following contents:

```python
filegroup(
    name = "extra_rules",
    srcs = glob(["*.swift"]),
    visibility = ["//visibility:public"],
)
```

To add a rule (for example, `MyPrivateRule`) add the following two files:

```swift
// ExtraRules.swift
func extraRules() -> [Rule.Type] {
    [
        MyPrivateRule.self,
    ]
}
```

```swift
// MyPrivateRule.swift
import SourceKittenFramework
import SwiftSyntax

struct MyPrivateRule: ConfigurationProviderRule {
    var configuration = SeverityConfiguration(.error)

    init() {}

        identifier: "my_private_rule",
        name: "My Private Rule",
        description: "This is my private rule.",
        kind: .idiomatic
    )

        // Perform validation here...
    }
}
```

Then you can reference the rule in your configuration or source files as though they were built in to the official SwiftLint repo.

This means that you have access to SwiftLintFramework's internal API. We make no guarantees as to the stability of these internal APIs, although if you end up using something that gets removed please reach out and we'll make a best effort to maintain some level of support.

This PR also improves the linter cache on macOS to make it correctly invalidate previous results when custom native rules are edited. This even works when doing local development of SwiftLint, where previous it was necessary to use `--no-cache` when working on SwiftLint, now the cache should always work.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -6,7 +6,10 @@ load(
 
 swift_library(
     name = "SwiftLintFramework",
-    srcs = glob(["Source/SwiftLintFramework/**/*.swift"]),
+    srcs = glob(
+        ["Source/SwiftLintFramework/**/*.swift"],
+        exclude = ["Source/SwiftLintFramework/Rules/ExcludedFromBazel/ExtraRules.swift"],
+    ) + ["@swiftlint_extra_rules//:extra_rules"],
     module_name = "SwiftLintFramework",
     visibility = ["//visibility:public"],
     deps = [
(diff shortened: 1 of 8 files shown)


jpsim:
This depends on #4037 & #4038

keith (approved):

keith (approved):

SimplyDanny (approved):

PR #4044, opened 2022-07-26
Fix false positive in self_in_property_initialization

### LINKED ISSUE #4041: False positive in self_in_property_initialization
### New Issue Checklist

```bash
$ swiftlint lint
Linting Swift files at paths file.swift
Linting 'file.swift' (1/1)
/Users/marcelo.fabri/project/file.swift:2:5: warning: Self in Property Initialization Violation: `self` refers to the unapplied `NSObject.self()` method, which is likely not expected. Make the variable `lazy` to be able to refer to the current instance or use `ClassName.self`. (self_in_property_initialization)
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.48.0
* Installation method used (Homebrew, CocoaPods, building from source, etc)? `portable_zip`

No

13.4.1

```swift
class Foo {
    var bar: Bool = false {
        didSet {
            value = {
                if bar {
                    return self.calculateA()
                } else {
                    return self.calculateB()
                }
            }()
            print(value)
        }
    }

    var value: String?

    func calculateA() -> String { "A" }
    func calculateB() -> String { "B" }
}
```

marcelofabri:
Fixes #4041

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@
 #### Bug Fixes
 
 * Fix false positive in `self_in_property_initialization` rule when using
-closures inside `didSet` and other accessors.  
+  closures inside `didSet` and other accessors.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#4041](https://github.com/realm/SwiftLint/issues/4041)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,10 @@
 
 #### Bug Fixes
 
-* None.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -21,7 +21,10 @@
 
 #### Bug Fixes
 
-* None.
+* Fix false positive in `self_in_property_initialization` rule when using
+closures inside `didSet` and other accessors.
```
jpsim, line 25: Nit:

```suggestion
  closures inside `didSet` and other accessors.  
```

jpsim (approved):

Issue #4027, opened 2022-07-16
comma_inheritance breaks protocol composition in associated types


sjmadsen:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

When two or more protocols are part of a protocol composition on an associated type, changing the `&` to `,` can cause a syntax error. For example:
 
```
protocol Transformable {
    associatedtype Unit: Equatable & CaseIterable & UnitConvertible

    ...
}

protocol UnitConvertible where Self: Equatable {
    ...
}
```

When using `,` instead of `&`, compilation results in `Redundant conformance constraint 'Self.Unit' : 'Equatable'`. In this example, it is important that the associated type `Unit` is declared as the composition of `Equatable`, `CaseIterable`, and `UnitConvertible`.

##### Complete output when running SwiftLint, including the stack trace and command used

n/a - SwiftLint runs fine. It's just too eager about suggesting that all `&` be changed to `,` in protocol compositions. Since it's in the default rule set, `swiftlint lint --fix` can introduce syntax errors.

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? **0.47.1**
* Installation method used (Homebrew, CocoaPods, building from source, etc)? **Homebrew**
* Paste your configuration file:

```yml
opt_in_rules:
  - closure_end_indentation
  - contains_over_first_not_nil
  - empty_count
  - empty_string
  - first_where
  - implicit_return
  - modifier_order
  - nimble_operator
  - operator_usage_whitespace
  - overridden_super_call
  - prefer_nimble
  - prohibited_super_call
  - sorted_first_last
  - trailing_closure
  - unused_import
disabled_rules:
  - identifier_name
  - private_over_fileprivate
  - type_name
inclusive_language:
  override_terms:
    - blacklist
    - whitelist
    - slave
line_length:
  warning: 300
  error: 400
nesting:
  type_level:
    warning: 2
type_body_length:
  warning: 220
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? **Yes, but not applicable to this problem.**
  If so, paste their relative paths and respective contents.
* Which Xcode version are you using (check `xcodebuild -version`)? **13.4.0**
* Do you have a sample that shows the issue? Run `echo "[string here]" | swiftlint lint --no-cache --use-stdin --enable-all-rules`
[...]

PR #4047, opened 2022-07-26
Make comma_inheritance opt-in

sjmadsen:
This addresses issue #4027.

The Swift language grammar doesn't allow a comma in place of an ampersand in all cases. Specifically, if a protocol's associated type is required to conform to a composition of protocols and one of those protocols in turn requires conformance to one of the same protocols, the compiler will generate an error: "redundant conformance constraint". For example:

    protocol Transformable {
        associatedtype Unit: Equatable & CaseIterable & UnitConvertible

        ...
    }

    protocol UnitConvertible where Self: Equatable {
        ...
    }

When `&` is replaced with `,`, this produces `Redundant conformance constraint: 'Self.Unit' : 'Equatable'`.

My first thought was to disable autocorrect, but this penalizes those that want to use the rule without solving the problem for problematic code bases. It feels better to make the rule opt-in, so that developers who want to enforce the rule can still take advantage of autocorrection, while code bases where it is an issue can either leave it disabled or opt-in and selectively disable where necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
   values.  
   [JP Simard](https://github.com/jpsim)
 
+* Make `comma_inheritance` an opt-in rule.  
+  [Steve Madsen](https://github.com/sjmadsen)
+  [#4027](https://github.com/realm/SwiftLint/issues/4027)
+
 #### Experimental
 
 * None.
@@ -31,10 +35,6 @@
   closures inside `didSet` and other accessors.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#4041](https://github.com/realm/SwiftLint/issues/4041)
-
-* Make `comma_inheritance` an opt-in rule.  
-  [Steve Madsen](https://github.com/sjmadsen)
-  [#4027](https://github.com/realm/SwiftLint/issues/4027)
 
 ## 0.48.0: Rechargeable Defuzzer
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#4041](https://github.com/realm/SwiftLint/issues/4041)
 
+* Make `comma_inheritance` an opt-in rule.  
+  [Steve Madsen](https://github.com/sjmadsen)
+  [#4027](https://github.com/realm/SwiftLint/issues/4027)
+
 ## 0.48.0: Rechargeable Defuzzer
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -32,6 +32,10 @@
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#4041](https://github.com/realm/SwiftLint/issues/4041)
 
+* Make `comma_inheritance` an opt-in rule.  
+  [Steve Madsen](https://github.com/sjmadsen)
+  [#4027](https://github.com/realm/SwiftLint/issues/4027)
```
SimplyDanny, line 37: This should rather be mentioned in the "Breaking" section since users need to change their configuration if they want to keep the rule enabled.

SimplyDanny (changes requested):

CHANGELOG.md
sjmadsen: Moved.

SimplyDanny (approved):

SimplyDanny:
Thank you!

jpsim:
Shouldn't the right fix here be to avoid reporting violations for associated types?

SimplyDanny:
> Shouldn't the right fix here be to avoid reporting violations for associated types?

There is no difference between associated types and protocol definitions. In the example
```swift
protocol UnitConvertible where Self: Equatable {}

protocol Transformable: Equatable, UnitConvertible {   // ⚠️
    associatedtype Unit: Equatable, UnitConvertible    // ⚠️
}
```
we get the same message from the compiler at both definitions. But: The messages are only warnings, as I see now. So I wonder why the SwiftLint rule and its fix does anything wrong here. Isn't it helpful to see that there are redundant conformances?

sjmadsen:
Now that you say that, I'm pretty sure I have "warnings as errors" turned on. I still don't think it's great for SwiftLint's default configuration to potentially introduce compiler warnings into a codebase when `--fix` is run.

SimplyDanny:
Well, I think introducing this kind of warning is fine. It is not caused by SwiftLint doing anything wrong but is just made visible to the compiler.

What is your view on this, @jpsim?

jpsim:
Yeah there's no hard policy, the obvious ideal would be for SwiftLint to never have any false positives, any false negatives and never correct violations to cause previously compiling code to not compile.

However, we're limited by operating on the pre-typechecked AST in lint rules, so we should strike a balance between offering broadly useful rules that as much as possible do the right thing.

If the false-positive rate is too high and can't easily be reduced, then marking a rule as opt-in seems like a reasonable compromise to me.

PR #4057, opened 2022-07-31
DuplicatedKeyInDictionaryLiteralRule: fix false positives scenarios.

OrEliyahu:
The current check doesn't take into consideration that keys can be
generated in runtime. For example using UUID initializer (as in the
added example) raises a false positive diagnosis. Note that the
suggested change will not raise on true scenarios which use call, such
as:
```swift
[
  generateSameKey(): "1",
  generateSameKey(): "2"
]
func generateSameKey() -> Int { return 0 }
```

However, it is better than raising false positives.

About the solution:
Apparently when dictionary is parsed all of its keys has
exactly the same kind, `source.lang.swift.structure.elem.expr`,
regardless of their content. So in order to figure out the what is the
actual key one must reparse it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 ## Master
 
+#### Bug Fixes
+
+* Fix false positive in `Duplicated Key in Dictionary Literal Violation` rule when using
+  keys that are generated at runtime with the same source code.  
+  [OrEliyahu](https://github.com/OrEliyahu)
+  [#4012](https://github.com/realm/SwiftLint/issues/4012)
+
 #### Breaking
 
 * SwiftLint now requires Swift 5.6 or higher to build, and macOS 12
--- a/Source/SwiftLintFramework/Rules/Lint/DuplicatedKeyInDictionaryLiteralRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/DuplicatedKeyInDictionaryLiteralRule.swift
@@ -28,6 +28,18 @@ public struct DuplicatedKeyInDictionaryLiteralRule: ASTRule, ConfigurationProvid
                     foo: "1",
                     bar: "2"
                 ]
+            """),
+            Example("""
+                [
+                    UUID(): "1",
+                    UUID(): "2"
+                ]
+            """),
+            Example("""
+                [
+                    #line: "1",
+                    #line: "2"
+                ]
             """)
         ],
         triggeringExamples: [
@@ -73,7 +85,7 @@ public struct DuplicatedKeyInDictionaryLiteralRule: ASTRule, ConfigurationProvid
             return []
         }
 
-        let keys = dictionaryKeys(with: file, dictionary: dictionary)
+        let keys = nonGeneratedDictionaryKeys(with: file, dictionary: dictionary)
         guard keys.count >= 2 else {
             return []
         }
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 ## Master
 
+#### Bug Fixes
+
+* Fix false positive in `Duplicated Key in Dictionary Literal Violation` rule when using
+  keys that are generated in runtime with the same code.
```
SimplyDanny, line 6:
```suggestion
  keys that are generated at runtime with the same source code.  
```

Source/SwiftLintFramework/Rules/Lint/DuplicatedKeyInDictionaryLiteralRule.swift
```diff
@@ -117,8 +132,36 @@ public struct DuplicatedKeyInDictionaryLiteralRule: ASTRule, ConfigurationProvid
         }
     }
 
+    private func isCodeGeneratedKey(keyExpression: String) -> Bool {
+      let contentFile = File(contents: keyExpression)
+      let request = Request.editorOpen(file: contentFile)
+
+      guard let response = try? request.send() else { return false }
```
SimplyDanny, line 139: I'm not absolutely sure, but another SourceKit request might be expensive. It doesn't provide much value either since we can already see from the string whether it is `#line` (the simplest check) or is a call (contains parentheses). Thus, another request is not actually needed here.

Source/SwiftLintFramework/Rules/Lint/DuplicatedKeyInDictionaryLiteralRule.swift
```diff
@@ -117,8 +132,36 @@ public struct DuplicatedKeyInDictionaryLiteralRule: ASTRule, ConfigurationProvid
...
+     }
+
     private struct DictionaryKey {
         let byteRange: ByteRange
         let content: String
     }
 }
+
+private extension Dictionary where Key == String, Value == SourceKitRepresentable {
+  var expressionKind: SwiftExpressionKind? {
```
SimplyDanny, line 151: Indentation should be 4 spaces.

SimplyDanny (commented):
Thank you for the fix! I think we can simplify it a bit avoiding the additional SourceKit request.

Source/SwiftLintFramework/Rules/Lint/DuplicatedKeyInDictionaryLiteralRule.swift
OrEliyahu: OK, note 2 things:
1.  Checking the `kind` is more general and also contains other directives like `#sourceLocation` and any other directive that will be added to the language in the future.
2. `SwiftExpressionKind.call` probably contains other scenarios that I don't think about (see new implementation)

SimplyDanny: Valid points. The cases we know of are covered now while still being very fast. If more scenarios come up or the language is extended we can enhance the implementation.

Another approach could be an implementation based on the real AST using SwiftSyntax. It is fast and as exact as we can be without a full compilation.

SimplyDanny (approved):

PR #4061, opened 2022-08-02
Verify all examples provided in rule descriptions

SimplyDanny:
All examples that are used in the documentation should be verified.

For some of the rules, this works without further work just by conforming to the `AutomaticTestableRule` protocol. Others require some customization of test execution, e.g. rules dealing with comments still trigger when wrapped into another comment.

### DIFF
--- a/Source/SwiftLintFramework/Models/Example.swift
+++ b/Source/SwiftLintFramework/Models/Example.swift
@@ -18,6 +18,12 @@ public struct Example {
     ///
     /// - SeeAlso: addEmoji(_:)
     public private(set) var testMultiByteOffsets: Bool
+    /// Whether tests shall verify that the example wrapped in a comment doesn't trigger
+    private(set) var testWrappingInComment: Bool
+    /// Whether tests shall verify that the example wrapped into a string doesn't trigger
+    private(set) var testWrappingInString: Bool
+    /// Whether tests shall verify that the disabled rule (comment in the example) doesn't trigger
+    private(set) var testDisableCommand: Bool
     /// Whether the example should be tested on Linux
     public private(set) var testOnLinux: Bool
     /// The path to the file where the example was created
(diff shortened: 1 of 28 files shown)


jpsim:
If we want to encourage automatically testing all examples, why not switch to an opt-out model for automatic test generation?

SimplyDanny:
> If we want to encourage automatically testing all examples, why not switch to an opt-out model for automatic test generation?

Good suggestion! Would you like to see the switch to the new model in this PR or is it a separate one? I would rather prefer to do it as a follow-up.

jpsim:
Yes, a follow-up is fine.

Tests/SwiftLintFrameworkTests/TestHelpers.swift
```diff
@@ -294,11 +294,17 @@ private func testCorrection(_ correction: (Example, Example),
 }
 
 private func addEmoji(_ example: Example) -> Example {
-    return example.with(code: "/* 👨‍👩‍👧‍👦👨‍👩‍👧‍👦👨‍👩‍👧‍👦 */\n\(example.code)")
+    if example.testMultiByteOffsets {
+        return example.with(code: "/* 👨‍👩‍👧‍👦👨‍👩‍👧‍👦👨‍👩‍👧‍👦 */\n\(example.code)")
+    }
+    return example
```
jpsim, line 300: Instead of returning the passed example here, why not avoid calling this function altogether if the example disables testing with multi-byte offsets?

Right now, doesn't this cause the example to be verified multiple times?

jpsim (changes requested):

Source/SwiftLintFramework/Models/Example.swift
```diff
@@ -73,6 +88,32 @@ public extension Example {
...
+    }
+
+    func skipDisableCommandTest() -> Example {
+        var new = self
+        new.testDisableCommand = false
+        return new
+    }
 
     /// Makes the current example focused.
     func focused() -> Example {
```
jpsim, line 119:
```
Source/SwiftLintFramework/Models/Example.swift:119:10: error: Unused Declaration Violation: Declarations should be referenced at least once within all files linted. (unused_declaration)
```

jpsim: I think we can just silence this one since it's designed to be a debug helper for local development.

Tests/SwiftLintFrameworkTests/TestHelpers.swift
SimplyDanny: Right. The example was run twice.

Source/SwiftLintFramework/Models/Example.swift
SimplyDanny: Silenced the warning and added a comment to make the intention clear.

jpsim (approved):
💯

PR #4063, opened 2022-08-04
Rewrite ColonRule with SwiftSyntax

jpsim:
Making it about 7x faster, finding some previously missed cases.

PR #4064, opened 2022-08-04
Add command to parse OSSCheck results

jpsim:


jpsim:
Not actually for merging, was just useful when auditing the OSSCheck results from https://github.com/realm/SwiftLint/pull/4063

PR #4065, opened 2022-08-04
Verify examples in rules by default and enforce explicit exclusion

SimplyDanny:
A rule must conform to ManuallyTestedExamplesRule to skip generation of a test for its examples.

Next step would be to get rid of the ManuallyTestedExamplesRule conformances of some rules.

### DIFF
--- a/Source/SwiftLintFramework/Protocols/Rule.swift
+++ b/Source/SwiftLintFramework/Protocols/Rule.swift
@@ -89,8 +89,8 @@ extension Rule {
 /// A rule that is not enabled by default. Rules conforming to this need to be explicitly enabled by users.
 public protocol OptInRule: Rule {}
 
-/// A rule that has unit tests automatically generated using Sourcery.
-public protocol AutomaticTestableRule: Rule {}
+/// A rule that does not automatically generate unit tests for its examples.
+public protocol ManuallyTestedExamplesRule: Rule {}
 
 /// A rule that is user-configurable.
 public protocol ConfigurationProviderRule: Rule {
--- a/Source/SwiftLintFramework/Rules/Idiomatic/DiscouragedObjectLiteralRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/DiscouragedObjectLiteralRule.swift
@@ -1,4 +1,4 @@
-public struct DiscouragedObjectLiteralRule: ASTRule, OptInRule, ConfigurationProviderRule {
+public struct DiscouragedObjectLiteralRule: ASTRule, OptInRule, ConfigurationProviderRule, ManuallyTestedExamplesRule {
     public var configuration = ObjectLiteralConfiguration()
 
     public init() {}
--- a/Source/SwiftLintFramework/Rules/Style/ColonRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/ColonRule.swift
@@ -2,8 +2,8 @@ import Foundation
 import SourceKittenFramework
 import SwiftSyntax
 
-public struct ColonRule: SubstitutionCorrectableRule, ConfigurationProviderRule, AutomaticTestableRule,
-                         SourceKitFreeRule {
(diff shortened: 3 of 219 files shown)


Source/SwiftLintFramework/Protocols/Rule.swift
```diff
@@ -89,8 +89,8 @@ extension Rule {
 /// A rule that is not enabled by default. Rules conforming to this need to be explicitly enabled by users.
 public protocol OptInRule: Rule {}
 
-/// A rule that has unit tests automatically generated using Sourcery.
-public protocol AutomaticTestableRule: Rule {}
+/// A rule that misses automatically generated unit tests for its (non-)triggering examples.
```
jpsim, line 92: Nit:

```suggestion
/// A rule that does not automatically generate unit tests for its examples.
```

Source/SwiftLintFramework/Rules/Idiomatic/DiscouragedObjectLiteralRule.swift
```diff
@@ -1,4 +1,4 @@
-public struct DiscouragedObjectLiteralRule: ASTRule, OptInRule, ConfigurationProviderRule {
+public struct DiscouragedObjectLiteralRule: ASTRule, OptInRule, ConfigurationProviderRule, ManuallyTestedExamplesRule {
```
jpsim, line 1: Seems like this one could remove this conformance if `DiscouragedObjectLiteralRuleTests.swift` was renamed to `DiscouragedObjectLiteralRuleConfigurationTests.swift` like you did with the other rules in #4061.

Source/SwiftLintFramework/Rules/Style/ColonRule.swift
```diff
@@ -2,8 +2,8 @@ import Foundation
 import SourceKittenFramework
 import SwiftSyntax
 
-public struct ColonRule: SubstitutionCorrectableRule, ConfigurationProviderRule, AutomaticTestableRule,
-                         SourceKitFreeRule {
+public struct ColonRule: SubstitutionCorrectableRule, ConfigurationProviderRule, SourceKitFreeRule,
+                         ManuallyTestedExamplesRule {
```
jpsim, line 6: How come this one changed from `AutomaticTestableRule` to `ManuallyTestedExamplesRule`?

jpsim (approved):
Needs a changelog entry in the breaking changes section since this changes the public API of SwiftLintFramework in a backwards-incompatible way.

jpsim:
There's also an opportunity here to audit the `ManuallyTestedExamplesRule` rules to see which ones can be easily migrated to have their tests automatically generated.

Source/SwiftLintFramework/Rules/Idiomatic/DiscouragedObjectLiteralRule.swift
SimplyDanny: I have already prepared another follow-up to do exactly that.

Source/SwiftLintFramework/Rules/Style/ColonRule.swift
SimplyDanny: Good question. The `AutomaticTestableRule` conformance is not true since there is no automatically generated test. Seems like the build step checking for changes caused by `make sourcery` does not yet work. I will investigate this.

SimplyDanny:
> There's also an opportunity here to audit the `ManuallyTestedExamplesRule` rules to see which ones can be easily migrated to have their tests automatically generated.

Good point! Coincidentally, I've already prepared another PR based on this one to remove all of the `ManuallyTestedExamplesRule` conformances again and let Sourcery take care of all the example verification tests. 😉

Source/SwiftLintFramework/Rules/Style/ColonRule.swift
SimplyDanny: See #4084.

PR #4067, opened 2022-08-07
Run all tests with Bazel

jpsim:
Other than the one test case that uses paths with spaces in them which Bazel doesn't support.


jpsim:
https://buildkite.com/swiftlint/swiftlint/builds/2847#018279fd-6826-4b4e-9fef-fb68458931b2

PR #4069, opened 2022-08-08
Migrate to Swift Concurrency

jpsim:
Appears to be 20%-40% slower than using GCD for concurrent computation, so not worth it right now.

Linting SwiftLint:

```
Benchmark 1: GCD
  Time (mean ± σ):      2.674 s ±  0.166 s    [User: 8.914 s, System: 0.296 s]
  Range (min … max):    2.514 s …  3.071 s    10 runs

Benchmark 2: Swift Concurrency
  Time (mean ± σ):      3.175 s ±  0.100 s    [User: 9.265 s, System: 0.904 s]
  Range (min … max):    3.081 s …  3.401 s    10 runs

Summary
  'GCD' ran
    1.19 ± 0.08 times faster than 'Swift Concurrency'
```

Linting Lyft:

```
Benchmark 1: GCD
  Time (mean ± σ):     59.847 s ±  0.141 s    [User: 247.340 s, System: 15.841 s]
  Range (min … max):   59.747 s … 59.947 s    2 runs

Benchmark 2: Swift Concurrency
  Time (mean ± σ):     84.071 s ±  0.333 s    [User: 292.524 s, System: 35.801 s]
  Range (min … max):   83.836 s … 84.307 s    2 runs

Summary
  'GCD' ran
    1.40 ± 0.01 times faster than 'Swift Concurrency'
```


jpsim:
The initial approach to replace all uses of `DispatchQueue.concurrentPerform` with CollectionConcurrencyKit had a significant performance penalty:

image

By reverting one use of it back, it appears to strike a better balance and actually outperform the previous GCD-only solution.

jpsim:
Going to break this up into smaller changes and assess the impact of each individually.

PR #4074, opened 2022-08-09
Use `concurrentFlatMap` in SwiftLintFramework

jpsim:
This has breaking changes to the SwiftLintFramework public API if anyone actually happens to be using that.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -20,6 +20,7 @@ swift_library(
     module_name = "SwiftLintFramework",
     visibility = ["//visibility:public"],
     deps = [
+        "@com_github_johnsundell_collectionconcurrencykit//:CollectionConcurrencyKit",
         "@com_github_jpsim_sourcekitten//:SourceKittenFramework",
         "@com_github_keith_swift_syntax//:SwiftSyntax",
         "@com_github_keith_swift_syntax//:SwiftSyntaxParser",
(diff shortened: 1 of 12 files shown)


jpsim:
Seems like the Swift Concurrency runtime doesn't play well with tsan at all: https://forums.swift.org/t/tsan-with-task-concurrency-warnings/57775

Someone says that's fixed in Xcode 14 / Swift 5.7.

jpsim:
> [](https://forums.swift.org/t/tsan-with-task-concurrency-warnings/57775)
> 
> Someone says that's fixed in Xcode 14 / Swift 5.7.

Confirmed SwiftLint tests pass with tsan enabled with Xcode 14 beta 5.

jpsim:
@SimplyDanny what do you think?

On one hand, it seems like the Swift Concurrency scheduler is more likely to see improvements than GCD concurrent perform, it seems like this is pretty close perf-wise to the previous approach, but on the other hand this makes async keywords spread through the API for no significant gain right now.

SimplyDanny:
> @SimplyDanny what do you think?

Without performance gains, the cluttered function interfaces and API changes I would actually postpone the update until the benefit is more visible.

Given that you have made all the work to update the interfaces already, it's also fine for me to merge it right now, however.

SimplyDanny (approved):

jpsim:
Yeah I'm not sold on this yet. Will hold off on merging until there's a compelling reason to do so.

PR #4076, opened 2022-08-09
Let all example verification tests be generated by Sourcery

SimplyDanny:
With this change, the protocol `ManuallyTestedExamplesRule` is no longer needed. All example tests are automatically generated. Shall the protocol be removed entirely after its short lifetime or is it still useful to keep it?


jpsim:
To fix the bazel tests, you'll need to change this line:

```diff
diff --git a/Tests/BUILD b/Tests/BUILD
index b92ed8d2e..30d18b315 100644
--- a/Tests/BUILD
+++ b/Tests/BUILD
@@ -8,7 +8,7 @@ swift_library(
         exclude = [
             "SwiftLintFrameworkTests/Resources/**",
             # Bazel doesn't support paths with spaces in them
-            "SwiftLintFrameworkTests/FileNameNoSpaceRuleTests.swift",
+            "SwiftLintFrameworkTests/FileNameNoSpaceRuleConfigurationTests.swift",
         ],
     ),
     module_name = "SwiftLintFrameworkTests",
```

jpsim:
> Shall the protocol be removed entirely after its short lifetime or is it still useful to keep it?

Let's remove it completely. You'll need to update the changelog entry to reflect the new state.

jpsim:
Instead of naming the custom tests `<rule name>ConfigurationTests`, what if they were consistently named `<rule name>CustomTests`? Seems like a better name for what these tests do, what do you think?

jpsim (approved):
+1 feel free to merge once you've addressed or considered my feedback

jpsim:
With adec1c76eea045137685550e999bbcd72c9c1293 bazel test failures will start being printed on CI which should make these failures easier to debug in the future

SimplyDanny:
> Instead of naming the custom tests `<rule name>ConfigurationTests`, what if they were consistently named `<rule name>CustomTests`? Seems like a better name for what these tests do, what do you think?

Alternatively, we could name the auto-generated tests something like `<rule name>ExampleTests` and allow the easier name `<rule name>Tests` for the manually written ones. Either way I would again prepare a follow-up for the renaming.

jpsim (approved):

PR #4077, opened 2022-08-10
chore(typo): fix typo on func

stavares843:


jpsim:
Thanks! But in the future, please only put user-facing changes in the changelog.

PR #4083, opened 2022-08-10
Rename test classes

SimplyDanny:


.sourcery/AutomaticRuleTests.stencil
```diff
@@ -5,7 +5,7 @@ import XCTest
 
 {% for rule in types.structs %}
 {% if rule.name|hasSuffix:"Rule" %}
-class {{ rule.name }}Tests: XCTestCase {
+class {{ rule.name }}ExamplesTests: XCTestCase {
```
jpsim, line 8: Maybe `GeneratedTests` instead? Since the hand-written tests may also use "examples".

```suggestion
class {{ rule.name }}GeneratedTests: XCTestCase {
```

And `AutomaticRuleTests.{stencil,swift}` could be renamed to `GeneratedTests.{stencil,swift}`

SimplyDanny: Good suggestion!

jpsim (approved):
Thanks for considering my feedback!

Issue #4090, opened 2022-08-14
closure_spacing false positive when using Swift Regex literals involving brackets


JonathanDowning:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

SwiftLint erroneously raises a `closure_spacing` warning when using the new Swift `Regex` literal syntax which includes an opening and closing bracket.

```swift
let r = /\{\}/ // ⚠️ Closure Spacing Violation: Closure expressions should have a single space inside each brace. (closure_spacing)
```

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint
main.swift:1:10: warning: Closure Spacing Violation: Closure expressions should have a single space inside each brace. (closure_spacing)

```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.47.1
* Installation method used (Homebrew, CocoaPods, building from source, etc)? Homebrew
* Paste your configuration file:

```yml
only_rules
  - closure_spacing
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? No
* Which Xcode version are you using (check `xcodebuild -version`)? Build version 14A5294e
* Do you have a sample that shows the issue? 
```sh
echo "let r = /\{\}/" | swiftlint lint --no-cache --use-stdin --enable-all-rules
```

jpsim:
Fixed in #4092.

PR #4091, opened 2022-08-14
Skip braces in regex literals

SimplyDanny:
This is a workaround for the false positive reported in #4090.

Regex literals will only be supported in Swift 6. For that reason, SourceKit does not yet know anything about the syntax type "Regex Literal". The correct fix would be skipping regex literals completely in the rule like it is done for strings and comments.

The lookarounds in the regex to find the braces make the rule significantly slower as we see in the report below.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,10 @@ accordingly._
   [SimplyDanny](https://github.com/SimplyDanny)
   [#4030](https://github.com/realm/SwiftLint/issues/4030)
 
+* Skip braces in regex literals to avoid false positive in `closure_spacing` rule.  
+  [SimplyDanny](https://github.com/SimplyDanny)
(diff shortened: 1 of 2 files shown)


jpsim:
Alternatively, here's a SwifSyntax implementation that I think mostly works, I just haven't implemented corrections yet: https://github.com/realm/SwiftLint/compare/jp-closure-spacing-rule-swift-syntax

Feel free to pick it up if you have time, otherwise I can probably finish it later this week.

SimplyDanny:
> Alternatively, here's a SwifSyntax implementation that I think mostly works, I just haven't implemented corrections yet: https://github.com/realm/SwiftLint/compare/jp-closure-spacing-rule-swift-syntax

Looks good to me so far!

> Feel free to pick it up if you have time, otherwise I can probably finish it later this week.

I'll let you know once I find the time to continue with it.

jpsim:
#4092 is ready to go now

PR #4101, opened 2022-08-16
WIP: Rewrite mark rule using SwiftSyntax

jpsim:
Still needs work, especially the correction logic.


jpsim:
https://github.com/realm/SwiftLint/pull/5408

PR #4105, opened 2022-08-16
Fix first_where false negative

KS1019:
When filtering against an array of dictionaries with String keys, SwiftLint does not trigger a violation. This PR adds a new triggering example and a fix to address the issue.


jpsim (approved):
Looks good, can you add a changelog entry please? Thanks!

KS1019:
@jpsim Hi, I just added a new entry in CHANGELOG.md. Can I just merge this PR after the checks pass?

PR #4109, opened 2022-08-18
Fix broken correction for `explicit_init` rule

KS1019:
When there is space or newline between class name and .init, correcting the Swift code currently causes a compilation error. This PR adds new examples and a fix to address the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -110,7 +110,7 @@ accordingly._
 
 * Fix broken correction for `explicit_init` rule
   [KS1019](https://github.com/KS1019/)
-  [#4109](https://github.com/realm/SwiftLint/pull/4109)
+
 
 ## 0.48.0: Rechargeable Defuzzer
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -108,6 +108,10 @@ accordingly._
   [KS1019](https://github.com/KS1019/)
   [#4105](https://github.com/realm/SwiftLint/pull/4105)
 
+* Fix broken correction for `explicit_init` rule
+  [KS1019](https://github.com/KS1019/)
+  [#4109](https://github.com/realm/SwiftLint/pull/4109)
+
 ## 0.48.0: Rechargeable Defuzzer
 
 This is the last release to support building with Swift 5.5.x and running on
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Rules/Idiomatic/ExplicitInitRule.swift
```diff
@@ -40,14 +40,21 @@ public struct ExplicitInitRule: SubstitutionCorrectableASTRule, ConfigurationPro
               obs2,
               resultSelector: { MyType.init($0, $1) }
             ).asMaybe()
-            """)
+            """),
+            Example("let int = Int\n.init(1.0)"),
```
jpsim, line 44: Can you please use multiline string literals here to make this easier to read?

KS1019: Fixed by 9a5c286

CHANGELOG.md
```diff
@@ -108,6 +108,10 @@ accordingly._
   [KS1019](https://github.com/KS1019/)
   [#4105](https://github.com/realm/SwiftLint/pull/4105)
 
+* Fix broken correction for `explicit_init` rule
+  [KS1019](https://github.com/KS1019/)
+  [#4109](https://github.com/realm/SwiftLint/pull/4109)
```
jpsim, line 113: Some minor nits here about changelog entry formatting. We usually have two trailing spaces to force a new line and only link to fixed issues, not PRs introducing the fix/change.

Source/SwiftLintFramework/Rules/Idiomatic/ExplicitInitRule.swift
```diff
@@ -40,14 +40,64 @@ public struct ExplicitInitRule: SubstitutionCorrectableASTRule, ConfigurationPro
               obs2,
               resultSelector: { MyType.init($0, $1) }
             ).asMaybe()
+            """),
+            Example("""
+            let int = Int
+            .init(1.0)
+            """),
```
jpsim, line 47: Can you please exclude these examples from the generated documentation?

```suggestion
            """).excludeFromDocumentation(),
```

jpsim (approved):
Thanks for the fix!

PR #4111, opened 2022-08-18
Replace manual backtracking by location information from the captured group

SimplyDanny:
I think this is a bit more straightforward and easier to understand than the backtracking. Since the regex has already been used for matching, the location information can be extracted from the captured groups.


jpsim (approved):
I don’t have a preference. You can merge this if you like.

PR #4114, opened 2022-08-19
Run Buildkite CI jobs with Xcode 13.4.1

jpsim:
Previously ran on Xcode 13.3.0


jpsim:
All the jobs appear to work well, so I've set Xcode 13.4.1 as the xcode-selected version on the CI machines.

PR #4118, opened 2022-08-23
feat: includePaths support star path, matching star path.

fengsh998:
add ./**/*Test* to the excluded configuration section.
included field can support eg: /**/*Test, **/*Test*, **/*Test, /Demo/**/*Tm*/**/mo.


fengsh998 (commented):
I think no problem

fengsh998:
path : /Example/**/*Test*/**/fr*work
** represent any path dirs,
*Test* : represent begin start any string, include Test key word, end any string.  not matching  Test 
fr*work: matching begin fr ,include any string and end work.

Issue #4117, opened 2022-08-23
Docker Linux container CLI hangs


josh:
It seems like some of the recent async changes have caused a regression in the published Docker Linux container. Testing from the `latest` tag, the linting runs as expected but the process just hangs and never quits. The most recently good container is the `5.5-latest` tag.

❌ ghcr.io/realm/swiftlint:latest
❌ ghcr.io/realm/swiftlint:0.49.0-rc.2
❌ ghcr.io/realm/swiftlint:0.49.0-rc.1
✅ ghcr.io/realm/swiftlint:5.5-latest
✅ ghcr.io/realm/swiftlint:5.5-0.48.0 

Doing git bisect rebuilding the [Dockerfile](https://github.com/realm/SwiftLint/blob/master/Dockerfile) leads to a97718f4c6a80e799116b99f53802c849273e162 as the commit introducing the hang. Guessing it has something to do with the new concurrency runtime on Linux and is just stuck waiting in the event loop.

The container also seems fine on `master` using a standard debug build with `swift build`. Maybe the [Dockerfile](https://github.com/realm/SwiftLint/blob/master/Dockerfile) needs some [build flag tweaks](https://github.com/realm/SwiftLint/blob/2960ae6a7acb254e36f784e954c7a973ed9c7ee3/Dockerfile#L18) or there's [some new concurrency runtime object file that needs to be copied into the runtime image](https://github.com/realm/SwiftLint/blob/2960ae6a7acb254e36f784e954c7a973ed9c7ee3/Dockerfile#L32-L36)?

jpsim:
Thanks for flagging this, I can repro:

```console
docker run -it ghcr.io/realm/swiftlint:0.49.0-rc.2 bash
$ swiftlint version          
0.49.0-rc.2
$ touch file.swift
$ swiftlint file.swift
Linting Swift files at paths file.swift
Linting 'file.swift' (1/1)
Done linting! Found 0 violations, 0 serious in 1 file.
^ never terminates
```

josh:
> Before I keep going, can you make sure this image works for you?

Yep, it works!

Another hacky patch I was experimenting with while debugging was to [always `exit(0)` where lint checks the error count](https://github.com/realm/SwiftLint/blob/7259c02ee5b3cb757e4210071b5a2fe94a8c5ffc/Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift#L103). Kinda ironic there is no bug if your source code has linting issues hehe.

[...]

PR #4123, opened 2022-08-23
Fix hang after linting or analyzing on Linux

jpsim:
Fixes https://github.com/realm/SwiftLint/issues/4117


josh (approved):
👍🏻 

I believe only the analyze and lint commands are the only ones affected since they were switched over to `AsyncParsableCommand`.

PR #4129, opened 2022-08-26
Migrate EmptyXCTestMethodRule to SwiftSyntax fixing some false-positives

### LINKED ISSUE #3647: `empty_xctest_method` incorrectly triggering for setUpWithError and tearDownWithError
### New Issue Checklist

- [/] Updated SwiftLint to the latest version
- [/] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

Currently on `master`, the `empty_xctest_method` rule is incorrectly triggered for overridden `setUpWithError` and `tearDownWithError` methods in some cases.

If you update `EmptyXCTestMethodRuleExamples.nonTriggeringExamples` to include a new non-triggering example with one of the following, the `EmptyXCTestMethodRuleTests.testWithDefaultConfiguration` unit test will fail:

```swift
Example("""
        class TotoTests: XCTestCase {
            var foobar: String = ""

                foobar = "anything"
            }
        }
""")
```
or
```swift
Example("""
        class TotoTests: XCTestCase {
            var foobar: String?

            override func tearDownWithError() throws {
                foobar = nil
            }
        }
""")
```

If we update it to add a (redundant) `super` call it fixes the issue:

```swift
  try setUpWithError()
  foobar = "anything"
}
```

However there are two issues with this: 
- This assumes that you have the rule enabled to enforce super calls, which is and should continue to be separate from this rule
- This assumes that the super call is necessary, which for XCTest setup and tear down functions, it isn't (unless you have your own superclass or something)

Additionally, this has only started failing since updating SwiftLint, we had this rule enabled in 0.38.0 but building the latest version from `master` and using that causes a failure. I'm not sure exactly what version since 0.38.0 this started happening.

Looking at the `EmptyXCTestMethodRule` implementation and putting a breakpoint where the `StyleViolation` is returned, we can see that both `enclosedVarParameters` and `substructure` are showing as empty for the `functionMethodInstance` `subDictionary`, and that's what's causing the violation.

If I also `po dictionary` at the same breakpoint, I can't see any way of detecting that the function isn't actually empty. It's like that `foobar = ""` doesn't exist in the source.

I'm not too savvy with SourceKitten to know if this is a bug with SourceKitten, but since the function isn't actually empty, this is a bit of an issue.

### Environment

- Latest master version as of now
- Unit test (also works building from source and creating a test case with the same issue and running SwiftLint lint)
- Xcode 12.5

### LINKED ISSUE #3691: False positive for empty_xctest_method when assigning statements to _
### New Issue Checklist

### Describe the bug

I'm getting a false positive for `empty_xctest_method` with code like the following:

```
    func testStuff() {
        _ = 4 + 4
    }
```

and 

```
    func testStuff() {
        _ = 4 + 4
        _ = 4 + 4
    }
```

But not

```
    func testStuff() {
        let aaa = 4 + 4
    }
```

or 

```
    func testStuff() {
        _ = 4 + 4
        let aaa = 4 + 4
    }
```

When these methods are members of a subclass of XCTestCase

Sample input:

```
import XCTest

class MyTests: XCTestCase {
    func testStuff() {
        _ = 4 + 4
    }
}
```

```bash
$ swiftlint lint --config swiftlint.yml  t.swift

Linting Swift files at paths t.swift
Linting 't.swift' (1/1)
/tmp/t.swift:4:5: warning: Empty XCTest Method Violation: Empty XCTest method should be avoided. (empty_xctest_method)
```

### Environment

* SwiftLint version - 0.43.1
* Installation method used - HomeBrew

```yml
opt_in_rules:
  - empty_xctest_method
```

No

```
Xcode 12.4
Build version 12D4e
```

```swift
import XCTest

class MyTests: XCTestCase {
    func testStuff() {
        _ = 4 + 4
    }
}
```

SimplyDanny:
Fixes #3647.
Fixes #3691.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ accordingly._
 * Migrate `empty_xctest_method` rule to SwiftSyntax fixing some false positives.  
   [SimplyDanny](https://github.com/SimplyDanny)
   [#3647](https://github.com/realm/SwiftLint/issues/3647)
-  [#3691](https://github.com/realm/SwiftLint/issues/3691))
+  [#3691](https://github.com/realm/SwiftLint/issues/3691)
 
 ## 0.49.0: Asynchronous Defuzzer
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,10 @@ accordingly._
 
 #### Bug Fixes
 
-* None.
+* Migrate `empty_xctest_method` rule to SwiftSyntax fixing some false positives.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+  [#3647](https://github.com/realm/SwiftLint/issues/3647)
+  [#3691](https://github.com/realm/SwiftLint/issues/3691))
 
 ## 0.49.0: Asynchronous Defuzzer
 
--- a/Source/SwiftLintFramework/Rules/Lint/EmptyXCTestMethodRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/EmptyXCTestMethodRule.swift
@@ -1,6 +1,6 @@
-import SourceKittenFramework
+import SwiftSyntax
 
-public struct EmptyXCTestMethodRule: Rule, OptInRule, ConfigurationProviderRule {
+public struct EmptyXCTestMethodRule: OptInRule, ConfigurationProviderRule, SwiftSyntaxRule {
     public var configuration = SeverityConfiguration(.warning)
 
     public init() {}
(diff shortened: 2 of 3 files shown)


jpsim:
Nice! Can you please audit the OSSCheck results to make sure there are no regressions?

SimplyDanny:
> Nice! Can you please audit the OSSCheck results to make sure there are no regressions?

They report that the rule does neither trigger on `static var allTests = ...` nor the false-positives reported in the linked issues anymore. That should be fine.

The second violation warning really was a bug that is now fixed.

CHANGELOG.md
```diff
@@ -18,7 +18,10 @@ accordingly._
 
 #### Bug Fixes
 
-* None.
+* Migrate `empty_xctest_method` rule to SwiftSyntax fixing some false positives.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+  [#3647](https://github.com/realm/SwiftLint/issues/3647)
+  [#3691](https://github.com/realm/SwiftLint/issues/3691))
```
jpsim, line 24:
```suggestion
  [#3691](https://github.com/realm/SwiftLint/issues/3691)
```

Source/SwiftLintFramework/Rules/Lint/EmptyXCTestMethodRule.swift
```diff
@@ -14,34 +14,64 @@ public struct EmptyXCTestMethodRule: Rule, OptInRule, ConfigurationProviderRule
...
     )
 
-    public func validate(file: SwiftLintFile) -> [StyleViolation] {
-        return testClasses(in: file).flatMap { violations(in: file, for: $0) }
+    public func makeVisitor(file: SwiftLintFile) -> ViolationsSyntaxVisitor? {
+        EmptyXCTestMethodRuleVisitor()
+    }
+}
+
+private class EmptyXCTestMethodRuleVisitor: SyntaxVisitor, ViolationsSyntaxVisitor {
```
jpsim, line 22: Nit:

```suggestion
private final class EmptyXCTestMethodRuleVisitor: SyntaxVisitor, ViolationsSyntaxVisitor {
```

Source/SwiftLintFramework/Rules/Lint/EmptyXCTestMethodRule.swift
```diff
@@ -14,34 +14,64 @@ public struct EmptyXCTestMethodRule: Rule, OptInRule, ConfigurationProviderRule
...
+    override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
+        guard let inheritanceList = node.inheritanceClause?.inheritedTypeCollection else {
+            return .skipChildren
+        }
+        let isTestClass = inheritanceList
+            .map(\.typeName)
+            .compactMap { $0.as(SimpleTypeIdentifierSyntax.self) }
+            .map(\.name)
+            .map(\.text)
+            .contains("XCTestCase")
```
jpsim, line 34: Two small perf optimizations you could do here:

1. Stop iterating over the inheritance list as soon as an `XCTestCase` type is found
2. Only iterate once

```suggestion
        let isTestClass = inheritanceList.contains { type in
            type.typeName.as(SimpleTypeIdentifierSyntax.self)?.name.text == "XCTestCase"
        }
```

Feel free to break down the closure in more statements, the important thing is to avoid the map/compactMap/map/map/contains to just a contains.

Source/SwiftLintFramework/Rules/Lint/EmptyXCTestMethodRule.swift
```diff
@@ -14,34 +14,64 @@ public struct EmptyXCTestMethodRule: Rule, OptInRule, ConfigurationProviderRule
...
+        .skipChildren
     }
 
-    // MARK: - Private
+    override func visitPost(_ node: FunctionDeclSyntax) {
+        if node.isOverride, node.hasEmptyBody {
+            violationPositions.append(node.funcKeyword.positionAfterSkippingLeadingTrivia)
+        } else if node.isTestMethod, node.hasEmptyBody {
+            violationPositions.append(node.funcKeyword.positionAfterSkippingLeadingTrivia)
+        }
```
jpsim, line 55: Nit:

```suggestion
        if (node.isOverride || node.isTestMethod) && node.hasEmptyBody {
            violationPositions.append(node.funcKeyword.positionAfterSkippingLeadingTrivia)
        }
```

Source/SwiftLintFramework/Rules/Lint/EmptyXCTestMethodRule.swift
```diff
@@ -14,34 +14,64 @@ public struct EmptyXCTestMethodRule: Rule, OptInRule, ConfigurationProviderRule
...
 
-    private func testClasses(in file: SwiftLintFile) -> [SourceKittenDictionary] {
-        let dict = file.structureDictionary
-        return dict.substructure.filter { dictionary in
-            guard dictionary.declarationKind == .class else { return false }
-            return dictionary.inheritedTypes.contains("XCTestCase")
+private extension FunctionDeclSyntax {
+    var isOverride: Bool {
+        if let modifiers = modifiers {
+            return modifiers.map(\.name).map(\.text).contains("override")
```
jpsim, line 62: Same here regarding not iterating multiple times and stopping iteration as soon as a match is found.

```suggestion
            return modifiers.contains { $0.name.text == "override" }
```

jpsim (approved):

Source/SwiftLintFramework/Rules/Lint/EmptyXCTestMethodRule.swift
SimplyDanny: This is a habit coming from other languages where the evaluation is always lazy. In Swift we need to start the chain with `.lazy` if the same behavior is desired.

Your suggestion is more concise, however, so I go with it. 😉

Issue #4127, opened 2022-08-25
swiftlint --fix path error


wasdkhan:
Is there anyway to support piping the auto correctable rules into stdout or another file, similar to how SwiftFormat does here: https://github.com/nicklockwood/SwiftFormat#command-line-tool ?

To my knowledge, SwiftLint autocorrect (--fix) only writes to disk, so we do not see the change before it makes it, only the warning/error.

Replication:
Colon spacing error sample file:
```
// Sample.swift
let x : String = " "
```

Command:
```
cat sample.swift | swiftlint lint --fix --use-stdin --path output.swift
```

Error:
```
file needs a path to call write(_:): file SwiftLintFile+Regex.swift, line 250
PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace.
Aborted (core dumped)
```

SimplyDanny:
There is currently no way to print corrected content to a file. The purpose of the `--path` argument you are using is only to specify input files. It will be ignored when `--use-stdin` is given in addition. To avoid this and other misunderstandings, the `--path` options is deprecated in the most recent versions already.

I propose a fix in #4132 for the error you report that makes sure autocorrected STDIN input is printed to STDOUT. That in turn allows you to pipe it into a file like:

```bash
cat sample.swift | swiftlint lint --fix --use-stdin > corrected-sample.swift
```

Would this fix be enough for you?

wasdkhan:
> Would this fix be enough for you?

Thanks, this looks great. I went ahead and wrote a wrapper script (copy sample.swift, grab the configs, fix the copy, output contents, delete copy), but official support is even better! 

Hopefully we can use this solution after we update our SwiftLint version.

SimplyDanny:
The change has been merged and should be included in the next release.

PR #4132, opened 2022-08-27
Print autocorrected STDIN input to STDOUT

SimplyDanny:
This fixes a fatal error when STDIN input is autocorrected and potentially satisfies #4127.


jpsim (approved):
The control flow is hard to reason about here, but I can’t think of a better or simpler way to do this, so I’m ok with this solution.

PR #4135, opened 2022-08-28
[CI] Run Jazzy with Bazel

jpsim:
To cache gems, the SwiftLint binary & the SourceKitten docs.


jpsim:
Works but seems to be a bit slower than running on the host machines directly.

Issue #2495, opened 2018-12-02
Rule Request: `guard self` (vs weakSelf, strongSelf)


freak4pc:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

1. Why should this rule be added? 

Since Swift 4.2 we can now do `guard let self = self` instead of `guard let strongSelf`, `guad let weakSelf`, `guard let 'self'` and all other sort of oddities. I'd like to enforce that. 

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

Trigger:
```swift
if let `self` = self
guard let `self` = self
guard let strongSelf = self
guard let [anything not "self"] = self
```

Doesn't trigger:
```swift
if let self = self
guard let self = self
guard let self = [anytime this is self]
```

3. Should the rule be configurable, if so what parameters should be configurable?
I don't think so.

4. Should the rule be opt-in or enabled by default? Why?
Opt-in

jpsim:
Thanks for the suggestion! I think this should be configurable on the rebinding identifier. There are still issues with Swift 4.2+ when rebinding `self` to `self`, especially around lldb (see [SR-6156](https://bugs.swift.org/browse/SR-6156)). For this reason, some projects have adopted a pattern to _always_ rebind `self` to something specifically different, in Lyft's case that's `guard let this = self`.

freak4pc:
Yup, that makes sense :)

lennartstolz:
I do feel like a grave desecrator - but I'd still love to see this rule becoming part of `SwiftLint`.

Especially because in the last ~3 yrs it almost came the standard in most codebases I'm working to use:

```swift
guard let self = self else { ... }
```

and the issue mentioned by @jpsim is also resolved by now. ✅ 

@marcelofabri : Do you think it's a valuable addition to SwiftLint? Is it "straightforward" to be implemented?
If so: I volunteer to implement it.

dwirandytlvk:
Any update related to this?

jpsim:
Adding a rule to do this in https://github.com/realm/SwiftLint/pull/4146

samrayner:
Thank you for this @jpsim! Would it be possible to put the configured bind_identifier into the warning message so that devs know what identifier to use when they see the warning?

PR #4146, opened 2022-08-30
Add new `self_binding` opt-in rule

jpsim:
To enforce that `self` identifiers are consistently re-bound to a common identifier name.

Configure `bind_identifier` to the name you want to use.

Defaults to `self`.

Addresses https://github.com/realm/SwiftLint/issues/2495

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,11 @@ accordingly._
 
 #### Enhancements
 
-* None.
+* Add new `self_binding` opt-in rule to enforce that `self` identifiers are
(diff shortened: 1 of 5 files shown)


jpsim:
You're almost there:

```yaml
opt_in_rules:
  - self_binding
self_binding:
  bind_identifier: this # Or whatever other identifier you like
```

samrayner:
Thank you for this @jpsim! Would it be possible to put the configured bind_identifier into the warning message so that devs know what identifier to use when they see the warning?

Issue #4048, opened 2022-07-27
Is there any way to specify an output file for SwiftLint reports instead of using '>' ?


tychop:
I want to save a json report of my SwiftLint BuildToolPlugin so I can use it elsewhere.

I am currently trying the following:

```
import PackagePlugin

@main
struct SwiftLintPlugin: BuildToolPlugin {
	func createBuildCommands(context: PluginContext, target: Target) async throws -> [Command] {
		return [
			.buildCommand(
				displayName: "Running SwiftLint for \(target.name)",
				executable: try context.tool(named: "swiftlint").path,
				arguments: [
					"lint",
					"--in-process-sourcekit",
					"--path", "\(target.directory.string)",
					"--config", "\(context.package.directory.string)/.swiftlint.yml",
					"--reporter", "json",
					">", "swiftlint.result.json"
				],
				environment: [:]
			)
		]
	}
}
```

But that basically boils down to the following command in which the quotes around the '>' character prevents creating an output file.

```
swiftlint lint \
	--in-process-sourcekit \
	--path BusinessModel \
	--config .swiftlint.yml \
	--reporter json ">" swiftlint.result.json
```

Is there any way to specify an output file for SwiftLint reports instead of using '>' ?

SimplyDanny:
As far as I see, there is currently no other way to write the report to a file.

One option is to support that in SwiftLint. Another one is to convince the maintainers of [swift-tools-support-core](https://github.com/apple/swift-tools-support-core) to avoid escaping of single arguments like `>`, `|`, `2>`, ... [This](https://github.com/apple/swift-tools-support-core/blob/0b77e67c484e532444ceeab60119b8536f8cd648/Sources/TSCBasic/StringConversions.swift#L66) should be the responsible line in their code.

jpsim:
@tychop can you please check out #4148 and let me know if it's what you're looking for?

tychop:
This works excellent !!! Thank you very much.

PR #4148, opened 2022-08-30
Add `--output` option to lint and analyze commands

jpsim:
To write to a file instead of to stdout.

Addresses https://github.com/realm/SwiftLint/issues/4048


Source/SwiftLintFramework/Extensions/Configuration+Remote.swift
```diff
@@ -141,11 +141,11 @@ internal extension Configuration.FileGraph.FilePath {
     ) throws -> String {
         if let cachedFilePath = cachedFilePath {
             if taskDone {
-                queuedPrint(
+                queuedPrintError(
```
SimplyDanny, line 144: Is the changed output stream for these messages an API change that should be mentioned in the CHANGELOG explicitly? Some users might refer to the output in their build/lint jobs.

SimplyDanny (approved):
LGTM! Just one question ...

Source/SwiftLintFramework/Extensions/Configuration+Remote.swift
jpsim: Yes, although I'd put this in the bug fixes section because the (unwritten?) policy has been for violations to be printed to stdout and any other logging to be printed to stderr so that you could always `swiftlint ... > outfile.txt` or `swiftlint --reporter json > outfile.json` and not have invalid formatting because there was some error/info logs mixed in with the reports.

jpsim: a0100ef

Issue #4094, opened 2022-08-15
`explicit_self` rule not working?


NachoSoto:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

Given this example in the rule:
```swift
struct ABC {
    func f1() {}
    func f2() {
        f1()
    }
}
```

No warnings are produced

##### Complete output when running SwiftLint, including the stack trace and command used
```bash
$ swiftlint
Linting Swift files in current working directory
Linting 'test.swift' (1/1)
Done linting! Found 0 violations, 0 serious in 1 file.
```

### Environment

* SwiftLint version: `0.48.0`
* Installation method used: Homebrew
* Paste your configuration file:

```yml
opt_in_rules:
  - explicit_self
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? No
* Which Xcode version are you using?  `Xcode 14.0 Build version 14A5294e`
* Do you have a sample that shows the issue?
```bash
echo "struct A { func f1() {} func f2() { f1() } }" | swiftlint lint --no-cache --use-stdin --enable-all-rules
```

NachoSoto:
I see now that this is an "analyzer rule". I hadn't heard about those before.
However this isn't working for me, so surely I'm doing something wrong:
```bash
swiftlint analyze
Error: Could not read compiler invocations
```

NachoSoto:
I see this is documented [here](https://github.com/realm/SwiftLint#analyze). Sorry for the noise!

jpsim:
Yeah analyzer rules are harder to set up and take along time to run but they allow you to run static analysis type rules that are quite powerful and accurate. Let me know if you need more help to set this up.

NachoSoto:
Thanks!

NachoSoto:
I wonder if it should be an error to add an invalid rule in `opt_in_rules`? That would have made me realize right away that I was doing something wrong.

SimplyDanny:
> I wonder if it should be an error to add an invalid rule in `opt_in_rules`? That would have made me realize right away that I was doing something wrong.

This is a good idea, indeed! Wrongly places rules in the configuration should be flagged.

jpsim:
> I wonder if it should be an error to add an invalid rule in `opt_in_rules`? That would have made me realize right away that I was doing something wrong.

Ah yes, good call, sorry I missed this comment.

PR #4151, opened 2022-09-02
Warn on rules listed in wrong section in configuration

SimplyDanny:
Opt-in rules do not need to be disabled, while rules active by default do not need to be enabled in the `opt_in_rules` section. Analyzer rules should be listed in their own section. In addition to misspelled rule names, warn for IDs listed in the wrong category.

This validation has been suggested in [this comment](https://github.com/realm/SwiftLint/issues/4094#issuecomment-1215901036),

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,14 @@
 * Add methods from SE-0348 to `UnusedDeclarationRule`.  
   [JP Simard](https://github.com/jpims)
 
+* Show warnings in the console for
+  * opt-in rules that are listed in the `disabled_rules` configuration section,
+  * active-by-default rules that are listed in the `opt_in_rules` configuration section and
+  * Analyzer rules that are listed in the `opt_in_rules` configuration section
+
+  as the user's intention might not be reflected correctly in the configuration.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+
 #### Bug Fixes
 
 * None.
--- a/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
@@ -57,6 +57,9 @@ extension Configuration {
             configurationDictionary: dict, disabledRules: disabledRules,
             optInRules: optInRules, onlyRules: onlyRules, ruleList: ruleList
         )
+        Self.warnAboutMisplacedAnalyzerRules(optInRules: optInRules, ruleList: ruleList)
+        Self.warnAboutMisplaced(optInRules: optInRules, ruleList: ruleList)
+        Self.warnAboutMisplaced(disabledRules: disabledRules, ruleList: ruleList)
 
         let allRulesWrapped: [ConfigurationRuleWrapper]
         do {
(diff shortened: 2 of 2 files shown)


Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
```diff
@@ -202,4 +203,15 @@ extension Configuration {
...
         }
     }
+
+    private static func warnAboutMisplacedAnalyzerRules(optInRules: [String], ruleList: RuleList) {
+        let analyzerRules = ruleList.list
+            .filter { $0.value.self is AnalyzerRule.Type }
+            .map(\.key)
+        let invalidKeys = Set(analyzerRules).intersection(optInRules)
+        if invalidKeys.isNotEmpty {
+            queuedPrintError("warning: Analyzer rules \(invalidKeys) are opt-in by default and should be " +
```
jpsim, line 213: Can this sort and pretty-print the rule IDs? right now it's a `Set` so the order could change every time this is run, and the way it's embedded in a sentence might not look good. Maybe a dashed list at the end?

> warning: Analyzer rules must be listed in the 'analyzer_rules' configuration section:
> - explicit_self
> - unused_declaration
> - unused_import

jpsim (commented):
I'm glad to see this overall, thanks for doing this.

Can we centralize the check in one place though?

Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
SimplyDanny: The set is now sorted. Like in the other warnings, for every misplaced rule, a separate warning is shown.

SimplyDanny:
> Can we centralize the check in one place though?

All checks are now centralized in `Configuration+Parser.swift`.

CHANGELOG.md
```diff
@@ -47,6 +47,14 @@
 * Add methods from SE-0348 to `UnusedDeclarationRule`.  
   [JP Simard](https://github.com/jpims)
 
+* Show warnings in the console for
+  * opt-in rules that are listed in the `disabled_rules` configuration section,
```
jpsim, line 51: How does this work with nested configurations? Like if an opt-in rule is enabled at the root configuration and disabled in a nested configuration?

SimplyDanny: Hm ... I guess it does not work at all. I did not think about nested configurations to be honest. Considering them now, my assumptions (no need to disable opt-in rules, no need to enable rules active by default) are not valid anymore. It makes sense to disable an opt-in rule if it was enabled in the parent configuration for example.

It would still be reasonable to validate merged configurations. But how to report warnings then? How to tell to which combination of configurations a warning belongs to?

The initial idea for these checks came from the fact that Analyzer rules were specified but were not considered because `swiftlint lint` was run instead of `swiftlint analyze`. It doesn't even matter whether they are listed in `analyzer_rules` or `opt_in_rules`. The important thing is to actually run SwiftLint in `analyze` mode.

Reflecting on all this, I think there is not much we can do to help the user. Analyzer rules must be triggered specially. SwiftLint cannot decide whether it is running in the right mode just by the configuration. And it also doesn't hurt to have rules listed redundantly or in the "wrong" section.

The only thing that could help is providing a hint like "`rule_id` as an Analyzer rule belongs into the `analyzer_rules` section to run it in the special `analyze` mode". Feels a bit over-engineered now, however. If all rules are listed in the correct section, there won't be a warning and you can still forget to run `swiftlint analyze` separately.

That said, it's probably better to skip implementation of these warnings entirely and close the PR. 😅

jpsim: Yes, sorry to say but I also think we should just close the PR.

SimplyDanny: Fine for me. Good you asked the question. Better than adding something unhelpful, confusing or even wrong.

At least, we though about it and know better now and for the future.

PR #4152, opened 2022-09-03
Add `excludes_trivial_init` for `missing_docs`

### LINKED ISSUE #4107: Option to disable missing_docs for trivial public init's
### New Issue Checklist

### New rule request

We often need trivial public init functions to allow an object to created. But this causes a missing_docs warning from SwiftLint unless we add doc comments with something generic like "Initializes xyz".

It would be nice if there was a way to disable missing_docs for public init functions with no parameters, such as excludes_trivial_init. This would let us keep missing_docs for the other situation.

Example where it should not trigger if excludes_trivial_init is set.
```
public init() {}
public init() {}
public init() {
    self.property1 = 123
}

/// Initializes with a name.
public init(name: String) {
    self.name = name
}
```

Example where it should still trigger even if excludes_trivial_init is set.
```
public init(name: String) {
    self.name = name
}
```

Yes, a configuration option, such as excludes_trivial_init, would be important.

Opt-in to avoid causing problems for existing code.

marcelofabri:
Fixes #4107

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,10 @@
 
 #### Enhancements
 
-* None.
+* Add new `excludes_trivial_init` configuration for `missing_docs` rule
+  to exclude initializers without any parameters.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [#4107](https://github.com/realm/SwiftLint/issues/4107)
 
 #### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


SimplyDanny (approved):
LGTM!

jpsim:
+1 and although you did the right thing naming this configuration parameter this way, I wonder how consistent we are across all rules about the tense of our verbs & keys (e.g. `exclude_trivial_init` vs `excludes_trivial_init`). Might not matter much in practice if it's inconsistent, or maybe it'll confuse someone. Probably not worth breaking anything now just in the name of consistency though.

marcelofabri:
> +1 and although you did the right thing naming this configuration parameter this way, I wonder how consistent we are across all rules about the tense of our verbs & keys (e.g. `exclude_trivial_init` vs `excludes_trivial_init`). Might not matter much in practice if it's inconsistent, or maybe it'll confuse someone. Probably not worth breaking anything now just in the name of consistency though.

Yeah, I noticed the same thing and was wondering what is our preference here

PR #4153, opened 2022-09-03
Exclude `weak_delegate` rule from autocorrection

### LINKED ISSUE #3577: Rule weak_delegate should not autocorrect
### New Issue Checklist

### Describe the bug

The [weak_delegate](https://realm.github.io/SwiftLint/weak_delegate.html) rule is unique in that it actually changes the behavior of the code by autocorrecting `weak` to delegates.  It would be unsafe to apply this rule to a codebase as objects may now released that had previously been retained, leading to undefined behavior and bugs.

This rule is still useful as a warning to developers, but autocorrect should be disabled.

SimplyDanny:
Fixes #3577.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 #### Breaking
 
-* None.
+* Exclude `weak_delegate` rule from autocorrection due to behavioral changes
+  leading to potential undefined behavior or bugs.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+  [#3577](https://github.com/realm/SwiftLint/issues/3577)
 
 #### Experimental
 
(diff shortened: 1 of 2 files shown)


SimplyDanny:
The build does not like me anymore. 😐 The TSan job is always failing with 
```
ERROR: Failed to query remote execution capabilities: UNAUTHENTICATED: no authentication metadata found
```
Anything I can do about it, @jpsim? It's the same for #4151 and #3989.

jpsim:
I wonder if this is because I need to flip a flag in Azure Pipelines to allow external repos to access these credentials 🤔

Looking into it.

jpsim:
Just checked this, let's see if that works.

image

jpsim:
I give up. I may need to disable the remote cache on Azure Pipelines completely.

jpsim:
ok I think it's working now.

SimplyDanny:
> Not great, but the rule is flawed to start with.

Indeed. Thank you for looking into the build issues!

PR #4154, opened 2022-09-04
Rewrite `for_where`, adding `allow_for_as_filter` config

### LINKED ISSUE #4040: Add configuration for `for_where` rule
### New Issue Checklist

----

While updating our code base to use SwiftLint 0.48.0, I realized `for_where` is now triggering for some cases where it wasn't before. It's great to fix false positives, but it made me realize that this rule might not be always wanted.

In several places of our app, we use an `if` inside a `for` to test if any object in a given array meets a condition. Some examples:

```swift
for clause in clauses {
    if clause.evaluate(object) {
        return true
    }
}
return false
```

```swift
for featureSet in minimalSupportedFeatureSets {
    if device.supportsFeatureSet(featureSet) {
        return true
    }
}

return false
```

While we could update them to use `where`, I kind of feel like it's against the spirit of the rule and it makes code harder to read:

```swift 
for clause in clauses where clause.evaluate(object) {
    return true
}
return false
```

Instead, the rule could have a configuration like `allow_single_return` to not trigger a violation if the only statement inside the `if` is a `return` statement.

marcelofabri:
Fixes #4040


jpsim:
What do you think about https://github.com/JohnCoates/Aerial/blob/97068951d2c67a56d42f705d405bd0578e45a672/Aerial/Source/Models/ManifestLoader.swift#L955:25 which was flagged by OSSCheck?

marcelofabri:
> What do you think about https://github.com/JohnCoates/Aerial/blob/97068951d2c67a56d42f705d405bd0578e45a672/Aerial/Source/Models/ManifestLoader.swift#L955:25 which was flagged by OSSCheck?

I could go either way honestly. I think complex `where` conditions are not great, so maybe not throwing a violation here and leaving it for the developer's judgement is better?

jpsim (approved):

PR #4155, opened 2022-09-04
Rewrite ForceUnwrappingRule using SwiftSyntax

### LINKED ISSUE #2915: force_unwrapping is not fired for trailing closures syntax
### New Issue Checklist

### Describe the bug

`force_unwrapping` is not fired for trailing closures syntax
```swift
let value = [1, 3, 5, 6].first { $0.isMultiple(of: 2) }!
```

Same case but with parentheses around works fine
```swift
let value = [1, 3, 5, 6].first({ $0.isMultiple(of: 2) })!
```

E.g. for this case `superfluous_disable_command` rule is fired
```swift
let value = [1, 3, 5, 6].first { $0.isMultiple(of: 2) }! // swiftlint:disable:this force_unwrapping
```

```bash
$ swiftlint lint

Superfluous Disable Command Violation: SwiftLint rule 'force_unwrapping' did not trigger a violation in the disabled region. Please remove the disable command. (superfluous_disable_command)
```

### Environment

0.35.0
CocoaPods
```yml
opt_in_rules:
  - force_unwrapping
```

no
Version 11.1 (11A1027)
* Do you have a sample that shows the issue? Run `echo "let value = [1, 3, 5, 6].first { $0.isMultiple(of: 2) }!" | swiftlint lint --no-cache --use-stdin --enable-all-rules`

```swift
// This triggers a violation:
let value = [1, 3, 5, 6].first { $0.isMultiple(of: 2) }!
```

marcelofabri:
Fixes #2915

### DIFF
--- a/Source/SwiftLintFramework/Models/Location.swift
+++ b/Source/SwiftLintFramework/Models/Location.swift
@@ -60,9 +60,9 @@ public struct Location: CustomStringConvertible, Comparable, Codable {
     /// - parameter position: The absolute position returned from SwiftSyntax.
     public init(file: SwiftLintFile, position: AbsolutePosition) {
         self.file = file.path
-        if let sourceLocation = file.locationConverter?.location(for: position) {
-            line = sourceLocation.line
-            character = sourceLocation.column
+        if let lineAndCharacter = file.stringView.lineAndCharacter(forByteOffset: ByteCount(position.utf8Offset)) {
+            line = lineAndCharacter.line
+            character = lineAndCharacter.character
         } else {
             line = nil
             character = nil
--- a/Source/SwiftLintFramework/Rules/Idiomatic/ForceUnwrappingRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/ForceUnwrappingRule.swift
@@ -54,6 +54,7 @@ public struct ForceUnwrappingRule: OptInRule, SwiftSyntaxRule, ConfigurationProv
             Example("dict[\"abc\"]↓!.bar(\"B\")"),
             Example("if dict[\"a\"]↓!↓!↓!↓! {}"),
             Example("var foo: [Bool]! = dict[\"abc\"]↓!"),
+            Example("realm.objects(SwiftUTF8Object.self).filter(\"%K == %@\", \"柱нǢкƱаم👍\", utf8TestString).first↓!"),
             Example("""
             context("abc") {
               var foo: [Bool]! = dict["abc"]↓!
--- a/Source/SwiftLintFramework/Rules/Idiomatic/ForceUnwrappingRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/ForceUnwrappingRule.swift
@@ -61,7 +61,9 @@ public struct ForceUnwrappingRule: OptInRule, SwiftSyntaxRule, ConfigurationProv
             }
             """),
             Example("open var computed: String { return foo.bar↓! }"),
-            Example("return self↓!")
+            Example("return self↓!"),
+            Example("[1, 3, 5, 6].first { $0.isMultiple(of: 2) }↓!"),
+            Example("map[\"a\"]↓!↓!")
         ]
     )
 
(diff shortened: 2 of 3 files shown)


Source/SwiftLintFramework/Models/Location.swift
```diff
@@ -60,9 +60,9 @@ public struct Location: CustomStringConvertible, Comparable, Codable {
     /// - parameter position: The absolute position returned from SwiftSyntax.
     public init(file: SwiftLintFile, position: AbsolutePosition) {
         self.file = file.path
-        if let sourceLocation = file.locationConverter?.location(for: position) {
-            line = sourceLocation.line
-            character = sourceLocation.column
+        if let lineAndCharacter = file.stringView.lineAndCharacter(forByteOffset: ByteCount(position.utf8Offset)) {
```
jpsim, line 63: @marcelofabri why this change? Did you benchmark it as being faster?

jpsim: And why not delegate to this init? https://github.com/realm/SwiftLint/blob/79347a17286476509ff79cf92f959e42597d4b56/Source/SwiftLintFramework/Models/Location.swift#L45

marcelofabri: I didn’t benchmark it, but the existing code was giving wrong column numbers for lines with UTF-8 characters. The added example was one case.

marcelofabri: (I should’ve delegated it to the other init though)

Source/SwiftLintFramework/Rules/Idiomatic/ForceUnwrappingRule.swift
```diff
@@ -39,151 +38,44 @@ public struct ForceUnwrappingRule: OptInRule, ConfigurationProviderRule {
...
-            return false
-        }
+private final class ForceUnwrappingVisitor: SyntaxVisitor, ViolationsSyntaxVisitor {
+    private(set) var violationPositions: [AbsolutePosition] = []
 
-        // range is type annotation of declaration
-        return true
+    override func visitPost(_ node: ForcedValueExprSyntax) {
+        violationPositions.append(node.exclamationMark.positionAfterSkippingLeadingTrivia)
     }
```
jpsim, line 80: chefs kiss gif

Source/SwiftLintFramework/Models/Location.swift
jpsim: Giving our caching of `StringView`s, it may actually be faster and more accurate than SwiftSyntax.

I'd encourage you to benchmark it in SwiftLint and to file a bug report towards SwiftSyntax if there's a correctness bug.

jpsim: > (I should’ve delegated it to the other init though)

Good opportunity for a follow-up PR 😄

marcelofabri: I’ll make that change and open an issue with SwiftSyntax later today 🙏

Source/SwiftLintFramework/Rules/Idiomatic/ForceUnwrappingRule.swift
SimplyDanny: Same reaction ... 😂

Source/SwiftLintFramework/Models/Location.swift
marcelofabri: I think we were using this wrong. From `SourceLocation`:

```
/// The UTF-8 byte offset from the beginning of the line where this location
  /// resides. 1-based.
  public var column: Int? {
    return compLoc?.column
  }
```

As far as I can tell, we expect `Location`'s `character` to be a UTF-16 offset, so this would explain the behavior I was seeing. Does this make sense?

jpsim: Hmmm, I expect it to be the logical character offset on the line so that Xcode renders it at the correct location.

PR #4156, opened 2022-09-04
Rewrite `large_tuple` rule with SwiftSyntax

marcelofabri:


marcelofabri:
I've checked the warnings from oss-check and it's a mix of violations that changed the position for variable declarations (I think the new behavior is how it should've been from the beginning) and fixing false negatives for things like `typealias` that were previously uncaught.

Source/SwiftLintFramework/Rules/Metrics/LargeTupleRule.swift
```diff
@@ -24,152 +14,29 @@ public struct LargeTupleRule: ASTRule, ConfigurationProviderRule {
...
-                let offset = byteRange.location + initialOffset
-                offsets.append((offset: offset, size: size))
-            }
-
-            let replacement = String(repeating: " ", count: substring.bridge().length)
-            text = text.replacingCharacters(in: range, with: replacement).bridge()
-        }
+    public func validate(file: SwiftLintFile) -> [StyleViolation] {
+        let visitor = LargeTupleRuleVisitor()
+        return visitor
```
jpsim, line 19: Nit: can use implicit return here

```suggestion
        LargeTupleRuleVisitor()
```

Up to you though, strictly a stylistic choice.

Source/SwiftLintFramework/Rules/Metrics/LargeTupleRule.swift
```diff
@@ -24,152 +14,29 @@ public struct LargeTupleRule: ASTRule, ConfigurationProviderRule {
...
     }
+}
 
-    private func containsReturnArrow(in text: String, range: NSRange) -> Bool {
-        let arrowRegex = SwiftVersion.current >= .fiveDotFive
-                        ? regex("\\A(?:\\s*async)?(?:\\s*throws)?\\s*->")
-                        : regex("\\A(?:\\s*throws)?\\s*->")
-        let start = NSMaxRange(range)
-        let restOfStringRange = NSRange(location: start, length: text.bridge().length - start)
+private final class LargeTupleRuleVisitor: SyntaxVisitor {
```
jpsim, line 36: Another great simplification thanks to SwiftSyntax!

By the way, I'd love to make improvements to SwiftSyntaxRule and the related SwiftSyntax convenience protocols so that this case could be covered. Let me know if you think of a way to do that!

jpsim (approved):

Source/SwiftLintFramework/Rules/Metrics/LargeTupleRule.swift
marcelofabri: Yeah, I'm sure as we see more rules being converted to SwiftSyntax we'll find more common patterns and opportunities to create abstractions.

PR #4157, opened 2022-09-04
Rewrite `trailing_closure` with SwiftSyntax

marcelofabri:


jpsim:
@marcelofabri I'm guessing you closed this because there were too many false positives?

marcelofabri:
> @marcelofabri I'm guessing you closed this because there were too many false positives?

Yeah, and catching them would be more complex than what was willing to handle 🫣

PR #4158, opened 2022-09-04
Rewrite void_return rule using SwiftSyntax

marcelofabri:


jpsim (approved):
Lovely

Source/SwiftLintFramework/Rules/Style/VoidReturnRule.swift
```diff
@@ -41,28 +40,83 @@ public struct VoidReturnRule: ConfigurationProviderRule, SubstitutionCorrectable
...
+            region.contains(node.positionAfterSkippingLeadingTrivia, locationConverter: locationConverter)
+        }
+
+        guard !isInDisabledRegion else {
+            return super.visit(node)
+        }
+
+        correctionPositions.append(tuple.positionAfterSkippingLeadingTrivia)
+
+        var returnType = SyntaxFactory.makeTypeIdentifier("Void")
```
jpsim, line 116: TIL about `SyntaxFactory`

jpsim:
@marcelofabri can we merge this?

marcelofabri:
> @marcelofabri can we merge this?

I need to think about what to do with these violations. My current thinking is that it's fine to have them fixed here and update `redundant_void_return` to catch them, wdyt?

jpsim:
Looking at the OSSCheck report, it looks like all the violations are safe because they're in closures with enough parent type context to infer the void return value. That's not the case with the current implementation of `redundant_void_return`.

ZevEisenberg:
+1 I can't speak to the rewrite, but this fixes a case that currently incorrectly triggers on main, and which might be good to add to the test cases:

```swift
func foo() -> () async -> Int
```

Failure on main:

    testWithDefaultConfiguration(): failed - nonTriggeringExample violated: 
    ```
    /* 👨‍👩‍👧‍👦👨‍👩‍👧‍👦👨‍👩‍👧‍👦 */
    func foo() -> () async -> Int
                  ^ warning: Void Return Violation: Prefer `-> Void` over `-> ()`. (void_return)
    ```

ZevEisenberg:
bump - I just rediscovered this thread because I dusted off an old project and found that this was still the case

lo1tuma:
Is there anything missing or something I could help with to get this PR merged?

SimplyDanny:
Rule has been rewritten in #5351.

ZevEisenberg:
I'm not at a computer right now so I can't test it, but do you know if it now correctly handles the case I mentioned [above](https://github.com/realm/SwiftLint/pull/4158#issuecomment-1250129148)?

SimplyDanny:
> I'm not at a computer right now so I can't test it, but do you know if it now correctly handles the case I mentioned [above](https://github.com/realm/SwiftLint/pull/4158#issuecomment-1250129148)?

Yes, the example is handled correctly now. See #5381.

PR #4159, opened 2022-09-04
Rewrite unneeded_parentheses_in_closure_argument with SwiftSyntax

marcelofabri:


marcelofabri:
i'll add a changelog entry before merging this

Source/SwiftLintFramework/Rules/Style/UnneededParenthesesInClosureArgumentRule.swift
```diff
@@ -60,75 +75,74 @@ public struct UnneededParenthesesInClosureArgumentRule: ConfigurationProviderRul
         ]
     )
 
-    public func validate(file: SwiftLintFile) -> [StyleViolation] {
-        return violationRanges(file: file).map {
-            StyleViolation(ruleDescription: Self.description,
-                           severity: configuration.severity,
-                           location: Location(file: file, characterOffset: $0.location))
+    public func makeVisitor(file: SwiftLintFile) -> ViolationsSyntaxVisitor? {
+        Visitor()
```
jpsim, line 79: I like this pattern to have nested Visitor/Rewriter types to simplify the naming for these visitors and rewriters.

Source/SwiftLintFramework/Rules/Style/UnneededParenthesesInClosureArgumentRule.swift
```diff
@@ -60,75 +75,74 @@ public struct UnneededParenthesesInClosureArgumentRule: ConfigurationProviderRul
...
+            return SyntaxFactory.makeClosureParam(
+                name: name,
+                trailingComma: isLast ? nil : SyntaxFactory.makeCommaToken(trailingTrivia: .spaces(1))
+            )
+        }
+
+        correctionPositions.append(clause.positionAfterSkippingLeadingTrivia)
+
+        let paramList = SyntaxFactory.makeClosureParamList(items).withTrailingTrivia(.spaces(1))
+        return Syntax(node.withInput(Syntax(paramList)))
```
jpsim, line 146: Should this be `return super.visit(...)`? I don't think the superclasses ever do anything here, so I think we can just always return in this func?

Source/SwiftLintFramework/Rules/Style/UnneededParenthesesInClosureArgumentRule.swift
```diff
@@ -60,75 +75,74 @@ public struct UnneededParenthesesInClosureArgumentRule: ConfigurationProviderRul
...
+        guard !isInDisabledRegion else {
+            return super.visit(node)
         }
+
+        let items = clause.parameterList.enumerated().compactMap { idx, param -> ClosureParamSyntax? in
+            guard let name = param.firstName else {
+                return nil
+            }
+
+            let isLast = idx == clause.parameterList.count - 1
```
jpsim, line 136: Nit:

```suggestion
            let isLast = idx == clause.parameterList.indices.last
```

jpsim (approved):

Source/SwiftLintFramework/Rules/Style/UnneededParenthesesInClosureArgumentRule.swift
marcelofabri:
```
Type '((DefaultIndices<FunctionParameterListSyntax>.Element) throws -> Bool) throws -> DefaultIndices<FunctionParameterListSyntax>.Element?' (aka '((SyntaxChildrenIndex) throws -> Bool) throws -> Optional<SyntaxChildrenIndex>') cannot conform to 'BinaryInteger'
```

😮

Source/SwiftLintFramework/Rules/Style/UnneededParenthesesInClosureArgumentRule.swift
marcelofabri: Calling the superclass is important so it does visit the node's children. In this case I don't think it matters, but decided to call `super.visit` for consistency

Source/SwiftLintFramework/Rules/Style/UnneededParenthesesInClosureArgumentRule.swift
jpsim: Ah not an array. Ok.

Source/SwiftLintFramework/Rules/Style/UnneededParenthesesInClosureArgumentRule.swift
jpsim: Oh I think I might need to fix some rules I’ve written based on this comment!

PR #4160, opened 2022-09-04
Rewrite `implicit_getter` rule with SwiftSyntax

marcelofabri:


Source/SwiftLintFramework/Rules/Style/ImplicitGetterRule.swift
```diff
@@ -16,134 +15,63 @@ public struct ImplicitGetterRule: ConfigurationProviderRule {
...
-
-                    return setDict.offset == dict.offset
-                }
-
-                guard !hasSetToken else {
-                    return nil
-                }
+        let visitor = ImplicitGetterRuleVisitor()
+
+        return visitor
```
jpsim, line 20: Nit: I like using implicit returns for these

```suggestion
        ImplicitGetterRuleVisitor()
```

jpsim (approved):

PR #4170, opened 2022-09-07
Rewrite `anyobject_protocol` with SwiftSyntax & deprecate

jpsim:


marcelofabri:
This is now a warning in the Swift compiler (since Swift 5.4 https://github.com/apple/swift/pull/34885):

```
file.swift:1:15: warning: using 'class' keyword to define a class-constrained protocol is deprecated; use 'AnyObject' instead
protocol Foo: class {}
              ^~~~~
              AnyObject
``` 

I wonder if we should just remove this rule?

jpsim:
> I wonder if we should just remove this rule?

interesting. I can make it opt-in, deprecate it for now and make it print a warning and fully remove it in a few releases.

PR #4177, opened 2022-09-07
Rewrite `empty_string` with SwiftSyntax

jpsim:


Source/SwiftLintFramework/Rules/Performance/EmptyStringRule.swift
```diff
@@ -17,23 +16,44 @@ public struct EmptyStringRule: ConfigurationProviderRule, OptInRule {
...
+}
+
+private extension EmptyStringRule {
+    final class Visitor: SyntaxVisitor, ViolationsSyntaxVisitor {
+        private(set) var violationPositions: [AbsolutePosition] = []
+
+        override func visitPost(_ node: StringLiteralExprSyntax) {
+            guard
+                // Empty string literal: `""`, `#""#`, etc.
+                node.segments.count == 1 && node.segments.first!.contentLength == .zero,
```
marcelofabri, line 39: nit: can we avoid the force unwrap here just in case this condition ever gets moved/copied etc?

Source/SwiftLintFramework/Rules/Performance/EmptyStringRule.swift
```diff
@@ -17,23 +16,44 @@ public struct EmptyStringRule: ConfigurationProviderRule, OptInRule {
...
+            violationPositions.append(violationPosition)
         }
     }
 }
+
+private extension TokenKind {
+    var isEqualityComparison: Bool {
+        self == .spacedBinaryOperator("==") ||
+            self == .spacedBinaryOperator("!=") ||
+            self == .unspacedBinaryOperator("==")
```
marcelofabri, line 57: isn't `unspacedBinaryOperator("!=")` possible here?

marcelofabri (approved):

Source/SwiftLintFramework/Rules/Performance/EmptyStringRule.swift
jpsim: No, it gets treated like a force unwrap

Source/SwiftLintFramework/Rules/Performance/EmptyStringRule.swift
jpsim: Sure

PR #4178, opened 2022-09-07
Rewrite `flatmap_over_map_reduce` with SwiftSyntax

jpsim:


marcelofabri (approved):

jpsim:
I’m not sure this one is correct since I’m no longer checking if there’s a `map` operation prior to the `reduce` but I can’t think of a counter example.

PR #4179, opened 2022-09-07
Rewrite `discouraged_optional_boolean` with SwiftSyntax

jpsim:


jpsim:
OSSCheck flagged some new false negatives that I'll address soon.

PR #4182, opened 2022-09-07
Rewrite `redundant_nil_coalescing` with SwiftSyntax

jpsim:


Source/SwiftLintFramework/Rules/Idiomatic/RedundantNilCoalescingRule.swift
```diff
@@ -16,30 +15,77 @@ public struct RedundantNilCoalescingRule: OptInRule, SubstitutionCorrectableRule
             Example("var myVar: Int?; myVar ?? 0\n")
         ],
         triggeringExamples: [
-            Example("var myVar: Int? = nil; myVar↓ ?? nil\n"),
-            Example("var myVar: Int? = nil; myVar↓??nil\n")
```
jpsim, line 20: This wasn't a valid nil coalescing operation

marcelofabri (approved):

jpsim:
I think we’re ok but I’ll double check the OSSCheck results before merging.

PR #4183, opened 2022-09-07
Re-delegate to ArrayInitRule

SimplyDanny:
Why not? 😄


jpsim:
> Why not?

Because of https://github.com/realm/SwiftLint/pull/4175 😅

It'd be great if you want to take a stab at making it compatible with that though.

SimplyDanny:
> > Why not?
> 
> Because of #4175 😅
> 
> It'd be great if you want to take a stab at making it compatible with that though.

Ahh ... I though it was already merged. Stupid me! 🤦‍♂️

PR #4185, opened 2022-09-07
Re-delegate to ArrayInitRule

SimplyDanny:
Next try ... 😉

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/ArrayInitRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/ArrayInitRule.swift
@@ -23,31 +23,31 @@ public struct ArrayInitRule: SwiftSyntaxRule, ConfigurationProviderRule, OptInRu
             Example("foo.map { /* a comment */ !$0 }\n")
         ],
         triggeringExamples: [
-            Example("↓foo.map({ $0 })\n"),
-            Example("↓foo.map { $0 }\n"),
-            Example("↓foo.map { return $0 }\n"),
+            Example("foo.↓map({ $0 })\n"),
+            Example("foo.↓map { $0 }\n"),
+            Example("foo.↓map { return $0 }\n"),
             Example("""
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Rules/Lint/ArrayInitRule.swift
```diff
@@ -69,7 +69,7 @@ extension ArrayInitRule {
                 return
             }
 
-            violationPositions.append(node.positionAfterSkippingLeadingTrivia)
+            violationPositions.append(memberAccess.name.positionAfterSkippingLeadingTrivia)
```
SimplyDanny, line 72: Changing the position where the violation triggers to the method made the delegating rule a bit easier to implement. Of course, this could be done differently having a separate method here for the type-safe variant. I think, the new position fits actually better, though.

jpsim: Fine by me.

SimplyDanny:
> Nicely done. Even better than before since the expensive operations are deferred only when there's a possible violation. 👏

Thank you!

Source/SwiftLintFramework/Rules/Lint/TypesafeArrayInitRule.swift
```diff
@@ -70,29 +69,33 @@ public struct TypesafeArrayInitRule: AnalyzerRule, ConfigurationProviderRule {
...
+    }
+
+    private func getOffset(in file: SwiftLintFile, at location: Location) -> ByteCount? {
+        guard let line = location.line, let offset = location.character else {
+            return nil
+        }
+        return file.stringView.byteOffset(forLine: Int64(line), bytePosition: Int64(offset))
     }
 
     private func buildIndex(for filePath: String, using compilerArguments: [String]) -> SourceKittenDictionary {
```
jpsim, line 101: I guess this is unused now

PR #4195, opened 2022-09-09
Add ability to filter rules for `generate-docs` subcommand

kattouf:
Added rules filtering options, like in the `rules` subcommand.

generate-docs--help

To achieve reuse with `rules` subcommand:
- extracted filtering logic into `RulesFilter` (and write tests)
- extracted filtering command line parameters into `RulesFilterOptions: ParsableArguments`


jpsim:
Thanks for the PR! I'll review shortly but in the meantime, here's some unused imports that CI flagged:

```
Source/swiftlint/Commands/Common/RulesFilterOptions.swift:2:1: error: Unused Import Violation: All imported modules should be required to make the file compile. (unused_import)
Source/swiftlint/Commands/Common/RulesFilter.ExcludingOptions+RulesFilterOptions.swift:1:1: error: Unused Import Violation: All imported modules should be required to make the file compile. (unused_import)
Source/swiftlint/Helpers/RulesFilter.swift:1:1: error: Unused Import Violation: All imported modules should be required to make the file compile. (unused_import)
```

kattouf:
Thanks for quick feedback, I'll fix CI issues and mark "Ready for review" again

jpsim:
That's why we don't have tests for the CLI target.

kattouf:
Hm, I think this will be fixed by adding flag inheritance to the definition of OTHER_SWIFT_FLAGS using $(inherited)

Doesn't works
```sh
 xcodebuild -scheme swiftlint test -destination "platform=macOS" OTHER_SWIFT_FLAGS='-D DISABLE_FOCUSED_EXAMPLES'
```
Works:
```sh
xcodebuild -scheme swiftlint test -destination "platform=macOS" OTHER_SWIFT_FLAGS='$(inherited) -D DISABLE_FOCUSED_EXAMPLES'
```

I will push this change to my branch to test this changes in CI

kattouf:
That helped ⬆️

jpsim:
Fantastic! To be honest I don't remember the exact reason why we didn't have CLI, it might have been something else that's since been resolved, or maybe I didn't configure things correctly. In any case, I'm glad you got it working here!

Package.swift
```diff
@@ -43,6 +43,12 @@ let package = Package(
                 "SwiftyTextTable",
             ]
         ),
+        .testTarget(
+          name: "swiftlintTests",
```
jpsim, line 47: Can this be renamed/moved to `CLITests`?

jpsim: If you could also mirror this in Bazel here that'd be appreciated: https://github.com/realm/SwiftLint/blob/main/Tests/BUILD

If you're not comfortable with Bazel don't worry about it, I can do this as a follow-up.

kattouf: Yes, I'm ok with this

jpsim, line 47: Nit: indent with 4 spaces for consistency with the rest of this file

jpsim:
Yeah so SwiftPM wasn't able to do this until recently: https://github.com/apple/swift-package-manager/issues/4639

azure-pipelines.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         sw_vers
         xcodebuild -version
       displayName: Version Informations
-    - script: xcodebuild -scheme swiftlint test -destination "platform=macOS" OTHER_SWIFT_FLAGS="-D DISABLE_FOCUSED_EXAMPLES"
+    - script: xcodebuild -scheme swiftlint test -destination "platform=macOS" OTHER_SWIFT_FLAGS="\$(inherited) -D DISABLE_FOCUSED_EXAMPLES"
```
jpsim, line 42: Nice!

Package.swift
kattouf: Sorry, I replied to first comment and didn't see the second about Basel.
I have no experience with Basel and would be grateful if you take this (:

jpsim:
Thanks for your contributions here, @kattouf! I'm especially happy to see tests for the CLI target for the first time ever.

PR #4196, opened 2022-09-09
Include the configured `bind_identifier` in `self_binding` violation

jpsim:
As suggested in https://github.com/realm/SwiftLint/pull/4146#issuecomment-1241793057


samrayner (approved):
Thank you so much!

PR #4198, opened 2022-09-10
Ignore comments starting right at the beginning of a line

### LINKED ISSUE #3430: indentation_width triggered when the previous line is commented at the beginning

SimplyDanny:
Might fix #3430. The comments used by Xcode only use `//` so there might not be the need to support block comments too. I think they are mainly what caused the issue. Thus, simply ignoring lines starting with `//` should be enough?


SimplyDanny:
The reported violations are not new. The must have already been there before the change.

PR #4203, opened 2022-09-13
Update SwiftSyntax & internal parser for Swift 5.7

jpsim:


jpsim:
Seems like the SwiftSyntax 5.7 release requires Swift 5.7 / Xcode 14 to compile.

PR #4206, opened 2022-09-14
Add new `if_let_shadowing` rule

### LINKED ISSUE #4202: if let from Swift 5.7
Suggestion to add a rule to help finding code that can be changed to use the short version of if let,

```
if let foo = foo {
  bar(foo)
}
```
to:
```
if let foo {
  bar(foo)
}
```

I've used this in my .yml which does the trick for my code, but I'm sure it can be improved to cover more cases:

```
  use_short_if_let:
    name: "Use short if let"
    regex: "if let (\\w*) = \\1(,| {)"
    capture_group: 0
    severity: warning
```

SimplyDanny:
Closes #4202.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,11 @@
   messages.  
   [JP Simard](https://github.com/jpims)
 
+* Add a new `if_let_shadowing` opt-in rule that triggers in Swift 5.7 when a
+  shadowing optional binding is created in an if- or guard-statement.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+  [#4202](https://github.com/realm/SwiftLint/issues/4202)
+
 #### Bug Fixes
 
 * Respect `validates_start_with_lowercase` option when linting function names.  
--- a/Source/SwiftLintFramework/Models/PrimaryRuleList.swift
+++ b/Source/SwiftLintFramework/Models/PrimaryRuleList.swift
@@ -81,6 +81,7 @@ public let primaryRuleList = RuleList(rules: [
     IBInspectableInExtensionRule.self,
     IdenticalOperandsRule.self,
     IdentifierNameRule.self,
+    IfLetShadowingRule.self,
     ImplicitGetterRule.self,
     ImplicitReturnRule.self,
     ImplicitlyUnwrappedOptionalRule.self,
--- a/Source/SwiftLintFramework/Models/SwiftVersion.swift
+++ b/Source/SwiftLintFramework/Models/SwiftVersion.swift
@@ -31,6 +31,8 @@ public extension SwiftVersion {
     static let fiveDotFive = SwiftVersion(rawValue: "5.5.0")
     /// Swift 5.6.x - https://swift.org/download/#swift-56
     static let fiveDotSix = SwiftVersion(rawValue: "5.6.0")
+    /// Swift 5.7.x - https://swift.org/download/#swift-57
+    static let fiveDotSeven = SwiftVersion(rawValue: "5.7.0")
 
     /// The current detected Swift compiler version, based on the currently accessible SourceKit version.
     ///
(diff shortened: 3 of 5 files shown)


marcelofabri:
Should this be opt-in?

SimplyDanny:
> Should this be opt-in?

You think so? In a Swift 5.7 code base, this would be a welcome rule and I feel like the false-positive rate should be low. This is no strong opinion, though. I'm also fine with making it optional.

marcelofabri:
> > Should this be opt-in?
> 
> You think so? In a Swift 5.7 code base, this would be a welcome rule and I feel like the false-positive rate should be low. This is no strong opinion, though. I'm also fine with making it optional.

IIRC, the evolution thread for that feature was somewhat controversial - this was even in the list of "things the Swift team won't consider proposals for".

I guess up to this day I'm still a little uncomfortable with making rules enabled by default as many developers consider SwiftLint the ultimate source of truth™ for styles vs using it as a tool and configuring it properly.

SimplyDanny:
Okay ... you convinced me. Let's make it optional.

Do you have ideas for more test cases? The ones I added are all quite simple. I cannot think of anything more elaborated, though.

marcelofabri:
Maybe it’s worth it temporarily removing the min Swift version so OSS check can run it

CraigSiemens:
> Do you have ideas for more test cases? The ones I added are all quite simple. I cannot think of anything more elaborated, though.

Here's some of the ones I've run into in our project that I think could be added.

Non triggering
```swift
if let i = i as? Foo
```

Triggering
```swift
if someCondition, ↓let j = j {
if someCondition,
   ↓let j = j {

// I'm not sure how multiple matches on a line work 🙃
if ↓let i = i, ↓let j = j {

if ↓let i = i,
   ↓let j = j {

// and the guard variants of those
```

marcelofabri:
Another one to watch out for: `where` conditions

SimplyDanny:
> Another one to watch out for: `where` conditions

Would you give me an example? I don't think optional bindings are allowed in where-clauses.

SimplyDanny:
> Maybe it’s worth it temporarily removing the min Swift version so OSS check can run it

Done. The samples I have checked look reasonable. Almost all are fairly standard.

CraigSiemens:
> > Another one to watch out for: `where` conditions
> 
> Would you give me an example, @marcelofabri? I don't think optional bindings are allowed in where-clauses.

I think he's referring to `where` and `repeat-where` loops that have conditions. I though where-clauses at first too 😄.

marcelofabri:
> > > Another one to watch out for: `where` conditions
> > 
> > 
> > Would you give me an example, @marcelofabri? I don't think optional bindings are allowed in where-clauses.
> 
> I think he's referring to `where` and `repeat-where` loops that have conditions. I thought where-clauses at first too 😄.

I was thinking about where clauses - I guess I didn’t sleep enough 😆

SimplyDanny:
> > > Another one to watch out for: `where` conditions
> > 
> > 
> > Would you give me an example, @marcelofabri? I don't think optional bindings are allowed in where-clauses.
> 
> I think he's referring to `where` and `repeat-where` loops that have conditions. I thought where-clauses at first too 😄.

I now think you are both referring to `while` and `repeat-while` loops actually. 🤔 If so, I've just added test cases for `while` loops at least. If you still mean something else, please let me know. 😅

Fun fact: The [Swift reference for `repeat-while` loops](https://docs.swift.org/swift-book/ReferenceManual/Statements.html#grammar_repeat-while-statement) states that the condition in the `while` can be an optional binding while the grammar rule expects expressions only. The grammar rule seem to be correct since this is also what SwiftSyntax and the Swift compiler accept. I'll create a fix for the documentation.

That being said, I'm confident now that all possible optional bindings are covered.

CraigSiemens:
> I now think you are both referring to `while` and `repeat-while` loops actually.

Yeah, I was thinking about `while` loops. Apparently I mix up words that look like `wh__e` 😆.

Otherwise, looks good to me.

eneko:
Will this rule be auto-correctable?

SimplyDanny:
> Will this rule be auto-correctable?

Yes, it will be.

CraigSiemens (approved):

Issue #4031, opened 2022-07-20
SwiftSyntax comment parsing error since 0.47.0+ on Linux builds


wasdkhan:
### Describe the bug

Swiftlint is ignoring disable comments since 0.47.0 and flagging warning and errors when it shouldn't.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint lint sample.swift
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.47.1
* Installation method used (Homebrew, CocoaPods, building from source, etc)? building from source on linux

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? no
  If so, paste their relative paths and respective contents.

### Example

```swift
// sample.swift
// swiftlint:disable colon
let noWarning :String = "" // No warning about colons immediately after variable names!
// swiftlint:enable colon
let hasWarning :String = "" // Warning generated about colons immediately after variable names
```
Expected: One warning
```
Linting Swift files at paths /home/wmkn/local/sample.swift
Linting 'sample.swift' (1/1)
/home/wmkn/local/sample.swift:5:5: warning: Colon Spacing Violation: Colons should be next to the identifier when specifying a type and next to the key in dictionary literals. (colon)
Done linting! Found 1 violation, 0 serious in 1 file.
```

Actual: Two warnings
```
Linting Swift files at paths /home/wmkn/local/sample.swift
Linting 'sample.swift' (1/1)
Could not parse the syntax tree for at least one file. Results may be invalid.
/home/wmkn/local/sample.swift:3:5: warning: Colon Spacing Violation: Colons should be next to the identifier when specifying a type and next to the key in dictionary literals. (colon)
/home/wmkn/local/sample.swift:5:5: warning: Colon Spacing Violation: Colons should be next to the identifier when specifying a type and next to the key in dictionary literals. (colon)
Done linting! Found 2 violations, 0 serious in 1 file.
```

wasdkhan:
> Would you please provide examples?

Updated my post of an example file where Swiftlint ignores the disable comments and flags all the lines as errors.

SimplyDanny:
You explicitly mention Linux. Did you try to the same on macOS?

jpsim:
> Could not parse the syntax tree for at least one file. Results may be invalid.

Where do you have the SwiftSyntax parser lib installed on your distribution of Linux?

[...]

PR #4216, opened 2022-09-19
Use SwiftSyntax's new SwiftParser

jpsim:
This new parser has a number of benefits for us:

1. Up to ~15% faster lint times on macOS (not measured on Linux)
2. Significantly smaller binaries: `portable_swiftlint.zip` is 11.3 MB down from 20.9 MB
3. No need to vendor the static `lib_InternalSwiftSyntaxParser` library and the various linker hacks needed to support that
4. Prevent Linux distribution errors like https://github.com/realm/SwiftLint/issues/4031
5. Clearly the future of SwiftSyntax considering how aggressively Apple is trying to remove the C++ parser: See https://github.com/apple/swift-syntax/pull/767
6. Possibly a lower barrier for new contributors given that it's 100% Swift

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -33,8 +33,8 @@
         "repositoryURL": "https://github.com/apple/swift-syntax.git",
         "state": {
           "branch": null,
-          "revision": "04d4497be6b88e524a71778d828790e9589ae1c4",
-          "version": "0.50700.0"
+          "revision": "093e5ee151d206454e2c1950d81333c4d4a4472e",
+          "version": null
         }
       },
       {
--- a/Source/SwiftLintFramework/Rules/Idiomatic/ToggleBoolRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/ToggleBoolRule.swift
@@ -1,5 +1,4 @@
 import SwiftSyntax
-import SwiftSyntaxBuilder
 
 public struct ToggleBoolRule: SwiftSyntaxCorrectableRule, ConfigurationProviderRule, OptInRule {
     public var configuration = SeverityConfiguration(.warning)
@@ -32,7 +31,7 @@ public struct ToggleBoolRule: SwiftSyntaxCorrectableRule, ConfigurationProviderR
     )
 
     public func makeVisitor(file: SwiftLintFile) -> ViolationsSyntaxVisitor? {
-        Visitor()
+        Visitor(viewMode: .sourceAccurate)
     }
 
     public func makeRewriter(file: SwiftLintFile) -> ViolationsSyntaxRewriter? {
@@ -81,19 +80,18 @@ private extension ToggleBoolRule {
 
             correctionPositions.append(node.positionAfterSkippingLeadingTrivia)
 
-            let functionCall = FunctionCallExprSyntax { functionCall in
-                functionCall.useCalledExpression(
-                    ExprSyntax(
-                        MemberAccessExprSyntax { memberAccess in
-                            memberAccess.useBase(node.first!.withoutTrivia())
-                            memberAccess.useDot(.period)
-                            memberAccess.useName(.identifier("toggle"))
-                        }
+            let functionCall = FunctionCallExprSyntax(
+                calledExpression: ExprSyntax(
+                    MemberAccessExprSyntax(
+                        base: node.first!.withoutTrivia(),
+                        dot: .periodToken(),
+                        name: .identifier("toggle"),
+                        declNameArguments: nil
                     )
-                )
-                functionCall.useLeftParen(.leftParen)
-                functionCall.useRightParen(.rightParen)
-            }
+                ),
+                leftParen: .leftParenToken(), argumentList: .init([]), rightParen: .rightParenToken(),
+                trailingClosure: nil, additionalTrailingClosures: nil
+            )
 
             let newNode = node
                 .replacing(childAt: 0, with: ExprSyntax(functionCall))
--- a/Source/SwiftLintFramework/Rules/Idiomatic/ToggleBoolRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/ToggleBoolRule.swift
@@ -1,4 +1,5 @@
 import SwiftSyntax
+import SwiftSyntaxBuilder
 
 public struct ToggleBoolRule: SwiftSyntaxCorrectableRule, ConfigurationProviderRule, OptInRule {
     public var configuration = SeverityConfiguration(.warning)
@@ -80,21 +81,8 @@ private extension ToggleBoolRule {
 
             correctionPositions.append(node.positionAfterSkippingLeadingTrivia)
 
-            let functionCall = FunctionCallExprSyntax(
-                calledExpression: ExprSyntax(
-                    MemberAccessExprSyntax(
-                        base: node.first!.withoutTrivia(),
-                        dot: .periodToken(),
-                        name: .identifier("toggle"),
-                        declNameArguments: nil
-                    )
-                ),
-                leftParen: .leftParenToken(), argumentList: .init([]), rightParen: .rightParenToken(),
-                trailingClosure: nil, additionalTrailingClosures: nil
-            )
-
             let newNode = node
-                .replacing(childAt: 0, with: ExprSyntax(functionCall))
+                .replacing(childAt: 0, with: "\(node.first!.withoutTrivia()).toggle()")
                 .removingLast()
                 .removingLast()
                 .withLeadingTrivia(node.leadingTrivia ?? .zero)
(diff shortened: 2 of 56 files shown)


Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
```diff
@@ -183,7 +183,7 @@ extension Configuration {
 
             switch rulesMode {
             case .allEnabled:
-                return
+                break
```
jpsim, line 186: I need to see if this is a rule that needs to be updated or a regression in SwiftSyntax.

jpsim:
Sadly OSSCheck reports that the new Swift parser is 31-69% slower than the previous C++ one 😢 

There are also some violations that were introduced and others that were removed compared to the 5.6 parser, which I’ll need to look into. These may be fixing or introducing false positives/negatives. It’s possible I’ll need to update either the rules or SwiftSyntax to fix the ones that are regressions.

jpsim:
There are new TSan races with this which I’ve reported in https://github.com/apple/swift-syntax/pull/807#issuecomment-1250700072

jpsim:
> Sadly OSSCheck reports that the new Swift parser is 31-69% slower than the previous C++ one 😢

Then again, linting Lyft's iOS monorepo locally is now faster (459 files/s vs 423 files/s before). I need to look more into the perf implications.

Source/SwiftLintFramework/Rules/Idiomatic/ToggleBoolRule.swift
```diff
@@ -81,19 +80,18 @@ private extension ToggleBoolRule {
...
 
-            let functionCall = FunctionCallExprSyntax { functionCall in
-                functionCall.useCalledExpression(
-                    ExprSyntax(
-                        MemberAccessExprSyntax { memberAccess in
-                            memberAccess.useBase(node.first!.withoutTrivia())
-                            memberAccess.useDot(.period)
-                            memberAccess.useName(.identifier("toggle"))
-                        }
+            let functionCall = FunctionCallExprSyntax(
```
DougGregor, line 83: You could probably do this with:

```swift
let functionCall: ExprSyntax = "\(node.first!.withoutTrivia()).toggle()"
```

Source/SwiftLintFramework/Rules/Idiomatic/VoidFunctionInTernaryConditionRule.swift
```diff
@@ -130,11 +130,26 @@ private class VoidFunctionInTernaryConditionVisitor: SyntaxVisitor, ViolationsSy
...
+
+    override func visitPost(_ node: UnresolvedTernaryExprSyntax) {
+        guard node.firstChoice.is(FunctionCallExprSyntax.self),
+              let parent = node.parent?.as(ExprListSyntax.self),
+              parent.last?.is(FunctionCallExprSyntax.self) == true,
+              !parent.containsAssignment,
+              let grandparent = parent.parent,
+              grandparent.is(SequenceExprSyntax.self),
+              let blockItem = grandparent.parent?.as(CodeBlockItemSyntax.self),
+              !blockItem.isImplicitReturn else {
```
DougGregor, line 142: One option you could consider would be to use the `SwiftOperators` library (https://github.com/apple/swift-syntax/blob/main/Sources/SwiftOperators/SwiftOperators.docc/SwiftOperators.md) to fold all of the sequence expressions into expression trees before you get here. Then you'll get a well-structured AST where you can check the parent of the ternary operator without guessing at its role in the sequence expression.

That said, `SwiftOperators` is new and less tested than the parser, so consider my suggestion a little bit aspirational at this point :)

Source/SwiftLintFramework/Rules/Idiomatic/ToggleBoolRule.swift
SimplyDanny: Nice one!

jpsim: Thanks for the suggestion, that's the same one I had demonstrated here: https://github.com/apple/swift-syntax/pull/782#issuecomment-1244064686

Source/SwiftLintFramework/Rules/Idiomatic/VoidFunctionInTernaryConditionRule.swift
jpsim: Yes, I believe that's being considered in other PRs by @marcelofabri

Package.resolved
```diff
@@ -24,17 +24,17 @@
...
-          "revision": "df9ee6676cd5b3bf5b330ec7568a5644f547201b",
-          "version": "1.1.3"
+          "revision": "9f39744e025c7d377987f30b03770805dcb0bcd1",
+          "version": "1.1.4"
         }
       },
       {
         "package": "SwiftSyntax",
-        "repositoryURL": "https://github.com/apple/swift-syntax.git",
+        "repositoryURL": "https://github.com/rintaro/swift-syntax.git",
```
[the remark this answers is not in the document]
jpsim: Yes, doing that in https://github.com/realm/SwiftLint/pull/4203

This PR is an exploration to update to `swift-syntax@main` and migrate to the Swift parser.

jpsim:
Turns out that when compiling SwiftLint with `bazel build -c opt swiftlint`, this change has a 30%-70% performance degradation.

However, when compiling with `swift build -c release`, this change has a 0%-8% performance improvement!

I'll be digging into the swiftc calls made by both build systems in order to identify where the discrepancies are.

Package.resolved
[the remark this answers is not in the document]
CodaFi: I've relaxed the version bounds on the argument parser (which we suspect might actually be a package manager dependency resolver bug, but that's a different story) which should unblock you here.

jpsim:
When I enable WMO with Bazel, I see the same improvement as with SwiftPM. I’ll be doing that shortly.

Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
jpsim: It's a regression: https://github.com/apple/swift-syntax/issues/851

I can work around it for now though.

jpsim:
There are 4 SwiftSyntax regressions that we're seeing here:

- `return_value_from_void_function` was reporting false positives because of https://github.com/apple/swift-syntax/issues/851. I have a workaround for now so this shouldn't block us.
- There's a file in the OSSCheck corpus that fails to parse properly, causing some false negatives in the `force_unwrapping` rule: https://github.com/apple/swift-syntax/issues/852
- There's another file in the OSSCheck corpus that fails to parse properly, causing some false negatives in the `syntactic_sugar` rule. I haven't filed an issue for this one yet.
- There appears to be quite a few regressions to how SwiftSyntax parses operators, causing false positives and false negatives in the `operator_usage_whitespace` rule. I haven't yet started filing issues for those.

Source/SwiftLintFramework/Extensions/Configuration+Parsing.swift
jpsim: Fixed in https://github.com/apple/swift-syntax/pull/842

jpsim:
Here's another source of false negatives: https://github.com/apple/swift-syntax/issues/860

jpsim:
With https://github.com/apple/swift-syntax/pull/869 we're down to just a single known regression, which affects only a single file and a single rule in OSSCheck:

[/KsApi/GraphAPI.swift:34:39: warning: Syntactic Sugar Violation: Shorthand syntactic sugar should be used, i.e. Int? instead of Optional<Int>. (syntactic_sugar)](https://github.com/kickstarter/ios-oss/blob/8af41af921b063961e87a3fc43911bac4d65daa1/KsApi/GraphAPI.swift#L34:39)

It's still not clear to me if it's the `syntactic_sugar` rule that needs to be updated, or if it's a regression in the SwiftSyntax parse tree.

Digging...

jpsim:
# 🎉 There are no more known regressions on either performance or correctness 🎉

Still requires https://github.com/apple/swift-syntax/pull/869 before we can point to `apple/swift-syntax@main`, but otherwise we're good!

---

That being said, I'm on the fence about merging this right away.

The fact that only a small proportion of SwiftLint rules today are using SwiftSyntax and that we were already hitting several regressions makes me concerned that as we migrate more rules, we'll uncover new regressions in the swift parser that wouldn't have been present in the 5.7 parser.

On the other hand, switching now has a number of other benefits:

1. Up to ~15% faster lint times on macOS (not measured on Linux)
2. Significantly smaller binaries: `portable_swiftlint.zip` is 11.3 MB down from 20.9 MB
3. No need to vendor the static `lib_InternalSwiftSyntaxParser` library and the various linker hacks needed to support that
4. Prevent Linux distribution errors like https://github.com/realm/SwiftLint/issues/4031
5. Clearly the future of SwiftSyntax considering how aggressively Apple is trying to remove the C++ parser: See https://github.com/apple/swift-syntax/pull/767
6. Possibly a lower barrier for new contributors given that it's 100% Swift

@SimplyDanny @marcelofabri what do you think?

SimplyDanny:
Fantastic! 🥳 I think you mentioned enough reasons to justify the merge. Uncovering regressions only helps to further improve the new parser.

PR #4218, opened 2022-09-19
Rewrite `dynamic_inline` with SwiftSyntax

jpsim:


Source/SwiftLintFramework/Rules/Lint/DynamicInlineRule.swift
```diff
@@ -25,47 +24,33 @@ public struct DynamicInlineRule: ASTRule, ConfigurationProviderRule {
...
+    public func makeVisitor(file: SwiftLintFile) -> ViolationsSyntaxVisitor? {
+        Visitor()
+    }
+}
+
+private extension DynamicInlineRule {
+    private final class Visitor: SyntaxVisitor, ViolationsSyntaxVisitor {
+        private(set) var violationPositions: [AbsolutePosition] = []
+
+        override func visitPost(_ node: FunctionDeclSyntax) {
```
marcelofabri, line 36: This won't cover subscripts, inits and getter/setters, which the rule currently covers (based on `functionKinds`, not sure if that actually works/is expected)

jpsim: Good catch! The example/test coverage for this rule is not great so it didn't catch any of that.

jpsim: I merged this by mistake without addressing your feedback. I'll be doing that in a new PR.

PR #4219, opened 2022-09-19
Rewrite `discouraged_object_literal` with SwiftSyntax

jpsim:


jpsim:
I'll add this to the changelog before cutting a release.

PR #4221, opened 2022-09-19
Rewrite `empty_enum_arguments` with SwiftSyntax

jpsim:


jpsim:
I'll add this to the changelog before cutting a release.

jpsim:
I need to fix this case: https://github.com/kickstarter/ios-oss/blob/28d69e7d6a13504c14ce4a6b00811fcd180e878d/Kickstarter-iOS/AppDelegateViewModel.swift#L383:42

but the other 4 reports from OSSCheck are improvements over the previous implementation. It happens when using labeled discarded arguments.

PR #4223, opened 2022-09-19
Rewrite `explicit_init` with SwiftSyntax

jpsim:


Source/SwiftLintFramework/Rules/Idiomatic/ExplicitInitRule.swift
```diff
@@ -164,52 +163,78 @@ public struct ExplicitInitRule: SubstitutionCorrectableASTRule, ConfigurationPro
...
+            }
 
-    public func violationRanges(in file: SwiftLintFile, kind: SwiftExpressionKind,
-                                dictionary: SourceKittenDictionary) -> [NSRange] {
-        guard kind == .call,
-              let name = dictionary.name,
-              let typeRange = findTypeRange(in: name),
-              let nameByteRange = dictionary.nameByteRange else {
-            return []
+            violationPositions.append(violationPosition)
```
SimplyDanny, line 195:
```suggestion
            violationPositions.append(calledBase.endPositionBeforeTrailingTrivia)
```
Should be possible, right?

jpsim: Yes, will do.

jpsim:
I've audited the OSSCheck results and the only changes are fixed false positives.

PR #4227, opened 2022-09-19
Rewrite `ibinspectable_in_extension` with SwiftSyntax

jpsim:


SimplyDanny:
To make the rule even faster especially in huge source files, traversing the AST could be stopped as soon as a class, struct or enum is encountered as only extension are relevant. This is done for example in `EmptyXCTestMethodRule`. I cannot tell, though, whether this really has a benefit in speed, but an advantage in huge ASTs is conceivable.

jpsim:
Yes, good point @SimplyDanny, I'll see if I can do that here.

PR #4230, opened 2022-09-20
Rewrite `fatal_error_message` rule with SwiftSyntax

marcelofabri:


Source/SwiftLintFramework/Rules/Idiomatic/FatalErrorMessageRule.swift
```diff
@@ -36,32 +36,38 @@ public struct FatalErrorMessageRule: ASTRule, ConfigurationProviderRule, OptInRu
...
+                return
+            }
+
+            violationPositions.append(node.positionAfterSkippingLeadingTrivia)
         }
+    }
+}
 
-        if bodyRange.length == 0 {
+private extension TupleExprElementListSyntax {
```
marcelofabri, line 60: at what point should we start extracting these extensions? this was copied from `XCTFailMessageRule`

PR #4214, opened 2022-09-18
Add `xctestcase_subclasses` option to `test_case_accessibility` rule

mildm8nnered:
Adds an `xctestcase_subclasses` option to the `test_case_accessibility` rules, partially resolving #4200 

For example, given

```
class MyTestCase: XCTestCase { }

class BrokenSwiftLintDemoTests: MyTestCase {
    func notATest() { }
}
```

```
test_case_accessibility:
  xctestcase_subclasses: ["MyTestCase"]
```

can be used so that `test_case_accessibility` can still work.

See https://github.com/realm/SwiftLint/issues/4200

Issue #4200, opened 2022-09-11
balanced_xctest_lifecycle, single_test_class, empty_xctest_method and test_case_accessibility do not work on when the parent class is a subclass of XCTestCase


mildm8nnered:
### New Issue Checklist

- [X ] Updated SwiftLint to the latest version
- [X ] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

balanced_xctest_lifecycle, single_test_class, empty_xctest_method and test_case_accessibility work fine for me on XCTestCase subclasses.

However if the parent class is a subclass of XCTestCase, then these rules do not fire.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ Pods/SwiftLint/swiftlint
```

```
Linting Swift files in current working directory
Linting 'AppDelegate.swift' (1/4)
Linting 'SwiftLintDemoTests.swift' (2/4)
Linting 'MyTestCase.swift' (3/4)
Linting 'BrokenSwiftLintDemoTests.swift' (4/4)
/Users/martin.redington/Documents/Hudl/Source/SwiftLintDemo/SwiftLintDemoTests/SwiftLintDemoTests.swift:11:7: warning: Balanced XCTest life-cycle Violation: Test classes must implement balanced setUp and tearDown methods. (balanced_xctest_lifecycle)
/Users/martin.redington/Documents/Hudl/Source/SwiftLintDemo/SwiftLintDemoTests/SwiftLintDemoTests.swift:16:5: warning: Empty XCTest Method Violation: Empty XCTest method should be avoided. (empty_xctest_method)
/Users/martin.redington/Documents/Hudl/Source/SwiftLintDemo/SwiftLintDemoTests/SwiftLintDemoTests.swift:11:1: warning: Single Test Class Violation: 2 test classes found in this file. (single_test_class)
/Users/martin.redington/Documents/Hudl/Source/SwiftLintDemo/SwiftLintDemoTests/SwiftLintDemoTests.swift:25:1: warning: Single Test Class Violation: 2 test classes found in this file. (single_test_class)
/Users/martin.redington/Documents/Hudl/Source/SwiftLintDemo/SwiftLintDemoTests/SwiftLintDemoTests.swift:20:5: warning: Test case accessibility Violation: Test cases should only contain private non-test members. (test_case_accessibility)
Done linting! Found 5 violations, 0 serious in 4 files.
```

[...]

PR #4233, opened 2022-09-21
Add `test_parent_classes` option to `single_test_class`, `balanced_xctest_lifecycle`, and `empty_xctest` rules

mildm8nnered:
Adds a `test_parent_classes` option to the `single_test_class`, `balanced_xctest_lifecycle`, and  `empty_xctest` rules, resolving the remaining rules mentioned in https://github.com/realm/SwiftLint/issues/4200.

Very similar to #4214 but see notes below.

For example, given

```
class MyTestCase: XCTestCase { }

class SwiftLintDemoTests: XCTestCase {
    // Unbalanced
        XCTFail("Hello")
    }

    // Empty
    func testExample() throws {

    }
}

// Second test class
class SwiftLintDemoTests2: XCTestCase {
    func testExample() throws {
        XCTFail("Hello")
    }
}
```

The following configurations will pick up the violations:

```
opt_in_rules:
  - balanced_xctest_lifecycle
  - single_test_class
  - empty_xctest_method

balanced_xctest_lifecycle:
  test_parent_classes: ["MyTestCase"]

single_test_class:
  test_parent_classes: ["MyTestCase"]

empty_xctest_method:
  test_parent_classes: ["MyTestCase"]
```

## [Notes](#notes)

There is a lot of redundancy in the configurations. `BalancedXCTestLifecycleConfiguration.swift` and `EmptyXCTestMethodConfiguration.swift` are identical, and `SingleTestClassConfiguration.swift` differs only in including `QuickSpec` as an additional default parent class.

I didn't roll anything up as yet, but obviously it would be trivial to either have an abstract parent class, or to have two of the configurations inherit from the other one, and to take the default parent classes as a constructor argument.

Also, I had to implement `makeViolation(file: SwiftLintFile, position: AbsolutePosition) -> StyleViolation` myself in `EmptyXCTestMethodRule.swift` because my rule's configuration was no longer a SeverityConfiguration - I wonder if there some way I could have avoided having to do that.

### DIFF
--- a/Source/SwiftLintFramework/Rules/RuleConfigurations/BalancedXCTestLifecycleConfiguration.swift
+++ b/Source/SwiftLintFramework/Rules/RuleConfigurations/BalancedXCTestLifecycleConfiguration.swift
@@ -1,0 +1,27 @@
+public struct BalancedXCTestLifecycleConfiguration: RuleConfiguration, Equatable {
+    public private(set) var severityConfiguration = SeverityConfiguration(.warning)
+    public private(set) var testParentClasses: Set<String> = ["XCTestCase"]
+    
+    public var consoleDescription: String {
+        return severityConfiguration.consoleDescription +
+            ", test_parent_classes: [\(testParentClasses)]"
+    }
+
+    public mutating func apply(configuration: Any) throws {
+        guard let configuration = configuration as? [String: Any] else {
+            throw ConfigurationError.unknownConfiguration
+        }
+
+        if let severityString = configuration["severity"] as? String {
+            try severityConfiguration.apply(configuration: severityString)
+        }
+
+        if let extraTestParentClasses = configuration["test_parent_classes"] as? [String] {
+            self.testParentClasses.formUnion(extraTestParentClasses)
+        }
+    }
+
+    public var severity: ViolationSeverity {
+        return severityConfiguration.severity
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,11 @@
   [Martin Redington](https://github.com/mildm8nnered)
   [#4200](https://github.com/realm/SwiftLint/issues/4200)
 
+* Add `test_parent_classes` option to `balanced_xctest_lifecycle`, `single_test_class`
+  and `empty_xctest_method` rules.  
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#4200](https://github.com/realm/SwiftLint/issues/4200)
+
 #### Bug Fixes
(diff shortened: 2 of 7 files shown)


Source/SwiftLintFramework/Rules/RuleConfigurations/BalancedXCTestLifecycleConfiguration.swift
```diff
@@ -0,0 +1,27 @@
+public struct BalancedXCTestLifecycleConfiguration: RuleConfiguration, Equatable {
+    public private(set) var severityConfiguration = SeverityConfiguration(.warning)
+    public private(set) var testParentClasses: Set<String> = ["XCTestCase"]
+    
+    public var consoleDescription: String {
+        return severityConfiguration.consoleDescription +
+            ", test_parent_classes: [\(testParentClasses)]"
```
SimplyDanny, line 7: The brackets are superfluous. Sorting the set makes the output deterministic:
```suggestion
            ", test_parent_classes: \(testParentClasses.sorted())"
```

SimplyDanny (commented):
> There is a lot of redundancy in the configurations. `BalancedXCTestLifecycleConfiguration` and `EmptyXCTestMethodConfiguration` are identical, and `SingleTestClassConfiguration` differs only in including `QuickSpec` as an additional default parent class.

`QuickSpec` also inherits from `XCTestCase` and thus should be in the default set of all of the configurations.

> I didn't roll anything up as yet, but obviously it would be trivial to either have an abstract parent class, or to have two of the configurations inherit from the other one, and to take the default parent classes as a constructor argument.

We could have one configuration struct used by all the rules.

> Also, I had to implement `makeViolation(file: SwiftLintFile, position: AbsolutePosition) -> StyleViolation` myself in `EmptyXCTestMethodRule` because my rule's configuration was no longer a `SeverityConfiguration` - I wonder if there some way I could have avoided having to do that.

Currently, there is no way to avoid it. Up to now, only a handful of rules require manual implementation.

Source/SwiftLintFramework/Rules/RuleConfigurations/BalancedXCTestLifecycleConfiguration.swift
mildm8nnered: So I added `QuickSpec` to all the defaults, and rolled the new ones up, with typealiases for the names you'd expect.

But I bodged rebasing, so I just moved to a new branch and PR -  #4262

mildm8nnered:
Replaced by #4262

PR #4237, opened 2022-09-24
Rewrite protocol_property_accessors_order with SwiftSyntax

marcelofabri:


Source/SwiftLintFramework/Rules/Style/ProtocolPropertyAccessorsOrderRule.swift
```diff
@@ -25,19 +24,75 @@ public struct ProtocolPropertyAccessorsOrderRule: ConfigurationProviderRule, Sub
...
+        }
+    }
+}
+
+private extension AccessorBlockSyntax {
+    var hasViolation: Bool {
+        guard accessors.count == 2,
+              accessors.allSatisfy({ $0.body == nil }),
+              let firstAccessor = accessors.first,
+              firstAccessor.accessorKind.tokenKind == .contextualKeyword("set") else {
```
SimplyDanny, line 92: `firstAccessor` could be inlined.

Source/SwiftLintFramework/Rules/Style/ProtocolPropertyAccessorsOrderRule.swift
```diff
@@ -25,19 +24,75 @@ public struct ProtocolPropertyAccessorsOrderRule: ConfigurationProviderRule, Sub
...
+                node.withAccessors(reversedAccessors)
+            )
+        }
+    }
+}
+
+private extension AccessorBlockSyntax {
+    var hasViolation: Bool {
+        guard accessors.count == 2,
+              accessors.allSatisfy({ $0.body == nil }),
```
SimplyDanny, line 90: What is the body of an accessor? 🤔

marcelofabri: an `AcessorDeclSyntax` is also used for computed properties, so having a body means it's not from a protocol. For example:

```swift
struct A {
  var b: Int {
    get { return 1 }
  }
}
```

SimplyDanny: I see. As only protocols are relevant for this rule, traversing the AST could be stopped as soon as a class, struct, enum or extension is encountered. This is done for example in `EmptyXCTestMethodRule`. This might also accelerate the rule a bit since processing can be stopped quite early. In addition, the check for existing bodies could be skipped.

PR #4238, opened 2022-09-24
Move method

SimplyDanny:


Source/SwiftLintFramework/Models/SwiftLintFile.swift
```diff
@@ -63,31 +63,13 @@ public final class SwiftLintFile {
...
-        }
-
-        return containsAttributesRequiringFoundation(dict: structureDictionary)
-    }
 }
 
 // MARK: - Hashable Conformance
 
-extension SwiftLintFile: Hashable {
+extension SwiftLintFile: Equatable, Hashable {
```
jpsim, line 70: FYI not a big deal but Hashable implies Equatable.

PR #4240, opened 2022-09-24
Pin Swift Docker image to 5.6.3-focal

jpsim:
Until I sort out some Docker 5.7 issues


jpsim:
This is meant to be temporary until I sort out some Docker 5.7 issues.

PR #4246, opened 2022-09-26
Explain where to get syntax kinds for custom rules from

### LINKED ISSUE #4239: Add `match_kinds` explanations with examples in documentations
### New Issue Checklist

```bash
$ swiftlint lint
```

### Environment

```yml
# insert yaml contents here
```

```swift
// This triggers a violation:
let foo = try! bar()
```

SimplyDanny:
Fixes #4239.


jpsim:
Thanks for documenting this!

PR #4247, opened 2022-09-26
Enable `if_let_shadowing` rule and fix all violations

SimplyDanny:


SimplyDanny:
Looks like the `duplicate_imports` rule is confused by the short optional binding syntax.

SimplyDanny:
> Looks like the `duplicate_imports` rule is confused by the short optional binding syntax.

SourceKit does not offer a node for bindings in `if` or `guard` statements - there is no `source.lang.swift.decl.var.local` and no `source.lang.swift.ref.var.*` to begin with.

That means the `duplicate_imports` rule is currently broken for Swift 5.7 code bases using the short optional binding syntax.

@jpsim, @marcelofabri: For your notice.

SimplyDanny:
Filed https://github.com/apple/swift/issues/61509 for the missing references.

SimplyDanny:
Now that Analyzer rules have been disabled in #4728, this rule can be enabled for the repository.

PR #4260, opened 2022-09-30
Generate Swift Syntax Dashboard in documentation

jpsim:
This is a useful overview of where the migration effort stands and what rules can be migrated next.

image


jpsim:
FYI @marcelofabri @SimplyDanny since you've been doing the bulk of the migrations

marcelofabri:
Nice! Can we group by default x opt-in rules as well? I’ve been trying to focus on riles where the impact would be greater

jpsim:
> Nice! Can we group by default x opt-in rules as well? I’ve been trying to focus on riles where the impact would be greater

thanks for the suggestion, I was considering doing that, so now you’ve convinced me!

jpsim:
Published here: https://realm.github.io/SwiftLint/swift-syntax-dashboard.html

PR #4263, opened 2022-10-03
Convert `duplicate_enum_cases` rule to SwiftSyntax

marcelofabri:


SimplyDanny (approved):

SimplyDanny:
Is this rule still required anyway? I don't know about the past, but now I see the compiler complaining already:
Screenshot 2022-10-03 at 12 28 53

marcelofabri:
Nice catch! I'll try to find out when this was fixed in the compiler and mark it as deprecated

marcelofabri:
I think you saw the warning because you had a `a` case without a label. Having only elements with labels don't trigger warnings even with Swift 5.7:

```swift
import Foundation

enum PictureImport {
    case add(int: Int)
    case add(data: Data)
}
```

Issue #4024, opened 2022-07-14
Rule Request: No magic number


amadeu01:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

Please describe the rule idea, format
this issue's title as `Rule Request: [Rule Name]` and describe:

1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.
https://eslint.org/docs/latest/rules/no-magic-numbers

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

Basically, when you use a number without giving it some context.

```swift

Circle()
            .fill(Color.primary.opacity(isAnimate ? 0.1 : 1.0 )) // magic numbers
            .frame(
              width: 10, // magic number
              height: 10
            )
```
4. Should the rule be configurable, if so what parameters should be configurable?
No

6. Should the rule be opt-in or enabled by default? Why?

It should be optional

douglasdrumz:
This is a good idea. One small comment, I think you should edit the title to `Rule Request: No magic number` to comply with the template

amadeu01:
> This is a good idea. One small comment, I think you should edit the title to `Rule Request: No magic number` to comply with the template

Done! Thanks

douglasdrumz:
About item 4: what do you think of having a configuration to allow some numbers? 0 and 1 are so common and understandable that this rule without configuration may trigger false positives.
Examples: 
- It's common to do `array[0]` to access the first element. Without this option, one would need to define a `let firstIndex = 0` just to do `array[firstIndex]`. Less readable than just using `array[0]`.
- In the case of 1, it's common to have some operation plus or minus 1 to avoid off-by-one errors. Having to define another constant may make the code more unreadable.

Using that link you provided as an example, eslint accepts an array of ignored numbers. That way, you can have a rule `no-magic-numbers ignore:[0,1]`

SimplyDanny:
> About item 4: what do you think of having a configuration to allow some numbers? 0 and 1 are so common and understandable that this rule without configuration may trigger false positives.

[...]

PR #4265, opened 2022-10-03
Add No Magic Numbers rule

thisIsTheFoxe:
Hi everyone,

I've been using SwiftLint for a while, great project! Now in the Hacktoberfest season I got the motivation to have a little look how it actually works..

Hope everything checks out.. 
Let me know what you think :)

Closes:
-  #4024 
<hr/>

## TL;DR: Number literals should only be used in variable declarations

I'm also trying to consider what was said in the issue (i.e. ignoring the numbers 1,0,0.0,and 1.0, and also -1 and -1.0). I also didn't implement a configuration to exclude other numbers. This might be something to look into in the future..
From Danger warnings bellow it seems like the most common issues are related to `@/#available` and `return/case`. Another issue could be `let myInt = Int16(10)`. I'm not entirely sure about all those and ideally they should probably be customisable in the config, but for now I felt like only `available` should be ignored, because it's very natural and overly common to use numbers to refer to OS versions.
I'm also not entirely sure about why the perforce is so bad. Trying it with regex seems to be even worse if you got any ideas lmk..


thisIsTheFoxe:
cc @amadeu01 @SimplyDanny wdyt? lmk if you got any ideas or if I'm missing sth :)

SimplyDanny:
First of all: Thank you for the contribution!

Let me give you some general advice:
* There are currently two approaches on how rules are written in SwiftLint. The first uses structural source information provided by [SourceKit](https://github.com/apple/swift/tree/main/tools/SourceKit). This is the one you picked, probably because most of the existing rules are still based on SourceKit. The information we get from SourceKit mainly has a structural nature and was the best available until some months ago. The new approach uses [SwiftSyntax](https://github.com/apple/swift-syntax) which makes a complete AST of the source code available to a rule. If you have a look in the recent commits you will find a lot of rules being rewritten to be based on SwiftSyntax. I recommend to base your rule on SwiftSyntax too. I also think the implementation will be more fun to do for you. 😉 
* I guess your implementation is very slow because it iterates over every token in every line in every analyzed source file. This needs to be improved and it likely will be automatically by working with the file's AST.
* The dedicated test class `NoMagicNumbersRuleTests` is not needed. Just inline the examples in the rule's description. All examples will be tested automatically.

thisIsTheFoxe:
@SimplyDanny there we go, thank you so much for the help :)

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -0,0 +1,92 @@
+import SwiftSyntax
+
+public struct NoMagicNumbersRule: SwiftSyntaxRule, OptInRule, ConfigurationProviderRule {
+    public init() {}
+
+    public var configuration = SeverityConfiguration(.warning)
+
+    private static let functionExample = Example("""
```
SimplyDanny, line 8: You can just inline these examples. It's better to have all examples in one sequence.

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -0,0 +1,92 @@
...
+    if #available(iOS 13, OSX 10.10, *) {
+        return
+    }
+}
+""")
+
+    public static let description = RuleDescription(
+        identifier: "no_magic_numbers",
+        name: "No Magic Numbers",
+        description: "‘Magic numbers’ should be replaced by named constants.",
```
SimplyDanny, line 35: Single quotes are often used to highlight code. I think we can go without them.

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -0,0 +1,92 @@
...
+
+private extension TokenSyntax {
+    var isMagicNumber: Bool {
+        let numerStr = withoutTrivia().text.replacingOccurrences(of: "_", with: "")
+
+        guard let number = Double(numerStr),
+              ![0, 1, -1].contains(number),
+              let parentToken = parent?.parent,
+              !parentToken.is(InitializerClauseSyntax.self),
+              !parentToken.is(CodeBlockItemSyntax.self) else {
```
SimplyDanny, line 87: This check seems unnecessary. A free-standing number is anyway a compiler warning.

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -0,0 +1,92 @@
...
+}
+
+private extension TokenSyntax {
+    var isMagicNumber: Bool {
+        let numerStr = withoutTrivia().text.replacingOccurrences(of: "_", with: "")
+
+        guard let number = Double(numerStr),
+              ![0, 1, -1].contains(number),
+              let parentToken = parent?.parent,
+              !parentToken.is(InitializerClauseSyntax.self),
```
SimplyDanny, line 86: I think this rule should also trigger on initializer expressions that aren't simple number expressions. That is, `let a = b + 1.0` should trigger while `let a = 7` is accepted. What do you think?

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+}
+
+private extension TokenSyntax {
+    var isMagicNumber: Bool {
+        let numerStr = withoutTrivia().text.replacingOccurrences(of: "_", with: "")
+
+        guard let number = Double(numerStr),
+              ![0, 1, -1].contains(number),
```
SimplyDanny, line 84: This list could also contain `1.0` and `0.0`. Also, `-1` is not needed since you are only checking number literals which do not comprise the prefix operator.

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -0,0 +1,92 @@
...
+            if node.digits.isMagicNumber {
+                self.violations.append(node.digits.positionAfterSkippingLeadingTrivia)
+            }
+        }
+    }
+}
+
+private extension TokenSyntax {
+    var isMagicNumber: Bool {
+        let numerStr = withoutTrivia().text.replacingOccurrences(of: "_", with: "")
```
SimplyDanny, line 81: No need to do this as long as this rule is only interested in a defined set of literals.

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -0,0 +1,92 @@
...
+            functionExample,
+            classExample,
+            availableExample
+        ],
+        triggeringExamples: [
+            Example("foo(123)"),
+            Example("bar(1_000.005_01)"),
+            Example("array[42]"),
+            Example("let box = array[12 + 14]"),
+            Example("Color.primary.opacity(isAnimate ? 0.1 : 1.5)")
```
SimplyDanny, line 54: Please also add violation markers `↓` to the places in these examples where the rule is supposed to trigger.

SimplyDanny (changes requested):
Well done! This is a lot clearer. Please have a look into my remarks.

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
thisIsTheFoxe: Yup, was just thinking in terms or playgrounds.. but who uses lint for playgrounds anyways? hehe :D

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
thisIsTheFoxe: Yes, that's the idea as you mentioned in the issue. Since those numbers (floating point and int) are used p often they could be excluded by default..

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
thisIsTheFoxe: Well, yes, technically. Just that imo `1.0` should be excluded by default. But `let a = b + 2.0` should and afaik, does trigger. I'll add a test case for that.

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -0,0 +1,85 @@
...
+            }
+        }
+        """)
+        ],
+        triggeringExamples: [
+            Example("foo(↓321)"),
+            Example("bar(↓1_000.005_01)"),
+            Example("array[↓42]"),
+            Example("let box = array[↓12 + ↓14]"),
+            Example("let a = b + 2.0"),
```
SimplyDanny, line 47:
```suggestion
            Example("let a = b + ↓2.0"),
```
And could you also add an example like `let a = b + 1.0` to the non-triggering section?

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+}
+
+private extension TokenSyntax {
+    var isMagicNumber: Bool {
+        let numerStr = text.replacingOccurrences(of: "_", with: "")
+
+        guard let number = Double(numerStr),
+              ![0, 1].contains(number),
```
SimplyDanny, line 78: What about `1.0` and `0.0`? Or is it that `Double("1.0")` returns `1`?

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -0,0 +1,85 @@
...
+              let parentToken = parent?.parent,
+              !parentToken.is(InitializerClauseSyntax.self) else {
```
SimplyDanny, line 80: Are you still thinking about excluding annotations like `@available` here too, and, as you suggested, things like `let i = Int64(10)`?

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
thisIsTheFoxe: Yes, `Double("1.0") == 1` ;)
[Try it online!](https://tio.run/##Ky7PTCsx@f8/M7cgv6hEwS2/NC8lsSQzP4@roCgzr0TDJb80KSdVQ8lQz0BJU8HWVsFQ8/9/AA "Swift – Try It Online")

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
thisIsTheFoxe: `@available` is apparently being ignored (see also test case).
With the other one I'm a bit unsure bc of how specific it is.. also, for cases like that `let i: Int64 = 10` could theoretically be used to avoid the warning. wdyt?

SimplyDanny: I see. The numbers in the `@available` attributes are special syntax elements. That's why they are automatically excluded.

Regarding the second part, you are right in that it's not obvious which case must be allowed and which should be ignored. Let's keep it how it is for now and come back to it when there are specific requirements.

SimplyDanny (approved):

SimplyDanny:
Thanks again for your work, @thisIsTheFoxe!

amadeu01:
thanks @thisIsTheFoxe

PR #4269, opened 2022-10-03
Migrate `switch_case_on_newline` to SwiftSyntax

marcelofabri:


marcelofabri:
I think the Sourcery violation that was fixed might be because that file contains invalid Swift code (see line 34: https://github.com/krzysztofzablocki/Sourcery/blob/b9b5981d87a38783846c60d98a58be4616e0a61e/Templates/Tests/Generated/AutoEquatable.generated.swift#L34)

SimplyDanny (approved):

PR #4270, opened 2022-10-03
Migrate `no_extension_access_modifier` to SwiftSyntax

marcelofabri:


SimplyDanny (approved):
Traversing the AST could be stopped whenever a struct, class, protocol or enum is encountered.

PR #4271, opened 2022-10-03
Migrate `static_operator` rule to SwiftSyntax

marcelofabri:


Source/SwiftLintFramework/Rules/Idiomatic/StaticOperatorRule.swift
```diff
@@ -79,23 +78,34 @@ public struct StaticOperatorRule: ASTRule, ConfigurationProviderRule, OptInRule
...
+}
 
-        return [
-            StyleViolation(ruleDescription: Self.description,
-                           severity: configuration.severity,
-                           location: Location(file: file, byteOffset: offset))
-        ]
+private extension FunctionDeclSyntax {
+    var isFreeFunction: Bool {
+        parent?.is(CodeBlockItemSyntax.self) ?? false
```
SimplyDanny, line 100: My usual recommendation. 😄

Instead of checking for the parent subsequently, prefer to skip function declarations contained in struct, class, ... declarations by stopping AST traversal right when they are encountered.

Source/SwiftLintFramework/Rules/Idiomatic/StaticOperatorRule.swift
```diff
@@ -79,23 +78,34 @@ public struct StaticOperatorRule: ASTRule, ConfigurationProviderRule, OptInRule
...
+    }
+
+    var isOperator: Bool {
+        switch identifier.tokenKind {
+        case .spacedBinaryOperator:
+            return true
+        default:
+            return false
+        }
```
SimplyDanny, line 109:
```suggestion
        identifier.tokenKind == .spacedBinaryOperator
```

Source/SwiftLintFramework/Rules/Idiomatic/StaticOperatorRule.swift
marcelofabri: my concern is that we might miss one of these cases - we can do it as an optimization, but I wouldn't want to rely on that

Source/SwiftLintFramework/Rules/Idiomatic/StaticOperatorRule.swift
marcelofabri: so this enum actually has an associated value, so we can't use an equality comparison. We could use `if case let` but I personally don't love that syntax.

SimplyDanny (approved):

PR #4275, opened 2022-10-03
[oss-check] Add DuckDuckGo and Signal

marcelofabri:
As we're rewriting a lot of rules recently, having more repos in oss-check can help us catch regressions earlier.


SimplyDanny (approved):

Dangerfile
```diff
@@ -14,7 +14,7 @@ modified_files = git.modified_files + git.added_files
 # including in a CHANGELOG for example
 has_app_changes = !modified_files.grep(/Source/).empty?
 has_test_changes = !modified_files.grep(/Tests/).empty?
-has_danger_changes = !modified_files.grep(/Dangerfile|script\/oss-check|Gemfile/).empty?
+has_danger_changes = !modified_files.grep(/Dangerfile|tools\/oss-check|Gemfile/).empty?
```
jpsim, line 17: Thanks!

Source/swiftlint/Commands/Analyze.swift
```diff
@@ -13,7 +13,7 @@ extension SwiftLint {
         var quiet = false
         @Option(help: "The path of the full xcodebuild log to use when running AnalyzerRules.")
         var compilerLogPath: String?
-        @Option(help: "The path of a compilation database to use when running AnalyzerRules.")
+        @Option(help: "The path of a compilation database to use when running AnalyzerRules")
```
jpsim, line 16: Why this change? Don't most help descriptions end with a period?

tools/oss-check
```diff
@@ -290,12 +290,14 @@ end
 @repos = [
   Repo.new('Aerial', 'JohnCoates/Aerial'),
   Repo.new('Alamofire', 'Alamofire/Alamofire'),
+  Repo.new('DuckDuckGo', 'duckduckgo/iOS'),
   Repo.new('Firefox', 'mozilla-mobile/firefox-ios'),
   Repo.new('Kickstarter', 'kickstarter/ios-oss'),
   Repo.new('Moya', 'Moya/Moya'),
   Repo.new('Nimble', 'Quick/Nimble'),
   Repo.new('Quick', 'Quick/Quick'),
   Repo.new('Realm', 'realm/realm-cocoa'),
```
jpsim, line 299: While we're in here we could update this repo to point to `realm-swift`:

```suggestion
  Repo.new('Realm', 'realm/realm-swift'),
```

I think it was renamed a few years ago.

jpsim (approved):
Unfortunately this will make OSSCheck take even longer, and it's already one of the longest CI jobs, but I think it's a worthwhile tradeoff.

I think cloning the repos is what takes the longest time, which we could speed up by somehow caching them on the CI machines, but that's also tricky.

We could have a CI job that runs on a cron job daily that updates the local cache that we clone from.

I'm open to suggestions.

Source/swiftlint/Commands/Analyze.swift
marcelofabri: this was just to trigger sos-check, otherwise it'd skip because both binaries would have the same signature. I'll revert it before merging

marcelofabri:
With the bazel cache, oss-check is now faster than analyze from what I've seen, so we have a little room here to make it slower. 

I wonder if downloading a zip from GitHub would be faster (ie https://github.com/realm/SwiftLint/archive/refs/heads/main.zip) than cloning the repo.

Source/swiftlint/Commands/Analyze.swift
jpsim: Should work again now that you've fixed `OSSCheck` to run when `tools/oss-check` changes.

marcelofabri: it didn't because of this check: https://github.com/realm/SwiftLint/blob/main/tools/oss-check#L324-L329

jpsim: Ah right.

jpsim:
> I wonder if downloading a zip from GitHub (ie https://github.com/realm/SwiftLint/archive/refs/heads/main.zip) would be faster than cloning the repo.

Wasn't faster in my testing here: https://github.com/realm/SwiftLint/pull/4106

jpsim:
> Wasn't faster in my testing here: #4106

Maybe I was doing something wrong, the tarballs were coming back empty and I don't remember if I actually got it working. You could check if you want.

PR #4278, opened 2022-10-04
Migrate `implicitly_unwrapped_optional` to SwiftSyntax

marcelofabri:


Source/SwiftLintFramework/Rules/RuleConfigurations/ImplicitlyUnwrappedOptionalConfiguration.swift
```diff
@@ -13,17 +13,17 @@ public enum ImplicitlyUnwrappedOptionalModeConfiguration: String {
     }
 }
 
-public struct ImplicitlyUnwrappedOptionalConfiguration: RuleConfiguration, Equatable {
-    private(set) var severity: SeverityConfiguration
-    private(set) var mode: ImplicitlyUnwrappedOptionalModeConfiguration
+public struct ImplicitlyUnwrappedOptionalConfiguration: RuleConfiguration, Equatable, SeverityBasedRuleConfiguration {
```
SimplyDanny, line 16:
```suggestion
public struct ImplicitlyUnwrappedOptionalConfiguration: SeverityBasedRuleConfiguration, Equatable {
```
Or would you really like to keep both? In this case, `SeverityBasedRuleConfiguration` could be renamed into `SeverityConfigurationProvider` or something similar.

SimplyDanny (approved):

marcelofabri:
Even though there are a lot of oss-check reports, I think the implementation here is simply enough to say that they're just due to the violation position changing.

PR #4281, opened 2022-10-04
Migrate `explicit_enum_raw_value` rule to SwiftSyntax

marcelofabri:


marcelofabri:
Those are real issues that I need to fix

PR #4282, opened 2022-10-04
Replace mutating for-in with `reduce(into:)`

jpsim:


jpsim:
Not a meaningful change, just stylistic.

PR #4284, opened 2022-10-04
Add column for SourceKit usage to `rules` command

jpsim:


Source/swiftlint/Commands/Rules.swift
```diff
@@ -109,14 +109,15 @@ private extension TextTable {
...
+            TextTableColumn(header: "uses sourcekit"),
             TextTableColumn(header: "configuration")
         ]
         self.init(columns: columns)
         let sortedRules = ruleList.list.sorted { $0.0 < $1.0 }
         func truncate(_ string: String) -> String {
             let stringWithNoNewlines = string.replacingOccurrences(of: "\n", with: "\\n")
             let minWidth = "configuration".count - "...".count
-            let configurationStartColumn = 124
+            let configurationStartColumn = 140
```
jpsim, line 120: I'm almost certain this value was wrong before, and is wrong now. To be correct I think we need to derive it programmatically.

PR #4293, opened 2022-10-05
Make `private_unit_test` rule correctable

SimplyDanny:


Source/SwiftLintFramework/Rules/Lint/PrivateUnitTestRule.swift
```diff
@@ -106,12 +106,48 @@ public struct PrivateUnitTestRule: SwiftSyntaxRule, ConfigurationProviderRule, C
                 private ↓func test4() {}
             }
             """)
+        ],
+        corrections: [
+            Example("""
+
+                private class Test: XCTestCase {}
```
jpsim, line 113: Are these empty newlines in these examples intentional?

Source/SwiftLintFramework/Rules/Lint/PrivateUnitTestRule.swift
```diff
@@ -106,12 +106,48 @@ public struct PrivateUnitTestRule: SwiftSyntaxRule, ConfigurationProviderRule, C
...
+                """): Example("""
+
+                    class Test: XCTestCase {}
+                    """),
+            Example("""
+                class Test: XCTestCase {
+
+                    private func test1() {}
```
jpsim, line 121: Should these examples have violation markers?

Source/SwiftLintFramework/Rules/Lint/PrivateUnitTestRule.swift
```diff
@@ -156,6 +192,77 @@ private class Visitor: SyntaxVisitor, ViolationsSyntaxVisitor {
...
+    let disabledRegions: [SourceRange]
+    private let parentClassRegex: NSRegularExpression
+
+    init(parentClassRegex: NSRegularExpression,
+         locationConverter: SourceLocationConverter,
+         disabledRegions: [SourceRange]) {
+        self.locationConverter = locationConverter
+        self.disabledRegions = disabledRegions
+        self.parentClassRegex = parentClassRegex
+    }
```
jpsim, line 207: Nit: keep the same order as the parameters?

```suggestion
        self.parentClassRegex = parentClassRegex
        self.locationConverter = locationConverter
        self.disabledRegions = disabledRegions
    }
```

Source/SwiftLintFramework/Rules/Lint/PrivateUnitTestRule.swift
```diff
@@ -156,6 +192,77 @@ private class Visitor: SyntaxVisitor, ViolationsSyntaxVisitor {
...
+
+    override func visit(_ node: ClassDeclSyntax) -> DeclSyntax {
+        guard node.hasParent(matching: parentClassRegex), node.isPrivate else {
```
jpsim, line 210: Nit: reverse condition order since I expect `node.isPrivate` to be cheaper than matching the parent class regex.

```suggestion
        guard node.isPrivate, node.hasParent(matching: parentClassRegex) else {
```

jpsim (approved):

Source/SwiftLintFramework/Rules/Lint/PrivateUnitTestRule.swift
SimplyDanny: Yes. In the example
```swift
class Test {
    private fund test() {}
}
```
the `private` token has leading trivia that must be preserved when removing it by attaching the trivia to the next token. This is what the unpleasant `withoutPrivate(modifiers:declKeyword:)` method does. The empty lines in the tests make sure this works.

If there is a better way to deal with the whitespace, please let me know.

Source/SwiftLintFramework/Rules/Lint/PrivateUnitTestRule.swift
SimplyDanny: Yes. They now have.

Source/SwiftLintFramework/Rules/Lint/PrivateUnitTestRule.swift
SimplyDanny: Good point!

PR #4296, opened 2022-10-05
Rewrite `trailing_semicolon` with SwiftSyntax

jpsim:
I'm open to suggestions for better ways to remove nodes in a rewriter.


jpsim:
All newly introduced violations are legitimate and were missed with the previous implementation.

PR #4310, opened 2022-10-06
Provide syntax visitor base class allowing for convenient skipping of declaration nodes

SimplyDanny:


SimplyDanny:
The single fixed violation is valid. All the violations in Signal seem to be unrelated.

jpsim:
I've removed Signal from OSSCheck because of these weird issues: https://github.com/realm/SwiftLint/pull/4313

Source/SwiftLintFramework/Protocols/SwiftSyntaxRule.swift
```diff
@@ -58,7 +58,53 @@ public extension SwiftSyntaxRule {
 }
 
 /// A SwiftSyntax `SyntaxVisitor` that produces absolute positions where violations should be reported.
-public protocol ViolationsSyntaxVisitor: SyntaxVisitor {
+public class ViolationsSyntaxVisitor: SyntaxVisitor {
```
jpsim, line 61: Let's make this `open` since it's designed to be subclassed.

```suggestion
open class ViolationsSyntaxVisitor: SyntaxVisitor {
```

jpsim, line 61: Can this be marked as `open` since it's designed to be subclassed?

```suggestion
open class ViolationsSyntaxVisitor: SyntaxVisitor {
```

jpsim (approved):

PR #4311, opened 2022-10-06
[bazel] For SwiftSyntax to compile with optimizations

keith:
This avoids potential stack overflows and slowness when just working on
SwiftLint itself with bazel. These compiles take a lot longer but as
long as you're not updating swift-syntax that should be a 1 time penalty

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -21,10 +21,7 @@ swift_library(
     visibility = ["//visibility:public"],
     deps = [
         "@com_github_jpsim_sourcekitten//:SourceKittenFramework",
-        "@com_github_apple_swift_syntax//:SwiftSyntax",
-        "@com_github_apple_swift_syntax//:SwiftSyntaxBuilder",
-        "@com_github_apple_swift_syntax//:SwiftParser",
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks Keith! Should this be moved to your swift-syntax-bazel repo and we can switch back to point to that?

This definitely makes otherwise `dbg` runs much faster.

Before

```
//Tests:CLITests                PASSED in 0.4s
//Tests:ExtraRulesTests         PASSED in 0.4s
//Tests:GeneratedTests          PASSED in 79.6s
//Tests:IntegrationTests        PASSED in 113.5s
//Tests:SwiftLintFrameworkTests PASSED in 13.7s
```

After

```
//Tests:CLITests                PASSED in 0.3s
//Tests:ExtraRulesTests         PASSED in 0.3s
//Tests:GeneratedTests          PASSED in 55.2s
//Tests:IntegrationTests        PASSED in 15.0s
//Tests:SwiftLintFrameworkTests PASSED in 9.9s
```

PS: There's something weird happening with OSSCheck & Signal, so ignore that, it's noise.

keith:
wow those integration tests!

PR #4313, opened 2022-10-06
Remove Signal from OSSCheck

jpsim:
It's misbehaving for some reason. It'd be appreciated if anyone wanted to look into that.


jpsim:
Example of misbehaving: https://github.com/realm/SwiftLint/pull/4311#issuecomment-1270586855

FYI @marcelofabri

PR #4317, opened 2022-10-07
Migrate `legacy_multiple` rule to SwiftSyntax

marcelofabri:


Source/SwiftLintFramework/Rules/Idiomatic/LegacyMultipleRule.swift
```diff
@@ -38,15 +39,53 @@ public struct LegacyMultipleRule: OptInRule, ConfigurationProviderRule {
...
+        private(set) var violationPositions: [AbsolutePosition] = []
+
+        override func visitPost(_ node: InfixOperatorExprSyntax) {
+            guard let operatorNode = node.operatorOperand.as(BinaryOperatorExprSyntax.self),
+                  operatorNode.operatorToken.tokenKind == .spacedBinaryOperator("%"),
+                  let parent = node.parent?.as(InfixOperatorExprSyntax.self),
+                  parent.leftOperand.as(InfixOperatorExprSyntax.self) == node,
+                  let parentOperatorNode = parent.operatorOperand.as(BinaryOperatorExprSyntax.self),
+                  parentOperatorNode.isEqualityOrInequalityOperator,
+                  parent.rightOperand.as(IntegerLiteralExprSyntax.self)?.digits.tokenKind == .integerLiteral("0") else {
```
SimplyDanny, line 76: Can also be the other way around with the `0` on the left, e.g. `0 == 4 % 2`.

SimplyDanny (approved):

Source/SwiftLintFramework/Rules/Idiomatic/LegacyMultipleRule.swift
marcelofabri: good callout! this is something that the existing code (before this PR) ignores, but it should be easy to update this PR to handle this case. I'll give it a try later

PR #4320, opened 2022-10-07
Rewrite `function_default_parameter_at_end` with SwiftSyntax

marcelofabri:


marcelofabri:
the oss-check results are due to the violation position for `static func`s changing + 3 new warnings on apple/swift

SimplyDanny (approved):

PR #4322, opened 2022-10-07
Rewrite `valid_ibinspectable` rule with SwiftSyntax

marcelofabri:


marcelofabri:
^ those are false positives, I'll fix that

jpsim (approved):

PR #4329, opened 2022-10-07
Make `nsobject_prefer_isequal` rule work for nested classes

SimplyDanny:


Source/SwiftLintFramework/Rules/Lint/NSObjectPreferIsEqualRule.swift
```diff
@@ -89,6 +83,14 @@ private extension FunctionDeclSyntax {
     }
 }
 
+private extension Syntax {
+    var isInObjcClass: Bool {
```
marcelofabri, line 87: maybe the existing rule had this problem already, but we should also check for `@objcmembers`

SimplyDanny: You are right. It didn't work before either. Added a CHANGELOG entry for the fix.

Source/SwiftLintFramework/Rules/Lint/NSObjectPreferIsEqualRule.swift
```diff
@@ -89,6 +83,14 @@ private extension FunctionDeclSyntax {
     }
 }
 
+private extension Syntax {
```
jpsim, line 86: I think if you extend SyntaxProtocol you don’t need to wrap with Syntax above.

SimplyDanny: Good hint! See #4332.

PR #4330, opened 2022-10-07
Use os_unfair_lock instead of NSLock on Darwin

jpsim:


marcelofabri:
Just curious: is this for performance reasons?

jpsim:
Yes, tends to be 2-3% faster in my tests.

PR #4331, opened 2022-10-07
Disregard whitespace differences in `identical_operands` rule

SimplyDanny:


jpsim (approved):
+1 if OSSCheck doesn’t report any performance degradation.

SimplyDanny:
> +1 if OSSCheck doesn’t report any performance degradation.

Even faster in most of the code bases. 🥳

PR #4333, opened 2022-10-09
Convert `number_separator` rule to SwiftSyntax

marcelofabri:
I expect a lot of warnings from oss-check because the violation position has changed for negative numbers


marcelofabri:
Checked the oss-check results. They're all due to the violation position changing when it's a negative number + 2 false negatives from `apple/swift` that are now caught.

SimplyDanny (approved):

PR #4334, opened 2022-10-09
Rename `if_let_shadowing` rule to `shorthand_optional_binding`

SimplyDanny:


Source/SwiftLintFramework/Rules/Idiomatic/ShorthandOptionalBindingRule.swift
```diff
@@ -1,11 +1,11 @@
 import SwiftSyntax
 
-public struct IfLetShadowingRule: OptInRule, SwiftSyntaxCorrectableRule, ConfigurationProviderRule {
+public struct ShorthandOptionalBindingRule: OptInRule, SwiftSyntaxCorrectableRule, ConfigurationProviderRule {
     public var configuration = SeverityConfiguration(.warning)
 
     public static var description = RuleDescription(
-        identifier: "if_let_shadowing",
-        name: "If-let Shadowing",
+        identifier: "shorthand_optional_binding",
```
jpsim, line 7: Want to add `if_let_shadowing` as a deprecated alias since we already shipped the previous version in a release candidate?

jpsim (approved):

Source/SwiftLintFramework/Rules/Idiomatic/ShorthandOptionalBindingRule.swift
jpsim: https://github.com/realm/SwiftLint/blob/7585aef91070536d044a766383d6803c1ab32fb2/Source/SwiftLintFramework/Models/RuleDescription.swift#L45

SimplyDanny: I'm fine with adding the alias. 

But do you think this is really necessary? Even if it has only been included in a pre-release version yet? Aren't they subject to change anytime?

jpsim: Yeah but it doesn't cost us anything to throw it in there.

jpsim: I don't feel strongly though, if you want to merge as-is because the PR is green and you don't want to wait for CI that's ok.

PR #4335, opened 2022-10-10
Rewrite `operator_whitespace` rule with SwiftSyntax

marcelofabri:


marcelofabri:
I've checked the oss-check results and IMO they're all false negatives that are now caught 🚨

SimplyDanny (approved):

PR #4339, opened 2022-10-10
Rewrite `unused_control_flow_label` with SwiftSyntax

marcelofabri:


Source/SwiftLintFramework/Rules/Lint/UnusedControlFlowLabelRule.swift
```diff
@@ -86,47 +85,85 @@ public struct UnusedControlFlowLabelRule: SubstitutionCorrectableASTRule, Config
...
 
-    public func validate(file: SwiftLintFile, kind: StatementKind,
-                         dictionary: SourceKittenDictionary) -> [StyleViolation] {
-        return self.violationRanges(in: file, kind: kind, dictionary: dictionary).map { range in
-            StyleViolation(ruleDescription: Self.description,
-                           severity: configuration.severity,
-                           location: Location(file: file, characterOffset: range.location))
+            let newNode = node.statement.withLeadingTrivia(node.leadingTrivia ?? .zero)
+            correctionPositions.append(violationPosition)
+            return visit(newNode).as(StmtSyntax.self) ?? newNode
```
marcelofabri, line 138: this feels weird, but I think we need to call `visit` on the node returned here to handle nested `LabeledStmtSyntax`s

Source/SwiftLintFramework/Rules/Lint/UnusedControlFlowLabelRule.swift
```diff
@@ -86,47 +85,85 @@ public struct UnusedControlFlowLabelRule: SubstitutionCorrectableASTRule, Config
...
-        var rangeToRemove = violationRange
-        let contentsNSString = file.stringView
-        if let byteRange = contentsNSString.NSRangeToByteRange(start: violationRange.location,
-                                                               length: violationRange.length),
-            let nextToken = file.syntaxMap.tokens.first(where: { $0.offset > byteRange.location }) {
-            let nextTokenLocation = contentsNSString.location(fromByteOffset: nextToken.offset)
-            rangeToRemove.length = nextTokenLocation - violationRange.location
+private extension LabeledStmtSyntax {
+    var violationPosition: AbsolutePosition? {
+        let visitor = BreakAndContinueLabelCollector(viewMode: .sourceAccurate)
```
marcelofabri, line 145: is this a good idea? having a separate visitor simplifies the implementation a lot, but not sure if we'd get huge impacts. 

This shouldn't matter that much in practice though because having a `LabeledStmtSyntax` in Swift is pretty rare to being with.

PR #4344, opened 2022-10-11
[Bazel] Set `swift.enable_library_evolution` for CryptoSwift

jpsim:
To avoid this compiler warning:

```
Sources/CryptoSwift/RSA/RSA+Cipher.swift:130:7: warning: default will never be executed
      @unknown default:
      ^
```


marcelofabri (approved):

jpsim:
This didn't have the intended effect. Going to go with https://github.com/realm/SwiftLint/pull/4345.

PR #4346, opened 2022-10-12
Update SwiftSyntax

jpsim:
Appears to have some improvements regarding parsing files with parser errors.


SimplyDanny:
> Looks like all the changes reported by OSSCheck are improvements over what was being reported before.

Indeed. Neat how resilient the parser is now. These false-negatives where accepted because this file isn't valid Swift code due to being a Sourcery template.

PR #4349, opened 2022-10-12
Reduce memory usage

jpsim:
When linting SwiftLint, this brings memory usage down by around 20%, going from 372MB to 297MB.

This is achieved by removing the unused `RebuildQueue` and by clearing cached data associated with files after processing them.

Depends on https://github.com/jpsim/SourceKitten/pull/749.

### DIFF
--- a/Source/SwiftLintFramework/Models/Linter.swift
+++ b/Source/SwiftLintFramework/Models/Linter.swift
@@ -249,6 +249,9 @@ public struct CollectedLinter {
                 "completely removed in a future release.")
         }
 
+        // Free some memory used for this file's caches. They shouldn't be needed after this point.
+        file.invalidateCache()
+
         return (violations, ruleTimes)
     }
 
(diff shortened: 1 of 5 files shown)


Source/SwiftLintFramework/Models/Linter.swift
```diff
@@ -249,6 +249,9 @@ public struct CollectedLinter {
                 "completely removed in a future release.")
         }
 
+        // Free some memory used for this file's caches. They shouldn't be needed after this point.
+        file.invalidateCache()
```
jpsim, line 253: I wonder if doing this asynchronously in a fire-and-forget way would speed things up.

```suggestion
        Task {
            file.invalidateCache()
        }
```

jpsim: https://github.com/realm/SwiftLint/pull/4351

PR #4351, opened 2022-10-12
Asynchronously invalidate file caches

jpsim:
To see if this fire-and-forget style cleanup improves lint times at all.


jpsim:
> To see if this fire-and-forget style cleanup improves lint times at all.

Not really.

PR #4352, opened 2022-10-12
Remove some file caches

jpsim:
These shouldn't be too expensive to compute on every access and should help reduce memory usage.


jpsim:
OSSCheck will benchmark this to determine if it's a win or not.

PR #4355, opened 2022-10-12
Rewrite `redundant_discardable_let` with SwiftSyntax

jpsim:


Source/SwiftLintFramework/Rules/Style/RedundantDiscardableLetRule.swift
```diff
@@ -29,62 +28,65 @@ public struct RedundantDiscardableLetRule: SubstitutionCorrectableRule, Configur
...
-        }
-        let nextUTF16Unit = fileContents.substring(with: NSRange(location: utf16Range.upperBound, length: 1))
-        return nextUTF16Unit == ":"
+private extension VariableDeclSyntax {
+    var hasRedundantDiscardableLetViolation: Bool {
+        letOrVarKeyword.tokenKind == .letKeyword &&
+            bindings.count == 1 &&
+            bindings.first!.pattern.is(WildcardPatternSyntax.self) &&
+            bindings.first!.typeAnnotation == nil &&
+            modifiers?.contains(where: { $0.name.text == "async" }) != true
```
SimplyDanny, line 90:
```suggestion
            modifiers?.contains(where: { $0.name.text == "async" }) == false
```

SimplyDanny (approved):

Source/SwiftLintFramework/Rules/Style/RedundantDiscardableLetRule.swift
jpsim: We want `nil` and `false` to both satisfy the guard.

Another way to write this would be:

```suggestion
            modifiers == nil || !modifiers.first!.contains(where: { $0.name.text == "async" })
```

But that's significantly harder to read.

SimplyDanny: I see. Optional booleans ... 😅

PR #4359, opened 2022-10-13
Rewrite `attributes` with SwiftSyntax

jpsim:
Not 100% the same as it previously was, but I think all the changes reported by OSSCheck are improvements or neutral.


jpsim:
Not 100% the same as it previously was, but I think all the changes reported by OSSCheck are improvements or neutral.

PR #4365, opened 2022-10-13
Rewrite `sorted_first_last` & `first_where` with SwiftSyntax

jpsim:
Add a `functionCallBase` helper.


jpsim:
All the newly reported violations were previously false negatives.

PR #4369, opened 2022-10-13
Rethink body line count calculation

jpsim:
A long-standing limitation with SourceKit's "editor open" request is that we weren't able to get certain tokens, such as braces, brackets and parentheses.

This meant that this code block would be counted as two lines:

```swift
print(
  "hi"
)
```

because the trailing `)` would be treated as a whitespace line.

This meant that our "body length" family of rules that measure the effective line count of declarations like functions, types or closures would often significantly under-count the number of content lines in a body.

Now with SwiftSyntax, we can get all tokens, including the ones SourceKit was previously ignoring, so we can get much more accurate line counts when ignoring whitespace and comments.

In addition, we weren't very thorough in how we measured body length.

As an exercise, how many lines long would you say the body of this function is?

```swift
func hello() {
  print("hello")
}
```

Does the body span one line or three lines?

I propose that we consistently ignore the left and right brace lines when calculating the body line count of these scopes so that we measure body line counts like this:

```swift
// 1 line
{ print("foo") }
// 1 line
{
}
// 1 line
{
  print("foo")
}
// 2 lines
{
  let sum = 1 + 2
  print(sum)
}
```

Now with those changes in place, in order to keep the default configuration thresholds to similar levels as before, we need to adjust them slightly. Here's what I'm suggesting:

|Rule|Before (warning/error)|After (warning/error)|
|-|-|-|
|closure_body_length|20/100|30/100|
|function_body_length|40/100|50/100|
|type_body_length|200/350|250/350|

This is a pretty significant breaking change and I suspect we'll hear from users who are surprised that some of their declarations now exceed the rule limits, but I believe this new approach to calculating body lines is more correct and intuitive compared to what we've had until now.

OSSCheck is also going to report a bazillion changes with this, which is expected given the scope of this change.

Finally, this removes one more use of a SourceKit request from our codebase, helping toward the SwiftSyntax migration efforts.


jpsim:
Given how significant this change is, I'd appreciate your thoughts and review @marcelofabri @SimplyDanny

Source/SwiftLintFramework/Helpers/BodyLineCounter.swift
```diff
@@ -0,0 +1,63 @@
...
+    /// - parameter leftBraceLine:  The line where the scope's left opening brace is located.
+    /// - parameter rightBraceLine: The line where the scope's right closing brace is located.
+    /// - parameter limit:          The body line limit to stay under, ignoring lines only containing comments and/or
+    ///                             whitespace.
+    ///
+    /// - returns: A tuple containing whether the body `exceeds` the specific `limit` and `lineCount`, the number of
+    ///            effective lines of the body ignoring lines only containing comments and/or whitespace.
+    static func lineCountIgnoringCommentsAndWhitespace(
+        file: SwiftLintFile, leftBraceLine: Int, rightBraceLine: Int, limit: Int
+    ) -> Result {
```
SimplyDanny, line 24: As the method name suggests it could just return the line count as an `Int` and the leave the assessment of it to the caller. This would also make the `limit` parameter unnecessary.

Source/SwiftLintFramework/Helpers/BodyLineCounter.swift
```diff
@@ -0,0 +1,63 @@
...
+    ///
+    /// - parameter file:           The source file.
+    /// - parameter leftBraceLine:  The line where the scope's left opening brace is located.
+    /// - parameter rightBraceLine: The line where the scope's right closing brace is located.
+    /// - parameter limit:          The body line limit to stay under, ignoring lines only containing comments and/or
+    ///                             whitespace.
+    ///
+    /// - returns: A tuple containing whether the body `exceeds` the specific `limit` and `lineCount`, the number of
+    ///            effective lines of the body ignoring lines only containing comments and/or whitespace.
+    static func lineCountIgnoringCommentsAndWhitespace(
```
SimplyDanny, line 22: Is there a convention on when to introduce a static function in a namespace and when to define it as a function in an extension (here an extension of `SwiftLintFile`)?

SimplyDanny (approved):
Better necessarily implies different. 😉

Source/SwiftLintFramework/Helpers/BodyLineCounter.swift
jpsim: No convention. I had this in an extension on `SwiftLintFile` in `SwiftLintFile+Regex.swift` at first but it wasn't related to regular expressions at all, so I wanted to move it elsewhere and came up with this namespace without thinking too deeply about it.

Now that you mention it, it's probably better as an extension on `SwiftLintFile` so we need fewer function parameters.

I'll move this to a `SwiftLintFile+BodyLineCount.swift` file.

Thanks for the idea!

Source/SwiftLintFramework/Helpers/BodyLineCounter.swift
jpsim: Yes that would probably be better. Let me try that.

PR #4370, opened 2022-10-14
Rewrite "body length" rules with SwiftSyntax

jpsim:
- `closure_body_length`
- `function_body_length`
- `type_body_length`

Depends on https://github.com/realm/SwiftLint/pull/4369


Source/SwiftLintFramework/Rules/Metrics/FunctionBodyLengthRule.swift
```diff
@@ -1,6 +1,6 @@
-import SourceKittenFramework
+import SwiftSyntax
 
-public struct FunctionBodyLengthRule: ASTRule, ConfigurationProviderRule {
+public struct FunctionBodyLengthRule: SourceKitFreeRule, ConfigurationProviderRule {
```
jpsim, line 3: I still need to handle initializers here.

jpsim: Done.

PR #4371, opened 2022-10-14
Rewrite `unused_closure_parameter` with SwiftSyntax

jpsim:


jpsim:
All the OSSCheck reports were previously false negatives 🙌

jpsim:
There's more code golf that could be played to clean this up. Others are free to tackle that if they like, but it gets the job done right now so I'm merging as-is.

PR #4372, opened 2022-10-14
Rewrite `legacy_objc_type` with SwiftSyntax

jpsim:


marcelofabri (approved):
nit: can we make `legacyObjcTypes` a `Set` now that's only used for `.contains`?

marcelofabri:
we probably should catch `NSNotification.Name` too (from oss-check)

jpsim:
> nit: can we make `legacyObjcTypes` a `Set` now that's only used for `.contains`?

Even if a collection is only ever used for membership checking, and its element type is Hashable, a linear array lookup may be faster than a hash+equality lookup for a small number of elements. That's because hashing may be more expensive than equality checks.

jpsim:
> we probably should catch `NSNotification.Name` too (from oss-check)

Is there an alternative to using `NSNotification.Name`? Seems like there's no Swift bridged equivalent for that one.

marcelofabri:
> > we probably should catch `NSNotification.Name` too (from oss-check)
> 
> Is there an alternative to using `NSNotification.Name`? Seems like there's no Swift bridged equivalent for that one.

`Notification.Name` should work - we have a few uses of it in our codebase

jpsim:
> `Notification.Name` should work - we have a few uses of it in our codebase

Ah indeed. Sorry I should have manually checked that before posting.

jpsim:
The only difference now with the previous implementation is that this doesn't catch references in import statements, but I think that's ok.

PR #4379, opened 2022-10-15
Introduce `ReasonedRuleViolation` type to associate a reason with a violation position

SimplyDanny:
With this new type we can get rid of most of the custom `StyleViolation` creators.

### DIFF
--- a/Source/SwiftLintFramework/Protocols/SwiftSyntaxRule.swift
+++ b/Source/SwiftLintFramework/Protocols/SwiftSyntaxRule.swift
@@ -11,11 +11,11 @@ public protocol SwiftSyntaxRule: SourceKitFreeRule {
 
     /// Produce a violation for the given file and absolute position.
     ///
-    /// - parameter file:     The file for which to produce the violation.
-    /// - parameter position: The absolute position in the file where the violation should be located.
+    /// - parameter file:      The file for which to produce the violation.
+    /// - parameter violation: A violation in the file.
     ///
     /// - returns: A violation for the given file and absolute position.
-    func makeViolation(file: SwiftLintFile, position: AbsolutePosition) -> StyleViolation
+    func makeViolation(file: SwiftLintFile, violation: ReasonedRuleViolation) -> StyleViolation
 
     /// Gives a chance for the rule to do some pre-processing on the syntax tree.
     /// One typical example is using `SwiftOperators` to "fold" the tree, resolving operators precedence.
@@ -29,11 +29,12 @@ public protocol SwiftSyntaxRule: SourceKitFreeRule {
 
 public extension SwiftSyntaxRule where Self: ConfigurationProviderRule,
                                        ConfigurationType: SeverityBasedRuleConfiguration {
-    func makeViolation(file: SwiftLintFile, position: AbsolutePosition) -> StyleViolation {
+    func makeViolation(file: SwiftLintFile, violation: ReasonedRuleViolation) -> StyleViolation {
         StyleViolation(
             ruleDescription: Self.description,
-            severity: configuration.severity,
-            location: Location(file: file, position: position)
+            severity: violation.severity ?? configuration.severity,
+            location: Location(file: file, position: violation.position),
+            reason: violation.reason
         )
     }
 }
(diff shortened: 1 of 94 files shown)


Source/SwiftLintFramework/Protocols/SwiftSyntaxRule.swift
```diff
@@ -48,16 +49,49 @@ public extension SwiftSyntaxRule {
...
+    }
+
+    public static func < (lhs: Self, rhs: Self) -> Bool {
+        lhs.position < rhs.position
+    }
+}
+
+/// Extension for arrays of `ReasonedRuleViolation`s that provides the automatic conversion of
+/// `AbsolutePosition`s into `ReasonedRuleViolation`s (without a specific reason).
+extension Array where Element == ReasonedRuleViolation {
```
marcelofabri, line 81: nice! i was looking into doing something similar, but this was the missing piece. I went with the protocol/associated type route and then faced a bunch of issues, so this is really clever ✨

Source/SwiftLintFramework/Protocols/SwiftSyntaxRule.swift
```diff
@@ -48,16 +49,49 @@ public extension SwiftSyntaxRule {
...
+    mutating func append(_ position: AbsolutePosition) {
+        append(ReasonedRuleViolation(position: position))
+    }
+
+    mutating func append(contentsOf positions: [AbsolutePosition]) {
+        positions.forEach { append($0) }
```
marcelofabri, line 87: My gut feeling is that `append(contentsOf: positions.map {  ReasonedRuleViolation(position: $0) })` would be faster

Source/SwiftLintFramework/Protocols/SwiftSyntaxRule.swift
```diff
@@ -48,16 +49,49 @@ public extension SwiftSyntaxRule {
...
+            .map { makeViolation(file: file, violation: $0) }
+    }
+}
+
+/// A violation produced by `ViolationsSyntaxVisitor`s.
+public struct ReasonedRuleViolation: Comparable {
+    /// The violation's position.
+    public let position: AbsolutePosition
+    /// A specific reason for the violation.
+    public let reason: String?
```
marcelofabri, line 63: while we're at this, can we add the `severity`? this is the other reason rules can't use `ViolationsSyntaxVisitor` currently

Source/SwiftLintFramework/Rules/Idiomatic/NoFallthroughOnlyRule.swift
```diff
@@ -24,7 +24,7 @@ private extension NoFallthroughOnlyRule {
         override func visitPost(_ node: SwitchCaseListSyntax) {
             let cases = node.compactMap { $0.as(SwitchCaseSyntax.self) }
 
-            let violations = cases.enumerated()
+            cases.enumerated()
```
marcelofabri, line 27: nit: can we revert this change? i feel like it's slightly easier to debug/read

Source/SwiftLintFramework/Rules/Style/ClosureParameterPositionRule.swift
```diff
@@ -117,16 +117,15 @@ private extension ClosureParameterPositionRule {
                 return
             }
 
-            let violations = signature.positionsToCheck
+            signature.positionsToCheck
```
marcelofabri, line 120: same nit about this change

marcelofabri (approved):
This is great! Can you please also rebase before merging as some rules got merged after you created your branch? 🙏

Source/SwiftLintFramework/Protocols/SwiftSyntaxRule.swift
SimplyDanny: I already had it in there but cannot remember why I decided to leave it out at the end.

Could you mention an example rule for which this would be useful?

marcelofabri: any of the "Count" violations, like `FunctionParameterCountRule`

marcelofabri (approved):

PR #4381, opened 2022-10-16
Add way to preprocess syntaxTree in SwiftSyntaxRule

marcelofabri:
This is another reason why some rules can't use `SwiftSyntaxRule` directly, so adding some convenience for them to avoid duplicating the whole `walk` dance.

### DIFF
--- a/Source/SwiftLintFramework/Protocols/SwiftSyntaxRule.swift
+++ b/Source/SwiftLintFramework/Protocols/SwiftSyntaxRule.swift
@@ -16,6 +16,15 @@ public protocol SwiftSyntaxRule: SourceKitFreeRule {
     ///
     /// - returns: A violation for the given file and absolute position.
     func makeViolation(file: SwiftLintFile, position: AbsolutePosition) -> StyleViolation
+
+    /// Gives a chance for the rule to do some pre-processing on the syntax tree.
+    /// One typical example is using `SwiftOperators` to "fold" the tree, resolving operators precedence.
+    /// By default, returns this just returns the same `syntaxTree`.
+    ///
+    /// - parameter syntaxTree: The syntaxTree to run pre-processing on
+    ///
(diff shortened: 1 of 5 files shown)


Source/SwiftLintFramework/Protocols/SwiftSyntaxRule.swift
```diff
@@ -16,6 +16,15 @@ public protocol SwiftSyntaxRule: SourceKitFreeRule {
     ///
     /// - returns: A violation for the given file and absolute position.
     func makeViolation(file: SwiftLintFile, position: AbsolutePosition) -> StyleViolation
+
+    /// Gives a chance for the rule to do some pre-processing on the syntax tree.
+    /// One typical example is using `SwiftOperators` to "fold" the tree, resolving operators precedence.
+    /// By default, returns this just returns the same `syntaxTree`.
```
SimplyDanny, line 22:
```suggestion
    /// By default, it just returns the same `syntaxTree`.
```

Source/SwiftLintFramework/Protocols/SwiftSyntaxRule.swift
```diff
@@ -16,6 +16,15 @@ public protocol SwiftSyntaxRule: SourceKitFreeRule {
...
+    ///
+    /// - parameter syntaxTree: The syntaxTree to run pre-processing on
```
SimplyDanny, line 24: Nit:
```suggestion
    /// - parameter syntaxTree: The syntax tree to run pre-processing on
```

SimplyDanny (approved):
Cool!

I prepared something similar already: The `ViolationsSyntaxVisitor` got a new property `requiresFoldedTree: Bool` which the `validate(file: SwiftLintFile)` method used to either walk the original or the folded tree.

Your solution is more generic, though, allowing also other preprocessing steps. I like it more. 👍

PR #4382, opened 2022-10-16
Rewrite `control_statement` rule with SwiftSyntax

marcelofabri:


marcelofabri:
One case that the previous implementation would ignore is:

```swift
if (edges.contains { edge in edges.filter { $0.parent == edge.parent }.count > 1 }) { ... }
```

IMO we should catch it but not automatically fix since it'd generate invalid code. I think using this form is preferable:

```swift
if edges.contains(where: { edge in edges.filter { $0.parent == edge.parent }.count > 1 }) { ... }
```

Thoughts?

jpsim:
> Thoughts?

I tend to agree that the spirit of this rule is such that trailing closures wrapped in parentheses still violates this rule, but that it's impossible to automatically fix because we can't know what the closure parameter label is in all cases.

PR #4384, opened 2022-10-16
Support .focused() in keys for corrections

marcelofabri:
I never remember that using `.focused()` only works in the right side of `corrections`, so just making this a little bit more flexible


SimplyDanny (approved):
I feel with you. 😅 Thank you for doing this!

PR #4390, opened 2022-10-16
Migrate `inclusive_language` rule to SwiftSyntax

marcelofabri:


marcelofabri:
This is still probably worth it merging even with small performance regressions since they'll pay off once we don't have any more SourceKit rules. However, I'd love to make it faster - any ideas? 

I'm thinking about maybe using a regex instead of iterating over the terms, but not sure if that'd be faster.

jpsim (approved):

PR #4395, opened 2022-10-17
Rewrite `redundant_set_access_control` with SwiftSyntax

marcelofabri:


jpsim:
I think there's a regression here:

```swift
extension Color {
    /// Some docs.
    public internal(set) static var someColor = Color.anotherColor
}
```

jpsim (approved):

PR #4396, opened 2022-10-17
Rewrite `duplicated_key_in_dictionary_literal` with SwiftSyntax

marcelofabri:


Source/SwiftLintFramework/Rules/Lint/DuplicatedKeyInDictionaryLiteralRule.swift
```diff
@@ -79,74 +79,59 @@ public struct DuplicatedKeyInDictionaryLiteralRule: ASTRule, ConfigurationProvid
...
+private extension DuplicatedKeyInDictionaryLiteralRule {
+    final class Visitor: ViolationsSyntaxVisitor {
+        override func visitPost(_ list: DictionaryElementListSyntax) {
+            let keys = list.map(\.keyExpression).compactMap { expr -> DictionaryKey? in
+                expr.stringContent.map {
+                    DictionaryKey(position: expr.positionAfterSkippingLeadingTrivia, content: $0)
                 }
-
-                let existingKeyKinds = file.syntaxMap.kinds(inByteRange: existingKey.byteRange)
-                let keyKinds = file.syntaxMap.kinds(inByteRange: key.byteRange)
```
marcelofabri, line 102: we're not doing this check anymore, which in theory could lead to false positives, but we also don't do it in other rules like `identical_operands` so we should be fine?

jpsim:
I also took a stab at rewriting this one in #4464 forgetting to check open PRs. Sharing in case there's something useful in its implementation, but yours seems more sophisticated.

Source/SwiftLintFramework/Rules/Lint/DuplicatedKeyInDictionaryLiteralRule.swift
```diff
@@ -79,74 +79,59 @@ public struct DuplicatedKeyInDictionaryLiteralRule: ASTRule, ConfigurationProvid
...
-        return keys.compactMap { key -> DictionaryKey? in
-            guard let range = key.byteRange,
-                  let substring = contents.substringWithByteRange(range) else {
-                return nil
-            }
+            var existingKeys: [String: DictionaryKey] = [:]
+            let newViolations = keys
+                .filter { key in
+                    guard existingKeys[key.content] != nil else {
+                        existingKeys[key.content] = key
```
jpsim, line 104: Nit, personal opinion: this mutation side-effect doesn't belong in a `filter`

marcelofabri: updated!

jpsim (approved):

PR #4401, opened 2022-10-17
Let more rules rely on `SwiftSyntaxRule`'s defaults

SimplyDanny:


Source/SwiftLintFramework/Protocols/SwiftSyntaxRule.swift
```diff
@@ -63,6 +63,22 @@ public extension SwiftSyntaxRule {
             .map { makeViolation(file: file, violation: $0) }
     }
 
+    func makeViolation(file: SwiftLintFile, violation: ReasonedRuleViolation) -> StyleViolation {
+        guard let severity = violation.severity else {
+            // This error will only be thrown in tests. It cannot come up at runtime.
+            queuedFatalError("""
+                A severity must be provided. Either define it in the violation or make the rule configuration \
+                conform to `SeverityBasedRuleConfiguration` to take the default.
+            """)
```
SimplyDanny, line 72: @jpsim, @marcelofabri: Is this acceptable for you? This can also be ensured by compile-time checks with some protocol-, associatedtype- and generics-voodoo. As I've written in the comment, however, a missing severity will be noticed immediately when running a single test. So some complexity could be saved here (for now).

jpsim (approved):

Source/SwiftLintFramework/Protocols/SwiftSyntaxRule.swift
jpsim: I’m ok with this

marcelofabri: we could also extract this default implementation to a new protocol so using it is explicit and is less likely we make a mistake

SimplyDanny: > we could also extract this default implementation to a new protocol so using it is explicit and is less likely we make a mistake 

You mean a sub-protocol of `SwiftSyntaxRule`? 

> I don't see a reason why every rule couldn't be configurable by severity level too. But it feels like doing either of those things is out of scope for this PR.

I agree. Some of the configurations provide multiple severity levels, however.

marcelofabri: > You mean a sub-protocol of SwiftSyntaxRule?

Yeah, just so this behavior is opt-in and explicit

SimplyDanny: I'm not yet convinced this is worth another rule type (we already have an involved hierarchy of them) as I don't see "hidden surprises" as long as every rule has at least one triggering example.

Let's really keep it out of this PR. 😅

PR #4406, opened 2022-10-17
Only lint Swift source files within the provided target

tonyarnold:
When running with the SwiftPM build plugin provided in #4176, I noticed that regardless of which target was being built, the entire working directory was being linted. 

This PR proposes to only lint Swift source files within the passed target by passing them as arguments to the build plugin process. 

It also bails out early if the type of target it not suitable for linting, or there are no Swift source files to lint.

### DIFF
--- a/Plugins/SwiftLintPlugin/Path+Helpers.swift
+++ b/Plugins/SwiftLintPlugin/Path+Helpers.swift
@@ -2,7 +2,7 @@ import Foundation
 import PackagePlugin
 
 extension Path {
-    /// Scans the receiver, then all of it's parents looking for a configuration file with the name ".swiftlint.yml".
+    /// Scans the receiver, then all of its parents looking for a configuration file with the name ".swiftlint.yml".
     /// - Returns: Path to the configuration file, or nil if one cannot be found.
     func firstConfigurationFileInParentDirectories() -> Path? {
         // TODO: Consider linking to the framework to get the default configuration file name
--- a/Plugins/SwiftLintPlugin/Path+Helpers.swift
+++ b/Plugins/SwiftLintPlugin/Path+Helpers.swift
@@ -1,0 +1,18 @@
+import Foundation
+import PackagePlugin
+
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks for the PR, Tony!

This seems like a win at first glance, but may lead to some surprising results depending on how Xcode deals with diagnostics for files not in the currently selected target.

For example, what if you make change to a file that's not in your selected target but still covered by SwiftLint? The IDE wouldn't show you any diagnostics for that. Maybe it already doesn't?

It'd be surprising if Xcode says the project has no linter violations, only for then CI or a CLI invocation to say there are some.

If Xcode ignores non-target files regardless, then sure I'd say this is a nice win and we should do something like this.

tonyarnold:
Hey JP! :wave: 

> For example, what if you make change to a file that's not in your selected target but still covered by SwiftLint? The IDE wouldn't show you any diagnostics for that. Maybe it already doesn't?

To be honest, I think that what you've described is exactly what I'd expect. I don't want files outside of my current build target to be assessed/linted by the current target's tooling. Listing of those files should be taken care of by whatever target those files are a part of. 

CI should probably make use of a command plugin, rather than a build plugin? Especially if you want a customised/CI specific reporter, right?

jpsim:
I really don't have much experience with SwiftPM plugins. I'll merge this once you add a changelog entry describing the changes.

Thanks!

Plugins/SwiftLintPlugin/Path+Helpers.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+import PackagePlugin
+
+extension Path {
+    /// Scans the receiver, then all of it's parents looking for a configuration file with the name ".swiftlint.yml".
```
jpsim, line 5: Nit: `it's` -> `its`

```suggestion
    /// Scans the receiver, then all of its parents looking for a configuration file with the name ".swiftlint.yml".
```

PR #4413, opened 2022-10-18
Rewrite `return_arrow_whitespace` with SwiftSyntax

marcelofabri:
~I'm having issues thinking about how to best implement this Rewriter, so putting up this WIP PR if anyone wants to chime in~

Decided to use the good old String substitution

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -141,6 +141,7 @@
   - `redundant_objc_attribute`
   - `redundant_string_enum_value`
   - `required_deinit`
+  - `return_arrow_whitespace`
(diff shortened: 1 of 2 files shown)


jpsim:
I remember looking into this one and other spacing related rules and having a hard time because SwiftSyntax associates some whitespace with tokens before or after, and we want to operate on the combined whitespace between tokens.

Maybe if we scan the SwiftSyntax repo for tests or examples we'll find a new way to think about whitespace that can help us with these spacing rules.

jpsim (approved):
Creative solution

PR #4425, opened 2022-10-20
Rewrite `prefixed_toplevel_constant` with SwiftSyntax

marcelofabri:


Source/SwiftLintFramework/Rules/Style/PrefixedTopLevelConstantRule.swift
```diff
@@ -34,15 +32,20 @@ public struct PrefixedTopLevelConstantRule: ASTRule, OptInRule, ConfigurationPro
...
+            }
+            """)
         ],
         triggeringExamples: [
             Example("private let ↓Foo = 20.0"),
             Example("public let ↓Foo = false"),
             Example("internal let ↓Foo = \"Foo\""),
             Example("let ↓Foo = true"),
             Example("let ↓foo = 2, ↓bar = true"),
-            Example("var foo = true, let ↓Foo = true"),
```
marcelofabri, line 45:
```
error: 'let' cannot appear nested inside another 'var' or 'let' pattern
var foo = true, let Foo = true
```

jpsim (approved):

PR #4428, opened 2022-10-20
Rewrite `type_name` rule with SwiftSyntax

marcelofabri:
~This also now validates protocols, which I think was just a historical oversight.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,6 +132,7 @@
   - `trailing_comma`
   - `trailing_semicolon`
   - `type_body_length`
+  - `type_name`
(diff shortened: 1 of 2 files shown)


marcelofabri:
That's a lot of violations. I guess I'll keep the existing behavior for now.

marcelofabri:
Checked the [logs in Buildkite](https://buildkite.com/swiftlint/swiftlint/builds/4104#0183f4da-3591-44e6-9a1b-3d073cad89b9) and all oss-check results are in `apple/Swift` in files that were probably crashing SourceKit before

PR #4432, opened 2022-10-20
Rewrite `lower_acl_than_parent` with SwiftSyntax

jpsim:
And fix violations in SwiftLint.


jpsim:
I suspect this will catch quite a few previous false negatives.

PR #4440, opened 2022-10-21
Rewrite `trailing_closure` with SwiftSyntax

marcelofabri:
- Rewrite `trailing_closure` with SwiftSyntax
- Fix false positives


marcelofabri:
I think I was able to handle all the false positives from the previous PR - checking oss-check these seem all valid violations.

SimplyDanny:
Rule has been rewritten in #5414.

Issue #4415, opened 2022-10-18
lint command paths argument is unexpectedly case sensitive when applying exclude rules


alexbbrown:
swiftlint lint [<options>] [<paths> ...]

If my code is held in `/home/a/b/c` and I pass in the argument `/home/a/b/C` then SwiftLint scans the directory `/home/a/b/c/`. 

That's cool, and expected.  it's a Mac, it's not case sensitive.

BUT.  If my `config.yml` file contains an exclusion rule:
```
excluded: # paths to ignore during linting.
  - build
```

😀 then the exclusion works  when the invocation looks like this:
```
swiftlint lint --config config.yml /home/a/b/C
```
😡 but the exclusion is *not honored* in this case:
```
swiftlint lint --config config.yml /home/a/b/C
```
and then my /home/a/b/c/build directory gets scanned, with horrible errors as a result.

alexbbrown:
0.49.1

PR #4444, opened 2022-10-22
Recompute the absolute path supplied on the command line as relative,…

alexbbrown:
… so that it correctly resolves path case consistent with the exclude and include values.

To address this issue: https://github.com/realm/SwiftLint/issues/4415 - where a trivial use of excludes can fail if the (root) path supplied to lint is not case-identical to the path on the filesystem.

# Detailed Issue

In detail: currently the included and excluded paths are always computed as relative paths, like so:

```
        includedPaths = includedPaths.map {
            $0.bridge().absolutePathRepresentation(rootDirectory: previousBasePath).path(relativeTo: newBasePath)
        }

        excludedPaths = excludedPaths.map {
            $0.bridge().absolutePathRepresentation(rootDirectory: previousBasePath).path(relativeTo: newBasePath)
        }
    }
```

The includedPaths, excludedPaths and the rootPath (aka `path`) are searched to find all lintable files

```
    let pathsForPath = includedPaths.isEmpty
            ? fileManager.filesToLint(inPath: path, rootDirectory: nil)
            : []

       let includedPaths = self.includedPaths
            .flatMap(Glob.resolveGlob)

        return excludeByPrefix
            ? filterExcludedPathsByPrefix(in: pathsForPath, includedPaths)
            : filterExcludedPaths(fileManager: fileManager, in: pathsForPath, includedPaths)
```

In filterExcludedPathsByPrefix the set of lintable files reachable from the excludedPaths is subtracted from the set reachable from includedPaths plus rootPath.

```
let excludedPaths = self.excludedPaths(fileManager: fileManager)
result.minusSet(Set(excludedPaths))
```

This subtraction can fail if rootPath case does not exactly match the actual case of files on disk, leading to excluded paths being unexpectedly linted.  

This is because rootPath is the only absolute path, and case errors in its path are not corrected.

However, the Include and Exclude paths are computed as paths relative to the CWD, and when they are converted back to absolute paths, any case errors are corrected.

Resolution
###

Compute the path as a path relative to CWD, and then turn it back to absolute.  

Now the exact same processing that happens to the include/exclude paths is also applied to the rootPath, which means it has the same case.

Alternatives considered:
###

This change could also be applied to sourceKitten's `absolutePathRepresentation(rootDirectory: String = FileManager.default.currentDirectoryPath)` to make its response to the different values more consistent.

### DIFF
--- a/Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
@@ -36,7 +36,14 @@ extension Configuration {
         // If path is a file and we're not forcing excludes, skip filtering with excluded/included paths
         if path.isFile && !forceExclude { return [path] }
 
-        let pathsForPath = includedPaths.isEmpty ? fileManager.filesToLint(inPath: path, rootDirectory: nil) : []
+        let relativePath: String
+        if path.bridge().isAbsolutePath {
+            relativePath = path.bridge().path(relativeTo: rootDirectory)
+        } else {
+            relativePath = path
+        }
+
+        let pathsForPath = includedPaths.isEmpty ? fileManager.filesToLint(inPath: relativePath, rootDirectory: nil) : []
         let includedPaths = self.includedPaths
             .flatMap(Glob.resolveGlob)
             .parallelFlatMap { fileManager.filesToLint(inPath: $0, rootDirectory: rootDirectory) }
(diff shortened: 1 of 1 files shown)


alexbbrown:
Closing while I fix a couple issues

alexbbrown:
It's hard to write a test for this without actually creating files on the fileSystem, because the behaviour that changes the case of the paths actually comes from NSURL (not fileManager) reading the disk

alexbbrown:
I have re-opened this pull request.  If anyone reviews this, I would appreciate suggestions on how to write a test that constructs a sample set of folders and swift files in a temporary directory (see the issue for a simple example) so I can write a test.

I can see only one example - in sandbox test code - where this is done so far in the SwiftLint tests.

ehtisham-badar (approved):

PR #4447, opened 2022-10-23
Rewrite `reduce_into` rule with SwiftSyntax

marcelofabri:


Source/SwiftLintFramework/Rules/Performance/ReduceIntoRule.swift
```diff
@@ -89,61 +88,91 @@ public struct ReduceIntoRule: ASTRule, ConfigurationProviderRule, OptInRule {
...
-        let initMatch = initExpression.firstMatch(in: contents.string, options: [], range: range)
-        return initMatch != nil
+        return nil
+    }
+}
+
+private extension IdentifierExprSyntax {
+    var isCopyOnWriteType: Bool {
+        let type = identifier.text
+        return type == "Array" || type == "Dictionary"
```
jpsim, line 176: Should we add `Set`?

jpsim (approved):

PR #4449, opened 2022-10-23
Rewrite `unused_optional_binding` with SwiftSyntax

marcelofabri:


marcelofabri:
i've looked at the oss-check and they seem legit. `while let _ = await next() {}` was suspicious to me, but afaict `while await next() != nil {}` can be used instead

Source/SwiftLintFramework/Rules/Style/UnusedOptionalBindingRule.swift
```diff
@@ -38,56 +37,51 @@ public struct UnusedOptionalBindingRule: ASTRule, ConfigurationProviderRule {
             "}\n"),
             Example("if let ↓(_, _, _) = getOptionalTuple(), let bar = Foo.optionalValue {\n" +
             "}\n"),
-            Example("func foo() {\nif let ↓_ = bar {\n}\n"),
-            Example("if case .some(let ↓_) = self {}")
```
marcelofabri, line 42: this would be caught by `empty_enum_arguments`

Source/SwiftLintFramework/Rules/Style/UnusedOptionalBindingRule.swift
```diff
@@ -38,56 +37,51 @@ public struct UnusedOptionalBindingRule: ASTRule, ConfigurationProviderRule {
             "}\n"),
             Example("if let ↓(_, _, _) = getOptionalTuple(), let bar = Foo.optionalValue {\n" +
             "}\n"),
-            Example("func foo() {\nif let ↓_ = bar {\n}\n"),
```
jpsim, line 41: Why was this one removed?

jpsim (approved):

Source/SwiftLintFramework/Rules/Style/UnusedOptionalBindingRule.swift
marcelofabri: it wasn't, I just removed the trailing comma

jpsim: Got it

PR #4453, opened 2022-10-23
Add more repos to oss-check

marcelofabri:


tools/oss-check
```diff
@@ -30,6 +30,9 @@ OptionParser.new do |opts|
   opts.on('--skip-clean', 'skip cleaning on completion') do |skip_clean|
     @options[:skip_clean] = skip_clean
   end
+  opts.on('--force', 'run oss-check even if binaries are equal') do |force|
```
jpsim, line 33: Nice

jpsim (approved):
Approving but I think we’re really due for an optimization for how we clone these repos. A daily scheduled CI job that keeps them locally updated on the machine or something.

marcelofabri:
> Approving but I think we’re really due for an optimization for how we clone these repos. A daily scheduled CI job that keeps them locally updated on the machine or something.

I agree, but from my recent experience the analyze job is still the slowest to run since we've moved oss-check to bazel, so it hasn't been _too_ painful (other than making queuing worse)

jpsim (approved):
The added coverage here is worth making this a bit longer to run, but let's not add any more repos until we've optimized how we clone them.

jpsim:
https://github.com/realm/SwiftLint/pull/4456

PR #4454, opened 2022-10-23
Update SwiftSyntax to `60c7037`

jpsim:
This has changes to how comments are associated to nodes.

See https://github.com/apple/swift-syntax/pull/985


jpsim:
There are some regressions here I need to fix.

PR #4457, opened 2022-10-23
Move TSan Runs CI job to Buildkite

jpsim:
Now that we have more Buildkite machines, these should be faster than running on Azure Pipelines.

Also remove the `//bazel:xcode_config` configurations since they should no longer be in use.


jpsim:
Ran in 4m47s on Buildkite vs 16m8s on Azure Pipelines.

PR #4458, opened 2022-10-23
[CI] Generate docs with Buildkite instead of Azure Pipelines

jpsim:


jpsim:
rules_ruby giving me a hard time here

PR #4459, opened 2022-10-23
Rewrite `discarded_notification_center_observer` with SwiftSyntax

jpsim:


jpsim:
The two new violations reported by OSSCheck were previously false negatives.

PR #4465, opened 2022-10-24
Rewrite `identifier_name` rule with SwiftSyntax

marcelofabri:


SimplyDanny:
Rule has been rewritten in #5354.

PR #4456, opened 2022-10-23
[OSSCheck] Cache repos

jpsim:
This should speed up OSSCheck runs considerably.

In my local testing, cloning repos goes from 77s to 29s on a 100 Mbps connection.

PR #4473, opened 2022-10-24
Revert "[OSSCheck] Cache repos"

jpsim:
Reverts realm/SwiftLint#4456

This might need more error handling / recovery: https://github.com/realm/SwiftLint/pull/4456#issuecomment-1288792588

### DIFF
--- a/tools/oss-check
+++ b/tools/oss-check
@@ -60,10 +60,6 @@ class Repo
     "https://github.com/#{github_location}"
   end
 
-  def cached_tarball
-    "/tmp/#{name}.tgz"
-  end
-
   def to_s
     @name
   end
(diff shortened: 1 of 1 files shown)


tools/oss-check
```diff
@@ -143,20 +139,7 @@ def setup_repos
...
     dir = "#{@working_dir}/#{repo.name}"
     puts "Cloning #{repo}"
-    if File.exists?(repo.cached_tarball)
-      perform("tar -xzvf #{repo.cached_tarball} #{dir} 2> /dev/null")
-      Dir.chdir(dir) do
-        commit_before_pull = `git rev-parse HEAD`.strip
-        perform("git pull 2> /dev/null")
-        commit_after_pull = `git rev-parse HEAD`.strip
-        if commit_before_pull != commit_after_pull
-          perform("tar -czvf #{repo.cached_tarball} #{dir} 2> /dev/null")
```
jpsim, line 153: Now I'm wondering if this should create a tarball to a temporary file and then move it atomically to prevent any interruptions from producing a corrupt tarball in the final destination.

PR #4477, opened 2022-10-24
[OSSCheck] Cache repos (#4456)

jpsim:
This should speed up OSSCheck runs considerably.

In my local testing, cloning repos goes from 77s to 29s on a 100 Mbps connection.

This should be resilient to interruptions now.

### DIFF
--- a/tools/oss-check
+++ b/tools/oss-check
@@ -60,6 +60,10 @@ class Repo
     "https://github.com/#{github_location}"
   end
 
+  def cached_tarball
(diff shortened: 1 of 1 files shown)


marcelofabri:
Still seeing issues with oss-check (eg https://buildkite.com/swiftlint/swiftlint/builds/4351#01842b5b-ed9e-4879-a33e-6c172bf4ed17), going to revert this for now.

PR #4482, opened 2022-10-26
Create source location converters from source

jpsim:
Rather than from syntax trees.

I'll only merge this if it has a measurable performance improvement.


jpsim:
Between 3% slower to 1% faster according to OSSCheck, let's pass.

PR #4485, opened 2022-10-28
Fixes SwiftLint support on Xcode Cloud

### LINKED ISSUE #4484: SwiftLint as Xcode Plugin doesn't work on Xcode Cloud
### New Issue Checklist

### Describe the bug

When setting up SwiftLint as an Xcode Plugin it always fails in Xcode Cloud builds. This happens because Xcode Cloud doesn't have write access to `pluginWorkDirectory` or `.cachesDirectory`.

```
…
2022-10-27T11:29:16.891006545Z	Done linting! Found 13 violations, 0 serious in 612 files.
2022-10-27T11:29:16.891087138Z	Error: You don’t have permission to save the file “30cde6e7dbce7eecb2a554e9f928c329b93
2022-10-27T11:29:16.891194169Z	5109f468fa23b621c67e13581183e.plist” in the folder “B0249
2022-10-27T11:29:16.891286022Z	CAF-F
2022-10-27T11:29:16.891350071Z	879-3BE8-8269-FC397BB
2022-10-27T11:29:16.891423633Z	A4E12\342\200
2022-10-27T11:29:16.891483188Z	\235.
2022-10-27T11:29:16.891612278Z	Command PhaseScriptExecution failed with a nonzero exit code
…
```

### Environment

* SwiftLint version:
  [0.50.0-rc.2](https://github.com/realm/SwiftLint/tree/0.50.0-rc.2)
* Installation method used:
  SPM, installed as Xcode Plugin

  Not relevant

  No
  Version 14.1 (14B47)

  No, only reproducible on Xcode Cloud

JagCesar:
This PR fixes so SwiftLint installed as Xcode Plugin can be used on Xcode Cloud. The problem was that Xcode Cloud doesn't have write access to the `.cachesDirectory`. We solved this by always using ` NSTemporaryDirectory()` if we're being run on a CI.

This will have the side-effect of doing the same on all CI's, but we couldn't come up with a way to distinguish Xcode Cloud from other CI's.

Solved this together with @westerlund

fixes #4484

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@
 
 #### Enhancements
 
+* Fix SwiftLint support on Xcode Cloud.  
+  [JagCesar](https://github.com/JagCesar)
+  [westerlund](https://github.com/westerlund)
+  [#4484](https://github.com/realm/SwiftLint/issues/4484)
+
 * Add `no_magic_numbers` rule to avoid "Magic Numbers".  
   [Henrik Storch](https://github.com/thisisthefoxe)
   [#4031](https://github.com/realm/SwiftLint/issues/4024)
--- a/Source/swiftlint/Extensions/ProcessInfo+XcodeCloud.swift
+++ b/Source/swiftlint/Extensions/ProcessInfo+XcodeCloud.swift
@@ -1,0 +1,28 @@
+import Foundation
+
+extension ProcessInfo {
+    var isLikelyXcodeCloudEnvironment: Bool {
+        // https://developer.apple.com/documentation/xcode/environment-variable-reference
+        let requiredKeys: Set = [
+            "CI",
+            "CI_BUILD_ID",
+            "CI_BUILD_NUMBER",
+            "CI_BUNDLE_ID",
+            "CI_COMMIT",
+            "CI_DERIVED_DATA_PATH",
+            "CI_PRODUCT",
+            "CI_PRODUCT_ID",
+            "CI_PRODUCT_PLATFORM",
+            "CI_PROJECT_FILE_PATH",
+            "CI_START_CONDITION",
+            "CI_TEAM_ID",
+            "CI_WORKFLOW",
+            "CI_WORKSPACE",
+            "CI_XCODE_PROJECT",
+            "CI_XCODE_SCHEME",
+            "CI_XCODEBUILD_ACTION"
+        ]
+
+        return requiredKeys.isSubset(of: environment.keys)
+    }
+}
--- a/Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
+++ b/Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
@@ -262,7 +262,11 @@ private class LintOrAnalyzeResultBuilder {
         }
         configuration = config
         reporter = reporterFrom(identifier: options.reporter ?? config.reporter)
-        cache = options.ignoreCache ? nil : LinterCache(configuration: config)
+        if options.ignoreCache || ProcessInfo.processInfo.isLikelyXcodeCloudEnvironment {
+            cache = nil
+        } else {
+            cache = LinterCache(configuration: config)
+        }
         self.options = options
 
         if let outFile = options.output {


Source/SwiftLintFramework/Extensions/Configuration+Cache.swift
```diff
@@ -71,7 +71,14 @@ extension Configuration {
 
     internal var cacheURL: URL {
         let baseURL: URL
-        if let path = cachePath {
+
+        if ProcessInfo.processInfo.environment["CI"] == "TRUE" {
```
jpsim, line 75: CI is a super generic environment variable name. Is there a more reliable way to detect if this is running in an Xcode Cloud environment? I won't want everyone running SwiftLint in their CI to suddenly lose caching.

jpsim (changes requested):
Calling `NSTemporaryDirectory()` will return a new directory every time, effectively disabling caching of linter results. If we do something like this, it should be by disabling caching explicitly.

westerlund:
> Calling `NSTemporaryDirectory()` will return a new directory every time, effectively disabling caching of linter results. If we do something like this, it should be by disabling caching explicitly.

Yeah we thought of that and the problem is that we can't send any parameters to Xcode Plugins (as you know) and the only other solution would be to introduce another plugin target which explicitly disables caching, what do you think? Is it better to introduce a second SwiftLintPlugin or disable caching entirely only on Xcode Cloud?

jpsim:
> If not, then we should find the most reliable way to detect we're running on that platform and disable caching completely.

Here: https://github.com/realm/SwiftLint/blob/0.50.0-rc.3/Source/swiftlint/Commands/Lint.swift#L53

Could be

```swift
ignoreCache: noCache || ProcessInfo.processInfo.environment["SWIFTLINT_DISABLE_CACHE"] == "TRUE"
```

westerlund:
> Does Xcode Cloud have any way to persist state across CI jobs at all? If not, then we should find the most reliable way to detect we're running on that platform and disable caching completely.

In theory DerivedData will be cached between builds, however it seems like we don't have write access to that path.

> > If not, then we should find the most reliable way to detect we're running on that platform and disable caching completely.
> 
> Here: https://github.com/realm/SwiftLint/blob/0.50.0-rc.3/Source/swiftlint/Commands/Lint.swift#L53
> 
> Could be
> 
> ```swift
> ignoreCache: noCache || ProcessInfo.processInfo.environment["SWIFTLINT_DISABLE_CACHE"] == "TRUE"
> ```

I did some digging and it seems like Xcode Cloud is the only CI setting the variable `CI_BUNDLE_ID`, so expanding on your suggestion it could be something along the lines of:
```swift
ignoreCache: noCache || ProcessInfo.processInfo.environment["CI_BUNDLE_ID"] != nil
```
We'd much rather have this working automatically than manually setting an environment value on the Target itself, what do you think? Would it be reliable enough? At least until we get write permissions on the Xcode Cloud VM.

westerlund:
> `CI_BUNDLE_ID` is super generic too and possible that either some other CI or even user's dev machines have that set for some reason, and then they'd lose SwiftLint caching. So I'm not comfortable disabling caching based on the presence of that environment variable.
> 
> Xcode Cloud supports setting [custom environment variables](https://developer.apple.com/documentation/xcode/environment-variable-reference), so I'd like to propose that users define the `SWIFTLINT_DISABLE_CACHE` variable and we can check for it in SwiftLint.
> 
> > In addition to predefined environment variables, set custom environment variables in a workflow’s Environment section and access them in your custom build scripts or test actions.

Alright, I'll update the PR and also update the README to reflect this change. Thanks for the input!

jpsim:
You know what, looking at the [Xcode Cloud docs](https://developer.apple.com/documentation/xcode/environment-variable-reference) again, I think we can be reasonably sure that we're running on Xcode Cloud if all of the following environment variables are set:

```
CI
CI_BUILD_ID
CI_BUILD_NUMBER
CI_BUNDLE_ID
CI_COMMIT
CI_DERIVED_DATA_PATH
CI_PRODUCT
CI_PRODUCT_ID
CI_PRODUCT_PLATFORM
CI_PROJECT_FILE_PATH
CI_START_CONDITION
CI_TEAM_ID
CI_WORKFLOW
CI_WORKSPACE
CI_XCODE_PROJECT
CI_XCODE_SCHEME
CI_XCODEBUILD_ACTION
CI_XCODEBUILD_EXIT_CODE
```

So you could extract a function that determines if we're running on Xcode Cloud.

If someone or some other CI sets all the same environment variables, then I think it's reasonable to disable caching.

westerlund:
> You know what, looking at the [Xcode Cloud docs](https://developer.apple.com/documentation/xcode/environment-variable-reference) again, I think we can be reasonably sure that we're running on Xcode Cloud if all of the following environment variables are set:
> 
> ```
> CI
> CI_BUILD_ID
> CI_BUILD_NUMBER
> CI_BUNDLE_ID
> CI_COMMIT
> CI_DERIVED_DATA_PATH
> CI_PRODUCT
> CI_PRODUCT_ID
> CI_PRODUCT_PLATFORM
> CI_PROJECT_FILE_PATH
> CI_START_CONDITION
> CI_TEAM_ID
> CI_WORKFLOW
> CI_WORKSPACE
> CI_XCODE_PROJECT
> CI_XCODE_SCHEME
> CI_XCODEBUILD_ACTION
> CI_XCODEBUILD_EXIT_CODE
> ```
> 
> So you could extract a function that determines if we're running on Xcode Cloud.
> 
> If someone or some other CI sets all the same environment variables, then I think it's reasonable to disable caching.

Good idea, we'll go ahead and fix that. I'm glad we could come up with a solution that disables caching automatically , because then we can remove it later when the environment grants us write access!

Ignore my commits, I'll get back with the fix later!

I think it's reasonable to have a section in the README about caching not working on Xcode Cloud, what do you think? I also saw that the README is available in a few other languages than English, how do we properly update all of them?

jpsim:
> I think it's reasonable to have a section in the README about caching not working on Xcode Cloud, what do you think? I also saw that the README is available in a few other languages than English, how do we properly update all of them?

I don't think this needs a readme entry, and when we update the readme we typically only update the English one, the other ones are occasionally updated by other contributors who speak that language.

jpsim:
Looks good @JagCesar, thanks for the PR.

I made a few small modifications and will be merging once CI is 🟢

* Moved the check so it also applies when running `analyze`
* Updated `ProcessInfo.isLikelyXcodeCloudEnvironment` to only access `environment` once, in case there's any overhead involved, so it's not done repeatedly.

jpsim (approved):

PR #4488, opened 2022-10-28
Rewrite `optional_enum_case_matching` with SwiftSyntax

jpsim:


jpsim:
All the newly reported violations from OSSCheck appear to have previously been false negatives.

PR #4503, opened 2022-11-02
Add `Sequence.adjacentPairs()` from Swift Algorithms

jpsim:
To see if it improves performance over `zip(self, Array(dropFirst()) + [nil])`.


jpsim:
No meaningful improvement, if any.

PR #4504, opened 2022-11-02
Rewrite `prefer_self_in_static_references` with SwiftSyntax

SimplyDanny:
Fixes #4489.

This implementation ignores quite a few type references that could easily be found by the rule. As a first step, it should behave as before (while fixing the linked bug and some more false-positives/-negatives) even if this requires more work to purposefully ignore parts. In a next step, it can be enabled for constructor calls `MyType(...)`, key paths `\MyType.myProperty`, types in general like in `func f(inst: MyType) {...}` and other constructs.

PR #4505, opened 2022-11-02
Use more `Self`

SimplyDanny:


jpsim:
I think this makes the code less clear in many cases.

SimplyDanny:
> I think this makes the code less clear in many cases.

Hm ... I like it more actually. It shows that the type is nothing special but just what is currently implemented. So less reasoning about what it is and where it comes from.

But yeah, if this is not common sense we may disable the rule in the repository or generally exclude some cases from its implementation. The occurrences "fixed" in this PR are a preparation for #4504 since the SwiftSyntax-based rule finds some more (valid) cases (and could be extended even further).

Issue #4507, opened 2022-11-03
DiscouragedDirectInitRule should apply to NSError


jszumski:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

DiscouragedDirectInitRule should apply to `NSError()`.  Directly initializing `NSError` results in a runtime warning:
> -[NSError init] called; this results in an invalid NSError instance. It will raise an exception in a future release. Please call errorWithDomain:code:userInfo: or initWithDomain:code:userInfo:. This message shown only once.

##### Complete output when running SwiftLint, including the stack trace and command used

n/a

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? **0.49.1**
* Installation method used (Homebrew, CocoaPods, building from source, etc)? **Homebrew**
* Paste your configuration file: **uses defaults**

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)?
  If so, paste their relative paths and respective contents. **no**
* Which Xcode version are you using (check `xcodebuild -version`)? **Xcode 14.1**
* Do you have a sample that shows the issue? Run `echo "[string here]" | swiftlint lint --no-cache --use-stdin --enable-all-rules`
  to quickly test if your example is really demonstrating the issue. If your example is more
  complex, you can use `swiftlint lint --path [file here] --no-cache --enable-all-rules`.

```swift
let foo = NSError()
```

PR #4508, opened 2022-11-03
Adds `NSError` to the list of types in `discouraged_direct_init`

jszumski:
Fixes https://github.com/realm/SwiftLint/issues/4507.  

Using `NSError()` at runtime shows this warning and should be discouraged: 

> -[NSError init] called; this results in an invalid NSError instance. It will raise an exception in a future release. Please call errorWithDomain:code:userInfo: or initWithDomain:code:userInfo:. This message shown only once.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@
 
 #### Enhancements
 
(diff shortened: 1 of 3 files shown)


jszumski:
Updated `CHANGELOG.md` 👍

Issue #4510, opened 2022-11-03
`multiline_arguments_brackets` requires last tailing closure to contain a newline


CraigSiemens:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

It seems like the `multiline_arguments_brackets` assumes that the end of the `callBody` is still within the parenthesis for the arguments.

In the case where call contains a trailing closure, the rule will generate a false positive when the closure is a single line.

In the case where the call contains multiple trailing closure, the rule will generate a false positive when the last trailing closure is a single line. Earlier trailing closures can be one one or multiple lines and it wont affect the results.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
➜ swiftlint lint example.swift --no-cache --enable-all-rules 
Linting Swift files at paths example.swift
Linting 'example.swift' (1/1)
The `anyobject_protocol` rule is now deprecated and will be completely removed in a future release.
/Users/neo/Desktop/example.swift:3:25: warning: Multiline Arguments Brackets Violation: Multiline arguments should have their surrounding brackets in a new line. (multiline_arguments_brackets)
Done linting! Found 1 violation, 0 serious in 1 file.
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.50.0
* Installation method used (Homebrew, CocoaPods, building from source, etc)? home-brew
* Paste your configuration file: n/a
* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? No
* Which Xcode version are you using (check `xcodebuild -version`)? 14.1
* Do you have a sample that shows the issue? Yes

```zsh
echo "SomeType(\n    a: 1\n) { print("completion") }" | swiftlint lint --no-cache --use-stdin --enable-all-rules
```

```swift
SomeType(
    a: 1
) { print("completion") }
```

PR #4511, opened 2022-11-03
#4510 - Fix false-positives from `multiline_arguments_brackets`

CraigSiemens:
- #4510

Fixes an issue where `multiline_arguments_brackets` would give a false positive when the function is called with a trailing closure that was on a single line.

In that case, the `callBody` could be the string unto but not including the closing `}` for the closure. The `expectedBodyEndRegex` would then fail presumably because it was expecting the `callBody` to only go upto the `)`.


jpsim:
Closed in favor of #4512.

PR #4512, opened 2022-11-03
Rewrite `multiline_arguments_brackets` with SwiftSyntax

### LINKED ISSUE #4510: `multiline_arguments_brackets` requires last tailing closure to contain a newline

CraigSiemens:
While I was making #4511 I decided to try rewriting it using `SwiftSyntax` since the repo is moving all the rules that way anyway.

---

Between this PR and #4511, only one would need to be accepted. Both resolve #4510 which was my original goal.

This PR builds also contains the commit from the other one to get the added examples and changelog update.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Style/MultilineArgumentsBracketsRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/MultilineArgumentsBracketsRule.swift
@@ -1,7 +1,7 @@
 import Foundation
-import SourceKittenFramework
+import SwiftSyntax
 
-public struct MultilineArgumentsBracketsRule: ASTRule, OptInRule, ConfigurationProviderRule {
+public struct MultilineArgumentsBracketsRule: SwiftSyntaxRule, OptInRule, ConfigurationProviderRule {
     public var configuration = SeverityConfiguration(.warning)
 
     public init() {}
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Rules/Style/MultilineArgumentsBracketsRule.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
```
jpsim, line 1: This import isn't needed.

```suggestion
```

jpsim:
Thanks for the PR(s)! I do think we should go with this one since as you've said, we're aiming to move off of SourceKit entirely, so this definitely helps towards that goal.

I just removed an unused Foundation import, now I'll be reviewing your changes and if everything from OSSCheck looks good I'll merge.

jpsim (approved):
I reviewed the code and audited the OSSCheck report. Looks great!

PR #4518, opened 2022-11-05
Introduce `onlyElement` property on `Collection`

SimplyDanny:


Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
```diff
@@ -67,4 +67,11 @@ extension Collection {
     var isNotEmpty: Bool {
         return !isEmpty
     }
+
+    /// Get the only element in the collection.
+    ///
+    /// If the collection is empty or contains more than one element the result will be `nil`.
+    var singleElement: Element? {
```
jpsim, line 74: Nit: I think `onlyElement` might be a better name but I don’t feel strongly.

jpsim (approved):

Source/SwiftLintFramework/Extensions/Array+SwiftLint.swift
SimplyDanny: Agree, it's more on point.

PR #4520, opened 2022-11-06
Rewrite `balanced_xctest_lifecycle` with SwiftSyntax

marcelofabri:


marcelofabri:
There's a CI machine that has a full disk, but I ran `make analyze` and double checked that there are no violations, so merging this.

SimplyDanny:
> There's a CI machine that has a full disk, but I ran `make analyze` and double checked that there are no violations, so merging this.

What can we do about the full disk?

PR #4524, opened 2022-11-06
Run rules serially

jpsim:
When rules were mostly SourceKit-based, linting all rules concurrently on any given file had a performance boost.

Now that most rules are SwiftSyntax-based, it's possible this is no longer helping and in fact may be increasing thread/core contention or otherwise adding overhead.

Best way to know is to measure, so I'm doing that here.


jpsim:
At least right now, it doesn't seem to help in the general case. It does help in _some_ cases, like when linting the Wire repo, but that should be improved by https://github.com/apple/swift-syntax/pull/1053

PR #4530, opened 2022-11-08
Introduce a "rule registry" concept

jpsim:
This will allow for registering rules that aren't compiled as part of SwiftLintFramework.

Specifically this will allow us to split the built-in and extra rules into separate modules, leading to faster incremental compilation when working on rules since the rest of the framework won't need to be rebuilt on every compilation.


jpsim:
Marking as draft as I fix some test failures

jszumski (approved):
For easier review it might be better to separate out the `SwiftLintTestCase` introduction as a non-functional change and then build the registry on top of that.

jpsim:
> For easier review it might be better to separate out the `SwiftLintTestCase` introduction as a non-functional change and then build the registry on top of that.

I split the commits to make this easier to review for that reason.

Thanks for the review.

Tests/SwiftLintTestHelpers/SwiftLintTestCase.swift
```diff
@@ -0,0 +1,9 @@
+import SwiftLintFramework
+import XCTest
+
+open class SwiftLintTestCase: XCTestCase {
```
mildm8nnered, line 4: A lot of the unit test related rules will not fire for custom test case classes, unless they are configured to do so via the `test_parent_classes` parameter, so you may want to update `.swiftlint.yml` appropriately.

jpsim: Thanks, I configured this:

```yaml
single_test_class:
  test_parent_classes:
    - SwiftLintTestCase
    - XCTestCase
```

But there are other unit test rules that should be configured in this same way.

jpsim: YAML anchors to the rescue

```yaml
balanced_xctest_lifecycle: &unit_test_configuration
  test_parent_classes:
    - SwiftLintTestCase
    - XCTestCase
empty_xctest_method: *unit_test_configuration
single_test_class: *unit_test_configuration
```

mildm8nnered: nice - I did not know you could do that.

Issue #4531, opened 2022-11-08
Please provide linux/arm64 Binary Release


naraesk:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

Since v5.6 swift releases arm64 binaries (aarch64) for Linux/Ubuntu. The swiftlint binary for Linux only runs on amd64. For arm, I have to build swiftlint myself. This is of course possible, but it would be great if arm64 binaries would be provided. This would simplify the integration into arm-based docker containers and Github Actions.

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 5.7.1
* Installation method used (Homebrew, CocoaPods, building from source, etc)? building from source

jpsim:
Yes this would be great, and we're already using Docker Buildx so in theory it should be possible to update our CI to run that.

I tried a few months ago but the Swift 5.6.x Docker images didn't come in arm64 flavours at the time. Now I think the Swift 5.7 images do have an arm64 variant. I'll give this another shot.

naraesk:
Great, thanks for already working on this. :-)

To clarify, I'm primarily interested in an arm64 binary published on https://github.com/realm/SwiftLint/releases which I can integrate into my own docker container.

doozMen:
A grate blog post can be found https://www.polpiella.dev/binary-targets-in-modern-swift-packages

jpsim:
The Linux binaries in the release assets are generated in Docker, so that’s the first step.

Unfortunately it looks like with QEMU and the default GitHub Actions machines they’re not beefy enough to finish building both architectures, and the CI job appears to be getting killed.

doozMen:
Yes I have similar issues, ... shitty

doozMen:
I managed to fix it by manually building an artifactbundle as described here https://github.com/airbnb/swift/discussions/197#discussioncomment-4109195

Would it be possible for swiftlint to provide this in the future?

inket:
Sorry to hijack/bump. I noticed newer versions also don't have the Linux x86 binary. (We use SwiftLint in GitHub actions with the Linux runner as it's a lot cheaper than macOS runners.)

Is there anything I can help with?

If it's too complicated for the foreseeable future, may I suggest building the Linux versions in a separate GitHub action?

jpsim:
[...]

PR #4532, opened 2022-11-08
Publish a multi-platform Docker image

jpsim:
Addresses https://github.com/realm/SwiftLint/issues/4531

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,10 @@
 *
-!Plugins
+!.bazelrc
+!.bazelversion
+!bazel
+!tools/bazelw
+!BUILD
+!ci.bazelrc
+!MODULE.bazel
 !Source
-!Tests
-!Package.*
+!WORKSPACE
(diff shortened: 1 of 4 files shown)


jpsim:
I think this is finally working. I had to switch to building with Bazel with remote caching because otherwise the GitHub Actions CI job was taking too long (or too much memory) to build SwiftLint for arm64 with QEMU, and there's less static linking with the resulting binaries, but I think that's a worthwhile tradeoff.

keith (approved):

GNMoseke:
Is there anything stopping this from going in? Looks to be almost a year old but I would love the ability to use a multi-arch swiftlint image

jpsim:
> Is there anything stopping this from going in? Looks to be almost a year old but I would love the ability to use a multi-arch swiftlint image

The GitHub runners used to be too underpowered to successfully cross-compile for multiple platforms.

I tried leveraging Bazel and the remote build cache to reduce the amount of work done by the workers themselves, but ran into other issues. Then I lost interest in making this happen.

@GNMoseke please feel free to pick this up again!

Issue #4539, opened 2022-11-09
Malformed SwiftLint directives should generate an error or warning


mildm8nnered:
### New Issue Checklist

- [ x] Updated SwiftLint to the latest version
- [x ] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

I've just been reviewing our code base for "blanket disables" where an engineer disabled some rule in a stretch of code (or at the top of the file) and then "forgot" to enable it after the relevant stretch of code.

I came across quite a few typos where someone (possibly even me) had written

`// swiftlint:disaable some_rule`

or more commonly

`// swiftlint:enaable some_rule`

which kind of makes sense, because an un-disabled rule that's needed would likely be noticed straight away, but an un-reenabled rule could remain silently re-enabled forever.

It would be great if SwiftLint could print a warning when it sees `// swiftlint:something` where `something` is not a valid directive.

##### Complete output when running SwiftLint, including the stack trace and command used

N/A

### Environment

N/A

jpsim:
Great idea. This would be a great first issue if you felt inclined to contribute this to SwiftLint yourself.

mildm8nnered:
So I think I can see where I can detect these, in `Command` `init`, so something like (very hackily)

```
    public init?(actionString: String, line: Int, character: Int) {
        let scanner = Scanner(string: actionString)
        _ = scanner.scanString("swiftlint:")
        // (enable|disable)(:previous|:this|:next)
        guard let actionAndModifierString = scanner.scanUpToString(" ") else {
            queuedPrintError("Found a swiftlint comment directive without an action")
            return nil
        }
        let actionAndModifierScanner = Scanner(string: actionAndModifierString)
        guard let actionString = actionAndModifierScanner.scanUpToString(":") else {
            queuedPrintError("Found a swiftlint comment directive without an action")
            return nil
        }        
        guard let action = Action(rawValue: actionString) else {
            queuedPrintError("Found a swiftlint comment with an invalid action")
            return nil
        }
```

And I have the line number to use in the message.

What I couldn't work out was how to get the filename, as without that the error message is not that much use ("somewhere in your files there is a malformed directive ...).

That initialiser seems to get called from here:

```
private extension Trivia {
    func commands(offset: AbsolutePosition, locationConverter: SourceLocationConverter) -> [Command] {
        var triviaOffset = SourceLength.zero
        var results: [Command] = []
        for trivia in self {
[...]

PR #4543, opened 2022-11-11
Warn on malformed SwiftLint directives

mildm8nnered:
WIP, but comments welcome.

Detects malformed SwiftLint directives in source code. See https://github.com/realm/SwiftLint/issues/4539

For example

Screenshot 2022-11-11 at 23 19 29

This feels like it could actually be a rule, and that would make reporting easier. At the same time, we're parsing the directives anyway, so why do another separate pass.

Error messages and reporting - these aren't really violations per se (unless this should actually be a rule). More like the `warning: Found a configuration for 'balanced_xctest_lifecycle' rule, but it is not enabled on 'opt_in_rules'.` type messages.

But these issues have locations in source files, and it's really nice to have them show up in Xcode, so I just went with the Xcode format for right now.

I'm not sure if that is the right thing to do in all cases though - of course, if this was a rule, it could just use whatever reporter the user had configured.

### DIFF
--- a/Source/SwiftLintFramework/Models/Command.swift
+++ b/Source/SwiftLintFramework/Models/Command.swift
@@ -69,18 +69,22 @@ public struct Command: Equatable {
     /// - parameter line:         The line in the source file where this command is defined.
     /// - parameter character:    The character offset within the line in the source file where this command is
     ///                           defined.
-    public init?(actionString: String, line: Int, character: Int) {
+    public init?(actionString: String, file: String?, line: Int, character: Int) {
         let scanner = Scanner(string: actionString)
         _ = scanner.scanString("swiftlint:")
         // (enable|disable)(:previous|:this|:next)
         guard let actionAndModifierString = scanner.scanUpToString(" ") else {
+            queuedPrintError("Found a swiftlint directive with no action at \(file ?? "Unknown file"), line \(line)")
             return nil
         }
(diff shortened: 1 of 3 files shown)


mildm8nnered:
https://github.com/realm/SwiftLint/pull/4546 implements similar checks, but as a rule

0xWDG:
> #4546 implements similar checks, but as a rule

TBH I prefer this as a standard swiftlint option then a separate (opt in)rule.

mildm8nnered:
> > #4546 implements similar checks, but as a rule
> 
> TBH I prefer this as a standard swiftlint option then a separate (opt in)rule.

In the rule implementation in #4546 I made it enabled by default - would that cover your concerns? The rule implementation feels a bit cleaner to me, and makes the output format of the warnings more flexible.

0xWDG:
> > > #4546 implements similar checks, but as a rule
> > 
> > 
> > TBH I prefer this as a standard swiftlint option then a separate (opt in)rule.
> 
> In the rule implementation in #4546 I made it enabled by default - would that cover your concerns? The rule implementation feels a bit cleaner to me, and makes the output format of the warnings more flexible.

Sounds great 😊

mildm8nnered:
Closing this in favour of https://github.com/realm/SwiftLint/pull/4546

Issue #4540, opened 2022-11-09
Feature Request: Mechanism to enable all opt-in rules by default


mildm8nnered:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

I wasn't sure whether to file this as a new rule request or not. I guess an `all` "pseudo-rule" would do what I want.

I'd like to be able to enable all opt-in rules in my configuration file;

```
opt_in_rules:
  - all
```

would be fine (then I can disable the ones I don't want via `disabled_rules`).

This would save me having to check for new opt-in rules when I upgrade to a new version of SwiftLint

##### Complete output when running SwiftLint, including the stack trace and command used

N/A

### Environment

N/A

mildm8nnered:
See https://github.com/realm/SwiftLint/pull/4544

jpsim:
Have you tried the `—enable-all-rules` CLI flag? Does that do what you want?

By the way, we’ve historically been pretty liberal with allowing opt-in rules so I wouldn’t recommend actually keeping all rules enabled.

mildm8nnered:
So `--enable-all-rules` does not do what I want, because it ignores `disabled_rules:`, and in practice, there are almost no cases where a user would really want all rules enabled.

The intent here is to enable all opt-in rules by default, but to allow the user to disable any that they don't want.

Our internal policy (at my place of work) is to pretty much enable all rules (and all new rules), except any that we find annoying/disagree with the premise of/currently produce too many hits in our codebase (we operate a zero warnings policy).

When we upgrade SwiftLint, we use `swiftlint rules` to detect un-enabled opt-in rules, and decide whether to include them or not.

This is annoying because we can't tell from `rules` output whether an un-enabled rule is actually new, or whether it's an older rule that we evaluated previously and decided not to enable. As a hacky workaround for this, we actually list all opt-in rules in `opt_in_rules:`, and comment out those we don't want, but there's still a process of discovery and reconciliation to go through.

With `all`, any new opt-in rules in a new version of SwiftLint would be picked up automatically, with no need for us to discover them from `rules` output or the release notes (and our config would be a lot simpler).

We can then just run SwiftLint, and decide, based on any hits, whether to keep them (and patch any violations), or to explicitly disable them.

jpsim:
[...]

PR #4544, opened 2022-11-12
Added "all" pseudo-rule for opt in rules - enables all opt in rules

mildm8nnered:
Adds an "all" pseudo-rule - if specified in `opt_in_rules", this will enable all opt in rules.

For example: 

```
opt_in_rules:
  - all
```

Any other entries in `opt_in_rules` will be ignored (e.g. duplicates will not be reported).

See https://github.com/realm/SwiftLint/issues/4540

### DIFF
--- a/Source/SwiftLintFramework/Extensions/Configuration+RulesMode.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+RulesMode.swift
@@ -61,7 +61,7 @@ public extension Configuration {
                 self = .only(Set(onlyRules + analyzerRules))
             } else {
                 warnAboutDuplicates(in: disabledRules)
-                
+
                 let effectiveOptInRules: [String]
                 if optInRules.contains("all") {
                     effectiveOptInRules = primaryRuleList.list.compactMap { ruleID, ruleType in
@@ -70,7 +70,7 @@ public extension Configuration {
                 } else {
                     effectiveOptInRules = optInRules
                 }
-                                
+
                 warnAboutDuplicates(in: effectiveOptInRules + analyzerRules)
                 self = .default(disabled: Set(disabledRules), optIn: Set(effectiveOptInRules + analyzerRules))
             }
--- a/Source/SwiftLintFramework/Extensions/Configuration+RulesMode.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+RulesMode.swift
@@ -61,8 +61,18 @@ public extension Configuration {
                 self = .only(Set(onlyRules + analyzerRules))
             } else {
                 warnAboutDuplicates(in: disabledRules)
-                warnAboutDuplicates(in: optInRules + analyzerRules)
-                self = .default(disabled: Set(disabledRules), optIn: Set(optInRules + analyzerRules))
+                
+                let effectiveOptInRules: [String]
+                if optInRules.contains("all") {
+                    effectiveOptInRules = primaryRuleList.list.compactMap { ruleID, ruleType in
+                        ruleType is OptInRule.Type ? ruleID : nil
+                    }
+                } else {
+                    effectiveOptInRules = optInRules
+                }
+                                
+                warnAboutDuplicates(in: effectiveOptInRules + analyzerRules)
+                self = .default(disabled: Set(disabledRules), optIn: Set(effectiveOptInRules + analyzerRules))
             }
         }
 
--- a/Source/SwiftLintFramework/Extensions/Configuration+RulesMode.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+RulesMode.swift
@@ -65,7 +65,7 @@ public extension Configuration {
                 let effectiveOptInRules: [String]
                 if optInRules.contains("all") {
                     effectiveOptInRules = primaryRuleList.list.compactMap { ruleID, ruleType in
-                        ruleType is OptInRule.Type ? ruleID : nil
+                        ruleType is OptInRule.Type && !(ruleType is AnalyzerRule.Type) ? ruleID : nil
                     }
                 } else {
                     effectiveOptInRules = optInRules
--- a/Source/SwiftLintFramework/Extensions/Configuration+RulesMode.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+RulesMode.swift
@@ -63,7 +63,7 @@ public extension Configuration {
                 warnAboutDuplicates(in: disabledRules)
 
                 let effectiveOptInRules: [String]
-                if optInRules.contains("all") {
+                if optInRules.contains(RuleIdentifier.all.stringRepresentation) {
                     let allOptInRules = primaryRuleList.list.compactMap { ruleID, ruleType in
                         ruleType is OptInRule.Type && !(ruleType is AnalyzerRule.Type) ? ruleID : nil
                     }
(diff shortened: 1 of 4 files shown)


mildm8nnered:
test failing because ` IO failure on output stream: No space left on device`

mildm8nnered:
I added a CHANGELOG notice, and a little bit of documentation to the README. I did see a Chinese and Korean README as well - not sure how you handle updating these ...

Source/SwiftLintFramework/Extensions/Configuration+RulesMode.swift
```diff
@@ -61,8 +61,18 @@ public extension Configuration {
...
+                let effectiveOptInRules: [String]
+                if optInRules.contains("all") {
+                    effectiveOptInRules = primaryRuleList.list.compactMap { ruleID, ruleType in
+                        ruleType is OptInRule.Type ? ruleID : nil
+                    }
+                } else {
+                    effectiveOptInRules = optInRules
+                }
+
+                warnAboutDuplicates(in: effectiveOptInRules + analyzerRules)
```
SimplyDanny, line 74: What about Analyzer rules? Are they meant to be activated by `all`? If so the "warn about duplicates" should not include them as they have their own section. A warning would be confusing. Admittedly, I think we should not include them.

In case they are skipped the README should mention that.

SimplyDanny:
> I added a CHANGELOG notice, and a little bit of documentation to the README. I did see a Chinese and Korean README as well - not sure how you handle updating these ...

You may safely ignore them. Some attentive users familiar with the languages will come by at some point to translate and update the README.

Source/SwiftLintFramework/Extensions/Configuration+RulesMode.swift
mildm8nnered: Actually that did occur to me the other day when I saw your PR about analyzer rule handling - I'll check tomorrow and update appropriately

mildm8nnered: So I changed the code to not include analyser rules, but this leads to a weird edge case where the user has listed analyzer rules in `opt_in_rules`, but then `all` would enable all the opt in rules, but disable the explicitly listed analyser rules in the `opt_in_rules` section.

So I went with ignoring the analyser rules, but then taking the union of all the optin rules and the users explicit optin rules, so if the user explicitly listed them, they should remain enabled.

Updated the README as well, but not 100% happy with the language.

SimplyDanny: This semantic makes sense to me. Alternatively, since we have the warning about Analyzer rules listed in the `opt_in_rules` section, we could ignore them in the `all` case. The warning should be enough hint. I'm fine with both ways. Let @jpsim confirm.

mildm8nnered:
@SimplyDanny - I just spotted that I'm getting an `"all" is not a valid rule identifier` at runtime.

coming from `Configuration.validate(ruleIds: Set<String>, valid: Set<String>, silent: Bool = false) -> Set<String>`

I can suppress that by doing something like

`optInRuleIdentifiers = validate(ruleIds: optInRuleIdentifiers, valid: validRuleIdentifiers.union(["all"]))` 

at the two call sites relating to optin rules, but that feels a tiny bit hacky. Any suggestions?

SimplyDanny:
Adding `all` in all method calls is error prone. Either hard-code it in the method body or put it by default into `invalidRuleIdsWarnedAbout`. All not quite clean, but a special identifier requires special treatment. I currently have no better idea.

Thinking about this case, the `disabled_rules` section comes into my mind. We should also make sure that using `all` there leads to an error or a warning, because that wouldn't make any sense.

mildm8nnered:
> Adding `all` in all method calls is error prone. Either hard-code it in the method body or put it by default into `invalidRuleIdsWarnedAbout`. All not quite clean, but a special identifier requires special treatment. I currently have no better idea.
> 
> Thinking about this case, the `disabled_rules` section comes into my mind. We should also make sure that using `all` there leads to an error or a warning, because that wouldn't make any sense.

How about something like

```
        private func validateOptInRuleIdentifiers(ruleIds: Set<String>, valid: Set<String>, silent: Bool = false) -> Set<String> {
            validate(ruleIds: ruleIds, valid: ruleIds.union(["all"]), silent: silent)
        }
```

Source/SwiftLintFramework/Extensions/Configuration+RulesWrapper.swift
```diff
@@ -113,6 +113,10 @@ internal extension Configuration {
         }
 
         // MARK: - Methods: Validation
+        private func validateOptInRuleIdentifiers(ruleIds: Set<String>, valid: Set<String>, silent: Bool = false) -> Set<String> {
```
SimplyDanny, line 116:
```suggestion
        private func validate(optInRuleIds: Set<String>, valid: Set<String>, silent: Bool = false) -> Set<String> {
```

SimplyDanny:
> Thinking about this case, the disabled_rules section comes into my mind. We should also make sure that using all there leads to an error or a warning, because that wouldn't make any sense.

Did you have a chance to think about this part, @mildm8nnered?

mildm8nnered:
> > Thinking about this case, the disabled_rules section comes into my mind. We should also make sure that using all there leads to an error or a warning, because that wouldn't make any sense.
> 
> Did you have a chance to think about this part, @mildm8nnered?

 So I think my last commit [d88a5fb](https://github.com/realm/SwiftLint/pull/4544/commits/d88a5fb583f9f9ac6291c49c6e94402e59d16b14) (which I've applied your suggestion to since) should handle that, as I'm only ignoring `all` for optin rules, so `disabled_rules` should still get a warning.

SimplyDanny:
> > > Thinking about this case, the disabled_rules section comes into my mind. We should also make sure that using all there leads to an error or a warning, because that wouldn't make any sense.
> > 
> > 
> > Did you have a chance to think about this part, @mildm8nnered?
> 
> So I think my last commit [d88a5fb](https://github.com/realm/SwiftLint/pull/4544/commits/d88a5fb583f9f9ac6291c49c6e94402e59d16b14) (which I've applied your suggestion to since) should handle that, as I'm only ignoring `all` for optin rules, so `disabled_rules` should still get a warning.

I see. Makes sense to me now. So the PR looks good to me now (again). I would still like to give @jpsim the chance to intervene. So let's wait for his feedback.

mildm8nnered:
Is there some kind of general issue with the thread sanitiser checks?

I'm seeing similar failures on other PRs to the CI failures showing up on this one. For example, #4647 and #4657

SimplyDanny:
> Is there some kind of general issue with the thread sanitiser checks?

I first thought it was somehow caused by my change. When the second one failed too, it became clear that this must be a general issue. Now I see you have the same problem ...

mildm8nnered:
I seem to be hitting thread sanitiser issues again on this branch :-(

SimplyDanny:
The responsible job has been removed in #4701. Another rebase should fix your PR build.

mildm8nnered:
Thanks @SimplyDanny - that got me a clean build again.

@jpsim - any thoughts on merging this?

SimplyDanny:
Merged. 😊 Thanks again for all the work @mildm8nnered!

shuguenot:
Has this been tested with nested configurations?

mildm8nnered:
> Has this been tested with nested configurations?

yes - are you seeing any problems?

Issue #3769, opened 2021-11-30
[unused_declaration rule] `// swiftlint:enable unused_declaration` does not work


acecilia:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

When running the `unused_declaration` analyze rule over below code, there is no violation being reported:

```swift
// swiftlint:disable all
// swiftlint:enable unused_declaration
private func unusedPrivate() { }
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? `0.45.0`
* Installation method used (Homebrew, CocoaPods, building from source, etc)? `mint`
* Paste your configuration file:

```yml
analyzer_rules:
  - unused_declaration
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)?
  If so, paste their relative paths and respective contents. `No`
* Which Xcode version are you using (check `xcodebuild -version`)? `Xcode 12.5.1`
* Do you have a sample that shows the issue? Yes. Clone https://github.com/acecilia/swift-compilation-database and then do `make go_with_xcodebuild`. It is possible to observe that [here](https://github.com/acecilia/swift-compilation-database/blob/d9b237561e2d9f1232ac852e51ad762405b8548b/Sources/ModuleA/ModuleA1.swift#L12) there is no violation reported

mildm8nnered:
I think this is actually working as intended.

If I remove `// swiftlint:disable all`, the `unused_declaration` triggers for me.

Possible the confusion stems from the `// swiftlint:enable unused_declaration` - I believe SwiftLint will ignore this because of the `// swiftlint:disable all`. This will disable all rules until a matching `// swiftlint:enable all` is seen - rules can't be re-enabled one-by-one.

A couple of interesting oddities I came across while playing with this:

In the following code:

```
import Cocoa

// swiftlint:disable all
// swiftlint:enable unused_declaration
private func unusedPrivate() { }

@main
class AppDelegate: NSObject, NSApplicationDelegate {
    @IBOutlet var window: NSWindow!
    
    // swiftlint:disable:next nonexistent_rule
    let y = 0
    
    // swiftlint:disable:next force_try
    let z = 0
    
    // swiftlint:enable:next force_try
}
```

[...]

PR #4546, opened 2022-11-13
Adds invalid swiftlint command rule

mildm8nnered:
Rule-based approach to https://github.com/realm/SwiftLint/pull/4543

Adds an `invalid_swiftlint_command` rule to detect malformed `swiftlint` commands. Bungled `enable`'s in particular, after a valid `disable` can lead to stretches of code going unintentionally un-linted.

nonTriggeringExamples:

```
// swiftlint:disable unused_import
// swiftlint:enable unused_import
// swiftlint:disable:next unused_import
// swiftlint:disable:previous unused_import
// swiftlint:disable:this unused_import
```

triggeringExamples:

```
// swiftlint:
// swiftlint: 
// swiftlint::
// swiftlint:: 
// swiftlint:disable
// swiftlint:dissable unused_import
// swiftlint:enaaaable unused_import
// swiftlint:disable:nxt unused_import
// swiftlint:enable:prevus unused_import
// swiftlint:enable:ths unused_import
// swiftlint:enable
// swiftlint:enable:
// swiftlint:enable: 
// swiftlint:disable: unused_import
```

## Command Parsing

This PR also slightly changes the parsing of commands. Previously cases like

```
// swiftlint:disable: unused_import
// swiftlint:disable:nxt unused_import
// swiftlint:enable:prevus unused_import
// swiftlint:enable:ths unused_import
```

Would have been parsed, incorrectly, as

```
// swiftlint:disable unused_import
// swiftlint:disable unused_import
// swiftlint:enable unused_import
// swiftlint:enable unused_import
```

These cases will now be ignored, and will be warned about by this rule.

### DIFF
--- a/Source/SwiftLintFramework/Models/Command.swift
+++ b/Source/SwiftLintFramework/Models/Command.swift
@@ -8,6 +8,8 @@ public struct Command: Equatable {
         case enable
         /// The rule(s) associated with this command should be disabled by the SwiftLint engine.
         case disable
+        /// The action string was invalid
+        case invalid
 
         /// - returns: The inverse action that can cancel out the current action, restoring the SwifttLint engine's
         ///            state prior to the current action.
@@ -15,6 +17,7 @@ public struct Command: Equatable {
             switch self {
             case .enable: return .disable
             case .disable: return .enable
+            case .invalid: return .invalid
             }
         }
     }
@@ -27,6 +30,8 @@ public struct Command: Equatable {
         case this
         /// The command should only apply to the line following its definition.
         case next
+        /// The modifier string was invalid
+        case invalid
     }
 
     /// Text after this delimiter is not considered part of the rule.
@@ -35,6 +40,10 @@ public struct Command: Equatable {
     ///
     ///     swiftlint:disable:next force_try - Explanation here
     private static let commentDelimiter = " - "
+
+    var isValid: Bool {
+        action != .invalid && modifier != .invalid && !ruleIdentifiers.isEmpty
+    }
 
     internal let action: Action
     internal let ruleIdentifiers: Set<RuleIdentifier>
@@ -53,7 +62,7 @@ public struct Command: Equatable {
     ///                              defined.
     /// - parameter modifier:        This command's modifier, if any.
     /// - parameter trailingComment: The comment following this command's `-` delimiter, if any.
-    public init(action: Action, ruleIdentifiers: Set<RuleIdentifier>, line: Int = 0,
+    public init(action: Action, ruleIdentifiers: Set<RuleIdentifier> = [], line: Int = 0,
                 character: Int? = nil, modifier: Modifier? = nil, trailingComment: String? = nil) {
         self.action = action
         self.ruleIdentifiers = ruleIdentifiers
@@ -74,19 +83,19 @@ public struct Command: Equatable {
         _ = scanner.scanString("swiftlint:")
         // (enable|disable)(:previous|:this|:next)
         guard let actionAndModifierString = scanner.scanUpToString(" ") else {
-            return nil
+            self.init(action: .invalid, line: line, character: character)
+            return
         }
         let actionAndModifierScanner = Scanner(string: actionAndModifierString)
         guard let actionString = actionAndModifierScanner.scanUpToString(":"),
-            let action = Action(rawValue: actionString)
-            else {
-                return nil
+              let action = Action(rawValue: actionString)
+        else {
+            self.init(action: .invalid, line: line, character: character)
+            return
         }
-        self.action = action
-        self.line = line
-        self.character = character
 
         let rawRuleTexts = scanner.scanUpToString(Self.commentDelimiter) ?? ""
+        var trailingComment: String?
         if scanner.isAtEnd {
             trailingComment = nil
         } else {
@@ -103,23 +112,30 @@ public struct Command: Equatable {
             return component.isNotEmpty && component != "*/"
         }
 
-        ruleIdentifiers = Set(ruleTexts.map(RuleIdentifier.init(_:)))
-
-        guard !ruleIdentifiers.isEmpty else {
-            return nil
-        }
+        let ruleIdentifiers = Set(ruleTexts.map(RuleIdentifier.init(_:)))
 
         // Modifier
         let hasModifier = actionAndModifierScanner.scanString(":") != nil
+        let modifier: Modifier?
         if hasModifier {
             let modifierString = String(actionAndModifierScanner.string[actionAndModifierScanner.currentIndex...])
-            guard let modifier = Modifier(rawValue: modifierString) else {
-                return nil
+            if let modifierTemp = Modifier(rawValue: modifierString) {
+                modifier = modifierTemp
+            } else {
+                modifier = .invalid
             }
-            self.modifier = modifier
         } else {
             modifier = nil
         }
+
+        self.init(
+            action: action,
+            ruleIdentifiers: ruleIdentifiers,
+            line: line,
+            character: character,
+            modifier: modifier,
+            trailingComment: trailingComment
+        )
     }
 
     /// Expands the current command into its fully descriptive form without any modifiers.
@@ -146,6 +162,8 @@ public struct Command: Equatable {
                 Self(action: action, ruleIdentifiers: ruleIdentifiers, line: line + 1),
                 Self(action: action.inverse(), ruleIdentifiers: ruleIdentifiers, line: line + 1, character: Int.max)
             ]
+        case .invalid:
+            return []
         }
     }
 }
--- a/Source/SwiftLintFramework/Models/Command.swift
+++ b/Source/SwiftLintFramework/Models/Command.swift
@@ -119,11 +119,7 @@ public struct Command: Equatable {
         let modifier: Modifier?
         if hasModifier {
             let modifierString = String(actionAndModifierScanner.string[actionAndModifierScanner.currentIndex...])
-            if let modifierTemp = Modifier(rawValue: modifierString) {
-                modifier = modifierTemp
-            } else {
-                modifier = .invalid
-            }
+            modifier = Modifier(rawValue: modifierString) ?? .invalid
         } else {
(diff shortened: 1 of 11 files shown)


mildm8nnered:
So I managed to work out `make sourcery` to get the new rule into PrimaryRuleList, but I'm failing the generated tests because of the nature of the rule, it needs to be validated with 

```
    func testExamples() {
        verifyRule(InvalidSwiftLintCommandRule.description, skipCommentTests: true, skipDisableCommandTests: true)
    }
```

as I've done in `InvalidSwiftLintCommandRuleTests`, but there doesn't seem to be any way to pass these parameters into the generated tests, or to exclude that rule from the generated tests.

Not sure what the best way is to go from here ...

SimplyDanny:
> Not sure what the best way is to go from here ...

See for example `ExpiringTodoRule`. You can pass the options to skip certain tests to every example or an array of examples.

mildm8nnered:
So I guess I could special case this rule in  `.sourcery/GeneratedTests.stencil`, something like

```
@_spi(TestHelper)
@testable import SwiftLintFramework
import SwiftLintTestHelpers
import XCTest

// swiftlint:disable file_length single_test_class type_name

{% for rule in types.structs %}
{% if rule.name|hasSuffix:"Rule" and rule.name != "InvalidSwiftLintCommandRule" %}
class {{ rule.name }}GeneratedTests: XCTestCase {
    func testWithDefaultConfiguration() {
        verifyRule({{ rule.name }}.description)
    }
}
{% if not forloop.last %}

{% endif %}
{% endif %}
{% endfor %}
```

and rely on the handwritten test. 

Or I could further edit the stencil to generate the appropriate code for this case. That doesn't feel very graceful. I will also face a similar problem for https://github.com/realm/SwiftLint/issues/4684, so ideally any mechanism here would be easily extensible.

Source/SwiftLintFramework/Rules/Lint/InvalidSwiftLintCommandRule.swift
```diff
@@ -0,0 +1,126 @@
...
+            Example("// ↓swiftlint:enable:ths unused_import"),
+            Example("// ↓swiftlint:enable"),
+            Example("// ↓swiftlint:enable:"),
+            Example("// ↓swiftlint:enable: "),
+            Example("// ↓swiftlint:disable: unused_import")
+        ]
+    )
+
+    func makeVisitor(file: SwiftLintFile) -> ViolationsSyntaxVisitor {
+        Visitor(viewMode: .sourceAccurate)
```
SimplyDanny, line 40: Can we just use `file.commands` instead of a custom visitor?

mildm8nnered: I don't that will work, as we're looking for invalid commands, that would have failed parsing, so they won't end up in `file.commands`

SimplyDanny: See this small example test case:
```swift
class CommandVisitorTest: XCTestCase {
    func test() {
        let file = SwiftLintFile(contents: """
            // swiftlint:disable:next foo bar
        """)

        XCTAssertEqual(file.commands.map(\.ruleIdentifiers), [[.single(identifier: "foo"), .single(identifier: "bar")]])
    }
}
```
It shows that even unknown identifiers are parsed and collected.

mildm8nnered: So the `invalid_swiftlint_command` will not actually pick up invalid rule identifiers, but `superfluous_disable` already reports those:

`Superfluous Disable Command Violation: 'x' is not a valid SwiftLint rule. Please remove it from the disable command. (superfluous_disable_command)`

`invalid_swiftlint_command` picks up some cases that will fail parsing in the CommandVisitor, and never end up in `file.commands`

```
class CommandVisitorTest: XCTestCase {
    func testInvalidActions() {
        let file = SwiftLintFile(contents: """
            // swiftlint:dissable:next foo bar
            // swiftlint:
            // swiftlint:
            // swiftlint::
            // swiftlint:disable
            // swiftlint:dissable unused_import
            // swiftlint:enaaaable unused_import
            // swiftlint:enable
            """)
        XCTAssertTrue(file.commands.isEmpty)
    }
}
```

In fact, it seems really dodgy that something like

`// swiftlint:disable:nxt function_parameter_count`

end up in `file.commands`, as the invalid modifier will be ignored there, and then `function_parameter_count` will be disabled for the rest of the file, which is surely not what the user intended.

This rule will at least help flag those cases up to the user.

mildm8nnered: I've actually updated Command now, so that it's initialiser will return nil if the modifier is invalid, so 

`// swiftlint:disable:nxt unused_import` 

will have no effect, instead of disabling `unused_import` for the rest of the file. This seems more reasonable than the current behaviour, and makes the rule code a lot simpler, because previously I had a lot of code just for detecting and parsing the modifier to see if it was bad.

mildm8nnered:
```

class CommandVisitorTest: XCTestCase {
    func testInvalidActions() {
        let file = SwiftLintFile(contents: """
            // swiftlint:dissable:next foo bar
            // swiftlint:
            // swiftlint:
            // swiftlint::
            // swiftlint::
            // swiftlint:disable
            // swiftlint:dissable unused_import
            // swiftlint:enaaaable unused_import
            // swiftlint:enable
            // swiftlint:enable:
            // swiftlint:enable:
            // swiftlint:disable: unused_import
            // swiftlint:disable:nxt unused_import
            // swiftlint:enable:prevus unused_import
            // swiftlint:enable:ths unused_import
            """)
        XCTAssertTrue(file.commands.isEmpty)
    }
}
```

SimplyDanny: Okay, I got that. Thank you for the clarification.

I still think `CommandVisitor` should parse these commands once and for all and the newly added rule just evaluates all commands parsed from a file (`file.commands`). That can be achieved by extending `Command.Action` and `Command.Modifier` with `.invalid(String)` enum cases. We save another parsing pass and partial duplication of code in that way. It boils down to validating the available commands.

Does that make sense or am I missing something again?

mildm8nnered: Yep, I totally agree one pass would be better and even less code.

It does seem a bit mean to leave invalid commands in `file.commands` though, and make callers check to see if they are valid or not.

How about something like this: 

Parse all the commands in CommandVisitor, and valid or not (including those cases which would fail parsing today), store the results in the `commandsCache`, and give `Command` an `isValid` property (which can look at the modifier and action to see if they're valid). There's also a case where no rules are supplied, but I guess I can always check the number of ruleIdentifiers for that.

Define `commands` as follows:

`internal var commands: [Command] { commandsCache.get(self).filter { $0.isValid } }`

and add `invalidCommands` defined as

`internal var invalidCommands: [Command] { commandsCache.get(self).filter { !$0.isValid } }`

That way everything is parsed just once, but callers of `command` can reasonably expect that those commands are valid.

mildm8nnered: Something like this:

https://github.com/realm/SwiftLint/compare/main...mildm8nnered:SwiftLint:mildm8nnered-invalid-swiftlint-directives-rule-one-pass?expand=1

SimplyDanny: Yes, I like it more! Would you like to apply it to this PR for review?

mildm8nnered: Sure - just merged those changes into this branch!

mildm8nnered: Any thoughts on this @SimplyDanny ?

mildm8nnered:
See also my comment at `https://github.com/realm/SwiftLint/issues/3769#issuecomment-1445466113` - it looks like all kinds of odd strings, like `// hjkhjkkswiftlint:disable all` will be treated as valid, which seems off.

Source/SwiftLintFramework/Models/Command.swift
```diff
@@ -74,19 +83,19 @@ public struct Command: Equatable {
         _ = scanner.scanString("swiftlint:")
```
SimplyDanny, line 83: The initializer doesn't need to be optional (`init?`) anymore, right?

Source/SwiftLintFramework/Models/Command.swift
```diff
@@ -27,6 +30,8 @@ public struct Command: Equatable {
         case this
         /// The command should only apply to the line following its definition.
         case next
+        /// The modifier string was invalid
```
SimplyDanny, line 33: Nit: In documentation, prefered style is a period at the end of sentences/descriptions.

Source/SwiftLintFramework/Models/Command.swift
```diff
@@ -8,13 +8,16 @@ public struct Command: Equatable {
         case enable
         /// The rule(s) associated with this command should be disabled by the SwiftLint engine.
         case disable
+        /// The action string was invalid
```
SimplyDanny, line 11: Same here: Prefer a period at the end.

SimplyDanny (approved):
Sorry to miss your last comment. Apart from some smaller findings, it looks good to me now. 

I like that the rule itself is very easy now. Good work!

Source/SwiftLintFramework/Models/Command.swift
mildm8nnered: There should be a rule for that :-)

SimplyDanny:
Merged. 😊 Thanks for all the work @mildm8nnered!

PR #4563, opened 2022-11-20
Fix building with `swift build -c release`

jpsim:
Which previously tried to build the SwiftLintTestHelpers module, which is a test-only module.

This should fix building with precommit and Mint, both of which assume you can build the CLI targets in a Swift package with the `swift build -c release` command.


jpsim:
Needs a changelog entry.

PR #4575, opened 2022-11-22
Fix `lower_acl_than_parent` false positives

### LINKED ISSUE #4564: False positive `lower_acl_than_parent` in public extension
### New Issue Checklist

### Describe the bug

`lower_acl_than_parent` rule reports a false positive for members of types defined in `public extension`.

```bash
$ swiftlint --config swiftlint.yml lower_acl_than_parent.swift
Linting Swift files at paths lower_acl_than_parent.swift
Linting 'lower_acl_than_parent.swift' (1/1)
lower_acl_than_parent.swift:5:9: warning: Lower ACL than parent Violation: Ensure declarations have a lower access control level than their enclosing parent (lower_acl_than_parent)
```

### Environment

  0.50.0
  Homebrew

```yml
opt_in_rules:
  - lower_acl_than_parent
```

  No
  Xcode 14.1
  Build version 14B47b

```swift
public enum Namespace {}

public extension Namespace {
    struct Box {
        public let value: String // This line triggers a false positive
    }
}
```

HeEAaD:
Fix false positives in `lower_acl_than_parent` when the nominal parent is an extension.

The idea is to check for the nominal parent extension when the modifiers of the nominal parent don't contain a ACL modifier.

Fixes #4564

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@
   [#4559](https://github.com/realm/SwiftLint/issues/4559)
   [#4560](https://github.com/realm/SwiftLint/issues/4560)
 
+* Fix false positives in `lower_acl_than_parent` when the nominal parent is an extension.  
+  [Steffen Matthischke](https://github.com/heeaad)
+  [#4564](https://github.com/realm/SwiftLint/issues/4564)
+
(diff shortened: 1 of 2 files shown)


jpsim:
Thanks for the fix!

Issue #4597, opened 2022-11-27
False positive in rule `empty_enum_arguments`


HeEAaD:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

SwiftLint reports `empty_enum_arguments` violations when matching static methods with empty initializers arguments.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint lint --config example_swiftlint.yml --no-cache example.swift
Linting Swift files at paths example.swift
Linting 'example.swift' (1/1)
[...]/example.swift:5:35: warning: Empty Enum Arguments Violation: Arguments can be omitted when matching enums with associated values if they are not used. (empty_enum_arguments)
[...]/example.swift:7:36: warning: Empty Enum Arguments Violation: Arguments can be omitted when matching enums with associated values if they are not used. (empty_enum_arguments)
Done linting! Found 2 violations, 0 serious in 1 file.
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)?  
0.50.1
* Installation method used (Homebrew, CocoaPods, building from source, etc)?  
Mint
* Paste your configuration file:

```yml
only_rules:
    - empty_enum_arguments
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)?  
No
* Which Xcode version are you using (check `xcodebuild -version`)?  
Xcode 14.1
Build version 14B47b

* Do you have a sample that shows the issue? Run `echo "[string here]" | swiftlint lint --no-cache --use-stdin --enable-all-rules`
  to quickly test if your example is really demonstrating the issue. If your example is more
  complex, you can use `swiftlint lint --path [file here] --no-cache --enable-all-rules`.

```swift
import Foundation

func example(foo: FooEnum) {
    switch foo {
    case .aStaticMethod(from: Date()): // triggers empty_enum_arguments violation
        break
    case .aStaticMethod(from: .init()): // triggers empty_enum_arguments violation
        break
    case .aStaticMethod(from: Date(timeIntervalSinceNow: 0)):  // does not trigger empty_enum_arguments violation
        break
    case .case2:
        break
    default:
        break
    }
}

enum FooEnum {
    case case0
    case case1
    case case2
    static func aStaticMethod(from date: Date) -> Self {
        // Returns case based on `date`.
[...]

PR #4600, opened 2022-11-28
Fix empty_enum_arguments false positives

HeEAaD:
Fix false positive in `empty_enum_arguments` when the called expression is an identifier or an `.init()` call.

Fixes https://github.com/realm/SwiftLint/issues/4597

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,10 @@
 
 #### Bug Fixes
 
-* None.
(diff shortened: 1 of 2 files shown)


jpsim:
Thanks for the fix! Please add a changelog entry.

HeEAaD:
> Thanks for the fix! Please add a changelog entry.

added :)

PR #4603, opened 2022-11-29
Use a binary target for the build tool plugin

tonyarnold:
This will save users of the build tool plugin from having to compile the CLI tool before running the plugin. 

I'm still having a look at how to automate the updating of the binary details in the Package description.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -39,7 +39,7 @@ let package = Package(
             name: "SwiftLintPlugin",
             capability: .buildTool(),
             dependencies: [
-                .target(name: "swiftlint")
+                .target(name: "SwiftLintBinary")
             ]
         ),
         .executableTarget(
@@ -98,5 +98,10 @@ let package = Package(
                 "SwiftLintTestHelpers"
             ]
         ),
+        .binaryTarget(
+            name: "SwiftLintBinary",
+            url: "https://github.com/realm/SwiftLint/releases/download/0.50.1/SwiftLintBinary-macos.artifactbundle.zip",
+            checksum: "487c57b5a39b80d64a20a2d052312c3f5ff1a4ea28e3cf5556e43c5b9a184c0c"
+        )
     ]
 )


Package.swift
```diff
@@ -98,5 +98,10 @@ let package = Package(
                 "SwiftLintTestHelpers"
             ]
         ),
+        .binaryTarget(
+            name: "SwiftLintBinary",
+            url: "https://github.com/realm/SwiftLint/releases/download/0.50.1/SwiftLintBinary-macos.artifactbundle.zip",
+            checksum: "487c57b5a39b80d64a20a2d052312c3f5ff1a4ea28e3cf5556e43c5b9a184c0c"
```
jpsim, line 104: There’s a chicken and egg problem here. We won’t know the checksum until after the artifact is created.

tonyarnold: This is very, very correct. I need to do some reading, because this seems like a thing that Apple would have thought of. 

SwiftPM probably allows a local binary target at a path within the repository.

keith: Technically I guess you could build the binary, update this file, tag it, and still add the binary to the release tag even though it was built from 1 commit before the actual tag ¯\_(ツ)_/¯

tonyarnold: That's true. 

What's the release process look like now? Is it automated?

tonyarnold: Ah yep, I can see the "Releasing.md" file. 

I'll propose some updates to the Makefile to calculate the checksum and pipe it into the Package description as part of the release process.

jpsim: I hope you don't mind I pushed some commits to your branch in order to update the release workflow based on this conversation.

Let me know what you think.

tonyarnold: The additions look good to me, JP! 

Thanks for stepping in and taking care of the scripting 😊

jpsim: Could you please try pointing to the most recent commit on this branch (677a0a01bd510c35ac6428919bcc9711d25f57b6) to confirm that it works in your usage of SwiftPM plugins?

tonyarnold: I can see that it's selecting the correct tool in both Xcode and SwiftPM builds: 

```
working directory:
   /Users/tonyarnold/Developer/Itty Bitty Apps/AccessibilityExamples
tool mapping:
   swiftlint: /Users/tonyarnold/Library/Developer/Xcode/DerivedData/Accessibility_Examples-fzzfvojshuaewrausnzxfwtgijig/SourcePackages/artifacts/swiftlint/SwiftLintBinary.artifactbundle/swiftlint-0.50.1-macos/bin/swiftlint
tool search paths:
   /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
   /Applications/Xcode.app/Contents/Developer/usr/bin
   /bin
   /sbin
   /usr/bin
   /usr/sbin
```

I'm seeing issues when running the Xcode plugin (it stalls indefinitely at the "Apply build tool plugin" build phase), but that appears to be unrelated to the changes we've made here. I'll dig into that in a separate PR.

jpsim: ok I'll hold off on merging until you've looked into this.

tonyarnold: Okay, I've just pushed a few commits that resolve an infinite recursion bug for filesystems that don't have a configuration, and quietened the output so that Xcode only prints violations, not the checked files.

jpsim: There’s a lint failure due to a long line and a build failure on Linux. If you don’t get around to fixing those I can do that later.

tonyarnold: All done, let's see if it passes CI now 😉

tonyarnold: It's your call, but it might be worth cutting a release for that infinite recursion bug that I fixed. That's present in 0.50.1, as well.

jpsim: I’ll keep it in mind. Frequent releases means frequent cache invalidations, which impacts all users, not just the ones that would be impacted by this bug.

PR #4613, opened 2022-12-02
Don’t trigger shorthand_operator inside operator declaration

### LINKED ISSUE #4611: `shorthand_operator` should not trigger for forwarding operator implementations.
### New Issue Checklist

### Describe the bug

When implementing our own shorthand operators, it is common to implement it by delegating to non-shorthand operators, eg.
```swift
    public static func -= (lhs: inout Self, rhs: Int) {
        lhs = lhs - rhs
    }
```

This pattern should not trigger the linter, since the suggestion is non-operative (it would break the code by introducing a recursive loop), and the pattern is sufficiently common that a general exclusion is warranted.

```bash
$ swiftlint lint
error: Shorthand Operator Violation: Prefer shorthand operators (+=, -=, *=, /=) over doing the operation and assigning. (shorthand_operator)
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? `0.49.1`
* Installation method used (Homebrew, CocoaPods, building from source, etc)? `brew`
* Which Xcode version are you using (check `xcodebuild -version`)? `14.1 14B47b`

marcelofabri:
Fixes #4611

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,11 @@
   [JP Simard](https://github.com/jpsim)
   [#4558](https://github.com/realm/SwiftLint/issues/4558)
 
+* Don't trigger `shorthand_operator` violations inside a shorthand operator
+  function declaration.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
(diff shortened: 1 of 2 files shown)


jpsim:
Nice fix

PR #4616, opened 2022-12-03
Consider custom attributes in `attributes` rule

### LINKED ISSUE #4599: False positive for `attributes` rule when using `@MainActor`
### New Issue Checklist

### Describe the bug

It seems using `@MainActor` triggers a false positive depending on where `@MainActor` is exactly. See example below.

```bash
$ swiftlint lint
Linting 'test.swift' (1/1)
/Users/marco/Desktop/test/test.swift:5:7: warning: Attributes Violation: Attributes should be on their own lines in functions and types, but on the same line as variables and imports. (attributes)
```

### Environment

* SwiftLint version 0.50.1

```yml
opt_in_rules: 
  - attributes
```

  > No
  > Xcode 14.1
  > Build version 14B47b

```swift
// This triggers a violation:
import AppKit

@NSApplicationMain
@MainActor
final class AppDelegate: NSAppDelegate {}
```

The order of the attributes is important:
```swift
// This does NOT trigger a violation:
import AppKit

@MainActor
@NSApplicationMain
final class AppDelegate: NSAppDelegate {}
```

I also tried using `@main` instead of `@NSApplicationMain` with the same result.

Interestingly, using e.g. `@main` _and_ `@NSApplicationMain` but _without_ `@MainActor` does not trigger a warning, regardless of the order. So it seems like the problem is `@MainActor`.

SimplyDanny:
Fixes #4599.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -78,6 +78,9 @@ opt_in_rules:
   - xct_specific_matcher
   - yoda_condition
 
+attributes:
+  always_on_line_above:
+    - "@OptionGroup"
 identifier_name:
   excluded:
     - id
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,10 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#4645](https://github.com/realm/SwiftLint/issues/4645)
 
+* Consider custom attributes in `attributes` rule.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+  [#4599](https://github.com/realm/SwiftLint/issues/4599)
+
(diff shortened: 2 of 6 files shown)


.swiftlint.yml
```diff
@@ -77,6 +77,9 @@ opt_in_rules:
   - xct_specific_matcher
   - yoda_condition
 
+attributes:
+  always_on_line_above:
+    - "@OptionGroup"
```
SimplyDanny, line 82: Opinions on that? I think the list of command line arguments looks better if attributes are all above the variable declarations.

SimplyDanny: Fine for you, @jpsim?

Issue #4615, opened 2022-12-02
False positive `inert_defer` with condition compilation


yuri-qualtie:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

Rule `inert_defer` is false positive when defer is wrapped in condition compilation. For example: `#if os(macOS)`

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint Test.swift

Linting Swift files at paths Test.swift
Linting 'Test.swift' (1/1)
.../Test.swift:6:9: warning: Inert Defer Violation: If defer is at the end of its parent scope, it will be executed right where it is anyway. (inert_defer)
Done linting! Found 1 violation, 0 serious in 1 file.
```

### Environment

* SwiftLint version - `0.50.1`
* Installation method used  - Homebrew
* Paste your configuration file - default 

* Xcode version - `Xcode 14.0.1 Build version 14A400`
* Run - `swiftlint lint --path Test.swift --no-cache --enable-all-rules`
```
../Test.swift:4:5: warning: Explicit ACL Violation: All declarations should specify Access Control Level keywords explicitly. (explicit_acl)
../Test.swift:6:9: warning: Inert Defer Violation: If defer is at the end of its parent scope, it will be executed right where it is anyway. (inert_defer)
../Test.swift:3:8: warning: Missing Docs Violation: public declarations should be documented. (missing_docs)
../Test.swift:3:8: warning: Required Deinit Violation: Classes should have an explicit deinit method. (required_deinit)
```

```swift
public class Test {
    func hello(name: String) -> String {
        #if os(macOS)
        // this trigger violation: Inert Defer Violation: If defer is at the end of its parent scope, it will be executed right where it is anyway. (inert_defer)
        defer {
            print(self)
        }
        #endif

        print(name)

        return "Hello " + name
    }
}
```

stasswtf:
Also false positive:

```swift
let task = Task {
	#if DEBUG
		let stopwatch = Stopwatch()
		defer { Logger.default.debug("\(name ?? "Tool"): \(stopwatch).") }
	#endif
	// …
}
```

SimplyDanny:
The rule seems to be obsolete now given that the compiler warns already:
[...]

PR #4618, opened 2022-12-04
Deprecate `inert_defer` rule in favor of the Swift compiler warning

SimplyDanny:
See https://github.com/realm/SwiftLint/issues/4615#issuecomment-1336348568.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Lint/InertDeferRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/InertDeferRule.swift
@@ -1,6 +1,18 @@
 import SwiftSyntax
 
-struct InertDeferRule: ConfigurationProviderRule, SwiftSyntaxRule {
+private let warnDeprecatedOnceImpl: Void = {
+    queuedPrintError("""
+        The `\(InertDeferRule.description.identifier)` rule is now deprecated and will be completely \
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Rules/Lint/InertDeferRule.swift
```diff
@@ -1,5 +1,17 @@
 import SwiftSyntax
 
+private let warnDeprecatedOnceImpl: Void = {
+    queuedPrintError("""
+        The `\(InertDeferRule.description.identifier)` rule is now deprecated and will be completely \
```
marcelofabri, line 5: can you add that it was deprecated as it's a built-in Swift compiler?

Source/SwiftLintFramework/Rules/Lint/InertDeferRule.swift
```diff
@@ -1,5 +1,17 @@
...
+        removed in a future release.
+        """
+    )
+}()
+
+private func warnDeprecatedOnce() {
+    _ = warnDeprecatedOnceImpl
+}
+
 struct InertDeferRule: ConfigurationProviderRule, SwiftSyntaxRule {
```
marcelofabri, line 15: maybe we should make it an opt-in rule as part of the deprecation process

PR #4622, opened 2022-12-05
Build with `--swiftcopt=-warnings-as-errors` when using Bazel

jpsim:


jpsim:
Should pass once https://github.com/realm/SwiftLint/pull/4623 is merged and this is rebased.

PR #4624, opened 2022-12-06
Add rule for single space after period on comments

Julioacarrettoni:
Because even prose needs a style guide, and according to [owl.purdue.edu](https://owl.purdue.edu/owl/subject_specific_writing/journalism_and_journalistic_writing/ap_style.html#:~:text=million%2C%206%20cents.-,Punctuation,Use%20a%20single%20space%20after%20a%20period.,-Do%20not%20use) we should be using a single space after punctuation, I'm introducing `period_spacing`.

> Punctuation
> Use a single space after a period.

The rule simply checks and removes extra spaces after the first space following a period in `.blockComment`, `.docBlockComment`, `.lineComment`, and `.docLineComment`.

Per the [contributing guidelines](https://github.com/realm/SwiftLint/blob/4adabd8e4b0506e83937a128132eec292d8cc54a/CONTRIBUTING.md#tests) :
> SwiftLint supports building via Xcode and Swift Package Manager on macOS, and with Swift Package Manager on Linux. When contributing code changes, please ensure that all three supported build methods continue to work and pass tests.

- [x] $ xcodebuild -scheme swiftlint test -destination platform=macOS,arch=arm64,id=00006001-000411983E09801E
- [x] $ swift test
- [x] $ make docker_test (passes when applying the change from https://github.com/realm/SwiftLint/pull/4625)

This rule opt-in because it triggers in a lot of the linked repositories and it fits the definition of [opt-in-rules](https://github.com/realm/SwiftLint#opt-in-rules) nevertheless the rule does not trigger in this repository right now.

> A rule that is not general consensus or is only useful in some cases (e.g. force_unwrapping)

triggers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3315,6 +3315,11 @@ This is the last release to support building with Swift 4.0 and Swift 4.1.
 
 #### Enhancements
 
+* Add `period_spacing` opt-in rule that checks periods are not followed
+  by 2 or more spaces in comments.
+  [Julioacarrettoni](https://github.com/Julioacarrettoni)
+  [#4624](https://github.com/realm/SwiftLint/pull/4624)  s
+
 * Add optional filename verification to the `file_header` rule.
   All occurrences in the pattern of the `SWIFTLINT_CURRENT_FILENAME`
   placeholder are replaced by the name of the validated file.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,11 @@
   `single_test_class` and `empty_xctest_method` rules.  
   [Martin Redington](https://github.com/mildm8nnered)
   [#4200](https://github.com/realm/SwiftLint/issues/4200)
+  
+  * Add `period_spacing` opt-in rule that checks periods are not followed
+  by 2 or more spaces in comments.  
+  [Julioacarrettoni](https://github.com/Julioacarrettoni)
+  [#4624](https://github.com/realm/SwiftLint/pull/4624)
 
 #### Bug Fixes
 
@@ -3314,11 +3319,6 @@ This is the last release to support building with Swift 4.0 and Swift 4.1.
   [#1892](https://github.com/realm/SwiftLint/issues/1892)
 
 #### Enhancements
-
-* Add `period_spacing` opt-in rule that checks periods are not followed
-  by 2 or more spaces in comments.
-  [Julioacarrettoni](https://github.com/Julioacarrettoni)
-  [#4624](https://github.com/realm/SwiftLint/pull/4624)  s
 
 * Add optional filename verification to the `file_header` rule.
   All occurrences in the pattern of the `SWIFTLINT_CURRENT_FILENAME`
--- a/Source/SwiftLintFramework/Extensions/SyntaxClassification+isComment.swift
+++ b/Source/SwiftLintFramework/Extensions/SyntaxClassification+isComment.swift
@@ -1,0 +1,15 @@
+import IDEUtils
+
+extension SyntaxClassification {
+    // True if it is any kind of comment.
+    var isComment: Bool {
+        switch self {
+        case .lineComment, .docLineComment, .blockComment, .docBlockComment:
+            return true
+        case .none, .keyword, .identifier, .typeIdentifier, .operatorIdentifier, .dollarIdentifier, .integerLiteral,
+             .floatingLiteral, .stringLiteral, .stringInterpolationAnchor, .poundDirectiveKeyword, .buildConfigId,
+             .attribute, .objectLiteral, .editorPlaceholder:
+            return false
+        }
+    }
+}
--- a/Source/SwiftLintFramework/Rules/Lint/CommentSpacingRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/CommentSpacingRule.swift
@@ -122,6 +122,7 @@ struct CommentSpacingRule: SourceKitFreeRule, ConfigurationProviderRule, Substit
     func violationRanges(in file: SwiftLintFile) -> [NSRange] {
         // Find all comment tokens in the file and regex search them for violations
         file.syntaxClassifications
+            .filter(\.kind.isComment)
             .map { $0.range.toSourceKittenByteRange() }
             .compactMap { (range: ByteRange) -> [NSRange]? in
                 return file.stringView
--- a/Source/SwiftLintFramework/Rules/Lint/CommentSpacingRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/CommentSpacingRule.swift
@@ -122,15 +122,8 @@ struct CommentSpacingRule: SourceKitFreeRule, ConfigurationProviderRule, Substit
     func violationRanges(in file: SwiftLintFile) -> [NSRange] {
         // Find all comment tokens in the file and regex search them for violations
         file.syntaxClassifications
-            .compactMap { (classifiedRange: SyntaxClassifiedRange) -> [NSRange]? in
-                switch classifiedRange.kind {
-                case .blockComment, .docBlockComment, .lineComment, .docLineComment:
-                    break
-                default:
-                    return nil
-                }
-
-                let range = classifiedRange.range.toSourceKittenByteRange()
+            .map { $0.range.toSourceKittenByteRange() }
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -3315,6 +3315,11 @@ This is the last release to support building with Swift 4.0 and Swift 4.1.
 
 #### Enhancements
 
+* Add `period_spacing` opt-in rule that checks periods are not followed
```
jpsim, line 3318: Can you move this to the topmost section? This is in the release notes section for the 0.26 release from 2018 👴

Source/SwiftLintFramework/Rules/Lint/PeriodSpacingRule.swift
```diff
@@ -0,0 +1,100 @@
...
+                default:
+                    return nil
+                }
+
+                let range = classifiedRange.range.toSourceKittenByteRange()
+                return file.stringView
+                    .substringWithByteRange(range)
+                    .map(StringView.init)
+                    .map { commentBody in
+                        // Look for a period followed by two or more whitespaces but not new line or caret returns
```
jpsim, line 68:
```suggestion
                        // Look for a period followed by two or more whitespaces but not new line or carriage returns
```

Source/SwiftLintFramework/Rules/Lint/PeriodSpacingRule.swift
```diff
@@ -0,0 +1,100 @@
...
+            **/
+            """)
+        ],
+        triggeringExamples: [
+            Example("/* Only god knows why.  This symbol does nothing. */", testWrappingInComment: false),
+            Example("// Only god knows why.  This symbol does nothing.", testWrappingInComment: false),
+            Example("// Single. Double.  End.", testWrappingInComment: false),
+            Example("// Single. Double.  Triple.   End.", testWrappingInComment: false),
+            Example("// Triple.   Quad.    End.", testWrappingInComment: false),
+            Example("///   - code: Identifier of the error.  Integer.", testWrappingInComment: false)
```
SimplyDanny, line 40: Please also add violation markers `↓` to these triggering examples.

Source/SwiftLintFramework/Rules/Lint/PeriodSpacingRule.swift
```diff
@@ -0,0 +1,100 @@
...
+        ]
+    )
+
+    func violationRanges(in file: SwiftLintFile) -> [NSRange] {
+        // Find all comment tokens in the file and regex search them for violations
+        file.syntaxClassifications
+            .compactMap { (classifiedRange: SyntaxClassifiedRange) -> [NSRange]? in
+                switch classifiedRange.kind {
+                case .blockComment, .docBlockComment, .lineComment, .docLineComment:
+                    break
```
SimplyDanny, line 58: Control flow is a bit unusual here. Can we extract this into a function or an extension method on `SyntaxClassifiedRange` and use a `guard` or `if` statement here instead?

CHANGELOG.md
```diff
@@ -3315,6 +3315,11 @@ This is the last release to support building with Swift 4.0 and Swift 4.1.
...
+  by 2 or more spaces in comments.
```
SimplyDanny, line 3319: Funnily, entries in the CHANGELOG must end with 2 spaces so that they are separated from the links following in the rendered document:
```suggestion
  by 2 or more spaces in comments.  
```

SimplyDanny (changes requested):

Source/SwiftLintFramework/Rules/Lint/PeriodSpacingRule.swift
```diff
@@ -0,0 +1,100 @@
...
+            Example("// Why. ↓ Symbol does nothing."): Example("// Why. Symbol does nothing."),
+            Example("// Single. Double. ↓ End."): Example("// Single. Double. End."),
+            Example("// Single. Double. ↓ Triple. ↓  End."): Example("// Single. Double. Triple. End."),
+            Example("// Triple. ↓  Quad. ↓   End."): Example("// Triple. Quad. End."),
+            Example("///   - code: Identifier. ↓ Integer."): Example("///   - code: Identifier. Integer.")
+        ]
+    )
+
+    func violationRanges(in file: SwiftLintFile) -> [NSRange] {
+        // Find all comment tokens in the file and regex search them for violations
```
SimplyDanny, line 53: I'd love to see a rule enforcing periods to finalize comments. 😉 Can we have a period after each sentence to be consistent?

CHANGELOG.md
Julioacarrettoni: 🤦 Sure thing!

Source/SwiftLintFramework/Rules/Lint/PeriodSpacingRule.swift
Julioacarrettoni: Uhhhh! I like that!
Should it be a separate rule though?

CHANGELOG.md
Julioacarrettoni: Now that weird comment I got before adding the entry to the changelog makes way more sense.

Source/SwiftLintFramework/Rules/Lint/CommentSpacingRule.swift
```diff
@@ -122,15 +122,8 @@ struct CommentSpacingRule: SourceKitFreeRule, ConfigurationProviderRule, Substit
...
-            .compactMap { (classifiedRange: SyntaxClassifiedRange) -> [NSRange]? in
-                switch classifiedRange.kind {
-                case .blockComment, .docBlockComment, .lineComment, .docLineComment:
-                    break
-                default:
-                    return nil
-                }
-
-                let range = classifiedRange.range.toSourceKittenByteRange()
+            .map { $0.range.toSourceKittenByteRange() }
```
SimplyDanny, line 125: The `isComment` seems to be missing here.

Source/SwiftLintFramework/Rules/Lint/PeriodSpacingRule.swift
SimplyDanny: Well, I don't think such a rule would be feasible. It's hard to recognize full sentences. A comment can also be a link, a code snippet or a few words which don't result into a sentence.

My comment was rather meant as a joke referring to your comments in the new rule ... 😄

Source/SwiftLintFramework/Extensions/SyntaxClassification+isComment.swift
```diff
@@ -0,0 +1,15 @@
...
+
+extension SyntaxClassification {
+    // True if it is any kind of comment.
+    var isComment: Bool {
+        switch self {
+        case .lineComment, .docLineComment, .blockComment, .docBlockComment:
+            return true
+        case .none, .keyword, .identifier, .typeIdentifier, .operatorIdentifier, .dollarIdentifier, .integerLiteral,
+             .floatingLiteral, .stringLiteral, .stringInterpolationAnchor, .poundDirectiveKeyword, .buildConfigId,
+             .attribute, .objectLiteral, .editorPlaceholder:
```
SimplyDanny, line 11: Why the full set of cases here? To make sure the compiler warns if there are new kinds?

Source/SwiftLintFramework/Rules/Lint/PeriodSpacingRule.swift
Julioacarrettoni: I'm glad that's the case, I had a mini crisis trying to figure out the rule for that 😅

Source/SwiftLintFramework/Extensions/SyntaxClassification+isComment.swift
Julioacarrettoni: Exactly 💯

Source/SwiftLintFramework/Rules/Lint/CommentSpacingRule.swift
Julioacarrettoni: That's why I get for doing manual copy&paste.
Thank you so much for the thorough review!

I'm kinda surprised by the lack of false positives with this mistake. I mean, the list of examples is not exhaustive but the rule was also run on all those other repos right?

Julioacarrettoni:
So… what happens next? 🙃

PR #4625, opened 2022-12-06
Update docker_test to swift 5.7

Julioacarrettoni:
> macOS Ventura 13.0.1 (22A400)
> swift-driver version: 1.62.15 Apple Swift version 5.7.1 (swiftlang-5.7.1.135.3 clang-1400.0.29.51)
> Docker 4.15.0 (93002)

When running `make docker_test` it fails with
>'swiftlint': error: package 'swiftlint' is using Swift tools version 5.7.0 but the installed version is 5.6.3

Updating `docker_test` to use `swift:5.7-focal` fixes this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,7 +136,7 @@ docker_image:
 	docker build --platform linux/amd64 --force-rm --tag swiftlint .
 
 docker_test:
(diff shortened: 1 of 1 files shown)


marcelofabri (approved):
thanks!

PR #4642, opened 2022-12-13
fix false positives for enums and private functions with `redundant_objc_attribute` rule

### LINKED ISSUE #4633: `redundant_objc_attribute` false positives after updating 0.49.1->0.50.1
### New Issue Checklist

### Describe the bug

`redundant_objc_attribute` false positives after updating 0.49.1->0.50.1

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.50.1
* Installation method used (Homebrew, CocoaPods, building from source, etc)? pre-commit

```yml
# insert yaml contents here
```

  If so, paste their relative paths and respective contents. No
* Which Xcode version are you using (check `xcodebuild -version`)? 14.0.1

```swift
@objc public extension Foo {
  // This shouldn't trigger the warning
  @objc enum Bar: Int {
    case bar
  }
  // This will have compiler error without the @objc
  var bar: Bar { .bar }
  // This also triggers a warning and also causes compiler error but a better fix would be to mark other private interfaces @nonobjc
  @objc private enum Baz: Int {
    case baz
  }
  private var baz: Baz { .baz }
}
```

Cyberbeni:
fixes #4633

other examples of this rule might also need updating

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@
   [#4643](https://github.com/realm/SwiftLint/issues/4643)
 * Fix false positives for enums and private functions with
   `redundant_objc_attribute` rule.  
+* Fix false positives for enums, private functions and private stored variables
+  with `redundant_objc_attribute` rule.  
   [Cyberbeni](https://github.com/Cyberbeni)
   [#4633](https://github.com/realm/SwiftLint/issues/4633)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@
   subjects inside functions.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#4643](https://github.com/realm/SwiftLint/issues/4643)
+* Fix false positives for enums and private functions with
+  `redundant_objc_attribute` rule.  
+  [Cyberbeni](https://github.com/Cyberbeni)
+  [#4633](https://github.com/realm/SwiftLint/issues/4633)
 
 ## 0.50.3: Bundle of Towels
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,6 @@
   subjects inside functions.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#4643](https://github.com/realm/SwiftLint/issues/4643)
-* Fix false positives for enums and private functions with
-  `redundant_objc_attribute` rule.  
 * Fix false positives for enums, private functions and private stored variables
   with `redundant_objc_attribute` rule.  
   [Cyberbeni](https://github.com/Cyberbeni)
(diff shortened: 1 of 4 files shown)


Cyberbeni:
ah, private is also bad for stored property

Cyberbeni:
Should probably also also check for other types, not just enums: https://github.com/realm/SwiftLint/blob/449190d32406f7ae6aa9fdb7ecb2360c517e9483/Source/SwiftLintFramework/Rules/Idiomatic/RedundantObjcAttributeRule.swift#L66

jpsim:
Thanks for the PR!

> Should probably also also check for other types, not just enums:

Do you want to fix this in this PR or as a follow-up?

Cyberbeni:
I'll check what the compiler says about them tomorrow and fix them.

Cyberbeni:
seems like class doesn't need `@objc`, it just needs to be subclassed from NSObject (tested with declaring an `@objc` variable), same with typealias, haven't tried struct/actor, enum requires `@objc` (and also having an Int rawValue)

functions (for `#selector`) need `@objc` if private

variables (for NSObject's value(forKey:)) only need to be marked `@objc` if inside class declaration and private, stored or calculated doesn't matter

tests would probably still need some more fixing up but I won't have time for that anytime soon. for example some of them don't even have valid syntax, like `Example("private @GKInspectable var foo: String! {}"),`

mildm8nnered:
I've been playing around with this issue and then found this branch.

I think there are some problems here:

Consider the non-triggering example

```
@objcMembers
class Foo {
    class Bar: NSObject {
        @objc var foo: Any
    }
}
```

Actually, consider a slight variant, which will actually compile:

```
@objcMembers
class Foo: NSObject {
    class Bar: NSObject {
        @objc var foo: Any?
    }
}
```

This does not trigger, which is ok, but in the `-Swift.h` header, this will produce:

```
SWIFT_CLASS("_TtC12Sportscode643Foo")
@interface Foo : NSObject
- (nonnull instancetype)init OBJC_DESIGNATED_INITIALIZER;
@end
```

Now consider

```
@objcMembers
class Foo: NSObject {
    @objc
    class Bar: NSObject {
        @objc var foo: Any?
    }
}
```

this will trigger, but the `@objc` is required here to expose Bar to ObjC - this code produces

```
SWIFT_CLASS("_TtC12Sportscode643Foo")
@interface Foo : NSObject
- (nonnull instancetype)init OBJC_DESIGNATED_INITIALIZER;
@end


SWIFT_CLASS("_TtCC12Sportscode643Foo3Bar")
@interface Bar : NSObject
@property (nonatomic) id _Nullable foo;
- (nonnull instancetype)init OBJC_DESIGNATED_INITIALIZER;
@end
```

These were not errors before this branch. 

Similarly, the triggering example

```
        @objcMembers
        class Foo {
            ↓@objc var bar: Any?
            ↓@objc var foo: Any?
            ↓@objc
            class Bar {
                @objc
                var foo: Any?
            }
        }
```

I believe this should not trigger for the `@objc` before the class.

PR #4646, opened 2022-12-16
Fix false positive on `private_subject`

### LINKED ISSUE #4643: `private_subject` false positives in function scope
### New Issue Checklist

### Describe the bug

Starting in SwiftLint 0.50.0 the `private_subject` rule triggers on function local variables. This was not the case in 0.49.1 and earlier. 

```bash
$ swiftlint lint
```

### Environment

  * 0.50.3
  * CocoaPods

```yml
opt_in_rules:
  - private_subject
```

  * No
  * 14.1

```swift
func foo() {
    // This triggers a violation "Private Combine Subject Violation: Combine Subject should be private. (private_subject)":
    let goodSubject = PassthroughSubject<Bool, Never>(true)
}
```

marcelofabri:
Fixes #4643


Source/SwiftLintFramework/Rules/Lint/PrivateSubjectRuleExamples.swift
```diff
@@ -100,7 +100,14 @@ internal struct PrivateSubjectRuleExamples {
...
             """#
-        )
+        ),
+        Example(
+            """
+            func foo() {
+                let goodSubject = PassthroughSubject<Bool, Never>(true)
+            }
+            """
+        ),
```
SimplyDanny, line 110: This evil trailing comma is causing the test failures. 😬

SimplyDanny (approved):

Source/SwiftLintFramework/Rules/Lint/PrivateSubjectRuleExamples.swift
marcelofabri: muscle memory, on my daily job we require it 😅

PR #4655, opened 2022-12-22
Add `excluded_regular_expressions` option for identifier_name rule

kyounh12:
There already exists an option called `excluded` to prevent some unwanted checks.

However, `excluded` option has limitations when we need to exclude names with similar patterns since it requires exact match.

So we may solve this problem by adding `excluded_regular_expressions` option to `identifier_name` rule and exclude some names by regex.

This would be extremely helpful in situations like below:
* exclude identifier names with other languages. ( Xcode allows identifier names with some other languages )
* exclude identifier names having similar patterns
* ...

```
identifier_name:
  excluded_regular_expressions:
    - Apple
    - ^[가-힣_]$
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@
 * Separate analyzer rules as an independent section in the rule directory of the reference.  
   [Ethan Wong](https://github.com/GetToSet)
   [#4664](https://github.com/realm/SwiftLint/pull/4664)
+* Add `excluded_regular_expressions` option to `identifier_name` rule.  
+  [Moly](https://github.com/kyounh12)
+  [#4655](https://github.com/realm/SwiftLint/pull/4655)
 
 #### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@
   [#4664](https://github.com/realm/SwiftLint/pull/4664)
 * Add `excluded_regular_expressions` option to `identifier_name` rule.  
 * Interpret strings in `excluded` option of `identifier_name`, 
+
+* Interpret items in `excluded` option of `identifier_name`, 
   `type_name` and `generic_type_name` rules as regex.  
   [Moly](https://github.com/kyounh12)
   [#4655](https://github.com/realm/SwiftLint/pull/4655)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@
   [Ethan Wong](https://github.com/GetToSet)
   [#4664](https://github.com/realm/SwiftLint/pull/4664)
 * Add `excluded_regular_expressions` option to `identifier_name` rule.  
+* Interpret strings in `excluded` option of `identifier_name`, 
+  `type_name` and `generic_type_name` rules as regex.  
   [Moly](https://github.com/kyounh12)
   [#4655](https://github.com/realm/SwiftLint/pull/4655)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,10 +25,8 @@
 * Separate analyzer rules as an independent section in the rule directory of the reference.  
   [Ethan Wong](https://github.com/GetToSet)
   [#4664](https://github.com/realm/SwiftLint/pull/4664)
-* Add `excluded_regular_expressions` option to `identifier_name` rule.  
+  
 * Interpret strings in `excluded` option of `identifier_name`, 
-
-* Interpret items in `excluded` option of `identifier_name`, 
   `type_name` and `generic_type_name` rules as regex.  
   [Moly](https://github.com/kyounh12)
   [#4655](https://github.com/realm/SwiftLint/pull/4655)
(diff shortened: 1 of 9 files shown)


Source/SwiftLintFramework/Rules/Style/IdentifierNameRule.swift
```diff
@@ -38,6 +38,13 @@ struct IdentifierNameRule: ASTRule, ConfigurationProviderRule {
                 return []
             }
 
+            guard !configuration.excludedRegularExpressions.contains(where: {
+                guard let regex = try? NSRegularExpression(pattern: $0) else { return false }
```
SimplyDanny, line 42: Creating the regex should already be done when parsing the configuration so that it only happens once and not every time an identifier is checked.

SimplyDanny (changes requested):
I like the idea of supporting patterns for excluded names. However, the `NameConfiguration` already contains a few options and I would like to avoid adding a new one. The existing list of `excluded` names would only become a subset of the new `excluded_pattern` option.

What do you think about just interpreting the strings in the `excluded` list as regular expression that must match a whole name? If there is a list element `"MyName"` it would only match the name `MyName` but not `MyName_v1`. That makes sure that existing configurations work as before. With the possibility to add regular expressions to the list, one could add `"^MyName_v\d+$"` to match the other (versioned) name, too.

Furthermore, `NameConfiguration` is also used in `GenericTypeNameRule` and `TypeNameRule`. This should be tested as well.

Source/SwiftLintFramework/Rules/Style/IdentifierNameRule.swift
kyounh12: Thank you for your review and I totally agree with your idea to consider `excluded` list as regular expressions :)

I pushed some commits applying the changes, but I'm not pretty sure if interpreting `excluded` list as regex in this way would be clear to other people too. ( By adding `^` to the beginning and `$` to the end )

Source/SwiftLintFramework/Rules/Style/IdentifierNameRule.swift
```diff
@@ -34,7 +34,11 @@ struct IdentifierNameRule: ASTRule, ConfigurationProviderRule {
         }
 
         return validateName(dictionary: dictionary, kind: kind).map { name, offset in
-            guard !configuration.excluded.contains(name), let firstCharacter = name.first else {
+            guard let firstCharacter = name.first else { return [] }
+
+            guard !configuration.excludedRegularExpressions.contains(where: {
+                return !$0.matches(in: name, options: [], range: NSRange(name.startIndex..., in: name)).isEmpty
```
SimplyDanny, line 40: This should be a method in the configuration to avoid duplications.

CHANGELOG.md
```diff
@@ -10,7 +10,9 @@
 
 #### Enhancements
 
-* None.
+* Add `excluded_regular_expressions` option to `identifier_name` rule.
```
SimplyDanny, line 13: Please update the description.

Source/SwiftLintFramework/Rules/Style/IdentifierNameRule.swift
SimplyDanny: I cannot think of an example in which this would cause a problem. Up to now, only identifiers are supported. Wrapping them now into `^` and `$` and interpreting them as regular expressions should not cause any trouble even people are not aware of this.

Source/SwiftLintFramework/Rules/Style/IdentifierNameRule.swift
kyounh12: That would be much better, thanks :)

SimplyDanny (approved):

Source/SwiftLintFramework/Models/ExcludedRegexExpression.swift
```diff
@@ -0,0 +1,33 @@
+import Foundation
+
+/// Represents regex from `exclude` option in `NameConfiguration`.
+/// Using NSRegularExpression causes failure on equality check between two `NameConfiguration`s.
+/// This class compares pattern only when checking its equality
+public final class ExcludedRegexExpression: NSObject {
```
SimplyDanny, line 6: Could you please explain why this class is necessary? There also exist other configurations with `NSRegularExpression`-typed fields.

junseok-l2: I found `RuleConfigurationTests.testNameConfigurationSetsCorrectly` fails on CI machine when using `NSRegularExpression` type in `NameConfiguration`. ( It succeeds on my local machine )

The test compares two `NameConfiguration` with XCTAssertEqual and those configurations holds two different `NSRegularExpression` instances and causes the test to fail.

kyounh12: I found `RuleConfigurationTests.testNameConfigurationSetsCorrectly` fails on CI machine when using `NSRegularExpression` type in `NameConfiguration`. ( It succeeds on my local machine )

The test compares two `NameConfiguration` with XCTAssertEqual and those configurations holds two different `NSRegularExpression` instances and causes the test to fail.

SimplyDanny:
Well done! Thank you for all the work and your patience.

PR #4656, opened 2022-12-22
Deprecate `unused_capture_list` rule in favor of the Swift compiler warning

Cyberbeni:
similarly to #4618 

Tested all the triggering examples with Xcode 14.0.1


jpsim (approved):
Thanks!

Issue #3712, opened 2021-09-10
Compiler Directives makes opening brace rule violation missed in subsequent function


StevenGurrSky:
### New Issue Checklist

- [X] Updated SwiftLint to the latest version
- [X] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

If I have a function containing a compiler directive and a function below that should trigger the opening brace rule, it gets missed.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint lint
Linting Swift files in current working directory
Linting 'TestFile.swift' (1/1)
Done linting! Found 0 violations, 0 serious in 1 file.
```

### Environment

* SwiftLint version: 0.44.0
* Installation method used: Downloaded from the `portable_swiftlint.zip` from the 0.44.0 release tag
* Paste your configuration file: None/default/anything with `opening_brace` enabled

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? No
* Which Xcode version are you using? 12.5.1 (although not relevant)
* Do you have a sample that shows the issue?

```swift
class TestFile {
    func problemFunction() {
        #if DEBUG
        #endif
    }

    func openingBraceViolation()
    {
        print("Brackets")
    }
}
```

The `openingBraceViolation` function does not trigger the opening_brace rule as expected.

PR #4658, opened 2022-12-24
opening_brace rule bug fixes

mildm8nnered:
Fix for #3712

In the example given for the defect

```
             class TestFile {
                   func problemFunction() {
                       #if DEBUG
                       #endif
                   }
                   func openingBraceViolation()
                   {
                       print("Brackets")
                   }
               }
```

The `if` in `#endif` was being matched by the regex. I added in a word boundary (`\b`), to make sure that `if`, `guard` and `while` are only matched if they start at a word boundary


SimplyDanny (approved):
LGTM!

PR #4664, opened 2023-01-02
Separate analyzer rules as an independent section in the rule directory.

GetToSet:
PR #4620 gives warnings on misplaced analyzer rules. This PR separates analyzer rules as an independent section in the rule directory so that developers may be easier to recognize those rules should be separately treated.

### DIFF
--- a/Source/SwiftLintFramework/Documentation/RuleListDocumentation.swift
+++ b/Source/SwiftLintFramework/Documentation/RuleListDocumentation.swift
@@ -34,7 +34,8 @@ public struct RuleListDocumentation {
 
     private var indexContents: String {
         let defaultRuleDocumentations = ruleDocumentations.filter { !$0.isOptInRule }
-        let optInRuleDocumentations = ruleDocumentations.filter { $0.isOptInRule }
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Documentation/RuleListDocumentation.swift
```diff
@@ -33,8 +33,9 @@ public struct RuleListDocumentation {
     // MARK: - Private
 
     private var indexContents: String {
-        let defaultRuleDocumentations = ruleDocumentations.filter { !$0.isOptInRule }
-        let optInRuleDocumentations = ruleDocumentations.filter { $0.isOptInRule }
+        let defaultRuleDocumentations = ruleDocumentations.filter { !$0.isOptInRule && !$0.isAnalyzerRule }
```
SimplyDanny, line 36:
```suggestion
        let defaultRuleDocumentations = ruleDocumentations.filter { !$0.isOptInRule }
```
Analyzer rules are opt-in by default.

SimplyDanny (approved):
The separate section makes sense to me. Good suggestion!

SimplyDanny:
Thanks, @GetToSet!

PR #4662, opened 2022-12-31
Depend on swift-syntax prerelease tag

technocidal:
Updated the SwiftLint package to depend on the [swift-syntax prerelease](https://github.com/apple/swift-syntax/releases/tag/0.50800.0-SNAPSHOT-2022-12-20-a).

PR #4674, opened 2023-01-06
Pin SwiftSyntax to a tagged version

jpsim:
Specifically this tag: https://github.com/apple/swift-syntax/releases/tag/0.50800.0-SNAPSHOT-2022-12-29-a

This should make SwiftPM happier about the stability of the pinned version.


SimplyDanny:
See also https://github.com/realm/SwiftLint/pull/4662.

jpsim:
> See also #4662.

Thanks for the heads up @SimplyDanny , but that PR doesn't update the bazel dependency, nor does it update the GitHub Action and bash script to automatically update the SwiftSyntax version, which this PR now does.

SimplyDanny:
Sure thing. My comment was just intended to serve as a hint that there exists something similar already without any assessment of the two. 😉

SimplyDanny:
Can we be sure that these tags are going to be created on a regular basis for SwiftSyntax from now on?

As far as I have seen they only recently started to provide them.

I liked that we closely kept up with the development there by updating the dependency weekly.

SimplyDanny (approved):

jpsim:
> Can we be sure that these tags are going to be created on a regular basis for SwiftSyntax from now on?

No, I have no indication that more pre-release tags will come other than the two that were cut in the last 2 weeks (`2022-12-20-a` & `2022-12-29-a`).

> I liked that we closely kept up with the development there by updating the dependency weekly.

Yes, I like that too, but the SwiftPM plugin story was making relying on commits painful.

jpsim:
Hmmm, I need to find a way to update the resolved commit in `Package.resolved` 🤔.

Issue #4641, opened 2022-12-12
Cannot add SwiftLint as package for Xcode plugin


fmoraes74:
### New Issue Checklist

- [ ] Updated SwiftLint to the latest version
- [ ] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

I am trying to add SwiftLint as package dependency to run it as a build plugin. It fails resolution with invalid manifest.

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? Trying to pull latest
* Installation method used (Homebrew, CocoaPods, building from source, etc)? SPM

* Which Xcode version are you using (check `xcodebuild -version`)? 
 ```Xcode 14.1
Build version 14B47b```

jpsim:
I can look into this if you share more information, steps to reproduce, exact issue you're seeing, etc.

fmoraes74:
@jpsim This is what I see when I tried to add it:
<img width="1089" alt="image" src="https://user-images.githubusercontent.com/7808154/207132529-3503999d-77a6-44f8-856b-6765f116c6da.png">

Not tried anything more than just add the `SwiftLint` package dependency and it failed with the above error.

jpsim:
We’re not using BrightFuture or FileKit, so those errors are coming from something else.

fmoraes74:
This only happens when I add SwiftLint package. I even tried with the commit from the latest version:

<img width="804" alt="image" src="https://user-images.githubusercontent.com/7808154/207136782-1e611e7a-bcab-4bbb-b716-2c7b0161a8c2.png">
<img width="412" alt="image" src="https://user-images.githubusercontent.com/7808154/207136811-08847b1e-22a8-4d60-b488-5bba79f4aab1.png">

When I try to add it, I see it come up with version 0.9.2:

<img width="840" alt="image" src="https://user-images.githubusercontent.com/7808154/207136941-2d6708c2-3efe-4ad1-bbec-aae218fb5f50.png">

jpsim:
I think your SwiftPM caches may be out of date. Please clear those and try again.

fmoraes74:
I cleaned up the directory and retried, same issue, it still shows version 0.9.2. I also noticed the following:
<img width="731" alt="image" src="https://user-images.githubusercontent.com/7808154/207141235-09d55268-80aa-4b48-b851-ab4b416d9b15.png">

jpsim:
[...]

Issue #4628, opened 2022-12-07
SwiftPM plugin doesn't work when integrated in the build phases of an Xcode project that contains nested SwiftPM packages


nkristek:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

I have an Xcode project for an iOS app that contains nested SwiftPM packages. Until now SwiftLint was configured to run as a build phase that uses the installed version of the system, now I want to use the new SwiftPM plugin integration instead (to benefit from easier versioning control across multiple developers).

When using the SwiftLint plugin in the nested SwiftPM packages using the following code in the corresponding `Package.swift`:
```swift
dependencies: [
    .package(url: "https://github.com/realm/SwiftLint.git", branch: "main")
]
// ...
plugins: [
    .plugin(name: "SwiftLintPlugin", package: "SwiftLint")
]
```
the plugin works as intended.

However, when I try to integrate the plugin in the project (in order to lint the main app target) using the instructions [here](https://github.com/realm/SwiftLint#xcode-1) (adding the plugin in the `Run Build Tool Plug-ins` build phase), building the project fails with the error:
```
ld: entry point (_main) undefined. for architecture arm64
```

I'm building the project on an Apple Silicon based machine, but running Xcode under Rosetta2 results in the same error, but listing the Intel architecture instead. I've tried this in multiple codebases including a fresh iOS app (using the Xcode template) and I found out it always occurs once the project has a nested SwiftPM package.

#### Steps to reproduce

- Using Xcode 14.1
- Create a new project using the `App` template in the `iOS` category
  - Select `Storyboard` and `Swift` without Core Data or Tests
- Now we'll add the nested SwiftPM package
- Using `Finder` create a new folder called `NestedModule` inside the project folder
- In that folder create a new file called `Package.swift` with these contents:
```swift
// swift-tools-version:5.7

import PackageDescription

let package = Package(
    name: "NestedModule",
    platforms: [
        .iOS(.v14),
    ],
    products: [
        .library(name: "Nested", targets: ["Nested"])
    ],
    targets: [
        .target(name: "Nested")
    ]
)
```
- In the same folder create a new folder called `Sources`
- Inside `Sources` create a folder called `Nested`
- Inside `Nested` create an empty file called `File.swift` (to silence the warning about no source files)
- Drag the folder `NestedModule` into Xcode directly below the project
[...]

PR #4680, opened 2023-01-10
Use prebuildCommand for SwiftLint plugin

evfemist:
using `.prebuildCommand` instead `.buildCommand` fixes error `Entry point (_main) undefined. for architecture arm64` that was mention in [#4641](https://github.com/realm/SwiftLint/issues/4641) and [#4628](https://github.com/realm/SwiftLint/issues/4628)

Also in this PR I fix issue that cache files was added to bundle.


ckuburlis:
Works well for me. Without this I was getting some weird errors `Executable path is a Directory` and my app was trying to add swiftLint as a resource. This solved both of the issues for me.

ryanjfar:
Is it possible to have this Merged? it fixed the error I was getting from the Xcode build plugin as well, `Executable path is a Directory`

ApolloZhu:
This change also fixes missing "ProduceName-Swift.h" error in mixed Swift and Objective-C project

evfemist:
Merged with latest `main`, all checks have passed 😉

SimplyDanny (approved):

SimplyDanny:
Merged. Thank you @evfemist!

tonyarnold:
Sadly, this change makes SwiftLint unusable in my project - exclude paths are entirely ignored, and I'm seeing hundreds of thousands of errors and warnings from things that should be excluded. 

FYI: The issue with `Entry point (_main) undefined. for architecture arm64` is an Xcode bug that can be worked around using #4732 until the SwiftPM team have implemented additional API to handle how input files are consumed.

tonyarnold:
It looks like appending `--force-exclude` as an argument fixes exclude paths being ignored. I'll put together a PR for it now. 

I'm still not sure that `prebuildCommand` is the correct tool for this job, and the documentation for it makes mention of the performance impacts:

> Because prebuild commands are run on every build, they are can have significant performance impact and should only be used when there is no way to know the names of the outputs before the command is run.

Also, while my understanding was correct about the differences between the two approaches, given that we're not writing any "real" outputs, both of the approaches are going to run on every build. So there's no performance gain to be had from `buildCommand` at this time.

evfemist:
Hi @tonyarnold, looks like there are difference when `swiftlint` runs in Xcode and when in terminal.
I have tried to run same command as produced by plugin in terminal - and all works fine without `--force-exclude`

Issue #4637, opened 2022-12-11
number_separator rule changed 


Nef10:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

With https://github.com/realm/SwiftLint/pull/4583 the rule change - I think in an unindented way:

| Code | SwiftLint 0.49.0 | SwiftLint 0.50.0 | SwiftLint 0.50.1 |
| ------------- | ------------- | ------------- | ------------- |
| `Date(timeIntervalSince1970: 1497078000)` | error | error | ok |
| `Date(timeIntervalSince1970: 1_497_078_000)` | ok | ok | error |
| `#colorLiteral(red: 0.3543982506, green: 1.0, blue: 1.0, alpha: 1)` | ok | error | ok

I think @jpsim thought that it broke for all decimals (e.g. like line one behaved like line 3) - however the refactoring from 0.50.0 actually only seems to impact the parsing of the colorLiteral.

Additionally, I think the error on `Date(timeIntervalSince1970: 1_497_078_000)`  could be improved, it currently says `Underscores should be used as thousand separator in large decimal numbers. (number_separator)` - would be nice if that is negated if we enforce no `_` for decimals.

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? see above
* Installation method used (Homebrew, CocoaPods, building from source, etc)? portable from GitHub releases
* Paste your configuration file:

```yml
opt_in_rules:
  - number_separator
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? no
* Which Xcode version are you using (check `xcodebuild -version`)? 14.1
* Do you have a sample that shows the issue? Run `echo "[string here]" | swiftlint lint --no-cache --use-stdin --enable-all-rules` 
  to quickly test if your example is really demonstrating the issue. If your example is more
  complex, you can use `swiftlint lint --path [file here] --no-cache --enable-all-rules`. see above

SimplyDanny:
Checking with 0.50.1 without any configuration I get exactly the same results as in 0.49.0:
| Code | SwiftLint 0.50.1 |
| ------------- | ------------- |
| `Date(timeIntervalSince1970: 1497078000)` | error |
| `Date(timeIntervalSince1970: 1_497_078_000)` | ok |
[...]

PR #4685, opened 2023-01-11
Let `number_separator` rule trigger on misplaced separators

SimplyDanny:
Partial fix for #4637.

The `number_separator` rule now also triggers on misplaced separators with a different violation message.


SimplyDanny:
All reported (fixed) violations in OSS are due to the changed violation reason. There is not violation for a misplaced separator.

jpsim (approved):
Thanks for the improvements!

PR #4696, opened 2023-01-16
Add new `superfluous_else` rule

SimplyDanny:
This PR adds a new `superfluous_else` rule that triggers on `if`-statements when an attached `else`-block can be removed, because all branches of the previous `if`-block(s) would certainly exit the current scope already.

While auto-fixes are imaginable, comments make the rewrite functionality quite difficult. Apart from comments, it’s generally not easy to decrease the indentation of a complete code block. All spaces after a newline would have to be adjusted.

Therefore, the rule comes without automatic rewrite for now. This can be added at a later point in time perhaps even as a general algorithm performing indentations on entire code blocks.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Style/SuperfluousElseRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/SuperfluousElseRule.swift
@@ -104,20 +104,20 @@ struct SuperfluousElseRule: SwiftSyntaxRule, ConfigurationProviderRule, OptInRul
 private class Visitor: ViolationsSyntaxVisitor {
     override var skippableDeclarations: [DeclSyntaxProtocol.Type] { [ProtocolDeclSyntax.self] }
 
-    override func visitPost(_ node: IfStmtSyntax) {
+    override func visitPost(_ node: IfExprSyntax) {
         if node.violatesRule {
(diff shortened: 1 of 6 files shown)


SimplyDanny:
I've checked the visible OSS findings and all of them are reasonable.

Source/SwiftLintFramework/Rules/Style/SuperfluousElseRule.swift
```diff
@@ -0,0 +1,152 @@
...
+                        }
+                    }
+                } else ↓if i < 12 {
+                    return 2
+                } else ↓if i < 24 {
+                    return 8
+                } else {
+                    return 3
+                }
+            """)
```
1in1, line 95: Is it possible to make this work for `break` and `continue` statements in loops also?

SimplyDanny: Yes, it's planned to support these, too, in a follow-up extension to keep the number of OSS findings small.

SimplyDanny: FYI @1in1: #5425

PR #4699, opened 2023-01-17
[OSSCheck] Cache git repos

jpsim:
By maintaining persistent clones of the repos in `/tmp/SwiftLint-OSSCheck-Repos` which are incrementally synced instead of cloned from scratch on every run.


jpsim:
Seeing issues with this. Reverting.

PR #4704, opened 2023-01-18
Add support for bzlmod

keith:


jpsim:
This is looking great.

Can you please update `make release` in the Makefile to update the `MODULE.bazel` version like I did in https://github.com/jpsim/SourceKitten/pull/756?

What's the state of the rules_xcodeproj integration? In https://github.com/buildbuddy-io/rules_xcodeproj/issues/1598#issuecomment-1397176820 Brentley hints that this can't really be worked around in the current versions of Bazel.

keith:
added the make update. so rules_xcodeproj will work with 0.12.1 for in this repo, in the case you're pulling in sources from external repos it still needs more fixes

keith:
we can wait for that, but we can't add an archive_override because that isn't allowed for folks depending on this

keith:
ok updated to a version of rules_xcodeproj that works

jpsim:
Could you please rebase to pull in https://github.com/realm/SwiftLint/pull/4711?

MODULE.bazel
```diff
@@ -0,0 +1,26 @@
...
+    name = "swiftlint",
+    version = "0.50.3",
+    compatibility_level = 1,
+    repo_name = "SwiftLint",
+)
+
+bazel_dep(name = "platforms", version = "0.0.5")
+bazel_dep(name = "rules_apple", version = "2.0.0", repo_name = "build_bazel_rules_apple")
+bazel_dep(name = "rules_swift", version = "1.5.1", repo_name = "build_bazel_rules_swift")
+bazel_dep(name = "rules_xcodeproj", version = "0.12.2", repo_name = "com_github_buildbuddy_io_rules_xcodeproj")
```
jpsim, line 11: Can this be marked as a dev dependency?

jpsim, line 11: Do we have any benefits from renaming these repos? Can we just use the default names?

MODULE.bazel
```diff
@@ -0,0 +1,26 @@
...
+use_repo(
+    swiftlint_repos,
+    "com_github_apple_swift_syntax",
+    "com_github_johnsundell_collectionconcurrencykit",
+    "com_github_krzyzanowskim_cryptoswift",
+    "swiftlint_com_github_scottrhoyt_swifty_text_table",
+)
+
+extra_rules = use_extension("//bazel:extensions.bzl", "extra_rules")
+use_repo(extra_rules, "swiftlint_extra_rules")
```
jpsim, line 26: 😍

WORKSPACE
```diff
@@ -2,47 +2,6 @@ workspace(name = "SwiftLint")
...
-
-load("//bazel:repos.bzl", "swiftlint_repos")
-
-swiftlint_repos()
-
-load("//bazel:deps.bzl", "swiftlint_deps")
-
-swiftlint_deps()
-
 http_archive(
```
jpsim, line 5: Any value in keeping this in here?

jpsim:
Could you please update the bazel instructions in the readme? https://github.com/realm/SwiftLint#using-bazel

MODULE.bazel
keith: for folks still using the WORKSPACE on any names that would overlap i think we need to keep the old names, on the ones that are much more specifically named we could change it as long as we sync'd that with the workspace downloads, this was the path of least resistance

keith: AFAICT no because it's used in the root BUILD file, if it was in one that wasn't evaluated by downstream consumers then you could

WORKSPACE
keith: nice catch, nope

keith:
added this method to the readme too

jpsim (approved):
This is great, thank you.

keith:
Thanks!

Issue #4707, opened 2023-01-19
Expand emoji reporter to be more useful for humans reading it


levibostian:
I have a suggestion for a feature request. Open to suggestions on the solution, so let me jump into the problem 😄. 

Our Swift code is setup to run `swiftlint` on a git hook. When our git hooks run swiftlint, we use the emoji reporter since it's the reporter that is most human-readable. 

The emoji reporter output is quite friendly:
```
/code/Tests/Foo/IntegrationTest.swift
⛔️ Line 19: Lines should not have trailing whitespace.
⛔️ Line 19: Limit vertical whitespace to a single empty line. Currently 5.
```

The emoji reporter has a lot of great things going for it. 

However, these error messages of the reporter don't satisfy certain common use cases our team runs into. 

1. Sometimes it's helpful to view more information about a rule that was broken. Maybe I don't fully understand the error message "Lines should not have trailing whitespace" and an example would be helpful? 

The [swiftlint rule docs](https://realm.github.io/SwiftLint/file_length.html) are really good! When someone on my team gets confused about a lint rule, I point them to the official docs. 

A link to the docs from the CLI emoji reporter error message could be considered? 

2. Maybe I want to consider [disabling this rule for this 1 file in my code base](https://github.com/realm/SwiftLint#disable-rules-in-code)? 

If you want to [disable a rule for a particular file](https://github.com/realm/SwiftLint#disable-rules-in-code), you need to know the rule name (example: `file_length`). 

The emoji reporter error message could benefit from adding the name of the rule that was broken in the output? 

---

My workaround to do either of these 2 things is:
* Change the swiftlint reporter from "emoji" to "xcode", re-run swiftlint, see the rule name get printed to me. I then look up the docs or disable the rule in code from the rule name. 
* Change the reporter back to emoji. 

This workaround is annoying for me, but also something I need to share with new team members being onboarded to our team. 

Thanks for the time!

SimplyDanny:
Adding the rule identifier to the output can be done easily, see #4708.

A link to the docs is possible too, but would pollute the minimal output with lots of URLs. That doesn't sound attractive to me, admittedly. With the rule identifier, however, the command `swiftlint rules <id>` can be used to print the rule's description.

Also, `swiftlint docs` makes it quite easy to jump to the website. What I could think of is supporting `swiftlint docs <id>` to directly jump to the website specific to the rule. As of now, this is not possible.

levibostian:
[...]

PR #4710, opened 2023-01-19
Allow to pass a rule identifier to the `swiftlint docs` command

SimplyDanny:
See #4707.


SimplyDanny:
> Useful! Maybe if an unknown rule identifier is passed we can link to the rule directory? https://realm.github.io/SwiftLint/rule-directory.html

Good idea! I've added that.

PR #4713, opened 2023-01-20
Add bazel BCR config

keith:
This is based on sourcekitten's

### DIFF
--- a/.bcr/source.template.json
+++ b/.bcr/source.template.json
@@ -1,0 +1,5 @@
+{
+    "url": "https://github.com/realm/SwiftLint/releases/download/{TAG}/bazel.tar.gz",
+    "integrity": "",
+    "strip_prefix": ""
+}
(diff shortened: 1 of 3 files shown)


.bcr/source.template.json
```diff
@@ -0,0 +1,5 @@
+{
+    "url": "https://github.com/realm/SwiftLint/releases/download/{TAG}/bazel.tar.gz",
```
keith, line 2: I assume the bazel archive is the right one?

jpsim: Yes

jpsim (approved):

jpsim:
Oh I think I’ll need to add a fixed releaser so the BCR app uses my fork of the registry.

jpsim:
36c24c822c0944813ad841b18e3028d7ce17493f

PR #4717, opened 2023-01-22
Add new `direct_return` opt-in rule

SimplyDanny:


SimplyDanny:
I checked around 70% of the OSS findings and all were valid.

jpsim:
This is great @SimplyDanny! I've turned this on for Lyft's codebase and the consensus is that folks like it.

I ran into 3 kinds of minor issues with the rule:

## 1. Corrections "eats" some comments

For example this:

```swift
private func getNodes(bar: Bar) -> [(Node, Int)] {
    // This is an important comment
    let allNodes = bar.rootNodes
        .flatMap { $0.transform1() }
        .flatMap { $0.transform2() }
        .map { ($0, $0.someIntProp) }

    return allNodes
}
```

Corrects to this:

```swift
private func getNodes(bar: Bar) -> [(Node, Int)] {
    return bar.rootNodes
        .flatMap { $0.transform1() }
        .flatMap { $0.transform2() }
        .map { ($0, $0.someIntProp) }
}
```

Ideally it would keep the `// This is an important comment` comment.

## 2. `disable:next` commands aren't respected when correcting

This is probably a side-effect of 1, but this:

```swift
private func getNodes(bar: Bar) -> [(Node, Int)] {
    // swiftlint:disable:next direct_return - Local var needed to help the compiler type check the expression
    let allNodes = bar.rootNodes
        .flatMap { $0.transform1() }
        .flatMap { $0.transform2() }
        .map { ($0, $0.someIntProp) }

    return allNodes
}
```

Corrects to this:

```swift
private func getNodes(bar: Bar) -> [(Node, Int)] {
    return bar.rootNodes
        .flatMap { $0.transform1() }
        .flatMap { $0.transform2() }
        .map { ($0, $0.someIntProp) }
}
```

But really because of the disable comment, this shouldn't be corrected at all.

## 3. Some typechecking information is lost

For example this:

```swift
var values: [Foo] {
  let values: [Bar] = [.first, .second, .third]
  return values
}
```

Corrects to this:

```swift
var values: [Foo] {
  return [.first, .second, .third]
}
```

Which may or may not typecheck.

This typechecks though:

```swift
var values: [Foo] {
  return [.first, .second, .third] as [Bar]
}
```

SimplyDanny:
> This is great @SimplyDanny! I've turned this on for Lyft's codebase and the consensus is that folks like it.

Glad to hear that you like the new rule generally!

> I ran into 3 kinds of minor issues with the rule's correction logic:

I've addressed these issues in #4783. Correct handling of trivia is a really unpleasant task. 😬

jpsim:
> Correct handling of trivia is a really unpleasant task.

I agree. SwiftSyntax doesn't make it easy. I keep hoping I'm going to have a "lightbulb" moment at some point where it's going to click how to actually operate over whitespace elegantly with SwiftSyntax.

SimplyDanny:
> > Correct handling of trivia is a really unpleasant task.
> 
> 
> 
> I agree. SwiftSyntax doesn't make it easy. I keep hoping I'm going to have a "lightbulb" moment at some point where it's going to click how to actually operate over whitespace elegantly with SwiftSyntax.

Haha ... same for me. 😂 

It would probably make sense to have our own "trivia library" in SwiftLint. Something like "(un-)indent this list of statements" would be very useful to have, for example. But it would presumably feel like a set of heuristics that could potentially break with every new version of SwiftSyntax.

jpsim:
> It would probably make sense to have our own "trivia library" in SwiftLint.

I wonder if this would make sense in the SwiftSyntax repo itself.

To start we could start a discussion in the [Source Tooling section of the swift forums](https://forums.swift.org/c/development/tooling/39) about our challenges working with trivia using SwiftSyntax and if folks have recommendations, or if folks would be interested in having a "trivia library" in SwiftSyntax.

With SwiftSyntax powering the upcoming Swift macros features, I suspect there will be increasing interest and adoption of SwiftSyntax in the coming months, so we're probably far from the only folks thinking about this.

SimplyDanny:
The SwiftSyntax repo would definitely be the better place where such a library should belong.

Not sure if it should start with an initial implementation, though, to have something to talk about. For experimentation, the SwiftLint repo would be a good starting point.

But asking folks in the forum first is definitely a good idea. I may prepare a post when there is some spare time left.

PR #4731, opened 2023-01-29
Add blanket_disable_command rule

### LINKED ISSUE #4684: Rule Request: no_blanket_disables - disallow disabling SwiftLint rules for the entire file
### New Issue Checklist

- [x ] Updated SwiftLint to the latest version
- [ x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

Sometimes it's necessary to disable a SwiftLint rule in code, with the `// swiftlint:disable rule_name` directive, but engineers intentionally just put the disable at the top of, and applying to the, the entire file, when only a short stretch of code is relevant.

At other times, developers may intend to re-enable the SwiftLint rule, but forget to do so.

In either case, swiftlint rules can end up being disabled for stretches of code that we actually probably do want to lint.

This rule would warn on any cases of `// swiftlint:disable rule_name` that do not have a matching `enable` before the end of the file.

Triggering examples:

```
// swiftlint:disable some_rule
```

```
// swiftlint:disable some_rule some_other_rule
// swiftlint:enable some_rule
```

Non-triggering examples:

```
// swiftlint:disable some_rule
// swiftlint:enable some_rule
```

```
// swiftlint:disable some_rule some_other_rule
// swiftlint:enable some_rule
// swiftlint:enable some_other_rule
```

```
// swiftlint:disable:this some_rule
```

```
// swiftlint:disable:previous some_rule
```

```
// swiftlint:disable:next some_rule
```

For some rules, such as `file_length` it doesn't make sense to reenable them.

An `allowed_rules` parameter would specify rules that will be ignored by this rule.

opt-in, as users should never really want to disable rules for irrelevant stretches of code.

mildm8nnered:
Adds a new SwiftLint rule: `blanket_disable_command `

Resolves #4684 

This warns (by default) when a SwiftLint rule is disabled, and then never later re-enabled. In many cases this is because the developer forgot to re-enable the rule, or somehow bodged it (e.g. `// swftlint:enable some_rule`), resulting in stretches of code going unintentionally un-linted.

Violations of this rule are surprisingly common. I found a bunch in the SwiftLint source code, which I've patched in various ways in this PR - some could simply be turned into `// swiftlint:disable:next some_rule`'s and some I suppressed.

There are some rules that it doesn't make sense to re-enable, and that should apply to the whole file. An `allowed_rules` configuration parameter allows these to be specified. By default, `file_length` and `single_test_class` are allowed.

For these and possibly other rules, we may want to disable a particular rule always for the entire file, and to warn if it that rule is disabled with a modifier or re-enabled. The `always_blanket_disable` configuration parameter allows a set of rules to be specified that should always be warned on anything except a blanket disable. By default the set is empty.

This rule will also warn if you use `swiftlint:disable` to disable a rule that's already disabled, or `swiftlint:enable` to enable a rule that isn't disabled.

nonTriggeringExamples:

```
// swiftlint:disable unused_import
// swiftlint:enable unused_import
```

```
// swiftlint:disable unused_import unused_declaration
// swiftlint:enable unused_import
// swiftlint:enable unused_declaration
```

```
// swiftlint:disable:this unused_import")
```

```
// swiftlint:disable:next unused_import")
```

```
// swiftlint:disable:previous unused_import")
```

triggeringExamples:

```
// swiftlint:disable unused_import
```

```
// swiftlint:disable unused_import unused_declaration
// swiftlint:enable unused_import
```

```
// swiftlint:disable unused_import
// swiftlint:disable unused_import
// swiftlint:enable unused_import
```

```
// swiftlint:enable unused_import
```

See https://github.com/realm/SwiftLint/issues/4684


Source/SwiftLintFramework/Rules/Lint/NoBlanketDisablesRule.swift
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    private func validateAlwaysBlanketDisable(file: SwiftLintFile) -> [StyleViolation] {
+        var violations: [StyleViolation] = []
+
+        guard configuration.alwaysBlanketDisableRuleIdentifiers.isEmpty == false else {
+            return []
+        }
+
+        for command in file.commands where command.modifier != nil && command.action == .enable {
```
SimplyDanny, line 116: Is `command.action == .enable` correct here?

Are there any tests for this kind of validation?

SimplyDanny:
In general, I like this rule a lot! The OSS findings and the findings in SwiftLint itself splendidly show how much SwiftLint misses it.

Source/SwiftLintFramework/Rules/Lint/NoBlanketDisablesRule.swift
mildm8nnered: So `command.action == enable` was correct, but `command.modifier != nil` was not :-( . The intent here is that if we specify `always_blanket_disable` for a rule, then the only valid command is `// swiftlint:disable some_rule` - enable, or using a modifier will trigger.

I've improved the reason's to make this clearer (and I wasn't warning on modifiers before, but I think it makes sense - `// swiftlint:disable:next file_length` is almost certainly not intended).

I also added a unit test to cover the `always_blanket_disable` functionality.

mildm8nnered:
> In general, I like this rule a lot! The OSS findings and the findings in SwiftLint itself splendidly show how much SwiftLint misses it.

Thanks! I just saw so many people just disabling stuff for the remainder of the file to fix one warning that it got annoying - we  have a custom Danger script to do (some of) this, but much nicer to just have it baked in. And yes, it does seem to get a lot of hits :-)

Not entirely sure that `no_blanket_disable` is the best naming. Possibly `blanket_disable_command`, to echo `superfluous_disable_command`?

Source/SwiftLintFramework/Rules/Lint/NoBlanketDisablesRule.swift
```diff
@@ -0,0 +1,135 @@
...
+                violations.append(contentsOf: alreadyDisabledRuleIdentifiers.map {
+                    let reason = "The disabled '\($0.stringRepresentation)' SwiftLint rule was already disabled"
+                    return violation(forPath: file.file.path, command: command, reason: reason)
+                })
+            }
+
+            if command.action == .enable {
+                let notDisabledRuleIdentifiers = command.ruleIdentifiers.subtracting(disabledRuleIdentifiers)
+                violations.append(contentsOf: notDisabledRuleIdentifiers.map {
+                    let reason = "The enabled '\($0.stringRepresentation)' SwiftLint rule was not disabled"
```
SimplyDanny, line 63:
```suggestion
                    let reason = "The enabled '\($0.stringRepresentation)' rule was not disabled"
```

Source/SwiftLintFramework/Rules/Lint/NoBlanketDisablesRule.swift
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        let allowedRuleIdentifiers = configuration.allowedRuleIdentifiers
+        for disabledRuleIdentifier in disabledRuleIdentifiers {
+            if allowedRuleIdentifiers.contains(disabledRuleIdentifier.stringRepresentation) {
+                continue
+            }
+
+            if let command = ruleIdentifierToCommandMap[disabledRuleIdentifier] {
+                let reason = "The disabled '\(disabledRuleIdentifier.stringRepresentation)' SwiftLint rule " +
```
SimplyDanny, line 89:
```suggestion
                let reason = "The disabled '\(disabledRuleIdentifier.stringRepresentation)' rule " +
```

Source/SwiftLintFramework/Rules/Lint/NoBlanketDisablesRule.swift
```diff
@@ -0,0 +1,135 @@
...
+            return []
+        }
+
+        for command in file.commands {
+            let ruleIdentifiers: Set<String> = Set(command.ruleIdentifiers.map { $0.stringRepresentation })
+            let intersection = ruleIdentifiers.intersection(configuration.alwaysBlanketDisableRuleIdentifiers)
+            if command.action == .enable {
+                violations.append(contentsOf: intersection.map {
+                    let reason = "The '\($0)' SwiftLint rule should be disabled once for the whole file. " +
+                                 "You don't need to re-enable it"
```
SimplyDanny, line 122:
```suggestion
                    let reason = "The '\($0)' rule applies to the whole file and thus doesn't need to be re-enabled"
```

Source/SwiftLintFramework/Rules/Lint/NoBlanketDisablesRule.swift
```diff
@@ -0,0 +1,135 @@
...
+                    return violation(forPath: file.file.path, command: command, reason: reason)
+                })
+            } else if command.modifier != nil {
+                violations.append(contentsOf: intersection.map {
+                    let reason = "Do not use 'previous', 'this' or 'next' when disabling the '\($0)' SwiftLint rule"
```
SimplyDanny, line 127:
```suggestion
                    let reason = "The '\($0)' rule applies to the whole file and thus cannot be disabled locally with 'previous', 'this' or 'next'"
```

SimplyDanny (commented):
Indeed, `blanket_disable_command` sounds better to me. Also, the wording of the violation messages can be improved. I made a few suggestions. Please let me know what you think.

mildm8nnered:
> Indeed, `blanket_disable_command` sounds better to me. Also, the wording of the violation messages can be improved. I made a few suggestions. Please let me know what you think.

Cool - I went with `blanket_disable_command` and all your reason suggestions.

Source/SwiftLintFramework/Rules/Lint/NoBlanketDisablesRule.swift
```diff
@@ -0,0 +1,135 @@
...
+            Example("""
+            // swiftlint:disable unused_import unused_declaration
+            // swiftlint:enable unused_import
+            // swiftlint:enable unused_declaration
+            """),
+            Example("// swiftlint:disable:this unused_import"),
+            Example("// swiftlint:disable:next unused_import"),
+            Example("// swiftlint:disable:previous unused_import")
+        ],
+        triggeringExamples: [
```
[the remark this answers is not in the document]
mildm8nnered: I did play with that before, but ran into some issues because the Command location was not where I expected it to be. I'll have another crack at it.

SimplyDanny (approved):

Source/SwiftLintFramework/Rules/Lint/BlanketDisableCommandRule.swift
```diff
@@ -0,0 +1,143 @@
...
+            // swiftlint:enable unused_declaration
+            """),
+            Example("// swiftlint:disable:this unused_import"),
+            Example("// swiftlint:disable:next unused_import"),
+            Example("// swiftlint:disable:previous unused_import")
+        ],
+        triggeringExamples: [
+            Example("// ↓swiftlint:disable unused_import"),
+            Example("""
+            // ↓swiftlint:disable unused_import unused_declaration
```
SimplyDanny, line 32: Would it be possible to have the marker at the identifier like
```suggestion
            // swiftlint:disable unused_import ↓unused_declaration
```
so I don't have to read the complete violation message and extract the identifier name from there.

Source/SwiftLintFramework/Rules/Lint/BlanketDisableCommandRule.swift
```diff
@@ -0,0 +1,143 @@
+import SwiftSyntax
+
+struct BlanketDisableCommandRule: ConfigurationProviderRule {
+    typealias ConfigurationType = BlanketDisableCommandConfiguration
```
SimplyDanny, line 4: The alias is not needed.

Source/SwiftLintFramework/Rules/Lint/BlanketDisableCommandRule.swift
```diff
@@ -0,0 +1,143 @@
...
+            """),
+            Example("""
+            // swiftlint:disable unused_import
+            // ↓swiftlint:disable unused_import
+            // swiftlint:enable unused_import
+            """),
+            Example("""
+            // ↓swiftlint:enable unused_import
+            """)
+        ].skipWrappingInCommentTests().skipDisableCommandTests()
```
SimplyDanny, line 43: Can the `.skipWrappingInCommentTests()` be removed with your fixes in #4799?

Source/SwiftLintFramework/Rules/RuleConfigurations/BlanketDisableCommandConfiguration.swift
```diff
@@ -0,0 +1,33 @@
+public struct BlanketDisableCommandConfiguration: SeverityBasedRuleConfiguration, Equatable {
+    public private(set) var severityConfiguration = SeverityConfiguration(.warning)
+    public private(set) var allowedRuleIdentifiers: Set<String> = ["file_length", "single_test_class"]
+    public private(set) var alwaysBlanketDisableRuleIdentifiers: Set<String> = []
```
SimplyDanny, line 4: Isn't it the other way around?
```suggestion
    public private(set) var allowedRuleIdentifiers: Set<String> = []
    public private(set) var alwaysBlanketDisableRuleIdentifiers: Set<String> = ["file_length", "single_test_class"]
```

Source/SwiftLintFramework/Rules/Lint/BlanketDisableCommandRule.swift
```diff
@@ -0,0 +1,143 @@
...
+        var ruleIdentifierToCommandMap: [RuleIdentifier: Command] = [:]
+        var disabledRuleIdentifiers: Set<RuleIdentifier> = []
+
+        for command in file.commands {
+            if command.action == .disable {
+                let alreadyDisabledRuleIdentifiers = command.ruleIdentifiers.intersection(disabledRuleIdentifiers)
+                violations.append(contentsOf: alreadyDisabledRuleIdentifiers.map {
+                    let reason = "The disabled '\($0.stringRepresentation)' rule was already disabled"
+                    return violation(forFile: file, command: command, reason: reason)
+                })
```
SimplyDanny, line 57: It would help me understand the different checks better if this was a method call with an expressive name:
```suggestion
                violations.append(contentsOf: validateAlreadyDisabledRules(...))
```
Same for the other branches/checks. What do you think?

SimplyDanny (commented):
Just found some more points I'd like to discuss. Sorry! 😅

mildm8nnered:
> Just found some more points I'd like to discuss. Sorry! 😅

No worries :-) I'll try and get to these today ...

Source/SwiftLintFramework/Rules/RuleConfigurations/BlanketDisableCommandConfiguration.swift
mildm8nnered: This is actually correct - the intention is that by default, the rule will not complain about 

`// swiftlint:disable file_length`, for example, even though it's not re-enabled, because it makes sense to just disable it once for the whole file, and there's no point in doing `// swiftlint:enable file_length` at the end of the file.

Adding a rule to the `alwaysBlanketDisableRuleIdentifiers` goes one step further, by warning the user on any SwiftLint command affecting the rule, that is anything other than `// swiftlint:disable file_length`. I didn't want to enable that by default.

Source/SwiftLintFramework/Rules/Lint/BlanketDisableCommandRule.swift
mildm8nnered: Yep, that is much better.

Source/SwiftLintFramework/Rules/RuleConfigurations/BlanketDisableCommandConfiguration.swift
mildm8nnered: I have made one slight modification though - any rules in `alwaysBlanketDisableRuleIdentifiers` are automatically "allowed"

Source/SwiftLintFramework/Rules/Lint/BlanketDisableCommandRule.swift
mildm8nnered: Yep, I've made this change - definitely an improvement

Source/SwiftLintFramework/Rules/Lint/BlanketDisableCommandRule.swift
mildm8nnered: Possibly - I'll give that a go. Re #4799, not sure if you read my recent comment there, but I think just removing support for commands in multiline comments completely might be the way to go.

mildm8nnered: Yep, so the fixes I have in #4799 do not resolve `.skipWrappingInCommentTests` generally, partly because the general issue with commands in block comments (only the first one will be recognised) still remain, and because commands should still work in comments.

mildm8nnered:
> > Just found some more points I'd like to discuss. Sorry! 😅
> 
> No worries :-) I'll try and get to these today ...

ok - so I think I've addressed everything from that last batch :-)

Source/SwiftLintFramework/Rules/Lint/BlanketDisableCommandRule.swift
SimplyDanny: Got it. 👍

Source/SwiftLintFramework/Rules/RuleConfigurations/BlanketDisableCommandConfiguration.swift
SimplyDanny: Hm, got it ... this is tricky, though.

Source/SwiftLintFramework/Rules/Lint/BlanketDisableCommandRule.swift
```diff
@@ -0,0 +1,198 @@
...
+            in: file,
+            disabledRuleIdentifiers: disabledRuleIdentifiers,
+            ruleIdentifierToCommandMap: ruleIdentifierToCommandMap
+        )
+        violations += validateAlwaysBlanketDisable(file: file)
+
+        return violations
+    }
+
+    private func violation(
```
SimplyDanny, line 90: Is this indirection really needed for only one saved conversion?

mildm8nnered: So in three cases I have a RuleIdentifier (so they use the indirection), and then in two cases, in `validateAlwaysBlanketDisable`, I have the string.

I couldn't see anyway round that - the indirection felt cleaner than making the three call sites call `stringRepresentation` themselves.

Source/SwiftLintFramework/Rules/RuleConfigurations/BlanketDisableCommandConfiguration.swift
mildm8nnered: I think this is the right behaviour, but maybe the naming isn't quite perfect. I went with `allowed_rules` and `always_blanket_disable`, but maybe there's something better, although I don't have any good suggestions.

I've also just added `file_header`, `file_name`, and `file_name_no_space` to the default allowed rules list.

SimplyDanny:
Well done. Thanks @mildm8nnered!

PR #4732, opened 2023-01-29
Implement a workaround for Swift Package build plugins consuming input files

tonyarnold:
If a build tool plugin is applied to an Xcode target and lists `.swift` source files as inputs of the build commands it generates, then those `.swift` source files aren't seen by Xcode's build system (as they've been "consumed" by the plugin command). 

The workaround creates a new symbolic link to the project directory within the build plugin's working directory, then references every input file via that symbolically linked directory as the input files for the build command. This prevents Xcode "consuming" the input files, and allows further processing such as compilation.

The SwiftLint executable is still passed the unchanged input file paths. 

The workaround is only run for Xcode targets, and is unnecessary for SwiftPM targets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,12 @@
 
 #### Bug Fixes
 
+* Work around an issue where the Swift Package Manger build plugin would 
+  "consume" input files causing Xcode's build system to ignore them in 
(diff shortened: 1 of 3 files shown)


tonyarnold:
Superseded by #4757.

Issue #4660, opened 2022-12-27
Allow shortening output in Terminal


DaniSpringer:
Running `swiftlint --progress` in my repos folder, outputs various warnings for each folder.
However, many of the outputted lines wrap to the next line, because the full absolute path is printed (and that is too long for a single line).
It would be nice if we can have swiftlint only print the relative path (`pwd` + `/warning_folder`), so as to stay on a single line more often.

So instead of printing this:
>/Users/dani/Documents/repos/mathee/AppDelegate.swift:16:8: warning: Todo Violation: TODOs should be resolved (Shabbos). (todo)

To print this:
>/mathee/AppDelegate.swift:16:8: warning: Todo Violation: TODOs should be resolved (Shabbos). (todo)

SimplyDanny:
The output format depends in the reporter in use that can be specified by the `--reporter` option. The default is `xcode`. I guess Xcode requires the full path to display the messages in the editor properly or at all.

Would another reporter like `relative-path`, `xcode-relative-path`, `short-console`, ... be an option for you?

SimplyDanny:
> Would another reporter like `relative-path`, `xcode-relative-path`, `short-console`, ... be an option for you?

This is not a suggestion but a question. The idea would be to add another reporter with any of the mentioned names (or another one) that can be enabled by `--reported <name>`. Is this something that would help you?

By the way, I just noticed that the default reporter can even be specified in the `.swiftlint.yml` configuration.

DaniSpringer:
@SimplyDanny yes I think that would help! :)

> By the way, the default reporter can even be specified in the `.swiftlint.yml` configuration.
Ok great to know, too.

Thank you

SimplyDanny:
The new reporter has just been merged. It will be available in the next release.

DaniSpringer:
> The new reporter has just been merged. It will be available in the next release.

Sounds great! But I'm not sure I follow 😅 Anything I should do?

Thanks
Dani

SimplyDanny:
> Anything I should do?

Not necessarily. The new reporter will be available in the upcoming release. If you'd like to try it, you may wait for the new version or directly test it on the current HEAD.

That's up to you, though. I just wanted to let you know.

mildm8nnered:
You can install `0.51.0` via Homebrew, or any of the usual methods, once upgraded (`swiftlint version` to check), at the top of some source tree (I used SwiftLint itself here), try 

[...]

PR #4739, opened 2023-02-03
Allow shortening output in Terminal

roya1v:
Added a new reporter that uses relative paths. As described in #4660. 
I've updated the unit tests but not sure that they make a lot of sense as with the current mocked location the output is identical to the xcode reporter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@
   [David Steinacher](https://github.com/stonko1994)
   [#4626](https://github.com/realm/SwiftLint/issues/4626)
   
-* Add `relative-path` reporter to generate reports with relative paths
+* Add `relative-path` reporter to generate reports with relative file paths.  
   [Roya1v](https://github.com/roya1v)
   [#4660](https://github.com/realm/SwiftLint/issues/4660)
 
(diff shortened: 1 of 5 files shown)


SimplyDanny (commented):
Thank you for the contribution, @roya1v!

Could you please add another test case verifying that paths are really made relative by this reporter?

roya1v:
@SimplyDanny thanks for checking out my pr so quickly! Frankly I'm not entirely sure what the best way to implement this test case. 
Let me know what you think and if you want me to fix something!
Thanks <3

SimplyDanny (approved):
> Let me know what you think and if you want me to fix something!

The new test case looks sufficient to me. Would you like to add a CHANGELOG entry to credit your contribution?

roya1v:
For sure! Just did that

CHANGELOG.md
```diff
@@ -41,6 +41,10 @@
   `quick_discouraged_call`.  
   [David Steinacher](https://github.com/stonko1994)
   [#4626](https://github.com/realm/SwiftLint/issues/4626)
+  
+* Add `relative-path` reporter to generate reports with relative paths
```
SimplyDanny, line 45:
```suggestion
* Add `relative-path` reporter to generate reports with relative file paths.  
```
Sorry, but the two spaces at the end are important to force a line break. 😅

SimplyDanny:
Thank you, @roya1v!

PR #4742, opened 2023-02-05
Fix redundant objc attribute false alarms

mildm8nnered:
Fixes some false alarms in the redundant_objc_attribute that seem to have come in around 0.50.1

Resolves https://github.com/realm/SwiftLint/issues/4633

We've seen a few of these in our own codebase.

I picked this up and fixed most of it, when I noticed @Cyberbeni 's branch. My final solution is a mix of my original, and some stuff I cribbed from his fixes at https://github.com/realm/SwiftLint/pull/4642

I also patched up some of the examples that weren't valid Swift code or didn't make sense.

These are the examples that were triggering false alarms:

```
        @objcMembers
        public class Foo: NSObject {
            @objc
            private func handler(_ notification: Notification) {
            }
            func registerForNotifications() {
                NotificationCenter.default.addObserver(self, selector: #selector(handler(_:)), name: nil, object: nil)
            }
        }
```

```
        class Foo: NSObject { }
        @objc extension Foo {
            @objc enum Bar: Int {
               case bar
            }
            var bar: Bar { .bar }
        }
```

```
        class Foo: NSObject { }
        @objc extension Foo {
          @objc private enum Baz: Int {
            case baz
          }
          private var baz: Baz { .baz }
        }
```

```
        @objcMembers
        internal class Foo: NSObject {
            @objc
            private var baz: Int = 1
            var x: Any? {
                value(forKey: "baz")
            }
        }
```

```
        @objcMembers
        class Foo: NSObject {
            @objc enum Bar: Int {
               case bar
            }
        }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -389,6 +389,10 @@
 * Fix correction on `lower_acl_than_parent` rule for `open` declarations.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#4753](https://github.com/realm/SwiftLint/issues/4753)
+* Fix false positives on `redundant_objc_attribute` rule for enums
+  and private members.  
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#4633](https://github.com/realm/SwiftLint/issues/4633)
 
 * Fix `void_return` rule to support async and async throws functions.  
   [Mathias Schreck](https://github.com/lo1tuma)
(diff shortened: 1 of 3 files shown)


mildm8nnered:
See also https://github.com/apple/swift/issues/63442

SimplyDanny:
Thank you and sorry for letting you wait for so long! 😶

PR #4744, opened 2023-02-06
Cache folded syntax tree

marcelofabri:


marcelofabri:
I was hoping to get more speed gains tbh, not sure if it's worth merging this. Probably it is given that the number of rules using `folded()` will in theory only go up?

SimplyDanny (approved):

SimplyDanny:
Makes sense to me. I expect more rules using it over time.

marcelofabri:
Merging because this also has the nice benefit of allowing a custom rule to skip validation based on the file path by returning `nil` in `preprocess`.

Source/SwiftLintFramework/Protocols/SwiftSyntaxRule.swift
```diff
@@ -24,7 +24,7 @@ public protocol SwiftSyntaxRule: SourceKitFreeRule {
     /// - parameter syntaxTree: The syntax tree to run pre-processing on
```
jpsim, line 24: This is now out of date.

jpsim (commented):
+1 other than the outdated docs

Source/SwiftLintFramework/Protocols/SwiftSyntaxRule.swift
marcelofabri: good catch, will update

PR #4745, opened 2023-02-06
Avoid withoutTrivia() where it’s not necessary

marcelofabri:


marcelofabri:
it's interesting that the violations seem to be valid, but not sure why this change would start flagging them

jpsim:
Have you figured out the OSSCheck results?

marcelofabri:
> Have you figured out the OSSCheck results?

Not really, my best guess is that there's something about these files that perhaps trigger a bug in SwiftSyntax when calling `withoutTrivia()`?

SimplyDanny:
All changes are contained in 2f0e537.

PR #3438, opened 2020-11-19
Add Bool violation reporting in `redundant_type_annotation`

agarmash:
Fixed the issue and refactored the code a bit.
Resolves #3423

Issue #3423, opened 2020-11-10
Valid violation not reported by redundant_type_annotation for Bool


revolter:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

Using the `redundant_type_annotation` and `var isActive: Bool = false` doesn't trigger a warning.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ ./swiftlint lint Test.swift
Linting Swift files at paths Test.swift
Linting 'Test.swift' (1/1)
Done linting! Found 0 violations, 0 serious in 1 file.
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.41.0
* Installation method used (Homebrew, CocoaPods, building from source, etc)? Using a pre-built package
* Paste your configuration file:

```yml
opt_in_rules:
    - redundant_type_annotation
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? No
  If so, paste their relative paths and respective contents.
* Which Xcode version are you using (check `xcodebuild -version`)? Xcode 12.0 Build version 12A7209
* Do you have a sample that shows the issue? Run `echo "[string here]" | swiftlint lint --no-cache --use-stdin --enable-all-rules`
  to quickly test if your example is really demonstrating the issue. If your example is more
  complex, you can use `swiftlint lint --path [file here] --no-cache --enable-all-rules`.

```swift
class Test {
	var isActive: Bool = false
}
```

revolter:
@jpsim, Same for this one.

PR #4756, opened 2023-02-08
Add consider_default_literal_types_redundant option to RedundantTypeAnnotationRule

garricn:
- Add `consider_default_literal_types_redundant` option to `RedundantTypeAnnotationConfiguration`
- Add examples in rule description
- Improve control flow in `visitPost` methods
- Refactor `isRedundant` method to include and use `considerLiteralsRedundant` for `Bool`, `Double`, `Int` and `String`

By default, literals for `Bool`, `Double`, `Int` and `String` will be considered **not** redundant. For example: `var isEnabled: Bool = true`. This is an intentional change in behavior. See the comments below for additional details.

> [!IMPORTANT]
> The `consider_default_literal_types_redundant ` option must be set to `true` to maintain the previous behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,8 @@
 * Add `consider_default_literal_types_redundant` option to the
   `redundant_type_annotation` rule supporting `Bool`, `Double`, `Int` and
   `String`. Setting the option to `true` lets the rule consider said types in
-  declarations like `let i: Int = 1` or `let s: String = ""` as redundant.  
+  declarations like `let i: Int = 1` or `let s: String = ""` as redundant.
+  Note that `Bool` literals will no longer be considered redundant by default.  
   [Garric Nahapetian](https://github.com/garricn)
 
 #### Experimental
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,11 +21,10 @@
   [Martin Redington](https://github.com/mildm8nnered)
   [#4792](https://github.com/realm/SwiftLint/issues/4792)
 
-* Add `consider_default_literal_types_redundant` option to the
-  `redundant_type_annotation` rule supporting `Bool`, `Double`, `Int` and
-  `String`. Setting the option to `true` lets the rule consider said types in
-  declarations like `let i: Int = 1` or `let s: String = ""` as redundant.
-  Note that `Bool` literals will no longer be considered redundant by default.  
+* With the introduction of the `consider_default_literal_types_redundant`
+  option to the `redundant_type_annotation` rule, `Bool` literals will no
+  longer be considered redundant by default. Set this option to true to
+  preserve the previous behavior.  
   [Garric Nahapetian](https://github.com/garricn)
 
 #### Experimental
@@ -194,6 +193,14 @@
   `list.enumerated().map { $1 }`.  
   [Martin Redington](https://github.com/mildm8nnered)
   [#5470](https://github.com/realm/SwiftLint/issues/5470)
+
+* Include `Double`, `Int` and `String` to the exiting redundant type validation
+  check of `Bool` in the `redundant_type_annotation` rule. Add
+  `consider_default_literal_types_redundant` option supporting `Bool`,
+  `Double`, `Int` and `String`. Setting this option to `true` lets the rule
+  consider said types in declarations like `let i: Int = 1` or
+  `let s: String = ""` as redundant.  
+  [Garric Nahapetian](https://github.com/garricn)
 
 #### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,12 @@
   [Martin Redington](https://github.com/mildm8nnered)
   [#4792](https://github.com/realm/SwiftLint/issues/4792)
 
+* Add `consider_default_literal_types_redundant` option to
+  `RedundantTypeAnnotationRule` supporting `Bool`, `Double`, `Int` and `String`.
+  Note that this option must be set to `true` to maintain the previous behavior.
+  [Garric Nahapetian](https://github.com/garricn)
+  [#4756](https://github.com/realm/SwiftLint/pull/4756)
+
 #### Experimental
 
 * Add two new options to the `lint` and `analyze` commands: `--write-baseline`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,11 +21,11 @@
   [Martin Redington](https://github.com/mildm8nnered)
   [#4792](https://github.com/realm/SwiftLint/issues/4792)
 
-* Add `consider_default_literal_types_redundant` option to
-  `RedundantTypeAnnotationRule` supporting `Bool`, `Double`, `Int` and `String`.
-  Note that this option must be set to `true` to maintain the previous behavior.
+* Add `consider_default_literal_types_redundant` option to the
+  `redundant_type_annotation` rule supporting `Bool`, `Double`, `Int` and
+  `String`. Setting the option to `true` lets the rule consider said types in
+  declarations like `let i: Int = 1` or `let s: String = ""` as redundant.  
   [Garric Nahapetian](https://github.com/garricn)
-  [#4756](https://github.com/realm/SwiftLint/pull/4756)
 
 #### Experimental
 
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
@@ -234,20 +234,25 @@ private extension TypeAnnotationSyntax {
         // Consider the type annotation redundant if `considerLiteralsRedundant` is true and the expression is one of
         // the supported compiler-inferred literals.
         if considerLiteralsRedundant {
-            return switch initializer {
-            case let expr where expr.is(BooleanLiteralExprSyntax.self):
-                type.trimmedDescription == "Bool"
-            case let expr where expr.is(FloatLiteralExprSyntax.self):
-                type.trimmedDescription == "Double"
-            case let expr where expr.is(IntegerLiteralExprSyntax.self):
-                type.trimmedDescription == "Int"
-            case let expr where expr.is(StringLiteralExprSyntax.self):
-                type.trimmedDescription == "String"
-            default:
-                false
-            }
+            return isCompilerInferredLiteral(expression: initializer)
         }
         return initializer.accessedNames.contains(type.trimmedDescription)
+    }
+
+    // Returns true if the expression is one of the supported compiler-inferred literals.
+    private func isCompilerInferredLiteral(expression: ExprSyntax) -> Bool {
+        return switch expression {
+        case let expr where expr.is(BooleanLiteralExprSyntax.self):
+            type.trimmedDescription == "Bool"
+        case let expr where expr.is(FloatLiteralExprSyntax.self):
+            type.trimmedDescription == "Double"
+        case let expr where expr.is(IntegerLiteralExprSyntax.self):
+            type.trimmedDescription == "Int"
+        case let expr where expr.is(StringLiteralExprSyntax.self):
+            type.trimmedDescription == "String"
+        default:
+            false
+        }
     }
 }
 
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
@@ -221,32 +221,6 @@ private extension RedundantTypeAnnotationRule {
     }
 }
 
-private extension SyntaxKind {
-    var compilerInferredLiteralType: String? {
-        switch self {
-        case .booleanLiteralExpr:
-            "Bool"
-        case .floatLiteralExpr:
-            "Double"
-        case .integerLiteralExpr:
-            "Int"
-        case .stringLiteralExpr:
-            "String"
-        default:
-            nil
-        }
-    }
-
-    func isLiteralExpression() -> Bool {
-        switch self {
-        case .booleanLiteralExpr, .floatLiteralExpr, .integerLiteralExpr, .stringLiteralExpr:
-            true
-        default:
-            false
-        }
-    }
-}
-
 private extension ExprSyntax {
     /// An expression can represent an access to an identifier in one or another way depending on the exact underlying
     /// expression type. E.g. the expression `A` accesses `A` while `f()` accesses `f` and `a.b.c` accesses `a` in the
@@ -277,5 +251,31 @@ private extension ExprSyntax {
             expr = forceUnwrap.expression
         }
         return expr.accessedNames.contains(type.trimmedDescription)
+    }
+}
+
+private extension SyntaxKind {
+    var compilerInferredLiteralType: String? {
+        switch self {
+        case .booleanLiteralExpr:
+            "Bool"
+        case .floatLiteralExpr:
+            "Double"
+        case .integerLiteralExpr:
+            "Int"
+        case .stringLiteralExpr:
+            "String"
+        default:
+            nil
+        }
+    }
+
+    func isLiteralExpression() -> Bool {
+        switch self {
+        case .booleanLiteralExpr, .floatLiteralExpr, .integerLiteralExpr, .stringLiteralExpr:
+            true
+        default:
+            false
+        }
     }
 }
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
@@ -242,7 +242,7 @@ private extension ExprSyntax {
     }
 
     func hasRedundantLiteralType(_ type: TypeSyntax) -> Bool {
-        kind.isLiteralExpression() && kind.compilerInferredLiteralType == type.trimmedDescription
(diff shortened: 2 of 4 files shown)


SimplyDanny:
An observation for this rule in general: The exception for `Bool` seems to be quite arbitrary. What about `var j: Int = 7` or `var d: Double = 1.3` or `let s: String = "my string"`?

I wonder if we should rather extend the rule for these cases and rename the new option `ignore_booleans` to `ignore_literals`.

@revolter: You added the exception for `Bool` back then. Anything that speaks against my suggestions?

garricn:
> An observation for this rule in general: The exception for `Bool` seems to be quite arbitrary. What about `var j: Int = 7` or `var d: Double = 1.3` or `let s: String = "my string"`?
> 
> I wonder if we should rather extend the rule for these cases and rename the new option `ignore_booleans` to `ignore_literals`.
> 
> @revolter: You added the exception for `Bool` back then. Anything that speaks against my suggestions?

Hi @SimplyDanny, thank you for your observation. The examples you gave are not currently triggering examples and booleans are the only literals that currently violate this rule. It makes sense that numeric and string literals are not violations since you may need to be explicit that the type in `let foo = 0` is actually an `Int`, `Double`, or `Float`. Or, in `let foo = "someString"`, you may need to be explicit that `foo` is a `String` or `StaticString`. This is why we made the determination to implement this as `ignore_booleans` instead of something like `ignore_literals`.

SimplyDanny:
Well, without the explizit types, the variables `j`, `d` and `s` would have the same type as specified. They are the defaults.

The current implementation checks exactly for the combinations `Bool = true` and `Bool = false`. The same could be done for `Int = <any integer>`, `Double = <any float>`, ...

Moreover, a variable being assigned a Boolean literal is not necessarily a `Bool`. That also works for types conforming to `ExpressibleByBooleanLiteral`.

The exception for `Bool` is still not clear to me yet.

garricn:
> Well, without the explizit types, the variables `j`, `d` and `s` would have the same type as specified. They are the defaults.
> 
> The current implementation checks exactly for the combinations `Bool = true` and `Bool = false`. The same could be done for `Int = <any integer>`, `Double = <any float>`, ...
> 
> Moreover, a variable being assigned a Boolean literal is not necessarily a `Bool`. That also works for types conforming to `ExpressibleByBooleanLiteral`.
> 
> The exception for `Bool` is still not clear to me yet.

Thank you for your reply @SimplyDanny. The original intent for the special handling of Bool is not discernible from the [PR](https://github.com/realm/SwiftLint/pull/3438) or the [issue](https://github.com/realm/SwiftLint/issues/3423).

My guess is that we would not want any of the following to be violations because, in certain cases, we might need to be explicit about the type:

```swift
let someInt: Int = 0
let someDouble: Double = 0
let someFloat: Float = 0
let someString: String = "foo"
let someStaticString: StaticString =  "bar"
let someCustomBool: CustomBool = false
```

However, in the case of Bool, we do not need to be explicit when defining one because `Bool` is the [only conforming type](https://developer.apple.com/documentation/swift/expressiblebybooleanliteral#conforming-types) to `ExpressibleByBooleanLiteral` in the Swift Standard Library.

https://developer.apple.com/documentation/swift/expressiblebystringliteral#conforming-types
https://developer.apple.com/documentation/swift/expressiblebyintegerliteral#conforming-types
https://developer.apple.com/documentation/swift/expressiblebyfloatliteral#conforming-types

Granted, you can create your own type that is `ExpressibleByBooleanLiteral` (like CustomBool above), but in that case you would want to be explicit about the type and you would not want the rule to trigger a violation.

Although I think your point is a valuable one, I believe it is outside the scope of this PR and so I'd like to suggest the following way forward:

1. Move ahead with this PR as is
2. Separately from this PR, other PRs in the future could consider removing custom handling of booleans or alternatively adding custom handling for all literal types

SimplyDanny:
> Although I think your point is a valuable one, I believe it is outside the scope of this PR and so I'd like to suggest the following way forward:
> 
> 1. Move ahead with this PR as is
> 2. Separately from this PR, other PRs in the future could consider removing custom handling of booleans or alternatively adding custom handling for all literal types

That might be an acceptable way to go. However, I still take exception to the option's name `ignore_booleans` which is obviously very specific for Booleans. Changing an options name later on might cause trouble for users, unfortunately. That's why I'm still opposed to just adding it immediately.

tinder-cfuller:
Reading through the comments, it really feels like the decision to have `let someBool: Bool = true` violate a rule named `RedundantTypeAnnotation` was possibly not the best decision in the first place, and is likely the cause for some confusion I am seeing in the comments.

I would like to reiterate that `var j: Int = 7`, `var d: Double = 1.3` and `let s: String = "my string"` do NOT violate `RedundantTypeAnnotation`. For some reason only the boolean check was added to `RedundantTypeAnnotation` (and hence why this PR is proposing only adding the exception for booleans – there would be no reason for the option to be `ignore_literals` since the other literals are not violations).

I would like to kindly suggest an alternative way forward:

- Revise this pull request to instead completely remove the boolean check from `RedundantTypeAnnotation`
- Submit another pull request to add a new opt-in rule to treat boolean literals with explicit type annotations as violations, maybe named `BooleanLiteralTypeAnnotation`

@SimplyDanny Would this alternative approach be accepted? Thank you 🙏

tinder-cfuller:
Reading through the comments, it really feels like the decision to have `let someBool: Bool = true` violate a rule named `RedundantTypeAnnotation` was possibly not the best decision in the first place, and is likely the cause for some confusion I am seeing in the comments.

I would like to reiterate that `var j: Int = 7`, `var d: Double = 1.3` and `let s: String = "my string"` do NOT violate `RedundantTypeAnnotation`. For some reason only the boolean check was added to `RedundantTypeAnnotation` (and hence why this PR is proposing only adding the exception for booleans – there would be no reason for the option to be `ignore_literals` since the other literals are not violations).

I would like to kindly suggest an alternative way forward:

- Revise this pull request to instead completely remove the boolean check from `RedundantTypeAnnotation`.
- Submit another pull request to add a new opt-in rule to treat boolean literals with explicit type annotations as violations named `BooleanLiteralTypeAnnotation` – or `LiteralTypeAnnotation` if it is preferred that all literals with explicit types are violations.

With this approach, there is no need for a configuration option at all.

@SimplyDanny Would this alternative approach be accepted? And do you prefer `BooleanLiteralTypeAnnotation` or `LiteralTypeAnnotation` for the new rule?

Thank you 🙏

revolter:
For clarity: I was the one that requested this (in #3423), but I didn't make any code change for it.

---

As far as the current situation goes, I would be fine with reverting that change completely.

> `Bool` is the [only conforming type](https://developer.apple.com/documentation/swift/expressiblebybooleanliteral#conforming-types) to `ExpressibleByBooleanLiteral` in the Swift Standard Library

but this can change in the future, so I would say that considering

```swift
let test: Bool = true
```

to be redundant is a guess, while considering

```swift
let test: URL = URL()
```

to be redundant is a certainty.

SimplyDanny:
I'd like to avoid exceptions for certain cases that actually should follow the same rules as others. Breaking changes are acceptable if they really fix an inconstancy or harmonize implementations.

The exception for booleans bugs me, with or without the `ignore_booleans` option.

Types can conform to any `ExpressibleBy...Literal` to make them be initializable with the corresponding literal. But all literals are as well associated with a default type. So if the rule talks about "redundant types" I'd consider explicit default types for literals redundant. Booleans follow the same logic; I don't see why there ought to be any special handling for them (as of now or with an option).

Since there might be people not agreeing on the statement that default types can be considered redundant, my favorite is still this `ignore_literals` or perhaps the better `consider_default_literal_types_redundant` to be more specific. With that set to `true`,

```swift
let k: Int = 0
let d: Double = 0.0
let s: String = ""
let b: Bool = false
```

would all trigger while

```swift
let k: Int8 = 0
let d: Double = 0
let s: Name = ""
let b: OnOff = true
```

would not.

tinder-cfuller:
@SimplyDanny Sounds good 👍 … `consider_default_literal_types_redundant` will default to `false`. Thank you for confirming that.

And may treating `Int`, `Double` and `String` literals as redundant be added on a separate PR (potentially by someone else), to limit the scope of the change on this PR?

SimplyDanny:
> And may treating `Int`, `Double` and `String` literals as redundant be added on a separate PR (potentially by someone else), to limit the scope of the change on this PR?

Well, ideally we would have all types properly supported right from the new option's introduction on. 😉

garricn:
@SimplyDanny Thanks again for all the feedback. I investigated adding validations for Int, Double and String literals and determined that an overhaul of the rule, potentially adopting SwiftSyntax, would likely be required. Since that would significantly increase the scope and complexity of the PR, I propose that the PR is limited to adding the new config option only. I've updated to use the option renamed as you requested, and it's working really well! My hope is that you will consider merging this as is and allow the additional type support to be added later. I look forward to hearing back from you. Thank you 😄

SimplyDanny:
> @SimplyDanny Thanks again for all the feedback. I investigated adding validations for Int, Double and String literals and determined that an overhaul of the rule, potentially adopting SwiftSyntax, would likely be required. Since that would significantly increase the scope and complexity of the PR, I propose that the PR is limited to adding the new config option only. I've updated to use the option renamed as you requested, and it's working really well! My hope is that you will consider merging this as is and allow the additional type support to be added later. I look forward to hearing back from you. Thank you 😄

The rewrite with SwiftSyntax is now done with #5389 just being merged. I would still appreciate you implementing all cases to have this in a complete and reasonable shape. With the rewrite, this ought not to be that complex. Otherwise, we need to wait for me or another contributor to find the time to address this.

tinder-cfuller:
@SimplyDanny The rewrite of the rule with SwiftSyntax makes this much more straightforward! Very cool! 👍

@garricn Thank you for implementing this! ❤️

CHANGELOG.md
```diff
@@ -21,6 +21,12 @@
   [Martin Redington](https://github.com/mildm8nnered)
   [#4792](https://github.com/realm/SwiftLint/issues/4792)
 
+* Add `consider_default_literal_types_redundant` option to
+  `RedundantTypeAnnotationRule` supporting `Bool`, `Double`, `Int` and `String`.
+  Note that this option must be set to `true` to maintain the previous behavior.
+  [Garric Nahapetian](https://github.com/garricn)
+  [#4756](https://github.com/realm/SwiftLint/pull/4756)
```
SimplyDanny, line 28:
```suggestion
* Add `consider_default_literal_types_redundant` option to the
  `redundant_type_annotation` rule supporting `Bool`, `Double`, `Int` and `String`.
  Setting the option to `true` lets the rule consider said types in
  declarations like `let i: Int = 1` or `let s: String = ""` as redundant.
  [Garric Nahapetian](https://github.com/garricn)
```

We only mention issue numbers in the changelog.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -168,47 +183,69 @@ private extension RedundantTypeAnnotationRule {
...
+            case let expr where expr.is(FloatLiteralExprSyntax.self):
+                type.trimmedDescription == "Double"
+            case let expr where expr.is(IntegerLiteralExprSyntax.self):
+                type.trimmedDescription == "Int"
+            case let expr where expr.is(StringLiteralExprSyntax.self):
+                type.trimmedDescription == "String"
+            default:
+                false
+            }
         }
```
SimplyDanny, line 249: This part should be a separate method that's only called when the option is `true`.

SimplyDanny (changes requested):

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -168,50 +183,77 @@ private extension RedundantTypeAnnotationRule {
...
-
-        // If the initializer is a boolean expression, we consider using the `Bool` type
-        // annotation as redundant.
-        if initializer.is(BooleanLiteralExprSyntax.self) {
-            return type.trimmedDescription == "Bool"
+        // Consider the type annotation redundant if `considerLiteralsRedundant` is true and the expression is one of
+        // the supported compiler-inferred literals.
+        if considerLiteralsRedundant {
+            return isCompilerInferredLiteral(expression: initializer)
```
tinder-cfuller, line 237:
```suggestion
            return isCompilerInferredLiteral(initializer)
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -168,50 +183,77 @@ private extension RedundantTypeAnnotationRule {
...
         }
         return initializer.accessedNames.contains(type.trimmedDescription)
     }
+
+    // Returns true if the expression is one of the supported compiler-inferred literals.
+    private func isCompilerInferredLiteral(expression: ExprSyntax) -> Bool {
+        return switch expression {
```
tinder-cfuller, line 244:
```suggestion
    private func isCompilerInferredLiteral(_ expr: ExprSyntax) -> Bool {
        switch expr {
```

tinder-cfuller (approved):
Thanks again Garric! Looking forward to using this new option! 👍

I added two small suggestions, but they are very minor, so only use if you see a benefit.

I do have one question though, which is whether we need to indicate somewhere that `Bool` literals will no longer be considered redundant by default. Since this is a change in behavior, does it need to be listed as a "breaking" change?

SimplyDanny:
> I do have one question though, which is whether we need to indicate somewhere that `Bool` literals will no longer be considered redundant by default. Since this is a change in behavior, does it need to be listed as a "breaking" change?

Yes, I'd like that.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -234,14 +234,14 @@ private extension TypeAnnotationSyntax {
...
+            return isCompilerInferredLiteral(initializer)
         }
         return initializer.accessedNames.contains(type.trimmedDescription)
     }
 
     // Returns true if the expression is one of the supported compiler-inferred literals.
-    private func isCompilerInferredLiteral(expression: ExprSyntax) -> Bool {
-        return switch expression {
+    private func isCompilerInferredLiteral(_ expr: ExprSyntax) -> Bool {
+        switch expr {
```
SimplyDanny, line 244: Switching over `expr.kind` instead would be cheaper.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -234,14 +234,14 @@ private extension TypeAnnotationSyntax {
         // Consider the type annotation redundant if `considerLiteralsRedundant` is true and the expression is one of
         // the supported compiler-inferred literals.
         if considerLiteralsRedundant {
-            return isCompilerInferredLiteral(expression: initializer)
+            return isCompilerInferredLiteral(initializer)
```
SimplyDanny, line 237: I would actually expect this to be called outside of this method separately to avoid passing the option around.

SimplyDanny (changes requested):

CHANGELOG.md
```diff
@@ -21,6 +21,13 @@
   [Martin Redington](https://github.com/mildm8nnered)
   [#4792](https://github.com/realm/SwiftLint/issues/4792)
 
+* Add `consider_default_literal_types_redundant` option to the
+  `redundant_type_annotation` rule supporting `Bool`, `Double`, `Int` and
+  `String`. Setting the option to `true` lets the rule consider said types in
+  declarations like `let i: Int = 1` or `let s: String = ""` as redundant.
+  Note that `Bool` literals will no longer be considered redundant by default.
```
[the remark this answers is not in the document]
garricn: The existing lines 24 to 28 are currently all in the Breaking section already. If you can clarify your meaning, I would be happy to adjust it. Thank you.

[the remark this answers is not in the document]
SimplyDanny: Ah, I see. I though that we should split the entry in an improvement for all literals except booleans and a breaking change for booleans.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -233,4 +240,42 @@ private extension ExprSyntax {
             []
         }
     }
+
+    func hasRedundantLiteralType(_ type: TypeSyntax) -> Bool {
+        kind.isLiteralExpression() && type.trimmedDescription == kind.compilerInferredLiteralType
```
SimplyDanny, line 245:
```suggestion
        type.trimmedDescription == kind.compilerInferredLiteralType 
```

This is actually enough because comparison to `nil` fails anyway.

garricn: Great catch!

SimplyDanny (approved):
I like how that turned out. The new option makes a lot of sense to me and resolves and aged inconsistency.

tinder-cfuller:
@SimplyDanny Thanks for all your guidance and attention on this! Very much appreciated ⭐

@garricn Great work! Thank you again for making this improvement.

SimplyDanny:
Just to let you guys know: I made a few cleanups to reduce some duplications in #5554. Didn't want to prolong the review here while being pedantic. 😅

tinder-cfuller:
@SimplyDanny That great! 👏

tinder-cfuller:
@SimplyDanny That's great! 👏

PR #4757, opened 2023-02-08
Ensure that configured exclusion rules are respected by the build plugin

tonyarnold:
#4680 changed the build plugin to use a `prebuildCommand` rather than a `buildCommand`. For some reason, this change altered the behaviour and configured exclusions were ignored. 

Given that we always pass all of the Swift source files in the target to the tool within the build plugin, we need to ensure that any exclusion rules in the configuration are respected. 

This PR proposes to add the `--force-exclude` option to the build plugin's invocation of the tool.

### DIFF
--- a/Plugins/SwiftLintPlugin/SwiftLintPlugin.swift
+++ b/Plugins/SwiftLintPlugin/SwiftLintPlugin.swift
@@ -29,6 +29,10 @@ struct SwiftLintPlugin: BuildToolPlugin {
         var arguments = [
             "lint",
             "--quiet",
+            // We always pass all of the Swift source files in the target to the tool,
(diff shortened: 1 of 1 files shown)


evfemist:
Found that if I paste generated command to the Terminal - it will correctly exclude files without `--force-exclude`. This means that there are some difference how swiftlint runs during build process and when just in Terminal.
But I don't know what this can be. So probably `--force-exclude` is the easiest way to fix this.

tonyarnold:
Yeah, it seems necessary for the build plugin within Xcode to specify `--force-exclude`.

If you haven't already, take a look at #4758, it looks like that will supersede this PR.

SimplyDanny (approved):
Thanks! Adding the option in this context makes sense to me.

PR #4764, opened 2023-02-13
Update README.md

mhasan341:
Added Change to error regarding permission on linking


SimplyDanny:
I don't think it's SwiftLint's responsibility to explain that. It should be clear to every user that certain directories require elevated access rights.

So either we propose `sudo ln ...` by default or we leave it as is. A "permission denied" warning should be enough indication on what's wrong.

Harry-KNIGHT:
LGTM

Issue #4737, opened 2023-01-31
Bazel instructions result in missing `rules_xcodeproj`


mattrobmattrob:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

Lack of `rules_xcodeproj` definition when following the `WORKSPACE` + `bazel run -c opt @SwiftLint//:swiftlint` README.md definition in a workspace that doesn't already define `rules_xcodeproj`.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ bazel run -c opt @SwiftLint//:swiftlint
INFO: Invocation ID: 56dcdaec-eba1-4d83-badf-90936d230d2d
ERROR: Skipping '@SwiftLint//:swiftlint': error loading package '@SwiftLint//': Unable to find package for @com_github_buildbuddy_io_rules_xcodeproj//xcodeproj:xcodeproj.bzl: The repository '@com_github_buildbuddy_io_rules_xcodeproj' could not be resolved: Repository '@com_github_buildbuddy_io_rules_xcodeproj' is not defined.
WARNING: Target pattern parsing failed.
ERROR: error loading package '@SwiftLint//': Unable to find package for @com_github_buildbuddy_io_rules_xcodeproj//xcodeproj:xcodeproj.bzl: The repository '@com_github_buildbuddy_io_rules_xcodeproj' could not be resolved: Repository '@com_github_buildbuddy_io_rules_xcodeproj' is not defined.
INFO: Elapsed time: 5.806s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
    currently loading: @SwiftLint//
ERROR: Build failed. Not running target
```

### Environment

`WORKSPACE` to use `bazel run -c opt @SwiftLint//:swiftlint`:
```yml
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

# iOS/macOS

http_archive(
    name = "build_bazel_rules_apple",
    sha256 = "43737f28a578d8d8d7ab7df2fb80225a6b23b9af9655fcdc66ae38eb2abcf2ed",
    url = "https://github.com/bazelbuild/rules_apple/releases/download/2.0.0/rules_apple.2.0.0.tar.gz",
)

load(
    "@build_bazel_rules_apple//apple:repositories.bzl",
    "apple_rules_dependencies",
)

[...]

PR #4768, opened 2023-02-16
Update BUILD file for SwiftSyntax

thii:
Based on https://github.com/apple/swift-syntax/blob/0.50800.0-SNAPSHOT-2022-12-29-a/Package.swift

Fixes the second error in https://github.com/realm/SwiftLint/issues/4737.


jpsim (approved):
Thanks!

Issue #3173, opened 2020-04-06
Rule Request: Enforce private properties in SwiftUI `View`'s


sindresorhus:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.

Apple recommends that state properties in SwiftUI [`View`](https://developer.apple.com/documentation/swiftui/view)'s are private:

> You should only access a state property from inside the view’s body, or from methods called by it. For this reason, declare your state properties as private, to prevent clients of your view from accessing it. - https://developer.apple.com/documentation/swiftui/state

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

**Would trigger:**

```swift
struct ContentView: View {
	@State var foo = ""
}
```

```swift
struct ContentView: View {
	@ObservedObject var foo = Foo()
}
```

```swift
struct ContentView: View {
	@EnvironmentObject var foo
}
```

**Would NOT trigger:**

```swift
struct ContentView: View {
	var foo = ""
}
```

3. Should the rule be configurable, if so what parameters should be configurable?

I would like to see an option to enforce that all properties that are not initializers to be private.

4. Should the rule be opt-in or enabled by default? Why?
   See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a rule as opt-in.

Yes, since it's what Apple recommends.

christiankm:
I created a simple custom rule for this, as a starting point. I will try to make a PR for an official rule.

```
state_private:
    name: "Private SwiftUI State"
    regex: "\\@State\\s*var"
    message: "SwiftUI @State variables should always be marked private."
    severity: warning
```

sindresorhus:
Please keep it open.

sindresorhus:
Please keep it open.

mt00chikin:
FYI, I created https://github.com/realm/SwiftLint/pull/4769 as a starting point to try and address a subset of these concerns. My team was also looking at a custom rule to detect @State properties and require them to be private. It doesn't cover every concern addressed in here, but it could definitely be updated or configured to cover additional cases if desired.

PR #4769, opened 2023-02-16
Add new `private_swiftui_state_property` rule

mt00chikin:
This creates a new opt-in rule that requires that SwiftUI state properties be declared as private. This is inspired by Apple's [documentation](https://developer.apple.com/documentation/swiftui/state) for State:
> To avoid this, always declare state as private, and place it in the highest view in the view hierarchy that needs access to the value

This only covers @State and not other SwiftUI property wrappers like @StateObject or @EnvironmentObject.

This MR attempts to address requests from issue: https://github.com/realm/SwiftLint/issues/3173


CHANGELOG.md
```diff
@@ -18,6 +18,11 @@
 
 #### Enhancements
 
+* Add new `private_swiftui_state_property` opt-in rule to encourage setting 
+  SwiftUI @State properties to private.
```
SimplyDanny, line 22: Two spaces at the end are mandatory to enforce a line break.
```suggestion
  SwiftUI `@State` properties to private.  
```

CHANGELOG.md
```diff
@@ -18,6 +18,11 @@
...
+  [mt00chikin](https://github.com/mt00chikin)
+  [#issue_number](https://github.com/realm/SwiftLint/issues/3173)
```
SimplyDanny, line 24:
```suggestion
  [#3173](https://github.com/realm/SwiftLint/issues/3173)
```

Source/SwiftLintFramework/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -0,0 +1,99 @@
+import SwiftSyntax
+
+/// Rule to require that any state properties in SwiftUI be declared as private.
+/// State properties should only be accessible from inside the View's body, or from methods called by it
+struct PrivateSwiftUIStatePropertyRule: SwiftSyntaxRule, OptInRule, ConfigurationProviderRule {
+    var configuration = SeverityConfiguration(.warning)
+
+    static let description = RuleDescription(
+        identifier: "private_swiftui_state",
+        name: "Private SwiftUI @State properties",
```
SimplyDanny, line 10: We use headline style for these names:
```suggestion
        name: "Private SwiftUI @State Properties",
```

Source/SwiftLintFramework/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -0,0 +1,99 @@
...
+        description: "SwiftUI state properties should be private",
```
SimplyDanny, line 11: Does that sound better?
```suggestion
        description: "SwiftUI's state properties should be private",
```

Source/SwiftLintFramework/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -0,0 +1,99 @@
...
+                  let identifier = stateAttr.attributeName.as(SimpleTypeIdentifierSyntax.self) else {
+                return false
+            }
+
+            return identifier.name.text == "State"
+        }
+    }
+}
+
+private extension ModifierListSyntax {
```
SimplyDanny, line 89: Please check if some of these extensions are already defined in `SwiftSyntax+SwiftLint.swift`. If so, please use those. If not, think about integrating yours with the existing ones.

SimplyDanny (commented):
Thank you for the contribution @mt00chikin! It's a good starting point.

Are you going to add more SwiftUI attributes to this rule?

It should be hard to make this rule correctable, that is offer a rewriter to add the `private` modifier automatically when running `swiftlint --fix`. What do you think about that?

mt00chikin:
I updated the branch based on your feedback and suggestions, they all looked good to me. 
As far as making this correctable, that seems like a good addition to this rule. I'll have to dig into the Rewriter code a bit to figure out how it's implemented, but I'd be happy to try and get that working as well.

mt00chikin:
Right now this rule only requires that SwiftUI `@State` properties be private since this is explicitly recommended in Apple's documentation, but I could potentially see an argument for adding @StateObject as well. 
I didn't want the scope of the rule to be too broad, so I left that one (along with @EnvironmentObject and @ObservedObject) out of this initial pass.

DivineDominion:
@SimplyDanny  With the advent of Swift macros and thus `@StateObject` being phased out, I believe this is a good-enough starting point, no more attributed needed 👍

SimplyDanny:
> @SimplyDanny With the advent of Swift macros and thus `@StateObject` being phased out, I believe this is a good-enough starting point, no more attributed needed 👍

Convinced. Are you willing to rebase the PR and resolve all conflicts, @mt00chikin?

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRuleExamples.swift
```diff
@@ -0,0 +1,42 @@
...
+            }
+            """
+        )
+    ]
+
+    static let triggeringExamples: [Example] = [
+        Example(
+            """
+            struct ContentView: View {
+                @State var isPlaying: Bool = false
```
SimplyDanny, line 37: The example misses a violation marker `↓` to see where exactly the rule triggers.

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -0,0 +1,52 @@
...
+/// State properties should only be accessible from inside the View's body, or from methods called by it
+struct PrivateSwiftUIStatePropertyRule: SwiftSyntaxRule, OptInRule, ConfigurationProviderRule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "private_swiftui_state",
+        name: "Private SwiftUI @State Properties",
+        description: "SwiftUI's state properties should be private",
+        kind: .lint,
+        nonTriggeringExamples: PrivateSwiftUIStatePropertyRuleExamples.nonTriggeringExamples,
```
SimplyDanny, line 13: If a rule's implementation is short and the number of examples is rather small, we typically inline the examples.

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -0,0 +1,52 @@
...
+        triggeringExamples: PrivateSwiftUIStatePropertyRuleExamples.triggeringExamples
+    )
+
+    func makeVisitor(file: SwiftLintFile) -> ViolationsSyntaxVisitor {
+        Visitor(viewMode: .sourceAccurate)
+    }
+}
+
+private extension PrivateSwiftUIStatePropertyRule {
+    final class Visitor: ViolationsSyntaxVisitor {
```
SimplyDanny, line 23: I wonder if the rule should also check that the declaring type conforms to `View`. It doesn't seem too uncommon that there's an attribute called `State` which isn't the one SwiftUI defines.

mt00chikin: Good question, I believe I had played around with that at one point. Let me see if I can revive that code.

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
mt00chikin: 👍 I wasn't certain on the convention here. I can revert the change, it's an easy one!

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -0,0 +1,108 @@
...
+        override var skippableDeclarations: [DeclSyntaxProtocol.Type] {
+            .allExcept(StructDeclSyntax.self)
```
SimplyDanny, line 64: Final classes can conform to `View` as well if I remember correctly.

Other than that, we need to consider structs in structs and structs in classes and ... as well. Hence, it'd be better to exclude certain types instead of accepting structs exclusively.

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        override func visit(_ node: StructDeclSyntax) -> SyntaxVisitorContinueKind {
+            guard let inheritedTypeCollection = node.inheritanceClause?.inheritedTypeCollection else {
+                return .skipChildren
+            }
+
+            let inheritedTypes = Set(inheritedTypeCollection
```
SimplyDanny, line 72: I think wrapping into `Set` is not necessary. The list of inherited types is typically very short. So there won't be a performance improvement for `contains` by using a set instead of an array.

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
mt00chikin: 👍 I'll make a change. Do you think that skipping extensions and protocol declarations might be sufficient here? Neither of those declarations supported nested classes, so that might cover any of this rule's cases.

mt00chikin: Never mind, I don't think that's a good assumption on my part either. I'm going to update the visitor accordingly and get a commit up in a bit.

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -0,0 +1,137 @@
...
+
+    func makeVisitor(file: SwiftLintFile) -> ViolationsSyntaxVisitor {
+        Visitor(viewMode: .sourceAccurate)
+    }
+}
+
+private extension PrivateSwiftUIStatePropertyRule {
+    final class Visitor: ViolationsSyntaxVisitor {
+        override var skippableDeclarations: [DeclSyntaxProtocol.Type] {
+            [EnumDeclSyntax.self, FunctionDeclSyntax.self, ProtocolDeclSyntax.self, VariableDeclSyntax.self]
```
SimplyDanny, line 82: Even though unlikely, types can be defined almost everywhere, especially also in enum, function and variable declarations. That said, protocols are actually the only declaration type that can be ignored.

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
+            guard let inheritedTypeCollection = node.inheritanceClause?.inheritedTypeCollection else {
+                return .skipChildren
+            }
+
+            let inheritedTypes = inheritedTypeCollection.typeNames
+            return inheritedTypes.contains("View") ? .visitChildren : .skipChildren
```
SimplyDanny, line 91: Could be extracted into a function or extension.

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -0,0 +1,137 @@
...
+    }
+}
+
+private extension PrivateSwiftUIStatePropertyRule {
+    final class Visitor: ViolationsSyntaxVisitor {
+        override var skippableDeclarations: [DeclSyntaxProtocol.Type] {
+            [EnumDeclSyntax.self, FunctionDeclSyntax.self, ProtocolDeclSyntax.self, VariableDeclSyntax.self]
+        }
+
+        override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
```
SimplyDanny, line 85: Also think about actors ... 😉

SimplyDanny (commented):
I'm afraid I still have some remarks.

mt00chikin:
No worries, @SimplyDanny! I appreciate the feedback. Let me know if there's anything else in my latest commit that might have been overlooked.

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -0,0 +1,157 @@
...
+
+            return inheritedTypeCollection.conformsToViewProtocol ? .visitChildren : .skipChildren
+        }
+
+        override func visit(_ node: StructDeclSyntax) -> SyntaxVisitorContinueKind {
+            guard let inheritedTypeCollection = node.inheritanceClause?.inheritedTypeCollection else {
+                return .skipChildren
+            }
+
+            return inheritedTypeCollection.conformsToViewProtocol ? .visitChildren : .skipChildren
```
SimplyDanny, line 108: Maybe we can reduce duplication by implementing this as
```suggestion
           node.inheritanceClause?.conformsToViewProtocol == true ? .visitChildren : .skipChildren
```

mt00chikin: Seems like a reasonable request to me, I'll move this logic to TypeInheritanceClauseSyntax

SimplyDanny:
There are actually two more things that I've just thought of ...

When a declaration is nested inside another declaration implementing `View` like

```swift
struct ContentView: View {
    struct OtherStruct {
        @State var isPlaying: Bool = false   // rule doesn't trigger
    }
}
```

the rule doesn't trigger in the nested type. That's fine. However, if it's the other way around as in

```swift
struct OtherStruct {
    struct ContentView: View {
        @State var isPlaying: Bool = false    // rule doesn't trigger
    }
}
```

it also doesn't trigger and that'd be incorrect. It should trigger. We'd have to keep track of the scope we are currently in instead of skipping the type's body completely.

Other than that, Apple's documentation you referred to states

> Create a state value in an [App](https://developer.apple.com/documentation/swiftui/app), [Scene](https://developer.apple.com/documentation/swiftui/scene), or [View](https://developer.apple.com/documentation/swiftui/view) by applying the `@State` attribute to a property declaration and providing an initial value.

which means the rule should not only check for conformance to `View` but to `App` and `Scene` as well.

SimplyDanny:
Let's see if the following example works:
```swift
struct MyStruct {
    struct ContentView: View {
        @State ↓var isPlaying: Bool = false
    }
    @State var isPlaying: Bool = false
}
```

mt00chikin:
Ok, I fundamentally misinterpreted `visitPost(_:)` before today. Needed to read the docs more thoroughly, that was my mistake. I believe I have your latest test case working correctly now, but I'm going to do some additional testing and try to clean the code up a bit where possible.

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -0,0 +1,246 @@
...
+private extension PrivateSwiftUIStatePropertyRule {
+    final class Visitor: ViolationsSyntaxVisitor {
+        override var skippableDeclarations: [DeclSyntaxProtocol.Type] {
+            [ProtocolDeclSyntax.self]
+        }
+
+        /// LIFO stack that is stores type inheritance clauses for each visited node
+        /// The last value is the inheritance clause for the most recently visited node
+        /// A nil value indicates that the node does not provide any inheritance clause
+        private var visitedTypeInheritances: [TypeInheritanceClauseSyntax?] = []
```
SimplyDanny, line 169: There is a `Stack` type in SwiftLint specifically for such use cases.

SimplyDanny (approved):
One more tip ...

Apart from this little improvement, I think this looks good now. At least I cannot think of any more cases to be considered for the time being.

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
mt00chikin: Nice! Updated

SimplyDanny:
This is going to be merged soon. Thank you for your patience!

mt00chikin:
Thank you! I appreciate the thorough and patient review feedback. I definitely learned a lot about SwiftSyntax from this experience.

calebwilson706:
Has this rule been released? I'm trying to use it but I'm getting 'warning: 'private_swiftui_state' is not a valid rule identifier' and the response from 'swiftlint rules' doesn't contain it. Im on version '0.52.4'

SimplyDanny:
> Has this rule been released? I'm trying to use it but I'm getting 'warning: 'private_swiftui_state' is not a valid rule identifier' and the response from 'swiftlint rules' doesn't contain it. Im on version '0.52.4'

No, it's not yet been released.

ObjectiveCesar:
I personally think the same rule should apply for `@StateObject`.

DivineDominion:
You and Apple both :)
https://developer.apple.com/documentation/swiftui/stateobject#

Just want to say that this is not a matter of preference whether a `@StateObject` should be changed from "outside" its owning view. One can try, but it's a source of problems.

SimplyDanny:
> I personally think the same rule should apply to `@StateObject`.

This is easy to achieve. Anyone willing to open a PR? 😊

mt00chikin:
If no one has already started on that work, I can probably find some time to add it this week or next! I actually had considered adding this from the outset, but it just got lost along the way. I'd be happy to add it, I think that it's a good addition to the rule

ObjectiveCesar:
> If no one has already started on that work, I can probably find some time to add it this week or next! I actually had considered adding this from the outset, but it just got lost along the way. I'd be happy to add it, I think that it's a good addition to the rule

Thank you for making the effort. I certainly appreciate it.

mt00chikin:
FYI, I've opened https://github.com/realm/SwiftLint/pull/5200 to address this feedback and apply the rule to StateObject as well. Feel free to comment on that MR with any additional concerns or requests

PR #4770, opened 2023-02-17
Remove the need for rules_xcodeproj dependency for downstream consumers

thii:
This was needed because the `xcodeproj` target was declared in the same
BUILD file with the `@SwiftLint//:swiftlint` target, so when users run
`bazel run @SwiftLint//:swiftlint`, bazel would try to load
`rules_xcodeproj`. `rules_xcodeproj` 1.0 allows us to specify the
location of the generated Xcode project, so we don't need this target in
the top-level BUILD file anymore.

This only fixes for non-bzlmod case now.

Fixes https://github.com/realm/SwiftLint/issues/4737

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -4,11 +4,6 @@ load(
     "swift_binary",
     "swift_library",
 )
-load(
-    "@com_github_buildbuddy_io_rules_xcodeproj//xcodeproj:defs.bzl",
-    "xcode_schemes",
-    "xcodeproj",
-)
 
 # Targets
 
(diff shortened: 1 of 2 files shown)


jpsim:
I'm happy to merge this now but the PR is marked as draft. Did you still want to work on this before I merge?

thii:
It should be ready now. I was waiting for CI to be green before marking it as ready to reduce notification noises.

jpsim:
Merged in https://github.com/realm/SwiftLint/pull/5112, thanks for your contribution!

Issue #4666, opened 2023-01-02
Rule Request: no_duplicated_condition


lo1tuma:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

1. Why should this rule be added?

This rule should flag conditions that are used multiple times in the same branching statement like if-else-if or switch statement.

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

Violating Examples

```swift
if x < 5 {
  foo()
} else if x < 2 {
  bar()
} else if x < 5 { // duplicated condition
  baz()
}

switch x {
case "a":
  foo()
case "a" // duplicated condition
  bar()
}
```

Non-Violating Examples

```swift
if x < 5 {
  foo()
  baz()
} else if x < 2 {
  bar()
}

switch x {
case "a":
  foo()
  bar()
}
```

3. Should the rule be configurable, if so what parameters should be configurable?

This rule doesn’t need to be configurable.

4. Should the rule be opt-in or enabled by default? Why?
   See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a rule as opt-in.

This would be a good candidate for a builtin rule as it points out bad structured code.

PR #4771, opened 2023-02-17
Add new `duplicate_conditions` rule

1in1:
Long time user, first time contributor - let me know if anything's wrong!

__I can't run Sourcery at the moment__ (I don't have access to a mac for a while, this change was made on a linux box -- Sourcery is... non-trivial to install, shall we say); so if that's required for this change, and a maintainer is able to help with that when this is otherwise ready, I'd be appreciative!

This PR adds the `duplicate_conditions` rule described [here](https://github.com/realm/SwiftLint/issues/4666)

`if-else` and `switch` branching statements are considered. Compiler directives are not considered. The behaviour is very slightly different between `if-else` and `switch` statements:
- For `if-else` statements, a violation is triggered if the conditions of one branch are the same as the conditions of another branch, ignoring ordering. 
  - I considered flagging individual conditions, but I think something like the following is reasonable stylistically, in some scenarios, and so should not be a violation:
```
if x < 5, y == "s" { foo() }
else if x < 5 { bar() }
```
Additionally, to cover this case generally, some reasonably complicated special casing would be needed to prevent the following from being a violation:
```
if let x = maybeX1, y = x.p { foo() }
else if let x = maybeX2, y = x.p { bar() }
```
- For `switch` statements, a violation is triggered if any of the conditions (including `when` clause, if applicable) in any of the case labels match

See the examples for more :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@
 * Add `duplicate_conditions` rule which warns when a condition is duplicated
   in separate branches of the same branching statement (if-else, or switch).  
   [1in1](https://github.com/1in1)
-  [#issue_number](https://github.com/realm/SwiftLint/issues/4666)
+  [#4666](https://github.com/realm/SwiftLint/issues/4666)
 
 * Add local links to rule descriptions to every rule listed
   in `Rule Directory.md`.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@
 
 #### Enhancements
 
+* Add `duplicate_conditions` rule which warns when a condition is duplicated
+  in separate branches of the same branching statement (if-else, or switch).  
+  [1in1](https://github.com/1in1)
+  [#issue_number](https://github.com/realm/SwiftLint/issues/4666)
+
 * Add local links to rule descriptions to every rule listed
   in `Rule Directory.md`.  
   [kattouf](https://github.com/kattouf)
--- a/Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
@@ -8,7 +8,7 @@ struct DuplicateConditionsRule: SwiftSyntaxRule, ConfigurationProviderRule {
     static let description = RuleDescription(
         identifier: "duplicate_conditions",
         name: "Duplicate Conditions",
-        description: "Duplicate conditions in the same branching statement should be avoided",
+        description: "Duplicate sets of conditions in the same branching statement should be avoided",
         kind: .lint,
         nonTriggeringExamples: [
             Example("""
@@ -27,6 +27,13 @@ struct DuplicateConditionsRule: SwiftSyntaxRule, ConfigurationProviderRule {
                 }
             """),
             Example("""
+                if x < 5, y == "s" {
+                  foo()
+                } else if x < 5 {
+                  bar()
+                }
+            """),
+            Example("""
                 switch x {
                 case \"a\":
                   foo()
@@ -34,9 +41,24 @@ struct DuplicateConditionsRule: SwiftSyntaxRule, ConfigurationProviderRule {
                 }
             """),
             Example("""
+                switch x {
+                case \"a\" where y == "s":
+                  foo()
+                case \"a\" where y == "t":
+                  bar()
+                }
+            """),
+            Example("""
                 if let x = maybeAbc {
                   foo()
                 } else if let x = maybePqr {
+                  bar()
+                }
+            """),
+            Example("""
+                if let x = maybeAbc, let z = x.maybeY {
+                  foo()
+                } else if let x = maybePqr, let z = x.maybeY {
                   bar()
                 }
             """),
@@ -54,16 +76,16 @@ struct DuplicateConditionsRule: SwiftSyntaxRule, ConfigurationProviderRule {
                   foo()
                 } else if y == "s" {
                   bar()
-                } else if ↓x < 5 { 
+                } else if ↓x < 5 {
                   baz()
                 }
             """),
             Example("""
-                if x < 5, ↓y == "s" {
+                if ↓x < 5, y == "s" {
                   foo()
                 } else if x < 10 {
                   bar()
-                } else if ↓y == "s", x < 15 {
+                } else if ↓y == "s", x < 5 {
                   baz()
                 }
             """),
@@ -76,9 +98,24 @@ struct DuplicateConditionsRule: SwiftSyntaxRule, ConfigurationProviderRule {
                 }
             """),
             Example("""
+                switch x {
+                case ↓\"a\" where y == "s":
+                  foo()
+                case ↓\"a\" where y == "s":
+                  bar()
+                }
+            """),
+            Example("""
                 if ↓let xyz = maybeXyz {
                   foo()
                 } else if ↓let xyz = maybeXyz {
+                  bar()
+                }
+            """),
+            Example("""
+                if let x = maybeAbc, let z = x.maybeY {
+                  foo()
+                } else if let x = maybeAbc, let z = x.maybeY {
                   bar()
                 }
             """),
@@ -120,34 +157,35 @@ private extension DuplicateConditionsRule {
                 maybeCurr = curr.elseBody?.as(IfStmtSyntax.self)
             }
 
-            let positionsByCondition = statementChain
-                .flatMap { $0.conditions }
-                .compactMap(extract)
-                .reduce(into: [[UInt8]: [AbsolutePosition]](), { xs, x in
-                    xs[x.text, default: []].append(x.location)
-                })
+            let positionsByConditions = statementChain
+                .reduce(into: [Set<[UInt8]>: [AbsolutePosition]]()) { acc, elt in
+                    let conditions = elt.conditions.compactMap(extract)
+                    let location = elt.conditions.positionAfterSkippingLeadingTrivia
+                    acc[Set(conditions), default: []].append(location)
+                }
 
-            addViolations(positionsByCondition)
+            addViolations(positionsByConditions)
         }
 
         override func visitPost(_ node: SwitchCaseListSyntax) {
             let switchCases = node.compactMap { $0.as(SwitchCaseSyntax.self) }
 
             let positionsByCondition = switchCases
-                .reduce(into: [[UInt8]: [AbsolutePosition]]()) { xs, x in
+                .reduce(into: [[UInt8]: [AbsolutePosition]]()) { acc, elt in
                     // Defaults don't have a condition to worry about
-                    guard case let .case(caseLabel) = x.label else { return }
-                    for item in caseLabel.caseItems {
-                        let pattern = item.pattern.withoutTrivia().syntaxTextBytes
-                        let location = item.positionAfterSkippingLeadingTrivia
-                        xs[pattern, default: []].append(location)
+                    guard case let .case(caseLabel) = elt.label else { return }
+                    for caseItem in caseLabel.caseItems {
+                        let pattern = caseItem.pattern.withoutTrivia().syntaxTextBytes
+                        let whereClause = caseItem.whereClause?.withoutTrivia().syntaxTextBytes ?? []
+                        let location = caseItem.positionAfterSkippingLeadingTrivia
+                        acc[pattern + whereClause, default: []].append(location)
                     }
                 }
 
             addViolations(positionsByCondition)
         }
 
-        private func extract(_ node: ConditionElementSyntax) -> (text: [UInt8], location: AbsolutePosition)? {
+        private func extract(_ node: ConditionElementSyntax) -> [UInt8]? {
             let text: [UInt8]
             switch node.condition {
             case .availability(let node):
@@ -162,10 +200,10 @@ private extension DuplicateConditionsRule {
                 return nil
             }
 
-            return (text: text, location: node.positionAfterSkippingLeadingTrivia)
+            return text
         }
 
-        private func addViolations(_ positionsByCondition: [[UInt8]: [AbsolutePosition]]) {
+        private func addViolations<T>(_ positionsByCondition: [T: [AbsolutePosition]]) {
             let duplicatedPositions = positionsByCondition
                 .filter { $0.value.count > 1 }
                 .flatMap { $0.value }
--- a/Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
+++ b/Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
@@ -154,22 +154,22 @@ struct DuplicateConditionsRule: SwiftSyntaxRule, ConfigurationProviderRule {
 
 private extension DuplicateConditionsRule {
     final class Visitor: ViolationsSyntaxVisitor {
-        override func visitPost(_ node: IfStmtSyntax) {
-            if  node.parent?.is(IfStmtSyntax.self) == true {
+        override func visitPost(_ node: IfExprSyntax) {
+            if  node.parent?.is(IfExprSyntax.self) == true {
                 // We can skip these cases - they will be picked up when we visit the top level `if`
                 return
             }
 
-            var maybeCurr: IfStmtSyntax? = node
-            var statementChain: [IfStmtSyntax] = []
+            var maybeCurr: IfExprSyntax? = node
+            var statementChain: [IfExprSyntax] = []
             while let curr = maybeCurr {
                 statementChain.append(curr)
-                maybeCurr = curr.elseBody?.as(IfStmtSyntax.self)
+                maybeCurr = curr.elseBody?.as(IfExprSyntax.self)
             }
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,11 @@
 
 #### Enhancements
 
+* Add `duplicate_conditions` rule which warns when a condition is duplicated
+  in separate branches of the same branching statement (if-else, or switch).  
+  [1in1](https://github.com/1in1)
+  [#issue_number](https://github.com/realm/SwiftLint/issues/4666)
```
SimplyDanny, line 24:
```suggestion
  [#4666](https://github.com/realm/SwiftLint/issues/4666)
```

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+
+struct DuplicateConditionsRule: SwiftSyntaxRule, ConfigurationProviderRule {
+    var configuration = SeverityConfiguration(.warning)
+
+    init() {}
+
+    static let description = RuleDescription(
+        identifier: "duplicate_conditions",
+        name: "Duplicate Conditions",
+        description: "Duplicate sets of conditions in the same branching statement should be avoided",
```
SimplyDanny, line 11:
```suggestion
        description: "Duplicate sets of conditions in the same branch condition should be avoided",
```

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+    func makeVisitor(file: SwiftLintFile) -> ViolationsSyntaxVisitor {
+        Visitor(viewMode: .sourceAccurate)
+    }
+}
+
+private extension DuplicateConditionsRule {
+    final class Visitor: ViolationsSyntaxVisitor {
+        override func visitPost(_ node: IfStmtSyntax) {
+            if let prevToken = node.previousToken,
+               prevToken.tokenKind == .elseKeyword {
```
SimplyDanny, line 148: We typically check that like `node.parent.is(IfExprSyntax.self)` which is a bit more direct.

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+                }
+
+            addViolations(positionsByCondition)
+        }
+
+        private func extract(_ node: ConditionElementSyntax) -> [UInt8]? {
+            let text: [UInt8]
+            switch node.condition {
+            case .availability(let node):
+                text = node.withoutTrivia().syntaxTextBytes
```
SimplyDanny, line 192: `.syntaxTextBytes` fails when condition elements incorporate different whitespaces. While presumably a bit slower `.debugDescription(includeChildren: true, includeTrivia: false)` would catch these cases.

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+
+            return text
+        }
+
+        private func addViolations<T>(_ positionsByCondition: [T: [AbsolutePosition]]) {
+            let duplicatedPositions = positionsByCondition
+                .filter { $0.value.count > 1 }
+                .flatMap { $0.value }
+
+            violations.append(contentsOf: duplicatedPositions)
```
SimplyDanny, line 211: Perhaps a matter of taste, but the violations should only occur at real duplicates. Thus, the first condition shouldn't have a violation marker.

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+            case .optionalBinding(let node):
+                text = node.withoutTrivia().syntaxTextBytes
+            default:
+                return nil
+            }
+
+            return text
+        }
+
+        private func addViolations<T>(_ positionsByCondition: [T: [AbsolutePosition]]) {
```
SimplyDanny, line 206: Would be enough to pass a `[[AbsolutePosition]]` in here.

SimplyDanny (changes requested):
Thank you for contributing this new rule!

It looks good already. Yet, I have some remarks you may consider:

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
1in1: I don't think my original choice of phrasing was very good, but I'm also not certain about `branch condition`; to me this means, for example, `x < 5` in `if x < 5 { foo() } else { bar() }`. What do you think about `branch instruction` instead?

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
1in1: Much nicer! Thank you

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
1in1: Good spot, changed

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
1in1: I agree in principle, and originally wrote it this way. But in practice, these errors happen when there are lots of cases, and spotting the one that's wrong might be tricky without indicating both of the pair (see some of the examples picked up by SwiftLintBot). This would also be in line with `DuplicateEnumCasesRule`.

I don't have strong feelings on this at all though, and I'm very happy to change this if you reckon it's cleaner :)

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
1in1: Generic dropped by moving to `any Collection` (using an Array would mean wrapping `.values` in an Array initialiser)

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
SimplyDanny: You brought a convincing argument. I'm fine with the current way of reporting on every duplicate. 👍

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
```diff
@@ -0,0 +1,227 @@
...
+                            ?? ""
+                        let location = caseItem.positionAfterSkippingLeadingTrivia
+                        acc[pattern + whereClause, default: []].append(location)
+                    }
+                }
+
+            addViolations(positionsByCondition.values)
+        }
+
+        private func extract(_ node: ConditionElementSyntax) -> String {
```
SimplyDanny, line 203: I think this method now boils down to only the single expression/statement `node.debugDescription(includeChildren: true, includeTrivia: false)`, i.e. the switch statement isn't needed anymore.

SimplyDanny:
I've just added a commit to fix the Sourcery test failure.

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
1in1: Yup sounds good; I think they're very mildly different, e.g., 
```
ConditionElementSyntax children=1
  0: IdentifierExprSyntax children=1
    0: identifier("z")
```
vs
```
IdentifierExprSyntax children=1
  0: identifier("z")
```

But I'd agree, I don't think the distinction matters in this context. Thanks!

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
1in1: Perfect :)

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
1in1: Have just been playing with this, and I think it's correct as is; it's subtle, though.

The use of `node` in `node.debugDescription` is referring to the inner `node`, from the `case let` in the switch statements; e.g., of type `SequenceExprSyntax`, or `ExprListSyntax`, etc. If we get rid of the switch statements, we are throwing away the 'unwrapping', and looking just at the outer `ConditionElementSyntax` type.

This is an issue when we have multiple conditions on the same line, because the outer `ConditionElementSyntax` keeps information such as trailing commas. E.g., comparing strings such as

```
ConditionElementSyntax children=2
  0: SequenceExprSyntax children=1
    0: ExprListSyntax children=3
      0: IdentifierExprSyntax children=1
        0: identifier("y")
      1: BinaryOperatorExprSyntax children=1
        0: binaryOperator("==")
      2: StringLiteralExprSyntax children=3
        0: stringQuote
        1: StringLiteralSegmentsSyntax children=1
          0: StringSegmentSyntax children=1
            0: stringSegment("s")
        2: stringQuote
  1: comma
```

Behaves differently to comparing strings such as
```
SequenceExprSyntax children=1
  0: ExprListSyntax children=3
    0: IdentifierExprSyntax children=1
      0: identifier("y")
    1: BinaryOperatorExprSyntax children=1
      0: binaryOperator("==")
    2: StringLiteralExprSyntax children=3
      0: stringQuote
      1: StringLiteralSegmentsSyntax children=1
        0: StringSegmentSyntax children=1
          0: stringSegment("s")
      2: stringQuote
```

I'll rename the inner `node` to something else to avoid the name overload, but I think the behaviour as written is correct

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
1in1: Ha, frustratingly, Jazzy is freaking out over the use of `any` or `some` when trying to generate the docs:
```
error: cannot specialize protocol type 'Collection'
```

`[[AbsolutePosition]]`` it is :+1:

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
```diff
@@ -0,0 +1,227 @@
+import SwiftSyntax
+
+struct DuplicateConditionsRule: SwiftSyntaxRule, ConfigurationProviderRule {
+    var configuration = SeverityConfiguration(.warning)
```
lo1tuma, line 4: I wonder if this should be even a error and on by default. Since a duplicated condition in a branching statement usually means one of the branches gets never executed, thus it is in the best case just dead code and in the worst case it is a bug.

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
```diff
@@ -0,0 +1,227 @@
...
+                }
+            """),
+            Example("""
+                if case .p = x {
+                  foo()
+                } else if case .q = x {
+                  bar()
+                }
+            """)
+        ],
```
lo1tuma, line 72: What do you think about adding a non-triggering example of two nested if-statements with the same condition?

```swift
if true {
    if true { foo() }
}
```

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
```diff
@@ -0,0 +1,227 @@
...
+                }
+            """),
+            Example("""
+                if ↓x < 5, y == "s" {
+                  foo()
+                } else if x < 10 {
+                  bar()
+                } else if ↓y == "s", x < 5 {
+                  baz()
+                }
```
lo1tuma, line 101: Is it also possible to mix `,` and `&&` and still get the duplicated condition detected?

```swift
if x < 5, y == "s" {
  foo()
} else if x < 10 {
  bar()
} else if y == "s" && x < 5 {
  baz()
}
```

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
```diff
@@ -0,0 +1,227 @@
...
+                if ↓#available(macOS 10.15, *) {
+                  foo()
+                } else if ↓#available(macOS 10.15, *) {
+                  bar()
+                }
+            """),
+            Example("""
+                if ↓case .p = x {
+                  foo()
+                } else if ↓case .p = x {
```
lo1tuma, line 143: I would suggest to add one more triggering example, that has more than one duplication, e.g.:

```swift
if ↓true {} else if ↓true {} else if ↓true {} else {}
```

lo1tuma (commented):
Thank you for implementing this rule. Can’t wait to enable it in my projects.

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
1in1: So, in theory, yes. But I think doing it right makes the implementation much, much trickier

When we have `x < 5, y == "s"`, the syntax tree pulls these apart for us as distinct `SequenceExprSyntax` nodes. We can easily process them separately. If we have `x < 5 && y == "s"`, this is now a single `SequenceExprSyntax` node, with an `ExprListSyntax` with a bunch of children of differing types. One of these is the `BinaryOperatorExprSyntax` we want, the `&&` operator, so we'd have to break apart the list on those operators, and look at those individually:
```
0: SequenceExprSyntax children=1
  0: ExprListSyntax children=7
    0: IdentifierExprSyntax children=1
      0: identifier("y")
    1: BinaryOperatorExprSyntax children=1
      0: binaryOperator("==")
    2: StringLiteralExprSyntax children=3
      0: stringQuote
      1: StringLiteralSegmentsSyntax children=1
        0: StringSegmentSyntax children=1
          0: stringSegment("s")
      2: stringQuote
    3: BinaryOperatorExprSyntax children=1
      0: binaryOperator("&&")
    4: IdentifierExprSyntax children=1
      0: identifier("x")
    5: BinaryOperatorExprSyntax children=1
      0: binaryOperator("<")
    6: IntegerLiteralExprSyntax children=1
      0: integerLiteral("5")
```


But then what happens if we wrap in brackets? If we instead have `if (x < 5 && y == "s")`, then even though this is "the same", we now have a `TupleExprSyntax` wrapping the original `SequenceExprSyntax`. So we'd have to figure out that we need to drill down through these as many times as we need, before figuring out the `&&`'s. I would imagine there are also concerns about operator precedence that would need to be accounted for (in particular, note the `==` is at the same level as the `&&` in the above breakdown; is it obvious how to split up `p && q || r && s`?), and that there could be issues I haven't thought of...

I don't know this codebase, so if you know of any tools that already exist to help with this complexity, I'm very open to trying this out! But otherwise, I think limiting the scope to ensure it's correct might be safer. Thoughts?

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
1in1: Makes sense to me --- will change!

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
lo1tuma: You are totally right. I don’t know if there is any easy way to achieve this already, since I’m also new to the codebase.
I think this feature would be nice to have, but for an first iteration it is probably not needed. Maybe we can create a follow-up issue to track this feature.

SimplyDanny: > I don't know this codebase, so if you know of any tools that already exist to help with this complexity, I'm very open to trying this out! But otherwise, I think limiting the scope for now to ensure it's correct might be safer. Thoughts?

I agree that the current scope is sufficient as a first implementation. However, the functionality you are looking for, that is converting an expression list into a syntax tree taking precedence rules into account, already exists. See `IdenticalOperandsRule` as an example on how this works.

Source/SwiftLintFramework/Rules/Lint/DuplicateConditionsRule.swift
SimplyDanny: You are right. The comma might interfere. So let me correct my proposal to
```swift
node.with(\.trailingComma, nil).debugDescription(includeChildren: true, includeTrivia: false)
```
Honestly, the only thing that bothers me in the current switch is the duplication of the `.debugDescription(includeChildren: true, includeTrivia: false)` calls.

1in1: My bad; I wasn't taking `.condition`. You're right and we can do away with the switch. Done

SimplyDanny (approved):
Looks good to me now. Well done!

PR #4773, opened 2023-02-17
Add imports for typealias dependencies

keith:
Starting with Swift 5.8 there is a warning if you use a typealias that
aliases a type from a module that you don't import. This handles that
case by extracting the module name from the type alias' type.

Example:

```
// Module A
public struct Foo { ... }
```

```
// Module B
import A
public typealias Bar = Foo
```

```
// Module C
import B

func baz(arg: Bar) { ... }
```

In this example module C must now `import A` in order to avoid this
warning.


keith:
dropping this for now because this warning went away after beta 1

PR #4777, opened 2023-02-20
Proof-of-concept: Async Rules

jpsim:
Using the new SourceKitten async send support introduced in https://github.com/jpsim/SourceKitten/pull/762.


jpsim:
I expect this to perform much worse than `main` right now because we're now using multiple caches for SourceKit editor-open requests. For more realistic performance numbers I'd need to get rid of the previous synchronous cache completely and migrate everything over.

lo1tuma:
What would be the use-case for an async rule?

lo1tuma:
Ah thanks for the explanation. Better performance would be nice to have. My naive thinking was that it would be enough for the rule runner to by async and parallelize the execution of synchronous rules. But I’m still a swift newbie and don’t know the concurrency model in detail yet.

PR #4784, opened 2023-02-25
Improved `identifier_name` documentation

### LINKED ISSUE #4767: identifier_name issue didn't show warning
### New Issue Checklist

### Describe the bug

 In an exception to the above, variable names may start with a capital letter when they are **declared static and immutable.**
Above is the sentence on SwiftLint Rules, I found out that when I declared  `static var AbCvar = 0`, and Xcode didn't show the warning because I realize the rule mention static and immutable. What is the reason ?

### Environment

```yml
excluded:
  - Pods
disabled_rules:
  - trailing_whitespace
identifier_name:
  excluded: # excluded via string array
    - id
    - URL
    - url
    - app
file_length:
  warning: 500
  error: 800
type_body_length:
  - 500 # warning
  - 700 # error
```

mildm8nnered:
Resolves #4767 

I added a few more examples, and removed the reference to `immutable` in the documentation.

It still does seem a bit weird that

`static let Bar = 0` (outside a class) will trigger the identifier_name rule, but that 

```
class Foo {
    static let Bar = 0
}
```

Does not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -141,10 +141,6 @@
 
 * Fix false positives in `attributes` rule when using property wrappers
   with keypath arguments.  
-  [JP Simard](https://github.com/jpsim)
-
-* Fix false positives in `attributes` rule when using property wrappers
-  with keypath arguments.  
   [JP Simard](https://github.com/jpsim)	
 
 ## 0.50.3: Bundle of Towels
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,10 @@
 * Catch more valid `no_magic_numbers` violations.  
   [JP Simard](https://github.com/jpsim)
 
+* Improve `identifier_name` documentation.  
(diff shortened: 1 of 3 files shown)


SimplyDanny:
> It still does seem a bit weird that `static let Bar = 0` (outside a class) will trigger the `identifier_name` rule, ...

What do you mean by that? Static variables cannot be declared outside of a type.

SimplyDanny (approved):

SimplyDanny:
Thank you for the fix!

mildm8nnered:
> > It still does seem a bit weird that `static let Bar = 0` (outside a class) will trigger the `identifier_name` rule, ...
> 
> What do you mean by that? Static variables cannot be declared outside of a type.

In actual swift code, yep, but if you provide it as an Example, but not wrapped in a class, it shows up as a `.varGlobal` instead of a `.varStatic`. I guess this is related to the fact that many of the examples are not actually valid swift code, by themselves (e.g. see #3348).

PR #4793, opened 2023-03-03
Add new `no_constant_condition` rule

### LINKED ISSUE #4668: RuleRequest: no_constant_condition
### New Issue Checklist

### New rule request

Conditional branches where the condition constantly evaluates to `true` or `false` are useless because only one branch is executed always or never.

Violating Examples

```swift
if true {
    print("foo")
}

if false {
    print("foo")
}

if x && false {
    print("foo")
}

if x || true {
    print("foo")
}

switch "b" {
case "a":
    print("foo")
case "c":
    print("bar")
default:
    break
}
```

Non-Violating Examples

```swift
if x {
    print("foo")
}

switch x {
case "a":
    print("foo")
case "c":
    print("bar")
default:
    break
}
```

This rule doesn’t need to be configurable.

I think this rule would be a good candidate for the default rules since it points out a possible programming error.

lo1tuma:
This fixes #4668

### DIFF
--- a/Source/SwiftLintFramework/Rules/Idiomatic/NoConstantConditionRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/NoConstantConditionRule.swift
@@ -129,278 +129,277 @@ private enum ExpressionResult: Equatable {
     case dynamic
 }
 
-private func getTokenText(_ token: TokenSyntax) -> String {
-    return token.text
-}
-
 private enum StringLiteralResult {
     case constant(_ value: String)
     case dynamic
 }
 
-private func analyzeStringLiteral(_ expression: StringLiteralExprSyntax) -> StringLiteralResult {
-    var value = ""
+private extension StringLiteralExprSyntax {
+    func analyze() -> StringLiteralResult {
+        var value = ""
 
-    for literalSegment in expression.segments {
-        switch literalSegment {
-        case .expressionSegment:
-            return .dynamic
-        case let .stringSegment(stringSegment):
-            value += getTokenText(stringSegment.content)
-        }
-    }
-
-    return .constant(value)
-}
-
-private func analyzePrimitiveLiteralExpression(_ element: ExprSyntax) -> ExpressionResult {
-    if let booleanExpression = element.as(BooleanLiteralExprSyntax.self) {
-        let value = getTokenText(booleanExpression.booleanLiteral)
-        return .evaluatedConstant(.booleanValue(value == "true"))
-    }
-
-    if let integerExpression = element.as(IntegerLiteralExprSyntax.self) {
-        let value = getTokenText(integerExpression.digits)
-        return .evaluatedConstant(.integerValue(Int(value) ?? 0))
-    }
-
-    if let floatExpression = element.as(FloatLiteralExprSyntax.self) {
-        let value = getTokenText(floatExpression.floatingDigits)
-        return .evaluatedConstant(.floatValue(Float(value) ?? 0.0))
-    }
-
-    if let stringExpression = element.as(StringLiteralExprSyntax.self) {
-        let stringLiteralResult = analyzeStringLiteral(stringExpression)
-        switch stringLiteralResult {
-        case let .constant(value):
-            return .evaluatedConstant(.stringValue(value))
-        case .dynamic:
-            return .dynamic
-        }
-    }
-
-    if element.is(NilLiteralExprSyntax.self) {
-        return .evaluatedConstant(.nilValue)
-    }
-
-    return .dynamic
-}
-
-private func isBinaryOperator(_ operatorExpression: ExprSyntax, _ expectedOperator: String) -> Bool {
-    if let binaryOperator = operatorExpression.as(BinaryOperatorExprSyntax.self) {
-        let operatorText = getTokenText(binaryOperator.operatorToken)
-
-        return operatorText == expectedOperator
-    }
-
-    return false
-}
-
-private func isLogicalAndOperator(_ operatorExpression: ExprSyntax) -> Bool {
-    return isBinaryOperator(operatorExpression, "&&")
-}
-
-// swiftlint:disable:next function_body_length cyclomatic_complexity
-private func evaluateConstantInfixExpression(
-    operatorExpression: ExprSyntax,
-    leftOperandValue: ConstantValue,
-    rightOperandValue: ConstantValue) -> ConstantEvaluationResult {
-    if isLogicalAndOperator(operatorExpression) {
-        if case let .booleanValue(leftBooleanValue) = leftOperandValue,
-           case let .booleanValue(rightBooleanValue) = rightOperandValue {
-            return .evaluated(.booleanValue(leftBooleanValue && rightBooleanValue))
-        }
-    }
-
-    if isBinaryOperator(operatorExpression, "||") {
-        if case let .booleanValue(leftBooleanValue) = leftOperandValue,
-           case let .booleanValue(rightBooleanValue) = rightOperandValue {
-            return .evaluated(.booleanValue(leftBooleanValue || rightBooleanValue))
-        }
-    }
-
-    if isBinaryOperator(operatorExpression, "+") {
-        if case let .stringValue(leftStringValue) = leftOperandValue,
-           case let .stringValue(rightStringValue) = rightOperandValue {
-            return .evaluated(.stringValue(leftStringValue + rightStringValue))
-        }
-    }
-
-    if isBinaryOperator(operatorExpression, "==") {
-        if case let .booleanValue(leftBooleanValue) = leftOperandValue,
-           case let .booleanValue(rightBooleanValue) = rightOperandValue {
-            return .evaluated(.booleanValue(leftBooleanValue == rightBooleanValue))
-        }
-        if case let .stringValue(leftStringValue) = leftOperandValue,
-           case let .stringValue(rightStringValue) = rightOperandValue {
-            return .evaluated(.booleanValue(leftStringValue == rightStringValue))
-        }
-        if case let .integerValue(leftIntValue) = leftOperandValue,
-           case let .integerValue(rightIntValue) = rightOperandValue {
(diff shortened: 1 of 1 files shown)


Source/SwiftLintFramework/Rules/Idiomatic/NoConstantConditionRule.swift
```diff
@@ -0,0 +1,566 @@
+// swiftlint:disable file_length cyclomatic_complexity
+import SwiftSyntax
+
+struct NoConstantConditionRule: ConfigurationProviderRule, SwiftSyntaxRule {
+    var configuration = SeverityConfiguration(.warning)
+
+    init() {}
+
+    static let description = RuleDescription(
+        identifier: "no_constant_condtion",
```
mildm8nnered, line 10: `no_constant_condition`

Source/SwiftLintFramework/Rules/Idiomatic/NoConstantConditionRule.swift
```diff
@@ -0,0 +1,566 @@
...
+        name: "No constant condition",
```
mildm8nnered, line 11: "No Constant Condition"

Source/SwiftLintFramework/Rules/Idiomatic/NoConstantConditionRule.swift
```diff
@@ -0,0 +1,566 @@
...
+        description: "Disallow branching statements when the branching condition is constant",
```
mildm8nnered, line 12: Style for the description feels a tiny bit off (for example, none start with "Disallow"). Maybe something like "Branching statements with constant conditions should be avoided"

CHANGELOG.md
```diff
@@ -18,6 +18,10 @@
 
 #### Enhancements
 
+* Add `no_constant_condition` rule which warns when a `if`, `guard` or `switch` condition is a constant value that would never change at runtime.
```
mildm8nnered, line 21: This should be wrapped like the other entries

Source/SwiftLintFramework/Rules/Idiomatic/NoConstantConditionRule.swift
```diff
@@ -0,0 +1,566 @@
...
+            Example("guard ↓foo() && false {} else {}"),
+            Example("guard ↓2 < 3 {} else {}"),
+            Example("switch ↓2 < 3 { default: break }"),
+            Example("switch ↓\"foo\" { default: break }"),
+            Example("switch ↓foo() && false { default: break }")
+        ]
+    )
+
+    func preprocess(file: SwiftLintFile) -> SourceFileSyntax? {
+        return file.foldedSyntaxTree
```
mildm8nnered, line 104: returns not required in one liners

Source/SwiftLintFramework/Rules/Idiomatic/NoConstantConditionRule.swift
lo1tuma: Fixed.

Source/SwiftLintFramework/Rules/Idiomatic/NoConstantConditionRule.swift
lo1tuma: Done

Source/SwiftLintFramework/Rules/Idiomatic/NoConstantConditionRule.swift
lo1tuma: Thanks. I’ve used the wording from [`ESLint`](https://eslint.org/docs/latest/rules/no-constant-condition) as an inspiration, but I’m ok with your suggestion as well.

CHANGELOG.md
lo1tuma: Done

Source/SwiftLintFramework/Rules/Idiomatic/NoConstantConditionRule.swift
lo1tuma: I know, but personally I prefer being explicit. I think it improves readability because the person reading to code can immediately see the `return` statement and doesn’t have to count the amount of lines. That’s being said, if this is a requirement that this project wants to enforce I’m happy to change it (sounds like there should be linting rule for that 😁).

mildm8nnered: https://realm.github.io/SwiftLint/implicit_return.html will detect that, but it's disabled in SwiftLint's `.swiftlint.yml` :-)

I don't think there's any policy here, as I can see plenty of examples, elsewhere in the code, so I think you're fine to keep it if you prefer.

Source/SwiftLintFramework/Rules/Idiomatic/NoConstantConditionRule.swift
```diff
@@ -0,0 +1,566 @@
...
+    }
+
+    if element.is(NilLiteralExprSyntax.self) {
+        return .evaluatedConstant(.nilValue)
+    }
+
+    return .dynamic
+}
+
+private func isBinaryOperator(_ operatorExpression: ExprSyntax, _ expectedOperator: String) -> Bool {
```
mildm8nnered, line 189: If you look at the structure of some of the other rules, in cases with a pure function like this, operating on an `ExprSyntax` (or lots of the other common objects), they'll often go for

```
private extension ExprSyntax {
    func isBinaryOperator(_ expectedOperator: String) -> Bool {
        // some code here
    }
}
```

Just search for `extension ExprSyntax` to see some examples.

I think a lot of code here *could* be given a similar treatment.

I don't think that's really documented anywhere, or mandatory in anyway, but just an observation that a lot of the existing code seems to follow that pattern.

SimplyDanny:
Thank you for bringing up the idea and the contribution!

I have a few remarks regarding usefulness of this rule and implementation:

Meanwhile, SwiftLint scans 18 open source projects and apart from the one in NetNewsWire (which is quite strange but valid code) the findings are little and rather unimportant (test code). I guess the reason is that constant conditions can easily be found in reviews.

So let me ask: How did you come up with this rule? Did you experience good examples where such a rule would have been helpful?

Regarding implementation: The current structure is not easy to follow. It's hard to reason about completeness. Since we are dealing with a well structured model (an AST), it seems reasonable to base such an analysis on the infrastructure that's provided by SwiftSyntax, namely the different kinds of visitors. I definitely would prefer that over a hand-crafted AST traversal that is hard to follow.

I've prepared the first steps of an implementation that seems more logical to me [here](https://github.com/realm/SwiftLint/compare/main...SimplyDanny:SwiftLint:expression-evaluation-state). What do you think about that? Can you imagine to continue on it in the context of your rule implementation?

Apart from structure I see two more advantages:
* Expression evaluation is a reusable component.
* It can be tested in separation from any rule.

lo1tuma:
@SimplyDanny thanks for your feedback

> So let me ask: How did you come up with this rule? Did you experience good examples where such a rule would have been helpful?

I’m quite new Swift (started 6 months ago) and started to look into a big swift legacy project in my team. I’ve tried to remove a few unused features. Since there was a huge lack of automated tests, I’ve tried to change as little code as possible in order to avoid introducing bug. So every change I make is usually a baby-step and if the IDE is able to make the change automatically (like renaming a variable) I use that. So at some point I had to touch a complicated `if` condition, where the `then`-branch was not supposed to run for the feature I was trying to remove. It looked really complicated and since I was new to the project I didn’t even understood everything. So I’ve only removed a small part of the condition that controlled the feature I was removing (something like `&& featureFooIsEnabled`). After that, I’ve tried to understand the remaining condition a bit better and inlined a few variables. Then I realized the remaining condition always evaluates to `true`, so I can completely get rid of which improves the readability of the code significantly.

Then I wondered why the swift compiler as well as `SwiftLint` did not tell me about this. I’ve been quite involved with linters since more than 10 years, I’ve contributed to `ESLint` and some ESLint plugins and I was also a heavy user of it over the last 10 years. So I was somehow expecting a linter (or the compiler) in a different language should tell me about such things. As you can see [here](https://eslint.org/docs/latest/rules/no-constant-condition) ESLint has the same rule.


> Regarding implementation: The current structure is not easy to follow. 

What kind of structure are you referring to? The code structure? If so, it would be nice if you could elaborate a little bit. Since I’ve used plain functions most of the time, I don’t see a big problem with the structure.

> It's hard to reason about completeness. 

Do you mean completeness in terms of the expression evaluation algorithm? If so, then I agree. But I think we don’t need to aim for completeness. I think we can start with something and extend to functionality over time. The current approach treats everything it cannot evaluate as a dynamic expression. So in the worst case, the rule doesn’t detect some constant expressions. Another caveat could be, when people are overloading binary operators, so they behave differently as their default behavior.


> Since we are dealing with a well structured model (an AST), it seems reasonable to base such an analysis on the infrastructure that's provided by SwiftSyntax, namely the different kinds of visitors. I definitely would prefer that over a hand-crafted AST traversal that is hard to follow.

I’m not sure if it can be done with visitors only. For the evaluation we often need the direct relation of AST nodes with the parents and children and the visitor pattern doesn’t really guarantee that there are no other AST nodes in-between. But I would be open to it and would be happy to give it a try 😉.


> I've prepared the first steps of an implementation that seems more logical to me [here](https://github.com/realm/SwiftLint/compare/main...SimplyDanny:SwiftLint:expression-evaluation-state). What do you think about that? Can you imagine to continue on it in the context of your rule implementation?

Usually I prefer an iterative approach to solve something. Since this is the only rule right now, that needs expression evaluation, I don’t see an immediate need to make this a re-usable component. First step is, to make it work for the given use-case and then it can be extracted and used for other use-cases. But that’s just how I work. In this particular case I think extracting this makes sense (that’s also what ESLint did). Maybe it could be useful to extend the [`duplicate_conditions` rule](https://github.com/realm/SwiftLint/pull/4771#discussion_r1124324015). So long story short: the work you did on your branch looks quite good so far, and I would happy to have a generic and re-usable solution in the long-term.

PR #4799, opened 2023-03-04
Fix/improve command parsing in multiline comments

### LINKED ISSUE #4798: superfluous_disable_command in a multiline comment generates multiple false alarms for `x is not a valid SwiftLint rule`
### New Issue Checklist

### Describe the bug

Given the following input:

```
/*extension SidebarViewController {
    // swiftlint:disable cyclomatic_complexity
    // Make sure we're right clicking a menu entry
}
*/
```

`superfluous_disable_command` will generate multiple false positives.

#4791 fixes the case of this that occurs in Aerial purely by chance (see the osstools output on that PR), because the problem stretch of code there happens to contain the word `all` which is interpreted as suppress all rules. However #4791 does not fix the problem for the above case, or any other cases in general.

```bash
% swiftlint lint --no-cache --use-stdin --enable-all-rules <<EOF
/*extension SidebarViewController {
    // swiftlint:disable cyclomatic_complexity
    // Make sure we're right clicking a menu entry
}
*/
EOF
<nopath>:5:3: warning: Superfluous Disable Command Violation: SwiftLint rule 'cyclomatic_complexity' did not trigger a violation in the disabled region. Please remove the disable command. (superfluous_disable_command)
<nopath>:5:3: warning: Superfluous Disable Command Violation: 'sure' is not a valid SwiftLint rule. Please remove it from the disable command. (superfluous_disable_command)
<nopath>:5:3: warning: Superfluous Disable Command Violation: '//' is not a valid SwiftLint rule. Please remove it from the disable command. (superfluous_disable_command)
<nopath>:5:3: warning: Superfluous Disable Command Violation: 'a' is not a valid SwiftLint rule. Please remove it from the disable command. (superfluous_disable_command)
<nopath>:5:3: warning: Superfluous Disable Command Violation: 'clicking' is not a valid SwiftLint rule. Please remove it from the disable command. (superfluous_disable_command)
<nopath>:5:3: warning: Superfluous Disable Command Violation: 'entry' is not a valid SwiftLint rule. Please remove it from the disable command. (superfluous_disable_command)
<nopath>:5:3: warning: Superfluous Disable Command Violation: 'right' is not a valid SwiftLint rule. Please remove it from the disable command. (superfluous_disable_command)
<nopath>:5:3: warning: Superfluous Disable Command Violation: '}' is not a valid SwiftLint rule. Please remove it from the disable command. (superfluous_disable_command)
<nopath>:5:3: warning: Superfluous Disable Command Violation: 'menu' is not a valid SwiftLint rule. Please remove it from the disable command. (superfluous_disable_command)
<nopath>:5:3: warning: Superfluous Disable Command Violation: 'Make' is not a valid SwiftLint rule. Please remove it from the disable command. (superfluous_disable_command)
<nopath>:5:3: warning: Superfluous Disable Command Violation: 'we're' is not a valid SwiftLint rule. Please remove it from the disable command. (superfluous_disable_command)
Done linting! Found 11 violations, 0 serious in 1 file.
```

### Environment

* Paste your configuration file: none

```
Xcode 14.2
Build version 14C18
```

See description above.

mildm8nnered:
Fixes #4798 

The easiest fix for this problem would be to ignore commands in block comments. With both the existing code, and with the changes proposed here, only the first command within a block comment will be detected in any case.

I guess there could be some people relying on commands working in block comments, and dropping that would definitely be a breaking change for them.

What I've done here is, in block comments, is to only parse the command up to the end of the line, instead of up to the end of the block comment, which preserves the current behaviour without the bugs.

I used the annoyingly verbose Swift String getLineStart() API to get the end of the lines while accounting for all possible line separators, but there is probably a better way to do that.

Of course, if we can just drop block comment support altogether, this PR would be a lot lot simpler ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@
 
 * Fix false positives in `indentation_width` rule.  
   [Sven Münnich](https://github.com/svenmuennich)
+
 * Fix false positives for superfluous_disable_command, and improved
   command parsing in multiline strings.  
   [Martin Redington](https://github.com/mildm8nnered)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -253,6 +253,10 @@
   by `disable` commands.  
   [Martin Redington](https://github.com/mildm8nnered)
   [#4788](https://github.com/realm/SwiftLint/issues/4788)
+* Fix false positives for superfluous_disable_command, and improved
+  command parsing in multiline strings.  
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#4798](https://github.com/realm/SwiftLint/issues/4798)
 
 * Fixed correction for `trailing_comma` rule wrongly removing trailing
   comments.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -253,6 +253,7 @@
   by `disable` commands.  
   [Martin Redington](https://github.com/mildm8nnered)
   [#4788](https://github.com/realm/SwiftLint/issues/4788)
+
 * Fix false positives for superfluous_disable_command, and improved
   command parsing in multiline strings.  
   [Martin Redington](https://github.com/mildm8nnered)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,10 @@
 
 * Fix false positives in `indentation_width` rule.  
   [Sven Münnich](https://github.com/svenmuennich)
+* Fix false positives for superfluous_disable_command, and improved
+  command parsing in multiline strings.  
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#4798](https://github.com/realm/SwiftLint/issues/4798)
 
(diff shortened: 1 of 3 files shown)


mildm8nnered:
The more I think about this, the more I think that just dropping command support in block comments is the way to go.

1) commands in multiline comments don't really work properly anyway - only the first command will be registered.
2) they're not documented in the readme, which only describes the `lineComment` mode.
3) if commands in multiline comments did actually work, they would just trigger a lot of `superfluous_disable_command` hits, as whatever they're suppressing would most likely not be triggering
4) block commenting out a bunch of code is not great, but is something that people probably do a lot in real life, and they should be able to do that and just forget about it (for a while)
5) The required changes in CommandVisitor would then be to just remove `.blockComment(let comment)` from the `case` statement.

mildm8nnered:
#4803 is probably a better solution

PR #4800, opened 2023-03-05
Use `all` optin rules facility for SwiftLint's own `.swiftlint.yml`

mildm8nnered:
Changes SwiftLint's own `.swiftlint.yml` to use the new `all` pseudo rule introduced in #4544 

By default, all opt-in (non-analyser) rules will be enabled, except for those specified in `disabled_rules`.

I started by setting up `disabled_rules` so that the exact same set of rules were enabled and disabled as before.

Then I enabled any rules which don't actually fire in the codebase (see [535d28](https://github.com/realm/SwiftLint/commit/535d28497f992d4c1f0a95952d433a25ea9b52b0)).

Finally, for rules that had simple fixes, or one a few violations, I either fixed them, or suppressed the violations. This included `legacy_objc_type`, `let_var_whitespace`, `prefer_self_in_static_references`, `balanced_xctest_lifecycle`, `discouraged_optional_boolean`, and `optional_enum_case_matching`.


Source/SwiftLintFramework/Rules/Lint/MissingDocsRule.swift
```diff
@@ -49,6 +49,7 @@ struct MissingDocsRule: OptInRule, ConfigurationProviderRule {
     }
 
     typealias ConfigurationType = MissingDocsRuleConfiguration
+
```
mildm8nnered, line 52: Blank line required by `let_var_whitespace`

Source/swiftlint/Helpers/LintOrAnalyzeArguments.swift
```diff
@@ -20,11 +20,12 @@ struct LintOrAnalyzeArguments: ParsableArguments {
     var config = [String]()
     @Flag(name: [.long, .customLong("autocorrect")], help: "Correct violations whenever possible.")
     var fix = false
+
     @Flag(help: """
```
mildm8nnered, line 24: I think there's a bug in `let_var_whitespace`, which is being triggered by the multiline `@Flag` - I'll open a defect for that once I have a good test case.

Source/swiftlint/Helpers/Signposts.swift
```diff
@@ -1,5 +1,6 @@
 #if canImport(os)
 import os.signpost
+
```
mildm8nnered, line 3: `let_var_whitespace`

Source/swiftlint/Helpers/LintOrAnalyzeArguments.swift
mildm8nnered: See #4801

.swiftlint.yml
```diff
@@ -40,7 +40,6 @@ disabled_rules:
   - no_magic_numbers
   - optional_enum_case_matching
   - prefer_nimble
-  - prefer_self_in_static_references
```
SimplyDanny: There is no common sense among the collaborators of this project whether this rule is helpful. It was disabled intentionally. We should rather keep it inactive.

SimplyDanny (approved):
Dogfooding! 🐶

.swiftlint.yml
mildm8nnered: Cool - I've re-disabled it.

PR #4812, opened 2023-03-11
Adds `unneeded_synthesized_initializer` rule.

mildm8nnered:
Shamelessly stolen from swift-format


Tests/GeneratedTests/GeneratedTests.swift
```diff
@@ -1,1364 +1,1370 @@
...
 
-class AccessibilityLabelForImageRuleGeneratedTests: SwiftLintTestCase {
+class AccessibilityLabelForImageRuleGeneratedTests: XCTestCase {
     func testWithDefaultConfiguration() {
         verifyRule(AccessibilityLabelForImageRule.description)
     }
 }
 
-class AccessibilityTraitForButtonRuleGeneratedTests: SwiftLintTestCase {
+class AccessibilityTraitForButtonRuleGeneratedTests: XCTestCase {
```
SimplyDanny, line 17: Unexpected changes ...

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
```diff
@@ -0,0 +1,252 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+
+import SwiftSyntax
+
+struct UnneededSynthesizedInitializerRule: SwiftSyntaxRule, ConfigurationProviderRule {
+    var configuration = SeverityConfiguration(.warning)
+
+    init() {}
```
SimplyDanny, line 19: Unneeded synthesized initializer. 😉

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
```diff
@@ -0,0 +1,252 @@
...
+// Adapted from swift-format's UseSynthesizedInitializer.swift
+//
+// https://github.com/apple/swift-format
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
```
SimplyDanny, line 12: My gut feeling says this header provides all the necessary information the licence text requires in case of redistribution.

Let @jpsim confirm or object.

SimplyDanny (commented):
I like this rule! Did you think about an auto-fix removing the initializers automatically?

mildm8nnered:
> I like this rule! Did you think about an auto-fix removing the initializers automatically?

I had not really thought about that, but will take a look once I'm confident in how I'm treating default initialisers.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
mildm8nnered: and the new warnings seem to check out as ok.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/ExpiringTodoConfiguration.swift
```diff
@@ -4,11 +4,6 @@ struct ExpiringTodoConfiguration: RuleConfiguration, Equatable {
 
         fileprivate(set) var opening: String
         fileprivate(set) var closing: String
-
```
mildm8nnered, line 7: automatically corrected by the rule

Source/SwiftLintCore/Protocols/SwiftSyntaxCorrectableRule.swift
```diff
@@ -19,8 +19,7 @@ public extension SwiftSyntaxCorrectableRule {
             return []
         }
 
-        file.write(newTree.description)
-        return rewriter
+        let corrections = rewriter
```
mildm8nnered, line 22: This change was necessary because given an input like

```
internal struct Foo {
    var bar: String

  ↓internal init(bar: String) {
        self.bar = bar
    }
}
```

and a correction like

```
internal struct Foo {
    var bar: String
}
```

The violation location is out of bounds in the corrected code. We now resolve the correction location in the original code, and then write out the corrected code.

mildm8nnered:
> I like this rule! Did you think about an auto-fix removing the initializers automatically?

ok - I think I've got correction working properly as well, so hopefully this should be good to go.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
```diff
@@ -0,0 +1,312 @@
...
+            }
+            return .skipChildren
+        }
+    }
+
+    final class Rewriter: SyntaxRewriter, ViolationsSyntaxRewriter {
+        var correctionPositions: [AbsolutePosition] = []
+        private var unneededInitializers: [InitializerDeclSyntax] = []
+
+        // private(set) var correctionPositions: [AbsolutePosition] = []
```
SimplyDanny: Remove?

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
```diff
@@ -0,0 +1,312 @@
...
+
+import SwiftSyntax
+
+struct UnneededSynthesizedInitializerRule: SwiftSyntaxCorrectableRule, ConfigurationProviderRule {
+    var configuration = SeverityConfiguration(.warning)
+
+    static let description = RuleDescription(
+        identifier: "unneeded_synthesized_initializer",
+        name: "Unneeded Synthesized Initializer",
+        description: "This initializer would be synthesized automatically - you do not need to define it",
```
SimplyDanny, line 22: Hint: If the description is not used as the violation message, it can be formulated in a more "descriptive" way explaining more details if that makes sense. You may even use full sentences.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
```diff
@@ -0,0 +1,312 @@
...
+        kind: .lint,
```
SimplyDanny, line 23: Not very important, but for me it seems that this rule would better fit into the `.idiomatic` category.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
```diff
@@ -0,0 +1,312 @@
...
+            let member = memberItem.decl
+            // Collect all stored variables into a list
+            if let varDecl = member.as(VariableDeclSyntax.self) {
+                let modifiers = varDecl.modifiers
+                if modifiers == nil {
+                    storedProperties.append(varDecl)
+                    continue
+                }
+                guard !modifiers.isStatic else { continue }
+                storedProperties.append(varDecl)
```
SimplyDanny, line 113: Can be condensed to
```suggestion
                if modifiers == nil || !modifiers.isStatic {
                    storedProperties.append(varDecl)
                }
```
can't it?

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
```diff
@@ -0,0 +1,312 @@
...
+                // Collect any possible redundant initializers into a list
```
SimplyDanny: Comment looks a bit misplaced here. Would it be better to put it into the block right after the condition?

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
```diff
@@ -0,0 +1,312 @@
...
+            } else if let initDecl = member.as(InitializerDeclSyntax.self) {
+                guard initDecl.optionalMark == nil else { continue }
+                guard initDecl.hasThrowsOrRethrowsKeyword == false else { continue }
```
SimplyDanny, line 117:
```suggestion
                if initDecl.hasThrowsOrRethrowsKeyword { continue }
```
or merge with previous `guard`.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
```diff
@@ -0,0 +1,312 @@
...
+                initializerBody(initializer.body, matches: storedProperties)
+            else {
+                continue
+            }
+            guard initializerModifiers(initializer.modifiers, match: storedProperties) else {
+                continue
+            }
+            guard initializer.isInlinable == false else {
+                continue
+            }
```
SimplyDanny, line 140:
```suggestion
            if initializer.isInlinable {
                continue
            }
```
or try to merge all `guard` statements into one.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
```diff
@@ -0,0 +1,312 @@
...
+            unneededInitializers.append(initializer)
+        }
+        return unneededInitializers
```
SimplyDanny, line 143: Could be `return initilizers.filter { ... }`.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
```diff
@@ -0,0 +1,312 @@
...
+}
+
+private extension VariableDeclSyntax {
+    var identifiers: [IdentifierPatternSyntax] {
+        var ids: [IdentifierPatternSyntax] = []
+        for binding in bindings {
+            guard let id = binding.pattern.as(IdentifierPatternSyntax.self) else { continue }
+            ids.append(id)
+        }
+        return ids
```
SimplyDanny, line 281:
```suggestion
        bindings.compactMap { $0.pattern.as(IdentifierPatternSyntax.self) }
```

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRuleExamples.swift
SimplyDanny, line 1: Indentation in the example snippets is 8 spaces instead of the preferred 4 spaces.

Other than that: Very well written comprehensive list of test cases! 👍

SimplyDanny (changes requested):
I have some more remarks being aware of the fact that some of them refer to copied code.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
mildm8nnered: Much nicer :-)

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRuleExamples.swift
mildm8nnered: > Indentation in the example snippets is 8 spaces instead of the preferred 4 spaces.
> 
> Other than that: Very well written comprehensive list of test cases! 👍

Thanks!

I went with all your suggestions, which tightened the code up a fair bit :-)

re indentation in the samples, I couldn't quite see what you mean - they look like 4 space indentations to me in Xcode and on GitHub, but maybe I'm missing something?

I think the only other thing (hopefully) is the CHANGELOG entry - there isn't a Main entry yet in `main` , and I didn't want to presume to add one ...

mildm8nnered: Fixed the change log now that the main section exists ...

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRuleExamples.swift
```diff
@@ -0,0 +1,426 @@
...
+    static let corrections = [
+        Example("""
+                struct Foo {
+                    let bar: String
+
+                   ↓init(bar: String) {
+                        self.bar = bar
+                    }
+
+                    func baz() {
```
SimplyDanny, line 304: Could you add a comment above the method declaration to proof that it is not removed by the rewriter?

SimplyDanny (commented):
There are still a few things that stick out. Are you still in the mood to do some more cleanup? 😬

mildm8nnered:
> There are still a few things that stick out. Are you still in the mood to do some more cleanup? 😬

Sure 😀

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
```diff
@@ -0,0 +1,286 @@
...
+import SwiftSyntax
+
+struct UnneededSynthesizedInitializerRule: SwiftSyntaxCorrectableRule, ConfigurationProviderRule {
+    var configuration = SeverityConfiguration(.warning)
+
+    static let description = RuleDescription(
+        identifier: "unneeded_synthesized_initializer",
+        name: "Unneeded Synthesized Initializer",
+        description: "Default or memberwise initializers that will be automatically synthesized " +
+                     "do not need to be manually defined",
```
SimplyDanny, line 23:
```suggestion
                     "do not need to be manually defined.",
```

mildm8nnered:
Worked through most of the CR comments, but still have a few left to do. Some of those `if` statements are indeed, very twisty.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
```diff
@@ -0,0 +1,286 @@
...
+                switch Syntax(element).as(SyntaxEnum.self) {
+                case .memberAccessExpr(let element):
+                    guard let base = element.base,
+                          base.description.trimmingCharacters(in: .whitespacesAndNewlines) == "self"
+                    else {
+                        return false
+                    }
+                    leftName = element.name.text
+                case .assignmentExpr(let element):
+                    guard element.assignToken.tokenKind == .equal else { return false }
```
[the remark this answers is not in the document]
mildm8nnered: So 

```
case .assignmentExpr(let element) where element.assignToken.tokenKind != .equal:
    return false
case .identifierExpr(let element):
    rightName = element.identifier.text
default:
    return false
}
```

does not work the same, because any `.assignmentExpr` that does not match the `where` clause will end up returning `false` in the `default` case, which is wrong.

To work, we'd need something like

```
case .assignmentExpr(let element) where element.assignToken.tokenKind != .equal:
    return false
case .assignmentExpr(_):
    break
case .identifierExpr(let element):
    rightName = element.identifier.text
default:
    return false
}
```

more than happy to change it to that, but at that point is it actually better/clearer than the original (ported) code?

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
```diff
@@ -0,0 +1,286 @@
...
+            // Ensure that parameters that correspond to properties declared using 'var' have a default
+            // argument that is identical to the property's default value. Otherwise, a default argument
+            // doesn't match the memberwise initializer.
+            let isVarDecl = property.bindingKeyword.tokenKind == .keyword(.var)
+            if isVarDecl, let initializer = property.bindings.first?.initializer {
+                guard let defaultArg = parameter.defaultArgument else { return false }
+                guard initializer.value.description == defaultArg.value.description else { return false }
+            } else if parameter.defaultArgument != nil {
+                return false
+            }
```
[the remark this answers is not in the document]
mildm8nnered: So I didn't manage to get it down to one statement, but I have tried to simplify this as much as possible

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
```diff
@@ -0,0 +1,286 @@
...
+
+            if
+                propertyId.identifier.text != parameter.firstName.text
+                    || (propertyType != nil && propertyType?.description.trimmingCharacters(in: .whitespaces) !=
+                    parameter.type.description.trimmingCharacters(in: .whitespacesAndNewlines))
+            {
+                return false
+            }
```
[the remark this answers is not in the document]
mildm8nnered: I've rolled this up into the preceding statements now

mildm8nnered:
I think that should be all CR comments addressed now, but let me know if there are any further changes required @SimplyDanny (and thanks for all the detailed feedback).

SimplyDanny (approved):

SimplyDanny:
> I think that should be all CR comments addressed now, but let me know if there are any further changes required @SimplyDanny (and thanks for all the detailed feedback).

I've just pushed two more commits. They are rather unimportant changes regarding formatting, inlining and reuse. [b3d3b7e](https://github.com/realm/SwiftLint/pull/4812/commits/b3d3b7e00dabbd8e8c06f3c460b071711237571d) could make the rule a bit faster, though.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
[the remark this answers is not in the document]
SimplyDanny: The second (and implemented) one looks clearer to me. Thanks!

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededSynthesizedInitializerRule.swift
jpsim: Yes, I think this is fine. We've done this before with operator folding which we had taken from swift-syntax at some point, but now just use it directly.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnneededSynthesizedInitializerRule.swift
```diff
@@ -0,0 +1,308 @@
...
+        override var skippableDeclarations: [DeclSyntaxProtocol.Type] { .allExcept(StructDeclSyntax.self) }
+
+        override func visit(_ node: StructDeclSyntax) -> SyntaxVisitorContinueKind {
+            violations += node.unneededInitializers.map {
+                let initializerType = $0.parameterList.isEmpty ? "default" : "memberwise"
+                let reason = "This \(initializerType) initializer would be synthesized automatically - " +
+                "you do not need to define it"
+                return ReasonedRuleViolation(position: $0.positionAfterSkippingLeadingTrivia, reason: reason)
+            }
+            return .skipChildren
```
jpsim, line 53: This means that nested structs won't trigger a violation, right?

```swift
struct Outer {
  struct Inner {
    let prop: Int
    init(prop: Int) {
      self.prop = prop
    }
  }
}
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnneededSynthesizedInitializerRule.swift
```diff
@@ -0,0 +1,308 @@
...
+
+private extension UnneededSynthesizedInitializerRule {
+    final class Visitor: ViolationsSyntaxVisitor {
+        override var skippableDeclarations: [DeclSyntaxProtocol.Type] { .allExcept(StructDeclSyntax.self) }
+
+        override func visit(_ node: StructDeclSyntax) -> SyntaxVisitorContinueKind {
+            violations += node.unneededInitializers.map {
+                let initializerType = $0.parameterList.isEmpty ? "default" : "memberwise"
+                let reason = "This \(initializerType) initializer would be synthesized automatically - " +
+                "you do not need to define it"
```
jpsim, line 50: Nit, indentation:

```suggestion
                    "you do not need to define it"
```

jpsim (approved):
I'm ok with doing this, it's a very useful rule to have, and I think the attribution is sufficient.

SimplyDanny:
Are you fine with squashing all the commits, @mildm8nnered?

mildm8nnered:
> Are you fine with squashing all the commits, @mildm8nnered?

Absolutely

PR #4818, opened 2023-03-13
Fix for `trailing_semicolon` rule wrongly also removing trailing comments

### LINKED ISSUE #4817: Auto-fixing `trailing_semicolon` rule deletes trailing comments
### New Issue Checklist

### Describe the bug

When SwiftLint auto-fixes a `trailing_semicolon` violation, it also deletes any comment after the semicolon. It shouldn't.

#### Original line:
```
private let pulseVelocity: CGFloat = 878 / 2;  // 878 px / 2 seconds
```

#### Command:
```bash
$ swiftlint lint --fix
```

#### Expected:
```
private let pulseVelocity: CGFloat = 878 / 2  // 878 px / 2 seconds
```

#### Actual:
```
private let pulseVelocity: CGFloat = 878 / 2
```

### Environment

* 0.50.3
* Homebrew

```yml
disabled_rules:
 - identifier_name
 - todo
 - trailing_comma
 - type_body_length
 - function_body_length

included:
 - Encircle

excluded:
 - Pods

# Rule configurations
force_cast: warning

class_delegate_protocol: error

cyclomatic_complexity:
  error: 34

identifier_name:
  min_length:
    error: 0
    warning: 3
  excluded:
    - id

line_length: 120
```

* Which Xcode version are you using (check `xcodebuild -version`)? 14.2

mildm8nnered:
Resolves #4817

When replacing the semi-colon, preserve the trailing trivia.

Ironically, the original comment read 

`// Is there a better way to remove a node? Should we somehow keep trailing trivia?`

### DIFF
--- a/Source/SwiftLintFramework/Rules/Idiomatic/TrailingSemicolonRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/TrailingSemicolonRule.swift
@@ -66,7 +66,6 @@ private extension TrailingSemicolonRule {
             }
 
             correctionPositions.append(node.positionAfterSkippingLeadingTrivia)
-            // Keep the trailing trivia
             return ("" as TokenSyntax).with(\.trailingTrivia, node.trailingTrivia ?? .zero)
         }
     }
--- a/Source/SwiftLintFramework/Rules/Idiomatic/TrailingSemicolonRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/TrailingSemicolonRule.swift
@@ -66,7 +66,7 @@ private extension TrailingSemicolonRule {
             }
 
             correctionPositions.append(node.positionAfterSkippingLeadingTrivia)
-            return ("" as TokenSyntax).with(\.trailingTrivia, node.trailingTrivia ?? .zero)
+            return .unknown("").with(\.trailingTrivia, node.trailingTrivia ?? .zero)
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Rules/Idiomatic/TrailingSemicolonRule.swift
```diff
@@ -65,8 +66,8 @@ private extension TrailingSemicolonRule {
             }
 
             correctionPositions.append(node.positionAfterSkippingLeadingTrivia)
-            // Is there a better way to remove a node? Should we somehow keep trailing trivia?
-            return super.visit(TokenSyntax(.semicolon, presence: .missing))
+            // Keep the trailing trivia
+            return ("" as TokenSyntax).with(\.trailingTrivia, node.trailingTrivia ?? .zero)
```
mildm8nnered, line 70: Not sure If this is the best way to do this

SimplyDanny: Not sure either. This seems to boil down to `.identifier("")` which is a bit strange. Do you like `.unknown("").with(\.trailingTrivia, ...)` more?

SimplyDanny (approved):

Source/SwiftLintFramework/Rules/Idiomatic/TrailingSemicolonRule.swift
mildm8nnered: I think so - it feels less hacky than the cast, so I went with that.

SimplyDanny:
Thank you for the prompt fix!

PR #4828, opened 2023-03-19
Fix crash when parsing apple/swift

### LINKED ISSUE #4827: When scanning apple/swift, linting multiple files fails with an array index out of bounds
### New Issue Checklist

### Describe the bug

```
brew install swiftlint
git clone https://github.com/apple/swift
cd swift
swiftlint ./validation-test/compiler_crashers_fixed/00432-swift-nominaltypedecl-getdeclaredtypeincontext.swift
```

```
Linting Swift files at paths ./validation-test/compiler_crashers_fixed/00432-swift-nominaltypedecl-getdeclaredtypeincontext.swift
Linting '00432-swift-nominaltypedecl-getdeclaredtypeincontext.swift' (1/1)
Swift/StringRangeReplaceableCollection.swift:302: Fatal error: String index is out of bounds
PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace.
zsh: trace trap  swiftlint 
```

The following files are affected:

```
./validation-test/compiler_crashers_fixed/00432-swift-nominaltypedecl-getdeclaredtypeincontext.swift
./validation-test/compiler_crashers_fixed/02004-swift-diagnosticengine-flushactivediagnostic.swift
./validation-test/compiler_crashers_fixed/00789-swift-typebase-isexistentialtype.swift
./validation-test/compiler_crashers_fixed/01275-swift-diagnosticengine-diagnose.swift
./validation-test/compiler_crashers_fixed/01334-swift-constraints-constraintsystem-gettypeofmemberreference.swift
./validation-test/compiler_crashers_fixed/00138-swift-modulefile-maybereadpattern.swift
./validation-test/compiler_crashers_fixed/01747-swift-constraints-constraintsystem-getfixedtyperecursive.swift
./validation-test/compiler_crashers_fixed/01315-swift-parser-parsebraceitems.swift
./validation-test/compiler_crashers_fixed/01512-swift-optional-swift-infixoperatordecl.swift
./validation-test/compiler_crashers_fixed/01246-getselftypeforcontainer.swift
./validation-test/compiler_crashers_fixed/01253-swift-tuplepattern-createsimple.swift
./validation-test/compiler_crashers_fixed/01156-getselftypeforcontainer.swift
./validation-test/compiler_crashers_fixed/01159-swift-declcontext-lookupqualified.swift
./validation-test/compiler_crashers_fixed/01453-swift-constraints-constraintgraph-change-undo.swift
./validation-test/compiler_crashers_fixed/01843-llvm-bitstreamcursor-readrecord.swift
./validation-test/compiler_crashers_fixed/00360-swift-parser-parseexprlist.swift
./validation-test/compiler_crashers_fixed/00362-clang-stmt-statisticsenabled.swift
./validation-test/compiler_crashers_fixed/01020-swift-lexer-lexoperatoridentifier.swift
./validation-test/compiler_crashers_fixed/01185-swift-parser-parsetypeidentifier.swift
./validation-test/compiler_crashers_fixed/01361-swift-clangmoduleunit-getadaptermodule.swift
./validation-test/compiler_crashers_fixed/28365-swift-constraints-constraintgraphnode-getmembertype.swift
./validation-test/compiler_crashers_fixed/01278-swift-modulefile-maybereadconformance.swift
./validation-test/compiler_crashers_fixed/00184-swift-modulefile-lookupvalue.swift
./validation-test/compiler_crashers_fixed/26761-swift-constraints-constraintsystem-solvesimplified.swift
./validation-test/compiler_crashers_fixed/01981-swift-modulefile-maybereadgenericparams.swift
./validation-test/compiler_crashers_fixed/02127-swift-modulefile-maybereadpattern.swift
./validation-test/compiler_crashers_fixed/01157-swift-typebase-getcanonicaltype.swift
./validation-test/compiler_crashers_fixed/00095-swift-clangimporter-implementation-mergepropinfointoaccessor.swift
./validation-test/compiler_crashers_fixed/01588-swift-constraints-constraintsystem-gettypeofmemberreference.swift
./validation-test/compiler_crashers_fixed/01815-getselftypeforcontainer.swift
./validation-test/compiler_crashers_fixed/01284-swift-parser-parseidentifier.swift
./validation-test/compiler_crashers_fixed/00934-swift-type-walk.swift
./validation-test/compiler_crashers_fixed/01873-swift-pattern-foreachvariable.swift
./validation-test/compiler_crashers_fixed/01263-no-stacktrace.swift
./validation-test/compiler_crashers_fixed/01151-getselftypeforcontainer.swift
./validation-test/compiler_crashers_fixed/01262-swift-modulefile-maybereadconformance.swift
./validation-test/compiler_crashers_fixed/00983-a.swift
./validation-test/compiler_crashers_fixed/01155-swift-nominaltypedecl-getdeclaredtypeincontext.swift
./validation-test/compiler_crashers_fixed/02193-swift-typechecker-resolvetypeincontext.swift
./validation-test/compiler_crashers_fixed/01280-swift-type-walk.swift
./validation-test/compiler_crashers_fixed/01254-swift-constraints-constraintsystem-diagnosefailurefromconstraints.swift
./test/SourceKit/CodeFormat/indent-trailing/trailing-generic-param-comma.swift
```

See above

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? `0.50.3`

none

```
Xcode 14.2
Build version 14C18
```

See above for sample files that cause the issue

mildm8nnered:
Fixes #4827

With especially funky files (e.g. Apple's test cases for issues that formerly crashed the compiler), `name` could end up being empty, which means that `name[name.startIndex]` will instantly crash.

### DIFF
--- a/Source/SwiftLintFramework/Rules/Idiomatic/GenericTypeNameRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/GenericTypeNameRule.swift
@@ -66,7 +66,7 @@ private extension GenericTypeNameRule {
 
         override func visitPost(_ node: GenericParameterSyntax) {
             let name = node.name.text
-            guard !configuration.shouldExclude(name: name) else { return }
+            guard !name.isEmpty, !configuration.shouldExclude(name: name) else { return }
 
             let allowedSymbols = configuration.allowedSymbols.union(.alphanumerics)
             if !allowedSymbols.isSuperset(of: CharacterSet(charactersIn: name)) {


Source/SwiftLintFramework/Rules/Idiomatic/GenericTypeNameRule.swift
```diff
@@ -66,7 +66,7 @@ private extension GenericTypeNameRule {
 
         override func visitPost(_ node: GenericParameterSyntax) {
             let name = node.name.text
-            guard !configuration.shouldExclude(name: name) else { return }
+            guard !name.isEmpty, !configuration.shouldExclude(name: name) else { return }
```
SimplyDanny, line 69: An empty generic parameter name isn't valid Swift. This is what I was referring to in my other comments regarding this topic. Supporting invalid Swift code would lead to a lot of strange checks in SwiftLint's rules. This either clutters the code significantly or will just be forgotten all over the place causing "bugs". Compilable Swift input code is really not a too hard requirement we should rely upon.

mildm8nnered: So I totally get that we want to avoid invalid code, and that the bus errors and segfault might not be worth the effort to chase down, but I would have thought this case, and https://github.com/jpsim/SourceKitten/issues/765 would be quite uncontroversial.

In the latter case, if File (in SourceKitten) is initialised immediately, we `try` to read the file as UTF-8, and fail. In the deferred reading case, we `try!`, and therefore crash.

If a user accidentally introduces a non-UTF-8 character into a file, or somehow end up accidentally removing read permissions from the file, and SwiftLint just crashes, that is not a great experience for the user, when we could pretty trivially guard against that and tell them what's wrong.

Similarly here, granted the Swift code will be invalid, but clearly it is possible for us to arrive at `GenericTypeNameRule`'s `visitPost` method, with `node.name.text` being empty, which will definitely crash. Now the user has to try and work out why it's crashing, whether they should open a defect for this behaviour, etc.

A single ` !name.isEmpty` check added to the `guard` statement doesn't seem like a lot of "strange code" or "clutter", but just standard defensive programming, and a much better user experience for anyone who happens to hit these unusual cases.

SimplyDanny: > In the latter case, if File (in SourceKitten) is initialised immediately, we `try` to read the file as UTF-8, and fail. In the deferred reading case, we `try!`, and therefore crash.
>
> If a user accidentally introduces a non-UTF-8 character into a file, or somehow end up accidentally removing read permissions from the file, and SwiftLint just crashes, that is not a great experience for the user, when we could pretty trivially guard against that and tell them what's wrong.

No objections. The tool can definitely report something useful.

> A single ` !name.isEmpty` check added to the `guard` statement doesn't seem like a lot of "strange code" or "clutter", but just standard defensive programming, and a much better user experience for anyone who happens to hit these unusual cases.

I'm still struggling with this one, though. It's a single check here, yes. Then we could go on and add similar checks for all kinds of identifiers like class/struct/enum/.../variable/type names all over the place. I don't like "dead code" given that it is only there to protect against anyway invalid input.

I'm planning to let SwiftSyntax do this job before SwiftLint jumps in. Most of the rules already build on it. If there are any parsing errors, we can ask SwiftSyntax to report them. The output is going to be very helpful and the rules can trust on valid input assured by the parser.

mildm8nnered: I totally get the thin end of the wedge argument.

I think if you consider the **entire universe** of possible inputs, then, yep, maybe we would end up putting guards in a lot more places.

I think there are two ways to look at it:

Firstly, if you're doing an array access, and you got the array or the index from somewhere else, you probably shouldn't trust those, simply as a matter of form - i.e. defensive programming.

Alternatively, if you consider the array of **actual inputs** (to SwiftLint), it turns out that so far, the only place that we hit an array index out of bounds issue (that we know about it) is this place. The cost of patching this is pretty low, and it fixes a whole bunch of cases.

If a load more suddenly turned up, then maybe we would want to consider whether it's worth the cost, or whether we should be taking a different approach to this.

re `SwiftSyntax`, should we be reporting this issue upstream to them?

Largely orthogonal to this conversation, but I did see an interesting mention of SwiftSyntax in a thread in the iOS Developers slack channel the other day, which caught my eye. I have no experience with SwiftSyntax outside of within SwiftLint a tiny bit, and no view on its quality, but I'll post the thread below:

Thread is about some swift-evolution Macro proposal

Poster A:

"As someone who worked quite a bit with SwiftSyntax: this [Macros] is going to be used by very few people. Perhaps the pre-made macros, but once you get down to SwiftSyntax, the library itself is still young, buggy and undocumented (I did most of my work following Apple's repos, their tests etc). Def not ready to be part of the language via macros imho."

Poster B:

"This mirrors our experience with using SwiftSyntax. Apple seems to have put some effort recently in gradually improving the documentation side of things but the state of the library itself is exactly how you describe it. Hopefully we’ll be able to pin down the exact version of the library when declaring macros (although some recent pitches from Swift Evolution seem to suggest otherwise) rather than this library being coupled to the toolchain."

Which make It sound like we might now want to rely on SwiftSyntax taking care of this themselves, in the near-term.

SimplyDanny: > re `SwiftSyntax`, should we be reporting this issue upstream to them?

Nothing to report to them. The parser already collects diagnostics. SwiftLint just has to use/show them.

> Which make It sound like we might now want to rely on SwiftSyntax taking care of this themselves, in the near-term.

I don't really get what you mean. Could you elaborate?

SimplyDanny (approved):

Source/SwiftLintFramework/Rules/Idiomatic/GenericTypeNameRule.swift
mildm8nnered: So I took your previous comment to mean "we should never actually get called in this scenario [with an empty name], and it's SwiftSyntax's fault that we do - they should be taking care of this", if that makes sense.

SimplyDanny: > Nothing to report to them. The parser already collects diagnostics. SwiftLint just has to use/show them.

I was a bit too fast posing this statement. Actually, SwiftLint does not handle empty parameter clauses so well. I've reported an issue.

That said, SwiftSyntax is going to become better over time. Whenever it complains about invalid input, SwiftLint should stop continuing its analysis. If we find a case that is not reported by SwiftSyntax that's a defect on their side which we should report. That way, the models the rules work on can be assumed to be valid.

PR #4845, opened 2023-03-28
Add sorted enum cases rule

kimdv:
Added a new rule that will show an error when enum cases are not sorted

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -42,6 +42,7 @@ disabled_rules:
   - prefixed_toplevel_constant
   - required_deinit
   - self_binding
+  - sorted_enum_cases
   - strict_fileprivate
   - switch_case_on_newline
   - trailing_closure
--- a/Source/SwiftLintFramework/Rules/Style/SortedEnumCasesRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/SortedEnumCasesRule.swift
@@ -1,0 +1,120 @@
+import SwiftSyntax
+
+struct SortedEnumCasesRule: ConfigurationProviderRule, SwiftSyntaxRule, OptInRule {
+    var configuration = SeverityConfiguration(.warning)
+
+    init() {}
+
+    static let description = RuleDescription(
+        identifier: "sorted_enum_cases",
+        name: "Sorted Enum Cases",
+        description: "Enum cases should be sorted",
+        kind: .style,
+        nonTriggeringExamples: [
+            Example("""
+            enum foo {
+                case a
+                case b
+                case c
+            }
+            """),
+            Example("""
+            enum foo {
+                case a, b, c
+            }
+            """),
+            Example("""
+            enum foo {
+                case a
+                case b, c
+            }
+            """),
+            Example("""
+            enum foo {
+                case a(foo: Foo)
+                case b(String), c
+            }
+            """),
+            Example("""
+            @frozen
+            enum foo {
+                case b
+                case a
+                case c, f, d
+            }
+            """)
+        ],
+        triggeringExamples: [
+            Example("""
+            enum foo {
+                ↓case b
+                ↓case a
+                case c
+            }
+            """),
+            Example("""
+            enum foo {
+                case ↓b, ↓a, c
+            }
+            """),
+            Example("""
+            enum foo {
+                ↓case b, c
+                ↓case a
+            }
+            """),
+            Example("""
+            enum foo {
+                case a
+                case b, ↓d, ↓c
+            }
+            """),
+            Example("""
+            enum foo {
+                case a(foo: Foo)
+                case ↓c, ↓b(String)
+            }
+            """)
+        ]
+    )
+
+    func makeVisitor(file: SwiftLintFile) -> ViolationsSyntaxVisitor {
+        Visitor(viewMode: .sourceAccurate)
+    }
+}
+
+private extension SortedEnumCasesRule {
+    final class Visitor: ViolationsSyntaxVisitor {
+        override var skippableDeclarations: [DeclSyntaxProtocol.Type] {
+            return .allExcept(EnumDeclSyntax.self)
+        }
+
+        override func visit(_ node: EnumDeclSyntax) -> SyntaxVisitorContinueKind {
+            guard !node.attributes.contains(attributeNamed: "frozen") else {
+                return .skipChildren
+            }
+
+            let cases = node.members.members.compactMap { $0.decl.as(EnumCaseDeclSyntax.self) }
+            let sortedCases = cases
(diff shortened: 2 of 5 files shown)


khawajafarooqahmed (approved):

Source/SwiftLintFramework/Rules/Style/SortedEnumCasesRule.swift
```diff
@@ -0,0 +1,95 @@
...
+    func makeVisitor(file: SwiftLintFile) -> ViolationsSyntaxVisitor {
+        Visitor(viewMode: .sourceAccurate)
+    }
+}
+
+private extension SortedEnumCasesRule {
+    final class Visitor: ViolationsSyntaxVisitor {
+        override var skippableDeclarations: [DeclSyntaxProtocol.Type] {
+            return [VariableDeclSyntax.self]
+        }
```
SimplyDanny, line 71: You probably rather want:
```suggestion
        override var skippableDeclarations: [DeclSyntaxProtocol.Type] { .allExcept(EnumDeclSyntax.self) }
```
That way, it skips protocols, classes, structs, ... but not enums.

SimplyDanny (changes requested):
Nice little rule! And good idea to split the sorting of cases and case elements. 👍 

I think this rule should ignore `@frozen` enums.

And please also add a test case where some enum cases have arguments.

SimplyDanny:
To get the tests running, please ignore the new rule in the `.swiftlint.yml` configuration. We may enable it in a separate PR afterwards.

.swiftlint.yml
```diff
@@ -48,6 +48,7 @@ disabled_rules:
   - type_contents_order
   - unused_capture_list
   - vertical_whitespace_between_cases
+  - sorted_enum_cases
```
SimplyDanny, line 51: Please sort alphabetically (like your enum cases 😉).

Source/SwiftLintFramework/Rules/Style/SortedEnumCasesRule.swift
```diff
@@ -0,0 +1,119 @@
+import Foundation
```
SimplyDanny, line 1: Foundation is not needed.

Source/SwiftLintFramework/Rules/Style/SortedEnumCasesRule.swift
```diff
@@ -0,0 +1,119 @@
...
+                case a(foo: Foo)
+                case b(String), c
+            }
+            """),
+            Example("""
+            @frozen
+            enum foo {
+                case b
+                case a
+                case c
```
SimplyDanny, line 44: `case c, f, d` will still trigger the rule.

Source/SwiftLintFramework/Rules/Style/SortedEnumCasesRule.swift
```diff
@@ -0,0 +1,119 @@
...
+            Example("""
+            enum foo {
+                case a
+                case b, ↓d, ↓c
+            }
+            """),
+            Example("""
+            enum foo {
+                case a(foo: Foo)
+                case c, b(String)
```
SimplyDanny, line 76: Missing violation markers `↓`.

SimplyDanny (changes requested):

Source/SwiftLintFramework/Rules/Style/SortedEnumCasesRule.swift
```diff
@@ -0,0 +1,123 @@
...
+                case b(String), c
+            }
+            """),
+            Example("""
+            @frozen
+            enum foo {
+                case b
+                case a
+                case c
+                case c, f, d
```
SimplyDanny, line 44: We like to have valid Swift in the examples. Now `c` is redeclared.

Source/SwiftLintFramework/Rules/Style/SortedEnumCasesRule.swift
```diff
@@ -0,0 +1,123 @@
...
+            Example("""
+            enum foo {
+                case a
+                case b, ↓d, ↓c
+            }
+            """),
+            Example("""
+            enum foo {
+                case a(foo: Foo)
+                case ↓c, b(String)
```
SimplyDanny, line 76: Both violate:
```suggestion
                case ↓c, ↓b(String)
```

Source/SwiftLintFramework/Rules/Style/SortedEnumCasesRule.swift
```diff
@@ -0,0 +1,123 @@
...
+            let cases = node.members.members.compactMap { $0.decl.as(EnumCaseDeclSyntax.self) }
+            let sortedCases = cases
+                .sorted(by: { $0.elements.first!.identifier.text < $1.elements.first!.identifier.text })
+
+            zip(sortedCases, cases).forEach { sortedCase, currentCase in
+                if sortedCase.elements.first?.identifier.text != currentCase.elements.first?.identifier.text {
+                    violations.append(currentCase.positionAfterSkippingLeadingTrivia)
+                }
+            }
+        }
```
SimplyDanny, line 107: This can be solved more elegantly in my humble opinion:
```suggestion
        override func visit(_ node: EnumDeclSyntax) -> SyntaxVisitorContinueKind {
            guard !node.attributes.contains(attributeNamed: "frozen") else {
                return .skipChildren
            }
            
            // ...
            
            return .visitChildren
        }
```

Source/SwiftLintFramework/Rules/Style/SortedEnumCasesRule.swift
```diff
@@ -0,0 +1,123 @@
...
+
+        override func visitPost(_ node: EnumCaseDeclSyntax) {
+            guard node.parent?.parent?.parent?.parent?.as(EnumDeclSyntax.self)?.attributes.contains(attributeNamed: "frozen") == false else {
```
SimplyDanny, line 110: Can be removed with my previous suggestion.

Source/SwiftLintFramework/Rules/Style/SortedEnumCasesRule.swift
```diff
@@ -0,0 +1,123 @@
...
+            """),
+            Example("""
+            @frozen
+            enum foo {
+                case b
+                case a
+                case c
+                case c, f, d
+            }
+            """).focused()
```
SimplyDanny, line 46:
```suggestion
            """)
```
This fails the builds.

SimplyDanny (changes requested):

SimplyDanny (approved):
Thank you, @kimdv!

PR #4846, opened 2023-03-29
Fix README.md example for custom_rules#{in,ex}cluded

sashaweiss:
Firstly, thanks for maintaining this tool!

There's an error in `README.md` in depicting how to use `{in,ex}cluded` with `custom_rules`. Specifically, it suggests that `{in,ex}cluded` should key to a list of values, but the implementation expects a string type.

`README_KR.md`'s example is correct.

`README_CN.md` is missing this example entirely, and while I'd like to add it now I wouldn't be able to add an appropriate comment and so I did not do so here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -568,10 +568,8 @@ following syntax:
 ```yaml
 custom_rules:
   pirates_beat_ninjas: # rule identifier
-    included: 
-      - ".*\\.swift" # regex that defines paths to include during linting. optional.
-    excluded: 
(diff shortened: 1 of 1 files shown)


SimplyDanny:
They support both a single pattern or a list of patterns. In that case, I like the list syntax more even though the example only assumes a single pattern. It hints that multiple patterns are possible.

SimplyDanny:
Well, there might still be a bug in the implementation. The documentation is correct, however.

If you have a simple reproducing example at hand, please do not hesitate to open an issue.

Issue #4823, opened 2023-03-16
Processing of command line paths is unintuitive and confusing


mildm8nnered:
### New Issue Checklist

- [X] Updated SwiftLint to the latest version
- [X] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

Consider the case where I have a `.swiftlint.yml` in my project, which specifies some `included` paths (like SwiftLint's own configuration), and I'm running `swiftlint` on the command line from the top level directory.

```
% head .swiftlint.yml
included:
  - Plugins
  - Source
  - Tests
excluded:
  - Tests/SwiftLintFrameworkTests/Resources
analyzer_rules:
  - unused_declaration
  - unused_import
opt_in_rules:
```

(`excluded:` can be ignored for the moment).

Some of the invocations of `swiftlint` do not work as expected for me. All examples using SwiftLint itself as the project.

1) `swiftlint`

Expected: Should run swiftlint using the configuration file.

Works as expected (`Linting Swift files in current working directory` could perhaps be clearer, because it's actually linting the files specified in the config).

```
% swiftlint                      
Linting Swift files in current working directory
Linting 'SwiftLintPlugin.swift' (1/544)
Linting 'Path+Helpers.swift' (2/544)
Linting 'BodyLengthRuleVisitor.swift' (3/544)
Linting 'CommandVisitor.swift' (4/544)
```

2) `swiftlint Source/SwiftLintFramework/Reporters/CSVReporter.swift` (single file(s))

Expected: Should run swiftlint using the configuration file, but only on the specified file(s). 

Works as expected:

```
 % swiftlint Source/SwiftLintFramework/Reporters/CSVReporter.swift 
Linting Swift files at paths Source/SwiftLintFramework/Reporters/CSVReporter.swift
Linting 'CSVReporter.swift' (1/1)
Done linting! Found 0 violations, 0 serious in 1 file.
```

3) `swiftlint Source/SwiftLintFramework/Reporters/CSVReporter.swift a.swift` (single files, where some do not exist)

Expected: Should run `swiftlint` using the configuration file, but only on the specified file(s), and should report any files that do not exist.

Observed: SwiftLint lints `Source/SwiftLintFramework/Reporters/CSVReporter.swift`, and then lints all 544 files defined in the configuration (so presumably CSVReporter gets linted twice).

[...]

PR #4847, opened 2023-03-29
fix configuration unit tests

mildm8nnered:
I was looking into making some changes to how we parse files on the command line (see #4823), when I came across some issues with the unit tests. This PR does not change behaviour at all - just addresses these unit test issues.

The initial problem test is `testForceExcludesFile` in `ConfigurationTests`, which looks like this:

```
func testForceExcludesFile() {
    let configuration = Configuration(excludedPaths: ["directory/ExcludedFile.swift"])
    let paths = configuration.lintablePaths(inPath: "directory/ExcludedFile.swift", forceExclude: true,
                                            fileManager: TestFileManager())
    XCTAssertEqual([], paths)
}
```

`TestFileManager` looks like this:

```
private class TestFileManager: LintableFileManager {
    func filesToLint(inPath path: String, rootDirectory: String? = nil) -> [String] {
        switch path {
        case "directory": return ["directory/File1.swift", "directory/File2.swift",
                                  "directory/excluded/Excluded.swift",
                                  "directory/ExcludedFile.swift"]
        case "directory/excluded": return ["directory/excluded/Excluded.swift"]
        case "directory/ExcludedFile.swift": return ["directory/ExcludedFile.swift"]
        default: break
        }
        XCTFail("Should not be called with path \(path)")
        return []
    }

    func modificationDate(forFileAtPath path: String) -> Date? {
        return nil
    }
}
```

When this test is run, in `Configuration+LintableFiles.swift`, `lintablePaths` will get called with arguments `path = directory/ExcludedFile.swift`, `forceExclude = true`, and `excludeByPrefix = false` (and the `TestFileManager`).

```
internal func lintablePaths(
    inPath path: String,
    forceExclude: Bool,
    excludeByPrefix: Bool = false,
    fileManager: LintableFileManager = FileManager.default
) -> [String] {
    if path.isFile {
        if forceExclude {
            return excludeByPrefix
                ? filterExcludedPathsByPrefix(in: [path.absolutePathStandardized()])
                : filterExcludedPaths(fileManager: fileManager, in: [path.absolutePathStandardized()])
        }
        // If path is a file and we're not forcing excludes, skip filtering with excluded/included paths
        return [path]
    }

    let pathsForPath = includedPaths.isEmpty ? fileManager.filesToLint(inPath: path, rootDirectory: nil) : []
    let includedPaths = self.includedPaths
        .flatMap(Glob.resolveGlob)

    return excludeByPrefix
        ? filterExcludedPathsByPrefix(in: pathsForPath, includedPaths)
        : filterExcludedPaths(fileManager: fileManager, in: pathsForPath, includedPaths)
}
```

`path.isFile` would be `true` on a real filesystem, but will (almost certainly) be `false` in the unit test case, so the test will fall through to the remaining code.

That code will end up returning `[]` anyway, but the test is not exercising the code that it's supposed to be.

To work around this, I added a `isFile(atPath:)` method to `LintableFileManager`, which the mock can implement to fall down the correct path.

This caused `testForceExcludesFile` to start failing.

The problem here is that in `filterExcludedPaths(fileManager: fileManager, in: [path.absolutePathStandardized()])`, for example, we'll actually pass in `/path/to/current_directory/directory/ExcludedFile.swift`, but our Mock TestFileManager will return relative paths when checking the exclusions.

In real world execution, I think all the path comparisons for exclusion are done with absolute paths.

I changed TestFileManager to return absolute paths, but this then caused other unit tests to fail until they were similarly adjusted to expect absolute paths.

### DIFF
--- a/Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
+++ b/Source/SwiftLintFramework/Extensions/Configuration+LintableFiles.swift
@@ -33,7 +33,7 @@ extension Configuration {
         excludeByPrefix: Bool = false,
         fileManager: LintableFileManager = FileManager.default
     ) -> [String] {
-        if path.isFile {
+        if fileManager.isFile(atPath: path) {
             if forceExclude {
                 return excludeByPrefix
                     ? filterExcludedPathsByPrefix(in: [path.absolutePathStandardized()])
--- a/Source/SwiftLintFramework/Extensions/FileManager+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/FileManager+SwiftLint.swift
@@ -51,17 +51,6 @@ extension FileManager: LintableFileManager {
     }
 
     public func isFile(atPath path: String) -> Bool {
-        exists(atPath: path, isDirectory: false)
-    }
-
-    private func exists(atPath path: String, isDirectory: Bool) -> Bool {
-        guard !path.isEmpty else {
-            return false
-        }
-        var isDirectoryObjC: ObjCBool = false
-        if fileExists(atPath: path, isDirectory: &isDirectoryObjC) {
-            return isDirectoryObjC.boolValue == isDirectory
-        }
-        return false
+        path.isFile
     }
 }
--- a/Source/SwiftLintFramework/Extensions/FileManager+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/FileManager+SwiftLint.swift
@@ -19,6 +19,13 @@ public protocol LintableFileManager {
     ///
     /// - returns: A date, if one was determined.
     func modificationDate(forFileAtPath path: String) -> Date?
+
+    /// Returns true if a file (but not a directory) exists at the specified path.
+    ///
+    /// - parameter path: The path that should be checked to see if it is a file.
+    ///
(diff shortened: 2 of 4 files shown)


SimplyDanny (approved):
Alternatively, we could skip the mock entirely and instead create the directory structure in the `Resources` folder next to the test classes. It avoids any issues caused by incorrect mocking.

What do you think about that?

mildm8nnered:
> Alternatively, we could skip the mock entirely and instead create the directory structure in the `Resources` folder next to the test classes. It avoids any issues caused by incorrect mocking.
> 
> What do you think about that?

So I think extending the `LintableFilesManager` to make the DI more "complete" is the right thing to do regardless.

I'm a massive unfan of mocking the filesystem (as opposed to say a network service) as there's a perfectly good filesystem right there, and because of bugs like this, but I can see that there are some advantages in making the exact dependencies explicit, and having the mock behaviour as an additional constraint on the code.

Given that the code is already there and now apparently works correctly, with these relatively small changes, I think I'm happy to go with this, and maybe consider abandoning the mock in a later pass.

I want to propose some later changes to how we expand lintable files later anyway (see #4823), so will definitely be revisiting this area soon.

PR #4849, opened 2023-03-30
Breaking change: paths specified on the command line should take precedence over configurations

### LINKED ISSUE #4823: Processing of command line paths is unintuitive and confusing

mildm8nnered:
This PR changes how SwiftLint treats files or directories passed as command line arguments, and resolves  #4823

Typically, one would expect that paths explicitly supplied on the command line would take precedence over paths specified in a configuration file.

SwiftLint behaves like this in most cases, but there are some scenarios where the existing behavior is not intuitive or expected.

I'll enumerate some scenarios below (some of which have not changed at all).

1. `swiftlint` - if there is a configuration file present, this will be attended to. Otherwise, lint all `.swift` files under the current directory - unchanged

2. `swiftlint /relative/or/absolute/path/to/some/file` - the specified file(s) will be linted - unchanged

3. `swiftlint /relative/or/absolute/path/to/some/file /relative/or/absolute/path/to/some/file_that_does_not_actually exist`

     Old behaviour: SwiftLint would lint the file that exists, and then scan all files and directories specified in the configuration file if one was present. If no configuration file was present, SwiftLint would just lint the existing file.

    New behaviour: SwiftLint will ignore the configuration's specified files and directories, and lint any specified files that actually do exist. A warning will be printed for any path supplied on the command line that does not exist:

    `warning: No file or directory found at '/tmp/Z.swift'.`

4. `swiftlint /relative/or/absolute/path/to/some/directory`

    Old behaviour - if a configuration file was detected, and specified any directories to scan, the configuration file's specification of which files/directories to scan would be used, and the command line directory (or directories) would be ignored. If no configuration file was present, or if it specified no paths to scan, then the directory specified on the command line would be scanned.

    New behaviour - the directory or directories specified on the command line will be scanned. Any directories specified in a configuration file will be ignored. The remainder of the configuration file, if any, for example, exclusions, rule enablement or disablement, will be respected.

#### Messaging

If there is no configuration file, or no paths are specified in the configuration file, then SwiftLint will print

`Linting Swift files in current working directory`

as before. However if a configuration file is specified, or the default is present, and any paths are specified in the configuration, then SwiftLint will now print

`Linting Swift files specified in the configuration`

instead.

### DIFF
--- a/Source/SwiftLintCore/Extensions/Configuration+LintableFiles.swift
+++ b/Source/SwiftLintCore/Extensions/Configuration+LintableFiles.swift
@@ -48,11 +48,10 @@ extension Configuration {
         }
 
         if path.isEmpty {
-            var pathsForPath = includedPaths
-            if pathsForPath.isEmpty {
-                pathsForPath = fileManager.filesToLint(inPath: path, rootDirectory: nil)
+            if includedPaths.isEmpty {
+                return excludePaths(from: fileManager.filesToLint(inPath: "", rootDirectory: rootDirectory))
             }
-            let includedPaths = pathsForPath
+            let includedPaths = includedPaths
                 .flatMap(Glob.resolveGlob)
                 .parallelFlatMap { fileManager.filesToLint(inPath: $0, rootDirectory: rootDirectory) }
             return excludePaths(from: includedPaths)
--- a/Source/SwiftLintCore/Extensions/Configuration+LintableFiles.swift
+++ b/Source/SwiftLintCore/Extensions/Configuration+LintableFiles.swift
@@ -1,6 +1,11 @@
 import Foundation
 
 extension Configuration {
+    public enum ExcludeBy {
+        case prefix
+        case paths(excludedPaths: [String])
+    }
+
     // MARK: Lintable Paths
     /// Returns the files that can be linted by SwiftLint in the specified parent path.
     ///
@@ -12,8 +17,8 @@ extension Configuration {
     ///
     /// - returns: Files to lint.
     public func lintableFiles(inPath path: String, forceExclude: Bool,
-                              excludeByPrefix: Bool = false) -> [SwiftLintFile] {
-        return lintablePaths(inPath: path, forceExclude: forceExclude, excludeByPrefix: excludeByPrefix)
+                              excludeBy: ExcludeBy) -> [SwiftLintFile] {
+        return lintablePaths(inPath: path, forceExclude: forceExclude, excludeBy: excludeBy)
             .compactMap(SwiftLintFile.init(pathDeferringReading:))
     }
 
@@ -33,21 +38,29 @@ extension Configuration {
         excludeBy: ExcludeBy,
         fileManager: some LintableFileManager = FileManager.default
     ) -> [String] {
+        func excludePaths(from includedPaths: [String]) -> [String] {
+            switch excludeBy {
+            case .prefix:
+                return filterExcludedPathsByPrefix(in: includedPaths)
+            case .paths(let excludedPaths):
+                return filterExcludedPaths(excludedPaths, in: includedPaths)
+            }
+        }
+
         if path.isEmpty {
-            let includedPaths = self.includedPaths
+            var pathsForPath = includedPaths
+            if pathsForPath.isEmpty {
+                pathsForPath = fileManager.filesToLint(inPath: path, rootDirectory: nil)
+            }
+            let includedPaths = pathsForPath
                 .flatMap(Glob.resolveGlob)
                 .parallelFlatMap { fileManager.filesToLint(inPath: $0, rootDirectory: rootDirectory) }
-
-            return excludeByPrefix
-                ? filterExcludedPathsByPrefix(in: includedPaths)
-                : filterExcludedPaths(fileManager: fileManager, in: includedPaths)
+            return excludePaths(from: includedPaths)
         }
 
         if fileManager.isFile(atPath: path) {
             if forceExclude {
-                return excludeByPrefix
-                    ? filterExcludedPathsByPrefix(in: [path.absolutePathStandardized()])
-                    : filterExcludedPaths(fileManager: fileManager, in: [path.absolutePathStandardized()])
+                return excludePaths(from: [path.absolutePathStandardized()])
             }
             // If path is a file and we're not forcing excludes, skip filtering with excluded/included paths
             return [path]
@@ -55,11 +68,8 @@ extension Configuration {
 
         if fileManager.isDirectory(atPath: path) {
             let pathsForPath = fileManager.filesToLint(inPath: path, rootDirectory: nil)
-            return excludeByPrefix
-                ? filterExcludedPathsByPrefix(in: pathsForPath)
-                : filterExcludedPaths(fileManager: fileManager, in: pathsForPath)
+            return excludePaths(from: pathsForPath)
         }
-
         return []
     }
 
@@ -70,7 +80,7 @@ extension Configuration {
     ///
     /// - returns: The input paths after removing the excluded paths.
     public func filterExcludedPaths(
-        fileManager: LintableFileManager = FileManager.default,
+        _ excludedPaths: [String],
         in paths: [String]...
     ) -> [String] {
         let allPaths = paths.flatMap { $0 }
@@ -81,7 +91,6 @@ extension Configuration {
         let result = NSMutableOrderedSet(array: allPaths)
         #endif
 
-        let excludedPaths = self.excludedPaths(fileManager: fileManager)
         result.minusSet(Set(excludedPaths))
         // swiftlint:disable:next force_cast
         return result.map { $0 as! String }
--- a/Source/SwiftLintCore/Extensions/Configuration+LintableFiles.swift
+++ b/Source/SwiftLintCore/Extensions/Configuration+LintableFiles.swift
@@ -33,6 +33,11 @@ extension Configuration {
         excludeBy: ExcludeBy,
         fileManager: some LintableFileManager = FileManager.default
     ) -> [String] {
+        func excludePaths(from includedPaths: [String]) -> [String] {
+            excludeByPrefix
+            ? filterExcludedPathsByPrefix(in: includedPaths)
+            : filterExcludedPaths(fileManager: fileManager, in: includedPaths)
+        }
         if path.isEmpty {
             var pathsForPath = includedPaths
             if pathsForPath.isEmpty {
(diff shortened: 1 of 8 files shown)


Source/swiftlint/Extensions/Configuration+CommandLine.swift
```diff
@@ -236,7 +236,11 @@ extension Configuration {
         if !visitor.quiet {
             let filesInfo: String
             if visitor.paths.isEmpty || visitor.paths == [""] {
-                filesInfo = "in current working directory"
+                if self != Self.default, includedPaths.isNotEmpty {
```
SimplyDanny, line 239: Why is the check `self != Self.default` necessary? Isn't `includedPaths.isNotEmpty` enough?

Source/SwiftLintCore/Extensions/Configuration+LintableFiles.swift
```diff
@@ -38,30 +38,41 @@ extension Configuration {
...
+        if path.isEmpty {
+            var pathsForPath = includedPaths
+            if pathsForPath.isEmpty {
+                pathsForPath = fileManager.filesToLint(inPath: path, rootDirectory: nil)
+            }
+            let includedPaths = pathsForPath
+                .flatMap(Glob.resolveGlob)
+                .parallelFlatMap { fileManager.filesToLint(inPath: $0, rootDirectory: rootDirectory) }
+            return excludePaths(from: includedPaths)
+        }
```
SimplyDanny, line 59: This looks suspicious to me. `fileManager.filesToLint` (a very expensive operation) is called twice if there are no `includePaths`. This would be my idea of what this should look like:
```suggestion
        if path.isEmpty {
            if includedPaths.isEmpty {
                return excludePaths(from: fileManager.filesToLint(inPath: "", rootDirectory: rootDirectory))
            }
            let includedPaths = includedPaths
                .flatMap(Glob.resolveGlob)
                .parallelFlatMap { fileManager.filesToLint(inPath: $0, rootDirectory: rootDirectory) }
            return excludePaths(from: includedPaths)
        }
```

Source/SwiftLintCore/Extensions/Configuration+LintableFiles.swift
```diff
@@ -38,30 +38,41 @@ extension Configuration {
...
-        case .prefix:
-            return filterExcludedPathsByPrefix(in: pathsForPath, includedPaths)
-        case .paths(let excludedPaths):
-            return filterExcludedPaths(excludedPaths, in: pathsForPath, includedPaths)
+        if fileManager.isDirectory(atPath: path) {
+            let pathsForPath = fileManager.filesToLint(inPath: path, rootDirectory: nil)
+            return excludePaths(from: pathsForPath)
         }
+
+        Issue.noSuchFileOrDirectory(path: path).print()
```
SimplyDanny, line 74: Like that! 👍

SimplyDanny (commented):
I like the proposed change. However, we have to be very careful with this change. It might break people's workflows.

Source/swiftlint/Extensions/Configuration+CommandLine.swift
mildm8nnered: Good point!

Source/SwiftLintCore/Extensions/Configuration+LintableFiles.swift
mildm8nnered: Ah yes - that definitely does look better

mildm8nnered:
> I like the proposed change. However, we have to be very careful with this change. It might break people's workflows.

Absolutely. If someone is inadvertently passing a non-existent file or a directory on the command line in their CI scripts, they would previously get a scan based on their config file settings, whereas now SwiftLint will potentially scan nothing (in the case of a non-existent path), or just the specified directory.

SimplyDanny:
> > I like the proposed change. However, we have to be very careful with this change. It might break people's workflows.
> 
> Absolutely. If someone is inadvertently passing a non-existent file or a directory on the command line in their CI scripts, they would previously get a scan based on their config file settings, whereas now SwiftLint will potentially scan nothing (in the case of a non-existent path), or just the specified directory.

If the passed file doesn't exist, SwiftLint shows a warning now. That's nice. So it's not just silently doing nothing. What would be even better is when SwiftLint was returning a non-zero result code. With that, CI scripts calling SwiftLint would clearly report back if there was nothing to lint. Do you think that's achievable without changing too much?

mildm8nnered:
> > > I like the proposed change. However, we have to be very careful with this change. It might break people's workflows.
> > 
> > 
> > Absolutely. If someone is inadvertently passing a non-existent file or a directory on the command line in their CI scripts, they would previously get a scan based on their config file settings, whereas now SwiftLint will potentially scan nothing (in the case of a non-existent path), or just the specified directory.
> 
> If the passed file doesn't exist, SwiftLint shows a warning now. That's nice. So it's not just silently doing nothing. What would be even better is when SwiftLint was returning a non-zero result code. With that, CI scripts calling SwiftLint would clearly report back if there was nothing to lint. Do you think that's achievable without changing too much?

I think that should be achievable - I think we will actually return a non zero status right now, if allow zero lintable files is off, purely by chance, but we might want to ignore zero lintable files if paths are supplied and none of them exist/can be read.

SimplyDanny:
> > > > I like the proposed change. However, we have to be very careful with this change. It might break people's workflows.
> > > 
> > > 
> > > Absolutely. If someone is inadvertently passing a non-existent file or a directory on the command line in their CI scripts, they would previously get a scan based on their config file settings, whereas now SwiftLint will potentially scan nothing (in the case of a non-existent path), or just the specified directory.
> > 
> > 
> > If the passed file doesn't exist, SwiftLint shows a warning now. That's nice. So it's not just silently doing nothing. What would be even better is when SwiftLint was returning a non-zero result code. With that, CI scripts calling SwiftLint would clearly report back if there was nothing to lint. Do you think that's achievable without changing too much?
> 
> I think that should be achievable - I think we will actually return a non zero status right now, if allow zero lintable files is off, purely by chance, but we might want to ignore zero lintable files if paths are supplied and none of them exist/can be read.

That means the current behavior is already good. Whenever paths specified on the command line or in the configuration result in an empty set of files to lint, SwiftLint fails with a message. Also, your new warning lets users know that the path given doesn't exist.

The `allow_zero_lintable_files` option can only be specified in a configuration, so it should only apply to sessions initiated by the include/exclude paths defined in the same. The option should not have an influence on anything specified on the command line. This is what you mean, isn't it?

mildm8nnered:
>  The `allow_zero_lintable_files` option can only be specified in a configuration, so it should only apply to sessions initiated by the include/exclude paths defined in the same. The option should not have an influence on anything specified on the command line. This is what you mean, isn't it?

Not quite - the way I see it, specifying paths on the command line means that any `included` paths from the config file, if it's present, should be ignored.

But the other settings in the config file should still be respected (IMO) - e.g. the rules settings, and also, and I think this is right, `allow_zero_lintable_files` should be respected as well - if I have that setting on in my config, and I specify on the command line directories that exist and are readable, but contain no swift files, I would not expect to see an error - I specified in the config that no swift files was ok.

Similarly, currently, we will respect the `excluded` settings from the config file.

So even though this "works" in this branch currently, if `allow_zero_lintable_files` is off, I think we actually want a separate error, which is "paths were supplied on the command line, but none of them were readable (or existed).

I think I have a potential fix for this, which is not too awful - still playing around with it

mildm8nnered:
So this latest commit (https://github.com/realm/SwiftLint/pull/4849/commits/8650b5719c2add62bba8e4f6113190df49d52cfd) will produce a non-zero exit status if none of the supplied paths is exists, or is readable.

(Actually, currently, if a file is not readable, SwiftLint will crash, but there is a fix for that in SourceKitten, that hasn't made it into a release yet).

Not the most graceful - I couldn't see any elegant way to propagate the fact that the path did not exist back up the call tree, so if paths were supplied, but we ended up with no lintable files, we check whether any of the supplied paths exists, or is readable, and if not, throw the error.

jpsim:
Hey Martin, thanks so much for taking this on and for your patience on this.

It feels like ever time I've touched the precedence of included/excluded files in the past, it's angered some user somewhere that was relying on the previous behavior.

One example in which I could foresee this breaking people's SwiftLint workflows is when they have some sort of integration with another tool or are calling SwiftLint from a script. For example, if someone had a shell script that called SwiftLint on all files changed in a pull request, they might still want their configuration to take precedence over file paths that were passed on the CLI.

On the surface, I would tend to agree that a path specified on the command line interface seems like a stronger signal to the tool that the user actually wants to see lint results for that file, regardless of if it's ignored in the configuration, but in practice because of use cases like the one I've mentioned above, this might be counter-intuitive to some folks.

Happy to discuss this further if you like, but for the time being I'm inclined to avoid making this change.

mildm8nnered:
> Hey Martin, thanks so much for taking this on and for your patience on this.
> 
> It feels like ever time I've touched the precedence of included/excluded files in the past, it's angered some user somewhere that was relying on the previous behavior.
> 
> One example in which I could foresee this breaking people's SwiftLint workflows is when they have some sort of integration with another tool or are calling SwiftLint from a script. For example, if someone had a shell script that called SwiftLint on all files changed in a pull request, they might still want their configuration to take precedence over file paths that were passed on the CLI.
> 
> On the surface, I would tend to agree that a path specified on the command line interface seems like a stronger signal to the tool that the user actually wants to see lint results for that file, regardless of if it's ignored in the configuration, but in practice because of use cases like the one I've mentioned above, this might be counter-intuitive to some folks.
> 
> Happy to discuss this further if you like, but for the time being I'm inclined to avoid making this change.

So I can totally understand the sensitivity around changes like this.

Would you consider a change that fixed case 3. (where a non-existent file path on the command line causes us to fall back to the configuration), but left the directory case (4) alone?

PR #4852, opened 2023-03-31
Update SwiftSyntax

SimplyDanny:


Package.swift
```diff
@@ -31,7 +31,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser.git", .upToNextMinor(from: "1.2.1")),
-        .package(url: "https://github.com/apple/swift-syntax.git", exact: "0.50900.0-swift-DEVELOPMENT-SNAPSHOT-2023-02-06-a"),
+        .package(url: "https://github.com/apple/swift-syntax.git", exact: "509.0.0-swift-DEVELOPMENT-SNAPSHOT-2023-03-17-a"),
```
jpsim, line 34: Does SwiftPM consider this tag format to be stable? I can’t make heads or tails of how it chooses what is stable and what isn’t.

SimplyDanny: As I understand it, only `.revision` is a problem. `.exact` is fine as long as it can be parsed according to the [Semantic Versioning 2.0.0](https://semver.org) rules. Which seems to be the case. The package wouldn't even build if it was not.

The version I picked is the most recent one that is accepted by SPM (because it starts with an actual version) and is not for Swift 5.9.

PR #4853, opened 2023-03-31
Update swift-argument-parser

SimplyDanny:


jpsim:
Ha, I ran into this recently too. Someone (you or me or Keith) could add the newer version in the bazel-central-registry and then we could update here. I don’t think there’s anything we actually need in that release though so I think we’re not in any rush.

jamesvanhallen:
Hello, could you please update to latest 1.3.0
now I have conflicts

PR #4857, opened 2023-04-02
Add example to proof that shorthand optional binding works with `unused_declaration` rule

SimplyDanny:
This now only works with Swift 5.8/Xcode 14.3. Is it okay to restrict the rule to Swift 5.8 subsequently?

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/UnusedDeclarationRuleExamples.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/UnusedDeclarationRuleExamples.swift
@@ -128,7 +128,7 @@ struct UnusedDeclarationRuleExamples {
           "hello"
         }
         """)
-    ] + platformSpecificNonTriggeringExamples
+    ] + platformSpecificNonTriggeringExamples + versionSpecificNonTriggeringExamples
 
     static let triggeringExamples = [
         Example("""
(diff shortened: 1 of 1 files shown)


SimplyDanny:
Okay for you, @jpsim?

CHANGELOG.md
```diff
@@ -2,7 +2,10 @@
 
 #### Breaking
 
-* None.
+* The `unused_declaration` rule requires Swift 5.8/Xcode 14.3 to work properly
```
jpsim, line 5: I think this can go in the Bug Fixes category.

Source/SwiftLintBuiltInRules/Rules/Lint/UnusedDeclarationRule.swift
```diff
@@ -27,6 +27,7 @@ struct UnusedDeclarationRule: ConfigurationProviderRule, AnalyzerRule, Collectin
         name: "Unused Declaration",
         description: "Declarations should be referenced at least once within all files linted",
         kind: .lint,
+        minSwiftVersion: .fiveDotEight,
```
jpsim, line 30: I don't think this needs to be a breaking change, since the rule works fine on older Swift versions if you're not using shorthand optional bindings.

Source/SwiftLintBuiltInRules/Rules/Lint/UnusedDeclarationRuleExamples.swift
```diff
@@ -44,6 +44,15 @@ struct UnusedDeclarationRuleExamples {
         }
         _ = ResponseModel()
         """),
+        Example("""
```
jpsim, line 47: If this test fails on Swift < 5.8, you can conditionally add this example only when compiling with 5.8+ like we do for other rules.

jpsim (approved):
I don't think this needs to be a breaking change.

jpsim:
In fact I don't even think this needs a changelog entry since there are no code changes, it's just that using unused_declaration with shorthand optional bindings in Swift < 5.8 has false positives.

PR #4862, opened 2023-04-03
Fix false positive in rule `indentation_width`

svenmuennich:
We noticed a lot of false positives for rule `indentation_width` when setting the newly introduced option `include_multiline_strings` to `false`. 

Here's an example where this would produce a false positive:
```swift
let x = [
    "key": [
        ["nestedKey": "string"],
    ],
]
```

This PR fixes the rule's detection of lines belonging to a multiline line string by considering only this lines that represent the content of the string, not its delimiters.

The indentation of the closing delimiter of a multiline string is always tested for rule violations (independently from `include_multiline_strings`), as IMHO the delimiter acts more like a closing brace than a string. Hence the following example always produces a violation:
```swift
let x = """
    content
   """
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#4860](https://github.com/realm/SwiftLint/issues/4860)
 
+* Fix false positives in `indentation_width` rule.  
+  [Sven Münnich](https://github.com/svenmuennich)
+
 ## 0.51.0: bzllint
 
 #### Breaking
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -28,6 +28,8 @@
 * Fix `lower_acl_than_parent` rule rewriter by preserving leading whitespace.  
   [SimplyDanny](https://github.com/SimplyDanny)
   [#4860](https://github.com/realm/SwiftLint/issues/4860)
+* Fix false positives in `indentation_width` rule.
```
SimplyDanny, line 31: Please add an empty line between the entries.

SimplyDanny (approved):
Looks good to me!

SimplyDanny:
Thank you, @svenmuennich!

PR #4863, opened 2023-04-04
Disabled rules in child configs are no longer automatically ignored if they are opt-in

### LINKED ISSUE #4859: `disabled_rules` does not work sometimes in a child config
### New Issue Checklist

### Describe the bug

To give some context, I have a primary `.swiftlint.yml`, and then I have a `.swiftlint-incremental.yml` that enables more rules than my usual linting.

I can get this to work, but lots of other configurations are behaving in surprising and unexpected ways for me.

So, what works:

`.swiftlint.yml`:

```
optin_rules:
  - all # you don't have to use all, ordinary rule names will work too
disabled_rules:
  - rule1
  - rule2
  - rule3
```

`.swiftlint-incremental.yml`:

```
parent_config: .swiftlint.yml

disabled_rules:
#  - rule1
  - rule2
  - rule3
```

When SwiftLint is run with the `.swiftlint-incremental.yml` file, all the rules, except `rule2` and `rule3` will be run, which is exactly what I want.

Here are some cases that don't work as I expect (all assume we are at the top of the SwiftLint source tree, and the file names are `.swiftlint-new.yml` and `.swiftlint-child.yml`). I'm just using `implicit_return` as an example - it could be any opt-in rule.

```
% cat .swiftlint-new.yml
included:
  - Plugins
  - Source
  - Tests
excluded:
  - Tests/SwiftLintFrameworkTests/Resources
disabled_rules:
  - type_body_length
  - function_body_length
  - large_tuple
```

As expected, `implicit_return` is not enabled.

```
 % swiftlint rules --config .swiftlint-new.yml | perl -ae 'print if $. < 4 || $F[1] eq "implicit_return" || $. == 229'
+------------------------------------------+--------+-------------+------------------------+-------------+----------+----------------+---------------+
| identifier                               | opt-in | correctable | enabled in your config | kind        | analyzer | uses sourcekit | configuration |
+------------------------------------------+--------+-------------+------------------------+-------------+----------+----------------+---------------+
| implicit_return                          | yes    | yes         | no                     | style       | no       | yes            | severity: ... |
+------------------------------------------+--------+-------------+------------------------+-------------+----------+----------------+---------------+
```

Then:

```
 % cat .swiftlint-child.yml                                                                                           
parent_config: .swiftlint-new.yml

opt_in_rules:
  - implicit_return

disabled_rules:
  - implicit_return
```

```
% swiftlint rules --config .swiftlint-child.yml | perl -ae 'print if $. < 4 || $F[1] eq "implicit_return" || $. == 229'
+------------------------------------------+--------+-------------+------------------------+-------------+----------+----------------+---------------+
| identifier                               | opt-in | correctable | enabled in your config | kind        | analyzer | uses sourcekit | configuration |
+------------------------------------------+--------+-------------+------------------------+-------------+----------+----------------+---------------+
| implicit_return                          | yes    | yes         | yes                    | style       | no       | yes            | severity: ... |
+------------------------------------------+--------+-------------+------------------------+-------------+----------+----------------+---------------+
```

Even though it is explicitly disabled in the child configuration, SwiftLint is still enabling the `implicit_return` rule.

If I remove or comment out the `parent_config:` entry from `.swiftlint-child.yml` the problem goes away, and `implicit_return` is correctly disabled.

And this is the basic crux of the problem it seems - if you have a `parent_config:` entry and an `opt_in_rules`  section, then `disabled_rules` no longer works ...

Maybe similar-ish to https://github.com/realm/SwiftLint/issues/3645, but that is files, not rules.

See above

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? `0.51.0`

See above

I'm using `parent_config`, but not nesting per se.

```
Xcode 14.2
Build version 14C18
```

See above.

### LINKED ISSUE #4876: Question: What is the expected behaviour of `opt_in_rules` and `disabled_rules` in a child configuration?
### New Issue Checklist

### Describe the bug

Replaces #4859

Some of the behaviours of `opt_in_rules` and `disabled_rules` seem, at face value, to be incorrect.

For example, consider a case where the parent.yml looks like:

```
opt_in_rules:
  - implicit_return
disabled_rules:
  - implicit_return
```

and the child.yml looks like:

```
parent_config: parent.yml
```

Naively, I would expect the specified opt-in rule to be disabled in the parent configuration, and to also be disabled in the child, but in practice, with `0.51.0`, the rule will be **enabled** in the child, according to the `rules` subcommand:

```
% swiftlint rules  --config parent.yml | grep implicit_return
| implicit_return                          | yes    | yes         | NO                     | style       | no       | yes            | severity: ... |
```

```
% swiftlint rules  --config child.yml | grep implicit_return
| implicit_return
```

And observing actual linting behaviour, given

```
class ImplicitReturn {
   var foo: Int {
       return 0
   }
}
```

```
 % swiftlint lint  --config parent.yml ImplicitReturn.swift 
Linting Swift files at paths ImplicitReturn.swift
Linting 'ImplicitReturn.swift' (1/1)
%
% swiftlint lint --config child.yml ImplicitReturn.swift 
Linting Swift files at paths ImplicitReturn.swift
Linting 'ImplicitReturn.swift' (1/1)
/Users/martin.redington/Documents/Hudl/Source/SwiftLint/ImplicitReturn.swift:3:8: warning: Implicit Return Violation: Prefer implicit returns in closures, functions and getters (implicit_return)
```

Similarly, if the child configuration is

```
parent_config: parent.yml

opt_in_rules:
  - implicit_return
disabled_rules:
  - implicit_return

```

and the parent has a "default" configuration (using opt_in_rules and disabled_rules), no matter what the parent settings, the specified rule will always be **enabled** in the child, which seems wrong.

Below is the complete truth table for whether an optin rule is enabled in the merged configuration, as a function of whether it is listed in the opt_in_rules, and/or the disabled_rules sections of the parent and child configurations.

The "unexpected" cases are asterisked.

```
        parent    parent      child      child
        opt_in   disabled    opt_in     disabled    current     proposed

0       false      false      false      false       false       false
1        true      false      false      false        true        true
2       false       true      false      false       false       false
3        true       true      false      false        true       false *
4       false      false       true      false        true        true
5        true      false       true      false        true        true
6       false       true       true      false        true        true
7        true       true       true      false        true        true
8       false      false      false       true       false       false
9        true      false      false       true       false       false
10      false       true      false       true       false       false
11       true       true      false       true       false       false
12      false      false       true       true        true       false *
13       true      false       true       true        true       false *
14      false       true       true       true        true       false *
15       true       true       true       true        true       false *
```

The current behaviour is actually "better" for some use cases - if I want my child configuration to be stricter than the parent, the current behaviour can actually end up being more compact than the new configuration, and I suspect that fixing this would be a breaking change for some users, but the current behaviour nevertheless seems counter-intuitive to me.

My "question" is, is the current behaviour intentional, or it should it actually behave as the `proposed` column above, or in some other way?

A similar situation occurs when the parent configuration uses `only_rules`, and the child has a default configuration.

The truth table for this case (of whether the rule is enabled in the merged configuration) is:

```
       child       child
      opt_in     disabled    current   proposed

0      false      false       true       true
1       true      false       true       true
2      false       true      false      false
3       true       true       true      false *
```

Examining case 3

parent.yml:

```
only_rules:
   - implicit_return
```

and child.yml:

```
parent_config: parent.yml

opt_in_rules:
  - implicit_return
disabled_rules:
  - implicit_return
```

Currently (in main/0.51.0) the rule would end up **enabled** for the child, even though it is explicitly disabled in the child configuration.

See above

### Environment

`0.51.0`

See above

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? Nested no, but parent and child yes.

```
Xcode 14.2
Build version 14C18
```

See description above

mildm8nnered:
Fixes some unexpected behaviors in the interaction between parent and child configurations.

For example, if a rule is enabled and disabled in the parent config, and not mentioned in the child config (case 3 below), currently the rule will become re-enabled in the merged configuration.

Resolves #4876, which documents the unexpected behaviors.

Below is the truth table for whether an optin rule is enabled in the child configuration, as a function of whether it is listed in the `opt_in_rules`, and/or the `disabled_rules` sections of the parent and child configuration.

`current` is the value for `main`/`0.51.0`, and `new` is the value produced by this branch.

asterisked rows have changed.

```
        parent    parent      child      child
        opt_in   disabled    opt_in     disabled    current       new

0       false      false      false      false       false       false
1        true      false      false      false        true        true
2       false       true      false      false       false       false
3        true       true      false      false        true       false *
4       false      false       true      false        true        true
5        true      false       true      false        true        true
6       false       true       true      false        true        true
7        true       true       true      false        true        true
8       false      false      false       true       false       false
9        true      false      false       true       false       false
10      false       true      false       true       false       false
11       true       true      false       true       false       false
12      false      false       true       true        true       false *
13       true      false       true       true        true       false *
14      false       true       true       true        true       false *
15       true       true       true       true        true       false *
```

To take case 3 in detail, if an optin rule is listed in opt_in_rules and disabled_rules in the parent configuration, and not mentioned in the child, then currently (in `main`/`0.51.0`) the rule will be disabled for the parent configuration, but **enabled** in the child.

parent.yml:

```
opt_in_rules:
  - implicit_return
disabled_rules:
  - implicit_return
```

child.yml:

```
parent_config: parent.yml
```

checking enablement:

```
% swiftlint rules  --config parent.yml | grep implicit_return
| implicit_return                          | yes    | yes         | NO                     | style       | no       | yes            | severity: ... |
```

```
% swiftlint rules  --config child.yml | grep implicit_return
| implicit_return                          | yes    | yes         | YES                    | style       | no       | yes            | severity: ... |
```

and actual behaviour, given:

```
class ImplicitReturn {
   var foo: Int {
       return 0
   }
}
```

```
 % swiftlint lint  --config parent.yml ImplicitReturn.swift 
Linting Swift files at paths ImplicitReturn.swift
Linting 'ImplicitReturn.swift' (1/1)
%
% swiftlint lint --config child.yml ImplicitReturn.swift 
Linting Swift files at paths ImplicitReturn.swift
Linting 'ImplicitReturn.swift' (1/1)
/Users/martin.redington/Documents/Hudl/Source/SwiftLint/ImplicitReturn.swift:3:8: warning: Implicit Return Violation: Prefer implicit returns in closures, functions and getters (implicit_return)
```

Similarly, in cases 12 through 15, the rule is both mentioned in the child's `opt_in_rules`, and `disabled_rules`, which should surely mean the rule ends up disabled in the child, no matter what the parent says.

This PR changes the processing of these configurations, to produce the truth values shown in the `new` column above, which seem more sensible than the current values.

There is a related issue when the parent has only rules, and the child has a default configuration.

Here is the truth table for this case:

```
       child       child
       optin     disabled    current     new

0      false      false       true       true
1       true      false       true       true
2      false       true      false      false
3       true       true       true      false *
```

Examining case 3

parent.yml:

```
only_rules:
   - implicit_return
```

and child.yml:

```
parent_config: parent.yml

opt_in_rules:
  - implicit_return
disabled_rules:
  - implicit_return
```

Currently (in `main`/`0.51.0`) the rule would end up enabled for the child, even though it is explicitly disabled in the child configuration.

### DIFF
--- a/Tests/SwiftLintFrameworkTests/ConfigurationTests+MultipleConfigs.swift
+++ b/Tests/SwiftLintFrameworkTests/ConfigurationTests+MultipleConfigs.swift
@@ -302,6 +302,25 @@ extension ConfigurationTests {
     }
     
     func testParentChildOptInAndDisable() {
+        func isEnabledInChild(
+            ruleType: Rule.Type,
+            enabledInParent: Bool,
+            disabledInParent: Bool,
+            enabledInChild: Bool,
+            disabledInChild: Bool
+        ) -> Bool {
+            let ruleIdentifier = ruleType.description.identifier
+            let parentConfiguration = Configuration(rulesMode: .default(
+                disabled: disabledInParent ? [ruleIdentifier] : [],
+                optIn: enabledInParent ? [ruleIdentifier] : []
+            ))
+            let childConfiguration = Configuration(rulesMode: .default(
+                disabled: disabledInChild ? [ruleIdentifier] : [],
+                optIn: enabledInChild ? [ruleIdentifier] : []
+            ))
+            let mergedConfiguration = parentConfiguration.merged(withChild: childConfiguration, rootDirectory: "")
+            return mergedConfiguration.contains(rule: ruleType)
+        }
 //        let expectedResults = "0101111100001111"
         let expectedResults = "0100111100000000"
         XCTAssertEqual(expectedResults.count, 4 * 4)
@@ -325,25 +344,6 @@ extension ConfigurationTests {
             XCTAssertEqual(expectedResult, result)
         }
     }
-
-    private func isEnabledInChild(
-        ruleType: Rule.Type,
-        enabledInParent: Bool,
-        disabledInParent: Bool,
-        enabledInChild: Bool,
-        disabledInChild: Bool
-    ) -> Bool {
-        let ruleIdentifier = ruleType.description.identifier
-        let parentConfiguration = Configuration(rulesMode: .default(
-            disabled: disabledInParent ? [ruleIdentifier] : [],
-            optIn: enabledInParent ? [ruleIdentifier] : []
-        ))
-        let childConfiguration = Configuration(rulesMode: .default(
-            disabled: disabledInChild ? [ruleIdentifier] : [],
-            optIn: enabledInChild ? [ruleIdentifier] : []
-        ))
-        let mergedConfiguration = parentConfiguration.merged(withChild: childConfiguration, rootDirectory: "")
-        return mergedConfiguration.contains(rule: ruleType)
     }
     
     // MARK: - Remote Configs
--- a/Tests/SwiftLintFrameworkTests/ConfigurationTests+MultipleConfigs.swift
+++ b/Tests/SwiftLintFrameworkTests/ConfigurationTests+MultipleConfigs.swift
@@ -337,7 +337,8 @@ extension ConfigurationTests {
         ]
         XCTAssertEqual(testCases.unique.count, 4 * 4)
         let ruleType = ImplicitReturnRule.self
-        XCTAssertTrue(ruleType is OptInRule)
+        let isOptIn = ruleType is OptInRule.Type
+        XCTAssertTrue(isOptIn)
         let ruleIdentifier = ruleType.description.identifier
         for testCase in testCases {
             let parentConfiguration = Configuration(rulesMode: .default(
@@ -371,7 +372,7 @@ extension ConfigurationTests {
         ]
         XCTAssertEqual(testCases.unique.count, 2 * 2)
         let ruleType = BlanketDisableCommandRule.self
-        XCTAssertFalse(ruleType is OptInRule)
+        XCTAssertFalse(ruleType is OptInRule.Type)
         let ruleIdentifier = ruleType.description.identifier
         for testCase in testCases {
             let parentConfiguration = Configuration(
@@ -403,7 +404,7 @@ extension ConfigurationTests {
         ]
         XCTAssertEqual(testCases.unique.count, 2 * 2)
         let ruleType = ImplicitReturnRule.self
-        XCTAssertTrue(ruleType is OptInRule)
+        XCTAssertTrue(ruleType is OptInRule.Type)
         let ruleIdentifier = ruleType.description.identifier
         let parentConfiguration = Configuration(rulesMode: .only([ruleIdentifier]))
         for testCase in testCases {
--- a/Tests/SwiftLintFrameworkTests/ConfigurationTests+MultipleConfigs.swift
+++ b/Tests/SwiftLintFrameworkTests/ConfigurationTests+MultipleConfigs.swift
@@ -304,19 +304,19 @@ extension ConfigurationTests {
     func testParentChildOptInAndDisable() {
         func isEnabledInChild(
             ruleType: Rule.Type,
-            enabledInParent: Bool,
+            optedInInParent: Bool,
             disabledInParent: Bool,
-            enabledInChild: Bool,
+            optedInInChild: Bool,
             disabledInChild: Bool
         ) -> Bool {
             let ruleIdentifier = ruleType.description.identifier
             let parentConfiguration = Configuration(rulesMode: .default(
                 disabled: disabledInParent ? [ruleIdentifier] : [],
-                optIn: enabledInParent ? [ruleIdentifier] : []
+                optIn: optedInInParent ? [ruleIdentifier] : []
             ))
             let childConfiguration = Configuration(rulesMode: .default(
                 disabled: disabledInChild ? [ruleIdentifier] : [],
-                optIn: enabledInChild ? [ruleIdentifier] : []
+                optIn: optedInInChild ? [ruleIdentifier] : []
             ))
             let mergedConfiguration = parentConfiguration.merged(withChild: childConfiguration, rootDirectory: "")
             return mergedConfiguration.contains(rule: ruleType)
@@ -332,18 +332,18 @@ extension ConfigurationTests {
             (value ? " " : "") + " \(value)"
         }
         for i in 0..<expectedResults.count {
-            let enabledInParent = i & 1 == 1
+            let optedInInParent = i & 1 == 1
             let disabledInParent = i & 2 == 2
-            let enabledInChild = i & 4 == 4
+            let optedInInChild = i & 4 == 4
             let disabledInChild = i & 8 == 8
             let result = isEnabledInChild(
                 ruleType: ImplicitReturnRule.self,
-                enabledInParent: enabledInParent,
+                optedInInParent: optedInInParent,
                 disabledInParent: disabledInParent,
-                enabledInChild: enabledInChild,
+                optedInInChild: optedInInChild,
                 disabledInChild: disabledInChild
             )
-            // print(">>>> \(i) \(format(enabledInParent)) \(format(disabledInParent)) \(format(enabledInChild)) \(format(disabledInChild)) \(format(result))")
+            // print(">>>> \(i) \(format(optedInInParent)) \(format(disabledInParent)) \(format(optedInInChild)) \(format(disabledInChild)) \(format(result))")
             XCTAssertEqual(expectedResults[i], result)
         }
     }
--- a/Tests/SwiftLintFrameworkTests/ConfigurationTests+MultipleConfigs.swift
+++ b/Tests/SwiftLintFrameworkTests/ConfigurationTests+MultipleConfigs.swift
@@ -340,7 +340,7 @@ extension ConfigurationTests {
                 optedInInChild: optedInInChild,
                 disabledInChild: disabledInChild
             )
-            XCTAssertEqual(expectedResults[index], result)
+            XCTAssertEqual(result, expectedResults[index])
         }
     }
 
@@ -369,7 +369,7 @@ extension ConfigurationTests {
                 optedInInChild: optedInInChild,
                 disabledInChild: disabledInChild
             )
-            XCTAssertEqual(expectedResults[index], result)
+            XCTAssertEqual(result, expectedResults[index])
         }
     }
 
--- a/Tests/SwiftLintFrameworkTests/ConfigurationTests+MultipleConfigs.swift
+++ b/Tests/SwiftLintFrameworkTests/ConfigurationTests+MultipleConfigs.swift
@@ -337,8 +337,7 @@ extension ConfigurationTests {
         ]
         XCTAssertEqual(testCases.unique.count, 4 * 4)
         let ruleType = ImplicitReturnRule.self
-        let isOptIn = ruleType is OptInRule.Type
-        XCTAssertTrue(isOptIn)
+        XCTAssertTrue(ruleType.isOptInRule)
         let ruleIdentifier = ruleType.description.identifier
         for testCase in testCases {
             let parentConfiguration = Configuration(rulesMode: .default(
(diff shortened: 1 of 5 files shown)


Tests/SwiftLintFrameworkTests/ConfigurationTests+MultipleConfigs.swift
SimplyDanny, line 1: I'm struggling a bit with these tests. While I get the intention and understand that you save a lot of code with the way the test cases are constructed, I actually would prefer explicit separate test cases with a descriptive name.

SimplyDanny (commented):
The change looks good to me! However, I'm having a hard time following the tests. 🫣

Tests/SwiftLintFrameworkTests/ConfigurationTests+MultipleConfigs.swift
mildm8nnered: I hear you - I do want to cover all of the cases here, because they are quite twisty and hard to understand, but I'm not super-enthusiastic about writing 16 tests (for the one truth table) with names like

`testOptedInParentOptedInChildDisabledInParent()`

Could something like the below be a reasonable compromise, where the truth table is more obvious?

```
    func testParentChildOptInAndDisable() {
        typealias TestCase = (
            optedInInParent: Bool,
            disabledInParent: Bool,
            optedInInChild: Bool,
            disabledInChild: Bool,
            isEnabled: Bool
        )
        let testCases: [TestCase] = [
            (false, false, false, false, false),
            (true, false, false, false, true),
            (false, true, false, false, false),
            (true, true, false, false, false),
            (false, false, true, false, true),
            (true, false, true, false, true),
            (false, true, true, false, true),
            (true, true, true, false, true),
            (false, false, false, true, false),
            (true, false, false, true, false),
            (false, true, false, true, false),
            (true, true, false, true, false),
            (false, false, true, true, false),
            (true, false, true, true, false),
            (false, true, true, true, false),
            (true, true, true, true, false)
        ]
        XCTAssertEqual(testCases.count, 4 * 4)
        let ruleType = ImplicitReturnRule.self
        let ruleIdentifier = ruleType.description.identifier
        for testCase in testCases {
            let parentConfiguration = Configuration(rulesMode: .default(
                disabled: testCase.disabledInParent ? [ruleIdentifier] : [],
                optIn: testCase.optedInInParent ? [ruleIdentifier] : []
            ))
            let childConfiguration = Configuration(rulesMode: .default(
                disabled: testCase.disabledInChild ? [ruleIdentifier] : [],
                optIn: testCase.optedInInChild ? [ruleIdentifier] : []
            ))
            let mergedConfiguration = parentConfiguration.merged(withChild: childConfiguration, rootDirectory: "")
            let isEnabled = mergedConfiguration.contains(rule: ruleType)
            let message = "optedInInParent = \(testCase.optedInInParent), disabledInParent = \(testCase.disabledInParent), " +
            "optedInInChild = \(testCase.optedInInChild), disabledInChild = \(testCase.disabledInChild)"
            XCTAssertEqual(isEnabled, testCase.isEnabled, message)
        }
    }
```

SimplyDanny: Yes better, looks like a good compromise. 👍

The line asserting the number of test cases `XCTAssertEqual(testCases.count, 4 * 4)` should probably also make sure all lines are unique like `XCTAssertEqual(testCases.unique.count, 4 * 4)`.

Tests/SwiftLintFrameworkTests/ConfigurationTests+MultipleConfigs.swift
```diff
@@ -301,6 +301,113 @@ extension ConfigurationTests {
...
+            TestCase(optedInInParent: false, disabledInParent: true, optedInInChild: false, disabledInChild: true, isEnabled: false),
+            TestCase(optedInInParent: true, disabledInParent: true, optedInInChild: false, disabledInChild: true, isEnabled: false),
+            TestCase(optedInInParent: false, disabledInParent: false, optedInInChild: true, disabledInChild: true, isEnabled: false),
+            TestCase(optedInInParent: true, disabledInParent: false, optedInInChild: true, disabledInChild: true, isEnabled: false),
+            TestCase(optedInInParent: false, disabledInParent: true, optedInInChild: true, disabledInChild: true, isEnabled: false),
+            TestCase(optedInInParent: true, disabledInParent: true, optedInInChild: true, disabledInChild: true, isEnabled: false)
+            // swiftlint:enable line_length
+        ]
+        XCTAssertEqual(testCases.unique.count, 4 * 4)
+        let ruleType = ImplicitReturnRule.self
```
SimplyDanny, line 333: It's probably helpful to assert that this exemplary rule is an `OptInRule` in case that changes at some time later and suddenly the test fails.

Same for the other test case where the rule is active by default.

Tests/SwiftLintFrameworkTests/ConfigurationTests+MultipleConfigs.swift
```diff
@@ -301,6 +301,113 @@ extension ConfigurationTests {
...
+            ))
+            let childConfiguration = Configuration(rulesMode: .default(
+                disabled: testCase.disabledInChild ? [ruleIdentifier] : [],
+                optIn: testCase.optedInInChild ? [ruleIdentifier] : []
+            ))
+            let mergedConfiguration = parentConfiguration.merged(withChild: childConfiguration, rootDirectory: "")
+            let isEnabled = mergedConfiguration.contains(rule: ruleType)
+            let message = "optedInInParent = \(testCase.optedInInParent) " +
+                          "disabledInParent = \(testCase.disabledInParent), " +
+            "optedInInChild = \(testCase.optedInInChild), disabledInChild = \(testCase.disabledInChild)"
```
SimplyDanny, line 348:
```suggestion
            let message = "optedInInParent = \(testCase.optedInInParent) " +
                          "disabledInParent = \(testCase.disabledInParent), " +
                          "optedInInChild = \(testCase.optedInInChild), " +          
                          "disabledInChild = \(testCase.disabledInChild)"
```

Alternatively, you may add the message to the `TestCase` struct as a computed property which would avoid the `testCase.`.

SimplyDanny (approved):

Tests/SwiftLintFrameworkTests/ConfigurationTests+MultipleConfigs.swift
```diff
@@ -301,6 +301,122 @@ extension ConfigurationTests {
...
+        }
+        let testCases: [TestCase] = [
+            TestCase(disabledInParent: false, disabledInChild: false, isEnabled: true),
+            TestCase(disabledInParent: true, disabledInChild: false, isEnabled: false),
+            TestCase(disabledInParent: false, disabledInChild: true, isEnabled: false),
+            TestCase(disabledInParent: true, disabledInChild: true, isEnabled: false)
+        ]
+        XCTAssertEqual(testCases.unique.count, 2 * 2)
+        let ruleType = BlanketDisableCommandRule.self
+        XCTAssertFalse(ruleType is OptInRule.Type)
```
SimplyDanny, line 374: Either use `ruleType.isOptInRule` here or inline it also in the other two test methods (which I'd prefer over the extension).

mildm8nnered: The problem I was having was that the obvious

```
        let ruleType = ImplicitReturnRule.self
        XCTAssertTrue(ruleType is OptInRule.Type)
```

produces a compilation warning - `'is' test is always true`

I went with

```
        let ruleType = ImplicitReturnRule.self
        XCTAssertTrue((ruleType as Any) is OptInRule.Type)
```

as workaround (unless you have a better one).

SimplyDanny: I see. Fair enough to remove the assertion again. I leave the decision up to you. 😉

mildm8nnered: I think the workaround is not too ugly, so I'm happy to leave it in. I definitely like the idea of validating that the rules are still optin (or not).

SimplyDanny:
Thanks for all the effort, @mildm8nnered!

PR #4872, opened 2023-04-06
Add new `disallow_top_level_variables` rule

shivatinker:
Added new `disallow_top_level_variables` rule, that we want to use in our team. For now, we are using `prefixed_toplevel_constant` rule to locate global constants and move them to scopes (enums/classes/structs). 

Feel free to ask any questions or suggest improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@
 
 * Trigger `prefer_self_in_static_references` rule on more type references.
   [SimplyDanny](https://github.com/SimplyDanny)
+  
+* Add new `disallow_top_level_variables` rule that disallows declaring variables
+  and constants as the top-level declarations.
+  [Andrii Zinoviev](https://github.com/shivatinker)
 
 #### Bug Fixes
 
--- a/Source/SwiftLintFramework/Rules/Style/DisallowTopLevelVariablesRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/DisallowTopLevelVariablesRule.swift
@@ -1,0 +1,64 @@
+import SwiftSyntax
+
+struct DisallowTopLevelVariablesRule: SwiftSyntaxRule, OptInRule, ConfigurationProviderRule {
+    var configuration: SeverityConfiguration = .init(.warning)
+
+    static let description: RuleDescription = .init(
+        identifier: "disallow_top_level_variables",
+        name: "Disallow Top-Level Variables",
+        description: "Top-Level variables and constants should be avoided",
+        kind: .style,
+        nonTriggeringExamples: [
+            Example(
+                """
+                struct A {
+                    public var a: Int = 12
+                }
+                """
+            )
+        ],
+        triggeringExamples: [
+            Example("var ↓a: Int = 11"),
+            Example("let ↓a: Int = 11"),
+            Example("private var ↓a: Int = 11"),
+            Example("public var ↓a: String?"),
+            Example("public var ↓b: Int { return 42 }")
+        ])
+
+    func makeVisitor(file: SwiftLintFile) -> ViolationsSyntaxVisitor {
+        Visitor()
+    }
+}
+
+private extension DisallowTopLevelVariablesRule {
+    final class Visitor: ViolationsSyntaxVisitor {
+        init() {
+            super.init(viewMode: .sourceAccurate)
+        }
+
+        override var skippableDeclarations: [DeclSyntaxProtocol.Type] { .all }
+
+        override func visitPost(_ node: VariableDeclSyntax) {
+            let kind = node.bindingKeyword.tokenKind
+            guard kind == .keyword(.let) || kind == .keyword(.var) else {
+                return
+            }
+
+            for binding in node.bindings {
+                guard let pattern = binding.pattern.as(IdentifierPatternSyntax.self) else {
+                    continue
+                }
+
+                violations.append(pattern.positionAfterSkippingLeadingTrivia)
+            }
+        }
+
+        override func visit(_ node: CodeBlockSyntax) -> SyntaxVisitorContinueKind {
+            .skipChildren
+        }
+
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,10 @@
 
 * Trigger `prefer_self_in_static_references` rule on more type references.
   [SimplyDanny](https://github.com/SimplyDanny)
+  
+* Add new `disallow_top_level_variables` rule that disallows declaring variables
+  and constants as the top-level declarations.
```
SimplyDanny, line 30:
```suggestion
* Add new `disallow_top_level_variables` rule that disallows variable and
  constant declarations on top level.  
```
The two spaces at the end are important to enforce a line break.

Source/SwiftLintFramework/Rules/Style/DisallowTopLevelVariablesRule.swift
```diff
@@ -0,0 +1,64 @@
+import SwiftSyntax
+
+struct DisallowTopLevelVariablesRule: SwiftSyntaxRule, OptInRule, ConfigurationProviderRule {
+    var configuration: SeverityConfiguration = .init(.warning)
+
+    static let description: RuleDescription = .init(
+        identifier: "disallow_top_level_variables",
+        name: "Disallow Top-Level Variables",
+        description: "Top-Level variables and constants should be avoided",
```
SimplyDanny, line 9:
```suggestion
        description: "Top-level variables and constants should be avoided",
```

Source/SwiftLintFramework/Rules/Style/DisallowTopLevelVariablesRule.swift
```diff
@@ -0,0 +1,64 @@
...
+    func makeVisitor(file: SwiftLintFile) -> ViolationsSyntaxVisitor {
+        Visitor()
+    }
+}
+
+private extension DisallowTopLevelVariablesRule {
+    final class Visitor: ViolationsSyntaxVisitor {
+        init() {
+            super.init(viewMode: .sourceAccurate)
+        }
```
SimplyDanny, line 37: It's preferred to skip this constructor and initialize the visitor like `Visitor(viewMode: .sourceAccurate)` in line 29.

Source/SwiftLintFramework/Rules/Style/DisallowTopLevelVariablesRule.swift
```diff
@@ -0,0 +1,64 @@
...
+
+        override var skippableDeclarations: [DeclSyntaxProtocol.Type] { .all }
+
+        override func visitPost(_ node: VariableDeclSyntax) {
+            let kind = node.bindingKeyword.tokenKind
+            guard kind == .keyword(.let) || kind == .keyword(.var) else {
```
SimplyDanny, line 43: Is there a need to check for the binding keywords explicitly? Might be redundant as we are already dealing with a `VariableDeclSyntax`.

Source/SwiftLintFramework/Rules/Style/DisallowTopLevelVariablesRule.swift
```diff
@@ -0,0 +1,64 @@
...
+                return
+            }
+
+            for binding in node.bindings {
+                guard let pattern = binding.pattern.as(IdentifierPatternSyntax.self) else {
```
SimplyDanny, line 48: What about tuples? Here again, I don't think this explicit check is required.

Source/SwiftLintFramework/Rules/Style/DisallowTopLevelVariablesRule.swift
```diff
@@ -0,0 +1,64 @@
...
+                violations.append(pattern.positionAfterSkippingLeadingTrivia)
+            }
+        }
+
+        override func visit(_ node: CodeBlockSyntax) -> SyntaxVisitorContinueKind {
+            .skipChildren
+        }
+
+        override func visit(_ node: ClosureExprSyntax) -> SyntaxVisitorContinueKind {
+            .skipChildren
```
SimplyDanny, line 61: Would be great to have a test example for closures with variable declarations inside.

Source/SwiftLintFramework/Rules/Style/DisallowTopLevelVariablesRule.swift
```diff
@@ -0,0 +1,64 @@
...
+                guard let pattern = binding.pattern.as(IdentifierPatternSyntax.self) else {
+                    continue
+                }
+
+                violations.append(pattern.positionAfterSkippingLeadingTrivia)
+            }
+        }
+
+        override func visit(_ node: CodeBlockSyntax) -> SyntaxVisitorContinueKind {
+            .skipChildren
```
SimplyDanny, line 57: I cannot think of an example for this case? If you have one in mind, please add a test example for it.

Source/SwiftLintFramework/Rules/Style/DisallowTopLevelVariablesRule.swift
```diff
@@ -0,0 +1,64 @@
+import SwiftSyntax
+
+struct DisallowTopLevelVariablesRule: SwiftSyntaxRule, OptInRule, ConfigurationProviderRule {
+    var configuration: SeverityConfiguration = .init(.warning)
```
SimplyDanny, line 4: A matter of taste, but the preferred/consistent style would be `var configuration = SeverityConfiguration(.warning)`.

SimplyDanny (changes requested):
I think that's a neat little rule. Thank you for the idea and implementation!

shivatinker:
For now I do not have a lot of time to work on this, so I think we can close it until I will have time to finish this rule. 

Thanks for the review, @SimplyDanny , I will take your comments into account and create new PR when I will be ready.

PR #4873, opened 2023-04-07
Add automatic architecture detection to Xcode script configuration

FradSer:
This pull request adds automatic detection of the current architecture (Apple or Intel) to the Xcode script configuration. With this change, the script will now use the appropriate path for the swiftlint executable based on the current architecture.

The updated code replaces the hard-coded `/opt/homebrew/bin` path with a conditional statement that sets the `prefie` variable to the appropriate path based on the current architecture. If the current architecture is `x86_64*`, the `prefie` variable is set to `/usr/local/bin/`, otherwise it is set to `/opt/homebrew/bin/`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,9 +158,14 @@ folder by default. To instruct Xcode where to find SwiftLint, you can either add
 `/opt/homebrew/bin` to the `PATH` environment variable in your build phase
 
 ```bash
-export PATH="$PATH:/opt/homebrew/bin"
-if which swiftlint > /dev/null; then
-  swiftlint
+if [[ "$(uname -m)" = x86_64* ]]; then
+  prefie="/usr/local/bin/"
+else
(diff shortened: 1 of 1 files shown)


SimplyDanny (commented):
I actually like the `export PATH` more. What do you think about the following?
```bash
if [[ "$(uname -m)" == arm64 ]]; then
    export PATH="/opt/homebrew/bin:$PATH"
fi
if which swiftlint > /dev/null; then
  swiftlint
else
  echo "..."
fi
```

FradSer:
@SimplyDanny I have no objections, I will modify the code accordingly and use `export PATH` as suggested.

SimplyDanny (approved):

SimplyDanny:
Could you please rebase? That might fix the build issues.

PR #4882, opened 2023-04-12
Use Ubuntu 20.04 to release Linux binaries

gferon:
Using the latest version of Ubuntu leads to incompatibilities with `glibc` versions from LTS versions.


gferon:
Similar change to https://github.com/nicklockwood/SwiftFormat/pull/1371, although I'm not sure if these are all the places you'd have to change.

jpsim:
Thanks!

PR #4892, opened 2023-04-14
Update Ruby gems

jpsim:
* cocoapods: 1.11.3 -> 1.12.0
* jazzy: 0.14.2 -> 0.14.3


jpsim:
I'll probably need to update the Ruby version on the CI machines to do this. 🤔

PR #4893, opened 2023-04-14
[CI] Use rbenv to manage Ruby

jpsim:
And pin the Ruby version to 2.7.5. rbenv was installed on all CI machines using Homebrew, with Ruby 2.7.5 installed and set as the global version. The `pre-command` script was also updated on all agents to add rbenv binaries to the path and run `rbenv init`.

Remove rules_ruby integration and just use rbenv and ruby directly.


jpsim:
I had to install rbenv on all the CI machines for this to work.

PR #4897, opened 2023-04-15
Mildm8nnered skip unit tests in magic numbers rule

### LINKED ISSUE #4896: `no_magic_numbers` should not trigger test code
### Description

In test code, it's reasonable to use raw numbers in assertions. The `no_magic_numbers` rule thus reports a lot of findings in tests.

```swift
class MyTest: XCTestCase {
  func test() {
    let x = 4
    XCTAssertEqual(x, 7)  // Violation here
  }
}
```

Defining another variable `let expected = 7` would not improve style, but would rather make the code more opaque and difficult to reason about.

It's probably required to add a configuration option to specify which parent classes make a test class a test class. In the example, it's `XCTestCase` (the default of the option). This can be taken over from already existing rules specific for tests.

### Environment

* 0.51.0
* Homebrew

```yml
only_rules:
  - no_magic_numbers
```

mildm8nnered:
WIP

Fixes #4896

Adds a `test_parent_classes` to the Magic Number Rule.

Violations within a class that inherits directly from one of the specified parent classes (by default `XCTestCase` and `QuickSpec`) will be ignored.

This includes violations within inner classes of a class that inherits directly from one of the specified parent classes.

If users wish to detect violations of the rule within unit tests, they can simply set `test_parent_classes` to `[]`.

I also "aligned" some of the labels in other classes that also use a similar configuration (actually the same configuration, but typealiased). In some places we were referring to `testClasses`, which is actually `testParentClasses`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@
   [Martin Redington](https://github.com/mildm8nnered)
   [#4819](https://github.com/realm/SwiftLint/issues/4819)
 
-* Adds `test_parent_classes` option to the `magic_numbers` rule.
+* Adds `test_parent_classes` option to the `no_magic_numbers` rule.
   Violations within test classes will now be ignored by default.  
   [Martin Redington](https://github.com/mildm8nnered)
   [#4896](https://github.com/realm/SwiftLint/issues/4896)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,11 @@
   [Martin Redington](https://github.com/mildm8nnered)
   [#4819](https://github.com/realm/SwiftLint/issues/4819)
 
+* Adds `test_parent_classes` option to the `magic_numbers` rule.
+  Violations within test classes will now be ignored by default.  
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#4896](https://github.com/realm/SwiftLint/issues/4896)
+
 #### Bug Fixes
 
 * Fix `lower_acl_than_parent` rule rewriter by preserving leading whitespace.  
--- a/Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
@@ -120,18 +120,12 @@ private extension ExprSyntaxProtocol {
         while parent != nil {
             if
                 let classDecl = parent?.as(ClassDeclSyntax.self),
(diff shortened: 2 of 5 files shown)


mildm8nnered:
Should be resolved by #4897

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -92,3 +113,25 @@ private extension TokenSyntax {
...
+                return true
+            }
+            parent = parent?.parent
+        }
+        return false
+    }
+}
+
+private extension ClassDeclSyntax {
+    func isTestClass(testParentClasses: Set<String>) -> Bool {
```
SimplyDanny, line 134: The `isXCTestCase` already exists on `ClassDeclSyntax` objects.

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -44,6 +44,20 @@ struct NoMagicNumbersRule: SwiftSyntaxRule, OptInRule, ConfigurationProviderRule
...
+            """),
+            Example("""
+            class FooTests: XCTestCase {
+                let array: [Int] = []
+                let bar = array[42]
+            }
+            """),
+            Example("""
+            class FooTests: XCTestCase {
+                class Bar {
```
SimplyDanny, line 56: Nested classes are discussable if they are not themselves test classes. But I agree that this rule should be disabled in test code generally. So ignoring them as well is fine.

CHANGELOG.md
```diff
@@ -31,6 +31,11 @@
   [Martin Redington](https://github.com/mildm8nnered)
   [#4819](https://github.com/realm/SwiftLint/issues/4819)
 
+* Adds `test_parent_classes` option to the `magic_numbers` rule.
```
SimplyDanny, line 34:
```suggestion
* Adds `test_parent_classes` option to the `no_magic_numbers` rule.
```

SimplyDanny (approved):

Source/SwiftLintFramework/Rules/Idiomatic/NoMagicNumbersRule.swift
mildm8nnered: I should have remembered that :-)

mildm8nnered:
CR comments addressed

SimplyDanny:
Thank you for the fast implementation!

Issue #4782, opened 2023-02-24
Class  X is implemented in both A and B.  One of the two will be used. Which one is undefined.


egesucu:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

This will happen when you analyze the code with "swiftlint" or "swiftlint analyze". It will not break the build or give any crashes.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
objc[36713]: Class _TtC11SwiftSyntax16BumpPtrAllocator is implemented in both /opt/homebrew/Cellar/swiftlint/0.50.3/bin/swiftlint (0x10140c0d0) and /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/host/libSwiftSyntax.dylib (0x1051847c8). One of the two will be used. Which one is undefined.
objc[36713]: Class _TtC11SwiftSyntax35IncrementalParseReusedNodeCollector is implemented in both /opt/homebrew/Cellar/swiftlint/0.50.3/bin/swiftlint (0x10140c288) and /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/host/libSwiftSyntax.dylib (0x105184978). One of the two will be used. Which one is undefined.
objc[36713]: Class _TtC11SwiftSyntax26IncrementalParseTransition is implemented in both /opt/homebrew/Cellar/swiftlint/0.50.3/bin/swiftlint (0x10140c320) and /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/host/libSwiftSyntax.dylib (0x105184a10). One of the two will be used. Which one is undefined.
objc[36713]: Class _TtC11SwiftSyntax23SourceLocationConverter is implemented in both /opt/homebrew/Cellar/swiftlint/0.50.3/bin/swiftlint (0x10140e188) and /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/host/libSwiftSyntax.dylib (0x105184b50). One of the two will be used. Which one is undefined.
[...]

PR #4901, opened 2023-04-17
Workaround dyld warning about SwiftSyntax classes

keith:
This uses a recent but unannounced (also read as private) feature of
dyld where it ignores duplicate Objective-C classes when they're in a
special format in the binary. https://github.com/apple-oss-distributions/dyld/blob/c8a445f88f9fc1713db34674e79b00e30723e79d/dyld/PrebuiltLoader.cpp#L1660-L1662

I think this is generally safe because hopefully people aren't actually
using the SwiftSyntax classes through the Objective-C runtime, but if
they are we'd still probably prefer to silence the noise and accept the
UB.

Fixes https://github.com/realm/SwiftLint/issues/4782

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -52,10 +52,12 @@ let package = Package(
                 .product(name: "SwiftParser", package: "swift-syntax"),
                 .product(name: "SwiftOperators", package: "swift-syntax"),
                 "SwiftyTextTable",
+                .target(name: "DyldWarningWorkaround", condition: .when(platforms: [.macOS])),
                 "Yams",
                 .product(name: "CryptoSwift", package: "CryptoSwift", condition: .when(platforms: [.linux]))
             ]
         ),
+        .target(name: "DyldWarningWorkaround"),
         .target(
             name: "SwiftLintTestHelpers",
             dependencies: [
--- a/BUILD
+++ b/BUILD
@@ -21,13 +21,13 @@ swift_library(
     module_name = "SwiftLintFramework",
     visibility = ["//visibility:public"],
     deps = [
-        "@com_github_jpsim_sourcekitten//:SourceKittenFramework",
         "@com_github_apple_swift_syntax//:optlibs",
+        "@com_github_jpsim_sourcekitten//:SourceKittenFramework",
         "@sourcekitten_com_github_jpsim_yams//:Yams",
         "@swiftlint_com_github_scottrhoyt_swifty_text_table//:SwiftyTextTable",
     ] + select({
         "@platforms//os:linux": ["@com_github_krzyzanowskim_cryptoswift//:CryptoSwift"],
-        "//conditions:default": [],
+        "//conditions:default": [":DyldWarningWorkaround"],
     }),
 )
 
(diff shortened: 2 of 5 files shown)


Package.swift
```diff
@@ -52,10 +52,12 @@ let package = Package(
                 .product(name: "SwiftParser", package: "swift-syntax"),
                 .product(name: "SwiftOperators", package: "swift-syntax"),
                 "SwiftyTextTable",
+                "DyldWarningWorkaround",
```
jpsim, line 55: Should this be a conditional dependency so it’s not pulled in on Linux?

keith: done

jpsim (approved):
I dont *love* this solution but I don’t have any better ideas. Thanks!

keith:
I think the other option is to link against Xcode's version, but that obviously has the same downside as the previous internal dylib

jpsim:
Yeah and I’d rather have this workaround if it means we control the version of SwiftSyntax we ship with.

jpsim:
Doesn't seem to apply to tests:

```console
$ bazel test --test_output=streamed //Tests:GeneratedTests
WARNING: Streamed test output requested. All tests will be run locally, without sharding, one at a time
INFO: Analyzed target //Tests:GeneratedTests (0 packages loaded, 0 targets configured).
INFO: Found 1 test target...
Test Suite 'All tests' started at 2023-04-18 13:21:47.833
Test Suite 'GeneratedTests.xctest' started at 2023-04-18 13:21:47.834
Test Suite 'AccessibilityLabelForImageRuleGeneratedTests' started at 2023-04-18 13:21:47.834
Test Case '-[GeneratedTests.AccessibilityLabelForImageRuleGeneratedTests testWithDefaultConfiguration]' started.
objc[25929]: Class _TtC11SwiftSyntax16BumpPtrAllocator is implemented in both /private/var/tmp/_bazel_jsimard/7034bb326caf3934edced22f6f4dd368/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/Tests/GeneratedTests.xctest/Contents/MacOS/GeneratedTests (0x1079759c0) and /Applications/Xcode-14.3.0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/host/libSwiftSyntax.dylib (0x10a100a10). One of the two will be used. Which one is undefined.
objc[25929]: Class _TtC11SwiftSyntax35IncrementalParseReusedNodeCollector is implemented in both /private/var/tmp/_bazel_jsimard/7034bb326caf3934edced22f6f4dd368/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/Tests/GeneratedTests.xctest/Contents/MacOS/GeneratedTests (0x107975b78) and /Applications/Xcode-14.3.0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/host/libSwiftSyntax.dylib (0x10a100bc0). One of the two will be used. Which one is undefined.
objc[25929]: Class _TtC11SwiftSyntax26IncrementalParseTransition is implemented in both /private/var/tmp/_bazel_jsimard/7034bb326caf3934edced22f6f4dd368/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/Tests/GeneratedTests.xctest/Contents/MacOS/GeneratedTests (0x107975c10) and /Applications/Xcode-14.3.0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/host/libSwiftSyntax.dylib (0x10a100c58). One of the two will be used. Which one is undefined.
...
```

keith:
Unfortunately it looks like the handling is such that only the main binary is consulted for this:

https://github.com/apple-oss-distributions/dyld/blob/c8a445f88f9fc1713db34674e79b00e30723e79d/dyld/PrebuiltLoader.cpp#L1658-L1660

My assumption is that in this case the main binary is `xctest` instead of our test bundle. Maybe adding a macOS test host would solve that, but that probably wouldn't be worth it :/

PR #4905, opened 2023-04-18
Allow configuring `xct_specific_matcher` with matchers

jpsim:
So that either `one-argument-asserts` or `two-argument-asserts` or both can be enabled.

The following configuration effectively reverts back to the rule behavior prior to https://github.com/realm/SwiftLint/pull/3858:

```yaml
xct_specific_matcher:
  matchers:
    - two-argument-asserts
```

### DIFF
--- a/Source/SwiftLintFramework/Rules/Idiomatic/XCTSpecificMatcherRuleExamples.swift
+++ b/Source/SwiftLintFramework/Rules/Idiomatic/XCTSpecificMatcherRuleExamples.swift
@@ -50,7 +50,13 @@ internal struct XCTSpecificMatcherRuleExamples {
         Example("XCTAssertEqual(foo?.bar, toto())"),
         Example("XCTAssertEqual(foo?.bar, .toto(.zoo))"),
         Example("XCTAssertEqual(toto(), foo?.bar)"),
-        Example("XCTAssertEqual(.toto(.zoo), foo?.bar)")
+        Example("XCTAssertEqual(.toto(.zoo), foo?.bar)"),
+
+        // Configurations Disabled
+        Example("XCTAssertEqual(foo, true)",
+                configuration: ["matchers": ["one-argument-asserts"]]),
+        Example("XCTAssert(foo == bar)",
+                configuration: ["matchers": ["two-argument-asserts"]])
     ]
 
     static let triggeringExamples = [
--- a/Source/SwiftLintFramework/Rules/RuleConfigurations/XCTSpecificMatcherRuleConfiguration.swift
+++ b/Source/SwiftLintFramework/Rules/RuleConfigurations/XCTSpecificMatcherRuleConfiguration.swift
@@ -8,8 +8,8 @@ struct XCTSpecificMatcherRuleConfiguration: SeverityBasedRuleConfiguration, Equa
     }
 
     private enum ConfigurationKey: String {
-        case severity = "severity"
-        case matchers = "matchers"
+        case severity
+        case matchers
     }
 
     var consoleDescription: String {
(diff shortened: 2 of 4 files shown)


jpsim:
Naming suggestions welcome.

Source/SwiftLintFramework/Rules/RuleConfigurations/XCTSpecificMatcherRuleConfiguration.swift
```diff
@@ -0,0 +1,35 @@
+struct XCTSpecificMatcherRuleConfiguration: SeverityBasedRuleConfiguration, Equatable {
+    private(set) var severityConfiguration = SeverityConfiguration(.warning)
+    private(set) var matchers = Set(Matcher.allCases)
+
+    enum Matcher: String, Hashable, CaseIterable {
+        case singleArgument = "single-argument"
+        case doubleArgument = "double-argument"
```
SimplyDanny, line 7: What about `one-argument-asserts` and `two-arguments-asserts` like used in the rule?

SimplyDanny (commented):
We could also think about splitting the rule into two separate rules. But that's semantically not very sensible.

What's the benefit for a user to activate the rule only for one/two arguments though? I guess it's only the incremental activation, right?

jpsim:
> What's the benefit for a user to activate the rule only for one/two arguments though? I guess it's only the incremental activation, right?

Two reasons come to mind:

1. Incremental adoption of the new cases caught by #3858
2. For some codebases (like Lyft's) #3858 might have too many false positives

When I tried #3858 on Lyft's codebase it caught 711 new violations, I tried migrating a few manually (the rule isn't correctable) and 2 out of the 10 I did were false positives in the sense that they couldn't be migrated to `XCTAssertEqual` so I'm assuming I'd probably need a few hundred disable commands to fully enable the `single-argument` cases.

Source/SwiftLintFramework/Rules/RuleConfigurations/XCTSpecificMatcherRuleConfiguration.swift
jpsim: Sounds good.

jpsim: Minor nit: I'll go with `two-argument-asserts` because it represents two-argument asserts and it makes the slugs more similar in formatting (`{one,two}-argument-asserts`).

SimplyDanny:
> When I tried #3858 on Lyft's codebase it caught 711 new violations, I tried migrating a few manually (the rule isn't correctable) and 2 out of the 10 I did were false positives in the sense that they couldn't be migrated to `XCTAssertEqual` so I'm assuming I'd probably need a few hundred disable commands to fully enable the `single-argument` cases.

I see. So even though the rule causes false positives in both cases, compile errors seem to be more likely for the "one argument" variant. That's a good argument for the configuration.

SimplyDanny (approved):

SimplyDanny:
Should we have tests for the new options?

Source/SwiftLintFramework/Rules/Idiomatic/XCTSpecificMatcherRuleExamples.swift
```diff
@@ -50,7 +50,13 @@ internal struct XCTSpecificMatcherRuleExamples {
...
         Example("XCTAssertEqual(foo?.bar, .toto(.zoo))"),
         Example("XCTAssertEqual(toto(), foo?.bar)"),
-        Example("XCTAssertEqual(.toto(.zoo), foo?.bar)")
+        Example("XCTAssertEqual(.toto(.zoo), foo?.bar)"),
+
+        // Configurations Disabled
+        Example("XCTAssertEqual(foo, true)",
+                configuration: ["matchers": ["one-argument-asserts"]]),
+        Example("XCTAssert(foo == bar)",
+                configuration: ["matchers": ["two-argument-asserts"]])
```
SimplyDanny, line 59: Should better be excluded from documentation. Otherwise it's confusing since we don't see the configuration in the rule directory.

PR #4908, opened 2023-04-18
Baseline support, a.k.a. disabling rules based on file name regexps

esilverberg:
This is another approach to implementing the "baseline" functionality that is introduced in this [old PR from 2020](https://github.com/realm/SwiftLint/pull/3421).

To use it, you simply add to `.swiftlint.yml` the following section:

```
disabled_rules_for_files:
  rule_name:
    - filename_regexp
```

Where `rule_name` is the name of the rule, and `filename_regexp` is a regular expression evaluated against a filename to determine if the rule should be skipped on the given file.

In the last comment regarding this kind of functionality from [this old PR](https://github.com/realm/SwiftLint/pull/3421), this was written:

> Looking at this again for the first time in a while, there's a few things I'd like the author(s) to think through before we can consider something like this, which I think overall the concept is a good idea.
> 
> 1. Should users be able to specify the baseline file path?
> 2. Should baselines apply to analyzer rules, or just lint rules?
> 3. Right now the baseline implementation is fairly brittle in the sense that if the line numbers change, the baseline record becomes out of date. Do we want to consider a mechanism to minimize the thrash when say a violation is moved down by a few lines because new code was added above it?
> 4. Should there be an easy way to reset the baseline? Right now you have to find the baseline file and delete it.
> 5. Can we somehow leverage the lint results cache? We're already saving lint results for faster incremental linting operations, so perhaps snapshotting that would be a new command distinct from swiftlint lint?
> 6. Unit tests, documentation and changelog entries would be important to add before considering merging something like this.

I don't think questions 2-6 apply to my new PR, but I'm happy to answer/respond to any additional feedback you may have! Would love to see functionality like this incorporated 🙏


SimplyDanny:
Sorry that this PR has been completely ignored. Meanwhile #5475 has superseded this PR. With that being merged, we now have a baseline functionality experimentally available in SwiftLint. It's different from your approach, but the results ought to be the same.

Thank you for your efforts anyway!

PR #4911, opened 2023-04-20
Add new `redundant_self_in_closure` rule

SimplyDanny:
This adds the new (and long awaited by #59) `redundant_self_in_closure` rule that triggers in closures on explicitly used `self` when it's actually not needed due to:

* Strongly captured `self` (`{ [self] in ... }`) according to [SE-0269](https://github.com/apple/swift-evolution/blob/main/proposals/0269-implicit-self-explicit-capture.md)
* Closure used in a struct declaration (`self` can always be omitted)
* Anonymous closures that are directly called (`{ ... }()`) as they are
    definitly not escaping
* Weakly captured `self` with explicit unwrapping according to [SE-0365](https://github.com/apple/swift-evolution/blob/main/proposals/0365-implicit-self-weak-capture.md)

What's missing is the detection of redundant `self` in classes when the closure isn't `@escaping`. This case can only be resolved by a separate Analyzer rule that has full knowledge about the closure's type. It's not possible to implement that reliably on bare source code level only.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -40,6 +40,7 @@ disabled_rules:
   - prefer_nimble
   - prefer_self_in_static_references
   - prefixed_toplevel_constant
+  - redundant_self_in_closure
   - required_deinit
   - self_binding
   - sorted_enum_cases
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
   * Closure used in a struct declaration (`self` can always be omitted)
   * Anonymous closures that are directly called (`{ ... }()`) as they are
     definitly not escaping
+  * Weakly captured `self` with explicit unwrapping
 
   [SimplyDanny](https://github.com/SimplyDanny)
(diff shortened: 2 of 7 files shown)


SimplyDanny:
There is a case when this rule causes false positives as can be seen [here](https://github.com/brave/brave-ios/blob/a04229d62e448b671dadfdff991a27bbf0d9000a/Sources/Brave/Frontend/Settings/Brave%20Search/BraveSearchDebugMenu.swift#L64:7). A simple example would be
```swift
struct S {
  var x = 1
  func f() {
    acceptClosure {
      let x = 2
      self.x = x // rule triggers here
    }
  }
}
```
The rule would trigger on `self` which is not correct here due the other `x` in scope.

jpsim:
Regarding the false positives described in https://github.com/realm/SwiftLint/pull/4911#issuecomment-1516122706, solving that fully would be impossible without rebuilding the full Swift compiler, since `x` could also be defined as a global variable somewhere far away 😅 

Don't sweat it.

PR #4913, opened 2023-04-21
Correct organization name in README files

enebin:
raywenderlich.com changed their name to Kodeco so I changed it accordingly.


SimplyDanny (approved):

SimplyDanny:
Thanks!

PR #4915, opened 2023-04-21
Move CocoaPods CI job to Buildkite

jpsim:
Should run more quickly with bundler and CocoaPods caches.


jpsim:
I'd need to install tvOS simulators on the CI machines for this to work.

PR #4900, opened 2023-04-17
[bazel] Remove custom SwiftSyntax BUILD file

keith:
Something similar to this has been merged upstream instead now. This
also renames the repo name to SwiftSyntax in preparation for it being in
the BCR

PR #4920, opened 2023-04-22
[SwiftSyntax] Update to latest `release/5.9` commit

jpsim:
Pulls in changes from https://github.com/apple/swift-syntax/pull/1556.

Also includes changes from https://github.com/realm/SwiftLint/pull/4900 to use the upstream SwiftSyntax BUILD files.

Can't merge until a new `509.0.0-swift-5.9-DEVELOPMENT-SNAPSHOT-2023-04-...` development snapshot tag is published.


SimplyDanny:
What's wrong with SourceKitten by the way? New findings and fixes have appeared in all builds for a few days.

jpsim:
I don't know. There might be some non-determinism in SwiftLint that's surfaced more when linting SourceKitten or maybe some CI machine has a bad checkout? A bad checkout would impact both `main` and PR branches though so I don't think that's it.

Issue #4923, opened 2023-04-23
Some SwiftLint warning messages are not prepended with `warning: `, and do not show up in Xcode


mildm8nnered:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

Consider the following configuration file:

```
allow_zero_lintable_files: true

implicit_return:
  included:
    - closure

custom_rules:
  no_system_loggers:
    included: ".*.swift"
    name: "Avoid system loggers"
    message: "Use something else"
#    regex: ( print|NSLog)
    severity: warning
```

When run with, for example

`% swiftlint lint --config broken.yml  /tmp/`

This will produce the following output:

```
Invalid configuration for custom rule 'no_system_loggers'.
warning: Found a configuration for 'implicit_return' rule, but it is not enabled on 'opt_in_rules'.
Linting Swift files at paths /tmp/
Done linting! Found 0 violations, 0 serious in 0 files.
```

When running within Xcode, the message about the `"Found a configuration for 'implicit_return' rule, but it is not enabled on 'opt_in_rules'.` is prepended with `warning:`, Xcode will display that message as a warning.

However the `Invalid configuration for custom rule 'no_system_loggers'.` message will not be displayed.

This, and other similar messages should be prepended with `warning: `, so that they also appear in Xcode.

##### Complete output when running SwiftLint, including the stack trace and command used

See above

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? `0.51.0`
* Installation method used (Homebrew, CocoaPods, building from source, etc)? Homebrew
* Paste your configuration file:

See above

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? No

* Which Xcode version are you using (check `xcodebuild -version`)?

```
Xcode 14.2
Build version 14C18
```

* Do you have a sample that shows the issue? 

See above.

PR #4927, opened 2023-04-24
Prepend `warning: ` to error messages so that they show in Xcode

whiteio:
Some error messages aren't being shown in Xcode, `warning: ` should be prepended to these warning messages so that they will show in Xcode.

Fixes https://github.com/realm/SwiftLint/issues/4923


mildm8nnered:
These ones definitely look good (and are the only "Invalid ..." ones). You beat me to it, lol.

I think there are some more that could arguably use the same treatment - for example

`queuedPrintError("Could not write to file at path \(outFile)")` in LintOrAnalyzeCommand.swift

SimplyDanny (approved):
LGTM. Thank you, @whiteio!

PR #4928, opened 2023-04-24
Adds test verifying that the file_header rule can require an empty file header

simonbs:
This PR adds a test to the FileHeaderRuleTests to verify that the `file_header` rule can be used to require an empty file header using `forbidden_pattern: .`.

For the longest time I have been wanting a SwiftLint rule to require empty file headers and only now did I realize that I can use the `file_header` rule and it's `forbidden_pattern` to achieve this. I thought I wanted to add a unit test to document this behavior and ensure that it doesn't break in future updates.

The two tests ensures that:

1. The configuration `forbidden_pattern: .` detects non-empty file headers.
2. Documentation for a type does not count as a file header.


SimplyDanny (approved):
A very helpful and valuable addition. Thank you!

PR #4929, opened 2023-04-24
Switch Case Sortation

atacan:
Correctable rule that sorts the cases of swift alphabetically.

```swift
  Example("""
  ↓switch foo {
  case "c":
      break
  case "a":
      break
  case "b":
      break
  }
  """)
```
```swift
// will be
  Example("""
  switch foo {
  case "a":
      break
  case "b":
      break
  case "c":
      break
  }
  """)
```

### DIFF
--- a/Source/SwiftLintFramework/Rules/Style/SwitchCaseSortRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/SwitchCaseSortRule.swift
@@ -317,7 +317,7 @@ private let examples: (triggering: [Example], nonTriggering: [Example], correcti
             break
         case .a:
             break
-        case default:
+        default:
             break
         }
         """)
@@ -329,17 +329,18 @@ private let examples: (triggering: [Example], nonTriggering: [Example], correcti
             break
         case .d:
             break
-        case default:
+        default:
             break
         }
         """)
     )
 //    triggering.append(
+//        // this is a compiler error, so need to handle it: Additional 'case' blocks cannot appear after the 'default' block of a 'switch'
 //        Example("""
 //        ↓switch foo {
 //        case .d:
 //            break
-//        case default:
+//        default:
 //            break
 //        case .a:
 //            break
@@ -353,7 +354,7 @@ private let examples: (triggering: [Example], nonTriggering: [Example], correcti
 //            break
 //        case .d:
 //            break
-//        case default:
+//        default:
 //            break
 //        }
 //        """)
--- a/Source/SwiftLintFramework/Rules/Style/SwitchCaseSortRule.swift
+++ b/Source/SwiftLintFramework/Rules/Style/SwitchCaseSortRule.swift
@@ -29,55 +29,6 @@ struct SwitchCaseSortRule: ConfigurationProviderRule, SwiftSyntaxCorrectableRule
 }
 
 private extension SwitchCaseSortRule {
-//    final class Visitor: ViolationsSyntaxVisitor {
-//        override func visitPost(_ node: SwitchExprSyntax) {
-//            guard node.cases.isNotEmpty else {
-//                return
-//            }
-//
-//            var caseNamesToSort = [String]()
-//
-//            for caseListSyntax in node.cases {
-//                switch caseListSyntax {
-//                case let .switchCase(caseSyntax):
-//                    switch caseSyntax.label {
-//                    case let .default(defaultLabelSyntax):
-//                        // xcode already warns if default is not at the end?
-//                        continue
-//                    case let .case(caseLabelSyntax):
-//                        if caseLabelSyntax.caseItems.isEmpty {
-//                            continue
-//                        }
-//                        if caseLabelSyntax.caseItems.count == 1 {
-//                            // get the name for top level sortation
-//                            let name = sortableName(for: caseLabelSyntax.caseItems.first!)
-//                            caseNamesToSort.append(name)
-//                        } else { // multiple items in one case expression
-//                            // sort them among themselves
-//                            var caseNames = [String]()
-//                            for caseItem in caseLabelSyntax.caseItems {
-//                                let name = sortableName(for: caseItem)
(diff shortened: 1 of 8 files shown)


Source/SwiftLintFramework/Rules/Style/SwitchCaseSortRule.swift
```diff
@@ -0,0 +1,243 @@
...
+    case (.default, _):
+        return false
+    case (_, .default):
+        return true
+    case let (.case(lhsCaseLabel), .case(rhsCaseLabel)):
+        return byName(lhsCaseLabel, rhsCaseLabel)
+    }
+}
+
+private func byName(_ lhs: SwitchCaseListSyntax.Element, _ rhs: SwitchCaseListSyntax.Element) -> Bool {
```
SimplyDanny, line 203: What about extensions for these types that implement `Comparable` instead of these freestanding functions used for sorting?

The same extensions could contain the `sortableName` methods/accessors.

`sortableName` could also be extracted by another `SyntaxTransformVisitor` that would then contain all this logic.

Freestanding functions appear without context. They look like arbitrary functions but actually here a specific extraction logic is implemented for all these types. So the functions all belong together. But the don't as each of them stands independent of the others.

Source/SwiftLintFramework/Rules/Style/SwitchCaseSortRule.swift
```diff
@@ -0,0 +1,243 @@
...
+ > case `ifConfigDecl`(IfConfigDeclSyntax) // `case` expression is inside #if...#endif
+ >> IfConfigDeclSyntax
+ >>> \.clauses: IfConfigClauseListSyntax
+ >>>> IfConfigClauseSyntax
+ >>>>> case `statements`(CodeBlockItemListSyntax)
+ >>>>> case `switchCases`(SwitchCaseListSyntax)
+ >>>>> case `decls`(MemberDeclListSyntax)
+ >>>>> case `postfixExpression`(ExprSyntax)
+ >>>>> case `attributes`(AttributeListSyntax)
+ */
```
SimplyDanny, line 243: Should be removed.

Source/SwiftLintFramework/Rules/Style/SwitchCaseSortRuleExamples.swift
```diff
@@ -0,0 +1,340 @@
...
+        //        switch foo {
+        //        case .a:
+        //            break
+        //        case .d:
+        //            break
+        //        default:
+        //            break
+        //        }
+        //        """)
+        //    )
```
SimplyDanny, line 207: Should be removed or enabled.

Source/SwiftLintFramework/Rules/Style/SwitchCaseSortRuleExamples.swift
```diff
@@ -0,0 +1,340 @@
+enum SwitchCaseSortRuleExamples {
+    // to see the triggering and correction subsequently
+    static let examples: (triggering: [Example], nonTriggering: [Example], corrections: [Example: Example]) = {
```
SimplyDanny, line 3: I'd like to see examples where the `switch` statement is contained (indented) in a function. Also, some comments and whitespace should be used in cases to proof that the rewriter doesn't break or remove them.

Source/SwiftLintFramework/Rules/Style/SwitchCaseSortRule.swift
```diff
@@ -0,0 +1,243 @@
...
+        } else { // default or single cases
+            elements.append(caseListElement)
+        }
+    }
+
+    return SwitchCaseListSyntax(elements.sorted(by: byName))
+}
+
+private func sort(_ caseLabel: SwitchCaseLabelSyntax) -> SwitchCaseLabelSyntax {
+    // remove trailing commas, then sort them, then join them with commas
```
SimplyDanny, line 115: The comments are helpful. However, we like to use full sentences ending with a period.

Source/SwiftLintFramework/Rules/Style/SwitchCaseSortRule.swift
```diff
@@ -0,0 +1,243 @@
...
+        .sorted(by: byName)
+        .enumerated()
+        .map { index, item in
+            index == caseLabel.caseItems.count - 1 ? item : item.with(\.trailingComma, ", ")
+        }
+    return caseLabel.with(\.caseItems, CaseItemListSyntax(sortedItems))
+}
+
+private func sortableName(for expressionPattern: ExpressionPatternSyntax) -> String? {
+    if let memberAccessExpression = expressionPattern.expression
```
SimplyDanny, line 127: Could be implemented as a `switch` statement with type-casting patterns. Or see my other comment mentioning `SyntaxTransformVisitor`.

Source/SwiftLintFramework/Rules/Style/SwitchCaseSortRule.swift
```diff
@@ -0,0 +1,243 @@
...
+    } else if let floatExpression = expressionPattern.expression
+        .as(FloatLiteralExprSyntax.self) {
+        return floatExpression.floatingDigits.text
+    } else if let identifierExpression = expressionPattern.expression
+        .as(IdentifierExprSyntax.self) {
+        return identifierExpression.identifier.text
+    } else if let functionCall = expressionPattern.expression.as(FunctionCallExprSyntax.self),
+              let memberAccess = functionCall.calledExpression.as(MemberAccessExprSyntax.self) {
+        return memberAccess.name.text
+    }
```
SimplyDanny, line 154: What about type-casting, tuple and wildcard patterns?

SimplyDanny (changes requested):
Thank you for this contribution!

I'm not yet so convinced that this rule is going to be very useful despite for very simple `switch` statements that wouldn't appear too often.

However, I've just added some review comments you might already want to take into account.

Issue #4932, opened 2023-04-25
Xcode 14.3 Support - Set Higher Minimum OS Versions


LowAmmo:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

The issue is with publishing a cocoapod that depends on SwiftLint with Xcode 14.3. Looks like Xcode 14.3 is requiring the minimum iOS version to be 11.

Should be able to just set the minimum version of SwiftLint to at least 11.0 (I'd suggest maybe 12.0 or 14.0, as 11.0 is already 5+ years old).

```bash
warning: The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 9.0, but the range of supported deployment target versions is 11.0 to 16.4.99. (in target 'SwiftLint' from project 'Pods')
```

##### Complete output when running SwiftLint, including the stack trace and command used
n/a

### Environment
Xcode 14.3

jpsim:
Why do you want to publish a CocoaPod that depends on SwfitLint?

The SwiftLint CocoaPod isn't a "real" CocoaPod in the sense that it's not meant to be packaged in apps built with CocoaPods, but rather a hack to allow users who prefer managing SwiftLint via CocoaPods to do so.

jpsim:
> But, I'm guessing that if you try to update the podspec to the cocoapod trunk repo, that you'll end up running into the same issue... Just need to update the minimum versions across the board...

I can still target those old versions if I publish the pod with Xcode 14.2 selected.

LowAmmo:
@jpsim  - A...good point... Yeah, as long as you delay moving to Xcode 14.3, you can keep pushing with the lower versions.

But, would think it makes sense to just go ahead and update to be able to fully support Xcode 14.3 sooner rather, than later... (I doubt Apple is going to change their mind and have Xcode 14.4 go back to supporting back to iOS 9, etc...)

jpsim:
Yeah it's fine to update now, can you make a PR bumping the minimum versions as of Xcode 14.3?

LowAmmo:
Sure, I'll get a PR bumping up...  Guessing you would rather keep versions at the bare minimum versus going up to like 15.5 or 14.5??

jpsim:
[...]

PR #4934, opened 2023-04-25
Update minimum OS versions for Xcode 14.3 (#4932)

LowAmmo:
* iOS - 11.0
* MacOS - 10.13
* tvOS - 11.0
* WatchOS - 7.0

This is to address Issue #4932


jpsim (approved):
Thanks!

Issue #4810, opened 2023-03-11
Feature request: lexicographic sorting options for `sorted_imports`


mildm8nnered:
### New Issue Checklist

- [X] Updated SwiftLint to the latest version
- [X] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

```
import AppTypes
import AVFoundation
import Base
import Logging
```

Will pass `sorted_imports`, but `swift-format`'s equivalent of this rule wants lexicographic sorting, and expects

```
import AppTypes
import Base
import Logging
import AVFoundation
```

It would be nice if SwiftLint offered a configuration option that preferred a sort order compatible with `swift-format`

##### Complete output when running SwiftLint, including the stack trace and command used

N/A

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.50.3
* Installation method used (Homebrew, CocoaPods, building from source, etc)? Homebrew
* Paste your configuration file:

none

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)?
  If so, paste their relative paths and respective contents.

no

* Which Xcode version are you using (check `xcodebuild -version`)?

```
Xcode 14.2
Build version 14C18
```

* Do you have a sample that shows the issue? Run `echo "[string here]" | swiftlint lint --no-cache --use-stdin --enable-all-rules`
  to quickly test if your example is really demonstrating the issue. If your example is more
  complex, you can use `swiftlint lint --path [file here] --no-cache --enable-all-rules`.

N/A

PR #4935, opened 2023-04-25
Add `grouping` option to the `sorted_imports` rule

hiltonc:
Adds a `grouping` option to the `sorted_import` rule. With this configuration:

```
sorted_imports:
  grouping: attributes
```

It puts all the `@testable` imports at the top of the import groups.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -183,7 +183,6 @@
   to sort groups of imports defined by their preceding attributes
   (e.g. `@testable`, `@_exported`, ...).  
   [hiltonc](https://github.com/hiltonc)
-  [#4935](https://github.com/realm/SwiftLint/pull/4935)
 
 #### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#4990](https://github.com/realm/SwiftLint/issues/4990)
 
+* Add `grouping` option to the `sorted_imports` rule allowing
+  to sort groups of imports defined by their preceding attributes
+  (e.g. `@testable`, `@_exported`, ...).  
+  [hiltonc](https://github.com/hiltonc)
+
 #### Bug Fixes
 
 * Do not trigger `prefer_self_in_static_references` rule on `typealias`
@@ -178,11 +183,6 @@
   This results in errors being silently discarded, which may be unexpected.
   See this forum thread for more details: https://forums.swift.org/t/56066  
   [kylebshr](https://github.com/kylebshr)
-* Adds `grouping` option to the `sorted_imports` rule.  
-* Add `grouping` option to the `sorted_imports` rule allowing
-  to sort groups of imports defined by their preceding attributes
-  (e.g. `@testable`, `@_exported`, ...).  
-  [hiltonc](https://github.com/hiltonc)
 
 #### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -179,6 +179,9 @@
   See this forum thread for more details: https://forums.swift.org/t/56066  
   [kylebshr](https://github.com/kylebshr)
 * Adds `grouping` option to the `sorted_imports` rule.  
+* Add `grouping` option to the `sorted_imports` rule allowing
+  to sort groups of imports defined by their preceding attributes
+  (e.g. `@testable`, `@_exported`, ...).  
   [hiltonc](https://github.com/hiltonc)
   [#4935](https://github.com/realm/SwiftLint/pull/4935)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -178,6 +178,9 @@
   This results in errors being silently discarded, which may be unexpected.
   See this forum thread for more details: https://forums.swift.org/t/56066  
   [kylebshr](https://github.com/kylebshr)
+* Adds `grouping` option to the `sorted_imports` rule.  
+  [hiltonc](https://github.com/hiltonc)
+  [#4935](https://github.com/realm/SwiftLint/pull/4935)
 
 #### Bug Fixes
 
--- a/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/SortedImportsConfiguration.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/SortedImportsConfiguration.swift
@@ -1,0 +1,42 @@
+
+struct SortedImportsConfiguration: RuleConfiguration, Equatable {
+    typealias Parent = SortedImportsRule
+
+    enum SortedImportsGroupingConfiguration: String {
+        /// Sorts import lines based on any import attributes (e.g. `@testable`, `@_exported`, etc.), followed by a case
+        /// insensitive comparison of the imported module name.
+        case attributes
+        /// Sorts import lines based on a case insensitive comparison of the imported module name.
+        case name
+
+        init(value: Any) throws {
+            if let string = value as? String,
+               let value = Self(rawValue: string) {
+                self = value
+            } else {
+                throw Issue.unknownConfiguration(ruleID: Parent.identifier)
+            }
+        }
+    }
+
+    private(set) var severity = SeverityConfiguration<Parent>(.warning)
+    private(set) var grouping = SortedImportsGroupingConfiguration.name
+
+    var consoleDescription: String {
+        return "severity: \(severity.consoleDescription)"
+            + ", grouping: \(grouping)"
+    }
+
+    mutating func apply(configuration: Any) throws {
+        guard let configuration = configuration as? [String: Any] else {
+            throw Issue.unknownConfiguration(ruleID: Parent.identifier)
+        }
+
+        if let grouping = configuration["grouping"] as? String {
+            self.grouping = try SortedImportsGroupingConfiguration(value: grouping)
+        }
+        if let severityString = configuration["severity"] as? String {
+            try severity.apply(configuration: severityString)
+        }
+    }
+}
--- a/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/SortedImportsConfiguration.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/SortedImportsConfiguration.swift
@@ -1,4 +1,3 @@
-
 struct SortedImportsConfiguration: RuleConfiguration, Equatable {
     typealias Parent = SortedImportsRule
 
@@ -7,12 +6,12 @@ struct SortedImportsConfiguration: RuleConfiguration, Equatable {
         /// insensitive comparison of the imported module name.
         case attributes
         /// Sorts import lines based on a case insensitive comparison of the imported module name.
-        case name
+        case names
 
         init(value: Any) throws {
             if let string = value as? String,
                let value = Self(rawValue: string) {
-                self = value
+               self = value
             } else {
                 throw Issue.unknownConfiguration(ruleID: Parent.identifier)
             }
@@ -20,7 +19,7 @@ struct SortedImportsConfiguration: RuleConfiguration, Equatable {
     }
 
     private(set) var severity = SeverityConfiguration<Parent>(.warning)
-    private(set) var grouping = SortedImportsGroupingConfiguration.name
+    private(set) var grouping = SortedImportsGroupingConfiguration.names
 
     var consoleDescription: String {
         return "severity: \(severity.consoleDescription)"
--- a/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/SortedImportsConfiguration.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/SortedImportsConfiguration.swift
@@ -22,13 +22,14 @@ struct SortedImportsConfiguration: RuleConfiguration, Equatable {
             throw Issue.unknownConfiguration(ruleID: Parent.identifier)
         }
 
-        if let rawGrouping = configuration["grouping"] as? String {
-            if let grouping = SortedImportsGroupingConfiguration(rawValue: rawGrouping) {
-                self.grouping = grouping
-            } else {
+        if let rawGrouping = configuration["grouping"] {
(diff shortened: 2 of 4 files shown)


mildm8nnered:
So `swift-format`'s `OrderedImports` rule takes the view that regular imports should precede testable imports.

It also sorts imports slightly differently as well - https://github.com/realm/SwiftLint/issues/4810

Not sure what the right answer is here, but it's a shame that we can't have some compatibility so that people can use both together, without the linters fighting over which is the correct order.

hiltonc:
@mildm8nnered There’s definitely room for different import sort strategies. I’m open to changing the naming to allow for adding a swift-format compatible strategy down the line.

mildm8nnered:
> @mildm8nnered There’s definitely room for different import sort strategies. I’m open to changing the naming to allow for adding a swift-format compatible strategy down the line.
> 
> Maybe `grouping` -> `strategy` and `attributes` -> `testableFirst`.

I think I wouldn't worry about it too much for now. I don't think that many people actually run `swift-format` linting anyway, but it is just kind of annoying that they have different conventions, or at least, cannot be configured to be compatible, although `swift-format` offers very little in the way of configurability.

CHANGELOG.md
```diff
@@ -44,6 +44,10 @@
   [AndrewDMontgomery](https://github.com/andrewdmontgomery)
   [#4875](https://github.com/realm/SwiftLint/pull/4875)
 
+* Adds `grouping` option to the `sorted_imports` rule.
```
SimplyDanny, line 47:
```suggestion
* Add `grouping` option to the `sorted_imports` rule allowing
  to sort groups of imports defined by their preceding attributes
  (e.g. `@testable`, `@_exported`, ...).  
```

CHANGELOG.md
```diff
@@ -44,6 +44,10 @@
   [AndrewDMontgomery](https://github.com/andrewdmontgomery)
   [#4875](https://github.com/realm/SwiftLint/pull/4875)
 
+* Adds `grouping` option to the `sorted_imports` rule.  
+  [hiltonc](https://github.com/hiltonc)
+  [#4935](https://github.com/realm/SwiftLint/pull/4935)
```
SimplyDanny, line 49: This should be an issue that is addressed by this PR. If there isn't any, we omit this line.

Source/SwiftLintFramework/Rules/RuleConfigurations/SortedImportsConfiguration.swift
```diff
@@ -0,0 +1,43 @@
+enum SortedImportsGroupingConfiguration: String {
+    /// Sorts import lines based on any import attributes (e.g. `@testable`, `@_exported`, etc.), followed by a case
+    /// insensitive comparison of the imported module name.
+    case attributes
+    /// Sorts import lines based on a case insensitive comparison of the imported module name.
+    case name
+
+    init(value: Any) throws {
```
SimplyDanny, line 8: This initializer isn't needed. You may just call `SortedImportsGroupingConfiguration(rawValue:)`.

Source/SwiftLintFramework/Rules/RuleConfigurations/SortedImportsConfiguration.swift
```diff
@@ -0,0 +1,43 @@
+enum SortedImportsGroupingConfiguration: String {
+    /// Sorts import lines based on any import attributes (e.g. `@testable`, `@_exported`, etc.), followed by a case
+    /// insensitive comparison of the imported module name.
+    case attributes
+    /// Sorts import lines based on a case insensitive comparison of the imported module name.
+    case name
```
SimplyDanny, line 6:
```suggestion
    case names
```
for consistency.

Source/SwiftLintFramework/Rules/RuleConfigurations/SortedImportsConfiguration.swift
```diff
@@ -0,0 +1,43 @@
...
+struct SortedImportsConfiguration: RuleConfiguration, Equatable {
+    private(set) var severity = SeverityConfiguration(.warning)
+    private(set) var grouping: SortedImportsGroupingConfiguration
+
+    var consoleDescription: String {
+        return "severity: \(severity.consoleDescription)"
+            + ", grouping: \(grouping)"
+    }
+
+    init(grouping: SortedImportsGroupingConfiguration) {
```
SimplyDanny, line 27: Not needed when initializing the field in line 20 with the default value.

Source/SwiftLintFramework/Rules/Style/SortedImportsRule.swift
```diff
@@ -1,30 +1,36 @@
...
-    fileprivate var contentRange: NSRange {
+fileprivate extension Line {
+    var contentRange: NSRange {
         return NSRange(location: range.location, length: content.bridge().length)
     }
 
+    // `Line` in this rule always contains word import
+    // This method returns contents of line that are before import
+    func importAttributes() -> String {
+        return content[importAttributesRange()].trimmingCharacters(in: .whitespaces)
```
SimplyDanny, line 12: The attributes don't have to be on the same line with the import statement. They can also come in the lines above. Also, there can be multiple attributes for a single import.

Source/SwiftLintFramework/Rules/Style/SortedImportsRule.swift
```diff
@@ -47,96 +53,25 @@ private extension Sequence where Element == Line {
 }
 
 struct SortedImportsRule: CorrectableRule, ConfigurationProviderRule, OptInRule {
-    var configuration = SeverityConfiguration(.warning)
+    var configuration = SortedImportsConfiguration(grouping: .name)
 
+    private static let examples = SortedImportsRuleExamples.self
```
SimplyDanny, line 58: Please inline this variable, i.e. use `SortedImportsRuleExamples.nonTriggeringExamples`, ...

Source/SwiftLintFramework/Rules/Style/SortedImportsRule.swift
```diff
@@ -173,6 +108,24 @@ struct SortedImportsRule: CorrectableRule, ConfigurationProviderRule, OptInRule
...
     }
 
+    /// - returns: whether `lhs` should come before `rhs` based on a comparison of the contents of the import lines
+    private func should(_ lhs: Line, comeBefore rhs: Line) -> Bool {
+        switch configuration.grouping {
+        case .attributes:
+            let lhsAttributes = lhs.importAttributes()
+            let rhsAttributes = rhs.importAttributes()
+            if lhsAttributes != rhsAttributes {
+                if lhsAttributes.isEmpty != rhsAttributes.isEmpty {
```
SimplyDanny, line 118: It's enough to check for `rhsAttributes.isEmpty` and, if so, return `true`. In other words, the whole `if` statement can be reduced to `return rhsAttributes.isEmpty || lhsAttributes < rhsAttributes`.

Source/SwiftLintFramework/Rules/Style/SortedImportsRule.swift
```diff
@@ -173,6 +108,24 @@ struct SortedImportsRule: CorrectableRule, ConfigurationProviderRule, OptInRule
...
+                    return rhsAttributes.isEmpty
+                }
+                return lhsAttributes < rhsAttributes
+            }
+            return lhs.importModule().lowercased() <= rhs.importModule().lowercased()
+        case .name:
+            return lhs.importModule().lowercased() <= rhs.importModule().lowercased()
```
SimplyDanny, line 125: This line is duplicated and should therefore be moved outside the `switch` statement.

Source/SwiftLintFramework/Rules/Style/SortedImportsRuleExamples.swift
```diff
@@ -0,0 +1,96 @@
...
+        #else
+        import CCC
+        import DDD.B
+        #endif
+        import AAA
+        import BBB
+        """),
+        Example("@testable import AAA\n  @testable import BBB", configuration: groupByAttributesConfiguration),
+        Example("@testable import BBB\n  import AAA", configuration: groupByAttributesConfiguration),
+        Example("@_exported import BBB\n  @testable import AAA", configuration: groupByAttributesConfiguration)
```
SimplyDanny, line 26: Examples with non-default configurations should be excluded from the rule's documentation as the configuration isn't shown in the rendered example on the website. There is the flag `excludeFromDocumentation` for this case.

SimplyDanny (changes requested):
Thank you for the contribution! I have a few remarks.

The most important point is that attributes do not necessarily have to stand on the same line like the import statement.

Source/SwiftLintFramework/Rules/RuleConfigurations/SortedImportsConfiguration.swift
hiltonc: I would still need to handle an unknown value for `grouping`, either ignoring it or throwing:

```
        if let rawGrouping = configuration["grouping"] as? String,
           let grouping = SortedImportsGroupingConfiguration(rawValue: rawGrouping) {
            self.grouping = grouping
        }
```

or

```
        if let rawGrouping = configuration["grouping"] as? String {
            if let grouping = SortedImportsGroupingConfiguration(rawValue: rawGrouping) {
                self.grouping = grouping
            } else {
                throw ConfigurationError.unknownConfiguration
            }
        }
```

Any preference?

Source/SwiftLintFramework/Rules/Style/SortedImportsRule.swift
hiltonc: Doing this fails the `  import AAA\n@testable import ↓BBB` test:

```swift
            let lhsAttributes = lhs.importAttributes()
            let rhsAttributes = rhs.importAttributes()
            if lhsAttributes != rhsAttributes {
                return rhsAttributes.isEmpty || lhsAttributes < rhsAttributes
            }
```

`rhsAttributes` is `@testable`, which is not empty, and `"" < "@testable"`. So it returns `true`. But we actually want it to return `false` to put attributes first.

Source/SwiftLintFramework/Rules/Style/SortedImportsRule.swift
hiltonc: I just pushed a commit with test cases for multiple attributes on the same line, seems to work fine.

That's a great point about declaration over multiple lines though. Oof. It does appear to be an existing issue, with the default config we get this failure and correction:

```
        Example("@testable\nimport BBB\nimport ↓AAA"):
            Example("@testable\nimport AAA\nimport BBB"),
```

Do you consider this a blocking issue?

Source/SwiftLintFramework/Rules/RuleConfigurations/SortedImportsConfiguration.swift
SimplyDanny: I prefer an error for an unknown value (option 2).

Source/SwiftLintBuiltInRules/Rules/Style/SortedImportsRuleExamples.swift
```diff
@@ -0,0 +1,157 @@
...
+            import CCC
+            import DDD.B
+            #endif
+            import AAA
+            import BBB
+            """),
+        Example("""
+          @testable import BBB
+        @testable import ↓AAA
+        """, configuration: groupByAttributesConfiguration, excludeFromDocumentation: true):
```
SimplyDanny, line 119: Corrections are not contained in the documentation, so you may omit the `excludeFromDocumentation` flag.

Source/SwiftLintFramework/Rules/Style/SortedImportsRule.swift
SimplyDanny: Got it. 👍

Source/SwiftLintFramework/Rules/Style/SortedImportsRule.swift
SimplyDanny: Since it's an existing issue it shouldn't block your contribution. However, the issue can probably only be fixed by basing the rule on SwiftSyntax. With all the available test cases this is quite a fun thing to do. So in case you are interested ... another contribution is highly welcomed. 😉

Source/SwiftLintBuiltInRules/Rules/Style/SortedImportsRuleExamples.swift
```diff
@@ -0,0 +1,157 @@
+internal struct SortedImportsRuleExamples {
+    private static let groupByAttributesConfiguration = ["grouping": "attributes"]
+
+    static let nonTriggeringExamples = [
+        Example("import AAA\nimport BBB\nimport CCC\nimport DDD"),
+        Example("import Alamofire\nimport API"),
+        Example("import labc\nimport Ldef"),
+        Example("import BBB\n// comment\nimport AAA\nimport CCC"),
+        Example("@testable import AAA\nimport   CCC"),
+        Example("import AAA\n@testable import   CCC"),
```
SimplyDanny, line 10: Would be easier if these examples (and the others below) would use text blocks too.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/SortedImportsConfiguration.swift
```diff
@@ -29,8 +20,12 @@ struct SortedImportsConfiguration: RuleConfiguration, Equatable {
             throw ConfigurationError.unknownConfiguration
         }
 
-        if let grouping = configuration["grouping"] as? String {
-            self.grouping = try SortedImportsGroupingConfiguration(value: grouping)
+        if let rawGrouping = configuration["grouping"] as? String {
```
SimplyDanny, line 23: If value for `grouping` is not a `String` there should also be an exception.

Source/SwiftLintBuiltInRules/Rules/Style/SortedImportsRuleExamples.swift
SimplyDanny: Thank you for reformatting this!

SimplyDanny (approved):
Thank you for your work and your patience!

I've just rebased your branch since there were some conflicts caused by recent refactorings on `main`.

hiltonc:
Thanks @SimplyDanny!

PR #4946, opened 2023-04-26
Improve StyleViolation reason formatting validation

jpsim:
So it checks violations for all rules in the registry, not just the built-in rules.


SimplyDanny:
I added this check only for built-in rules intentionally, so that SwiftLint is not enforcing a certain style for custom messages. I'm fine with doing that, but I'm not sure if all people with custom rules are going to like it.

Issue #4632, opened 2022-12-08
`statement_position` fails on one-line `catch`


lhunath:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

Assuming `uncuddled_else`, the following passes validation:
```
do {
    x
}
catch {
    y
}
```

But the following fails validation:
```
do { x }
catch { y }
```

This is inconsistent and should be fixed. One-line statements are very beneficial and shouldn't be punished.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint test.swift
Linting Swift files at paths test.swift
Linting 'test.swift' (1/1)
test.swift:1:8: warning: Statement Position Violation: Else and catch should be on the same line, one space after the previous declaration. (statement_position)
Done linting! Found 1 violation, 0 serious in 1 file.
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? `0.50.1`
* Installation method used (Homebrew, CocoaPods, building from source, etc)? `brew`
* Paste your configuration file:

```yml
statement_position:
  statement_mode: uncuddled_else
```

lhunath:
This is still true. With `uncuddled_else`, the following correctly raises a warning:
```
// cuddled
if one {
    then()
} else if two {
    then()
} else {
    then()
}
```

The following correctly passes the linter:
```
// uncuddled
if one {
    then()
}
else if two {
    then()
}
else {
    then()
}
```

But the following incorrectly raises a warning, even though the `else` is uncuddled, as the rule requires:
```
// uncuddled
if one { then() }
else if two { then() }
else { then() }
```

lhunath:
The issue is that the rule requires there to be white-space in front of the `else`/`catch` which is equal to the amount of white-space in front of the `}`, a check which blindly assumes the `}` and `else` are both alone on a line with nothing but indentation in front of them.
https://github.com/realm/SwiftLint/blob/23d6a7c3f56f2d0815774415a12b5405dfd1d9b0/Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRule.swift#L153

Funnily, the following code is deemed "legal":
```
// uncuddled
if one { then() }
 else if two { then() }
 else { then() }
```

The white-space check should not be performed if the `}` is not alone on a line but preceded by another statement.

SimplyDanny:
[...]

PR #4954, opened 2023-05-02
Refactor `statement_position` rule to use `SwiftSyntax`

enebin:
This PR aims to resolve the issue #4632 by refactoring the code to use SwiftSyntax instead of Regex.

- Solve the issue using regex was quite challenging. The new implementation provides a more reliable and robust solution.
- Throughout the refactoring process, I tried to maintain the original error and test cases.
- I also added more examples with complete syntax.
- With these changes,  the issue #4632 is fixed and the library no longer throws a warning (or error) for this case while supporting all other features.

This is my first pull request for this library's source code so... please let me know if I did anything incorrectly 🙏


lhunath:
Much appreciated!

Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRule.swift
```diff
@@ -1,217 +1,384 @@
...
+            return nil
+        }
+
+        // Check if `else` doesn't have newline in leading trivia of `else`'and
+        // multiple spaces in trailing trivia of `body`'s `rightBrace`
+        guard
+            elseKeyword.leadingTrivia.isEmpty,
+            body.rightBrace.trailingTrivia.isSingleSpace == true
+        else {
+            violationPosition = body.rightBrace.endPositionBeforeTrailingTrivia
```
SimplyDanny, line 206: Having the violation markers on the keywords (`else` and `catch`) seems more natural since their positions actually cause the rule to trigger.

Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRule.swift
```diff
@@ -1,217 +1,384 @@
...
+                clause.previousToken(viewMode: .sourceAccurate)?.isRightBrace == true,
+                let previousRightBrace = clause.previousToken(viewMode: .sourceAccurate)
+            else {
+                return
+            }
+
+            let hasEmptyLeadingTrivia = clause.catchKeyword.leadingTrivia.isEmpty
+            let trailingTriviaSingleSpace = (index == 0
+                                             ? body.rightBrace
+                                             : previousRightBrace).trailingTrivia.isSingleSpace
```
SimplyDanny, line 283: Can this not always be `previousRightBrace` even if `index == 0`?

Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRule.swift
```diff
@@ -1,217 +1,384 @@
...
+    var elseBeforeIfKeyword: TokenSyntax? {
+        if
+            ifKeyword.previousToken(viewMode: .sourceAccurate)?.isElseKeyword == true,
+            let elseBeforeIfKeyword = ifKeyword.previousToken(viewMode: .sourceAccurate)
+        {
+            return elseBeforeIfKeyword
+        }
+
+        return nil
+    }
```
SimplyDanny, line 189:
```suggestion
        if let previous = ifKeyword.previousToken(viewMode: .sourceAccurate), previous.isElseKeyword {
            return previous
        }
        return nil
    }
```

SimplyDanny (commented):
I've just rebased the PR and added a few changes which improve the violation messages making them more specific, readability of the examples and general structure. For example, the uncuddled case should be handled in a separate test. The second description of the rule did not make much sense, so I removed it.

Other than that, I have a few remarks/questions but need some more time understanding the implementation fully. What is already clear is that any comments between the closing brace `}` and the keywords `else`/`catch` will get lost in the rewrite.

enebin:
> @enebin: Is there an interest on your side to continue working on this PR?

@SimplyDanny  I apologize. I had completely forgotten about this PR 😅
Thank you for reminding me. Could you possibly give me a bit more time? I will come back with a review and response to this PR as soon as possible.

SimplyDanny:
> Could you possibly give me a bit more time? I will come back with a review and response to this PR as soon as possible.

Sure. Take your time. I only wanted to make sure this PR is not abandoned.

Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRule.swift
SimplyDanny: Yes, that would be different and, because of that, it's something we shouldn't do in this rewrite as it would make the reported OSS findings unusable. So let's keep this for another PR afterwards. 😉

enebin:
I apologize for the delay. I've been going through a busy period personally, so even a simple request took too long. I've now requested a review again @SimplyDanny

SimplyDanny (changes requested):
Overall it looks good. Glad you found the time to continue with the PR. 🙏

Still, comments will be removed by the rewrite. That might be okay, though, as developers enabling the rule enforce a certain style which would be violated by comments between braces and `else`/`catch` keywords. Not sure if the previous implementation did better. A way to address this, is to skip the rewrite if comments are part of the trivia between the keywords and any braces. The rule would still trigger, but let the fix up to a human to deal with the comments properly.

There are a lot of OSS findings which we need to review thoroughly. Some of them are caused by the changed violation message. While you did great in specializing the message, it's always helpful to only do this in a later step. By keeping the previous message, the OSS report contains only real findings. This is important for a rewrite to make sure that nothing broke. After the review (even in a follow-up PR), you may then change the message. The findings are then exclusively due to the changed wording. Long story short, I suggest to revert the message change and update it later separately.

Issue #4844, opened 2023-03-28
Xcode hanging when applying SwiftLint to many targets 


Mistakx:
### New Issue Checklist

- [X] Updated SwiftLint to the latest version
- [X] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug
I'm currently trying to add SwiftLint as a Build Tool Plug-In in Xcode.
The repo has a Package, and that package has around 70 targets. 
Adding the plugin to every target currently makes it so Xcode gets stuck while building the project.

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.51.0
* Installation method used (Homebrew, CocoaPods, building from source, etc)? Build Tool Plug-In
* Paste your configuration file: N/A

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? No
* Which Xcode version are you using (check `xcodebuild -version`)? 14.2.0
* Do you have a sample that shows the issue? N/A

geertbleyen:
I'm experiencing the same thing. It takes up to 50 seconds in the Prepare packages part before the build actually starts, where all this time is spent on SwiftLint.
![image](https://user-images.githubusercontent.com/71486361/228853972-2ae7068e-6129-4510-8bbd-5657a0a0e317.png)

Using Xcode 14.3 and SwiftLint 0.51.0

geertbleyen:
Additional info: previously we were using cdd891a4a29cfd3473737857385f79c972702293, which didn't have this issue.
Reason for using that version previously was that:
- we had an in-house version for the SPM buildtool plugin to run SwiftLint on our SPM packages. This in-house version also used the binaryTarget and had the same name, so we had to stick with a version of SwiftLint that didn't use the binaryTarget for the buildtool plugin.
- Getting rid of our in-house version wasn't possible until the SPM dependency issue with relying on 'unstable' versions was resolved (which happened only in 0.51.0 if I'm not mistaken).

I've now spent hours in switching all our SPM packages to use SwiftLint 0.51.0 (which by themselves works ok as seen in the screenshot above, although individual SPM packages are rather small for us).
Only at the end of my work, where the 'big' application starts now consuming those packages and integrates SwiftLint itself, did this issue show up.

geertbleyen:
For now I decided to temporarily use SwiftLint as script in the target's buildphases to mitigate this issue.
[...]

PR #4955, opened 2023-05-02
Move `excludedPaths` out of iteration loop in favor of dependency inj…

andyyhope:
# Problem 
When using SwiftLint's Swift Plugin feature for moderately sized Packages (1000+ files), build times would regress considerably in comparison to the script integration - even when a cache was provided for successive builds.

### Issue
- [#4844](https://github.com/realm/SwiftLint/issues/4844)

# Cause
This build regression comes from the `excludedPaths(fileManager:)` function being called when iterating the files array for the lint process. For Swift packages, this can result in a considerably large computation time (0.33s on an M1 mac) because it can potentially return thousands of files.

# Solution
Because this doesn't appear to change between iterations, I've moved the API call up the stack and outside the iteration loop and passed it down via arguments. There are only a few call-sites which needed to be updated and it doesn't change any functionality.

To maintain code behaviour, I've created a new `enum` type `ExcludeBy`. This maintains mutual exclusivity between the types of exclusion filepaths: `filterExcludedPathsByPrefix` and `filterExcludedPaths`. This is the type that gets passed down to the linter.

# Tests
I've also updated all tests with the new API changes and they're all passing. ✅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,10 @@
   [keith](https://github.com/keith)
   [#4782](https://github.com/realm/SwiftLint/issues/4782)
 
+* Move `excludedPaths` out of iteration loop in favor of dependency injection  
+  [andyyhope](https://github.com/andyyhope)
+  [#4844](https://github.com/realm/SwiftLint/issues/4844)
+
 ## 0.51.0: bzllint
 
 #### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,8 +70,8 @@
   [keith](https://github.com/keith)
   [#4782](https://github.com/realm/SwiftLint/issues/4782)
 
-* Move `excludedPaths(fileManager:)` operation out of linting iterations
-  in favor of dependency injection to improve build speeds for Swift Plugin.   
+* Improve lint times of SwiftLintPlugin by moving the
+  `excludedPaths(fileManager:)` operation out of the linting iterations.  
   [andyyhope](https://github.com/andyyhope)
   [#4844](https://github.com/realm/SwiftLint/issues/4844)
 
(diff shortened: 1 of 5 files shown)


Source/SwiftLintCore/Extensions/Configuration+LintableFiles.swift
```diff
@@ -98,7 +108,7 @@ extension Configuration {
     /// - parameter fileManager: The file manager to get child paths in a given parent location.
     ///
     /// - returns: The expanded excluded file paths.
-    private func excludedPaths(fileManager: LintableFileManager) -> [String] {
+    public func excludedPaths(fileManager: LintableFileManager) -> [String] {
```
SimplyDanny, line 111: Perhaps add a default argument (`fileManager: LintableFileManager = FileManager.default)`) here?

andyyhope: Sure. Added in a new commit

CHANGELOG.md
```diff
@@ -70,6 +70,10 @@
   [keith](https://github.com/keith)
   [#4782](https://github.com/realm/SwiftLint/issues/4782)
 
+* Move `excludedPaths` out of iteration loop in favor of dependency injection
```
SimplyDanny, line 73: This could be more informative in that it describes what the change means for users. The current entry is quite technical and an internal detail.

Also, please finish the entry with a period and two spaces to enforce a line break in Markdown.

Sorry missed that in the last review ...

SimplyDanny (approved):

CHANGELOG.md
andyyhope: Sure np. I've changed it to:
> Move `excludedPaths(fileManager:)` operation out of linting iterations
  in favor of dependency injection to improve build speeds for Swift Plugin.

SimplyDanny:
Thank you, @andyyhope!

PR #4958, opened 2023-05-04
Add `unhandled_throwing_task` rule

kylebshr:
This rule will check Task's that are not explicitly annotated with success and failure types for unhandled `try` expressions. These `try`s will silently fail if an error is thrown. 

I decided against prescribing a fix, but decided this should be an `error` by default. It would be cool if you could configure it to, perhaps:

- Replace these `Task`s with `Task<Void, Never>`
- Wrap the try expression in a `do`/`catch`

However I think this is quite useful on its own.

Thank you so much @jpsim for getting this started for me! Let me know if you think it needs any improvements.


Source/SwiftLintBuiltInRules/Rules/Lint/UnhandledTryInTaskRule.swift
```diff
@@ -0,0 +1,111 @@
...
+
+private extension UnhandledTryInTaskRule {
+    final class Visitor: ViolationsSyntaxVisitor {
+        override func visitPost(_ node: FunctionCallExprSyntax) {
+            if let typeIdentifier = node.calledExpression.as(IdentifierExprSyntax.self) {
+                if typeIdentifier.identifier.text == "Task" {
+                    let throwsVisitor = ThrowsVisitor(viewMode: .sourceAccurate)
+                    throwsVisitor.walk(node)
+                    if throwsVisitor.doesThrow {
+                        violations.append(node.calledExpression.positionAfterSkippingLeadingTrivia)
```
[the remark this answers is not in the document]
kylebshr: Seems like it can be a follow up IMO

kylebshr:
oooh just realized this will catch `try?`, we don't want that!

kylebshr:
and a bug in my catch checking, will push some follow ups

jpsim:
Could you please review the results found by OSSCheck in this comment to see if it includes any false positives? https://github.com/realm/SwiftLint/pull/4958#issuecomment-1533976824

jpsim:
Should a top-level `throw` in the Task body also trigger a rule violation?

kylebshr:
> Should a top-level throw in the Task body also trigger a rule violation?

Yeah good call!

Source/SwiftLintCore/Models/Example.swift
```diff
@@ -120,7 +120,7 @@ extension Example {
     }
 
     /// Makes the current example focused. This is for debugging purposes only.
-    func focused() -> Example { // swiftlint:disable:this unused_declaration
+    public func focused() -> Example { // swiftlint:disable:this unused_declaration
```
jpsim, line 123: Could you please revert this? I'm making it public in a separate PR: https://github.com/realm/SwiftLint/pull/4959

Source/SwiftLintBuiltInRules/Rules/Lint/UnhandledTryInTaskRule.swift
```diff
@@ -0,0 +1,170 @@
+import SwiftSyntax
+
+struct UnhandledTryInTaskRule: ConfigurationProviderRule, SwiftSyntaxRule {
+    var configuration = SeverityConfiguration(.error)
+
+    static let description = RuleDescription(
+        identifier: "unhandled_try_in_task",
+        name: "Unhandled Try in Task",
+        description: "Errors thrown inside of this task are not handled",
```
jpsim, line 9: Nit:

```suggestion
        description: "Errors thrown inside this task are not handled",
```

Source/SwiftLintBuiltInRules/Rules/Lint/UnhandledTryInTaskRule.swift
```diff
@@ -0,0 +1,170 @@
+import SwiftSyntax
+
+struct UnhandledTryInTaskRule: ConfigurationProviderRule, SwiftSyntaxRule {
+    var configuration = SeverityConfiguration(.error)
+
+    static let description = RuleDescription(
+        identifier: "unhandled_try_in_task",
+        name: "Unhandled Try in Task",
```
jpsim, line 8: Since there might not be a `try` keyword in the body of the task, this might be misleading.

Perhaps the rule should be `unhandled_throwing_task`?

```suggestion
        identifier: "unhandled_throwing_task",
        name: "Unhandled Throwing Task",
```

Source/SwiftLintBuiltInRules/Rules/Lint/UnhandledTryInTaskRule.swift
```diff
@@ -0,0 +1,170 @@
...
+    func makeVisitor(file: SwiftLintFile) -> ViolationsSyntaxVisitor {
+        Visitor(viewMode: .sourceAccurate)
+    }
+}
+
+private extension UnhandledTryInTaskRule {
+    final class Visitor: ViolationsSyntaxVisitor {
+        override func visitPost(_ node: FunctionCallExprSyntax) {
+            if let typeIdentifier = node.calledExpression.as(IdentifierExprSyntax.self) {
+                if typeIdentifier.identifier.text == "Task" {
```
jpsim, line 119: Should we also check to make sure that placeholder generics aren't used? E.g. `Task<_, _>`

Source/SwiftLintBuiltInRules/Rules/Lint/UnhandledThrowingTaskRule.swift
```diff
@@ -0,0 +1,238 @@
+import SwiftSyntax
+
+struct UnhandledThrowingTaskRule: ConfigurationProviderRule, SwiftSyntaxRule {
+    var configuration = SeverityConfiguration(.error)
+
+    static let description = RuleDescription(
+        identifier: "unhandled_throwing_task",
+        name: "Unhandled Throwing Task",
+        description: "Errors thrown inside this task are not handled",
```
jpsim, line 9: Is there a link for more information that you could include in this description? I feel like someone new to Swift Concurrency might not understand what the problem is or how to fix it.

E.g.

```suggestion
        description: "Errors thrown inside this task are not handled (See this forum thread for more details https://forums.swift.org/...)",
```

Source/SwiftLintBuiltInRules/Rules/Lint/UnhandledThrowingTaskRule.swift
```diff
@@ -0,0 +1,238 @@
...
+        return parent.name.text == "value"
+    }
+
+    var doesThrow: Bool {
+        ThrowsVisitor(viewMode: .sourceAccurate)
+            .walk(tree: self, handler: \.doesThrow)
+    }
+}
+
+private final class ThrowsVisitor: SyntaxVisitor {
```
jpsim, line 201: Maybe a little docstring here would be useful to say what this detects at a high level.

Source/SwiftLintBuiltInRules/Rules/Lint/UnhandledThrowingTaskRule.swift
```diff
@@ -0,0 +1,238 @@
...
+    var doesThrow = false
+
+    override func visit(_ node: DoStmtSyntax) -> SyntaxVisitorContinueKind {
+        // If there are no catch clauses, visit children to see if there are any try expressions.
```
jpsim, line 205: Does this handle these cases?

```swift
Task {
  do {
    try foo()
  } catch {
    try bar()
  }
}
```

or:

```swift
Task {
  do {
    try foo()
  } catch {
    throw BarError()
  }
}
```

jpsim:
> do you think we need to address the case where the Task is returned to ship this?

I'd appreciate if you take a stab at looking at the [implicit_return rule](https://github.com/realm/SwiftLint/blob/e86c06c3902cc5d2f09b71bc0c985b4231f4a26e/Source/SwiftLintBuiltInRules/Rules/Style/ImplicitReturnRule.swift) to see if you can reuse that logic to detect if the task is returned, but if it's too challenging to implement this check, I don't think it should block this rule from landing.

Source/SwiftLintBuiltInRules/Rules/Lint/UnhandledThrowingTaskRule.swift
kylebshr: It did not, but with my latest changes it does, good call out!

kylebshr:
@jpsim I did take a look at the different return rules, none of them were particularly helpful for this case. I hope the return situation is uncommon enough that most code bases can still enable this rule without any issues

Source/SwiftLintBuiltInRules/Rules/Lint/UnhandledThrowingTaskRule.swift
```diff
@@ -0,0 +1,260 @@
+import SwiftSyntax
+
+struct UnhandledThrowingTaskRule: ConfigurationProviderRule, SwiftSyntaxRule {
+    var configuration = SeverityConfiguration(.error)
+
+    static let description = RuleDescription(
+        identifier: "unhandled_throwing_task",
+        name: "Unhandled Throwing Task",
+        description: "Errors thrown inside this task are not handled, which may be unexpected. " +
```
jpsim, line 9: Could you please use a multiline string literal here? This also fixes some spacing issues in the string.

```swift
description: """
    Errors thrown inside this task are not handled, which may be unexpected. \
    Handle errors inside the task, or use `try await` to access the Tasks value and handle errors. \
    See this forum thread for more details: \
    https://forums.swift.org/t/task-initializer-with-throwing-closure-swallows-error/56066
    """,
```

Source/SwiftLintBuiltInRules/Rules/Lint/UnhandledThrowingTaskRule.swift
```diff
@@ -0,0 +1,260 @@
...
+}
+
+/// If the `doesThrow` property is true after visiting, then this node throws an error that is "unhandled."
+/// Try statements inside a `do` with a `catch` that handles all errors will not be marked as throwing.
+private final class ThrowsVisitor: SyntaxVisitor {
+    var doesThrow = false
+
+    override func visit(_ node: DoStmtSyntax) -> SyntaxVisitorContinueKind {
+        // If there are no catch clauses, visit children to see if there are any try expressions.
+        guard let lastCatchClause = node.catchClauses?.last else {
```
jpsim, line 229: Should we exit early and return `.skipChildren` if `doesThrow` is true?

jpsim (approved):
Looking good! Minor suggestions, and needs another commit at least to re-trigger CI which was down when you pushed your last commit.

Source/SwiftLintBuiltInRules/Rules/Lint/UnhandledThrowingTaskRule.swift
kylebshr: Ah yeah, good catch - was just following what I saw in another rule :)

Source/SwiftLintBuiltInRules/Rules/Lint/UnhandledThrowingTaskRule.swift
kylebshr: Good idea!

jpsim:
One last thing, can you please add a changelog entry following the formatting described in `CONTRIBUTING.md`?

PR #4965, opened 2023-05-06
feat: add implicitly_unwrapped_weak rule

rickytan:
This rule checks dangerous usage of force unwrapping weak member, such as:
```swift
class MyClass {
  weak var bar: SomeObject!
}
```

Except `@IBOutlet` member

### DIFF
--- a/Source/SwiftLintBuiltInRules/Models/BuiltInRules.swift
+++ b/Source/SwiftLintBuiltInRules/Models/BuiltInRules.swift
@@ -86,6 +86,7 @@ public let builtInRules: [Rule.Type] = [
     IdentifierNameRule.self,
     ImplicitGetterRule.self,
     ImplicitReturnRule.self,
+    ImplicitlyUnwrapWeakVariableRule.self,
     ImplicitlyUnwrappedOptionalRule.self,
     InclusiveLanguageRule.self,
     IndentationWidthRule.self,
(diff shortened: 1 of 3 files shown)


SimplyDanny:
How is this rule different from the already existing rule `implicitly_unwrapped_optional`?

rickytan:
Oh, I didn't notice that this rule also worked on this case. So this PR can be closed.

rickytan:
Actually , I want to bypass the following cases:
```swift
class MyClass {
   var bar: SomeObject!
}
```
and
```swift
func someFunc(_ param0: SomeObject!)
```
But if I opt-in `implicitly_unwrapped_optional`, it will make too many unintended violations

SimplyDanny:
I get your point. A `weak` implicitly unwrapped optional is a contradiction. Instead of a new rule, what do you think about an option for `implicitly_unwrapped_optional`?

Side question: What would be a useful example for an implicitly unwrapped function parameter as in `func f(param: SomeObject!)`?

rickytan:
Ok, I'll add a new mode for `implicitly_unwrapped_optional` rule.

For the side question: We had a lot of objc protocol function to be implemented in Swift, but there didn't annotated with `nonnull` or `nullable` properly

rickytan:
See https://github.com/realm/SwiftLint/pull/4975 , and this can be closed @SimplyDanny

PR #5030, opened 2023-05-22
Windows Port

compnerd:
This is currently not possible to build on Windows.  It requires that SourceKitten changes are merged (https://github.com/jpsim/SourceKitten/pull/769) and some changes to the toolchain that are still in the process of getting merged upstream (https://github.com/apple/swift-package-manager/pull/6573) which depends on a toolchain that is not yet tagged.

It is not possible to build the tests on Windows due to the use of the executable target as a dependency for the tests which results in multiple definitions of `main`.

Running the generated binary over the SwiftLint source base currently identifies a crash at some point, which needs to be investigated.


Source/SwiftLintCore/Helpers/Reachability.swift
```diff
@@ -1,4 +1,4 @@
-#if !os(Linux)
+#if !os(Linux) && !os(Windows)
```
SimplyDanny, line 1:
```suggestion
#if os(macOS)
```

Source/swiftlint/Commands/Rules.swift
```diff
@@ -149,12 +154,20 @@ private extension TextTable {
 
 private struct Terminal {
     static func currentWidth() -> Int {
+#if os(Windows)
+        var csbi: CONSOLE_SCREEN_BUFFER_INFO = CONSOLE_SCREEN_BUFFER_INFO()
```
SimplyDanny, line 158:
```suggestion
        var csbi = CONSOLE_SCREEN_BUFFER_INFO()
```

SimplyDanny (approved):
This basically looks fine to me. What about the open issues/PRs from other repositories you mentioned in the PR description? Is there anything more required to make SwiftLint work on Windows?

compnerd:
The only thing needed is to get the SourceKitten changes in (but haven't been able to get response from @jpsim to get those resolved and merged).

jpsim:
> The only thing needed is to get the SourceKitten changes in (but haven't been able to get response from @jpsim to get those resolved and merged).

Last I checked the windows PRs for some SwiftLint dependencies failed to build or failed tests for macOS or Linux. I’m happy to review again when those failures are addressed.

Source/SwiftLintFramework/Helpers/Glob.swift
```diff
@@ -20,6 +24,28 @@ struct Glob {
 
         return expandGlobstar(pattern: pattern)
             .reduce(into: [String]()) { paths, pattern in
+#if os(Windows)
+                URL(fileURLWithPath: pattern).withUnsafeFileSystemRepresentation {
+                    var ffd: WIN32_FIND_DATAW = WIN32_FIND_DATAW()
+
+                    let hDirectory: HANDLE = String(cString: $0!).withCString(encodedAs: UTF16.self) {
+                        FindFirstFileW($0, &ffd)
```
roman-bcny, line 32: This doesn't appear to work for globs that include directories

```
SwiftLintPackageTests.xctest	21520	CreateFile	C:\workspace\SwiftLint\Tests\FileSystemAccessTests\Resources\ProjectMock\Level1\*\*.swift	NAME INVALID	Desired Access: Generic Read, Disposition: Open, Options: Synchronous IO Non-Alert, Open Reparse Point, Attributes: n/a, ShareMode: Read, Write, Delete, AllocationSize: n/a
```

testExcludeByPrefixGlobExcludePaths runs this for `Level1/*/*.swift`, and on Windows only `Level1/*.swift` would work but `Level1/*/*.swift` gets treated as `Level1/*` dir with the `*.swift` glob, which can't find the dir and fails.

From a cursory research it seems Windows doesn't expose this as an api and we'll have to manually split the string and query recursively to handle such cases.

compnerd:
@SimplyDanny - okay, I think that I've addressed the lint issue which means that this should pass now. Once this is merged, we should be able to start investigating and fixing the test failures.

Source/SwiftLintFramework/Helpers/Glob.swift
SimplyDanny: I might be tougher than it seems as this is an area that has caused a lot of headaches in the past (even in the Unix case). See #6342 as an example where I try to improve performance (again). It also lists (some) issues related to that topic.

On the other hand, getting it to work at all on Windows should be doable. Performance is another topic then ...

PR #5040, opened 2023-05-30
Rewrite `todo` rule with SwiftSyntax

woxtu:
Hi,

As the title says, this PR rewrites `todo` rule with SwiftSyntax.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/TodoRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/TodoRule.swift
@@ -1,14 +1,15 @@
 import Foundation
 import SourceKittenFramework
+import SwiftSyntax
 
-extension SyntaxKind {
+extension SourceKittenFramework.SyntaxKind {
     /// Returns if the syntax kind is comment-like.
     var isCommentLike: Bool {
         return Self.commentKinds.contains(self)
     }
 }
 
-struct TodoRule: ConfigurationProviderRule {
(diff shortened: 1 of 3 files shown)


SimplyDanny:
We should check the new OSS violations and violation fixes carefully. Some of them look suspicious. I wouldn't expect any behavioural changes of a rewrite of this rule.

Source/SwiftLintBuiltInRules/Rules/Lint/TodoRule.swift
```diff
@@ -32,54 +25,90 @@ struct TodoRule: ConfigurationProviderRule {
...
+        switch self {
+        case
+                .blockComment(let comment),
+                .lineComment(let comment),
+                .docBlockComment(let comment),
+                .docLineComment(let comment):
+            var startIndex = comment.startIndex
+            var violations = [ReasonedRuleViolation]()
+            while let result = try? NSRegularExpression.cached(pattern: #"\b(TODO|FIXME)(?::|\b)([^\n]*)"#)
+                .firstMatch(in: comment, range: NSRange(startIndex..., in: comment)) {
```
SimplyDanny, line 66: Out of interest: Can we use `matches` here instead of `firstMatch` to avoid the `while` loop?

woxtu: Probably.

jpsim:
Thanks for your contribution, @woxtu!

jpsim (approved):

PR #5054, opened 2023-06-11
Add rule suggesting static immutable property

snofla:
Adds a rule that suggests to use a static modifier for immutable initialized properties.

So this:
```
struct Foo {
   let property = 2
}
```
is suggested to be:
```
struct Foo {
   static let property = 2
}
```


snofla:
Needs some extra work - closing..

PR #5065, opened 2023-06-17
Merge allowed symbols

SimplyDanny:


SimplyDanny:
The OSS changes have its source only in the changed violation messages.

PR #5068, opened 2023-06-19
Silence duplicate symbol runtime warning

jpsim:
Similar to #4901 but updated with a new warning from Xcode 15 beta 1.

Unfortunately this doesn't appear to work, even with this change the warning is still logged:

```
objc[18947]: Class _TtC11SwiftParserP33_78149DB072C20084E7D780D86E26C3AF41StringLiteralExpressionIndentationChecker is implemented in both /private/var/tmp/_bazel_jp/d6a0d3abed4579623f79fc46ad4a7a79/execroot/_main/bazel-out/darwin_arm64-opt/bin/swiftlint (0x1059a79a0) and /Applications/Xcode-15.0.0-Beta.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/host/libSwiftParser.dylib (0x10703da50). One of the two will be used. Which one is undefined.
```

It might have something to do with how long the symbol name is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,12 @@
   [woxtu](https://github.com/woxtu)
   [#3023](https://github.com/realm/SwiftLint/issues/3023)
 
+* Work around dyld warning about duplicate SwiftSyntax classes with Xcode 15
+  betas.  
+  [keith](https://github.com/keith)
+  [JP Simard](https://github.com/jpsim)
+  [#4782](https://github.com/realm/SwiftLint/issues/4782)
+
 ## 0.52.2: Crisper Clearer Pleats
 
 #### Breaking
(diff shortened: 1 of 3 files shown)


jpsim:
Any thoughts @keith ?

keith:
Soo, the issue is related to the warning you're disabling. It's actually fixed if you change the name size from `64` to `128` (or probably just anything longer than this string) in objc_dupclass.h. I don't know _why_ that works because it sure seems like dyld expects the strings to only be 64 characters https://github.com/apple-oss-distributions/dyld/blob/c8a445f88f9fc1713db34674e79b00e30723e79d/common/MachOAnalyzer.cpp#L4381 but it seems to so I guess I'd go with that!

jpsim:
Thanks for digging into this, @Keith. Indeed increasing the max dup class name length appears to resolve the issue with Xcode 15 beta 1 and beta 2.

PR #5076, opened 2023-06-22
Fix for #5075

### LINKED ISSUE #5075: False positive in unneeded_synthesized_initializer
### New Issue Checklist

### Describe the bug

`unneeded_synthesized_initializer` reports a false positive when the init has side effects. In the attached sample, the `init` prints something, and removing it would remove the print. In the example where I encountered it in a real app, a SwiftUI view modifier's `init` contained a call to `UIScrollView.appearance().bounces = false`, which we definitely don't want to lose by deleting the `init`!

```bash
$ swiftlint lint
Linting 'InitTest.swift' (1/1)
/Users/igeek/Downloads/InitTest/InitTest.swift:2:5: warning: Unneeded Synthesized Initializer Violation: This default initializer would be synthesized automatically - you do not need to define it (unneeded_synthesized_initializer)
```

### Environment

* SwiftLint version: **0.52.3**
* Installation method used: **homebrew**
* Paste your configuration file: **n/a (default configuration)**
* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? **No**
* Which Xcode version are you using? **Xcode 14.3.1 (14E300c) on macOS 13.4.1 (22F82)**

```swift
struct MyStruct {
    init() {
        print("perform side effect")
    }
}
```

mildm8nnered:
Fixes #5075

For the no argument case, we weren't checking for side-effects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,10 @@
 
 #### Bug Fixes
 
-* None.
+* Fix false positives for the `unneeded_synthesized_initializer` rule, when
+  no argument initializers had side-effects.  
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#5075](https://github.com/realm/SwiftLint/issues/5075)
 
 ## 0.52.3: Duplicate Hampers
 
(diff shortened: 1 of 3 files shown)


SimplyDanny (approved):

SimplyDanny:
Thank you for the prompt fix!

PR #5083, opened 2023-06-26
Improves correction for cases with trailing comments

### LINKED ISSUE #5081: Rule `direct_return` comments the whole line.
### New Issue Checklist

### Describe the bug

When using `swiftlint --fix` with the `direct_return`rule enabled, the following code lines are mishandled:
```swift
let view = nib.instantiate(withOwner: self, options: nil).first as! UIView // swiftlint:disable:this force_cast
return view
``` 

The rule currently updates the code to
```swift
 // swiftlint:disable:this force_cast        return nib.instantiate(withOwner: self, options: nil).first as! UIView // swiftlint:disable:this force_cast
```
adding the comment both before and after the statement.
An expected update would be 
```swift
return nib.instantiate(withOwner: self, options: nil).first as! UIView // swiftlint:disable:this force_cast
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.52.3
* Installation method used (Homebrew, CocoaPods, building from source, etc)? brew

```yml
only_rules:
  - direct_return
```

* Which Xcode version are you using (check `xcodebuild -version`)? Version 14.3.1 (14E300c)

mildm8nnered:
Fixes #5081

Appends the trailing trivia from the variable declaration and the return statement to the new return statement.

This works great for the case reported in the report:

```
func f() -> UIView {
    let view = instantiateView() as! UIView // swiftlint:disable:this force_cast
    return view
}
 ```

becomes

```
func f() -> UIView {
    return instantiateView() as! UIView // swiftlint:disable:this force_cast
}
```

When both have trailing comments, it's not so great, but not completely awful, and certainly a lot better than what we had before.

```
func f() -> UIView {
    let view = instantiateView() as! UIView // swiftlint:disable:this force_cast
    return view // return the view
}
```

now becomes

```
func f() -> UIView {
   return instantiateView() as! UIView // swiftlint:disable:this force_cast // return the view
}
```

as opposed to (on 0.52.2), somewhat crazily

```
func f() -> UIView {
 // swiftlint:disable:this force_cast    return instantiateView() as! UIView // swiftlint:disable:this force_cast
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@
   [Martin Redington](https://github.com/mildm8nnered)
   [#5075](https://github.com/realm/SwiftLint/issues/5075)
 
+* Fix auto-correction for the `direct_return` rule, when statements have
+  trailing comments.  
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#5081](https://github.com/realm/SwiftLint/issues/5081)
+
 ## 0.52.3: Duplicate Hampers
 
 #### Breaking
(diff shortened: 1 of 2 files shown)


SimplyDanny (approved):
Thank you for the prompt fix! Handling trivia is always fun ... 😄

PR #5084, opened 2023-06-26
Fix false positive for `prefer_self_in_static_references` rule on class declarations

kasrababaei:
### Description of changes:

The `prefer_self_in_static_references` rule triggers a warning for class `Movie` in the following example:

```Swift
class Movie<Genre> {
    let genre: Genre
    
    init(genre: Genre) {
        self.genre = genre
    }
}

class StarWars: Movie<StarWars.Genre> { // <- triggers a warning
    enum Genre {
        case sciFi
    }
}
```

However, changing `class StarWars: Movie<StarWars.Genre>` to `class StarWars: Movie<Self.Genre>` would throw the following compile time error:

> Covariant 'Self' or 'Self?' can only appear as the type of a property, subscript or method result; did you mean 'StarWars'?
> Replace 'Self' with 'StarWars'

I think the issue is the grand parent syntax isn't used to checked to see if it's a `GenericArgumentSyntax` or not.

Maybe there's a better solution, and I'm open to suggestions, but this one didn't break any tests and is working.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -172,6 +172,10 @@
   [JP Simard](https://github.com/jpsim)
   [#4985](https://github.com/realm/SwiftLint/issues/4985)
 
+* Fix false positive for `prefer_self_in_static_references` when a class inherits from
+  another class with generic types.  
+  [kasrababaei](https://github.com/kasrababaei)
+
 ## 0.52.0: Crisp Clear Pleats
 
 #### Breaking
--- a/Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRule.swift
@@ -179,6 +179,13 @@ private class Visitor: ViolationsSyntaxVisitor {
         parentDeclScopes.pop()
     }
 
+    override func visit(_ node: GenericArgumentListSyntax) -> SyntaxVisitorContinueKind {
+        if case .likeClass = parentDeclScopes.peek() {
+            return .skipChildren
+        }
+        return .visitChildren
+    }
+
     override func visitPost(_ node: SimpleTypeIdentifierSyntax) {
         guard let parent = node.parent else {
             return
--- a/Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRuleExamples.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRuleExamples.swift
@@ -72,6 +72,33 @@ enum PreferSelfInStaticReferencesRuleExamples {
                 let c: C = C()
                 func f(c: C) -> KeyPath<C, Int> { \\Self.i }
             }
+        """, excludeFromDocumentation: true),
+        Example("""
+            class C1<T> {}
+            class C2: C1<C2> {}
+        """, excludeFromDocumentation: true),
+        Example("""
+                class C1<T> {}
+                class C2: C1<C2.C3> {
+                    class C3 {}
+                }
+                """, excludeFromDocumentation: true),
+        Example("""
+            class C1<T> {}
+            class C2: C1<C2.C3.C4> {
+                class C3 {
+                    class C4 {}
+                }
+            }
+        """, excludeFromDocumentation: true),
+        Example("""
+            class S1<T> {
+                class S2 {}
+                func f() {
+                    let s1 = S1<S1.S2>()
+                    let s2 = S1<S1>()
+                }
+            }
         """, excludeFromDocumentation: true)
     ]
 
@@ -184,6 +211,14 @@ enum PreferSelfInStaticReferencesRuleExamples {
                     return f
                 }
                 func g(a: [↓S]) -> [↓S] { a }
+            }
+        """, excludeFromDocumentation: true),
+        Example("""
+            class T {
+                let child: T
+                init(input: Any) {
+                    child = (input as! T).child
+                }
             }
         """, excludeFromDocumentation: true)
     ]


Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRuleExamples.swift
```diff
@@ -72,6 +72,9 @@ enum PreferSelfInStaticReferencesRuleExamples {
                 let c: C = C()
                 func f(c: C) -> KeyPath<C, Int> { \\Self.i }
             }
+        """, excludeFromDocumentation: true),
+        Example("""
+            class C: A<C.i> {}
```
SimplyDanny, line 77: We like to have the examples as valid/compilable Swift code. This example isn't valid because `C.i` doesn't exist.

Also there are a few other cases I'd like to see test cases for. Your example, having it as

```swift
class C1<T> {}
class C2: C1<C2.C3> { // rule not triggering here
    class C3 {}
}
```

is one case. Another one is just

```swift
class C1<T> {}
class C2: C1<C2> {} // rule not triggering here
```

without the inner type.

Now, your fix also causes new issues in some [OSS projects](https://github.com/realm/SwiftLint/pull/5084#issuecomment-1608180503) which we have to check and make sure the findings and fixes are okay. There are at least two other cases we need to have tests for. These are:

```swift
class S1<T> {
    class S2 {}
    func f() {
        let s1 = S1<S1.S2>() // rule not triggering here
        let s2 = S1<S1>()    // rule could trigger here but shouldn't
    }
```

This applies to classes only. In structs, the rule may trigger in both cases.

CHANGELOG.md
```diff
@@ -172,6 +172,10 @@
   [JP Simard](https://github.com/jpsim)
   [#4985](https://github.com/realm/SwiftLint/issues/4985)
 
+* Do not trigger `prefer_self_in_static_references` rule on class declarations.  
+  [kasrababaei](https://github.com/kasrababaei)
+  [#5084](https://github.com/realm/SwiftLint/issues/5084)
```
SimplyDanny, line 177: This line is supposed to refer to an issue given that it exist. For your finding, there isn't an issue so the line can be omitted.

SimplyDanny (changes requested):
Thank you for bringing this issue up and proposing an immediate fix! I have a some remarks though.

Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRuleExamples.swift
kasrababaei: Added two more examples and edited the one that was added.

Sorry for doing a force-push, now you can't do "review changes since your last review". Just wanted to fix it in one commit.

Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRule.swift
```diff
@@ -180,6 +180,11 @@ private class Visitor: ViolationsSyntaxVisitor {
     }
 
     override func visitPost(_ node: SimpleTypeIdentifierSyntax) {
+        if let grandParent = node.parent?.parent,
+           case .likeClass = parentDeclScopes.peek(), grandParent.is(GenericArgumentSyntax.self) {
+            // Type is a generic parameter in a class.
+            return
+        }
```
SimplyDanny, line 187: What about this (admittedly rather pathological) example?

```swift
class C1<T> {}
class C2: C1<C2.C3.C4> {
    class C3 {
        class C4 {}
    }
}
```

I assume it would fail the test.

It's easier to skip `GenericArgumentListSyntax` elements entirely when we're in a class scope. That can be done by another `visit` method. Looking up parents is only useful for specific cases. When recursion comes into play this typically fails.

CHANGELOG.md
```diff
@@ -172,6 +172,10 @@
   [JP Simard](https://github.com/jpsim)
   [#4985](https://github.com/realm/SwiftLint/issues/4985)
 
+* Fix false positive for `prefer_self_in_static_references` when a class inherits from
+  another class with generic types.
```
SimplyDanny, line 176:
```suggestion
  another class with generic types.  
```
Two spaces at the end are important for rendering.

SimplyDanny (changes requested):

Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRule.swift
kasrababaei: Yeah, I don't think recursion makes sense here. I just removed checking `GenericArgumentListSyntax` as suggested and tests now pass. Just couldn't understand this part:

> That can be done by another `visit` method

SimplyDanny: With that change, the rule doesn't trigger on cases like [this](https://github.com/JohnCoates/Aerial/blob/f5a42f790e7e09e9461c2faf0f1b79ea9c98c67e/Aerial/Source/Views/Layers/AnimationLayer.swift#L35:35) anymore. We might need another test case for that. 😅

This is the reason why I suggested:

> That can be done by another `visit` method.

Without verifying it myself, I think adding the method

```swift
override func visit(_ node: GenericArgumentListSyntax) -> SyntaxVisitorContinueKind {
    if case .likeClass = parentDeclScopes.peek() {
        return .skipChildren
    }
    return .visitChildren
}
```

would resolve the conflict.

kasrababaei: You basically resolved it yourself 😄. Pushed the changes, it worked locally, waiting for CI to finish.

SimplyDanny: No worries. That's called _knowledge sharing_. 😉

SimplyDanny (approved):
Looks good now. Thank you for your cooperation!

PR #5088, opened 2023-06-29
Document the case-sensitivity of included and excluded parameters

revolter:
Because git is case-insensitive, I was able to rename a folder long ago without having my colleague notice it, and after that, I added some include and exclude rules which didn't apply to him the same way they did for me. So, after an ~1h long debugging, we finally figured out the cause, and I thought of making sure that others notice this earlier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -506,9 +506,9 @@ opt_in_rules: # some rules are turned off by default, so you need to opt-in
 analyzer_rules: # Rules run by `swiftlint analyze`
   - explicit_self
 
-included: # paths to include during linting. `--path` is ignored if present.
+included: # case-sensitive paths to include during linting. `--path` is ignored if present.
   - Source
-excluded: # paths to ignore during linting. Takes precedence over `included`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -506,9 +506,9 @@ opt_in_rules: # some rules are turned off by default, so you need to opt-in
 analyzer_rules: # Rules run by `swiftlint analyze`
   - explicit_self
 
-included: # paths to include during linting. `--path` is ignored if present.
+included: # case-sensitive paths to include during linting. `--path` is ignored if present.
```
SimplyDanny, line 509: Since it ends with a period, the first letter should be uppercase I think:

```suggestion
included: # Case-sensitive paths to include during linting. `--path` is ignored if present.
```

SimplyDanny (approved):

README.md
revolter: The majority of the inline comments start with lowercase letters and don't end in a full stop, so I thought to go the other way, and change some other exceptions to get them inline (pun not intended). What do you think?

PR #5096, opened 2023-07-03
Fix false positives for the private_subject rule

kasrababaei:
The `private_subject` rule triggers a warning when creating a subject inside of an initializer. 

Triggering example:
```Swift
class Foo {
    let isEnabled: AnyPublisher<Bool, Never>

    init() {
        let isEnabled = PassthroughSubject<Bool, Never>() // This triggers a warning but it shouldn't
        self.isEnabled = isEnabled.eraseToAnyPublisher()
    }
}
```


SimplyDanny (approved):
Looks good. Thank you!

Issue #5103, opened 2023-07-06
False positive on @Environment(\.presentationMode)


chrisngabp:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

The line `@Environment(\.presentationMode) private var presentationMode` is complaining with an attributes violation: `Attributes Violation: Attributes should be on their own lines in functions and types, but on the same line as variables and imports (attributes)`.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint lint
warning: Attributes Violation: Attributes should be on their own lines in functions and types, but on the same line as variables and imports (attributes)
Done linting! Found 1 violation, 0 serious in 1639 files.
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? `0.52.3`
* Installation method used (Homebrew, CocoaPods, building from source, etc)? `CocoaPods`
* Which Xcode version are you using (check `xcodebuild -version`)? `14.3`

```swift
// This triggers a violation:
@Environment(\.presentationMode) private var presentationMode
```

SimplyDanny:
The rule has an option `attributes_with_arguments_always_on_line_above` which is `true` by default.

chrisngabp:
@SimplyDanny maybe we can improve the text `Attributes should be on their own lines in functions and types, but on the same line as variables and imports (attributes)` to be `Attributes should be on their own lines in functions, variables with attributes and types, but on the same line as variables and imports (attributes)`

PR #5108, opened 2023-07-07
Improving attributes error message

chrisngabp:
The value of attributesWithArgumentsAlwaysOnNewLine is true by default. But the current error is not intuitive enough.

This is an improvement that came from this issue:
https://github.com/realm/SwiftLint/issues/5103

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Style/AttributesRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/AttributesRule.swift
@@ -9,7 +9,7 @@ struct AttributesRule: SwiftSyntaxRule, OptInRule, ConfigurationProviderRule {
         description: """
             Attributes should be on their own lines in functions and types, but on the same line as variables and \
             imports except when they have properties and attributes_with_arguments_always_on_line_above is true \
-            which is the default.
+            which is the default
             """,
         kind: .style,
         nonTriggeringExamples: AttributesRuleExamples.nonTriggeringExamples,
--- a/Source/SwiftLintBuiltInRules/Rules/Style/AttributesRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/AttributesRule.swift
@@ -52,7 +52,10 @@ private extension AttributesRule {
 
             switch hasViolation {
             case .argumentsAlwaysOnNewLineViolation:
-                let reason = "Attributes with arguments or inside always_on_line_above must be on a new line instead of the same line"
+                let reason = """
+                    Attributes with arguments or inside always_on_line_above must be on a new line \
+                    instead of the same line
+                    """
 
                 violations.append(
                     ReasonedRuleViolation(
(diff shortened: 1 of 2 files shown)


Source/SwiftLintBuiltInRules/Rules/Style/AttributesRule.swift
```diff
@@ -8,7 +8,8 @@ struct AttributesRule: SwiftSyntaxRule, OptInRule, ConfigurationProviderRule {
         name: "Attributes",
         description: """
             Attributes should be on their own lines in functions and types, but on the same line as variables and \
-            imports
+            imports except when they have properties and attributes_with_arguments_always_on_line_above is true \
+            which is the default
```
SimplyDanny, line 12: This seems a bit too long for a violation message. Better to have a dedicated message for the cases where the option comes into play.

chrisngabp: Does the codebase have an example that I can reference to? I tried to do something like that on my first attempt but I saw the message is a static let and the other rules I saw didn't have something like that. I totally agree it'd be the best option!

SimplyDanny: This list of violations in a rule either accepts a violation with its own message and severity or just a position in which case the rule's description is taken as the message.

An example would be `FunctionParameterCountRule`. In line 76, a `ReasonedRuleViolation` is appended to the list of violations.

chrisngabp: @SimplyDanny thanks for the help! I just pushed some changes with your proposed approach and added some tests. Let me know if it looks good!

Tests/SwiftLintFrameworkTests/AttributesRuleTests.swift
```diff
@@ -95,4 +95,36 @@ class AttributesRuleTests: SwiftLintTestCase {
...
+        verifyRule(argumentsAlwaysOnLineDescription,
+                   ruleConfiguration: ["attributes_with_arguments_always_on_line_above": false])
+    }
+
+    func testAttributesWithArgumentsAlwaysOnLineAboveTrue() {
+        let nonTriggeringExamples = [
+            Example("@Environment(\\.presentationMode)\nprivate var presentationMode")
+        ]
+        let triggeringExamples = [
+            Example("@Environment(\\.presentationMode) private var presentationMode")
```
SimplyDanny, line 120: There should be violation markers `↓` in the non-triggering examples.

SimplyDanny (approved):
Looks good! Just one nit ...

Tests/SwiftLintFrameworkTests/AttributesRuleTests.swift
chrisngabp: Cool! I didn't realize I didn't put them on my tests. Done!

SimplyDanny (approved):

Issue #4898, opened 2023-04-16
Feature Request: Add a configure subcommand to SwiftLint


mildm8nnered:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

The requested `configure` command is slightly akin to `swift-format`'s [`dump-configuration`](https://github.com/apple/swift-format#viewing-the-default-configuration) command, or `Periphery`'s [`--setup`](https://github.com/peripheryapp/periphery#the-scan-command) flag.

When run, `configure` will construct an `.swiftlint.yml` file. By default, `configure` would run interactively.

Additional command line options would allow `configure` to run non-interactively, accepting all the recommended defaults.

Additional command line options would allow `configure`'s behaviour to be modifier still further - e.g., a threshold value, indicating that any rules with more then X violations should be disabled.

Additional command line options would allow `configure` to modify an existing configuration file.

A shell script follows in a comment which provides a proof-of-concept implementation of what `configure` might look like.

Additional features that could be added could be detection of files that should likely be excluded: files under `Pods`, or where `generated` or `build` appears in the path name, for example, and allowing the user to select a reporter.

Sample output when it's run at the top level of SwiftLint's source tree follows below:

```
% ./configure_swiftlint.sh
Welcome to the SwiftLint configuration assistant. Press any key to begin.
Checking for existing configuration files...
Found an existing configuration in .swiftlint.yml
Do you wish to continue? (Y/n) 
Found other configuration files:

./Tests/SwiftLintFrameworkTests/Resources/ProjectMock/NestedConfig/Test/.swiftlint.yml
./Tests/SwiftLintFrameworkTests/Resources/ProjectMock/NestedConfig/Test/Sub/.swiftlint.yml
./Tests/SwiftLintFrameworkTests/Resources/ProjectMock/ChildConfig/Cycle1/.swiftlint.yml
./Tests/SwiftLintFrameworkTests/Resources/ProjectMock/ChildConfig/Cycle2/.swiftlint.yml
./Tests/SwiftLintFrameworkTests/Resources/ProjectMock/ChildConfig/Cycle3/Main/.swiftlint.yml
./Tests/SwiftLintFrameworkTests/Resources/ProjectMock/ParentConfig/Cycle1/.swiftlint.yml
[...]

PR #5113, opened 2023-07-09
Print YAML configuration in output of `rules` command

SimplyDanny:
I wonder if this is actually worth a separate `swiftlint config [--default] <rule>` command? 🤔

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,10 @@
   [keith](https://github.com/keith)
   [5139](https://github.com/realm/SwiftLint/pull/5139)
 
+* Show a rule's active YAML configuration in output of 
+  `swiftlint rules <rule>`.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+
 #### Bug Fixes
 
 * Fix false positive in `control_statement` rule that triggered on conditions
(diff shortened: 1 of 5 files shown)


mildm8nnered:
> I wonder if this is actually worth a separate `swiftlint config [--default] <rule>` command? 🤔

So not arguing against the concept, but I actually had my eye (but not my heart set) on `configure` as a new command for a Periphery style configuration wizard - see #4898 and #5041 - happy to pick an alternative, but if we had `config` and `configure`, that would potentially be a bit confusing.

SimplyDanny:
> > I wonder if this is actually worth a separate `swiftlint config [--default] <rule>` command? 🤔
> 
> So not arguing against the concept, but I actually had my eye (but not my heart set) on `configure` as a new command for a Periphery style configuration wizard - see #4898 and #5041 - happy to pick an alternative, but if we had `config` and `configure`, that would potentially be a bit confusing.

Good point. My idea is to offer a way to append a default configuration for a rule to the `.swiftlint.yml` file like `swiftlint config <rule> >> .swiftlint.yml`. However, I also see this could be added to the `rules` command by adding two new options to it: 

* `--config-only` to print only the rule's configuration as YAML
* `--default` to use the rule's default configuration (instead of what's configured) wherever it is printed

I think the `--default` also makes sense without the `--config-only` option for this command.

PR #5119, opened 2023-07-14
Ignore initializers of computed properties in `prefer_self_in_static_references` rule

### LINKED ISSUE #5118: False positive for rule `prefer_self_in_static_references`
### New Issue Checklist

### Describe the bug

The following is a triggering example with this rule enabled, yet it results in a compiler error: 
`error: covariant ‘Self’ type cannot be referenced from a stored property initializer`

```
class A {
        static let start: Int = 0

        private(set) var current: Int = A.start {
             didSet {
                 self.foo()
             }
        }
}
```
This will incorrectly convert `private(set) var current: Int = A.start {` to `private(set) var current: Int = Self.start {`

However this rule works correctly and doesn't trigger without the `didSet` block:
```
class A {
        static let start: Int = 0
        private(set) var current: Int = A.start
}
```

### Environment

* SwiftLint version: 0.52.4
* Which Xcode version are you using: Xcode 14.3

SimplyDanny:
Fixes #5118.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,12 @@
 
 #### Bug Fixes
 
+* Fix bug in `prefer_self_in_static_references` rule that triggered on
+  initializers of computed properties in classes when the property had an
+  accessor block.  
+  [SimplyDanny](https://github.com/SimplyDanny)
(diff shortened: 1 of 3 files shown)


pm-dev:
Thanks!

PR #5121, opened 2023-07-15
Fixed false positives for nested structs in classes

### LINKED ISSUE #5120: unneeded_synthesized_initializer rule does not detect violations in a struct nested inside a class
### New Issue Checklist

### Describe the bug

Given

```
class Foo {
    struct Bar {
        let baz: Int

        init(baz: Int) {
            self.baz = baz
        }
    }
}
```

`unneeded_synthesized_initializer` will fail to detect the violation here.

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.52.4
* Installation method used (Homebrew, CocoaPods, building from source, etc)? CocoaPods

14.2

See above

mildm8nnered:
Resolves #5120

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,11 @@
   [JP Simard](https://github.com/jpsim)
   [#4737](https://github.com/realm/SwiftLint/issues/4737)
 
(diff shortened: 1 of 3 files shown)


SimplyDanny (approved):
Thank you!

PR #5122, opened 2023-07-15
Hide reporters from public interface

SimplyDanny:
Or is there a good reason to expose them?


mildm8nnered:
So I have a use case where I want to use the Summary reporter

> Or is there a good reason to expose them?

So I have a use case in https://github.com/realm/SwiftLint/pull/5041, where I'd like to use the Summary reporter as part of a configuration wizard `configure` command, to show the user what violations are currently present in their tree ...

SimplyDanny:
> So I have a use case in #5041, where I'd like to use the Summary reporter as part of a configuration wizard `configure` command, to show the user what violations are currently present in their tree ...

You are free to make some of them public again as needed. Most reporters don't need to be publicly visible, though. `reporterFrom(identifier:)` is the only interface that must be accessible. So having the implementations `internal` by default seems more logical.

PR #5131, opened 2023-07-17
Update README_CN.md

Jerry23011:
I got mislead by the Chinese README at first because I'm running on an Apple silicon. Hope this update can help others.


SimplyDanny (approved):
Thank you for the update!

PR #5139, opened 2023-07-21
Add unneeded_override rule

keith:
This rule flags functions where the only thing they do is call their
super function and therefore could be omitted. For example:

```swift
override func foo() {
    super.foo()
}
```

This can get pretty complex since there are a lot of slight variations
the subclasses' functions can call the superclasses' functions with, but
this covers many of the cases. Ideally this would handle variable
overrides too but it doesn't currently.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/UnneededOverrideRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/UnneededOverrideRule.swift
@@ -1,0 +1,130 @@
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+struct UnneededOverrideRule: ConfigurationProviderRule, SwiftSyntaxCorrectableRule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "unneeded_override",
+        name: "Unneeded Overridden Functions",
+        description: "Remove overridden functions that don't do anything except call their super",
+        kind: .lint,
+        nonTriggeringExamples: UnneededOverrideRuleExamples.nonTriggeringExamples,
+        triggeringExamples: UnneededOverrideRuleExamples.triggeringExamples,
+        corrections: UnneededOverrideRuleExamples.corrections
+    )
+
+    func makeVisitor(file: SwiftLintFile) -> ViolationsSyntaxVisitor {
+        Visitor(viewMode: .sourceAccurate)
+    }
+
+    func makeRewriter(file: SwiftLintFile) -> ViolationsSyntaxRewriter? {
+        Rewriter(
+            locationConverter: file.locationConverter,
+            disabledRegions: disabledRegions(file: file)
+        )
+    }
+}
+
+private func simplify(_ node: SyntaxProtocol) -> ExprSyntaxProtocol? {
+    if let expr = node.as(AwaitExprSyntax.self) {
+        return expr.expression
+    } else if let expr = node.as(TryExprSyntax.self) {
+        // Assume using try! / try? changes behavior
+        if expr.questionOrExclamationMark != nil {
+            return nil
+        }
+
+        return expr.expression
+    } else if let expr = node.as(FunctionCallExprSyntax.self) {
+        return expr
+    } else if let stmt = node.as(ReturnStmtSyntax.self) {
+        return stmt.expression
+    }
+
+    return nil
+}
+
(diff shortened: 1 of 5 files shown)


Source/SwiftLintBuiltInRules/Rules/Lint/UnneededOverrideRule.swift
```diff
@@ -0,0 +1,89 @@
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+struct UnneededOverrideRule: ConfigurationProviderRule, SwiftSyntaxRule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "unneeded_override",
+        name: "Remove empty overridden functions",
```
SimplyDanny, line 9: This should be more like a title, e.g. "Unneeded Overridden Functions".

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededOverrideRule.swift
```diff
@@ -0,0 +1,89 @@
...
+    } else if let stmt = node.as(ReturnStmtSyntax.self) {
+        return stmt.expression
+    }
+
+    return nil
+}
+
+private final class Visitor: ViolationsSyntaxVisitor {
+    override func visitPost(_ node: FunctionDeclSyntax) {
+        guard let modifiers = node.modifiers, modifiers.contains(where: { $0.name.text == "override" }),
```
SimplyDanny, line 42:
```suggestion
        guard let modifiers = node.modifiers, modifiers.containsOverride,
```

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededOverrideRule.swift
```diff
@@ -0,0 +1,89 @@
...
+            node.body?.statements.count == 1, let statement = node.body?.statements.first else {
```
SimplyDanny, line 43:
```suggestion
            let statement = node.body?.statements.onlyElement else {
```

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededOverrideRule.swift
```diff
@@ -0,0 +1,89 @@
...
+        // Extract the function call from other expressions like try / await / return.
+        // If this returns a non-super calling function that will get filtered out later
+        var syntax: ExprSyntaxProtocol? = simplify(statement.item)
+        while let nestedSyntax = syntax {
+            if nestedSyntax.as(FunctionCallExprSyntax.self) != nil {
+                break
+            }
+
+            syntax = simplify(nestedSyntax)
+        }
```
SimplyDanny, line 61: Maybe extract this into a function to avoid casting to `FunctionCallExprSyntax` twice.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededOverrideRule.swift
```diff
@@ -0,0 +1,89 @@
...
+
+        let overridenFunctionName = node.identifier.text
+        guard let call = syntax?.as(FunctionCallExprSyntax.self),
+            let member = call.calledExpression.as(MemberAccessExprSyntax.self),
+            member.base?.as(SuperRefExprSyntax.self) != nil,
```
SimplyDanny, line 66:
```suggestion
            member.base?.is(SuperRefExprSyntax.self) == true,
```

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededOverrideRuleExamples.swift
```diff
@@ -0,0 +1,164 @@
+struct UnneededOverrideRuleExamples {
+    static let nonTriggeringExamples = [
+        Example("""
+class Foo {
+    override func bar() {
+        super.bar()
+        print("hi")
+    }
+}
+"""),
```
SimplyDanny, line 10: You may indent the snippets to be aligned with the `Example`.

SimplyDanny (commented):
I like this!

Did you think about adding a rewriter removing the superfluous overrides when running in `--fix` mode?

Also, don't forget the CHANGELOG entry. 😉

keith:
Thanks for the review! I implemented fixing

SimplyDanny (approved):
All triggering examples miss the violation marker `↓`.

Other than that, it seems to be good to go.

PR #5142, opened 2023-07-25
Rules.Idomatic: update switch for new `macro` enumerator

compnerd:
The introduction of the `macro` enumerator in the enumeration causes a build break due to an uncovered switch.  Conservatively treat the macros as being exempt from the ACLing.


SimplyDanny:
I think this is no longer required with the SwiftSyntax rewrite in #5382.

compnerd:
@SimplyDanny sounds good!  BTW, I wouldn't mind some early feedback on https://github.com/realm/SwiftLint/pull/5030 (I'm happy to split that into separate PRs, but want to gauge the interest and any easy things that can be addressed up front).

PR #5144, opened 2023-07-25
Indentation style

mredig:
Add rule indentation_style. Warns when tabs and spaces are mixed within the same file and allows for setting your entire project to either tabs OR spaces. Also allows for spaces to finesse placement at the end of a tabbed indentation.

I'll admit that I'm a little apprehensive about this PR because I feel like this feature should somehow already be here, but I cannot figure out how to use it if it is. Either way, I hope it's helpful!


Tests/GeneratedTests/GeneratedTests.swift
```diff
@@ -518,6 +518,12 @@ class InclusiveLanguageRuleGeneratedTests: SwiftLintTestCase {
     }
 }
 
+class IndentationStyleRuleGeneratedTests: SwiftLintTestCase {
```
Harry-KNIGHT, line 521:
```suggestion
final class IndentationStyleRuleGeneratedTests: SwiftLintTestCase {
```

Harry-KNIGHT:
Can you add code example of your PR please ?

mredig:
Sorry, I don't know how this slipped through my radar. I didn't notice a response until now.

> Can you add code example of your PR please ?

I'm not sure I understand what you mean. Are you asking for something like the [rule description](https://github.com/realm/SwiftLint/blob/b643b48490ecf713535decdd1425c0917525d7b9/Source/SwiftLintBuiltInRules/Rules/Style/IndentationStyleRule.swift)?

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/IndentationStyleConfiguration.swift
```diff
@@ -0,0 +1,61 @@
...
+	typealias Parent = IndentationStyleRule
+
+	@ConfigurationElement(key: "severity")
+	private(set) var severityConfiguration = SeverityConfiguration<Parent>.warning
+	@ConfigurationElement(key: "include_multiline_strings")
+	private(set) var includeMultilineStrings = false
+	@ConfigurationElement(key: "include_multiline_comments")
+	private(set) var includeMultilineComments = true
+	@ConfigurationElement(key: "per_file")
+	private(set) var perFile = true
```
SimplyDanny, line 13: Not sure about this option. I tend to keep it simple for the first and only have a project-wide preferred setting instead of individual variants per file.

Source/SwiftLintBuiltInRules/Rules/Style/IndentationStyleRule.swift
```diff
@@ -0,0 +1,385 @@
...
+            } else {
+                switch firstLineIndentation {
+                case " ":
+                    fileStyle = .spaces
+                    confirmedFileStyle = .spaces
+                case "\t":
+                    fileStyle = .tabs
+                    confirmedFileStyle = .tabs
+                default:
+                    queuedFatalError(
```
SimplyDanny, line 227: Prefer to report another violation instead of crashing.

Source/SwiftLintBuiltInRules/Rules/Style/IndentationStyleRule.swift
```diff
@@ -0,0 +1,385 @@
...
+    }
+
+    // MARK: - Properties
+    var configuration = IndentationStyleConfiguration()
+
+    static let description = RuleDescription(
+        identifier: "indentation_style",
+        name: "Indentation Style",
+        description: """
+            Indent code using either tabs or spaces, not a chaotic mix. Can be configured to per file or project wide.
```
SimplyDanny, line 18: I had to laugh about the "chaotic mix". Yet, I think it's a bit too offensive for a documentation. Better leave it out. 😉 

```suggestion
            Code should be indented using either tabs or spaces. Can be configured to per file or project wide.
```

SimplyDanny (commented):
First a few general remarks and questions. Once they are clarified, I may check the actual implementation.

Source/SwiftLintBuiltInRules/Rules/Style/IndentationStyleRule.swift
```diff
@@ -0,0 +1,385 @@
...
+    /// or
+    /// ```swift
+    /// let foo = [1,
+    ///            2,
+    ///            3]
+    /// ```
+    ///
+    /// These don't necessarily line up nicely on the 2,3,4,8 space indentation or whatever tab width you have set.
+    private
+    func previousLineSetsNonstandardIndentationExpectation(_ currentLine: Line, in file: SwiftLintFile) -> Bool {
```
Harry-KNIGHT, line 318:
```suggestion
    private func previousLineSetsNonstandardIndentationExpectation(_ currentLine: Line, in file: SwiftLintFile) -> Bool {
```

Source/SwiftLintBuiltInRules/Rules/Style/IndentationStyleRule.swift
mredig: > I had to smile when I read the "chaotic mix"

Mission accomplished :) I can now close this PR!

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/IndentationStyleConfiguration.swift
```diff
@@ -0,0 +1,61 @@
...
+
+struct IndentationStyleConfiguration: SeverityBasedRuleConfiguration, Equatable {
+	typealias Parent = IndentationStyleRule
+
+	@ConfigurationElement(key: "severity")
+	private(set) var severityConfiguration = SeverityConfiguration<Parent>.warning
+	@ConfigurationElement(key: "include_multiline_strings")
+	private(set) var includeMultilineStrings = false
+	@ConfigurationElement(key: "include_multiline_comments")
+	private(set) var includeMultilineComments = true
```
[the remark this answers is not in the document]
mredig: If I recall correctly, it's because Xcode automatically indents multiline comments content with an additional space, even if you use tabs for indentation. As a tabs evangelist, even I don't always care to fix that since the smallest tweak can cause Xcode to reset that.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/IndentationStyleConfiguration.swift
```diff
@@ -0,0 +1,61 @@
...
+	@ConfigurationElement(key: "per_file")
+	private(set) var perFile = true
+	@ConfigurationElement(key: "preferred_style")
+	private(set) var preferredStyle = PreferredStyle.spaces
+	@ConfigurationElement(key: "tab_width")
+	private(set) var tabWidth: Int?
```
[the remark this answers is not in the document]
mredig: Maybe it would be better with a different name. This isn't for setting how many spaces a tab should consist of, it's to tell swiftlint how wide your tabs are, since I don't think that can be derived from anywhere...

The reason being, if you are using tab indentation, sometimes you use spaces to provide a little nudging after the tab indentation. This setting allows for no more than `tabWidth - 1` spaces following any tab indentation.

I'll point out that, if you are using tabs in what I consider to be the most optimal way, you don't need this. However, forcing tabs projects to use tabs *my* way goes entirely against the philosophy of tabs usage in the first place, so this setting now exists.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/IndentationStyleConfiguration.swift
```diff
@@ -0,0 +1,61 @@
...
+	enum PreferredStyle: String, AcceptableByConfigurationElement {
+		case tabs
+		case spaces
+
+		func asOption() -> OptionType {
+			.string(rawValue)
+		}
+	}
+
+	mutating func apply(configuration: Any) throws {
```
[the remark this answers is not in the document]
mredig: If I recall correctly, I think it's for this bit:

```swift
if perFile == false {
	if let preferredStyle = configurationDict[$preferredStyle] as? String {
		self.preferredStyle = PreferredStyle(rawValue: preferredStyle) ?? .spaces
	}
}
```

Source/SwiftLintBuiltInRules/Rules/Style/IndentationStyleRule.swift
```diff
@@ -0,0 +1,385 @@
...
+        ]
+    )
+
+    // MARK: - Initializers
+    // MARK: - Methods: Validation
+    func validate(file: SwiftLintFile) -> [StyleViolation] { // swiftlint:disable:this cyclomatic_complexity
+        var violations: [StyleViolation] = []
+
+        var fileStyle: ConfigurationType.PreferredStyle?
+        if configuration.perFile == false {
```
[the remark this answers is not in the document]
mredig: I understand the irony of me arguing any style preferences on this project, but the reason I do this is for cognitive overhead while reading code. I used to always prefer the conciseness of the bang operator for testing conditions, but another programmer shared that they found reading new code much faster when there was a big ol' `false` staring you in the face instead of a tiny little `!`.

I'll go ahead and change it, but I just wanted to share my reasoning :)

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/IndentationStyleConfiguration.swift
mredig: I've found that many projects I've come across have a hodgepodge mix. I included this so migration to project wide settings can be a little more gradual. Not to mention that Xcode allows setting tabs or spaces on a file by file basis, I don't want to cause violations in areas where teams may have already deliberately gone against the grain.

I can remove, but the work is already done!

Source/SwiftLintBuiltInRules/Rules/Style/IndentationStyleRule.swift
mredig: Oooh smart! Blind spot there. Thanks!

Tests/GeneratedTests/GeneratedTests.swift
mredig: I reverted this change because these are autogenerated tests.

mredig:
Hmm - Looks like there've been some substantial changes since this PR was made. Working on getting it to build and pass tests with the latest updates....

mredig:
> 1 Error
> 🚫	Could not build branch
> Generated by 🚫 [Danger](https://danger.systems/)

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/IndentationStyleConfiguration.swift
SimplyDanny: Got that. Perhaps this can be combined with `preferredStyle` in the sense of having `spaces`, `tabs` and `automatic` (or `perFile`) as options?

This would have the pretty side effect, that the `apply` method can be auto-generated.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/IndentationStyleConfiguration.swift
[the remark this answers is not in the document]
SimplyDanny: If trailing spaces are only used to refine the indentation, but otherwise the file's default is respected by Xcode so that either spaces or tabs are used for the main chunk of the indentation, this seems like a normal case that should be supported anyway also in other areas, e.g.

```swift
if cond1,
   cond2 {
    // ...
}
```

wherein `cond2` might be indented by one tab and one space.

In other words, is there any reason to make this optional?

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/IndentationStyleConfiguration.swift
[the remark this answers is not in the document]
SimplyDanny: This would also be required, if the we offer an automatic rewrite for the rule.

But I wonder, what happens if it remains `nil`. Then the rule would have to assume something, wouldn't it? Is it better to specify 2 or 4 as defaults right away then?

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/IndentationStyleConfiguration.swift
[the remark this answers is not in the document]
SimplyDanny: See my other suggestion about extending `PreferredStyle` to circumvent this special case.

PR #5145, opened 2023-07-25
Add no_default_switch_in_operators rule

keith:
This rule validates that `default` is not used in switch cases in
operator definitions. This could be expanded to do more than operator
definitions but operators feel a bit special in that forgetting to add
your case to them as you add new cases probably leads to more subtle
bugs than other areas.

### DIFF
--- a/Source/SwiftLintCore/Models/Location.swift
+++ b/Source/SwiftLintCore/Models/Location.swift
@@ -98,7 +98,7 @@ private extension Optional where Wrapped: Comparable {
             return lhs < rhs
         case (nil, _?):
             return true
-        default:
+        case (_, _):
             return false
         }
     }
--- a/Source/SwiftLintBuiltInRules/Models/BuiltInRules.swift
+++ b/Source/SwiftLintBuiltInRules/Models/BuiltInRules.swift
@@ -125,6 +125,7 @@ public let builtInRules: [Rule.Type] = [
     NSObjectPreferIsEqualRule.self,
     NestingRule.self,
     NimbleOperatorRule.self,
+    NoDefaultSwitchInOperatorsRule.self,
     NoExtensionAccessModifierRule.self,
     NoFallthroughOnlyRule.self,
     NoGroupingExtensionRule.self,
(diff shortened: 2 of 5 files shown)


keith:
this rule might be a bit too specific, happy to just keep this in our internal configuration if folks think so

ryandemo (approved):

SimplyDanny:
There are indeed a lot of findings in OSS projects. However, I'd like to understand why the rule is useful. What's wrong about default cases? What if the `switch` statement doesn't have to do anything with the operator's arguments? 

I actually suspected this rule to have a high rate of false positives. Looking into the OSS findings it hasn't, however. 🤔

keith:
> However, I'd like to understand why the rule is useful. What's wrong about default cases?

Yea I think one could argue against using `default` cases in general as a benefit so that when you add a new case to any enum you are required to reconsider each place where it is `switch`ed and decide how your new case should behave. The reason we discovered this internally was because we saw some very subtle cases of this where folks specifically forgot to update the switch in the equatable definition. This one can be a bit more subtle just because they also might assume the Equatable conformance is auto-generated. 

> What if the switch statement doesn't have to do anything with the operator's arguments?

Yea this would definitely be a false positive, but I don't see any spot checking this list since I guess that's probably unlikely.

This is definitely an opinionated rule so if you think we should just keep this internally that's fine with us.

SimplyDanny:
> This is definitely an opinionated rule so if you think we should just keep this internally that's fine with us.

It's hard to argue with a high false positive rate if there seemingly weren't any violations found in the OSS scan. I also see that it can have a value. Yet, it still seems to be of limited and quite specific use only. So I rather tend to not have it as a built-in rule.

Before we close this request, let's hear @jpsim's opinion.

keith:
(also note it is an opt in rule)

Source/SwiftLintCore/Models/Location.swift
```diff
@@ -98,7 +98,7 @@ private extension Optional where Wrapped: Comparable {
             return lhs < rhs
         case (nil, _?):
             return true
-        default:
+        case (_, _):
```
Harry-KNIGHT, line 101: Why use this instead of default ?

PR #5149, opened 2023-07-31
Disable Violation Test Inline Error Highlighting

mredig:
When creating examples, most test failures highlight *in* the example code, which is very helpful. However, if a rule fails to enable or disable according to `// swiftlint:disable` commands, the test failure unhelpfully just highlights in the generic unit test code.

This PR just refactors the location of the error highlighting to point to the failing example, just as the other test failures do.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,10 +33,6 @@
   [keith](https://github.com/keith)
   [5139](https://github.com/realm/SwiftLint/pull/5139)
 
-* Tests now highlight disable and superfluous failures inline in example code  
-  [mredig](https://github.com/mredig)
-  [5149](https://github.com/realm/SwiftLint/pull/5149)
-
 #### Bug Fixes
 
 * Fix false positive in `control_statement` rule that triggered on conditions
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -33,6 +33,10 @@
   [keith](https://github.com/keith)
   [5139](https://github.com/realm/SwiftLint/pull/5139)
 
+* Tests now highlight disable and superfluous failures inline in example code  
+  [mredig](https://github.com/mredig)
+  [5149](https://github.com/realm/SwiftLint/pull/5149)
```
SimplyDanny, line 38: An entry in the CHANGELOG is not necessary as this is nothing user-visible.

SimplyDanny (approved):
Very useful. Thank you!

CHANGELOG.md
mredig: Ah - sounds good. I'll get that updated soon.

mredig: d u n Done!

PR #5154, opened 2023-08-03
Allow to initialize `Example`s by string literals

SimplyDanny:
This seems so obvious. Why didn't we do this earlier? 😅

I only converted a few example sets as a starting point to show this works well.


marcelofabri:
This breaks the highlighting that we had on the failed example:

Screenshot 2023-08-03 at 14 15 02

SimplyDanny:
Indeed. Didn't think about that ... Can you think of a way to make this work?

PR #5170, opened 2023-08-11
Added bare_init rule

### LINKED ISSUE #2960: Rule Request: [explicit_init] - avoid using `let cl: Class = .init()`
### New Issue Checklist

### New rule request

Improve rule `explicit_init` to force user to use `let cl = Class()` instead of `let cl: Class = .init()`

> Code is more readable when you initialize with class name instead of .init() only

> NO:
```
let view: UIView = .init(frame: .init(origin: .init(x: 0, y: 0), size: .init(width: 50, height: 50)))
```
> YES:
```
let view = UIView(frame: CGRect(origin: CGPoint(x: 0, y: 0), size: CGSize(width: 50, height: 50)))
```

> I guess no.

> I guess it should be enabled by default.

mildm8nnered:
Potentially resolves  #2960

I went for `bare_init`, as we already have an `explicit_init` rule.

Not massively happy with the name, so if anyone has a better suggestion ...

I did think of just adding a configuration option to `explicit_init` to extend its behavior, but then I would have needed another configuration flag to turn the rules normal behavior off, and a separate rule seemed easier.

Triggers a lot of violations in the open source samples, and in SwiftLint itself, as this is a very common construct, so I made it opt-in.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/BareInitRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/BareInitRule.swift
@@ -43,8 +43,8 @@ private extension BareInitRule {
 
 private extension MemberAccessExprSyntax {
     var bareInitPosition: AbsolutePosition? {
-        if base == nil, name.text == "init" {
-            return dot.positionAfterSkippingLeadingTrivia
+        if base == nil, declName.baseName.text == "init" {
+            return period.positionAfterSkippingLeadingTrivia
         } else {
             return nil
         }
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/BareInitRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/BareInitRule.swift
@@ -7,11 +7,12 @@ struct BareInitRule: ConfigurationProviderRule, SwiftSyntaxRule, OptInRule {
     static let description = RuleDescription(
         identifier: "bare_init",
         name: "Bare Init",
-        description: "Use named constructors over .init",
+        description: "Prefer named constructors over .init and type inference",
         kind: .idiomatic,
         nonTriggeringExamples: [
             Example("let foo = Foo()"),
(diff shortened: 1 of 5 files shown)


Source/SwiftLintBuiltInRules/Rules/Idiomatic/BareInitRule.swift
```diff
@@ -0,0 +1,52 @@
+import SwiftSyntax
+import SwiftSyntaxBuilder
```
SimplyDanny, line 2: This import isn't need, is it?

Source/SwiftLintBuiltInRules/Rules/Idiomatic/BareInitRule.swift
```diff
@@ -0,0 +1,52 @@
...
+                return
+            }
+
+            violations.append(violationPosition)
+        }
+    }
+}
+
+private extension MemberAccessExprSyntax {
+    var bareInitPosition: AbsolutePosition? {
```
SimplyDanny, line 45: Perhaps inline this method. I don't see a real benefit to have it separated.

SimplyDanny (approved):
Looks good!

SimplyDanny:
> I did think of just adding a configuration option to `explicit_init` to extend its behavior, but then I would have needed another configuration flag to turn the rules normal behavior off, and a separate rule seemed easier.

Thinking again about this ... Both rules are actually very similar, so I tend to like a configuration flag more. The rewriter of `explicit_init` would just do nothing for `.init(...)`.

mildm8nnered:
> > I did think of just adding a configuration option to `explicit_init` to extend its behavior, but then I would have needed another configuration flag to turn the rules normal behavior off, and a separate rule seemed easier.
> 
> Thinking again about this ... Both rules are actually very similar, so I tend to like a configuration flag more. The rewriter of `explicit_init` would just do nothing for `.init(...)`.

I've tried implementing the option in #5203 ...

PR #5174, opened 2023-08-13
Ignore the second operands of binary left shift operations

### LINKED ISSUE #5171: no_magic_number - Ignore numbers involved in bit-wise operations (like << or >>  or | or &)
### New Issue Checklist

### New rule request

(I debated if this is technically a bug or request, but it's definitely not a "new" rule, more of just improving an existing rule.)

What I'm running into is the "no_magic_number" rule triggering on bitwise structs or enums that we have in our code - 

    struct LogType: OptionSet {
        /// Returns the raw bitmask value of the option and satisfies the `RawRepresentable` protocol.
        public let rawValue: UInt
    
        /// Logging to the operating system
        public static let operatingSystem: LogType= LogType(rawValue: 1 << 0)
    
        /// Logging to the debug console
        public static let debugConsole: LogType= LogType(rawValue: 1 << 1)
    
        /// Logging to a file
        public static let file: LogType= LogType(rawValue: 1 << 2) // swiftlint:disable:this no_magic_numbers
    
        /// Storing Logs in memory
        public static let inMemory: LogType= LogType(rawValue: 1 << 3) // swiftlint:disable:this no_magic_numbers
    
        /// Testing logger
        public static let testing: LogType= ulogLogType(rawValue: 1 << 4) // swiftlint:disable:this no_magic_numbers
    
        /// Visually presents logs
        public static let visual: LogType= LogType(rawValue: 1 << 5) // swiftlint:disable:this no_magic_numbers
    
        /// The log event is reporting to another delegate/system
        public static let reporting: LogType= LogType(rawValue: 1 << 6) // swiftlint:disable:this no_magic_numbers
    
        /// Other (bucket for any 'other' type of logging not applicable to any of the existing types
        public static let other: LogType= LogType(rawValue: 1 << 7) // swiftlint:disable:this no_magic_numbers
    
        /**
         Creates a new instance of the options set from the specified raw value
         */
        public init(rawValue: UInt) {
            self.rawValue = rawValue
        }
    }

Obviously, the bit-wise left shift operators appear to have "magic numbers", but being part of a left (or right shift) operation kind of excludes it from being "magic".

To make it helpful, could enforce that the number is less than 32 (or 64) (or whatever number of bits are in that specified storage type), depending on how smart the rule can be.

An additional idea would be to just make it a configuration point on the no_magic_number rule. Along the lines of "ignore_bitwise_operations: true"

Hopefully that makes sense. As you can see from the code sample...it's not the hardest thing in the world to just disable the rule for the lines where it applies. But it makes the code messier than it needs to be for cases that (I think) are universally okay for having "magic numbers" (as long as they are less than 64).

mildm8nnered:
Resolves #5171, at least in part.

```
struct LogType: OptionSet {
    /// Returns the raw bitmask value of the option and satisfies the `RawRepresentable` protocol.
    public let rawValue: UInt
    public static let operatingSystem: LogType= LogType(rawValue: 1 << 0)
    public static let debugConsole: LogType= LogType(rawValue: 1 << 1)
    public static let file: LogType= LogType(rawValue: 1 << 2)
    public static let inMemory: LogType= LogType(rawValue: 1 << 3)
    public static let testing: LogType= ulogLogType(rawValue: 1 << 4)
    public static let visual: LogType= LogType(rawValue: 1 << 5)
    public static let reporting: LogType= LogType(rawValue: 1 << 6)
    public static let other: LogType= LogType(rawValue: 1 << 7)

    public init(rawValue: UInt) {
        self.rawValue = rawValue
    }
}
```

Will no longer generate any warnings. Only the left shift operator will receive this treatment, and only for its second operand.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,10 +103,7 @@
   when calling `NSNumber.init(value:)` directly.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#5172](https://github.com/realm/SwiftLint/issues/5172)
-* The `no_magic_numbers` rule will not trigger for the second operand of
-  bitwise left shift operations.  
-* The `no_magic_numbers` rule will not trigger for the either operand of
-  bitwise shift operations.  
+
 * The `no_magic_numbers` rule will not trigger for bitwise shift
   operations.  
   [Martin Redington](https://github.com/mildm8nnered)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,6 +107,8 @@
   bitwise left shift operations.  
 * The `no_magic_numbers` rule will not trigger for the either operand of
   bitwise shift operations.  
+* The `no_magic_numbers` rule will not trigger for bitwise shift
+  operations.  
   [Martin Redington](https://github.com/mildm8nnered)
   [#5171](https://github.com/realm/SwiftLint/issues/5171)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,6 +103,10 @@
   when calling `NSNumber.init(value:)` directly.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#5172](https://github.com/realm/SwiftLint/issues/5172)
+* The `no_magic_numbers` rule will not trigger for the second operand of
+  bitwise left shift operations.  
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#5171](https://github.com/realm/SwiftLint/issues/5171)
 
 ## 0.52.4: Lid Switch
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,8 @@
   [#5172](https://github.com/realm/SwiftLint/issues/5172)
 * The `no_magic_numbers` rule will not trigger for the second operand of
   bitwise left shift operations.  
+* The `no_magic_numbers` rule will not trigger for the either operand of
+  bitwise shift operations.  
   [Martin Redington](https://github.com/mildm8nnered)
   [#5171](https://github.com/realm/SwiftLint/issues/5171)
 
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
@@ -213,9 +213,8 @@ private extension ExprSyntaxProtocol {
         }
 
         let operatorIndex = siblings.index(after: siblings.startIndex)
-        if let tokenKind = siblings[operatorIndex].as(BinaryOperatorExprSyntax.self)?.operatorToken.tokenKind,
-           tokenKind == .binaryOperator("<<") || tokenKind == .binaryOperator(">>") {
-            return true
+        if let tokenKind = siblings[operatorIndex].as(BinaryOperatorExprSyntax.self)?.operatorToken.tokenKind {
+            return tokenKind == .binaryOperator("<<") || tokenKind == .binaryOperator(">>")
         }
 
         return false
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
@@ -205,20 +205,17 @@ private extension ExprSyntaxProtocol {
     }
 
     func isOperandOfBitwiseShiftOperation() -> Bool {
-        guard let siblings = parent?.as(ExprListSyntax.self)?.children(viewMode: .sourceAccurate) else {
+        guard
+            let siblings = parent?.as(ExprListSyntax.self)?.children(viewMode: .sourceAccurate),
+            siblings.count == 3
+        else {
             return false
         }
-        if siblings.count == 3 {
-            let index = siblings.index(after: siblings.startIndex)
-            if let tokenKind = siblings[index].as(BinaryOperatorExprSyntax.self)?.operatorToken.tokenKind {
-                if tokenKind == .binaryOperator("<<") || tokenKind == .binaryOperator(">>") {
-                    let lastIndex = siblings.index(after: index)
-                    let selfAsSyntax = self.as(Syntax.self)
-                    if siblings[siblings.startIndex] == selfAsSyntax || siblings[lastIndex] == selfAsSyntax {
-                        return true
-                    }
-                }
-            }
+
+        let operatorIndex = siblings.index(after: siblings.startIndex)
+        if let tokenKind = siblings[operatorIndex].as(BinaryOperatorExprSyntax.self)?.operatorToken.tokenKind,
+           tokenKind == .binaryOperator("<<") || tokenKind == .binaryOperator(">>") {
+            return true
         }
 
         return false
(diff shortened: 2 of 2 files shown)


mildm8nnered:
Seems to be failing CI for reasons unrelated to this branch.

SimplyDanny:
> Seems to be failing CI for reasons unrelated to this branch.

Yes, the GitHub token expired. It will hopefully be updated shortly.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -196,4 +202,22 @@ private extension ExprSyntaxProtocol {
         }
         return nil
     }
+
+    func isSecondOperandOfLeftShiftOperation() -> Bool {
```
LowAmmo, line 206: Any reason to not just do Right sift, also?

Rename method to "isSecondOperandOfShiftOperation", and look for ">>" also?

mildm8nnered: ok - so now we'll allow **either** operand of a bitwise shift operation to be magic

CHANGELOG.md
```diff
@@ -78,6 +78,11 @@
   [Martin Redington](https://github.com/mildm8nnered)
   [#5137](https://github.com/realm/SwiftLint/issues/5137)
 
+* The `no_magic_numbers` rule will not trigger for the either operand of
```
LowAmmo, line 81: nit: "...rule will not trigger for bitwise shift operations"

LowAmmo (approved):

CHANGELOG.md
mildm8nnered: nice catch - thanks! I've cleaned it up

LowAmmo (approved):

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -196,4 +203,24 @@ private extension ExprSyntaxProtocol {
...
+        guard let siblings = parent?.as(ExprListSyntax.self)?.children(viewMode: .sourceAccurate) else {
+            return false
+        }
+        if siblings.count == 3 {
+            let index = siblings.index(after: siblings.startIndex)
+            if let tokenKind = siblings[index].as(BinaryOperatorExprSyntax.self)?.operatorToken.tokenKind {
+                if tokenKind == .binaryOperator("<<") || tokenKind == .binaryOperator(">>") {
+                    let lastIndex = siblings.index(after: index)
+                    let selfAsSyntax = self.as(Syntax.self)
+                    if siblings[siblings.startIndex] == selfAsSyntax || siblings[lastIndex] == selfAsSyntax {
```
SimplyDanny, line 217: I cannot think of an example in which this condition would not be true. Omit it?

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -68,7 +68,11 @@ struct NoMagicNumbersRule: SwiftSyntaxRule, OptInRule, ConfigurationProviderRule
                 let a = Int(3)
             }
             class MyTest: XCTestCase {}
-            """)
+            """),
+            Example("let foo = 1 << 2"),
+            Example("let foo = 1 >> 2"),
+            Example("let foo = 2 >> 2"),
+            Example("let foo = 2 << 2")
```
SimplyDanny, line 75: Negative numbers like in `-2 << 2` and `2 >> -2` would still trigger. Is this intentional out of being a bit questionable?

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -196,4 +203,24 @@ private extension ExprSyntaxProtocol {
         }
         return nil
     }
+
+    func isOperandOfBitwiseShiftOperation() -> Bool {
+        guard let siblings = parent?.as(ExprListSyntax.self)?.children(viewMode: .sourceAccurate) else {
+            return false
+        }
+        if siblings.count == 3 {
```
SimplyDanny, line 211: Add this to the `guard` above to avoid deep nesting.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -196,4 +203,21 @@ private extension ExprSyntaxProtocol {
...
+            let siblings = parent?.as(ExprListSyntax.self)?.children(viewMode: .sourceAccurate),
+            siblings.count == 3
+        else {
+            return false
+        }
+
+        let operatorIndex = siblings.index(after: siblings.startIndex)
+        if let tokenKind = siblings[operatorIndex].as(BinaryOperatorExprSyntax.self)?.operatorToken.tokenKind,
+           tokenKind == .binaryOperator("<<") || tokenKind == .binaryOperator(">>") {
+            return true
```
SimplyDanny, line 218:
```suggestion
        if let tokenKind = siblings[operatorIndex].as(BinaryOperatorExprSyntax.self)?.operatorToken.tokenKind {
            return tokenKind == .binaryOperator("<<") || tokenKind == .binaryOperator(">>")
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
mildm8nnered: So that was not intentional, but I think I could live with it, on the basis that negative numbers are quite a rare case.

What do you think?

SimplyDanny: Negative number used in shift operations can be considered "magic" I think. 😉

SimplyDanny (approved):

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
LowAmmo: Agreed... a negative number in a shift operation seems like it would be an error case, actually...  So, if we warn on it, we're just making the world a better place! 😄

LowAmmo (approved):

CHANGELOG.md
```diff
@@ -103,6 +103,14 @@
   when calling `NSNumber.init(value:)` directly.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#5172](https://github.com/realm/SwiftLint/issues/5172)
+* The `no_magic_numbers` rule will not trigger for the second operand of
```
LowAmmo, line 106: Guessing a merge issue? I think the one at line 110 is the best...

LowAmmo (changes requested):

CHANGELOG.md
mildm8nnered: nice spot! Thanks!

LowAmmo (approved):

SimplyDanny:
Thank you for the discussions, implementation and review to both of you, @LowAmmo and @mildm8nnered!

PR #5175, opened 2023-08-13
Removed `Image(systemName:)` special case

### LINKED ISSUE #561: valid_docs: Possible false negative
```swift
  /**
   Fetch the current value/state of a preference with the given identifier.

   - Parameter identifier: The identifier of the preference item

   - Returns: The current value of the preference
   */
  public static func get(identifier: String) -> AnyObject? {
```

...triggers this warning on the "Returns" line:

`Valid Docs Violation: Documented declarations should be valid. (valid_docs)`

### LINKED ISSUE #5165: accessibility_label_for_image doesn't require a11y label with Image(systemName:)
### New Issue Checklist

### Describe the bug

When opting in to use `accessibility_label_for_image`, I would expect to get a warning every time a default accessibility label is used, but there's no warning for missing labels when the image is an SF Symbol.

While this is documented behaviour in [the rules](https://realm.github.io/SwiftLint/accessibility_label_for_image.html), I consider it a bug because the rule is opt-in for people who want to have better accessibility labels and just describing the icon is not generally a good label.

It is true that some SF Symbols have default accessibility labels, like how `plus` has the default label "Add", many others simply fall back to the name of the icon, like `clock.arrow.circlepath`.

If others disagree, this might just be a feature request to allow us to also require a11y labels when `Image(systemName:)` is used.

mildm8nnered:
Resolves #5165

The `Image(systemName:)` constructor will now trigger a violation, as many system images do not have good accessibility labels.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,6 +106,11 @@
   when calling `NSNumber.init(value:)` directly.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [#5172](https://github.com/realm/SwiftLint/issues/5172)
+* The `accessibility_label_for_image` rule will no longer ignore the
+  `Image(systemName:)` constructor, as many system images do not
+  have good accessibility labels.  
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#5165](https://github.com/realm/SwiftLint/issues/5165)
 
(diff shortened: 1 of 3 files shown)


SimplyDanny (approved):
Thank you!

PR #5203, opened 2023-09-02
Add `include_bare_init` option to `explicit_init` rule

### LINKED ISSUE #2960: Rule Request: [explicit_init] - avoid using `let cl: Class = .init()`

mildm8nnered:
Adds `include_bare_init` option to the `explicit_init` rule.

This will produce violations where `.init` is called using type inference to determine the type.

For example:

```
let foo: Foo = ↓.init()
let foo: [Foo] = [↓.init(), ↓.init()]
foo(↓.init())
```

This option is off by default (as many code bases use the `.init` construct), and these violations are not correctable, as SwiftLint cannot infer the intended type.

Also adds an `include_explicit_init` option, on by default, which allows the original behavior of `explicit_init` to be suppressed, so that users can tune the rule's behavior to suit their particular use case. 

Replaces #5170, and resolves #2960

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,9 +68,8 @@
   values for configurations being printed for a single rule or all rules.  
   [SimplyDanny](https://github.com/SimplyDanny)
 
-* Add `include_bare_init` and `include_explicit_init` options to the
-  `explicit_init` rule. `include_bare_init` encourages using named
-  constructors over `.init()` and type inference.  
+* Add `include_bare_init` option to the `explicit_init` rule. `include_bare_init`
+  encourages using named constructors over `.init()` and type inference.  
   [Martin Redington](https://github.com/mildm8nnered)
   [#5203](https://github.com/realm/SwiftLint/issues/5203)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@
 * Add `--default-config` option to `rules` command allowing to use default
   values for configurations being printed for a single rule or all rules.  
   [SimplyDanny](https://github.com/SimplyDanny)
+
 * Add `include_bare_init` and `include_explicit_init` options to the
   `explicit_init` rule. `include_bare_init` encourages using named
   constructors over `.init()` and type inference.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@
   Bash, Zsh and fish.  
   [SimplyDanny](https://github.com/SimplyDanny)
 
-* Add new `private_swiftui_state_property` opt-in rule to encourage setting 
+* Add new `private_swiftui_state_property` opt-in rule to encourage setting
   SwiftUI `@State` and `@StateObject` properties to private.  
   [mt00chikin](https://github.com/mt00chikin)
   [#3173](https://github.com/realm/SwiftLint/issues/3173)
@@ -53,10 +53,10 @@
   [keith](https://github.com/keith)
   [5139](https://github.com/realm/SwiftLint/pull/5139)
 
-* Show a rule's active YAML configuration in output of 
+* Show a rule's active YAML configuration in output of
   `swiftlint rules <rule>`.  
   [SimplyDanny](https://github.com/SimplyDanny)
-  
+
 * Add `invokeTest()` to `overridden_super_call` defaults.  
   [DylanBettermannDD](https://github.com/DylanBettermannDD)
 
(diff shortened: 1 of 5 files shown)


Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/ExplicitInitConfiguration.swift
```diff
@@ -0,0 +1,23 @@
...
+        guard let configuration = configuration as? [String: Any] else {
+            throw Issue.unknownConfiguration(ruleID: Parent.identifier)
+        }
+
+        if let severityString = configuration[$severityConfiguration] as? String {
+            try severityConfiguration.apply(configuration: severityString)
+        }
+
+        includeExplicitInit = configuration[$includeExplicitInit] as? Bool ?? true
+        includeBareInit = configuration[$includeBareInit] as? Bool ?? false
```
SimplyDanny, line 21: It's better (though a bit more code) to set the properties only if the `configuration` specifies them, because it avoids to use the right defaults at two places.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/ExplicitInitConfiguration.swift
```diff
@@ -0,0 +1,23 @@
+struct ExplicitInitConfiguration: SeverityBasedRuleConfiguration, Equatable {
+    typealias Parent = ExplicitInitRule
+
+    @ConfigurationElement(key: "severity")
+    private(set) var severityConfiguration = SeverityConfiguration<Parent>(.warning)
+    @ConfigurationElement(key: "include_explicit_init")
+    private(set) var includeExplicitInit = true
+    @ConfigurationElement(key: "include_bare_init")
```
SimplyDanny, line 8: Swift calls expression like `.init(...)` "implicit member expression". Not sure we should use the term in an **Explicit**InitRule, though. So "bare" might be okay. What do you think?

mildm8nnered: So I'm not super-happy with "bare". "inferred" sounds a little bit better, and is at least in keeping with existing terms of art, but "inferred_init" doesn't feel quite right, as it's not the "init" that is inferred ... totally open to suggestions if anyone has something better.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/ExplicitInitConfiguration.swift
```diff
@@ -0,0 +1,23 @@
+struct ExplicitInitConfiguration: SeverityBasedRuleConfiguration, Equatable {
+    typealias Parent = ExplicitInitRule
+
+    @ConfigurationElement(key: "severity")
+    private(set) var severityConfiguration = SeverityConfiguration<Parent>(.warning)
+    @ConfigurationElement(key: "include_explicit_init")
```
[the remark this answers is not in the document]
mildm8nnered: So I just added `include_explicit_init` for completeness, even though it does seem unlikely that someone would want to allow `SomeClass.init()` - have removed it now.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/ExplicitInitConfiguration.swift
SimplyDanny: I'm struggling as well. `include_implict_init_members` maybe?

CHANGELOG.md
```diff
@@ -64,6 +64,12 @@
   values for configurations being printed for a single rule or all rules.  
   [SimplyDanny](https://github.com/SimplyDanny)
 
+* Add `include_bare_init` and `include_explicit_init` options to the
```
SimplyDanny, line 67:
```suggestion
* Add `include_bare_init` option to the
```

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/ExplicitInitConfiguration.swift
mildm8nnered: Maybe `bare` is ok - "implicit member expression" feels a little obscure

SimplyDanny: The name would not matter so much if there was a description added to each option. The infrastructure is there already. I may prepare a PR in the upcoming days to open it up and add a few examples. More can then be attached over time.

Long story short, let's keep "bare". 😉

PR #5204, opened 2023-09-04
check string before command for `invalid_swiftlint_command` rule

mildm8nnered:
Some minor improvements to the `invalid_swiftlint_command` rule.

Location is now more accurate - we point to the start of the `swiftlint` command.

Separate reasons are now provided for invalid actions, modifier, or missing rules.

We now check whether `swiftlint` is preceded by whitespace or a comment `/` character, so `sswiftlint:disable some_rule` for example, would now generate a warning.


SimplyDanny (approved):
Great improvements!

PR #5214, opened 2023-09-09
README configuration section updated

chibombo:
## Context
When the developer creates a new Swift package project by default the code directory is `Sources` instead of `Source`.
### Why I modify the READMEs
- When some try to use the example as a first approach, the swiftlint command will fail because the path does exist. The `allow_zero_lintable_files: false` is causing this unexpected behavior.

We can accept the change for future developers or we can change the `allow_zero_lintable_files` rule to `true` in the example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@
 * None.
 
 #### Enhancements
+* Update README configuration for new projects. The `included` used to have the old `Source` path, 
+  the new default is `Sources`.  
+  [chibombo](https://github.com/chibombo)
 
 * Show specific violation message for the `attributes` rule when the option
   `always_on_line_above` or `attributes_with_arguments_always_on_line_above`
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,9 @@
 * None.
 
 #### Enhancements
+* Update README configuration for new projects. The `included` used to have the old `Source` path, 
+  the new default is `Sources`.  
+  [chibombo](https://github.com/chibombo)
```
SimplyDanny, line 25: No need for a changelog entry. Please remove it.

SimplyDanny (changes requested):
Thank you! The change makes sense.

CHANGELOG.md
chibombo: Done 🚀

SimplyDanny (approved):

PR #5215, opened 2023-09-09
`no_magic_numbers` rule should ignore 100

### LINKED ISSUE #5185: no_magic_numbers - Ignore "100" when part of percent conversions
### New Issue Checklist

### New rule request

This one probably requires making too broad of an assumption, but would be awesome if the no_magic_numbers logic was smart enough to ignore "100" when it's part of a percent conversion.

For example:

     let batteryPercent = Int(batteryLevel * 100)

     public var mediaVolume: Int {
        return Int(self.outputVolume * 100)
     }

My idea would be that if  the "magic number" is 100 (or 100.0), and used for multiplication or division:

     a * 100
     100 * b
     c / 100

Then ignore the "100"?

Might be able to get away with ignoring any conversion numbers "100, 1000, 1000000", etc. (any instance of 1 followed by any number of 0's)

Totally okay if all this did was make you laugh at my laziness of not wanting to put in a disable comment. But, figured it couldn't hurt to ask!

-Thanks!

mildm8nnered:
The `no_magic_numbers` rule now ignores 100

I didn't go for any power of 10, as 100 is 97% plus of the cases in one of my large codebases (100 = 4,000+ [includes user readable strings with 14 languages, which bumps it up a fair bit, but still indicative], 1,000 = 94, 10,000 = 12, 100,000 = 10, 1000,000 = 4).

Resolves #5185

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,11 @@
   [Martin Redington](https://github.com/mildm8nnered)
   [#5204](https://github.com/realm/SwiftLint/issues/5204)
 
+* 100 is no longer considered to be a magic number by the `no_magic_numbers`
+  rule.  
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#5215](https://github.com/realm/SwiftLint/issues/5215)
+
 #### Bug Fixes
 
 * Fix false positive in `control_statement` rule that triggered on conditions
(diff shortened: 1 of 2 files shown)


SimplyDanny (approved):
Thank you!

PR #5217, opened 2023-09-09
Add examples for `superfluous_disable_command`

laralove143:
I'm not sure how to document the cases involving configuration files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@
 
 #### Enhancements
 
+* Add examples for `superfluous_disable_command`.  
+  [laralove143](https://github.com/laralove143)
+  [#issue_number](https://github.com/realm/SwiftLint/issues/issue_number)
+
 * Show specific violation message for the `attributes` rule when the option
   `always_on_line_above` or `attributes_with_arguments_always_on_line_above`
   is involved.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,10 +21,6 @@
 
 #### Enhancements
 
-* Add examples for `superfluous_disable_command`.  
-  [laralove143](https://github.com/laralove143)
-  [#issue_number](https://github.com/realm/SwiftLint/issues/issue_number)
-
 * Show specific violation message for the `attributes` rule when the option
   `always_on_line_above` or `attributes_with_arguments_always_on_line_above`
   is involved.  
--- a/Source/SwiftLintCore/Rules/SuperfluousDisableCommandRule.swift
+++ b/Source/SwiftLintCore/Rules/SuperfluousDisableCommandRule.swift
@@ -14,13 +14,13 @@ public struct SuperfluousDisableCommandRule: ConfigurationProviderRule, SourceKi
         kind: .lint,
         nonTriggeringExamples: [
             Example("↓let abc:Void // swiftlint:disable:this colon"),
-            Example("↓// swiftlint:disable colon (If it is enabled in the configuration file"),
-            Example("↓// swiftlint:enable colon (If it is disabled in the configuration file")
+            Example("↓// swiftlint:disable colon (If it is enabled in the configuration file)"),
+            Example("↓// swiftlint:enable colon (If it is disabled in the configuration file)")
         ],
         triggeringExamples: [
             Example("↓let abc: Void // swiftlint:disable:this colon"),
-            Example("↓// swiftlint:disable colon (If it is disabled in the configuration file"),
-            Example("↓// swiftlint:enable colon (If it is enabled in the configuration file")
+            Example("↓// swiftlint:disable colon (If it is disabled in the configuration file)"),
+            Example("↓// swiftlint:enable colon (If it is enabled in the configuration file)")
         ]
     )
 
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,10 @@
 
 #### Enhancements
 
+* Add examples for `superfluous_disable_command`.  
+  [laralove143](https://github.com/laralove143)
+  [#issue_number](https://github.com/realm/SwiftLint/issues/issue_number)
```
SimplyDanny, line 26: An entry in the changelog is not needed as this is no user-facing change.

Source/SwiftLintCore/Rules/SuperfluousDisableCommandRule.swift
```diff
@@ -12,6 +12,16 @@ public struct SuperfluousDisableCommandRule: ConfigurationProviderRule, SourceKi
             in the disabled region. Use " - " if you wish to document a command.
             """,
         kind: .lint
```
SimplyDanny, line 14:
```suggestion
        kind: .lint,
```
This fails the build.

Source/SwiftLintCore/Rules/SuperfluousDisableCommandRule.swift
```diff
@@ -12,6 +12,16 @@ public struct SuperfluousDisableCommandRule: ConfigurationProviderRule, SourceKi
...
+        nonTriggeringExamples: [
+            Example("let abc:Void // swiftlint:disable:this colon"),
+            Example("// swiftlint:disable colon (If it is enabled in the configuration file"),
+            Example("// swiftlint:enable colon (If it is disabled in the configuration file")
+        ],
+        triggeringExamples: [
+            Example("let abc: Void // swiftlint:disable:this colon"),
+            Example("// swiftlint:disable colon (If it is disabled in the configuration file"),
+            Example("// swiftlint:enable colon (If it is enabled in the configuration file")
```
SimplyDanny, line 23: Please add violation markers `↓` to indicate where exactly the rule triggers.

Source/SwiftLintCore/Rules/SuperfluousDisableCommandRule.swift
```diff
@@ -12,6 +12,16 @@ public struct SuperfluousDisableCommandRule: ConfigurationProviderRule, SourceKi
             in the disabled region. Use " - " if you wish to document a command.
             """,
         kind: .lint
+        nonTriggeringExamples: [
+            Example("let abc:Void // swiftlint:disable:this colon"),
+            Example("// swiftlint:disable colon (If it is enabled in the configuration file"),
+            Example("// swiftlint:enable colon (If it is disabled in the configuration file")
```
SimplyDanny, line 18: I wonder how the example can at the same time trigger and not trigger the rule. 🤔

SimplyDanny (changes requested):
Thanks for taking over this job! Please see my review comments ...

Source/SwiftLintCore/Rules/SuperfluousDisableCommandRule.swift
laralove143: It depends on the configuration, if it's already enabled/disabled in the configuration, the enable/disable command doesn't do anything and is therefore linted. I haven't tested this personally but that's how I imagined it'd work.

laralove143: I went ahead and tested it manually, that is indeed the case.

Source/SwiftLintCore/Rules/SuperfluousDisableCommandRule.swift
```diff
@@ -11,7 +11,17 @@ public struct SuperfluousDisableCommandRule: ConfigurationProviderRule, SourceKi
...
+        nonTriggeringExamples: [
+            Example("↓let abc:Void // swiftlint:disable:this colon"),
+            Example("↓// swiftlint:disable colon (If it is enabled in the configuration file)"),
+            Example("↓// swiftlint:enable colon (If it is disabled in the configuration file)")
+        ],
+        triggeringExamples: [
+            Example("↓let abc: Void // swiftlint:disable:this colon"),
+            Example("↓// swiftlint:disable colon (If it is disabled in the configuration file)"),
+            Example("↓// swiftlint:enable colon (If it is enabled in the configuration file)")
+        ]
```
SimplyDanny, line 24: I'd like to make this a bit clearer, in that "non-triggering" really shows an example in which the `colon` rule would trigger while the command disables it. Similarly, the "triggering" case presents the same example but now with the rule triggering which draws the `disable` command useless. That's what `superfluous_disable_command` is there for and thus the examples should depict.

I also think it's clear that the `colon` rule must be enabled in the first place. We might not need to mention this in extra comments.

Contrary to one of my last review comments, violations markers are not helpful here as they would appear at a confusing location. The exact location anyway doesn't matter so much for this rule in my opinion. Sorry for the back and forth.

That said, my suggestion would be like:

```suggestion
        nonTriggeringExamples: [
            Example("let abc:Void // swiftlint:disable:this colon"),
            Example("""
                // swiftlint:disable colon
                let abc:Void
                // swiftlint:enable colon
                """)
        ],
        triggeringExamples: [
            Example("let abc: Void // swiftlint:disable:this colon"),
            Example("""
                // swiftlint:disable colon
                let abc: Void
                // swiftlint:enable colon
                """)
        ]
```

SimplyDanny:
In order to make the tests run through, you need to wait for #5225 to be merged. After rebasing you PR on top of it, tests are expected to be successful.

SimplyDanny (approved):

PR #5219, opened 2023-09-13
Document issue regarding the change in default value for sandboxing

technocidal:
Xcode 15 enforces Sandboxing for User Scripts by default. This means that SwiftLint does not run anymore as it can't access any of the necessary files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,15 @@ Insert the following as the script:
 
 ![](https://raw.githubusercontent.com/realm/SwiftLint/main/assets/runscript.png)
 
+> Xcode 15 made a significant change by setting the default value of the ENABLE_USER_SCRIPT_SANDBOXING Build Setting from NO to YES.
+> As a result, SwiftLint encounters an error related to missing file permissions,
+> which typically manifests as follows: `error: Sandbox: swiftlint(19427) deny(1) file-read-data.`
+>
+> To resolve this issue, it is necessary to manually adjust the ENABLE_USER_SCRIPT_SANDBOXING setting to NO for the specific target that SwiftLint is being configured for.
+> This adjustment will ensure that SwiftLint functions properly.
+>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -149,6 +149,15 @@ Insert the following as the script:
...
 
+> Xcode 15 made a significant change by setting the default value of the ENABLE_USER_SCRIPT_SANDBOXING Build Setting from NO to YES.
+> As a result, SwiftLint encounters an error related to missing file permissions,
+> which typically manifests as follows: `error: Sandbox: swiftlint(19427) deny(1) file-read-data.`
+>
+> To resolve this issue, it is necessary to manually adjust the ENABLE_USER_SCRIPT_SANDBOXING setting to NO for the specific target that SwiftLint is being configured for.
+> This adjustment will ensure that SwiftLint functions properly.
+>
+> Special thanks to [Michal Tynior](https://thisdevbrain.com/swiftlint-permission-issue/) for providing this solution.
+
```
SimplyDanny, line 160: I'd like to keep it a bit shorter. Quoting the text is not required. It can be added as normal text.

Special thanks can go into the PR description or the commit message. Better to leave it out of the README. Otherwise, we'd have to add kudos at many more places. 😉 

```suggestion
Xcode 15 made a significant change by setting the default value of the `ENABLE_USER_SCRIPT_SANDBOXING` Build Setting from `NO` to `YES`.
As a result, SwiftLint encounters an error related to missing file permissions,
which typically manifests as follows: `error: Sandbox: swiftlint(19427) deny(1) file-read-data.`

To resolve this issue, it is necessary to manually set the `ENABLE_USER_SCRIPT_SANDBOXING` setting to `NO` for the specific target that SwiftLint is being configured for.

```

SimplyDanny (changes requested):
Thanks!

README.md
technocidal: @SimplyDanny Thanks for the feedback! I agree with your points and have accepted your suggestion.

SimplyDanny (approved):

PR #5226, opened 2023-09-18
Add strict configuration file setting

### LINKED ISSUE #5220: Error severity for every rule using .swiftlint.yml
Would it be possible to have a configuration in `.swiftlint.yml` that treats all violations as errors? It'll be nice to have the option to enable what the `--strict` flag does but from the `.swiftlint.yml` config file. I'd like to run swiftlint on strict mode but I'm using SPM to add it into my project and I can't find a way to pass the `--strict` flag to the plugin command. I'm thinking on forking the repository and updating the plugin code on my side, but that solution isn't ideal since I'd have to keep updating that fork to get updates.

mildm8nnered:
Resolves #5220 

Adds a `strict` configuration file setting, equivalent to the `--strict` command line option.

If `--lenient` is passed on the command line, this will take precendence over the configuration file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,7 +86,7 @@
 * Adds a `strict` configuration file setting, equivalent to the `--strict`
   command line option.
   [Martin Redington](https://github.com/mildm8nnered)
-  [#XXXX](https://github.com/realm/SwiftLint/issues/XXXX)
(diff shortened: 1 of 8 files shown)


SimplyDanny (approved):
Thank you!

PR #5230, opened 2023-09-20
Require Swift 5.9

SimplyDanny:


jpsim:
I sorted out the CI issues in https://github.com/realm/SwiftLint/pull/5235 and that will merge once CI passes.

PR #5242, opened 2023-09-27
Preserve `@_exported` import statements in `unused_imports`

jszumski:
Preserve `@_exported` import statements even if unused to avoid breaking downstream modules depending on the re-exported module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@
 #### Bug Fixes
 
 * Fix false positive in `unused_import` rule that triggered on 
-  `@_exported` imports which could break downstream modules even if removed.  
+  `@_exported` imports which could break downstream modules if removed.  
   [jszumski](https://github.com/jszumski)
   [#5242](https://github.com/realm/SwiftLint/pull/5242)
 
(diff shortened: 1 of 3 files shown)


Source/SwiftLintBuiltInRules/Rules/Lint/UnusedImportRuleExamples.swift
```diff
@@ -122,7 +125,7 @@ struct UnusedImportRuleExamples {
             dispatchMain()
             """),
         Example("""
-        ↓@_exported import Foundation
+        ↓@_implementationOnly import Foundation
```
jszumski, line 128: The original intent in https://github.com/realm/SwiftLint/pull/2993 seemed to make sure _any_ attribute was correctable, so I arbitrarily chose `@_implementationOnly` as a replacement.

CHANGELOG.md
```diff
@@ -19,7 +19,10 @@
 
 #### Bug Fixes
 
-* None.
+* Fix false positive in `unused_import` rule that triggered on 
+  `@_exported` imports which could break downstream modules even if removed.
```
SimplyDanny, line 23:
```suggestion
  `@_exported` imports which could break downstream modules if removed.  
```

SimplyDanny (approved):
Thank you!

PR #5246, opened 2023-09-28
Fix `unused_import` when using a constructor defined transitively

jszumski:
Fixes a false positive in `unused_import` when using a constructor defined in a transitive module.

https://github.com/apple/swift/commit/3ea9bed415526f1ed27affac6fe87c868941a376 added additional cursor info for the constructor in `key.secondary_symbols`.

## Example

```swift
// Module A
class Thing {}

// Module B
import ModuleA
extension Thing {
   init(model: [String: Any]) {}
}

// Module C
import ModuleA
import ModuleB // this gets marked as unused

let foo = Thing(model: ...)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,11 @@
   [jszumski](https://github.com/jszumski)
   [#5242](https://github.com/realm/SwiftLint/pull/5242)
 
+* Fix false positive in `unused_import` rule when using a constructor 
+  defined in a transitive module.  
+  [jszumski](https://github.com/jszumski)
+  [#5246](https://github.com/realm/SwiftLint/pull/5246)
+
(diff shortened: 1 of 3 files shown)


jszumski:
What is the best way to model this fix in a test?

SimplyDanny:
> What is the best way to model this fix in a test?

I can only think of a separate test that runs the rule on a dummy project portraying exactly this setup of dependencies, that is an Xcode project with three modules like you've used in the description. But it's also fine for me to accept this PR without a dedicated test as all existing examples still seem to work.

jszumski:
I'm good to merge as is then, thanks!

SimplyDanny:
Please rebase. Danger doesn't like merge commits. 😉

SimplyDanny (approved):

PR #5255, opened 2023-10-03
Configure expression folding by argument of @SwiftSyntaxRule macro

SimplyDanny:


Source/SwiftLintBuiltInRules/Rules/Performance/EmptyCountRule.swift
```diff
@@ -37,6 +36,10 @@ struct EmptyCountRule: ConfigurationProviderRule, OptInRule, SwiftSyntaxRule {
     func makeVisitor(file: SwiftLintFile) -> ViolationsSyntaxVisitor {
         Visitor(onlyAfterDot: configuration.onlyAfterDot)
     }
+
+    func preprocess(file: SwiftLintFile) -> SourceFileSyntax? {
+        file.foldedSyntaxTree
+    }
```
SimplyDanny, line 42: We might be able to find a solution for this explicit implementation as well. Configurations need to be passed to the visitor somehow automatically.

jpsim (approved):
This is nicer, thanks!

Source/SwiftLintBuiltInRules/Rules/Performance/EmptyCountRule.swift
SimplyDanny: That's done with #5275.

Issue #5254, opened 2023-10-02
install via brew on silicon macOS fails


netskink:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

Updated brew
installed SwiftLint via brew
My path already had /opt/ fix for silicon macOS

works 
```
$ swiftlint -h
```
 
fails
```
$ swiftlint
SourceKittenFramework/library_wrapper.swift:31: Fatal error: Loading sourcekitdInProc.framework/Versions/A/sourcekitdInProc failed
```

Adding to build phase like so:

```
# Type a script or drag a script file from your workspace to insert its path.
if which swiftlint > /dev/null; then
    swiftlint
else
    echo "Warning: swiftlint not installed. Use brew install swiftlint"
fi
```

Does nothing.  I don't see any messages on pass or fail. I suspect it fails sine running at prompt fails.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash

$ swiftlint lint
SourceKittenFramework/library_wrapper.swift:31: Fatal error: Loading sourcekitdInProc.framework/Versions/A/sourcekitdInProc failed
[1]    12738 trace trap  swiftlint lint
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)?
    - 0.53.0 
* Installation method used (Homebrew, CocoaPods, building from source, etc)?
   - brew install swiftlint 
* Paste your configuration file:

no mods other that brew install. No knowledge of whee this file is located

```yml
# insert yaml contents here
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)?
  If so, paste their relative paths and respective contents.
   - got me

* Which Xcode version are you using (check `xcodebuild -version`)?
    - >  xcode -version
    - > zsh: command not found: Xcode
    - Xcode uI says 14.3
    
* Do you have a sample that shows the issue? Run `echo "[string here]" | swiftlint lint --no-cache --use-stdin --enable-all-rules`
  to quickly test if your example is really demonstrating the issue. If your example is more
  complex, you can use `swiftlint lint --path [file here] --no-cache --enable-all-rules`.

```swift
// This triggers a violation:
let foo = try! bar()
```

SimplyDanny:
Is Xcode installed? [This](https://github.com/realm/SwiftLint/issues/1466#issuecomment-297265863) might help.

[...]

PR #5260, opened 2023-10-03
update to specify the proper active directory to docs

netskink:
This in reference to this [issue](https://github.com/realm/SwiftLint/issues/5254)

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,36 @@ unacceptable behavior to [info@realm.io](mailto:info@realm.io).
 brew install swiftlint
 ```
 
+Afterwords ensure that xcode has the active developer directory set to
+point where the Swift toolchain is located.
+
+This is the default active directory but it does not have the Swift toolchain:
+
+```
+$ xcode-select -p
+/Library/Developer/CommandLineTools
+```
+
+Use this command to specify the default active directory that includes
+the Swift toolchain:
+
+```
+❯ sudo xcode-select -s /Applications/Xcode.app/Contents/Developer
+Password:
+❯ xcode-select -p
+/Applications/Xcode.app/Contents/Developer
+```
+
+Afterwards, `swiftlint` with default analyze command will work:
+
+```
+> swiftlint
+.... <lines trucated>
+empty line; currently 4 (vertical_whitespace)
+Done linting! Found 897 violations, 10 serious in 39 files.
+```
+
+
 ### Using [CocoaPods](https://cocoapods.org):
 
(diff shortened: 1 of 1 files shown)


SimplyDanny:
I think these steps are already covered by [this section](https://github.com/realm/SwiftLint#working-with-multiple-swift-versions). Do you agree?

netskink:
No that is why I'm adding it to that section. Please search on Xcode select in that section and show me where it's mentioned.

SimplyDanny:
`xcode-select` isn't something specific for SwiftLint. It's the common way to switch between Xcode toolchains. Thus, we don't need to explain.

The section I mentioned definitely relates to it. The point is that Xcode must be installed; nothing more. SwiftLint looks for SourceKit at the locations listed there. `xcode-select` isn't needed at all. You may just set `$XCODE_DEFAULT_TOOLCHAIN_OVERRIDE`, `$TOOLCHAIN_DIR` or `$TOOLCHAINS` to make SwiftLint find special Xcodes at exotic locations.

That said, what might be left for the README is the fact that a full Xcode must be installed (preferably) at a default location. But this applies to [all installation methods](https://github.com/realm/SwiftLint#installation) and should at most be mentioned there.

netskink:
Very good.  Is these environment variables specified in the docs?  If not would it be wise to specify?

SimplyDanny:
> Very good. Is these environment variables specified in the docs? If not would it be wise to specify?

Yes, they are [here](https://github.com/realm/SwiftLint#working-with-multiple-swift-versions).

netskink:
In the section titled "working with multiple swift versions"?

SimplyDanny:
> In the section titled "working with multiple swift versions"?

Yes.

netskink:
So I only have one swift version installed.  That section is a long way off from the initial installation instructions.  Not to mention the section you referenced is not explicit to provide the correct procedure.

With that said do you still feel the documentation does not need to be modified?

After all this is a tool to make higher quality code and one aspect is that code be simplified and easier to understand, right?

SimplyDanny:
> After all this is a tool to make higher quality code and one aspect is that code be simplified and easier to understand, right?

Yes, and one aspect of high quality is to avoid redundancy. I just want to make sure the README contains the relevant points but not more.

If you think the section is too far away from the "Installation" section, link to it from there or move it. Fine with me, if you mention `xcode-select` as a tool to change default Xcode installations. I really wouldn't do more here though.

PR #5270, opened 2023-10-09
Add option for `unneeded_override` to affect initializers

### LINKED ISSUE #5265: unneeded_override doesn't check init
### New Issue Checklist

- [ ] Updated SwiftLint to the latest version

### Describe the bug

unneeded_override rule does not check  init

```bash
$ swiftlint lint
```

### Environment

* SwiftLint version 0.52.3
* Installation method used CocoaPods

In the following code, `override init()` will not trigger:
```swift

    override init() {
        super.init()
    }

    override func printHi() {
        super.printHi()
    }
```

leonardosrodrigues0:
## Changes

These changes add an `affect_initializers` option to `unneeded_override` rule . Closes #5265.

When the option is enabled, the overridden initializers are checked in the same way methods are. In addition, failable (`init?`) and implicitly unwrapped (`init!`) initializers never trigger the rule, as it is allowed to override one with the other and vice versa. Finally, `private` overrides also don't trigger the rule, as they can be used to hide the `super` initializer.

## Implementation

A new `OverridableDecl` protocol was built to allow both `FunctionDeclSyntax` and `InitializerDeclSyntax` to be checked using the same function. The check for `init` name on the member access expression was implemented by comparing directly to the `"init"` string, as it is done in `ExplicitInitRule`, `NSNumberInitAsFunctionReferenceRule` and `ReduceIntoRule`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,11 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#4801](https://github.com/realm/SwiftLint/pull/4801)
 
+* Add `affect_initializers` option to allow `unneeded_override` rule
+  to affect initializers.  
+  [leonardosrodrigues0](https://github.com/leonardosrodrigues0)
+  [#5265](https://github.com/realm/SwiftLint/issues/5265)
+
 #### Bug Fixes
 
 * Ignore overridden functions with default parameters in the `unneeded_override`
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/UnneededOverrideRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/UnneededOverrideRule.swift
@@ -3,7 +3,7 @@ import SwiftSyntaxBuilder
 
 @SwiftSyntaxRule(explicitRewriter: true)
 struct UnneededOverrideRule: Rule {
-    var configuration = SeverityConfiguration<Self>(.warning)
+    var configuration = UnneededOverrideRuleConfiguration()
 
     static let description = RuleDescription(
         identifier: "unneeded_override",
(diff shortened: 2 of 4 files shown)


leonardosrodrigues0:
Any thoughts on this?

leonardosrodrigues0:
@SimplyDanny Could you review this enhancement? This would be my first contribution here and any suggestion are welcome.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededOverrideRule.swift
```diff
@@ -16,103 +16,136 @@ struct UnneededOverrideRule: Rule {
...
+    var declText: String { get }
 }
 
-private extension UnneededOverrideRule {
-    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
-        override func visitPost(_ node: FunctionDeclSyntax) {
-            if isUnneededOverride(node) {
-                self.violations.append(node.positionAfterSkippingLeadingTrivia)
-            }
+extension OverridableDeclSyntax {
```
SimplyDanny, line 62: All the extensions should be `private`.

SimplyDanny (commented):
Very well done! Not much to say. 👍

To check the rule against a few OSS project, please set the new option to `true` by default. With the report produced by the build we are able to judge on the option's impact. Typically we find some special cases that need to be considered in the implementation. After the report is checked, you can turn off the option again.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededOverrideRule.swift
```diff
@@ -16,103 +16,136 @@ struct UnneededOverrideRule: Rule {
...
         }
+
+        return true
     }
 }
 
-private func isUnneededOverride(_ node: FunctionDeclSyntax) -> Bool {
-    guard node.modifiers.contains(keyword: .override), let statement = node.body?.statements.onlyElement else {
-        return false
+extension FunctionDeclSyntax: OverridableDeclSyntax {
```
SimplyDanny, line 106: Please make these two extensions `private` as well.

leonardosrodrigues0: I tried making them `private` while I was implementing this, but we can't:
```
'private' modifier cannot be used with extensions that declare protocol conformances
```

leonardosrodrigues0:
Thank you for reviewing this! 

> Typically we find some special cases that need to be considered in the implementation.

I made the change and checked the report. It seems like the differences found were the expected.

Cases with `override`s that change the visibility of the `init` caught my attention, but those could also happen with a method and it seems impossible to me to catch them while linting.

There are also the `required` inheritance of `init`s that do not need the `override` keyword, but if the subclass implement another initializer, then it is mandatory to implement the `required init` even if it only calls the `super`, so I don't think we can trigger in that case.

> After the report is checked, you can turn off the option again.

I was going for off by default, but now I'm thinking that it would be expected for this to be on by default...

SimplyDanny:
```swift
override init!(arg: ...) {
    super.init(arg: arg)
}
```

and

```swift
override init?(arg: ...) {
    super.init(arg: arg)
}
```

should not trigger. It's allowed to override a failable initializer with an initializer that's implicitly unwrapped and vice versa.

The `private` keyword in

```swift
private override init() {
    super.init()
}
```

is an indicator that this override exists to hide the initializer. If the parent was `private` as well, the override would be prohibited by the compiler.

> I was going for off by default, but now I'm thinking that it would be expected for this to be on by default...

The thing is that folks might already have adopted the rule and it's always surprising if the same rule suddenly triggers on more cases. Sure, then they can set the option to `false`. But just as much they can activate it intentionally.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededOverrideRule.swift
```diff
@@ -16,103 +16,136 @@ struct UnneededOverrideRule: Rule {
...
+            let expr: DeclSyntax = ""
+            return expr
+                .with(\.leadingTrivia, node.leadingTrivia)
+                .with(\.trailingTrivia, node.trailingTrivia)
+        }
     }
+}
 
-    return syntax?.as(FunctionCallExprSyntax.self)
+private protocol OverridableDeclSyntax: DeclSyntaxProtocol {
```
SimplyDanny, line 54: Conforming to `DeclSyntaxProtocol` is not allowed. You can use `some OverridableDeclSyntax & DeclSyntaxProtocol` in `visitOverridable` instead.

SimplyDanny, line 54:
```suggestion
private protocol OverridableDeclSyntax: WithAttributesSyntax, WithModifiersSyntax {
```

saves you two declarations in the protocol. 😉

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededOverrideRule.swift
SimplyDanny: I see. Please avoid to add the `declText` to the types and make it visible to the whole module. Rather pass the name to `isUnneededOverride` which you'd need to make a function.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededOverrideRule.swift
```diff
@@ -16,103 +16,136 @@ struct UnneededOverrideRule: Rule {
...
+    var declText: String { get }
 }
 
-private extension UnneededOverrideRule {
-    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
-        override func visitPost(_ node: FunctionDeclSyntax) {
-            if isUnneededOverride(node) {
-                self.violations.append(node.positionAfterSkippingLeadingTrivia)
-            }
+private extension OverridableDeclSyntax {
```
SimplyDanny, line 62: Please omit the `Syntax` suffix. This is what SwiftSyntax uses for their types.

SimplyDanny (changes requested):

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededOverrideRule.swift
leonardosrodrigues0: Done!

leonardosrodrigues0:
Thanks for the insights on failable, implicitly unwrapped and private initializers, I'm fixing those.

I found a solution to not expose the private protocol contents while conforming the types to it:

```swift
class Foo {}

private protocol Bar {
    var value: String { get }
}

private extension Foo {
    var value: String { "bar" }
}

extension Foo: Bar {}
```

I'm thinking this should be good for us here, let me know if you judge otherwise.

SimplyDanny:
> I'm thinking this should be good for us here, let me know if you judge otherwise.

It's good enough in the sense of a workaround. But is it really easy to follow? If it doesn't bother you too much, I'd still prefer the extra `name` argument. Not elegant but easy to comprehend.

leonardosrodrigues0:
> It's good enough in the sense of a workaround. But is it really easy to follow? If it doesn't bother you too much, I'd still prefer the extra `name` argument. Not elegant but easy to comprehend.

Interesting point, I agree... I just changed it and added the checks for failable, implicitly unwrapped and private initializers.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededOverrideRule.swift
leonardosrodrigues0: Done.

CHANGELOG.md
```diff
@@ -29,6 +29,10 @@
 * Rewrite `multiline_literal_brackets` rule using SwiftSyntax, fixing some
   false positives that would happen when comments are present.  
   [Marcelo Fabri](https://github.com/marcelofabri)
+  
+* Add option to make `unneeded_override` affect initializers.
```
SimplyDanny, line 33: Please mention the option's name.

SimplyDanny (approved):

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/UnneededOverrideRuleConfiguration.swift
```diff
@@ -0,0 +1,11 @@
+import SwiftLintCore
+
+@AutoApply
+struct UnneededOverrideRuleConfiguration: SeverityBasedRuleConfiguration {
+    typealias Parent = UnneededOverrideRule
+
+    @ConfigurationElement(key: "severity")
+    private(set) var severityConfiguration = SeverityConfiguration<Parent>(.warning)
+    @ConfigurationElement(key: "affect_inits")
```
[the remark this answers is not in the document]
leonardosrodrigues0: Done!

CHANGELOG.md
leonardosrodrigues0: Done!

SimplyDanny:
Thanks, @leonardosrodrigues0! That's a great first contribution. 👍

leonardosrodrigues0:
Thank you @SimplyDanny for the reviews and the support!!

PR #5272, opened 2023-10-10
Harmonize rule extensions

SimplyDanny:
~This allows to reduce existential types later.~

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/BlockBasedKVORule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/BlockBasedKVORule.swift
@@ -35,7 +35,7 @@ struct BlockBasedKVORule: ConfigurationProviderRule {
     )
 }
 
-extension BlockBasedKVORule {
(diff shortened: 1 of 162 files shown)


jpsim:
> This allows to reduce existential types later.

How? On its own, I preferred keeping private implementation details of these rules private and inaccessible to the rest of the module.

SimplyDanny:
> > This allows to reduce existential types later.
> 
> How? On its own, I preferred keeping private implementation details of these rules private and inaccessible to the rest of the module.

The method that creates the rewriter for a rule

```swift
func makeRewriter(file: SwiftLintFile) -> ViolationsSyntaxRewriter?
```

returns an existential. If the rewriters are `internal` this can become an opaque result type:

```swift
func makeRewriter(file: SwiftLintFile) -> (some Rewriter)? {
```

SimplyDanny:
Turns out, the compiler confused me. Visibility can remain `private` for this to work (which makes more sense admittedly). A few things can still be harmonized, though. Let me update the PR keeping the extensions private ...

PR #5273, opened 2023-10-11
Enforce `any` on existential types

SimplyDanny:
This makes syntactically clear which types are rather expensive. `any` will be mandatory in Swift 6.

Especially on return types, existential types can be avoided by having generic protocols (adding an `associatedtype`). That has been done for `SwiftSyntaxCorrectableRule` and `Rule`.

@jpsim: Okay for you to enforce `any` in the Bazel and SPM builds?

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -6,7 +6,7 @@ load(
     "swift_compiler_plugin"
 )
 
-swiftFeatures = ["-enable-upcoming-feature", "ExistentialAny"]
+copts = ["-enable-upcoming-feature", "ExistentialAny"]
 
 # Targets
 
@@ -19,7 +19,7 @@ swift_library(
         "@SwiftSyntax//:SwiftCompilerPlugin_opt",
         "@SwiftSyntax//:SwiftSyntaxMacros_opt",
     ],
-    copts = swiftFeatures,
+    copts = copts,
 )
 
 swift_compiler_plugin(
@@ -29,7 +29,7 @@ swift_compiler_plugin(
         "@SwiftSyntax//:SwiftCompilerPlugin_opt",
         "@SwiftSyntax//:SwiftSyntaxMacros_opt",
     ],
-    copts = swiftFeatures,
+    copts = copts,
 )
 
 swift_library(
@@ -53,7 +53,7 @@ swift_library(
     plugins = [
         ":SwiftLintCoreMacros",
     ],
-    copts = swiftFeatures,
+    copts = copts,
 )
 
 swift_library(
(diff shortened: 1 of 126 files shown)


BUILD
```diff
@@ -6,6 +6,8 @@ load(
     "swift_compiler_plugin"
 )
 
+swiftFeatures = ["-enable-upcoming-feature", "ExistentialAny"]
```
jpsim, line 9: Nit: call this `copts` since in rules_swift, "features" are a different thing. E.g. https://github.com/bazelbuild/rules_swift/blob/1.12.0/swift/internal/feature_names.bzl

```suggestion
copts = ["-enable-upcoming-feature", "ExistentialAny"]
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/StrictFilePrivateRule.swift
```diff
@@ -185,7 +185,7 @@ private extension StrictFilePrivateRule {
             violations.append(node.positionAfterSkippingLeadingTrivia)
         }
 
-        private func implementedTypesInDecl(of node: SyntaxProtocol?) -> [String] {
+        private func implementedTypesInDecl(of node: (any SyntaxProtocol)?) -> [String] {
```
jpsim, line 188: This one can't be `some`?

Source/SwiftLintBuiltInRules/Rules/Idiomatic/SyntacticSugarRule.swift
```diff
@@ -256,7 +256,7 @@ private final class SyntacticSugarRuleVisitor: SyntaxVisitor {
 }
 
 private struct CorrectingContext {
-    let rule: Rule
+    let rule: any Rule
```
jpsim, line 259: This can trivially avoid the existential

```swift
private struct CorrectingContext<R: Rule> {
    let rule: R
```

jpsim (approved):
I'm fine if we do this, but I spot checked a few of the `any` uses in this PR and it seems like there's probably more that could be easily converted to `some`. I suggest you run through them to pick off the easy ones before merging.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/StrictFilePrivateRule.swift
SimplyDanny: It can. 👍

SimplyDanny:
> I'm fine if we do this, but I spot checked a few of the `any` uses in this PR and it seems like there's probably more that could be easily converted to `some`. I suggest you run through them to pick off the easy ones before merging.

Good catches! I replaced a few more `any`s. So the change already works as intended: It's now easier to spot unnecessary existentials or existentials that can be get rid of by a small refactoring.

PR #5274, opened 2023-10-12
Let all rules be configurable

SimplyDanny:
`ConfigurationProviderRule` does not serve a purpose anymore as all rules are configurable (if only with a severity).

This is one more step to let `@SwiftSyntaxRule` generate even more boilerplate code by relying on the existence of a `configuration` variable that can be passed to a rule's `Visitor` and `Rewriter` automatically.


jpsim (approved):
Nice cleanup. Were there any rules that couldn't previously be configured that now can? If so this could warrant a note in the changelog.

PR #5275, opened 2023-10-12
Support arbitrary configurations in `@SwiftSyntaxRule`

SimplyDanny:
Almost all rules based on SwiftSyntax can be set up now by just adding `@SwiftSyntaxRule` to the rule struct.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -3,7 +3,8 @@ import CompilerPluginSupport
 import PackageDescription
 
 let swiftFeatures: [SwiftSetting] = [
-    .enableUpcomingFeature("ExistentialAny")
+    .enableUpcomingFeature("ExistentialAny"),
+    .unsafeFlags(["-cross-module-optimization"])
 ]
 
 let package = Package(
--- a/Package.swift
+++ b/Package.swift
@@ -3,8 +3,7 @@ import CompilerPluginSupport
 import PackageDescription
 
 let swiftFeatures: [SwiftSetting] = [
-    .enableUpcomingFeature("ExistentialAny"),
-    .unsafeFlags(["-cross-module-optimization"])
+    .enableUpcomingFeature("ExistentialAny")
 ]
 
 let package = Package(
--- a/BUILD
+++ b/BUILD
@@ -6,7 +6,7 @@ load(
     "swift_compiler_plugin"
 )
 
-copts = ["-enable-upcoming-feature", "ExistentialAny"]
+copts = ["-enable-upcoming-feature", "ExistentialAny", "-cross-module-optimization"]
 
 # Targets
 
--- a/BUILD
+++ b/BUILD
@@ -6,7 +6,7 @@ load(
     "swift_compiler_plugin"
 )
 
-copts = ["-enable-upcoming-feature", "ExistentialAny", "-cross-module-optimization"]
+copts = ["-enable-upcoming-feature", "ExistentialAny"]
 
 # Targets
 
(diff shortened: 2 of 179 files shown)


SimplyDanny:
Hm ... why does this have an effect on runtime performance? 🤔

jpsim:
> Hm ... why does this have an effect on runtime performance?

Because of generics across module boundaries that can't be specialized. I think.

jpsim:
> > Hm ... why does this have an effect on runtime performance?
> 
> Because of generics across module boundaries that can't be specialized. I think.

Ah, no if I had to guess it's that now every rule calls `file.locationConverter`, which is managed by a lock: https://github.com/realm/SwiftLint/blob/0.53.0/Source/SwiftLintCore/Extensions/SwiftLintFile%2BCache.swift#L166

SimplyDanny:
> > > Hm ... why does this have an effect on runtime performance?
> 
> > 
> 
> > Because of generics across module boundaries that can't be specialized. I think.
> 
> 
> 
> Ah, no if I had to guess it's that now every rule calls `file.locationConverter`, which is managed by a lock: https://github.com/realm/SwiftLint/blob/0.53.0/Source/SwiftLintCore/Extensions/SwiftLintFile%2BCache.swift#L166

Yeah, that is my guess too. I'm working on a commit to verify that.

Are you basically happy with this change?

jpsim:
> Are you basically happy with this change?

I'm not convinced this is better. Tastes a bit like generic soup to me and I don't find that the boilerplate it helps remove is all that worth it.

Seems like we might be able to do something like this:

```swift
@SwiftSyntaxRule(locationConverter: true)
struct AnonymousArgumentInMultilineClosureRule: OptInRule { /* ...* / }
```

and only change the codegen, not the visitor declarations, and avoid all the new generics.

SimplyDanny:
> Seems like we might be able to do something like this:
> 
> ```swift
> @SwiftSyntaxRule(locationConverter: true)
> struct AnonymousArgumentInMultilineClosureRule: OptInRule { /* ...* / }
> ```
> 
> and only change the codegen, not the visitor declarations, and avoid all the new generics.

The reason for the generics is the configuration. As long as the visitor doesn't know the type of a rule's configuration, it's impossible to generate the `makeVisitor` for it. Once a rule has a custom configuration, the `@SwiftSyntaxRule` becomes unusable.
The generic soup is mainly invisible to people working on rule implementations.

jpsim:
I can come around on the generics ergonomics aspect, it's not as big as I thought it was at first glance.

Although there still appears to be a ~1%-3% perf impact with this change. Maybe now it's due to the generic APIs across module boundaries?

Package.swift
```diff
@@ -3,7 +3,8 @@ import CompilerPluginSupport
 import PackageDescription
 
 let swiftFeatures: [SwiftSetting] = [
-    .enableUpcomingFeature("ExistentialAny")
+    .enableUpcomingFeature("ExistentialAny"),
+    .unsafeFlags(["-cross-module-optimization"])
```
jpsim, line 7: Does using unsafe flags affect warnings end users see when integrating SwiftLint as a SwiftPM plugin?

jpsim (approved):
+1 Assuming SwiftPM is happy consuming this package even with unsafe flags.

Nice job finding the `-cross-module-optimization` flag. As if Swift has a giant "fix perf" button 😅

Package.swift
SimplyDanny: As per the [documentation](https://developer.apple.com/documentation/packagedescription/swiftsetting/unsafeflags(_:_:)#discussion), it does cause errors for version-based dependencies. While that might be okay on macOS due to the use of `SwiftLintBinary`, it's an issue on Linux where everything has to be built.

SimplyDanny:
> +1 Assuming SwiftPM is happy consuming this package even with unsafe flags.
> 
> Nice job finding the `-cross-module-optimization` flag. As if Swift has a giant "fix perf" button 😅

Yeah, it makes total sense for modules only used to structure a project; not the ones that are delivered as standalone libraries. Yet, it still seems to be an internal experimental feature. 

As I understand, it can be used during development, at least, to check if specializations of generic types and methods at compile time generally have an effect on performance. If they have, one can try to make a few declarations `@inlinable` explicitly to achieve the same performance gain. But that's quite tedious.

SimplyDanny:
Performance looks pretty good with the explicit annotations now ...

PR #5276, opened 2023-10-13
Port more rules to the `@SwiftSyntaxRule` macro

jpsim:
By adding `needsLocationConverter` and `needsConfiguration` parameters. An alternative to #5275.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/AnonymousArgumentInMultilineClosureRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/AnonymousArgumentInMultilineClosureRule.swift
@@ -1,6 +1,7 @@
 import SwiftSyntax
 
-struct AnonymousArgumentInMultilineClosureRule: SwiftSyntaxRule, OptInRule {
+@SwiftSyntaxRule(needsLocationConverter: true)
+struct AnonymousArgumentInMultilineClosureRule: OptInRule {
     var configuration = SeverityConfiguration<Self>(.warning)
 
     static let description = RuleDescription(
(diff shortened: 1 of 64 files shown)


SimplyDanny:
I'm not so happy with this. For me, macros should not be coupled too much to the context they are used in. Preferably, the macro implementation can evaluate if the macro is allowed on a declaration, as an expression, ... and report a diagnostic accordingly. In my mind, a macro can be added or not. In case it can, it should just work and not dictate how the surrounding code must be structured, named and ordered.

This is why I did #5275. People working on a rule don't need to know anything about the macro's implementation. They can just refer to the configuration or the location converter if they need to. How all this works, is not important.

With the proposal here, the macro defines how to name the parameters and their order. I presumably need to look up the macro's implementation to really know what I have to do. And the amount of code I can spare to write is quite low. Fields and the initialzer don't come automatically, even though they always look the same. Adding the `makeVisitor` myself (without the macro) is not a big deal either then.

That said, I still belief in #5275. 😄 Just need to figure out how to avoid runtime generics.

jpsim:
I get what you're saying, and somewhat agree and somewhat disagree 😄

I don't necessarily think the macro needs to be 100% decoupled from the code in the surrounding area of the usage site when it's already a macro super specific to writing SwiftLint rules. Before this PR we're already making an assumption that there's a private `Visitor` type available.

On the note of this not saving that much in terms of lines of code, yeah it's not groundbreaking but what is more valuable than saving lines of code is enforcing a certain amount of consistency across all rules. This does that.

So overall, while you make excellent points, I think this approach makes a really nice "DSL" for writing rules conveniently and consistently, without incurring any performance penalties.

Maybe we can find a middle ground between this and #5275 where we can both be happy 😄.

Thanks so much for sharing your opinions, I really appreciate the discussion.

jpsim:
Closing in favor of #5275.

PR #5279, opened 2023-10-15
Ignore completely empty enum arguments

### LINKED ISSUE #5269: empty_enum_arguments false positive, --fix command breaks code preventing successful compile
### New Issue Checklist

### Describe the bug

The empty_enum_arguments has false positives when using HealthKit types. HealthKit still has a number of types using static functions instead of static vars. The --fix command will remove the required parenthesis and make the code not compile.

Screenshot 2023-10-09 at 7 46 11 AM

https://developer.apple.com/documentation/healthkit/hkunit

Since there are system APIs in the health space and probably other built-in sdks outside of the developers control, this rule should allow a customization to provide an allow list or ignore list.

```bash
$ swiftlint lint
```

### Environment

0.53.0
pkg install

```yml
# insert yaml contents here
```

empty_enum_arguments is on by default

No
Version 15.0 (15A240d)
See screenshot above or sample code below

```swift
    static func swiftLintFalsePositive(for unit: HKUnit) {
        switch unit {
        case .foot(), .inch():
            print("foot or inch")
            
        case .meter(), .meterUnit(with: .kilo):
            print("metric units")
            
        default:
            print("Something else")
        }
    }
```

SimplyDanny:
The Swift compiler now complains about these anyway.

Fixes #5269 indirectly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,12 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#5277](https://github.com/realm/SwiftLint/issues/5277)
 
+* Remove arguments in `empty_enum_arguments` rule only if placeholders are
+  used. The Swift compiler complains now anyway about empty arguments on enums
+  cases without associated values.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+  [#5269](https://github.com/realm/SwiftLint/issues/5269)
+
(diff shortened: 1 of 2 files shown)


SimplyDanny:
For enum cases with associated values of type `Void` it's allowed to keep the empty parentheses. That's impossible to know on syntax-level. So we have to accept false positives - probably the existing ones.

PR #5286, opened 2023-10-17
Update README with the macro validation skipping instructions

slavikus:
Xcode 15 introduces an additional security check that requires macros to be explicitely enabled, causing the following error during the build:

> Target 'SwiftLintCoreMacros' must be enabled before it can be used.

In CI environments, this is normally not possible. Fortunately, there are new user defaults key to bypass this that I just reverse engineered. Updating the README to include this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,13 +221,13 @@ Select `SwiftLintPlugin` from the list and add it to the project.
 
 ![](https://raw.githubusercontent.com/realm/SwiftLint/main/assets/select-swiftlint-plugin.png)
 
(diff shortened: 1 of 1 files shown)


SimplyDanny (approved):
Thanks for adding these details!

jaredsinclair:
@slavikus Have you, or someone you know, actually used these two flags specifically with Xcode Cloud? I'm not able to get things to work, and Google search turns up so few results that this PR here is the closest I've come to finding someone who may actually have attempted these flags with Xcode Cloud.

Issue #5205, opened 2023-09-05
Xcode 15 Beta 8 + Xcode Cloud Error


dcilia:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

Hello.  We are utilizing the SwiftLint plugin as a build phase in our project containing:

- iOS App Target
- iOS App Clip Target  

When building with Xcode Cloud using Xcode Beta (8 in this case) a build error is encountered related to SwiftLint.

```
2023-09-05T18:13:05.391803418Z	2023-09-05 11:12:56.500 xcodebuild[10687:38398]  DVTFilePathFSEvents: Failed to start fs event stream.
2023-09-05T18:13:05.391812273Z	Error: You don’t have permission to save the file “f50a50e51ef5bdd780b98209347c4360c491f7299bcc424e3316162e78de3c24.plist” in the folder “478B1767-7422-3827-9D90-A09D769C708F”.
2023-09-05T18:13:05.391819938Z	
2023-09-05T18:13:05.391824842Z	/usr/bin/sandbox-exec -p "(version 1)
2023-09-05T18:13:05.391830105Z	(deny default)
2023-09-05T18:13:05.391849244Z	(import \"system.sb\")
2023-09-05T18:13:05.391854386Z	(allow file-read*)
2023-09-05T18:13:05.391859099Z	(allow process*)
2023-09-05T18:13:05.391863573Z	(allow file-write*
2023-09-05T18:13:05.391886045Z	    (subpath \"/private/tmp\")
2023-09-05T18:13:05.391896318Z	    (subpath \"/private/var/folders/x7/fs459gdd67132_hw5t3cz_s00000gn/T\")
2023-09-05T18:13:05.391913502Z	)
2023-09-05T18:13:05.391918327Z	(deny file-write*
2023-09-05T18:13:05.391923035Z	    (subpath \"/Volumes/workspace/repository\")
2023-09-05T18:13:05.391928353Z	)
2023-09-05T18:13:05.391933923Z	(allow file-write*
2023-09-05T18:13:05.391938967Z	    (subpath \"/Volumes/workspace/DerivedData/SourcePackages/plugins/**REDACTED**.output/**REDACTED**/SwiftLintPlugin\")
```

```
[...]

PR #5287, opened 2023-10-18
#5205 Explicitely disable cache for SwiftLintPlugin runs on Xcode Cloud

slavikus:
In Xcode Cloud environment, SwiftLint’s cache cannot be written.

When using the SwiftLinkPlugin, there is no way to disable the cache. Previously, a solution was made for the SwiftLint CLI itself where it looks at a set of environment variables (#4485).

This solution offers a cleaner approach where the plugin itself decides whether it needs to enable or disable the cache based on the `CI_XCODE_CLOUD` environment variable.

### DIFF
--- a/Plugins/SwiftLintPlugin/SwiftLintPlugin.swift
+++ b/Plugins/SwiftLintPlugin/SwiftLintPlugin.swift
@@ -32,9 +32,16 @@ struct SwiftLintPlugin: BuildToolPlugin {
             // We always pass all of the Swift source files in the target to the tool,
             // so we need to ensure that any exclusion rules in the configuration are
             // respected.
-            "--force-exclude",
-            "--cache-path", "\(workingDirectory)"
+            "--force-exclude"
         ]
+
+        // Determine whether we need to enable cache or not (for Xcode Cloud we don't)
+        if ProcessInfo.processInfo.environment["CI_XCODE_CLOUD"] == "TRUE" {
+            arguments.append("--no-cache")
+        } else {
+            arguments.append("--cache-path")
+            arguments.append("\(workingDirectory)")
+        }
 
         // Manually look for configuration files, to avoid issues when the plugin does not execute our tool from the
         // package source directory.


SimplyDanny:
Do you say that the approach from #4485 doesn't work anymore?

slavikus:
> Do you say that the approach from #4485 doesn't work anymore?

Correct. I suspect either the set of environmental variables has changed, or they are no longer passed down to the invoked process in the cloud run. Unfortunately there is no way to get any output from the launched swiftlint tool, but either way I believe gating cache from the SwiftLintPlugin is the right thing to do as it's being directly invoked by the xcodebuild during the Xcode cloud builds. If someone is using a shell script phase approach to directly launch SwiftLint CLI, they are free to determine and provide heuristics themselves as they have finer control over the launch parameters as opposed to the plugin which operates with a hardcoded set.

simba909:
For reference (I know it's in the title but GitHub doesn't link that way): this fixes #5205

SimplyDanny:
And why can't we just update `isLikelyXcodeCloudEnvironment`?

Or are you saying SwiftLint should get rid of `isLikelyXcodeCloudEnvironment` in the long run and let calling scripts decide about caching?

slavikus:
> Or are you saying SwiftLint should get rid of `isLikelyXcodeCloudEnvironment` in the long run and let calling scripts decide about caching?

Conceptually I do think it's best to offload this selection to a particular tool that interacts with Xcode directly (in this case, a plugin). The command line linter should be platform-agnostic and perhaps making it know about the environment is not the most efficient approach.

If you'd like, I can extend this PR to remove the CLI checks and include some instructions on invoking the swiftlint directly from the run script phase for the Xcode cloud (if someone prefers that approach).

The reason I didn't right away is because I try to minimize negating work already done by others who has invested significant time in it.

SimplyDanny:
So this makes sense to me then. In the long run, SwiftLint should get rid of `isLikelyXcodeCloudEnvironment` entirely. I don't much like that the decision for using the cache is now spread over (at least) three places.

SimplyDanny (approved):

PR #5292, opened 2023-10-19
Check disabled regions for all syntax types in parent rewriter class

SimplyDanny:


jpsim (approved):
Finally! This has been bugging me ever since we started using SwiftSyntax. Thanks 🙏

PR #5298, opened 2023-10-23
Fix CocoaPods job + Swift 5.9.1

marcelofabri:
We don't really need to test `pod lib lint` against an iOS simulator, so let's just use `macos` to make CI happy again


jpsim (approved):

SimplyDanny:
Aha ... that's the way. 😅

PR #5300, opened 2023-10-25
Fix implicit_getter false positive

kabiroberai:
The `implicit_getter` rule wasn't aware of the `init` accessor ([SE-0400](https://github.com/apple/swift-evolution/blob/main/proposals/0400-init-accessors.md)), causing it to incorrectly trigger on the following example

```swift
var foo: String {
  init { ... }
  get { ... } // implicit_getter violation
}
```

Rather than adding a check for `initAccessor` I updated the check to bail early unless `accessorsList` has a single element, that being `get`. This allows for forwards-compatibility if another accessor variant were to be introduced in the future. Semantically, IIUC this rule should only trigger in the aforementioned case that there's one accessor and that accessor is `get` — otherwise it isn't possible to elide the keyword.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Style/ImplicitGetterRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/ImplicitGetterRule.swift
@@ -30,8 +30,8 @@ private enum ViolationKind {
 private extension ImplicitGetterRule {
     final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
         override func visitPost(_ node: AccessorBlockSyntax) {
-            guard let getAccessor = node.getAccessor,
-                  node.accessorsList.count == 1,
+            guard node.accessorsList.count == 1,
+                  let getAccessor = node.getAccessor,
                   getAccessor.effectSpecifiers == nil,
                   getAccessor.modifier == nil,
                   getAccessor.attributes.isEmpty == true,
--- a/Source/SwiftLintBuiltInRules/Rules/Style/ImplicitGetterRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/ImplicitGetterRule.swift
@@ -31,7 +31,7 @@ private extension ImplicitGetterRule {
     final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
         override func visitPost(_ node: AccessorBlockSyntax) {
             guard let getAccessor = node.getAccessor,
-                  node.setAccessor == nil,
+                  node.accessorsList.count == 1,
                   getAccessor.effectSpecifiers == nil,
                   getAccessor.modifier == nil,
                   getAccessor.attributes.isEmpty == true,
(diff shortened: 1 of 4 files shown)


Source/SwiftLintBuiltInRules/Rules/Style/ImplicitGetterRuleExamples.swift
```diff
@@ -30,6 +30,15 @@ struct ImplicitGetterRuleExamples {
                 }
             }
             """),
+        Example("""
+            class Foo {
+                var foo: Int {
+                    @storageRestrictions(initializes: _foo)
```
SimplyDanny, line 36: `_foo` should be a property of the class to make it valid Swift, right?

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitGetterRule.swift
```diff
@@ -31,7 +31,7 @@ private extension ImplicitGetterRule {
     final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
         override func visitPost(_ node: AccessorBlockSyntax) {
             guard let getAccessor = node.getAccessor,
-                  node.setAccessor == nil,
+                  node.accessorsList.count == 1,
```
SimplyDanny, line 34: As an optimization, we could do the check for the number of elements first to avoid searching for `get` in `getAccessor`.

SimplyDanny (approved):
Looks good to me. Thanks!

Please also add an entry to the CHANGELOG to credit yourself and let others know about the fix.

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitGetterRuleExamples.swift
kabiroberai: Yep, it gets through parsing but it would probably fail during sema. I left it out from the example following suit with the previous example that referred to `_abc`; assumed that was for brevity. Want me to add an explicit `_foo` property here (and maybe an `_abc` to the example right before this one)?

kabiroberai: Updated this particular instance — thinking about it a bit I'm not sure if the other examples _need_ to be updated considering that `_abc` etc. could conceivably be declared elsewhere (e.g. global computed vars). Seems almost all the examples in this file use `_abc`/`_foo`/... and adding the backing prop could add syntactic noise. That said if you do feel strongly about the change I'd be happy to add the backing props elsewhere.

SimplyDanny: Fair enough. Other examples require some cleanup too, but that's not your job in this PR. Follow-ups are always welcome though.

SimplyDanny:
You need to remove the `swiftlint:disable` command [here](https://github.com/realm/SwiftLint/blob/main/Source/SwiftLintCore/Models/RuleConfigurationDescription.swift#L400) now to make the tests pass.

kabiroberai:
@SimplyDanny do you know what Danger is complaining about?

SimplyDanny:
> @SimplyDanny do you know what Danger is complaining about?

It doesn't like merge commits. Please rebase and force-push instead.

kabiroberai:
@SimplyDanny looks like we're green!

PR #5301, opened 2023-10-25
Use `_modify` accessor to allow `ConfigurationElement` to be a struct type

SimplyDanny:


jpsim (approved):

Source/SwiftLintCore/Models/RuleConfigurationDescription.swift
```diff
@@ -390,12 +390,16 @@ public protocol InlinableOptionType: AcceptableByConfigurationElement {}
 ///            error: 2
 ///    ```
 @propertyWrapper
-public class ConfigurationElement<T: AcceptableByConfigurationElement & Equatable> {
+public struct ConfigurationElement<T: AcceptableByConfigurationElement & Equatable> {
```
jpsim, line 393: Should `AcceptableByConfigurationElement` itself have the `Equatable` requirement since it doesn't really seem useful to conform non-equatable values to that protocol?

PR #5302, opened 2023-10-27
Speed up closure_parameter_position when there are no violations

marcelofabri:
In a large codebase without violations, this change made this rule by ~25% (~8.4s -> ~6s). Let's see if oss-check agrees.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@
 * Make `unneeded_break_in_switch` auto correctable.  
   [KS1019](https://github.com/KS1019/)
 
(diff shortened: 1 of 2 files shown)


SimplyDanny (approved):
Looks like the OSS checks agree in parts.

marcelofabri:
> Looks like the OSS checks agree in parts.

oss-check is running all the rules which makes a change in a single rule hard to visualize. I'll rebase after I merge https://github.com/realm/SwiftLint/pull/5303

marcelofabri:
Well, still not showing the gains I was hoping. Since it doesn't make things worse, I'll merge it because I think it can matter for large projects.

PR #5304, opened 2023-10-27
Conform InvalidSwiftLintCommandRule to SourceKitFreeRule

marcelofabri:
This rule doesn't use SourceKit under the hood, so it should conform to `SourceKitFreeRule`, making it easier to figure out which rules still depend on SourceKit.

(For context, in our codebase we're trying to only use rules without SourceKit to avoid paying its fixed cost)

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/InvalidSwiftLintCommandRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/InvalidSwiftLintCommandRule.swift
@@ -1,4 +1,4 @@
-struct InvalidSwiftLintCommandRule: Rule {
+struct InvalidSwiftLintCommandRule: Rule, SourceKitFreeRule {
     var configuration = SeverityConfiguration<Self>(.warning)
 
     static let description = RuleDescription(


SimplyDanny:
Why is this so much faster now? 🤔 Is it because of #5303?

SimplyDanny (approved):

marcelofabri:
> Why is this so much faster now? 🤔 Is it because of #5303?

This is weird. The only possible explanation I could think of is that we're not making any SourceKit requests anymore when this is the only enabled rule, but that should've already been the case before this PR. Searching for SourceKitFreeRule also doesn't show anything that would explain this speed up 🤔

marcelofabri:
OK I think I was both right and wrong at the same time :)

It's faster because we _are_ making SourceKit requests, it just wasn't obvious to me.

In `Linter`, we have this condition check:

https://github.com/realm/SwiftLint/blob/8e34619576b3212d2ed5fbc8bef2a259eacfe637/Source/SwiftLintCore/Models/Linter.swift#L70-L73

What I didn't get initially is that `file.sourcekitdFailed` actually can make a SourceKit request:

https://github.com/realm/SwiftLint/blob/8e34619576b3212d2ed5fbc8bef2a259eacfe637/Source/SwiftLintCore/Extensions/SwiftLintFile%2BCache.swift#L104-L107

https://github.com/realm/SwiftLint/blob/8e34619576b3212d2ed5fbc8bef2a259eacfe637/Source/SwiftLintCore/Extensions/SwiftLintFile%2BCache.swift#L13-L22

So this explains the performance boost when `InvalidSwiftLintCommandRule` is the only enabled rule (or the only enabled rule that uses SourceKit).

PR #5308, opened 2023-10-29
Rewrite cyclomatic_complexity with SwiftSyntax

marcelofabri:


marcelofabri:
I've looked into the [only change oss-check is reporting](https://github.com/brave/brave-ios/blob/b14b83e7373dc07e7a31f534223f648e5f8cfbbc/Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift#L192:10) and I think this PR is actually fixing a bug.

Looks like the old implementation could end up excluding `fallthrough` multiple times given how it'd call itself recursively without excluding `fallthrough`s that were already counted - the old impl also didn't guarantee `fallthrough` as a keyword :)

Source/SwiftLintBuiltInRules/Rules/Metrics/CyclomaticComplexityRule.swift
```diff
@@ -69,70 +70,109 @@ struct CyclomaticComplexityRule: ASTRule {
...
         }
 
-        return complexity
-    }
+        override func visitPost(_ node: WhileStmtSyntax) {
+            complexity += 1
+        }
 
-    // Switch complexity is reduced by `fallthrough` cases
+        override func visitPost(_ node: CatchClauseSyntax) {
```
marcelofabri, line 144: this is a bit weird, and maybe not intentional but this snippet:

```swift
do {} catch {}
```

contains a `case` according to SourceKitten:

```json
{
  "key.length" : 14,
  "key.offset" : 0,
  "key.substructure" : [
    {
      "key.bodylength" : 0,
      "key.bodyoffset" : 4,
      "key.kind" : "source.lang.swift.stmt.brace",
      "key.length" : 2,
      "key.offset" : 3
    },
    {
      "key.elements" : [
        {
          "key.kind" : "source.lang.swift.structure.elem.pattern",
          "key.length" : 1,
          "key.offset" : 12
        }
      ],
      "key.kind" : "source.lang.swift.stmt.case",
      "key.length" : 8,
      "key.offset" : 6
    }
  ]
}
```

since it's kind of a different branch I decided to keep counting it.

jpsim (approved):
Good job! I had tried this one a while back and didn't crack it, but reading through this one looks correct to me!

Source/SwiftLintBuiltInRules/Rules/Metrics/CyclomaticComplexityRule.swift
jpsim: The fact that SourceKit thinks that's a case statement is weird, but it's true that there are two branches in `do {} catch {}`: one where only `do` is invoked and another where `do` and `catch` are invoked.

PR #5313, opened 2023-10-30
Rewrite DiscouragedOptionalCollectionRule with SwiftSyntax

jpsim:


jpsim:
I have to make some adjustments based on the OSSCheck results.

jpsim:
I spot checked the differences reported by OSSCheck when ignoring columns and only looking at lines for rule violations and it all seems to be legit. Would like another maintainer to review and sign off before I merge though.

SimplyDanny (approved):
The new implementation looks impressively easy, especially compared to the previous one.

The OSS report looks good as well. I cannot think of another case that might be missing. The examples seem quite exhaustive, so the likelihood must be low for something to be not taken care of.

jpsim:
Thanks for taking a look @SimplyDanny 🙏

PR #5315, opened 2023-10-31
Sample change

mildm8nnered:
testing oss-check

please ignore


jpsim:
Is it possible that this rule has some non-determinism that causes these differences in OSSCheck?

mildm8nnered:
> Is it possible that this rule has some non-determinism that causes these differences in OSSCheck?

Something weird is definitely going on - 1,537 warnings after rebasing - I was pretty close to `main` before rebasing anyway.

I'll take a closer look.

mildm8nnered:
So I get absolutely consistent results when running SwiftLint locally on the command line, either from home-brew, or building from `main`, or when running this branch.

I get completely inconsistent results when running `oss-check` locally, and in CI on this PR.

On this PR, pushing empty commits:

```
1,573
1,582
2
```

Locally I have

```
61
1,575
1,276
```

This file appears to be causing a lot of the errors: https://raw.githubusercontent.com/apple/swift/a8d43c7fcbf2cd6e412e4979a926acaec2f20ceb/benchmark/multi-source/PrimsSplit/Prims_main.swift

So ... something weird definitely seems to be going on

PR #5317, opened 2023-10-31
Rewrite DuplicateImportsRule with SwiftSyntax

jpsim:


jpsim:
The two newly reported violations from OSSCheck were false negatives, so I think this is good to go.

PR #5320, opened 2023-11-01
Compile with `-strict-concurrency=complete`

jpsim:
Only in Bazel for now, because this is considered an unsafe flag in SwiftPM which would lead to warnings for downstream consumers of SwiftLint using SwiftPM.

Some imports of SwiftSyntax need the `@preconcurrency` annotation until https://github.com/apple/swift-syntax/pull/2322 is available in a release.

The following SwiftLint libraries have `-strict-concurrency=complete` applied:

* SwiftLintCoreMacros
* SwiftLintBuiltInRules (CI only)
* SwiftLintExtraRules

The following SwiftLint libraries don't have the flag applied and need to be migrated:

* SwiftLintCore
* swiftlint (CLI target)

So really the rules and macros are now being compiled with `-strict-concurrency=complete`, but the core infrastructure of SwiftLint is not.

Still, given that Swift 6 will eventually make these warnings errors by default, it's good to prevent issues from creeping in earlier rather than later.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -5,6 +5,14 @@ steps:
       - bazel build :swiftlint
       - echo "+++ Test"
       - bazel test --test_output=errors //Tests/...
+  - label: "Build With Strict Concurrency"
+    commands:
+      - echo "+++ Add @preconcurrency imports"
+      - ./tools/add-preconcurrency-imports.sh
+      - echo "+++ Build"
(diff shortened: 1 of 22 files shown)


jpsim:
Sadly this leads to warnings in SwiftPM since it doesn't have the flag:

> remark: '@preconcurrency' attribute on module 'SwiftSyntax' is unused

So might not want to enable that. 🤔

Issue #5305, opened 2023-10-28
`no_magic_numbers` doesnt work with tuple destructuring


laralove143:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

code like
```swift
let (httpStatusCodeErrorLowerBound, httpStatusCodeErrorUpperBound) = (400, 599)
```
still reports
`No Magic Numbers Violation: Magic numbers should be replaced by named constants (no_magic_numbers)`
while similar code
```swift
let httpStatusCodeErrorLowerBound = 400
let httpStatusCodeErrorUpperBound = 599
```
doesnt cause any lint violations

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint lint
Linting Swift files in current working directory
Linting 'Test.swift' (1/1)
Test.swift:2:75: warning: No Magic Numbers Violation: Magic numbers should be replaced by named constants (no_magic_numbers)
Test.swift:21:80: warning: No Magic Numbers Violation: Magic numbers should be replaced by named constants (no_magic_numbers)
Done linting! Found 2 violations, 0 serious in 1 file.
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.53.0
* Installation method used (Homebrew, CocoaPods, building from source, etc)? SPM Plugin
* Paste your configuration file:
```yml
disabled_rules:
  - force_try
  - trailing_comma

opt_in_rules:
  - accessibility_label_for_image
  - accessibility_trait_for_button
  - anonymous_argument_in_multiline_closure
  - array_init
  - attributes
  - balanced_xctest_lifecycle
  - closure_body_length
  - closure_end_indentation
  - closure_spacing
  - collection_alignment
  - comma_inheritance
  - conditional_returns_on_newline
  - contains_over_filter_count
  - contains_over_filter_is_empty
  - contains_over_first_not_nil
  - contains_over_range_nil_comparison
  - convenience_type
  - direct_return
  - discarded_notification_center_observer
  - discouraged_assert
  - discouraged_none_name
  - discouraged_object_literal
  - discouraged_optional_boolean
  - discouraged_optional_collection
  - empty_collection_literal
  - empty_count
  - empty_string
  - empty_xctest_method
  - enum_case_associated_values_count
  - expiring_todo
  - explicit_enum_raw_value
  - explicit_init
  - extension_access_modifier
[...]

PR #5327, opened 2023-11-04
skip tuples for `no_magic_numbers` rule

mildm8nnered:
Now ignores magic numbers in tuple assignments like

`let (httpStatusCodeErrorLowerBound, httpStatusCodeErrorUpperBound) = (400, 599)`

OSS checks look good to me.

Resolves https://github.com/realm/SwiftLint/issues/5305 by skipping magic numbers in tuple expressions altogether.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 * Trigger `nsobject_prefer_isequal` and `redundant_self_in_closure` even in case
   the surrounding declaration is nested in an extension.  
   [SimplyDanny](https://github.com/SimplyDanny)
+
 * Fixed false positives for the `no_magic_numbers` rule, when they
   are defined in a tuple like `let (a, b) = (5, 10)` or `let a = (2, 3)`.  
   [Martin Redington](https://github.com/mildm8nnered)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,7 @@
 * `superfluous_disable_command` violations are now triggered for
   custom rules.  
   [Marcelo Fabri](https://github.com/marcelofabri)
+  [Martin Redington](https://github.com/mildm8nnered)
   [#4754](https://github.com/realm/SwiftLint/issues/4754)
 
 * Fixed false positives for the `no_magic_numbers` rule, when they
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,20 +82,30 @@
   [BB9z](https://github.com/BB9z)
   [#5289](https://github.com/realm/SwiftLint/issues/5289)
 
-* Fix invalid corrections for opaque and existential optionals in 
+* Fix invalid corrections for opaque and existential optionals in
   `syntactic_sugar` rule.  
   [SimplyDanny](https://github.com/SimplyDanny)
   [#5277](https://github.com/realm/SwiftLint/issues/5277)
 
-* Fix false positive in `unused_import` rule that triggered on 
+* Fix false positive in `unused_import` rule that triggered on
   `@_exported` imports which could break downstream modules if removed.  
   [jszumski](https://github.com/jszumski)
   [#5242](https://github.com/realm/SwiftLint/pull/5242)
 
-* Fix false positive in `unused_import` rule when using a constructor 
+* Fix false positive in `unused_import` rule when using a constructor
   defined in a transitive module.  
   [jszumski](https://github.com/jszumski)
   [#5246](https://github.com/realm/SwiftLint/pull/5246)
+
+* `superfluous_disable_command` violations are now triggered for
+  custom rules.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#4754](https://github.com/realm/SwiftLint/issues/4754)
+* Fixed false positives for the `no_magic_numbers` rule, when they
+  are defined in a tuple like `let (a, b) = (5, 10)`.  
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#5305](https://github.com/realm/SwiftLint/pull/5305)
 
 ## 0.53.0: Laundry List
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@
 * Trigger `nsobject_prefer_isequal` and `redundant_self_in_closure` even in case
   the surrounding declaration is nested in an extension.  
   [SimplyDanny](https://github.com/SimplyDanny)
+* Fixed false positives for the `no_magic_numbers` rule, when they
+  are defined in a tuple like `let (a, b) = (5, 10)` or `let a = (2, 3)`.  
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#5305](https://github.com/realm/SwiftLint/pull/5305)
 
 ## 0.54.0: Macro-Economic Forces
 
@@ -96,17 +100,6 @@
   defined in a transitive module.  
   [jszumski](https://github.com/jszumski)
   [#5246](https://github.com/realm/SwiftLint/pull/5246)
-
-* `superfluous_disable_command` violations are now triggered for
-  custom rules.  
-  [Marcelo Fabri](https://github.com/marcelofabri)
-  [Martin Redington](https://github.com/mildm8nnered)
-  [#4754](https://github.com/realm/SwiftLint/issues/4754)
-
-* Fixed false positives for the `no_magic_numbers` rule, when they
-  are defined in a tuple like `let (a, b) = (5, 10)` or `let a = (2, 3)`.  
-  [Martin Redington](https://github.com/mildm8nnered)
-  [#5305](https://github.com/realm/SwiftLint/pull/5305)
 
 ## 0.53.0: Laundry List
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,10 +100,10 @@
 * `superfluous_disable_command` violations are now triggered for
   custom rules.  
   [Marcelo Fabri](https://github.com/marcelofabri)
-  [Martin Redington](https://github.com/mildm8nnered)
   [#4754](https://github.com/realm/SwiftLint/issues/4754)
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -50,21 +50,26 @@
   [BB9z](https://github.com/BB9z)
   [#5289](https://github.com/realm/SwiftLint/issues/5289)
 
-* Fix invalid corrections for opaque and existential optionals in
```
mildm8nnered, line 53: Fixed some trailing spaces while I was here

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -74,7 +74,8 @@ struct NoMagicNumbersRule: OptInRule {
             Example("let foo = 1 >> 2"),
             Example("let foo = 2 >> 2"),
             Example("let foo = 2 << 2"),
-            Example("let a = b / 100.0")
+            Example("let a = b / 100.0"),
+            Example("let (httpStatusCodeErrorLowerBound, httpStatusCodeErrorUpperBound) = (400, 599)")
```
SimplyDanny, line 78: The variable names could be shorter.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -88,7 +89,13 @@ struct NoMagicNumbersRule: OptInRule {
                     extension NSObject {
                         let a = Int(↓3)
                     }
-            """)
+            """),
+            Example("""
+            if (fileSize > ↓1000000) {
+                try FileManager.default.removeItem(at: cacheFileUrl)
```
SimplyDanny, line 95: That's unnecessary code actually.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
SimplyDanny, line 78: What about the case `let a = (2, 3)`? Is it not supposed to be ignored by the rule?

mildm8nnered: So I wasn't intending to cover that case, but it would be easy to. Do you think we should?

SimplyDanny: I think so. `let (a, b) = (2, 3)` is only a special case of `let a = (2, 3)`. But with regards to magic numbers it should not make a difference.

CHANGELOG.md
```diff
@@ -50,21 +50,26 @@
...
   [#5242](https://github.com/realm/SwiftLint/pull/5242)
 
-* Fix false positive in `unused_import` rule when using a constructor 
+* Fix false positive in `unused_import` rule when using a constructor
   defined in a transitive module.  
   [jszumski](https://github.com/jszumski)
   [#5246](https://github.com/realm/SwiftLint/pull/5246)
 
+* Fixed false positives for the `no_magic_numbers` rule, when they
+  are defined in a tuple like `let (a, b) = (5, 10)`.
```
SimplyDanny, line 69:
```suggestion
  are defined in a tuple like `let (a, b) = (5, 10)` or `let a = (2, 3).  
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -212,3 +224,20 @@ private extension ExprSyntaxProtocol {
...
+            return false
+        }
+        let secondChildIndex = children.index(after: children.startIndex)
+
+        guard let secondChild = children[secondChildIndex].as(InitializerClauseSyntax.self),
+              let tupleExpression = secondChild.value.as(TupleExprSyntax.self),
+              tupleExpression.elements.count > 1  else {
+            return false
+        }
+        return true
```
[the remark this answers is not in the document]
mildm8nnered: So much nicer

SimplyDanny:
Looking at the OSS findings, there's another case I haven't thought of before. If there are more complicated expression in a tuple initialization like in `let a = (2 + 3, f(arg: 47))`, the rule won't trigger now. Is this expected? In `let a = f(4)` the rule would trigger on `4`, so not triggering in tuples seems like an inconsistency then.

mildm8nnered:
> Looking at the OSS findings, there's another case I haven't thought of before. If there are more complicated expression in a tuple initialization like in `let a = (2 + 3, f(arg: 47))`, the rule won't trigger now. Is this expected? In `let a = f(4)` the rule would trigger on `4`, so not triggering in tuples seems like an inconsistency then.

So `let a = f(4)` does trigger, even though it's technically a tuple, because there were OSS samples like `let imageHeight = (width - ↓24)` that I wanted to still warn on.

I *thought* the OSS violations we ended up with looked ok, but `let a = (2 + 3, f(arg: 47))` does bother me. That would not violate anymore.

Maybe drop back to just allowing

`let (a, b) = (5, 10)`

That would still allow

`let (a, b) = (2 + 3, f(arg: 47))`

or we could go further and require that the tuple elements all be simple expressions.

SimplyDanny:
> ... or we could go further and require that the tuple elements all be simple expressions.

I think that's the way to go. The rule currently triggers on `let a = 2 + 3` but not on `let a = 3`. It should be the same for tuples in my opinion. I'm open for better arguments though.

mildm8nnered:
> > ... or we could go further and require that the tuple elements all be simple expressions.
> 
> I think that's the way to go. The rule currently triggers on `let a = 2 + 3` but not on `let a = 3`. It should be the same for tuples in my opinion. I'm open for better arguments though.

Sounds good to me -  `let (a, b) = (5, 10)` all good. And a nope for anything else, or complex expressions.

SimplyDanny:
> > > ... or we could go further and require that the tuple elements all be simple expressions.
> >
> > I think that's the way to go. The rule currently triggers on `let a = 2 + 3` but not on `let a = 3`. It should be the same for tuples in my opinion. I'm open for better arguments though.
> 
> Sounds good to me -  `let (a, b) = (5, 10)` all good. And a nope for anything else, or complex expressions.

Just to be sure: Also `let a = (3, 2)` wouldn't trigger, right?

mildm8nnered:
> > > > ... or we could go further and require that the tuple elements all be simple expressions.
> > > 
> > > 
> > > I think that's the way to go. The rule currently triggers on `let a = 2 + 3` but not on `let a = 3`. It should be the same for tuples in my opinion. I'm open for better arguments though.
> > 
> > 
> > Sounds good to me -  `let (a, b) = (5, 10)` all good. And a nope for anything else, or complex expressions.
> 
> Just to be sure: Also `let a = (3, 2)` wouldn't trigger, right?

ok - so I think I have this working. 

As long as every element of the RHS tuple is an Integer or Float, we will allow it through, so

`let a = (3, 2)` is fine, as is `let (a, b) = (3, 2)`

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -74,7 +74,9 @@ struct NoMagicNumbersRule: OptInRule {
             Example("let foo = 1 >> 2"),
             Example("let foo = 2 >> 2"),
             Example("let foo = 2 << 2"),
-            Example("let a = b / 100.0")
+            Example("let a = b / 100.0"),
+            Example("let (lowerBound, upperBound) = (400, 599)"),
+            Example("let a = (5, 10)")
```
SimplyDanny, line 79: Let's also have a test for labeled tuple elements and a tuple used somewhere else that triggers the rule.

After that, I promise to accept the change. 😇

mildm8nnered: So I added those examples, and also extended the allowed tuple elements to include string literals, as cases like `let notFound = (404, "Not Found")` seemed reasonable to skip as well.

mildm8nnered: But hold off on merging for right now @SimplyDanny  - I want to checkout the OSS results, which now seem to have suspiciously lower fixes compared to before

mildm8nnered: ok - that looks better. Should be good to go now once CI passes

SimplyDanny: The strings and booleans make me think again. We now have:

```swift
let a = (3, 4)      // ok
let a = (3, "a")    // ok 
let a = (3, b)      // triggers on 3
let a = (3, (4, 5)) // triggers on all numbers
let a = (true, 4)   // ok
let a = (3, 4 + 5)  // triggers on all numbers
```

Does this make sense? Is it consistent and logical? 🤔 We might need to come up with a better set of rules for tuples I'm afraid.

mildm8nnered: So I totally get your point, but actually, all of the above examples would have been violations before, and there were enough reasonably allowable fixed violations in the OSS tests, that I think the change will add some value for some users.

```
let a = (3, b)      // triggers on 3
let a = (3, (4, 5)) // triggers on all numbers
let a = (3, 4 + 5)  // triggers on all numbers
```

will still be violations, but in the OSS code, there were few or none of these.

Users of the above examples would not be put at any disadvantage by this change - they would just need to disable the rule, like they would today.

I could definitely see a case for allowing `let a = (3, (4, 5))`, but if someone wants that to pass, they can always request that we extend the coverage.

SimplyDanny: Fair enough. Let's have it merged then. Good job!

SimplyDanny (approved):

PR #5328, opened 2023-11-05
Rewrite no_grouping_extension rule with SwiftSyntax

marcelofabri:


Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoGroupingExtensionRule.swift
```diff
@@ -22,38 +23,100 @@ struct NoGroupingExtensionRule: OptInRule {
     )
 
     func validate(file: SwiftLintFile) -> [StyleViolation] {
-        let collector = NamespaceCollector(dictionary: file.structureDictionary)
-        let elements = collector.findAllElements(of: [.class, .enum, .struct, .extension])
+        return Visitor(configuration: configuration, file: file)
```
jpsim, line 26: Totally up to you but I find that using implicit returns and if-expressions from Swift 5.9 to make this more readable:

```swift
Visitor(configuration: configuration, file: file)
    .walk(file: file) { visitor in
        visitor.extensionDeclarations.compactMap { decl in
            if visitor.typeDeclarations.contains(decl.name) {
                ReasonedRuleViolation(position: decl.position)
            } else {
                nil
            }
        }
    }
    .sorted()
    .map { makeViolation(file: file, violation: $0) }
```

jpsim (approved):
Fantastic cleanup!

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoGroupingExtensionRule.swift
```diff
@@ -22,38 +23,100 @@ struct NoGroupingExtensionRule: OptInRule {
...
+        private(set) var extensionDeclarations = Set<ExtensionDeclaration>()
+
+        override var skippableDeclarations: [any DeclSyntaxProtocol.Type] {
+            [
+                ProtocolDeclSyntax.self,
+                FunctionDeclSyntax.self,
+                VariableDeclSyntax.self,
+                InitializerDeclSyntax.self,
+                SubscriptDeclSyntax.self
+            ]
```
SimplyDanny, line 59: I think it's better to use the `allExcept(...)` method here. With that, there is a direct mapping between declarations not to skip and the implemented `visit` methods.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoGroupingExtensionRule.swift
```diff
@@ -22,38 +23,100 @@ struct NoGroupingExtensionRule: OptInRule {
...
+
+private extension NoGroupingExtensionRule {
+    struct ExtensionDeclaration: Hashable {
+        let name: String
+        let position: AbsolutePosition
+    }
+
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        private(set) var typeDeclarations = Set<String>()
+        private var typeScope: [String] = []
```
SimplyDanny, line 49: There is a `Stack` type in SwiftLint.

marcelofabri: That type doesn't expose a way to iterate through the contents like I needed - and using an array is simple enough that I didn't want to bother adding one

marcelofabri:
I'll merge as is because honestly I don't care enough about this particular rule, just want to move the needle in moving us off of SourceKit :) 

Feel free to make any style changes in any other PRs

PR #5335, opened 2023-11-07
Adapt linter to prepare project for Swift6.

johnno1962:
Hi @jpsim, @marcelofabri,

I don't know what you'll make of this but this PR adds a function to the inter to prepare a project for the upcoming Swift 6 accessed by a new option `--experimental-swift6 <Package.Swift/Project.xcproject>`. It was born out of [this conversation](https://forums.swift.org/t/existential-any-migration-script/67875/11) on the Swift forums where someone was looking for a migration script to add the new any/some annotations on protocol references in Swift sources required by [SE0335](https://github.com/apple/swift-evolution/blob/main/proposals/0335-existential-any.md). This should hopefully save people a good deal of time as for some projects this can mean hundreds or even thousands of edits are necessary to migrate. It's been extensively tested with the top 1000 Swift packages from the Swift package index (ordered by GitHub stars as of two years ago). The results of this testing are available [here](https://johnholdsworth.com/opaqueify.txt) and you can see the majority of projects compile after running the opaqueify script I've incorporated into the linter. Those that do not generally have only few errors to correct.

The problem I imagine from your point of view is maintenance as the code rates fairly highly on the eccentricity scale as it uses a number of gnarly regular expressions, expressed using a library of my own making [SwiftRegex5](https://github.com/johnno1962/SwiftRegex5) from a time when this was not supported directly in the language. Although the primary analysis is done using SourceKit, these regular expressions cover edge cases you encounter processing a larger corpus of software. Ultimately, though the overall rule is simple: elide to some for procedure argument and return values and add any for everything else, you can improve the odds of a complete build by adding exceptions. The final stage of processing is running the project though the Xcode15 compiler with option `-enable-upcoming-feature ExistentialAny` which is the acid test. I've hard coded the path to /Applications/Xcode.app for want of a better solution.

I've made the library a package in itself and can support it for you as it has been extensively tested on real code.

Let me know your thoughts,

Cheers,

John


johnno1962:
Hi, not sure what the problems are but I'll be needing some help with Bazel it seems to survive CI.

johnno1962:
I can see this is going to be a non-starter as it pulls in packages which will not build on Linux. More thought required...

PR #4755, opened 2023-02-07
Trigger superfluous_disable_command for custom_rules

marcelofabri:
Fixes #4754

Need to write tests, but wanna make sure nothing breaks horribly first.

PR #5336, opened 2023-11-08
Fix custom_rules regression introduced in #4755

marcelofabri:
Here's what was happening:

- `CustomRules.swift` filters out violations outside of disabled regions
- This means that we won't get that violation in `Linter` so we'll always trigger `superfluous_disable_command`
- `Region.isRuleEnabled` doesn't understand custom rules

We can change the logic a little bit to make it work for custom rules. I don't love this, but I _think_ it works (and we already special case `CustomRules` in a few places).

### DIFF
--- a/Source/SwiftLintCore/Models/Linter.swift
+++ b/Source/SwiftLintCore/Models/Linter.swift
@@ -102,8 +102,13 @@ private extension Rule {
 
         let (disabledViolationsAndRegions, enabledViolationsAndRegions) = violations.map { violation in
             return (violation, regions.first { $0.contains(violation.location) })
-        }.partitioned { _, region in
-            return region?.isRuleEnabled(self) ?? true
+        }.partitioned { violation, region in
+            if self is CustomRules {
+                let disabled = region?.isRuleDisabled(customRuleIdentifier: violation.ruleIdentifier) ?? false
+                return !disabled
+            } else {
+                return region?.isRuleEnabled(self) ?? true
+            }
         }
 
         let customRulesIDs: [String] = {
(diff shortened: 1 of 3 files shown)


keith:
Looks like this might not be enough, with this commit I still see the issue, here's another example:

```swift
        if UIAccessibility.isVoiceOverRunning { // swiftlint:disable:this use_voice_over_provider
            return
        }
```

```
  use_voice_over_provider:
    name: "Use `getVoiceOverProvider()`."
    regex: '\bUIAccessibility\.isVoiceOverRunning\b'
    match_kinds:
      - identifier
      - typeidentifier
```

marcelofabri:
took a quick look and couldn't repro it with a unit test:

```swift
    func testSuperfluousDisableCommandDoesntTriggerWithCustomRulesDisableThis() throws {
        let customRulesConfiguration: [String: Any] = [
            "use_voice_over_provider": [
                "regex": "\\bUIAccessibility\\.isVoiceOverRunning\\b",
                "match_kinds": ["identifier", "typeidentifier"]
            ]
        ]

        let example = Example(
            """
            if UIAccessibility.isVoiceOverRunning {} // swiftlint:disable:this use_voice_over_provider
            """,
            configuration: customRulesConfiguration
        ).skipWrappingInCommentTest()
        let configuration = try XCTUnwrap(makeConfig(["custom_rules": customRulesConfiguration], "custom_rules"))
        let violations = violations(example, config: configuration)

        XCTAssertEqual(violations.count, 0)
    }
```

i'll try to dig deeper tonight, but i'm inclined to just revert the PR.

PR #5351, opened 2023-11-14
Rewrite `void_return` rule with SwiftSyntax

SimplyDanny:


jpsim (approved):
Textbook example of migrating a messy rule to a clean SwiftSyntax implementation 👌

PR #5354, opened 2023-11-15
Rewrite `identifier_name` rule with SwiftSyntax

SimplyDanny:


jpsim:
Thanks for tackling this migration! Lots of OSSCheck results here. Do you have any thoughts for how to audit those?

SimplyDanny:
Some of them are due to changed violation positions (e.g when declarations have attributes). That's an easy change.

For many though, I'm not sure if they are expected or wrong, e.g. variable declarations in `if let`. The existing test base is not very comprehensive in that regard.

While function declarations and enum elements are rather clear, I wonder what shall be considered variable declarations: class/struct fields, local variables, function parameters, `guard let`/`if let`, loop variables?

jpsim:
> For many though, I'm not sure if they are expected or wrong, e.g. variable declarations in `if let`. The existing test base is not very comprehensive in that regard.

I'm less concerned with how comprehensive the test suite is and I think minimizing the SwiftLint version upgrade churn is more important here.

If we can expand the rule's coverage to catch more kinds of identifiers like you've done here, but have a mode where we can trim that to be closer to what the rule previously caught, it could help folks migrate at their own pace instead of being hit with a ton of new violations when they update SwiftLint.

SimplyDanny:
Puh ... 😅 This has been much more work than I expected initially.

The remaining OSS findings are due to:

* Valid findings that were previously not discovered (e.g. functions declared in functions)
* Different message reported (the actual identifier is now used in the issue reason)
* Tuple declarations are now fully supported

On top of these, the syntax tree looks incorrect for code like `f { @Sendable (e: Int) in }` where the `(e: Int)` is parsed as the argument list of `@Sendable` instead of the closure's parameters. [This finding](https://github.com/realm/realm-swift/blob/95cb711d60801e82542b866d16d75608c40d7b2e/Realm/ObjectServerTests/SwiftObjectServerTests.swift#L369:53) stems from the potential bug in the parser.

The set of examples has been extended so that more cases are documented and tested. What's up for a follow-up PR is the adjustment of violation positions. They are quite inconsistent at the moment and complicate the code. Harmonizing them should not cause any harm, but would have made the OSS findings hard to review here.

jpsim:
I really appreciate how you approached this rewrite by first matching the old implementation and then making your improvements.

PR #5357, opened 2023-11-16
[bazel] Add optional universal macro support

keith:
This allows users to pass `--@SwiftLint//:universal_tools=true` in order
to compile the underlying macros as a universal binary, which aids in
caching across intel and arm macs


jpsim (approved):
TIL

PR #5361, opened 2023-11-19
Trigger on both first and second name

SimplyDanny:
Not sure I like this. While the first argument name is actually the user-visible one, it's often used to create a descriptive overall function name like `doSomething(if condition: Bool)`. This often leads to relatively short prepositions like `as`, `if`, `on`, `by`, ... which would now all trigger the rule.

If added, it should probably be hidden behind an option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 #### Enhancements
 
+* Given a function `f(a b: Int)`, the `identifier_name` rule now warns on both
+  parameter names `a` and `b` for being too short (in this case).  
+  [SimplyDanny](https://github.com/SimplyDanny)
+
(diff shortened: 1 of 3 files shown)


SimplyDanny:
Not sure I like this. While the first argument name is actually the user-visible one, it's often used to create a descriptive overall function name like `doSomething(if condition: Bool)`. This often leads to relatively short prepositions like `as`, `if`, `on`, `by`, ... which would now all trigger the rule.

If added, it should probably be hidden behind an option.

jpsim:
I don’t think this is in keeping with the spirit or intention of the rule.

SimplyDanny:
> I don’t think this is in keeping with the spirit or intention of the rule.

Agree.

PR #5367, opened 2023-11-27
Rewrite `let_var_whitespace` rule with SwiftSyntax

### LINKED ISSUE #4801: Multiline property wrappers cause false positives for `let_var_whitespace`
### New Issue Checklist

### Describe the bug

Given the following code, `let_var_whitespace` will fire at lines 4 and 8.

```
internal struct Foo {
    @Flag(name: [.long, .customLong("autocorrect")], help: "Correct violations whenever possible.")
    internal var fix = false
    @Flag(help: """
        Should reformat the Swift files using the same mechanism used by Xcode (via SourceKit).
        Only applied with `--fix`/`--autocorrect`.
        """)
    internal var format = false
    @Flag(help: "Use an alternative algorithm to exclude paths for `excluded`, which may be faster in some cases.")
    internal var useAlternativeExcluding = false
}
```

This code is actually from LintOrAnalyzeArguments.swift

```bash
% swiftlint lint --no-cache --enable-all-rules Foo.swift
Linting Swift files at paths Foo.swift
Linting 'Foo.swift' (1/1)
/tmp/Foo.swift:3:14: warning: Explicit Type Interface Violation: Properties should have a type interface (explicit_type_interface)
/tmp/Foo.swift:8:14: warning: Explicit Type Interface Violation: Properties should have a type interface (explicit_type_interface)
/tmp/Foo.swift:10:14: warning: Explicit Type Interface Violation: Properties should have a type interface (explicit_type_interface)
/tmp/Foo.swift:4:5: warning: Variable Declaration Whitespace Violation: Let and var should be separated from other statements by a blank line. (let_var_whitespace)
/tmp/Foo.swift:8:5: warning: Variable Declaration Whitespace Violation: Let and var should be separated from other statements by a blank line. (let_var_whitespace)
Done linting! Found 5 violations, 0 serious in 1 file.
```

### Environment

none

no

```
Xcode 14.2
Build version 14C18
```

See above.

SimplyDanny:
Fixes #4801.

What's more:

* Examples have been refactored to use multiline strings for better readability.
* Additional examples check more special cases.
* `typealias` and `#if`/`#endif` are also considered declarations to maintain a distance to.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,11 @@
 * Rewrite `multiline_literal_brackets` rule using SwiftSyntax, fixing some
   false positives that would happen when comments are present.  
   [Marcelo Fabri](https://github.com/marcelofabri)
+  
+* Rewrite `let_var_whitespace` rule using SwiftSyntax, fixing false positives
+  when attributes attached to declarations were spread over multiple lines.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+  [#4801](https://github.com/realm/SwiftLint/pull/4801)
 
 #### Bug Fixes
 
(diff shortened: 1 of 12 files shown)


SimplyDanny:
Remaining OSS findings are all valid.

PR #5382, opened 2023-12-10
Rewrite `explicit_acl` rule with SwiftSyntax

SimplyDanny:


SimplyDanny:
All remaining OSS findings are previous false positives that are now fixed.

PR #5385, opened 2023-12-10
Append to `.swiftlint.yml` config file in OSS check

SimplyDanny:


jpsim:
I’m not sure this is correct to do, like if any of these projects add / change their swiftlint config.

SimplyDanny:
> I’m not sure this is correct to do, like if any of these projects add / change their swiftlint config.

The configuration [is deleted](https://github.com/SimplyDanny/SwiftLint/blob/634bda3756e577e8cee055b91c37206dff789174/tools/oss-check#L144) before we add/append our specifics.

jpsim:
Ah great if we’re still doing that. Carry on!

PR #5389, opened 2023-12-11
RedundantTypeAnnotationRule: Rewrite with SwiftSyntax and allow ignoring certain attributes

### LINKED ISSUE #5366: SwiftLint Incorrectly Flags @Parameter Type Annotation as Redundant in TipKit Integration
### New Issue Checklist

### Describe the bug

While integrating TipKit, as documented [here](https://developer.apple.com/documentation/tipkit), a conflict arises with SwiftLint's handling of redundant type annotations. Specifically, SwiftLint incorrectly flags the type annotation in the @Parameter attribute as redundant, which, when auto-corrected, leads to a compilation failure.

```bash
$ swiftlint lint
# Output indicating 'redundant_type_annotation' violation for the @Parameter attribute
```

### Environment

* SwiftLint version 0.53.0
* Installation method used CocoaPods
* Xcode 15.0.1 - Build version 15A507

```swift
// This triggers a violation:
@Parameter
static var isLoggedIn: Bool = false
```
image
image

### Additional Context
SwiftLint's `redundant_type_annotation` rule suggests the removal of the for type annotation in the `@Parameter` attribute, which, if applied, results in a compilation error. A temporary workaround is to disable this specific rule using `// swiftlint:disable redundant_type_annotation` for the affected line. Are there any better solutions to keep this rule and support macros?

tonell-m:
Hi, first time contributing here so just let me know if something is wrong or missing in my PR 🙂 

This PR resolves #5366 by adding a new `ignored_annotations` configuration to this rule that accepts a set of strings listing annotations that should disable this rule for properties that are marked with it. Currently this is already done arbitrarily for `@IBInspectable` but now it is configurable. To solve the linked issue the configuration would be:

```yaml
redundant_type_annotations:
  ignored_attributes:
    - Parameter
```

This configuration still defaults to only `IBDesignable` for retrocompatibility.

And, while I was at it and as @SimplyDanny suggested in the issue's comments, I took the opportunity to rewrite this rule using SwiftSyntax. Let me know what you think!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,7 +71,10 @@
   Extensions are an exception; more than one is allowed.  
   [Muhammad Zeeshan](https://github.com/mzeeshanid)
   [#2802](https://github.com/realm/SwiftLint/issues/2802)
-* Rewrite `redundant_type_annotation` rule with SwiftSyntax and add a new `ignored_attributes` option that allows disabling the rule for properties that are marked with at least one of the configured attributes
+
+* Add new `ignored_attributes` option to `redundant_type_annotation` rule
+  that allows disabling the rule for properties that are marked with at least
+  one of the configured attributes
   [tonell-m](https://github.com/tonell-m)
   [#5366](https://github.com/realm/SwiftLint/issues/5366)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,7 +74,7 @@
 
 * Add new `ignore_attributes` option to `redundant_type_annotation` rule
   that allows disabling the rule for properties that are marked with at least
-  one of the configured attributes.
+  one of the configured attributes.  
   [tonell-m](https://github.com/tonell-m)
   [#5366](https://github.com/realm/SwiftLint/issues/5366)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,9 +89,8 @@
   * `nimble_operator`
   * `opening_brace`
   * `orphaned_doc_comment`
-  * `trailing_closure`
-  * `void_return`
   * `redundant_type_annotation`
+  * `trailing_closure`
   * `void_return`
 
   [SimplyDanny](https://github.com/SimplyDanny)  
@@ -99,7 +98,6 @@
   [Marcelo Fabri](https://github.com/marcelofabri)  
   [swiftty](https://github.com/swiftty)  
   [KS1019](https://github.com/KS1019)  
-  [Marcelo Fabri](https://github.com/marcelofabri)
   [tonell-m](https://github.com/tonell-m)
 
 * Print invalid keys when configuration parsing fails.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,7 +74,7 @@
 
 * Add new `ignore_attributes` option to `redundant_type_annotation` rule
   that allows disabling the rule for properties that are marked with at least
-  one of the configured attributes
+  one of the configured attributes.
   [tonell-m](https://github.com/tonell-m)
   [#5366](https://github.com/realm/SwiftLint/issues/5366)
 
@@ -92,6 +92,7 @@
   * `trailing_closure`
   * `void_return`
   * `redundant_type_annotation`
+  * `void_return`
 
   [SimplyDanny](https://github.com/SimplyDanny)  
   [kishikawakatsumi](https://github.com/kishikawakatsumi)  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,9 @@
   Extensions are an exception; more than one is allowed.  
   [Muhammad Zeeshan](https://github.com/mzeeshanid)
   [#2802](https://github.com/realm/SwiftLint/issues/2802)
+* Rewrite `redundant_type_annotation` rule with SwiftSyntax and add a new `ignored_attributes` option that allows disabling the rule for properties that are marked with at least one of the configured attributes
+  [tonell-m](https://github.com/tonell-m)
+  [#5366](https://github.com/realm/SwiftLint/issues/5366)
 
 * Rewrite the following rules with SwiftSyntax:
   * `explicit_acl`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,12 +91,15 @@
   * `orphaned_doc_comment`
   * `trailing_closure`
   * `void_return`
+  * `redundant_type_annotation`
 
   [SimplyDanny](https://github.com/SimplyDanny)  
   [kishikawakatsumi](https://github.com/kishikawakatsumi)  
   [Marcelo Fabri](https://github.com/marcelofabri)  
   [swiftty](https://github.com/swiftty)  
   [KS1019](https://github.com/KS1019)  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [tonell-m](https://github.com/tonell-m)
 
 * Print invalid keys when configuration parsing fails.  
   [SimplyDanny](https://github.com/SimplyDanny)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,7 +72,7 @@
   [Muhammad Zeeshan](https://github.com/mzeeshanid)
   [#2802](https://github.com/realm/SwiftLint/issues/2802)
 
-* Add new `ignored_attributes` option to `redundant_type_annotation` rule
+* Add new `ignore_attributes` option to `redundant_type_annotation` rule
   that allows disabling the rule for properties that are marked with at least
   one of the configured attributes
   [tonell-m](https://github.com/tonell-m)
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
@@ -2,12 +2,9 @@ import Foundation
 import SwiftLintCore
 import SwiftSyntax
 
-struct RedundantTypeAnnotationRule: SwiftSyntaxCorrectableRule, OptInRule {
+@SwiftSyntaxRule
+struct RedundantTypeAnnotationRule: OptInRule {
     var configuration = RedundantTypeAnnotationConfiguration()
-
-    func makeVisitor(file: SwiftLintFile) -> ViolationsSyntaxVisitor<ConfigurationType> {
-        Visitor(configuration: configuration, file: file)
-    }
 
     func makeRewriter(file: SwiftLintFile) -> (some ViolationsSyntaxRewriter)? {
         Rewriter(
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
@@ -13,7 +13,6 @@ struct RedundantTypeAnnotationRule: SwiftSyntaxCorrectableRule, OptInRule {
         nonTriggeringExamples: [
             Example("var url = URL()"),
             Example("var url: CustomStringConvertible = URL()"),
-            Example("@IBInspectable var color: UIColor = UIColor.white"),
             Example("var one: Int = 1, two: Int = 2, three: Int"),
             Example("guard let url = URL() else { return }"),
             Example("if let url = URL() { return }"),
@@ -45,7 +44,13 @@ struct RedundantTypeAnnotationRule: SwiftSyntaxCorrectableRule, OptInRule {
 
             var direction = Direction.up
             """),
-            Example("var values: Set<Int> = Set([0, 1, 2])")
+            Example("@IgnoreMe var a: Int = Int(5)", configuration: ["ignore_attributes": ["IgnoreMe"]]),
+            Example("""
+            var a: Int = {
+                @IgnoreMe let i: Int = Int(1)
+                return i
+            }
+            """, configuration: ["ignore_attributes": ["IgnoreMe"]])
         ],
         triggeringExamples: [
             Example("var url↓:URL=URL()"),
@@ -67,6 +72,7 @@ struct RedundantTypeAnnotationRule: SwiftSyntaxCorrectableRule, OptInRule {
             Example("guard var set↓: Set<Int> = Set<Int>([]) else { return }"),
             Example("if var set↓: Set<Int> = Set<Int>.init([]) { return }"),
             Example("var set↓: Set = Set<Int>([]), otherSet: Set<Int>"),
+            Example("var num↓: Int = Int.random(0..<10)"),
             Example("""
             class ViewController: UIViewController {
               func someMethod() {
@@ -83,7 +89,14 @@ struct RedundantTypeAnnotationRule: SwiftSyntaxCorrectableRule, OptInRule {
 
             var direction↓: Direction = Direction.up
             """),
-            Example("var num↓: Int = Int.random(0..<10")
+            Example("@DontIgnoreMe var a↓: Int = Int(5)", configuration: ["ignore_attributes": ["IgnoreMe"]]),
+            Example("""
+            @IgnoreMe
+            var a: Int = {
+                let i↓: Int = Int(1)
+                return i
+            }
+            """, configuration: ["ignore_attributes": ["IgnoreMe"]])
         ],
         corrections: [
             Example("var url↓: URL = URL()"): Example("var url = URL()"),
@@ -124,7 +137,21 @@ struct RedundantTypeAnnotationRule: SwiftSyntaxCorrectableRule, OptInRule {
               }
             }
             """),
-            Example("var num: Int = Int.random(0..<10)"): Example("var num = Int.random(0..<10)")
+            Example("var num: Int = Int.random(0..<10)"): Example("var num = Int.random(0..<10)"),
+            Example("""
+            @IgnoreMe
+            var a: Int = {
+                let i↓: Int = Int(1)
+                return i
+            }
+            """, configuration: ["ignore_attributes": ["IgnoreMe"]]):
+            Example("""
+            @IgnoreMe
+            var a: Int = {
+                let i = Int(1)
+                return i
+            }
+            """)
         ]
     )
 
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
@@ -3,7 +3,7 @@ import SwiftLintCore
 import SwiftSyntax
 
 @SwiftSyntaxRule
-struct RedundantTypeAnnotationRule: OptInRule {
+struct RedundantTypeAnnotationRule: SwiftSyntaxCorrectableRule, OptInRule {
     var configuration = RedundantTypeAnnotationConfiguration()
 
     static let description = RuleDescription(
@@ -15,6 +15,9 @@ struct RedundantTypeAnnotationRule: OptInRule {
             Example("var url = URL()"),
             Example("var url: CustomStringConvertible = URL()"),
             Example("@IBInspectable var color: UIColor = UIColor.white"),
+            Example("var one: Int = 1, two: Int = 2, three: Int"),
+            Example("guard let url = URL() else { return }"),
+            Example("if let url = URL() { return }"),
             Example("""
             enum Direction {
                 case up
@@ -40,6 +43,9 @@ struct RedundantTypeAnnotationRule: OptInRule {
             Example("let url↓: URL = URL()"),
             Example("lazy var url↓: URL = URL()"),
             Example("let url↓: URL = URL()!"),
+            Example("var one: Int = 1, two↓: Int = Int(5), three: Int"),
+            Example("guard let url↓: URL = URL() else { return }"),
+            Example("if let url↓: URL = URL() { return }"),
             Example("let alphanumerics↓: CharacterSet = CharacterSet.alphanumerics"),
             Example("""
             class ViewController: UIViewController {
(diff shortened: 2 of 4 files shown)


tonell-m:
Just finished resolving test and linking issues, sorry for the noise

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -1,9 +1,120 @@
 import Foundation
-import SourceKittenFramework
+import SwiftLintCore
+import SwiftSyntax
 
-struct RedundantTypeAnnotationRule: OptInRule, SubstitutionCorrectableRule {
-    var configuration = SeverityConfiguration<Self>(.warning)
+struct RedundantTypeAnnotationRule: SwiftSyntaxCorrectableRule, OptInRule {
+    var configuration = RedundantTypeAnnotationConfiguration()
```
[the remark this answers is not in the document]
tonell-m: This var can't be private because it's part of the [`Rule`](https://github.com/realm/SwiftLint/blob/6366fb85669c7a3e487fffea80285bafdccb7a8c/Source/SwiftLintCore/Protocols/Rule.swift#L20) protocol

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@
   Extensions are an exception; more than one is allowed.  
   [Muhammad Zeeshan](https://github.com/mzeeshanid)
   [#2802](https://github.com/realm/SwiftLint/issues/2802)
+* Rewrite `redundant_type_annotation` rule with SwiftSyntax and add a new `ignored_attributes` option that allows disabling the rule for properties that are marked with at least one of the configured attributes
```
SimplyDanny, line 18: You may the rule's ID and your name to the list in line 19/22 and only mention the new attribute as an enhancement.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -1,9 +1,143 @@
 import Foundation
-import SourceKittenFramework
+import SwiftLintCore
+import SwiftSyntax
 
-struct RedundantTypeAnnotationRule: OptInRule, SubstitutionCorrectableRule {
-    var configuration = SeverityConfiguration<Self>(.warning)
+struct RedundantTypeAnnotationRule: SwiftSyntaxCorrectableRule, OptInRule {
```
SimplyDanny, line 5: Please use `@SwiftSyntaxRule` and remove the explicit `makeVisitor` implementation.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -1,9 +1,143 @@
...
+            // If the type is implicit, `base` will be nil, meaning there is no redundancy.
+            return false
+        }
+
+        return base.baseName.text == typeName
+    }
+}
+
+extension RedundantTypeAnnotationRule {
     static let description = RuleDescription(
```
SimplyDanny, line 141: Why is that moved down here? Let's have it right at the beginning after the `configuration` again.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -1,9 +1,143 @@
...
+        let doesNotContainIgnoredAttributes = configuration.ignoredAnnotations.allSatisfy {
+            !self.attributes.contains(attributeNamed: $0)
+        }
+
+        // Only take the last binding into account in case multiple
+        // variables are declared on a single line.
+        // This binding must have both a type declaration and an initializer
+        // sequence for it to be potentially redundant.
+        guard doesNotContainIgnoredAttributes,
+              let binding = bindings.last,
```
SimplyDanny, line 87: Might be fine for now, but actually restricting the rule to the last binding only is not necessary. It could as well check all other bindings.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -1,9 +1,143 @@
...
+                return calledExpression.baseName.text == typeName + genericArguments
+            }
+
+            // If the function call is a member access expression, check if it is a violation
+            return isMemberAccessViolation(node: functionCall.calledExpression, typeName: typeName)
+        }
+
+        // If the initializer is a boolean expression, we consider using the `Bool` type
+        // annotation as redundant.
+        if initializer.as(BooleanLiteralExprSyntax.self) != nil {
```
SimplyDanny, line 118:
```suggestion
        if initializer.is(BooleanLiteralExprSyntax.self) {
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -1,9 +1,143 @@
...
+        }
+
+        // If the initializer is a function call (generally a constructor or static builder),
+        // check if the base type is the same as the one from the type annotation.
+        if let functionCall = initializer.as(FunctionCallExprSyntax.self) {
+            if let calledExpression = functionCall.calledExpression.as(DeclReferenceExprSyntax.self) {
+                // Parse generic arguments if there are any.
+                var genericArguments = ""
+                if let genericArgumentsClauseBytes = type.genericArguments?.trimmed.syntaxTextBytes {
+                    genericArguments = String(bytes: genericArgumentsClauseBytes, encoding: .utf8) ?? ""
```
SimplyDanny, line 107: The aspect of generic arguments is actually more involved, e.g. `let s: Set<Int> = Set.init([])` should not trigger while `let s: Set = Set<Int>([])` should. An there are more special cases here. We may keep it as simple as it is here, but need to make sure that all OSS findings remain stable with the rewrite.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -1,9 +1,143 @@
...
+            initializer = forceUnwrap.expression
+        }
+
+        // If the initializer is a function call (generally a constructor or static builder),
+        // check if the base type is the same as the one from the type annotation.
+        if let functionCall = initializer.as(FunctionCallExprSyntax.self) {
+            if let calledExpression = functionCall.calledExpression.as(DeclReferenceExprSyntax.self) {
+                // Parse generic arguments if there are any.
+                var genericArguments = ""
+                if let genericArgumentsClauseBytes = type.genericArguments?.trimmed.syntaxTextBytes {
```
SimplyDanny, line 106: Something like `type.genericArguments?.trimmedDescription` might be helpful to avoid the explicit parsing.

SimplyDanny (changes requested):
Thanks for taking up this rewrite! This is actually a bit more involved than adding the option for ignored attributes only. So I have a few comments ...

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/RedundantTypeAnnotationConfiguration.swift
```diff
@@ -0,0 +1,11 @@
+import SwiftLintCore
+
+@AutoApply
+struct RedundantTypeAnnotationConfiguration: SeverityBasedRuleConfiguration {
+    typealias Parent = RedundantTypeAnnotationRule
+
+    @ConfigurationElement(key: "severity")
+    var severityConfiguration = SeverityConfiguration<Parent>(.warning)
+    @ConfigurationElement(key: "ignored_annotations")
+    var ignoredAnnotations = Set<String>(["IBInspectable"])
```
garricn, line 10: The word `ignore` might be more appropriate than `ignored` because the configuration element is "configuring the rule to _ignore_ certain annotations". Also, the code base [appears](https://github.com/search?q=repo%3Arealm%2FSwiftLint%20ignore&type=code) to use `ignore`/`ignores` more than `ignored`.

Side note, the PR title and description mention ignore "attributes", not "annotations". I'm just pointing out for consistency purposes.

Great work!

```suggestion
    @ConfigurationElement(key: "ignore_annotations")
    var ignoreAnnotations = Set<String>(["IBInspectable"])
```

garricn:
I'm really excited about this work! After this merges, perhaps we can also introduce another configuration element named `ignore_literals` which can address the issues raised by this [PR](https://github.com/realm/SwiftLint/pull/4756) as well as the [concerns](https://github.com/realm/SwiftLint/pull/4756#issuecomment-1452585211) of @SimplyDanny therein in terms of not limiting the configuration element to just ignoring booleans.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/RedundantTypeAnnotationConfiguration.swift
tonell-m: You're right it should be attributes and not annotations, it took me a while to figure out which was the actually correct wording for Swift 😅 I'll fix that up!

As for the use of past tense, for me `ignored_attributes` suggests that the given attributes will be ignored by the linter while `ignore_attributes` suggests that this property is a flag that enables ignoring all attributes. Though I agree that for the sake of concistency it is better here to use `ignore` 👍

tonell-m:
@SimplyDanny Thanks for the review!

After your comments I've rewrote the rule with a new approach that seems to be more flexible. Instead of visiting `VariableDeclSyntax` nodes I use `PatternBindingSyntax` and `OptionalBindingConditionSyntax` as entry points and moved the logic into a `TypeAnnotationSyntax` extension. That way it will check for all binding patterns even if there are multiple for a single var/let keyword, and it also works the same of guard/if clauses. 
I've also improved the handling of generics to handle more complex cases as you suggested, so `var s: Set = Set<Int>([])` will trigger and not `var s: Set<Int> = Set([])`. I've added a bunch of tests with various configurations to ensure that this is working as expected, let me know if you think of other cases that haven't been handled here :)

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/RedundantTypeAnnotationConfiguration.swift
SimplyDanny: > As for the use of past tense, for me `ignored_attributes` suggests that the given attributes will be ignored by the linter while `ignore_attributes` suggests that this property is a flag that enables ignoring all attributes.

I'm with @tonell-m about this. `ignore_attributes` sounds like a boolean flag.

> Though I agree that for the sake of concistency it is better here to use `ignore` 👍

Can you provide an example where it is really used like that? I can only find flags following this name pattern. `ignored_attributes` seems like a better fit, to be honest.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -72,98 +124,183 @@ struct RedundantTypeAnnotationRule: OptInRule, SubstitutionCorrectableRule {
...
     }
+}
 
-    func substitution(for violationRange: NSRange, in file: SwiftLintFile) -> (NSRange, String)? {
-        return (violationRange, "")
-    }
+private extension RedundantTypeAnnotationRule {
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        override func visit(_ node: VariableDeclSyntax) -> SyntaxVisitorContinueKind {
+            guard node.doesNotContainIgnoredAttributes(for: configuration) else {
```
SimplyDanny, line 144: This is a good idea. However, it skips all other redundant type annotations nested in the variable declaration, e.g.

```swift
@IgnoreMe
var j: Int {
    let i: Int = Int(1) // will not trigger, but should
    return i
}
```

Therefore it's better to check the parent's parent of `PatternBindingSyntax` for the attribute.

SimplyDanny (changes requested):

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
tonell-m: Good call! I've made the update as you suggested

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -1,8 +1,10 @@
 import Foundation
```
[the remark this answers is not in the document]
tonell-m: Yes indeed

CHANGELOG.md
```diff
@@ -16,17 +16,25 @@
   [Muhammad Zeeshan](https://github.com/mzeeshanid)
   [#2802](https://github.com/realm/SwiftLint/issues/2802)
 
+* Add new `ignore_attributes` option to `redundant_type_annotation` rule
+  that allows disabling the rule for properties that are marked with at least
+  one of the configured attributes
```
SimplyDanny, line 21: The two spaces at the end are important:
```suggestion
* Add new `ignore_attributes` option to `redundant_type_annotation` rule
  that allows disabling the rule for properties that are marked with at least
  one of the configured attributes.  
```

CHANGELOG.md
```diff
@@ -16,17 +16,25 @@
...
+  [#5366](https://github.com/realm/SwiftLint/issues/5366)
+
 * Rewrite the following rules with SwiftSyntax:
   * `explicit_acl`
   * `identifier_name`
   * `let_var_whitespace`
   * `multiline_literal_brackets`
   * `opening_brace`
   * `void_return`
+  * `redundant_type_annotation`
```
SimplyDanny, line 32: Sorted. 😬
```suggestion
  * `opening_brace`
  * `redundant_type_annotation`
  * `void_return`
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -28,15 +43,36 @@ struct RedundantTypeAnnotationRule: OptInRule, SubstitutionCorrectableRule {
             }
 
             var direction = Direction.up
-            """)
+            """),
+            Example("@IgnoreMe var a: Int = Int(5)", configuration: ["ignore_attributes": ["IgnoreMe"]]),
+            Example("""
+            var a: Int = {
```
SimplyDanny, line 49:
```suggestion
            var a: Int {
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -52,13 +88,41 @@ struct RedundantTypeAnnotationRule: OptInRule, SubstitutionCorrectableRule {
             }
 
             var direction↓: Direction = Direction.up
-            """)
+            """),
+            Example("@DontIgnoreMe var a↓: Int = Int(5)", configuration: ["ignore_attributes": ["IgnoreMe"]]),
+            Example("""
+            @IgnoreMe
+            var a: Int = {
```
SimplyDanny, line 95:
```suggestion
            var a: Int {
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -72,98 +136,188 @@ struct RedundantTypeAnnotationRule: OptInRule, SubstitutionCorrectableRule {
                 let myVar = Int(5)
               }
             }
+            """),
+            Example("var num: Int = Int.random(0..<10)"): Example("var num = Int.random(0..<10)"),
+            Example("""
+            @IgnoreMe
+            var a: Int = {
```
SimplyDanny, line 143:
```suggestion
            var a: Int {
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -72,98 +136,188 @@ struct RedundantTypeAnnotationRule: OptInRule, SubstitutionCorrectableRule {
...
     }
 
-    func violationRanges(in file: SwiftLintFile) -> [NSRange] {
-        return file
-            .match(pattern: expressionPattern)
-            .filter {
-                $0.1 == [.keyword, .identifier, .typeidentifier, .identifier] ||
-                $0.1 == [.keyword, .identifier, .typeidentifier, .keyword]
+    final class Rewriter: ViolationsSyntaxRewriter {
+        private let configuration: RedundantTypeAnnotationConfiguration
```
SimplyDanny, line 194:
```suggestion
        private let configuration: ConfigurationType
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -72,98 +136,188 @@ struct RedundantTypeAnnotationRule: OptInRule, SubstitutionCorrectableRule {
...
+
+        init(configuration: RedundantTypeAnnotationConfiguration,
```
SimplyDanny, line 196:
```suggestion
        init(configuration: ConfigurationType,
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -72,98 +136,188 @@ struct RedundantTypeAnnotationRule: OptInRule, SubstitutionCorrectableRule {
...
     }
+}
 
-    private func getPartsOfExpression(
-        in file: SwiftLintFile, range: NSRange
-    ) -> (variableTypeName: String, assigneeName: String)? {
-        let substring = file.stringView.substring(with: range)
-        let components = substring.components(separatedBy: "=")
+private extension TypeAnnotationSyntax {
+    func isRedundant(for configuration: RedundantTypeAnnotationConfiguration,
```
SimplyDanny, line 250: `configuration` not needed in here, is it?

SimplyDanny (changes requested):

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
tonell-m: Indeed!

SimplyDanny:
Looking at the OSS findings, most of them look valid. 

However, [there are cases](https://github.com/wordpress-mobile/WordPress-iOS/blob/12065e659d39e59a7e9a01a45caf31112dee49b8/WordPress/Classes/ViewRelated/Gutenberg/Collapsable%20Header/Collabsable%20Header%20Filter%20Bar/CollabsableHeaderFilterCollectionViewCell.swift#L83:26) where a sequence of `MemberAccessExpression`s leads to the final element/function. An easy example would be `let a: A = A.b.c` They were previously discovered by the rule, so they should also be discoverable by the rewritten rule.

Admittedly, these cases are debatable, but for the sake of a rewrite, the rule should behave as close to the previous version as it can.

@tonell-m: Would you consider these cases in your PR?

CHANGELOG.md
```diff
@@ -16,17 +16,25 @@
   [Muhammad Zeeshan](https://github.com/mzeeshanid)
   [#2802](https://github.com/realm/SwiftLint/issues/2802)
 
+* Add new `ignore_attributes` option to `redundant_type_annotation` rule
+  that allows disabling the rule for properties that are marked with at least
+  one of the configured attributes.
```
SimplyDanny, line 21: Please append two spaces to the end of the description. They are required to enforce a line break in the rendered Markdown.

tonell-m:
> Looking at the OSS findings, most of them look valid.
>
> However, [there are cases](https://github.com/wordpress-mobile/WordPress-iOS/blob/12065e659d39e59a7e9a01a45caf31112dee49b8/WordPress/Classes/ViewRelated/Gutenberg/Collapsable%20Header/Collabsable%20Header%20Filter%20Bar/CollabsableHeaderFilterCollectionViewCell.swift#L83:26) where a sequence of MemberAccessExpressions leads to the final element/function. An easy example would be let a: A = A.b.c They were previously discovered by the rule, so they should also be discoverable by the rewritten rule.
>
> Admittedly, these cases are debatable, but for the sake of a rewrite, the rule should behave as close to the previous version as it can.
>
> @tonell-m: Would you consider these cases in your PR?

@SimplyDanny This case is now handled as well. Do you think we also need to handle cases where there is a mix of `MemberAccessExprSyntax` and `FunctionCallExprSyntax`? Like `let a: A = A.f().b`

SimplyDanny:
> > Looking at the OSS findings, most of them look valid.
> > However, [there are cases](https://github.com/wordpress-mobile/WordPress-iOS/blob/12065e659d39e59a7e9a01a45caf31112dee49b8/WordPress/Classes/ViewRelated/Gutenberg/Collapsable%20Header/Collabsable%20Header%20Filter%20Bar/CollabsableHeaderFilterCollectionViewCell.swift#L83:26) where a sequence of MemberAccessExpressions leads to the final element/function. An easy example would be let a: A = A.b.c They were previously discovered by the rule, so they should also be discoverable by the rewritten rule.
> > Admittedly, these cases are debatable, but for the sake of a rewrite, the rule should behave as close to the previous version as it can.
> > @tonell-m: Would you consider these cases in your PR?
> 
> @SimplyDanny This case is now handled as well. Do you think we also need to handle cases where there is a mix of `MemberAccessExprSyntax` and `FunctionCallExprSyntax`? Like `let a: A = A.f().b`

That depends on what the previous implementation did. I guess, it triggered, hence the rewritten rule should trigger too.

SimplyDanny:
> That depends on what the previous implementation did. I guess, it triggered, hence the rewritten rule should trigger too.

Rebased and addressed this point.

Issue #5373, opened 2023-12-04
`switch_case_alignment` failing on alignment in one-line case expressions


lhunath:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

A new syntax is now available for `if` and `switch` allowing for one-line expression statements with these. As such, the following are now completely acceptable and normal:
```
return if a { a } else { b }
return switch a { case .x: a default: b }
```

SwiftLint should be updated to allow for these, eg:
```
warning: Switch and Case Statement Alignment Violation: Case statements should be indented within their closing brace (switch_case_alignment)
```

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint lint
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.53.0
* Installation method used (Homebrew, CocoaPods, building from source, etc)? Homebrew
* Paste your configuration file:

```yml
included:
  - Hubstaff
excluded:
  - Hubstaff/Modules/Services/Activities/Server/GraphQL # https://github.com/realm/SwiftLint/issues/2329
  - Hubstaff/Modules/Services/Server/GraphQL
  - Hubstaff/App/Tests/Mocks
line_length:
  warning: 140
  ignores_comments: true
  ignores_interpolated_strings: true
colon:
  flexible_right_spacing: true
type_name:
  allowed_symbols: "_"
  min_length: 1
identifier_name:
  allowed_symbols: "_"
  min_length: 1
switch_case_alignment:
  indented_cases: true
function_parameter_count: 8
file_length: 1500
cyclomatic_complexity:
  warning: 15
  ignores_case_statements: true
function_body_length: 80
trailing_comma:
  mandatory_comma: true
nesting:
  type_level: 4
statement_position:
  statement_mode: uncuddled_else
opening_brace:
  allowMultilineFunc: true
disabled_rules:
  - todo              # TODOs are permitted as annotated markers for future improvements.
  - large_tuple       # We allow large tuples as cheap and easy anonymous struct substitutes.
  - opening_brace     # https://github.com/realm/SwiftLint/issues/4026
  - type_body_length  # https://github.com/realm/SwiftLint/issues/4631
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? no
[...]

PR #5390, opened 2023-12-12
Add a new option to `switch_case_alignment` rule to ignore switch statements written in a single line

tonell-m:
As discussed in #5373 this PR adds a new `ignore_one_liners` option to the `switch_case_alignment` rule to allow this new one liner syntax for switch statements:

```swift
switch s { case .x: a default: b }
```

This option is still disabled by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@
 
 #### Enhancements
 
-* Add a new `ignore_one_liners` option to `switch_case_alignment` rule to ignore switch statements written in a single line.  
+* Add new `ignore_one_liners` option to `switch_case_alignment` 
+  rule to ignore switch statements written in a single line.  
   [tonell-m](https://github.com/tonell-m)
   [#5373](https://github.com/realm/SwiftLint/issues/5373)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 #### Enhancements
 
+* Add a new `ignore_one_liners` option to `switch_case_alignment` rule to ignore switch statements written in a single line.  
+  [tonell-m](https://github.com/tonell-m)
+  [#5373](https://github.com/realm/SwiftLint/issues/5373)
+
 * Add new `one_declaration_per_file` rule that allows only a
   single class/struct/enum/protocol declaration per file.
   Extensions are an exception; more than one is allowed.  
--- a/Source/SwiftLintBuiltInRules/Rules/Style/SwitchCaseAlignmentRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/SwitchCaseAlignmentRule.swift
@@ -43,14 +43,22 @@ struct SwitchCaseAlignmentRule: Rule {
 extension SwitchCaseAlignmentRule {
     final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
         override func visitPost(_ node: SwitchExprSyntax) {
-            let closingBracePosition = node.rightBrace.positionAfterSkippingLeadingTrivia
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
 
 #### Enhancements
 
+* Add a new `ignore_one_liners` option to `switch_case_alignment` rule to ignore switch statements written in a single line.
```
SimplyDanny, line 13:
```suggestion
* Add new `ignore_one_liners` option to `switch_case_alignment`
  rule to ignore switch statements written in a single line.  
```

SimplyDanny (changes requested):
Looks good to me!

Please also add some tests for the new option.

tonell-m:
@SimplyDanny about adding tests for the new option, I'm not sure how to make examples that are taking different configuration values, could you guide me a bit on how to do that? Like adding an example with a one liner that would trigger with the option set to false and not trigger when it's set to true

SimplyDanny:
> @SimplyDanny about adding tests for the new option, I'm not sure how to make examples that are taking different configuration values, could you guide me a bit on how to do that? Like adding an example with a one liner that would trigger with the option set to false and not trigger when it's set to true

Every `Example` accepts a `configuration` argument. It would be `["ignore_one_liners": true]` here. With that, you can add a normal example to the rule's description which is being tested as normal like all other examples.

Source/SwiftLintBuiltInRules/Rules/Style/SwitchCaseAlignmentRule.swift
```diff
@@ -216,5 +227,14 @@ extension SwitchCaseAlignmentRule {
                 """)
             ]
         }
+
+        private func getOneLinerExample(ignoreOneLiners: Bool) -> Example {
+            let marker = ignoreOneLiners ? "" : violationMarker
+
+            return Example(
```
SimplyDanny, line 234: I wish there were more examples, e.g. some were the opening and closing braces are on different lines.

SimplyDanny (changes requested):

SimplyDanny:
Please rebase onto the current `main` branch to get all builds running.

Source/SwiftLintBuiltInRules/Rules/Style/SwitchCaseAlignmentRule.swift
tonell-m: Good call, I've added some extra tests and tweaked the rule so that it checks for the line of the switch keyword instead of opening brace to determine if it's a one liner, so that this doesn't count as one

```swift
switch i
{ case .x: 1 default 0 }
```

Source/SwiftLintBuiltInRules/Rules/Style/SwitchCaseAlignmentRule.swift
```diff
@@ -216,5 +227,46 @@ extension SwitchCaseAlignmentRule {
...
+                """, configuration: ["ignore_one_liners": true]),
+                Example("""
+                let a = switch i {
+                case .x: 1 default: 0
+                }
+                """, configuration: ["ignore_one_liners": true]),
+                Example("""
+                let a = switch i {
+                case .x: 1 default: 0 }
+                """, configuration: ["ignore_one_liners": true])
```
SimplyDanny, line 268: Good examples! They should have violation markers, though.

SimplyDanny (approved):

Source/SwiftLintBuiltInRules/Rules/Style/SwitchCaseAlignmentRule.swift
tonell-m: Fair point 😁

Issue #5359, opened 2023-11-18
Detect redundant extensions


inotetw:
## New rule request

extension without any function will be avoid

### Bad
```
extension Dog {}
```

SimplyDanny:
I would be interested in how regular you really stumble upon empty extensions. Have you seen this often, @inotetw?

inotetw:
I often find the code  such as following, may be my team coder use xcode template to generate code

// MARK: UITableViewDelegate
extension TableVM: UITableViewDelegate {}

KaitoMuraoka:
I will take on this Issue.

mzeeshanid:
Its a first attempt based on variable and function declarations. There are more cases that needs to be address as well. A part from variable and function declarations, extensions can have other declarations as well. Any suggestions?

tonell-m:
@mzeeshanid you could also check for type declarations (structs, enums, classes), and perhaps associated types and type aliases?

Also idk if subscripts count as function declaration in SwiftSyntax but if not you could check for those as well 🙂

mzeeshanid:
I found a 1 liner solution i.e. if memberBlock is empty then the extension is redundant. See the last commit :)

tonell-m:
Oh neat. Good one! 👏

PR #5397, opened 2023-12-18
- rule added i.e. redundant_extension #5359

mzeeshanid:
Rule added for avoiding redundant extensions resolve

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -42,6 +42,7 @@ disabled_rules:
   - prefer_nimble
   - prefer_self_in_static_references
   - prefixed_toplevel_constant
+  - redundant_extension
   - redundant_self_in_closure
   - required_deinit
   - self_binding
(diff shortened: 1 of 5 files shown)


SimplyDanny:
To be honest, I don't deem this rule to be as useful as I intentionally thought. The OSS report contains a lot of false positives. It's pretty common to conform a foreign type to a local protocol. And quite often the type already comes with the requirements defined in the protocol causing an empty body.

So blindly triggering on every extension with an empty body isn't the way to go in my opinion. What might be acceptable is that the rule only triggers on empty extension if the extended type is defined in the same file. In that case, the conformance can just be moved to the type's definition. This could be a convention that projects follow, and the rule could support them with it.

That said, I think this rule requires some more thought and, admittedly, the implementation for what I suggest above is a bit more involved.

jaredgrubb:
It seems that the rule is really _trying_ to flag:
```
extension Foo {}
```
but is _also_ trapping good and correct code like:
```
extension Foo: Sendable {}
```

I think the rule is just being too loose for what the Issue and Examples suggest. The OSS failures are calling out this deficiency.

PR #5398, opened 2023-12-18
Add Accessibility Font Size Rule

MartijnAmbagtsheer:
The text below has been taken from the `rule request` template.

**1. Why should this rule be added? Share links to existing discussion about what the community thinks about this.**  
The increasing focus on accessibility in both [America through the ADA (Americans with Disabilities Act)](https://www.ada.gov/resources/web-guidance/) and the upcoming [European legislation](https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=uriserv%3AOJ.L_.2019.151.01.0070.01.NLD&toc=OJ%3AL%3A2019%3A151%3ATOC) underscores the growing importance of accessible applications. To assist in this effort, we can identify existing inaccessible elements using tools such as SwiftLint, among others.

There are already two rules for testing the accessibility of images and buttons. With this new rule, I aim to align with the [WCAG (Web Content Accessibility Guidelines) 1.4.4](https://www.w3.org/WAI/WCAG21/Understanding/resize-text.html). This guideline stipulates that every text must be scalable.

**2. Provide several examples of what _would_ and _wouldn't_  trigger violations.**  
Insofar as I have been able to find, SwiftUI inherently scales well on its own, but this can be forcibly disabled using the `.font(.system())` modifier.

```Swift
struct TestView: View {
	var body: some View {
		Text("Hello World!")
			.font(.system(size: 20))
	}
}
```

Using the default Text, without the .font(.system(size:)) modifier would work and not trigger the linter.

```Swift
struct TestView: View {
	var body: some View {
		Text("Hello World!")
	}
}
```

This also applies to other view elements like a Button, TextField or SecureField.

**3. Should the rule be configurable, if so what parameters should be configurable?**   
For this rule, no additional configuration is required.  

**4. Should the rule be opt-in or enabled by default? Why?
    See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a rule as opt-in.**  
As stated in the rules in the README, this rule complies with the guidelines to be opt-in, as the rule, like the other accessibility rules, is opt-in due to potential false positives given the many possibilities to set up a view within SwiftUI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,6 +109,10 @@
 
 #### Enhancements
 
+* Add `Accessibility Font Size` rule to warn if a SwiftUI Text
+  has a fixed font size.  
+  [MartijnAmbagtsheer](https://github.com/MartijnAmbagtsheer)
+
 * Add `only` configuration option to `todo` rule which allows to specify
   whether the rule shall trigger on `TODO`s, `FIXME`s or both.  
   [gibachan](https://github.com/gibachan)
--- a/Source/SwiftLintBuiltInRules/Models/BuiltInRules.swift
+++ b/Source/SwiftLintBuiltInRules/Models/BuiltInRules.swift
@@ -3,6 +3,7 @@
 
 /// The rule list containing all available rules built into SwiftLint.
 public let builtInRules: [any Rule.Type] = [
+    AccessibilityFontSizeRule.self,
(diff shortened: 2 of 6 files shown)


MartijnAmbagtsheer:
The ownership of the code has been modified. A new PR can be found here: https://github.com/realm/SwiftLint/pull/5436

PR #5400, opened 2023-12-21
Add `discouraged_void_return` rule

SimplyDanny:


marcelofabri:
Isn’t this the same as https://github.com/realm/SwiftLint/blob/77a9d3ccc6bced1299e608622fe79b4d28f5e549/Source/SwiftLintBuiltInRules/Rules/Idiomatic/ReturnValueFromVoidFunctionRule.swift#L4?

SimplyDanny:
> Isn’t this the same as
> 
> https://github.com/realm/SwiftLint/blob/77a9d3ccc6bced1299e608622fe79b4d28f5e549/Source/SwiftLintBuiltInRules/Rules/Idiomatic/ReturnValueFromVoidFunctionRule.swift#L4
> 
> ?

It is indeed. Thank you for the pointer. I was looking for similar rules but couldn't find this one for unclear reasons. Both rules will be merged in #5403 with a favor on `return_value_from_void_function`.

PR #5402, opened 2023-12-23
Document modified configurations used for examples

SimplyDanny:
Example:

```swift
for user in users {
  if user.id == 1 { return true }
}

//
// allow_for_as_filter: true
//
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,9 @@
   [Ueeek](https://github.com/Ueeek)
   [#5615](https://github.com/realm/SwiftLint/issues/5615)
 
+* Add modified configurations to examples in rule documentation.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+
 * Add new option `evaluate_effective_access_control_level` to `missing_docs`
   rule. Setting it to `true` stops the rule from triggering on declarations
   inside of types with lower visibility. These declarations effectively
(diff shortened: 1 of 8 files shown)


SimplyDanny:
@mildm8nnered: I would like to hear your opinion on this. Is the presentation style clear enough?

mildm8nnered:
> @mildm8nnered: I would like to hear your opinion on this. Is the presentation style clear enough?

So I think it's a matter of taste, but I'd probably put the context before the example, so something like this:

```
// allow_for_as_filter: true
for user in users {
  if user.id == 1 { return true }
}
```

although that could collide with swift documentation comments in some cases ...

I guess there's also a bit of a risk that people might think these are directives like `// swiftlint:disable` somehow, and end up pasting them into their own code.

SimplyDanny:
> > @mildm8nnered: I would like to hear your opinion on this. Is the presentation style clear enough?
> 
> So I think it's a matter of taste, but I'd probably put the context before the example, so something like this:
> 
> ```
> // allow_for_as_filter: true
> for user in users {
>   if user.id == 1 { return true }
> }
> ```
> 
> although that could collide with swift documentation comments in some cases ...
> 
> I guess there's also a bit of a risk that people might think these are directives like `// swiftlint:disable` somehow, and end up pasting them into their own code.

All the variants I can think of are not optimal. I went with your suggestion to have the configuration first. There is an empty line between them and the code. That should be a clear enough separation. Example:

```swift
//
// allow_for_as_filter: true
//

for user in users {
  if user.id == 1 { return true }
}
```

The documentation is (hopefully) not something people rely upon as an API, so we can update it once we come up with a better idea later.

Issue #5372, opened 2023-12-04
swiftlint and SwiftLintFramework being build when only using SwiftLintPlugin on MacOS


geertbleyen:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug
I've noticed that while we are only using and referring SwiftLintPlugin form the SwiftLint SPM products, Xcode is building the SwiftLint and SwiftLintFramework targets regardless. These targets take quite some time to build, and don't add anything to the result as they are not used anywhere, and the plugin uses the binaryTarget dependency on MacOS.
Tbh, I suspect this to be an Xcode/Apple related issue, but wanted to report it nonetheless to verify the finding.
Are we doing something wrong, or this is just how it currently works in Xcode 

### Environment

* SwiftLint version: 0.54.0
* Xcode version: 15.0.1
* Installation method: SPM integration
<img width="860" alt="image" src="https://github.com/realm/SwiftLint/assets/71486361/1c4a6d23-33be-49d8-b6d0-7803414b53ca">
<img width="432" alt="image" src="https://github.com/realm/SwiftLint/assets/71486361/caa8b73f-f388-4925-8266-a75f865728f8">
<img width="812" alt="image" src="https://github.com/realm/SwiftLint/assets/71486361/70b0ad45-fa3e-4a89-af43-3462a0af025c">
Don't mind the 0.51.0 version specced here, as our package.resolved has 0.54.0 locked in.
<img width="601" alt="image" src="https://github.com/realm/SwiftLint/assets/71486361/8bd4cadc-6b67-452f-ba2d-6e6af97261b5">

* Paste your configuration file: NA

julien-baillon:
Hi @SimplyDanny, I'm facing the same issue. I have created this [PR](https://github.com/realm/SwiftLint/pull/5404) to fix this issue. Could you test it by using `.package(url: "https://github.com/julien-baillon/SwiftLint", branch: "jb-plugin-dependencies"),` in you `Package.swift`.

Issue #5099, opened 2023-07-05
SwiftLint Compile Times Extremely Small


DomenicBianchi01:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

- When compiling the app with the SwiftLint SPM plugin installed, the compile times are **extremely** slow.
- It looks like the culprit is `SwiftSyntax`

![Screenshot 2023-07-05 at 10 04 15 AM](https://github.com/realm/SwiftLint/assets/25673485/27dcaf43-07ec-4733-83a9-b9aa6fe7b7d8)

### Environment

* SwiftLint version: 0.52.3
* Xcode version: 14.3.1
* Installation method: SPM

DomenicBianchi01:
Related issue: https://github.com/realm/SwiftLint/issues/4261

marcelofabri:
Not really related, that issue is about the runtime performance of SwiftLint. Optimizing for compilation times hasn't been a goal of SwiftLint. IMO the real issue here is how SPM works if this is happening in every build.

I personally always recommend people using a pre-built binary if they don't want to pay the compilation cost.

jpsim:
You're correct to note that SwiftSyntax is responsible for the bulk of the compilation time. That's also tracked here: https://github.com/apple/swift-syntax/issues/1194

This is why most users should be using pre-built SwiftLint binaries, which the SwiftPM plugin should be using by default. If that's not happening for you for some reason, please share more information.

anton-plebanovich:
Not sure if related or if should I create a separate issue. I'm facing an issue of huge lint times when I try to pass concrete files to lint. In my project, it takes ~1 second to perform full lint of ~250 files and ~12 seconds when I pass ~60 files. Version 0.52.4. 

It takes ~1 second on Version 0.51.0 for example for the same 60 files.

My teammate on a different project reported ~10 mins lint time on bigger file counts and on slower machines.

jpsim:
How does that compare with running `swiftlint` as installed from Homebrew from your command line?

anton-plebanovich:
> How does that compare with running swiftlint as installed from Homebrew from your command line?
If that's a question for me, I use Cocoapods and binary there so not sure if it makes sense to check binary from Homebrew.

Mordil:
[...]

PR #5404, opened 2023-12-26
Prevent from building SwiftLint when using SwiftLint Plugin on macOS

julien-baillon:
When we use `SwiftLintPlugin` with SwiftPM on macOS, SwiftLint and all its dependencies are built even if it's useless on macOS because there is `SwiftLintBinary` instead. 

Here is an example project that only use `SwiftLintPlugin`, but takes about one minute to compile because it compiles SwiftLint for nothing.

```swift
// swift-tools-version: 5.9

import PackageDescription

let package = Package(
    name: "MyLibrary",
    platforms: [.macOS(.v14)],
    products: [
        .library(
            name: "MyLibrary",
            targets: ["MyLibrary"]),
    ],
    dependencies: [
        .package(url: "https://github.com/realm/SwiftLint", exact: Version("0.54.0")),
    ],
    targets: [
        .target(
            name: "MyLibrary",
            plugins: [.plugin(name: "SwiftLintPlugin", package: "SwiftLint")]
        )
    ]
)
```

[MyLibrary.zip](https://github.com/realm/SwiftLint/files/13772645/MyLibrary.zip)

I believe it's due to `.target(name: "swiftlint", condition: .when(platforms: [.linux]))` on SwiftLint's `Package.swift`. The condition seams not to work as expected for plugins. 

Related issues: 
- [swiftlint and SwiftLintFramework being build when only using SwiftLintPlugin on MacOS](https://github.com/realm/SwiftLint/issues/5372)
- [SwiftLint Compile Times Extremely Slow](https://github.com/realm/SwiftLint/issues/5099)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,13 +10,14 @@
 
 #### Enhancements
 
+* Prevent from compiling `SwiftLint` target when only using `SwiftLintPlugin` on macOS.  
+  [Julien Baillon](https://github.com/julien-baillon)
+  [#5372](https://github.com/realm/SwiftLint/issues/5372)
+
 * Add new `ignore_one_liners` option to `switch_case_alignment` 
   rule to ignore switch statements written in a single line.  
   [tonell-m](https://github.com/tonell-m)
   [#5373](https://github.com/realm/SwiftLint/issues/5373)
-* Prevent from compiling `SwiftLint` target when only using `SwiftLintPlugin` on macOS.
-  [Julien Baillon](https://github.com/julien-baillon)
-  [#5372](https://github.com/realm/SwiftLint/issues/5372)
 
 * Add new `one_declaration_per_file` rule that allows only a
   single class/struct/enum/protocol declaration per file.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
 
 #### Enhancements
 
+* Prevent from compiling `SwiftLint` target when only using `SwiftLintPlugin` on macOS.
```
SimplyDanny, line 13: Please append two spaces to the end of the description. They are required to enforce a line break in the rendered Markdown.

SimplyDanny (approved):
Looks good to me, even though it's a pity the conditional dependencies don't seem to work in this case. They would be the more elegant solution.

I checked that your branch indeed increases build times. However, all dependencies are still fetched and checked out. This seems to be the least that must be done.

CHANGELOG.md
julien-baillon-ma: done

Thank you for the tip, now I know.

julien-baillon:
Thank you @SimplyDanny for this review and the work you do for SwiftLint, It's a really important project for me and my team 🙏🏻.

SimplyDanny (approved):

jpsim:
It would be great if you could report these issues to the SwiftPM project @julien-baillon

julien-baillon:
Sorry @jpsim, I spent a bunch of hours trying to reproduce the issue with a test package, in order to report the issue to SwiftPM, but I failed. The `.when(platforms: [.linux])` does the job on my package, I don't know why SwiftLint Package ignores it.

PR #5407, opened 2023-12-29
Ignore absence of a non-initial local config instead of falling back to default.

kohtenko:
This change is helpful for Swift Package Manager (SPM) users. When one sets up the parent config to be shared between packages and keeps it in a working directory like following:
```
- WorkingDirectory
   - parent_config.yml
   - Package1
      - .swiftlint.yaml
      - ... 
   - Package2
      - .swiftlint.yaml
      - ... 
   - .... 
```

The current configuration works as expected. The challenge comes when the package is not local anymore (not in active development anymore). SPM checks out the repository in project's DerivedData instead, and no parent config is available at that new folder structure. Swiftlint plugin runs normally by SPM, but the config is now corrupted and fallback to default, which might bring unexpected warnings and errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 * None.
 
 #### Enhancements
+* Ignore absence of a non-initial local config instead of
+  falling back to default.  
+  [kohtenko](https://github.com/kohtenko)
+  [#5407](https://github.com/realm/SwiftLint/pull/5407)
 
 * Add new option `ignore_typealiases_and_associatedtypes` to
   `nesting` rule. It excludes `typealias` and `associatedtype`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,10 @@
 * None.
 
 #### Enhancements
+
 * Ignore absence of a non-initial local config instead of
   falling back to default.  
   [kohtenko](https://github.com/kohtenko)
-  [#5407](https://github.com/realm/SwiftLint/pull/5407)
 
 * Add new option `ignore_typealiases_and_associatedtypes` to
   `nesting` rule. It excludes `typealias` and `associatedtype`
--- a/Source/SwiftLintCore/Extensions/Configuration+FileGraph.swift
+++ b/Source/SwiftLintCore/Extensions/Configuration+FileGraph.swift
@@ -88,17 +88,43 @@ package extension Configuration {
             )
 
             if !ignoreParentAndChildConfigs {
-                try processPossibleReference(
+                try processPossibleReferenceIgnoringFileAbsence(
                     ofType: .childConfig,
+                    from: vertex,
+                    remoteConfigTimeoutOverride: remoteConfigTimeoutOverride,
+                    remoteConfigTimeoutIfCachedOverride: remoteConfigTimeoutIfCachedOverride)
+
+                try processPossibleReferenceIgnoringFileAbsence(
+                    ofType: .parentConfig,
+                    from: vertex,
+                    remoteConfigTimeoutOverride: remoteConfigTimeoutOverride,
+                    remoteConfigTimeoutIfCachedOverride: remoteConfigTimeoutIfCachedOverride)
+            }
+        }
+
+        private mutating func processPossibleReferenceIgnoringFileAbsence(
+            ofType type: EdgeType,
+            from vertex: Vertex,
+            remoteConfigTimeoutOverride: TimeInterval?,
+            remoteConfigTimeoutIfCachedOverride: TimeInterval?
+        ) throws {
+            do {
+                try processPossibleReference(
+                    ofType: type,
                     from: vertex,
                     remoteConfigTimeoutOverride: remoteConfigTimeoutOverride,
                     remoteConfigTimeoutIfCachedOverride: remoteConfigTimeoutIfCachedOverride
                 )
-                try processPossibleReference(
-                    ofType: .parentConfig,
-                    from: vertex,
-                    remoteConfigTimeoutOverride: remoteConfigTimeoutOverride,
-                    remoteConfigTimeoutIfCachedOverride: remoteConfigTimeoutIfCachedOverride
+            } catch {
+                // If a child or parent config file doesn't exist, do not fail the rest of the config tree.
+                // Instead, just ignore this leaf of the config. Otherwise, rethrow the error.
+                guard case let Issue.fileNotFound(path) = error else {
+                    throw error
+                }
+                queuedPrintError("""
+                    A local configuration at \(path) was not found. \
+                    Ignoring this part of the configuration.
+                    """
                 )
             }
         }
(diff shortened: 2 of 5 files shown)


Source/SwiftLintCore/Extensions/Configuration+FileGraphSubtypes.swift
```diff
@@ -74,8 +74,8 @@ internal extension Configuration.FileGraph {
         private func read(at path: String) throws -> String {
             guard !path.isEmpty && FileManager.default.fileExists(atPath: path) else {
                 throw isInitialVertex ?
-                    Issue.initialFileNotFound(path: path) :
-                    Issue.genericWarning("File \(path) can't be found.")
+                Issue.initialFileNotFound(path: path) :
+                Issue.fileNotFound(path: path)
```
SimplyDanny, line 78: If you'd like to change the format, please have it like:
```swift
throw isInitialVertex
    ? Issue.initialFileNotFound(path: path)
    : Issue.fileNotFound(path: path)
```

Source/SwiftLintCore/Models/Issue.swift
```diff
@@ -125,6 +128,8 @@ public enum Issue: LocalizedError, Equatable {
             return "Cannot get cursor info from file at path '\(path ?? "...")' within '\(id)' rule."
         case let .yamlParsing(message):
             return "Cannot parse YAML file: \(message)"
+        case let .fileNotFound(path):
+            return "File not found: \(path)"
```
SimplyDanny, line 132:
```suggestion
            return "File at path '\(path)' not found."
```

Tests/SwiftLintFrameworkTests/Resources/ProjectMock/ChildConfig/Test1/Main/main.yml
```diff
@@ -9,3 +9,4 @@ excluded:
 line_length: 80
 
 child_config: child1.yml
+parent_config: nonExistingParent.yml
```
SimplyDanny, line 12: Would be nice if something was asserted in a test as well.

Source/SwiftLintCore/Extensions/Configuration+FileGraph.swift
```diff
@@ -89,18 +89,29 @@ public extension Configuration {
...
-                    remoteConfigTimeoutIfCachedOverride: remoteConfigTimeoutIfCachedOverride
-                )
+                do {
+                    try processPossibleReference(
+                        ofType: .childConfig,
+                        from: vertex,
+                        remoteConfigTimeoutOverride: remoteConfigTimeoutOverride,
+                        remoteConfigTimeoutIfCachedOverride: remoteConfigTimeoutIfCachedOverride
+                    )
+                    try processPossibleReference(
```
SimplyDanny, line 99: I think you should have both in a separate try-catch. Otherwise, we don't check the parent configuration when there was a problem with reading a child configuration.

SimplyDanny (changes requested):
The change makes sense to me.

@mildm8nnered: I would like to check with you as the configuration expert. Please have a look.

mildm8nnered:
> The change makes sense to me.
> 
> @mildm8nnered: I would like to check with you as the configuration expert. Please have a look.

I think this seems reasonable.

Tests/SwiftLintFrameworkTests/Resources/ProjectMock/ChildConfig/Test1/Main/main.yml
kohtenko: `testValidChildConfig` validates it already. It fails to load the config with a non existent parent and fails. After the change is applied it succeeds as expected.

Let me know if you want some specific test scenario covered, though I can't think of one other than successfully loading the config 🤷‍♂️

Source/SwiftLintCore/Extensions/Configuration+FileGraph.swift
kohtenko: Good catch. Split into two try/catch

SimplyDanny: Please extract this to a function.

SimplyDanny (changes requested):

SimplyDanny:
Danger doesn't like merge commits. Please rebase your branch onto `main` instead of merging `main` in.

Source/SwiftLintCore/Extensions/Configuration+FileGraph.swift
```diff
@@ -88,18 +88,42 @@ package extension Configuration {
...
+                try processPossibleReference(
+                    ofType: type,
+                    from: vertex,
+                    remoteConfigTimeoutOverride: remoteConfigTimeoutOverride,
                     remoteConfigTimeoutIfCachedOverride: remoteConfigTimeoutIfCachedOverride
                 )
+            } catch {
+                // If a child or parent config file doesn't exist, do not fail the rest of the config tree
+                // Instead just ignore this leaf of the config
+                // Otherwise, rethrow the error.
```
SimplyDanny, line 121:
```suggestion
                // If a child or parent config file doesn't exist, do not fail the rest of the config tree.
                // Instead, just ignore this leaf of the config. Otherwise, rethrow the error.
```

Source/SwiftLintCore/Extensions/Configuration+FileGraph.swift
```diff
@@ -88,18 +88,42 @@ package extension Configuration {
...
+                guard case let Issue.fileNotFound(path) = error else {
+                    throw error
+                }
+                queuedPrintError("A local config at \(path) was not found. " +
+                                 "Ignoring the part of the config instead of failing back to default")
```
SimplyDanny, line 126:
```suggestion
                queuedPrintError("""
                    A local configuration at \(path) was not found. \
                    Ignoring this part of the configuration.
                    """
                )
```

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
 * None.
 
 #### Enhancements
+* Ignore absence of a non-initial local config instead of falling back to default..  
+  [kohtenko](https://github.com/kohtenko)
+  [#issue_number](https://github.com/realm/SwiftLint/issues/issue_number)
```
SimplyDanny, line 14:
```suggestion
```

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
 * None.
 
 #### Enhancements
+* Ignore absence of a non-initial local config instead of falling back to default..
```
SimplyDanny, line 12:
```suggestion
* Ignore absence of a non-initial local config instead of
  falling back to default.  
```

SimplyDanny (changes requested):

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
 * None.
 
 #### Enhancements
+* Ignore absence of a non-initial local config instead of
+  falling back to default.  
+  [kohtenko](https://github.com/kohtenko)
+  [#5407](https://github.com/realm/SwiftLint/pull/5407)
```
SimplyDanny, line 15:
```suggestion
#### Enhancements

* Ignore absence of a non-initial local config instead of
  falling back to default.  
  [kohtenko](https://github.com/kohtenko)
```

SimplyDanny (approved):

PR #5408, opened 2023-12-30
Rewrite `mark` rule using SwiftSyntax

swiftty:
- rewrite `mark` rule using SwiftSyntax.
- add `// MARK: ---` pattern to non triggering examples.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,10 +23,12 @@
   * `multiline_literal_brackets`
   * `opening_brace`
   * `void_return`
+  * `mark`
 
   [SimplyDanny](https://github.com/SimplyDanny)  
   [kishikawakatsumi](https://github.com/kishikawakatsumi)  
-  [Marcelo Fabri](https://github.com/marcelofabri)
+  [Marcelo Fabri](https://github.com/marcelofabri)  
+  [swiftty](https://github.com/swiftty)  
 
 * Print invalid keys when configuration parsing fails.  
   [SimplyDanny](https://github.com/SimplyDanny)
@@ -54,9 +56,6 @@
   non-failable conversions of UTF-8 `String` <-> `Data`.    
   [Ben P](https://github.com/ben-p-commits)
   [#5263](https://github.com/realm/SwiftLint/issues/5263)
-
-* Rewrite `mark` rule using SwiftSyntax.  
-  [swiftty](https://github.com/swiftty)
 
 #### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,7 +55,7 @@
   [Ben P](https://github.com/ben-p-commits)
   [#5263](https://github.com/realm/SwiftLint/issues/5263)
 
-* Rewrite `mark` rule using SwiftSyntax.
+* Rewrite `mark` rule using SwiftSyntax.  
   [swiftty](https://github.com/swiftty)
 
 #### Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,9 @@
   [Ben P](https://github.com/ben-p-commits)
   [#5263](https://github.com/realm/SwiftLint/issues/5263)
 
+* Rewrite `mark` rule using SwiftSyntax.
+  [swiftty](https://github.com/swiftty)
+
 #### Bug Fixes
 
 * Silence `discarded_notification_center_observer` rule in closures. Furthermore,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,10 +20,10 @@
   * `explicit_acl`
   * `identifier_name`
   * `let_var_whitespace`
+  * `mark`
   * `multiline_literal_brackets`
   * `opening_brace`
   * `void_return`
-  * `mark`
 
   [SimplyDanny](https://github.com/SimplyDanny)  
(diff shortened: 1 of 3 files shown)


Source/SwiftLintBuiltInRules/Rules/Lint/MarkRule.swift
```diff
@@ -9,239 +10,143 @@ struct MarkRule: CorrectableRule {
...
 
-    func validate(file: SwiftLintFile) -> [StyleViolation] {
-        return violationRanges(in: file, matching: pattern).map {
-            StyleViolation(ruleDescription: Self.description,
-                           severity: configuration.severity,
-                           location: Location(file: file, characterOffset: $0.location))
+            if pieces != newToken.leadingTrivia.pieces {
+                newToken.leadingTrivia = .init(pieces: pieces)
+            }
+            return super.visit(newToken)
```
SimplyDanny, line 44:
```suggestion
            return super.visit(token.with(\.leadingTrivia, Trivia(pieces: pieces)))
```

That way, you can omit the explicit copy `newToken` and the previous `if` statement.

Source/SwiftLintBuiltInRules/Rules/Lint/MarkRule.swift
```diff
@@ -9,239 +10,143 @@ struct MarkRule: CorrectableRule {
...
+            for result in token.violationResults() {
+                violations.append(result.position)
+            }
+            return .skipChildren
+        }
+    }
 
-    private let invalidLowercasePattern = "(?:// ?[Mm]ark:)"
+    final class Rewriter: ViolationsSyntaxRewriter {
+        private var correctedPositions: Set<AbsolutePosition> = []
```
SimplyDanny, line 30: Use `correctionPositions` from the parent class instead.

Source/SwiftLintBuiltInRules/Rules/Lint/MarkRule.swift
```diff
@@ -9,239 +10,143 @@ struct MarkRule: CorrectableRule {
...
-                let syntaxToken = syntaxTokens.first,
-                let syntaxKind = syntaxToken.kind,
-                SyntaxKind.commentKinds.contains(syntaxKind),
-                case let tokenLocation = Location(file: file, byteOffset: syntaxToken.offset),
-                case let matchLocation = Location(file: file, characterOffset: matchRange.location),
-                // Skip MARKs that are part of a multiline comment
-                tokenLocation.line == matchLocation.line
-            else {
-                return false
+    func violationResults(withCorrect: Bool = false) -> [ViolationResult] {
```
SimplyDanny, line 122: It seems like you don't need the parameter `withCorrect`, because all corrections are evaluated lazily anyway. You need to move the `correct()` call into the correction closure though.

SimplyDanny (commented):
Looks very good. Thanks! I have only a few small remarks ...

CHANGELOG.md
```diff
@@ -55,6 +55,9 @@
   [Ben P](https://github.com/ben-p-commits)
   [#5263](https://github.com/realm/SwiftLint/issues/5263)
 
+* Rewrite `mark` rule using SwiftSyntax.
```
SimplyDanny, line 58: You may want to add the rule and your name to the lists in line 19 instead.

swiftty:
@SimplyDanny Thanks for the review! I modified the codes.

CHANGELOG.md
```diff
@@ -23,10 +23,12 @@
   * `multiline_literal_brackets`
   * `opening_brace`
   * `void_return`
+  * `mark`
```
SimplyDanny, line 26: Alphabetically please. 🙂

Source/SwiftLintBuiltInRules/Rules/Lint/MarkRule.swift
```diff
@@ -9,239 +10,131 @@ struct MarkRule: CorrectableRule {
...
-            spaceStartPattern,
-            invalidEndSpacesPattern,
-            invalidSpacesAfterHyphenPattern,
-            invalidLowercasePattern,
-            missingColonPattern,
-            threeSlashesInsteadOfTwo
-        ].joined(separator: "|")
+private extension MarkRule {
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        override func visit(_ token: TokenSyntax) -> SyntaxVisitorContinueKind {
```
SimplyDanny, line 21: You can use `visitPost` as well. It allows you to omit the `return` and the semantic should be the same as tokens don't have children.

Source/SwiftLintBuiltInRules/Rules/Lint/MarkRule.swift
```diff
@@ -9,239 +10,131 @@ struct MarkRule: CorrectableRule {
...
+            return super.visit(token.with(\.leadingTrivia, Trivia(pieces: pieces)))
         }
     }
+}
 
-    func correct(file: SwiftLintFile) -> [Correction] {
-        var result = [Correction]()
+private struct ViolationResult {
+    var position: AbsolutePosition
+    var correct: ((inout [TriviaPiece]) -> Void)?
```
SimplyDanny, line 45: After the adaptions, `correct` cannot be `nil` anymore, can it?

SimplyDanny (approved):

PR #5412, opened 2024-01-07
Enable `superfluous_else` rule

SimplyDanny:
I obviously like this style. Okay to enforce it?


jpsim (approved):
Feel free to enable this

PR #5414, opened 2024-01-08
Rewrite `trailing_closure` rule with SwiftSyntax

KS1019:
This PR rewrites the `trailing_closure` rule with SwiftSyntax.

I understand there's #4440 already but I accidentally started working on this before checking if there are any other previous attempts for this already. I believe this PR still has some value since it's been a while since the last commit on #4440 and now we have the macro to write rules like this PR does.

Newly added examples were taken from #4440.

@marcelofabri Please let me know if this would be a problem. Thank you!


SimplyDanny (changes requested):
Thank you for the contribution! I think it's fine to take #4040 as a basis and adapt it to the current state and style of the project.

The exclusion of function calls in `if` conditions, loops and the like doesn't work for more complex expressions. You might find some ideas in #4040 that'll help you implement these cases properly. Please also add examples for them.

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
```diff
@@ -1,6 +1,8 @@
 import Foundation
```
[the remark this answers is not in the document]
KS1019: Fixed at 3c313ad

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
```diff
@@ -27,118 +34,48 @@ struct TrailingClosureRule: OptInRule {
...
 
-    func validate(file: SwiftLintFile) -> [StyleViolation] {
-        let dict = file.structureDictionary
-        return violationOffsets(for: dict, file: file).map {
-            StyleViolation(ruleDescription: Self.description,
-                           severity: configuration.severityConfiguration.severity,
-                           location: Location(file: file, byteOffset: $0))
-        }
-    }
+extension TrailingClosureRule {
```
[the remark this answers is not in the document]
KS1019: Fixed at af4cc0c

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
```diff
@@ -27,118 +34,48 @@ struct TrailingClosureRule: OptInRule {
...
+    class Visitor: ViolationsSyntaxVisitor<TrailingClosureConfiguration> {
```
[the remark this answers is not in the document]
KS1019: Fixed at e4eb519 and ce774bd

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
```diff
@@ -27,118 +34,48 @@ struct TrailingClosureRule: OptInRule {
...
-            let range = file.stringView.byteRangeToNSRange(byteRange),
-            let match = regex("\\s*\\(\\s*\\{").firstMatch(in: file.contents, options: [], range: range)?.range,
-            match.location == range.location {
-            return shouldTrigger()
-        }
-
-        return false
+private extension SyntaxProtocol {
+    var isNotNextToLeftBrace: Bool {
+        nextToken(viewMode: .sourceAccurate)?.tokenKind != .leftBrace
```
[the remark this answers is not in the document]
KS1019: Fixed at 186eb0f

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
```diff
@@ -27,118 +34,48 @@ struct TrailingClosureRule: OptInRule {
...
-        else {
-            return false
-        }
-
-        let pattern = regex("\\)\\s*\\)\\z")
-        return pattern.numberOfMatches(in: file.contents, range: range) > 0
+    var shouldTrigger: Bool {
+        // If at least last two arguments were ClosureExprSyntax, a violation should not be triggered.
+        arguments.count <= 1
+        || !arguments.dropFirst(arguments.count - 2).allSatisfy({ $0.expression.is(ClosureExprSyntax.self) })
```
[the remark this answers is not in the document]
KS1019: Fixed at 9350607

CHANGELOG.md
```diff
@@ -55,6 +55,9 @@
   [Ben P](https://github.com/ben-p-commits)
   [#5263](https://github.com/realm/SwiftLint/issues/5263)
 
+* Rewrite `trailing_closure` rule with SwiftSyntax
```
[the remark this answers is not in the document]
KS1019: Fixed

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
```diff
@@ -18,127 +19,78 @@ struct TrailingClosureRule: OptInRule {
...
-            let nameLength = dictionary.nameLength,
-            case let start = nameOffset + nameLength,
-            case let length = totalLength + offset - start,
-            case let byteRange = ByteRange(location: start, length: length),
-            let range = file.stringView.byteRangeToNSRange(byteRange),
-            let match = regex("\\s*\\(\\s*\\{").firstMatch(in: file.contents, options: [], range: range)?.range,
-            match.location == range.location {
-            return shouldTrigger()
+        override func visit(_ node: ForStmtSyntax) -> SyntaxVisitorContinueKind {
+            .skipChildren
```
SimplyDanny, line 65: I don't think the following example would trigger with this change:

```swift
for n in list {
    ↓n.forEach({ print($0) })
}
```

Think about how to keep traversing the loop's body while skipping everything else. 😉

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
```diff
@@ -18,127 +19,78 @@ struct TrailingClosureRule: OptInRule {
...
-    private func isAnonymousClosureCall(dictionary: SourceKittenDictionary, file: SwiftLintFile) -> Bool {
-        guard let byteRange = dictionary.byteRange,
-            let range = file.stringView.byteRangeToNSRange(byteRange)
-        else {
-            return false
-        }
+    var shouldTrigger: Bool {
+        // If at least last two arguments were ClosureExprSyntax, a violation should not be triggered.
+        arguments.count <= 1
+        || !arguments.dropFirst(arguments.count - 2).allSatisfy({ $0.expression.is(ClosureExprSyntax.self) })
```
SimplyDanny, line 88: If I understand it correctly, this method only makes sense together with `node.lastArgumentIsClosure` being checked in front. So you should better combine both into one.

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
```diff
@@ -18,127 +19,78 @@ struct TrailingClosureRule: OptInRule {
...
-                let match = regex("\\s*\\{").firstMatch(in: file.contents, options: [], range: range)?.range,
-                match.location == range.location
-            else {
-                return false
-            }
-
-            return true
-        }
+private extension SyntaxProtocol {
+    var isNotNextToLeftBrace: Bool {
```
SimplyDanny, line 71: No longer needed.

CHANGELOG.md
```diff
@@ -34,11 +34,13 @@
   * `nimble_operator`
   * `opening_brace`
   * `void_return`
+  * `trailing_closure`
```
SimplyDanny, line 37: Alphabetically please. 😊

SimplyDanny (changes requested):

CHANGELOG.md
KS1019: Fixed at 3fb95bc

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
KS1019: Fixed at ca8e2e3

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
KS1019: Fixed at 6b1d0f9

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
KS1019: Fixed at 4636f40

(Finally understanding the point of the `walk()` function. It is pretty useful. 😀)

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
```diff
@@ -18,127 +19,81 @@ struct TrailingClosureRule: OptInRule {
...
             Example("↓foo.reduce(0, combine: { $0 + 1 })"),
             Example("↓offsets.sorted(by: { $0.offset < $1.offset })"),
-            Example("↓foo.something(0, { $0 + 1 })")
+            Example("↓foo.something(0, { $0 + 1 })"),
+            Example("↓foo.something(param1: { _ in true }, param2: 0, param3: { _ in false })"),
+            Example("""
+            for n in list {
+                ↓n.forEach({ print($0) })
+            }
+            """)
```
SimplyDanny, line 50: Let's exclude this example from being documented. It isn't special from a user's perspective:
```suggestion
            """, excludeFromDocumentation: true)
```

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
```diff
@@ -18,127 +19,81 @@ struct TrailingClosureRule: OptInRule {
...
+        override func visitPost(_ node: FunctionCallExprSyntax) {
+            guard node.trailingClosure == nil else { return }
 
-        if dictionary.expressionKind == .call,
-            shouldBeTrailingClosure(dictionary: dictionary, file: file),
-            let offset = dictionary.offset {
-            results = [offset]
-        }
+            if configuration.onlySingleMutedParameter {
+                guard node.containsOnlySingleMutedParameter else { return }
```
SimplyDanny, line 61: Condition can be combined with the `if` above. So it's clearer that both conditions must hold.

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
```diff
@@ -18,127 +19,81 @@ struct TrailingClosureRule: OptInRule {
...
 
-        if let kind = dictionary.statementKind, kind != .brace {
-            // trailing closures are not allowed in `if`, `guard`, etc
-            results += dictionary.substructure.flatMap { subDict -> [ByteCount] in
-                guard subDict.statementKind == .brace else {
-                    return []
+                violations.append(node.positionAfterSkippingLeadingTrivia)
+            } else {
+                if node.shouldTrigger {
```
SimplyDanny, line 65: Use `else if`.

SimplyDanny (approved):
Just a few more nits ...

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
KS1019: I'm not sure if I understand your point correctly. 

If we combine two conditions into one with `&&`, it makes sense for the case when both of them are `true` but it won't for the case when `configuration.onlySingleMutedParameter` is `true` and `node.containsOnlySingleMutedParameter` is `false` because it will incorrectly evaluate `else` or `else if` below and can append an incorrect violation when we want it to exit the function without adding any. Doing this causes `TrailingClosureRuleTests` to fail (at least on my machine).

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
KS1019: Fixed at f371ffb

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
KS1019: Fixed at e1efa80

Source/SwiftLintBuiltInRules/Rules/Style/TrailingClosureRule.swift
SimplyDanny: You are right. I confused myself thinking about the whole method and how it could be written differently.

KS1019: Fixed at f371ffb

SimplyDanny:
Thank you for the contribution and the patience taking up all my review comments, @KS1019!

PR #5417, opened 2024-01-13
Rewrite `orphaned_doc_comment` rule with SwiftSyntax

swiftty:
This PR rewrites the `orphaned_doc_comment` rule with SwiftSyntax.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/OrphanedDocCommentRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/OrphanedDocCommentRule.swift
@@ -58,8 +58,8 @@ struct OrphanedDocCommentRule: Rule {
             """),
             Example("""
             extension Nested {
-                ///
-                ↓/// Look here for more info: https://github.com.
+                ↓///
+                /// Look here for more info: https://github.com.
 
                 // Not a doc string
                 var myGreatProperty: String!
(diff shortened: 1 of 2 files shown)


Source/SwiftLintBuiltInRules/Rules/Lint/OrphanedDocCommentRule.swift
```diff
@@ -40,36 +41,77 @@ struct OrphanedDocCommentRule: SourceKitFreeRule {
...
+            ↓/// Look here for more info: https://github.com.
+            // Not a doc string
+            ↓/// My great property
+            // Not a doc string
+            var myGreatProperty: String!
+            """),
+            Example("""
+            extension Nested {
+                ///
+                ↓/// Look here for more info: https://github.com.
```
SimplyDanny, line 62: I think it was more appropriate if the rule would trigger on the first doc comment instead of the last. In this example, that would be one line above. What do you think?

SimplyDanny (approved):
Looks very good. Thanks!

There is only one question from my side:

Source/SwiftLintBuiltInRules/Rules/Lint/OrphanedDocCommentRule.swift
swiftty: Thanks for the review!

My first implementation triggers on the first line of the doc, but the original implementation triggers on the last line.
Therefore, I changed the behavior, but I prefer it to be triggered on the first line!
https://github.com/realm/SwiftLint/pull/5417/commits/d75e29e7e14f4d2c2f7576e8305d504158aa5e1c

SimplyDanny: As the findings in OSS project haven't changed with your rewrite, we can consider it correct. 👍 

So we may change the violation position now and ignore all OSS findings that are going to be reported due to the position change.

Are you motivated to do that? Otherwise, I'm also fine with merging the PR as is.

PR #5429, opened 2024-01-18
Don't trigger the return_value_from_void_function warning from initializers

mrbkap:
Consider the code:
```swift
func foo() {
   enum Bar: RawRepresentable() {
       case one
       case two

       init?(rawValue: String) {
           return nil
       }
   }
}
```

This example should not trigger `return_value_from_void_function` in the initializer because the initializer is nullable and can return `nil`. However, using the latest version of `SwiftLint` I'm seeing this rule trigger on this example.

The problem appears to be that `Syntax.enclosingFunction` extension climbs the AST's parent chain looking for a `FunctionDecl` but initializers are neither blocks nor variable declarations. By bailing when we see an initializer, we won't trigger this rule incorrectly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,7 +130,7 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#3852](https://github.com/realm/SwiftLint/pull/3852)
 
-* Don't trigger the return_value_from_void_function warning from initializers.  
+* Don't trigger the `return_value_from_void_function` warning from initializers.  
   [mrbkap](https://github.com/mrbkap)
   [#5429](https://github.com/realm/SwiftLint/pull/5429)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,7 +132,6 @@
 
 * Don't trigger the `return_value_from_void_function` warning from initializers.  
   [mrbkap](https://github.com/mrbkap)
-  [#5429](https://github.com/realm/SwiftLint/pull/5429)
 
 ## 0.54.0: Macro-Economic Forces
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -130,6 +130,10 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#3852](https://github.com/realm/SwiftLint/pull/3852)
 
+* Don't trigger the return_value_from_void_function warning from initializers.
```
SimplyDanny, line 133:
```suggestion
* Don't trigger the `return_value_from_void_function` warning from initializers.  
```

CHANGELOG.md
```diff
@@ -130,6 +130,10 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#3852](https://github.com/realm/SwiftLint/pull/3852)
 
+* Don't trigger the return_value_from_void_function warning from initializers.  
+  [mrbkap](https://github.com/mrbkap)
+  [#5429](https://github.com/realm/SwiftLint/pull/5429)
```
SimplyDanny, line 135:
```suggestion
  [mrbkap](https://github.com/mrbkap)
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/ReturnValueFromVoidFunctionRuleExamples.swift
```diff
@@ -45,6 +45,17 @@ internal struct ReturnValueFromVoidFunctionRuleExamples {
...
+        func baz() {
+            enum Foo {
+                case bar
+
+                init?(arg: String?) {
+                    return nil
+                }
+            }
+        }
+        """),
```
SimplyDanny, line 57:
```suggestion
        """, excludeFromDocumentation: true),
```

SimplyDanny (approved):
Looks good to me. Thanks!

mrbkap:
Thanks for the quick action and cleanups!

PR #5437, opened 2024-01-23
Allow line_length rule configuration accept regex patterns to exclude

kasrababaei:
We noticed there's no way to define a set of regex to skip running the `line_length` rule on lines that match the pattern. 

This is to avoid disabling the rule in many places. A good example, that actually made me put up a PR, is having long test names. We use Quick and Nimble in our project and don't want to discourage developers to shorten test descriptions. 

So, provided the configuration is:
```
line_length:
  warning: 30
  error: 50
  ignores_comments: true
  ignores_urls: true
  excluded_patterns:
    - "^print"
```

swiflint won't trigger a warning for this sample code:
```
enum Foo {
    static func testFunc() {
        print("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA") // This wouldn't trigger a warning
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,10 @@
   [Oleg Kokhtenko](https://github.com/kohtenko)
   [#5405](https://github.com/realm/SwiftLint/issues/issue_number)
 
+* Add `excluded_patterns` to `line_length` to avoid linting lines
+  that match the patterns.
+  [kasrababaei](https://github.com/kasrababaei)
+
 #### Bug Fixes
 
 * Silence `discarded_notification_center_observer` rule in closures. Furthermore,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,8 +104,8 @@
   [Oleg Kokhtenko](https://github.com/kohtenko)
   [#5405](https://github.com/realm/SwiftLint/issues/issue_number)
 
-* Add `excluded_patterns` to `line_length` to avoid linting lines
-  that match the patterns.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -104,6 +104,10 @@
   [Oleg Kokhtenko](https://github.com/kohtenko)
   [#5405](https://github.com/realm/SwiftLint/issues/issue_number)
 
+* Add `excluded_patterns` to `line_length` to avoid linting lines
+  that match the patterns.
```
SimplyDanny, line 108: Please append two spaces to the end of the description. They are required to enforce a line break in the rendered Markdown.


You say here that the patterns must **match** the line but the implementation checks for **containment**.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/LineLengthConfiguration.swift
```diff
@@ -14,6 +14,8 @@ struct LineLengthConfiguration: RuleConfiguration {
     private(set) var ignoresComments = false
     @ConfigurationElement(key: "ignores_interpolated_strings")
     private(set) var ignoresInterpolatedStrings = false
+    @ConfigurationElement(key: "excluded_patterns")
```
SimplyDanny, line 17: The name should indicate that the exclusion applies to lines:
```suggestion
    @ConfigurationElement(key: "excluded_lines_patterns")
```

SimplyDanny (changes requested):
Sounds like a reasonable addition.

SimplyDanny (approved):

PR #5438, opened 2024-01-25
Add failing ClosureEndIndentationRule example

keith:


keith:
if someone wants to take a fix for this one please do!

PR #5439, opened 2024-01-26
Rewrite `nesting` rule with SwiftSyntax

swiftty:
This PR rewrites the `nesting` rule with SwiftSyntax.


swiftty:
In the current implementation, it is triggered after the access level control https://github.com/realm/SwiftLint/pull/5439/commits/9891aef6623c2a74155cd35b10f05146df32735b, but I prefer the behavior where the node is triggered at its start position. What do you think?

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -12,137 +13,171 @@ struct NestingRule: Rule {
...
+        override func visitPost(_ node: EnumDeclSyntax) {
+            validationData.pop()
+        }
 
-    private func levelViolation(file: SwiftLintFile, dictionary: SourceKittenDictionary,
-                                previousKind: SwiftStructureKind?, level: Int, forFunction: Bool) -> StyleViolation? {
-        guard let offset = dictionary.offset else {
-            return nil
+        override func visit(_ node: ExtensionDeclSyntax) -> SyntaxVisitorContinueKind {
+            validate(node, inFunction: validationData.isFunction)
```
SimplyDanny, line 80: Even though the rule cannot trigger on extensions, you may pass the `extensionKeyword` to the `validate` method to avoid the nilable argument.

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -12,137 +13,171 @@ struct NestingRule: Rule {
...
+            let isFunction = validationData.isFunction
+            let (level, targetLevel) = if isFunction {
+                (validationData.functionLevel, configuration.functionLevel)
+            } else {
+                (validationData.typeLevel, configuration.typeLevel)
+            }
+
+            var violatingSeverity: ViolationSeverity?
+
+            if configuration.alwaysAllowOneTypeInFunctions, !inFunction {
```
SimplyDanny, line 150: Could be an `if` expression that is assigned to `violatingSeverity` directly.

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -12,137 +13,171 @@ struct NestingRule: Rule {
...
-            ruleDescription: Self.description,
-            severity: severity,
-            location: Location(file: file, byteOffset: offset),
-            reason: "\(targetName) should be nested at most \(threshold) level\(pluralSuffix) deep"
-        )
+        // MARK: -
+        private func validate(_ node: some DeclSyntaxProtocol, triggeringToken: TokenSyntax? = nil, inFunction: Bool) {
+            validationData.push(node)
+            let isFunction = validationData.isFunction
+            let (level, targetLevel) = if isFunction {
```
SimplyDanny, line 142: While I'm not agains the new `if` expressions, I'm still unsure how to format them in a clear and readable way. Either

```swift
let x = if ... {
            a
        } else {
            b
        }
```
or
```swift
let x =
    if ... {
        a
    } else {
        b
    }
```
look a bit better, though.

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -12,137 +13,171 @@ struct NestingRule: Rule {
...
+            guard let severity = violatingSeverity else {
+                return
+            }
+
+            let targetName = isFunction ? "Functions" : "Types"
+            let threshold = configuration.threshold(with: targetLevel, for: severity)
+            let pluralSuffix = threshold > 1 ? "s" : ""
+            let position = (triggeringToken?.positionAfterSkippingLeadingTrivia
+                            ?? node.positionAfterSkippingLeadingTrivia)
+            violations.append(.init(
```
SimplyDanny, line 167: Please specify the type:
```suggestion
            violations.append(ReasonedRuleViolation(
```

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -12,137 +13,171 @@ struct NestingRule: Rule {
...
+        }
+
+        // MARK: - configuration for ignoreTypealiasesAndAssociatedtypes
+        override func visitPost(_ node: TypeAliasDeclSyntax) {
+            guard !configuration.ignoreTypealiasesAndAssociatedtypes else { return }
+            validate(node, triggeringToken: node.typealiasKeyword, inFunction: false)
+            validationData.pop()
+        }
+
+        // associatedtype is only permitted within a protocol,
```
SimplyDanny, line 113: That's true for Swift versions < 5.10. Swift 5.10 [allows](https://github.com/apple/swift-evolution/blob/main/proposals/0404-nested-protocols.md) to nest protocols inside of other non-generic contexts. Only then, the option starts to make sense eventually.

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -12,137 +13,171 @@ struct NestingRule: Rule {
...
+        // so it maybe not necessary to check the configuration.
+        override func visitPost(_ node: AssociatedTypeDeclSyntax) {
+            guard !configuration.ignoreTypealiasesAndAssociatedtypes else { return }
+            validate(node, triggeringToken: node.associatedtypeKeyword, inFunction: false)
+            validationData.pop()
+        }
+
+        // MARK: - configuration for checkNestingInClosuresAndStatements
+        override func visit(_ node: ClosureExprSyntax) -> SyntaxVisitorContinueKind {
+            if !configuration.checkNestingInClosuresAndStatements {
```
SimplyDanny, line 123: Use a `guard` to avoid the negation.

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -12,137 +13,171 @@ struct NestingRule: Rule {
...
+                return .skipChildren
+            } else {
+                return super.visit(node)
+            }
         }
 
-        guard let severity = violatingSeverity else {
-            return nil
+        override func visit(_ node: CodeBlockItemSyntax) -> SyntaxVisitorContinueKind {
+            if !configuration.checkNestingInClosuresAndStatements && node.parent?.inStatement ?? false {
```
SimplyDanny, line 131:
```suggestion
            if !configuration.checkNestingInClosuresAndStatements, node.parent?.inStatement ?? false {
```

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -12,137 +13,171 @@ struct NestingRule: Rule {
...
+    private(set) var functionLevel: Int = -1
+    private var declStack = Stack<any DeclSyntaxProtocol>()
 
-    private struct ValidationArgs {
-        var typeLevel: Int = -1
-        var functionLevel: Int = -1
-        var previousKind: SwiftStructureKind?
-        var violations: [StyleViolation] = []
+    var isFunction: Bool {
+        declStack.peek()?.is(FunctionDeclSyntax.self) ?? false
```
SimplyDanny, line 24: The only fact to know from the top stack object is whether it is a function. So let's only store this information and avoid to deal with potentially expensive existential types.

SimplyDanny (changes requested):
Looks good. Thanks for the rewrite!

Please consider my remarks ...

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
swiftty: Thanks!, The former alignment is too much work, so I changed to the latter.

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -18,23 +18,18 @@ struct NestingRule: Rule {
 private struct ValidationData {
     private(set) var typeLevel: Int = -1
     private(set) var functionLevel: Int = -1
-    private var declStack = Stack<any DeclSyntaxProtocol>()
+    private(set) var functionOrNotStack = Stack<Bool>()
 
-    var isFunction: Bool {
```
SimplyDanny: I liked this property. Why not keep it to "hide" the implementation details? 😉 

Perhaps name it `lastIsFunction`.

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -18,23 +18,18 @@ struct NestingRule: Rule {
 private struct ValidationData {
     private(set) var typeLevel: Int = -1
     private(set) var functionLevel: Int = -1
-    private var declStack = Stack<any DeclSyntaxProtocol>()
+    private(set) var functionOrNotStack = Stack<Bool>()
```
SimplyDanny, line 21: I think, they can all be `private` without `(set)`.

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -12,137 +13,161 @@ struct NestingRule: Rule {
...
-    private let omittedStructureKinds = SwiftDeclarationKind.variableKinds
-        .union([.enumcase, .enumelement])
-        .map(SwiftStructureKind.declaration)
-
-    private struct ValidationArgs {
-        var typeLevel: Int = -1
-        var functionLevel: Int = -1
-        var previousKind: SwiftStructureKind?
-        var violations: [StyleViolation] = []
+private struct ValidationData {
```
SimplyDanny, line 18:
```suggestion
private struct Levels {
```

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -12,137 +13,161 @@ struct NestingRule: Rule {
...
-        }
+    mutating func push(_ isFunction: Bool) {
+        functionOrNotStack.push(isFunction)
+        updateLevel(with: 1, isFunction: isFunction)
     }
 
-    func validate(file: SwiftLintFile) -> [StyleViolation] {
-        return validate(file: file, substructure: file.structureDictionary.substructure, args: ValidationArgs())
+    mutating func pop() {
+        updateLevel(with: -1, isFunction: functionOrNotStack.pop()!)
```
SimplyDanny, line 29: Please avoid force-unwrapping.

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -12,137 +13,161 @@ struct NestingRule: Rule {
...
-            location: Location(file: file, byteOffset: offset),
-            reason: "\(targetName) should be nested at most \(threshold) level\(pluralSuffix) deep"
-        )
+        override func visitPost(_ node: StructDeclSyntax) {
+            validationData.pop()
+        }
+
+        // MARK: - configuration for ignoreTypealiasesAndAssociatedtypes
+        override func visitPost(_ node: TypeAliasDeclSyntax) {
+            guard !configuration.ignoreTypealiasesAndAssociatedtypes else { return }
```
SimplyDanny, line 103: Nit: Use `if` to avoid negation.

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -12,137 +13,161 @@ struct NestingRule: Rule {
...
+            validate(forFunction: false, triggeringToken: node.typealiasKeyword)
+            validationData.pop()
+        }
+
+        override func visitPost(_ node: AssociatedTypeDeclSyntax) {
+            guard !configuration.ignoreTypealiasesAndAssociatedtypes else { return }
```
SimplyDanny, line 109: Nit: Use `if` to avoid negation.

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -12,137 +13,161 @@ struct NestingRule: Rule {
...
+            functionLevel += value
+        } else {
+            typeLevel += value
         }
     }
+}
+
+private extension NestingRule {
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        override var skippableDeclarations: [any DeclSyntaxProtocol.Type] { [ProtocolDeclSyntax.self] }
```
SimplyDanny, line 43: This is contra productive later when nested protocols are allowed. Perhaps you can add a test for this case already.

SimplyDanny (changes requested):

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
swiftty: I made all properties private scoped! Could you review it?

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -12,137 +13,175 @@ struct NestingRule: Rule {
...
-        .union([.enumcase, .enumelement])
-        .map(SwiftStructureKind.declaration)
+    private var lastIsFunction: Bool { functionOrNotStack.peek() == true }
 
-    private struct ValidationArgs {
-        var typeLevel: Int = -1
-        var functionLevel: Int = -1
-        var previousKind: SwiftStructureKind?
-        var violations: [StyleViolation] = []
+    mutating func validate(
```
SimplyDanny, line 25: Oh sorry, I missed the fact that access to the internal levels is needed in this method. In that case, `private(set)` was actually just fine. The `validate` method doesn't really fit into the `Levels` type otherwise as its sole purpose is to keep track of levels - nothing more.

swiftty: Thanks! I moved `validate` logic into Visitor's method.

SimplyDanny:
> In the current implementation, it is triggered after the access level control [9891aef](https://github.com/realm/SwiftLint/commit/9891aef6623c2a74155cd35b10f05146df32735b), but I prefer the behavior where the node is triggered at its start position. What do you think?

Yes, currently the rule triggers on the declaration keyword. Better would be the real beginning of it or at its name. Note that the order of attributes is not strict, e.g. there could also be a declaration written as `static private func f() {}`.

Either way, this rewrite is good to go. A potential violation position change is up to another PR.

SimplyDanny:
Looks like the builds fail due to missing disk space. This might take a while to fix.

SimplyDanny (approved):

PR #5440, opened 2024-01-26
Add new opt-in rule `discouraged_orphan_init`

otaviocc:
This pull request introduces a new **opt-in** rule, `discouraged_orphan_init`, aimed at improving code readability and clarity within Swift projects. The rule discourages the use of `.init` without explicit type declaration, addressing a common readability issue in code reviews and collaborative environments.

In many instances, using `.init` alone can make it challenging to understand the type being initialized, especially when the type cannot be easily inferred from the context. This lack of clarity can lead to confusion and increased time spent on understanding and reviewing code. By enforcing explicit type declaration with initializers, this rule helps ensure that the code is more self-documenting and easier to read.

Some triggering examples:

```swift
callSomething(↓.init())
callSomething(↓.init(parameter: 42))

callSomething(
    parameter: ↓.init()
)

someClosure {
    ↓.init()
}
```

I believe this rule will be a valuable addition to our SwiftLint configuration, and I look forward to your feedback and suggestions on this pull request.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -16,6 +16,7 @@ disabled_rules:
   - conditional_returns_on_newline
   - convenience_type
   - discouraged_optional_collection
+  - discouraged_orphan_init
   - explicit_acl
   - explicit_enum_raw_value
   - explicit_top_level_acl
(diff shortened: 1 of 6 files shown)


SimplyDanny:
I think this is already addressed by the `include_bare_init` option in the `explicit_init` rule, isn't it?

otaviocc:
> I think this is already addressed by the `include_bare_init` option in the `explicit_init` rule, isn't it?

Oh, that's perfect! Nice catch @SimplyDanny, thanks!

PR #5447, opened 2024-01-31
Make private_swiftui_state correctable

mt00chikin:
The `private_swiftui_state` rule that was introduced in 0.53.0 ensures that any SwiftUI State variables are declared as private, per Apple's [documentation](https://developer.apple.com/documentation/swiftui/state).

However, the first iteration of this rule did not support autocorrect. In an existing project with a significant SwiftUI footprint, this could make the task of fixing any existing violations somewhat difficult. It should be possible to insert the `private` modifier on a State variable declaration automatically, so this adds a Rewriter to the rule that can replace any existing modifiers on a State variable declaration with `private`.


Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRuleExamples.swift
```diff
@@ -0,0 +1,243 @@
+internal struct PrivateSwiftUIStatePropertyRuleExamples {
```
mt00chikin, line 1: I added a couple new test cases to cover existing behavior and some corrections to validate the new Rewriter, but that put the rule declaration file over the 400 line limit. So I went ahead and took the opportunity to move the examples into their own file. Let me know if there's a better option!

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -247,6 +77,64 @@ private extension PrivateSwiftUIStatePropertyRule {
...
+            let privateModifier = DeclModifierSyntax(
+                leadingTrivia: node.bindingSpecifier.leadingTrivia,
+                name: .keyword(.private),
+                trailingTrivia: .space
+            )
+            let newBindingSpecifier = node.bindingSpecifier.with(\.leadingTrivia, [])
+
+            let newModifiers: DeclModifierListSyntax = [privateModifier]
+            let newNode = node
+                .with(\.modifiers, newModifiers)
```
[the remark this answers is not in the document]
mt00chikin: Good question, my thought was that we'd want to replace any existing ACL modifiers on State properties when applying this correction, since the rule expects any state property to be declared as private.
That might be a bad assumption. I haven't sent any examples of private(set) being used for state properties, but I could try to account for that possibility. 
Any thoughts on how to handle existing modifiers that might conflict with the corrected private ACL modifier?

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -247,6 +77,64 @@ private extension PrivateSwiftUIStatePropertyRule {
             violations.append(decl.bindingSpecifier.positionAfterSkippingLeadingTrivia)
         }
     }
+
+    final class Rewriter: ViolationsSyntaxRewriter {
+        /// LIFO stack that stores type inheritance clauses for each visited node
+        /// The last value is the inheritance clause for the most recently visited node
+        /// A nil value indicates that the node does not provide any inheritance clause
+        private var visitedTypeInheritances = Stack<InheritanceClauseSyntax?>()
```
SimplyDanny, line 85: This is some duplication. There is a mechanism that allows you to collect replacements and additions in the visitor already, sparing you to implement a rewriter (and duplicate code). It's a bit restricted but might be sufficient for this rule. Check out https://github.com/realm/SwiftLint/blob/4078e41f1b1000a43cee065f3cd4d7160dd74911/Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRule.swift#L218 to get an idea of how this works.

mt00chikin: Thanks for the feedback. I was originally unsure how to avoid the duplication problem, but I didn't realize I didn't need an explicit rewriter and could use Visitor to handle corrections. 

I've tried to change this up a bit to use that pattern by taking advantage of some existing extensions on DeclModifierListSyntax. I'm just checking for the presence of a single access level modifier and substituting `private` in its place if it's higher than allowed for the scope of this rule. It doesn't necessarily solve for the `private(set) public` modifier situation, though.

SimplyDanny: I wonder if `private(set) public` is a constellation, the rule should not complain about anyway. Are SwiftUI states allowed to be read by external clients? 🤔

SimplyDanny: > So perhaps I could consider private(set) to be a non-triggering example...

Sounds reasonable. I'd support that.

mt00chikin: It turns out that declaring the access level of the property's setter as private causes the property to be exposed in the memberwise initializer. 

The following code will build:
```
public struct SwiftUIView: View {
    @State private(set) public var isPlaying: Bool = false

    public var body: some View {
        Text(/*@START_MENU_TOKEN@*/"Hello, World!"/*@END_MENU_TOKEN@*/)
        Toggle(isOn: $isPlaying) {
            Text("Playing")
        }
    }
}

#Preview {
    SwiftUIView(isPlaying: true)
}
```

Declaring the State property as private prevents the value from being set in a memberwise initializer entirely. Based on that observation, I think that this rule needs to use strict enforcement of the private access level on each state property.

SimplyDanny: I see. Thanks for investigating. In that case, an explicit rewriter would be the way to go.

mt00chikin:
I've updated the branch to return to an explicit rewriter. 
I also made a slight change to how violation positions are determined that should hopefully make failures more intuitive - if an existing violating access level modifier is present, then it'll mark the violation at the access level modifier. Otherwise, it will continue to mark the violation at the binding specifier.

CHANGELOG.md
```diff
@@ -121,6 +121,11 @@
 
 * Make `empty_count` auto-correctable.  
   [KS1019](https://github.com/KS1019/)
+  
+* Make `private_swiftui_state` auto-correctable. 
+  Access control has been made stricter and treats `fileprivate` as a violation.
```
SimplyDanny, line 126: Please append two spaces to the end of the description. They are required to enforce a line break in the rendered Markdown:
```suggestion
* Make `private_swiftui_state` auto-correctable. Access control
  has been made stricter and treats `fileprivate` as a violation.  
```

CHANGELOG.md
```diff
@@ -121,6 +121,11 @@
...
+  [mt00chikin](https://github.com/mt00chikin)
+  [#5447](https://github.com/realm/SwiftLint/pull/5447)
```
SimplyDanny, line 128: No need to link to a PR. The number will be appended to the commit message upon merge anyway:
```suggestion

```

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -244,7 +74,74 @@ private extension PrivateSwiftUIStatePropertyRule {
...
+        override func visit(_ node: ActorDeclSyntax) -> DeclSyntax {
+            visitedTypeInheritances.push(node.inheritanceClause)
+            return super.visit(node)
+        }
+
+        override func visitPost(_ node: Syntax) {
+            guard node.is(ClassDeclSyntax.self) ||
+                    node.is(StructDeclSyntax.self) ||
+                    node.is(ActorDeclSyntax.self) else { return }
+            visitedTypeInheritances.pop()
```
SimplyDanny, line 110:
```suggestion
            if node.is(ClassDeclSyntax.self) ||
               node.is(StructDeclSyntax.self) ||
               node.is(ActorDeclSyntax.self) { 
                visitedTypeInheritances.pop()
            }
```

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -244,7 +74,74 @@ private extension PrivateSwiftUIStatePropertyRule {
...
+        }
+
+        override func visit(_ node: VariableDeclSyntax) -> DeclSyntax {
+            guard
+                let parent = node.parent,
+                parent.is(MemberBlockItemSyntax.self),
+                let inheritanceClause = visitedTypeInheritances.peek() as? InheritanceClauseSyntax,
+                inheritanceClause.conformsToApplicableSwiftUIProtocol,
```
SimplyDanny, line 118: Would be better to store just this attribute instead of the whole object (also in the visitor).

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -244,7 +74,74 @@ private extension PrivateSwiftUIStatePropertyRule {
...
+                node.attributes.hasStateAttribute,
+                !node.modifiers.containsPrivateOrFileprivate()
```
SimplyDanny, line 120: The conditions deciding about a violation could be extracted into an extension on `VariableDeclSyntax`, so that they can be reused in both the visitor and the rewriter. Let the visitor stop for `VariableDeclSyntax` as well to have more common parts.

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRuleExamples.swift
```diff
@@ -0,0 +1,247 @@
...
+            @State
+            ↓var isPlaying: Bool = false
+        }
+        """): Example("""
+                        struct ContentView: View {
+                            @State
+                            private var isPlaying: Bool = false
+                        }
+                        """)
+    ]
```
SimplyDanny, line 246: Please check that comments between the modifiers are kept.

SimplyDanny (changes requested):

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRuleExamples.swift
mt00chikin: Good call, it was dropping leading and trailing trivia from any existing modifiers. 
I updated it so that any trivia attached to the removed modifiers is persisted and appended to the inserted `private` access modifier. 
Can you think of any issues that might arise from that, or other situations that I might need to account for?

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
mt00chikin: I modified the stack so that it just stores the Bool result of this computed property. `true` tells the visitor or rewriter that the current scope should be evaluated for violations.

CHANGELOG.md
```diff
@@ -121,6 +121,10 @@
 
 * Make `empty_count` auto-correctable.  
   [KS1019](https://github.com/KS1019/)
+  
+* Make `private_swiftui_state` auto-correctable.  
+  Access control has been made stricter and treats `fileprivate` as a violation.
```
SimplyDanny, line 126: 2 spaces at the end of the last line of the description are sufficient.
```suggestion
* Make `private_swiftui_state` auto-correctable. Access control
  has been made stricter and treats `fileprivate` as a violation.  
```

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -202,58 +32,170 @@ private extension PrivateSwiftUIStatePropertyRule {
         }
 
         /// LIFO stack that stores type inheritance clauses for each visited node
-        /// The last value is the inheritance clause for the most recently visited node
-        /// A nil value indicates that the node does not provide any inheritance clause
-        private var visitedTypeInheritances = Stack<InheritanceClauseSyntax?>()
+        /// `true` indicates the the most recently visited decl syntax 
+        /// should be evaluated for SwiftUI State access levels
+        private var visitedScopes = Stack<Bool>()
```
SimplyDanny, line 37:
```suggestion
        private var swiftUITypeScopes = Stack<Bool>()
```

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -202,58 +32,170 @@ private extension PrivateSwiftUIStatePropertyRule {
...
             return .visitChildren
         }
 
         override func visitPost(_ node: ActorDeclSyntax) {
-            visitedTypeInheritances.pop()
+            visitedScopes.pop()
+        }
+
+        override func visit(_ node: VariableDeclSyntax) -> SyntaxVisitorContinueKind {
+            guard let parent = node.parent, parent.is(MemberBlockItemSyntax.self) else {
```
SimplyDanny, line 67: This condition needs to go into `visitPost`. Please have a test for it, e.g.:
```swift
struct MyView: View {
    var s: Int {
        @State // This would trigger
        var s: Int = 3
        return s
    }

    var body: some View { Text("") }
}
```

Not sure this is valid. But currently it would trigger and `private` is not allowed there.

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -202,58 +32,170 @@ private extension PrivateSwiftUIStatePropertyRule {
...
+        override func visitPost(_ node: VariableDeclSyntax) {
             guard
-                let decl = node.decl.as(VariableDeclSyntax.self),
-                let inheritanceClause = visitedTypeInheritances.peek() as? InheritanceClauseSyntax,
-                inheritanceClause.conformsToApplicableSwiftUIProtocol,
-                decl.attributes.hasStateAttribute,
-                !decl.modifiers.containsPrivateOrFileprivate()
+                let shouldCheckAccessLevel = visitedScopes.peek(),
+                shouldCheckAccessLevel,
+                node.containsSwiftUIStateAccessLevelViolation
```
SimplyDanny, line 78:
```suggestion
                visitedScopes.peek(),
                node.containsSwiftUIStateAccessLevelViolation
```

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -202,58 +32,170 @@ private extension PrivateSwiftUIStatePropertyRule {
         }
 
         /// LIFO stack that stores type inheritance clauses for each visited node
-        /// The last value is the inheritance clause for the most recently visited node
-        /// A nil value indicates that the node does not provide any inheritance clause
-        private var visitedTypeInheritances = Stack<InheritanceClauseSyntax?>()
+        /// `true` indicates the the most recently visited decl syntax 
+        /// should be evaluated for SwiftUI State access levels
```
SimplyDanny, line 36:
```suggestion
        /// LIFO stack that stores if a type conforms to SwiftUI protocols.
        /// `true` indicates that SwiftUI state properties should be
        /// checked in the scope of the last entered declaration.
```

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -202,58 +32,170 @@ private extension PrivateSwiftUIStatePropertyRule {
...
+            } else {
+                violations.append(node.bindingSpecifier.positionAfterSkippingLeadingTrivia)
+            }
+        }
+    }
+
+    final class Rewriter: ViolationsSyntaxRewriter<ConfigurationType> {
+        /// LIFO stack that stores type inheritance clauses for each visited node
+        /// `true` indicates the the most recently visited decl syntax
+        /// should be evaluated for SwiftUI State access levels
```
SimplyDanny, line 94:
```suggestion
        /// LIFO stack that stores if a type conforms to SwiftUI protocols.
        /// `true` indicates that SwiftUI state properties should be
        /// checked in the scope of the last entered declaration.
```

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -202,58 +32,170 @@ private extension PrivateSwiftUIStatePropertyRule {
...
+        private var visitedScopes = Stack<Bool>()
```
SimplyDanny, line 95:
```suggestion
        private var swiftUITypeScopes = Stack<Bool>()
```

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -202,58 +32,170 @@ private extension PrivateSwiftUIStatePropertyRule {
...
+
+        override func visit(_ node: ActorDeclSyntax) -> DeclSyntax {
+            visitedScopes.push(node.inheritanceClause.conformsToApplicableSwiftUIProtocol)
+            return super.visit(node)
+        }
+
+        override func visitPost(_ node: Syntax) {
+            if node.is(ClassDeclSyntax.self) ||
+                node.is(StructDeclSyntax.self) ||
+                node.is(ActorDeclSyntax.self) {
```
SimplyDanny, line 115:
```suggestion
            if node.is(ClassDeclSyntax.self) ||
               node.is(StructDeclSyntax.self) ||
               node.is(ActorDeclSyntax.self) {
```

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -202,58 +32,170 @@ private extension PrivateSwiftUIStatePropertyRule {
...
+                visitedScopes.pop()
+            }
+        }
+
+        override func visit(_ node: VariableDeclSyntax) -> DeclSyntax {
+            guard
+                let parent = node.parent,
+                parent.is(MemberBlockItemSyntax.self),
+                let shouldCheckAccessLevel = visitedScopes.peek(),
+                shouldCheckAccessLevel,
```
SimplyDanny, line 125:
```suggestion
                visitedScopes.peek(),
```

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -202,58 +32,170 @@ private extension PrivateSwiftUIStatePropertyRule {
...
+                node.is(StructDeclSyntax.self) ||
+                node.is(ActorDeclSyntax.self) {
+                visitedScopes.pop()
+            }
+        }
+
+        override func visit(_ node: VariableDeclSyntax) -> DeclSyntax {
+            guard
+                let parent = node.parent,
+                parent.is(MemberBlockItemSyntax.self),
```
SimplyDanny, line 123:
```suggestion
                node.parent?.is(MemberBlockItemSyntax.self) == true,
```

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -202,58 +32,170 @@ private extension PrivateSwiftUIStatePropertyRule {
...
+                // Extract the leading trivia from the binding specifier and apply it to the private modifier
+                let privateModifier = DeclModifierSyntax(
+                    leadingTrivia: node.bindingSpecifier.leadingTrivia,
+                    name: .keyword(.private),
+                    trailingTrivia: .space
+                )
+
+                let bindingSpecifier = node.bindingSpecifier.with(\.leadingTrivia, [])
+
+                let newModifiers: DeclModifierListSyntax = [privateModifier]
```
SimplyDanny, line 146: Inline.

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -202,58 +32,170 @@ private extension PrivateSwiftUIStatePropertyRule {
...
+            else {
+                return DeclSyntax(node)
+            }
+
+            correctionPositions.append(node.bindingSpecifier.positionAfterSkippingLeadingTrivia)
+
+            // If there are no modifiers present on the current syntax node,
+            // then we should retain the binding specifier's leading trivia
+            // by appending it to our inserted private access level modifier
+            guard !node.modifiers.isEmpty else {
```
SimplyDanny, line 136: Use `if` to avoid negation.

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -202,58 +32,170 @@ private extension PrivateSwiftUIStatePropertyRule {
...
+
+            let filteredModifiers = node.modifiers.filter { $0.asAccessLevelModifier == nil }
+            // Extract the leading trivia from the binding specifier and apply it to the private modifier
+            let privateModifier = DeclModifierSyntax(
+                leadingTrivia: previousAccessModifierLeadingTrivia,
+                name: .keyword(.private),
+                trailingTrivia: previousAccessModifierTrailingTrivia.merging(.space)
+            )
+
+            let newModifiers: DeclModifierListSyntax = filteredModifiers + [privateModifier]
```
SimplyDanny, line 179: ACL modifier should appear first.

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -202,58 +32,170 @@ private extension PrivateSwiftUIStatePropertyRule {
...
+            let newNode = node
```
SimplyDanny, line 180: Inline.

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
```diff
@@ -202,58 +32,170 @@ private extension PrivateSwiftUIStatePropertyRule {
...
+    }
+}
+
+private extension InheritanceClauseSyntax? {
     static let applicableSwiftUIProtocols: Set<String> = ["View", "App", "Scene"]
 
     var conformsToApplicableSwiftUIProtocol: Bool {
-        inheritedTypes.containsInheritedType(inheritedTypes: Self.applicableSwiftUIProtocols)
+        guard let self else { return false }
+        return self.inheritedTypes.containsInheritedType(inheritedTypes: Self.applicableSwiftUIProtocols)
```
SimplyDanny, line 198:
```suggestion
        self?.inheritedTypes.containsInheritedType(inheritedTypes: Self.applicableSwiftUIProtocols) ?? false
```

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRuleExamples.swift
SimplyDanny: Looks good now. This might not address all special case that could arise around comments, but that's fine.

SimplyDanny (changes requested):

Source/SwiftLintBuiltInRules/Rules/Lint/PrivateSwiftUIStatePropertyRule.swift
mt00chikin: `peek()` returns an optional value, so I'd have to nil coalesce to a default value to get this to compile. I could nil coalesce to `false` to get it down to a single line if that's preferable, though.

mt00chikin:
Thanks for the feedback, Danny. I've updated the branch with your suggestions and pushed the changes to the remote.

CHANGELOG.md
```diff
@@ -126,6 +126,10 @@
 
 * Make `empty_count` auto-correctable.  
   [KS1019](https://github.com/KS1019/)
+  
+* Make `private_swiftui_state` auto-correctable. Access control
+  has been made stricter and treats `fileprivate` as a violation.
```
SimplyDanny, line 131: I think that's not correct, it is? The rule still checks `modifiers.containsPrivateOrFileprivate()` and I'd better leave it that way (at least in this change as it's about auto-correction).

SimplyDanny (approved):

CHANGELOG.md
mt00chikin: You're correct, this was an artifact of my intermediate changes when I reverted back to the explicit rewriter approach. I'll update this so that it reflects the actual change.

PR #4607, opened 2022-11-29
Fix building tests in Xcode

ryanaveo:
Fix #4606

SwiftLintTestHelpers is a testTarget in order to use @testable imports for SwiftLintFramework, which had previously failed to compile for release when the target was declared as a regular target with the following error:
```
> swift build -c release

error: module 'SwiftLintFramework' was not compiled for testing
@testable import SwiftLintFramework
```

However, this breaks building/testing the target in Xcode with the following error:
```
Unable to resolve build file: XCBCore.BuildFile (The workspace has a reference to a missing target with GUID 'PACKAGE-TARGET:SwiftLintTestHelpers')
```

In order for SwiftLintTestHelpers to use internal symbols in SwiftLintFramework without using @testable import, we use @_spi imports instead. This fixes the issues when building on Xcode and for release.

**Test Plan**

- swift build -c release
```
Compiling plugin Generate Manual...
Compiling plugin SwiftLintPlugin...
Building for production...
[32/32] Compiling SwiftLintTestHelpers RuleDescription+Examples.swift
Build complete! (225.61s)
```
- Xcode builds successfully

**Notes**
Rule.swift needed to be broken into two files to pass the file length rule

PR #5466, opened 2024-02-10
Prefer `package` ACL over `@_spi`

SimplyDanny:
The `@_spi` workaround was introduced in #4607 to make the build work for `SwiftLintTestHelper` in Xcode and SPM.

With the new `package` access control modifier in Swift 5.9, the internal `@_spi` attribute doesn't seem to be needed to give `SwiftLintTestHelper` access to internal symbols.

@ryanaveo: You contributed #4607 back then. In your environment, would this change work?


jpsim (approved):
Nice

PR #5475, opened 2024-02-25
Baseline implementation

mildm8nnered:
Inspired by #3421 and its predecessors.

Adds a new baseline capability to SwiftLint, in the form of two new options to the lint and analyze commands.

This may be helpful for people who have large numbers of pre-existing violations in their codebase, that they cannot easily fix, and do not want to suppress individually.

This should allow rules to be enabled, and existing violations be ignored, but any new violations detected.

#### Usage:

```
  --baseline <baseline>   The path to a baseline file, which will be used to filter out detected violations.

  --write-baseline <write-baseline>
                          The path to save detected violations to as a new baseline.
```

For example:

`swiftlint --baseline /path/to/baselinefile --write-baseline /path/to/baselinefile` 

Will read the existing baseline from `baselinefile`, and also write the full set of detected violations back to the same file.

If the baseline cannot be read, that is a fatal error, unless you've also specified the same file with `--write-baseline` (so that you can use that invocation, and still have it work first time around).

In practice, in CI, I expect that you would probably keep a current baseline from your main branch, and then compare feature branches against that baseline.

Baselines can be quite large (with `--enable-all-rules` for SwiftLint, it's about 6MB), so in practice I would probably compress them before committing them to my source tree, and then uncompress them on the fly in CI before use.

We could compress and decompress them from SwiftLint itself, but I like the simplicity of the JSON format.

#### Baseline Comparisons

A baseline is a dictionary, keyed on relative file path, whose values are the array of `StyleViolation`'s in each file, and additionally the text of the line of source code that triggered the violation. Storing the text means that violations can still be matched, even if they have been moved around in the source file, changing their line numbers.

On disk, the baseline is actually stored as a sorted array, so that identical baselines will have identical on-disk representations.

When filtering detected violations according to the baseline, we first compare for exact equality between the detected and baseline violations, including line numbers.

If the sets of violations are not exactly equal, then we remove any violations that are exactly identical (again including line numbers).

We then group the violations by rule, and then again by the original source code text. For each detected violation, if the baseline does not contain a matching violation of the same rule, with the same reason, text, and severity, then the detected violation is reported. If the detected violations contain multiple violations of the same rule, with the same reason, text, and severity, then all those violations are reported, even if the baseline contains a match, as we cannot tell which of the violations map to which of the baseline violations.

It may be the case that the user fixed one violation of a rule, but introduced a completely new violation of that rule, with the same text, elsewhere, or fixed two existing violations, but added a new one. In the current implementation, as long as the number of violations of that rule (additionally matching on the reason and text) did not increase, we will ignore them.

#### Threshold Violations

Currently threshold violations are not filtered or stored in the baseline, and are calculated against the filtered violations.

#### Inspecting a Baseline

There is also a new command - `baseline`

```
$ swiftlint  help baseline 
OVERVIEW: Reports the violations in a baseline or the violations that are present in another baseline, but not in the original.

USAGE: swiftlint baseline <subcommand>

OPTIONS:
  --version               Show the version.
  -h, --help              Show help information.

SUBCOMMANDS:
  report (default)        Reports the violations in a baseline.
  compare                 Reports the violations that are present in another baseline but not in the original baseline.

  See 'swiftlint help baseline <subcommand>' for detailed help.
$ 
$ 
$ swiftlint  help baseline report
OVERVIEW: Reports the violations in a baseline.

USAGE: swiftlint baseline report <baseline> [--reporter <reporter>] [--output <output>]

ARGUMENTS:
  <baseline>              The path to the baseline file.

OPTIONS:
  --reporter <reporter>   The reporter used to report violations. The 'summary' reporter can be useful to provide an overview.
  --output <output>       The file where violations should be saved. Prints to stdout by default.
  --version               Show the version.
  -h, --help              Show help information.

$
$ swiftlint  help baseline compare
OVERVIEW: Reports the violations that are present in another baseline but not in the original baseline.

USAGE: swiftlint baseline compare <baseline> --other-baseline <other-baseline> [--reporter <reporter>] [--output <output>]

ARGUMENTS:
  <baseline>              The path to the baseline file.

OPTIONS:
  --other-baseline <other-baseline>
                          The path to a second baseline to compare against the baseline. Violations in the second baseline that are not present in the original baseline
                          will be reported.
  --reporter <reporter>   The reporter used to report violations. The 'summary' reporter can be useful to provide an overview.
  --output <output>       The file where violations should be saved. Prints to stdout by default.
  --version               Show the version.
  -h, --help              Show help information.

```

This uses the  existing reporters to report the violations contained in the baseline (`--reporter summary` can be useful with the following commands, to get an overview without having to wade through individual violations).

For example, to inspect a baseline:

`swiftlint baseline report Baseline.json `

To see what new violations have been introduced, you can run:

`swiftlint baseline compare Baseline.json --other-baseline NewBaseline.json`

This will produce the same output as:

`swiftlint lint --baseline Baseline.json`

To see what violations have been fixed, simply reverse the baselines

`swiftlint baseline compare NewBaseline.json --other-baseline Baseline.json`


mildm8nnered:
@SimplyDanny and @jpsim - would love to get some feedback on this ...

Source/SwiftLintCore/Models/Baseline.swift
```diff
@@ -0,0 +1,169 @@
...
+
+    init(violations: [StyleViolation]) {
+        baselineViolations = violations.baselineViolations.groupedByFile()
+    }
+
+    /// Writes a `Baseline` to disk in JSON format.
+    ///
+    /// - parameter violations: The violations to save.
+    /// - parameter toPath: The path to write to.
+    public static func write(_ violations: [StyleViolation], toPath path: String) throws {
```
SimplyDanny, line 48: I'd expect the call-side to be like `Baseline(violations:).write(toPath:)`, in other words, the `static` can be avoided.

Source/SwiftLintCore/Models/Baseline.swift
```diff
@@ -0,0 +1,169 @@
...
+    /// Filters out violations that are present in the `Baseline`.
+    ///
+    /// Assumes that all violations are from the same file.
+    ///
+    /// - parameter violations: The violations to filter.
+    /// - Returns: The new violations.
+    public func filter(_ violations: [StyleViolation]) -> [StyleViolation] {
+        guard let firstViolation = violations.first,
+              let baselineViolations = baselineViolations[firstViolation.location.relativeFile ?? ""],
+              baselineViolations.isNotEmpty else {
```
SimplyDanny, line 66: Please help me to understand this. Why is only the first violation checked for existence in the baseline?

Source/SwiftLintCore/Models/Baseline.swift
```diff
@@ -0,0 +1,169 @@
...
+        for (ruleIdentifier, ruleViolations) in violationsByRuleIdentifier {
+            guard
+                let baselineViolations = baselineViolationsByRuleIdentifier[ruleIdentifier],
+                    baselineViolations.isNotEmpty else {
+                filteredViolations.formUnion(ruleViolations)
+                continue
+            }
+
+            let groupedRuleViolations = Dictionary(grouping: ruleViolations) { $0.key }
+            let groupedBaselineViolations = Dictionary(grouping: baselineViolations) { $0.key }
```
SimplyDanny, line 93:
```suggestion
            let groupedRuleViolations = Dictionary(grouping: ruleViolations, by: \.key)
            let groupedBaselineViolations = Dictionary(grouping: baselineViolations, by: \.key)
```

Tests/SwiftLintFrameworkTests/BaselineTests.swift
```diff
@@ -0,0 +1,182 @@
...
+    return result
+#endif
+}
+
+private var sourceFilePath: String = {
+    temporaryFilePath + ".swift"
+}()
+
+final class BaselineTests: XCTestCase {
+    private var violations: [StyleViolation] {
```
SimplyDanny, line 23: Can be a `static let` constant.

Tests/SwiftLintFrameworkTests/BaselineTests.swift
```diff
@@ -0,0 +1,182 @@
...
+        [
+            ArrayInitRule.description,
+            BlockBasedKVORule.description,
+            ClosingBraceRule.description,
+            DirectReturnRule.description
+        ].violations
+    }
+
+    private var baseline: Baseline {
```
SimplyDanny, line 32: Can be a `static let` constant.

Tests/SwiftLintFrameworkTests/BaselineTests.swift
```diff
@@ -0,0 +1,182 @@
...
+        Baseline(violations: violations)
+    }
+
+    func testWritingAndReading() throws {
+        try testBlock {
+            let baselinePath = temporaryFilePath
+            let baseline = baseline
```
SimplyDanny, line 39:
```suggestion

```

Tests/SwiftLintFrameworkTests/BaselineTests.swift
```diff
@@ -0,0 +1,182 @@
...
+            let violation = StyleViolation(
+                ruleDescription: newViolationRuleDescription,
+                location: Location(file: sourceFilePath, line: line, character: 1)
+            )
+            newViolations.insert(violation, at: insertionIndex)
+            XCTAssertEqual(baseline.filter(newViolations), [violation])
+        }
+    }
+
+    private func testBlock(_ block: () throws -> Void) throws {
```
SimplyDanny, line 129: I wonder if this can also be achieved by implementing appropriate `setUp` and `tearDown` methods in a test class.

Tests/SwiftLintFrameworkTests/Resources/BaselineFixtures/BaselineExample.swift
SimplyDanny, line 1: You may just inline this file and have it as a text block instead.

Tests/SwiftLintFrameworkTests/BaselineTests.swift
```diff
@@ -0,0 +1,182 @@
...
+            DirectReturnRule.description
+        ].violations
+    }
+
+    private var baseline: Baseline {
+        Baseline(violations: violations)
+    }
+
+    func testWritingAndReading() throws {
+        try testBlock {
```
SimplyDanny, line 37: Is `testBlock` needed here?

SimplyDanny (changes requested):
Very good idea to store and compare the violating line for more context. To increase precision even further, could this theoretically be extended to the violating line itself plus the previous and the subsequent line? Not sure what the benefit of this would be though ...

CHANGELOG.md
```diff
@@ -139,6 +139,13 @@
   [phlippieb](https://github.com/phlippieb)
   [#5471](https://github.com/realm/SwiftLint/issues/5471)
 
+* Add two new options to the `lint` and `analyze` commands: `--write-baseline`
+  to save a baseline to disk, and `--baseline` to read a saved baseline and
+  use it to filter out detected pre-existing violations. A new `report-baseline`
+  command uses the reporters to print the violations in a baseline.  
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#5475](https://github.com/realm/SwiftLint/pull/5475)
```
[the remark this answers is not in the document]
mildm8nnered: I've added a reference to #3421 and moved it to the Experimental section

Source/SwiftLintCore/Models/Baseline.swift
mildm8nnered: So we want to check whether there are any incoming violations, and whether we have any violations in the baseline for the file. When called by the LintOrAnalyzeCommand, the method will be passed the list of violations for a specific file ("Assumes that all violations are from the same file."), so we only need to look at the file for the first violation to look up the baseline violations for the file.

Source/SwiftLintCore/Models/Baseline.swift
```diff
@@ -0,0 +1,169 @@
...
+                text = ""
+            }
+            baselineViolations.append(BaselineViolation(violation: violation, text: text))
+        }
+        return baselineViolations
+    }
+}
+
+private extension Sequence where Element == BaselineViolation {
+    var violationsWithAbsolutePaths: [StyleViolation] {
```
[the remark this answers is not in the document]
mildm8nnered: I've moved the inner logic to an extension on StyleViolation, so we're just calling `map { $0.violation.withAbsolutePath }` here now.

Source/swiftlint/Commands/ReportBaseline.swift
```diff
@@ -0,0 +1,37 @@
+import ArgumentParser
+import Foundation
+import SwiftLintFramework
+
+extension SwiftLint {
+    struct ReportBaseline: ParsableCommand {
```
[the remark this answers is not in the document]
mildm8nnered: So something like `swiftlint baseline --report`?

mildm8nnered:
> Very good idea to store and compare the violating line for more context. To increase precision even further, could this theoretically be extended to the violating line itself plus the previous and the subsequent line? Not sure what the benefit of this would be though ...

So we could definitely store more context, and that would definitely improve accuracy in some situations.

But having more context would mean that a change in the preceding or following line would then show up as "new", at least on the basis of exact equality. The same is true for the line the violation is on of course - someone could change whitespace there, but still leave the violation, and that would fail our equality test. We could try to filter intelligently for preceding and following line changes, but I think the line the violation is on is going to work well for most cases (one can always construct a pathological set of violations that will fool whichever algorithm you use).

If this does turn out to be a problem in practice, we can always update it later.

Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
```diff
@@ -52,6 +55,23 @@ struct LintOrAnalyzeCommand {
     private static func collectViolations(builder: LintOrAnalyzeResultBuilder) async throws -> [SwiftLintFile] {
         let options = builder.options
         let visitorMutationQueue = DispatchQueue(label: "io.realm.swiftlint.lintVisitorMutation")
+        let baseline: Baseline?
+        if let baselinePath = options.baseline {
+            do {
+                baseline = try Baseline(fromPath: baselinePath)
+            } catch {
+                Issue.genericWarning(
```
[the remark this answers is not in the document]
mildm8nnered: I went with

```
    /// The baseline file at `path` is not readable or cannot be opened.
    case baselineNotReadable(path: String)
```

Source/SwiftLintCore/Models/Baseline.swift
```diff
@@ -0,0 +1,169 @@
...
+            }
+        }
+
+        let violationsWithAbsolutePaths = Set(filteredViolations.violationsWithAbsolutePaths)
+        return violations.filter { violationsWithAbsolutePaths.contains($0) }
+    }
+}
+
+private extension Sequence where Element == StyleViolation {
+    var baselineViolations: [BaselineViolation] {
```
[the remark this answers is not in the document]
mildm8nnered: So I've refactored this a little bit, but in this case, as we're processing the array, we're building up a cache of the lines of each file in `var lines: [String: [String]] = [:]`, so we can't just map from one sequence to the next.

Tests/SwiftLintFrameworkTests/BaselineTests.swift
mildm8nnered: So it will pass without it, but unless the sample code is on disk, `Baseline(violations: Self.violations)` will be unable to read the file to obtain the lines of text - it will go with `""` in these cases, but `File.swift` in SourceKittenFramework will print to stderr:

```
Could not read contents of `/private/var/folders/7h/f16htcls4l391mnfhbw21628y1r4t3/T/B5A0ACEE-62F4-40B6-B22E-96E7BCF2D75F.swift`
Could not read contents of `/private/var/folders/7h/f16htcls4l391mnfhbw21628y1r4t3/T/B5A0ACEE-62F4-40B6-B22E-96E7BCF2D75F.swift`
Could not read contents of `/private/var/folders/7h/f16htcls4l391mnfhbw21628y1r4t3/T/B5A0ACEE-62F4-40B6-B22E-96E7BCF2D75F.swift`
Could not read contents of `/private/var/folders/7h/f16htcls4l391mnfhbw21628y1r4t3/T/B5A0ACEE-62F4-40B6-B22E-96E7BCF2D75F.swift
```

Tests/SwiftLintFrameworkTests/BaselineTests.swift
mildm8nnered: So I've moved the directory changing methods into `setUpWithError` and `tearDownWithError`.

For the test file, which we need to create on disk before each test, setUp would almost have worked, but in `testViolationDetection` we need to perform that setup before testing each rule in each position.

Source/SwiftLintCore/Models/Baseline.swift
```diff
@@ -0,0 +1,169 @@
...
+    }
+}
+
+private extension Sequence where Element == BaselineViolation {
+    var violationsWithAbsolutePaths: [StyleViolation] {
+        map {
+            let location = $0.violation.location
+            let absolutePath: String?
+            if let relativePath = location.file {
+                absolutePath = FileManager.default.currentDirectoryPath + "/" + relativePath
```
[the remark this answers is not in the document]
mildm8nnered: So that is true - we also suffer from that problem in some of the reporters (see screenshot below). I think `relativeFile` is going to be important in a group setting, where the baseline is likely generated on a different machine from the developers.

I guess we could have an option to use absolutePaths instead (Periphery, in general, seems to use absolute paths, but has an option to use relative paths) if you think we need it.

Screen Shot 2024-04-13 at 19 58 27

Source/SwiftLintCore/Models/Baseline.swift
```diff
@@ -0,0 +1,169 @@
...
+    }
+}
+
+/// A set of violations that can be used to filter newly detected violations.
+public struct Baseline: Equatable {
+    private let baselineViolations: GroupedViolations
+
+    /// The stored violations.
+    public var violations: [StyleViolation] {
+        baselineViolations.keys.sorted().flatMap({ baselineViolations[$0]! }).violationsWithAbsolutePaths
```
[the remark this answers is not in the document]
mildm8nnered: I went with

```
baselineViolations.map { ($0, $1) }.sorted { $0.0 < $1.0 }.flatMap { $0.1 }.violationsWithAbsolutePaths
```

mildm8nnered:
I swapped saving the dictionaries of baseline violations, keyed on file, to saving a sorted array of baseline violations  - https://github.com/realm/SwiftLint/pull/5475/commits/2834540c33ba3be7b178e6baaf038073ec0d7cc2

So exactly identical baselines should also now be identical on disk.

Source/swiftlint/Commands/ReportBaseline.swift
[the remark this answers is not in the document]
mildm8nnered: I went with:

`swiftlint baseline [<action>] --baseline <baseline> [--reporter <reporter>] [--output <output>]`

Where `report` is the default (and only) action

Tests/SwiftLintFrameworkTests/BaselineTests.swift
```diff
@@ -0,0 +1,229 @@
...
+                                 """
+
+    private static let violations: [StyleViolation] = {
+        [
+            ArrayInitRule.description,
+            BlockBasedKVORule.description,
+            ClosingBraceRule.description,
+            DirectReturnRule.description
+        ].violations
+    }()
```
SimplyDanny, line 54:
```suggestion
    private static let violations = [
            ArrayInitRule.description,
            BlockBasedKVORule.description,
            ClosingBraceRule.description,
            DirectReturnRule.description
        ].violations
```

Tests/SwiftLintFrameworkTests/BaselineTests.swift
```diff
@@ -0,0 +1,229 @@
...
+
+    private static let baseline: Baseline = {
+        Baseline(violations: violations)
+    }()
```
SimplyDanny, line 58:
```suggestion
    private static let baseline = Baseline(violations: violations)
```

mildm8nnered:
I've added a `compare` action to the `baseline` command, that will print violations that occur in another baseline, but not in the original.

Source/SwiftLintCore/Models/Baseline.swift
```diff
@@ -0,0 +1,186 @@
...
+        var baselineViolations: [BaselineViolation] = []
+        for violation in self {
+            guard let absolutePath = violation.location.file,
+                  let lineNumber = violation.location.line != nil ? (violation.location.line ?? 0) - 1 : nil,
+                  lineNumber > 0 else {
+                baselineViolations.append(violation.baselineViolation())
+                continue
+            }
+            if let fileLines = lines[absolutePath] {
+                let text = (!fileLines.isEmpty && lineNumber < fileLines.count ) ? fileLines[lineNumber] : ""
```
SimplyDanny, line 135:
```suggestion
                let text = (fileLines.isNotEmpty && lineNumber < fileLines.count ) ? fileLines[lineNumber] : ""
```

Source/SwiftLintCore/Models/Baseline.swift
```diff
@@ -0,0 +1,186 @@
...
+                baselineViolations.append(violation.baselineViolation(text: text))
+                continue
+            }
+            let text: String
+            if let fileLines = SwiftLintFile(path: absolutePath)?.lines.map(\.content),
+               lineNumber < fileLines.count {
+                text = fileLines[lineNumber]
+                lines[absolutePath] = fileLines
```
SimplyDanny, line 143: Shouldn't the content be cached even if `lineNumber < fileLines.count` is not fulfilled?

Source/SwiftLintCore/Models/Baseline.swift
```diff
@@ -0,0 +1,186 @@
...
+                baselineViolations.append(violation.baselineViolation())
+                continue
+            }
+            if let fileLines = lines[absolutePath] {
+                let text = (!fileLines.isEmpty && lineNumber < fileLines.count ) ? fileLines[lineNumber] : ""
+                baselineViolations.append(violation.baselineViolation(text: text))
+                continue
+            }
+            let text: String
+            if let fileLines = SwiftLintFile(path: absolutePath)?.lines.map(\.content),
```
SimplyDanny, line 140: This is potentially time-consuming and memory-hungry as it's loading a whole project in the worst case ...

We may start with this, but I suspect that, in a next step, violations might get assigned the offending line upon creation or in a modification shortly after creation depending on the baseline functionality being active or not.

Source/SwiftLintCore/Models/Baseline.swift
```diff
@@ -0,0 +1,186 @@
...
+               lineNumber < fileLines.count {
+                text = fileLines[lineNumber]
+                lines[absolutePath] = fileLines
+            } else {
+                text = ""
+            }
+            baselineViolations.append(violation.baselineViolation(text: text))
+        }
```
SimplyDanny, line 148: This whole loop is effectively all about getting the `text` to subsequently do a `baselineViolations.append(violation.baselineViolation(text: text))`. Perhaps there should be some "Text Extractor" to handle the caching internally. In fact, I would like to give things names to better understand the process.

SimplyDanny (changes requested):

Source/SwiftLintCore/Models/Baseline.swift
```diff
@@ -0,0 +1,175 @@
...
+
+    func groupedByRuleIdentifier(filteredBy existingViolations: [BaselineViolation]) -> GroupedViolations {
+        Set(self).subtracting(existingViolations).groupedByRuleIdentifier()
+    }
+}
+
+private extension StyleViolation {
+    var withAbsolutePath: StyleViolation {
+        let absolutePath: String?
+        if let relativePath = location.file {
```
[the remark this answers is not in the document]
mildm8nnered: I think that's the first one I've written

Source/SwiftLintCore/Models/Baseline.swift
mildm8nnered: So there definitely is an overhead - on my main project converting 78,785 violations (`--enable-all-rules`) in 2,197 files, c 300,000+ loc on my 16" 2021 MBP with an Apple M1 Max takes 1.5 - 2.5 seconds, so not great, but not terrible.

Just having a play around with memory now.

Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
```diff
@@ -261,6 +283,7 @@ private class LintOrAnalyzeResultBuilder {
     var fileBenchmark = Benchmark(name: "files")
     var ruleBenchmark = Benchmark(name: "rules")
     var violations = [StyleViolation]()
+    var allViolations = [StyleViolation]()
```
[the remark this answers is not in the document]
mildm8nnered: `unfilteredViolations` perhaps?

Source/SwiftLintCore/Models/Baseline.swift
mildm8nnered: So I've moved the caching into it's own LineCache, which pretty much reduces this operation to a `map`

Source/SwiftLintCore/Models/Baseline.swift
mildm8nnered: So I definitely see a spike memory wise, but its not too massive compared to the rest of our consumption (with the same dataset as above).

Screen Shot 2024-04-15 at 22 20 28

mildm8nnered:
I think I've addressed everything so far, but let me know if I've missed anything ...

Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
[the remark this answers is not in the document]
SimplyDanny: Sounds good. And some doc comments perhaps?

Source/SwiftLintCore/Models/Baseline.swift
SimplyDanny: Okay. So let's go with the current implementation and optimize afterwards if necessary.

Source/SwiftLintCore/Models/Baseline.swift
```diff
@@ -0,0 +1,184 @@
...
+    /// - Returns: The new violations.
+    public func filter(_ violations: [StyleViolation]) -> [StyleViolation] {
+        guard let firstViolation = violations.first,
+              let baselineViolations = baselineViolations[firstViolation.location.relativeFile ?? ""],
+              baselineViolations.isNotEmpty else {
+            return violations
+        }
+
+        let relativePathViolations = violations.baselineViolations
+        guard relativePathViolations != baselineViolations else {
```
SimplyDanny, line 73: Invert and use `if` to avoid negation.

Source/SwiftLintCore/Models/Baseline.swift
```diff
@@ -0,0 +1,184 @@
...
+        if let fileLines = lines[absolutePath] {
+            return (fileLines.isNotEmpty && lineNumber < fileLines.count) ? fileLines[lineNumber] : ""
+        }
+        if let fileLines = SwiftLintFile(path: absolutePath)?.lines.map(\.content) {
+            lines[absolutePath] = fileLines
+            return lineNumber < fileLines.count ? fileLines[lineNumber] : ""
+        }
+        return ""
+    }
+}
```
SimplyDanny, line 150: This ought to serve the same purpose but with caching and text extraction not mixed up in a single method:
```suggestion
private struct LineCache {
    private var lines: [String: [String]] = [:]

    mutating func text(at location: Location) -> String {
        let line = (location.line ?? 0) - 1
        if line > 0, let file = location.file, let content = cached(file: file), line < content.count {
            return content[line]
        }
        return ""
    }

    private mutating func cached(file: String) -> [String]? {
        if let fileLines = lines[file] {
            return fileLines
        }
        if let fileLines = SwiftLintFile(path: file)?.lines.map(\.content) {
            lines[file] = fileLines
            return fileLines
        }
        return nil
    }
}
```

Source/SwiftLintCore/Models/Baseline.swift
```diff
@@ -0,0 +1,184 @@
...
+
+private extension StyleViolation {
+    var withAbsolutePath: StyleViolation {
+        let absolutePath: String? = if let relativePath = location.file {
+            FileManager.default.currentDirectoryPath + "/" + relativePath
+        } else {
+            nil
+        }
```
SimplyDanny, line 176: Always unclear how to format this. Better?
```suggestion
        let absolutePath: String? =
            if let relativePath = location.file {
                FileManager.default.currentDirectoryPath + "/" + relativePath
            } else {
                nil
            }
```

Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
```diff
@@ -115,6 +121,21 @@ struct LintOrAnalyzeCommand {
...
+    private static func baseline(_ options: LintOrAnalyzeOptions) throws -> Baseline? {
+        if let baselinePath = options.baseline {
+            do {
+                return try Baseline(fromPath: baselinePath)
+            } catch {
+                Issue.baselineNotReadable(path: baselinePath).print()
+                if options.writeBaseline == options.baseline {
+                    return nil
+                }
+                throw error
```
SimplyDanny, line 133: Does it work if we throw `Issue.baselineNotReadable(path: baselinePath)` here instead of printing it dedicatedly?

Source/swiftlint/Commands/Baseline.swift
```diff
@@ -0,0 +1,92 @@
...
+            switch action {
+            case .report:
+                try report()
+            case .compare:
+                try compare()
+            }
+            ExitHelper.successfullyExit()
+        }
+
+        func validate() throws {
```
SimplyDanny, line 48: You could make `Action` two sub-commands to explicitly be able to specify which arguments are allowed in either case.

SimplyDanny:
> I think I've addressed everything so far, but let me know if I've missed anything ...

I kept the open points as "unresolved" and added a few more comments.

Source/SwiftLintCore/Models/Baseline.swift
mildm8nnered: yep, I think that makes it clearer

Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
mildm8nnered: So it would work, but by throwing the error we get more detailed error messaging as well.

```
$ swiftlint --baseline BBB.json
warning: Cannot open or read the baseline file at path 'BBB.json'.
Error: The file “BBB.json” couldn’t be opened because there is no such file.
$ touch BBB.json
$ chmod -r  BBB.json
$
$ swiftlint --baseline BBB.json
warning: Cannot open or read the baseline file at path 'BBB.json'.
Error: The file “BBB.json” couldn’t be opened because you don’t have permission to view it.
```

Also, in the special case where the user is writing to and reading from the same baseline, we print the warning, but proceed anyway, as the next time they invoke the command it will work.

Tests/SwiftLintFrameworkTests/BaselineTests.swift
```diff
@@ -0,0 +1,225 @@
...
+#else
+    return result
+#endif
+}
+
+private var sourceFilePath: String = {
+    temporaryFilePath + ".swift"
+}()
+
+final class BaselineTests: XCTestCase {
```
mildm8nnered, line 22: Should this be `SwiftLintTestCase`? - I couldn't work out why most tests are, but some tests aren't.

Source/swiftlint/Commands/Baseline.swift
mildm8nnered: nice - I will give that a go.

mildm8nnered: Ah - that does seem much nicer

Tests/SwiftLintFrameworkTests/BaselineTests.swift
SimplyDanny: `SwiftLintTestCase` only make sure all rules are registered. As violations are created manually in your tests, registering rules don't seem to be needed. It's a convenience not a requirement.

Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
SimplyDanny: Okay, got that. But isn't this also happening due to the `throw error` (when `nil` isn't returned)?

Source/swiftlint/Commands/Baseline.swift
```diff
@@ -0,0 +1,89 @@
+import ArgumentParser
+import Foundation
+import SwiftLintFramework
+
+extension SwiftLint {
+    struct Baseline: ParsableCommand {
+        static let configuration = CommandConfiguration(
+            abstract: "Reports the violations in a baseline or the violations " +
+                      "that are present in another baseline, but not in the original.",
```
SimplyDanny, line 9: The existing abstracts for other commands are quite short (and come without a final dot). For a the sake of consistency, let's do it likewise here:
```suggestion
            abstract: "Operations on existing baselines",
```

Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
mildm8nnered: So everyone will see

`warning: Cannot open or read the baseline file at path 'BBB.json'.`

if the input and output baseline are different filenames, then `throw error` will happen and SwiftLint will quit with the more detailed error message. e.g.

`Error: The file “BBB.json” couldn’t be opened because there is no such file.`

So people who just get the warning do not get the detail. This is probably not going to happen much, but I guess really we should only be ignoring that if the file doesn't exist. If the file is not writable, the user probably does care.

mildm8nnered: I've rejigged this slightly, so we are now explicitly checking for `CocoaError.fileReadNoSuchFile` in the warning only case.

aiKrice:
Very eager to see it :)

Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
```diff
@@ -115,6 +121,23 @@ struct LintOrAnalyzeCommand {
...
+                return try Baseline(fromPath: baselinePath)
+            } catch CocoaError.fileReadNoSuchFile {
+                Issue.baselineNotReadable(path: baselinePath).print()
+                if options.writeBaseline != options.baseline {
+                    throw error
+                }
+            } catch {
+                Issue.baselineNotReadable(path: baselinePath).print()
+                throw error
+            }
```
SimplyDanny, line 136:
```suggestion
            } catch {
                Issue.baselineNotReadable(path: baselinePath).print()
                if (error as? CocoaError)?.code != CocoaError.fileReadNoSuchFile || options.writeBaseline != options.baseline {
                    throw error
                }
            }
```

Tests/SwiftLintFrameworkTests/BaselineTests.swift
```diff
@@ -0,0 +1,225 @@
...
+            let violation = StyleViolation(
+                ruleDescription: newViolationRuleDescription,
+                location: Location(file: sourceFilePath, line: line, character: 1)
+            )
+            newViolations.insert(violation, at: insertionIndex)
+            XCTAssertEqual(baseline.filter(newViolations), [violation])
+        }
+    }
+
+    private func testBlock(_ block: () throws -> Void) throws {
```
SimplyDanny, line 177: Ideally, the path to the example path is something local. So I propose to manage file creation and cleanup only in here and pass the path to the caller: 
```suggestion
    private func withExampleFileCreated(_ block: (filePath: String) throws -> Void) throws {
```

Tests/SwiftLintFrameworkTests/BaselineTests.swift
```diff
@@ -0,0 +1,225 @@
+@testable import SwiftLintBuiltInRules
+@testable import SwiftLintCore
+import XCTest
+
+private var temporaryFilePath: String {
```
SimplyDanny, line 5: Let this only be a (fixed?) folder
```suggestion
private var temporaryFolderPath: String {
```
and take care of changing the active directory there in `setUp`; switch back in `tearDown`. Have `withExampleFileCreated` create the example and manage its removal.

mildm8nnered:
Is there anything left here that you'd like me to address @SimplyDanny ?

Tests/SwiftLintFrameworkTests/BaselineTests.swift
```diff
@@ -0,0 +1,222 @@
...
+
+        for ruleDescription in ruleDescriptions {
+            for insertionIndex in 0..<violationRuleDescriptions.count {
+                try testViolationDetection(
+                    violationRuleDescriptions: violationRuleDescriptions,
+                    newViolationRuleDescription: ruleDescription,
+                    insertionIndex: insertionIndex
+                )
+            }
+            break
```
SimplyDanny, line 137: The `break` is unintended here, isn't it?

Tests/SwiftLintFrameworkTests/BaselineTests.swift
```diff
@@ -0,0 +1,222 @@
...
+            )
+            newViolations.insert(violation, at: insertionIndex)
+            XCTAssertEqual(baseline.filter(newViolations), [violation])
+        }
+    }
+
+    private func withExampleFileCreated(_ block: (String) throws -> Void) throws {
+        let sourceFilePath = temporaryDirectoryPath.stringByAppendingPathComponent("\(UUID().uuidString).swift")
+        guard let data = Self.example.data(using: .utf8) else {
+            XCTFail("Could not convert example code to data using utf8 encoding")
```
SimplyDanny, line 180:
```suggestion
            XCTFail("Could not convert example code to data using UTF-8 encoding")
```

SimplyDanny (approved):
Have you thought about [this comment](https://github.com/realm/SwiftLint/pull/5475#discussion_r1564905460) yet?

Tests/SwiftLintFrameworkTests/BaselineTests.swift
mildm8nnered: let me check that ...

mildm8nnered: Ah yes 🤦 - it was left over from testing

mildm8nnered:
> Have you thought about [this comment](https://github.com/realm/SwiftLint/pull/5475#discussion_r1564905460) yet?

will reply on the original comment ...

Source/SwiftLintCore/Models/Baseline.swift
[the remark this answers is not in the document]
mildm8nnered: Could you clarify that a little bit @SimplyDanny - I wasn't sure exactly what you were getting at, or what that would look like in code.

I think it's clean enough now (especially after your fixes for the line caching) that it should be easy to rejig it slightly.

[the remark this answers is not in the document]
mildm8nnered: One thing I do like about the current calling convention is that it's really clean at the call-sites:

`baselineViolations = violations.baselineViolations.groupedByFile()`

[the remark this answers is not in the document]
SimplyDanny: What I don't like so much about this, is that the innocent looking "property access" `violations.baselineViolations` is a quite expensive operation. I'd like to make this a bit more obvious by using some kind of type conversion using an initializer call instead and perhaps even having an own type for a set of `BaselineViolation`s.

A combination of both approaches could be:

```swift
private typealias BaselineGroup = [BaselineViolation]

private extension Sequence where Element == BaselineViolation {
    init(_ violations: [StyleViolation]) where Self == BaselineGroup {
        var lineCache = LineCache()
        self = violations.map { $0.baselineViolation(text: lineCache.text(at: $0.location)) }
    }

    var violationsWithAbsolutePaths: [StyleViolation] {
        map { $0.violation.withAbsolutePath }
    }

    func groupedByFile() -> ViolationsPerFile {
        Dictionary(grouping: self) { $0.violation.location.relativeFile ?? "" }
    }

    func groupedByRuleIdentifier(filteredBy existingViolations: [BaselineViolation] = []) -> ViolationsPerRule {
        Dictionary(grouping: Set(self).subtracting(existingViolations), by: \.violation.ruleIdentifier)
    }
}
```

It's not a separate type requiring some boilerplate to implement but creating an instance of it needs an initializer call like `BaselineGroup(styleViolations)`. I'm not yet fully happy with it, but at least it's making things more explicit.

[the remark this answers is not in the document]
SimplyDanny: Also not that the two methods called `groupedByRuleIdentifier` can be combined to one using a default argument.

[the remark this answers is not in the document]
mildm8nnered: > What I don't like so much about this, is that the innocent looking "property access" `violations.baselineViolations` is a quite expensive operation. I'd like to make this a bit more obvious by using some kind of type conversion using an initializer call instead and perhaps even having an own type for a set of `BaselineViolation`s.
> 
> A combination of both approaches could be:
> 
> ```swift
> private typealias BaselineGroup = [BaselineViolation]
> 
> private extension Sequence where Element == BaselineViolation {
>     init(_ violations: [StyleViolation]) where Self == BaselineGroup {
>         var lineCache = LineCache()
>         self = violations.map { $0.baselineViolation(text: lineCache.text(at: $0.location)) }
>     }
> 
>     var violationsWithAbsolutePaths: [StyleViolation] {
>         map { $0.violation.withAbsolutePath }
>     }
> 
>     func groupedByFile() -> ViolationsPerFile {
>         Dictionary(grouping: self) { $0.violation.location.relativeFile ?? "" }
>     }
> 
>     func groupedByRuleIdentifier(filteredBy existingViolations: [BaselineViolation] = []) -> ViolationsPerRule {
>         Dictionary(grouping: Set(self).subtracting(existingViolations), by: \.violation.ruleIdentifier)
>     }
> }
> ```
> 
> It's not a separate type requiring some boilerplate to implement but creating an instance of it needs an initializer call like `BaselineGroup(styleViolations)`. I'm not yet fully happy with it, but at least it's making things more explicit.

So I went with pretty much exactly this in https://github.com/realm/SwiftLint/pull/5475/commits/e48485a65a92549881df299c602bbc03cd67ff7c  - I'd tried to play around with something similar last night, but couldn't quite get the syntax right.

I like the initializer, and it doesn't add too much at the call sites - still one liners.

Not super-happy with `BaselineGroup` - I tried `BaselineViolations`, but then there's an ivar called `baselineViolations` with a different type.

lmk what you think.

lmk what you think

SimplyDanny:
This is a very good implementation to start with and collect some feedback for. Having it as an incubation allows us to further enhance it if that's required.

Thanks a lot, @mildm8nnered, for picking up and refining previous PRs!

mildm8nnered:
> This is a very good implementation to start with and collect some feedback for. Having it as an incubation allows us to further enhance it if that's required.
> 
> Thanks a lot, @mildm8nnered, for picking up and refining previous PRs!

Thank you for all your help and feedback @SimplyDanny 

We may be able to close these old baseline PRs now:

#3421 and #4908 ...

aiKrice:
Hello @mildm8nnered , First of all great job.
I would like to be sure how should I use the baseline. Naively, I try in the same "spirit" than Detekt (for kotlint).

So here is my commands:
1. Generate a baseline with `swiftlint --write-baseline config/swiftlint/baseline.json`
2. Apply it: swiftlint --silence-deprecation-warnings --strict --baseline config/swiftlint/baseline.json --config config/swiftlint/swiftlint.yml  myswiftfile_with_violations_baselined.swift`

But I still have the violations output. Could you tell me what I am wrong ?

mildm8nnered:
> Hello @mildm8nnered , First of all great job. I would like to be sure how should I use the baseline. Naively, I try in the same "spirit" than Detekt (for kotlin).
> 
> So here is my commands:
> 
> 1. Generate a baseline with `swiftlint --write-baseline config/swiftlint/baseline.json`
> 2. Apply it: `swiftlint --silence-deprecation-warnings --strict --baseline config/swiftlint/baseline.json --config config/swiftlint/swiftlint.yml  myswiftfile_with_violations_baselined.swift`
> 
> But I still have the violations output. Could you tell me what I am wrong ?

So you need to use the same config file to generate the baseline that you will use when using it to filter violations.

I would have expected your commands to be:

`swiftlint --write-baseline config/swiftlint/baseline.json --config config/swiftlint/swiftlint.yml myswiftfile_with_violations_baselined.swift`

and

`swiftlint --silence-deprecation-warnings --strict --baseline config/swiftlint/baseline.json --config config/swiftlint/swiftlint.yml  myswiftfile_with_violations_baselined.swift`

`--strict` should not make a difference - it will mean that the baseline violations and detected violations will not be exact matches (because severity will differ), but when we filter non-exact matches, we ignore severity (and line number).

Please let us know if this does not fix it for you.

PR #5478, opened 2024-02-29
Allow rules to lint empty files

marcelofabri:
https://github.com/realm/SwiftLint/pull/3885 made it so no violations ever trigger when a file is empty. 

While this makes sense in general, we'd like to introduce a custom rule in our codebase to avoid empty files. But that doesn't work as expected because the rule isn't even called.

The simplest thing I could think of is to introduce `shouldLintEmptyFiles` on `Rule` (defaulting it to `false`) and using it when deciding whether to lint a file or not.

### DIFF
--- a/Source/SwiftLintCore/Models/Linter.swift
+++ b/Source/SwiftLintCore/Models/Linter.swift
@@ -62,8 +62,13 @@ private extension Rule {
               configuration: Configuration,
               superfluousDisableCommandRule: SuperfluousDisableCommandRule?,
               compilerArguments: [String]) -> LintResult? {
+        // We shouldn't lint if the current Swift version is not supported by the rule
+        guard SwiftVersion.current >= Self.description.minSwiftVersion else {
+            return nil
+        }
+
         // Empty files shouldn't trigger violations
-        guard !file.isEmpty, SwiftVersion.current >= Self.description.minSwiftVersion else {
+        guard !shouldLintEmptyFiles, !file.isEmpty else {
             return nil
         }
 
@@ -209,7 +214,12 @@ public struct CollectedLinter {
 
     private func getStyleViolations(using storage: RuleStorage,
                                     benchmark: Bool = false) -> ([StyleViolation], [(id: String, time: Double)]) {
-        guard !file.isEmpty else {
+        var rules = rules
+        if file.isEmpty {
+            rules = rules.filter(\.shouldLintEmptyFiles)
+        }
+
+        guard !rules.isEmpty else {
             // Empty files shouldn't trigger violations
             return ([], [])
         }
--- a/Source/SwiftLintCore/Models/Linter.swift
+++ b/Source/SwiftLintCore/Models/Linter.swift
@@ -67,8 +67,8 @@ private extension Rule {
             return nil
         }
 
-        // Empty files shouldn't trigger violations
-        guard !shouldLintEmptyFiles, !file.isEmpty else {
+        // Empty files shouldn't trigger violations if `shouldLintEmptyFiles` is `false`
+        if file.isEmpty, !shouldLintEmptyFiles {
             return nil
         }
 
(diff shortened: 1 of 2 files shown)


Source/SwiftLintCore/Models/Linter.swift
```diff
@@ -209,7 +214,12 @@ public struct CollectedLinter {
 
     private func getStyleViolations(using storage: RuleStorage,
                                     benchmark: Bool = false) -> ([StyleViolation], [(id: String, time: Double)]) {
-        guard !file.isEmpty else {
+        var rules = rules
+        if file.isEmpty {
+            rules = rules.filter(\.shouldLintEmptyFiles)
+        }
```
SimplyDanny, line 220: This could be done in the initializer. The file and the rules list exist there already.

Also, this could save one copy by using `rules = file.isEmpty ? linter.rules.filter(\.shouldLintEmptyFiles) : linter.rules`.

Source/SwiftLintCore/Models/Linter.swift
```diff
@@ -62,8 +62,13 @@ private extension Rule {
...
               superfluousDisableCommandRule: SuperfluousDisableCommandRule?,
               compilerArguments: [String]) -> LintResult? {
+        // We shouldn't lint if the current Swift version is not supported by the rule
+        guard SwiftVersion.current >= Self.description.minSwiftVersion else {
+            return nil
+        }
+
         // Empty files shouldn't trigger violations
-        guard !file.isEmpty, SwiftVersion.current >= Self.description.minSwiftVersion else {
+        guard !shouldLintEmptyFiles, !file.isEmpty else {
```
SimplyDanny, line 71: Does `shouldLintEmptyFiles` mean "lint **only** empty files"? If so, the name should make that clear and this condition should use XOR, shouldn't it? So I think then it should be:
```suggestion
        if file.isEmpty XOR !shouldLintEmptyFiles {
```

SimplyDanny (commented):
I have a few questions.

If a custom rule is the goal, this should also be an option in the definition of a custom rule. Or do you mean native custom rules?

marcelofabri:
> I have a few questions.
> 
> If a custom rule is the goal, this should also be an option in the definition of a custom rule. Or do you mean native custom rules?

Native custom rules!

Source/SwiftLintCore/Models/Linter.swift
marcelofabri: sure, I can make that change -- I was mostly following the existing logic

Source/SwiftLintCore/Models/Linter.swift
marcelofabri: > Does shouldLintEmptyFiles mean "lint only empty files"?

I don't think so. For example, in my case I'd like to also lint against files that only have comments (so they're not "empty" per the definition in `file.isEmpty`)

Source/SwiftLintCore/Models/Linter.swift
```diff
@@ -62,8 +62,13 @@ private extension Rule {
               configuration: Configuration,
               superfluousDisableCommandRule: SuperfluousDisableCommandRule?,
               compilerArguments: [String]) -> LintResult? {
+        // We shouldn't lint if the current Swift version is not supported by the rule
+        guard SwiftVersion.current >= Self.description.minSwiftVersion else {
+            return nil
+        }
+
         // Empty files shouldn't trigger violations
```
SimplyDanny, line 70: Comment needs an update now.

Source/SwiftLintCore/Models/Linter.swift
SimplyDanny, line 71: This would be easier to grasp with an `if` statement:
```suggestion
        if shouldLintEmptyFiles || file.isEmpty {
```

But then, apparently, nothing is linted when the rule "should lint empty files". That doesn't seem to be correct.

SimplyDanny (approved):

PR #5479, opened 2024-03-03
Update SwiftSyntax to version 510.0.0

SimplyDanny:


jpsim:
The 9-13% perf hit is unfortunate. Would be good to share those numbers with the SwiftSyntax folks.

jpsim (approved):

SimplyDanny:
While the improvements in memory management in https://github.com/apple/swift-syntax/pull/2537 look very promising, [the decision](https://github.com/apple/swift-syntax/issues/2155#issuecomment-1986677731) is against adding them to any Swift 510.x.y release. We might only see it in 600.0.0 earliest.

So what to do? Skip 510.x.y entirely and wait for 600.x.y or accept a performance drop of ~10%?

jpsim:
We should stay on the latest SwiftSyntax release because it’ll parse Swift 5.10 code more accurately. Higher accuracy is worth a 10% hit in perf in the uncached case.

Issue #5471, opened 2024-02-23
Rule Request: Prefer `static` over `final class`


SimplyDanny:
The modifiers `final` and `class` on a method/property have the same meaning as `static`. The rule should flag appearances of `final class` on the same declaration and replace them by `static`.

I don't think there is any need for configuration. The rule could be enabled by default, but this can be decided in review when browsing through the reported OSS findings.

Triggering:

```swift
class C {
    final class func f() {}
}

final class C {
    class func f() {}
}
```

Non-triggering:

```swift
class C {
    static func f() {}
}

class C {
    class func f() {}
}

final class C {}
```

omelics:
The rule conflicts with nested classes. For example:

```swift
extension MyView {
    final class SelectionView: UIView { }
}
```

This obviously cannot be replaced with the `static` keyword and will lead to a compilation error.

SimplyDanny:
> The rule conflicts with nested classes. For example:
> 
> ```swift
> extension A {
>     final class B { }
> }
> ```
> 
> This obviously cannot be replaced with the `static` keyword and will lead to a compilation error.

Which part is replaced by the rule in this example?

omelics:
> > The rule conflicts with nested classes. For example:
> > ```swift
> > extension A {
> >     final class B { }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This obviously cannot be replaced with the `static` keyword and will lead to a compilation error.
> 
> Which part is replaced by the rule in this example?

Sorry, the previous example was incorrect.

The following code is triggered by the rule, while it should not, I believe:

```swift
final class B: UIView {
    override class var layerClass: AnyClass {
        return CALayer.self
    }
}
```

Seems like it happens because of combination of a `final class` and a `class` property override. Swift do allow replacing `class` keyword with `static` in this scenario, but Xcode suggests using `override class var` by default.

SimplyDanny:
> Seems like it happens because of combination of a `final class` and a `class` property override. Swift do allow replacing `class` keyword with `static` in this scenario, but Xcode suggests using `override class var` by default.

Just checked with

```swift
class B {
    class var a: Int { 1 }
}

final class A: B {
    override static var a: Int { 2 }
}
```

which compiles fine in Swift 5.10.

omelics:
[...]

PR #5487, opened 2024-03-06
Add new `static_over_final_class` rule

phlippieb:
Implements rule requested in https://github.com/realm/SwiftLint/issues/5471


Source/SwiftLintBuiltInRules/Rules/Idiomatic/FinalClassRule.swift
```diff
@@ -0,0 +1,64 @@
...
+               node.modifiers.contains(keyword: .class) {
+                violations.append(node.positionAfterSkippingLeadingTrivia)
+            }
+
+            return .visitChildren
+        }
+
+        override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
+            if node.modifiers.contains(keyword: .final) {
+                node.memberBlock.members
```
SimplyDanny, line 55: Better to maintain a stack of booleans that can be looked up to check whether the current scope is a final class. There are some examples of rules using a stack. Look for the `Stack` type and its references in rules.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FinalClassRule.swift
```diff
@@ -0,0 +1,64 @@
...
+                class func f() {}
+            }
+            """)
+        ]
+    )
+}
+
+private extension FinalClassRule {
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        override func visit(_ node: FunctionDeclSyntax) -> SyntaxVisitorContinueKind {
```
SimplyDanny, line 44: Note that this is not only about functions but also variables and subscripts.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FinalClassRule.swift
```diff
@@ -0,0 +1,64 @@
+import SwiftSyntax
+
+@SwiftSyntaxRule
+struct FinalClassRule: Rule {
```
SimplyDanny, line 4: Perhaps extends the name to `StaticOverFinalClass` and the identifier to `static_over_final_class` accordingly.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FinalClassRule.swift
```diff
@@ -0,0 +1,64 @@
...
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "final_class",
+        name: "Final Class",
+        description: "Prefer `static` over `final class`",
```
SimplyDanny, line 10:
```suggestion
        description: "Prefer `static` over `final class` to declare as non-overridable",
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FinalClassRule.swift
```diff
@@ -0,0 +1,64 @@
...
+            class C {
+                final class func f() {}
+            }
+            """),
+            Example("""
+            final class C {
+                class func f() {}
+            }
+            """)
+        ]
```
SimplyDanny, line 38: Please also consider nested classes:
```swift
final class C1 {
    class func f() {}      // Triggers
    class C2 {
        class func f() {}  // Does not triggers
    }
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FinalClassRule.swift
```diff
@@ -0,0 +1,64 @@
...
+            }
+            """),
+            Example("""
+            final class C {}
+            """),
+        ],
+        triggeringExamples: [
+            Example("""
+            class C {
+                final class func f() {}
```
SimplyDanny, line 30: Please add violation markers `↓`:
```suggestion
                ↓final class func f() {}
```

CHANGELOG.md
```diff
@@ -125,6 +125,10 @@
 * Make `trailing_closure` correctable.  
   [KS1019](https://github.com/KS1019/)
 
+* Add new `final_class` rule to prefer `static` over `final class` methods.
```
SimplyDanny, line 128: Please append two spaces to the end of the description. They are required to enforce a line break in the rendered Markdown:
```suggestion
* Add new `final_class` rule to prefer `static` over `final class` declarations.  
```

SimplyDanny (changes requested):
Thank you for the contribution. A few remarks ...

SimplyDanny:
Run `IntegrationTests` locally to get rid of the build failures. Also exclude the new rule from the `.swiftlint.yml` file.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FinalClassRule.swift
phlippieb: Ok cool, I have loosely followed an example I found in NestingRule.

A question about this approach though: to me it seems duplicative to build up a new stack (however minimal) to track information that can be derived directly from the existing syntax tree. The duplication opens the door to errors, such as forgetting to pop the stack after visiting a node, meaning that the stack could accidentally become out of sync with the actual syntax tree. Why do you prefer to build a new stack? I'm sure there's a good reason, so just asking to learn :)

phlippieb:
@SimplyDanny thanks for the review! I have pushed changes to address all comments, and also noted a [question here](https://github.com/realm/SwiftLint/pull/5487#discussion_r1515959154) if you have time to explain your approach to me. 

If you are happy with the PR, I will squash the commits and update the PR to clean up the history.

CHANGELOG.md
```diff
@@ -125,6 +125,10 @@
 * Make `trailing_closure` correctable.  
   [KS1019](https://github.com/KS1019/)
 
+* Add new `static_over_final_class` rule to prefer `static` over `final class` declerations.
```
SimplyDanny, line 128:
```suggestion
* Add new `static_over_final_class` rule to prefer `static` over
  `final class` declaration.  
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/StaticOverFinalClassRule.swift
```diff
@@ -0,0 +1,136 @@
...
+        private func validateNode(at position: AbsolutePosition, with modifiers: DeclModifierListSyntax) {
+            if modifiers.contains(keyword: .final),
+               modifiers.contains(keyword: .class) {
+                violations.append(position)
+            } else if modifiers.contains(keyword: .class),
+                      levels.lastIsFinalClass {
+                violations.append(position)
+            }
+
+            levels.push(false)
```
SimplyDanny, line 133: The stack is only needed to track classes. The constant `false` being added already indicates that this is a redundant operation.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/StaticOverFinalClassRule.swift
```diff
@@ -0,0 +1,136 @@
...
+// Stack of flags indicating whether each level in the tree is a final class
+private typealias LevelIsFinalClassStack = Stack<Bool>
+
+private extension LevelIsFinalClassStack {
+    var lastIsFinalClass: Bool { peek() == true }
+}
+
+private extension StaticOverFinalClassRule {
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        private var levels = LevelIsFinalClassStack()
```
SimplyDanny, line 85: I think this can be simplified:
```suggestion
        private var classContexts = Stack<Bool>
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/StaticOverFinalClassRule.swift
```diff
@@ -0,0 +1,136 @@
...
+            """)
+        ]
+    )
+}
+
+// Stack of flags indicating whether each level in the tree is a final class
+private typealias LevelIsFinalClassStack = Stack<Bool>
+
+private extension LevelIsFinalClassStack {
+    var lastIsFinalClass: Bool { peek() == true }
```
SimplyDanny, line 80:
```suggestion
    var lastIsFinal: Bool { peek() == true }
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/StaticOverFinalClassRule.swift
```diff
@@ -0,0 +1,136 @@
...
+            levels.push(node.modifiers.contains(keyword: .final))
+            return .visitChildren
+        }
+
+        override func visitPost(_ node: ClassDeclSyntax) {
+            levels.pop()
+        }
+
+        override func visit(_ node: FunctionDeclSyntax) -> SyntaxVisitorContinueKind {
+            validateNode(at: node.positionAfterSkippingLeadingTrivia, with: node.modifiers)
```
SimplyDanny, line 97: Do this in the `visitPost`s and remove the `visit` overrides.

SimplyDanny (changes requested):

SimplyDanny:
> If you are happy with the contents of the PR, I will squash the commits and update the PR to clean up the history.

Manual squashing is not necessary. That's done by GitHub when merging.

SimplyDanny (approved):

PR #5494, opened 2024-03-12
Explicit types for build time optimization

janborowski-paramount:
Hello everyone!

_This PR intended to fix compilation time issue with DeclaredIdentifiersTrackingVisitor.swift but it seems that this problem has been resolved [here](https://github.com/realm/SwiftLint/commit/5bfe329d09c6816fcc490899c38a83a55aafa2c0#diff-90801f3c759de18bb6ddbae9401e05a5f4c603c92a114bec2ca477d2208cf4e1)._

This pull request offers small improvement to SwiftPMCompilationDB via introducing explicit types. 
Before optimizations:
```
251.79ms    Source/swiftlint/Helpers/SwiftPMCompilationDB.swift:29:17	static method parse(yaml:)
```

After optimizations
```
29.33ms     Source/swiftlint/Helpers/SwiftPMCompilationDB.swift:29:17	static method parse(yaml:)
```

Please let me know what you think about this change.  I'm uncertain about how to conduct thorough testing. I've executed automated tests locally, and all have passed. I'm encountering an issue only with `make docker_test`, but it appears to be related to the Docker configuration on my machine, as similar issues are present on the main branch as well.


SimplyDanny:
I don't really see the benefit of this. Yes, compile times can be reduced considerably for selected functions, yet the code becomes cluttered with types nobody needed before. What's more, most of the compile time is spent to build SwiftSyntax anyway. Thus the total win would not be significant.

PR #5497, opened 2024-03-15
Introduce SwiftLintCommandPlugin

garricn:
## Changelog

- Introduce `SwiftLintCommandPlugin`
- Rename `SwiftLintBuildToolPlugin`
- Add Swift Package Manager installation instructions
- Improve README.md

## Benefits

- Allows running SwiftLint from the command line when added as a package dependency
- Command line usage can be pinned to a specific version (as compared to Homebrew, for example)
- The plugins are named `SwiftLintCommandPlugin` and `SwiftLintBuildToolPlugin` for clarity
- The README now explains how to add SwiftLint as a package dependency

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@
   [kohtenko](https://github.com/kohtenko)
 * Introduce SwiftLintCommand plugin.  
 * Introduce SwiftLintCommandPlugin plugin.  
+* Introduce SwiftLintCommandPlugin.
+  Rename SwiftLintBuildToolPlugin.
+  Add Swift Package Manager installation instructions.
+  Improve README.md.  
   [garricn](https://github.com/garricn)
 
 * Add new option `ignore_typealiases_and_associatedtypes` to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@
 * Rewrite `SwiftLintPlugin` using `BUILD_WORKSPACE_DIRECTORY` without relying
   on the `--config` option.  
   [Garric Nahapetian](https://github.com/garricn)
+  
+* Introduce SwiftLintCommandPlugin.
+  Rename SwiftLintBuildToolPlugin.
+  Add Swift Package Manager installation instructions.    
+  [garricn](https://github.com/garricn)
 
 #### Experimental
 
@@ -15,12 +20,6 @@
 * Ignore absence of a non-initial local config instead of
   falling back to default.  
   [kohtenko](https://github.com/kohtenko)
-* Introduce SwiftLintCommand plugin.  
-* Introduce SwiftLintCommandPlugin plugin.  
-* Introduce SwiftLintCommandPlugin.
-  Rename SwiftLintBuildToolPlugin.
-  Add Swift Package Manager installation instructions.    
-  [garricn](https://github.com/garricn)
 
 * Add new option `ignore_typealiases_and_associatedtypes` to
   `nesting` rule. It excludes `typealias` and `associatedtype`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
 * Ignore absence of a non-initial local config instead of
   falling back to default.  
   [kohtenko](https://github.com/kohtenko)
+* Introduce SwiftLintCommand plugin.  
+  [garricn](https://github.com/garricn)
 
 * Add new option `ignore_typealiases_and_associatedtypes` to
   `nesting` rule. It excludes `typealias` and `associatedtype`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
   falling back to default.  
   [kohtenko](https://github.com/kohtenko)
 * Introduce SwiftLintCommand plugin.  
+* Introduce SwiftLintCommandPlugin plugin.  
   [garricn](https://github.com/garricn)
 
 * Add new option `ignore_typealiases_and_associatedtypes` to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
   
 * Introduce SwiftLintCommandPlugin.
   Rename SwiftLintBuildToolPlugin.
-  Add Swift Package Manager installation instructions.    
+  Add Swift Package Manager installation instructions.  
   [garricn](https://github.com/garricn)
 
 #### Experimental
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,7 @@
 * Introduce SwiftLintCommandPlugin plugin.  
 * Introduce SwiftLintCommandPlugin.
   Rename SwiftLintBuildToolPlugin.
-  Add Swift Package Manager installation instructions.
-  Improve README.md.  
+  Add Swift Package Manager installation instructions.    
   [garricn](https://github.com/garricn)
 
 * Add new option `ignore_typealiases_and_associatedtypes` to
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,12 @@
 
 #### Enhancements
 
+* Introduce SwiftLintCommandPlugin.
+  Rename SwiftLintBuildToolPlugin.
```
SimplyDanny, line 14: This is a breaking change and should therefore be mentioned in the `Breaking` section.

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@
...
+  Add Swift Package Manager installation instructions.
+  Improve README.md.
```
SimplyDanny, line 16: While this is appreciated, it's not relevant enough to inform users about it:
```suggestion
```

Plugins/SwiftLintCommandPlugin/SwiftLintCommandPlugin.swift
```diff
@@ -0,0 +1,40 @@
...
+
+@main
+internal struct SwiftLintCommandPlugin: CommandPlugin {
+    internal func performCommand(
+        context: PluginContext,
+        arguments: [String]
+    ) throws {
+        let tool: PluginContext.Tool = try context.tool(named: "swiftlint")
+        // Caching is managed internally because the cache must be located within the `pluginWorkDirectory`.
+        guard !arguments.contains("--cache-path") else {
```
SimplyDanny, line 12: Use `if` to avoid negation.

README.md
```diff
@@ -20,6 +20,14 @@ unacceptable behavior to [info@realm.io](mailto:info@realm.io).
...
 ## Installation
 
+### Using [Swift Package Manager](https://github.com/apple/swift-package-manager):
+
+> Replace `<version>` with the desired minimum version.
+
+```swift
+.package(url: "https://github.com/realm/SwiftLint.git", from: "<version>")
+```
+
```
SimplyDanny, line 30: Could you mention the plugins here and link to the sections explaining their usage?

SimplyDanny (changes requested):

CHANGELOG.md
tinder-cfuller: Is it actually a breaking change if has not been released yet?

Plugins/SwiftLintCommandPlugin/SwiftLintCommandPlugin.swift
```diff
@@ -0,0 +1,40 @@
+import Foundation
+import PackagePlugin
+
+@main
+internal struct SwiftLintCommandPlugin: CommandPlugin {
+    internal func performCommand(
```
tinder-cfuller, line 6: May we take advantage of implicit internal?
```suggestion
struct SwiftLintCommandPlugin: CommandPlugin {
    func performCommand(
```

Plugins/SwiftLintCommandPlugin/SwiftLintCommandPlugin.swift
```diff
@@ -0,0 +1,40 @@
...
+        let tool: PluginContext.Tool = try context.tool(named: "swiftlint")
+        // Caching is managed internally because the cache must be located within the `pluginWorkDirectory`.
+        guard !arguments.contains("--cache-path") else {
+            Diagnostics.error("Setting Cache Path Not Allowed")
+            return
+        }
+        let process: Process = .init()
+        process.currentDirectoryURL = URL(fileURLWithPath: context.package.directory.string)
+        process.executableURL = URL(fileURLWithPath: tool.path.string)
+        // The analyze command does not support the --cache-path argument
```
tinder-cfuller, line 19:
```suggestion
        // The analyze command does not support the `--cache-path` argument
```

Plugins/SwiftLintCommandPlugin/SwiftLintCommandPlugin.swift
SimplyDanny: Oh yeah, please. 👍

CHANGELOG.md
SimplyDanny: The plugin has existed for about a year already.

tinder-cfuller: Oh, right 🤦 Sorry!

To avoid the breaking change then, we could consider the following:

- instead of `SwiftLintBuildToolPlugin`, use `SwiftLintPlugin` (as it has been named)
- instead of `SwiftLintCommandPlugin`, use `SwiftLintCommand` (without the word `plugin`)

@SimplyDanny Do you prefer `SwiftLintPlugin`/`SwiftLintCommand` (avoids breaking change, but not clear that the command is also a plugin) or `SwiftLintBuildToolPlugin`/`SwiftLintCommandPlugin` (breaking change, but makes it very clear what each one is)?

SimplyDanny: As a third alternative, `SwiftLintPlugin` could be deprecated (printing a message) and   `SwiftLintBuildToolPlugin` could be introduced as the new name, having two `*Plugins` in the `Package.swift` for some time.

However, I actually like the `SwiftLintCommand` name. So of all those variants that's probably the easiest.

tinder-cfuller: Thinking more about this … the recent change to the build tool plugin could be considered a breaking change for [unexpected project structures](https://github.com/realm/SwiftLint#unexpected-project-structures) where the configuration file is located outside of the package/project directory.

Because of this, it could be seen as a benefit that a name change to `SwiftLintBuildToolPlugin` is also breaking, bringing attention to the plugin's revamped implementation.

I will leave it up to you both (@SimplyDanny and @garricn) to decide. I just wanted to surface this additional consideration so you were aware. Thank you 🙏

garricn: The new names clearly distinguish that one is a build tool plugin and the other is a command plugin. It is unfortunate that it will require a change to - at least - the Package.swift files of consumers. However, I agree with @tinder-cfuller that this could be a benefit to make the plugin implementation change (which may be breaking) more apparent when consumers upgrade.

Do either of you anticipate some greater challenge posed by this breaking change than just simply updating the name in the Package.swift file? For example, it's also possible that `SwiftLintPlugin` is also referenced in the documentation of consumers. If so, then perhaps we should consider an alternative. If not, I don't see this breaking change posing a drastic challenge on upgrading consumers.

Nevertheless, I defer to the judgment of you all. Let me know your thoughts.

tinder-cfuller: I simply wanted to surface some options so an informed decision would be made. Thanks for considering the choices. Much appreciated. Thank you.

The new names are fine by me 👍

tinder-cfuller (approved):
Thank you for contributing this! ❤️

CHANGELOG.md
SimplyDanny: The only problem is that people tend to ignore changelogs. But in this case, there is no other choice but to rename the plugin when updating the SwiftLint version as it just won't resolve. That's fine then. Let's stay with what you propose in the PR.

SimplyDanny (approved):

tinder-garricnahapetian:
> Thank you for contributing this! ❤️

No, thank you 😉

garricn:
> Thank you for contributing this! ❤️

No, thank you 😉

CHANGELOG.md
garricn: Sounds good. Thank you for merging! I will be available for any Issues related to this command as well as the rename.

Nomad00:
This looks awesome! Just hit the point where we wanted this to exist.
Is the release plan known for this by chance? (My apologies if this is defined somewhere already.)

SimplyDanny:
Indeed it seems to be about time. A lot things have changed since the last release. @jpsim, what do you think?

mstanzianoinsulet:
Forgive me if I'm missing something obvious, but are there instructions for how to use the `SwiftLintCommandPlugin` specifically?

SimplyDanny:
> Forgive me if I'm missing something obvious, but are there instructions for how to use the `SwiftLintCommandPlugin` specifically?

The README state it. `swift package plugin swiftlint` should make it run.

Issue #5470, opened 2024-02-22
`unused_enumerated` rule should also check closure parameters


SimplyDanny:
The `unused_enumerated` rule triggers when either of the two values returned by `.enumerated()` isn't used (replaced by an underscore `_`). The rule currently checks only `for` loops. It could be extended to support closure parameters as well.

The rule would only check the immediate closure after the `.enumerated()` call. Otherwise, the number of parameters and their types could have changed. 

Triggering:

```swift
list.enumerated().map { idx, _ in idx }

list.enumerated().map { _, elem in elem }

list.enumerated().forEach { print($0) }

list.enumerated().map { $1 }
```

Non-triggering:

```swift
list.enumerated().map { idx, elem in "\(idx): \(elem)" }

list.enumerated().map { $0 + $1 }

list.enumerated().something().map { _, elem in elem }
```

PS: This enhancement was motivated by [this comment](https://github.com/nicklockwood/SwiftFormat/issues/1607#issuecomment-1928863362).

mildm8nnered:
I am playing around with some fixes for this ...

PR #5498, opened 2024-03-17
Unused enumerated rule improvements

mildm8nnered:
Addresses #5470 

Extends `unused_enumerated` rule to cover cases like

```
list.enumerated().map { idx, _ in idx }
list.enumerated().map { _, elem in elem }
```

and

```
list.enumerated().forEach { print($0) }
list.enumerated().map { $1 }
```

Non-triggering cases are:

```
list.enumerated().map { idx, elem in "\(idx): \(elem)" }
list.enumerated().map { $0 + $1 }
list.enumerated().something().map { _, elem in elem }
```

Of the newly introduced OSS violations, some are false positives. In cases like 

```
   let selection = button.itemIcons.enumerated()
       .filter { _, icon in point.x > rectForItem(icon)!.origin.x }
       .map { index, _ in index }
       .last ?? 0
```

We will only inspect the first closure, and have no way of telling if arguments are accessed in later closures.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/DuplicateImportsRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/DuplicateImportsRule.swift
@@ -153,7 +153,7 @@ private extension DuplicateImportsRule {
 
         override func visit(_ node: CodeBlockItemListSyntax) -> CodeBlockItemListSyntax {
             let itemsToRemove = node
-                .enumerated()
+                .enumerated() // swiftlint:disable:next unused_enumerated
                 .filter { !$1.isContainedIn(regions: disabledRegions, locationConverter: locationConverter) }
                 .map { ($0, $1.item.positionAfterSkippingLeadingTrivia) }
                 .filter { importPositionsToRemove.contains($1) }
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/DuplicateImportsRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/DuplicateImportsRule.swift
@@ -153,7 +153,7 @@ private extension DuplicateImportsRule {
 
         override func visit(_ node: CodeBlockItemListSyntax) -> CodeBlockItemListSyntax {
             let itemsToRemove = node
-                .enumerated() // swiftlint:disable:next unused_enumerated
+                .enumerated()
                 .filter { !$1.isContainedIn(regions: disabledRegions, locationConverter: locationConverter) }
                 .map { ($0, $1.item.positionAfterSkippingLeadingTrivia) }
                 .filter { importPositionsToRemove.contains($1) }
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
@@ -78,13 +78,11 @@ struct UnusedEnumeratedRule: Rule {
 
 private extension UnusedEnumeratedRule {
     private struct Closure {
-        let id: SyntaxIdentifier
         let enumeratedPosition: AbsolutePosition?
         var zeroPosition: AbsolutePosition?
         var onePosition: AbsolutePosition?
 
-        init(id: SyntaxIdentifier, enumeratedPosition: AbsolutePosition? = nil) {
-            self.id = id
+        init(enumeratedPosition: AbsolutePosition? = nil) {
             self.enumeratedPosition = enumeratedPosition
         }
     }
@@ -155,11 +153,11 @@ private extension UnusedEnumeratedRule {
 
         override func visit(_ node: ClosureExprSyntax) -> SyntaxVisitorContinueKind {
             if let nextClosureId, nextClosureId == node.id, let lastEnumeratedPosition {
-                closures.push(Closure(id: nextClosureId, enumeratedPosition: lastEnumeratedPosition))
+                closures.push(Closure(enumeratedPosition: lastEnumeratedPosition))
                 self.nextClosureId = nil
                 self.lastEnumeratedPosition = nil
             } else {
-                closures.push(Closure(id: node.id))
+                closures.push(Closure())
             }
             return .visitChildren
         }
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
@@ -64,7 +64,7 @@ struct UnusedEnumeratedRule: Rule {
                 }
                 return $0
             }
-            """)
+            """, excludeFromDocumentation: true)
         ]
     )
 }
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
@@ -64,7 +64,8 @@ private extension UnusedEnumeratedRule {
 
         override func visit(_ node: FunctionCallExprSyntax) -> SyntaxVisitorContinueKind {
             guard node.isEnumerated,
-                  let trailingClosure = node.parent?.parent?.as(FunctionCallExprSyntax.self)?.trailingClosure
+                  let trailingClosure = node.parent?.parent?.as(FunctionCallExprSyntax.self)?.trailingClosure,
+                  node.parent?.as(MemberAccessExprSyntax.self)?.declName.baseName.text != "filter"
             else {
                 return .visitChildren
             }
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
@@ -22,7 +22,13 @@ struct UnusedEnumeratedRule: Rule {
             Example("list.enumerated().map { idx, elem in \"\\(idx): \\(elem)\" }"),
             Example("list.enumerated().map { $0 + $1 }"),
             Example("list.enumerated().something().map { _, elem in elem }"),
-            Example("list.map { ($0.offset, $0.element) }")
+            Example("list.map { ($0.offset, $0.element) }"),
+            Example("""
+            list.enumerated().map {
+                $1.enumerated().forEach { print($0, $1) }
+                return $0
+            }
+            """)
         ],
         triggeringExamples: [
             Example("for (↓_, foo) in bar.enumerated() { }"),
@@ -32,16 +38,64 @@ struct UnusedEnumeratedRule: Rule {
             Example("list.enumerated().map { idx, ↓_ in idx }"),
             Example("list.enumerated().map { ↓_, elem in elem }"),
             Example("list.enumerated().forEach { print(↓$0) }"),
-            Example("list.enumerated().map { ↓$1 }")
+            Example("list.enumerated().map { ↓$1 }"),
+            Example("""
+            list.enumerated().map {
+                $1.enumerated().forEach { print(↓$1) }
+                return $0
+            }
+            """),
+            Example("""
+            list.enumerated().map {
+                ↓$1.enumerated().forEach { print($0, $1) }
+                return 1
+            }
+            """),
+            Example("""
+            list.enumerated().map {
+                list.enumerated().forEach { print($0, $1) }
+                return ↓$0
+            }
+            """),
+            Example("""
+            list.enumerated().map {
+                $1.enumerated().forEach {
+                    print($0, $1)
+                     $1.enumerated().forEach {
+                         print(↓$1)
+                     }
+                }
+                return $0
+            }
+            """),
+            Example("""
+            list.enumerated().map {
+                $1.enumerated().filter {
+                    print($0, $1)
+                    $1.enumerated().forEach {
+                         if ↓$1 == 2 {
+                             return true
+                         }
+                    }
+                    return false
+                }
+                return $0
+            }
+            """)
         ]
     )
 }
 
 private extension UnusedEnumeratedRule {
+    private struct Closure {
+        var closure: ClosureExprSyntax
+        var zeroPosition: AbsolutePosition?
+        var onePosition: AbsolutePosition?
+    }
+
     final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
-        private var trailingClosure: ClosureExprSyntax?
-        private var zeroPosition: AbsolutePosition?
-        private var onePosition: AbsolutePosition?
+        private var nextClosure: ClosureExprSyntax?
+        private var closures: [Closure] = []
 
         override func visitPost(_ node: ForStmtSyntax) {
             guard let tuplePattern = node.pattern.as(TuplePatternSyntax.self),
@@ -95,41 +149,47 @@ private extension UnusedEnumeratedRule {
                     onePosition: firstTokenIsUnderscore ? nil : secondElement.positionAfterSkippingLeadingTrivia
                 )
             } else {
-                self.trailingClosure = trailingClosure
+                nextClosure = trailingClosure
             }
 
             return .visitChildren
         }
 
+        override func visit(_ node: ClosureExprSyntax) -> SyntaxVisitorContinueKind {
+            if node == nextClosure {
+                closures.append(Closure(closure: node))
+                nextClosure = nil
+            }
+            return .visitChildren
+        }
+
         override func visitPost(_ node: ClosureExprSyntax) {
-            guard node == trailingClosure else {
+            guard let closure = closures.last, node == closure.closure else {
                 return
             }
-            defer {
-                trailingClosure = nil
-                zeroPosition = nil
-                onePosition = nil
-            }
-            guard (zeroPosition != nil) != (onePosition != nil) else {
+            defer { closures.removeLast() }
+
+            guard (closure.zeroPosition != nil) != (closure.onePosition != nil) else {
                 return
             }
 
-            addViolation(zeroPosition: zeroPosition, onePosition: onePosition)
+            addViolation(zeroPosition: closure.zeroPosition, onePosition: closure.onePosition)
         }
 
         override func visitPost(_ node: DeclReferenceExprSyntax) {
-            guard trailingClosure != nil else {
+            guard var closure = closures.last, node.baseName.text == "$0" || node.baseName.text == "$1" else {
                 return
             }
             if node.baseName.text == "$0" {
                 if node.parent?.as(MemberAccessExprSyntax.self)?.declName.baseName.text == "element" {
-                    onePosition = node.positionAfterSkippingLeadingTrivia
+                    closure.onePosition = node.positionAfterSkippingLeadingTrivia
                 } else {
-                    zeroPosition = node.positionAfterSkippingLeadingTrivia
+                    closure.zeroPosition = node.positionAfterSkippingLeadingTrivia
                 }
-            } else if node.baseName.text == "$1" {
-                onePosition = node.positionAfterSkippingLeadingTrivia
+            } else {
+                closure.onePosition = node.positionAfterSkippingLeadingTrivia
             }
+            closures[closures.count - 1] = closure
         }
 
         private func addViolation(zeroPosition: AbsolutePosition?, onePosition: AbsolutePosition?) {
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
@@ -92,10 +92,6 @@ private extension UnusedEnumeratedRule {
             return .visitChildren
         }
 
-        override func visit(_ node: ClosureExprSyntax) -> SyntaxVisitorContinueKind {
-            node == trailingClosure ? .visitChildren : .skipChildren
-        }
-
         override func visitPost(_ node: ClosureExprSyntax) {
             guard node == trailingClosure else {
                 return
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
@@ -19,7 +19,7 @@ struct UnusedEnumeratedRule: Rule {
             Example("for (idx, _) in bar.something() { }"),
             Example("for idx in bar.indices { }"),
             Example("for (section, (event, _)) in data.enumerated() {}"),
-            // Example("list.enumerated().map { idx, elem in \"\(idx): \(elem)\" }"),
+//            Example("list.enumerated().map { idx, elem in \"\(idx): \(elem)\" }"),
             Example("list.enumerated().map { $0 + $1 }"),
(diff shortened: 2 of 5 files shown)


SimplyDanny:
> They are all variations of the following idiom:
> 
> ```
>    let selection = button.itemIcons.enumerated()
>        .filter { _, icon in point.x > rectForItem(icon)!.origin.x }
>        .map { index, _ in index }
>        .last ?? 0
> ```

Well, that's a case I didn't think of. This cannot be recognized reliably, I fear. Not sure the extension of the existing rule is useful then. 🤔

mildm8nnered:
> > They are all variations of the following idiom:
> > ```
> >    let selection = button.itemIcons.enumerated()
> >        .filter { _, icon in point.x > rectForItem(icon)!.origin.x }
> >        .map { index, _ in index }
> >        .last ?? 0
> > ```
> 
> Well, that's a case I didn't think of. This cannot be recognized reliably, I fear. Not sure the extension of the existing rule is useful then. 🤔
> 
> A way around would be to check subsequent `.map` and `.forEach` calls only. Anything more that makes sense?

So I don't think it blows it out of the water necessarily.

Two false alarms on the SwiftLint codebase and 5/17 hits on the OSS - ok, that's not great, but I suspect the problematic idiom is not that common.

We could make this behaviour configurable perhaps ...

SimplyDanny:
> Two false alarms on the SwiftLint codebase and 5/17 hits on the OSS - ok, that's not great, but I suspect the problematic idiom is not that common.

I still think, the rule should only check a few specific cases like `.map`, `.forEach`, ... then. Alternatively, it could stop analysis at `.filter` already.

mildm8nnered:
So I found quite a subtle bug around the 

```
        return Dictionary(
            uniqueKeysWithValues: contents.bridge()
                .components(separatedBy: "\n") // swiftlint:disable:next unused_enumerated
                .map { Int64($0.bridge().lengthOfBytes(using: .utf8)) }
                .reduce(into: [0]) { result, length in
                    let newLineCharacterLength = Int64(1)
                    let lineLength = length + newLineCharacterLength
                    result.append(contentsOf: [(result.last ?? 0) + lineLength])
                }
                .enumerated()
                .map { ($0.offset, $0.element) }
        )
```

case from SwiftLint's own code. We should be detecting a violation here - `.map { ($0.offset, $0.element) }` instead of above.

SimplyDanny:
The fixed violation is not correct, is it?

SimplyDanny:
> Secondly, we were not being very discriminating about which closures we ended up looking at - now we will only inspect the trailingClosure of the call immediately after the enumeration.

I think that the rule should stop on `.filter`. There might always come something after the filter that uses the other value returned by `.enumerated()` even though `.filter` doesn't.

mildm8nnered:
> The fixed violation is not correct, is it?

`list.map { ($0.offset, $0.element) }` this one? 

That should be non-triggering, as the index and element can be accessed via `offset` and `element` on the tuple values that`enumerated()` returns (which I did not know before).

mildm8nnered:
> [The last and only fix](https://github.com/wordpress-mobile/WordPress-iOS/blob/2537c045b9854bea324d159a3ac25815ee083656/WordPress/Classes/Stores/StatsPeriodStore.swift#L1432:30) in the OSS report a mean.

Ah - I didn't spot that - yes, that looks problematic :-( I'll see if I can work out what I've broken.

mildm8nnered:
> > [The last and only fix](https://github.com/wordpress-mobile/WordPress-iOS/blob/2537c045b9854bea324d159a3ac25815ee083656/WordPress/Classes/Stores/StatsPeriodStore.swift#L1432:30) in the OSS report a mean.

That was quite a lucky spot - resolved by https://github.com/realm/SwiftLint/pull/5498/commits/32c445bf939b25d2641f607bb3eef92639622b67 - we were skipping closures that didn't match our trailing closure.

mildm8nnered:
> > Secondly, we were not being very discriminating about which closures we ended up looking at - now we will only inspect the trailingClosure of the call immediately after the enumeration.
> 
> I think that the rule should stop on `.filter`. There might always come something after the filter that uses the other value returned by `.enumerated()` even though `.filter` doesn't.

So I believe that is what the code now does, both for cases like

```
            itemIcons.enumerated()
                   .filter { ↓_, icon in icon == x }
                   .map { index, _ in index }
                   .last
```

```
            itemIcons.enumerated()
                   .filter { ↓$0 == x }
                   .map { $1 }
                   .last
```

We'll only inspect the first closure after the enumerated node in both cases.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/DuplicateImportsRule.swift
```diff
@@ -153,7 +153,7 @@ private extension DuplicateImportsRule {
 
         override func visit(_ node: CodeBlockItemListSyntax) -> CodeBlockItemListSyntax {
             let itemsToRemove = node
-                .enumerated()
+                .enumerated() // swiftlint:disable:next unused_enumerated
```
SimplyDanny, line 156: I still wonder why this needs to be skipped. I think that the rule should either stop immediately when encountering a `filter` method or it should skip it and continue checking the subsequent `map`-like methods. If there aren't any of the latter, the rule remains silent.

mildm8nnered: Ah - my misunderstanding - I would love to continue checking, but I couldn't find an elegant way to do that - I'll look again.

Or we can just stop on filter and ignore (as opposed to stopping after filter, which is what we're actually doing in the current code here).

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -44,17 +56,87 @@ private extension UnusedEnumeratedRule {
...
+            addViolation(
+                zeroPosition: firstTokenIsUnderscore ? firstElement.positionAfterSkippingLeadingTrivia : nil,
+                onePosition: firstTokenIsUnderscore ? nil : secondElement.positionAfterSkippingLeadingTrivia
+            )
+        }
+
+        override func visit(_ node: FunctionCallExprSyntax) -> SyntaxVisitorContinueKind {
+            guard node.isEnumerated,
+                  let trailingClosure = node.parent?.parent?.as(FunctionCallExprSyntax.self)?.trailingClosure,
+                  node.parent?.as(MemberAccessExprSyntax.self)?.declName.baseName.text != "filter"
```
SimplyDanny, line 68: Add `let parent = node.parent` and check the last condition first.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -44,17 +56,87 @@ private extension UnusedEnumeratedRule {
...
+                return .visitChildren
+            }
+
+            if let parameterClause = trailingClosure.signature?.parameterClause {
+                guard let parameterClause = parameterClause.as(ClosureShorthandParameterListSyntax.self),
+                      parameterClause.count == 2,
+                      let firstElement = parameterClause.firstParameter,
+                      let secondElement = parameterClause.lastParameter,
+                      case let firstTokenIsUnderscore = firstElement.isUnderscore,
+                      case let lastTokenIsUnderscore = secondElement.isUnderscore,
```
SimplyDanny, line 79: I don't like the `case let` so much. Perhaps split the `guard` condition.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -44,17 +56,87 @@ private extension UnusedEnumeratedRule {
...
+                  let trailingClosure = node.parent?.parent?.as(FunctionCallExprSyntax.self)?.trailingClosure,
+                  node.parent?.as(MemberAccessExprSyntax.self)?.declName.baseName.text != "filter"
+            else {
+                return .visitChildren
+            }
+
+            if let parameterClause = trailingClosure.signature?.parameterClause {
+                guard let parameterClause = parameterClause.as(ClosureShorthandParameterListSyntax.self),
+                      parameterClause.count == 2,
+                      let firstElement = parameterClause.firstParameter,
```
SimplyDanny, line 76:
```suggestion
                      let firstElement = parameterClause.first,
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -44,17 +56,87 @@ private extension UnusedEnumeratedRule {
...
+                      let secondElement = parameterClause.lastParameter,
```
SimplyDanny, line 77:
```suggestion
                      let secondElement = parameterClause.last,
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -83,3 +165,19 @@ private extension TuplePatternElementSyntax {
...
+
+private extension ClosureShorthandParameterListSyntax {
+    var firstParameter: ClosureShorthandParameterSyntax? {
+        children(viewMode: .sourceAccurate).first?.as(ClosureShorthandParameterSyntax.self)
+    }
+
+    var lastParameter: ClosureShorthandParameterSyntax? {
+        children(viewMode: .sourceAccurate).last?.as(ClosureShorthandParameterSyntax.self)
+    }
+}
```
SimplyDanny, line 183: That's just `.first` and `.last`. 😉 
```suggestion
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -44,17 +56,87 @@ private extension UnusedEnumeratedRule {
...
             } else {
-                position = secondElement.positionAfterSkippingLeadingTrivia
-                reason = "When the item is not used, `.indices` should be used instead of `.enumerated()`"
+                self.trailingClosure = trailingClosure
+            }
+
+            return .visitChildren
+        }
+
+        override func visitPost(_ node: ClosureExprSyntax) {
```
SimplyDanny, line 96: This not going to work for closures in closures. I recommend to go with a separate visitor only collecting appearances of `$0`, `$0.element`, ...

Please also add tests.

SimplyDanny (changes requested):
This is the most the rule can check, I assume. Looks good now! 👍

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
mildm8nnered: 🤦

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
mildm8nnered: Cool - will take a bit of time to work this out

mildm8nnered: Do you have an example that you think will fail @SimplyDanny ?

SimplyDanny: Without trying:

```swift
list.enumerated().map {
    $1.enumerated().forEach { print($0, $1) }
    return $0
}
```

I assume this would trigger the rule on the outer `enumerated()`, even though it's actually validly used.

mildm8nnered: ok - so I think I have it working for closures within closures, indefinitely. The basic idea is to keep a stack that tracks   the zero and one position for each closure, and then push and pop the data structures onto the stack as we encounter and depart each closure.

It feels to me like it adds a fair bit of complexity for something that is a bit of an edge case, but it does pass all the tests and some new ones. First pass of code is on a PR targeting this branch at 

https://github.com/mildm8nnered/SwiftLint/pull/11

If you think that generally looks good-ish, I'll tidy it up and merge it into this branch.

SimplyDanny: I added some comments in the other PR.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -18,43 +18,177 @@ struct UnusedEnumeratedRule: Rule {
...
+
+            guard (closure.zeroPosition != nil) != (closure.onePosition != nil) else {
+                return
+            }
+
+            addViolation(zeroPosition: closure.zeroPosition, onePosition: closure.onePosition)
+        }
+
+        override func visitPost(_ node: DeclReferenceExprSyntax) {
+            guard var closure = closures.peek(), node.baseName.text == "$0" || node.baseName.text == "$1" else {
```
mildm8nnered, line 163: > How does it make sure that the closure is the right one?

So when we entered the closure, in `visit(_ node: ClosureExprSyntax)`, if the closure was the trailing closure of an enumerated function call (whose id is stored in `nextClosureId`) we push it onto the stack, so that is what `closures.peek()` will return.

And when we leave the closure, in `visitPost(_ node: ClosureExprSyntax)`, we pop that closure off the stack, and update state for that enclosing closure.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -18,43 +18,177 @@ struct UnusedEnumeratedRule: Rule {
...
             Example("for (↓_, foo) in abc.something().enumerated() { }"),
-            Example("for (idx, ↓_) in bar.enumerated() { }")
+            Example("for (idx, ↓_) in bar.enumerated() { }"),
+            Example("list.enumerated().map { idx, ↓_ in idx }"),
+            Example("list.enumerated().map { ↓_, elem in elem }"),
+            Example("list.enumerated().forEach { print(↓$0) }"),
+            Example("list.enumerated().map { ↓$1 }"),
+            Example("""
+            list.enumerated().map {
+                $1.enumerated().forEach { print(↓$1) }
```
mildm8nnered, line 44: > It feels a bit strange that the rule triggers on the identifiers (that are not missing) even though the rule is about the `enumerated()` method.

I know what you mean - it's not like the `_` case where it's obvious. I think of it as `$0` (or `$1`) being the thing that indicates there's a problem, as we didn't expect to see only one of `$0` or `$1`.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
mildm8nnered: Thanks! I've either adopted them wholesale, or replied inline on this PR ...

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
SimplyDanny: I'd still expect the rule to trigger on the `enumerated()` call to be honest. Changing that shouldn't be a big deal, right?

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
SimplyDanny: Consider
```swift
list.enumerated().map {
    $1.forEach { print($0) }
    return $1 
}
```
which is expected to trigger on the last `$1`, right? But does it actually? 🤓

SimplyDanny (changes requested):

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
mildm8nnered: Dammit. I think I might need to go back to the drawing board on this one :-(

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
mildm8nnered: Yep, that should be possible.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
SimplyDanny: Hint: `ShorthandArgumentRule`

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
mildm8nnered: So I didn't go the `ShorthandArgumentRule` route. Instead, when I see a closure that is not the one I'm looking for, I  push it onto the stack anyway, and then stop counting `$0`'s and `$1`'s. Seems to work ok for that test case.

mildm8nnered:
So we now place the violation just before the `enumerated()` call for the `$0` and `$1` cases, and `$0` or `$1` should be ignored in irrelevant closures.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -18,43 +18,211 @@ struct UnusedEnumeratedRule: Rule {
...
+
+            guard (closure.zeroPosition != nil) != (closure.onePosition != nil) else {
+                return
+            }
+
+            addViolation(
+                zeroPosition: closure.zeroPosition,
+                onePosition: closure.onePosition,
+                enumeratedPosition: closure.enumeratedPosition
+            )
```
SimplyDanny, line 181: The case that the closure IDs don't match seem impossible. Simplify: 
```suggestion
            if let closure = closures.pop(), (closure.zeroPosition != nil) != (closure.onePosition != nil) {
                addViolation(
                    zeroPosition: closure.zeroPosition,
                    onePosition: closure.onePosition,
                    enumeratedPosition: closure.enumeratedPosition
                )
            }
```

SimplyDanny (approved):

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -18,57 +18,221 @@ struct UnusedEnumeratedRule: Rule {
...
+                return $0
+            }
+            """, excludeFromDocumentation: true)
+            ,
+            Example("""
+            list.↓enumerated().map {
+                $1.forEach { print($0) }
+                return $1
+            }
+            """, excludeFromDocumentation: true)
```
SimplyDanny, line 74: Thanks for excluding the examples! 👍

SimplyDanny:
The [last OSS finding](https://github.com/wireapp/wire-ios/blob/5d2f52ccf3ee0afaf32f64c2ce50c01081415284/wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController/ViewModel/ConversationListViewModel.swift#L551:22) is a false positive. The implementation needs to check likewise for `$0.offset` in addition to `$0.element`.

mildm8nnered:
> The [last OSS finding](https://github.com/wireapp/wire-ios/blob/5d2f52ccf3ee0afaf32f64c2ce50c01081415284/wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController/ViewModel/ConversationListViewModel.swift#L551:22) is a false positive. The implementation needs to check likewise for `$0.offset` in addition to `$0.element`.

So I think that is actually a hit:

```
            sections.enumerated().forEach {
                delegate?.listViewModel(self, didUpdateSection: $0.offset)
            }
```

`$0.element` isn't referenced, so this could be

```
            sections.indices.forEach {
                delegate?.listViewModel(self, didUpdateSection: $0)
            }
```

I was providing the wrong reason for the `$0` and `offset/element` cases though 🤦 which I've fix in the last commit.

SimplyDanny:
> > The [last OSS finding](https://github.com/wireapp/wire-ios/blob/5d2f52ccf3ee0afaf32f64c2ce50c01081415284/wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController/ViewModel/ConversationListViewModel.swift#L551:22) is a false positive. The implementation needs to check likewise for `$0.offset` in addition to `$0.element`.
> 
> 
> 
> So I think that is actually a hit:
> 
> 
> 
> ```
> 
>             sections.enumerated().forEach {
> 
>                 delegate?.listViewModel(self, didUpdateSection: $0.offset)
> 
>             }
> 
> ```
> 
> 
> 
> `$0.element` isn't referenced, so this could be
> 
> 
> 
> ```
> 
>             sections.indices.forEach {
> 
>                 delegate?.listViewModel(self, didUpdateSection: $0)
> 
>             }
> 
> ```

You're right. 👍 The implementation assumes that any member access on `$0` is either `element` or `offset`. Seems correct, but wasn't instantly clear to me.

mildm8nnered:
> You're right. 👍 The implementation assumes that any member access on `$0` is either `element` or `offset`. Seems correct, but wasn't instantly clear to me.

So this logic here should handle that:


```
              if node.baseName.text == "$0" {
                    if node.parent?.as(MemberAccessExprSyntax.self)?.declName.baseName.text == "element" {
                        $0.onePosition = node.positionAfterSkippingLeadingTrivia
                    } else {
                        $0.zeroPosition = node.positionAfterSkippingLeadingTrivia
                    }
                } else {
                    $0.onePosition = node.positionAfterSkippingLeadingTrivia
                }
            }
```

`$0.element` we effective treat as though we saw `$1` ...

SimplyDanny:
> You're right. 👍 The implementation assumes that any member access on `$0` is either `element` or `offset`. Seems correct, but wasn't instantly clear to me.

Well, actually `$0.offset` can also be addressed by `$0.0` or even crazy things like `$0.self.0` or `$0.self.self.0` or ... While the last examples seem abnormal, I've seen `$0.0` and `$0.1` quite often.

mildm8nnered:
> > You're right. 👍 The implementation assumes that any member access on `$0` is either `element` or `offset`. Seems correct, but wasn't instantly clear to me.
> 
> Well, actually `$0.offset` can also be addressed by `$0.0` or even crazy things like `$0.self.0` or `$0.self.self.0` or ... While the last examples seem abnormal, I've seen `$0.0` and `$0.1` quite often.

We could probably cover `$0.1` in the same way that we do `element` if you think we should ...

SimplyDanny:
> > > You're right. 👍 The implementation assumes that any member access on `$0` is either `element` or `offset`. Seems correct, but wasn't instantly clear to me.
> > 
> > 
> > Well, actually `$0.offset` can also be addressed by `$0.0` or even crazy things like `$0.self.0` or `$0.self.self.0` or ... While the last examples seem abnormal, I've seen `$0.0` and `$0.1` quite often.
> 
> We could probably cover `$0.1` in the same way that we do `element` if you think we should ...

Doesn't seem to be a big deal. I think we should.

mildm8nnered:
> Doesn't seem to be a big deal. I think we should.

Cool - should be good now ...

PR #4758, opened 2023-02-08
Use BUILD_WORKSPACE_DIRECTORY in SwiftLintPlugin

garricn:
## Changelog

- Rewrite `SwiftLintPlugin` using `BUILD_WORKSPACE_DIRECTORY` without relying on the `--config` option
- Rewrite `Plugins/SwiftLintPlugin/Path+Helpers.swift`
- Update Plug-in Support in `README.md`

## Benefits

- Supports parent/child configurations
- Supports nested configurations

## Background

The current version of the plugin uses the `--config` option which is incompatible with project structures using parent/child configurations or nested configurations.

- References:
    - https://github.com/realm/SwiftLint/blob/0.50.3/Plugins/SwiftLintPlugin/SwiftLintPlugin.swift#L33
    - https://github.com/realm/SwiftLint/blob/0.50.3/Plugins/SwiftLintPlugin/SwiftLintPlugin.swift#L79

Instead of relying on the `--config` option, the plugin can leverage the `BUILD_WORKSPACE_DIRECTORY` environment variable (recently added to support Bazel) which sets SwiftLint’s working directory.

- Reference: https://github.com/realm/SwiftLint/blob/0.50.3/Source/swiftlint/Commands/SwiftLint.swift#L7

## Implementation Details

The plugin determines the SwiftLint working directory by locating the topmost config file within the package/project directory. If a config file is not found therein, the package/project directory is used as the working directory.

## Project Structures

The plugin throws an error when it is unable to resolve the SwiftLint working directory. For example, this will occur in Xcode projects where the target's Swift files are not located within the project directory.

To maximize compatibility with the plugin, avoid project structures that require the use of the `--config` option.

### Unexpected Project Structures

Project structures where SwiftLint's configuration file is located outside of the package/project directory are not directly supported by the build tool plugin. This is because it isn't possible to pass arguments to build tool plugins (e.g., passing the config file path). 

If your project structure doesn't work directly with the build tool plugin, please consider one of the following options:

- To use a config file located outside the package/project directory, a config file may be added to that directory specifying a parent config path to the other config file, e.g., `parent_config: path/to/.swiftlint.yml`.
- You can also consider the use of a Run Script Build Phase in place of the build tool plugin.

PR #5500, opened 2024-03-18
Move “Rewrite SwiftLintPlugin” to Breaking section

garricn:
The CHANGELOG entry for #4758 should have been added to the Breaking section since the plugin no longer uses the `--config` option.


SimplyDanny:
Thanks!

PR #5506, opened 2024-03-23
Report specific issues

SimplyDanny:
Use different issues to make inlineable configurations work in both scenarios, namely being used standalone and nested as part of another configuration. In the latter case, finding no values to parse in a raw configuration is okay, while as a standalone configuration it's not.

### DIFF
--- a/Tests/MacroTests/AutoApplyTests.swift
+++ b/Tests/MacroTests/AutoApplyTests.swift
@@ -122,8 +122,12 @@ final class AutoApplyTests: XCTestCase {
                 var eC = 3
 
                 mutating func apply(configuration: Any) throws {
+                    do {
                     try eB.apply(configuration, ruleID: Parent.identifier)
                     try $eB.performAfterParseOperations()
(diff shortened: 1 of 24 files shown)


Tests/MacroTests/AutoApplyTests.swift
```diff
@@ -122,8 +122,12 @@ final class AutoApplyTests: XCTestCase {
                 var eC = 3
 
                 mutating func apply(configuration: Any) throws {
+                    do {
                     try eB.apply(configuration, ruleID: Parent.identifier)
                     try $eB.performAfterParseOperations()
+                    } catch let issue as Issue where issue == Issue.nothingApplied(ruleID: Parent.identifier) {
+                    // Acceptable. Continue.
+                }
```
SimplyDanny, line 130: Indentation isn't working so well when concatenating code represented as strings. I'll investigate if result builders work better ...

SimplyDanny: #5507

Issue #5450, opened 2024-02-02
Is blanket_disable_command really necessary?


tomhuettmann:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version (updated via brew, brew currently still uses `0.53.0`)
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Question / New rule request

I stumbled across the paragraph in the `README.md:L432-433`:

> ...
> The rules will be disabled until the end of the file or until the linter sees a matching enable comment:
> ...

and see a collision with the in https://github.com/realm/SwiftLint/pull/4731 introduced `blanket_disable_command` rule (https://realm.github.io/SwiftLint/blanket_disable_command.html).

My question now is if re-enabling rules before the end of file is really necessary or not? I tried it out in a simple project, and it seems like written in the README.md, all the disabled files are automatically enabled again before linting the next file.

If so, I would like to recommend a new rule to get rid of those introduced unnecessary enable commands at file endings:

```yaml
no_swiftlint_enable_at_end_of_file:
  name: 'No SwiftLint enable at end of file'
  message: 'There is no swiftlint:enable necessary at the end of a file'
  regex: 'swiftlint:enable[^\n]*\n*\Z'
```

Non triggering example:
```
// swiftlint:disable foo
... some code ...
// swiftlint:enable foo
... some code ...
```

Triggering example:
```
// swiftlint:disable foo
... some code ...
// swiftlint:enable foo

```
, where the `// swiftlint:enable foo` is the last thing before EOF.

SimplyDanny:
The idea behind the `blanket_disable_command` rule is like: People might just write `// swiftlint:disable` while the they should use `// swiftlint:disable:this`, `// swiftlint:disable:next` or `// swiftlint:disable:previous` to be very explicit about the location to ignore; to make this a real exception. The rule assumes, a blanket `// swiftlint:disable` is unintentional and so notifies users about it.

Enabling a disabled rule at the end of the file is one way to circumvent `blanket_disable_command` from triggering in the whole file. Another one is

```
// swiftlint:disable:next blanket_disable_command
// swiftlint:disable nesting
```

at the beginning of the file.

I like your idea to complain about enabled rules at the end of files. The recommended alternative would be to disable the rule consciously at the beginning of files. This could be an option for `blanket_disable_command`.

@mildm8nnered: You are the expert for `blanket_disable_command`. What do you think?

mildm8nnered:
[...]

PR #5518, opened 2024-03-30
Updated blanket_disable_command reasons

mildm8nnered:
Updates the rule description and a reason for the `blanket_disable_rule`, avoiding language about the end of the file.

Partly addresses #5450 

New top level rule description (I think this will only appear in the documentation perhaps).

old: "swiftlint:disable commands should be re-enabled before the end of the file"

new: "`swiftlint:disable` commands should use `next`, `this` or `previous` to disable rules for a
                     single line, or `swiftlint:enable` to re-enable the rules immediately after the violations
                     to be ignored, instead of disabling the rule for the rest of the file"

and the reason:

old: "The disabled 'XXXX' rule should be re-enabled before the end of the file"

new: "Use 'next', 'this' or 'previous' instead to disable the 'XXXX' rule once or re-enable it as soon as possible"

The OSS violations and fixes should match exactly - they've just been triggered by the reason change.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/BlanketDisableCommandRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/BlanketDisableCommandRule.swift
@@ -5,8 +5,8 @@ struct BlanketDisableCommandRule: Rule, SourceKitFreeRule {
         identifier: "blanket_disable_command",
         name: "Blanket Disable Command",
         description: """
-                     `swiftlint:disable` commands should use `next`, `this` or `previous` to disable rules for a
-                     single line, or `swiftlint:enable` to re-enable the rules immediately after the violations
+                     `swiftlint:disable` commands should use `next`, `this` or `previous` to disable rules for a \
+                     single line, or `swiftlint:enable` to re-enable the rules immediately after the violations \
                      to be ignored, instead of disabling the rule for the rest of the file
                      """,
         kind: .lint,
@@ -148,9 +148,11 @@ struct BlanketDisableCommandRule: Rule, SourceKitFreeRule {
             }
 
             if let command = ruleIdentifierToCommandMap[disabledRuleIdentifier] {
-                let reason = "Use 'next', 'this' or 'previous' instead to disable the " +
-                              "'\(disabledRuleIdentifier.stringRepresentation)' rule once " +
-                              " or re-enable it as soon as possible"
+                let reason = """
+                             Use 'next', 'this' or 'previous' instead to disable the \
+                             '\(disabledRuleIdentifier.stringRepresentation)' rule once, \
+                             or re-enable it as soon as possible`
+                             """
                 return violation(for: command, ruleIdentifier: disabledRuleIdentifier, in: file, reason: reason)
             }
             return nil
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/BlanketDisableCommandRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/BlanketDisableCommandRule.swift
@@ -4,7 +4,9 @@ struct BlanketDisableCommandRule: Rule, SourceKitFreeRule {
     static let description = RuleDescription(
         identifier: "blanket_disable_command",
         name: "Blanket Disable Command",
-        description: "swiftlint:disable commands should be re-enabled before the end of the file",
+        description: "'swiftlint:disable' commands should use 'next', 'this' or 'previous' to disable rules for a " +
+                     "single line, or 'swiftlint:enable' to re-enable the rules immediately after the violations " +
+                     "to be ignored",
         kind: .lint,
         nonTriggeringExamples: [
             Example("""
@@ -144,8 +146,9 @@ struct BlanketDisableCommandRule: Rule, SourceKitFreeRule {
             }
 
             if let command = ruleIdentifierToCommandMap[disabledRuleIdentifier] {
-                let reason = "The disabled '\(disabledRuleIdentifier.stringRepresentation)' rule " +
-                             "should be re-enabled before the end of the file"
+                let reason = "Use 'next', 'this' or 'previous' to disable the " +
+                             "'\(disabledRuleIdentifier.stringRepresentation)' rule for a single line or " +
+                             "'swiftlint:enable' to re-enable the rule immediately after the violations to be ignored"
                 return violation(for: command, ruleIdentifier: disabledRuleIdentifier, in: file, reason: reason)
             }
             return nil
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/BlanketDisableCommandRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/BlanketDisableCommandRule.swift
@@ -148,10 +148,9 @@ struct BlanketDisableCommandRule: Rule, SourceKitFreeRule {
             }
 
             if let command = ruleIdentifierToCommandMap[disabledRuleIdentifier] {
-                let reason = "Use 'next', 'this' or 'previous' to disable the " +
-                             "'\(disabledRuleIdentifier.stringRepresentation)' rule for a single line or " +
-                             "'swiftlint:enable' to re-enable it as soon as possible"
-
+                let reason = "Use 'next', 'this' or 'previous' instead to disable the " +
+                              "'\(disabledRuleIdentifier.stringRepresentation)' rule once " +
+                              " or re-enable it as soon as possible"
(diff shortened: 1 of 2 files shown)


Source/SwiftLintBuiltInRules/Rules/Lint/BlanketDisableCommandRule.swift
```diff
@@ -144,8 +146,9 @@ struct BlanketDisableCommandRule: Rule, SourceKitFreeRule {
             }
 
             if let command = ruleIdentifierToCommandMap[disabledRuleIdentifier] {
-                let reason = "The disabled '\(disabledRuleIdentifier.stringRepresentation)' rule " +
-                             "should be re-enabled before the end of the file"
+                let reason = "Use 'next', 'this' or 'previous' to disable the " +
+                             "'\(disabledRuleIdentifier.stringRepresentation)' rule for a single line or " +
+                             "'swiftlint:enable' to re-enable the rule immediately after the violations to be ignored"
```
SimplyDanny, line 151: Hm ... these messages are typically short in nature. They succinctly mention what's wrong and sometimes give advice for the fix. The latter is rare and that's okay. For the reasons behind a violation, examples and ideas to fix it, one can have a look into the full documentation.

That said, I think this message is too long.

Source/SwiftLintBuiltInRules/Rules/Lint/BlanketDisableCommandRule.swift
```diff
@@ -4,7 +4,9 @@ struct BlanketDisableCommandRule: Rule, SourceKitFreeRule {
     static let description = RuleDescription(
         identifier: "blanket_disable_command",
         name: "Blanket Disable Command",
-        description: "swiftlint:disable commands should be re-enabled before the end of the file",
+        description: "'swiftlint:disable' commands should use 'next', 'this' or 'previous' to disable rules for a " +
+                     "single line, or 'swiftlint:enable' to re-enable the rules immediately after the violations " +
+                     "to be ignored",
```
SimplyDanny, line 9: That's what's dumped into the documentation. Here, you can elaborate. You may use multiple sentences and explain the rule in more detail.

Since that's going to be interpreted as Markdown, you may put code into backticks.

Source/SwiftLintBuiltInRules/Rules/Lint/BlanketDisableCommandRule.swift
mildm8nnered: So I've managed to trim it back to 132 characters plus the rule name (from 165 before).

`"Use 'next', 'this' or 'previous' to disable the 'colon' rule for a single line or 'swiftlint:enable' to re-enable it as soon as possible"`

Which is long, but not the longest reason I could find.

Even shorter would be:

`"Use 'next', 'this' or 'previous' to disable the 'colon' rule for one line or 'swiftlint:enable' to re-enable it as soon as possible"`

Happy to go back and forth on this for a bit to get it right. I'm trying to avoid saying anything about the "rest of the file" or the "end of the file".

SimplyDanny: I like these more. Here is another shorter one:

`"Use 'next', 'this' or 'previous' instead to disable the 'colon' rule once or re-enable it as soon as possible"`

I think it's mentioning the possibilities. Details can be taken from the description.

Source/SwiftLintBuiltInRules/Rules/Lint/BlanketDisableCommandRule.swift
```diff
@@ -4,7 +4,9 @@ struct BlanketDisableCommandRule: Rule, SourceKitFreeRule {
     static let description = RuleDescription(
         identifier: "blanket_disable_command",
         name: "Blanket Disable Command",
-        description: "swiftlint:disable commands should be re-enabled before the end of the file",
+        description: "`swiftlint:disable` commands should use `next`, `this` or `previous` to disable rules for a " +
+                     "single line, or `swiftlint:enable` to re-enable the rules immediately after the violations " +
+                     "to be ignored",
```
SimplyDanny, line 9:
```suggestion
        description: """
                     `swiftlint:disable` commands should use `next`, `this` or `previous` to disable rules for a
                     single line only, or `swiftlint:enable` to re-enable the rules immediately after the violations
                     to be ignored.
                     """,
```

SimplyDanny: Perhaps mention that disabling a rule for the rest of a file is not encouraged?

mildm8nnered:
Adopted your suggestions pretty much whole-sale - definitely looks much better now

Source/SwiftLintBuiltInRules/Rules/Lint/BlanketDisableCommandRule.swift
```diff
@@ -144,8 +148,9 @@ struct BlanketDisableCommandRule: Rule, SourceKitFreeRule {
             }
 
             if let command = ruleIdentifierToCommandMap[disabledRuleIdentifier] {
-                let reason = "The disabled '\(disabledRuleIdentifier.stringRepresentation)' rule " +
-                             "should be re-enabled before the end of the file"
+                let reason = "Use 'next', 'this' or 'previous' instead to disable the " +
+                              "'\(disabledRuleIdentifier.stringRepresentation)' rule once " +
+                              " or re-enable it as soon as possible"
```
SimplyDanny, line 153: Might be easier to read with a text block.

Source/SwiftLintBuiltInRules/Rules/Lint/BlanketDisableCommandRule.swift
```diff
@@ -4,7 +4,11 @@ struct BlanketDisableCommandRule: Rule, SourceKitFreeRule {
     static let description = RuleDescription(
         identifier: "blanket_disable_command",
         name: "Blanket Disable Command",
-        description: "swiftlint:disable commands should be re-enabled before the end of the file",
+        description: """
+                     `swiftlint:disable` commands should use `next`, `this` or `previous` to disable rules for a \
+                     single line, or `swiftlint:enable` to re-enable the rules immediately after the violations \
+                     to be ignored, instead of disabling the rule for the rest of the file
```
SimplyDanny, line 10:
```suggestion
                     to be ignored, instead of disabling the rule for the rest of the file.
```

SimplyDanny (approved):

mildm8nnered:
Ah - I think maybe that period should not be on the end of the reason https://github.com/realm/SwiftLint/pull/5518/commits/7039fba188cff7cdab5c24f7b036dd39e82e5263

https://github.com/realm/SwiftLint/pull/5475 is now failing CI with 

```
Test Suite 'BaselineTests' started at 2024-04-04 17:49:36.549.
Test Case '-[SwiftLintFrameworkTests.BaselineTests testNewViolation]' started.
Reasons shall not end with a period. Got "`swiftlint:disable` commands should use `next`, `this` or `previous` to disable rules for a single line, or `swiftlint:enable` to re-enable the rules immediately after the violations to be ignored, instead of disabling the rule for the rest of the file.". Either rewrite the rule's description or set a custom reason in the StyleViolation's constructor.: file StyleViolation.swift, line 46
```

mildm8nnered:
Hmm. Looks like some descriptions do have a period, and some don't. I think that test is failing because I'm creating a BlanketDisableCommand violation without a specific reason in the test, and it's picking up the description - I can probably work around that.

SimplyDanny:
Ups ... 😬 There are a few things to explain here.

One is that without a specific reason for a violation, the description is used. At the same time, the description appears in the documentation for every rule.

It is reasonable to prefer complete sentences in descriptions and explanations, but we also once agreed on violation messages to be as short as possible and without a full stop to mimic what other tools usually do.

Combining both, there is an unfortunate conflict that can only be resolved by using explicit reasons everywhere or enhance every rule description with a default reason distinct from the description.

That said, sorry for breaking tests. I didn't notice there are violations without their own reason.

mildm8nnered:
> Ups ... 😬 There are a few things to explain here.
> 
> One is that without a specific reason for a violation, the description is used. At the same time, the description appears in the documentation for every rule.
> 
> It is reasonable to prefer complete sentences in descriptions and explanations, but we also once agreed on violation messages to be as short as possible and without a full stop to mimic what other tools usually do.
> 
> Combining both, there is an unfortunate conflict that can only be resolved by using explicit reasons everywhere or enhance every rule description with a default reason distinct from the description.
> 
> That said, sorry for breaking tests. I didn't notice there are violations without their own reason.

no worries - I just switched to a different rule for my tests for now.

I think this particular failure is just because of the way I was using the rule in my tests - users would never see the description as a reason in this case.

PR #5521, opened 2024-04-01
Add option for `opening_brace` rule not to trigger with multiline statements

### LINKED ISSUE #3720: Update `opening_brace` for multiline if statements
### New Issue Checklist

### Rule update request

Similar to #3068 I'd like to propose a change to `opening_brace` to allow the following format:

```swift
if let child = parent.children.first,
    let grandchild = child.children.first,
    let greatGrandchild = grandchild.children.first
{
    // Some code
}
```

or 
```swift
if 
    let child = parent.children.first,
    let grandchild = child.children.first,
    let greatGrandchild = grandchild.children.first
{
    // Some code
}
```

Currently the autocorrect will fix it to the following which is difficult to read:

```swift
if 
    let child = parent.children.first,
    let grandchild = child.children.first,
    let greatGrandchild = grandchild.children.first {
    // Some code
}
```

leonardosrodrigues0:
In case of multiline `if` and `while` statements, many adopt the style of putting the opening brace in a new line. This change add a new option to the rule to allow that style.

Closes #3720

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/OpeningBraceConfiguration.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/OpeningBraceConfiguration.swift
@@ -6,10 +6,18 @@ struct OpeningBraceConfiguration: SeverityBasedRuleConfiguration {
 
     @ConfigurationElement(key: "severity")
     private(set) var severityConfiguration = SeverityConfiguration<Parent>(.warning)
-    @ConfigurationElement(key: "allow_multiline_func")
-    private(set) var allowMultilineFunc = false
     @ConfigurationElement(key: "ignore_multiline_type_headers")
     private(set) var ignoreMultilineTypeHeaders = false
     @ConfigurationElement(key: "ignore_multiline_statement_conditions")
     private(set) var ignoreMultilineStatementConditions = false
+    @ConfigurationElement(key: "ignore_multiline_function_signatures")
+    private(set) var ignoreMultilineFunctionSignatures = false
+    // TODO: [08/23/2026] Remove deprecation warning after ~2 years.
+    @ConfigurationElement(key: "allow_multiline_func", deprecationNotice: .suggestAlternative(
+        ruleID: Parent.identifier, name: "ignore_multiline_function_signatures"))
+    private(set) var allowMultilineFunc = false
+
+    var shouldIgnoreMultilineFunctionSignatures: Bool {
+        ignoreMultilineFunctionSignatures || allowMultilineFunc
+    }
 }
--- a/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/OpeningBraceConfiguration.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/OpeningBraceConfiguration.swift
@@ -8,4 +8,6 @@ struct OpeningBraceConfiguration: SeverityBasedRuleConfiguration {
     private(set) var severityConfiguration = SeverityConfiguration<Parent>(.warning)
     @ConfigurationElement(key: "allow_multiline_func")
     private(set) var allowMultilineFunc = false
+    @ConfigurationElement(key: "ignore_multiline_type_headers")
+    private(set) var ignoreMultilineTypeHeaders = false
 }
--- a/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/OpeningBraceConfiguration.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/OpeningBraceConfiguration.swift
@@ -10,4 +10,6 @@ struct OpeningBraceConfiguration: SeverityBasedRuleConfiguration {
     private(set) var allowMultilineFunc = false
     @ConfigurationElement(key: "ignore_multiline_type_headers")
     private(set) var ignoreMultilineTypeHeaders = false
+    @ConfigurationElement(key: "ignore_multiline_statement_conditions")
+    private(set) var ignoreMultilineStatementConditions = false
 }
--- a/Source/SwiftLintBuiltInRules/Rules/Style/OpeningBraceRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/OpeningBraceRule.swift
@@ -35,7 +35,7 @@ private extension OpeningBraceRule {
                 return
             }
 
-            collectViolations(for: node.memberBlock)
+            super.visitPost(node)
         }
 
         override func visitPost(_ node: ClassDeclSyntax) {
@@ -46,7 +46,7 @@ private extension OpeningBraceRule {
                 return
             }
 
-            collectViolations(for: node.memberBlock)
+            super.visitPost(node)
         }
 
         override func visitPost(_ node: EnumDeclSyntax) {
@@ -57,7 +57,7 @@ private extension OpeningBraceRule {
                 return
             }
 
-            collectViolations(for: node.memberBlock)
+            super.visitPost(node)
         }
 
         override func visitPost(_ node: ExtensionDeclSyntax) {
@@ -68,7 +68,7 @@ private extension OpeningBraceRule {
                 return
             }
 
-            collectViolations(for: node.memberBlock)
+            super.visitPost(node)
         }
 
         override func visitPost(_ node: ProtocolDeclSyntax) {
@@ -79,7 +79,7 @@ private extension OpeningBraceRule {
                 return
             }
 
-            collectViolations(for: node.memberBlock)
+            super.visitPost(node)
         }
 
         override func visitPost(_ node: StructDeclSyntax) {
@@ -90,8 +90,44 @@ private extension OpeningBraceRule {
                 return
             }
 
-            collectViolations(for: node.memberBlock)
+            super.visitPost(node)
         }
+
+        // MARK: - Conditional Statements
+
+        override func visitPost(_ node: ForStmtSyntax) {
+            if
+                configuration.ignoreMultilineStatementConditions,
+                hasMultilinePredecessors(node.body, keyword: node.forKeyword)
+            {
+                return
+            }
+
+            super.visitPost(node)
+        }
+
+        override func visitPost(_ node: IfExprSyntax) {
+            if
+                configuration.ignoreMultilineStatementConditions,
+                hasMultilinePredecessors(node.body, keyword: node.ifKeyword)
+            {
+                return
+            }
+
+            super.visitPost(node)
+        }
+
+        override func visitPost(_ node: WhileStmtSyntax) {
+            if
+                configuration.ignoreMultilineStatementConditions,
+                hasMultilinePredecessors(node.body, keyword: node.whileKeyword)
+            {
+                return
+            }
+
+            super.visitPost(node)
+        }
+
 
         // MARK: - Functions and Initializers
 
@@ -102,7 +138,7 @@ private extension OpeningBraceRule {
             if configuration.allowMultilineFunc, hasMultilinePredecessors(body, keyword: node.funcKeyword) {
                 return
             }
-            collectViolations(for: body)
+            super.visitPost(node)
         }
 
         override func visitPost(_ node: InitializerDeclSyntax) {
@@ -112,7 +148,7 @@ private extension OpeningBraceRule {
             if configuration.allowMultilineFunc, hasMultilinePredecessors(body, keyword: node.initKeyword) {
                 return
             }
-            collectViolations(for: body)
+            super.visitPost(node)
         }
 
         // MARK: - Auxiliar
--- a/Source/SwiftLintBuiltInRules/Rules/Style/OpeningBraceRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/OpeningBraceRule.swift
@@ -28,10 +28,8 @@ private extension OpeningBraceRule {
         // MARK: - Type Declarations
 
         override func visitPost(_ node: ActorDeclSyntax) {
-            if
-                configuration.ignoreMultilineTypeHeaders,
-                hasMultilinePredecessors(node.memberBlock, keyword: node.actorKeyword)
-            {
+            if configuration.ignoreMultilineTypeHeaders,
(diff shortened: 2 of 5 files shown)


leonardosrodrigues0:
@SimplyDanny Could you take a look at this when you have some time? It would benefit many that are used to adopting this style.

I left the new option on by default just to check with the OSS projects, all the changes are the expected fixed violations.

SimplyDanny (changes requested):
I wonder why this new option does not apply likewise to `guard` and `for` (and maybe other statements). Any reason for that?

> I left the new option on by default just to check with the OSS projects, all the changes are the expected fixed violations.

Very good! That helps a lot in assessment.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/OpeningBraceConfiguration.swift
```diff
@@ -8,4 +8,6 @@ struct OpeningBraceConfiguration: SeverityBasedRuleConfiguration {
     private(set) var severityConfiguration = SeverityConfiguration<Parent>(.warning)
     @ConfigurationElement(key: "allow_multiline_func")
     private(set) var allowMultilineFunc = false
+    @ConfigurationElement(key: "allow_multiline_statement")
+    private(set) var allowMultilineStatement = true
```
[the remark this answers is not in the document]
leonardosrodrigues0: Agreed. Just followed the pattern from the existing option, but should've thought about it. Changed to `ignore_multiline_condition_statements`, what do you think?

leonardosrodrigues0:
> I wonder why this new option does not apply likewise to `guard` and `for` (and maybe other statements). Any reason for that?

From the issue and my experience, the main focus would be `if`, `while` and `guard`, but `guard` has the else keyword that is used in the same line as the opening brace in this type of cases:

```swift
guard
    let child = parent.children.first,
    let grandchild = child.children.first,
    let greatGrandchild = grandchild.children.first
else {
    // Some code
}
```

I did, however, take a look at the OSS differences when we ignore the rule for multiline "predecessors of the body" for all statements affected by the rule. I checked all differences and it seems that this style is frequently adopted in type declarations (and extensions), as in:

```swift
extension RawRepresentable
where
    Self: AtomicOptionalRepresentable,
    RawValue: AtomicOptionalRepresentable
{
    // Some code
}
```

Considering this, I think that the new direction should be to let the option ignore all statements that are "multiline before the opening brace" (except for single keywords like `do` and `defer`). And if we go this way, I don't see a reason why we would leave functions and initializers (that are already ignored by the `allowMultilineFunc` option) with an independent option of all other statements.

Changing the existing option name and make it affect all statements would be a breaking change, and I don't have experience on how these are made in this project. Would we still support `allowMultilineFunc` but deprecate it?

Additionally, the new option name could be something like `ignoreMultilineBracePredecessors`, `ignoreMultilineOpeningBracePredecessors` or `ignoreMultilineBodyPredecessors`. I'm not sure if "predecessors" fits well here, but it is the best description I could create. On the other hand, `ignoreMultilineStatements` would be simpler but imprecise, as you pointed out.

SimplyDanny:
Thank you for the write-up! So the topic applies to functions/initializers (already supported), types and statements (implemented by this PR). 

I think that an existing option shouldn't suddenly support much more than it was intended for. But we may rename `allow_multiline_func` to something better matching while still supporting the current name for the time being.

Statements like `if`, `while`, `guard` and `for` should be associated with a separate option.

Types might be yet another category. We could merge them with the control-flow statements from the previous group (as both groups will be newly introduced). However, this doesn't really match, which would already be problematic when looking for an option name.

Talking about names: How about `ignoreMultilineFunctionSignatures`, `ignoreMultilineStatementConditions` and `ignoreMultilineTypeHeaders`?

leonardosrodrigues0:
I like the functions/initializers, types and statements separation and the name for the configuration options, I'll implement that.

> But we may rename `allow_multiline_func` to something better matching while still supporting the current name for the time being.

How would we do that? I found this example being removed in #5107, should we mark the option as deprecated like this?

```swift
mutating func apply(configuration: Any) throws {
    // ...
    } else if let statementLevelConfiguration = configurationDict["statement_level"] {
        queuedPrintError(
            """
            warning: 'statement_level' has been renamed to 'function_level' and will be completely removed \
            in a future release.
            """
        )
        try functionLevel.apply(configuration: statementLevelConfiguration)
    }
    // ...
}
```

SimplyDanny:
> How would we do that? I found this example being removed in https://github.com/realm/SwiftLint/pull/5107, should we mark the option as deprecated like this?

The new macro approach doesn't support deprecation warnings well yet. I'm working on it though. For our PR, just leave the existing option as is for now. I'll take care of the rename once I found a good solution.

SimplyDanny:
> > How would we do that? I found this example being removed in #5107, should we mark the option as deprecated like this?
> 
> The new macro approach doesn't support deprecation warnings well yet. I'm working on it though. For our PR, just leave the existing option as is for now. I'll take care of the rename once I found a good solution.

I've added a way to mark options as deprecated in #5540. Just in case you want to handle the rename here as well ... 😉

Source/SwiftLintBuiltInRules/Rules/Style/OpeningBraceRule.swift
```diff
@@ -206,14 +246,33 @@ private extension OpeningBraceRule {
...
-        private func refersToMultilineFunction(_ body: CodeBlockSyntax, functionIndicator: TokenSyntax) -> Bool {
+        private func isIgnoredMultilineFunc(_ body: CodeBlockSyntax, keywordToken: TokenSyntax) -> Bool {
+            guard configuration.allowMultilineFunc else {
+                return false
+            }
+
+            return isBodyPredecessorMultiline(body, keywordToken: keywordToken)
+        }
+
+        private func isIgnoredMultilineConditionStatement(_ body: some BracedSyntax, keywordToken: TokenSyntax) -> Bool {
```
roksollc, line 257: Since this function determines whether the `violationCorrection` function _should_ be called/processed, and in consideration of other suggestion, I believe something similar to the following would better reflect this function's intention/usage:
- `shouldIgnoreFromMultilineStatementConditions`

Source/SwiftLintBuiltInRules/Rules/Style/OpeningBraceRule.swift
```diff
@@ -37,6 +43,9 @@ private extension OpeningBraceRule {
 
         override func visitPost(_ node: EnumDeclSyntax) {
             let body = node.memberBlock
+            if isIgnoredMultilineConditionStatement(body, keywordToken: node.enumKeyword) {
+                return
+            }
```
roksollc, line 48: An _enum_ definition could have multiline protocol adherence, but I"m not aware of a condition list. So, is this needed? 

The same question applies to all other applications of this logic that do not have an associated condition list that precedes the opening brace.

Source/SwiftLintBuiltInRules/Rules/Style/OpeningBraceRule.swift
```diff
@@ -224,18 +283,24 @@ private extension BracedSyntax {
...
-            }
+        guard let previousToken = leftBrace.previousToken(viewMode: .sourceAccurate) else {
+            return nil
         }
+
+        let previousLocation = previousToken.endLocation(converter: locationConverter)
+        let leftBraceLocation = leftBrace.startLocation(converter: locationConverter)
+
+        let leftBraceIsInSameLine = previousLocation.line == leftBraceLocation.line
+        let leftBraceHasOneLeadingSpace = previousLocation.column + 1 == leftBraceLocation.column
```
roksollc, line 294:
```suggestion
        let isOpeningBraceOnSameLine = previousLocation.line == leftBraceLocation.line
        let doesOpeningBraceHaveOneLeadingSpace = previousLocation.column + 1 == leftBraceLocation.column
```

roksollc (commented):
Hello everyone! I've never contributed here, but my team would also like to see these changes. Great idea! While looking at the proposed solution, I added a couple of suggestions / questions. Thank you @leonardosrodrigues0 for your efforts!

SimplyDanny:
@leonardosrodrigues0: Do you fancy to continue with this PR?

leonardosrodrigues0:
Hi @SimplyDanny, I apologize for the long absence, I want to continue with this from now on. Thanks @roksollc for your POV as well.

I saw recent changes with the addition of `brace_on_new_line`, it's a great new option in my opinion. Before diving into the implementation, I'm considering how this option would interact with the "ignore" ones. It seems to me that if `brace_on_new_line` is enabled, then the "ignore" options shouldn't be considered: being multiline or not, the opening brace should be aligned with the parent's start column and one line after the previous location. This is different from the current interaction, in which `allow_multiline_func` works even with `brace_on_new_line`.

This made me think if these new options should actually enforce the opening brace position similar to `brace_on_new_line` instead of simply ignoring the rule for those multiline cases. My first conclusion, however, is that I'm overcomplicating this a bit, and ignoring should be fine. Let me know if you guys disagree.

Source/SwiftLintBuiltInRules/Rules/Style/OpeningBraceRule.swift
leonardosrodrigues0: The last version of the code I pushed applied this change for all syntax options so that we could see the differences in some open-source projects in the report that SwiftLintBot provides us. I had a long absence from this PR and didn't push a new version nor notified about this, sorry about that. Thanks for your contribution!

SimplyDanny:
> Hi @SimplyDanny, I apologize for the long absence, I want to continue with this from now on. Thanks @roksollc for your POV as well.

Nothing to apologize for. Glad you are back!

> I saw recent changes with the addition of `brace_on_new_line`, it's a great new option in my opinion. Before diving into the implementation, I'm considering how this option would interact with the "ignore" ones. It seems to me that if `brace_on_new_line` is enabled, then the "ignore" options shouldn't be considered: being multiline or not, the opening brace should be aligned with the parent's start column and one line after the previous location. This is different from the current interaction, in which `allow_multiline_func` works even with `brace_on_new_line`.
> 
> This made me think if these new options should actually enforce the opening brace position similar to `brace_on_new_line` instead of simply ignoring the rule for those multiline cases. My first conclusion, however, is that I'm overcomplicating this a bit, and ignoring should be fine. Let me know if you guys disagree.

I see your point. With the introduction of `brace_on_new_line` this gets slightly confusing. We basically have three modes:

1. Braces always on the same line with the declaration.
2. Braces always on a single line after the declaration disregarding other options.
3. Braces always on the same line with the declaration with "don't care" exceptions.

The "don't care", i.e. no real enforcement of a certain style, is a bit strange, but it's like it has always been. `brace_on_new_line` ignoring all other options also isn't too nice. Probably, `brace_on_new_line` should be its own rule to avoid these quirks.

SimplyDanny:
> Probably, `brace_on_new_line` should be its own rule to avoid these quirks.

That will be done with #5723. So you don't have to worry your brain on `brace_on_new_line` when continuing here, @leonardosrodrigues0. 😉

leonardosrodrigues0:
> That will be done with #5723. So you don't have to worry your brain on `brace_on_new_line` when continuing here, @leonardosrodrigues0. 😉

@SimplyDanny thank you!!

I just pushed a new version with the two new options. Also renamed the old `allow_multiline_func` one to `ignore_multiline_function_signatures` and marked the former as deprecated. I added a computed variable `shouldIgnoreMultilineFunctionSignatures` to keep supporting the original option for now, let me know in case that isn't the best approach.

Once again, the new options are set to `true` just to check the changes in the OSS projects, which I did, and all the differences are expected.

Source/SwiftLintBuiltInRules/Rules/Style/OpeningBraceRule.swift
```diff
@@ -25,31 +25,123 @@ struct OpeningBraceRule: SwiftSyntaxCorrectableRule {
...
+
+            super.visitPost(node)
+        }
+
+        // MARK: - Other Methods
+
+        /// Checks if a `BracedSyntax` has a multiline predecessor.
+        /// For type declarations, the predecessor is the header. For conditional statements,
+        /// it is the condition list, and for functions, it is the signature.
+        private func hasMultilinePredecessors(_ body: some BracedSyntax, keyword: TokenSyntax) -> Bool {
```
SimplyDanny, line 140: The name `Predecessors` is not too obvious. But I cannot think of anything better and the method is documented very well. So that's fine. 👍

SimplyDanny (changes requested):
Looks good. Well done!

What about `guard`? Any reason why it's ignored?

leonardosrodrigues0:
> Looks good. Well done!
> 
> What about `guard`? Any reason why it's ignored?

Thanks! Yes, `guard`'s opening braces are always preceded by `else`, so this rule doesn't affect those who adopt this style for multiline statements. Borrowing one of the examples in the issue:

```swift
guard
    let child = parent.children.first,
    let grandchild = child.children.first,
    let greatGrandchild = grandchild.children.first
else {
    // Some code
}
```

Source/SwiftLintBuiltInRules/Rules/Style/OpeningBraceRule.swift
leonardosrodrigues0: I agree, same feeling here... I'm not happy with the name but couldn't describe it better.

SimplyDanny (approved):
Looks good from my side. Ready to merge? Or is there anything you'd like to adapt?

leonardosrodrigues0:
Ready to merge! Thanks for the reviews and the patience with the delay 😅

pyrtsa:
> This made me think if these new options should actually enforce the opening brace position similar to `brace_on_new_line` instead of simply ignoring the rule for those multiline cases.

I think what we'd need is another `brace_on_new_line_after_multiline_statement` option to `opening_brace` rule.

SimplyDanny:
> > This made me think if these new options should actually enforce the opening brace position similar to `brace_on_new_line` instead of simply ignoring the rule for those multiline cases.
> 
> I think what we'd need is another `brace_on_new_line_after_multiline_statement` option to `opening_brace` rule.

Please open a ticket to discuss this, @pyrtsa.

PR #5523, opened 2024-04-02
Adjust case of major and minor severity for CodeClimate reporting

waitbutY:
Currently Gitlab rejects Swiftlint CodeClimate reports because the `MAJOR` and `MINOR` severity strings are all caps.  Changing these strings to lower case allows developers to submit Swiftlint CodeClimate json files directly to Gitlab to track code quality issues without manually manipulating the file.

See manual `grep` fix here: https://forum.gitlab.com/t/code-quality-widget-doesnt-load/90442

### DIFF
--- a/Tests/SwiftLintFrameworkTests/ReporterTests.swift
+++ b/Tests/SwiftLintFrameworkTests/ReporterTests.swift
@@ -92,12 +92,8 @@ class ReporterTests: SwiftLintTestCase {
     }
 
     func testCodeClimateReporter() {
-        let expectedOutput = stringFromFile("CannedCodeClimateReporterOutput.json").filter {
-            !$0.isWhitespace
-        }
-        let result = CodeClimateReporter.generateReport(generateViolations()).filter {
-            !$0.isWhitespace
-        }
+        let expectedOutput = stringFromFile("CannedCodeClimateReporterOutput.json")
+        let result = CodeClimateReporter.generateReport(generateViolations())
         XCTAssertEqual(result, expectedOutput)
     }
 
--- a/Tests/SwiftLintFrameworkTests/Resources/CannedCodeClimateReporterOutput.json
+++ b/Tests/SwiftLintFrameworkTests/Resources/CannedCodeClimateReporterOutput.json
@@ -1,63 +1,62 @@
 [
-    {
-      "check_name" : "Line Length",
-      "description" : "Violation Reason",
-      "engine_name" : "SwiftLint",
-      "fingerprint" : "917a85854a9500cfd28520fa4875a3ecbba171e522f13452c7adec71fc14497a",
-      "location" : {
-        "lines" : {
-          "begin" : 1,
-          "end" : 1
-        },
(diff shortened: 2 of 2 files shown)


SimplyDanny (approved):
Thanks! While the all-uppercase version is obviously incorrect according to the spec, I wonder how this has went unnoticed for such a long time. 🤔 

Please add an entry to the CHANGELOG to let people know about the change. Potentially even in the "Breaking" section just in case folks are relying on the previous version in some context.

waitbutY:
@SimplyDanny CHANGELOG entry added, and automated tests fixed!

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@
   Rename SwiftLintBuildToolPlugin.
   Add Swift Package Manager installation instructions.  
   [garricn](https://github.com/garricn)
+  
+* Adjusting CodeClimate Reporter output to have lower case severity
+  values for use in Gitlab and other CodeClimate compatible tools can 
+  import the report without modification.
+  [waitButY](https://github.com/waitbutY)
```
SimplyDanny, line 17: Please append two spaces to the end of the description. They are required to enforce a line break in the rendered Markdown. Also, the important aspect is compliance with the spec:
```suggestion
* Fix Code Climate reporter output by having lower case severity
  values to comply with the Code Climate specification.  
  [waitButY](https://github.com/waitbutY)
```

Tests/SwiftLintFrameworkTests/ReporterTests.swift
```diff
@@ -92,8 +92,12 @@ class ReporterTests: SwiftLintTestCase {
...
 
     func testCodeClimateReporter() {
-        let expectedOutput = stringFromFile("CannedCodeClimateReporterOutput.json")
-        let result = CodeClimateReporter.generateReport(generateViolations())
+        let expectedOutput = stringFromFile("CannedCodeClimateReporterOutput.json").filter {
+            !$0.isWhitespace
+        }
+        let result = CodeClimateReporter.generateReport(generateViolations()).filter {
+            !$0.isWhitespace
+        }
```
SimplyDanny, line 100: Why is the filtering suddenly necessary? It's just a few strings that have changed their case and need an update in the reference file.

SimplyDanny (changes requested):

Tests/SwiftLintFrameworkTests/ReporterTests.swift
waitbutY: I am not sure why i was struggling so hard to get all the whitespace to match, ill give it another shot

waitbutY: Got it! resolved, removed these whitespace filters

SimplyDanny (approved):

SimplyDanny:
As easy as it can be. Thank you!

waitbutY:
> As easy as it can be. Thank you!

loved having the chance to contribute a little!

Issue #5461, opened 2024-02-07
Adding a TypeContent case of `type_contents_order`


PangYenChen:
- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

Currently the TypeContent of `type_contents_order` is:

```
enum TypeContent: String {
    case `case` = "case"
    case typeAlias = "type_alias"
    case associatedType = "associated_type"
    case subtype = "subtype"
    case typeProperty = "type_property"
    case instanceProperty = "instance_property"
    case ibOutlet = "ib_outlet"
    case ibInspectable = "ib_inspectable"
    case initializer = "initializer"
    case typeMethod = "type_method"
    case viewLifeCycleMethod = "view_life_cycle_method"
    case ibAction = "ib_action"
    case otherMethod = "other_method"
    case `subscript` = "subscript"
    case deinitializer = "deinitializer"
}
```

It distinguishes between ib_outlet and instance_property, ib_action and other_method.

Would it be nice to have ib_segue_action as a case?

dk-talks:
Hi
Do we have to add a new case for 'ib_segue_action'. If so then I am not able to figure out what else is needed to change.
I have added an image for the changes I could figure out to make.
Thanks
<img width="731" alt="Screenshot 2024-03-31 at 1 46 40 AM" src="https://github.com/realm/SwiftLint/assets/65244911/bdf40f4d-7e5d-4747-a544-69cc322893e4">

PangYenChen:
The change in the image is enough. 

By the way, I don't know the proper order of ibSegueAction in the default configuration.

SimplyDanny:
[...]

PR #5524, opened 2024-04-02
#5461 added ib_segue_action case in TypeContent

dk-talks:
Hey! I have added a new case to TypeContent enum for ib_segue_action. And also adapted the change in [TypeContentsOrderRule.swift](https://github.com/realm/SwiftLint/blob/main/Source/SwiftLintBuiltInRules/Rules/Style/TypeContentsOrderRule.swift).

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/TypeContentsOrderConfiguration.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/TypeContentsOrderConfiguration.swift
@@ -17,6 +17,7 @@ enum TypeContent: String {
     case otherMethod = "other_method"
     case `subscript` = "subscript"
     case deinitializer = "deinitializer"
+    case ibSegueAction = "ib_segue_action"
 }
 
 @AutoConfigParser
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@
 * Support linting only provided file paths with command plugins.  
   [DanSkeel](https://github.com/DanSkeel)
 
+* Add new category for `@IBSegueAction` to `type_contents_order` rule.  
+  [dk-talks](https://github.com/dk-talks)  
+  [SimplyDanny](https://github.com/SimplyDanny)
+
 #### Bug Fixes
 
 * Ignore super calls with trailing closures in `unneeded_override` rule.  
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/UnusedDeclarationRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/UnusedDeclarationRule.swift
@@ -211,6 +211,7 @@ private extension SwiftLintFile {
 
     private func shouldIgnoreEntity(_ indexEntity: SourceKittenDictionary, relatedUSRsToSkip: Set<String>) -> Bool {
         let declarationAttributesToSkip: Set<SwiftDeclarationAttributeKind> = [
+            .ibsegueaction,
             .ibaction,
             .main,
             .nsApplicationMain,
--- a/Source/SwiftLintBuiltInRules/Rules/Style/TypeContentsOrderRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/TypeContentsOrderRule.swift
@@ -133,6 +133,9 @@ struct TypeContentsOrderRule: OptInRule {
             if typeContentStructure.enclosedSwiftAttributes.contains(SwiftDeclarationAttributeKind.ibaction) {
                 return .ibAction
             }
+            if typeContentStructure.enclosedSwiftAttributes.contains(SwiftDeclarationAttributeKind.ibsegueaction) {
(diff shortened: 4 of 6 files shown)


Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/TypeContentsOrderConfiguration.swift
```diff
@@ -41,5 +42,6 @@ struct TypeContentsOrderConfiguration: SeverityBasedRuleConfiguration {
         [.otherMethod],
         [.subscript],
         [.deinitializer]
```
SimplyDanny, line 44: @dk-talks: There is a comma missing here.

SimplyDanny (changes requested):

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/TypeContentsOrderConfiguration.swift
```diff
@@ -40,6 +41,7 @@ struct TypeContentsOrderConfiguration: SeverityBasedRuleConfiguration {
         [.ibAction],
         [.otherMethod],
         [.subscript],
-        [.deinitializer]
+        [.deinitializer],
+        [.ibSegueAction]
```
SimplyDanny, line 45: This needs to be removed. We try to avoid changing the defaults subsequently if possible.

SimplyDanny (changes requested):
Thanks! This needs some tests. See `TypeContentsOrderRuleTests`.

Also, you need to update the test reference for `IntegrationTests.testDefaultConfigurations` to make the build pass.

dk-talks:
Hi, @SimplyDanny still the build is not passing, is there something else needed to be done?

SimplyDanny:
> Hi, @SimplyDanny still the build is not passing, is there something else needed to be done?

Now that you have removed the entry from the default list, the test reference update needs to be reverted.

dk-talks:
Hey @SimplyDanny. Now that the build has passed, is there anything else needs to be done or is it enough to merge and close the issue?

SimplyDanny:
> Hey @SimplyDanny. Now that the build has passed, is there anything else needs to be done or is it enough to merge and close the issue?

1. This change needs tests. See `TypeContentsOrderRuleTests`.
2. A changelog entry needs to be added. See `CHANGELOG.md`.

SimplyDanny (approved):

guillaumealgis:
Hi, thanks for adding this new case to the `type_contents_order` rule :)

It seems [the documentation](https://realm.github.io/SwiftLint/type_contents_order.html) has not been updated, and I could not find where `ib_segue_action` is located in [the default configuration for this rule](https://github.com/realm/SwiftLint/blob/f045130e3e8e5bcaf09cc336b7b947d4bdc4b7c1/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/TypeContentsOrderConfiguration.swift#L30), making it difficult to know where SwiftLint expects me to put my `@IBSegueAction` methods 😕

SimplyDanny:
> Hi, thanks for adding this new case to the `type_contents_order` rule :)
> 
> It seems [the documentation](https://realm.github.io/SwiftLint/type_contents_order.html) has not been updated, and I could not find where `ib_segue_action` is located in [the default configuration for this rule](https://github.com/realm/SwiftLint/blob/f045130e3e8e5bcaf09cc336b7b947d4bdc4b7c1/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/TypeContentsOrderConfiguration.swift#L30), making it difficult to know where SwiftLint expects me to put my `@IBSegueAction` methods 😕
> 
> EDIT: It seems that putting the `@IBSegueAction` methods at the end of the enclosing type's implementation makes SwiftLint happy.

We intentionally didn't specify a default expectation because the new type was added on top of the already existing ones. And what's not explicitly specified is expected last by the rule.

@guillaumealgis: Do you think a default would be reasonable (also for documentation purposes) instead of leaving it unspecified? If so, what would be a decent order? Is putting it on the same level as `@IBAction` a good idea?

guillaumealgis:
> Do you think a default would be reasonable (also for documentation purposes) instead of leaving it unspecified? 

I think SwiftLint should explicitly define an order for all defined cases. This makes the expected position of the elements clearer for everyone.

I have honestly no strong opinion on where to put `ib_segue_action` in the list. Same level / right after `ib_action` seems fine, between `other_method` and `subscript` would be fine also, end of list is not outrageous 🤷🏼‍♂️

> what's not explicitly specified is expected last by the rule

Seems pretty logical, but it's not documented, so when SwiftLint started complaining after I updated it was not clear where I should put my code to satisfy it :)

SimplyDanny:
> I have honestly no strong opinion on where to put `ib_segue_action` in the list. Same level / right after `ib_action` seems fine, between `other_method` and `subscript` would be fine also, end of list is not outrageous 🤷🏼‍♂️

Same level as `ib_action` should be fine then: #5956.

PR #5527, opened 2024-04-05
Use SwiftSyntax version 600.0.0-prerelease-2024-04-02

SimplyDanny:
We can also switch back to an exact SwiftSyntax version now with the plugins in a separate repository. In fact, using the plugin no dependency to SwiftSyntax is required whatsoever.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -5,13 +5,13 @@ module(
     repo_name = "SwiftLint",
 )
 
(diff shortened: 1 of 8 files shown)


SimplyDanny:
OSS findings look all valid. Most of them stem from a fix that parses `@Sendable` used on closure parameters as an attribute on the argument list as opposed to an attribute with arguments as before.

Other differences are caused by support for new keywords like `__consuming` and typed-throws.

PR #5534, opened 2024-04-18
Fix swift-syntax dependency issue

### LINKED ISSUE #5533: Conflict with other macro package
### New Issue Checklist

### Describe the bug

We are some existing package dependency(eg. swift-testing and swift-power-assert) which has a dependency on swift-syntax.

eg. .from("509.0.0") .from("510.0.1")

But the latest version of SwiftLint has a exact version dependency on "swift-syntax".

eg. SwiftLint 0.54.0 has a exact "509.0.2"

https://github.com/realm/SwiftLint/blob/f17a4f9dfb6a6afb0408426354e4180daaf49cee/Package.swift#L19

This will cause a package resolution failing due to conflicting requirement.

### Solution

Use more flexible requirement of "from" and adapt the diff API internally so that both swift-syntax 509 and 510 will work here.

(eg. #if canImport(SwiftSyntax509) and #if canImport(SwiftSyntax510))

https://github.com/apple/swift-syntax/blob/728e2f6d5f9fd4a8e45eab6e52e86ad1519a7e2a/Sources/SwiftSyntax/Documentation.docc/Macro%20Versioning.md?plain=1#L9-L33

Kyle-Ye:
Close #5533

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -24,7 +24,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.2.1"),
-        .package(url: "https://github.com/apple/swift-syntax.git", exact: "510.0.2"),
+        .package(url: "https://github.com/apple/swift-syntax.git", from: "510.0.2"),
         .package(url: "https://github.com/jpsim/SourceKitten.git", .upToNextMinor(from: "0.35.0")),
         .package(url: "https://github.com/jpsim/Yams.git", from: "5.0.6"),
         .package(url: "https://github.com/scottrhoyt/SwiftyTextTable.git", from: "0.9.0"),


Package.swift
```diff
@@ -24,7 +24,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser.git", .upToNextMinor(from: "1.2.1")),
-        .package(url: "https://github.com/apple/swift-syntax.git", exact: "510.0.0"),
+        .package(url: "https://github.com/apple/swift-syntax.git", "510.0.0"..<"511.0.0"),
```
Kyle-Ye, line 27: We can use "509.0.0"..<"511.0.0" as the swift-syntax doc is suggesting to support a wider consumer.

But since we have already bumped the version requirement in #5479, we can just keep using "510.0.0"..<"511.0.0" here.

SimplyDanny:
It's definitely valid to bring this up. Thank you for that.

Yet, this needs some discussion among the maintainers first, because it's somehow opposed to the way SwiftLint currently handles this dependency.

Package.swift
SimplyDanny: Isn't this the same as `.upToNextMajor(from: "510.0.0")`?

Kyle-Ye: Nope, we prefer to manually to specify the range here. Say we'd like to support multi major version, then it would be "508.0.0"..<"511.0.0".

Use a version range gives us more freedom here.

Kyle-Ye: You can read the articles I linked in the issue for more detail.

SimplyDanny: The thing is that SwiftLint relies on SwiftSyntax heavily and every major release of SwiftSyntax comes with many changes that either cause deprecation warnings or even behavioral changes.

I'm not inclined to analyze reported issues down to the SwiftSyntax version used. Every SwiftLint release could be bound to major SwiftSyntax release up to the next major version to allow for some flexibility. That's why `.upToNextMajor` makes the most sense to me.

Kyle-Ye: > I'm not inclined to analyze reported issues down to the SwiftSyntax version used. Every SwiftLint release could be bound to major SwiftSyntax release up to the next major version to allow for some flexibility. That's why .upToNextMajor makes the most sense to me.

Got your idea here. But as we usually support 2 or 3 major Swift version. I believe it makes sense we support 2 or 3 corresponding SwiftSyntax here.

https://www.pointfree.co/blog/posts/116-being-a-good-citizen-in-the-land-of-swiftsyntax

For example, imagine there is an app called DemoApp. It has 3 macro package (A, B and C) dependency. (use 5.11 to replace 6.0 for simplicity)

Package A is a good citizen and the requirement for SwiftSyntax is "508.0.0"..<"511.0.0" which means it support 5.8, 5.9 and 5.10 release of SwiftSyntax.

Package B is "509.0.0"..<"511.0.0" which means it support 5.9 and 5.10 release of SwiftSyntax.

Package C is "508.0.0"..<"510.0.0" which means it support 5.8 and 5.9 release of SwiftSyntax.

So we'll resolve to use the latest 509.x.x version here.

But if Package A use "upToNextMajor(510.0.0)" here we then have to downgrade Package A to a lower version which use "upToNextMajor(509.0.0)".

So generally I think it would be great if Package A can pay some extra effort to be a good citizen in the land of swiftsyntax.

IMO, supporting multi version of SwiftSyntax is like supporting multi version of Swift. Unless you like to use the latest feature, most of the case the extra maintain effort would be small and trivial.

Kyle-Ye: > The thing is that SwiftLint relies on SwiftSyntax heavily and every major release of SwiftSyntax comes with many changes that either cause deprecation warnings or even behavior changes.
> 
> I'm not inclined to analyze reported issues down to the SwiftSyntax version used. Every SwiftLint release could be bound to a major SwiftSyntax release up to the next major version to allow for some flexibility. That's why `.upToNextMajor` makes the most sense to me.

Got it. If the upstream think maintaining multi version of SwiftSyntax is an issue.

We can bump to use `.upToNextMajor(from: "510.0.0")` or `from: "510.0.0"` here.

SimplyDanny: > We can bump to use .upToNextMajor(from: "510.0.0") or from: "510.0.0" here.

That would be best for now. Actually, SwiftLint is not like a library providing some macros. Its features and especially its performance is closely bound to SwiftSyntax. So ideally, we could keep this dependency in a narrow range.

In the long run, the SPM plugin should move into its own repository so that it can depend on a binary release of SwiftLint and the dependency to SwiftSyntax isn't visible to people using the plugin at all. Unfortunately, providing all products in one `Package.swift` isn't working well at the moment with SPM.

SimplyDanny (approved):

PR #5542, opened 2024-04-24
Build Plugin name was out of date in README

Diggory:
README stated that the Build Plugin name was `SwiftLintBuildToolPlugin` but it is actually `SwiftLintPlugin` - must have been renamed at some point, and docs weren't updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 #### Breaking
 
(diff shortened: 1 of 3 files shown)


Diggory:
Ah!  I see...
You'll need to update the screenshot as well.  Sorry, I can't be bothered to do that!

PR #5547, opened 2024-04-29
Enable rules that currently have no violations

mildm8nnered:
Enables the following rules for SwiftLint itself, which have no current violations:

```
anyobject_protocol
inert_defer
no_fallthrough_only
static_over_final_class
unused_capture_list
```

It's a shame there's no tooling to pick up these cases automatically, but I'm not sure what that would look like.


SimplyDanny (approved):
Great. Thank you!

PR #5548, opened 2024-04-29
Add new `prefer_key_path` rule

SimplyDanny:
This new rule triggers on code like
```swift
list.filter { $0.isValid }
```
which can be replaced by
```swift
list.filter(\.isValid)
```
as of Swift 5.2.

Automatic fixes fail once the argument label for the closure is named. The fix would need to know its name and insert it in the function call. But that's impossible for a syntax-based rule. We could do auto-fixes for simple cases like `map` and `filter` (which represent most of the findings in OSS projects and SwiftLint itself) at least. I'm not sure if partial fixes are a good thing though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,11 @@
   [woxtu](https://github.com/woxtu)  
   [SimplyDanny](https://github.com/SimplyDanny)  
 
+* Add new `prefer_key_path` rule that triggers when a trailing closure on a function 
+  call is only hosting a (chained) member access expression since the closure can be
+  replaced with a key path argument.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+
 * Adds `baseline` and `write_baseline` configuration file settings, equivalent
   to the `--baseline` and `--write-baseline` command line options.  
   [Martin Redington](https://github.com/mildm8nnered)
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -37,6 +37,7 @@ disabled_rules:
   - no_grouping_extension
   - no_magic_numbers
   - one_declaration_per_file
+  - prefer_key_path
   - prefer_nimble
   - prefixed_toplevel_constant
   - required_deinit
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -5,15 +5,15 @@ module(
     repo_name = "SwiftLint",
 )
 
-bazel_dep(name = "apple_support", version = "1.13.0", repo_name = "build_bazel_apple_support")
-bazel_dep(name = "bazel_skylib", version = "1.5.0")
-bazel_dep(name = "platforms", version = "0.0.8")
-bazel_dep(name = "rules_apple", version = "3.3.0", repo_name = "build_bazel_rules_apple")
-bazel_dep(name = "rules_swift", version = "1.16.0", repo_name = "build_bazel_rules_swift")
+bazel_dep(name = "apple_support", version = "1.16.0", repo_name = "build_bazel_apple_support")
+bazel_dep(name = "bazel_skylib", version = "1.7.1")
+bazel_dep(name = "platforms", version = "0.0.10")
+bazel_dep(name = "rules_apple", version = "3.6.0", repo_name = "build_bazel_rules_apple")
+bazel_dep(name = "rules_swift", version = "1.18.0", repo_name = "build_bazel_rules_swift")
 bazel_dep(name = "sourcekitten", version = "0.35.0", repo_name = "com_github_jpsim_sourcekitten")
 bazel_dep(name = "swift-syntax", version = "600.0.0-prerelease-2024-07-24", repo_name = "SwiftSyntax")
 bazel_dep(name = "swift_argument_parser", version = "1.3.1", repo_name = "sourcekitten_com_github_apple_swift_argument_parser")
-bazel_dep(name = "yams", version = "5.0.6", repo_name = "sourcekitten_com_github_jpsim_yams")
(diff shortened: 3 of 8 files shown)


SimplyDanny:
@mildm8nnered: I'd like to hear your opinion. Any more cases or shortcomings you can think of?

mildm8nnered:
> @mildm8nnered: I'd like to hear your opinion. The rule doesn't seem to be so involved. Any more cases or shortcomings you can think of?

Just trying it on my local codebase, which I expect to have a lot of violations, pretty much like the OSS findings.

mildm8nnered:
> > @mildm8nnered: I'd like to hear your opinion. The rule doesn't seem to be so involved. Any more cases or shortcomings you can think of?
> 
> Just trying it on my local codebase, which I expect to have a lot of violations, pretty much like the OSS findings.

So most hits look good, but I think this might be a false alarm (where there is no property named `otherResults`). I adapted this from proprietary code that I saw the hit with - I have not tested this specific example code here, but if it doesn't get a hit I can dig in deeper on my side.

```
            guard let otherResults = results as? [SomeClass] else {
                return
            }

            let dictionary = Dictionary(grouping: ↓otherResults) {
                $0.someProperty
            }
```

SimplyDanny:
> > > @mildm8nnered: I'd like to hear your opinion. The rule doesn't seem to be so involved. Any more cases or shortcomings you can think of?
> > 
> > 
> > Just trying it on my local codebase, which I expect to have a lot of violations, pretty much like the OSS findings.
> 
> So most hits look good, but I think this might be a false alarm (where there is no property named `otherResults`). I adapted this from proprietary code that I saw the hit with - I have not tested this specific example code here, but if it doesn't get a hit I can dig in deeper on my side.
> 
> ```
>             guard let otherResults = results as? [SomeClass] else {
>                 return
>             }
> 
>             let dictionary = Dictionary(grouping: ↓otherResults) {
>                 $0.someProperty
>             }
> ```

The rule would complain about the closure and suggest `Dictionary(grouping: otherResult, by: \.someProperty)`. Looks okay to me.

CHANGELOG.md
```diff
@@ -61,6 +61,11 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#70](https://github.com/realm/SwiftLint/issues/70)
 
+* Add new `prefer_key_path` rule that triggers when a trailing closure on a function
```
ikelax, line 64: Should it be mentioned that the rule is opt-in or disabled by default?

SimplyDanny: No, that's not necessary as it can be looked up in the [Rule Directory](https://realm.github.io/SwiftLint/rule-directory.html).

SimplyDanny:
I'm still unsure about this rule. As we see in the list of OSS findings, there will potentially be a lot of violations in existing code bases (78 in SwiftLint), most of them on `.map`, `.filter`, ... These are cases that could be and should be (due to their sheer amount) auto-fixed. However, fixing all findings will very likely lead to wrong code due to missing parameter names for custom functions accepting closures.

Fixing only a few (but most of the) findings seems to be against the philosophy of SwiftLint. Are there other rules that fix only partially? Would this be acceptable? Or do people expect a `swiftlint lint --fix` run to repair all violations?

SimplyDanny:
So I restricted the rule to known standard functions by default and there is an option to enable it for all other closures as well. I actually doubt that anyone would enable it to be honest, especially when it comes to auto-fixes.
 
@mildm8nnered: May I ask you again to run this rule with `--fix` on your project? In SwiftLint, there has only been one rewrite that caused a compile error afterwards - unfortunately one that SwiftLint cannot recognize as a false positive due to missing type information.

SimplyDanny:
> In SwiftLint, there has only been one rewrite that caused a compile error afterwards - unfortunately one that SwiftLint cannot recognize as a false positive due to missing type information.

Luckily, there is [an accepted proposal](https://github.com/apple/swift-evolution/blob/main/proposals/0438-metatype-keypath.md) that is going to make the auto-fixed code compile in this special case.

SimplyDanny:
> > @mildm8nnered: I'd like to hear your opinion. The rule doesn't seem to be so involved. Any more cases or shortcomings you can think of?
> 
> Just trying it on my local codebase, which I expect to have a lot of violations, pretty much like the OSS findings.

Could you check again with the `--fix` argument, @mildm8nnered? I've so far applied auto-fixes to a few smaller code bases and at least the default case (`restrict_to_standard_functions = true`) worked except for the one issue mentioned in the previous comments.

mildm8nnered:
> > > @mildm8nnered: I'd like to hear your opinion. The rule doesn't seem to be so involved. Any more cases or shortcomings you can think of?
> > 
> > 
> > Just trying it on my local codebase, which I expect to have a lot of violations, pretty much like the OSS findings.
> 
> Could you check again with the `--fix` argument, @mildm8nnered? I've so far applied auto-fixes to a few smaller code bases and at least the default case (`restrict_to_standard_functions = true`) worked except for the one issue mentioned in the previous comments.

So `--fix` worked for me - 167 files changed everything still compiles ...

SimplyDanny:
> > > > @mildm8nnered: I'd like to hear your opinion. The rule doesn't seem to be so involved. Any more cases or shortcomings you can think of?
> > > 
> > > 
> > > Just trying it on my local codebase, which I expect to have a lot of violations, pretty much like the OSS findings.
> > 
> > 
> > Could you check again with the `--fix` argument, @mildm8nnered? I've so far applied auto-fixes to a few smaller code bases and at least the default case (`restrict_to_standard_functions = true`) worked except for the one issue mentioned in the previous comments.
> 
> So `--fix` worked for me - 167 files changed everything still compiles ...

Thank you! That's good news. So the PR can finally land.

PR #5552, opened 2024-05-01
Adds `baseline` and `write_baseline` config file options

mildm8nnered:
So baselines can be used by people who can't pass command line arguments (e.g. plugin users).

`baseline` and `write_baseline` - these behave exactly the same way as their command line equivalents. The command line will take preference if there is a conflict, but there is not way to "unset" the config file entries from the command line - if there is a `baseline` config file entry, you can override it to be a different value, but there is no way to tell SwiftLint "don't read a baseline at all".

This is very similar to the PR from a while ago adding a `strict` config option - #5226

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -292,7 +292,7 @@
 * Adds `baseline` and `write_baseline` configuration file settings, equivalent
   to the `--baseline` and `--write-baseline` command line options.  
   [Martin Redington](https://github.com/mildm8nnered)
-  [#XXXX](https://github.com/realm/SwiftLint/issues/XXXX)
+  [#5552](https://github.com/realm/SwiftLint/issues/5552)
 
 #### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -289,6 +289,10 @@
   consider said types in declarations like `let i: Int = 1` or
   `let s: String = ""` as redundant.  
   [Garric Nahapetian](https://github.com/garricn)
+* Adds `baseline` and `write_baseline` configuration file settings, equivalent
+  to the `--baseline` and `--write-baseline` command line options.  
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#XXXX](https://github.com/realm/SwiftLint/issues/XXXX)
 
 #### Bug Fixes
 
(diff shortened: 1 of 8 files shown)


Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
```diff
@@ -44,7 +44,8 @@ struct LintOrAnalyzeCommand {
     private static func lintOrAnalyze(_ options: LintOrAnalyzeOptions) async throws {
         let builder = LintOrAnalyzeResultBuilder(options)
         let files = try await collectViolations(builder: builder)
-        if let baselineOutputPath = options.writeBaseline {
+        let baselineOutputPath = options.writeBaseline ?? builder.configuration.writeBaseline
+        if let baselineOutputPath {
```
SimplyDanny, line 48:
```suggestion
        if let baselineOutputPath = options.writeBaseline ?? builder.configuration.writeBaseline {
```

SimplyDanny (commented):
Is `write_baseline` really useful to have in the configuration? That would mean every new violation is directly added to the baseline. But new appearing violations should rather be fixed.

mildm8nnered:
> Is `write_baseline` really useful to have in the configuration? That would mean every new violation is directly added to the baseline. But new appearing violations should rather be fixed.

So I don't expect it to get that much use, but I could see a few possible use cases.

If you want to generate the Baselines in your CI environment, and can't pass command line options, then this would let you do so without having to reconstruct that environment locally. I could imagine people having a monthly or even yearly job for this.

If you wanted to track whether existing violations were getting fixed, you could write out a baseline for the build, and then use `swiftlint baseline compare` against the main baseline to detect fixed violations.

CHANGELOG.md
```diff
@@ -290,6 +290,11 @@
   `let s: String = ""` as redundant.  
   [Garric Nahapetian](https://github.com/garricn)
 
+* Adds `baseline` and `write_baseline` configuration file settings, equivalent
+  to the `--baseline` and `--write-baseline` command line options.  
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#5552](https://github.com/realm/SwiftLint/issues/5552)
```
SimplyDanny, line 296: This needs to be moved up into the "Main" section.

SimplyDanny (approved):

Issue #5424, opened 2024-01-16
Cannot run SwiftLint on any directory other than current working directory


tahirmt:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

I want to be able to run `swiftlint lint Project/Directory` and be able to have it read the config file located at `Project/Directory/.swiftlint.yml` and this does not work. It seems to not read the config file at all. If I pass in the config file using `--config` argument then it doesn't read nested `.swiftlint.yml` config files so that's not a possible alternative either.

Ideally, it should read the config files from the directory but if that's not possible then alternatively I would like a `--working-directory` argument so we can execute it from not the current directory. I have no option to cd to the working directory in my scenario as I am in a controlled environment.

##### Complete output when running SwiftLint, including the stack trace and command used

This is also irrelevant. It shows me thousands of violations for paths that I have ignored in my config file but it doesn't apply the config file.

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.54.0
* Installation method used (Homebrew, CocoaPods, building from source, etc)? Homebrew
* Paste your configuration file:

The configuration file itself is not relevant.

SoKamil:
In my case I want to traverse back two levels with `swiftlint ../../` and it reads properties from `.swiftlint.yml` like `disabled_rules` but then it completely ignores properties where path matters like `excluded` or `parent_config`.

mildm8nnered:
So ... ideally SwiftLint should work from anywhere, but there are definitely some spots in the code that assume that you will be executing at the top level of your project (maybe - at least it looks like that in the code), and there is also some complexity around how paths are treated on the command line, for historical reasons. It can be hard to change some of this behaviour, because existing users may be relying on them, even where they don't make sense when looked at afresh.

I think SwiftLint should read the `.swiftlint.yml`, just as though it was a nested configuration, although some settings may not work in that context (which sounds like it might be what's biting @SoKamil )- that is definitely something I can look into, time permitting.

mildm8nnered:
So I'm interested to know what kind of environment you're in @tahirmt that won't let you `cd`. I think the best solution for you here may be a `--working-directory` option, if you really can't `cd'.

[...]

PR #5560, opened 2024-05-04
Adds `--working-directory` command line option

mildm8nnered:
Adds a `--working-directory` argument to `lint` and `analyze`, for users who cannot easily control the current directory when executing SwiftLint (e.g. in some plugin and CI scenarios). 

Addresses #5424

This is required because `.swiftlint.yml`'s in the current directory are treated slightly differently to  `.swiftlint.yml`'s found deeper in the tree. 

For example:

```
$ pwd
/Users/some.user/Documents/Source/SwiftLint
$
$ swiftlint --quiet --reporter summary
+-----------------+--------+-------------+--------+----------+--------+------------------+-----------------+
| rule identifier | opt-in | correctable | custom | warnings | errors | total violations | number of files |
+-----------------+--------+-------------+--------+----------+--------+------------------+-----------------+
| Total           |        |             |        |        0 |      0 |                0 |               0 |
+-----------------+--------+-------------+--------+----------+--------+------------------+-----------------+
$
$
$ cd ../..
$ pwd
/Users/some.user/Documents
$
$
$ swiftlint  --quiet --reporter summary Source/SwiftLint
Linting Swift files at paths Source/SwiftLint
+---------------------+--------+-------------+--------+----------+--------+------------------+-----------------+
| rule identifier     | opt-in | correctable | custom | warnings | errors | total violations | number of files |
+---------------------+--------+-------------+--------+----------+--------+------------------+-----------------+
| file_name           | yes    | no          | no     |       18 |      0 |               18 |              18 |
| type_name           | no     | no          | no     |        8 |      0 |                8 |               8 |
| file_header         | yes    | no          | no     |        5 |      0 |                5 |               5 |
| trailing_comma      | no     | yes         | no     |        4 |      0 |                4 |               1 |
| empty_xctest_method | yes    | no          | no     |        1 |      0 |                1 |               1 |
| legacy_objc_type    | yes    | no          | no     |        1 |      0 |                1 |               1 |
| let_var_whitespace  | yes    | no          | no     |        1 |      0 |                1 |               1 |
+---------------------+--------+-------------+--------+----------+--------+------------------+-----------------+
| Total               |        |             |        |       38 |      0 |               38 |              21 |
+---------------------+--------+-------------+--------+----------+--------+------------------+-----------------+
Done linting! Found 38 violations, 0 serious in 646 files.
$
$
$ swiftlint --quiet --reporter summary --working-directory Source/SwiftLint 
+-----------------+--------+-------------+--------+----------+--------+------------------+-----------------+
| rule identifier | opt-in | correctable | custom | warnings | errors | total violations | number of files |
+-----------------+--------+-------------+--------+----------+--------+------------------+-----------------+
| Total           |        |             |        |        0 |      0 |                0 |               0 |
+-----------------+--------+-------------+--------+----------+--------+------------------+-----------------+
```

We could assume that the first `.swiftlint.yml`  we encountered was the "main" one, but this would not be true in the unusual, but valid case, where users intentionally did not have a "main" `.swiftlint.yml`. Additionally, some SwiftLint code, specifically some reporters, assume that the current directory is also the top-level directory of the project.

All relative paths on the command line and in configuration files will be treated as relative to the specified working directory.

### DIFF
--- a/Source/SwiftLintCore/Models/Configuration.swift
+++ b/Source/SwiftLintCore/Models/Configuration.swift
@@ -254,7 +254,7 @@ public struct Configuration {
             if useDefaultConfigOnFailure ?? !hasCustomConfigurationFiles {
                 // No files were explicitly specified, so maybe the user doesn't want a config at all -> warn
                 queuedPrintError(
-                    "\(Issue.wrap(error: error).errorDescription) – Falling back to default configuration"
+                    "\(Issue.wrap(error: error).errorDescription ?? "") – Falling back to default configuration"
                 )
                 self.init(rulesMode: rulesMode, cachePath: cachePath)
             } else {
--- a/Source/SwiftLintCore/Models/Configuration.swift
+++ b/Source/SwiftLintCore/Models/Configuration.swift
@@ -254,7 +254,7 @@ public struct Configuration {
             if useDefaultConfigOnFailure ?? !hasCustomConfigurationFiles {
                 // No files were explicitly specified, so maybe the user doesn't want a config at all -> warn
                 queuedPrintError(
-                    "\(Issue.wrap(error: error).errorDescription ?? "") – Falling back to default configuration"
+                    "\(Issue.wrap(error: error).errorDescription) – Falling back to default configuration"
                 )
                 self.init(rulesMode: rulesMode, cachePath: cachePath)
             } else {
--- a/Source/swiftlint/Helpers/LintOrAnalyzeArguments.swift
+++ b/Source/swiftlint/Helpers/LintOrAnalyzeArguments.swift
@@ -42,6 +42,8 @@ struct LintOrAnalyzeArguments: ParsableArguments {
     var baseline: String?
     @Option(help: "The path to save detected violations to as a new baseline.")
     var writeBaseline: String?
+    @Option(help: "The working directory to use when running SwiftLint.")
+    var workingDirectory: String?
     @Flag(help: "Use the in-process version of SourceKit.")
     var inProcessSourcekit = false
     @Option(help: "The file where violations should be saved. Prints to stdout by default.")
--- a/Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
+++ b/Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
@@ -29,7 +29,7 @@ struct LintOrAnalyzeCommand {
         if let workingDirectory = options.workingDirectory {
             if !FileManager.default.changeCurrentDirectoryPath(workingDirectory) {
                 throw SwiftLintError.usageError(
-                    description: "Could not change working directory to '\(workingDirectory)'"
+                    description: "Could not change working directory to '\(workingDirectory)'. Make sure it exists and is accessible."
                 )
             }
         }
--- a/Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
+++ b/Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
@@ -29,8 +29,8 @@ struct LintOrAnalyzeCommand {
         if let workingDirectory = options.workingDirectory {
             if !FileManager.default.changeCurrentDirectoryPath(workingDirectory) {
                 throw SwiftLintError.usageError(
-                    description:"Could not change working directory to '\(workingDirectory)'. " +
-                                "Make sure it exists and is accessible."
+                    description: "Could not change working directory to '\(workingDirectory)'. " +
+                                 "Make sure it exists and is accessible."
                 )
             }
         }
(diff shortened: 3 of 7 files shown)


Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
```diff
@@ -26,6 +26,11 @@ enum LintOrAnalyzeMode {
 
 struct LintOrAnalyzeCommand {
     static func run(_ options: LintOrAnalyzeOptions) async throws {
+        if let workingDirectory = options.workingDirectory {
+            if !FileManager.default.changeCurrentDirectoryPath(workingDirectory) {
+                throw Issue.couldNotChangeToWorkingDirectory(path: workingDirectory)
```
mildm8nnered, line 31: So originally this `throw` produced the following eventual output:

`Error: couldNotChangeToWorkingDirectory(path: "Source/SwiftLintt")`

because the thrown error eventually gets passed to `MessageInfo(error: Error, type: ParsableArguments.Type)` from the `swift-argument-parser`. Ideally it would get caught there by

```
      case let error as LocalizedError where error.errorDescription != nil:
        self = .other(message: error.errorDescription!, exitCode: .failure)
```

but it was not, because `Issue`'s implementation of `errorDescription` (`var errorDescription: String`) does not match the `LocalizedError` protocol requirement (`var errorDescription: String?`).

I've fixed that, but because of the way the thrown error is now printed, we get 

`Error: warning: Could not change working directory to 'Source/SwiftLintt'.`

One way around that would just be to brute-force:

```
if !FileManager.default.changeCurrentDirectoryPath(workingDirectory) {
    Issue.couldNotChangeToWorkingDirectory(path: workingDirectory).print()
    exit(2)
}
```

Another possibility would be to remove our `warning: ` and `error: ` prefixes from `errorDescription`, but to add them back in `Issue.print()` - most usages look like that would be ok, but I'm not sure how confident I am that all of them would be.

Source/SwiftLintCore/Models/Configuration.swift
```diff
@@ -235,7 +235,7 @@ public struct Configuration {
             if useDefaultConfigOnFailure ?? !hasCustomConfigurationFiles {
                 // No files were explicitly specified, so maybe the user doesn't want a config at all -> warn
                 queuedPrintError(
-                    "\(Issue.wrap(error: error).errorDescription) – Falling back to default configuration"
+                    "\(Issue.wrap(error: error).errorDescription ?? "") – Falling back to default configuration"
```
mildm8nnered, line 238: Not sure what a good default is here

Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
mildm8nnered: Switched to `SwiftLintError.usageError`, which produces: `Error: Could not change working directory to 'Source/SwiftLintt'

I would kind of expected. throwing an `Issue` to work though.

Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
```diff
@@ -26,6 +26,13 @@ enum LintOrAnalyzeMode {
 
 struct LintOrAnalyzeCommand {
     static func run(_ options: LintOrAnalyzeOptions) async throws {
+        if let workingDirectory = options.workingDirectory {
+            if !FileManager.default.changeCurrentDirectoryPath(workingDirectory) {
+                throw SwiftLintError.usageError(
+                    description: "Could not change working directory to '\(workingDirectory)'"
```
SimplyDanny, line 32:
```suggestion
                    description: "Could not change working directory to '\(workingDirectory)'. Make sure it exists and is accessible."
```

Source/swiftlint/Helpers/LintOrAnalyzeArguments.swift
```diff
@@ -42,6 +42,8 @@ struct LintOrAnalyzeArguments: ParsableArguments {
     var baseline: String?
     @Option(help: "The path to save detected violations to as a new baseline.")
     var writeBaseline: String?
+    @Option(help: "The working directory to use when running SwiftLint.")
```
SimplyDanny, line 45: Maybe this should elaborate a bit more on what "working directory" means. Or is this obvious?

SimplyDanny (approved):

Source/swiftlint/Helpers/LintOrAnalyzeArguments.swift
mildm8nnered: Well, "working directory" has it's own wikipedia page (https://en.wikipedia.org/wiki/Working_directory) which is more than we do right now (just checked and my submission is still in review).

```
% man pwd
PWD(1)                                                                         General Commands Manual                                                                         PWD(1)

NAME
     pwd – return working directory name
```

So for at least some users, I think it should be obvious

Source/swiftlint/Helpers/LintOrAnalyzeCommand.swift
```diff
@@ -26,6 +26,14 @@ enum LintOrAnalyzeMode {
 
 struct LintOrAnalyzeCommand {
     static func run(_ options: LintOrAnalyzeOptions) async throws {
+        if let workingDirectory = options.workingDirectory {
+            if !FileManager.default.changeCurrentDirectoryPath(workingDirectory) {
+                throw SwiftLintError.usageError(
+                    description: "Could not change working directory to '\(workingDirectory)'. " +
+                                 "Make sure it exists and is accessible."
```
SimplyDanny, line 33: I'd prefer a text block.

Issue #5295, opened 2023-10-21
Swift type checking using is


inotetw:
## New rule request

type ching using is

### Bad
```
if  a as? Dog  != nil { 
   doSomeThing()
}
```
### Good
```
if  a is Dog  { 
   doSomeThing()
}

if  let dog = dog as? Dog  { 
   dog.run()
}
```

type checking  using "is"  is more easy to understand then type casting and check nil

AballahNsour:
I will take this one

SimplyDanny:
> I will take this one

There has been some work in https://github.com/realm/SwiftLint/pull/5561 already.

PR #5561, opened 2024-05-05
Swift type checking using is

ikelax:
Addresses #5295.

This new rule triggers on code like
```swift
if a as? Dog != nil { 
   doSomeThing()
}
```

which can be replaced by

```swift
if a is Dog  { 
   doSomeThing()
}
```

or

```swift
if let a = a as? Dog  { 
   a.run()
}
```

However, so far there is not a Rewriter for fixing this automatically. If I am not mistaken, a as? Dog != nil can always be replaced by a is Dog.


ikelax:
I added a `Rewriter` for fixing this automatically. If I am not mistaken, `a as? Dog != nil` can always be replaced by `a is Dog`.

I checked also the finding of SwiftLintBot. As far as I can tell, there is not anything that speaks against this rule.

mildm8nnered:
So I wrote #5649 the other day because I had not spotted this PR. I've left mine open as a draft - please feel free to steal anything you like from it - I overrode `visitPost(_ node: UnresolvedAsExprSyntax)`, and then look at the siblings of the parent `ExprListSyntax` to look for `!= nil`, which might be a bit more efficient that looking at every `ExprListSyntax` (although I have not measured that).

I think the solutions are quite similar, but mine seem to pick up 26 OSS violations, versus 10 for this branch. When I looked at a couple, they did not seem to be false alarms. For example

https://github.com/wireapp/wire-ios/blob/ff4e0b2b87c7f78d638e3dc1791d587eaab8b6ed/wire-ios/Wire-iOS/Sources/UserInterface/Components/Views/MarkdownTextView+Recognizers.swift#L41:61

and 

https://github.com/kickstarter/ios-oss/blob/eb2e07004a6ffb7be3764dfa6f23d518c0b38290/Library/TestHelpers/MockStripeIntentService.swift#L17:41

ikelax:
Thank you. I prefer your idea so far. As soon as I am free again, I will work on it.

ikelax:
@mildm8nnered I stole a lot from #5649. What do you think about the current solution? Do you mind if I mention you also in the changelog?

mildm8nnered:
> @mildm8nnered I stole a lot from #5649. What do you think about the current solution? Do you mind if I also mention you in the changelog?

Hi @ikelax - absolutely fine to credit me, but make sure you put yourself first!

I'm on vacation right now, but back home at the end of this week, so I'll be able to review it then, but hopefully @SimplyDanny will take a look too, as he is the most stringent code reviewer.

CHANGELOG.md
```diff
@@ -433,7 +433,7 @@
   [Garric Nahapetian](https://github.com/garricn)
 
 * Add new `prefer_type_checking` rule to prefer `a is X` over `a as? X != nil`.  
-  [ikelax](https://github.com/ikelax)
+  [ikelax](https://github.com/ikelax) and [mildm8nnered](https://github.com/mildm8nnered)
```
SimplyDanny, line 436: We usually just list the names:
```suggestion
  [ikelax](https://github.com/ikelax)
  [mildm8nnered](https://github.com/mildm8nnered)
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferTypeCheckingRule.swift
```diff
@@ -0,0 +1,107 @@
...
+        }
+        return node.questionOrExclamationMark?.tokenKind == .postfixQuestionMark
+            && parent.dropLast().last?.as(BinaryOperatorExprSyntax.self)?.operator.tokenKind == .binaryOperator("!=")
+    }
+}
+
+private extension ExprListSyntax {
+    var unresolvedAsExprIsBeingComparedToNotNil: Bool {
+        guard
+            let node = dropFirst().first?.as(UnresolvedAsExprSyntax.self),
```
SimplyDanny, line 99: I don't think this covers cases like `2*x as? X != nil` which are rare but possible. I recommend setting `foldExpressions: true` in `@SwiftSyntaxRule` so you can work on a folded expression tree. That should simplify catching such cases.

Also, rewriting the syntax tree should become more straightforward if you start with the parent node, which is `InfixOperatorExprSyntax` with a `!=` `BinaryOperatorExprSyntax`. You can then replace the entire node and use the same entry point for both the visitor and the rewriter.

ikelax: I added such a case as an example.

By using `InfixOperatorExprSyntax`, I was able to simplify the code. Thank you!

mildm8nnered (approved):
lgtm!

mildm8nnered:
Congratulations on your first rule @ikelax !

jumhyn-browser:
@mildm8nnered FYI we've just updated SwiftLint and have disabled this rule globally because it is triggering on some examples which result in invalid code. The particular case we encountered looked like:
```
if let parseData(data) as MyType? != nil {
  // ...
}
```
It appears this rule is triggering on `as` in addition to `as?` and when `as` is converted to `is` it's possible for the expression to lose type context that was necessary to compile the expression (since `as` provides static type coercion while `is` is only a runtime type check).

mildm8nnered:
> @mildm8nnered FYI we've just updated SwiftLint and have disabled this rule globally because it is triggering on some examples which result in invalid code. The particular case we encountered looked like:
> 
> ```
> if let parseData(data) as MyType? != nil {
>   // ...
> }
> ```
> 
> It appears this rule is triggering on `as` in addition to `as?` and when `as` is converted to `is` it's possible for the expression to lose type context that was necessary to compile the expression (since `as` provides static type coercion while `is` is only a runtime type check).

@jumhyn-browser - could you file this as a new issue, and I'll take a look when I get a chance

jumhyn-browser:
@mildm8nnered filed https://github.com/realm/SwiftLint/issues/5854

Issue #5564, opened 2024-05-09
Rule Request: Newline between function, array, dictionary, tuple parameters


kimdv:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

After I read this [proposal](https://github.com/apple/swift-evolution/pull/2344) I tought it would be nice to have a rule like "if there is more than x parameters in a function/tuple/array/dictionary there must be a newline between".

This rule can also be implemented without this proposal implemented. 

The reason for this is makes it easier to review code. So the rule should be opt-in.
There could also be a configuration to state how many parameters there must be before newline is required.

It could also say that the trailing `)` should be on a newline. 
We could also have a configuration that require the `(` to have a trailing newline

Examples:
```swift
// Not OK
func foo(bar: Bar, baz: Baz)
func foo(
  bar: Bar,
  baz: Baz)

// OK
func foo(bar: Bar)
func foo(
  bar: Bar,
  baz: Baz
)
```

Not sure if it should be the same rule, but when calling a function we could also require newline.

Examples:
```swift
// Not OK
someClass.foo(bar: "bar", baz: "baz")
someClass.foo(
  bar: "bar",
  baz: "baz")

// OK
someClass.foo(bar: "bar")
someClass.foo(
  bar: "bar",
  baz: "baz"
)
```

kimdv:
If it makes sense and it's a good idea I'm willing to try to implement it

kimdv:
Is already in SwiftLint

PR #5569, opened 2024-05-12
[New rule] Add new rule for functions parameters

kimdv:
Fixes: https://github.com/realm/SwiftLint/issues/5564

This rule will ensure that all parameters are on a newline. 
This will enhance the diff when doing reviews. 

Not sure if we also should expand this rule for arrays and dictionaries. But I think it would be best to split those rules into multiple.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/FunctionParametersNewlineConfiguration.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/FunctionParametersNewlineConfiguration.swift
@@ -1,0 +1,13 @@
+import SwiftLintCore
+
+@AutoApply
+struct FunctionParametersNewlineConfiguration: SeverityBasedRuleConfiguration {
+    typealias Parent = FunctionParametersNewlineRule
+
+    @ConfigurationElement(key: "severity")
+    private(set) var severityConfiguration = SeverityConfiguration<Parent>(.warning)
+    @ConfigurationElement(key: "leading_paren_on_newline")
+    private(set) var leadingParenOnNewline = false
+    @ConfigurationElement(key: "trailing_paren_on_newline")
+    private(set) var trailingParenOnNewline = false
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@
 
 #### Enhancements
 
-* None.
+* [New rule] Add new rule for functions parameters.  
+  [kimdv](https://github.com/kimdv)
+  [#issue_number](https://github.com/realm/SwiftLint/issues/5564)
 
 #### Bug Fixes
 
--- a/Source/SwiftLintBuiltInRules/Models/BuiltInRules.swift
+++ b/Source/SwiftLintBuiltInRules/Models/BuiltInRules.swift
@@ -81,6 +81,7 @@ public let builtInRules: [any Rule.Type] = [
     FunctionBodyLengthRule.self,
     FunctionDefaultParameterAtEndRule.self,
     FunctionParameterCountRule.self,
+    FunctionParametersNewlineRule.self,
     GenericTypeNameRule.self,
     IBInspectableInExtensionRule.self,
     IdenticalOperandsRule.self,
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/FunctionParametersNewlineRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/FunctionParametersNewlineRule.swift
@@ -1,0 +1,91 @@
+import SwiftSyntax
+
+@SwiftSyntaxRule
+struct FunctionParametersNewlineRule: OptInRule {
+    var configuration = FunctionParametersNewlineConfiguration()
+
+    static let description = RuleDescription(
+        identifier: "function_parameters_newline",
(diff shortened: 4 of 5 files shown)


Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/FunctionParametersNewlineConfiguration.swift
```diff
@@ -0,0 +1,13 @@
+import SwiftLintCore
+
+@AutoApply
+struct FunctionParametersNewlineConfiguration: SeverityBasedRuleConfiguration {
+    typealias Parent = FunctionParametersNewlineRule
```
kimdv, line 5: Not sure if it makes sense. 
But we could also include a configuration so that the user could define how many params is needed before this rule is takes into account? Some people properly will allow `self.foo(bar: "bar", baz: "baz")` etc.

mildm8nnered:
I think this might be covered already by existing `multiline_xxxx` rules.

[multiline_arguments](https://realm.github.io/SwiftLint/multiline_arguments.html): Multiline Arguments
[multiline_arguments_brackets](https://realm.github.io/SwiftLint/multiline_arguments_brackets.html): Multiline Arguments Brackets
[multiline_function_chains](https://realm.github.io/SwiftLint/multiline_function_chains.html): Multiline Function Chains
[multiline_literal_brackets](https://realm.github.io/SwiftLint/multiline_literal_brackets.html): Multiline Literal Brackets
[multiline_parameters](https://realm.github.io/SwiftLint/multiline_parameters.html): Multiline Parameters
[multiline_parameters_brackets](https://realm.github.io/SwiftLint/multiline_parameters_brackets.html): Multiline Parameters Brackets

I *think* that `multiline_parameters` with `allows_single_line` set to false, and either `multiline_arguments_brackets` or `multiline_parameters_brackets` (they appear to do much the same thing) will be almost equivalent to this rule.

`multiline_literal_brackets` will do the same for arrays.

Rules about multiline parameters etc probably do belong in the `multiline_xxxx` namespace, but it is getting a bit crowded.

kimdv:
> I think this might be covered already by existing `multiline_xxxx` rules.
> 
> [multiline_arguments](https://realm.github.io/SwiftLint/multiline_arguments.html): Multiline Arguments [multiline_arguments_brackets](https://realm.github.io/SwiftLint/multiline_arguments_brackets.html): Multiline Arguments Brackets [multiline_function_chains](https://realm.github.io/SwiftLint/multiline_function_chains.html): Multiline Function Chains [multiline_literal_brackets](https://realm.github.io/SwiftLint/multiline_literal_brackets.html): Multiline Literal Brackets [multiline_parameters](https://realm.github.io/SwiftLint/multiline_parameters.html): Multiline Parameters [multiline_parameters_brackets](https://realm.github.io/SwiftLint/multiline_parameters_brackets.html): Multiline Parameters Brackets
> 
> I _think_ that `multiline_parameters` with `allows_single_line` set to false, and either `multiline_arguments_brackets` or `multiline_parameters_brackets` (they appear to do much the same thing) will be almost equivalent to this rule.
> 
> `multiline_literal_brackets` will do the same for arrays.
> 
> Rules about multiline parameters etc probably do belong in the `multiline_xxxx` namespace, but it is getting a bit crowded.

Oh! 
Thanks! 

I might look into those.

kimdv:
@mildm8nnered is it right that it's not possible to require all items inside an array etc to be on newline?
I see that `multiline_literal_brackets` have a "allow_single_line" like `multiline_parameters`

mildm8nnered:
> @mildm8nnered is it right that it's not possible to require all items inside an array etc to be on newline? I see that `multiline_literal_brackets` have a "allow_single_line" like `multiline_parameters`

So I'm not sure whether it's "right" or not - I suspect that the `multiline_xxxx` rules grew over time, and that there's not as much consistency as there could be between their various options.

It's quite likely that they be consolidated and rationalized a bit, but that's hard to do without potentially breaking existing deployments.

kimdv:
> > @mildm8nnered is it right that it's not possible to require all items inside an array etc to be on newline? I see that `multiline_literal_brackets` have a "allow_single_line" like `multiline_parameters`
> 
> So I'm not sure whether it's "right" or not - I suspect that the `multiline_xxxx` rules grew over time, and that there's not as much consistency as there could be between their various options.
> 
> It's quite likely that they be consolidated and rationalized a bit, but that's hard to do without potentially breaking existing deployments.

If they should be consolidated, we could keep them as is, and then reintroduce a new rule hat is more consistent :)

mildm8nnered:
> If they should be consolidated, we could keep them as is, and then reintroduce a new rule hat is more consistent :)

So that would be a possible way forward, but given that they have pretty good coverage across them already, would that be a good use of time?

I think if there's something that you can't do with some combination of the existing rules, new options would be the path of least resistance maybe.

SimplyDanny:
Let me add my two cents: I don't have a complete overview on all the `multiline_*` rules, so don't know what would have to be touched. But if we can make them more consistent and align their configuration options (making them more understandable for users and for contributors), go for it. Changed behavior should be documented in the CHANGELOG so that people get informed. Other than that, rules being in use already should not be blocking. Of course, discussing alternative options is always valid as well.

PR #5577, opened 2024-05-14
Silence `unused_enumerated` rule when `$0` in a closure is explicitly unpacked

### LINKED ISSUE #5573: [0.55.0] False positive on unused_enumerated with block + destructuring
### New Issue Checklist

### Describe the bug

The rule should not trigger when the index is used by destructuring the argument tuple.

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.55.0
* Installation method used (Homebrew, CocoaPods, building from source, etc)? bazel

```yml
# insert yaml contents here
```

No 

* Which Xcode version are you using (check `xcodebuild -version`)? 15.2.0

```swift
AlertOption.allCases.enumerated().map {
    let (i, option) = $0
    // ...
}
```

SimplyDanny:
The fix is quite specific for this single case. What are other cases where an "unpack" can happen? We could further check if `i` and `e` are actually used. But that's out of scope for this fix.

Fixes #5573.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,9 @@
 
 #### Bug Fixes
 
-* None.
+* Silence `unused_enumerated` rule when `$0` in a closure is explicitly unpacked.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+  [#5573](https://github.com/realm/SwiftLint/issues/5573)
 
 ## 0.55.0: Universal Washing Powder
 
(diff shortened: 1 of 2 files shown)


mildm8nnered (approved):
lgtm.

I think there are probably so many cases that it's very hard to catch them all, but if we can just fix most of the ones that cause people problems then maybe that's good enough (and the other violations can be suppressed, or people can turn the rule off).

PR #5582, opened 2024-05-15
Respect comments before opening brace in `opening_brace` rule

### LINKED ISSUE #5578: opening_brace triggers when the brace is preceded by a comment
### New Issue Checklist

### Describe the bug

When an opening brace is preceded by a comment (for example a clarifying comment in an if statement, a commented out inheritance, etc.), the `opening_brace` rule reports a warning, even when the line is technically compliant (one space between the end of the comment and the brace).

### Environment

* SwiftLint 0.55.0
* CocoaPods
* Configuration file does not configure `opening_brace`.
* Xode 15.4

### Samples

The following three examples produce the `Opening Brace Spacing Violation: Opening braces should be preceded by a single space and on the same line as the declaration (opening_brace)` warning:

```swift
// Here I added a clarifying comment since it might not be obvious that the unit here is seconds.
if timeSpent < 60.0 /* Seconds */ {
    // Do something here...
}
```
```swift
do {
    // Try something that may throw
} catch /*where error is CancellationError == false*/ { // The error handling logic here has changed and thus the catch condition is no longer needed, but kept here for a while for reference.
    // Error handling...
}
```
```swift
class SomeClass: SomeBaseClass/*, SomeProtocol*/ { // Inheritance of SomeProtocol is no longer needed.
    
}
```

I did notice that this may already be a known issue stemming from this rule's rewrite using SwiftSyntax (https://github.com/realm/SwiftLint/pull/5164#pullrequestreview-1620492516).

SimplyDanny:
Fixes #5578.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,12 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#5571](https://github.com/realm/SwiftLint/issues/5571)
 
+* Respect comments before opening brace in `opening_brace` rule when there is
+  one space before the brace after the comment. Everything else is still a
+  violation, yet the rewriter will not remove the comment anymore.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+  [#5578](https://github.com/realm/SwiftLint/issues/5578)
+
 ## 0.55.0: Universal Washing Powder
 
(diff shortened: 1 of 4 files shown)


mildm8nnered (approved):
lgtm

PR #5595, opened 2024-05-17
Add IfSwitchExpressionRule for enforcing if and switch statements. In…

AballahNsour:
…clude non-triggering and triggering examples.

Add IfSwitchExpressionRule for enforcing if and switch statements. Include non-triggering and triggering examples.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Style/IfSwitchExpressionRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/IfSwitchExpressionRule.swift
@@ -1,0 +1,30 @@
+import Foundation
+import SwiftSyntax
+import SwiftLintCore
+
+
+@SwiftSyntaxRule(explicitRewriter: true)
+struct IfSwitchExpressionRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "if_switch_expression",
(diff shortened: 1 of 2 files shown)


Naveen-C-Ramachandrappa (approved):

SimplyDanny:
Hi @AballahNsour! Your examples are good, but the implementation of the rule is definitely not sufficient. Are you going to continue on this PR?

AballahNsour:
hi @SimplyDanny , If you can give more guidance I can continue working on this RP. if not I not continue working on it

SimplyDanny:
So if we start with `if` statements only, we expect

```swift
func f(value: Bool) -> Int {
  if value {
    return 1
  } else {
    return 2
  }
}
```

to trigger the rule as it can be rewritten to

```swift
func f(value: Bool) -> Int {
  return if value {
    1
  } else {
    2
  }
}
```

or even

```swift
func f(value: Bool) -> Int {
  if value {
    1
  } else {
    2
  }
}
```

if one prefers implicit returns. Is that correct?

Your current implementation triggers on each and every `if` statement/expression, though. You need to add more logic to find the correct cases on which the rule shall trigger on considering the different examples you have already provided.

I have to say that his is not an easy rule to implement. Hence, it's fine for me if you disregard it. At the moment, though, I cannot see at least a direction of where this is heading given the rudimentary implementation.

PR #5264, opened 2023-10-04
Add a new `non_optional_string_data_conversion` rule

ben-p-commits:
This rule was inspired by this issue: https://github.com/realm/SwiftLint/issues/5263.

I'm fairly new to `SwiftSyntax` so more efficient ideas/approaches are welcome 🙏🏻.

Issue #5263, opened 2023-10-04
Rule Request: Prefer non-optional UTF8 String <-> Data conversion


samrayner:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

Please describe the rule idea, format
this issue's title as `Rule Request: [Rule Name]` and describe:

1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.

`String.data(using:)` and `String.init(data:encoding:)` return optional values due to supporting a large number of encodings, some of which can fail conversion.

However, conversion of UTF8 encoded Strings cannot fail. E.g. `Data("foo".utf8)` and `String(decoding: Data(), as: UTF8.self)` do not return Optionals.

https://www.swiftbysundell.com/tips/converting-between-string-and-data-without-optionals/

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

Would trigger violation:
```swift
"foo".data(using: .utf8) // Data?
String(data: data, encoding: .utf8) // String?
```

Would not trigger violation:
```swift
Data("foo".utf8) // Data
String(decoding: data, as: UTF8.self) // String
```

3. Should the rule be configurable, if so what parameters should be configurable?

I don't think any configuration is necessary.

4. Should the rule be opt-in or enabled by default? Why?
   See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a rule as opt-in.
   
I suggest this can be on by default. I think non-optional is always preferable to optional where it is completely safe. There is no sensible handling that one can do for an optional returned using the violating methods as that code will never run.

ben-p-commits:
I think there's an issue with the `.utf16` examples. The proposed `Data("".utf16)` example isn't supported by [Foundation's  Data struct.](https://developer.apple.com/documentation/foundation/data)

```swift
// works, an initializer that supports Sequence<UInt8> exists
let data = Data("foo".utf8)
            
// compiler error. no initializer for Data exists that supports Sequence<UInt16> out of the box
let data = Data("foo".utf16)
```

ben-p-commits:
@samrayner - wrapped up the day with this: #5264

ortekka:
[...]

PR #5601, opened 2024-05-23
Reverse Data -> String conversion rule

samrayner:
This reverses the Data -> String portion of [this PR](https://github.com/realm/SwiftLint/pull/5264) by introducing a new rule, following feedback on [the original issue](https://github.com/realm/SwiftLint/issues/5263). It leaves the String -> Data portion of the original rule unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@
   [Martin Redington](https://github.com/mildm8nnered)
   [#5769](https://github.com/realm/SwiftLint/issues/5769)
 
+* Revert the part of the `non_optional_string_data_conversion`
+  rule that enforces non-failable conversions of `Data` to UTF-8
+  `String`. This is due to the fact that the data to be converted
+  can be arbitrary and especially doesn't need to represent a valid
+  UTF-8-encoded string.  
+  [Sam Rayner](https://github.com/samrayner)
+  [#5263](https://github.com/realm/SwiftLint/issues/5263)
+
 #### Experimental
 
 * None.
@@ -18,8 +26,9 @@
   `ignore_multiline_function_signatures`.  
   [leonardosrodrigues0](https://github.com/leonardosrodrigues0)
   [#3720](https://github.com/realm/SwiftLint/issues/3720)
+
 * Add new `optional_data_string_conversion` rule to enforce
-  failable conversions of `Data` -> UTF-8 `String`.
+  failable conversions of `Data` to UTF-8 `String`.  
   [Sam Rayner](https://github.com/samrayner)
   [#5263](https://github.com/realm/SwiftLint/issues/5263)
 
@@ -237,11 +246,6 @@
   rule.  
   [Martin Redington](https://github.com/mildm8nnered)
   [#5709](https://github.com/realm/SwiftLint/issues/5709)
-
-* Revert `optional_data_string_conversion` enforcing
-  non-failable conversions of `Data` -> UTF-8 `String`.
-  [Sam Rayner](https://github.com/samrayner)
-  [#5263](https://github.com/realm/SwiftLint/issues/5263)
 
 ## 0.55.1: Universal Washing Powder
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@
   `ignore_multiline_function_signatures`.  
   [leonardosrodrigues0](https://github.com/leonardosrodrigues0)
   [#3720](https://github.com/realm/SwiftLint/issues/3720)
+* Add new `optional_data_string_conversion` rule to enforce
+  failable conversions of `Data` -> UTF-8 `String`.
+  [Sam Rayner](https://github.com/samrayner)
+  [#5263](https://github.com/realm/SwiftLint/issues/5263)
 
 #### Bug Fixes
 
@@ -233,6 +237,11 @@
   rule.  
   [Martin Redington](https://github.com/mildm8nnered)
   [#5709](https://github.com/realm/SwiftLint/issues/5709)
+
+* Revert `optional_data_string_conversion` enforcing
+  non-failable conversions of `Data` -> UTF-8 `String`.
+  [Sam Rayner](https://github.com/samrayner)
+  [#5263](https://github.com/realm/SwiftLint/issues/5263)
 
 ## 0.55.1: Universal Washing Powder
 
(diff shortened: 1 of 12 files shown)


Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift
```diff
@@ -0,0 +1,31 @@
...
+    )
+}
+
+private extension OptionalDataStringConversionRule {
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        override func visitPost(_ node: DeclReferenceExprSyntax) {
+            if node.baseName.text == "String",
+               let parent = node.parent?.as(FunctionCallExprSyntax.self),
+               parent.arguments.map({ $0.label?.text }) == ["decoding", "as"],
+               parent.arguments.last?.expression.as(IdentifierTypeSyntax.self)?.name == "UTF8.self" {
```
ben-p-commits, line 26: This fixed the test locally on my end, @samrayner.

```suggestion
               let expr = parent.arguments.last?.expression.as(MemberAccessExprSyntax.self),
               expr.base?.description == "UTF8",
               expr.declName.baseName.description == "self" {
```

CHANGELOG.md
```diff
@@ -24,6 +27,11 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#5153](https://github.com/realm/SwiftLint/issues/5153)
 
+* Revert `optional_data_string_conversion` enforcing
+  non-failable conversions of `Data` -> UTF-8 `String`.
```
SimplyDanny, line 31:
```suggestion
* Revert the part of the `non_optional_string_data_conversion`
  rule that enforces non-failable conversions of `Data` -> UTF-8
  `String`. This is due to the fact that the data to be converted
  can be arbitrary and especially doesn't need to represent a valid
  UTF-8-encoded string.    
```

Please append two spaces to the end of the description. They are required to enforce a line break in the rendered Markdown.

Also, this should go into the "Breaking" section.

Source/SwiftLintBuiltInRules/Rules/Lint/NonOptionalStringDataConversionRule.swift
```diff
@@ -5,16 +5,14 @@ struct NonOptionalStringDataConversionRule: Rule {
     var configuration = SeverityConfiguration<Self>(.warning)
     static let description = RuleDescription(
         identifier: "non_optional_string_data_conversion",
-        name: "Non-Optional String <-> Data Conversion",
-        description: "Prefer using UTF-8 encoded strings when converting between `String` and `Data`",
+        name: "Non-Optional String -> Data Conversion",
```
SimplyDanny, line 8:
```suggestion
        name: "Non-optional String -> Data Conversion",
```

Source/SwiftLintBuiltInRules/Rules/Lint/NonOptionalStringDataConversionRule.swift
```diff
@@ -5,16 +5,14 @@ struct NonOptionalStringDataConversionRule: Rule {
...
+        description: "Prefer using non-optional Data(_:) when converting a UTF-8 String to Data",
```
SimplyDanny, line 9:
```suggestion
        description: "Prefer non-optional `Data(_:)` initializer when converting `String` to `Data`",
```

Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift
```diff
@@ -0,0 +1,32 @@
+import SwiftSyntax
+
+@SwiftSyntaxRule
+struct OptionalDataStringConversionRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+    static let description = RuleDescription(
+        identifier: "optional_data_string_conversion",
+        name: "Optional Data -> String Conversion",
+        description: "Prefer using failable String(data:encoding:) when converting from `Data` to a UTF-8 `String`",
```
SimplyDanny, line 9:
```suggestion
        description: "Prefer failable `String(data:encoding:)` initializer when converting `Data` to `String`",
```

SimplyDanny (approved):
Other than some nits on formulation and style, this looks good and reasonable to me. Thank you for taking immediate (re)action on the comments in the original PR!

SimplyDanny:
Please run `IntegrationTests` locally and fix any violations/failures reported.

samrayner:
Really sorry @SimplyDanny, I can’t figure out how to fix the tests. The IntegrationTests seem to pass but I may be running them wrong somehow. I’ve addressed your other comments I think.

SimplyDanny:
> Really sorry @SimplyDanny, I can’t figure out how to fix the tests. The IntegrationTests seem to pass but I may be running them wrong somehow. I’ve addressed your other comments I think.

You need to update [`default_rule_configurations.yml`](https://github.com/realm/SwiftLint/blob/main/Tests/IntegrationTests/default_rule_configurations.yml). `IntegrationTests.testDefaultConfigurations` checks that it contains all rules with their default options.

SimplyDanny:
The remaining fail is a deprecation that has just become active. I'll remove the obsolete rule later.

samrayner:
Thanks @SimplyDanny! Sorry for the massive delay on fixing up this PR.

jrose-signal:
This is firing on collections that aren't Data (specifically `UnsafeBufferPointer<UInt8>`, where converting to Data first would involve a heap copy. It's unfortunate that Swift doesn't have a failable conversion operation for this, but switching to Data is not automatically the right recommendation either.

SimplyDanny:
> This is firing on collections that aren't Data (specifically `UnsafeBufferPointer<UInt8>`, where converting to Data first would involve a heap copy). ~It's unfortunate that Swift doesn't have a failable conversion operation for this, but switching to Data is not automatically the right recommendation either.~ It needs to use `init(bytes:encoding:)` instead in that case.
> 
> EDIT: Looking at the implementation of `init(data:encoding:)` in the new swift-foundation, it just [immediately turns around and calls `init(bytes:encoding:)`](https://github.com/apple/swift-foundation/blob/1a4b8efed1d499f25233f1be7032ad14ddf1edbe/Sources/FoundationEssentials/String/String%2BIO.swift#L53), so that's probably a safe recommendation all the time!

With #5794, the rule now suggests `String(bytes:encoding:)`.

Issue #5551, opened 2024-04-30
Feature request: Add an (opt in) update check to SwiftLint


mildm8nnered:
Periphery has an (opt-out) update check: `--disable-update-check`

I'm not sure how happy I'd be with an opt out network access, but I think a lot of people aren't updating SwiftLint that often (which is a shame because there are a lot of great new rules, bug fixes, and general improvements in every release), and at least offering the option of an opt-in update check would help those users who enabled it to be alerted to new versions.

I've not looked into what we could check - GitHub would be easiest maybe. Probably on a fast timeout, so as not to interfere with people's builds.

SimplyDanny:
I agree that having this check on by default might not be everybody's favorite. Yet, if one needs to explicitly opt-in with an argument when running `swiftlint` to perform the check feels the same like having no check at all. An option in the configuration might be a good middle course, though.

mildm8nnered:
So config file option - I think these should kind of be standard for most command line options anyway - that would help with awareness.

While this would be useless at first when no-one's using it, at least it would be there, and we can build up awareness over time.

PR #5613, opened 2024-05-29
Add update checker

mildm8nnered:
Adds an update check facility to SwiftLint - addresses #5551 

This has been added as a `--check-for-updates` command line option to the `version`, `lint`, and `analyze` commands.

It can also be enabled via the configuration file, as `check_for_updates`.

Although the usefulness of an opt-in option is limited, at least this provides a good story for how users _can_ keep their installs up to date. And it could easily be changed to opt-out either here or at some later date.

The update checker code is pretty closely based on Periphery's implementation, and should probably be credited in the source file.

The update check is performed after all other processing, so linting will never be delayed, and the code is completely synchronous, as we do not have to worry about blocking the main thread.

The messaging for a new version is: `A new version of SwiftLint is available: 0.51.1`

We currently do not print any messaging if the update check fails, which is probably wrong.

```
% swiftlint help version
OVERVIEW: Display the current version of SwiftLint

USAGE: swiftlint version [--verbose] [--check-for-updates]

OPTIONS:
  --check-for-updates     Check whether a later version of SwiftLint is available.

%
% ./swiftlint version --check-for-updates
0.55.1
Your version of SwiftLint is up to date
%
%
% swiftlint help lint
OVERVIEW: Print lint warnings and errors

USAGE: swiftlint lint [<options>] [<paths> ...]

ARGUMENTS:
  <paths>                 List of paths to the files or directories to lint.

OPTIONS:
  --check-for-updates     Check whether a later version of SwiftLint is available.
```

### DIFF
--- a/Source/SwiftLintCore/Models/Version.swift
+++ b/Source/SwiftLintCore/Models/Version.swift
@@ -17,4 +17,3 @@ public struct Version: VersionComparable {
         self.value = value
     }
 }
-
--- a/Source/SwiftLintCore/Models/Version.swift
+++ b/Source/SwiftLintCore/Models/Version.swift
@@ -5,4 +5,34 @@ public struct Version {
 
     /// The current SwiftLint version.
     public static let current = Version(value: "0.55.1")
+
+    /// Public initializer.
+    ///
+    /// - parameter value: the string value for this version.
+    public init(value: String) {
+        self.value = value
+    }
+}
+
+extension Version: Comparable {
+    public static func == (lhs: Version, rhs: Version) -> Bool {
+        lhs.value == rhs.value
+    }
+
+    public static func < (lhs: Version, rhs: Version) -> Bool {
+        if let lhsComparators = lhs.comparators, let rhsComparators = rhs.comparators {
+            return lhsComparators.lexicographicallyPrecedes(rhsComparators)
+        }
+        return lhs.value < rhs.value
+    }
+
+    private var comparators: [Int]? {
+        let components = value.split(separator: ".").compactMap { Int($0) }
+        guard let major = components.first else {
+            return nil
+        }
+        let minor = components.dropFirst(1).first ?? 0
+        let patch = components.dropFirst(2).first ?? 0
+        return [major, minor, patch]
+    }
 }
--- a/Source/SwiftLintCore/Models/Version.swift
+++ b/Source/SwiftLintCore/Models/Version.swift
@@ -1,7 +1,11 @@
 /// A type describing the SwiftLint version.
-public struct Version {
+public struct Version: VersionComparable {
     /// The string value for this version.
     public let value: String
+
+    public var rawValue: String {
+        value
+    }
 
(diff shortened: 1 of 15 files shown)


Source/swiftlint/Helpers/UpdateChecker.swift
```diff
@@ -0,0 +1,58 @@
...
+            return nil
+        }
+        return tagName
+    }
+
+    private static func sendSynchronousRequest(to url: URL) -> Data? {
+        var request = URLRequest(url: url)
+        request.setValue("SwiftLint", forHTTPHeaderField: "User-Agent")
+        request.setValue("application/vnd.github.v3+json", forHTTPHeaderField: "Accept")
+        let semaphore = DispatchSemaphore(value: 0)
```
SimplyDanny, line 46: Can we go with some Swift Concurrency magic here instead? I assume the whole method chain can be `async`.

Source/SwiftLintCore/Models/Version.swift
```diff
@@ -5,4 +5,34 @@ public struct Version {
...
+
+    private var comparators: [Int]? {
+        let components = value.split(separator: ".").compactMap { Int($0) }
+        guard let major = components.first else {
+            return nil
+        }
+        let minor = components.dropFirst(1).first ?? 0
+        let patch = components.dropFirst(2).first ?? 0
+        return [major, minor, patch]
+    }
```
SimplyDanny, line 37: There is a similar (or equal) implementation for `SwiftVersion`. Can we somehow reduce them to just one? Something like `VersionComparable` with a default implementation that only relies on the underlying version string.

Source/swiftlint/Helpers/UpdateChecker.swift
SimplyDanny: Tried this in 8a4e922. Looks pretty good. 🤩

SimplyDanny (approved):
Looks good to me! A CHANGELOG entry is all that's left.

Source/swiftlint/Helpers/UpdateChecker.swift
SimplyDanny: ... yet not on Linux. 😑

mildm8nnered:
> Looks good to me! A CHANGELOG entry is all that's left.

changelog entry added, but it looks like we're failing CI on the server on the Linux build and Danger as well maybe.

SimplyDanny:
Seems like this is only going to work in Swift 6 (with libraries updated accordingly). So we can revert my changes and go with the classic implementation for now until we switch to Swift 6.

mildm8nnered:
> > Looks good to me! A CHANGELOG entry is all that's left.
> 
> changelog entry added, but it looks like we're failing CI on the server on the Linux build and Danger as well maybe.

Finally worked this out - the new protocol unsurprisingly needed to be public (and documented).

SimplyDanny (approved):

PR #5614, opened 2024-06-03
Remove deprecated path argument

mildm8nnered:
Removes the deprecated path argument.

Added this as a breaking change to the changelog, as I think it may be in some cases. For example

https://github.com/fastlane/fastlane/blob/806c80c39754034c3dfcdade6beb51bb55c7f3b1/fastlane/spec/actions_specs/swiftlint_spec.rb#L126

### DIFF
--- a/README.md
+++ b/README.md
@@ -680,7 +680,9 @@ opt_in_rules: # some rules are turned off by default, so you need to opt-in
 analyzer_rules: # rules run by `swiftlint analyze`
   - explicit_self
 
-included: # case-sensitive paths to include during linting.
+# case-sensitive paths to include during linting. Directory paths supplied on the
+# command line will be ignored.
+included: 
   - Sources
 excluded: # case-sensitive paths to ignore during linting. Takes precedence over `included`
   - Carthage
--- a/README.md
+++ b/README.md
@@ -680,7 +680,7 @@ opt_in_rules: # some rules are turned off by default, so you need to opt-in
 analyzer_rules: # rules run by `swiftlint analyze`
   - explicit_self
 
-# case-sensitive paths to include during linting. Directory paths supplied on the
+# Case-sensitive paths to include during linting. Directory paths supplied on the
 # command line will be ignored.
 included: 
   - Sources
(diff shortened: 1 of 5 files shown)


Source/swiftlint/Commands/Analyze.swift
```diff
@@ -20,13 +18,7 @@ extension SwiftLint {
...
-                // TODO: [06/14/2024] Remove deprecation warning after ~2 years.
-                Issue.genericWarning(
-                    "The --path option is deprecated. Pass the path(s) to analyze last to the swiftlint command."
-                ).print()
-                allPaths = [path] + paths
-            } else if !paths.isEmpty {
+            if !paths.isEmpty {
                 allPaths = paths
             } else {
                 allPaths = [""] // Analyze files in current working directory if no paths were specified.
```
SimplyDanny, line 24: We can now initialize `allPaths` in line 20 directly.

README.md
```diff
@@ -668,7 +668,9 @@ opt_in_rules: # some rules are turned off by default, so you need to opt-in
 analyzer_rules: # rules run by `swiftlint analyze`
   - explicit_self
 
-included: # case-sensitive paths to include during linting. `--path` is ignored if present
+# case-sensitive paths to include during linting. Directory paths supplied on the
```
SimplyDanny, line 671:
```suggestion
# Case-sensitive paths to include during linting. Directory paths supplied on the
```

Source/swiftlint/Commands/Lint.swift
```diff
@@ -27,21 +25,12 @@ extension SwiftLint {
...
-
             if common.fix, let leniency = common.leniency {
                 Issue.genericWarning("The option --\(leniency) has no effect together with --fix.").print()
             }
 
             let allPaths =
-                if let path {
-                    [path] + paths
-                } else if !paths.isEmpty {
+                if !paths.isEmpty {
```
SimplyDanny, line 33: Perhaps use ternary operator now.

SimplyDanny (approved):

PR #5048, opened 2023-06-03
Rewrite `missing_docs` rule with SwiftSyntax

woxtu:
Hi,

As the title says, this PR rewrites `missing_docs` rule with SwiftSyntax.

PR #5617, opened 2024-06-04
add opt-in  non_empty_function rule

### LINKED ISSUE #5615: Rule Request: No Empty Function
### New Issue Checklist

### New rule request
A Function body should not be empty. Otherwise, show a warning.

An empty function, init and deinit can harm readability and lead to confusion because readers need to guess whether it's intentional or not.
Developers should put a comment to explain why the function/ init / deinit body is empty.

[SonarLint for Swift](https://rules.sonarsource.com/swift/tag/suspicious/RSPEC-1186) and [ESLint](https://eslint.org/docs/latest/rules/no-empty-function) have the same rule.
SwiftLint can be better by having this rule.

##### Triggering:
```
func f1 () {}

func f2 () {
}

// Comment
func f3 () {}

init() {}

deinit {}
```

##### Non-Triggering
```
func f4 () {
    let s = "xyz"
}

func f5 () { /* Comment*/  }

func f6() {
    // Comment
}

init() { /* comment */ }

deinit { /* comment */ }
```

Just so you know, this behavior is the same as SonarLint and ESLint.

### Should be configurable?
No parameter to configure.

### Should be opt-in?
In my opinion, this rule should be ~~"opt-in rule"~~ since there is no consensus, though it's active by default in SonarLint and ESLint.
Got feed back, and change  to default.

If this rule makes sense, I will work for this!
Thanks.

### Future enhancement
We can implement rules for closures and ~~initializers~~ as well.
But, they should be implemented as separate rules.

Ueeek:
resolve #5615 
Add `no_empty_function` rule

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,8 @@
 * Linting got around 20% faster due to the praisworthy performance
   improvements done in the [SwiftSyntax](https://github.com/apple/swift-syntax)
   library.
-*  Add `no_empty_block` default rule to validate that `function`, `init`, `deinit`, `catch` and `defer` should not have empty code block. They should at least contain a comment.
+* Add `no_empty_block` default rule to validate that code blocks should not be empty.
+They should at least contain a comment.
     [Ueeek](https://github.com/Ueeek)
     [#5615](https://github.com/realm/SwiftLint/issues/5615)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@
   improvements done in the [SwiftSyntax](https://github.com/apple/swift-syntax)
   library.
 * Add `no_empty_block` default rule to validate that code blocks should not be empty.
-They should at least contain a comment.
+  They should at least contain a comment.
     [Ueeek](https://github.com/Ueeek)
     [#5615](https://github.com/realm/SwiftLint/issues/5615)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
   library.
 *  Add `non_empty_function` opt-in rule to validate that functions should not have empty body. Otherwise show warning to write comments.
 *  Add `non_empty_function` opt-in rule to validate that `function`, `init` and `deinit` should not have empty body. Otherwise show warning to ask to write comments.
+*  Add `non_empty_function` default rule to validate that `function`, `init` and `deinit` should not have empty body. Otherwise show warning to ask to write comments.
     [Ueeek](https://github.com/Ueeek)
     [#5615](https://github.com/realm/SwiftLint/issues/5615)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@
 * Linting got around 20% faster due to the praisworthy performance
   improvements done in the [SwiftSyntax](https://github.com/apple/swift-syntax)
   library.
+*  Add `non_empty_function` opt-in rule to validate that functions should not have empty body. Otherwise show warning to write comments.
+    [Ueeek](https://github.com/Ueeek)
+    [#5615](https://github.com/realm/SwiftLint/issues/5615)
 
 * Adds `baseline` and `write_baseline` configuration file settings, equivalent
   to the `--baseline` and `--write-baseline` command line options.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,8 +24,9 @@
 * Linting got around 20% faster due to the praisworthy performance
   improvements done in the [SwiftSyntax](https://github.com/apple/swift-syntax)
   library.
-* Add `no_empty_block` default rule to validate that code blocks should not be empty.
-  They should at least contain a comment.
+
+* Add `no_empty_block` default rule to validate that code blocks are not empty.
+  They should at least contain a comment.  
     [Ueeek](https://github.com/Ueeek)
     [#5615](https://github.com/realm/SwiftLint/issues/5615)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@
 * Linting got around 20% faster due to the praisworthy performance
   improvements done in the [SwiftSyntax](https://github.com/apple/swift-syntax)
   library.
-*  Add `non_empty_function_body` default rule to validate that `function`, `init` and `deinit` should not have empty body. Otherwise show warning to ask to write comments.
+*  Add `no_empty_block` default rule to validate that `function`, `init`, `deinit`, `catch and `defer` should not have empty code block. They should at least contain a comment.
     [Ueeek](https://github.com/Ueeek)
     [#5615](https://github.com/realm/SwiftLint/issues/5615)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@
 * Linting got around 20% faster due to the praisworthy performance
   improvements done in the [SwiftSyntax](https://github.com/apple/swift-syntax)
   library.
-*  Add `no_empty_block` default rule to validate that `function`, `init`, `deinit`, `catch and `defer` should not have empty code block. They should at least contain a comment.
+*  Add `no_empty_block` default rule to validate that `function`, `init`, `deinit`, `catch` and `defer` should not have empty code block. They should at least contain a comment.
     [Ueeek](https://github.com/Ueeek)
     [#5615](https://github.com/realm/SwiftLint/issues/5615)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,15 +25,15 @@
   improvements done in the [SwiftSyntax](https://github.com/apple/swift-syntax)
   library.
 
-* Add `no_empty_block` default rule to validate that code blocks are not empty.
-  They should at least contain a comment.  
-    [Ueeek](https://github.com/Ueeek)
-    [#5615](https://github.com/realm/SwiftLint/issues/5615)
-
 * Adds `baseline` and `write_baseline` configuration file settings, equivalent
   to the `--baseline` and `--write-baseline` command line options.  
   [Martin Redington](https://github.com/mildm8nnered)
   [#5552](https://github.com/realm/SwiftLint/issues/5552)
+
+* Add `no_empty_block` default rule to validate that code blocks are not empty.
+  They should at least contain a comment.  
+    [Ueeek](https://github.com/Ueeek)
+    [#5615](https://github.com/realm/SwiftLint/issues/5615)
 
 #### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
   improvements done in the [SwiftSyntax](https://github.com/apple/swift-syntax)
   library.
 *  Add `non_empty_function` opt-in rule to validate that functions should not have empty body. Otherwise show warning to write comments.
+*  Add `non_empty_function` opt-in rule to validate that `function`, `init` and `deinit` should not have empty body. Otherwise show warning to ask to write comments.
     [Ueeek](https://github.com/Ueeek)
     [#5615](https://github.com/realm/SwiftLint/issues/5615)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,9 +24,7 @@
 * Linting got around 20% faster due to the praisworthy performance
   improvements done in the [SwiftSyntax](https://github.com/apple/swift-syntax)
   library.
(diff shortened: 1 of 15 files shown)


mildm8nnered:
Should `init` and `deinit` be checked as well?

SimplyDanny (changes requested):
Thank you for implementing the rule!

Such a check would actually also make sense for `defer` blocks and empty `struct`/`class`/... declarations. That can be up to an extension. However, to be prepared a name like `NoEmptyBlockRule` would leave room for these cases.

SimplyDanny:
> Such a check would actually also make sense for `defer` blocks and empty `struct`/`class`/... declarations. That can be up to an extension. However, to be prepared a name like `NoEmptyBlockRule` would leave room for these cases.

... and to `catch` blocks as well like requested in #5634.

Ueeek:
> Collaborator

@SimplyDanny  Thank you for your review and feedback!

And, thank you for suggesting `catch`, `defer`, `class`, ...

IMHO, it might be better to implement these cases as separate rules.
e.g.
- NoEmptyFunctionBody for `function`, `init` and `deinit` <- This PR.
- NoEmptyTypeDeclarationRule for `class`, `struct`, `enum` and `actor`.
- NoEmptyClosureRule for `closure`, `catch` and `defer`.

This separation will be consistent with `body length` rules. (functionBodyLength, closureBodyLength, and TypeBodyLength).

If it's better to implement them together in a single rule, I change the name of this rule to the `NoEmptyBlock` rule and I can implement them in this PR.
If it's better to implement them as separate rules, I change the name of this rule to the `NoEmptyFunctionBody` rule to avoid future confusion. And I can implement them in other PRs.

I would like to hear your feedback! Thank you!

Ueeek:
@SimplyDanny Thank you for reviewing ~.
I fixed the review points.
Also, support `catch` and `defer` as well.
And change the rule name as suggested!

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@
 * Linting got around 20% faster due to the praisworthy performance
   improvements done in the [SwiftSyntax](https://github.com/apple/swift-syntax)
   library.
+*  Add `no_empty_block` default rule to validate that `function`, `init`, `deinit`, `catch` and `defer` should not have empty code block. They should at least contain a comment.
```
mildm8nnered, line 18: Probably want to wrap this text the same as the rest of the file

Source/SwiftLintCore/Helpers/Stack.swift
```diff
@@ -3,7 +3,7 @@ public struct Stack<Element> {
     private var elements = [Element]()
 
     /// Creates an empty `Stack`.
-    public init() {}
+    public init() { /* Publish no-op initializer */ }
```
mildm8nnered, line 6: If this is a legitimate exception, because it's making the constructor public, could we detect and ignore it. I think this would only apply to `init` as far as I can think of.

Source/SwiftLintCore/Models/RuleRegistry.swift
```diff
@@ -12,7 +12,7 @@ public final class RuleRegistry: @unchecked Sendable {
     ///         accessed will not work.
     public private(set) lazy var list = RuleList(rules: registeredRules)
 
-    private init() {}
+    private init() { /* To guarantee that this is singleton. */ }
```
mildm8nnered, line 15: Also maybe this could be an exception too

Source/SwiftLintCore/Rules/SuperfluousDisableCommandRule.swift
```diff
@@ -1,7 +1,7 @@
 package struct SuperfluousDisableCommandRule: SourceKitFreeRule {
     package var configuration = SeverityConfiguration<Self>(.warning)
 
-    package init() {}
+    package init() { /* Make initializer as accessible as its type. */ }
```
mildm8nnered, line 4: Maybe here too - if an `init` has any accessibility modifier apart from `internal`, then there is a good reason it might be empty.

SimplyDanny:
Regarding accessibility modifiers on initializers: While it might be true that explicit modifiers have a potential meaning and are there for a good reason, it's not too easy to find out if the modifier is really required. See how I struggle in #5048 for example and consider

```swift
private struct S {
    private init() {}
}
```

where the `private` on the initializer is redundant.

So in the first shot of this rule, I'd require the comment in every empty block disregarding access control modifiers. It also doesn't hurt to enforce an explaining comment for these cases. Later we can add this specialization optionally.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoEmptyBlockRule.swift
```diff
@@ -0,0 +1,104 @@
...
+
+@SwiftSyntaxRule
+struct NoEmptyBlockRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "no_empty_block",
+        name: "No Empty Block",
+        // swiftlint:disable:next line_length
+        description: "Code Blocks(Function bodies, catch, defer) should not be empty; they should at least contain a comment",
```
SimplyDanny, line 11: Since this is used as the violations's message (as shown on the console or in IDEs and editors), it should be reasonable short. So I would say the following is sufficient:
```suggestion
        description: "Code blocks should not be empty; they should at least contain a comment",
```

SimplyDanny:
Any reason why

```swift
do {}
for _ in 1..<12 {}
while i < 12 {}
```

should not trigger the rule?

Ueeek:
Instead of validating each syntax, validate [`CodeBlockSyntax`](https://swiftpackageindex.com/swiftlang/swift-syntax/600.0.0-prerelease-2024-06-12/documentation/swiftsyntax/codeblocksyntax) so that we can support all type of CodeBlocks.(Accessor, DoCatch, Defer, Deinit, For, Function, if-else, init, repeat and while).

CHANGELOG.md
```diff
@@ -15,6 +15,10 @@
 * Linting got around 20% faster due to the praisworthy performance
   improvements done in the [SwiftSyntax](https://github.com/apple/swift-syntax)
   library.
+* Add `no_empty_block` default rule to validate that code blocks should not be empty.
```
SimplyDanny, line 18: Please add an empty line above.

CHANGELOG.md
```diff
@@ -15,6 +15,10 @@
...
+  They should at least contain a comment.
```
SimplyDanny, line 19:
```suggestion
  They should at least contain a comment.  
```

Please append two spaces to the end of the description. They are required to enforce a line break in the rendered Markdown.

CHANGELOG.md
SimplyDanny, line 18:
```suggestion
* Add `no_empty_block` default rule to validate that code blocks are not empty.
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoEmptyBlockRule.swift
```diff
@@ -0,0 +1,114 @@
+import SwiftSyntax
+
+@SwiftSyntaxRule
+struct NoEmptyBlockRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "no_empty_block",
+        name: "No Empty Block",
+        description: "Code blocks should not be empty; they should at least contain a comment",
```
SimplyDanny, line 10: I'm not too happy with this message yet. Twice the "should" bugs me. @mildm8nnered, could you help out?

SimplyDanny (approved):
Just a few more nits. Other than that, it looks good.

> Instead of validating each syntax, validate CodeBlockSyntax so that we can support all type of CodeBlocks.(Accessor, DoCatch, Defer, Deinit, For, Function, if-else, init, repeat and while).

Very good!

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoEmptyBlockRule.swift
mildm8nnered: So looking through the other rule descriptions, they are mostly single clauses, and negative ("should not"), but sometimes positive. How about something like:

"Code blocks should contain at least one statement or comment"

Or, using "avoid", which occurs in some existing rule descriptions:

"Empty blocks with no statements or comments should be avoided"

SimplyDanny: I like the first suggestion.

Ueeek:
@mildm8nnered @SimplyDanny  Thank you for comments! Fixed all point!

SimplyDanny:
Awesome first contribution! Thank you, @Ueeek.

marcelofabri:
I think it’s a bad idea to enable this rule by default. Besides the initializer cases that were caught here, a common case would be creating mock types that need to conform to a protocol, but don’t care about a specific method.

SimplyDanny:
> I think it’s a bad idea to enable this rule by default. Besides the initializer cases that were caught here, a common case would be creating mock types that need to conform to a protocol, but don’t care about a specific method.

I have no strong opinion on that. In my world, all rules could be optional and I enable them specifically. However, it seems like people are often not aware of new  optional rules. That's unfortunate. So having some default rules (with a low false positive rate) is actually a nice way to avoid helpful and likeable rules being unused.

Mocks usually appear in test code. Isn't it fine then to disable the rule in these files?

marcelofabri:
> > I think it’s a bad idea to enable this rule by default. Besides the initializer cases that were caught here, a common case would be creating mock types that need to conform to a protocol, but don’t care about a specific method.
> 
> I have no strong opinion on that. In my world, all rules could be optional and I enable them specifically. However, it seems like people are often not aware of new optional rules. That's unfortunate. So having some default rules (with a low false positive rate) is actually a nice way to avoid helpful and likeable rules being unused.
> 
> Mocks usually appear in test code. Isn't it fine then to disable the rule in these files?

It’s a big ask to disable the rule in some files in a large enough codebase. The number of violations in OSS check tells me this should be opt-in.

SimplyDanny:
> > > I think it’s a bad idea to enable this rule by default. Besides the initializer cases that were caught here, a common case would be creating mock types that need to conform to a protocol, but don’t care about a specific method.
> > 
> > 
> > I have no strong opinion on that. In my world, all rules could be optional and I enable them specifically. However, it seems like people are often not aware of new optional rules. That's unfortunate. So having some default rules (with a low false positive rate) is actually a nice way to avoid helpful and likeable rules being unused.
> > Mocks usually appear in test code. Isn't it fine then to disable the rule in these files?
> 
> It’s a big ask to disable the rule in some files in a large enough codebase. The number of violations in OSS check tells me this should be opt-in.

What about adding a configuration that allows to disable the rule for certain constructs? Options could be `initializer_bodies` (which includes deinitializers), `function_bodies` and `statement_blocks` (which includes `if`, `for`, `defer`, ...).

Ueeek:
@SimplyDanny @marcelofabri 
Thank you for suggesting the improvement!
The above solution of adding configurations looks good for me.
I will be happy to work for it after concluded.

SimplyDanny:
> What about adding a configuration that allows to disable the rule for certain constructs? Options could be `initializer_bodies` (which includes deinitializers), `function_bodies` and `statement_blocks` (which includes `if`, `for`, `defer`, ...).

@marcelofabri: Would that be acceptable for you? `initializer_bodies` would be the only one being enabled by default.

marcelofabri:
> > What about adding a configuration that allows to disable the rule for certain constructs? Options could be `initializer_bodies` (which includes deinitializers), `function_bodies` and `statement_blocks` (which includes `if`, `for`, `defer`, ...).
> 
> @marcelofabri: Would that be acceptable for you? `initializer_bodies` would be the only one being enabled by default.

I still think this should be opt-in. `initializer_bodies` being enabled by default will still trigger several false positives in cases where a `init` needs to be declared.

SimplyDanny:
> > > What about adding a configuration that allows to disable the rule for certain constructs? Options could be `initializer_bodies` (which includes deinitializers), `function_bodies` and `statement_blocks` (which includes `if`, `for`, `defer`, ...).
> > 
> > 
> > @marcelofabri: Would that be acceptable for you? `initializer_bodies` would be the only one being enabled by default.
> 
> I still think this should be opt-in. `initializer_bodies` being enabled by default will still trigger several false positives in cases where a `init` needs to be declared.

Fine for me. So let's make it opt-in and add the options as suggested in https://github.com/realm/SwiftLint/pull/5617#issuecomment-2198131979.

Ueeek:
Created PR https://github.com/realm/SwiftLint/pull/5685
It's contains
+ Change rule to OptIn
+ Add configuration

Issue #5598, opened 2024-05-21
`opening_brace` fix is broken


kapitoshka438:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

Source file:
```
func foo() {
    if q1, q2
    {
        do1()
    } else if q3, q4
    {
        do2()
    }
}

```

`swiftlint --fix`

Expected result:
```
func foo() {
    if q1, q2 {
        do1()
    } else if q3, q4 {
        do2()
    }
}

```

Actual result:
```
func foo() {
    if q1, q2 {
        do1()
    } else if q3, q4
          do2()
    }
}

```

### Environment

* SwiftLint version 0.55.1

PR #5620, opened 2024-06-11
Fix `opening_brace` correction

swiftty:
- fixes https://github.com/realm/SwiftLint/issues/5598 and disable command ignored on correcting

### DIFF
--- a/Source/SwiftLintCore/Extensions/SwiftLintFile+Regex.swift
+++ b/Source/SwiftLintCore/Extensions/SwiftLintFile+Regex.swift
@@ -273,6 +273,7 @@ extension SwiftLintFile {
         return violatingRanges.filter { range in
             let region = fileRegions.first {
                 $0.contains(Location(file: self, characterOffset: range.location))
+                || $0.contains(Location(file: self, characterOffset: range.location + range.length - 1))
             }
             return region?.isRuleEnabled(rule) ?? true
         }
--- a/Source/SwiftLintCore/Extensions/SwiftLintFile+Regex.swift
+++ b/Source/SwiftLintCore/Extensions/SwiftLintFile+Regex.swift
@@ -273,7 +273,6 @@ extension SwiftLintFile {
         return violatingRanges.filter { range in
             let region = fileRegions.first {
                 $0.contains(Location(file: self, characterOffset: range.location))
-                || $0.contains(Location(file: self, characterOffset: range.location + range.length - 1))
             }
             return region?.isRuleEnabled(rule) ?? true
         }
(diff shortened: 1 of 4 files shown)


Naveen-C-Ramachandrappa (approved):
Looks good.

Source/SwiftLintCore/Extensions/SwiftLintFile+Regex.swift
```diff
@@ -273,6 +273,7 @@ extension SwiftLintFile {
         return violatingRanges.filter { range in
             let region = fileRegions.first {
                 $0.contains(Location(file: self, characterOffset: range.location))
+                || $0.contains(Location(file: self, characterOffset: range.location + range.length - 1))
```
SimplyDanny, line 276: While this seems like a reasonable change, it has the potential of changing other corrections of rules as well (which we don't see in the OSS report). I'd rather like to avoid that. This will anyway become a thing if the past once these rules are converted to SwiftSyntax rules.

Actually, the decision if a rule is enabled or not should be the same for violations and corrections. In your example, a violations is reported, but no fix is applied. That's inconsistent. So we have to make sure that the "Is rule enabled?" check is implemented in the same way or at least can rely on the same positions in the "reporting" and "fix" cases, probably by extending `ViolationCorrection` with a `violationPosition`. Later, `ReasonedRuleViolation` and `ViolationCorrection` can even be merged.

SimplyDanny (changes requested):

Source/SwiftLintCore/Extensions/SwiftLintFile+Regex.swift
SimplyDanny: This has been addressed by #5680 and #5682.

Source/SwiftLintCore/Protocols/SwiftSyntaxCorrectableRule.swift
```diff
@@ -50,7 +50,7 @@ public extension SwiftSyntaxCorrectableRule {
                 }
             }
             .filter { file.ruleEnabled(violatingRange: $0.range, for: self) != nil }
-            .reversed()
+            .sorted(by: { $0.range.location > $1.range.location })
```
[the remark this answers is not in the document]
SimplyDanny: Rebased.

SimplyDanny:
@swiftty, sorry that I was too late with the review here. Parts of your PR had already been fixed by other measures or required more extensive refactoring. Thank you nevertheless for the valuable contribution! Please keep up the good work. 😊

I rebased the branch, so that everything should be good to go in a while.

swiftty:
@SimplyDanny, I appreciate your courteous response in fixing the code and merging the PR.

PR #5621, opened 2024-06-11
Fix type inside shell script in README.md

miltenkot:
I notice wrong typo which doesn't works properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ validations can be disabled with either of the following:
 * Setting Xcode defaults:
 
   ```bash
-  defaults write com.apple.dt.Xcode IDESkipPackagePluginFingerprintValidation -bool YES
+  defaults write com.apple.dt.Xcode IDESkipPackagePluginFingerprintValidatation -bool YES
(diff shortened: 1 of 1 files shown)


SimplyDanny (changes requested):
I don't see a typo in the original word. "Validation" is correct.

miltenkot:
> I don't see a typo in the original word. "Validation" is correct.

Yes, however, does not work correctly. I checked with this article, and it should be entered with a misspelling (validataion) to work with CI (Xcode Cloud). (https://forums.swift.org/t/telling-xcode-14-beta-4-to-trust-build-tool-plugins-programatically/59305/10)

SimplyDanny:
> > I don't see a typo in the original word. "Validation" is correct.
> 
> Yes, however, does not work correctly. I checked with this article, and it should be entered with a misspelling (validataion) to work with CI (Xcode Cloud). (https://forums.swift.org/t/telling-xcode-14-beta-4-to-trust-build-tool-plugins-programatically/59305/10)

I see, that wasn't obvious to me. It's still hard to approve a PR that introduces a typo ... I've never done that before. 😅

SimplyDanny (approved):

PR #5629, opened 2024-06-16
Temporarily run builds as GitHub actions

SimplyDanny:
Our build machines currently require some maintenance. During this time, all PRs are blocked. This change moves the builds to GitHub actions so that we get at least successful builds even though the performance results should not be trusted. We need to decide per change if it could have a performance impact and then rather wait until the more reliable build machines are back.

The commit changing the build scripts can just be reverted once the physical build machines are working again.


SimplyDanny:
Cannot run Danger from a fork branch due to https://github.com/danger/danger/issues/1103.

PR #5640, opened 2024-06-22
Enforce mandatory trailing comma and fix all violations

SimplyDanny:


mildm8nnered (approved):
Seems reasonable to me

PR #5649, opened 2024-07-01
Added prefer_type_checking rule

mildm8nnered:
I wrote this unaware of #5561

I think that PR should take precedence over this one, but I'm just leaving this here in case @ikelax wants to steal anything from it.


mildm8nnered:
Replaced by https://github.com/realm/SwiftLint/pull/5561

PR #5650, opened 2024-07-02
Add `ignore_codingkeys` parameter to nesting

### LINKED ISSUE #5641: Rule Request: Ignore CodingKey enums in Nesting
### New Issue Checklist

### New rule request

This is more of an edit to an existing rule than a new rule, so apologies if this is the wrong format... Add a boolean option to the Nesting rule. When this option is enabled, enum's conforming to `CodingKey` won't count as violations of nesting.

My main argument for this is that CodingKey's are less of a nested "type" and more of an annotation of the current type for `Codable` If you have types that have custom `CodingKey` types, you basically lose a 

**Triggering**
```swift
struct Wrapper: Codable {
    struct Inner: Codable {
        struct WayInner: Codable {
        }
    }
}
```

```swift
struct Wrapper: Codable {
    enum CodingKeys: String, CodingKey {
        case id = "identifier"

        struct TechnicallyOkayButBad {
            struct Violation {}
         }
    }
}
```

**Non Triggering**
```swift
struct Wrapper: Codable {
    struct Inner: Codable {
        let id: Int
        
        enum CodingKeys: String, CodingKey {
            case id = "identifier
        }
    }
}
```

I think I would propose this be an **opt-in setting** on the nesting rule.

 Looking at the criteria in the README
> A  rule that can have many false positives

I think this change would actually reduce false positives

> A rule that is too slow

I don't _think_ this would be an issue

> A rule that is not general consensus or is only useful in some cases

 I'm not sure if there's consensus here. And it's a very specific case...

braker1nine:
Resolves #5641 

This very explicitly targets `enum`'s named `CodingKeys` which conform to `CodingKey`. To provide an exception for `Codable` users who have to provide custom keys. 

OK ✅ 
```swift
struct Outer {
    struct Inner: Codable {
        let identifier: Int
        enum CodingKeys: String, CodingKey {
            case identifier = "id"
        }
    }
}
```
---

Not OK ❌ 
```swift
struct Outer {
    struct Inner {
        let identifier: Int
        enum SomeKeys: String, CodingKey {
            case identifier = "id"
        }
    }
}
```

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
@@ -64,7 +64,9 @@ private extension NestingRule {
         }
 
         override func visit(_ node: EnumDeclSyntax) -> SyntaxVisitorContinueKind {
-            validate(forFunction: false, triggeringToken: node.enumKeyword)
+            if !(configuration.ignoreCodingKeys && node.isCodingKey) {
+                validate(forFunction: false, triggeringToken: node.enumKeyword)
+            }
             return .visitChildren
         }
 
--- a/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/NestingConfiguration.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/NestingConfiguration.swift
@@ -15,6 +15,8 @@ struct NestingConfiguration: RuleConfiguration {
     private(set) var alwaysAllowOneTypeInFunctions = false
     @ConfigurationElement(key: "ignore_typealiases_and_associatedtypes")
     private(set) var ignoreTypealiasesAndAssociatedtypes = false
+    @ConfigurationElement(key: "ignore_codingkeys")
+    private(set) var ignoreCodingKeys = false
 
     func severity(with config: Severity, for level: Int) -> ViolationSeverity? {
         if let error = config.error, level > error {
--- a/Source/SwiftLintCore/Extensions/SwiftSyntax+SwiftLint.swift
+++ b/Source/SwiftLintCore/Extensions/SwiftSyntax+SwiftLint.swift
@@ -184,6 +184,25 @@ public extension EnumDeclSyntax {
             return rawValueTypes.contains(identifier)
         }
     }
+
+    /// True if this enum is a `CodingKey`
+    var isCodingKey: Bool {
+        guard let inheritedTypeCollection = inheritanceClause?.inheritedTypes else {
+            return false
+        }
+        
+        guard self.name.text == "CodingKeys" else {
+            return false
+        }
+
+        return inheritedTypeCollection.contains { element in
+            guard let identifier = element.type.as(IdentifierTypeSyntax.self)?.name.text else {
+                return false
+            }
+
+            return identifier == "CodingKey"
+        }
+    }
 }
 
 public extension FunctionDeclSyntax {
--- a/Tests/SwiftLintFrameworkTests/NestingRuleTests.swift
+++ b/Tests/SwiftLintFrameworkTests/NestingRuleTests.swift
@@ -551,4 +551,38 @@ final class NestingRuleTests: SwiftLintTestCase {
 
         verifyRule(description, ruleConfiguration: ["ignore_typealiases_and_associatedtypes": true])
     }
+    
+    func testNestingWithoutCodingKeys() {
+        var nonTriggeringExamples = NestingRule.description.nonTriggeringExamples
+        nonTriggeringExamples.append(contentsOf: [
+            .init("""
+                struct Outer {
+                    struct Inner {
+                        enum CodingKeys: String, CodingKey {
+                            case id
+                        }
+                    }
+                }
+                """
+             )
+        ])
+        
+        var triggeringExamples = NestingRule.description.triggeringExamples
+        triggeringExamples.append(contentsOf: [
+            .init("""
+                struct Outer {
+                    struct Inner {
+                        enum Example: String, CodingKey {
+                            case id
+                        }
+                    }
+                }
+                """)
+        ])
+        
+        let description = NestingRule.description.with(nonTriggeringExamples: nonTriggeringExamples, triggeringExamples: triggeringExamples)
+        
+        verifyRule(description, ruleConfiguration: ["ignore_codingkeys": true ])
+    }
+
 }


Source/SwiftLintCore/Extensions/SwiftSyntax+SwiftLint.swift
```diff
@@ -184,6 +184,25 @@ public extension EnumDeclSyntax {
             return rawValueTypes.contains(identifier)
         }
     }
+
+    /// True if this enum is a `CodingKey`
```
SimplyDanny, line 188:
```suggestion
    /// True if this enum is a `CodingKey`. For that, it has to be named `CodingKeys` and must conform to the `CodingKey` protocol. 
```

Source/SwiftLintCore/Extensions/SwiftSyntax+SwiftLint.swift
```diff
@@ -184,6 +184,25 @@ public extension EnumDeclSyntax {
...
+    var isCodingKey: Bool {
```
SimplyDanny, line 189: The enum itself is not a coding key, but it defines the coding keys, doesn't it? So what about the name `definesCodingKeys` instead?

Source/SwiftLintCore/Extensions/SwiftSyntax+SwiftLint.swift
```diff
@@ -184,6 +184,25 @@ public extension EnumDeclSyntax {
...
+        guard let inheritedTypeCollection = inheritanceClause?.inheritedTypes else {
+            return false
+        }
+        
+        guard self.name.text == "CodingKeys" else {
```
SimplyDanny, line 194:
```suggestion
        guard name.text == "CodingKeys" else {
```

Can also be added to the previous `guard`.

Source/SwiftLintCore/Extensions/SwiftSyntax+SwiftLint.swift
```diff
@@ -184,6 +184,25 @@ public extension EnumDeclSyntax {
...
+            return false
+        }
+
+        return inheritedTypeCollection.contains { element in
+            guard let identifier = element.type.as(IdentifierTypeSyntax.self)?.name.text else {
+                return false
+            }
+
+            return identifier == "CodingKey"
```
SimplyDanny, line 203:
```suggestion
            element.type.as(IdentifierTypeSyntax.self)?.name.text == "CodingKey"
```

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -64,7 +64,9 @@ private extension NestingRule {
         }
 
         override func visit(_ node: EnumDeclSyntax) -> SyntaxVisitorContinueKind {
-            validate(forFunction: false, triggeringToken: node.enumKeyword)
+            if !(configuration.ignoreCodingKeys && node.isCodingKey) {
```
SimplyDanny, line 67:
```suggestion
            if !configuration.ignoreCodingKeys || !node.isCodingKey {
```

Tests/SwiftLintFrameworkTests/NestingRuleTests.swift
```diff
@@ -551,4 +551,38 @@ final class NestingRuleTests: SwiftLintTestCase {
...
+                """
+             )
+        ])
+        
+        var triggeringExamples = NestingRule.description.triggeringExamples
+        triggeringExamples.append(contentsOf: [
+            .init("""
+                struct Outer {
+                    struct Inner {
+                        enum Example: String, CodingKey {
```
SimplyDanny, line 575:
```suggestion
                        ↓enum Example: String, CodingKey {
```

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/NestingConfiguration.swift
```diff
@@ -15,6 +15,8 @@ struct NestingConfiguration: RuleConfiguration {
     private(set) var alwaysAllowOneTypeInFunctions = false
     @ConfigurationElement(key: "ignore_typealiases_and_associatedtypes")
     private(set) var ignoreTypealiasesAndAssociatedtypes = false
+    @ConfigurationElement(key: "ignore_codingkeys")
```
SimplyDanny, line 18:
```suggestion
    @ConfigurationElement(key: "ignore_coding_keys")
```

SimplyDanny (changes requested):
Thank you for the contribution! Looks good in principle.

What about the following example?

```swift
struct Outer {
  enum CodingKeys: String, CodingKey {
    case id
    
    struct S {}
  }
}
```

I think, the rule should trigger on `S`.

You need to run `IntegrationTests` locally and fix all linting issues it reports and update the configuration set reference file.

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@
   improvements done in the [SwiftSyntax](https://github.com/apple/swift-syntax)
   library.
 
+* Add `ignore_codingkeys` parameter to `nesting` rule.
```
SimplyDanny, line 19: Please explain briefly what this option does.

```suggestion
* Add `ignore_coding_keys` parameter to `nesting` rule.  
```

braker1nine:
FYI I'm out of town until next weekend. I'll jump back into this once I'm back

braker1nine:
> FYI I'm out of town until next weekend. I'll jump back into this once I'm back

It wasn't a year long vacation but here we are a year later 😓 

 
> What about the following example?
> 
> ```swift
> struct Outer {
>   enum CodingKeys: String, CodingKey {
>     case id
>     
>     struct S {}
>   }
> }
> ```
> 
> I think, the rule should trigger on `S`.

I've updated this so that this will now trigger the rule.

> You need to run `IntegrationTests` locally and fix all linting issues it reports and update the configuration set reference file.

I think I've got this done? I moved the `EnumDeclSyntax` extension into its own file because the `SwiftSyntax+SwiftLint` file was too long and failing linting.

I don't fully understand what is failing now?

CHANGELOG.md
```diff
@@ -14,8 +14,11 @@
 
 ### Enhancements
 
-* None.
-
+* Add `ignore_coding_keys` parameter to `nesting` rule. Setting this to true prevents
+  `CodingKey` enums from violating the rule
```
```suggestion
  `CodingKey` enums from violating the rule.  
```

Source/SwiftLintCore/Extensions/SwiftSyntax+SwiftLint.swift
```diff
@@ -169,30 +169,6 @@ public extension VariableDeclSyntax {
     }
 }
 
-public extension EnumDeclSyntax {
```
SimplyDanny: You may keep this here and add you extension. A `// swiftlint:disable file_length` command is fine for this file.

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
```diff
@@ -152,6 +152,12 @@ private extension NestingRule {
             if configuration.alwaysAllowOneTypeInFunctions && inFunction && !forFunction {
                 return
             }
+
+            // if current defines coding keys and we're ignoring coding keys, then skip nesting rule
+            if configuration.ignoreCodingKeys && definesCodingKeys {
```
SimplyDanny: Can we move this into `visit(_ node: EnumDeclSyntax)` to avoid the extra argument?

Tests/BuiltInRulesTests/NestingRuleTests.swift
```diff
@@ -555,4 +555,48 @@ final class NestingRuleTests: SwiftLintTestCase {
 
         verifyRule(description, ruleConfiguration: ["ignore_typealiases_and_associatedtypes": true])
     }
+
+    func testNestingWithoutCodingKeys() {
```
SimplyDanny: You can move these examples directly into `NestingRuleExamples`. Any options deviating from the default will be added to the online documentation, so it's clear why an example triggers or why it doesn't.

Tests/BuiltInRulesTests/NestingRuleTests.swift
```diff
@@ -555,4 +555,48 @@ final class NestingRuleTests: SwiftLintTestCase {
...
+            """),
+            .init("""
+                struct Outer {
+                  enum CodingKeys: String, CodingKey {
+                    case id
+                    ↓struct S {}
+                  }
+                }
+            """),
+        ])
```
SimplyDanny: Let's have another example that triggers on a `CodingKeys` enum when the option is set to `false` (the default).

SimplyDanny (changes requested):
Glad you are back! 🥳 I have a few more comments ...

SimplyDanny:
For the build, please rebase onto the current `main`. The results are outdated and there is conflict as well.

Source/SwiftLintCore/Extensions/SwiftSyntax+SwiftLint.swift
braker1nine: 👍 is there a preferred place to put that disable? Top of the source file?

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
braker1nine: Hmmm yeah I think I could I would need to ensure we push another item onto `self.levels` in `visit(_ node: EnumDeclSyntax)`. To ensure the `CodingKeys` add another level to the nesting count without violating it themselves. 

```swift
override func visit(_ node: EnumDeclSyntax) -> SyntaxVisitorContinueKind {

    // if current defines coding keys and we're ignoring coding keys, then skip nesting rule
    // push another level on and proceed to visit children
    if configuration.ignoreCodingKeys && node.definesCodingKeys {
        levels.push(levels.lastIsFunction)
    } else {
        validate(forFunction: false, triggeringToken: node.enumKeyword)
    }

    return .visitChildren
}
```

But I that does seem a little simpler? Doesn't require all the other `visit` methods to add this parameter

Tests/BuiltInRulesTests/NestingRuleTests.swift
braker1nine: I might need just a little bit more clarification what you mean by this? I was thinking because these are not using the default configuration they shouldn't be in the default examples.

But you're saying they should?

Source/SwiftLintCore/Extensions/SwiftSyntax+SwiftLint.swift
SimplyDanny: Yes, typically below imports.

Tests/BuiltInRulesTests/NestingRuleTests.swift
SimplyDanny: In the documentation that includes these examples later, we now also show the configuration in case it deviates from the default one. So having them as normal examples is even nicer as people also see how options influence the rule's behavior.

Source/SwiftLintBuiltInRules/Rules/Metrics/NestingRule.swift
SimplyDanny: Not awesome, but better than the extra parameter.

Tests/BuiltInRulesTests/NestingRuleTests.swift
braker1nine: Ahhh okay I didn't realize that `Example` has a configuration option. That's why I was confused!

Okay I updated these and I removed the tests. Which I'm not 100% sure about... but I thought the testing covered all the default examples. Is that wrong?

CHANGELOG.md
```diff
@@ -63,6 +63,12 @@
   `max_number_of_single_line_parameters` and detect mixed formatting.  
   [GandaLF2006](https://github.com/GandaLF2006)
 
+* Add `ignore_coding_keys` parameter to `nesting` rule.  
+* Add `ignore_coding_keys` parameter to `nesting` rule. Setting this to true prevents
```
```suggestion
* Add `ignore_coding_keys` parameter to `nesting` rule. Setting this to true prevents
```

Tests/BuiltInRulesTests/NestingRuleTests.swift
SimplyDanny: The feature that configurations are added to the examples in the documentation is quite new and we won't convert existing tests. But we should prefer this way of documenting examples from now on.

SimplyDanny (approved):
Looks good now. Thanks for the updates, @braker1nine!

PR #5658, opened 2024-07-08
Enable `implicit_return` and fix all violations

mildm8nnered:
The only changes here are

1) Enabling (or not disabling) `implicit_return`
2) Fix all existing 300+ violations
3) Whitespace adjustments for parameter alignment where removing `return` threw things out.


SimplyDanny:
> 3. Whitespace adjustments for parameter alignment where removing return threw things out.

Can you give an example for such an adjustment? Is it anything that the rewriter could do better?

SimplyDanny (approved):

PR #5662, opened 2024-07-10
Enabled `prefer_self_in_static_references`, and fixed all current violations

mildm8nnered:
Fixed via `--fix` ...


SimplyDanny (approved):
Any issues with that after applying the auto-fixes?

mildm8nnered:
> Any issues with that after applying the auto-fixes?

I should have made that clear - it was all fine. I'm thinking of trying to apply a bunch of other autocorrects - even ones we're not interested in applying, just to see if they throw up any more problems.

PR #5664, opened 2024-07-10
Enable `multiline_parameters` rule, and fix all cases

mildm8nnered:
There were ~40 violations

In cases like 

```
func validate(file: SwiftLintFile, kind: SwiftDeclarationKind,
                  dictionary: SourceKittenDictionary) -> [StyleViolation] {
```

I mostly went for 

```
func validate(file: SwiftLintFile, 
              kind: SwiftDeclarationKind,
              dictionary: SourceKittenDictionary) -> [StyleViolation] {
```

instead of 

```
func validate(
    file: SwiftLintFile, 
     kind: SwiftDeclarationKind,
    dictionary: SourceKittenDictionary
) -> [StyleViolation] {
```

The latter would be enforced by `multiline_parameters_brackets` - we have about 200 violations of that rule, and about ~100 cases that comply with it (regex search for `func [a-zA-z]+\($`, but a lot of the hits may be examples), so I'm inclined to leave `multiline_parameters_brackets` off unless we really want to commit to the style in that last example.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/CaptureVariableRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/CaptureVariableRule.swift
@@ -158,11 +158,9 @@ struct CaptureVariableRule: AnalyzerRule, CollectingRule {
         file.declaredVariables(compilerArguments: compilerArguments)
     }
 
-    func validate(
-        file: SwiftLintFile,
-        collectedInfo: [SwiftLintFile: Self.FileInfo],
-        compilerArguments: [String]
-    ) -> [StyleViolation] {
+    func validate(file: SwiftLintFile,
+                  collectedInfo: [SwiftLintFile: Self.FileInfo],
+                  compilerArguments: [String]) -> [StyleViolation] {
         file.captureListVariables(compilerArguments: compilerArguments)
             .filter { capturedVariable in collectedInfo.values.contains { $0.contains(capturedVariable.usr) } }
             .map {
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/CaptureVariableRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/CaptureVariableRule.swift
@@ -158,8 +158,11 @@ struct CaptureVariableRule: AnalyzerRule, CollectingRule {
         file.declaredVariables(compilerArguments: compilerArguments)
     }
 
-    func validate(file: SwiftLintFile, collectedInfo: [SwiftLintFile: Self.FileInfo],
-                  compilerArguments: [String]) -> [StyleViolation] {
+    func validate(
+        file: SwiftLintFile,
+        collectedInfo: [SwiftLintFile: Self.FileInfo],
+        compilerArguments: [String]
+    ) -> [StyleViolation] {
         file.captureListVariables(compilerArguments: compilerArguments)
             .filter { capturedVariable in collectedInfo.values.contains { $0.contains(capturedVariable.usr) } }
             .map {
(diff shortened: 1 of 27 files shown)


Source/SwiftLintBuiltInRules/Rules/Lint/CaptureVariableRule.swift
```diff
@@ -158,8 +158,11 @@ struct CaptureVariableRule: AnalyzerRule, CollectingRule {
         file.declaredVariables(compilerArguments: compilerArguments)
     }
 
-    func validate(file: SwiftLintFile, collectedInfo: [SwiftLintFile: Self.FileInfo],
-                  compilerArguments: [String]) -> [StyleViolation] {
+    func validate(
+        file: SwiftLintFile,
+        collectedInfo: [SwiftLintFile: Self.FileInfo],
+        compilerArguments: [String]
+    ) -> [StyleViolation] {
```
SimplyDanny, line 165: Position of the return type is inconsistent here. Should it go to the end of the line above?

Source/SwiftLintBuiltInRules/Rules/Lint/UnusedDeclarationRule.swift
```diff
@@ -106,17 +107,20 @@ private extension SwiftLintFile {
...
         })
     }
 
-    func declaredUSR(indexEntity: SourceKittenDictionary, editorOpen: SourceKittenDictionary,
-                     compilerArguments: [String], configuration: UnusedDeclarationConfiguration)
+    func declaredUSR(indexEntity: SourceKittenDictionary,
+                     editorOpen: SourceKittenDictionary,
+                     compilerArguments: [String],
+                     configuration: UnusedDeclarationConfiguration)
     -> UnusedDeclarationRule.DeclaredUSR? {
```
SimplyDanny, line 124: Should it go to the end of the previous line?

Source/SwiftLintCore/Protocols/CollectingRule.swift
```diff
@@ -114,8 +118,11 @@ package protocol CollectingCorrectableRule: CollectingRule, CorrectableRule {
 }
 
 package extension CollectingCorrectableRule {
-    func correct(file: SwiftLintFile, collectedInfo: [SwiftLintFile: FileInfo],
-                 compilerArguments _: [String]) -> [Correction] {
+    func correct(
+        file: SwiftLintFile,
+        collectedInfo: [SwiftLintFile: FileInfo],
+        compilerArguments _: [String]
+    ) -> [Correction] {
```
SimplyDanny, line 125: Same question.

SimplyDanny (approved):

mildm8nnered:
So I put all those ones back to 

```
func foo(bar1: Int,
        bar2: Int) -> Int {
    1
}
```

Beyond the scope of this PR, across the codebase in general it is a bit of a mix. We have quite a lot of the above style, some (100 plus say) in the following `multiline_parameters_brackets` compliant style:

```
func foo(
    bar1: Int,
    bar2: Int
) -> Int {
    1
}
```

and then a few 

```
func foo(
    bar1: Int,
    bar2: Int) -> Int {
    1
}
```

I don't think there's any pressing need to regularize these, or strong position on which, if any, is better, but we could, for example, follow `multiline_parameters_brackets`.

SimplyDanny:
For my taste,
```swift
func foo(bar1: Int,
        bar2: Int) -> Int {
    1
}
```
is best and I wouldn't like to have the `multiline_parameters_brackets` style generally. But I don't think we have a rule for the other style, have we? Anyway, I agree that this is not pressing.

Issue #5659, opened 2024-07-08
Feature request - `--only-rule` argument for lint and analyze, to enable just one rule (possibly allowed multiple times)


mildm8nnered:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the feature

Sometime it would be handy to enable just one rule.

Consider trying to autocorrect just the `implicit_return` rule. It would be nice to be able to say

`swiftlint --fix --only-rule implicit_return`

The alternative would be to temporarily hack the configuration file.

Possibly could be added multiple times to enable a few rules, or as a comma separated list.

This is kind of in the spirit of `--enable-all-rules` - the configuration file rule enablement would be ignored, and SwiftLint would behave as though the rules were specified in `only_rules` in the config file.

mildm8nnered:
> The idea is good, but maybe there is a more general solution than the extra argument.
> 
> In case we go with the argument, would it also be allowed on the `lint` command? Why not, right?

Oh absolutely - `lint` and `analyze` - `--fix` is just the most obvious case, and the one I came across and thought - "editing .swiftlint.yml for this is really annoying".

I think the biggest question in my mind is should we allow multiple invocations of `--only-rule`, or make it a comma separated list, or just allow one invocation (probably easiest, and would satisfy a lot of use cases - you can always run it again specifying another rule, and if you want to run it regularly against a number of rules, maybe you should have a dedicated config file for that).

mildm8nnered:
But if you have another angle I'd love to hear it

SimplyDanny:
> But if you have another angle I'd love to hear it

No other angle. Reflecting on it, I always hacked the configuration. That's awkward. I'd like to have the argument as well. 😉 

> I think the biggest question in my mind is should we allow multiple invocations of `--only-rule`, or make it a comma separated list, or just allow one invocation (probably easiest, and would satisfy a lot of use cases - you can always run it again specifying another rule, and if you want to run it regularly against a number of rules, maybe you should have a dedicated config file for that).

Agree, one should be enough.

PR #5666, opened 2024-07-11
Add `--only-rule` command line option

mildm8nnered:
Addresses #5659

Adds an `--only-rule` option to the `lint` and `analyze` subcommands (`--fix` is also supported), which behaves as though that single rule was specified in `only_rules:` in the configuration file.

This is very similar in spirit to the `--enable-all-rules` command line option. Configuration file settings will be ignored.

A primary use case for this feature is to be able to autocorrect violations of a single rule at a time.

Previously the alternative was to hack the existing `.swiftlint.yml` file, or to provide a custom one.

```
 % swiftlint.debug help lint   
OVERVIEW: Print lint warnings and errors

USAGE: swiftlint lint [<options>] [<paths> ...]

ARGUMENTS:
  <paths>                 List of paths to the files or directories to lint.

OPTIONS:
  --only-rule <only-rule> Run only the specified rule, ignoring `only_rules`, `opt_in_rules` and `disabled_rules`.
```

```
% swiftlint help analyze
OVERVIEW: Run analysis rules

USAGE: swiftlint analyze [<options>] [<paths> ...]

ARGUMENTS:
  <paths>                 List of paths to the files or directories to analyze.

OPTIONS:
  --only-rule <only-rule> Run only the specified rule, ignoring `only_rules`, `opt_in_rules` and `disabled_rules`.
```

### DIFF
--- a/Source/swiftlint/Commands/Lint.swift
+++ b/Source/swiftlint/Commands/Lint.swift
@@ -19,6 +19,8 @@ extension SwiftLint {
         var noCache = false
         @Flag(help: "Run all rules, even opt-in and disabled ones, ignoring `only_rules`.")
         var enableAllRules = false
+        @Option(help: "Run only the specified rule, ignoring the configuration file.")
+        var onlyRule: String?
         @Argument(help: pathsArgumentDescription(for: .lint))
         var paths = [String]()
 
@@ -52,6 +54,7 @@ extension SwiftLint {
                 cachePath: cachePath,
                 ignoreCache: noCache,
                 enableAllRules: enableAllRules,
+                onlyRule: onlyRule,
                 autocorrect: common.fix,
                 format: common.format,
                 compilerLogPath: nil,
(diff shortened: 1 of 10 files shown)


Source/swiftlint/Commands/Lint.swift
```diff
@@ -19,6 +19,8 @@ extension SwiftLint {
         var noCache = false
         @Flag(help: "Run all rules, even opt-in and disabled ones, ignoring `only_rules`.")
         var enableAllRules = false
+        @Option(help: "Run only the specified rule, ignoring the configuration file.")
```
SimplyDanny, line 22: It's not that the configuration file is ignored completely, is it? Only the rules specified in there are overridden, yet configurations for the "only rule" still apply. Perhaps this should be made clearer here.

SimplyDanny (approved):
I really look forward to have this. Every so often I mangle the configuration file instead. 🫣

Source/swiftlint/Commands/Lint.swift
mildm8nnered: So I went with: "Run only the specified rule, ignoring `only_rules`, `optin_rules` and `disabled_rules`.", by analogy with `--enable-all-rules` "Run all rules, even opt-in and disabled ones, ignoring `only_rules`."

SimplyDanny: Sounds good. It's ~~`optin_rules`~~`opt_in_rules` though.

mildm8nnered: 🤦

Issue #2120, opened 2018-03-29
Rule Request: Unused variables in functions and classes


varunpm1:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Rule Request

1. Why should this rule be added?

I would prefer one rule to be added that can check for unused parameters passed into the function or unsued properties present inside a class or it's extensions. I understand that class variable tracking would be pretty difficult if they're public since they can be modified/accessed from other classes as well. I'm not sure if this is a feasible feature but would be great if it's present.

#### Triggering examples -

```
func testSomething(a: Int, b: Int, c: Int) -> Int { // c parameter isn't used at all
   let sum = a + b
   
    return sum
}
```

```
class Sample {
    var var1: Int = 0
    @IBOutlet weak var sampleCollectionView: UICollectionView!
    // since var1 and sampleCollectionView isn't being read or written into in entire class
}
```

It can be a enabled by default since there is no use of unused variables that are passed inside a function or in a class.

It could prevent some unused variables being left out during code refactoring/cleanup, that people might miss out :)

marcelofabri:
I think those should be two different rules.

+1 for ignoring function parameters in overrides. In protocol conformances it'll be harder given how SwiftLint works.

Checking unused properties would also be hard, but we could do it at least for `private` and `fileprivate` properties by now.

acecilia:
@marcelofabri some compilers allow to enable those warnings: for example in GCC by passing the flag `-Wunused` (see: https://stackoverflow.com/questions/386220/how-can-i-hide-defined-but-not-used-warnings-in-gcc). Do you think it should be a compiler feature or a Swiftlint feature?

mathaeus:
I'd also vote for a "unused method parameter rule". We accidentally introduced a bug because a method was refactored and afterwards a parameter was unused.

Moximillian:
Xcode supports checking for unused function parameters when using Objective-C, but not for Swift. This would be very nice addition to SwiftLint.

keith:
If you're interested in having a native Swift warning for unused parameters you should add your input about this here https://forums.swift.org/t/swift-and-gcc-warn-unused-parameter/14607/8

kirhgoff:
[...]

PR #5668, opened 2024-07-12
Add new `unused_parameter` rule

SimplyDanny:
Partially implements #2120.

Even though [Periphery](https://github.com/peripheryapp/periphery) does the same (and much more the like) globally, e.g. by checking all protocol implementors, due to the restricted local scope of a function/initializer/subscript, a basic version of this check operating on syntax level can be implemented in SwiftLint as well.

The only issue concerns auto-fixes and documentation. Xcode puts the second parameter name into the comment. If auto-fixing changes the parameter name to a wildcard, documentation would have to be adapted likewise. While it's possible to fix this, it feels just out-of-scope for the first rule proposal and perhaps this would go too far generally.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -48,6 +48,7 @@ disabled_rules:
   - trailing_closure
   - type_contents_order
   - unused_capture_list
+  - unused_parameter
   - vertical_whitespace_between_cases
 
 # Configurations
@@ -62,7 +63,7 @@ balanced_xctest_lifecycle: &unit_test_configuration
     - XCTestCase
 closure_body_length:
     warning: 50
-    error: 100    
+    error: 100
 empty_xctest_method: *unit_test_configuration
 file_name:
   excluded:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,11 @@
   [Ueeek](https://github.com/Ueeek)
   [#5615](https://github.com/realm/SwiftLint/issues/5615)
 
+* Add new `unused_parameter` rule that triggers on function/initializer/subscript
+  parameters that are not used inside of the function/initializer/subscript.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+  [#2120](https://github.com/realm/SwiftLint/issues/2120)
+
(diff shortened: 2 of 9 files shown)


SimplyDanny:
Fixed violations for `redundant_self_in_closures` rule look are valid and stem from the fact that `DeclaredIdentifiersTrackingVisitor` now considers parameters of initializers and subscripts as well.

The findings of the new rule are massive, but the ones listed in here all seem to be okay. Running the rule on SwiftLint itself also works nicely. There are no issues or compile errors afterwards. However, I'd like to check it on a bigger reference before I merge it. @mildm8nnered: Could you run the new rule on your code base and check if all auto-fixes are sane?

mildm8nnered:
> Fixed violations for `redundant_self_in_closures` rule look are valid and stem from the fact that `DeclaredIdentifiersTrackingVisitor` now considers parameters of initializers and subscripts as well.
> 
> The findings of the new rule are massive, but the ones listed in here all seem to be okay. Running the rule on SwiftLint itself also works nicely. There are no issues or compile errors afterwards. However, I'd like to check it on a bigger reference before I merge it. @mildm8nnered: Could you run the new rule on your code base and check if all auto-fixes are sane?

So on 170,000 loc,  1,915 violations I only found one thing:

Does not seem to deal with cases where variable names need to be backticked because they collide with reserved words:

```
    struct SomeStruct {
        let `operator`: SomeOperator

        public init(`operator`: SomeOperator) {
            self.operator = SomeOperator(rawValue: `operator`.rawValue)
        }
```

Will get corrected to:

```
    struct SomeStruct {
        let `operator`: SomeOperator

        public init(`operator` _: SomeOperator) {
            self.operator = SomeOperator(rawValue: `operator`.rawValue)
        }
```

I had several of these.

Everything else seemed fine and one I fixed these I got a clean compile.

mildm8nnered (approved):

SimplyDanny:
That's good news. Thanks a lot for checking!

I had the backticks in mind while implementing the rule and it partially considers such escaped variable names already. Should be fine to just disregard the backticks as only the name is important.

Issue #5667, opened 2024-07-12
Extraneous whitespace between attribute name and '(';


ConfusedVorlon:
### New Issue Checklist

- [ Y] Updated SwiftLint to the latest version
- [ Y] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

the following code

` private (set) var type: ClipType = .undetermined`

generates a warning in Xcode 16 beta 3

> warning: extraneous whitespace between attribute name and '('; this is an error in the Swift 6 language mode

1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.
I assume swiftlint could detect and fix this before Xcode needs to worry about it

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

triggers:
` private (set) var type: ClipType = .undetermined`

doesn't trigger
` private(set) var type: ClipType = .undetermined`

3. Should the rule be configurable, if so what parameters should be configurable?
no

4. Should the rule be opt-in or enabled by default? Why?
 default - because if swiftlint doesn't handle it, Xcode will

SimplyDanny:
This should be a pretty simple rule to implement. However, it's probably also not a big deal to fix this only as soon as the Swift 6 compiler complains.

PR #5669, opened 2024-07-13
Add new `attribute_name_spacing` rule

aryamansharda:
Resolves https://github.com/realm/SwiftLint/issues/5667

Seeing as `private (set)` would cause issues with the Swift 6 compiler in Xcode 16+, this PR implements the changes discussed in the issue linked above. 

The rule is now a default rule and prevents the following code from triggering errors:

Triggers:
`private (set) var type: ClipType = .undetermined`

Doesn't Trigger:
`private(set) var type: ClipType = .undetermined`

First time contributing to SwiftLint, so apologies for any contribution mistakes on my end. You managed to convince me to try and contribute @SimplyDanny (re. SwiftCraft)


SimplyDanny:
Hey @aryamansharda, nice to hear from you and even better via a contribution!

The implementation looks good to me. Yet, checking the commits that introduced this warning/error in the [Swift compiler](https://github.com/swiftlang/swift/commit/c6e425a55938dba840416e2ef1ae1491fc048082) and SwiftSyntax, I just found out that `<access-control-modifier> (set)` is only one special case. The warning/error is actually more general and applies to all modifiers and attributes likewise, e.g. `@available (*, deprecated)` or `nonisolated (unsafe)`.

For the attribute case, we could also check that there is no trivia between the `@` sign and the attribute name as this is another related compiler diagnostic. It also seems that while
```swift
let closure1 = { @MainActor (a, b) in
}
```
leads to the warning/error,
```swift
let closure2 = { @MainActor 
   (a, b) in
}
```
does not. In other words, a line break serves as a disambiguator.

I hope this isn't too much to ask for, but it seems that only supporting all these cases makes this rule a real help in preparation for Swift 6. I'm happy to help in case of any issues. 😉

aryamansharda:
Sounds good! I'll explore that today - just wanted to double-check my understanding before I continue. 

First, I should should update my existing implementation to check that there is no trivia between any `DeclModifier` and `DeclModifierDetail` which would handle cases like `(set)` and `nonisolated(unsafe)`

Next, based off the examples in the commit you referenced:
```
struct PropertyWrapperTest {
  @MyPropertyWrapper (param: 2)  // expected-warning {{extraneous whitespace between attribute name and '('; this is an error in Swift 6}}
  var x: Int
}

let closure1 = { @MainActor (a, b) in // expected-warning {{extraneous whitespace between attribute name and '('; this is an error in Swift 6}}
}

let closure2 = { @MainActor
  (a: Int, b: Int) in
}
```
When I plug them into the Swift AST Explorer, I see that `AttributeList` and `Attribute` are the type we're after. 

So, in addition to the existing check above, we want to add a new check for all `AttributeList->Attribute->IdentifierType` to ensure that there is no leading (ex. `@ MainActor` / `@ escaping`) or trailing trivia (ex. `@MyPropertyWrapper (param: 2)` / ` @MainActor (a, b)`)?

SimplyDanny:
Yes, you mentioned all the cases we can find in the related Swift compiler diagnostics.

mildm8nnered:
Not sure if this will work or not, but you might be able to avoid the Rewriter implementation by doing something like [ImplicitReturnRule](https://github.com/realm/SwiftLint/blob/cc4b569c5491b33963f3e0f435e7722f9311b77e/Source/SwiftLintBuiltInRules/Rules/Style/ImplicitReturnRule.swift#L1) does with `SwiftSyntaxCorrectableRule`.

aryamansharda:
@mildm8nnered Ah, very cool. Just updated the implementation to use that 👍

Source/SwiftLintBuiltInRules/Rules/Style/AttributeNameSpacingRule.swift
```diff
@@ -0,0 +1,142 @@
...
+                init(param: Int) {}
+            }
+            """),
+            Example("""
+            let closure2 = { @MainActor
+              (a: Int, b: Int) in
+            }
+            """),
+        ],
+        triggeringExamples: [
```
SimplyDanny, line 36: Let's check if `@MyProperty ()` triggers as well. 🤓

SimplyDanny (changes requested):
Looks good to me! All the main cases are covered. I have only a few small remarks.

SimplyDanny:
Hey @aryamansharda, anything that's blocking this PR? I'd like to include it into the next major release. Swift 6 is not too far off. To drive this rule useable, we shouldn't wait for too long.

aryamansharda:
Hey, sorry for the delay! I'll make the updates in the next day or two!

aryamansharda:
@SimplyDanny Sorry for the delay, but I've made the requested changes. If there's any other changes or documentation changes let me know - I promise the following turn around time will be much faster.

CHANGELOG.md
```diff
@@ -125,6 +125,11 @@
 
 #### Enhancements
 
+* Adds `access_control_setter_spacing` rule to enforce no trailing whitespace between attribute names and parentheses, ensuring compatibility with Swift 6, where this spacing causes compilation errors.
+  [aryamansharda](https://github.com/aryamansharda)
+  [#5667](https://github.com/realm/SwiftLint/issues/5667)
+* Linting got around 20% faster due to the praisworthy performance
+  improvements done in the [SwiftSyntax](https://github.com/apple/swift-syntax)
```
SimplyDanny, line 132: Please append two spaces to the end of the description. They are required to enforce a line break in the rendered Markdown.

```suggestion
* Add new `access_control_setter_spacing` rule to enforce no trailing whitespace between
  attribute names and parentheses, ensuring compatibility with Swift 6, where this spacing
  causes compilation errors.  
  [aryamansharda](https://github.com/aryamansharda)
  [#5667](https://github.com/realm/SwiftLint/issues/5667)

```

Source/SwiftLintBuiltInRules/Rules/Style/AttributeNameSpacingRule.swift
```diff
@@ -0,0 +1,144 @@
...
+import SwiftSyntax
+
+@SwiftSyntaxRule
+struct AttributeNameSpacingRule: SwiftSyntaxCorrectableRule {
+    var configuration = SeverityConfiguration<Self>(.error)
+
+    static let description = RuleDescription(
+        identifier: "attribute_name_spacing",
+        name: "Attribute Name Spacing",
+        description: "This rule prevents trailing spaces after attribute names, ensuring compatibility with Swift 6 where a space between an attribute name and the opening parenthesis results in a compilation error (e.g. `@MyPropertyWrapper ()`, `@escaping ()`).",
```
SimplyDanny, line 11: Use a string block:
```suggestion
        description: """
            This rule prevents trailing spaces after attribute names, ensuring compatibility \
            with Swift 6 where a space between an attribute name and the opening parenthesis \
            results in a compilation error (e.g. `@MyPropertyWrapper ()`, `private (set)`).
            """,
```

Source/SwiftLintBuiltInRules/Rules/Style/AttributeNameSpacingRule.swift
```diff
@@ -0,0 +1,144 @@
...
+            Example("fileprivate ↓(set) var foo: Bool = false"),
+            Example("public ↓(set) var foo: Bool = false"),
+            Example("  public  ↓(set) var foo: Bool = false"),
+            Example("@ ↓MainActor class Foo {}"),
+            Example("func funcWithEscapingClosure(_ x: @ ↓escaping () -> Int) {}"),
+            Example("func funcWithEscapingClosure(_ x: @escaping↓() -> Int) {}"),
+            Example("@available ↓(*, deprecated)"),
+            Example("@MyPropertyWrapper ↓(param: 2) "),
+            Example("nonisolated ↓(unsafe) var _value: X?"),
+            Example("@MyProperty ()"),
```
SimplyDanny, line 48:
```suggestion
            Example("@MyProperty ↓() class Foo {}"),
```

Source/SwiftLintBuiltInRules/Rules/Style/AttributeNameSpacingRule.swift
```diff
@@ -0,0 +1,144 @@
...
+            endPosition: AbsolutePosition,
+            replacement: String
+        ) {
+            let correction = ReasonedRuleViolation.ViolationCorrection(
+                start: startPosition,
+                end: endPosition,
+                replacement: replacement
+            )
+
+            violations.append(at: endPosition, correction: correction)
```
SimplyDanny, line 135: The description of this rule is very detailed which is good. It's impractical as a violation message, though. Better to set an explicit reason here that depends on the actual violation and is much shorter. That would look like:
```swift
violations.append(
    .init(
        position: endPosition, 
        reason: reason, 
        correction: .init(
            start: startPosition,
            end: endPosition,
            replacement: replacement
        )
    )
)
```

Incidentally, this is also why the tests fail.

SimplyDanny (changes requested):

Source/SwiftLintBuiltInRules/Rules/Style/AttributeNameSpacingRule.swift
aryamansharda: Updated!

SimplyDanny (approved):

SimplyDanny:
Thanks @aryamansharda! I only made the wording more serious to better match the `error` severity and the fact that these violation will be compiler errors in the future.

Once the build passes, this is ready to be merged.

Issue #4754, opened 2023-02-07
superfluous_disable_command not triggering for custom rule


marcelofabri:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

`superfluous_disable_command` is not triggering for custom rules. I don't know if this has always been the behavior or if it's a recent regression.

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint file.swift           
Linting Swift files at paths file.swift
Linting 'file.swift' (1/1)
/Users/marcelo.fabri/SwiftLint/test/file.swift:1:5: warning: forbidden Violation: This is forbidden (forbidden)
Done linting! Found 1 violation, 0 serious in 1 file.
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.50.3
* Installation method used (Homebrew, CocoaPods, building from source, etc)? brew
* Paste your configuration file:

```yml
custom_rules:
  forbidden:
    regex: 'FORBIDDEN'
    message: "This is forbidden"
    severity: warning
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? No
* Which Xcode version are you using (check `xcodebuild -version`)? 14.0.1
* Do you have a sample that shows the issue? Run `echo "[string here]" | swiftlint lint --no-cache --use-stdin --enable-all-rules`
  to quickly test if your example is really demonstrating the issue. If your example is more
  complex, you can use `swiftlint lint --path [file here] --no-cache --enable-all-rules`.

```swift
let FORBIDDEN = 1
// swiftlint:disable:next forbidden
let ALLOWED = 2
```

marcelofabri:
reopening because the PR that fixed it got reverted

mildm8nnered:
So I picked this up again the other day, and I think I can see where we were going wrong before ... should have a PR soon

mildm8nnered:
Resolved by #5670

PR #5670, opened 2024-07-13
Fix superfluous disable command for `custom_rules`

mildm8nnered:
So this is a kind of scary PR - addresses #4754 - The `superfluous_disable_command` rule should work for individual `custom_rules`. 

#4755 attempted to address this, but had issues, and was reverted in #5336

This PR is based on @marcelofabri 's work in #4755 with a few additional fixes.

A fair few times I thought I had this cracked, only to find a big hole. I think it's right now, but this PR definitely deserves additional scrutiny.

Consider the example below:

```
custom_rules:
  forbidden:
    regex: 'FORBIDDEN'
    message: "This is forbidden"
    severity: warning
```

The following code should trigger a `superfluous_disable_command` violation:

```
let FORBIDDEN = 1
// swiftlint:disable:next forbidden
let ALLOWED = 2
```

Today this would produce:

```
% swiftlint --config t.config t.swift --quiet
/Some/Path/To/t.swift:1:5: warning: forbidden Violation: This is forbidden (forbidden)
```

and after this PR is applied, we get:

```
% swiftlint.debug --config t.config t.swift --quiet
/Some/Path/To/t.swift:1:5: warning: forbidden Violation: This is forbidden (forbidden)
/Some/Path/To/t.swift:3:1: warning: Superfluous Disable Command Violation: SwiftLint rule 'forbidden' did not trigger a violation in the disabled region; remove the disable command (superfluous_disable_command)
```

It is currently also possible to disable all custom rules with:

```
// swiftlint:disable:next custom_rules
let FORBIDDEN = 1
```

That behaviour is still supported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,7 @@
   custom rules.  
   [Marcelo Fabri](https://github.com/marcelofabri)
   [Martin Redington](https://github.com/mildm8nnered)
+  [SimplyDanny](https://github.com/SimplyDanny)
   [#4754](https://github.com/realm/SwiftLint/issues/4754)
 
 * Align left closure brace with associated parent function call in `contrasted_opening_brace` rule.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,11 @@
 * Check `if` expressions nested arbitrarily deep in `contrasted_opening_brace` rule.  
   [SimplyDanny](https://github.com/SimplyDanny)
   [#5752](https://github.com/realm/SwiftLint/issues/5752)
+* `superfluous_disable_command` violations are now triggered for
+  custom rules.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#4754](https://github.com/realm/SwiftLint/issues/4754)
 
 * Align left closure brace with associated parent function call in `contrasted_opening_brace` rule.  
   [SimplyDanny](https://github.com/SimplyDanny)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@
 * Check `if` expressions nested arbitrarily deep in `contrasted_opening_brace` rule.  
   [SimplyDanny](https://github.com/SimplyDanny)
   [#5752](https://github.com/realm/SwiftLint/issues/5752)
+
 * `superfluous_disable_command` violations are now triggered for
   custom rules.  
   [Marcelo Fabri](https://github.com/marcelofabri)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,12 @@
 
 #### Bug Fixes
 
-* None.
+* `superfluous_disable_command` violations are now triggered for
+  custom rules.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [Martin Redington](https://github.com/mildm8nnered)
+  [SimplyDanny](https://github.com/SimplyDanny)
+  [#4754](https://github.com/realm/SwiftLint/issues/4754)
 
 ## 0.56.2: Heat Pump Dryer
 
@@ -63,13 +68,6 @@
 * Check `if` expressions nested arbitrarily deep in `contrasted_opening_brace` rule.  
   [SimplyDanny](https://github.com/SimplyDanny)
   [#5752](https://github.com/realm/SwiftLint/issues/5752)
-
-* `superfluous_disable_command` violations are now triggered for
-  custom rules.  
-  [Marcelo Fabri](https://github.com/marcelofabri)
-  [Martin Redington](https://github.com/mildm8nnered)
-  [SimplyDanny](https://github.com/SimplyDanny)
-  [#4754](https://github.com/realm/SwiftLint/issues/4754)
 
 * Align left closure brace with associated parent function call in `contrasted_opening_brace` rule.  
   [SimplyDanny](https://github.com/SimplyDanny)
--- a/Source/SwiftLintCore/Models/Linter.swift
+++ b/Source/SwiftLintCore/Models/Linter.swift
@@ -1,7 +1,6 @@
 import Foundation
 import SourceKittenFramework
 
-// swiftlint:disable file_length
 private let warnSourceKitFailedOnceImpl: Void = {
     Issue.genericWarning("SourceKit-based rules will be skipped because sourcekitd has failed.").print()
 }()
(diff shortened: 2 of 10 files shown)


Source/SwiftLintCore/Models/Region.swift
```diff
@@ -46,7 +46,10 @@ public struct Region: Equatable {
     ///
     /// - returns: True if the specified rule is disabled in this region.
     public func isRuleDisabled(_ rule: some Rule) -> Bool {
-        areRulesDisabled(ruleIDs: type(of: rule).description.allIdentifiers)
+        if let rule = rule as? CustomRules {
+            return areRulesDisabled(ruleIDs: rule.customRuleIdentifiers + [type(of: rule).description.identifier])
```
SimplyDanny, line 50: Does this mean a custom rule `custom_1` will be disabled if `custom_2` is actually named in the disable command? `rule.customRuleIdentifiers` are all custom identifiers defined, aren't they?

Source/SwiftLintCore/Rules/CustomRules.swift
```diff
@@ -79,19 +83,27 @@ struct CustomRules: Rule, CacheDescriptionProvider {
...
-    func isRuleDisabled(customRuleIdentifier: String) -> Bool {
+extension Region {
+    func isCustomRuleDisabled(customRuleIdentifier: String) -> Bool {
+        disabledRuleIdentifiers.intersection(
+            [
+                .all,
+                RuleIdentifier(customRuleIdentifier),
+                RuleIdentifier(CustomRules.description.identifier),
+            ]
+        ).isNotEmpty
```
SimplyDanny, line 99:
```suggestion
           areAllCustomRulesDisabled 
        || isCustomRuleSpecificallyDisabled(customRuleIdentifier: customRuleIdentifier)
        || disabledRuleIdentifiers.contains(.all)
```

Source/SwiftLintCore/Rules/CustomRules.swift
```diff
@@ -79,19 +83,27 @@ struct CustomRules: Rule, CacheDescriptionProvider {
                                severity: configuration.severity,
                                location: Location(file: file, characterOffset: $0.location),
                                reason: configuration.message)
-            }).filter { violation in
-                guard let region = file.regions().first(where: { $0.contains(violation.location) }) else {
```
SimplyDanny, line 83: Where is this checked now?

Source/SwiftLintCore/Models/Linter.swift
```diff
@@ -16,47 +17,87 @@ private struct LintResult {
...
-        }
         let regionsDisablingSuperfluousDisableRule = regions.filter { region in
             region.isRuleDisabled(superfluousDisableCommandRule)
         }
 
-        return regionsDisablingCurrentRule.compactMap { region -> StyleViolation? in
-            let isSuperfluousRuleDisabled = regionsDisablingSuperfluousDisableRule.contains {
-                $0.contains(region.start)
+        let identifiersWithRegions: [(String, [Region])] = {
+            if let customRules = self as? CustomRules {
```
SimplyDanny, line 32: `CustomRules` is a `Rule` as well. Can we just override this method for `CustomRules` instead of branching here?

Source/SwiftLintCore/Models/Region.swift
mildm8nnered: So this specific function will return `true` in that case, yes, but the actual identifiers will be checked afterwards, here in the construction of `identifiersWithRegions`

```
    func superfluousDisableCommandViolations(regions: [Region],
                                             superfluousDisableCommandRule: SuperfluousDisableCommandRule?,
                                             allViolations: [StyleViolation]) -> [StyleViolation] {
        guard regions.isNotEmpty, let superfluousDisableCommandRule else {
            return []
        }

        let regionsDisablingSuperfluousDisableRule = regions.filter { region in
            region.isRuleDisabled(superfluousDisableCommandRule)
        }

        let identifiersWithRegions: [(String, [Region])] = {
```

The proof of the pudding 🤞 is (slightly expanded for clarity):

```
    func testUnviolatedSpecificCustomRulesTriggersSuperfluousDisableCommand() throws {
        let customRules =         [
            "forbidden": [
                "regex": "FORBIDDEN",
            ],
            "forbidden2": [
                "regex": "FORBIDDEN2",
            ],
        ]

        let example = Example("""
                              // swiftlint:disable:next forbidden forbidden2
                              let FORBIDDEN = 1
                              """)

        let violations = try self.violations(forExample: example, customRules: customRules)
        XCTAssertEqual(violations.count, 1)
        XCTAssertTrue(violations[0].isSuperfluousDisableCommandViolation(for: "forbidden2"))
    }
```

Source/SwiftLintCore/Rules/CustomRules.swift
mildm8nnered: Everything now happens in `superfluousDisableCommandViolations` - I think this was the root of the problem with the previous PR.

Source/SwiftLintCore/Models/Linter.swift
mildm8nnered: yes, that would probably be better - I'll look at re-arranging this

Source/SwiftLintCore/Models/Region.swift
mildm8nnered: I'm wondering if a naming change might make this clearer, but I don't have a good candidate ...

Source/SwiftLintCore/Models/Linter.swift
mildm8nnered: So I've abtracted this logic into a `func disabledRuleIdentifiersWithRegions(regions: [Region]) -> [(String, [Region])]` method, which is implemented in an extension of `Rule`, and then also implemented on `CustomRules`.

SimplyDanny:
I still have my problems with the special handling of custom rules and some improvement ideas in my backlog.

Do you mind if I push some changes to your branch later?

mildm8nnered:
> I still have my concerns with the special handling of custom rules and some improvement ideas in my backlog.
> 
> Do you mind if I push some changes to your branch later?

Yes, I think there are still some oddities outside of superfluous_disable

Feel free to add more changes!

SimplyDanny:
I managed to avoid explicit type checks at a few places. There are more in other (untouched) code, though, that we should get rid of in the long run.

Please check if my changes make sense to you, @mildm8nnered.

Source/SwiftLintCore/Models/Linter.swift
```diff
@@ -93,16 +102,26 @@ private extension Rule {
...
-            region?.isRuleEnabled(self) ?? true
+        }.partitioned { violation, region in
+            if let region {
+                return isEnabled(in: region, for: violation.ruleIdentifier)
+            }
+            return true
         }
 
+        let customRulesIDs: [String] = {
+             guard let customRules = self as? CustomRules else {
```
SimplyDanny, line 113: Perhaps we could get rid of this by letting `Self.description.allIdentifiers` contain the custom IDs in case of custom rules.

SimplyDanny (approved):

mildm8nnered:
> I managed to avoid explicit type checks at a few places. There are more in other (untouched) code, though, that we should get rid of in the long run.
> 
> Please check if my changes make sense to you, @mildm8nnered.

So it may take me a while to digest those, but straight away it looks a lot shorter and cleaner.

One change request - I think you should add yourself to the authors in the changelog.

CHANGELOG.md
```diff
@@ -64,6 +64,12 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#5752](https://github.com/realm/SwiftLint/issues/5752)
 
+* `superfluous_disable_command` violations are now triggered for
+  custom rules.  
+  [Marcelo Fabri](https://github.com/marcelofabri)
+  [Martin Redington](https://github.com/mildm8nnered)
```
SimplyDanny, line 70:
```suggestion
  [Martin Redington](https://github.com/mildm8nnered)
  [Danny Mösch](https://github.com/SimplyDanny)
```

SimplyDanny:
> > I managed to avoid explicit type checks at a few places. There are more in other (untouched) code, though, that we should get rid of in the long run.
> > Please check if my changes make sense to you, @mildm8nnered.
> 
> So it may take me a while to digest those, but straight away it looks a lot shorter and cleaner.

Thanks! The tests you added gave me confidence and helped a lot in finding gaps.

> One change request - I think you should add yourself to the authors in the changelog.

Done.

Source/SwiftLintCore/Models/Linter.swift
mildm8nnered: I couldn't see an easy way to customize `allIdentifiers`, as the RuleDescription currently can't see the configuration :-(

SimplyDanny: Yeah, also what I struggled with. We'd need to move `allIdentifiers` into `Rule`, but this has other disadvantages. So let's not bother with that in this PR. I think it's good as it is. The special handling of custom rules is a broader topic we may address in follow-ups.

mildm8nnered:
> > I managed to avoid explicit type checks at a few places. There are more in other (untouched) code, though, that we should get rid of in the long run.
> > Please check if my changes make sense to you, @mildm8nnered.
> 
> So it may take me a while to digest those, but straight away it looks a lot shorter and cleaner.
> 
> One change request - I think you should add yourself to the authors in the changelog.

I think your changes look great @SimplyDanny. Am I ok to hit merge on this? Hopefully we'll do better than #4755 🤞

SimplyDanny:
> Am I ok to hit merge on this?

Oh, I misread this as a statement. Just merged it. 🫣

> Hopefully we'll do better than #4755 🤞 

People will let us know ... 😅

PR #5674, opened 2024-07-15
fix: void_function_in_ternary incorrectly triggered with if statement

u-abyss:


Source/SwiftLintBuiltInRules/Rules/Idiomatic/VoidFunctionInTernaryConditionRule.swift
```diff
@@ -60,6 +60,42 @@ struct VoidFunctionInTernaryConditionRule: Rule {
...
+            }
+            """),
+            Example("""
+            func exampleNestedIfExpr() -> String {
+                test()
+                if true {
+                  isTrue ? defaultValue() : defaultValue()
+                } else {
+                    "Default"
+                }
```
SimplyDanny, line 88: This is actually a triggering example, isn't?

SimplyDanny (changes requested):
Thanks taking up the issue!

The current implementation is very specific to implicit returns. However, `if` and `switch` expression can also appear as variable initializers and they can be nested into each other. I don't think the current solution addresses these cases, does it?

u-abyss:
@SimplyDanny 
Added handling of nested if and switch statements. I have a question.
I believe that when you initialize a variable in an if or switch expression, the compiler prevents you from writing a statement that returns void in it.
If there is an appropriate case for initializing a variable in an if or switch expression, I would appreciate it if you could let me know.

SimplyDanny:
> @SimplyDanny Added handling of nested if and switch statements. I have a question. I believe that when you initialize a variable in an if or switch expression, the compiler prevents you from writing a statement that returns void in it. If there is an appropriate case for initializing a variable in an if or switch expression, I would appreciate it if you could let me know.

Can you provide an example?

The code 

```swift
func f(i: Int) -> Int {
  if i > 1 {
    let j = i - 1
    j
  } else {
    0
  }
}
```

wouldn't compile. There you need the `return` before `j`.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/VoidFunctionInTernaryConditionRule.swift
```diff
@@ -199,6 +307,37 @@ private extension CodeBlockItemSyntax {
 
         return parent.children(viewMode: .sourceAccurate).count == 1
     }
+
+    // For codeBlockItem, recursively traverse it to determine if it is within its own FunctionDeclSyntax.
+    func getFunctionDeclSyntax(codeBlockItem: CodeBlockItemSyntax) -> FunctionDeclSyntax? {
+      let targetSyntax = codeBlockItem.parent?.parent?.parent?.parent
+      if let targetSyntax = targetSyntax?.as(FunctionDeclSyntax.self) {
```
SimplyDanny, line 314: This covers only functions, but there are also the other cases which are checked above in the `is...Return` properties.

SimplyDanny (changes requested):

Naveen-C-Ramachandrappa (approved):

Source/SwiftLintBuiltInRules/Rules/Idiomatic/VoidFunctionInTernaryConditionRule.swift
```diff
@@ -60,6 +60,56 @@ struct VoidFunctionInTernaryConditionRule: Rule {
...
+                        } else {
+                            return "5"
+                        }
+                    default:
+                        return "6"
+                    }
+                } else {
+                    return "7"
+                }
+            }
```
ankushkushwaha, line 111:
```suggestion
func exampleNestedIfExprAndSwitchExpr(index: Int, isTrue: Bool) -> String {
    guard index <= 3 else {
        return "7"
    }
    
    switch index {
    case 1:
        return isTrue ? "1" : "3"
    case 2:
        return isTrue ? "4" : "5"
    default:
        return "6"
    }
}
```

PR #5677, opened 2024-07-17
SwiftLintPlugin execution from Script Build Phase

frankois944:
When using the plugin from SPM, it's possible to run the switlint binary directly in the Script Build Phase, just like any other command.
The script locates the binary path and executes the command.

Sound like the CocoaPods integration.

I just added the script in the README and using the default value (xcode 15/16), these values can be wrong in previous xcode version or with custom package path.

```bash
SWIFT_PACKAGE_DIR="${BUILD_DIR%Build/*}SourcePackages/artifacts"
SWIFTLINT_CMD=$SWIFT_PACKAGE_DIR/swiftlintplugins/SwiftLintBinary/SwiftLintBinary.artifactbundle/swiftlint-*/bin/swiftlint

if test -f $SWIFTLINT_CMD 2>&1
then
    $SWIFTLINT_CMD
else
    echo "warning: `swiftlint` command not found - See https://github.com/realm/SwiftLint#installation for installation instructions."
fi
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,9 +347,11 @@ else
     echo "warning: `swiftlint` command not found - See https://github.com/realm/SwiftLint#installation for installation instructions."
 fi
 ```
+
 > [!NOTE]
-> The `SWIFTLINT_CMD` path use the default Xcode configuration and has been tested on Xcode 15/16, in case of another configuration like
-> custom Swift package path, please change the values accordinaly.
+> The `SWIFTLINT_CMD` path uses the default Xcode configuration and has been
+> tested on Xcode 15/16. In case of another configuration (e.g. a custom
+> Swift package path), please adapt the values accordingly.
 
 > [!TIP]
 > Uncheck `Based on dependency analysis` to run `swiftlint` on all incremental
--- a/README.md
+++ b/README.md
@@ -348,7 +348,7 @@ else
 fi
 ```
 > [!NOTE]
-> The `SWIFTLINT_CMD` path is using the default xcode condiguration and has been tested on xcode 15/16, in case of another configuration like
+> The `SWIFTLINT_CMD` path use the default xcode condiguration and has been tested on xcode 15/16, in case of another configuration like
 > custom package path, please change the values accordinaly.
 
 > [!TIP]
--- a/README.md
+++ b/README.md
@@ -332,8 +332,9 @@ else
 fi
 ```
 
-If you're using the SwiftLintPlugin from SPM. Use the
-following script implementation:
+If you're using the SwiftLintPlugin in a Swift package,
+you may refer to the `swiftlint` executable in the
+following way:
 
 ```bash
 SWIFT_PACKAGE_DIR="${BUILD_DIR%Build/*}SourcePackages/artifacts"
--- a/README.md
+++ b/README.md
@@ -342,7 +342,7 @@ SWIFTLINT_CMD=$SWIFT_PACKAGE_DIR/swiftlintplugins/SwiftLintBinary/SwiftLintBinar
 
 if test -f $SWIFTLINT_CMD 2>&1
 then
-    $SWIFTLINT_CMD
+    "$SWIFTLINT_CMD"
 else
     echo "warning: `swiftlint` command not found - See https://github.com/realm/SwiftLint#installation for installation instructions."
 fi
--- a/README.md
+++ b/README.md
@@ -347,8 +347,7 @@ else
     echo "warning: `swiftlint` command not found - See https://github.com/realm/SwiftLint#installation for installation instructions."
 fi
 ```
-
-The `SWIFTLINT_CMD` path is using the default xcode condiguration and has been tested on xcode 15/16, in case of another configuration like custom package path, change the values accordinaly.
+> The `SWIFTLINT_CMD` path is using the default xcode condiguration and has been tested on xcode 15/16, in case of another configuration like custom package path, please change the values accordinaly.
 
 > [!TIP]
 > Uncheck `Based on dependency analysis` to run `swiftlint` on all incremental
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -332,6 +332,21 @@ else
 fi
 ```
 
+If you're using the SwiftLintPlugin from SPM. Use the
+following script implementation:
```
SimplyDanny, line 336:
```suggestion
If you're using the SwiftLintPlugin in a Swift package,
you may refer to the `swiftlint` executable in the
following way:
```

README.md
```diff
@@ -332,6 +332,21 @@ else
...
+
+```bash
+SWIFT_PACKAGE_DIR="${BUILD_DIR%Build/*}SourcePackages/artifacts"
+SWIFTLINT_CMD=$SWIFT_PACKAGE_DIR/swiftlintplugins/SwiftLintBinary/SwiftLintBinary.artifactbundle/swiftlint-*/bin/swiftlint
```
SimplyDanny, line 340:
```suggestion
SWIFTLINT_CMD="$SWIFT_PACKAGE_DIR/swiftlintplugins/SwiftLintBinary/SwiftLintBinary.artifactbundle/swiftlint-*/bin/swiftlint"
```

README.md
```diff
@@ -332,6 +332,21 @@ else
...
+
+if test -f $SWIFTLINT_CMD 2>&1
+then
```
SimplyDanny, line 343:
```suggestion
if test -f "$SWIFTLINT_CMD" 2>&1
then
```

That's more consistent to the other Bash scripts in the README.

README.md
```diff
@@ -332,6 +332,21 @@ else
...
+    $SWIFTLINT_CMD
```
SimplyDanny, line 344:
```suggestion
    "$SWIFTLINT_CMD"
```

Quotes around variables are a good practice in case of any spaces in file paths.

SimplyDanny (changes requested):
An interesting and nifty way to get to a versioned binary. 👍 

However, it's probably not sure if this will work with future Xcode versions. Perhaps, we should also add a warning that this approach might fail with changes in Xcode.

README.md
Naveen-C-Ramachandrappa: I second what @SimplyDanny mentioned. Good to wrap it with spaces.

frankois944: Adding quote on script lines will break the script, give it a try and you will see.

frankois944:
> However, it's probably not sure if this will work with future Xcode versions. Perhaps, we should also add a warning that this approach might fail with changes in Xcode.

Yes, I know, it's possible, if Xcode change _again_ the project package path, but for now, it's working on the version 15 and 16.

We can replace this static path by a command to automatically find the binary, but i'm not a fan to execute it every time...

README.md
SimplyDanny: Yeah, then the second line needs to make sure the wildcard is properly resolved. I suggest
```bash
SWIFTLINT_CMD=$(ls "$SWIFT_PACKAGE_DIR"/swiftlintplugins/SwiftLintBinary/SwiftLintBinary.artifactbundle/swiftlint-*/bin/swiftlint | head -n 1)
```
to make sure potential spaces are quoted and there is only a single result in the command.

SimplyDanny:
> > However, it's probably not sure if this will work with future Xcode versions. Perhaps, we should also add a warning that this approach might fail with changes in Xcode.
> 
> Yes, I know, it's possible, if Xcode change _again_ the project package path, but for now, it's working on the version 15 and 16.
> 
> We can replace this static path by a command to automatically find the binary, but i'm not a fan to execute it every time...

Well, then we should mention exactly that: works in Xcode 15 and 16, but no guarantee for other versions or an alternative path.

README.md
frankois944: Using this command
```
SWIFTLINT_CMD=$(ls "$SWIFT_PACKAGE_DIR"/swiftlintplugins/SwiftLintBinary/SwiftLintBinary.artifactbundle/swiftlint-*/bin/swiftlint | head -n 1)
```

frankois944:
> > > However, it's probably not sure if this will work with future Xcode versions. Perhaps, we should also add a warning that this approach might fail with changes in Xcode.
> > 
> > 
> > Yes, I know, it's possible, if Xcode change _again_ the project package path, but for now, it's working on the version 15 and 16.
> > We can replace this static path by a command to automatically find the binary, but i'm not a fan to execute it every time...
> 
> Well, then we should mention exactly that: works in Xcode 15 and 16, but no guarantee for other versions or an alternative path.

Ok, I added this message.

> The SWIFTLINT_CMD path is using the default xcode condiguration and has been tested on xcode 15/16, in case of another configuration like custom package path, please change the values accordinaly.

Understand, I'm using the default Xcode values. The user must adapt his configuration.

README.md
```diff
@@ -332,6 +332,25 @@ else
...
+if test -f "$SWIFTLINT_CMD" 2>&1
+then
+    "$SWIFTLINT_CMD"
+else
+    echo "warning: `swiftlint` command not found - See https://github.com/realm/SwiftLint#installation for installation instructions."
+fi
+```
+> [!NOTE]
+> The `SWIFTLINT_CMD` path use the default Xcode configuration and has been tested on Xcode 15/16, in case of another configuration like
+> custom Swift package path, please change the values accordinaly.
```
SimplyDanny, line 352:
```suggestion

> [!NOTE]
> The `SWIFTLINT_CMD` path uses the default Xcode configuration and has been
> tested on Xcode 15/16. In case of another configuration (e.g. a custom
> Swift package path), please adapt the values accordingly.
```

SimplyDanny (approved):

Issue #4811, opened 2023-03-11
Feature request: Add a rationale to the rule description


mildm8nnered:
### New Issue Checklist

- [X] Updated SwiftLint to the latest version
- [X] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

The rule descriptions are quite sparse, and especially for novices, it may be very unclear what the motivation behind a given rule is, and why it might, or might not be, a good idea to enable or disable it.

The idea would be to add a text block to the rule descriptions that could be displayed in the documentation, explaining the rationale behind the rule, possibly with links etc.

Faux Pas (http://fauxpasapp.com) has a feature like this:

<img width="635" alt="Screenshot 2023-03-11 at 16 58 40" src="https://user-images.githubusercontent.com/435558/224499158-d578fa3f-fe52-406d-87e7-03126e94e7fa.png">

Even better would be if there was some way to jump directly to this from the warnings displayed in Xcode.

PR #5681, opened 2024-07-20
Started to add rationales

mildm8nnered:
Addresses #4811

The idea is to add a longer explanation of the purpose or motivation behind each rule.

You can now see the output via `swiftlint rules rule_name`

For example

```
% swiftlint.debug rules array_init
Array Init (array_init): Prefer using `Array(seq)` over `seq.map { $0 }` to convert a sequence into an Array

Rationale:

When converting the elements of a sequence directly into an `Array`, for clarity, prefer using the `Array` constructor over calling `map`. For example

    Array(foo)

rather than

    foo.↓map({ $0 })

If some processing of the elements is required, then using `map` is fine. For example

    foo.map { !$0 }

Constructs like

    enum MyError: Error {}
    let myResult: Result<String, MyError> = .success("")
    let result: Result<Any, MyError> = myResult.map { $0 }

may be picked up as false positives by the `array_init` rule. If your codebase contains constructs like this, consider using the `typesafe_array_init` analyzer rule instead.

Configuration (YAML):

  array_init:
    severity: warning

Triggering Examples (violations are marked with '↓'):

Example #1

    foo.↓map({ $0 })

Example #2

    foo.↓map { $0 }

Example #3

    foo.↓map { return $0 }

```

And when rendered via Jazzy:

Screenshot 2024-12-01 at 21 04 11

Rationales may contain MarkDown formatting.

When formatting for the console, any lines containing "```" will be stripped from the output, and any lines between them will be indented by four spaces - so in the raitionale definition, code samples should not be indented - no other formatting will be performed, so any MarkDown will appear as is.

When formatting for jazzy, line containing "```" that start a code block will be assumed to be swift code, and will have `swift` appended to them. If you want to format for a different language, then specify that explicitly when starting the code block.

This assures that any code samples in the rationale will be formatted in a very similar way to any examples that are provided for the rules - see output and screenshots above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -169,6 +169,12 @@
   [#5018](https://github.com/realm/SwiftLint/issues/5018)
 
 ### Bug Fixes
+* Add a new rationale property to rule descriptions, providing a more expansive
+  description of the motivation behind each rule.
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#5681](https://github.com/realm/SwiftLint/issues/5681)
+
+#### Bug Fixes
 
 * Ignore TipKit's `#Rule` macro in `empty_count` rule.  
   [Ueeek](https://github.com/Ueeek)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@
   function parameter can be replaced with an opaque `some` type.  
   [SimplyDanny](https://github.com/SimplyDanny)
 
+* Add a new rationale property to rule descriptions, providing a more expansive
+  description of the motivation behind each rule.  
+  [Martin Redington](https://github.com/mildm8nnered)
+  [#5681](https://github.com/realm/SwiftLint/issues/5681)
+
 ### Bug Fixes
 
 * Fix issue referencing the Tests package from another Bazel workspace.  
@@ -167,11 +172,6 @@
 * Improve performance when exclude patterns resolve to a large set of files.  
   [SimplyDanny](https://github.com/SimplyDanny)
   [#5018](https://github.com/realm/SwiftLint/issues/5018)
-
-* Add a new rationale property to rule descriptions, providing a more expansive
-  description of the motivation behind each rule.  
-  [Martin Redington](https://github.com/mildm8nnered)
-  [#5681](https://github.com/realm/SwiftLint/issues/5681)
 
 #### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -170,7 +170,7 @@
 
 ### Bug Fixes
 * Add a new rationale property to rule descriptions, providing a more expansive
-  description of the motivation behind each rule.
+  description of the motivation behind each rule.  
   [Martin Redington](https://github.com/mildm8nnered)
   [#5681](https://github.com/realm/SwiftLint/issues/5681)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -168,7 +168,6 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#5018](https://github.com/realm/SwiftLint/issues/5018)
 
-### Bug Fixes
 * Add a new rationale property to rule descriptions, providing a more expansive
   description of the motivation behind each rule.  
   [Martin Redington](https://github.com/mildm8nnered)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -173,7 +173,7 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#5018](https://github.com/realm/SwiftLint/issues/5018)
 
-#### Bug Fixes
+### Bug Fixes
 
 * Ignore TipKit's `#Rule` macro in `empty_count` rule.  
   [Ueeek](https://github.com/Ueeek)
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/AnonymousArgumentInMultilineClosureRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/AnonymousArgumentInMultilineClosureRule.swift
@@ -21,7 +21,7 @@ struct AnonymousArgumentInMultilineClosureRule: Rule {
 
         ```
         closure {
-        print(↓$0)
+            print(↓$0)
         }
         ```
         """,
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/AnonymousArgumentInMultilineClosureRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/AnonymousArgumentInMultilineClosureRule.swift
@@ -8,6 +8,23 @@ struct AnonymousArgumentInMultilineClosureRule: Rule {
         identifier: "anonymous_argument_in_multiline_closure",
         name: "Anonymous Argument in Multiline Closure",
         description: "Use named arguments in multiline closures",
+        rationale: """
+        In multiline closures, prefer
+
+        ```
+            closure { arg in
+                print(arg)
+            }
+        ```
+
+        to
+
+        ```
+            closure {
+                print(↓$0)
+            }
+        ```
+        """,
         kind: .idiomatic,
         nonTriggeringExamples: [
             Example("closure { $0 }"),
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/AnonymousArgumentInMultilineClosureRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/AnonymousArgumentInMultilineClosureRule.swift
@@ -12,17 +12,17 @@ struct AnonymousArgumentInMultilineClosureRule: Rule {
         In multiline closures, for clarity, prefer using named arguments
 
         ```
-        closure { arg in
-            print(arg)
-        }
+            closure { arg in
+                print(arg)
+            }
         ```
 
         to anonymous arguments
 
         ```
-        closure {
-            print(↓$0)
-        }
+            closure {
+               print(↓$0)
+            }
         ```
         """,
         kind: .idiomatic,
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/AnonymousArgumentInMultilineClosureRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/AnonymousArgumentInMultilineClosureRule.swift
@@ -9,7 +9,7 @@ struct AnonymousArgumentInMultilineClosureRule: Rule {
         name: "Anonymous Argument in Multiline Closure",
         description: "Use named arguments in multiline closures",
         rationale: """
-        In multiline closures, prefer using named arguments
+        In multiline closures, for clarity, prefer using named arguments
 
         ```
             closure { arg in
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/AnonymousArgumentInMultilineClosureRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/AnonymousArgumentInMultilineClosureRule.swift
@@ -9,7 +9,7 @@ struct AnonymousArgumentInMultilineClosureRule: Rule {
         name: "Anonymous Argument in Multiline Closure",
         description: "Use named arguments in multiline closures",
         rationale: """
-        In multiline closures, prefer
+        In multiline closures, prefer using named arguments
 
         ```
             closure { arg in
@@ -17,7 +17,7 @@ struct AnonymousArgumentInMultilineClosureRule: Rule {
             }
         ```
 
-        to
+        to anonymous arguments
 
         ```
             closure {
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/AnonymousArgumentInMultilineClosureRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/AnonymousArgumentInMultilineClosureRule.swift
@@ -12,17 +12,17 @@ struct AnonymousArgumentInMultilineClosureRule: Rule {
         In multiline closures, for clarity, prefer using named arguments
 
         ```
-            closure { arg in
-                print(arg)
-            }
+        closure { arg in
+            print(arg)
+        }
         ```
 
         to anonymous arguments
 
         ```
-            closure {
-                print(↓$0)
-            }
+        closure {
+            print(↓$0)
+        }
         ```
         """,
         kind: .idiomatic,
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/AnonymousArgumentInMultilineClosureRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/AnonymousArgumentInMultilineClosureRule.swift
@@ -12,17 +12,17 @@ struct AnonymousArgumentInMultilineClosureRule: Rule {
         In multiline closures, for clarity, prefer using named arguments
 
         ```
-            closure { arg in
-                print(arg)
-            }
+        closure { arg in
+            print(arg)
+        }
         ```
 
         to anonymous arguments
 
         ```
-            closure {
-               print(↓$0)
-            }
+        closure {
+        print(↓$0)
+        }
         ```
         """,
         kind: .idiomatic,
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/ArrayInitRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/ArrayInitRule.swift
@@ -13,27 +13,27 @@ struct ArrayInitRule: Rule, @unchecked Sendable {
         constructor over calling `map`. For example
 
         ```
-            Array(foo)
+        Array(foo)
         ```
 
         rather than
 
         ```
-            foo.↓map({ $0 })
+        foo.↓map({ $0 })
         ```
 
         If some processing of the elements is required, then using `map` is fine. For example
 
         ```
-            foo.map { !$0 }
+        foo.map { !$0 }
         ```
 
         Constructs like
 
         ```
-            enum MyError: Error {}
-            let myResult: Result<String, MyError> = .success("")
-            let result: Result<Any, MyError> = myResult.map { $0 }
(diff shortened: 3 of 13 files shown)


SimplyDanny:
That's an ambitious plan with 238 existing rules. 😅 But I very much appreciate the effort. This definitely something helpful and can be extended over time with the infrastructure as a starting point.

A few thoughts from my side:

* There are currently one description and sometimes additional more specific reasons per rule. Now, we would add another form of documentation with the `rational`. I wonder if we should rather extend the description and introduce something like a "default reason" or "violation reason".
* Something similar has been on my ToDo list for rule configuration options for quite a while.
* I think that the text should be entirely Markdown.

mildm8nnered:
> That's an ambitious plan with 238 existing rules. 😅 But I very much appreciate the effort. This definitely something helpful and can be extended over time with the infrastructure as a starting point.

238 is a lot yes, but I think definitely achievable. I've had a lot of fun writing the ones so far, and it's the kind of thing that you can knock some off on a rainy afternoon. Keep going, and before you know it you're done.

I think one think to talk about is "voice" - we could go for something very dry and technical, or somewhat more lighthearted, or even something a bit zany.

I think my take is that some kind of middle ground is good, but it would be nice to have some "character" to the rationale's, and to let them be a _little_ bit opionated in places ("almost no-one enables this rule").

> 
> A few thoughts from my side:
> 
> * There are currently one description and sometimes additional more specific reasons per rule. Now, we would add another form of documentation with the `rational`. I wonder if we should rather extend the description and introduce something like a "default reason" or "violation reason".

So I think `description` has a role, and it's short and widely used.

I kind of see `rationale` as more of a freeform piece of text, that might mention some original source of why X is bad/good, and/or also maybe give some tips about using the rule, or about configuration options that you might want to use, or that people commonly set.

"Why should I care about this rule/how should I get the most out of it", together with a bit more of a narrative explanation of what it actually is - sometimes the examples can be a little obtuse.

> * Something similar has been on my ToDo list for rule configuration options for quite a while.

Yep - I think it would be great to have something more formal for config options.

It's kind of a pain, because we probably have a lot, and someone has to write the text (see rationale's), but we could definitely do a better job of explaining what individual options do.

I think mentions of config options in rationale's are of the kind - "this option is one that you might want to use" rather than documenting the config options per se.

> * I think that the text should be entirely Markdown.

My kind of tacky solution was

1. needs to look good in `swiftlint rules rule_name` output (would be so nice if that picked up the system pager, if set)
2. needs to look good in Jazzy output

So strip "```" for the Terminal and add `swift` for Jazzy, and make sure you get your indentation right for the Terminal yourself.

I think the Terminal case is harder.

Adding `swift` to "```" I kind of saw a service to rationale authors - it's probably swift code, and I know we'll forget to add those manually.

SimplyDanny:
> > That's an ambitious plan with 238 existing rules. 😅 But I very much appreciate the effort. This definitely something helpful and can be extended over time with the infrastructure as a starting point.
> 
> 238 is a lot yes, but I think definitely achievable. I've had a lot of fun writing the ones so far, and it's the kind of thing that you can knock some off on a rainy afternoon. Keep going, and before you know it you're done.

Oh, perfect. Happy to learn you like writing documentation. 😉

> I think one think to talk about is "voice" - we could go for something very dry and technical, or somewhat more lighthearted, or even something a bit zany.
> 
> I think my take is that some kind of middle ground is good, but it would be nice to have some "character" to the rationale's, and to let them be a _little_ bit opionated in places ("almost no-one enables this rule").

I think all these styles are fine. And maybe even a mixture makes them interesting for readers. One consistent style throughout the whole set of rules is anyway almost impossible.
 
> > * I think that the text should be entirely Markdown.
> 
> My kind of tacky solution was
> 
> 1. needs to look good in `swiftlint rules rule_name` output (would be so nice if that picked up the system pager, if set)
> 2. needs to look good in Jazzy output
> 
> So strip "```" for the Terminal and add `swift` for Jazzy, and make sure you get your indentation right for the Terminal yourself.
> 
> I think the Terminal case is harder.

The thing is that without using Markdown, there is no good way to use different text styles and add links to other rules or further references. For the terminal output, all the styling modifiers could be removed afterwards.

> Adding `swift` to "```" I kind of saw a service to rationale authors - it's probably swift code, and I know we'll forget to add those manually.

Important is that it allows to choose other languages as well.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/AnonymousArgumentInMultilineClosureRule.swift
```diff
@@ -8,6 +8,23 @@ struct AnonymousArgumentInMultilineClosureRule: OptInRule {
         identifier: "anonymous_argument_in_multiline_closure",
         name: "Anonymous Argument in Multiline Closure",
         description: "Use named arguments in multiline closures",
+        rationale: """
+        In multiline closures, for clarity, prefer using named arguments
+
+        ```
+            closure { arg in
+                print(arg)
+            }
```
SimplyDanny, line 17: The indentation leads to indentation in the rendered Markdown as well. We should avoid that and rather indent the blocks before it's printed to the console.

Source/SwiftLintBuiltInRules/Rules/Lint/ArrayInitRule.swift
```diff
@@ -8,6 +8,35 @@ struct ArrayInitRule: OptInRule {
...
+
+        Constructs like
+
+        ```
+            enum MyError: Error {}
+            let myResult: Result<String, MyError> = .success("")
+            let result: Result<Any, MyError> = myResult.map { $0 }
+        ```
+
+        may be picked up as false positives by the `array_init` rule. If your codebase contains constructs like this, consider using the `typesafe_array_init` analyzer rule instead.
```
SimplyDanny, line 38: Looks a bit loooonnng ...

mildm8nnered:
> > > That's an ambitious plan with 238 existing rules. 😅 But I very much appreciate the effort. This definitely something helpful and can be extended over time with the infrastructure as a starting point.
> > 
> > 
> > 238 is a lot yes, but I think definitely achievable. I've had a lot of fun writing the ones so far, and it's the kind of thing that you can knock some off on a rainy afternoon. Keep going, and before you know it you're done.
> 
> Oh, perfect. Happy to learn you like writing documentation. 😉

:-)

> 
> > I think one think to talk about is "voice" - we could go for something very dry and technical, or somewhat more lighthearted, or even something a bit zany.
> > I think my take is that some kind of middle ground is good, but it would be nice to have some "character" to the rationale's, and to let them be a _little_ bit opionated in places ("almost no-one enables this rule").
> 
> I think all these styles are fine. And maybe even a mixture makes them interesting for readers. One consistent style throughout the whole set of rules is anyway almost impossible.

I hadn't thought of it like that, but actually I think that is absolutely the best approach. Nothing too prescriptive, and then we can catch anything that goes too far off-piste, or lacks clarity in individual CR.

I think comprehensibility to English as a second language speakers is important, so probably we wouldn't want rationale's in Yorkshire Dialect or Runyonese, sadly.

> 
> > > * I think that the text should be entirely Markdown.
> > 
> > 
> > My kind of tacky solution was
> > 
> > 1. needs to look good in `swiftlint rules rule_name` output (would be so nice if that picked up the system pager, if set)
> > 2. needs to look good in Jazzy output
> > 
> > So strip "```" for the Terminal and add `swift` for Jazzy, and make sure you get your indentation right for the Terminal yourself.
> > I think the Terminal case is harder.
> 
> The thing is that without using Markdown, there is no good way to use different text styles and add links to other rules or further references. For the terminal output, all the styling modifiers could be removed afterwards.

Any tips for how you think we should go about the styling modifiers stripping? Not entirely keen on having a markdown parser just so that I can de-markdown it :-)

> 
> > Adding `swift` to "```" I kind of saw a service to rationale authors - it's probably swift code, and I know we'll forget to add those manually.
> 
> Important is that it allows to choose other languages as well.

So the intention is to ignore any language code that's present, as an override.

Source/SwiftLintBuiltInRules/Rules/Lint/ArrayInitRule.swift
mildm8nnered: Oops yes - I wondered why my builds were failing CI

SimplyDanny:
> I hadn't thought of it like that, but actually I think that is absolutely the best approach. Nothing too prescriptive, and then we can catch anything that goes too far off-piste, or lacks clarity in individual CR.
> 
> I think comprehensibility to English as a second language speakers is important, so probably we wouldn't want rationale's in Yorkshire Dialect or Runyonese, sadly.

Fully support that, especially the last part! 🫣

> > The thing is that without using Markdown, there is no good way to use different text styles and add links to other rules or further references. For the terminal output, all the styling modifiers could be removed afterwards.
> 
> Any tips for how you think we should go about the styling modifiers stripping? Not entirely keen on having a markdown parser just so that I can de-markdown it :-)

Well, there is [swiftlang/swift-markdown](https://github.com/swiftlang/swift-markdown) from Apple which I had been playing around with a little at that time. A new dependency should be well thought of. As these rationales are going to provide such a huge benefit, that would be justified though I guess. Apart from removing any Markdown styling, it would also allow SwiftLint to check that the whole text is actually valid Markdown before rendering the docs in the build.

On the other hand, even unrendered Markdown is so well readable, that I wouldn't complain if we just print the text as is into the console (keeping also URLs). There are also console rendering tools available that could be used to prettify the output. That, however, oughtn't be SwiftLint's job then.

> > > Adding `swift` to "```" I kind of saw a service to rationale authors - it's probably swift code, and I know we'll forget to add those manually.
> > 
> > 
> > Important is that it allows to choose other languages as well.
> 
> So the intention is to ignore any language code that's present, as an override.

👍

mildm8nnered:
Are these CI failures something that I'm doing @SimplyDanny ? - https://buildkite.com/swiftlint/swiftlint/builds/8753 - they all seem to be concurrency related, but it feels like they come and go at random

SimplyDanny:
> Are these CI failures something that I'm doing @SimplyDanny ? - https://buildkite.com/swiftlint/swiftlint/builds/8753 - they all seem to be concurrency related, but it feels like they come and go at random

They fail randomly. It's annoying. A retrigger normally helps. I hope this gets fixed by having strict concurrency checking always on by default. But until then there's still a long way to go.

Optionally, we disable warnings reported as errors in Bazel builds.

mildm8nnered:
> > Are these CI failures something that I'm doing @SimplyDanny ? - https://buildkite.com/swiftlint/swiftlint/builds/8753 - they all seem to be concurrency related, but it feels like they come and go at random
> 
> They fail randomly. It's annoying. A retrigger normally helps. I hope this gets fixed by having strict concurrency checking always on by default. But until then there's still a long way to go.
> 
> Optionally, we can disable warnings reported as errors in Bazel builds for the time being.

It feel like mine always fail :-) Apart from pushing an empty commit, is there a better way to retrigger them?

SimplyDanny:
> > > Are these CI failures something that I'm doing @SimplyDanny ? - https://buildkite.com/swiftlint/swiftlint/builds/8753 - they all seem to be concurrency related, but it feels like they come and go at random
> > 
> > 
> > They fail randomly. It's annoying. A retrigger normally helps. I hope this gets fixed by having strict concurrency checking always on by default. But until then there's still a long way to go.
> > Optionally, we can disable warnings reported as errors in Bazel builds for the time being.
> 
> It feel like mine always fail :-) Apart from pushing an empty commit, is there a better way to retrigger them?

This should be much more stable now. I haven't seen the random failures since the update to Swift 6 and a few code adaptions.

Source/SwiftLintBuiltInRules/Rules/Style/AttributesRule.swift
```diff
@@ -11,6 +11,23 @@ struct AttributesRule: OptInRule {
             Attributes should be on their own lines in functions and types, but on the same line as variables and \
             imports
             """,
+        rationale: """
+        See https://ericasadun.com/2016/10/02/quick-style-survey/ for discussion.
+
+        Summarizing here: "[Erica Sadun's] take on things after the poll and after talking directly with a number of \
```
SimplyDanny, line 17: You could make use of quotes here like so:

```
Erica Sadun says:

> My name is Erica.
```

Source/SwiftLintCore/Models/RuleDescription.swift
```diff
@@ -54,6 +59,30 @@ public struct RuleDescription: Equatable, Sendable {
...
+            if line.contains("```") {
+                if insideMultilineString {
+                    insideMultilineString = false
+                } else {
+                    insideMultilineString = true
+                    if line.hasSuffix("```") {
+                        return line + "swift"
+                    }
+                }
+            }
```
SimplyDanny, line 78: Can we make this duplicated piece reusable somehow?

SimplyDanny (commented):
We can also merge this piece by piece. Every new rational improves the rule. No need to wait for all rules being updated.

mildm8nnered:
Screenshot 2024-11-05 at 17 52 14

This one took a bit more prompting - initially it just talked in generic terms about modern hash methods.

SimplyDanny:
Using the Copilot to craft rule descriptions is a very good idea. This can reduce the effort a lot. 👍

For coding in SwiftLint it's often not very helpful unfortunately.

Naveen-C-Ramachandrappa (approved):

Source/SwiftLintBuiltInRules/Rules/Metrics/ClosureBodyLengthRule.swift
```diff
@@ -1,10 +1,19 @@
...
     static let description = RuleDescription(
         identifier: "closure_body_length",
         name: "Closure Body Length",
         description: "Closure bodies should not span too many lines",
+        rationale: """
+        "Closure bodies should not span too many lines" says it all.
+
+        Possibly you could refactor your closure code and extract some of it into a function.
+
+        Many installations, including SwiftLint's, increase the default warning value from \
```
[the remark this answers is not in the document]
SimplyDanny: I think so. Better not motivate people be as lax as we are. 😅

CHANGELOG.md
```diff
@@ -56,6 +56,11 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#5018](https://github.com/realm/SwiftLint/issues/5018)
 
+* Add a new rationale property to rule descriptions, providing a more expansive
+  description of the motivation behind each rule.
```
SimplyDanny, line 60:
```suggestion
  description of the motivation behind each rule.  
```

Source/SwiftLintBuiltInRules/Rules/Lint/BalancedXCTestLifecycleRule.swift
```diff
@@ -8,6 +8,39 @@ struct BalancedXCTestLifecycleRule: Rule {
...
+        func setUp() {
+            foo = "Bar"
+        }
+
+        func tearDown() {
+            foo = nil
+        }
+        ```
+
+        No memory will be consumed by the value of the variable.
```
SimplyDanny, line 39:
```suggestion
        no memory will be consumed by the value of the variable.
```

Source/SwiftLintBuiltInRules/Rules/Lint/BalancedXCTestLifecycleRule.swift
```diff
@@ -8,6 +8,39 @@ struct BalancedXCTestLifecycleRule: Rule {
         identifier: "balanced_xctest_lifecycle",
         name: "Balanced XCTest Life Cycle",
         description: "Test classes must implement balanced setUp and tearDown methods",
+        rationale: """
+        The `setUp` method of `XCTestCase` can be used to set up variables and resources before \
+        each test is run (or with the `class` variant, before all tests are run).
+
+        The memory model for XCTestCase objects is non-obvious. An instance of the `XCTestCase` \
```
SimplyDanny, line 15: The memory model is important as well. However, this rule is about having a `tearDown` for every `setUp` method implemented for the sake of cleaning up/resetting resources that could otherwise cause effects on other tests (e.g. changed environment variables). Enforcing a `tearDown` at least makes you think about steps to perform after a test has run in order to leave everything in a consistent state and treat each test method as isolated. Basically, what's mentioned in the last sentence.

Source/SwiftLintBuiltInRules/Rules/Lint/BlanketDisableCommandRule.swift
```diff
@@ -9,6 +9,27 @@ struct BlanketDisableCommandRule: Rule, SourceKitFreeRule {
...
                      """,
+        rationale: """
+        The intent of this rule is to prevent code like
+
+        ```
+        // swiftlint:disable force_unwrapping
+        let foo = bar!
+        ```
+
+        which disables the `force_unwrapping` rule for the remainder the file, instead of just for the specific \
```
SimplyDanny, line 20:
```suggestion
        which disables the `force_unwrapping` rule for the remainder of the file, instead of just for the specific \
```

Source/SwiftLintBuiltInRules/Rules/Style/AttributesRule.swift
```diff
@@ -11,6 +11,22 @@ struct AttributesRule: Rule {
             Attributes should be on their own lines in functions and types, but on the same line as variables and \
             imports
             """,
+        rationale: """
+        Erica Sadun says:
+
+        > my take on things after the poll and after talking directly with a number of \
```
SimplyDanny, line 17:
```suggestion
        > My take on things after the poll and after talking directly with a number of \
```

Source/SwiftLintCore/Models/RuleDescription.swift
```diff
@@ -54,6 +60,16 @@ public struct RuleDescription: Equatable, Sendable {
     /// The console-printable string for this description.
     public var consoleDescription: String { "\(name) (\(identifier)): \(description)" }
 
+    /// The console-printable rationale for this description.
+    public var consoleRationale: String? {
+        rationale?.consoleRationale
+    }
+
+    /// The rationale for this description, with MarkDown formatting.
```
SimplyDanny, line 68:
```suggestion
    /// The rationale for this description, with Markdown formatting.
```

SimplyDanny (approved):
Sorry for the nitpicking. But other than that, I like this a lot! 👍

PR #5682, opened 2024-07-20
Check at the violation position if a correction can be applied

SimplyDanny:


Source/SwiftLintCore/Protocols/SwiftSyntaxCorrectableRule.swift
```diff
@@ -40,25 +40,32 @@ public extension SwiftSyntaxCorrectableRule {
         }
 
         // There is no rewriter. Falling back to the correction ranges collected by the visitor (if any).
-        let violationCorrections = makeVisitor(file: file)
+        let violations = makeVisitor(file: file)
             .walk(tree: syntaxTree, handler: \.violations)
-            .compactMap(\.correction)
-        if violationCorrections.isEmpty {
+        if violations.isEmpty {
```
mildm8nnered, line 45: So not specific to this PR, or a big deal generally, but more philosophically, should this be a `guard` - in this condition we want to return, and that would give us a compiler check.

Other philosophical approaches I've seen are "guard should be preconditions at the top of the function", or "I don't like guards".

SimplyDanny: I tend to use what's most straightforward to read, which is, if a `guard` avoids negations of conditions I go with it, otherwise I prefer an `if`.

However, you are right in that the benefit of a `guard` is actually the compiler check ensuring that the surrounding scope is really left. I'll update the code and try to remember that next time. 😅

PR #5685, opened 2024-07-21
make NoEmptyBlock rule to opt-in and add configuration

Ueeek:
As discussed in https://github.com/realm/SwiftLint/pull/5617,
This PR implements 2 changes on `no_empty_block` rule.
+ change rule type to `OptIn`
+ add configuration to disable the code block type. (initializer, function, accessor and statement)

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoEmptyBlockRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoEmptyBlockRule.swift
@@ -55,6 +55,42 @@ struct NoEmptyBlockRule: OptInRule {
             Example("""
             while i < 10 { /* do something */ }
             """),
+            Example("""
+            var flag = true {
+                willSet {}
+            }
+            """, configuration: ["disabled_block_types": ["function_bodies"]]),
+            Example("""
+            func f() {}
+            """, configuration: ["disabled_block_types": ["function_bodies"]]),
+            Example("""
+            deinit {}
+            """, configuration: ["disabled_block_types": ["initializer_bodies"]]),
+            Example("""
+            init() {}
+            """, configuration: ["disabled_block_types": ["initializer_bodies"]]),
+            Example("""
+            for _ in 0..<10 {}
+            """, configuration: ["disabled_block_types": ["statement_blocks"]]),
+            Example("""
+            do {
+            } catch {
+            }
+            """, configuration: ["disabled_block_types": ["statement_blocks"]]),
+            Example("""
+            defer {}
+            """, configuration: ["disabled_block_types": ["statement_blocks"]]),
+            Example("""
+            if flag {
+            } else {
+            }
+            """, configuration: ["disabled_block_types": ["statement_blocks"]]),
+            Example("""
+            repeat {} while (flag)
+            """, configuration: ["disabled_block_types": ["statement_blocks"]]),
+            Example("""
+            while i < 10 {}
+            """, configuration: ["disabled_block_types": ["statement_blocks"]]),
         ],
         triggeringExamples: [
             Example("""
@@ -100,9 +136,6 @@ struct NoEmptyBlockRule: OptInRule {
 private extension NoEmptyBlockRule {
     final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
         override func visitPost(_ node: CodeBlockSyntax) {
-            // No need to show a warning since Empty Block of `guard` is compile error.
-            guard node.parent?.kind != .guardStmt else { return }
-
             if let codeBlockType = node.codeBlockType, configuration.enabledBlockTypes.contains(codeBlockType) {
                 validate(node: node)
             }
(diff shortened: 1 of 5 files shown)


Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoEmptyBlockRule.swift
```diff
@@ -102,13 +103,36 @@ private extension NoEmptyBlockRule {
...
         }
     }
 }
+
+private extension CodeBlockSyntax {
+    var codeBlockType: NoEmptyBlockConfiguration.CodeBlockType? {
+        guard let kind = self.parent?.kind else { return nil }
+        switch kind {
+        case .accessorDecl:
+            return .accessorBodies
```
SimplyDanny, line 127: It's okay to have it separate, but it could also be included in `.functionBodies`.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoEmptyBlockRule.swift
```diff
@@ -102,13 +103,36 @@ private extension NoEmptyBlockRule {
...
-
             violations.append(node.leftBrace.positionAfterSkippingLeadingTrivia)
         }
     }
 }
+
+private extension CodeBlockSyntax {
+    var codeBlockType: NoEmptyBlockConfiguration.CodeBlockType? {
+        guard let kind = self.parent?.kind else { return nil }
+        switch kind {
```
SimplyDanny, line 125:
```suggestion
        switch parent?.kind {
```

Then you can make use of a `switch` expression and leave out the `return` keywords. 😉

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoEmptyBlockRule.swift
```diff
@@ -102,13 +103,36 @@ private extension NoEmptyBlockRule {
...
+        case .accessorDecl:
+            return .accessorBodies
+        case .functionDecl:
+            return .functionBodies
+        case .initializerDecl, .deinitializerDecl:
+            return .initializerBodies
+        case .forStmt, .doStmt, .whileStmt, .repeatStmt, .ifExpr, .guardStmt, .catchClause, .deferStmt:
+            return .statementBlocks
```
SimplyDanny, line 133:
```suggestion
        case .forStmt, .doStmt, .whileStmt, .repeatStmt, .ifExpr, .catchClause, .deferStmt:
            return .statementBlocks
        case .guardStmt:
            return nil
```

That way, you can skip the check in line 104.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/NoEmptyBlockConfiguration.swift
```diff
@@ -0,0 +1,26 @@
...
+        case statementBlocks = "statement_blocks"
+
+        static let all = Set(allCases)
+    }
+
+    @ConfigurationElement(key: "severity")
+    private(set) var severityConfiguration = SeverityConfiguration<Parent>(.warning)
+
+    @ConfigurationElement(key: "disabled")
+    private(set) var disabledBlockTypes: [CodeBlockType] = []
```
SimplyDanny, line 21: Please let the key's name coincide with the variable name:
```suggestion
    @ConfigurationElement(key: "disabled_block_types")
    private(set) var disabledBlockTypes: [CodeBlockType] = []
```
Later, we want to have the keys be inferred from the variable names.

SimplyDanny (changes requested):
Thanks for taking this up!

Besides me inline comments, there are two more general remarks:

* Please add some examples that show how the options work. A configuration deviating from the default will be added now into the rendered examples in the documentation.
* The rule in its current state doesn't consider `AccessorBlock`s at all.

Ueeek:
@SimplyDanny 
Thank you for giving kind feedback.
I fixed all the points you mentioned except for the one below!

> The rule in its current state doesn't consider AccessorBlocks at all.
Is this mean to add more examples related to "Accessor Blocks"?
It would be very appreciated if you could help me understand the comment well 🙇 
Thank you!

SimplyDanny:
> > The rule in its current state doesn't consider AccessorBlocks at all.
> 
> Is this mean to add more examples related to "Accessor Blocks"?
> 
> It would be very appreciated if you could help me understand the comment well 🙇 

I think that the rule wouldn't trigger in the example

```swift
struct S {
  var i: Int {}
}
```

But this is actually invalid Swift code anyway. So you can ignore my comment. Sorry for the confusion.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoEmptyBlockRule.swift
```diff
@@ -54,6 +55,42 @@ struct NoEmptyBlockRule: Rule {
...
+            if flag {
+            } else {
+            }
+            """, configuration: ["disabled_block_types": ["statement_blocks"]]),
+            Example("""
+            repeat {} while (flag)
+            """, configuration: ["disabled_block_types": ["statement_blocks"]]),
+            Example("""
+            while i < 10 {}
+            """, configuration: ["disabled_block_types": ["statement_blocks"]]),
```
SimplyDanny, line 93: Let's merge all the examples for one specific type of block into one, so that the rendering is a bit more dense.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoEmptyBlockRule.swift
```diff
@@ -54,6 +55,42 @@ struct NoEmptyBlockRule: Rule {
...
+            }
+            """, configuration: ["disabled_block_types": ["function_bodies"]]),
+            Example("""
+            func f() {}
+            """, configuration: ["disabled_block_types": ["function_bodies"]]),
+            Example("""
+            deinit {}
+            """, configuration: ["disabled_block_types": ["initializer_bodies"]]),
+            Example("""
+            init() {}
```
SimplyDanny, line 70: All examples should be valid Swift. `init`, `deinit` and `defer` need some declaration around them.

SimplyDanny (changes requested):

SimplyDanny (approved):
Thanks for all the updates!

Ueeek:
@SimplyDanny Thank you for kind feedbacks!

PR #5686, opened 2024-07-21
Rename macros to something more descriptive

SimplyDanny:
`AutoApply` was chosen to match with the method name that it implements. If one has never seen the `apply` method, the name might be confusing. `AutoConfigParser` seems more descriptive here.

The `Make` in `MakeAcceptableByConfigurationElement` was prepended to avoid name clashes with the protocol it implements. Obviously, Swift can differentiate between the macro with the `@` and the protocol. The measure was thus unjustified and we can just omit the prefix.

The previous names are still there but deprecated in case people use them to implement their own custom rules. That's hopefully only for one more release.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/AttributesConfiguration.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/AttributesConfiguration.swift
@@ -1,6 +1,6 @@
 import SwiftLintCore
 
-@AutoApply
+@AutoConfigParser
 struct AttributesConfiguration: SeverityBasedRuleConfiguration {
     typealias Parent = AttributesRule
 
--- a/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/BlanketDisableCommandConfiguration.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/BlanketDisableCommandConfiguration.swift
@@ -1,6 +1,6 @@
 import SwiftLintCore
 
-@AutoApply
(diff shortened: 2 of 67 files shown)


mildm8nnered (approved):
lgtm

SimplyDanny:
@mildm8nnered, I'm still thinking about the names, especially `AcceptableByConfigurationElement`. What about `ConfigurationElementType` or `ConfigurationElementCompatible`?

mildm8nnered:
> @mildm8nnered, I'm still thinking about the names, especially `AcceptableByConfigurationElement`. What about `ConfigurationElementType` or `ConfigurationElementCompatible`?

So I don't have a great mental model of the name-spacing here, or of current protocol naming conventions (there have been so many) but just based on:

```
/// Type of an object that can be used as a configuration element.
```

`ConfigurationElementType` seems to fit the bill, although I kind of feel it implies concreteness a bit in a way that protocol names usually try to avoid a little bit.

SimplyDanny:
> > @mildm8nnered, I'm still thinking about the names, especially `AcceptableByConfigurationElement`. What about `ConfigurationElementType` or `ConfigurationElementCompatible`?
> 
> So I don't have a great mental model of the name-spacing here, or of current protocol naming conventions (there have been so many) but just based on:
> 
> ```
> /// Type of an object that can be used as a configuration element.
> 
> ```
>
> `ConfigurationElementType` seems to fit the bill, although I kind of feel it implies concreteness a bit in a way that protocol names usually try to avoid a little bit.

Naming is hard. So maybe we just stay with the name as proposed in this PR without further changes. 😅

PR #5694, opened 2024-07-24
Add syntax to swiftlint_version to allow for specifying minimum and maximum versions

alex-taffe:
Currently, the configuration supports a `swiftlint_version` option. This option does not work for our workflow because we have many branches that may not be up to date with main/master. However, there are breaking changes that occur in SwiftLint, such as `redundant_optional_initialization` changing between 0.53.0 and 0.55.1, causing our team to not have consistency in their linting.

This adds new syntax to `swiftlint_version` that will compare to the current version installed and throw a warning/error if the installed version does not match the pattern the user specified

> warning: Currently running SwiftLint 0.55.1 but configuration specified minimum version 0.55.3.

The syntax is as follows:
```
swiftlint_version: 0.54.0
swiftlint_version: ">0.54.0"
swiftlint_version: ">=0.54.0"
swiftlint_version: "<0.54.0"
swiftlint_version: "<=0.54.0"
```

### DIFF
--- a/Source/SwiftLintCore/Models/Configuration.swift
+++ b/Source/SwiftLintCore/Models/Configuration.swift
@@ -300,18 +300,16 @@ public struct Configuration {
     static func checkVersion(against versionConstraint: String) { // swiftlint:disable:this function_body_length
         func showInvalidVersionStringError() -> Never {
             let invalidVersionString = """
-            error: swiftlint_version syntax invalid.
-            Please specify a version as follows:
-            0.54.0
-            >0.54.0
-            >=0.54.0
-            <0.54.0
-            <=0.54.0
-            """
+                error: swiftlint_version syntax invalid.
+                Please specify a version as follows:
+                0.54.0
+                >0.54.0
+                >=0.54.0
+                <0.54.0
+                <=0.54.0
+                """
 
-            queuedFatalError(
-                invalidVersionString
-            )
+            queuedFatalError(invalidVersionString)
         }
 
         func compareVersionString(_ versionString: String) {
@@ -350,8 +348,10 @@ public struct Configuration {
 
             if !comparator(Version.current, configVersion) {
                 queuedPrintError(
-                    "warning: Currently running SwiftLint \(Version.current.value) but " +
-                    "configuration specified \(errorDifferentiatorString) \(versionString)."
+                    """
+                    warning: Currently running SwiftLint \(Version.current.value) but \
+                    configuration specified \(errorDifferentiatorString) \(versionString).
+                    """
                 )
                 exit(2)
             }
--- a/Source/SwiftLintCore/Models/Configuration.swift
+++ b/Source/SwiftLintCore/Models/Configuration.swift
@@ -1,3 +1,6 @@
+// swiftlint:disable file_length
+// Everything in this file makes sense to be in this file
+
 import Foundation
 import SourceKittenFramework
 
@@ -124,7 +127,7 @@ public struct Configuration {
 
     /// Creates a `Configuration` by specifying its properties directly,
     /// except that rules are still to be synthesized from rulesMode, ruleList & allRulesWrapped
-    /// and a check against the pinnedVersion is performed if given.
+    /// and a check against the version in use is performed if a constraint is configured.
     ///
     /// - parameter rulesMode:              The `RulesMode` for this configuration.
     /// - parameter allRulesWrapped:        The rules with their own configurations already applied.
@@ -139,7 +142,7 @@ public struct Configuration {
     ///                                     lint as having failed.
     /// - parameter reporter:               The identifier for the `Reporter` to use to report style violations.
     /// - parameter cachePath:              The location of the persisted cache to use with this configuration.
-    /// - parameter pinnedVersion:          The SwiftLint version defined in this configuration.
+    /// - parameter versionConstraint:      The SwiftLint version constraint defined in this configuration.
     /// - parameter allowZeroLintableFiles: Allow SwiftLint to exit successfully when passed ignored or unlintable
     ///                                     files.
     /// - parameter strict:                 Treat warnings as errors.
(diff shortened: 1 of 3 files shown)


alex-taffe:
@SimplyDanny @mildm8nnered  if you could take a look at this when you get a chance, thanks!

SimplyDanny:
I see where you are coming from and why this would help, yet I was wondering whether a new separate key is really the way to go. The list is continuously growing and there is the problem of conflicting options. What about `minimum_swiftlint_version: 0.54.0` and `swiftlint_version: 0.53.0` for example? So there should probably be some consistency checks.

What I was thinking about is an extension of `swiftlint_version` in a way that it supports both fixed versions and a range of version given a minimum version. Something like the syntax `swiftlint_version: >=0.54.0` could be conceivable. What's your opinion on that?

alex-taffe:
Yeah I think I actually like that better, what're your thoughts on syntax? I kinda like [Dart's](https://dart.dev/tools/pub/dependencies). Traditional and carrot. Then if we can't parse, just throw a fatal error.

image


image

alex-taffe:
Also what's the minimum version of macOS that SwiftLint supports? Would love to use new SwiftRegex but it's macOS 13 or newer only

SimplyDanny:
> Also what's the minimum version of macOS that SwiftLint supports? Would love to use new SwiftRegex but it's macOS 13 or newer only

SwiftLint still supports macOS 12. We'd need good reasons to change that. While I fancy as well with SwiftRegex, it's not enough reason to change the minimum deployment version. But that might change when macOS 15 is out.

About the syntax: I personally don't like the caret style so much. Only the three variants

* `0.54.0`
* `>0.54.0`
* `>=0.54.0`

seem to be enough. There is no good reason for `<0.54.0` or `<=0.54.0`, is there?

alex-taffe:
In practice I’m guessing the less than or inclusive versions wouldn’t actually be used much (I doubt we ever will), but doesn’t seem like that much more work so might as well just do it while I’m in there. I’m fine with omitting caret syntax

SimplyDanny:
In your implementation it could be helpful to know that `Version` objects are `Comparable`. 😉

alex-taffe:
@SimplyDanny updated

Source/SwiftLintCore/Models/Configuration.swift
```diff
@@ -286,6 +285,96 @@ public struct Configuration {
...
+                invalidVersionString
+            )
+        }
+
+        func compareVersionString(_ versionString: String) {
+            var versionString = versionString
+            let comparator: (Version, Version) -> Bool
+            let errorDifferentiatorString: String
+
+            if versionString.starts(with: ">=") {
```
SimplyDanny, line 325: You could assign the variables directly making use of an `if` expression.

SimplyDanny (changes requested):
Good starting point! Please also add some tests for all the cases.

alex-taffe:
@SimplyDanny how do you suggest I approach tests on this? AFAIK, XCTest can't handle pulling out info from stdout nor can it intercept calls to `abort()`. I could refactor the method so it throws out an error enum with an intermediary that actually calls abort, but seems like a lot of boilerplate. Does bazel have some other better way to tackle this?

alex-taffe:
@SimplyDanny just seeing if you had any thoughts on this. Apologies for the follow up, this is just rapidly causing chaos amongst our 9000 internal branches (which is a whole separate issue but)

Source/SwiftLintCore/Models/Configuration.swift
```diff
@@ -286,6 +285,96 @@ public struct Configuration {
...
+            guard versionString.range(of: #"^\d+\.\d+\.\d+$"#, options: .regularExpression) != nil else {
+                showInvalidVersionStringError()
+            }
+
+            let configVersion = Version(value: versionString)
+
+            if !comparator(Version.current, configVersion) {
+                queuedPrintError(
+                    "warning: Currently running SwiftLint \(Version.current.value) but " +
+                    "configuration specified \(errorDifferentiatorString) \(versionString)."
```
SimplyDanny, line 354: Prefer a string block.

Source/SwiftLintCore/Models/Configuration.swift
```diff
@@ -286,6 +285,96 @@ public struct Configuration {
...
+                versionString = String(versionString.dropFirst(1))
+                errorDifferentiatorString = "less than"
+            } else {
+                comparator = (==)
+                errorDifferentiatorString = "exactly"
+            }
+
+            // make sure the remaining string is just a version string of numeral.numeral.numeral
+            versionString = versionString.trimmingCharacters(in: .whitespacesAndNewlines)
+            guard versionString.range(of: #"^\d+\.\d+\.\d+$"#, options: .regularExpression) != nil else {
```
SimplyDanny, line 345: Feels like the `Version` initializer should check this.

SimplyDanny (changes requested):

Source/SwiftLintCore/Models/Configuration.swift
```diff
@@ -286,6 +285,96 @@ public struct Configuration {
...
+        func showInvalidVersionStringError() -> Never {
+            let invalidVersionString = """
+            error: swiftlint_version syntax invalid.
+            Please specify a version as follows:
+            0.54.0
+            >0.54.0
+            >=0.54.0
+            <0.54.0
+            <=0.54.0
+            """
```
SimplyDanny, line 310: Indent:
```suggestion
                error: swiftlint_version syntax invalid.
                Please specify a version as follows:
                0.54.0
                >0.54.0
                >=0.54.0
                <0.54.0
                <=0.54.0
                """
```

Source/SwiftLintCore/Models/Configuration.swift
```diff
@@ -286,6 +285,96 @@ public struct Configuration {
...
+
+            queuedFatalError(
+                invalidVersionString
+            )
```
SimplyDanny, line 314:
```suggestion
            queuedFatalError(invalidVersionString)
```

Source/SwiftLintCore/Models/Configuration.swift
```diff
@@ -286,6 +285,96 @@ public struct Configuration {
...
+            guard versionString.range(of: #"^\d+\.\d+\.\d+$"#, options: .regularExpression) != nil else {
+                showInvalidVersionStringError()
+            }
+
+            let configVersion = Version(value: versionString)
+
+            if !comparator(Version.current, configVersion) {
+                queuedPrintError(
+                    "warning: Currently running SwiftLint \(Version.current.value) but " +
+                    "configuration specified \(errorDifferentiatorString) \(versionString)."
```
SimplyDanny, line 354:
```suggestion
                    """
                    warning: Currently running SwiftLint \(Version.current.value) but \
                    configuration specified \(errorDifferentiatorString) \(versionString).
                    """
```

SimplyDanny (commented):
Some more style issues ...

alex-taffe:
@SimplyDanny yeah sorry I pushed a bunch of stuff earlier then got distracted with some fires at work, I was working on refactoring into a struct anyways

alex-taffe:
We went a different direction on this, closing

Issue #5689, opened 2024-07-22
Custom rules as part of `only_rules` don't work (anymore?)


daniel-moh:
I have tried to add a custom rule to an `only_rules` setup. However, when running `swiftlint` (latest version 0.55.1) the custom rule doesn't get recognized in the `only_rules` list. 

I found a related issue indicating that this worked in the past (https://github.com/realm/SwiftLint/issues/5187). So I'm thinking either I am doing something terribly wrong (I tried my best to rule this out, see a reproducible example below) or this is a regression.

Of course, I also:
- [x] Updated SwiftLint to the latest version
- [x] Searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Minimal, reproducible example

Hers's my setup:

```
.
├── .swiftlint.yml
└── Ninja.swift

1 directory, 2 files
```

With a Swift file like so:

``` swift
// Ninja.swift

let name = "Ninja"
```

And a SwiftLint config file with the [exemplary custom rule from the readme](https://github.com/realm/SwiftLint?tab=readme-ov-file#2-regex-custom-rules) and the [`custom_rules` as an item under `only_rules`](https://github.com/realm/SwiftLint?tab=readme-ov-file#2-regex-custom-rules) as described in the docs:

``` yml
# swiftlint.yml

custom_rules:
  no_hiding_in_strings:
    regex: "([nN]inja)"
    match_kinds: string

only_rules:
  - no_hiding_in_strings
```

But it doesn't seem to work:

```
$ swiftlint --version
0.55.1

$ swiftlint lint --config .swiftlint.yml --no-cache Ninja.swift
warning: Found a configuration for 'custom_rules' rule, but it is not present in 'only_rules'.
Linting Swift files at paths Ninja.swift
Linting 'Ninja.swift' (1/1)
Done linting! Found 0 violations, 0 serious in 1 file.
```

For the record, [despite what the docs say](https://github.com/realm/SwiftLint?tab=readme-ov-file#2-regex-custom-rules), I also tried flipping the order in the config file with the `only_rules` item above the `custom_rules`:

``` yml
only_rules:
  - no_hiding_in_strings

custom_rules:
  no_hiding_in_strings:
    regex: "([nN]inja)"
    match_kinds: string
```

But no luck:

```
$  swiftlint --version
0.55.1

$ swiftlint lint --config .swiftlint.yml --no-cache Ninja.swift
[...]

PR #5712, opened 2024-07-29
Clarify usage of custom rules under only rules

daniel-moh:
See https://github.com/realm/SwiftLint/issues/5689.

Clarifies the usage of custom regex rules in combination with `only_rules` by rephrasing the sentence related to it in the readme and providing an explicit example.


SimplyDanny:
That should be clearer. Thanks!

SimplyDanny (approved):

PR #5717, opened 2024-07-31
Remove extra "case" syntax in documentation

UIApplicationMain:
The `wrapInFunc` method for the `EmptyEnumArgumentsRule.swift` code currently has an extra case statement in the output string, resulting in the following showing on the documentation

https://realm.github.io/SwiftLint/empty_enum_arguments.html
Screenshot 2024-07-30 at 9 45 37 PM

This removes that extra `case` syntax from the function, so that `wrapInFunc` is called like the other methods (i.e. `wrapInSwitch`) with the `case` as part of the param instead of from the `Example` output.


SimplyDanny (approved):

SimplyDanny:
Thank you!

PR #5718, opened 2024-07-31
Treat specialized imports with attributes as unique

### LINKED ISSUE #5716: `duplicate_imports` not properly taking `@testable` into account
### New Issue Checklist

### Describe the bug

`duplicate_imports` considers some `@testable` imports as duplicated even though they are not.

```swift
import Foo
@testable import struct Foo.Bar // warning: Duplicate Imports Violation: Imports should be unique (duplicate_imports)
```

The second import is NOT a duplication. `swiftlint --fix` will remove it but that breaks the code if the code relied on using some of `Foo.Bar`'s `internal` methods.

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.55.1
* Installation method used (Homebrew, CocoaPods, building from source, etc)? Mint

SimplyDanny:
Fixes #5716.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,11 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#5592](https://github.com/realm/SwiftLint/issues/5592)
 
+* Don't consider specialized imports with attributes as duplicates in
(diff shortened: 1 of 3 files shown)


vincentisambart:
Thank you!

PR #5723, opened 2024-08-03
Extract new `contrasted_opening_brace` rule from `opening_brace` rule

SimplyDanny:


marcelofabri:
Shouldn't `contrasted_opening_brace` be an opt-in rule? This caused 270k violations in our codebase.

SimplyDanny:
> Shouldn't `contrasted_opening_brace` be an opt-in rule? This caused 270k violations in our codebase.

It definitely should be. Fixed that and will release 0.56.1 in a minute.

Issue #5711, opened 2024-07-28
`--only-rule` option doesn't work without a `.swiftlint.yml` configuration


SimplyDanny:
As the title says, when you run for example `swiftlint --only-rule trailing_comma` in a folder that doesn't contain a `.swiftlint.yml` configuration file, SwiftLint disregards the `--only-rule` option and just runs with the default rule set.

mildm8nnered:
Resolved by #5773

PR #5725, opened 2024-08-03
Fix only rule config issues

mildm8nnered:
WIP

Partly addresses #5711 

`func merged(with child: RulesWrapper) -> RulesWrapper` is a problem, as this will not respect `--only-rules` - we may need a dedicated `onlyRule` RulesMode to get this to work in all cases.


mildm8nnered:
Replaced by #5773

PR #5736, opened 2024-08-07
Update Bazel bzlmod depedencies for `rules_swift` 2.x support

luispadron:
Updates to the required versions to support rules_swift 2.x, this does force users of this change to use rules_swift >= 2.1.1 because SourceKitten requires use of the new `swift_interop_hints` rule. This will take effect in new Bazel registry releases.

- [x] Depends on https://github.com/swiftlang/swift-syntax/pull/2790 being merged and released

I'll be making a registry patch for supporting older swift-syntax versions for SwiftLint 0.55.1

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -8,11 +8,11 @@ module(
 bazel_dep(name = "apple_support", version = "1.16.0", repo_name = "build_bazel_apple_support")
 bazel_dep(name = "bazel_skylib", version = "1.7.1")
 bazel_dep(name = "platforms", version = "0.0.10")
-bazel_dep(name = "rules_apple", version = "3.6.0", repo_name = "build_bazel_rules_apple")
-bazel_dep(name = "rules_swift", version = "1.18.0", repo_name = "build_bazel_rules_swift")
-bazel_dep(name = "sourcekitten", version = "0.35.0", repo_name = "com_github_jpsim_sourcekitten")
-bazel_dep(name = "swift-syntax", version = "600.0.0-prerelease-2024-07-24", repo_name = "SwiftSyntax")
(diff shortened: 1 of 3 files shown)


SimplyDanny:
The latest SwiftLint release relies on a SwiftSyntax pre-release able to scan Swift 6 code. So the `rules_swift` 2.x.y update should also be included in their Swift 6 line to make it actually useable from upcoming SwiftLint versions.

Now we have kind of an issue, in that 0.55.1.1 uses more recent Bazel dependencies than 0.56.0 would be able to be configured with once it's getting added to the Bazel registry.

SimplyDanny:
> > So the rules_swift 2.x.y update should also be included in their Swift 6 line to make it actually useable from upcoming SwiftLint versions.
> 
> swift-syntax doesn't need to use rules_swift 2 but it does need to set the compatibility level, which was done for 510.0.3 and is [already done in main](https://github.com/swiftlang/swift-syntax/blob/main/MODULE.bazel).
> 
> Once they create a new pre-release we can update SwiftLint to use that version and bump the rules_swift version here to 2.x.y

The `release/6.0` branch is separate from `release/5.10` and `main` and requires the change to be specifically cherry-picket to be included in Swift 6.

luispadron:
> The `release/6.0` branch is separate from `release/5.10` and `main` and requires the change to be specifically cherry-picket to be included in Swift 6.

Oh gotcha, I can get that cherry-picked to release/6.0, I'll put up a PR

luispadron:
https://github.com/swiftlang/swift-syntax/pull/2790

SimplyDanny:
You got my approval even though that doesn't count in this project. 😅

SimplyDanny:
Bazel build needs https://github.com/bazelbuild/bazel-central-registry/pull/2638.

PR #5754, opened 2024-08-18
fix: compilation with Xcode 16.1b1

### LINKED ISSUE #5738: Build failure since Xcode 16 Beta 5
### New Issue Checklist

### Bug Description

Screenshot 2024-08-08 at 16 51 45

```
/Users/giginet/.ghq/github.com/realm/SwiftLint/Source/SwiftLintCore/Extensions/Array+SwiftLint.swift:107:17 Mutable capture of 'inout' parameter 'buffer' is not allowed in concurrently-executing code
```

Since the latest Xcode 16 (Beta 5), the build has failed due to the compilation error.

`UnsafeMutableBufferPointer` can't be written from `Sendable` closure to avoid data-races even if products are built with Swift 5 mode.

### Environment

* SwiftLint version Current main (4a83c365b13f21928a340a2959ff6bfeef203be9)
* Xcode version Xcode 16 Beta 5
* Installation method used from Sources
* Configuration file: -

Brett-Best:
This reduces a Swift concurrency error to a warning which resolves compiling with Xcode 16.1b1.

Fix comes from https://github.com/realm/SwiftLint/issues/5738#issuecomment-2282211903.
Resolves #5738.

### DIFF
--- a/Source/SwiftLintCore/Extensions/Array+SwiftLint.swift
+++ b/Source/SwiftLintCore/Extensions/Array+SwiftLint.swift
@@ -103,12 +103,13 @@ public extension Array {
     func parallelMap<T>(transform: (Element) -> T) -> [T] {
         var result = ContiguousArray<T?>(repeating: nil, count: count)
         return result.withUnsafeMutableBufferPointer { buffer in
-            DispatchQueue.concurrentPerform(iterations: buffer.count) { idx in
-                buffer[idx] = transform(self[idx])
+            let localPointer = buffer
+            DispatchQueue.concurrentPerform(iterations: localPointer.count) { idx in
+                localPointer[idx] = transform(self[idx])
             }
-            return buffer.map { $0! }
+            return localPointer.map { $0! }
         }
-    }
+  }
 }
 
 public extension Collection {


Source/SwiftLintCore/Extensions/Array+SwiftLint.swift
```diff
@@ -103,12 +103,13 @@ public extension Array {
     func parallelMap<T>(transform: (Element) -> T) -> [T] {
         var result = ContiguousArray<T?>(repeating: nil, count: count)
         return result.withUnsafeMutableBufferPointer { buffer in
-            DispatchQueue.concurrentPerform(iterations: buffer.count) { idx in
-                buffer[idx] = transform(self[idx])
+            let localPointer = buffer
+            DispatchQueue.concurrentPerform(iterations: localPointer.count) { idx in
+                localPointer[idx] = transform(self[idx])
```
Brett-Best, line 108: > [!WARNING]
> Capture of 'localPointer' with non-sendable type 'UnsafeMutableBufferPointer<T?>' in a `@Sendable` closure; this is an error in the Swift 6 language mode

SimplyDanny (changes requested):
I wouldn't like to have a warning instead. It means people are unblocked, but it's again up to me to fix the warning.

The tooling around strict concurrency seems to be very buggy. The Swift compiler, Bazel and Xcode seem to have different interpretations of what is allowed in which mode. It feels like fixing something for one build method breaks another. I just fixed compile error spuriously appearing in the Bazel build without any other change causing it.

Because of that, I struggle to accept this change because some Xcode beta version (but apparently not the standalone Swift compiler) complains about it.

That said, if we find a solution that even works without a warning, I'm fine to accept it. Please be careful, though, as this part of the code is especially performance critical.

SimplyDanny:
Thanks @Brett-Best for creating the PR. As reported in #5738, the new Xcode beta seems to have this fixed. In this case, I prefer to change nothing for the time being here.

PR #5761, opened 2024-08-19
Align left brace of additional trailing closures with right brace of previous trailing closure

### LINKED ISSUE #5752: `contrasted_opening_brace` false positives in trailing closures and `else if` statements
### New Issue Checklist

I'm using the new `contrasted_opening_brace` rule. I ran across a potential false positive (at least in my opinion) when this rule is used on trailing closures that include parameters.

Examples include (lifted straight from my codebase):
```swift
return allDependencies.map
{ dependency in // <-- Violation here
    .init(name: dependency.fullName, version: dependency.version, directlyDeclared: dependency.declaredDirectly)
}
```

```swift
let originalServicesWithTheirUUIDs: [String: UUID] = services.reduce(into: [:])
{ result, originalService in // <-- Violation here
    result[originalService.name] = originalService.id
}
```

```swift
return AsyncStream
{ continuation in // <-- Violation here
}
```

I'd expect this rule to not consider these examples violations. Perhaps there could be a setting that would not flag these.

SimplyDanny:
Fixes #5752.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Style/ContrastedOpeningBraceRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/ContrastedOpeningBraceRule.swift
@@ -80,17 +80,26 @@ private extension BracedSyntax {
         }
         if let closure = `as`(ClosureExprSyntax.self),
            closure.keyPathInParent == \FunctionCallExprSyntax.trailingClosure {
-           var indentationDecidingToken = closure.leftBrace
-            repeat {
-                if let previousToken = indentationDecidingToken.previousToken(viewMode: .sourceAccurate) {
-                    indentationDecidingToken = previousToken
-                } else {
-                    break
-                }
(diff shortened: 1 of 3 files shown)


Source/SwiftLintBuiltInRules/Rules/Style/ContrastedOpeningBraceRule.swift
```diff
@@ -80,20 +80,29 @@ private extension BracedSyntax {
...
+private extension TokenSyntax {
+    var previousIndentationDecidingToken: TokenSyntax {
+        var indentationDecidingToken = self
+        repeat {
+            if let previousToken = indentationDecidingToken.previousToken(viewMode: .sourceAccurate) {
+                indentationDecidingToken = previousToken
+            } else {
+                break
+            }
+        } while indentationDecidingToken.leadingTrivia.containsNewlines() == false
```
SimplyDanny, line 101:
```suggestion
        } while !indentationDecidingToken.leadingTrivia.containsNewlines()
```

PR #5772, opened 2024-08-28
[Docs] Commented also when `excludes_trivial_init` is false in `missing_docs` rule

coffmark:
## Describe your changes

Already mentioned in docs when excludes_trivial_init is true, but not mentioned in docs when excludes_trivial_init is false.
At the moment, the documentation only mentions when excludes_trivial_init is true. ([Docs](https://realm.github.io/SwiftLint/missing_docs.html))

## Issue ticket number and link
None.


SimplyDanny:
Rule options are only mentioned in the docs if they deviate from their default, thus, an explicit `excludes_trivial_init = false` isn't going to be added to the example either.

The last example on the page basically proves that trivial initializers trigger by default.

coffmark:
@SimplyDanny 
Thanks for your reply, and maintaining SwiftLint.

I understood that the default options should not be mentioned.
I'll close this PR.

Issue #5776, opened 2024-08-31
Untranslated Content in README_CN.md


huixuhx:
### New Issue Checklist

- [x] I've Updated SwiftLint to the latest version.
- [x] I've searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues).

### Feature or Enhancement Proposal

Hi everyone,

I am a native Chinese speaker and recently tried to use SwiftLint by following the instructions in README_CN.md. However, I found that some important content is untranslated, which caused some difficulties for me in installation. Here are the untranslated sections I found:

- Installation using Swift Package Manager
- Installation using Xcode Package Dependency

- These untranslated contents were found on August 30, 2024.

I am willing to fix this issue and make the instructions more friendly to Chinese users.
Feel free to collaborate!

SimplyDanny:
@huixuhx: Is this resolved with your PR being merged?

PR #5777, opened 2024-08-31
Add instructions for installation in README_CN.md

huixuhx:
This change resolves untranslated content found in #5776 by adding translations for installation in README_CN.md

- Translated and added instructions for using Swif Package Manager
- Translated and added instrcutions for using Xcode Package Dependency


SimplyDanny (approved):
Thank you!

Issue #5775, opened 2024-08-30
Disable no_magic_numbers for preview code


rberggreen:
### New Issue Checklist

- [x] I've Updated SwiftLint to the latest version.
- [x] I've searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues).

### Feature or Enhancement Proposal
It would be useful to disable the rule no_magic_numbers for preview code, for the same reason that it's already disabled by default for test code. Previews typically use various hardcoded values. 

<img width="854" alt="Skärmavbild 2024-08-30 kl  14 03 04" src="https://github.com/user-attachments/assets/e811cf50-1ba3-44cc-a465-384932517f3a">

When expanding the `#Preview` macro, it can be observed that all preview code is in a struct implementing the protocol `DeveloperToolsSupport.PreviewRegistry`. However, adding this to the rule's configurable array of `test_parent_classes` does not silence magic number warnings in preview code. Perhaps SwiftLint is run before macro expansion and so perhaps we need a way to disable the rule for code inside certain macros. 

### Rule file example
```yml
only_rules:
  - no_magic_numbers

no_magic_numbers:
  test_parent_classes:
   - "QuickSpec"
   - "XCTestCase"
   - "DeveloperToolsSupport.PreviewRegistry"

```

mildm8nnered:
Resolved by #5778

PR #5778, opened 2024-08-31
no_magic_numbers rule now ignores violations in Preview macros

mildm8nnered:
Addresses #5775

violations of `no_magic_numbers` in SwiftUI's Preview macros (or any macro named `Preview`) will be ignored.


SimplyDanny:
I wonder if we should only ignore the `#Preview` macro specifically or come up with a more general "ignore these structures" setting which would allow to list names of structs, classes, macros, ...

mildm8nnered:
> I wonder if we should only ignore the `#Preview` macro specifically or come up with a more general "ignore these structures" setting which would allow to list names of structs, classes, macros, ...

So right now, we'd ignore nothing at all by default, although `Preview` is a good candidate for a default setting.

It would be pretty easy to say if you're a child of any struct/class/macro named `XXXX` then ignore, but I think macros is an obvious use case, because you will probably have many `#Preview`'s.

For a class or struct, there's only going to be one for any given name - I think it feels a bit cleaner for people to just

`// swiftlint:disable no_magic_numbers` at the top of that file (assuming that they configure `blanket_disable_command` to ignore `no_magic_numbers`, or suppress it locally).

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -241,6 +254,19 @@ private extension ExprSyntaxProtocol {
         }
         return false
     }
+
+    func isContainedInExcludedMacro(_ macrosToIgnore: Set<String>) -> Bool {
```
SimplyDanny, line 258: This could be done in an implementation of `visit(node: MacroExpansionExprSyntax)` that returns `.skipChildren` if the name matches.

SimplyDanny:
> > I wonder if we should only ignore the `#Preview` macro specifically or come up with a more general "ignore these structures" setting which would allow to list names of structs, classes, macros, ...
> 
> So right now, we'd ignore nothing at all by default, although `Preview` is a good candidate for a default setting.

I tend to like a default "`#Preview` macros are ignored" more. Right now, I can't see other use cases for the new option. So we could just ignore the macro content without any option.

SimplyDanny:
> > > I wonder if we should only ignore the `#Preview` macro specifically or come up with a more general "ignore these structures" setting which would allow to list names of structs, classes, macros, ...
> 
> > 
> 
> > So right now, we'd ignore nothing at all by default, although `Preview` is a good candidate for a default setting.
> 
> 
> 
> I tend to like a default "`#Preview` macros are ignored" more. Right now, I can't see other use cases for the new option. So we could just ignore the macro content without any option.

The question here is more whether we need an option at all. The rule could just ignore `#Preview` built-in. Anything that'd contradict that?

mildm8nnered:
> > > > I wonder if we should only ignore the `#Preview` macro specifically or come up with a more general "ignore these structures" setting which would allow to list names of structs, classes, macros, ...
> > 
> > 
> > > 
> > 
> > 
> > > So right now, we'd ignore nothing at all by default, although `Preview` is a good candidate for a default setting.
> > 
> > 
> > I tend to like a default "`#Preview` macros are ignored" more. Right now, I can't see other use cases for the new option. So we could just ignore the macro content without any option.
> 
> The question here is more whether we need an option at all. The rule could just ignore `#Preview` built-in. Anything that'd contradict that?

Not that I know of, but I don't really use macros much myself in my day job. I'll ditch the configuration - we can always add in back in if another use case turns up.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -121,6 +131,14 @@ private extension NoMagicNumbersRule {
             node.isSimpleTupleAssignment ? .skipChildren : .visitChildren
         }
 
+        override func visit(_ node: MacroExpansionExprSyntax) -> SyntaxVisitorContinueKind {
+            if case let .identifier(text) = node.macroName.tokenKind, text == "Preview" {
+                .skipChildren
+            } else {
+                .visitChildren
+            }
```
SimplyDanny, line 139:
```suggestion
            node.macroName.text == "Preview" ? .skipChildren : .visitChildren
```

SimplyDanny (approved):

PR #5781, opened 2024-09-02
Add max number if params allowed for `multiline_parameters`

kimdv:
This PR will configuration option for `multiline_parameters` so it's allowed to have like two params on the same line.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/MultilineParametersConfiguration.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/MultilineParametersConfiguration.swift
@@ -16,20 +16,21 @@ struct MultilineParametersConfiguration: SeverityBasedRuleConfiguration {
             return
         }
         guard maxNumberOfSingleLineParameters >= 1 else {
-            throw Issue.invalidConfiguration(
+            Issue.inconsistentConfiguration(
                 ruleID: Parent.identifier,
                 message: "Option '\($maxNumberOfSingleLineParameters.key)' should be >= 1."
-            )
+            ).print()
+            return
         }
 
         if maxNumberOfSingleLineParameters > 1, !allowsSingleLine {
-            throw Issue.invalidConfiguration(
+            Issue.inconsistentConfiguration(
                 ruleID: Parent.identifier,
                 message: """
                          Option '\($maxNumberOfSingleLineParameters.key)' has no effect when \
                          '\($allowsSingleLine.key)' is false.
                          """
-            )
+            ).print()
         }
     }
 }
--- a/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/MultilineParametersConfiguration.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/MultilineParametersConfiguration.swift
@@ -8,4 +8,28 @@ struct MultilineParametersConfiguration: SeverityBasedRuleConfiguration {
     private(set) var severityConfiguration = SeverityConfiguration<Parent>(.warning)
     @ConfigurationElement(key: "allows_single_line")
     private(set) var allowsSingleLine = true
+    @ConfigurationElement(key: "max_number_of_single_line_parameters")
+    private(set) var maxNumberOfSingleLineParameters: Int?
+
+    func validate() throws {
+        guard let maxNumberOfSingleLineParameters else {
+            return
+        }
+        guard maxNumberOfSingleLineParameters >= 1 else {
+            throw Issue.invalidConfiguration(
+                ruleID: Parent.identifier,
+                message: "Option '\($maxNumberOfSingleLineParameters.key)' should be >= 1."
+            )
+        }
+
+        if maxNumberOfSingleLineParameters > 1, !allowsSingleLine {
+            throw Issue.invalidConfiguration(
+                ruleID: Parent.identifier,
+                message: """
+                         Option '\($maxNumberOfSingleLineParameters.key)' has no effect when \
+                         '\($allowsSingleLine.key)' is false.
+                         """
+            )
+        }
+    }
 }
--- a/Source/SwiftLintBuiltInRules/Rules/Style/MultilineParametersRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/MultilineParametersRule.swift
@@ -43,6 +43,12 @@ private extension MultilineParametersRule {
                 numberOfParameters += 1
             }
 
+            if let maxNumberOfSingleLineParameters = configuration.maxNumberOfSingleLineParameters,
+               configuration.allowsSingleLine,
+               numberOfParameters > maxNumberOfSingleLineParameters {
+                return true
+            }
+
             guard linesWithParameters.count > (configuration.allowsSingleLine ? 1 : 0),
                   numberOfParameters != linesWithParameters.count else {
                 return false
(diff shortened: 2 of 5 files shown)


Source/SwiftLintBuiltInRules/Rules/Style/MultilineParametersRuleExamples.swift
```diff
@@ -336,5 +338,10 @@ internal struct MultilineParametersRuleExamples {
                 configuration: ["allows_single_line": false]),
         Example("func ↓foo(param1: Int, param2: Bool, param3: [String]) { }",
                 configuration: ["allows_single_line": false]),
+        Example("""
+        func ↓foo(param1: Int,
+                  param2: Bool, param3: [String]) { }
```
SimplyDanny, line 343: I'm trying to make sense of this new option. Why would it be helpful to allow this style given `max_number_of_params = 3`?

Source/SwiftLintBuiltInRules/Rules/Style/MultilineParametersRule.swift
```diff
@@ -34,6 +34,11 @@ private extension MultilineParametersRule {
                 return false
             }
 
+            if let maxNumberOfParams = configuration.maxNumberOfParams,
+                signature.parameterClause.parameters.count <= maxNumberOfParams {
```
SimplyDanny, line 38:
```suggestion
               signature.parameterClause.parameters.count <= maxNumberOfParams {
```

Source/SwiftLintBuiltInRules/Rules/Style/MultilineParametersRuleExamples.swift
kimdv: Hmm. Maybe we should only allow 3 params etc, if they are on the same line?
Would that make more sense?

kimdv: We could in this case do the other way around?

So if it's set to 3, and the there is a newline then it will show a warning to put on a single line?
But maybe it would be better in that case to show a warning that all params should be on a newline instead?

SimplyDanny: The default style the rule enforces is that all parameters are on the same line or every parameter is on a separate line.

Optionally, with `allows_single_line = false`, all parameters need to be placed on their own line.

With the new option, you want to allow a specified number of parameters to be placed on a single line. Once the threshold is exceeded, all parameters are required to be placed on a separate line again. That's basically `allows_single_line = true` up to a number of parameters.

I think that enforcing a single line is not practicable, because parameters and their types could consume a lot of space and so placing them on separate line, even without the rule enforcing it, should be fine.

Now, how does `max_number_of_parameters` (maybe we need a better name) play nicely with `allows_single_line = false`? It's a contradiction. On the other hand, `allows_single_line = true` is redundant with `max_number_of_parameters` being configured with a reasonable number. We somehow need to align both, deprecate or reuse `allows_single_line`.

kimdv: Okay so I rearranged some code. 

I still think that `allows_single_line` and `max_number_of_parameters` can live next to each other. 
It's just, when `allows_single_line` is `false` then `max_number_of_parameters` is ignored. I think that is fine. 

I added a test case, where allow single line is false but there is fewer params than  `max_number_of_parameters` define.
Also I added one where there is 3 params on each line. Also within the allow number in `max_number_of_parameters`.

kimdv: @SimplyDanny do you have time to review again? :)

SimplyDanny: So the intention is that single lines can be okay as long as they fit and a certain number of parameters is not exceeded. That seems to be fine.

What about the option's name? Would `max_number_of_single_line_parameters` be clearer in its meaning?

Another improvement would be a warning if `allows_single_line = false` and `max_number_of_parameters > 1`. That can be added later, though, as I need to think first about how to support that in a generic way.

SimplyDanny: > Another improvement would be a warning if `allows_single_line = false` and `max_number_of_parameters > 1`. That can be added later, though, as I need to think first about how to support that in a generic way.

Perhaps something like #5824?

kimdv: > What about the option's name? Would max_number_of_single_line_parameters be clearer in its meaning?

This is now added. 

> Another improvement would be a warning if allows_single_line = false and max_number_of_parameters > 1. That can be added later, though, as I need to think first about how to support that in a generic way.

Should like a good idea to have something where it's possible to give a warnings on clashing configurations. 
I will wait till it's merged, so we can add it right await

SimplyDanny: https://github.com/realm/SwiftLint/pull/5824 has just been merged. Let me know if it's sufficient.

kimdv: I've added configuration. Please revisit when you have time.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/MultilineParametersConfiguration.swift
```diff
@@ -8,4 +8,18 @@ struct MultilineParametersConfiguration: SeverityBasedRuleConfiguration {
...
+    func validate() throws {
+        if let maxNumberOfSingleLineParameters, !allowsSingleLine && maxNumberOfSingleLineParameters > 1 {
+            throw Issue.invalidConfiguration(
+                ruleID: Parent.identifier,
+                message: """
+                         Invalid configuration: 'allows_single_line' and 'max_number_of_single_line_parameters' are mutually exclusive.
+                         They cannot both be active with conflicting values.
+                         """
+            )
+        }
```
SimplyDanny, line 23: Note that the `Invalid configuration: ` part will be added to the message by default.

We could also directly check, that `max_number_of_single_line_parameters` has a sane value in any case.

My suggestion:

```suggestion
        guard let maxNumberOfSingleLineParameters else {
            return
        }
        guard maxNumberOfSingleLineParameters >= 1 else {
            throw Issue.invalidConfiguration(
                ruleID: Parent.identifier,
                message: "Option '\($maxNumberOfSingleLineParameters.key)' should be >= 1."
            )
        }
        if maxNumberOfSingleLineParameters > 1, !allowsSingleLine {
            throw Issue.inconsistentConfiguration(
                ruleID: Parent.identifier,
                message: """
                    Option '\($maxNumberOfSingleLineParameters.key)' has no effect when \
                    '\($allowsSingleLine.key)' is false.
                    """
            )
        }
```

SimplyDanny (changes requested):
The mechanism seems to work. 👍

Let's have a separate test verifying that these issues are thrown. You can intercept the console output with the help of `Issue.captureConsole { ... }`.

kimdv:
> `Issue.captureConsole`

I've tried to make some tests, but I can't seem to get it work as you suggest. 
I think it's because it throws the error?

I get this
```
XCTAssertEqual failed: threw error "invalidConfiguration(ruleID: "multiline_parameters", message: Optional("Option \'max_number_of_single_line_parameters\' should be >= 1."))"
```

I'm considering to do something like 

```swift
XCTAssertThrowsError(try config.apply(configuration: ["max_number_of_single_line_parameters": 2, "allows_single_line": false])) { error in
            
        }
```

SimplyDanny (approved):
Very good. Thanks for the updates and for adding tests!

SimplyDanny:
> > `Issue.captureConsole`
> 
> I've tried to make some tests, but I can't seem to get it work as you suggest. I think it's because it throws the error?

I fixed that by reporting a warning only instead of throwing an error. The rule still behaves reasonably even with confusing option combinations.

Issue #5788, opened 2024-09-11
superfluous_disable_command bug


artsimonyan23:
superfluous_disable_command incorrectly runs all custom rules.

SimplyDanny:
Please provide a concrete example.

mildm8nnered:
I wish I'd never touched this.

I think there is a problem here, where disabled regions are nested.

Given the following config:

```
only_rules:
  - custom_rules
  - superfluous_disable_command

# Custom rules
custom_rules:
  fatal_error:
    name: Fatal Error
    excluded: "Tests/*"
    message: Prefer using `queuedFatalError` over `fatalError` to avoid leaking compiler host machine paths.
    regex: \bfatalError\b
    match_kinds:
      - identifier
    severity: error
  fattal_error:
    name: Fattal Error
    excluded: "Tests/*"
    message: Prefer using `queuedFatalError` over `fatalError` to avoid leaking compiler host machine paths.
    regex: \bfattalError\b
    match_kinds:
      - identifier
    severity: error    
```

and the following input:

```
// swiftlint:disable fatal_error
// swiftlint:disable:next fattal_error
fatalError("something")
// swiftlint:enable fatal_error
```

then

```
% swiftlint tt.swift --config tt.config
Linting Swift files at paths tt.swift
Linting 'tt.swift' (1/1)
/Users/martin.redington/Documents/Source/SwiftLint/tt.swift:1:33: warning: Superfluous Disable Command Violation: SwiftLint rule 'fatal_error' did not trigger a violation in the disabled region; remove the disable command (superfluous_disable_command)
/Users/martin.redington/Documents/Source/SwiftLint/tt.swift:3:1: warning: Superfluous Disable Command Violation: SwiftLint rule 'fattal_error' did not trigger a violation in the disabled region; remove the disable command (superfluous_disable_command)
/Users/martin.redington/Documents/Source/SwiftLint/tt.swift:3:9223372036854775807: warning: Superfluous Disable Command Violation: SwiftLint rule 'fatal_error' did not trigger a violation in the disabled region; remove the disable command (superfluous_disable_command)
Done linting! Found 3 violations, 0 serious in 1 file.
```

But `fatalError` should not be triggering `superfluous_disable_command` here at all.

[...]

PR #5796, opened 2024-09-15
Fix nested disable commands

mildm8nnered:
Addresses #5788

The problem is that Regions are created with all of the rule identifiers that are disabled in them, with a new region being created for each `swiftlint:` command, more or less.

So an inner region may have multiple rules disabled, when disabled commands are nested, but not all of those rules might be violated within that specific region.

I guess people don't actually nest commands much in practice, at least for custom rules, or this would have created more noise.

This was not as hard to fix as I was afraid it might have been.

The basic idea is to remap the regions before passing them to `superfluousDisableCommandViolations`, so that the new regions describe the specific region that an individual rule is disabled within.

We could just calculate the regions that way in the first place, but the regions are used in multiple places, and I didn't want to interfere with any of the other call-sites right now.

Needs more tests in CustomRulesTests for sure.

Probably doesn't need to be a public method, if we're only using it for `superfluousDisableCommandViolations` - I just started with it there, and wanted it exposed right now - the RegionTests changes can probably go as well, once CustomRulesTests have more coverage, but I just wanted to make sure I'd got it right in all the existing cases and some new ones.

### DIFF
--- a/Source/SwiftLintCore/Extensions/SwiftLintFile+Regex.swift
+++ b/Source/SwiftLintCore/Extensions/SwiftLintFile+Regex.swift
@@ -57,6 +57,51 @@ extension SwiftLintFile {
         return regions
     }
 
+    public func remap(regions: [Region]) -> [Region] {
+        var remappedRegions = [Region]()
+        var startMap: [RuleIdentifier: Location] = [:]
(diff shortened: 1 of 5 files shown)


mildm8nnered:
Replaced by #5797

PR #5797, opened 2024-09-15
Fix nested disable commands improved

mildm8nnered:
Addresses https://github.com/realm/SwiftLint/issues/5788

The problem is that Regions are created with all of the rule identifiers that are disabled in them, with a new region being created for each `swiftlint:` command, more or less.

So an inner region may have multiple rules disabled, when disabled commands are nested, but not all of those rules might be violated within that specific region.

I guess people don't actually nest commands much in practice, at least for custom rules, or this would have created more noise.

This was not as hard to fix as I was afraid it might have been.

The basic idea is to remap the regions inside `superfluousDisableCommandViolations`, so that the new regions describe the specific region that an individual rule is disabled within.

### DIFF
--- a/Source/SwiftLintCore/Models/Linter.swift
+++ b/Source/SwiftLintCore/Models/Linter.swift
@@ -19,13 +19,10 @@ private extension Rule {
     func superfluousDisableCommandViolations(regions: [Region],
                                              superfluousDisableCommandRule: SuperfluousDisableCommandRule?,
                                              allViolations: [StyleViolation]) -> [StyleViolation] {
-        guard let superfluousDisableCommandRule else {
+        guard regions.isNotEmpty, let superfluousDisableCommandRule else {
             return []
         }
-        guard regions.isNotEmpty else {
-            return []
-        }
-
+        
         let regionsDisablingSuperfluousDisableRule = regions.filter { region in
             region.isRuleDisabled(superfluousDisableCommandRule)
         }
--- a/Source/SwiftLintCore/Models/Linter.swift
+++ b/Source/SwiftLintCore/Models/Linter.swift
@@ -1,6 +1,8 @@
 import Foundation
 import SourceKittenFramework
 
+// swiftlint:disable file_length
+
 private let warnSourceKitFailedOnceImpl: Void = {
     Issue.genericWarning("SourceKit-based rules will be skipped because sourcekitd has failed.").print()
 }()
@@ -22,7 +24,7 @@ private extension Rule {
         guard regions.isNotEmpty, let superfluousDisableCommandRule else {
             return []
         }
-        
+
         let regionsDisablingSuperfluousDisableRule = regions.filter { region in
             region.isRuleDisabled(superfluousDisableCommandRule)
         }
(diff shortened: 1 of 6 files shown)


Source/SwiftLintCore/Models/Linter.swift
```diff
@@ -145,6 +147,54 @@ private extension Rule {
                           ruleTime: ruleTime,
                           deprecatedToValidIDPairs: deprecatedToValidIDPairs)
     }
+
+    // Produces one region for each disabled rule identifier
```
SimplyDanny, line 151: Perhaps you can add some more documentation and examples here, because I'm having a hard time understanding what's happening and why.

SimplyDanny (approved):
The tests proof that things work as intended. That's probably sufficient here. 😅

Source/SwiftLintCore/Models/Linter.swift
mildm8nnered: Let me have think about what to add. The long-winded explanation is this:

Given these rules:

```
"rule1": [
    "regex": "pattern1"
],
"rule2": [
    "regex": "pattern2"
],
```

and this input:

```
// swiftlint:disable rule1
// swiftlint:disable rule2
let pattern2 = ""
// swiftlint:enable rule2
let pattern1 = ""
// swiftlint:enable rule1
```

The regions of the file will be (`index: start - end disabled rules`)

```
0: 1,27 - 2,26 rule1
1: 2,27 - 4,25 rule1, rule2        rule2 is violated in this region
2: 4,26 - 6,25 rule1                  rule1 is violated in this region
3: 6,26 - EOF  no rules disabled
```

These regions are disjoint/non-overlapping. Our algorithm for detecting superfluous disable violations works fine when the disabled areas are non-overlapping, or non-nested, but in this particular nested case, we will get a false superfluous disable command for `rule1` in region `0` and region `1`.

`decompose` is perhaps not quite the right word - what we do here is rewrite the regions, so that there is a region for each rule. These regions can be overlapping. In this above case, we get:

```
1,27 - 6,25 rule1                rule1 is violated in this region
2,27 - 4,25 rule2               rule2 is violated in this region
6,26 - EOF no rules disabled
```

So we correctly get no superfluous disable command violations in this case. I didn't want to change how we represent regions elsewhere, so we just convert them here for this special case.

SimplyDanny: Okay. Got that. I wonder if we should then generally treat regions like that right from the beginning.

> I didn't want to change how we represent regions elsewhere, so we just convert them here for this special case.

Understandable. We should keep that in mind, however. Looks like we can simplify a few things in this area.

Patrick-Kladek:
@SimplyDanny can we create a new (bugfix) release now that this is merged? 

We use https://github.com/lukepistrol/SwiftLintPlugin, which bundles a precompiled version of Swiftlint via SPM to speed up builds. The plugin repo mirrors the releases from this repo and automatically creates a precompiled version via GitHub actions. 

This would give everyone the benefits of this PR with fast builds.

Issue #5779, opened 2024-09-01
Add support for configuring `lenient: true` in `.swiftlint.yml`


F1248:
### New Issue Checklist

- [x] I've Updated SwiftLint to the latest version.
- [X] I've searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues).

### Feature or Enhancement Proposal

Since I can't pass command line arguments like `--lenient` when using SwiftLint through the SwiftLintBuildToolPlugin, writing `lenient: true` in `.swiftlint.yml` would be the perfect way to configure this, just as it was implemented with `strict` in #5226.

PR #5801, opened 2024-09-20
Added `lenient` command line option

mildm8nnered:
Addresses #5779 

So I basically just copied #5226 (which I wrote).

If you have `strict` and `lenient` as true in the config file, a `fatalError` will be thrown, but either can be overridden by the `--strict` or `--lenient` command line options, and they will also override the opposite setting in the config file.

For example, if you have 

`lenient: true` in your config file, but you specify `--strict` on the command line, you will get the `strict` behaviour, and your config file setting will be ignored.

I tested this manually - it was hard to do a unit test, because we `fatalError` instead of throwing an error right now.

As before, `--strict` and `--lenient` on the command line will also fatal error.

### DIFF
--- a/Tests/SwiftLintFrameworkTests/LintOrAnalyzeOptionsTests.swift
+++ b/Tests/SwiftLintFrameworkTests/LintOrAnalyzeOptionsTests.swift
@@ -1,9 +1,9 @@
 @testable import SwiftLintFramework
 import XCTest
 
-internal typealias Leniency = (strict: Bool, lenient: Bool)
-
 final class LintOrAnalyzeOptionsTests: XCTestCase {
+    typealias Leniency = LintOrAnalyzeOptions.Leniency
+
     func testLeniency() {
         let bothFalse = Leniency(strict: false, lenient: false)
         let bothTrue = Leniency(strict: true, lenient: true)
@@ -14,7 +14,7 @@ final class LintOrAnalyzeOptionsTests: XCTestCase {
         for configuration in parameters {
             for commandLine in parameters {
                 let options = LintOrAnalyzeOptions(leniency: configuration)
-                let leniency: Leniency = options.leniency(strict: commandLine.strict, lenient: commandLine.lenient)
+                let leniency = options.leniency(strict: commandLine.strict, lenient: commandLine.lenient)
                 if commandLine.strict {
                     // Command line takes precedence.
                     XCTAssertTrue(leniency.strict)
--- a/Tests/SwiftLintFrameworkTests/LintOrAnalyzeOptionsTests.swift
+++ b/Tests/SwiftLintFrameworkTests/LintOrAnalyzeOptionsTests.swift
@@ -1,4 +1,3 @@
-
 @testable import SwiftLintFramework
 import XCTest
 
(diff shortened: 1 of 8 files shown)


SimplyDanny:
The options turn more and more complex. It'd be really useful to have the combinations tested.

SimplyDanny:
> > The options turn more and more complex. It'd be really useful to have the combinations tested.
> 
> Did you have anything particular in mind - it would be quite easy to expose the logic in `applyLeniency` for the interactions between command line and config options, or are you thinking of something more.

Not sure if it's clean to move `applyLeniency` into `LintOrAnalyzeOptions`. In there, it would naturally be `internal` and can thus be tested.

mildm8nnered:
> > > The options turn more and more complex. It'd be really useful to have the combinations tested.
> > 
> > 
> > Did you have anything particular in mind - it would be quite easy to expose the logic in `applyLeniency` for the interactions between command line and config options, or are you thinking of something more.
> 
> Not sure if it's clean to move `applyLeniency` into `LintOrAnalyzeOptions`. In there, it would naturally be `internal` and can thus be tested.

Something like this perhaps, which tests the logic, if not the actual violations mapping?

```
extension LintOrAnalyzeOptions {
    // config file settings can be overridden by either `--strict` or `--lenient` command line options
    func leniency(strict: Bool, lenient: Bool) -> (Bool , Bool) {
        let strict = (strict && !self.lenient) || self.strict
        let lenient = (lenient && !self.strict) || self.lenient
        return (strict, lenient)
    }
}
```

Naveen-C-Ramachandrappa (approved):

Tests/SwiftLintFrameworkTests/LintOrAnalyzeOptionsTests.swift
```diff
@@ -0,0 +1,71 @@
+@testable import SwiftLintFramework
+import XCTest
+
+internal typealias Leniency = (strict: Bool, lenient: Bool)
```
SimplyDanny, line 4: Perhaps we can move this into the productive code?

Tests/SwiftLintFrameworkTests/LintOrAnalyzeOptionsTests.swift
```diff
@@ -0,0 +1,71 @@
...
+        let bothFalse = Leniency(strict: false, lenient: false)
+        let bothTrue = Leniency(strict: true, lenient: true)
+        let strict = Leniency(strict: true, lenient: false)
+        let lenient = Leniency(strict: true, lenient: false)
+        let parameters = [ bothFalse, bothTrue, strict, lenient ]
+
+        for configuration in parameters {
+            for commandLine in parameters {
+                let options = LintOrAnalyzeOptions(leniency: configuration)
+                let leniency: Leniency = options.leniency(strict: commandLine.strict, lenient: commandLine.lenient)
```
SimplyDanny, line 17:
```suggestion
                let leniency = options.leniency(strict: commandLine.strict, lenient: commandLine.lenient)
```

Tests/SwiftLintFrameworkTests/LintOrAnalyzeOptionsTests.swift
```diff
@@ -0,0 +1,71 @@
...
+import XCTest
+
+internal typealias Leniency = (strict: Bool, lenient: Bool)
+
+final class LintOrAnalyzeOptionsTests: XCTestCase {
+    func testLeniency() {
+        let bothFalse = Leniency(strict: false, lenient: false)
+        let bothTrue = Leniency(strict: true, lenient: true)
+        let strict = Leniency(strict: true, lenient: false)
+        let lenient = Leniency(strict: true, lenient: false)
```
SimplyDanny, line 11: Can be inlined.

SimplyDanny (approved):

Tests/SwiftLintFrameworkTests/LintOrAnalyzeOptionsTests.swift
mildm8nnered: Interesting - let me try that

Issue #5802, opened 2024-09-20
`prefer_type_checking` and optional types violation


cyberowl:
### New Issue Checklist

- [x] I've Updated SwiftLint to the latest version.
- [x] I've searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues).

### Bug Description

The following code causes false positive `prefer_type_checking` violation:

```swift
let value: String? = nil
print(value as Any? != nil) // -> false
```

Let's try fix it:
```
swiftlint . --fix
```

Now we got this:
```swift
let value: String? = nil
// xcodebuild warning: `'is' test is always true`
print(value is Any?) // -> now true, that's not expected
```

### Environment
- SwiftLint version: 0.57.0
- Xcode version: 16.0 (16A242d)
- Installation method used: Homebrew
- Configuration file: None

mildm8nnered:
At first I thought we should just ignore `SomeType?` completely, but actually I wonder if we should be triggering on `as` at all, as opposed to `as?`?

That would fix this case.

`someVar as? SomeType? != nil` would still get caught, but we could ignore `SomeType?` as well

mildm8nnered:
So I think this case will also be problematic

```
let x: String? = nil
print(x as String? != nil) // false
print(x is String?) // true
```

So I think we want to exclude the OptionalTypeSyntax cases as well

PR #5805, opened 2024-09-21
Dont trigger for non-optional type casting, or for optional types

mildm8nnered:
Addresses #5802

We no longer trigger unless the type casting is optional (`as?`), or if the type is optional.

Essentially, in either case, the correction would change the resulting `true` or `false` value. See #5802 for details.

We also now check for `nil != foo as? Bar` cases

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferTypeCheckingRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferTypeCheckingRule.swift
@@ -24,7 +24,8 @@ struct PreferTypeCheckingRule: Rule {
                 foo.run()
             }
             """),
-            Example("bar as Foo? != nil"),
+            Example("bar as Foo != nil"),
+            Example("nil != bar as Foo"),
             Example("bar as Foo? != nil"),
             Example("bar as? Foo? != nil"),
         ],
@@ -36,9 +37,12 @@ struct PreferTypeCheckingRule: Rule {
                 doSomeThing()
             }
             """),
+            Example("nil != bar ↓as? Foo"),
+            Example("nil != 2*x ↓as? X"),
         ],
         corrections: [
             Example("bar ↓as? Foo != nil"): Example("bar is Foo"),
+            Example("nil != bar ↓as? Foo"): Example("bar is Foo"),
             Example("2*x ↓as? X != nil"): Example("2*x is X"),
             Example("""
             if foo ↓as? Bar != nil {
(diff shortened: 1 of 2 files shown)


Naveen-C-Ramachandrappa (approved):

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferTypeCheckingRule.swift
```diff
@@ -24,6 +24,9 @@ struct PreferTypeCheckingRule: Rule {
                 foo.run()
             }
             """),
+            Example("bar as Foo? != nil"),
+            Example("bar as Foo? != nil"),
```
SimplyDanny, line 28: This line is a duplicate.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferTypeCheckingRule.swift
```diff
@@ -24,6 +24,9 @@ struct PreferTypeCheckingRule: Rule {
...
+            Example("bar as? Foo? != nil"),
```
SimplyDanny, line 29: This could actually be written as `bar is Foo`, couldn't it?

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferTypeCheckingRule.swift
```diff
@@ -79,9 +81,15 @@ private extension PreferTypeCheckingRule {
...
 
 private extension InfixOperatorExprSyntax {
-    var typeChecksWithAsCasting: Bool {
-        self.leftOperand.is(AsExprSyntax.self)
-        && self.operator.as(BinaryOperatorExprSyntax.self)?.operator.tokenKind == .binaryOperator("!=")
-        && self.rightOperand.is(NilLiteralExprSyntax.self)
+    var asExprWithOptionalTypeChecking: AsExprSyntax? {
+        if let asExpr = leftOperand.as(AsExprSyntax.self),
+           asExpr.questionOrExclamationMark?.tokenKind == .postfixQuestionMark,
+           asExpr.type.is(OptionalTypeSyntax.self) == false,
```
SimplyDanny, line 87:
```suggestion
           !asExpr.type.is(OptionalTypeSyntax.self),
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferTypeCheckingRule.swift
SimplyDanny, line 1: This rule doesn't consider that `!=` is commutative. So cases like `nil != bar as? Int` are not found.

SimplyDanny (changes requested):

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferTypeCheckingRule.swift
mildm8nnered: 🤦

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferTypeCheckingRule.swift
mildm8nnered: That is true - I'm not sure how common that is, but it should be pretty easy to cover

SimplyDanny (approved):

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferTypeCheckingRule.swift
SimplyDanny: There's at least one finding. 😉

PR #5806, opened 2024-09-21
Keep only command-line related functions in `swiftlint` module

SimplyDanny:
Swift 6 doesn't allow importing executable targets. So we need to move testable logic into a framework.


mildm8nnered (approved):
lgtm, and compiles and tests fine on my outdated Xcode 15.2

PR #5824, opened 2024-10-13
Add validation hook to configuration parsing

SimplyDanny:


SimplyDanny:
@mildm8nnered, could you have a look?

Source/SwiftLintCore/Models/Issue.swift
```diff
@@ -3,7 +3,7 @@ import Foundation
 /// All possible SwiftLint issues which are printed as warnings by default.
 public enum Issue: LocalizedError, Equatable {
     /// The configuration didn't match internal expectations.
-    case invalidConfiguration(ruleID: String)
+    case invalidConfiguration(ruleID: String, message: String? = nil)
```
mildm8nnered, line 6: I wonder if we could help the reader more with distinguishing between invalid rule configurations, and invalid swiftlint configurations (if you consider those to be different things) - `invalidRuleConfiguration` although I guess that is implied by `ruleID`).

mildm8nnered:
Might be good to include an "example" implementation of validate for an actual rule ...

Source/SwiftLintCore/Models/Issue.swift
SimplyDanny: Absolutely, I'm always a fan of tools providing clear feedback. This one can only be used inside of rules due to the `ruleID` parameter which is not otherwise available.

SimplyDanny:
> Might be good to include an "example" implementation of validate for an actual rule ...

https://github.com/realm/SwiftLint/pull/5781 is going to be the first candidate.

PR #5837, opened 2024-10-25
Replace `description.identifier` with `identifier`

mildm8nnered:
```
public extension Rule {
    /// The rule's unique identifier which is the same as `Rule.identifier`.
    static var identifier: String { description.identifier }
}
```

was defined a while ago, but only used in a few places. This PR replaces all occurrences of `Rule.description.identifier` with `Rule.identifier`

### DIFF
--- a/Source/SwiftLintCore/Protocols/Rule.swift
+++ b/Source/SwiftLintCore/Protocols/Rule.swift
@@ -147,7 +147,7 @@ public extension Rule {
 }
 
 public extension Rule {
-    /// The rule's unique identifier which is the same as `Rule.description.identifier`.
+    /// The rule's unique identifier which is the same as `Rule.identifier`.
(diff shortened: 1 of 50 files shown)


Source/SwiftLintCore/Protocols/Rule.swift
```diff
@@ -147,7 +147,7 @@ public extension Rule {
 }
 
 public extension Rule {
-    /// The rule's unique identifier which is the same as `Rule.description.identifier`.
```
SimplyDanny, line 150: This should not be changed.

SimplyDanny (approved):
Good one!

Source/SwiftLintCore/Protocols/Rule.swift
mildm8nnered: 🤦

Issue #3335, opened 2020-09-09
Rule Request: Cognitive Complexity


Nelyus:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

1. Why should this rule be added? Share links to existing discussion about what
   the community thinks about this.

I find myself regularly disabling the "Cyclomatic complexity" rule for method that only contain a switch, and are really easy to understand.

While discussing this with fellow developers, I was told and given a link about Cognitive Complexity, which aims exactly that : replace cyclomatic complexity to provide a metric best suited to evaluate the difficulty to understand a method. It has the added benefit of evaluating types as well as methods.

I don't have a lot of experience with it by lack of current tooling to mesure it, but for my methods that currently trigger cyclomatic complexity, it makes sense.

[White paper: Cognitive Complexity](https://www.sonarsource.com/resources/white-papers/cognitive-complexity.html)

2. Provide several examples of what _would_ and _wouldn't_ trigger violations.

Only methods having a score higher than the given threshold would trigger a violation.

The first exemple of the paper is the following 2 methods. They have the same cyclomatic complexity (4), but different cognitive complexity. 

```c++
int sumOfPrimes(int max) {
  int total = 0;
  OUT: for (int i = 1; i <= max; ++i) { // +1
    for (int j = 2; j < i; ++j) {       // +2 (for: +1, nesting: +1)
      if (i % j == 0) {                 // +3 (if: +1, nesting: +2)
        continue OUT;                   // +1 (break in the flow: +1)
      }
    }
    total += i;
  }
  return total;
} // Cognitive complexity 7
```

```c++
String getWords(int number) {
  switch (number) {         // +1
    case 1:
      return "one";
    case 2:
      return "a couple";
    case 3:
      return “a few”;
    default:
      return "lots";
  }
} // Cognitive complexity 1
```

It tends to favour code structures easier to read. For exemple *switch* are easier to read than there *if {} else if {}* counterpart, it has less noise, fewer repetition of the same variables.

3. Should the rule be configurable, if so what parameters should be configurable?

The rule would have 2 thresholds: the first for warnings, the second for errors.

They should have a default value, and be configurable.

4. Should the rule be opt-in or enabled by default? Why?

[...]

PR #5838, opened 2024-10-28
Add `Cognitive Complexity Rule`

lorwe:
Resolves [#3335](https://github.com/realm/SwiftLint/issues/3335)

Reference: https://www.sonarsource.com/resources/cognitive-complexity/


Sophrinix:
what needs to happen to get this merged? Is thing we can do to help on this issue. It would be amazing to have cognitive complexity rule enforcement in this library.

Sophrinix:
also why are these warnings being generated here when none of these projects have anything to do with this project?

mildm8nnered:
> what needs to happen to get this merged? Is thing we can do to help on this issue. It would be amazing to have cognitive complexity rule enforcement in this library.

So it will need to be code reviewed and approved, and then merged. Everyone who contributes to SwiftLint works on a volunteer basis, so it can take a while for people to find time to come back to it. If you look at previously merged PRs, you'll see that it can take a while.

I do have some comments on the PR, but I don't have time right now to take a proper look.

Generally if you can keep the branch clean and without merge conflicts that is good. The changelog can be a bit of a hotspot.

mildm8nnered:
> also why are these warnings being generated here when none of these projects have anything to do with this project?

So PRs that change any code in SwiftLint (some PRs are excluded but I can't remember the exact basis offhand) are run against a set of popular open source projects, and their violations compared against what `main` would detect.

This gives a kind of sanity check across a much wider codebase, and inspecting the new violations can be a big clue as to where we might be picking up false alarms, and "fixed" violations can be inspected to see if they really should have been fixed by the changes. In this case, it's a new rule, so you'd only expect new violations.

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@
   [Jordan Rose](https://github.com/jrose-signal)
   [SimplyDanny](https://github.com/SimplyDanny)
 
+* Add `Cognitive Complexity Rule`
```
mildm8nnered, line 19: If you look at previous rule addition changelog entries, they tend to be a bit wordier, and to specify the rule by its identifier. e.g.

```
* Add new `contrasted_opening_brace` rule that enforces opening
  braces to be on a separate line after the preceding declaration.  
```

Source/SwiftLintBuiltInRules/Rules/Metrics/CognitiveComplexityRule.swift
```diff
@@ -0,0 +1,266 @@
...
+            return .visitChildren
+        }
+
+        override func visitPost(_: ForStmtSyntax) {
+            nesting -= 1
+            complexity += nesting + 1
+        }
+
+        override func visit(_: IfExprSyntax) -> SyntaxVisitorContinueKind {
+            nesting += 1
```
mildm8nnered, line 155: It feels like there's quite a lot of repetition in the methods - it would be nice to abstract that way a bit if possible

SimplyDanny (changes requested):
Thank you for bringing up this rule, @lorwe!

Evaluating proper implementation of all the rules stated in the white paper was much easier if `ComplexityVisitor` would be defined in its own file accessible for dedicated testing. In other words, we should first test the complexity calculation on a set of code snippets. Once we are happy with it and find the rules describing Cognitive Complexity being considered, we may embed the visitor in the new SwiftLint rule where we only check for violations according to the computed complexity and the configured limits.

Issue #3750, opened 2021-10-29
The `redundant_type_annotation` rule does more harm than good


regexident:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

##### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ swiftlint lint
```

### Environment

* SwiftLint version: `0.45.0`
* Installation method used: `brew install swiftlint`
* Paste your configuration file:

```yml
opt_in_rules:
  - redundant_type_annotation
```

* Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)?
    No.

* Which Xcode version are you using (check `xcodebuild -version`)?
    ```
    Xcode 12.5.1
    Build version 12E507
    ```

* Do you have a sample that shows the issue? Run `echo "[string here]" | swiftlint lint --no-cache --use-stdin --enable-all-rules`
  to quickly test if your example is really demonstrating the issue. If your example is more
  complex, you can use `swiftlint lint --path [file here] --no-cache --enable-all-rules`.

```swift
struct Foo {}

struct SomeType{
    // This performs a wrong fix:
    public var foo: Foo = Foo()
    // This performs a fix, but really shouldn't:
    public var bar: Bool = true
    public var baz: Int = 42
    public var blee: String = "blee"
}
```

### Expected result

```swift
struct Foo {}

struct SomeType {
    public var foo: Foo = .init()
    public var bar: Bool = true
    public var baz: Int = 42
    public var blee: String = "blee"
}
```

### Actual result

```swift
struct Foo {}

struct SomeType {
    public var foo = Foo()
    public var bar = true
    public var baz: Int = 42
    public var blee: String = "blee"
}
```

The "fix" made the code **less consistent**: Some properties now have an explicit type, others don't. As a reader of that code you now have to parse each line, adapting to explicitly typed vs. implicitly inferred syntaxes on a line-by-line basis. And the selection seems arbitrary, too, given that both, `42` and `"blee"` could just as well have been type-inferred to `Int` and `String`. Not to mention `Double`, `[T]`, `Set<T>`, `[T: U]`, …

"But I think it just looks prettier" is never a good reason for picking a less correct/robust approach.

[...]

PR #5839, opened 2024-10-28
RedundantTypeAnnotation: add 'ignore_type_interfaces' option

jaredgrubb:
The `redundant_type_annotation` and `explicit_type_interface` rules conflict. For users that want to have an explicit type interface, but still be able to write simple code inside the bodies of functions, I want to add an option to `redundant_type_annotation` that has it ignore "type interfaces".

For example, this allows:
```
struct Foo {
    var bar: Bar = Bar() // OK: ignore this! I want explicit types on interfaces.
    func baz() {
       let bar: Bar = Bar() // WARN: redundant_type_annotation kicks in here though.
    }
}
```

I suggest that this addresses the complaint made in Issue #3750.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
@@ -62,7 +62,7 @@ struct RedundantTypeAnnotationRule: OptInRule, SwiftSyntaxCorrectableRule {
                 var url: URL = URL()
                 let myVar: Int? = 0, s: String = ""
             }
-            """, configuration: ["ignore_type_interfaces": true]),
+            """, configuration: ["ignore_properties": true]),
         ],
         triggeringExamples: [
             Example("var url↓:URL=URL()"),
@@ -100,7 +100,7 @@ struct RedundantTypeAnnotationRule: OptInRule, SwiftSyntaxCorrectableRule {
                 let myVar↓: Int = Int(5)
               }
             }
-            """, configuration: ["ignore_type_interfaces": true]),
+            """, configuration: ["ignore_properties": true]),
             Example("let a↓: [Int] = [Int]()"),
             Example("let a↓: A.B = A.B()"),
             Example("""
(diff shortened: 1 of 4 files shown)


Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/RedundantTypeAnnotationConfiguration.swift
```diff
@@ -8,6 +8,8 @@ struct RedundantTypeAnnotationConfiguration: SeverityBasedRuleConfiguration {
     var severityConfiguration = SeverityConfiguration<Parent>(.warning)
     @ConfigurationElement(key: "ignore_attributes")
     var ignoreAttributes = Set<String>(["IBInspectable"])
+    @ConfigurationElement(key: "ignore_type_interfaces")
```
SimplyDanny, line 11: Which other elements apart from properties does this include?

If it's only about properties, we could be more specific with the name by calling it `ignore_properties`.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantTypeAnnotationRule.swift
```diff
@@ -261,3 +274,17 @@ private extension SyntaxKind {
...
+    func shouldSkipRuleCheck(for varDecl: VariableDeclSyntax) -> Bool {
+        if ignoreAttributes.contains(where: { varDecl.attributes.contains(attributeNamed: $0) }) {
+            return true
+        }
+        if ignoreTypeInterfaces,
+           let parentNode = varDecl.parent,
+           parentNode.as(CodeBlockItemSyntax.self) == nil {
+            return true
+        }
+        return false
```
SimplyDanny, line 288:
```suggestion
        return ignoreTypeInterfaces && varDecl.parent?.is(MemberBlockItemSyntax.self) == true
```

SimplyDanny (changes requested):
The new option makes sense to me. Thanks for adding it!

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/RedundantTypeAnnotationConfiguration.swift
jaredgrubb: Yes, I think it is limited only to that. I was trying to echo the other rule (`explicit_type_interfaces`) by the name, but your suggestion is probably more accurate. I can change it!

SimplyDanny (approved):
Thank you, @jaredgrubb!

Issue #5840, opened 2024-10-28
`file_name` rule does not consider fully qualified names of nested types


fraioli:
### New Issue Checklist

- [x] I've Updated SwiftLint to the latest version.
- [x] I've searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues).

### Bug Description

When declaring a nested type in its own file, the `file_name` rule sees a violation if the file is named according to that nested type's fully qualified name. For example, naming the following file `Nested.MyType.swift` is a violation:

```swift
// Nested.MyType.swift

enum Nested {
    struct MyType {
    }
}
```

```bash
$ swiftlint --config config.yaml Nested.MyType.swift     
Linting Swift files at paths Nested.MyType.swift
Linting 'Nested.MyType.swift' (1/1)
/tmp/Nested.MyType.swift:1:1: warning: File Name Violation: File name should match a type or extension declared in the file (if any). (file_name)
Done linting! Found 1 violation, 0 serious in 1 file.
```

The `file_name` rule only allows the above file to be named `Nested.swift` or `MyType.swift`.

### Environment

* SwiftLint version: 0.57.0
* Xcode version: Xcode 16.0, Build version 16A242d
* Installation method used: Homebrew as well as building from source
* Configuration file:

```yml
opt_in_rules:
  - file_name
```

Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? No

PR #5841, opened 2024-10-28
Update `file_name` rule to match fully-qualified names of nested types

fraioli:
This PR is aimed to address Issue #5840. It does the following:

1. Allows the `file_name` rule to match nested types when using fully-qualified names, meaning naming the following file `Nested.MyType.swift` is no longer a violation:

```
// Nested.MyType.swift

enum Nested {
    struct MyType {
    }
}
```

2. Introduces a new option `fully_qualified` to have the `file_name` rule enforce using fully-qualified names, meaning naming the above file `MyType.swift` instead of `Nested.MyType.swift` would become a violation where it wasn't before (naming the file `Nested.swift` would still not be a violation).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,7 +65,7 @@
 
 * Fixes `file_name` rule to match fully-qualified names of nested types.
   Additionally adds a `require_fully_qualified_names` boolean option to enforce 
-  that file names match nested types only using their fully-qualified name.
+  that file names match nested types only using their fully-qualified name.  
   [fraioli](https://github.com/fraioli)
   [#5840](https://github.com/realm/SwiftLint/issues/5840)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,12 @@
   [Martin Redington](https://github.com/mildm8nnered)
   [#5711](https://github.com/realm/SwiftLint/issues/5711)
 
+* Fixes `file_name` rule to match fully-qualified names of nested types.
+  Additionally adds a `fully_qualified` boolean option to enforce that
+  file names match nested types only using their fully-qualified name.
+  [fraioli](https://github.com/fraioli)
+  [#issue_number](https://github.com/realm/SwiftLint/issues/issue_number)
+
 ## 0.57.0: Squeaky Clean Cycle
 
 #### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,10 +64,10 @@
   [#5711](https://github.com/realm/SwiftLint/issues/5711)
 
 * Fixes `file_name` rule to match fully-qualified names of nested types.
-  Additionally adds a `fully_qualified` boolean option to enforce that
-  file names match nested types only using their fully-qualified name.
+  Additionally adds a `require_fully_qualified_names` boolean option to enforce 
+  that file names match nested types only using their fully-qualified name.
   [fraioli](https://github.com/fraioli)
-  [#issue_number](https://github.com/realm/SwiftLint/issues/issue_number)
+  [#5840](https://github.com/realm/SwiftLint/issues/5840)
 
 ## 0.57.0: Squeaky Clean Cycle
 
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
@@ -74,102 +74,80 @@ private class TypeNameCollectingVisitor: SyntaxVisitor {
         super.init(viewMode: .sourceAccurate)
     }
 
-    /// Visits a node to collect its type name and store it as an ancestor type name to prepend to any
-    /// children to form their fully-qualified type names
-    private func visitNode(_ node: some TypeNameCollectible) -> SyntaxVisitorContinueKind {
-        let name = node.typeName
+    /// Calls `visit(name:)` using the name of the provided node
+    private func visit(node: some NamedDeclSyntax) -> SyntaxVisitorContinueKind {
+        visit(name: node.name.trimmedDescription)
+    }
+
+    /// Visits a node with the provided name, storing that name as an ancestor type name to prepend to
+    /// any children to form their fully-qualified names
+    private func visit(name: String) -> SyntaxVisitorContinueKind {
         let fullyQualifiedName = (ancestorNames + [name]).joined(separator: ".")
         names.insert(fullyQualifiedName)
 
+        // If the options don't require only fully-qualified names, then we will allow this node's
+        // name to be used by itself
         if !requireFullyQualifiedNames {
             names.insert(name)
         }
 
-        ancestorNames.push(node.typeName)
+        ancestorNames.push(name)
         return .visitChildren
     }
 
-    /// Removes a node's type name as an ancestor type name once all of its children have been visited
-    private func visitNodePost(_: some TypeNameCollectible) {
+    override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
+        visit(node: node)
+    }
+
+    override func visitPost(_: ClassDeclSyntax) {
         ancestorNames.pop()
     }
 
-    override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
-        visitNode(node)
-    }
-
-    override func visitPost(_ node: ClassDeclSyntax) {
-        visitNodePost(node)
-    }
-
     override func visit(_ node: ActorDeclSyntax) -> SyntaxVisitorContinueKind {
-        visitNode(node)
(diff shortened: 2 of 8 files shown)


Tests/SwiftLintFrameworkTests/FileNameRuleTests.swift
```diff
@@ -21,6 +22,8 @@ final class FileNameRuleTests: SwiftLintTestCase {
             rule = try FileNameRule(configuration: ["suffix_pattern": suffixPattern])
         } else if let nestedTypeSeparator {
             rule = try FileNameRule(configuration: ["nested_type_separator": nestedTypeSeparator])
+        } else if let fullyQualified {
+            rule = try FileNameRule(configuration: ["fully_qualified": fullyQualified])
```
fraioli, line 26: Open to other names here such as `require_fully_qualified`, `require_fully_qualified_names`, etc.

Tests/SwiftLintFrameworkTests/FileNameRuleTests.swift
```diff
@@ -52,6 +59,22 @@ final class FileNameRuleTests: SwiftLintTestCase {
...
+    func testMultipleLevelsDeeplyNestedTypeDoesntTrigger() {
+        XCTAssert(try validate(fileName: "Multiple.Levels.Deeply.Nested.MyType.swift").isEmpty)
+    }
+
+    func testNestedTypeNotFullyQualifiedDoesntTrigger() {
+        XCTAssert(try validate(fileName: "MyType.swift").isEmpty)
+    }
+
+    func testNestedTypeNotFullyQualifiedDoesTriggerWithOverride() {
+        XCTAssert(try !validate(fileName: "MyType.swift", fullyQualified: true).isEmpty)
```
SimplyDanny, line 75:
```suggestion
        XCTAssert(try validate(fileName: "MyType.swift", fullyQualified: true).isNotEmpty)
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
```diff
@@ -56,33 +56,98 @@ struct FileNameRule: OptInRule, SourceKitFreeRule {
...
+        let fullyQualifiedName = (ancestorNames + [name]).joined(separator: ".")
+        names.insert(fullyQualifiedName)
+
+        if !requireFullyQualifiedNames {
+            names.insert(name)
+        }
+    }
+
+    override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
+        ancestorNames.append(node.name.text)
```
SimplyDanny, line 85: Move this line into `addVisitedNodeNames` and call it right here.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
```diff
@@ -56,33 +56,98 @@ struct FileNameRule: OptInRule, SourceKitFreeRule {
 }
 
 private class TypeNameCollectingVisitor: SyntaxVisitor {
+    // All of a visited node's ancestor type names if that node is nested, starting with the furthest
+    // ancestor and ending with the direct parent
+    private var ancestorNames: [String] = []
```
SimplyDanny, line 61: You can use `Stack<String>` as its type.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
```diff
@@ -56,33 +56,98 @@ struct FileNameRule: OptInRule, SourceKitFreeRule {
...
+    // If true, nested types are only allowed in the file name when used by their fully-qualified name
+    // (e.g. `My.Nested.Type` and not just `Type`)
+    private let requireFullyQualifiedNames: Bool
+
+    init(requireFullyQualifiedNames: Bool) {
+        self.requireFullyQualifiedNames = requireFullyQualifiedNames
+        super.init(viewMode: .sourceAccurate)
+    }
+
+    private func addVisitedNodeName(_ name: String) {
```
SimplyDanny, line 75: I propose the following signature:

```suggestion
    private func addNames(for decl: some NamedDeclSyntax) {
```

So you can get rid of the `node.name.text` calls.

SimplyDanny (changes requested):
I like the extension of the rule. Thanks for the contribution!

Please find a few remarks as separate inline comment.

Tests/SwiftLintFrameworkTests/FileNameRuleTests.swift
SimplyDanny: I like the last one most.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
fraioli: Good suggestion. The only problem is with `func visitPost(_ node: ExtensionDeclSyntax)` since `ExtensionDeclSyntax` doesn't conform to `NamedDeclSyntax` (we grab the "name" via `node.extendedType.trimmedDescription`). I can still create a function that all the nodes that do conform to `NamedDeclSyntax` can call though

fraioli: Actually, I can use a protocol and conform some types to it to make this nicer

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
fraioli: Good point. I'm gonna make some changes to remove a lot of duplicate code and put them into common functions

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
```diff
@@ -56,33 +58,118 @@ struct FileNameRule: OptInRule, SourceKitFreeRule {
...
+    var typeName: String {
+        name.trimmedDescription
+    }
+}
+extension ClassDeclSyntax: TypeNameCollectible {}
+extension ActorDeclSyntax: TypeNameCollectible {}
+extension StructDeclSyntax: TypeNameCollectible {}
+extension TypeAliasDeclSyntax: TypeNameCollectible {}
+extension EnumDeclSyntax: TypeNameCollectible {}
+extension ProtocolDeclSyntax: TypeNameCollectible {}
```
SimplyDanny, line 169: That's a lot of overhead only to spare one more method. Honestly, I'd prefer to go with a `visit(name:String)` and a `visit(node:NamedDeclSyntax)`. The latter can just redirect to the first after extracting the name. And `visit(_: ExtensionDeclSyntax)` can call it directly.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
```diff
@@ -56,33 +58,118 @@ struct FileNameRule: OptInRule, SourceKitFreeRule {
...
+    private func visitNode(_ node: some TypeNameCollectible) -> SyntaxVisitorContinueKind {
+        let name = node.typeName
+        let fullyQualifiedName = (ancestorNames + [name]).joined(separator: ".")
+        names.insert(fullyQualifiedName)
+
+        if !requireFullyQualifiedNames {
+            names.insert(name)
+        }
+
+        ancestorNames.push(node.typeName)
```
SimplyDanny, line 88:
```suggestion
        ancestorNames.push(name)
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
fraioli: No problem. I'll go this route instead

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
fraioli: Good catch. This code goes away with the change to remove the protocol.

CHANGELOG.md
```diff
@@ -63,6 +63,12 @@
   [Martin Redington](https://github.com/mildm8nnered)
   [#5711](https://github.com/realm/SwiftLint/issues/5711)
 
+* Fixes `file_name` rule to match fully-qualified names of nested types.
+  Additionally adds a `require_fully_qualified_names` boolean option to enforce 
+  that file names match nested types only using their fully-qualified name.
```
SimplyDanny, line 68: Please append two spaces to the end of the description. They are required to enforce a line break in the rendered Markdown.

SimplyDanny (approved):

SimplyDanny:
Thank you again for your work, @fraioli!

PR #5843, opened 2024-10-29
Fix #3511: vertical_whitespace_between_cases handle '@unknown default'

### LINKED ISSUE #3511: vertical_whitespace_between_cases not triggering when using @unknown before default
### New Issue Checklist

### Describe the bug

The `vertical_whitespace_between_cases` rule doesn't trigger when using `@unknown` before `default` in a `switch`.

```bash
> ./swiftlint lint
Linting 'Test2.swift' (1/2)
Linting 'Test1.swift' (2/2)
/Users/revolt/Downloads/swiftlint-unknown/Test1.swift:11:1: warning: Vertical Whitespace Between Cases Violation: Include a single empty line between switch cases. (vertical_whitespace_between_cases)
/Users/revolt/Downloads/swiftlint-unknown/Test2.swift:11:1: warning: Vertical Whitespace Between Cases Violation: Include a single empty line between switch cases. (vertical_whitespace_between_cases)
/Users/revolt/Downloads/swiftlint-unknown/Test2.swift:13:1: warning: Vertical Whitespace Between Cases Violation: Include a single empty line between switch cases. (vertical_whitespace_between_cases)
Done linting! Found 3 violations, 0 serious in 2 files.
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? `0.43.0-rc.2`
* Installation method used (Homebrew, CocoaPods, building from source, etc)? `https://github.com/realm/SwiftLint/releases/download/0.43.0-rc.2/portable_swiftlint.zip`

```yml
opt_in_rules:
```

* Which Xcode version are you using (check `xcodebuild -version`)? `Xcode 12.2 Build version 12B45b`

`Test1.swift`:

```swift
enum Test {
    case first
    case second
}

let test = Test.first

switch test {
case .first:
    print("first")
case .second:
    print("second")
@unknown default:
    print("default")
}
```

`Test2.swift` _(that works correctly)_:

```swift
enum Test {
    case first
    case second
}

let test = Test.first

switch test {
case .first:
    print("first")
case .second:
    print("second")
default:
    print("default")
}
```

jaredgrubb:
This rule does not recognize the `@unknown default` variant. Fixes #3511 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -145,6 +145,11 @@
   [fraioli](https://github.com/fraioli)
   [#5840](https://github.com/realm/SwiftLint/issues/5840)
 
+* Fixes an issue where the `vertical_whitespace_between_cases` rule does not
+  recognize `@unknown default`.
+  [Jared Grubb](https://github.com/jaredtrubb)
+  [#5788](https://github.com/realm/SwiftLint/issues/3511)
+
 ## 0.57.0: Squeaky Clean Cycle
 
 #### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -146,7 +146,7 @@
   [#5840](https://github.com/realm/SwiftLint/issues/5840)
 
 * Fixes an issue where the `vertical_whitespace_between_cases` rule does not
-  recognize `@unknown default`.
+  recognize `@unknown default`.  
   [Jared Grubb](https://github.com/jaredtrubb)
   [#5788](https://github.com/realm/SwiftLint/issues/3511)
 
--- a/Source/SwiftLintBuiltInRules/Rules/Style/VerticalWhitespaceBetweenCasesRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/VerticalWhitespaceBetweenCasesRule.swift
@@ -23,6 +23,8 @@ struct VerticalWhitespaceBetweenCasesRule: Rule {
             default:
                 print("x is invalid")
 
+            @unknown default:
+                print("x is out of this world")
             }
             """),
         Example("""
@@ -42,6 +44,7 @@ struct VerticalWhitespaceBetweenCasesRule: Rule {
             case 0..<5: print("x is low")
             case 5..<10: print("x is high")
             default: print("x is invalid")
+            @unknown default: print("x is out of this world")
             }
             """),
         // Testing handling of trailing spaces: do not convert to """ style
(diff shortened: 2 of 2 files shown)


SimplyDanny (commented):
What about whitespace other than a single space between the attribute and `default`?

jaredgrubb:
Aka, you think the regex should be more like `@unknown[ ]+default`? It's a little odd, but I suppose the grammar would allow it? I can adjust if you prefer

SimplyDanny:
> Aka, you think the regex should be more like `@unknown[ ]+default`? It's a little odd, but I suppose the grammar would allow it? I can adjust if you prefer

I also think about newlines in between, e.g.

```swift
@unknown
default: ...
```

jaredgrubb:
Ah I see. I hadn't considered that, but I could imagine someone trying that. I'll add a unit test for that and adjust the regex.

jaredgrubb:
I gave it some attempts but I can't figure it out. The rule always seems to add an extra newline between the `@unknown` and `default`, which isn't at all what we want.

The issue is that `default:` is a suffix of `@unknown \n default:`. The Rule seems to loop until there are no more matches, and the rule flags the `default:` on its own without a newline ahead of it and inserts an extra newline between them.

I tried using a negative-lookbehind (`?<!`), but regex seems to require that such lookbehind are fixed-length, and  `@unknown[\s\n]+` is necessarily not fixed length.

I think supporting this would require reworking how the Rule operates (maybe requiring it to use the AST rather than regex?).

Happy to try something else if you have another idea.

I think this patch is strictly better than what is on-tree, even if it's not perfect. So maybe we can merge even with this limitation?

SimplyDanny:
> I think this patch is strictly better than what is on-tree, even if it's not perfect. So maybe we can merge even with this limitation?

It's definitely an improvement already. So I'm fine to merge it as is. However, I'm also open to review a rewrite using SwiftSyntax instead of regex in a next step. 😉

Please rebase and fix the conflicts, though.

CHANGELOG.md
```diff
@@ -145,6 +145,11 @@
   [fraioli](https://github.com/fraioli)
   [#5840](https://github.com/realm/SwiftLint/issues/5840)
 
+* Fixes an issue where the `vertical_whitespace_between_cases` rule does not
+  recognize `@unknown default`.
```
SimplyDanny, line 149:
```suggestion
  recognize `@unknown default`.  
```

SimplyDanny (approved):

PR #5866, opened 2024-11-23
Include AMD64 Linux binary in `artifactbundle`

### LINKED ISSUE #5514: SwiftLint not work with platform=linux/amd64
I download binary file of latest version. 
If I use `FROM --platform=linux/amd64 swift:5.9-jammy` in my CI/CD
It show to me - error: Tool ‘swiftlint’ is not supported on the target platform

bradleymackey:
- Resolves #5514
- **BREAKING**: `SwiftLintBinary-macos.artifactbundle.zip` is renamed to `SwiftLintBinary.artifactbundle.zip`
  - This is because an `artifactbundle` can contain binaries for different platforms, so having separate bundles for different platforms is kind of a leaky abstraction.
  - This also allows this single `artifactbundle` to work on macOS & Linux with no configuration changes on the user's end.
  - Follows SwiftFormat's convention of having a single artifactbundle.
- This change uses the existing Linux build, built in a Docker container, and already referenced by the `Makefile`. This is just a bit of shuffling around and injecting the built linux binary during `spm_artifactbundle`.
- Future work could add an `aarch64` Linux binary into the bundle, I may look into this soon.


SimplyDanny (commented):
Thanks!

> This is because an `artifactbundle` can contain binaries for different platforms, so having separate bundles for different platforms is kind of a leaky abstraction.

The benefit of this might be the reduced size of the bundle. It'll grow significantly with every new platform we support and ship in there.

bradleymackey:
@SimplyDanny That is true: the artifact bundle size with this changes goes from ~10MB to ~50MB. However, I feel like this is an acceptable tradeoff for now:

- 50MB still isn't huge and shouldn't significantly affect many people's workflows;
- `artifactbundle` is designed to contain cross-platform binaries, creating separate bundles for separate platforms would be going against best practices;
- [SwiftFormat does this](https://github.com/nicklockwood/SwiftFormat/releases/tag/0.55.2), a reasonable parallel;
- It's easier for users, as we don't need to conditionally change our dependency based on the platform.

```
// works on macOS and Linux without any further configuration
.binaryTarget(
    name: "swiftlint",
    url: "https://github.com/realm/SwiftLint/releases/download/0.58/SwiftLintBinary.artifactbundle.zip",
    checksum: "..."
),
```

It may be worth reassessing if the bundle grows beyond ~100MB, which is still a threshold that we should be under even with an aarch64 Linux binary as well.

bradleymackey:
Updated with commit that strips the binary (needed to be done in the Dockerfile/Linux context, macOS can't strip the Linux binary). This takes the raw binary from ~170MB to ~80MB, the `artifactbundle.zip` goes from ~50MB to ~40MB

Makefile
```diff
@@ -102,24 +113,24 @@ portable_zip: installables
...
 	cp -f "$(LICENSE_PATH)" "$(ARTIFACT_BUNDLE_PATH)"
-	(cd "$(TEMPORARY_FOLDER)"; zip -yr - "SwiftLintBinary.artifactbundle") > "./SwiftLintBinary-macos.artifactbundle.zip"
+	(cd "$(TEMPORARY_FOLDER)"; zip -yr - "SwiftLintBinary.artifactbundle") > "./SwiftLintBinary.artifactbundle.zip"
 
-zip_linux: docker_image
+zip_linux: docker_image build_linux
 	$(eval TMP_FOLDER := $(shell mktemp -d))
-	docker run swiftlint cat /usr/bin/swiftlint > "$(TMP_FOLDER)/swiftlint"
-	chmod +x "$(TMP_FOLDER)/swiftlint"
+	cp -f $(SWIFTLINT_EXECUTABLE_LINUX_AMD64) "$(TMP_FOLDER)"
```
SimplyDanny, line 127: The name is now `swiftlint_linux_amd64`. Shouldn't it be renamed?
```suggestion
	cp -f $(SWIFTLINT_EXECUTABLE_LINUX_AMD64) "$(TMP_FOLDER)/swiftlint"
```

Makefile
```diff
@@ -102,24 +113,24 @@ portable_zip: installables
...
 	cp -f "$(LICENSE_PATH)" "$(TMP_FOLDER)"
 	(cd "$(TMP_FOLDER)"; zip -yr - "swiftlint" "LICENSE") > "./swiftlint_linux.zip"
 
-zip_linux_release:
+zip_linux_release: build_linux
 	$(eval TMP_FOLDER := $(shell mktemp -d))
-	docker run --platform linux/amd64 "ghcr.io/realm/swiftlint:$(VERSION_STRING)" cat /usr/bin/swiftlint > "$(TMP_FOLDER)/swiftlint"
-	chmod +x "$(TMP_FOLDER)/swiftlint"
+	cp -f "$(SWIFTLINT_EXECUTABLE_LINUX_AMD64)" "$(TMP_FOLDER)"
```
SimplyDanny, line 133:
```suggestion
	cp -f "$(SWIFTLINT_EXECUTABLE_LINUX_AMD64)" "$(TMP_FOLDER)/swiftlint"
```

Makefile
```diff
@@ -102,24 +113,24 @@ portable_zip: installables
...
 	(cd "$(TEMPORARY_FOLDER)"; zip -yr - "swiftlint" "LICENSE") > "./portable_swiftlint.zip"
 
-spm_artifactbundle_macos: installables
+spm_artifactbundle: installables installables_linux
 	mkdir -p "$(ARTIFACT_BUNDLE_PATH)/swiftlint-$(VERSION_STRING)-macos/bin"
-	sed 's/__VERSION__/$(VERSION_STRING)/g' tools/info-macos.json.template > "$(ARTIFACT_BUNDLE_PATH)/info.json"
+	mkdir -p "$(ARTIFACT_BUNDLE_PATH)/swiftlint-$(VERSION_STRING)-linux-gnu/bin"
+	sed 's/__VERSION__/$(VERSION_STRING)/g' tools/info.json.template > "$(ARTIFACT_BUNDLE_PATH)/info.json"
 	cp -f "$(SWIFTLINT_EXECUTABLE)" "$(ARTIFACT_BUNDLE_PATH)/swiftlint-$(VERSION_STRING)-macos/bin"
+	cp -f "$(SWIFTLINT_EXECUTABLE_LINUX_AMD64)" "$(ARTIFACT_BUNDLE_PATH)/swiftlint-$(VERSION_STRING)-linux-gnu/bin"
```
SimplyDanny, line 121:
```suggestion
	cp -f "$(SWIFTLINT_EXECUTABLE_LINUX_AMD64)" "$(ARTIFACT_BUNDLE_PATH)/swiftlint-$(VERSION_STRING)-linux-gnu/bin/swiftlint"
```

SimplyDanny (approved):
Looks good to me! I only wonder about the name of the Linux binary. Can we go with just `swiftlint` as well?

bradleymackey:
@SimplyDanny The exported binary is now just called `swiftlint`. I've left `SWIFTLINT_EXECUTABLE_LINUX_AMD64` (used during the build process) named `swiftlint_linux_amd64` so we can differentiate the binary when it's placed in `BINARIES_FOLDER` (as the macOS binary is also stored in here during `installables`)

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@
 
 * The command plugin now requires write permissions so that it works with the `--fix` option without an error.  
   [SimplyDanny](https://github.com/SimplyDanny)
+* The artifactbundle name has changed.
+  `SwiftLintBinary-macos.artifactbundle.zip` is now called `SwiftLintBinary.artifactbundle.zip`.
+  This now includes an AMD64 Linux binary.
+  [Bradley Mackey](https://github.com/bradleymackey)
+  [#5514](https://github.com/realm/SwiftLint/issues/5514)
```
SimplyDanny, line 11:
```suggestion
* The command plugin now requires write permissions so that it works with the `--fix` option without an error.  
  [SimplyDanny](https://github.com/SimplyDanny)

* The artifact bundle name has changed. `SwiftLintBinary-macos.artifactbundle.zip` is now called 
  `SwiftLintBinary.artifactbundle.zip`. It now includes an AMD64 Linux binary.
  [Bradley Mackey](https://github.com/bradleymackey)
  [#5514](https://github.com/realm/SwiftLint/issues/5514)
```

SimplyDanny:
Thanks again for the contribution, @bradleymackey!

As you seem to be experienced in this field, are you interested in working on an arm64 Linux binary as well?

bradleymackey:
@SimplyDanny Thanks for your amazing work on this library! Sure, I'd be happy to look into it 😄

PR #5871, opened 2024-11-28
Replace identity expressions with `\.self`

SimplyDanny:


mildm8nnered (approved):
lgtm - tested out fine on my codebase

PR #5895, opened 2024-12-10
Docs: Update pre-commit-hook version

Kirade:
Updated pre-commit rev into recent version in README.


SimplyDanny (approved):
Thank you!

We should think about automating the version update somehow to avoid this inconsistency.

PR #5902, opened 2024-12-20
Add new `redundant_sendable` rule

SimplyDanny:
Inspired by https://www.massicotte.org/step-by-step-reading-from-storage.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/RedundantSendableRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/RedundantSendableRule.swift
@@ -1,0 +1,130 @@
+import SwiftSyntax
+
+@SwiftSyntaxRule(explicitRewriter: true)
+struct RedundantSendableRule: Rule {
+    var configuration = RedundantSendableConfiguration()
+
+    static let description = RuleDescription(
+        identifier: "redundant_sendable",
+        name: "Redundant Sendable",
+        description: "Sendable conformance is redundant on an actor-isolated type",
+        kind: .lint,
+        nonTriggeringExamples: [
+            Example("struct S: Sendable {}"),
+            Example("class C: Sendable {}"),
+            Example("actor A {}"),
+            Example("@MainActor struct S {}"),
+            Example("@MyActor enum E: Sendable { case a }"),
+        ],
+        triggeringExamples: [
+            Example("@MainActor struct ↓S: Sendable {}"),
+            Example("actor ↓A: Sendable {}"),
+            Example("@MyActor enum ↓E: Sendable { case a }", configuration: ["global_actors": ["MyActor"]]),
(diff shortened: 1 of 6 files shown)


SimplyDanny:
@mattmassicotte: I'd appreciate your opinion on the test cases/examples and the rule's description (which is also the message shown for a violation). Actually, I'm not super happy with the wording.

Source/SwiftLintBuiltInRules/Rules/Lint/RedundantSendableRule.swift
```diff
@@ -0,0 +1,130 @@
...
+        identifier: "redundant_sendable",
+        name: "Redundant Sendable",
+        description: "Sendable conformance is redundant on an isolated type",
+        kind: .lint,
+        nonTriggeringExamples: [
+            Example("struct S: Sendable {}"),
+            Example("class C: Sendable {}"),
+            Example("actor A {}"),
+            Example("@MainActor struct S {}"),
+            Example("@MyActor enum E: Sendable { case a }"),
```
mattmassicotte, line 17: Why should this last one not trigger?

Source/SwiftLintBuiltInRules/Rules/Lint/RedundantSendableRule.swift
```diff
@@ -0,0 +1,130 @@
...
+        ],
+        triggeringExamples: [
+            Example("@MainActor struct ↓S: Sendable {}"),
+            Example("actor ↓A: Sendable {}"),
+            Example("@MyActor enum ↓E: Sendable { case a }", configuration: ["global_actors": ["MyActor"]]),
```
mattmassicotte, line 22: Oh I think I get it, the actor has to be defined first?

mattmassicotte (approved):
This looks like a super-cool addition! And it's already catching (likely) real mistakes in projects today so well done!

mattmassicotte:
As for the description, I think you can simplify things a little if you want:

> Add new `redundant_sendable` rule that triggers on `Sendable` conformances of types that are implicitly already `Sendable` due to being actor-isolated.

Source/SwiftLintBuiltInRules/Rules/Lint/RedundantSendableRule.swift
SimplyDanny: Since the rule only operates on syntax-level, it needs to have global actors defined by name. There's no other way to know if the attribute is an actor or something else. `@MainActor` is clear and perhaps also `@...Actor` but that's too vague. Global actors aren't something that changes often, so people can just add their names to the configuration. It's always a trade-off.

SimplyDanny:
At the moment, this rule needs to be enabled explicitly. Feels like it's clear and safe enough to enable it by default though.

Issue #5904, opened 2024-12-22
The check for updates command line/configuration setting on the `lint` and `analyze` commands, no update check is performed


mildm8nnered:
### New Issue Checklist

- [x] I've Updated SwiftLint to the latest version.
- [x] I've searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues).

### Bug Description

When using the `--check-for-updates` command line option or configuration file setting on the `lint` and `analyze` commands, no update is performed, because we `exit(2)` once we've determined that violations are present

Create a file with some violations. For example

```swift
let foo = try! bar()
```
Then, run something like:

```bash
% swiftlint.debug lint --check-for-updates  t.swift
Linting Swift files at paths t.swift
Linting 't.swift' (1/1)
/tmp/t.swift:2:11: error: Force Try Violation: Force tries should be avoided (force_try)
Done linting! Found 1 violation, 1 serious in 1 file.
%
% echo $?
2
```

We should see a message relating to the update check as well, such as

```
Your version of SwiftLint is up to date.
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)

`0.57.1`

* Xcode version (run `xcodebuild -version` to be sure)

16.1

* Installation method used (Homebrew, CocoaPods, building from source, etc)

Cocoa Pods

PR #5905, opened 2024-12-22
Still check for updates if we have violations

mildm8nnered:
Addresses #5904

### DIFF
--- a/Source/SwiftLintFramework/LintOrAnalyzeCommand.swift
+++ b/Source/SwiftLintFramework/LintOrAnalyzeCommand.swift
@@ -145,11 +145,15 @@ package struct LintOrAnalyzeCommand {
         if let baselineOutputPath = options.writeBaseline ?? builder.configuration.writeBaseline {
             try Baseline(violations: builder.unfilteredViolations).write(toPath: baselineOutputPath)
         }
+        var numberOfSeriousViolations = 0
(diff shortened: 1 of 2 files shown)


SimplyDanny:
Alternatively, we could check for updates before processing violations.

mildm8nnered:
> Alternatively, we could check for updates before processing violations. Or is the workflow always like "first do the actual work, then check for updates"?

So I think check for updates is something that we want to happen last, or at least after the main action, in my mind.

SimplyDanny (approved):

Source/SwiftLintFramework/LintOrAnalyzeCommand.swift
```diff
@@ -145,12 +145,16 @@ package struct LintOrAnalyzeCommand {
         if let baselineOutputPath = options.writeBaseline ?? builder.configuration.writeBaseline {
             try Baseline(violations: builder.unfilteredViolations).write(toPath: baselineOutputPath)
         }
+        var numberOfSeriousViolations = 0
```
SimplyDanny, line 148: This will become an issue when we enforce strict concurrency checks. The API should allow to write this as:
```swift
let numberOfSeriousViolations = Signposts.record(...) {
    try postProcessViolations(files: files, builder: builder)
}
```

PR #5915, opened 2024-12-26
Add new `opaque_over_existential` rule

SimplyDanny:
Inspired by https://www.douggregor.net/posts/swift-for-cxx-practitioners-type-erasure.


SimplyDanny:
This rule leads to false positives when a protocol requires an `any` argument, e.g.

```swift
protocol P {}
protocol R {
  func f(p: any P)
}
struct S: R {
  func f(p: some P) {} // compile error
}
```

See https://forums.swift.org/t/existential-parameter-type-cannot-be-implemented-with-opaque-type/76895/1.

PR #5918, opened 2024-12-28
Fix: TipKit rule triggers an empty_count violation

### LINKED ISSUE #5883: Event-based TipKit rule triggers an `empty_count` violation
### New Issue Checklist

### Bug Description

TipKit uses the `#Rule` macro to define the conditions that must be fulfilled for a tip to be displayed:

```swift
import TipKit

struct SomeHelpfulTip: Tip {
    var rules: [Rule] {
        #Rule(Tips.someEventHappened) { $0.donations.count > 0 }
    }
}
```

`$0.donations.count > 0` triggers the `empty_count` rule violation. However, changing this to `$0.donations.isEmpty` causes a build error, "Event Rules do not support keyPath operations.", so I have to resort to disabling the `empty_count` rule here. `#Rule` is a Swift macro that expects an event donation count to be compared with a number, failing to generate code otherwise, so there aren't any alternatives.

### Environment

* SwiftLint version: 0.57.1
* Xcode version: 16.1
* Configuration file: 

<details>
  <summary>swiftlint.yml</summary>
  
```yml
  - Scripts
  - ReferenceImages
  - BuildSystem
  - Configurations
  - AppStoreConnect
  - node_modules
disabled_rules: # rule identifiers to exclude from running
  - line_length
  - identifier_name
  - todo
  - file_length
  - nesting
  - comma_inheritance
  - type_body_length
  - trailing_whitespace
  - static_over_final_class
opt_in_rules:
  - discouraged_assert
  - empty_count
  - first_where
  - empty_string
  - toggle_bool
  - convenience_type
  - overridden_super_call
  - shorthand_optional_binding
  - accessibility_label_for_image
  - accessibility_trait_for_button
  - array_init
  - attributes
  - closure_end_indentation
  - collection_alignment
  - yoda_condition
  - direct_return
analyzer_rules:
  - unused_import
custom_rules:
  invalid_hex_color:
    included: ".*\\.swift"
    name: "Invalid hex color"
    regex: 'hexString: "\#{0}[0-9a-fA-F]{3,8}"'
    message: "'hexString' should always be prefixed with a #"
    severity: error
  localizedDescription:
    included: ".*\\.swift"
    name: "Don't use localizedDescription"
    regex: '(\w+\.localizedDescription)'
    message: 'Don''t use .localizedDescription. Use "(error)" instead to get the full error string'
    severity: error
  no_ui_colors:
    name: "Don't use Standard UIColors"
    excluded: "(WidgetExtension)\\/.*|.*\\/WidgetExtension\\.swift|.*\\/UIColor\\+Extensions\\.swift|.*\\/Color\\+Extensions\\.swift"
    regex: '[.]\b(themeRed|themeBlue|themeWhite|themeBlack|systemBlue|systemGreen|systemIndigo|systemOrange|systemPink|systemPurple|systemRed|systemTeal|systemYellow|systemGray|systemGray2|systemGray3|systemGray4|systemGray5|systemGray6|black|blue|brown|cyan|darkGray|gray|green|lightGray|magenta|orange|purple|red|white|yellow)\b'
    message: "Only use the colors from UIColor+Extensions.swift"
    severity: warning
  no_color_initializers:
    name: "Don't use color initializers"
    excluded: "(WidgetExtension)\\/.*|.*\\/WidgetExtension\\.swift|.*\\/UIColor\\+Extensions\\.swift|.*\\/Color\\+Extensions\\.swift"
    regex: '\s(UIColor\(.*\)|.init\(hexString:.*\)|.init\(hue:.*saturation:.*brightness:.*\)|.init\(white:.*alpha:.*\)|.init\(red:.*green:.*blue:.*\)|.init\(red:.*green:.*blue:.*alpha:.*\))'
    message: "Only use the colors from UIColor+Extensions.swift"
    severity: warning
  no_color_generators:
    name: "Don't use color generators"
    excluded: "(WidgetExtension)\\/.*|.*\\/WidgetExtension\\.swift|.*\\/UIColor\\+Extensions\\.swift"
    regex: '\.(withAlphaComponent|darker|lighter)'
    message: "Only use the colors from UIColor+Extensions.swift"
    severity: warning
  no_optional_decode_initial_overridable:
    regex: 'decodeIfPresent\(InitialOverridable'
    name: "Don't decodeIfPresent InitialOverridable<T>"
    message: "InitialOverridable<T> should always be decoded as non-optional."
  no_raw_nsattributed_string_attributes:
    regex: 'NSAttributedString\(string: [\S]+,\s+attributes: [\S]+\)'
    name: "Do not use raw NSAttributedString attributes"
    message: "Use .init(string:safeAttributes:) instead"
  use_makefont:
    name: "Avoid UIFont(name:size)"
    regex: 'UIFont\(name: .*'
    message: "Use `UIFont.makeFont(name:size)` instead of `UIFont(name:size:)`"
    severity: warning
  disable_print:
    name: "print usage"
    regex: "((\\bprint)|(\\bdebugPrint)|(Swift\\.print))\\s*\\("
    message: "Don't use print statements in production code."
    severity: warning
    match_kinds:
      - identifier
overridden_super_call:
  included:
    - "*"
    - "viewSafeAreaInsetsDidChange()"
    - "viewLayoutMarginsDidChange()"
    - "viewDidLayoutSubviews()"
    - "viewWillTransition(to:with:)"
    - "collectionView(_:willDisplay:forItemAt:)"
    - "collectionView(_:didEndDisplaying:forItemAt:)"
attributes:
  attributes_with_arguments_always_on_line_above: false
  always_on_line_above:
    ["@MainActor", "@WidgetBundleBuilder", "@available", "@AppStorage"]

```
</details>

Ueeek:
resolve #5883 
Skip TipKit RuleMacro in EmptyCountRule

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,10 @@
 
 #### Bug Fixes
 
+* Ignore TipKit Rule Macro in `empty_count` rule.  
+  [Ueeek](https://github.com/Ueeek)
+  [#5883](https://github.com/realm/SwiftLint/issues/5883)
+
 * Ignore super calls with trailing closures in `unneeded_override` rule.  
   [SimplyDanny](https://github.com/SimplyDanny)
   [#5886](ttps://github.com/realm/SwiftLint/issues/5886)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,7 @@
 
 #### Bug Fixes
 
-* Ignore TipKit Rule Macro in `empty_count` rule.  
+* Ignore TipKit's' `#Rule` macro in `empty_count` rule.  
   [Ueeek](https://github.com/Ueeek)
   [#5883](https://github.com/realm/SwiftLint/issues/5883)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,7 @@
 
 #### Bug Fixes
 
-* Ignore TipKit's' `#Rule` macro in `empty_count` rule.  
+* Ignore TipKit's `#Rule` macro in `empty_count` rule.  
   [Ueeek](https://github.com/Ueeek)
   [#5883](https://github.com/realm/SwiftLint/issues/5883)
 
--- a/Source/SwiftLintBuiltInRules/Rules/Performance/EmptyCountRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Performance/EmptyCountRule.swift
@@ -20,6 +20,7 @@ struct EmptyCountRule: Rule {
             Example("[Int]().count == 0o07"),
             Example("discount == 0"),
             Example("order.discount == 0"),
+            Example("let rule = #Rule(Tips.Event(id: \"someTips\")) { $0.donations.count == 0 }"),
         ],
         triggeringExamples: [
             Example("[Int]().↓count == 0"),
@@ -32,6 +33,7 @@ struct EmptyCountRule: Rule {
             Example("[Int]().↓count == 0b00"),
             Example("[Int]().↓count == 0o00"),
             Example("↓count == 0"),
+            Example("let predicate =  #Predicate<SwiftDataModel> { $0.list.↓count == 0 }"),
         ],
         corrections: [
             Example("[].↓count == 0"):
@@ -62,6 +64,8 @@ struct EmptyCountRule: Rule {
                 Example("isEmpty && [Int]().isEmpty"),
             Example("[Int]().count != 3 && [Int]().↓count != 0 || ↓count == 0 && [Int]().count > 2"):
                 Example("[Int]().count != 3 && ![Int]().isEmpty || isEmpty && [Int]().count > 2"),
+            Example("let predicate =  #Predicate<SwiftDataModel> { $0.list.↓count == 0 }"):
+                Example("let predicate =  #Predicate<SwiftDataModel> { $0.list.isEmpty }"),
         ]
     )
 }
@@ -76,6 +80,10 @@ private extension EmptyCountRule {
             if let (_, position) = node.countNodeAndPosition(onlyAfterDot: configuration.onlyAfterDot) {
                 violations.append(position)
             }
+        }
+
+        override func visit(_ node: MacroExpansionExprSyntax) -> SyntaxVisitorContinueKind {
+            node.isTipsRuleMacro ? .skipChildren : .visitChildren
         }
     }
 
@@ -104,6 +112,14 @@ private extension EmptyCountRule {
                     }
             }
             return super.visit(node)
+        }
+
+        override func visit(_ node: MacroExpansionExprSyntax) -> ExprSyntax {
+            if node.isTipsRuleMacro {
+                ExprSyntax(Syntax(node).cast(MacroExpansionExprSyntax.self))
+            } else {
+                super.visit(node)
+            }
         }
     }
 }
@@ -134,6 +150,14 @@ private extension TokenSyntax {
         default:
             return nil
         }
+    }
+}
+
+private extension MacroExpansionExprSyntax {
+    var isTipsRuleMacro: Bool {
+        self.macroName.text == "Rule" &&
+        self.arguments.isNotEmpty &&
+        self.trailingClosure != nil
     }
 }
 
--- a/Source/SwiftLintBuiltInRules/Rules/Performance/EmptyCountRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Performance/EmptyCountRule.swift
@@ -83,14 +83,8 @@ struct EmptyCountRule: Rule {
                 Example("[Int]().count != 3 && ![Int]().isEmpty || isEmpty && [Int]().count > 2"),
             Example("#ExampleMacro { $0.list.↓count == 0 }"):
                 Example("#ExampleMacro { $0.list.isEmpty }"),
-            Example("#Rule { $0.donations.↓count == 0 }", excludeFromDocumentation: true):
-                Example("#Rule { $0.donations.isEmpty }"),
-            Example(
-                "#Rule(param1: \"param1\") { return $0.donations.↓count == 0 }",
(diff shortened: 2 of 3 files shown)


Source/SwiftLintBuiltInRules/Rules/Performance/EmptyCountRule.swift
```diff
@@ -137,6 +153,14 @@ private extension TokenSyntax {
     }
 }
 
+private extension MacroExpansionExprSyntax {
+    var isTipsRuleMacro: Bool {
+        self.macroName.text == "Rule" &&
+        self.arguments.isNotEmpty &&
+        self.trailingClosure != nil
```
SimplyDanny, line 160: Let's be even more restrictive here:
```suggestion
        macroName.text == "Rule" &&
        arguments.count == 1 &&
        trailingClosure.map { $0.statements.onlyElement?.item.is(ReturnStmtSyntax.self) == false } ?? false
```

Source/SwiftLintBuiltInRules/Rules/Performance/EmptyCountRule.swift
```diff
@@ -62,6 +64,8 @@ struct EmptyCountRule: Rule {
                 Example("isEmpty && [Int]().isEmpty"),
             Example("[Int]().count != 3 && [Int]().↓count != 0 || ↓count == 0 && [Int]().count > 2"):
                 Example("[Int]().count != 3 && ![Int]().isEmpty || isEmpty && [Int]().count > 2"),
+            Example("let predicate =  #Predicate<SwiftDataModel> { $0.list.↓count == 0 }"):
+                Example("let predicate =  #Predicate<SwiftDataModel> { $0.list.isEmpty }"),
```
SimplyDanny, line 68: Instead of `#Predicate<SwiftDataModel>` you may also use any other (simpler) macro that's not `#Rule`.

SimplyDanny (changes requested):
Thank you! Looks good in general, yet we can be even more restrictive in the check for the `#Rule` macro.

Source/SwiftLintBuiltInRules/Rules/Performance/EmptyCountRule.swift
Ueeek: Thank you for the suggestion!
I used "#ExampleMacro" as an example of Macro, referring NoMagicNumberRule.
Or, is it better to use an actual example?(e.g. #Preview?)

SimplyDanny: `#ExampleMacro` is great. The only reason I was suggesting another name is to keep the examples very simple and reduced down to the significant aspects. Otherwise, the name or its actual existence is not important as long as it's something different than `#Rule` to have a valid test case.

What comes to my mind now is that we should also test against `#Rule` without any arguments or more than one, the macro with a non-trailing closure, a return statement inside or more than one statement, ...

You can exclude these test cases from appearing in the documentation for simplicity by adding the argument `excludeFromDocumentation: true`. Only leave the ones you have now and skip the others as they are not helping users in understanding.

SimplyDanny:
Please rebase your branch to resolve the conflict.

Source/SwiftLintBuiltInRules/Rules/Performance/EmptyCountRule.swift
Ueeek: Nice suggestion, let me add the following tests.
* #Rule without any arguments
* #Rule with more than 1 argument
* #Rule with non-trailing closure
* #Rule with  multiple trailing closure
* #Rule with a closure with the return statement
* #Rule with a closure with multiple statements.

So far, writing the code of the above test cases with the #Rule macro from the Tips Kit results in a compile error.
However, we added the tests for the following reasons. Do I understand correctly? 🤔
* To cover all the conditional branches introduced in this PR.
* To prepare for the possibility that the #Rule macro might support {no/multiple} {arguments/closures} in the future.

SimplyDanny: Yes, we want to make very sure that the rule only ignores the TipKit `#Rule` macro, not any other `#Rule` macro. To know for sure, the rule would need full type information. That's expensive, so we try to encode all the syntactic details we know about the macro in order to reduce false positives. That's not 100% even now, but close enough.

The compiler complains about these cases for TipKit's `#Rule` macro. There could be other `#Rule` macros which are not as strict. We still want to check them.

If you remove `import TipKit` and define your own `#Rule` macro, the compiler will be happy. That's what you need to imagining as a test setup (without specifying that explicitly).

Makes sense?

Source/SwiftLintBuiltInRules/Rules/Performance/EmptyCountRule.swift
```diff
@@ -62,6 +81,40 @@ struct EmptyCountRule: Rule {
...
+                    doSomething()
+                    return $0.donations.↓count == 0
+                }
+            """, excludeFromDocumentation: true):
+                Example("""
+                #Rule(param1: "param1") {
+                    doSomething()
+                    return $0.donations.isEmpty
+                }
+            """),
```
SimplyDanny, line 117: To verify the correction mechanism, one test case is enough. We don't need to repeat all of them here again.

Source/SwiftLintBuiltInRules/Rules/Performance/EmptyCountRule.swift
Ueeek: Yes! Thank you for explain it! I'm learning lot from you. 😄

SimplyDanny (approved):
Two more nits. Then this is good to go. 👍

CHANGELOG.md
```diff
@@ -58,6 +58,10 @@
 
 #### Bug Fixes
 
+* Ignore TipKit's' `#Rule` macro in `empty_count` rule.
```
SimplyDanny, line 61:
```suggestion
* Ignore TipKit's `#Rule` macro in `empty_count` rule.  
```

PR #5961, opened 2025-01-14
[bazel] Fix CryptoSwift sha

keith:
Missed in https://github.com/realm/SwiftLint/commit/4189010afb8bd04c9d2e5f9a38fe44808e6727ad

### DIFF
--- a/bazel/repos.bzl
+++ b/bazel/repos.bzl
@@ -58,7 +58,7 @@ def swiftlint_repos(bzlmod = False):
 
     http_archive(
         name = "com_github_krzyzanowskim_cryptoswift",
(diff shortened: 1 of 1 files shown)


SimplyDanny (approved):
How can we make sure this doesn't happen again?

The build in BCR actually prints `--enable_bzlmod` read from `.bazelrc`. So `if not bzlmod` branch shouldn't be executed. No?

There will be a 0.58.1 shortly. We can just close the PR in BCR. I thought deleting the branch would close the PR. Obviously, that's not the case ...

keith:
looks like bazel CI might be macOS only? and this is a linux only dependency

SimplyDanny:
A new CI job running Bazel on Linux got merged with #5963, hopefully avoiding such an inconsistency from now on.

PR #5967, opened 2025-01-15
Parallelize SwiftLintFile construction, for a modest speedup

OneSadCookie:
`SwiftLintFile` is relatively expensive to construct; doing so in parallel provides a modest improvement to startup overhead.

### DIFF
--- a/Source/SwiftLintFramework/Configuration/Configuration+LintableFiles.swift
+++ b/Source/SwiftLintFramework/Configuration/Configuration+LintableFiles.swift
@@ -20,7 +20,9 @@ extension Configuration {
                               forceExclude: Bool,
                               excludeBy: ExcludeBy) -> [SwiftLintFile] {
         lintablePaths(inPath: path, forceExclude: forceExclude, excludeBy: excludeBy)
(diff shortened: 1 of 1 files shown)


SimplyDanny (changes requested):
Due to the `pathDeferringReading` initializer I don't think parallelizing has a relevance at this place. The content will be read upon request later when the file is first linted by any rule. Linting runs in parallel anyway. 

We also don't see any improvements in performance while scanning OSS repositories (comment above).

So we'd first need some kind of proof that shows the benefit of going parallel at this location.

SimplyDanny:
> I _am_ concerned though, that the benchmarks are only for _lint_ duration, not for _startup_ duration. Due to the efficacy of the cache, most swiftlint runs don't actually care nearly as much about lint duration as they do about startup overhead. My major interest in contribution is for startup overhead, which is inevitably a very small fraction of the benchmarks that are here.

The performance checks run with `--no-cache`, so we compare two independent runs, apart from caching performed by the file system or operating system. If there was a difference, we should see it in the comparison though.

At least we also don't see a negative effect of parallelization overhead. Therefore, I'd be open to accept the change.

SimplyDanny (approved):

Issue #5947, opened 2025-01-10
Test for emptiness is very slow


KeithBauerANZ:
### New Issue Checklist

- [x] I've Updated SwiftLint to the latest version.
- [x] I've searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues).

### Bug Description

I work on a project with a large number (~12000) of swift files. SwiftLint's cache does a good job of avoiding linting them when they haven't changed, which means that vast majority of the time we spend waiting for swiftlint is in its "setup costs". One of those costs is in determining whether to avoid linting the file because it is empty. This is done by this function at the bottom of `Linter.swift`:

```swift
private extension SwiftLintFile {
    var isEmpty: Bool {
        contents.isEmpty || contents == "\n"
    }
}
```

Because this accesses the `contents` property, this forces *reading* all 12000 swift files into memory, just to know if they're empty or not. This costs several seconds of execution time. That doesn't matter if they're then going to be linted anyway, but it *does* matter when they're unchanged from the last run and will skip linting.

I intended to contribute a fix for this, but it's not at all obvious to me what a good approach would be — this sits right at the intersection of SwiftLint and SourceKitten. Perhaps we should gather file sizes during the directory traversal, store them somewhere, and only read the file if the size == 1?

### Environment

SwiftLint version (run swiftlint version to be sure)
0.57.1

Xcode version (run xcodebuild -version to be sure)
Xcode 16.2
Build version 16C5032a

Installation method used (Homebrew, CocoaPods, building from source, etc)
Downloaded from release tag

Configuration file:
n/a

SimplyDanny:
After a brief glimpse on the code, it looks like the `isEmpty` check in `Linter.init` can just be removed. Without the filter, all rules would be collected. But before they do anything, there is the other check for `isEmpty`. It even seems cleaner to let the rule decide whether it wants to do things with empty files or skip them. Even better, this already happens in a parallel context.

Looks like we could give this a try.

OneSadCookie:
Unfortunately, that results in changed user-facing behavior:

> Test Case '-[IntegrationTests.IntegrationTests testSwiftLintAutoCorrects]' started.
/Users/....../SwiftLint/Source/SourceKittenFrameworkWrapper/Empty.swift:1: error: -[IntegrationTests.IntegrationTests testSwiftLintAutoCorrects] : failed - Files should have a single trailing newline
[...]

PR #5973, opened 2025-01-20
avoid pre-reading files

OneSadCookie:
Per #5947, files are being read unnecessarily when they have been previously linted and appear in the cache.

To avoid this,
* remove the test for file emptiness from `Linter.init`
* extract the redundant check for this in `Rule.lint` to a helper, which simplifies `Rule.lint` and allows it to return non-optional
* check the helper before calling either `Rule.lint` or `Rule.correct`.

### DIFF
--- a/Source/SwiftLintFramework/Models/Linter.swift
+++ b/Source/SwiftLintFramework/Models/Linter.swift
@@ -66,29 +66,33 @@ private extension Rule {
         return superfluousDisableCommandViolations
     }
 
+    func shouldRun(onFile file: SwiftLintFile) -> Bool {
+        // We shouldn't lint if the current Swift version is not supported by the rule
+        guard SwiftVersion.current >= Self.description.minSwiftVersion else {
+            return false
+        }
+
+        // Empty files shouldn't trigger violations if `shouldLintEmptyFiles` is `false`
+        if file.isEmpty, !shouldLintEmptyFiles {
+            return false
+        }
+
+        if !(self is any SourceKitFreeRule) && file.sourcekitdFailed {
+            warnSourceKitFailedOnce()
+            return false
+        }
+
+        return true
+    }
+
     // As we need the configuration to get custom identifiers.
-    // swiftlint:disable:next function_parameter_count function_body_length
+    // swiftlint:disable:next function_parameter_count
     func lint(file: SwiftLintFile,
               regions: [Region],
               benchmark: Bool,
               storage: RuleStorage,
               superfluousDisableCommandRule: SuperfluousDisableCommandRule?,
-              compilerArguments: [String]) -> LintResult? {
-        // We shouldn't lint if the current Swift version is not supported by the rule
-        guard SwiftVersion.current >= Self.description.minSwiftVersion else {
-            return nil
-        }
-
-        // Empty files shouldn't trigger violations if `shouldLintEmptyFiles` is `false`
-        if file.isEmpty, !shouldLintEmptyFiles {
-            return nil
-        }
-
-        if !(self is any SourceKitFreeRule) && file.sourcekitdFailed {
-            warnSourceKitFailedOnce()
-            return nil
-        }
-
+              compilerArguments: [String]) -> LintResult {
         let ruleID = Self.identifier
 
         let violations: [StyleViolation]
@@ -226,12 +230,7 @@ public struct Linter {
         self.configuration = configuration
         self.compilerArguments = compilerArguments
 
-        let fileIsEmpty = file.isEmpty
         let rules = configuration.rules.filter { rule in
-            if fileIsEmpty, !rule.shouldLintEmptyFiles {
-                return false
-            }
-
             if compilerArguments.isEmpty {
                 return !(rule is any AnalyzerRule)
             }
@@ -307,11 +306,15 @@ public struct CollectedLinter {
         let superfluousDisableCommandRule = rules.first(where: {
             $0 is SuperfluousDisableCommandRule
         }) as? SuperfluousDisableCommandRule
-        let validationResults = rules.parallelCompactMap {
-            $0.lint(file: file, regions: regions, benchmark: benchmark,
-                    storage: storage,
-                    superfluousDisableCommandRule: superfluousDisableCommandRule,
-                    compilerArguments: compilerArguments)
+        let validationResults: [LintResult] = rules.parallelCompactMap {
+            guard $0.shouldRun(onFile: file) else {
+                return nil
+            }
+
+            return $0.lint(file: file, regions: regions, benchmark: benchmark,
+                           storage: storage,
+                           superfluousDisableCommandRule: superfluousDisableCommandRule,
+                           compilerArguments: compilerArguments)
         }
         let undefinedSuperfluousCommandViolations = self.undefinedSuperfluousCommandViolations(
             regions: regions, configuration: configuration,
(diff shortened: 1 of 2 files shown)


Source/SwiftLintFramework/Models/Linter.swift
```diff
@@ -378,7 +381,13 @@ public struct CollectedLinter {
         }
 
         var corrections = [Correction]()
-        for rule in rules.compactMap({ $0 as? any CorrectableRule }) {
+        for rule in rules.compactMap({ (rule: any Rule) -> (any CorrectableRule)? in
```
SimplyDanny, line 384: Stylistically, I would prefer a separate `if` statement in the `for` block to check for `shouldRun`. Or you can make use of a `where` clause.

We could even check for `CorrectableRule` in the `for` block to avoid creating a new list and iterating twice.

SimplyDanny (changes requested):
Basically looks good to me!

Would be great to have a test for linting and correcting empty files with (mock) rules that do/don't support this.

OneSadCookie:
I've tried to write those tests, but pretty low confidence that that's "the right way"; pointers appreciated.

I tried to see if there was a way to write an integration test that the files' contents aren't read when the cache is populated, but short of modifying SourceKitten to report contents reads, or magical unix shenanigans (I think you can somehow get a pipe at a path so it gets opened like a regular file, maybe you could tell if data was consumed on the pipe?), I don't see how. Definitely open to ideas here, as regressions will be easy.

Tests/FrameworkTests/EmptyFileTests.swift
```diff
@@ -0,0 +1,75 @@
...
+            Correction(ruleDescription: Self.description, location: Location(file: file.path))
+        ]
+    }
+}
+
+final class EmptyFileTests: SwiftLintTestCase {
+    func testShouldLintEmptyFileRespectedDuringLint() {
+        let ruleList = RuleList(rules: RuleWithLintEmptyFilesMock<False>.self,
+                                RuleWithLintEmptyFilesMock<True>.self)
+        let configuration = try! Configuration(dict: [:], ruleList: ruleList) // swiftlint:disable:this force_try
```
SimplyDanny, line 53: The test can just throw. No need for force-try.

Tests/FrameworkTests/EmptyFileTests.swift
```diff
@@ -0,0 +1,75 @@
...
+        let file = SwiftLintFile(contents: "")
+        let linter = Linter(file: file, configuration: configuration)
+        let ruleStorage = RuleStorage()
+        let collectedLinter = linter.collect(into: ruleStorage)
+        let styleViolations = collectedLinter.styleViolations(using: ruleStorage)
+        XCTAssertEqual(styleViolations.count, 1)
+        XCTAssertEqual(styleViolations.first?.ruleIdentifier, "lint_empty_files_mock<true>")
+    }
+
+    func testShouldLintEmptyFileRespectedDuringCorrect() {
```
SimplyDanny, line 63: Would be fine to merge both tests to have the setup only once.

Tests/FrameworkTests/EmptyFileTests.swift
```diff
@@ -0,0 +1,75 @@
...
+
+private struct True: BooleanConstant {
+    static var boolValue: Bool { true }
+}
+
+private struct False: BooleanConstant {
+    static var boolValue: Bool { false }
+}
+
+private struct RuleWithLintEmptyFilesMock<ShouldLintEmptyFiles: BooleanConstant>: CorrectableRule {
```
SimplyDanny, line 16: A short name with this `ShouldLintEmptyFiles` parameter should reveal the rules purpose sufficiently well:

```suggestion
private struct RuleMock<ShouldLintEmptyFiles: BooleanConstant>: CorrectableRule {
```

Tests/FrameworkTests/EmptyFileTests.swift
```diff
@@ -0,0 +1,75 @@
+import SwiftLintFramework
+import XCTest
+
+private protocol BooleanConstant {
```
SimplyDanny, line 4: I'm not really a big fan of this, but looks like there is no other way due to the rules being statically initialized.

Tests/FrameworkTests/EmptyFileTests.swift
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    func correct(file: SwiftLintFile) -> [Correction] {
+        [
+            Correction(ruleDescription: Self.description, location: Location(file: file.path))
+        ]
+    }
+}
+
+final class EmptyFileTests: SwiftLintTestCase {
```
SimplyDanny, line 49: Let's have the test class first in the file.

Tests/FrameworkTests/EmptyFileTests.swift
```diff
@@ -0,0 +1,75 @@
...
+                        name: "",
+                        description: "",
+                        kind: .style,
+                        deprecatedAliases: ["mock"])
+    }
+
+    init() { /* conformance for test */ }
+    init(configuration _: Any) throws {
+        self.init()
+    }
```
SimplyDanny, line 30: Both can be removed.

SimplyDanny (changes requested):
> I've tried to write those tests, but pretty low confidence that that's "the right way"; pointers appreciated.

Looks okay. It shows that that `shouldLintEmptyFiles` property works at least.

> I tried to see if there was a way to write an integration test that the files' contents aren't read when the cache is populated, but short of modifying SourceKitten to report contents reads, or magical unix shenanigans (I think you can somehow get a pipe at a path so it gets opened like a regular file, maybe you could tell if data was consumed on the pipe?), I don't see how. Definitely open to ideas here, as regressions will be easy.

I don't know of a way to do this either.

Tests/FrameworkTests/EmptyFileTests.swift
OneSadCookie: Unfortunately, the caches get cleared after `styleViolations`, which nukes the `contents:` of the `SwiftLintFile` and sets it back to trying to read from the (nonexistent) path. So I don't know how to make this work.

SimplyDanny (approved):

Tests/FrameworkTests/EmptyFileTests.swift
```diff
@@ -0,0 +1,68 @@
+import SwiftLintFramework
+import XCTest
+
+final class EmptyFileTests: SwiftLintTestCase {
```
SimplyDanny, line 4:
```suggestion

// swiftlint:disable:next balanced_xctest_lifecycle
final class EmptyFileTests: SwiftLintTestCase {
```

Tests/FrameworkTests/EmptyFileTests.swift
```diff
@@ -1,36 +1,63 @@
...
+        let file = SwiftLintFile(contents: "")
+        let linter = Linter(file: file, configuration: configuration)
+        ruleStorage = RuleStorage()
+        collectedLinter = linter.collect(into: ruleStorage)
+    }
+
+    override func tearDownWithError() throws {
+        collectedLinter = nil
+        ruleStorage = nil
+    }
```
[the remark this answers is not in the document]
OneSadCookie: ah, I figured having a useless teardown was better than disabling the lint rule saying they have to balance; thanks

OneSadCookie:
Thanks for all your help getting this over the line, and attention to detail for the project. I really appreciate it!

Tests/FrameworkTests/EmptyFileTests.swift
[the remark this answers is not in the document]
SimplyDanny: Yeah, this is a matter of taste. Sorry for nitpicking. 😬

SimplyDanny:
> Thanks for all your help getting this over the line, and attention to detail for the project. I really appreciate it!

You are the one with an eye for the performance details. Keep it up!

PR #5976, opened 2025-01-21
Fix crash when disable command is preceded by unicode character

### LINKED ISSUE #5945: Crash with swiftlint directive after emoji with skin tone
### New Issue Checklist

### Bug Description

This crashes swiftlint:

```swift
"🤵🏼‍♀️" // swiftlint:disable:this nonexistent_rule
```

(whether the rule exists or not isn't relevant, only the directive following the emoji on the same line)

The crash is here:

https://github.com/realm/SwiftLint/blob/main/Source/SwiftLintCore/Extensions/String%2BSwiftLint.swift#L31

(My swiftlint doesn't have debug info, so the backtrace is useless, but that's the assertion message)

Just without any config at all:

```bash
$ swiftlint test.swift
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)
0.57.1

* Xcode version (run `xcodebuild -version` to be sure)
Xcode 16.2
Build version 16C5032a

* Installation method used (Homebrew, CocoaPods, building from source, etc)
Downloaded from release tag

* Configuration file:
n/a

SimplyDanny:
Fixes #5945.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,9 @@
 
 * Fix issue referencing the Tests package from another Bazel workspace.  
   [jszumski](https://github.com/jszumski)
+* Fix crash when a disable command is preceded by a unicode character.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+  [#5945](https://github.com/realm/SwiftLint/issues/5945)
 
 * Allow severity of `duplicate_imports` rule to be configurable.  
   [SimplyDanny](https://github.com/SimplyDanny)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 
 * Fix issue referencing the Tests package from another Bazel workspace.  
   [jszumski](https://github.com/jszumski)
+
 * Fix crash when a disable command is preceded by a unicode character.  
   [SimplyDanny](https://github.com/SimplyDanny)
   [#5945](https://github.com/realm/SwiftLint/issues/5945)
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/InvalidSwiftLintCommandRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/InvalidSwiftLintCommandRule.swift
@@ -64,7 +64,7 @@ struct InvalidSwiftLintCommandRule: Rule, SourceKitFreeRule {
         StyleViolation(
             ruleDescription: Self.description,
             severity: configuration.severity,
-            location: Location(file: file.path, line: command.line, character: command.character),
+            location: Location(file: file.path, line: command.line, character: command.range?.lowerBound),
             reason: reason
         )
     }
@@ -72,7 +72,7 @@ struct InvalidSwiftLintCommandRule: Rule, SourceKitFreeRule {
 
 private extension Command {
     func isPrecededByInvalidCharacter(in file: SwiftLintFile) -> Bool {
-        guard line > 0, let character, character > 1, line <= file.lines.count else {
+        guard line > 0, let character = range?.lowerBound, character > 1, line <= file.lines.count else {
             return false
         }
         let line = file.lines[line - 1].content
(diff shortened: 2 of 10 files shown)


Source/SwiftLintBuiltInRules/Rules/Lint/InvalidSwiftLintCommandRule.swift
```diff
@@ -60,35 +62,54 @@ struct InvalidSwiftLintCommandRule: Rule, SourceKitFreeRule {
 
     private func styleViolation(for command: Command, in file: SwiftLintFile, reason: String) -> StyleViolation {
         let character = command.startingCharacterPosition(in: file)
+        let characterOffset = character.flatMap {
```
mildm8nnered, line 65: So I think we could make the `characterOffset` retrieval a function on the `Command` extension, so this reads more nicely here, but having to do all of these line calculations feels a bit off in general.

The reason we have to do this calculations is because `Command`'s `character` property records the character beyond the end of the command. If that instead recorded the start of the command, which is where this rule places the violation, then we could ditch all of this code.

I think that might be possible, but something we could look at separately.

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@
 
 * Fix issue referencing the Tests package from another Bazel workspace.  
   [jszumski](https://github.com/jszumski)
+* Fix crash when a disable command is preceded by a unicode character.
```
mildm8nnered, line 21: Maybe a blank line here?

mildm8nnered (approved):

Source/SwiftLintBuiltInRules/Rules/Lint/InvalidSwiftLintCommandRule.swift
mildm8nnered: Or maybe just record the range of the Command at least both the start and end somehow ...

SimplyDanny: I changed it to record the beginning of the command instead. This indeed simplifies the code in the rule. However, at some point we need to perform the byte and position calculations, unfortunately, which is `CommandVisitor` now.

Going from an absolute byte position to a character doesn't seem to be obvious. That's what `characterPosition(of:)` is doing. In case you have a better idea, I'm open for it.

SimplyDanny: And I've just changed it to a range. We also need to know where the command end, because this is the beginning of a new region.

```swift
// swiftlint:disable all
```

would otherwise not trigger `blanket_disable_command` anymore as it were already disabled at the position where it triggered. 🤯

mildm8nnered: So I had this as a quick hack, storing `start` separately - next would have been to move it into a range - https://github.com/realm/SwiftLint/compare/main...mildm8nnered:mildm8nnered-invalid-swiftlint-command-rule?expand=1, so I think yours is fine.

Source/SwiftLintCore/Models/Command.swift
```diff
@@ -85,21 +84,20 @@ public struct Command: Equatable {
     ///
     /// - parameter actionString: The string in the command's definition describing its action.
     /// - parameter line:         The line in the source file where this command is defined.
-    /// - parameter character:    The character offset within the line in the source file where this command is
-    ///                           defined.
-    public init(actionString: String, line: Int, character: Int) {
+    /// - parameter range:        The range of the command in the line (0-based).
+    public init(actionString: String, line: Int, range: Range<Int>) {
```
mildm8nnered, line 88: I renamed the parameter to `commandString` in my branch, which I think is better - the comment is also wrong here - it's the local `actionString` on line 99/97 that contains the action.

AhmedZaghloul19 (approved):

PR #5977, opened 2025-01-21
Fix issue referencing the Tests package from another Bazel workspace

jszumski:
Allow `BuiltInRulesTests` data to be empty so a workspace can load the Tests package from a release archive.

Otherwise, Bazel users with `--incompatible_disallow_empty_glob` enabled will hit this error because the release archive only contains Swift files and not any test fixtures:

```
Error in glob: glob pattern 'BuiltInRulesTests/Resources/**' didn't match anything, but allow_empty is set to False (the default value of allow_empty can be set with --incompatible_disallow_empty_glob).
```

Note that `--incompatible_disallow_empty_glob` is enabled by default in Bazel 8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,9 @@
 
 ### Bug Fixes
 
-* None.
+* Fix issue referencing the Tests package from another Bazel workspace.  
+  [jszumski](https://github.com/jszumski)
+  [#5977](https://github.com/realm/SwiftLint/issues/5977)
 
 ## 0.58.2: New Year’s Fresh Fold
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,6 @@
 
 * Fix issue referencing the Tests package from another Bazel workspace.  
   [jszumski](https://github.com/jszumski)
-  [#5977](https://github.com/realm/SwiftLint/issues/5977)
 
 ## 0.58.2: New Year’s Fresh Fold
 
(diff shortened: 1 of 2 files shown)


luispadron (approved):

CHANGELOG.md
```diff
@@ -16,7 +16,9 @@
 
 ### Bug Fixes
 
-* None.
+* Fix issue referencing the Tests package from another Bazel workspace.  
+  [jszumski](https://github.com/jszumski)
+  [#5977](https://github.com/realm/SwiftLint/issues/5977)
```
SimplyDanny, line 21: We only link to issues here:

```suggestion
  [jszumski](https://github.com/jszumski)
```

SimplyDanny (approved):
Thanks!

Is this a regression? `BuiltInRulesTests` was newly added in 0.58.0. However, the resources were just moved from `FrameworkTests` and I don't see that `allow_empty = True` was previously specified there.

jszumski:
I believe this worked previously because the old glob was:

```
filegroup(
    name = "SwiftLintFrameworkTestsData",
    srcs = glob(
        ["**/**"],
        # Bazel doesn't support paths with spaces in them
        exclude = ["SwiftLintFrameworkTests/Resources/FileNameNoSpaceRuleFixtures/**"],
    ),
    visibility = ["//visibility:public"],
)
```

which would end up matching some Swift source files in the release distribution and thus wasn't empty.

PR #5981, opened 2025-01-28
Move `--only-rule` option to common arguments

SimplyDanny:


SimplyDanny:
@mildm8nnered: I don't see a reason to have them not in the common part. You?

PR #5983, opened 2025-01-30
File grouping is pretty CPU-heavy; parallelize it.

OneSadCookie:
You've heard the drill by now — I care deeply about startup time, and this particular thing SwiftLint does at startup is pretty slow.

`groupFiles` is doing a fair amount of CPU work on a single thread. For my project with 12k files and at least 10 different configs, it takes ~1.6s on my M1 MBP.

I've copied patterns from other functions in Array+SwiftLint, and not unit-tested them specifically because none of the other functions here are unit-tested, but maybe that's an oversight?

### DIFF
--- a/Source/SwiftLintCore/Extensions/Array+SwiftLint.swift
+++ b/Source/SwiftLintCore/Extensions/Array+SwiftLint.swift
@@ -84,7 +84,7 @@ public extension Array {
     /// `concurrentPerform`.
     ///
     /// - parameter transform: The transformation function to extract an element to its group key,
-    ///                        or exclude the element..
+    ///                        or exclude the element.
     ///
     /// - returns: The elements grouped by applying the specified transformation.
     func parallelFilterGroup<U: Hashable & Sendable>(by transform: @Sendable (Element) -> U?) ->
--- a/Source/SwiftLintCore/Extensions/Array+SwiftLint.swift
+++ b/Source/SwiftLintCore/Extensions/Array+SwiftLint.swift
@@ -66,7 +66,7 @@ public extension Array {
     /// Elements for which the `transform` returns a `nil` key are removed.
     ///
     /// - parameter transform: The transformation function to extract an element to its group key,
-    ///                        or exclude the element..
+    ///                        or exclude the element.
     ///
     /// - returns: The elements grouped by applying the specified transformation.
     func filterGroup<U: Hashable & Sendable>(by transform: (Element) -> U?) ->
--- a/Source/SwiftLintCore/Extensions/Array+SwiftLint.swift
+++ b/Source/SwiftLintCore/Extensions/Array+SwiftLint.swift
@@ -62,6 +62,46 @@ public extension Array {
         Dictionary(grouping: self, by: { transform($0) })
     }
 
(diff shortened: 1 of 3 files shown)


OneSadCookie:
Well! Lucky the test suite spotted those 😭

OneSadCookie:
`Configuration.RulesWrapper`'s `private static var isOptInRuleCache: [String: Bool] = [:]` isn't thread-safe. The crash only appears with `--enable-all-rules` for some reason.

OneSadCookie:
I don't think the `testInvalidMaxNumberOfSingleLineParametersWithSingleLineEnabled` failure has to do with my changes; I can't see any interaction. `Issue.captureConsole` looks highly suspect; that's an unlocked mutable global, and a `Task.yield()` which is seldom a good sign. The test passes thousands of times locally, though...

Source/SwiftLintCore/Extensions/Array+SwiftLint.swift
```diff
@@ -62,6 +62,46 @@ public extension Array {
         Dictionary(grouping: self, by: { transform($0) })
     }
 
+    /// Group the elements in this array into a dictionary, keyed by applying the specified `transform`.
+    /// Elements for which the `transform` returns a `nil` key are removed.
+    ///
+    /// - parameter transform: The transformation function to extract an element to its group key,
+    ///                        or exclude the element..
```
SimplyDanny, line 69:
```suggestion
    ///                        or exclude the element.
```

Same in the other method.

SimplyDanny:
> `Configuration.RulesWrapper`'s `private static var isOptInRuleCache: [String: Bool] = [:]` isn't thread-safe. The crash only appears with `--enable-all-rules` for some reason.

Yeah, `SwiftLintFramework` needs quite some refactoring to become thread-safe in the sense of Swift 6. Until then, manual checks and locking are unavoidable.

SimplyDanny:
> I don't think the `testInvalidMaxNumberOfSingleLineParametersWithSingleLineEnabled` failure has to do with my changes; I can't see any interaction. `Issue.captureConsole` looks highly suspect; that's an unlocked mutable global, and a `Task.yield()` which is seldom a good sign. The test passes thousands of times locally, though...

The current implementation is somewhat fragile, even though it works most of the time.

I try to improve it with #5986. Let me know what you think. For now, re-triggering the build made it pass.

SimplyDanny (approved):
Only one nit ... Otherwise this looks good and the benchmarks speak for themselves. 👍

Source/SwiftLintCore/Extensions/Array+SwiftLint.swift
```diff
@@ -62,6 +62,46 @@ public extension Array {
...
+            }
+        }
+        return result
+    }
+
+    /// Same as `filterGroup`, but spreads the work in the `transform` block in parallel using GCD's
+    /// `concurrentPerform`.
+    ///
+    /// - parameter transform: The transformation function to extract an element to its group key,
+    ///                        or exclude the element..
```
SimplyDanny, line 87:
```suggestion
    ///                        or exclude the element.
```

PR #5986, opened 2025-02-01
Ensure that content is complete using a queue

SimplyDanny:
The previous solution worked most of the time, but every now and then it failed due to laxly defined order.


OneSadCookie:
I'm not 100% on what's going on here, but at least this is concerning:
```swift
        actor Console {
            static var content = ""
        }
```

This is a mutable global without a lock. Placing it in an `actor` is doing nothing (except possibly confusing the reader; it may as well be an enum if all you want is the ability to declare a static at function scope).

The test probably passes so long as tests don't execute concurrently, but when they do, all bets are off as to which test receives what output. Adopting Swift Testing will likely make the problem more visible?

But I also don't think you need a queue to fix the problem; simply locking around access to `content` should make it just as reliable as the queue. (There's still a potential issue if the `runner` block spawns asynchronous work that prints issues from multiple tasks, that the order of prints won't be deterministic, but that doesn't seem to be the case currently.)

It seems the project doesn't have an equivalent of `OSAllocatedUnfairLock` (and can't use `Synchronization.Mutex`) yet; it might be time to add one?

SimplyDanny:
I got rid of the global variable. My assumption was that the actor would protect it but should have noticed that the compiler didn't require `await` when it was modified.

The local variable that's now used would be isolated to `@MainActor`, so that there are no races. Order can still be indeterministic, but that's okay.

The queue is needed to ensure all print operations to have finished before `captureConsole` returns the result. I'm not sure this can be guaranteed otherwise as `print` is non-isolated and can be called from everywhere anytime inside of `runner`.

I experimented with Swift Testing. However, adopting it would mean excluding people who are still on Xcode <16. That's perhaps a bit too early.

mattmassicotte:
That's a good catch on the actor static. Those are known unsafe https://github.com/swiftlang/swift/issues/78435

There might also be an issue with changing priority in that `Task`. It might also be possible to preserve ordering with a simple `AsyncStream`, if you don't want to take on another dependency.

SimplyDanny:
> It might also be possible to preserve ordering with a simple `AsyncStream`, if you don't want to take on another dependency.

Oh, that's great! `AsyncStream` works well and makes the code a bit easier, too.

Issue #5829, opened 2024-10-18
Semicolons: Avoid writing multiple statements on a single line separated with semicolons.


SimplyDanny:

val25-dev:
Hello @SimplyDanny,
Could I work on it? 😊

SimplyDanny:
> Hello @SimplyDanny,
> 
> Could I work on it? 😊

Sure. Please go ahead.

AhmedZaghloul19:
Hi @SimplyDanny 😊  it seems like it's not done yet, so I already implemented the code and want to create the PR for that, can you please assign it to me if it's not done yet?

Thanks in advance

PR #5990, opened 2025-02-05
Add SwiftLint rule to prevent multiple variable declarations on one line

AhmedZaghloul19:
for the issue: 
**Rule Request: Avoid multiple statements on a single line separated with semicolons #5829**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@
 
 ### Enhancements
 
-* None.
+* Add new `multiple_variable_declaration` opt-in rule that triggers when the there's any declaration of variable and/or constants at the same line.  
+  [AhmadZaghloul](https://github.com/AhmedZaghloul19)
 
 ### Bug Fixes
 
--- a/Source/SwiftLintBuiltInRules/Models/BuiltInRules.swift
+++ b/Source/SwiftLintBuiltInRules/Models/BuiltInRules.swift
@@ -245,4 +245,5 @@ public let builtInRules: [any Rule.Type] = [
     XCTFailMessageRule.self,
     XCTSpecificMatcherRule.self,
     YodaConditionRule.self,
+    MultipleVariableDeclarationRule.self,
 ]
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/MultipleVariableDeclarationRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/MultipleVariableDeclarationRule.swift
@@ -1,0 +1,41 @@
+import SwiftLintCore
+import SwiftSyntax
+
+@SwiftSyntaxRule(optIn: true)
+struct MultipleVariableDeclarationRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "multiple_variable_declaration",
+        name: "Multiple Variable Declaration",
+        description: "Variables should not be declared on the same line",
+        kind: .style,
+        nonTriggeringExamples: [
+            Example("let a = 1\nlet b = 2"),
+            Example("var x = 10\nvar y = 20"),
+        ],
+        triggeringExamples: [
+            Example("let a = 1; let b = 2"),
+            Example("var x = 10; var y = 20"),
+        ]
+    )
+
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        private var lastVariableLine: Int?
+
+        override func visitPost(_ node: CodeBlockItemSyntax) {
+            let converter = file.locationConverter
+            let currentLocation = converter
+                .location(for: node.positionAfterSkippingLeadingTrivia)
+            let currentLine = currentLocation.line
+            guard currentLine > 0 else {
+                return
+            }
+
+            if let lastLine = lastVariableLine, lastLine == currentLine {
+                violations.append(node.positionAfterSkippingLeadingTrivia)
+            }
+            lastVariableLine = currentLine
+        }
+    }
+}
--- a/Tests/BuiltInRulesTests/MultipleVariableDeclarationRuleTests.swift
+++ b/Tests/BuiltInRulesTests/MultipleVariableDeclarationRuleTests.swift
@@ -1,0 +1,49 @@
+import SwiftLintCore
+import SwiftSyntax
+import XCTest
+
+@testable import SwiftLintBuiltInRules
+
+final class MultipleVariableDeclarationRuleTests: XCTestCase {
+    func testNonTriggeringExamples() {
+        // Test cases where the rule should NOT trigger
+        let nonTriggeringExamples = [
+            Example("let a = 1\nlet b = 2"), // valid: variables on separate lines
+            Example("var x = 10\nvar y = 20"), // valid: variables on separate lines
+            Example("""
+                    var a = 1
+                    let b = 2
+                    """),
+        ]
+        let triggeringExamples = [
+            Example("let a = 1; let b = 2"), // invalid: variables on the same line
+            Example("var x = 10; var y = 20"), // invalid: variables on the same line
+            // invalid: multiple declarations on the same line should trigger multiple violations
+            Example("let a = 1; let b = 2; let c = 3 "),
+            Example("""
+                    let a = 1; var b = 2
+                    """),
+            Example("""
+                    func testFunction() {
+                        let a = 1; let b = 2
+                        var x = 10; var y = 20
+                    }
+                    """),
+        ]
+
+        nonTriggeringExamples.forEach { example in
+            XCTAssertFalse(violatesRule(example.code))
+        }
+
+        triggeringExamples.forEach { example in
+            XCTAssertTrue(violatesRule(example.code))
+        }
+    }
+
+    // MARK: - Helper method to check for rule violations
+    private func violatesRule(_ code: String) -> Bool {
+        let file = SwiftLintFile(contents: code)
+        let rule = MultipleVariableDeclarationRule()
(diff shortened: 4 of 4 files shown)


SimplyDanny:
Thank you for the contribution! Here are some general remarks:

* The rule is supposed to trigger on multiple **statements** on a single line. That is, it is not only about variable declarations, but all kinds of statements like `let a = 1; return a` or `if b { return }; let a = 1`.
* Refer to `TrailingSemicolonRule` to get an idea of how this could be implemented.
* Run `make sourcery` to get the order of tests and the rule list in the right order.
* 19980 warnings are surely too many. I expect a handful or warnings or even none.
* The explicit test `MultipleVariableDeclarationRuleTests` isn't required. Rather add more triggering and non-triggering examples to the rule's description.

Naveen-C-Ramachandrappa (approved):

AhmedZaghloul19:
Thank you @SimplyDanny, yes it was unnecessary to have unit tests for it, I think now it'll work properly.
and about this pipeline error, I really don't know why!, if you can help me fixing it it would be great.
it was successful before solving the merge conflict

SimplyDanny:
> and about this pipeline error, I really don't know why!, if you can help me fixing it it would be great.

Danger doesn't like merge commits, that's all. You need to rebase your branch onto `main` instead.

AhmedZaghloul19:
@SimplyDanny I'm trying to rebase on main but I'm afraid of messing everything up actually, as there is conflicts while rebasing and I'm afraid of not including your changes 😬

SimplyDanny:
> @SimplyDanny I'm trying to rebase on main but I'm afraid of messing everything up actually, as there is conflicts while rebasing and I'm afraid of not including your changes 😬

You definitely messed something up. You can try to rebase again and throw away all commits that are not yours or duplicates of commits you have created previously. There shouldn't be any conflicts apart from maybe CHANGELOG.md, which would be easy to resolve.

AhmedZaghloul19:
> > @SimplyDanny I'm trying to rebase on main but I'm afraid of messing everything up actually, as there is conflicts while rebasing and I'm afraid of not including your changes 😬
> 
> You definitely messed something up. You can try to rebase again and throw away all commits that are not yours or duplicates of commits you have created previously. There shouldn't be any conflicts apart from maybe CHANGELOG.md, which would be easy to resolve.

@SimplyDanny the git now is fixed and everything is ready 👍🏼

SimplyDanny:
> @SimplyDanny the git now is fixed and everything is ready 👍🏼 

Commits look okay now. But there's still a few things to consider. Did you read my [other comment](https://github.com/realm/SwiftLint/pull/5990#issuecomment-2640989662) carefully?

AhmedZaghloul19:
@SimplyDanny Now it's ready, but I think it's not good to merge with the 436 warnings which is the violations of the new rule in the project. 
do you think we have to fix it in this PR or create another issue for creating that and I'll do it also ?

Source/SwiftLintBuiltInRules/Rules/Idiomatic/MultipleStatementsDeclarationRule.swift
```diff
@@ -0,0 +1,92 @@
+import SwiftSyntax
+
+@SwiftSyntaxRule(explicitRewriter: true, optIn: true)
+struct MultipleStatementsDeclarationRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "multiple_statements_declaration",
+        name: "Multiple Statements Declaration",
+        description: "Statements should not be on the same line",
```
SimplyDanny, line 10: We can keep the name a bit shorter:
```suggestion
struct MultipleStatementsRule: Rule {
    var configuration = SeverityConfiguration<Self>(.warning)

    static let description = RuleDescription(
        identifier: "multiple_statements",
        name: "Multiple Statements",
        description: "Every statement should be on its own line",
```
Please also update all other places mentioning the name.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 
 ### Enhancements
 
+* Add `multiple_statements_declaration` opt-in rule 
+  that triggers when the there's multiple statements at the same line.
```
SimplyDanny, line 16:
```suggestion
  that triggers when there are multiple statements on the same line.  
```

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 
 ### Enhancements
 
+* Add `multiple_statements_declaration` opt-in rule
```
SimplyDanny, line 15:
```suggestion
* Add `multiple_statements` opt-in rule
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/MultipleStatementsDeclarationRule.swift
```diff
@@ -0,0 +1,92 @@
...
+        kind: .idiomatic,
+        nonTriggeringExamples: [
+            Example("let a = 1;\nlet b = 2;"),
```
SimplyDanny, line 13: Please use multi-line string blocks instead of explicit newline characters `\n`.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/MultipleStatementsDeclarationRule.swift
```diff
@@ -0,0 +1,92 @@
...
+            )
+            return super.visit(newNode)
+        }
+    }
+}
+
+private extension TokenSyntax {
+    var isThereStatementAfterSemicolon: Bool {
+        guard tokenKind == .semicolon,
+            !trailingTrivia.isEmpty else { return false }
```
SimplyDanny, line 65: I don't comprehend this condition. Why is no trivia after the semicolon allowed? I think an example like `let a = 0;let b = 0` should trigger as well.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/MultipleStatementsDeclarationRule.swift
```diff
@@ -0,0 +1,92 @@
...
+
+        if let nextToken = nextToken(viewMode: .sourceAccurate),
+           isFollowedOnlyByWhitespaceOrNewline {
+            return nextToken.leadingTrivia.containsNewlines() == false
```
SimplyDanny, line 69: I think what you actually want to know is whether the next token is on the same line as the semicolon. You can use `locationConverter` in every visitor and rewriter. It can tell you on which line a certain token is located. If you compare the line of the semicolon with the line of the next token, you know whether to trigger a violation or not.

Source/SwiftLintFramework/Configuration/Configuration+Remote.swift
```diff
@@ -86,7 +86,10 @@ internal extension Configuration.FileGraph.FilePath {
             // Block main thread until timeout is reached / task is done
             while true {
                 if taskDone { break }
-                if Date().timeIntervalSince(startDate) > timeout { task.cancel(); break }
+                if Date().timeIntervalSince(startDate) > timeout {
+                    task.cancel()
+                    break
+                }
```
SimplyDanny, line 92: This doesn't look it got auto-fixed by the rule. The only cases that can be fixed automatically are the ones where the statements are on their own line without leading tokens. We should probably restrict the rewriter to these cases.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/MultipleStatementsDeclarationRule.swift
```diff
@@ -0,0 +1,92 @@
...
+        override func visit(_ node: TokenSyntax) -> TokenSyntax {
+            guard node.isThereStatementAfterSemicolon else {
+                return super.visit(node)
+            }
+
+            correctionPositions.append(node.positionAfterSkippingLeadingTrivia)
+            let newNode = TokenSyntax(
+                .unknown(""),
+                leadingTrivia: node.leadingTrivia,
+                trailingTrivia: .newlines(1),
```
SimplyDanny, line 54: This doesn't indent the following statement. Consider the example
```swift
{
    let a = 0; let b = 0
}
```
which should be corrected to
```swift
{
    let a = 0;
    let b = 0
}
```

SimplyDanny (changes requested):
This looks much better! However, there are a few more remarks from my side ...

Source/SwiftLintBuiltInRules/Rules/Idiomatic/MultipleStatementsDeclarationRule.swift
AhmedZaghloul19: The autocorrect version is 
`{
    let a = 0
    let b = 0
}`
as we're removing the semicolon, as it's swift code styling 
We can keep the semicolon as well

SimplyDanny: Correct. My bad. Except this example wouldn't be formatted like that with the current implementation.

AhmedZaghloul19: @SimplyDanny So, do i have to do anything else for this rule except fixing the conflict?

SimplyDanny: The examples isn't formatted like you suggest. That needs to be fixed.

SimplyDanny:
@AhmedZaghloul19: Your branch is pretty outdated. Before you work on anything else, I highly recommend to rebase it.

Afterwards, please check my other comments.

PR #6009, opened 2025-02-25
Exclude swift-testing macro (`@Test`) in `identifier_name` rules

coffmark:
## Describe your changes
- Exclude swift-testing macro (`@Test`) in `identifier_name` rules.
- User can set `excluded_swift_testing` configuration as you needed.

## Motivation

Normally functions should be implemented starting with a lowercase letter, but when using swift-testing, there're cases where you want to write test cases in the function name that aren't lowercase.

The following example is a Japanese implementation of a function named "This is a test".

This PR, using the `excluded_swift_testing` flag, allows user to prevent errors (or warnings) from occurring in functions with `@Test` macros implemented.

```swift
// before
@Test func これはテストです() { } // swiftlint error (`identifier_name`)  

// after (when `excluded_swift_testing` is true)
@Test func これはテストです() { } // no error

// after (when `excluded_swift_testing` is false)
@Test func これはテストです() { } // swiftlint error (`identifier_name`)  
```

## Issue ticket number and link
None.


SimplyDanny:
Thanks for bringing this up, @coffmark!

I have two remarks:

1. Swift Testing allows you to put a more descriptive name inside of the `@Test` attribute like `@Test("This is a test")` which is exactly there to avoid mangling the method name.
2. If this option is still desired for other reasons, it should also respect XCTest test methods.

Issue #3723, opened 2021-09-22
legacy_objc_type rule triggers on NSNumber in the context of NumberFormatter which requires it


tylerbarnesTrexis:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

legacy_objc_type rule triggers on NSNumber in the context of NumberFormatter

##### Complete output when running SwiftLint, including the stack trace and command used

```zsh
% swiftlint
Linting Swift files in current working directory
Linting 
...
Linting
warning: Legacy Objective-C Reference Type Violation: Prefer Swift value types to bridged Objective-C reference types (legacy_objc_type)
Done linting! Found 1 violation, 0 serious in files.
```

### Environment

* SwiftLint version: 0.44.0
* Installation method used: Homebrew and SPM via Custom Package
* Paste your configuration file:

```yml
only_rules:
  - legacy_objc_type
```

```swift
// This triggers a violation:
NumberFormatter.string(from: Double as NSNumber)
```

_Originally posted by @PH9 in https://github.com/realm/SwiftLint/issues/3555#issuecomment-811875997_

CraigSiemens:
We recently ran into this as well with `NumberFormatter.string(from:)` and `UNMutableNotificationContent.badge` both still expecting a `NSNumber`.

SimplyDanny:
In my opinion, there is nothing we can really do about that. When working with certain frameworks it is simply unavoidable to reference these types in some way. The rule is opt-in and can be disabled by comments in the code in case of (rare) exceptions.

userException:
I got an error for NSNumber violation so changed:
NSNumber.init(value: someValue)
to
NSNumber.(value: someValue)

Worked !

rgoldberg:
@SimplyDanny Could a `ignore` config key accept an array of type names to ignore?

If you want to ignore this rule for `NSNumber`s because your Swift code must use many `NSNumber`s, just configure:

```
legacy_objc_type:
  ignore: [NSNumber]
```

The rule will still trigger for all other bridged Objective-C types.

SimplyDanny:
> @SimplyDanny Could an `ignore` or `exclude` config key accept an array of type names to ignore?

That's still not a fix but at least a way to reduce the number of false positives. I'd be supportive if someone wants to add such an option.

kapitoshka438:
[...]

PR #6012, opened 2025-03-05
Add new allowed_types option to legacy_objc_type rule

kapitoshka438:
This PR adds a new option to allow certain types, particularly when they are required by third-party dependencies.
https://github.com/realm/SwiftLint/issues/3723#issuecomment-2567548466


Tests/BuiltInRulesTests/LegacyObjcTypeRuleTests.swift
```diff
@@ -0,0 +1,20 @@
+@testable import SwiftLintBuiltInRules
+import TestHelpers
+import XCTest
+
+final class LegacyObjcTypeRuleTests: SwiftLintTestCase {
```
SimplyDanny, line 5: We prefer test cases to be part of the triggering and/or non-triggering example in the rule's description if possible. `Example` allows you to add a `configuration` argument if options influence the outcome. In other words, this dedicated test class isn't needed.

The advantage is that these examples also serve as documentation.

SimplyDanny (changes requested):
Looks good to me! There is only a small remark about the test class which can be improved ...

Tests/BuiltInRulesTests/LegacyObjcTypeRuleTests.swift
kapitoshka438: Thank you for review @SimplyDanny . I've made changes.

SimplyDanny (approved):

SimplyDanny:
Thank you, @kapitoshka438!

Issue #2689, opened 2019-03-26
Rule Request: [LineLengthConfiguration - Add ignoring rule for multiline strings] 


SergLam:
Currently, SwiftLint has next configuration rules for line_length:

line_length:
  warning: 120
  ignores_function_declarations: true
  ignores_comments: true
  ignores_interpolated_strings: true
  ignores_urls: true

https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules/RuleConfigurations/LineLengthConfiguration.swift

New feature request: ignore multi-line strings literals, for example:

`let emailRegEx = """
        (?:[a-z0-9!#$%\\&'*+/=?\\^_`{|}~-]+(?:\\.[a-z0-9!#$%\\&'*+/=?\\^_`{|}"+"~-]+)*|\"(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21\\x23-\\x5b\\x5d-\\"+"x7f]|\\\\[\\x01-\\x09\\x0b\\x0c\\x0e-\\x7f])*\")@(?:(?:[a-z0-9](?:[a-"+"z0-9-]*[a-z0-9])?\\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\\[(?:(?:25[0-5"+"]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-"+"9][0-9]?|[a-z0-9-]*[a-z0-9]:(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21"+"-\\x5a\\x53-\\x7f]|\\\\[\\x01-\\x09\\x0b\\x0c\\x0e-\\x7f])+)\\])
        """`

hafezs:
Any update on this?

PR #6014, opened 2025-03-06
Add `line_length` parameter: `ignores_multiline_strings`

thisIsTheFoxe:
- https://github.com/realm/SwiftLint/issues/2689
- ignore swift multiline strings marked by `"""`

The idea here is:
1. if line is a string only (within two `"""`)
2. and if next line that isn't a string only doesn't start with `"""` -> ignore as it's part of a multiline string
if 2 is false, then we might be within a string that's concatinated:
```swift
let a = "abc" +
"def"
+ "ghi"
```

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Metrics/LineLengthRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Metrics/LineLengthRule.swift
@@ -62,12 +62,9 @@ struct LineLengthRule: Rule {
                 return nil
             }
 
-            if configuration.ignoresMultilineStrings,
-               syntaxKindsByLine.value[line.index] == [.string] {
-                let lastStringLineIndex = (syntaxKindsByLine.value.dropFirst(line.index + 1).firstIndex(where: { $0 != [.string] }) ?? (file.lines.count + 1)) - 2
-                if file.lines[lastStringLineIndex].content.trimmingCharacters(in: .whitespaces).hasPrefix("\"\"\"") {
-                    return nil
-                }
+            if configuration.ignoresMultilineStrings &&
+               lineIsMultilineString(line, file: file, syntaxKindsByLine: syntaxKindsByLine.value) {
+                return nil
             }
 
             for pattern in configuration.excludedLinesPatterns where line.containsMatchingPattern(pattern) {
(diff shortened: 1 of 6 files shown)


Naveen-C-Ramachandrappa (approved):

thisIsTheFoxe:
@SimplyDanny sorry to ping you, but you seem the most active here and would really love to see this get in.. lmk what you think.. ^^

Tests/BuiltInRulesTests/LineLengthRuleTests.swift
```diff
@@ -21,6 +21,16 @@ final class LineLengthRuleTests: SwiftLintTestCase {
     private let interpolatedString = Example("print(\"\\(value)" + String(repeating: "A", count: 113) + "\" )\n")
     private let plainString = Example("print(\"" + String(repeating: "A", count: 121) + ")\"\n")
 
+    private let multilineString = Example("let multilineString = \"\"\"\n" +
+        String(repeating: "A", count: 121) + "\n" +
+        "\"\"\"\n")
+    private let multilineStringWithExpression = Example("let multilineString = \"\"\"\n" +
+        String(repeating: "A", count: 121) + "\n\n\"\"\"; let a = 1")
+    private let multilineStringWithNewlineExpression = Example("let multilineString = \"\"\"\n" +
+        String(repeating: "A", count: 121) + "\n\n\"\"\"\n; let a = 1")
```
SimplyDanny, line 30: Please add another test case with a function called on the string like
```swift
"""
AAA
""".functionCall()
```

SimplyDanny (approved):
Looks good to me! Please add a CHANGELOG entry to credit yourself and let other people know about the new option.

Source/SwiftLintBuiltInRules/Rules/Metrics/LineLengthRule.swift
```diff
@@ -88,6 +93,30 @@ struct LineLengthRule: Rule {
...
+    /// ```
+    /// let a = """
+    /// <line is somewhere in here>
+    /// """
+    /// ```
+    private func lineIsMultilineString(_ line: Line, file: SwiftLintFile, syntaxKindsByLine: [[SyntaxKind]]) -> Bool {
+        // contents of multiline strings only include one string element per line
+        guard syntaxKindsByLine[line.index] == [.string] else { return false }
+
+        // find the trailing delimiter `"""` in to make sure we're not in a list of concatinated strings
```
SimplyDanny, line 107:
```suggestion
        // find the trailing delimiter `"""` in order to make sure we're not in a list of concatenated strings
```

Tests/BuiltInRulesTests/LineLengthRuleTests.swift
```diff
@@ -79,6 +92,27 @@ final class LineLengthRuleTests: SwiftLintTestCase {
...
                    commentDoesntViolate: false, stringDoesntViolate: false)
     }
+
+    func testLineLengthWithIgnoreMultilineStringsTrue() {
+        let triggeringLines = [multilineStringFail]
+        let nonTriggeringLines = [
+            multilineString,
+            multilineStringWithExpression,
+            multilineStringWithNewlineExpression,
+            multilineStringWithFunction
```
thisIsTheFoxe, line 102:
```suggestion
            multilineStringWithFunction,
```

SimplyDanny (approved):

SimplyDanny:
Thank you, @thisIsTheFoxe!

PR #6015, opened 2025-03-08
Linux: static builds + arm64 build

bradleymackey:
- Linux builds with Docker are now built using the [static Linux SDK from Apple](https://www.swift.org/documentation/articles/static-linux-getting-started.html) so that Linux hosts no longer need to worry about runtime dependencies.
- Add an ARM64 build to the build artifacts and include this in linux release arifacts and the cross-platform artifact bundle.
   - The artifact paths of the original amd64 binary are unaffected, and the new arm64 binary is named `swiftlint_arm64`, located in the same directory as the amd64 binary.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -25,9 +25,6 @@ jobs:
   build:
     name: Build Docker Image
     runs-on: ubuntu-24.04
-    strategy:
-      matrix:
-        platform: [amd64, arm64]
     steps:
     - name: Define variables on push to `main`
       if: github.event_name == 'push'
@@ -62,6 +59,8 @@ jobs:
         REPOSITORY: ${{ github.repository }}
     - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@v3
+    - name: Install cross-binutils for aarch64
+      run: sudo apt install -y binutils-aarch64-linux-gnu
     - name: Login to GitHub registry
       uses: docker/login-action@v3
       with:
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -1,64 +1,20 @@
...
-COPY --from=builder /usr/lib/swift/linux/libFoundationNetworking.so /usr/lib
-COPY --from=builder /usr/lib/swift/linux/libFoundationXML.so /usr/lib
-COPY --from=builder /usr/lib/swift/linux/libswift_Concurrency.so /usr/lib
-COPY --from=builder /usr/lib/swift/linux/libswift_RegexParser.so /usr/lib
-COPY --from=builder /usr/lib/swift/linux/libswift_StringProcessing.so /usr/lib
-COPY --from=builder /usr/lib/swift/linux/libswiftCore.so /usr/lib
-COPY --from=builder /usr/lib/swift/linux/libswiftDispatch.so /usr/lib
-COPY --from=builder /usr/lib/swift/linux/libswiftGlibc.so /usr/lib
-COPY --from=builder /usr/lib/swift/linux/libswiftSynchronization.so /usr/lib
-COPY --from=builder /usr/lib/swift/linux/libswiftSwiftOnoneSupport.so /usr/lib
```
SimplyDanny, line 58: These runtime dependencies are not copied anymore?

Makefile
```diff
@@ -120,26 +127,29 @@ portable_zip: installables
...
 	cp -f "$(SWIFTLINT_EXECUTABLE)" "$(ARTIFACT_BUNDLE_PATH)/swiftlint-$(VERSION_STRING)-macos/bin/swiftlint"
 	cp -f "$(SWIFTLINT_EXECUTABLE_LINUX_AMD64)" "$(ARTIFACT_BUNDLE_PATH)/swiftlint-$(VERSION_STRING)-linux-gnu/bin/swiftlint"
+	cp -f "$(SWIFTLINT_EXECUTABLE_LINUX_ARM64)" "$(ARTIFACT_BUNDLE_PATH)/swiftlint-$(VERSION_STRING)-linux-gnu/bin/swiftlint_arm64"
 	cp -f "$(LICENSE_PATH)" "$(ARTIFACT_BUNDLE_PATH)"
 	(cd "$(TEMPORARY_FOLDER)"; zip -yr - "SwiftLintBinary.artifactbundle") > "./SwiftLintBinary.artifactbundle.zip"
 
-zip_linux: docker_image $(SWIFTLINT_EXECUTABLE_LINUX_AMD64)
+zip_linux: docker_image $(SWIFTLINT_EXECUTABLE_LINUX_AMD64) $(SWIFTLINT_EXECUTABLE_LINUX_ARM64)
 	$(eval TMP_FOLDER := $(shell mktemp -d))
 	cp -f $(SWIFTLINT_EXECUTABLE_LINUX_AMD64) "$(TMP_FOLDER)/swiftlint"
```
SimplyDanny, line 142: We should probably rename this to `swiftlint_amd64` once everything works.

tools/build-linux-release.sh
```diff
@@ -0,0 +1,27 @@
...
+
+set -eo pipefail
+
+pushd "$(dirname "${BASH_SOURCE[0]}")/.." > /dev/null
+
+BUILD_ARGS=(
+    --product swiftlint
+    --configuration release
+    -Xlinker -S
+)
```
SimplyDanny, line 11: Are the other linker options not needed anymore?

SimplyDanny (commented):
Looking forward to this!

To test the build you could make the `docker.yml` workflow trigger in pull requests.

.github/workflows/docker.yml
```diff
@@ -69,16 +71,29 @@ jobs:
...
-    - name: Rename binary artifact
+    - name: Strip and move binary artifacts
       if: contains(env.OUTPUT_TYPE, 'local')
-      run: mv artifacts/usr/bin/swiftlint artifacts/usr/bin/swiftlint_linux_amd64
-    - name: Upload binary artifact
+      run: |
+        strip artifacts/linux_amd64/usr/bin/swiftformat
+        mv artifacts/linux_amd64/usr/bin/swiftlint artifacts/usr/bin/swiftlint_linux_amd64
+        aarch64-linux-gnu-strip artifacts/linux_arm64/usr/bin/swiftformat
+        mv artifacts/linux_arm64/usr/bin/swiftlint artifacts/usr/bin/swiftlint_linux_arm64
```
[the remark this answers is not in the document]
bradleymackey: Oops! I'll double check all these

Dockerfile
bradleymackey: Still playing around here, most of these likely will need to still get copied into the container.

tools/build-linux-release.sh
bradleymackey: Also still playing around here. The main issue I'm having at the moment is getting libc statically linked while keeping the other dependencies dynamically linked. `--swift-sdk` is seemingly always producing a fully statically linked binary.

SimplyDanny (approved):

PR #6025, opened 2025-03-17
Add new opt-in `non_final_class` rule

JaviSoto:


Source/SwiftLintBuiltInRules/Rules/Style/NonFinalClassRule.swift
```diff
@@ -0,0 +1,51 @@
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+@SwiftSyntaxRule(correctable: true, optIn: true)
+struct NonFinalClassRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
```
SimplyDanny, line 8: We now support adding rationals to rule descriptions. This feels like a very good candidate for some more explanation, as it seems highly opinionated and project-specific considering that `open` implies `public` which is often not what libraries want instead.

Source/SwiftLintBuiltInRules/Rules/Style/NonFinalClassRule.swift
```diff
@@ -0,0 +1,51 @@
...
+            Example("public class MyClass {}"): Example("public final class MyClass {}"),
+        ]
+    )
+
+    func makeVisitor(file: SwiftLintFile) -> ViolationsSyntaxVisitor<ConfigurationType>? {
+        Visitor(configuration: configuration, file: file)
+    }
+
+    private final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
```
SimplyDanny, line 33: Prefer `visitPost` instead.

Source/SwiftLintBuiltInRules/Rules/Style/NonFinalClassRule.swift
```diff
@@ -0,0 +1,51 @@
...
+            let modifiers = node.modifiers
+            if !modifiers.contains(keyword: .final),
+               !modifiers.contains(keyword: .open) {
+                let classToken = node.classKeyword
```
SimplyDanny, line 37:
```suggestion
                let classTokenPosition = node.classKeyword.positionAfterSkippingLeadingTrivia
```

Source/SwiftLintBuiltInRules/Rules/Style/NonFinalClassRule.swift
```diff
@@ -0,0 +1,51 @@
...
+                violations.append(.init(
+                    position: classToken.positionAfterSkippingLeadingTrivia,
+                    reason: "Classes should be marked as `final` unless they are explicitly `open`",
```
SimplyDanny, line 40: Could be more succinct and targeted to the single class being checked:
```suggestion
                    reason: "Class should be marked as `final` or explicitly `open`",
```

Tests/SwiftLintBuiltInRulesTests/Rules/Style/NonFinalClassRuleTests.swift
SimplyDanny, line 1: This test isn't needed. All examples in the description are test cases by design.

Source/SwiftLintBuiltInRules/Rules/Style/NonFinalClassRule.swift
```diff
@@ -0,0 +1,51 @@
...
+        name: "Non-Final Class",
+        description: "Classes should be marked as `final` unless they are explicitly `open`.",
+        kind: .idiomatic,
+        nonTriggeringExamples: [
+            Example("final class MyClass {}"),
+            Example("open class MyClass {}"),
+            Example("public final class MyClass {}"),
+        ],
+        triggeringExamples: [
+            Example("class MyClass {}"),
```
SimplyDanny, line 19: Please add violation markers `↓` to indicate the exact positions where the violations are expected to appear.

SimplyDanny (changes requested):
Nice little rule that may have its use cases!

There are a few violations in SwiftLint's code base itself. They can likely be auto-fixed. However, if there is inheritance involved we shouldn't put an `open` but rather disable the rule locally or globally in the configuration (in case there are too many "false" positives).

Other than that, I've added a few remarks inline.

bcapps:
This would be awesome! Have been wanting this feature since this PR from 2018: https://github.com/realm/SwiftLint/pull/2124

Issue #5183, opened 2023-08-17
no_magic_number - Ignore UIColor initializers


LowAmmo:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New rule request

Requesting another enhancement to "no_magic_numbers", where it will ignore UIColor definitions.

Since the nature of definiting a color is going to have some "magic numbers", I think it would make sense to just ignore the values - 

    static let grayColor: UIColor = #colorLiteral(red: 0.7019607843, green: 0.7019607843, blue: 0.7019607843, alpha: 1)
    let myColor: UIColor = UIColor(red: 0.6, green: 1.0, blue: 0.2, alpha: 0.52)
    let yourColor: UIColor = UIColor(hue: 0.9, saturation: 0.6, brightness: 0.333334, alpha: 1.0)

(including all other UIColor constructors

I don't think this needs to be an "opt-in" configuration...I would think most people would want this by default...but not opposed to it being an opt-in rule.

Hopefully that makes sense!

-Thanks!

PR #6035, opened 2025-03-26
Fix Ignore UIColor initializers in no_magic_numbers rule issue

suojae:
### Description

**Resolves: #5183**

This PR addresses issue #5183 by making the `no_magic_numbers` rule ignore numerical literals used in UIColor initializers. Since color definitions inherently use numeric values (such as RGB components, alpha values, etc.), it makes sense to exempt these from being flagged as magic numbers.

### Implementation
- Added a new `isPartOfUIColorInitializer()` method to detect when a numeric literal is part of a UIColor initialize.
- Modified the `collectViolation` method to skip reporting violations for numbers used in color definitions.

### Test Coverage
Added several non-triggering examples to ensure different UIColor initializer formats are properly handled:
- Standard RGB with alpha: `UIColor(red: 0.6, green: 1.0, blue: 0.2, alpha: 0.52)`
- Color literal: `#colorLiteral(red: 0.7019607843, green: 0.7019607843, blue: 0.7019607843, alpha: 1)`
- HSB with alpha: `UIColor(hue: 0.9, saturation: 0.6, brightness: 0.333334, alpha: 1.0)`
- Display P3 color space: `UIColor(displayP3Red: 0.3, green: 0.8, blue: 0.5, alpha: 0.75)`
- Grayscale with alpha: `UIColor(white: 0.5, alpha: 0.8)`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 * Ignore `UIColor` initializers in `no_magic_numbers` rule.  
   [suojae](https://github.com/suojae)
+  [hyeffie](https://github.com/hyeffie)
   [#5183](https://github.com/realm/SwiftLint/issues/5183)
 
 * Exclude types with a `@Suite` attribute and functions annotated with `@Test` from `no_magic_numbers` rule.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 
 ### Enhancements
 
-* Fix Ignore UIColor initializers in no_magic_numbers rule issue.
+* Ignore `UIColor` initializers in `no_magic_numbers` rule.  
   [suojae](https://github.com/suojae)
   [#5183](https://github.com/realm/SwiftLint/issues/5183)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
 ### Enhancements
 
+* Fix Ignore UIColor initializers in no_magic_numbers rule issue.
+  [suojae](https://github.com/suojae)
+  [#5183](https://github.com/realm/SwiftLint/issues/5183)
+
 * Exclude types with a `@Suite` attribute and functions annotated with `@Test` from `no_magic_numbers` rule.
   Also treat a type as a `@Suite` if it contains `@Test` functions.  
   [SimplyDanny](https://github.com/SimplyDanny)
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
@@ -113,11 +113,9 @@ struct NoMagicNumbersRule: Rule {
             Example("""
             let myColor: UIColor = UIColor(red: 0.6, green: 1.0, blue: 0.2, alpha: 0.52)
             """),
-            
             Example("""
             let colorLiteral = #colorLiteral(red: 0.7019607843, green: 0.7019607843, blue: 0.7019607843, alpha: 1)
             """),
-
             Example("""
             let yourColor: UIColor = UIColor(hue: 0.9, saturation: 0.6, brightness: 0.333334, alpha: 1.0)
             """),
@@ -245,11 +243,9 @@ private extension NoMagicNumbersRule {
             if node.isOperandOfFreestandingShiftOperation() {
                 return
             }
-            
             if node.isPartOfUIColorInitializer() {
                 return
             }
-            
             let violation = node.positionAfterSkippingLeadingTrivia
             if let extendedTypeName = node.extendedTypeName() {
                 if !testClasses.contains(extendedTypeName) {
@@ -350,7 +346,6 @@ private extension ExprSyntaxProtocol {
         }
         return false
     }
-    
     func isPartOfUIColorInitializer() -> Bool {
         var parent = parent
         while let currentParent = parent, !currentParent.is(FunctionCallExprSyntax.self) {
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
@@ -368,9 +368,8 @@ private extension ExprSyntaxProtocol {
                 return true
             }
             if let memberAccess = call.calledExpression.as(MemberAccessExprSyntax.self),
-               memberAccess.declName.baseName.text == "init",
                let baseExpr = memberAccess.base?.as(DeclReferenceExprSyntax.self),
-               baseExpr.baseName.text == "UIColor" {
+               baseExpr.baseName.text == "UIColor" && memberAccess.declName.baseName.text == "init" {
                 return true
             }
         }
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
@@ -252,6 +252,7 @@ private extension NoMagicNumbersRule {
             if node.isOperandOfFreestandingShiftOperation() {
                 return
             }
+            
             if node.isPartOfUIColorInitializer() {
                 return
             }
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
@@ -245,6 +245,11 @@ private extension NoMagicNumbersRule {
             if node.isOperandOfFreestandingShiftOperation() {
                 return
             }
+            
+            if node.isPartOfUIColorInitializer() {
+                return
+            }
+            
             let violation = node.positionAfterSkippingLeadingTrivia
             if let extendedTypeName = node.extendedTypeName() {
                 if !testClasses.contains(extendedTypeName) {
@@ -342,6 +347,26 @@ private extension ExprSyntaxProtocol {
            let operatorSymbol = operation.operator.as(BinaryOperatorExprSyntax.self)?.operator.tokenKind,
            [.binaryOperator("<<"), .binaryOperator(">>")].contains(operatorSymbol) {
             return operation.parent?.isProtocol((any ExprSyntaxProtocol).self) != true
+        }
+        return false
+    }
+    
+    func isPartOfUIColorInitializer() -> Bool {
+        var parent = parent
+        while let currentParent = parent, !currentParent.is(FunctionCallExprSyntax.self) {
+            parent = currentParent.parent
+        }
+        guard let functionCall = parent?.as(FunctionCallExprSyntax.self) else {
+            return false
+        }
+        if let memberAccess = functionCall.calledExpression.as(MemberAccessExprSyntax.self),
+           let baseExpr = memberAccess.base?.as(DeclReferenceExprSyntax.self),
+           baseExpr.baseName.text == "UIColor" {
+            return true
+        }
+        if let identifierExpr = functionCall.calledExpression.as(DeclReferenceExprSyntax.self),
+           identifierExpr.baseName.text == "colorLiteral" {
+            return true
         }
         return false
     }
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
@@ -347,20 +347,30 @@ private extension ExprSyntaxProtocol {
         return false
     }
     func isPartOfUIColorInitializer() -> Bool {
-        var parent = parent
-        while let currentParent = parent, !currentParent.is(FunctionCallExprSyntax.self) {
-            parent = currentParent.parent
-        }
-        guard let functionCall = parent?.as(FunctionCallExprSyntax.self) else {
+        let uiColorInitializerLabels: Set<String> = [
+            "white", "alpha", "red", "displayP3Red", "green", "blue", "hue", "saturation", "brightness",
+            "cgColor", "ciColor", "resource", "patternImage"
+        ]
+        let colorLiteralLabels: Set<String> = ["red", "green", "blue", "alpha"]
+        guard let param = parent?.as(LabeledExprSyntax.self),
+              let label = param.label?.text else {
             return false
         }
-        if let memberAccess = functionCall.calledExpression.as(MemberAccessExprSyntax.self),
-           let baseExpr = memberAccess.base?.as(DeclReferenceExprSyntax.self),
-           baseExpr.baseName.text == "UIColor" {
-            return true
+        if uiColorInitializerLabels.contains(label),
+           let call = param.parent?.as(LabeledExprListSyntax.self)?.parent?.as(FunctionCallExprSyntax.self) {
+            if let calledExpr = call.calledExpression.as(DeclReferenceExprSyntax.self),
+               calledExpr.baseName.text == "UIColor" {
+                return true
+            }
+            if let memberAccess = call.calledExpression.as(MemberAccessExprSyntax.self),
+               let baseExpr = memberAccess.base?.as(DeclReferenceExprSyntax.self),
+               baseExpr.baseName.text == "UIColor" {
+                return true
+            }
         }
-        if let identifierExpr = functionCall.calledExpression.as(DeclReferenceExprSyntax.self),
-           identifierExpr.baseName.text == "colorLiteral" {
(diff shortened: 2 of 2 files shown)


Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -326,6 +346,26 @@ private extension ExprSyntaxProtocol {
         }
         return false
     }
+    
+    func isPartOfUIColorInitializer() -> Bool {
```
SimplyDanny, line 350: This does not look quite right (and it isn't as tests are failing). I'll give you an example implementation for you last test example `UIColor(white: 0.5, alpha: 0.8)`:

```swift
if let param = parent?.as(LabeledExprSyntax.self),
   ["white", "alpha"].contains(param.label?.text),
   let call = param.parent?.as(LabeledExprListSyntax.self)?.parent?.as(FunctionCallExprSyntax.self),
   call.calledExpression.as(DeclReferenceExprSyntax.self)?.baseName.text == "UIColor" {
    return true
}
```

Make especially sure that argument labels match as well. Looking for a specific parent node in a `while` loop is often not what you want given that the call can be at an arbitrarily high level totally out of scope of the expression to check.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -326,6 +346,26 @@ private extension ExprSyntaxProtocol {
         }
         return false
     }
+
```
SimplyDanny, line 349: Trailing whitespace.

SimplyDanny (changes requested):
Thank you for the adaption! Tests are currently failing. Please make sure your examples are valid and reflect the new behavior.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
suojae: Thanks your feedback!

commit log: 634c15edf383292849e2aebad21bda645a38b0f8

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 
 ### Enhancements
 
+* Fix Ignore UIColor initializers in no_magic_numbers rule issue.
```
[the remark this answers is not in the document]
hyeffie: Thanks for your feedback.
commit log: 607200e957377790bdea942434fc2fb3acdb6ecc

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
hyeffie: I’ve made the following improvements based on your feedback:
- Replaced the while loop with a more straightforward check using `parent?.as(LabeledExprSyntax.self)` to enhance readability and clarity.
- Introduced `uiColorInitializerLabels` and `colorLiteralLabels` sets to handle label matching more explicitly and cleanly.
- Refined the extraction of function calls to ensure more accurate detection of `UIColor` and colorLiteral initializers.
- Made sure to handle both `FunctionCallExprSyntax` and `MacroExpansionExprSyntax` consistently for better accuracy.

Thanks for your suggestions! Let me know if you need any further adjustments. 

commit log: 0275a24ef5fa5ade2ffbc35f09cd453c29c9376e

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -326,6 +346,35 @@ private extension ExprSyntaxProtocol {
         }
         return false
     }
+    func isPartOfUIColorInitializer() -> Bool {
+        let uiColorInitializerLabels: Set<String> = [
+            "white", "alpha", "red", "displayP3Red", "green", "blue", "hue", "saturation", "brightness",
+            "cgColor", "ciColor", "resource", "patternImage"
+        ]
+        let colorLiteralLabels: Set<String> = ["red", "green", "blue", "alpha"]
```
SimplyDanny, line 354: Please move this down just before the condition which uses it. You may even inline the variable.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -326,6 +346,35 @@ private extension ExprSyntaxProtocol {
...
+              let label = param.label?.text else {
+            return false
+        }
+        if uiColorInitializerLabels.contains(label),
+           let call = param.parent?.as(LabeledExprListSyntax.self)?.parent?.as(FunctionCallExprSyntax.self) {
+            if let calledExpr = call.calledExpression.as(DeclReferenceExprSyntax.self),
+               calledExpr.baseName.text == "UIColor" {
+                return true
+            }
+            if let memberAccess = call.calledExpression.as(MemberAccessExprSyntax.self),
```
SimplyDanny, line 365: Which kinds of expressions do you have in mind with this condition? Would be great to have test examples for them.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -326,6 +346,35 @@ private extension ExprSyntaxProtocol {
         }
         return false
     }
```
SimplyDanny, line 348: There should be an empty line between the functions. My previous comment was about superfluous whitespace in the line.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
suojae: Thanks for the comment. This condition checks for UIColor initializations that use member access syntax like `UIColor.init()`. I've added test examples for this pattern in the nonTriggeringExamples array, including both a variable declaration and a function return context.

commit log: a600255e00281930c1b98a99926916def3bf6699

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
hyeffie: Moved the variable closer to its usage and inlined it as suggested. Thank you for the feedback!

commit log: 8bd2e5276969cc3444a34e70487cfa2d9332350b

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
hyeffie: Apologies for the misunderstanding. Adjusted the spacing accordingly!

commit log: d2abd897433f64039b2b98634be71beac6dfa3f7

hyeffie: Apologies for the misunderstanding. Adjusted the spacing accordingly!

commit log: bcd854b1c3e7d99bcbbf541b6d12d1e43246344d

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
SimplyDanny: So let's also make sure that `memberAccess.declName` is `init` in order to avoid matching function calls we don't want to exclude from analysis.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -110,6 +110,32 @@ struct NoMagicNumbersRule: Rule {
             let a = 2
             #endif
             """),
+            Example("""
+            let myColor: UIColor = UIColor(red: 0.6, green: 1.0, blue: 0.2, alpha: 0.52)
+            """),
+            Example("""
+            let colorLiteral = #colorLiteral(red: 0.7019607843, green: 0.7019607843, blue: 0.7019607843, alpha: 1)
+            """),
+            Example("""
```
SimplyDanny, line 119: Let's keep the first two examples and add the `excludeFromDocumentation: true` argument to all the other example in order to avoid cluttering the documentation page too much.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -326,6 +355,31 @@ private extension ExprSyntaxProtocol {
         }
         return false
     }
+    
+    func isPartOfUIColorInitializer() -> Bool {
+        guard let param = parent?.as(LabeledExprSyntax.self),
+              let label = param.label?.text else {
+            return false
+        }
+        if ["white", "alpha", "red", "displayP3Red", "green", "blue", "hue", "saturation", "brightness","cgColor", "ciColor", "resource", "patternImage"].contains(label),
```
SimplyDanny, line 364: This line might be too long. It's fine to have the labels defined in a variable here.

SimplyDanny (approved):
A few more nits, then I'm going to be happy with that. 😉

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
hyeffie: I’ve updated the rule to explicitly check memberAccess.declName == "init". Now, only UIColor.init(...) calls are matched, preventing unintended function calls from being analyzed.

Let me know if you see any edge cases I might have missed!

commit log: 22f489ab8a5dc36cc174a2129ff8d76ead843900

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
hyeffie: Added array constants for label literals. 😃

commit log: 35d6acbd8482031ad60d79ca124843f2e0707520

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
suojae: I've and added `excludeFromDocumentation: true` to all UIColor examples except the first two. Thanks for the thorough review!

commit log: 3926fa90dc3b6046318415be6c89bf5887906f43

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
hyeffie: And I updated the condition to use short-circuit evaluation (&&) so that memberAccess.declName.baseName.text is only checked if baseExpr.baseName.text == "UIColor".

commit log: 3926fa90dc3b6046318415be6c89bf5887906f43

suojae: After adding the `memberAccess.declName.baseName.text == "init"` condition, linting performance for the Aerial project has decreased significantly. We've tried optimizing it by short-circuit evaluation with the `&&` operator, but the performance issue persists.

Accessing `memberAccess.declName.baseName.text` seems very expensive,  Could we consider using only the `baseExpr.baseName.text == "UIColor"` check instead? 

Allowing magic numbers in all UIColor methods (not just initializers) might be an acceptable trade-off for better performance. What do you think about this?

SimplyDanny: Using `&&` instead of separate conditions does not change the semantic of the code. Separate conditions are short-circuit evaluated, too, and are idiomatic Swift. Actually, this is a perfect idea for a new rule! 🤓

That said, the performance issue is some problem with the build machine I haven't figured out yet. It's always the Aerial performance that degrades. You do not need to worry about that.

While what you say about all methods on UIColor is valid, we'd need to make this explicit (potentially with a new option or only for a set of selected methods). People could extend UIColor to add new methods where they don't wish for this behavior. That's why I was also asking you to check the label names, not only the method name.

We can think about this, but preferably separate from the current PR.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -326,6 +354,33 @@ private extension ExprSyntaxProtocol {
         }
         return false
     }
+    
+    func isPartOfUIColorInitializer() -> Bool {
+        guard let param = parent?.as(LabeledExprSyntax.self),
+              let label = param.label?.text else {
+            return false
+        }
+        let uiColorInitializerLabels = ["white", "alpha", "red", "displayP3Red", "green", "blue", "hue", "saturation", "brightness","cgColor", "ciColor", "resource", "patternImage"]
```
SimplyDanny, line 363: I guess this line would be too long.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
suojae: Thanks for the clarification about the short-circuit evaluation. I understand now that the performance issue is unrelated to our code changes.

I agree that a more comprehensive solution for `UIColor` methods would be valuable. I'll keep the current implementation with label checks as you suggested, and I'll prepare a separate PR soon that addresses the UIColor magic numbers handling more systematically

Thanks for your review and guidance! ☺️

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
suojae: Thanks!

commit log: [d324e79](https://github.com/realm/SwiftLint/pull/6035/commits/d324e79f58830ebb59aa0684705c68b87e6fe6d5)

SimplyDanny:
There are still linting issues. You can find them as inline comments under the "Files changed" tab.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -326,6 +354,35 @@ private extension ExprSyntaxProtocol {
         }
         return false
     }
+    func isPartOfUIColorInitializer() -> Bool {
```
SimplyDanny, line 357:
```suggestion

    func isPartOfUIColorInitializer() -> Bool {
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -326,6 +354,35 @@ private extension ExprSyntaxProtocol {
...
+                return true
+            }
+            if let memberAccess = call.calledExpression.as(MemberAccessExprSyntax.self),
+               let baseExpr = memberAccess.base?.as(DeclReferenceExprSyntax.self),
+               baseExpr.baseName.text == "UIColor" && memberAccess.declName.baseName.text == "init" {
+                return true
+            }
+        }
+        let colorLiteralLabels = ["red", "green", "blue", "alpha"]
+        if colorLiteralLabels.contains(label),
```
SimplyDanny, line 379:
```suggestion
        if ["red", "green", "blue", "alpha"].contains(label),
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/NoMagicNumbersRule.swift
```diff
@@ -326,6 +354,35 @@ private extension ExprSyntaxProtocol {
...
+        ]
+        if uiColorInitializerLabels.contains(label),
+           let call = param.parent?.as(LabeledExprListSyntax.self)?.parent?.as(FunctionCallExprSyntax.self) {
+            if let calledExpr = call.calledExpression.as(DeclReferenceExprSyntax.self),
+               calledExpr.baseName.text == "UIColor" {
+                return true
+            }
+            if let memberAccess = call.calledExpression.as(MemberAccessExprSyntax.self),
+               let baseExpr = memberAccess.base?.as(DeclReferenceExprSyntax.self),
+               baseExpr.baseName.text == "UIColor" && memberAccess.declName.baseName.text == "init" {
```
SimplyDanny, line 374:
```suggestion
               baseExpr.baseName.text == "UIColor",
               memberAccess.declName.baseName.text == "init" {
```

SimplyDanny (approved):
This looks very good now. Thank you both for the team work! 👍 

There are only a few more styling issues you can ignore as I'll commit them myself.

PR #6049, opened 2025-04-08
Remove `opaque_over_existential` rule

### LINKED ISSUE #6046: `opaque_over_existential` breaks Codable and other conformances
### New Issue Checklist

### Bug Description

There are some cases where it is invalid to convert `any` to `some`, because it breaks protocol conformances. A minimum triggering example would be:

```diff
- init(from decoder: any Decoder) throws
+ init(from decoder: some Decoder) throws
```

To see why this is a problem, consider this file, called CodableLintBug.swift:

```swift
import Foundation

struct MyStruct {
  let name: String
}

extension MyStruct: Codable {
  enum CodingKeys: String, CodingKey {
    case name
  }
  // this "any" should not become "some"
  init(from decoder: any Decoder) throws {
    print(Self.self, #function, "customized implementation")
    let container = try decoder.container(keyedBy: CodingKeys.self)
    self.name = try container.decode(String.self, forKey: .name)
  }

  // this "any" should not become "some"
  func encode(to encoder: any Encoder) throws {
    print(Self.self, #function, "customized implementation")
    var container = encoder.container(keyedBy: CodingKeys.self)
    try container.encode(name, forKey: .name)
  }
}

let myStruct = MyStruct(name: "asdf")
let data = try JSONEncoder().encode(myStruct)
let decoded = try JSONDecoder().decode(MyStruct.self, from: data)
print("decoded \(decoded)")
```

When you run it with `swift CodableLintBug.swift`, you get this output:

```
MyStruct encode(to:) customized implementation
MyStruct init(from:) customized implementation
decoded MyStruct(name: "asdf")
```

However, if you enable the `opaque_over_existential` SwiftLint rule, it converts `any Decoder` and `any Encoder` to `some Decoder` and `some Encoder`, respectively. Running it again, it produces:

```
decoded MyStruct(name: "asdf")
```

Notice that it is still compiling without warnings or errors, but it is skipping over the customized conformances and instead going with the compiler-generated conformances. This could lead to some very confusing bugs! (This is also why people have proposed adding a keyword that indicates your intent to satisfy a protocol requirement, but that's for another discussion.)

⚠  this does not _only_ affect Codable/Decodable. It affects any function that takes an `any` parameter as part of satisfying a protocol conformance. I'm not sure if there's a way to detect this at a syntactic level, so I don't know if there's a completely correct way to implement this rule 🤔

```bash
$ swiftlint --no-cache --autocorrect
Correcting Swift files in current working directory
Correcting 'CodableLintBug.swift' (1/1)
/Users/zev/Downloads/CodableLintBug/CodableLintBug.swift: Corrected opaque_over_existential 2 times
Done correcting 1 file!
```

### Environment

* SwiftLint version: **0.59.0**
* Xcode version: **Xcode 16.3 (16E140) on macOS 15.4 (24E248)**
* Installation method used: **Homebrew and also install from standalone binaries provided on release page.**
* Configuration file:

```yml
opt_in_rules:
  - opaque_over_existential
```

Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? **No.**

SimplyDanny:
This partially reverts commit 82736d1925de388e6b5abb4c5a293393043608fe and resolves #6046.


SimplyDanny:
0.59.1 just got released, @417-72KI.

PR #6050, opened 2025-04-08
Fix issue referencing `FrameworkTests` from another Bazel workspace

jszumski:
Allow `FrameworkTests` data to be empty so a workspace can load the Tests package from a release archive.  Similar to https://github.com/realm/SwiftLint/pull/5977.

Two options to fix this long term:
- Add a smoke test of `ExtraRules` into the BCR pipeline.  One downside is you wouldn't detect issues until after the release was made.
- Move the definitions for `FrameworkTests` and `BuiltInRulesTests` into BUILD files one level deeper so they don't get eagerly loaded for anyone using custom `extra_rules`.

Happy to open a PR for either one if we decide on the best approach.

### DIFF
--- a/Tests/BUILD
+++ b/Tests/BUILD
@@ -135,6 +135,7 @@ swift_test(
     name = "FrameworkTests",
     data = glob(
         ["FrameworkTests/Resources/**"],
+        allow_empty = True,
     ),
     visibility = ["//visibility:public"],
     deps = [":FrameworkTests.library"],


SimplyDanny:
Why would Bazel eagerly load the other test bundles? There is no connection to them, is there?

SimplyDanny:
Constantly getting surprised by Bazel. It's quite hard to keep this running given that I don't work with it every day.

So I think the second option is probably the way to go. Would you be open to refactor the `BUILD` files ensuring that nothing breaks for other users?

SimplyDanny (approved):

Issue #6029, opened 2025-03-22
OnlyRule not work with custom_rules


Ernest0-Production:
### New Issue Checklist

- [x] I've Updated SwiftLint to the latest version.
- [x] I've searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues).

### Bug Description

If you specify a custom rule in the `--only-rule` argument, then no violations are detected during linting. (It does scan the files though)

When linting a built-in rule, everything works correctly

```bash
$ swiftlint lint --no-cache --only-rule my_rule
```

### Environment

* SwiftLint version 0.58.2
* Xcode version 16.2
* Installation method used mise-en-place

SimplyDanny:
Historically, handling of custom rules has its peculiarities and we have one of these examples here. Unexpectedly, the command to run a single custom rule is like `swiftlint lint --only-rule custom_rules --only-rule my_rule`.

PR #6057, opened 2025-04-21
only rule should work for individual custom rules

mildm8nnered:
Addresses #6029 and #6058

Allows individual custom rules to be specified in the `only_rules` configuration, or the `--only-rule` command line option, without also having to specify `custom_rules`.

I *think* this may bring individual custom_rules to be almost equivalent to "normal" rules, as they are now supported in `swiftlint:disable` commands, and most other cases as well.

The key change here is that when we calculate the resulting rules, if CustomRules is not in there, we manually add it in.

```
if !resultingRules.contains(where: { $0 is CustomRules }),
   !customRulesIdentifiers.isDisjoint(with: onlyRulesRuleIdentifiers),
   let customRules {
    resultingRules.append(customRules)
}
```

There was already pre-existing code later on to filter which custom rules are enabled.

Some further changes were required to make sure that the warning about configured but not enabled rules was fired for this configuration in the `only_rules` case.

### Other changes

We were calling `(allRulesWrapped.first { $0.rule is CustomRules })?.rule as? CustomRules` in about five different places, so I added a convenience `var` to `[ConfigurationRuleWrapper]`.

`validRuleIdentifiers` is now calculated lazily.

### Order of custom_rules violations

The order of violations reported by custom_rules was not deterministic - see #6058 - violations are now reported sorted by the custom rule identifier.

There is still some weirdness in the order that violations are output in, but I'll raise those issues elsewhere - they should at least be deterministic now.

### DIFF
--- a/Tests/FrameworkTests/ConfigurationTests.swift
+++ b/Tests/FrameworkTests/ConfigurationTests.swift
@@ -132,8 +132,7 @@ final class ConfigurationTests: SwiftLintTestCase {
         let customRules = [customRuleIdentifier: ["name": "A name for this custom rule", "regex": "this is illegal"]]
 
         let config = try Configuration(dict: ["only_rules": only, "custom_rules": customRules])
-        guard let resultingCustomRules = config.rules.customRules
-        else {
+        guard let resultingCustomRules = config.rules.customRules else {
             XCTFail("Custom rules are expected to be present")
             return
         }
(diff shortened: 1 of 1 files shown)


Tests/FrameworkTests/ConfigurationTests.swift
```diff
@@ -142,6 +143,35 @@ final class ConfigurationTests: SwiftLintTestCase {
...
+                "only_rules": only,
+                "custom_rules": customRules,
+            ],
+            onlyRule: [customRuleIdentifier]
+        )
+        guard let resultingCustomRules = config.rules.first(where: { $0 is CustomRules }) as? CustomRules
+        else {
+            XCTFail("Custom rules are expected to be present")
+            return
+        }
```
SimplyDanny:
```suggestion
        let resultingCustomRules = config.rules.customRules
        XCTAssertNotNil(resultingCustomRules)
```

SimplyDanny (approved):
Just a nit. Otherwise looks good to me!

Tests/FrameworkTests/ConfigurationTests.swift
mildm8nnered: So the existing extension on `[ConfigurationRuleWrapper]` was not quite right here - it  retrieves the custom rules **before** disablement, but I added a similar extension for `[any Rule]`, which enables `config.rules.customRules` and retrieves the configured rules, and replaces eight cases of `rules.first(where: { $0 is CustomRules }) as? CustomRules` in the other tests.

PR #6069, opened 2025-04-28
Add opt-in `unneeded_throws_rethrows` rule

tonyskansf:
Add a new opt-in rule that detects unnecessary `throws`, helping avoid misleading definitions and requirements for the caller to write unnecessary error handling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,9 @@
 
 ### Enhancements
 
-* None.
+* Add opt-in `unneeded_throws_rethrows` rules that triggers when declarations
+  marked `throws`/`rethrows` never actually throw or call any throwing code.
+  [Tony Ngo](https://github.com/tonyskansf)
 
 ### Bug Fixes
 
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
@@ -1,0 +1,165 @@
+import SwiftLintCore
+import SwiftSyntax
+
+@SwiftSyntaxRule(correctable: true, optIn: true)
+struct UnneededThrowsRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "unneeded_throws_rethrows",
+        name: "Unneeded (re)throws keyword",
+        description: "Non-throwing functions/variables should not me marked as `throws` or `rethrows`",
+        kind: .lint,
+        nonTriggeringExamples: UnneededThrowsRuleExamples.nonTriggeringExamples,
+        triggeringExamples: UnneededThrowsRuleExamples.triggeringExamples,
+        corrections: UnneededThrowsRuleExamples.corrections
+    )
+}
+
+private extension UnneededThrowsRule {
+    typealias Scope = [ThrowsClauseSyntax]
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        private var scopes = Stack<Scope>()
+
+        override var skippableDeclarations: [any DeclSyntaxProtocol.Type] { [
+            ProtocolDeclSyntax.self,
+        ] }
+
+        override func visit(_: FunctionParameterClauseSyntax) -> SyntaxVisitorContinueKind {
+            .skipChildren
+        }
+
+        override func visit(_ node: InitializerDeclSyntax) -> SyntaxVisitorContinueKind {
+            scopes.openScope(with: node.signature.effectSpecifiers?.throwsClause)
+            return .visitChildren
+        }
+
+        override func visitPost(_: InitializerDeclSyntax) {
+            if let closedScope = scopes.closeScope() {
+                validateScope(
+                    closedScope,
+                    reason: "The initializer does not throw any error"
+                )
+            }
+        }
+
+        override func visit(_ node: AccessorDeclSyntax) -> SyntaxVisitorContinueKind {
+            scopes.openScope(with: node.effectSpecifiers?.throwsClause)
+            return .visitChildren
+        }
+
+        override func visitPost(_: AccessorDeclSyntax) {
+            if let closedScope = scopes.closeScope() {
+                validateScope(
+                    closedScope,
+                    reason: "The accessor does not throw any error"
+                )
+            }
+        }
+
+        override func visitPost(_: VariableDeclSyntax) {
+            if let closedScope = scopes.closeScope() {
+                validateScope(
+                    closedScope,
+                    reason: "The variable does not throw any error"
+                )
+            }
+        }
+
+        override func visit(_ node: FunctionDeclSyntax) -> SyntaxVisitorContinueKind {
+            scopes.openScope(with: node.signature.effectSpecifiers?.throwsClause)
+            return .visitChildren
+        }
+
+        override func visitPost(_: FunctionDeclSyntax) {
+            if let closedScope = scopes.closeScope() {
+                validateScope(
+                    closedScope,
+                    reason: "The body of this function does not throw any error"
+                )
+            }
+        }
+
+        override func visit(_ node: FunctionTypeSyntax) -> SyntaxVisitorContinueKind {
+            scopes.openScope(with: node.effectSpecifiers?.throwsClause)
+            return .visitChildren
+        }
+
+        override func visitPost(_: FunctionTypeSyntax) {
+            if let closedScope = scopes.closeScope() {
+                validateScope(
+                    closedScope,
+                    reason: "The closure type does not throw any error"
+                )
+            }
+        }
+
+        override func visit(_: FunctionCallExprSyntax) -> SyntaxVisitorContinueKind {
+            scopes.openScope()
+            return .visitChildren
+        }
+
+        override func visitPost(_: FunctionCallExprSyntax) {
+            scopes.closeScope()
+        }
+
+        override func visit(_: DoStmtSyntax) -> SyntaxVisitorContinueKind {
+            scopes.openScope()
+            return .visitChildren
+        }
+
+        override func visitPost(_ node: CodeBlockSyntax) {
+            if node.parent?.is(DoStmtSyntax.self) == true {
+                scopes.closeScope()
+            }
+        }
+
+        override func visitPost(_ node: TryExprSyntax) {
+            if node.questionOrExclamationMark == nil {
+                scopes.markCurrentScopeAsThrowing()
+            }
+        }
+
+        override func visitPost(_: ThrowStmtSyntax) {
+            scopes.markCurrentScopeAsThrowing()
+        }
+
+        private func validateScope(_ scope: Scope, reason: String) {
+            guard let throwsToken = scope.last?.throwsSpecifier else { return }
+            violations.append(
+                ReasonedRuleViolation(
+                    position: throwsToken.positionAfterSkippingLeadingTrivia,
+                    reason: reason,
+                    correction: ReasonedRuleViolation.ViolationCorrection(
+                        start: throwsToken.positionAfterSkippingLeadingTrivia,
+                        end: throwsToken.endPosition,
+                        replacement: ""
+                    )
+                )
+            )
+        }
+    }
+}
+
+private extension Stack where Element == UnneededThrowsRule.Scope {
+    mutating func markCurrentScopeAsThrowing() {
+        modifyLast { currentScope in
+            _ = currentScope.popLast()
+        }
+    }
+
+    mutating func openScope() {
+        push([])
+    }
+
+    mutating func openScope(with throwsClause: ThrowsClauseSyntax?) {
+        if let throwsClause {
+            push([throwsClause])
+        }
+    }
+
+    @discardableResult
+    mutating func closeScope() -> [ThrowsClauseSyntax]? {
+        pop()
+    }
+}
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRuleExamples.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRuleExamples.swift
@@ -1,0 +1,258 @@
+internal struct UnneededThrowsRuleExamples {
+    static let nonTriggeringExamples = [
+        Example("""
+            func foo() throws {
+                try bar()
+            }
+            """),
+        Example("""
+            func foo() throws {
+                throw Example.failure
+            }
+            """),
+        Example("""
+            func foo(_ bar: () throws -> T) rethrows -> Int {
+                try items.map { try bar() }
+            }
+            """),
+        Example("""
+            func foo() {
+                func bar() throws {
+                    try baz()
+                }
+                try? bar()
+            }
+            """),
+        Example("""
+            protocol Foo {
+                func bar() throws
+            }
+            """),
+        Example("""
+            func foo() throws {
+                guard false else {
+                    throw Example.failure
+                }
+            }
+            """),
+        Example("""
+            func foo() throws {
+                do { try bar() } 
+                catch {
+                    throw Example.failure
+                }
+            }
+            """),
+        Example("""
+            func foo() throws {
+                do { try bar() } 
+                catch {
+                    try baz()
+                }
+            }
+            """),
+        Example("""
+        func foo() throws {
+            do {
+                throw Example.failure
+            } catch {
+                do {
+                    throw Example.failure
+                } catch {
+                    throw Example.failure
+                }
+            }
+        }
+        """),
+        Example("""
+            func foo() throws {
+                switch bar {
+                case 1: break
+                default: try bar()
+                }
+            }
+            """),
+        Example("""
+            var foo: Int {
+                get throws {
+                    try bar
+                }
+            }
+            """),
+    ]
+
+    static let triggeringExamples = [
+        Example("func foo() ↓throws {}"),
+        Example("let foo: () ↓throws -> Void = {}"),
+        Example("let foo: (() ↓throws -> Void)? = {}"),
+        Example("func foo(bar: () throws -> Void) ↓rethrows {}"),
+        Example("init() ↓throws {}"),
+        Example("""
+            func foo() ↓throws {
+                bar()
+            }
+            """),
+        Example("""
+            func foo() {
+                func bar() ↓throws {}
+                bar()
+            }
+            """),
+        Example("""
+            func foo() {
+                func bar() ↓throws {
+                    baz()
+                }
+                bar()
+            }
+            """),
+        Example("""
+            func foo() {
+                func bar() ↓throws {
+                    baz()
+                }
+                try? bar()
+            }
+            """),
+        Example("""
+            func foo() ↓throws {
+                func bar() ↓throws {
+                    baz()
+                }
+            }
+            """),
+        Example("""
+            func foo() ↓throws {
+                do { try bar() } 
+                catch {}
+            }
+            """),
+        Example("""
+            func foo() ↓throws {
+                do {} 
+                catch {}
+            }
+            """),
+        Example("""
+            func foo() {
+                do {
+                    func bar() ↓throws {}
+                    try bar()
+                } catch {}
+            }
+            """),
+        Example("""
+            func foo() ↓throws {
+                do {
+                    try bar()
+                    func baz() throws { try bar() }
+                    try baz()
+                } catch {}
+            }
+            """),
+        Example("""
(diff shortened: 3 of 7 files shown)


Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+import SwiftSyntax
+
+@SwiftSyntaxRule(correctable: true, optIn: true)
+struct UnneededThrowsRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "unneeded_throws_rethrows",
+        name: "Unneeded (re)throws keyword",
+        description: "Non-throwing functions/variables should not me marked as `throws` or `rethrows`",
```
msrutek-paylocity, line 11:
```suggestion
        description: "Non-throwing functions/variables should not be marked as `throws` or `rethrows`",
```

msrutek-paylocity (approved):

msrutek-paylocity:
@SimplyDanny Any thoughts on this? :)

msrutek-paylocity:
@tonyskansf It seems that some conflicts have arisen, would you mind resolving them? Thanks!

CHANGELOG.md
```diff
@@ -12,7 +12,9 @@
 
 ### Enhancements
 
-* None.
+* Add opt-in `unneeded_throws_rethrows` rules that triggers when declarations
+  marked `throws`/`rethrows` never actually throw or call any throwing code.
```

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+import SwiftSyntax
+
+@SwiftSyntaxRule(correctable: true, optIn: true)
+struct UnneededThrowsRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "unneeded_throws_rethrows",
+        name: "Unneeded (re)throws keyword",
+        description: "Non-throwing functions/variables should not be marked as `throws` or `rethrows`",
```
SimplyDanny:
```suggestion
        description: "Non-throwing functions/properties/closures should not be marked as `throws` or `rethrows`.",
```

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+        override func visitPost(_: ThrowStmtSyntax) {
+            scopes.markCurrentScopeAsThrowing()
+        }
+
+        private func validateScope(_ scope: Scope, reason: String) {
+            guard let throwsToken = scope.throwsClause?.throwsSpecifier else { return }
+            violations.append(
+                ReasonedRuleViolation(
+                    position: throwsToken.positionAfterSkippingLeadingTrivia,
+                    reason: reason,
```
SimplyDanny:
```suggestion
                    reason: "Superfluous 'throws'; " + reason,
```

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+                scopes.openScope(with: functionTypeSyntax.effectSpecifiers?.throwsClause)
+            }
+            return .visitChildren
+        }
+
+        override func visitPost(_: FunctionTypeSyntax) {
+            if let closedScope = scopes.closeScope() {
+                validateScope(
+                    closedScope,
+                    reason: "The closure type does not throw any error"
```
SimplyDanny: To keep it short:
```suggestion
                    reason: "closure type does not throw any error"
```

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+
+private extension UnneededThrowsRule {
+    struct Scope {
+        var throwsClause: ThrowsClauseSyntax?
+    }
+
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        private var scopes = Stack<Scope>()
+
+        override func visit(_: ProtocolDeclSyntax) -> SyntaxVisitorContinueKind {
```
SimplyDanny: As a shortcut, you can use `skippableDeclarations` to ignore declarations you don't need to consider.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+            if node.questionOrExclamationMark == nil {
+                scopes.markCurrentScopeAsThrowing()
+            }
+        }
+
+        override func visitPost(_: ThrowStmtSyntax) {
+            scopes.markCurrentScopeAsThrowing()
+        }
+
+        private func validateScope(_ scope: Scope, reason: String) {
```
SimplyDanny:
```suggestion
        private func validate(scope: Scope, reason: String) {
```

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+            }
+        }
+
+        override func visitPost(_ node: DoStmtSyntax) {
+            let doesNotContainCatchClauseWithoutPattern = !node.catchClauses.contains { catchClause in
+                catchClause.catchItems.isEmpty
+            }
+            if doesNotContainCatchClauseWithoutPattern {
+                scopes.markCurrentScopeAsThrowing()
+            }
```
SimplyDanny:
```suggestion
            if node.catchClauses.contains(where: { $0.catchItems.isEmpty }) {
                // All errors will be caught.
                return
            }
            scopes.markCurrentScopeAsThrowing()
```

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        override func visit(_ node: ForStmtSyntax) -> SyntaxVisitorContinueKind {
```
SimplyDanny: Could be `visitPost` too.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRuleExamples.swift
```diff
@@ -0,0 +1,328 @@
...
+            """),
+        Example("""
+            var foo: Int {
+                get throws {
+                    try bar
+                }
+            }
+            """),
+        Example("""
+        func foo() throws{
```
SimplyDanny:
```suggestion
        func foo() throws {
```

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRuleExamples.swift
```diff
@@ -0,0 +1,328 @@
...
+        typealias Foo = () throws -> Void
+        """),
+        Example("""
+        enum Foo {
+            case foo
+            case bar(() throws -> Void)
+        }
+        """),
+        Example("""
+        func foo() async rethrows {
```
SimplyDanny: `rethrows` needs a throwing function argument, otherwise it's invalid Swift. I see the intention of testing `rethrows` as well without much fuss. Yet, compilable Swift code is preferred.

Tests/IntegrationTests/default_rule_configurations.yml
```diff
@@ -1262,6 +1262,11 @@ unneeded_synthesized_initializer:
   meta:
     opt-in: false
     correctable: true
+unneeded_throws:
+  severity: warning
+  meta:
+    opt-in: true
+    correctable: false
```
SimplyDanny: The rule is actually `correctable`.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRuleExamples.swift
```diff
@@ -0,0 +1,328 @@
...
+            }
+            """),
+        Example("""
+        func foo() ↓throws {
+            do { try bar() } 
+            catch Example.failure {}
+            catch is SomeError {}
+            catch {}
+        }
+        """),
```
SimplyDanny, line 307: There should also be an example (especially a correction) making use of "typed throws".

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+            child.as(DeclReferenceExprSyntax.self)?.baseName.tokenKind == .identifier("Task")
+        }
+    }
+}
+
+private extension PatternBindingSyntax {
+    var containsInitializerClause: Bool {
+        children(viewMode: .sourceAccurate).contains { child in
+            child.is(InitializerClauseSyntax.self)
+        }
```
SimplyDanny:
```suggestion
        initializer != nil
```
Is that not possible?

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    var functionTypeSyntax: FunctionTypeSyntax? {
+        guard let typeAnnotation else { return nil }
+
+        var children = Set(typeAnnotation.children(viewMode: .sourceAccurate))
+
+        while let child = children.popFirst() {
```
SimplyDanny: That's dangerous. Consider this example:
```swift
let s: S<() throws -> Void> = S()
```
I think, you should be explicit about the syntax structure to match instead of iterating through all children.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+
+        private func validateScope(_ scope: Scope, reason: String) {
+            guard let throwsToken = scope.throwsClause?.throwsSpecifier else { return }
+            violations.append(
+                ReasonedRuleViolation(
+                    position: throwsToken.positionAfterSkippingLeadingTrivia,
+                    reason: reason,
+                    correction: ReasonedRuleViolation.ViolationCorrection(
+                        start: throwsToken.positionAfterSkippingLeadingTrivia,
+                        end: throwsToken.endPosition,
```
SimplyDanny: You need to use `endPositionBeforeTrailingTrivia` here to avoid removing trailing trivia.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+
+        override func visit(_ node: FunctionCallExprSyntax) -> SyntaxVisitorContinueKind {
+            if node.containsTaskDeclaration {
+                scopes.openScope()
+            }
+            return .visitChildren
+        }
+
+        override func visitPost(_ node: FunctionCallExprSyntax) {
+            if node.containsTaskDeclaration {
```
SimplyDanny: I don't understand this exclusion of tasks to be honest. Could you explain?

SimplyDanny (changes requested):
Looks very good already! You thought about many special cases already. The rule also makes sense. But adding it as opt-in is important as it can cause false positives in the context of protocol implementations (and maybe more occasions).

I've added a few comments. Please also check the findings reported by the OSS scan.

SimplyDanny:
You need to rebase onto `main` first to fix conflicts and build with the latest infrastructure. Ignore the conflict in `GeneratedTests.swift` and let the file be deleted. Eventually, run `swift run swiftlint-dev rules register` again.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
tonyskansf: Can you please give an example for which this would behave incorrectly? 

Tests are not passing if I change end position to the suggested.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
tonyskansf: Yes, thanks!

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
tonyskansf:
```swift
// triggering
func foo() ↓throws {
  Task {
    try bar()
  }
}

// non-triggering
func foo() throws -> Bar? {
  Bar(try baz())
}
```

If my memory is right, it is to handle cases like these ones

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
tonyskansf: I see that for you example it is incorrectly triggering. 

>be explicit about the syntax structure to match instead of iterating through all children

I'll try to come up with a better solution for these cases
- `let s: S<() throws -> Void> = S()`
- `let foo: () ↓throws -> Void = {}`

Thanks for pointing this out

tonyskansf: Please check out 25c4905.

Instead of iterating over all children, this implementation checks the `PatternBindingSyntax.typeAnnotation` type for only specific cases - that should be more explicit about which syntax structure we're interested in.

Please let me know if this is better approach or suggest other.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRuleExamples.swift
tonyskansf: Added ff94069. Using `scope.throwsClause` seems enough.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
SimplyDanny: `func f() throws /* comment */ {}`

CHANGELOG.md
```diff
@@ -82,6 +82,9 @@
   
 * Support deinitializers and subscripts in `function_body_length` rule.  
   [SimplyDanny](https://github.com/SimplyDanny)
+* Add opt-in `unneeded_throws_rethrows` rules that triggers when declarations
```
SimplyDanny: Missing newline.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
```diff
@@ -0,0 +1,214 @@
...
+                scopes.openScope(with: functionTypeSyntax.effectSpecifiers?.throwsClause)
+            }
+            return .visitChildren
+        }
+
+        override func visitPost(_: FunctionTypeSyntax) {
+            if let closedScope = scopes.closeScope() {
+                validate(
+                    scope: closedScope,
+                    reason: "closure type does not throw any error"
```
SimplyDanny: The "does not throw any error" part could be moved into the `validate` method.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
SimplyDanny: Looks better. `TupleTypeSyntax` is only relevant when the tuple has only one element, isn't it. Otherwise, it's hard to check the necessity of `throws` keywords.

Also, we should have test case for the cases in the switch of the types.

tonyskansf: That's true. The necessity of `throws` keywords could depend on a context we don't know.

I'll keep the check for single-element `TupleTypeSyntax` only. Making the rule less strict.

I've also added new test cases.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
tonyskansf: Should be fixed now 3416302ad6.

I also had to adjust start of the correction though.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
tonyskansf: Refactored to be part of the validate method.

c5cd3db

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
tonyskansf: Please check 84b7081

tonyskansf:
@SimplyDanny 
Sorry it took me a while to find time for the review comments. I've addressed all of them. I've gone through all the available warnings and identified one false-positive which I fixed in 032538b.

Please can you take a look again. Appreciate your time!

Thank you

CHANGELOG.md
```diff
@@ -83,6 +83,10 @@
 * Support deinitializers and subscripts in `function_body_length` rule.  
   [SimplyDanny](https://github.com/SimplyDanny)
 
+* Add opt-in `unneeded_throws_rethrows` rules that triggers when declarations
```
```suggestion
* Add opt-in `unneeded_throws_rethrows` rule that triggers when declarations
```

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
```diff
@@ -0,0 +1,220 @@
+import SwiftLintCore
+import SwiftSyntax
+
+@SwiftSyntaxRule(correctable: true, optIn: true)
+struct UnneededThrowsRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "unneeded_throws_rethrows",
+        name: "Unneeded (re)throws keyword",
```
SimplyDanny:
```suggestion
        name: "Unneeded (Re)Throws Keyword",
```

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
```diff
@@ -0,0 +1,220 @@
...
+                    validate(
+                        scope: closedScope,
+                        construct: "closure type"
+                    )
+                }
+            }
+        }
+
+        override func visit(_ node: FunctionCallExprSyntax) -> SyntaxVisitorContinueKind {
+            if node.containsTaskDeclaration {
```
SimplyDanny: Thinking again about this case ... it's not only relevant for tasks but for closures passed to functions in general, isn't it? So if we have

```swift
func h() throws {
  f(1) {
    try g()
  }
}
```

currently this doesn't trigger, yet it should.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRuleExamples.swift
```diff
@@ -0,0 +1,382 @@
...
+            case bar(() throws -> Void)
+        }
+        """),
+        Example("""
+        func foo() async throws {
+            for try await item in items {}
+        }
+        """),
+        Example("""
+        let foo: () throws -> Void
```
SimplyDanny, line 116: This would trigger

```swift
let foo: () throws -> Void = bar
```

but shouldn't as we don't know what `bar` exactly is.

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRule.swift
tonyskansf: Should be now covered with af24d81

Source/SwiftLintBuiltInRules/Rules/Lint/UnneededThrowsRuleExamples.swift
tonyskansf: Should be now covered with 844a710

PR #6082, opened 2025-05-19
Initial implementation of allowed_types option for `one-declaration-per-file` rule

### LINKED ISSUE #6072: `one_declaration_per_file`: Allow certain types in same file
### New Issue Checklist

Add an array property (tentatively called `allowed_types`) for `one_declaration_per_file` to allow certain types in same the file as other types. e.g.:

```yaml
one_declaration_per_file:
  allowed_types: [enum, protocol]
```

snofla:
WIP: This is going to resolve #6072

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/OneDeclarationPerFileRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/OneDeclarationPerFileRule.swift
@@ -2,8 +2,7 @@ import SwiftSyntax
 
 @SwiftSyntaxRule(optIn: true)
 struct OneDeclarationPerFileRule: Rule {
-    var configuration = SeverityConfiguration<Self>(.warning)
-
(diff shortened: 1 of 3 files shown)


SimplyDanny:
I wonder what the meaning of this option should be. Does it mean "these types do not count at all" or is it like "files with only these types are ignored"? In other words, with `allowed_types = ["struct"]`, would
```swift
class C {}
struct S {}
```
trigger or not?

snofla:
@SimplyDanny It would not trigger. A better name for the option would be: 

```
additionally_allowed_types
```

but a another, perhaps more correct but less verbose name escapes me right now :)

SimplyDanny:
`ignored_types`, `disregarded_types` or `skipped_types` (from the rule's point of view) would be other options.

SimplyDanny:
Before you continue, please rebase your PR onto the most recent `main` branch.

PR #6088, opened 2025-05-30
Update non_optional_string_data_conversion rule docs and examples to make it more clear

SergeyPekar:
**Motivation:**

The rule `non_optional_string_data_conversion` is bidirectional it is triggered when converting string to data as well as when converting from data to string so this behaviour is needed to be clearly documented.

**Changes:**

- Updated the description to clearly show that the rule will be triggered in both scenarios with the references to preferred initialisers
- Updated triggering and non-triggering examples

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/NonOptionalStringDataConversionRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/NonOptionalStringDataConversionRule.swift
@@ -5,14 +5,16 @@ struct NonOptionalStringDataConversionRule: Rule {
     var configuration = SeverityConfiguration<Self>(.warning)
     static let description = RuleDescription(
         identifier: "non_optional_string_data_conversion",
-        name: "Non-optional String -> Data Conversion",
-        description: "Prefer non-optional `Data(_:)` initializer when converting `String` to `Data`",
+        name: "Non-optional String <-> Data Conversion",
+        description: "Prefer the non-optional initializers when converting between `String` and `Data` (e.g. `Data(_:)` and `String(decoding:as:)`)",
         kind: .lint,
         nonTriggeringExamples: [
-            Example("Data(\"foo\".utf8)")
+            Example("Data(\"foo\".utf8)"),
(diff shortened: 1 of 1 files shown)


SergeyPekar:
> Thanks for the contribution!
> 
> However, this rule doesn't actually work in both directions. There are `optional_data_string_conversion` and `non_optional_string_data_conversion` which separate both cases from each other.



> Thanks for the contribution!
> 
> However, this rule doesn't actually work in both directions. There are `optional_data_string_conversion` and `non_optional_string_data_conversion` which separate both cases from each other.
Screenshot 2025-06-05 at 22 50 36


Yes. Looks like it was a bug of 0.55.1 (see the screenshot) that is already fixed

Source/SwiftLintBuiltInRules/Rules/Lint/NonOptionalStringDataConversionRule.swift
```diff
@@ -5,14 +5,16 @@ struct NonOptionalStringDataConversionRule: Rule {
     var configuration = SeverityConfiguration<Self>(.warning)
     static let description = RuleDescription(
         identifier: "non_optional_string_data_conversion",
-        name: "Non-optional String -> Data Conversion",
-        description: "Prefer non-optional `Data(_:)` initializer when converting `String` to `Data`",
+        name: "Non-optional String <-> Data Conversion",
+        description: "Prefer the non-optional initializers when converting between `String` and `Data` (e.g. `Data(_:)` and `String(decoding:as:)`)",
```
SimplyDanny, line 9: If you are still interested in providing extended documentation for this rule, please let the `description` property as it is and prefer the optional `rationale` property. It serves exactly this purpose.

SimplyDanny:
@SergeyPekar: Do you still fancy to add more documentation? If not, please close the PR at will.

SergeyPekar:
> @SergeyPekar: Do you still fancy to add more documentation? If not, please close the PR at will.

I will take a look during the week and close the PR if I won't come up with anything fancy

PR #6092, opened 2025-06-14
Add support for regex pattern in the  option of the  rule.

### LINKED ISSUE #6066: Support regex pattern matching for `file_name` `excluded` array in configuration file
### New Issue Checklist

The problem I am facing: I want to entirely disable the `file_name` rule for all Swift files in a codebase that ends with "Types". I guess this is not only specific to `file_name` **but all rules that have an `excluded` field**. I tried:

1. `.*Types\\.swift`
2. `Types`
3. `**/*Types`
4. `**/*Types.swift`

The examples included in [rules documentation](https://realm.github.io/SwiftLint/file_name.html) for the `excluded` field does not have any patterns, only hard-coded file names.

The top-level `excluded` array in a configuration file supports pattern matching (this field is for all the files swiftlint should ignore entirely).

Ueeek:
Resolve #6066

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
@@ -13,11 +13,10 @@ struct FileNameRule: OptInRule, SourceKitFreeRule {
 
     func validate(file: SwiftLintFile) -> [StyleViolation] {
         guard let filePath = file.path,
-            !shouldExclude(filePath: filePath),
-            case let fileName = filePath.bridge().lastPathComponent,
-            !configuration.excluded.contains(fileName) else {
+              !configuration.shouldExclude(filePath: filePath) else {
             return []
         }
+        let fileName = filePath.bridge().lastPathComponent
 
         let prefixRegex = regex("\\A(?:\(configuration.prefixPattern))")
         let suffixRegex = regex("(?:\(configuration.suffixPattern))\\z")
@@ -56,16 +55,6 @@ struct FileNameRule: OptInRule, SourceKitFreeRule {
                 location: Location(file: filePath, line: 1)
             ),
         ]
-    }
-}
-
-// MARK: - For `excluded` option
-private extension FileNameRule {
-    func shouldExclude(filePath: String) -> Bool {
-        configuration.excluded.compactMap {
-            let regex = try? RegularExpression(pattern: "^\($0)$")
-            return regex?.regex.firstMatch(in: filePath, options: [], range: filePath.fullNSRange)
-        }.isNotEmpty
     }
 }
 
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
@@ -1,3 +1,4 @@
+import SwiftLintCore
 import SwiftSyntax
 
 struct FileNameRule: OptInRule, SourceKitFreeRule {
@@ -12,6 +13,7 @@ struct FileNameRule: OptInRule, SourceKitFreeRule {
 
     func validate(file: SwiftLintFile) -> [StyleViolation] {
         guard let filePath = file.path,
+            !shouldExclude(filePath: filePath),
             case let fileName = filePath.bridge().lastPathComponent,
             !configuration.excluded.contains(fileName) else {
             return []
@@ -54,6 +56,16 @@ struct FileNameRule: OptInRule, SourceKitFreeRule {
                 location: Location(file: filePath, line: 1)
             ),
         ]
+    }
+}
+
+// MARK: - For `excluded` option
+private extension FileNameRule {
+    func shouldExclude(filePath: String) -> Bool {
+        configuration.excluded.compactMap {
+            let regex = try? RegularExpression(pattern: "^\($0)$")
+            return regex?.regex.firstMatch(in: filePath, options: [], range: filePath.fullNSRange)
+        }.isNotEmpty
     }
 }
 
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
@@ -15,11 +15,11 @@ struct FileNameRule: OptInRule, SourceKitFreeRule {
               !configuration.shouldExclude(filePath: filePath) else {
             return []
         }
-        let fileName = filePath.bridge().lastPathComponent
 
         let prefixRegex = regex("\\A(?:\(configuration.prefixPattern))")
         let suffixRegex = regex("(?:\(configuration.suffixPattern))\\z")
 
+        let fileName = filePath.bridge().lastPathComponent
         var typeInFileName = fileName.bridge().deletingPathExtension
 
         // Process prefix
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
@@ -1,4 +1,3 @@
-import SwiftLintCore
 import SwiftSyntax
 
 struct FileNameRule: OptInRule, SourceKitFreeRule {
--- a/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/FileNameConfiguration.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/FileNameConfiguration.swift
@@ -17,3 +17,19 @@ struct FileNameConfiguration: SeverityBasedRuleConfiguration {
     @ConfigurationElement(key: "require_fully_qualified_names")
     private(set) var requireFullyQualifiedNames = false
 }
+
+// MARK: - For `excluded` option
+extension FileNameConfiguration {
+    func shouldExclude(filePath: String) -> Bool {
+        let fileName = filePath.bridge().lastPathComponent
+        
+        // For backwards compatibility, `excluded` can have a fileName which is invalid as regex.(e.g. "NSString+Extension.swift")
+        if excluded.contains(fileName) {
+            return true
+        }
+        
+        return excluded.compactMap {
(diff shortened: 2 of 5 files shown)


Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
```diff
@@ -12,6 +13,7 @@ struct FileNameRule: OptInRule, SourceKitFreeRule {
 
     func validate(file: SwiftLintFile) -> [StyleViolation] {
         guard let filePath = file.path,
+            !shouldExclude(filePath: filePath),
             case let fileName = filePath.bridge().lastPathComponent,
             !configuration.excluded.contains(fileName) else {
```
SimplyDanny: Did you intentionally keep this check? I think it's a good way to ensure backwards compatibility even though an exclude name that is not a regular expression would match in `shouldExclude` already. We could make this part of `shouldExclude` then, probably the first check in there.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
```diff
@@ -57,6 +59,16 @@ struct FileNameRule: OptInRule, SourceKitFreeRule {
     }
 }
 
+// MARK: - For `excluded` option
+private extension FileNameRule {
+    func shouldExclude(filePath: String) -> Bool {
```
SimplyDanny: Let's make this a method of the configuration itself.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
```diff
@@ -57,6 +59,16 @@ struct FileNameRule: OptInRule, SourceKitFreeRule {
...
+        configuration.excluded.compactMap {
+            let regex = try? RegularExpression(pattern: "^\($0)$")
```
SimplyDanny: We could make the creation of regular expressions part of the configuration (`@ConfigurationElement` have a `postprocessor` argument) that can be used for that. By creating the regex in the configuration parsing process already, we ensure that they are correct and can report errors. Here, we would just ignore the exclude if regex creation fails.

SimplyDanny (changes requested):
Thank you for the implementation!

Please consider my remarks mainly about error handling in case of invalid exclude patterns.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
Ueeek: Yes, I keep "configuration.excluded.contains(fileName)" for backwards compatibility.
It's good idea to put it in `shouldExclude()` together.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/FileNameRule.swift
Ueeek: @SimplyDanny Thank you for suggestion!
 But, have doubt on it. In previous logic, users can pass something like "NSString+Extension.swift" into `excluded` params, which is invalid as Regex since `+` is not escaped.
As it will cause backwards-compatibility issue, I prefer not creating regex in the config parsing.
Or, should it be acceptable breaking-change? 🤔 
(Please correct me if I'm wrong :bow:)

SimplyDanny: That's a very valid example. You are right. The method should compare strings first to be backwards compatible and only afterwards try to match them as regular expressions.

And if the conversion to regex fails, we just assume it isn't supposed to
be a regex and just ignore it. There isn't more we can do.

Note that you can now also use the `Regex` library from Swift since macOS 13 is the lowest platform we support.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/FileNameConfiguration.swift
```diff
@@ -17,3 +17,20 @@ struct FileNameConfiguration: SeverityBasedRuleConfiguration {
...
+        let fileName = filePath.bridge().lastPathComponent
+
+        // For backwards compatibility,
+        // `excluded` can have a fileName which is invalid as regex.(e.g. "NSString+Extension.swift")
+        if excluded.contains(fileName) {
+            return true
+        }
+
+        return excluded.compactMap {
+            try? Regex("^\($0)$").firstMatch(in: filePath)
```
SimplyDanny: I was a bit too exited about the usage of `Regex`. At the moment, it's still better to go with `NSRegularExpression.cached(...)` because that would cache regex compilation.

Sorry for that!

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/FileNameConfiguration.swift
```diff
@@ -17,3 +17,20 @@ struct FileNameConfiguration: SeverityBasedRuleConfiguration {
...
+    func shouldExclude(filePath: String) -> Bool {
+        let fileName = filePath.bridge().lastPathComponent
+
+        // For backwards compatibility,
+        // `excluded` can have a fileName which is invalid as regex.(e.g. "NSString+Extension.swift")
+        if excluded.contains(fileName) {
+            return true
+        }
+
+        return excluded.compactMap {
```
SimplyDanny: I think `.contains {...}` is better here since that would stop once a match is found and not collect all matches first like mapping does.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/FileNameConfiguration.swift
```diff
@@ -17,3 +17,20 @@ struct FileNameConfiguration: SeverityBasedRuleConfiguration {
...
     private(set) var requireFullyQualifiedNames = false
 }
+
+// MARK: - For `excluded` option
+extension FileNameConfiguration {
+    func shouldExclude(filePath: String) -> Bool {
+        let fileName = filePath.bridge().lastPathComponent
+
+        // For backwards compatibility,
+        // `excluded` can have a fileName which is invalid as regex.(e.g. "NSString+Extension.swift")
```
SimplyDanny:
```suggestion
        // For backwards compatibility, `excluded` could contain a simple file name 
        // which might be invalid as a regex (e.g. `NSString+Extension.swift`).
```

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/FileNameConfiguration.swift
SimplyDanny: I wonder if we should really match against the whole file path. Until now, only file names are checked. The regex approach should probably do the same.

If that's a use case, we can still add that later. For now, I think that names are sufficient.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/FileNameConfiguration.swift
```diff
@@ -17,3 +17,20 @@ struct FileNameConfiguration: SeverityBasedRuleConfiguration {
...
+        if excluded.contains(fileName) {
```
SimplyDanny, line 26: I see a problem here. Your example `NSString+Extension.swift` wouldn't match against  `NSStringExtension.swift` here but then later in the regex-based match. That's undesired. 😶

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/FileNameConfiguration.swift
Ueeek: Ah, I also missed that point. Let me use existing one.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/FileNameConfiguration.swift
Ueeek: @SimplyDanny Good catch, thank you 🙇 .

From inputted `exclude`, we cannot distinguish whether a user intends `+` means `regex "+"` or `fileName "+"`.
Some possible options I came up with are
* Accept this behavior since it's not a false alerts.
* Add one new configuration param something like `exclude_pattern`, in addition to existing `exclude` param.

The second option seems better since it's working properly (though, the num of param increased) 🤔

SimplyDanny: > * Accept this behavior since it's not a false alerts.

It could cause false negatives though.

> * Add one new configuration param something like `exclude_pattern`, in addition to existing `exclude` param.

I also think that's the better solution. Unfortunately, it leads to yet another option. I was hoping to just extend existing functionality by supporting regex. But that doesn't seem to be possible without changing semantics.

If we add a new option anyway, we can also have it support complete path patterns. Then we not only support regex but also more specific exclusions (for whatever that's good for). With that, the name should include the "path" aspect like `exclude_path_patterns`.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/FileNameConfiguration.swift
SimplyDanny: You can just call `regex(...)`. It will do the caching and set a few default options.

Ueeek: It will crash with `force-try` if we pass any invalid regex. (e.g. "(")
Feels better to initialize `regex` in a safer way (`try? RegularExpression(pattern:)`)

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/FileNameConfiguration.swift
SimplyDanny: I wasn't so happy with the option's name and now went with `excluded_paths`. As a next step we could think about deprecating `excluded` to rename it to `excluded_file_names` which separates it better from the new exclude option working with paths.

SimplyDanny:
Rebased to take new required build steps into account.

SimplyDanny (approved):
Should be good to go now. Thanks again for all the work, @Ueeek!

Ueeek:
@SimplyDanny Thank you so much for your review and help😁😁

PR #6100, opened 2025-06-18
Migrate FileLengthRule from SourceKit to SwiftSyntax

jpsim:
Convert FileLengthRule to use SwiftSyntax instead of SourceKit for improved performance and fewer false positives.

- Uses ViolationsSyntaxVisitor pattern with token traversal
- Correctly handles multiline tokens by counting all spanned lines
- Properly excludes comment-only and whitespace-only lines
- Adds isWhitespace extension for TriviaPiece to support exclusion logic
- Accurately attributes trivia to correct line positions
- Extracts common line range logic to reduce code duplication

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Metrics/FileLengthRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Metrics/FileLengthRule.swift
@@ -61,13 +61,6 @@ private extension FileLengthRule {
 
                 // Process leading trivia
                 addTriviaLines(token.leadingTrivia, startingAt: token.position, to: &linesWithActualContent)
-
-                // Process trailing trivia
-                addTriviaLines(
-                    token.trailingTrivia,
-                    startingAt: token.endPositionBeforeTrailingTrivia,
-                    to: &linesWithActualContent
-                )
             }
             return linesWithActualContent.count
         }
@@ -104,28 +97,6 @@ private extension FileLengthRule {
             for line in startLine...endLine {
                 lines.insert(line)
             }
-        }
-    }
-}
-
-private extension TriviaPiece {
-    var isComment: Bool {
-        switch self {
-        case .lineComment, .blockComment, .docLineComment, .docBlockComment:
-            return true
-        default:
-            return false
-        }
-    }
-
-    var isWhitespace: Bool {
-        switch self {
-        case .spaces, .tabs, .newlines, .carriageReturns, .carriageReturnLineFeeds, .formfeeds, .verticalTabs:
-            return true
-        case .unexpectedText(let text):
-            return text.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
-        default:
-            return false
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Source/SwiftLintBuiltInRules/Rules/Metrics/FileLengthRule.swift
```diff
@@ -17,38 +18,114 @@ struct FileLengthRule: Rule {
...
+    var isComment: Bool {
+        switch self {
+        case .lineComment, .blockComment, .docLineComment, .docBlockComment:
+            return true
+        default:
+            return false
+        }
+    }
+
+    var isWhitespace: Bool {
```
pennig: `isComment` and `isWhitespace` are already members of `TriviaPiece` https://swiftpackageindex.com/swiftlang/swift-syntax/601.0.1/documentation/swiftsyntax/triviapiece

jpsim:
OSSCheck results show that this is mostly significantly faster, with no meaningful changes to the results, although there's lots of noise because of the addition of the column to the position (`:1`).

image

Source/SwiftLintBuiltInRules/Rules/Metrics/FileLengthRule.swift
```diff
@@ -17,38 +18,114 @@ struct FileLengthRule: Rule {
...
-        func lineCountWithoutComments() -> Int {
-            let commentKinds = SyntaxKind.commentKinds
-            return file.syntaxKindsByLines.filter { kinds in
-                !Set(kinds).isSubset(of: commentKinds)
-            }.count
+                // Process leading trivia
+                addTriviaLines(token.leadingTrivia, startingAt: token.position, to: &linesWithActualContent)
+
+                // Process trailing trivia
+                addTriviaLines(
```
pennig: trailing trivia doesn't contain newlines so we can probably skip this bit

Source/SwiftLintBuiltInRules/Rules/Metrics/FileLengthRule.swift
```diff
@@ -17,38 +18,114 @@ struct FileLengthRule: Rule {
...
+            let violation = ReasonedRuleViolation(
+                position: violationPosition,
+                reason: reason,
+                severity: severity
+            )
+            violations.append(violation)
+        }
+
+        private func countNonCommentLines(in node: SourceFileSyntax) -> Int {
+            var linesWithActualContent = Set<Int>()
```
pennig, line 57: consider `RangeSet`

jpsim: `RangeSet` is macOS 15+ but SwiftLint supports macOS 13.5+: https://github.com/realm/SwiftLint/blob/66930722ada01f630d4971be9260794bd6ad6ced/.bazelrc#L11

https://developer.apple.com/documentation/swift/rangeset

Source/SwiftLintBuiltInRules/Rules/Metrics/FileLengthRule.swift
jpsim: 8c2f2f8

Source/SwiftLintBuiltInRules/Rules/Metrics/FileLengthRule.swift
jpsim: 8c2f2f8

PR #6103, opened 2025-06-19
Migrate VerticalWhitespaceRule from SourceKit to SwiftSyntax

pennig:
Convert VerticalWhitespaceRule to use SwiftSyntax instead of SourceKit. It improves correctness while maintaining performance.

* Will now catch newlines interspersed with whitespace
* Operates only on structured elements of `leadingTrivia`, avoiding the need for regular expressions or logic to ignore false positives inside of block comments.
* Modifies only the affected trivia when correcting, instead of the entire file line-by-line.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@
   and fewer false positives.  
   [JP Simard](https://github.com/jpsim)
 
+* Migrate `vertical_whitespace` rule from SourceKit to SwiftSyntax for improved performance.
+  [Matt Pennig](https://github.com/pennig)
+
 ### Bug Fixes
 
 * Improved error reporting when SwiftLint exits, because of an invalid configuration file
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,9 @@
   and fewer false positives.  
   [JP Simard](https://github.com/jpsim)
 
+* Migrate `vertical_whitespace` rule from SourceKit to SwiftSyntax for improved performance.
+  [Matt Pennig](https://github.com/pennig)
```
jpsim, line 26: nit:

> 2 trailing spaces on the last line describing the change (so that Markdown renders each change [on its own line](https://daringfireball.net/projects/markdown/syntax#p)).

From https://github.com/realm/SwiftLint/blob/main/CONTRIBUTING.md#tracking-changes

jpsim (approved):
Thank you so much for the contribution! Slowly burning through the remaining SourceKit rules, only 20 left out of 239!!

CHANGELOG.md
jpsim: I'll fix this up after merging to save you the trouble.

jpsim: https://github.com/realm/SwiftLint/pull/6105

pennig: Thanks for the reminder (and the fix) re: the changelog

SimplyDanny:
@pennig: I found that the new implementation doesn't trigger anymore on too many empty lines at the top of a file, e.g. it doesn't trigger in [this line](https://github.com/kickstarter/ios-oss/blob/373155a92c21748af8bf6dbd6fce94899dd43583/KsApi%2Fmodels%2FDiscoveryEnvelope.swift#L2). Basically all the "This PR fixed a violation in Kickstarter" lines in the OSS report above.

~~Could you have a look into that?~~ Addressed by #6141.

PR #6111, opened 2025-06-20
Migrate LineLengthRule from SourceKit to SwiftSyntax

jpsim:
Convert LineLengthRule to use SwiftSyntax instead of SourceKit for
improved performance and better detection accuracy.

- Uses ViolationsSyntaxVisitor pattern with line-based validation
- Pre-computes ignored lines using helper visitors for efficiency
- Implements pure SwiftSyntax comment detection without SourceKit
- Correctly handles function declarations, multiline strings
- Maintains all configuration options including URL stripping
- Preserves exact line position reporting for violations


jpsim:
The OSSCheck reports are due to there being a column where previously it was only row in the reported location.

pennig (commented):
Breaking out the configuration-based Visitors makes the code super easy to read, but I wonder if we could eke out a bit more speed if those were combined into one Visitor. There's no overlap among their own visitPost methods, so the code would still be relatively easy to read.

On the other hand, it's already notably faster than the SourceKit approach as-is, so maybe it's fine.

jpsim:
> On the other hand, it's already notably faster than the SourceKit approach as-is

My thinking exactly! I'd rather spend cycles getting us off SourceKit than micro-optimizing specific SwiftSyntax rules further.

Source/SwiftLintBuiltInRules/Rules/Metrics/LineLengthRule.swift
```diff
@@ -24,149 +23,224 @@ struct LineLengthRule: Rule {
...
-                lineHasKinds(line: line,
-                             kinds: functionKinds,
-                             kindsByLine: swiftDeclarationKindsByLine.value) {
-                return nil
+        override func visit(_ node: SourceFileSyntax) -> SyntaxVisitorContinueKind {
+            // Populate functionDeclarationLines if ignores_function_declarations is true
+            if configuration.ignoresFunctionDeclarations {
+                let funcVisitor = FunctionLineVisitor(locationConverter: locationConverter)
+                funcVisitor.walk(node)
+                functionDeclarationLines = funcVisitor.lines
```
SimplyDanny:
```suggestion
                let funcVisitor = FunctionLineVisitor(locationConverter: locationConverter)
                functionDeclarationLines = funcVisitor.walk(tree: node, handler: \.lines)
```

Source/SwiftLintBuiltInRules/Rules/Metrics/LineLengthRule.swift
```diff
@@ -24,149 +23,224 @@ struct LineLengthRule: Rule {
...
-               lineIsMultilineString(line, file: file, syntaxKindsByLine: syntaxKindsByLine.value) {
-                return nil
+            // Populate commentOnlyLines if ignores_comments is true
+            if configuration.ignoresComments {
+                commentOnlyLines = findCommentOnlyLines(in: node, file: file, locationConverter: locationConverter)
             }
 
-            for pattern in configuration.excludedLinesPatterns where line.containsMatchingPattern(pattern) {
-                return nil
+            return .skipChildren // We'll do the main processing in visitPost
```
SimplyDanny, line 59: Processing could be done right here as well. But the separation of information collection and processing is actually quite nice. Doing everything in one function would spare all the `*Lines` properties, though.

Source/SwiftLintBuiltInRules/Rules/Metrics/LineLengthRule.swift
```diff
@@ -24,149 +23,224 @@ struct LineLengthRule: Rule {
...
+        }
+
+        override func visitPost(_ node: SourceFileSyntax) {  // swiftlint:disable:this unused_parameter
```
SimplyDanny:
```suggestion
        override func visitPost(_: SourceFileSyntax) {
```

Source/SwiftLintBuiltInRules/Rules/Metrics/LineLengthRule.swift
```diff
@@ -24,149 +23,224 @@ struct LineLengthRule: Rule {
...
+            let minLengthThreshold = configuration.params.map(\.value).min() ?? .max
+
+            for line in file.lines {
+                // Quick check to skip very short lines before expensive stripping
+                // `line.content.count` <= `line.range.length` is true.
+                // So, `check line.range.length` is larger than minimum parameter value.
+                // for avoiding using heavy `line.content.count`.
```
SimplyDanny:
```suggestion
                // So, check `line.range.length` is larger than minimum parameter value
                // for avoiding using heavy `line.content.count`.
```

Source/SwiftLintBuiltInRules/Rules/Metrics/LineLengthRule.swift
```diff
@@ -24,149 +23,224 @@ struct LineLengthRule: Rule {
...
+                if configuration.ignoresMultilineStrings && multilineStringLines.contains(line.index) {
+                    continue
+                }
+                if configuration.excludedLinesPatterns.contains(where: {
+                    regex($0).firstMatch(in: line.content, range: line.content.fullNSRange) != nil
+                }) {
+                    continue
+                }
+
+                // String stripping logic (remains largely the same as it's string-based)
```
SimplyDanny: That's helpful for review, but in a few days we might wonder what this is supposed to mean. 😅

Source/SwiftLintBuiltInRules/Rules/Metrics/LineLengthRule.swift
```diff
@@ -24,149 +23,224 @@ struct LineLengthRule: Rule {
...
+                    ))
+                    break // Only report one violation (the most severe one reached) per line
+                }
             }
+        }
 
-            var strippedString = line.content
-            if configuration.ignoresURLs {
-                strippedString = strippedString.strippingURLs
+        // This helper remains largely the same, as it's string manipulation.
```
SimplyDanny: See previous comment.

Source/SwiftLintBuiltInRules/Rules/Metrics/LineLengthRule.swift
```diff
@@ -24,149 +23,224 @@ struct LineLengthRule: Rule {
...
+              (node.openingPounds != nil && node.openingQuote.tokenKind == .stringQuote) else {
+            return
+        }
+
+        let startLocation = locationConverter.location(for: node.positionAfterSkippingLeadingTrivia)
+        let endLocation = locationConverter.location(for: node.endPositionBeforeTrailingTrivia)
 
-    init(_ computation: @escaping @autoclosure () -> Result) {
-        self.computation = computation
+        if startLocation.line < endLocation.line {
```
SimplyDanny: Is this check necessary? I guess the range is just empty otherwise.

SimplyDanny (commented):
LGTM!

Source/SwiftLintBuiltInRules/Rules/Metrics/LineLengthRule.swift
```diff
@@ -1,13 +1,12 @@
 import Foundation
 import SourceKittenFramework
```
SimplyDanny: Can be removed.

Source/SwiftLintBuiltInRules/Rules/Metrics/LineLengthRule.swift
jpsim: Going to keep this as-is.

PR #6112, opened 2025-06-20
Migrate FileHeaderRule from SourceKit to SwiftSyntax

jpsim:
## Summary

Convert FileHeaderRule to use SwiftSyntax instead of SourceKit for
improved performance and better handling of file header comments,
shebangs, and doc comments.

## Key Technical Improvements

- **Enhanced shebang support** properly skipping past `#!/usr/bin/env swift` lines
- **Better comment type discrimination** excluding doc comments from header analysis
- **Accurate position calculation** converting between UTF-8 and UTF-16 offsets for regex matching
- **Improved trivia traversal** for comprehensive header comment collection
- **SwiftLint command filtering** to exclude directive comments from header content

## Migration Details

- Replaced `OptInRule` with `@SwiftSyntaxRule(optIn: true)` annotation
- Implemented `ViolationsSyntaxVisitor` pattern for file-level analysis
- Added logic to start header collection after shebang.endPosition if present
- Distinguished between regular comments and doc comments (///, /** */)
- Maintained UTF-16 offset calculations for NSRegularExpression compatibility
- Added `skipDisableCommandTests: true` for SwiftSyntax disable command behavior
- Removed unnecessary SourceKittenFramework import


jpsim:
All the OSSCheck results are catching legitimate violations in DuckDuckGo because the file headers used in its Swift files don't match the required pattern set in its `.swiftlint.yml` file. In other words, the previous rule was broken.

PR #6116, opened 2025-06-20
Migrate StatementPositionRule from SourceKit to SwiftSyntax

jpsim:
Convert StatementPositionRule to use SwiftSyntax instead of SourceKit for
improved performance and more accurate positioning validation.

- Uses ViolationsSyntaxVisitor and ViolationsSyntaxRewriter patterns
- Validates else/catch keyword positioning with proper indentation checks
- Supports both default (same line) and uncuddled (next line) modes
- Correctly handles trailing/leading trivia for whitespace validation
- Implements corrections using explicit rewriter for both statement types
- Extracts common validation logic to reduce code duplication
- Skip correcting when there are comments

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRule.swift
@@ -129,7 +129,7 @@ private extension StatementPositionRule {
 
             violations.append(
                 ReasonedRuleViolation(
-                    position: result.previousToken.positionAfterSkippingLeadingTrivia,
+                    position: keyword.positionAfterSkippingLeadingTrivia,
                     reason: description.description,
                     severity: configuration.severity
                 )
--- a/Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRuleExamples.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRuleExamples.swift
@@ -67,60 +67,60 @@ struct StatementPositionRuleExamples {
 
     static let triggeringExamples = [
         // Single line examples
-        Example("if true { ↓}else if false { }"),
-        Example("if true { ↓}  else { }"),
+        Example("if true { }↓else if false { }"),
+        Example("if true { }  ↓else { }"),
         Example("""
-        do { ↓}
-        catch { }
+        do { }
+        ↓catch { }
         """),
         Example("""
         do {
             let a = 1
-        ↓}
-        \t  catch { }
+        }
+        \t  ↓catch { }
         """),
         // Multi-line examples
         Example("""
         if true {
             foo()
-        ↓}else {
+        }↓else {
             bar()
         }
         """),
         Example("""
         if true {
             foo()
-        ↓}    else if true {
+        }    ↓else if true {
             bar()
         }
         """),
         Example("""
         if true {
             foo()
-        ↓}
-        else {
+        }
+        ↓else {
             bar()
         }
         """),
         Example("""
         do {
             foo()
-        ↓}catch {
+        }↓catch {
             bar()
         }
         """),
         Example("""
         do {
             foo()
-        ↓}    catch {
+        }    ↓catch {
             bar()
         }
         """),
         Example("""
         do {
             foo()
-        ↓}
-        catch {
+        }
+        ↓catch {
             bar()
         }
         """),
@@ -128,17 +128,17 @@ struct StatementPositionRuleExamples {
         Example("""
         if true {
             foo()
-        ↓} // comment
-        else {
+        } // comment
+        ↓else {
             bar()
         }
         """),
         Example("""
         do {
             foo()
-        ↓}
+        }
         // comment
-        catch {
+        ↓catch {
             bar()
         }
         """),
@@ -147,23 +147,23 @@ struct StatementPositionRuleExamples {
     static let corrections = [
         // Single line examples
         Example("""
-        if true { ↓}
-         else { }
+        if true { }
+         ↓else { }
         """): Example("if true { } else { }"),
         Example("""
-        if true { ↓}
-           else if false { }
+        if true { }
+           ↓else if false { }
         """): Example("if true { } else if false { }"),
         Example("""
-        do { ↓}
-         catch { }
+        do { }
+         ↓catch { }
         """): Example("do { } catch { }"),
         // Multi-line examples
         Example("""
         if true {
             foo()
-        ↓}
-        else {
+        }
+        ↓else {
             bar()
         }
         """): Example("""
@@ -176,7 +176,7 @@ struct StatementPositionRuleExamples {
         Example("""
         if true {
             foo()
-        ↓}   else if true {
+        }   ↓else if true {
             bar()
         }
         """): Example("""
@@ -189,8 +189,8 @@ struct StatementPositionRuleExamples {
         Example("""
         do {
             foo()
-        ↓}
-        catch {
+        }
+        ↓catch {
             bar()
         }
         """): Example("""
@@ -203,7 +203,7 @@ struct StatementPositionRuleExamples {
         Example("""
         do {
             foo()
-        ↓}    catch {
+        }    ↓catch {
             bar()
         }
         """): Example("""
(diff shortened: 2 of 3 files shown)


Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRule.swift
```diff
@@ -36,175 +20,217 @@ struct StatementPositionRule: CorrectableRule {
...
 
-    func correct(file: SwiftLintFile) -> Int {
-        switch configuration.statementMode {
-        case .default:
-            defaultCorrect(file: file)
-        case .uncuddledElse:
-            uncuddledCorrect(file: file)
+    private static func calculateIndentation(_ trivia: Trivia) -> Int {
+        var indentation = 0
+        for piece in trivia.reversed() {
```
SimplyDanny, line 62: Why reversed?

Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRule.swift
```diff
@@ -36,175 +20,217 @@ struct StatementPositionRule: CorrectableRule {
...
+            var newCatchClauses: [CatchClauseSyntax] = []
+            var needsBodyUpdate = false
+
+            for catchClause in node.catchClauses {
+                if let corrected = correctCatchStatement(
+                    catchClause: catchClause,
+                    keyword: catchClause.catchKeyword
+                ) {
+                    newCatchClauses.append(corrected)
+                    needsBodyUpdate = true
```
SimplyDanny: Isn't this only required when the first `catch` keyword needs correction?

Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRule.swift
```diff
@@ -36,175 +20,217 @@ struct StatementPositionRule: CorrectableRule {
...
+
+            let validation = StatementValidation(keyword: keyword, previousToken: previousToken)
+            let needsCorrection = configuration.statementMode == .default ?
+                !validation.isValidForDefaultMode() :
+                !validation.isValidForUncuddledMode()
+
+            guard needsCorrection else { return nil }
+
+            // Skip correction if there are comments between brace and keyword
+            guard !validation.hasCommentsBetween else { return nil }
```
SimplyDanny: Could go into a method in one or the other way to be used here, in `correctIfStatement` and `validateStatement` in the visitor.

Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRuleExamples.swift
```diff
@@ -0,0 +1,322 @@
...
+        do {
+            foo()
+        ↓}
+        catch {
+            bar()
+        }
+        """),
+    ]
+
+    static let corrections = [
```
SimplyDanny, line 147: Misses examples that proof that corrections aren't applied if there are between closing brace and keyword.

SimplyDanny:
Wondering about the fixed violations in the Swift repository. The code doesn't look like it shouldn't trigger the rule anymore as both `else` are on lines below the closing brace.

Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRule.swift
jpsim: To calculate the indentation of the *last* line in the trivia, not the first. I'll add a comment.

Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRule.swift
jpsim: I think you're right.

Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRuleExamples.swift
jpsim: Our correction infra for generated tests doesn't support corrections that are skipped, but I'll add some triggering examples that wouldn't be corrected.

jpsim:
> Wondering about the fixed violations in the Swift repository. The code doesn't look like it shouldn't trigger the rule anymore as both `else` are on lines below the closing brace.

Not sure if you're saying this is a good thing or a bad thing.

Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRuleExamples.swift
SimplyDanny: In these cases, I typically add a correction pair where both examples are equal to show and test that nothing changes.

SimplyDanny:
> > Wondering about the fixed violations in the Swift repository. The code doesn't look like it shouldn't trigger the rule anymore as both `else` are on lines below the closing brace.
> 
> Not sure if you're saying this is a good thing or a bad thing.

Oh sorry, I think that these examples should still trigger a violation and not appear as "fixed" with the rewrite.

Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRule.swift
SimplyDanny: Without the `reversed()`, all tests still pass. So we should probably have a test case that makes sure reversed order is required.

jpsim:
> Oh sorry, I think that these examples should still trigger a violation and not appear as "fixed" with the rewrite.

Well really it's that these two violations merged and moved to a different location:

## Before

> This PR fixed a violation in Swift: [/stdlib/public/core/SwiftNativeNSArray.swift:401:7](https://github.com/apple/swift/blob/17ae36e67c5079ab77cc2b4db1f9872c0f86baa0/stdlib%2Fpublic%2Fcore%2FSwiftNativeNSArray.swift#L401): Warning: Else and catch should be on the same line, one space after the previous declaration (statement_position)

> This PR fixed a violation in Swift: [/stdlib/public/core/SwiftNativeNSArray.swift:452:7](https://github.com/apple/swift/blob/17ae36e67c5079ab77cc2b4db1f9872c0f86baa0/stdlib%2Fpublic%2Fcore%2FSwiftNativeNSArray.swift#L452): Warning: Else and catch should be on the same line, one space after the previous declaration (statement_position)

## After

> This PR introduced a violation in Swift: [/stdlib/public/core/SwiftNativeNSArray.swift:393:7](https://github.com/apple/swift/blob/17ae36e67c5079ab77cc2b4db1f9872c0f86baa0/stdlib%2Fpublic%2Fcore%2FSwiftNativeNSArray.swift#L393): Warning: Else and catch should be on the same line, one space after the previous declaration (statement_position)

SimplyDanny:
> ## Before
> > This PR fixed a violation in Swift: [/stdlib/public/core/SwiftNativeNSArray.swift:401:7](https://github.com/apple/swift/blob/17ae36e67c5079ab77cc2b4db1f9872c0f86baa0/stdlib%2Fpublic%2Fcore%2FSwiftNativeNSArray.swift#L401): Warning: Else and catch should be on the same line, one space after the previous declaration (statement_position)
> 
> > This PR fixed a violation in Swift: [/stdlib/public/core/SwiftNativeNSArray.swift:452:7](https://github.com/apple/swift/blob/17ae36e67c5079ab77cc2b4db1f9872c0f86baa0/stdlib%2Fpublic%2Fcore%2FSwiftNativeNSArray.swift#L452): Warning: Else and catch should be on the same line, one space after the previous declaration (statement_position)
> 
> ## After
> > This PR introduced a violation in Swift: [/stdlib/public/core/SwiftNativeNSArray.swift:393:7](https://github.com/apple/swift/blob/17ae36e67c5079ab77cc2b4db1f9872c0f86baa0/stdlib%2Fpublic%2Fcore%2FSwiftNativeNSArray.swift#L393): Warning: Else and catch should be on the same line, one space after the previous declaration (statement_position)

But in my understanding, all three lines should still trigger, because the keyword is always after the closing brace.

Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRuleExamples.swift
jpsim: Added tests and supporting infrastructure for this in 7e8763688.

jpsim:
> But in my understanding, all three lines should still trigger, because the keyword is always after the closing brace.

This was really a matter of violation location, but I think it makes more sense to report violations on the `else`/`catch` keyword rather than the preceding closing brace. I just pushed up that change, and now the `SwiftNativeNSArray.swift` violations are the same before and after this migration. The previous rule's violation locations were inconsistent.

PR #6119, opened 2025-06-20
Migrate VerticalWhitespaceOpeningBracesRule to SwiftSyntax

jpsim:
## Summary

Convert VerticalWhitespaceOpeningBracesRule to use SwiftSyntax instead
of SourceKit for improved performance and better detection of empty
lines after opening braces.

## Key Technical Improvements

- **Enhanced trivia analysis** for accurate empty line detection after
  opening braces
- **Proper handling of closure "in" keywords** with context-aware
  detection
- **Improved correction logic** handling all newline types (LF, CR,
  CRLF) consistently
- **SwiftSyntax visitor pattern** replacing regex-based detection for
  better accuracy
- **Comprehensive token analysis** supporting all opening brace types
  (`{`, `[`, `(`)
- **Accurate position tracking** for violation start and end positions

## Migration Details

- Replaced `CorrectableRule, OptInRule` with
  `@SwiftSyntaxRule(correctable: true, optIn: true)`
- Implemented `ViolationsSyntaxVisitor` for detecting violations in
  token trailing trivia
- Implemented `ViolationsSyntaxRewriter` for correcting violations
  across different syntax contexts
- Added proper handling of closures, code blocks, arrays, and tuples
- Maintained exact position reporting for violation locations
- Preserved all existing test cases and rule behavior

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Style/VerticalWhitespaceOpeningBracesRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/VerticalWhitespaceOpeningBracesRule.swift
@@ -1,12 +1,8 @@
-import Foundation
-import SourceKittenFramework
+// swiftlint:disable file_length
+import SwiftLintCore
+import SwiftSyntax
 
-private extension SwiftLintFile {
-    func violatingRanges(for pattern: String) -> [NSRange] {
-        match(pattern: pattern, excludingSyntaxKinds: SyntaxKind.commentAndStringKinds)
-    }
-}
-
+@SwiftSyntaxRule(correctable: true, optIn: true)
 struct VerticalWhitespaceOpeningBracesRule: Rule {
     var configuration = SeverityConfiguration<Self>(.warning)
 
@@ -22,6 +18,20 @@ struct VerticalWhitespaceOpeningBracesRule: Rule {
 
             }
         */
+        """),
+        Example("""
+        func foo() {
+            // This is a comment
+
+            let x = 5
+        }
+        """),
+        Example("""
+        if condition {
+            // Comment explaining the logic
+
+            performAction()
+        }
         """),
     ]
 
@@ -135,12 +145,32 @@ struct VerticalWhitespaceOpeningBracesRule: Rule {
               })
             }
             """),
+        Example("""
+        func foo() {
+        ↓
+            // This is a comment
+            let x = 5
+        }
+        """): Example("""
+            func foo() {
+                // This is a comment
+                let x = 5
+            }
+            """),
+        Example("""
+        if condition {
+        ↓
+            // Comment explaining the logic
+            performAction()
+        }
+        """): Example("""
+            if condition {
+                // Comment explaining the logic
+                performAction()
+            }
+            """),
     ]
 
-    private let pattern = "([{(\\[][ \\t]*(?:[^\\n{]+ in[ \\t]*$)?)((?:\\n[ \\t]*)+)(\\n)"
-}
-
-extension VerticalWhitespaceOpeningBracesRule: OptInRule {
     static let description = RuleDescription(
         identifier: "vertical_whitespace_opening_braces",
         name: "Vertical Whitespace after Opening Braces",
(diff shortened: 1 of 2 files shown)


SimplyDanny:
Rebased to get the better summary.

SimplyDanny:
The difference to the previous implementation are examples like

```swift
func f() { // comment

}
```

with comments after the opening brace. They trigger now.

Another difference is when the `in` keyword separating a closure's arguments from its body is on a line after the opening brace, like in:

```swift
f { foo,
    bar in

}
```

The previous implementation didn't trigger on the empty line, the new implementation does.

I think that both cases are acceptable without introducing an option since there are only ~20 examples of either nature in the scanned repositories.

Source/SwiftLintBuiltInRules/Rules/Style/VerticalWhitespaceOpeningBracesRule.swift
```diff
@@ -150,43 +180,371 @@ extension VerticalWhitespaceOpeningBracesRule: OptInRule {
...
+        switch tokenKind {
+        case .leftBrace, .leftParen, .leftSquare:
+            return true
+        default:
+            return false
+        }
+    }
+}
+
+private extension TriviaPiece {
```
SimplyDanny, line 541: We should move this to `SwiftSyntax+SwiftLint.swift` and reused it in other rules.

Source/SwiftLintBuiltInRules/Rules/Style/VerticalWhitespaceOpeningBracesRule.swift
```diff
@@ -150,43 +180,371 @@ extension VerticalWhitespaceOpeningBracesRule: OptInRule {
...
+            }
+            totalAdvance += bytesPerNewline
+        }
+
+        return (consecutiveNewlines, totalAdvance)
+    }
+}
+
+private extension VerticalWhitespaceOpeningBracesRule {
+    final class Visitor: ViolationsSyntaxVisitor<SeverityConfiguration<VerticalWhitespaceOpeningBracesRule>> {
```
SimplyDanny, line 323:
```suggestion
    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
```

Source/SwiftLintBuiltInRules/Rules/Style/VerticalWhitespaceOpeningBracesRule.swift
```diff
@@ -150,43 +180,371 @@ extension VerticalWhitespaceOpeningBracesRule: OptInRule {
...
+            for piece in trivia {
+                let positionAdvance = result.processTriviaPiece(piece, currentPosition: currentPosition)
+                currentPosition = currentPosition.advanced(by: positionAdvance)
+            }
+
+            return result
+        }
+    }
+
+    final class Rewriter: ViolationsSyntaxRewriter<SeverityConfiguration<VerticalWhitespaceOpeningBracesRule>> {
```
SimplyDanny, line 408:
```suggestion
    final class Rewriter: ViolationsSyntaxRewriter<ConfigurationType> {
```

Source/SwiftLintBuiltInRules/Rules/Style/VerticalWhitespaceOpeningBracesRule.swift
```diff
@@ -150,43 +180,371 @@ extension VerticalWhitespaceOpeningBracesRule: OptInRule {
...
+                let correctedTrivia = correctTrivia(trivia: leadingTrivia)
+                if correctedTrivia.hasCorrections {
+                    numberOfCorrections += correctedTrivia.correctionCount
+                    var newStatements = node.statements
+                    newStatements[newStatements.startIndex] = firstStatement
+                        .with(\.leadingTrivia, correctedTrivia.trivia)
+                    return node.with(\.statements, newStatements)
+                }
+            }
+            return super.visit(node)
```
SimplyDanny, line 422: These blocks look all very similar. Common parts could be extracted into a method.

Source/SwiftLintBuiltInRules/Rules/Style/VerticalWhitespaceOpeningBracesRule.swift
```diff
@@ -135,12 +145,32 @@ struct VerticalWhitespaceOpeningBracesRule: Rule {
...
+        ↓
+            // Comment explaining the logic
+            performAction()
+        }
+        """): Example("""
+            if condition {
+                // Comment explaining the logic
+                performAction()
+            }
+            """),
```
SimplyDanny, line 171: I tried with another example: 

```swift
if x == 5 { // comment
    
    print("x is 5")
}
```

The violation triggers at the comment and the correction removes on of the leading slashes `//`.

SimplyDanny (commented):
It's a bit irritating that the rule triggers on the opening brace in all cases except for closures where it triggers on the empty line. We should harmonize that (probably in a separate change afterwards).

Source/SwiftLintBuiltInRules/Rules/Style/VerticalWhitespaceOpeningBracesRule.swift
SimplyDanny, line 408: The rewriter is currently not used at all since the visitor does the replacement already and the rule doesn't specify `explicitRewriter: true`.

PR #6125, opened 2025-06-21
Remove dead code

jpsim:
In particular lots of stuff that used to be needed with SourceKit that we no longer need to keep around.

Identified using Periphery: https://github.com/peripheryapp/periphery


SimplyDanny (approved):
Periphery! 🥳

PR #6128, opened 2025-06-21
Add `RegexConfiguration.ExecutionMode`

jpsim:
To help migrate custom rules to SwiftSyntax. Not wired up yet, just the
configuration parsing and defaults. Wiring it up in https://github.com/realm/SwiftLint/pull/6129.

The diff looks big, but it's 500+ lines of tests, with ~45 lines of
actually new code.

### DIFF
--- a/Source/SwiftLintCore/RuleConfigurations/RegexConfiguration.swift
+++ b/Source/SwiftLintCore/RuleConfigurations/RegexConfiguration.swift
@@ -4,6 +4,11 @@ import SourceKittenFramework
 /// A rule configuration used for defining custom rules in yaml.
 public struct RegexConfiguration<Parent: Rule>: SeverityBasedRuleConfiguration, Hashable,
                                                 CacheDescriptionProvider, InlinableOptionType {
+    /// The execution mode for this custom rule.
+    public enum ExecutionMode: String, Codable, Sendable {
+        case swiftsyntax
+        case sourcekit
+    }
     /// The identifier for this custom rule.
     public let identifier: String
     /// The name for this custom rule.
@@ -24,6 +29,8 @@ public struct RegexConfiguration<Parent: Rule>: SeverityBasedRuleConfiguration, 
     public var severityConfiguration = SeverityConfiguration<Parent>(.warning)
     /// The index of the regex capture group to match.
     public var captureGroup = 0
+    /// The execution mode for this rule (nil means use global default).
+    public var executionMode: ExecutionMode?
 
     public var cacheDescription: String {
         let jsonObject: [String] = [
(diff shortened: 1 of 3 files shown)


Source/SwiftLintCore/RuleConfigurations/RegexConfiguration.swift
```diff
@@ -24,6 +31,8 @@ public struct RegexConfiguration<Parent: Rule>: SeverityBasedRuleConfiguration,
     public var severityConfiguration = SeverityConfiguration<Parent>(.warning)
     /// The index of the regex capture group to match.
     public var captureGroup = 0
+    /// The execution mode for this rule (nil means use global default).
```
SimplyDanny: Instead of `nil`, what about a default case in the enum?

Source/SwiftLintFramework/Rules/CustomRules.swift
```diff
@@ -7,23 +7,42 @@ struct CustomRulesConfiguration: RuleConfiguration, CacheDescriptionProvider {
...
+        }
+
+        for (key, value) in configurationDict where key != "default_execution_mode" {
             var ruleConfiguration = RegexConfiguration<Parent>(identifier: key)
 
             do {
                 try ruleConfiguration.apply(configuration: value)
+
+                // Apply default execution mode if the rule doesn't specify its own
+                if ruleConfiguration.executionMode == nil {
```
SimplyDanny: Can be moved after the `do-catch` block.

Source/SwiftLintFramework/Rules/CustomRules.swift
```diff
@@ -50,15 +69,19 @@ struct CustomRules: Rule, CacheDescriptionProvider, ConditionallySourceKitFree {
         name: "Custom Rules",
         description: """
             Create custom rules by providing a regex string. Optionally specify what syntax kinds to match against, \
-            the severity level, and what message to display.
+            the severity level, and what message to display. Rules default to SwiftSyntax mode for improved \
+            performance. Use `mode: sourcekit` or `default_execution_mode: sourcekit` for SourceKit mode.
```
SimplyDanny: If there's `default_execution_mode`, should we use `execution_mode` instead of `mode`?

Tests/FrameworkTests/CustomRulesTests.swift
```diff
@@ -512,6 +512,387 @@ final class CustomRulesTests: SwiftLintTestCase {
...
+        let configDict = [
+            "regex": "pattern",
+            "mode": "swiftsyntax",
+        ]
+
+        var regexConfig = Configuration(identifier: "test_rule")
+        do {
+            try regexConfig.apply(configuration: configDict)
+            XCTAssertEqual(regexConfig.executionMode, .swiftsyntax)
+        } catch {
```
SimplyDanny: I'd prefer a `throws` on the test function instead of catching the error and triggering a fail explicitly. The result is the same – a failing test.

SimplyDanny (approved):

PR #6130, opened 2025-06-21
Add caching to Swift linting job

SimplyDanny:
Let's see if that really makes reporting lint results faster. Currently it takes almost 6 minutes.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,11 +13,20 @@ jobs:
   lint-swift:
     name: Lint Swift
     runs-on: ubuntu-24.04 # "Noble Numbat"
-    container: swift:6.1-noble
     steps:
       - uses: actions/checkout@v4
+      - name: Create ci.bazelrc file
+        run: echo "$CI_BAZELRC_FILE_CONTENT" | base64 -d > ci.bazelrc
+        env:
+          CI_BAZELRC_FILE_CONTENT: ${{ secrets.CI_BAZELRC_FILE_CONTENT }}
       - name: Lint
-        run: swift run swiftlint --reporter github-actions-logging --strict 2> /dev/null
+        run: |
+          export PATH="/usr/share/swift/usr/bin:$PATH"
+          git apply --ignore-whitespace .bcr/patches/no-warnings-as-errors.patch
+          bazel build --config release //:swiftlint
+          ./bazel-bin/swiftlint lint --reporter github-actions-logging --strict 2> /dev/null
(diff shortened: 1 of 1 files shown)


SimplyDanny:
Down to 4 minutes. Not very impressive ...

.github/workflows/lint.yml
```diff
@@ -13,11 +13,18 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
+      - name: Create ci.bazelrc file
+        run: echo "$CI_BAZELRC_FILE_CONTENT" | base64 -d > ci.bazelrc
+        env:
+          CI_BAZELRC_FILE_CONTENT: ${{ secrets.CI_BAZELRC_FILE_CONTENT }}
       - name: Lint
-        run: swift run swiftlint --reporter github-actions-logging --strict 2> /dev/null
+        run: |
+          bazel build //:swiftlint
```
jpsim:
```suggestion
          bazel build --config release //:swiftlint
```

.github/workflows/lint.yml
```diff
@@ -13,11 +13,19 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
+      - name: Create ci.bazelrc file
+        run: echo "$CI_BAZELRC_FILE_CONTENT" | base64 -d > ci.bazelrc
+        env:
+          CI_BAZELRC_FILE_CONTENT: ${{ secrets.CI_BAZELRC_FILE_CONTENT }}
       - name: Lint
-        run: swift run swiftlint --reporter github-actions-logging --strict 2> /dev/null
+        run: |
+          export PATH="/usr/share/swift/usr/bin:$PATH"
```
jpsim, line 24: You'll need a Swift Docker container because Swift isn’t part of the standard Ubuntu Docker image.

SimplyDanny: Swift 6.1.2 is listed in their docu: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#language-and-runtime

With the Docker container, there's no Bazel.

jpsim: Ah got it these are the GHA images, not Docker.

SimplyDanny:
Down to 47 seconds. That's something. 👍

Since this now builds SwiftLint on Linux with Bazel, we can remove the explicit job doing the same on Azure or make it use the remote cache as well.

Thank you for suggesting the Bazel build, @jpsim!

PR #6131, opened 2025-06-23
Migrate MultilineParametersBracketsRule from SourceKit to SwiftSyntax

jpsim:
## Summary

Convert MultilineParametersBracketsRule to use SwiftSyntax instead of
SourceKit for improved performance and better detection of multiline
parameter formatting violations.

## Key Technical Improvements

- **Enhanced multiline detection** distinguishing between structurally
  multiline parameters and parameters with multiline default values
- **Accurate position reporting** using SwiftSyntax's precise token
  locations
- **Better handling of default values** by focusing on parameter
  structure rather than content
- **Improved performance** using visitor pattern over regex-based
  SourceKit analysis
- **Reduced false positives** for single parameters with multiline
  default values

## Migration Details

- Replaced `ASTRule` with `@SwiftSyntaxRule(optIn: true)` annotation
- Implemented `ViolationsSyntaxVisitor` pattern for function and
  initializer parameter analysis
- Added helper methods to extract significant tokens (name/type)
  excluding default values
- Converted regex-based bracket detection to SwiftSyntax position
  comparisons
- Maintained full compatibility with existing rule behavior and test
  cases

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Style/MultilineParametersBracketsRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/MultilineParametersBracketsRule.swift
@@ -1,7 +1,8 @@
-import Foundation
-import SourceKittenFramework
+import SwiftLintCore
+import SwiftSyntax
 
-struct MultilineParametersBracketsRule: OptInRule {
+@SwiftSyntaxRule(optIn: true)
+struct MultilineParametersBracketsRule: Rule {
     var configuration = SeverityConfiguration<Self>(.warning)
 
     static let description = RuleDescription(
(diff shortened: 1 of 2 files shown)


jpsim:
An analysis of the OSS Check results from Claude:

> The OSS Check results show that all violations are message changes only. The
> violations are still correctly detected at the same locations, but with more
> descriptive messages:
> 
> * Old: "Multiline parameters should have their surrounding brackets in a new
>   line"
> * New: "Opening parenthesis should be on a separate line when using
>   multiline parameters" OR "Closing parenthesis should be on a separate line
>   when using multiline parameters"
> 
> This is an improvement because:
> 
> * Users get more specific guidance about which bracket needs fixing
> * The violation location is more precise
> * No false positives or false negatives were introduced

SimplyDanny:
> An analysis of the OSS Check results from Claude:
> 
> > The OSS Check results show that all violations are message changes only. The
> > violations are still correctly detected at the same locations, but with more
> > descriptive messages:
> > 
> > * Old: "Multiline parameters should have their surrounding brackets in a new
> >   line"
> > * New: "Opening parenthesis should be on a separate line when using
> >   multiline parameters" OR "Closing parenthesis should be on a separate line
> >   when using multiline parameters"
> > 
> > This is an improvement because:
> > 
> > * Users get more specific guidance about which bracket needs fixing
> > * The violation location is more precise
> > * No false positives or false negatives were introduced

That cannot be entirely true. The numbers of fixed and new violations is not equal. So there must be a few that are only fixed and a few that are only new and thus have not just their message changed.

The ones I found look valid though. I'm trying to improve the categorization further in #6139.

Source/SwiftLintBuiltInRules/Rules/Style/MultilineParametersBracketsRule.swift
```diff
@@ -89,97 +90,153 @@ struct MultilineParametersBracketsRule: OptInRule {
...
-        // find violations at deeper levels
-        for substructure in substructure.substructure {
-            violations += self.violations(in: substructure, file: file)
+        private func significantEndToken(for parameter: FunctionParameterSyntax) -> TokenSyntax? {
+            // End of ellipsis, or type, or name (in that order of preference)
+            if let ellipsis = parameter.ellipsis {
+                return ellipsis
+            }
+            // Type is not optional, so directly use it
+            return parameter.type.lastToken(viewMode: .sourceAccurate) // Gets the very last token of the type
```
SimplyDanny, line 119: Due to the `.type` an example like 
```swift
func foo(a: T,
         b: String
             = "")
```
would only trigger on `a` but not at the closing parenthesis where it actually should. Not sure if that's relevant, though, as it's a quite strange way of formatting parameters. However, for pretty long names and types and short lines, the default value could have to be placed on a new line like in the example. Probably depends on what the previous behavior was ...

SimplyDanny:
> > An analysis of the OSS Check results from Claude:
> > > The OSS Check results show that all violations are message changes only. The
> > > violations are still correctly detected at the same locations, but with more
> > > descriptive messages:
> > > 
> > > * Old: "Multiline parameters should have their surrounding brackets in a new
> > >   line"
> > > * New: "Opening parenthesis should be on a separate line when using
> > >   multiline parameters" OR "Closing parenthesis should be on a separate line
> > >   when using multiline parameters"
> > > 
> > > This is an improvement because:
> > > 
> > > * Users get more specific guidance about which bracket needs fixing
> > > * The violation location is more precise
> > > * No false positives or false negatives were introduced
> 
> That cannot be entirely true. The numbers of fixed and new violations is not equal. So there must be a few that are only fixed and a few that are only new and thus have not just their message changed.
> 
> The ones I found look valid though. I'm trying to improve the categorization further in #6139.

[oss-check-summary.md](https://github.com/user-attachments/files/20911668/oss-check-summary.md) is the report with improved categorization and this PR cherry-picked on top. There are indeed a few fixed and new violations that differ in more than the message.

PR #6132, opened 2025-06-23
Migrate Linux builds from Azure to GitHub Actions

SimplyDanny:
This allows for better reusability and integration. macOS builds still run on Azure due to GitHub's limitation of up to 5 macOS jobs running concurrently.


SimplyDanny:
We need to add branch protection rules once that's merged.

SimplyDanny:
> I don't love that we actively use so many CI platforms and providers, but ok.

Fair point. I mean, there are now 3 builds left on Azure. Jazzy could probably be moved to Linux as well. And with some caching, we could also make the macOS builds faster. Eventually, we would only have Buildkite and GitHub Actions as CI platforms left.

SimplyDanny:
> > I don't love that we actively use so many CI platforms and providers, but ok.
> 
> Fair point. I mean, there are now 3 builds left on Azure. Jazzy could probably be moved to Linux as well. And with some caching, we could also make the macOS builds faster. Eventually, we would only have Buildkite and GitHub Actions as CI platforms left.

Jobs migrated from Azure to GitHub Actions in #6149.

Issue #5611, opened 2024-05-29
void_function_in_ternary incorrectly triggered with if and switch expressions


tristanSIF:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### Describe the bug

```swift
func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, sizeForItemAt indexPath: IndexPath) -> CGSize {
    switch indexPath.section {
    // Void Function in Ternary Violation: Using ternary to call Void functions should be avoided (void_function_in_ternary)
    case 0: isEditing ? CGSize(width: 150, height: 20) : CGSize(width: 100, height: 20) 
    default: .zero
    }
}

func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, sizeForItemAt indexPath: IndexPath) -> CGSize {
    if indexPath.section == 0 {
        // Void Function in Ternary Violation: Using ternary to call Void functions should be avoided (void_function_in_ternary)
        isEditing ? CGSize(width: 150, height: 20) : CGSize(width: 100, height: 20)
    } else {
        .zero
    }
}

### Environment

* SwiftLint version: 0.55.1
* Installation method used:  Homebrew

**Comment by mathieulb** [comment]
I second this, I have the problem with a function (x: MyEnum) -> String when using it in both sides of a ternary in a case. When I use the same function on only one side, the warning is not triggered. E.g.:

```
return switch {
  case 1: b ? f(.hello) : f(.world) // spurious warning
  case 2: b ? f(.hello) : "world" // OK
  case 3: b ? "hello" : f(.world) // OK
}
```

**Comment by vadimbelyaev** [comment]
Just encountered it with the latest SwiftLint version to date (0.59.1).
Looks like this issue and the linked PR are abandoned.

Here's my minimal example that triggers a false positive:
```swift
struct MyStruct {
    let someField: Int
}

func makeMyStruct() -> MyStruct {
    if true {
        true ? MyStruct(someField: 0) : MyStruct(someField: 1) // Void Function in Ternary Violation: Using ternary to call Void functions should be avoided (void_function_in_ternary)
    } else {
        true ? MyStruct(someField: 2) : MyStruct(someField: 3) // Void Function in Ternary Violation: Using ternary to call Void functions should be avoided (void_function_in_ternary)
    }
}
```

[...]

PR #6133, opened 2025-06-23
Ignore various assignment operators in `void_function_in_ternary`

SimplyDanny:
Addresses https://github.com/realm/SwiftLint/issues/5611#issuecomment-2994077467.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@
   [imsonalbajaj](https://github.com/imsonalbajaj)
   [#6054](https://github.com/realm/SwiftLint/issues/6054)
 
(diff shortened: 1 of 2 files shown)


SimplyDanny:
@jpsim: Do you have a clue why the initial Danger scan is often so slow? It's only CI 1 where it is like that.

> Linting Aerial with this PR took 0.74 s vs 0.18 s on main (311% slower).

jpsim:
> Do you have a clue why the initial Danger scan is often so slow? It's only CI 1 where it is like that.

Probably operating system level caching would be my hand wavy answer.

Running the Aerial repo twice, discarding the first one, might be the easiest way to account for this.

PR #6134, opened 2025-06-24
Add warmup step to OSS check

SimplyDanny:
Avoid initial delay probably caused by OS level caching.


jpsim:
> Linting Aerial with this PR took 1.69 s vs 1.49 s on `main` (13% slower).

🤔

PR #6140, opened 2025-06-25
Report remaining fixed and new violations

SimplyDanny:
Remaining are all violations that have not been put into other categories by mapping them to each other.


SimplyDanny:
Seems to work. See #6139 testing it with an example change.

PR #6165, opened 2025-07-17
New Rule: Implicit Optional Initialisation

### LINKED ISSUE #1940: Rule Request: Implicit initialization of optional bindings
### New Issue Checklist

### Rule Request

> 1. Why should this rule be added? Share links to existing discussion about what

Swift supports both default implicit initialization of optional bindings and explicit.

These are equivalent:

```swift
struct A {
  var x: Int?
}

struct A {
  var x: Int? = nil
}
```

I think you should be able to enforce it either way, so either `"always"` or `"never"` implicit init of optional bindings.

For `"always"`, this would trigger a violation:

```swift
struct A {
  var x: Int? = nil
}
```

For `"never"`, this would trigger a violation:

```swift
struct A {
  var x: Int?
}
```

   rule as opt-in.

Opt-in as it's opinionated.

leo-lem:
## Add `implicit_optional_initialization` Rule

This PR introduces a new rule: `implicit_optional_initialization`, replacing the previous `redundant_optional_initialization` rule, which has been deprecated using the `deprecationAlias` field.

---

### About the new rule

The rule enforces whether optional variables should *always* be implicitly initialized (i.e., rely on Swift’s default behavior) or *never* by always adding `= nil`.  
The default is always (matching the old rule).  
The rule is **NOT** enabled by default.  

It is configurable via the `style` parameter:

```yaml
implicit_optional_initialization:
  severity: warning
  style: always # or never
```

---

### ✅ Non-triggering examples

```swift
let bar: String? = nil
lazy var baz: Bool? = nil

// style: always
var foo: Int?
var bar: Optional<String>

// style: never
var foo: Int? = nil
var bar: Optional<String> = nil
```

---

### 🔴 Triggering examples

```swift
// style: always
var foo: Int?↓ = nil

// style: never
var foo: Int?↓
```

---

### 🛠 Corrections

The rule is correctable in both directions depending on the `style`:

- Adds or removes `= nil` as appropriate
- Skips computed properties and closures

---

### Implementation notes

- Built using SwiftSyntax (`@SwiftSyntaxRule`)
- Includes both visitor and rewriter
- Fully tested and integrated into `default_rule_configurations.yml`
- Improves on `redundant_optional_initialization` with finer control.

---

Closes #1940

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRule.swift
@@ -49,15 +49,14 @@ private extension ImplicitOptionalInitializationRule {
                         \.initializer,
                         InitializerClauseSyntax(
                             equal: .equalToken(
-                                leadingTrivia: (node.typeAnnotation?.trailingTrivia.isEmpty ?? true)
-                                ? .space
-                                : Trivia(),
+                                leadingTrivia: .space,
                                 trailingTrivia: .space
                             ),
                             value: ExprSyntax(NilLiteralExprSyntax(nilKeyword: .keyword(.nil))),
                             trailingTrivia: node.typeAnnotation?.trailingTrivia ?? Trivia()
                         )
                     )
+                    .with(\.typeAnnotation, node.typeAnnotation?.with(\.trailingTrivia, Trivia()))
             case .always:
                 node
                     .with(\.initializer, nil)
--- a/Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRuleExamples.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRuleExamples.swift
@@ -38,7 +38,6 @@ enum ImplicitOptionalInitializationRuleExamples { // swiftlint:disable:this type
                 var myVar: String? = nil
             }
             """, configuration: ["style": "never"]),
-
         Example("var x: Int? = nil // comment", configuration: ["style": "never"]),  // with comment after
 
         // always style
@@ -100,10 +99,10 @@ enum ImplicitOptionalInitializationRuleExamples { // swiftlint:disable:this type
 
     static let corrections = [
         // never style
-        Example("var ↓myVar: Int? ", configuration: ["style": "never"]):
-            Example("var myVar: Int? = nil "),
-        Example("var ↓myVar: Optional<Int> ", configuration: ["style": "never"]):
-            Example("var myVar: Optional<Int> = nil "),
+        Example("var ↓myVar: Int? // comment", configuration: ["style": "never"]):
+            Example("var myVar: Int? = nil // comment"),
+        Example("var ↓myVar: Optional<Int> // comment", configuration: ["style": "never"]):
+            Example("var myVar: Optional<Int> = nil // comment"),
         Example(
             """
             var ↓myVar: String? {
@@ -129,10 +128,10 @@ enum ImplicitOptionalInitializationRuleExamples { // swiftlint:disable:this type
             }
             """),
 
-        Example("var ↓myVar: Int? = nil", configuration: ["style": "always"]):
-            Example("var myVar: Int?"),
-        Example("var ↓myVar: Optional<Int> = nil", configuration: ["style": "always"]):
-            Example("var myVar: Optional<Int>"),
+        Example("var ↓myVar: Int? = nil // comment", configuration: ["style": "always"]):
+            Example("var myVar: Int? // comment"),
+        Example("var ↓myVar: Optional<Int> = nil // comment", configuration: ["style": "always"]):
+            Example("var myVar: Optional<Int> // comment"),
         Example(
             """
             var ↓myVar: String? = nil {


SimplyDanny:
As a general remark: There is a way of deprecating an existing rule, adding `deprecatedAliases` to its description. That's better than removing it. This would require that the rewritten rule implements the existing behavior as its default and only adds new functionality.

SimplyDanny:
> As a general remark: There is a way of deprecating an existing rule, adding `deprecatedAliases` to its description. That's better than removing it. This would require that the rewritten rule implements the existing behavior as its default and only adds new functionality.

Are you going to merge the new rule into the existing one, @leo-lem?

leo-lem:
> > As a general remark: There is a way of deprecating an existing rule, adding `deprecatedAliases` to its description. That's better than removing it. This would require that the rewritten rule implements the existing behavior as its default and only adds new functionality.
> 
> Are you going to merge the new rule into the existing one, @leo-lem?

We could. The new rule does cover all cases of the old one. The difference is that it wouldn't be active by default.

I was going off of the comment to the issue:
> Note that the "never" case already exists as the [Redundant Optional Initialization Rule](https://github.com/realm/SwiftLint/blob/master/Rules.md#redundant-optional-initialization).
> 
> I'm ok if we want to rename this rule and make it configurable to both cases.

But I am also open to introducing it as a new style rule in addition to the old rule. What are your thoughts on this, @SimplyDanny?

SimplyDanny:
> > > As a general remark: There is a way of deprecating an existing rule, adding `deprecatedAliases` to its description. That's better than removing it. This would require that the rewritten rule implements the existing behavior as its default and only adds new functionality.
> 
> > 
> 
> > Are you going to merge the new rule into the existing one, @leo-lem?
> 
> 
> 
> We could. The new rule does cover all cases of the old one. The difference is that it wouldn't be active by default.
> 
> 
> 
> I was going off of the comment to the issue:
> 
> > Note that the "never" case already exists as the [Redundant Optional Initialization Rule](https://github.com/realm/SwiftLint/blob/master/Rules.md#redundant-optional-initialization).
> 
> > 
> 
> > I'm ok if we want to rename this rule and make it configurable to both cases.
> 
> 
> 
> But I am also open to introducing it as a new style rule in addition to the old rule. What are your thoughts on this, @SimplyDanny?

I see. That `redundant_optional_initialization` still exists irritated me. If the new rule behaves like it by default, we don't need it anymore. With the alias, there would be no change for anyone.

leo-lem:
so remove the `redundant_optional_init` rule and enable the new rule by default, to match the behavior of the old?

SimplyDanny:
> so remove the `redundant_optional_init` rule and enable the new rule by default, to match the behavior of the old?

Yes. I didn't check, but all the test cases from `redundant_optional_init` should be part of the new rule then. The [OSS report](https://github.com/realm/SwiftLint/pull/6165#issuecomment-3085013164) should reflect the change in that there should be fixed violations for the old rule that are replaced with the same violations in the new rule. This, then, is "proof" for their functional equivalence. 

There is also [a better report](https://buildkite.com/organizations/swiftlint/pipelines/swiftlint/builds/10136/jobs/01987677-25f1-497e-acb2-bc61979feafa/artifacts/01987690-2088-4411-bc48-08af2f8afffd) categorizing fixed and new violations. I can post it here after the removal of the old rule in case you cannot access it.

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRule.swift
```diff
@@ -0,0 +1,297 @@
+import SwiftSyntax
+
+@SwiftSyntaxRule(explicitRewriter: true)
+struct ImplicitOptionalInitializationRule: Rule {
+  var configuration = ImplicitOptionalInitConfiguration()
```
SimplyDanny: Indentation should be 4 spaces.

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRule.swift
```diff
@@ -0,0 +1,297 @@
...
+
+  static let description = RuleDescription(
+    identifier: "implicit_optional_initialization",
+    name: "Implicit Optional Initialization",
+    description: "Enforce implicit initialization of optional variables (always or never)",
```
SimplyDanny: This is not only a description, but also the message shown in a violation. So it should be direct and clear. Since it depends on the new option now, we should report targeted messages specific for the `always` and `never` options. You can achieve this by adding a custom reason when you create a violation.

leo-lem:
summary looks good to me :)

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRule.swift
```diff
@@ -0,0 +1,297 @@
...
+    for style: ImplicitOptionalInitConfiguration.Style
+  ) -> [AbsolutePosition]? {
+    guard
+      bindingSpecifier.tokenKind == .keyword(.var),
+      !modifiers.contains(keyword: .lazy)
+    else { return nil }
+
+    let bindings = bindings.compactMap { $0.violationPosition(for: style) }
+
+    return bindings.isEmpty ? nil : bindings
```
SimplyDanny: Should be fine to return just an empty list here instead of `nil`.

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRule.swift
```diff
@@ -0,0 +1,297 @@
...
+        }) {  // we have not only willSet or didSet
+          return nil
+        }
+      } else {  // we have not only willSet or didSet
+        return nil
+      }
+    }
+
+    if (style == .never && !initializer.isNil) || (style == .always && initializer.isNil) {
+      return typeAnnotation.endPositionBeforeTrailingTrivia
```
SimplyDanny: I think it would be more reasonable to trigger on the variable name.

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRule.swift
```diff
@@ -0,0 +1,297 @@
...
+    for style: ImplicitOptionalInitConfiguration.Style
+  ) -> AbsolutePosition? {
+    guard
+      let typeAnnotation, typeAnnotation.isOptionalType
+    else { return nil }
+
+    // ignore properties with accessors unless they have only willSet or didSet
+    if let accessorBlock {
+      if let accessors = accessorBlock.accessors.as(AccessorDeclListSyntax.self) {
+        if accessors.allSatisfy({
```
SimplyDanny: Shouldn't that be `accessors.contains { $0 != .willSet && $0 != .didSet }` (pseudo code)?

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRule.swift
```diff
@@ -0,0 +1,297 @@
...
+          $0.accessorSpecifier.tokenKind != .keyword(.willSet)
+            && $0.accessorSpecifier.tokenKind != .keyword(.didSet)
+        }) {  // we have not only willSet or didSet
+          return nil
+        }
+      } else {  // we have not only willSet or didSet
```
SimplyDanny:
```suggestion
      } else {  // code block, i.e. getter
```

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRule.swift
```diff
@@ -0,0 +1,297 @@
...
+      }
+
+      violations.append(contentsOf: violationPositions)
+    }
+  }
+}
+
+private extension ImplicitOptionalInitializationRule {
+  final class Rewriter: ViolationsSyntaxRewriter<ConfigurationType> {
+    override func visitAny(_: Syntax) -> Syntax? { nil }
```
SimplyDanny: That shouldn't be needed. You might need to visit a pattern binding instead of a variable declaration.

SimplyDanny (changes requested):
OSS report looks good. It shows that the new rule behaves like the replaces one by default.

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRule.swift
```diff
@@ -0,0 +1,160 @@
+import SwiftSyntax
+
+@SwiftSyntaxRule(explicitRewriter: true)
+struct ImplicitOptionalInitializationRule: Rule {
+    var configuration = ImplicitOptionalInitConfiguration()
+
+    static let description = RuleDescription(
+        identifier: "implicit_optional_initialization",
+        name: "Implicit Optional Initialization",
+        description: "Optionals should be consistently initialised",
```
SimplyDanny:
```suggestion
        description: "Optionals should be consistently initialized, either with `= nil` or without.",
```

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRule.swift
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        var reason: String {
+            let recommendation =
+                switch configuration.style {
+                case .always: "explicitly initialized to nil"
+                case .never: "implicitly initialized without nil"
+                }
+
+            return "Optionals should be \(recommendation) in variable declarations"
```
SimplyDanny:
```suggestion
            return "Optional should be \(recommendation)"
```

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRule.swift
```diff
@@ -0,0 +1,160 @@
...
+private extension ImplicitOptionalInitializationRule {
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        override func visitPost(_ node: VariableDeclSyntax) {
+            violations.append(
+                contentsOf: node.violationPositions(for: configuration.style).map {
+                    ReasonedRuleViolation(position: $0, reason: reason)
+                })
+        }
+
+        var reason: String {
```
SimplyDanny: Can we have this as a computed property of the visitor? That would make it easier to find.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/ImplicitOptionalInitConfiguration.swift
```diff
@@ -0,0 +1,17 @@
+import SwiftLintCore
+
+@AutoConfigParser
+struct ImplicitOptionalInitConfiguration: SeverityBasedRuleConfiguration {
```
SimplyDanny: Please have the name exactly like the rule.

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRule.swift
```diff
@@ -0,0 +1,160 @@
...
+
+private extension ImplicitOptionalInitializationRule {
+    final class Rewriter: ViolationsSyntaxRewriter<ConfigurationType> {
+        override func visit(_ node: VariableDeclSyntax) -> DeclSyntax {
+            let violationPositions = node.violationPositions(for: configuration.style)
+            guard !violationPositions.isEmpty else { return super.visit(node) }
+
+            numberOfCorrections +=
+                violationPositions.filter {
+                    !$0.isContainedIn(
```
SimplyDanny: You wouldn't need this if you were to visit `PatternBindingSyntax` elements here instead of `VariableDeclSyntax`. Construction of the replacement could also be easier that way. Same for the visitor.

With `node.parent` you can access the variable declaration to retrieve the type.

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitRuleExamples.swift
```diff
@@ -0,0 +1,158 @@
...
+            func funcName() {
+              var myVar: String?
+            }
+            """, configuration: ["style": "always"]),
+    ]
+
+    static let triggeringExamples = [
+        // never style
+        Example("var myVar↓: Int? ", configuration: ["style": "never"]),
+        Example("var myVar↓: Optional<Int> ", configuration: ["style": "never"]),
```
SimplyDanny: I think that the violation position should either be at the name

```swift
Example("var ↓myVar: Optional<Int> ", configuration: ["style": "never"]),
```

or after the type

```swift        
Example("var myVar: Optional<Int>↓ ", configuration: ["style": "never"]),        
```

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/ImplicitOptionalInitConfiguration.swift
leo-lem: i get lint violation for type name too long then, what's you recommendation here?

disable the type name length rule? or go with something less descriptive for the rule name, like OptionalInitialization?

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRule.swift
leo-lem: it's on the visitor currently. do you mean to move it to the configuration? or move recommendation to an extra computed property?

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/ImplicitOptionalInitConfiguration.swift
SimplyDanny: Just disable the rule. These names follow defined patterns. Following the rule would only cause confusions.

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRule.swift
SimplyDanny: I see. Then it's the order that tricked me. I'm too used to properties being located before functions. 😅

leo-lem: ah, fair enough. i'll move it up.

leo-lem:
@SimplyDanny there seems to be no space left on the buildkite device, how can we fix that?

https://buildkite.com/swiftlint/swiftlint/builds/10196/steps/canvas?sid=01989972-3289-400f-9d21-c70f06911036

SimplyDanny:
> @SimplyDanny there seems to be no space left on the buildkite device, how can we fix that?
> 
> https://buildkite.com/swiftlint/swiftlint/builds/10196/steps/canvas?sid=01989972-3289-400f-9d21-c70f06911036

I can try to clean this up later.

Unrelated: Please prefer rebase over merge of `main` to avoid merge commits.

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRule.swift
```diff
@@ -0,0 +1,129 @@
...
+
+private extension ImplicitOptionalInitializationRule {
+    final class Rewriter: ViolationsSyntaxRewriter<ConfigurationType> {
+        override func visit(_ node: PatternBindingSyntax) -> PatternBindingSyntax {
+            guard
+                let violationPosition = node.violationPosition(for: configuration.style),
+                !violationPosition.isContainedIn(regions: disabledRegions, locationConverter: locationConverter)
+            else {
+                return super.visit(node)
+            }
```
SimplyDanny:
```suggestion
            guard node.violationPosition(for: configuration.style) != nil else {
                return super.visit(node)
            }
```

SimplyDanny (approved):
Looks good to me now. Thanks for all the updates.

You seem to handle comments after the type annotation and after the `nil` literal property. However, there are no test cases for that. That would be the only thing missing.

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRule.swift
SimplyDanny: This is already handled in the parent class.

Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRuleExamples.swift
```diff
@@ -39,6 +39,8 @@ enum ImplicitOptionalInitializationRuleExamples { // swiftlint:disable:this type
             }
             """, configuration: ["style": "never"]),
 
+        Example("var x: Int? = nil // comment", configuration: ["style": "never"]),  // with comment after
```
SimplyDanny, line 41: This is especially important in `corrections` to make sure everything will get replaced in the intended way.

SimplyDanny:
Thank you for all the work, @leo-lem! That's good to be merged now.

PR #6166, opened 2025-07-18
Refactor Exclude by behavior into a Strategy pattern

rlf-doordash:
Right now there are only 2 ways of excluding files managed by an enum. The logic deciding between those 2 is split along several places of code, which makes it difficult to standardize between them and pose a scalability challenge if we were to add new options. 

This PR creates a common protocol among those 2 options (by prefix and by expanding subpaths) and refactors them so their logic can be called generically instead of exposing inner implementation details in different places of code


rlf-doordash:
@SimplyDanny Do you happen to know how to get a review on proposed changes?

SimplyDanny:
> @SimplyDanny Do you happen to know how to get a review on proposed changes?

I can have a look. But it'll take a while.

rlf-doordash:
> > @SimplyDanny Do you happen to know how to get a review on proposed changes?
> 
> I can have a look. But it'll take a while.

Sounds good, would appreciate it. This is part of a larger improvement I have prepared to allow faster parsing of large codebases :D

Source/SwiftLintFramework/Configuration/Configuration+LintableFiles.swift
```diff
@@ -18,7 +13,7 @@ extension Configuration {
     /// - returns: Files to lint.
     public func lintableFiles(inPath path: String,
                               forceExclude: Bool,
-                              excludeBy: ExcludeBy) -> [SwiftLintFile] {
+                              excludeBy: any ExcludeByStrategy) -> [SwiftLintFile] {
```
SimplyDanny: Can be `some`, can't it?

Source/SwiftLintFramework/ExcludeByStrategy/ExcludeByPathsByExpandingSubPaths.swift
```diff
@@ -0,0 +1,33 @@
+//
+//  ExcludeByPathsByExpandingSubPaths.swift
+//
```
SimplyDanny: Please remove the header.

Source/SwiftLintFramework/ExcludeByStrategy/ExcludeByPathsByExpandingSubPaths.swift
```diff
@@ -0,0 +1,33 @@
...
+
+import Foundation
+
+public struct ExcludeByPathsByExpandingSubPaths: ExcludeByStrategy {
```
SimplyDanny: We don't need `public`, do we?

Source/SwiftLintFramework/ExcludeByStrategy/ExcludeByStrategyFactory.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+class ExcludeByStrategyFactory {
+    static func createExcludeByStrategy(options: LintOrAnalyzeOptions,
```
SimplyDanny: Could be part of the protocol as an extension. Otherwise, `enum`s are typically the go-to types for factories.

SimplyDanny:
> > > @SimplyDanny Do you happen to know how to get a review on proposed changes?
> > 
> > 
> > I can have a look. But it'll take a while.
> 
> Sounds good, would appreciate it. This is part of a larger improvement I have prepared to allow faster parsing of large codebases :D

That motivated me come back to this PR faster. 😉

Source/SwiftLintFramework/ExcludeByStrategy/ExcludeByStrategyFactory.swift
rlf-doordash: What do you have in mind? Something like: 

```
enum ExcludeByStrategyType {
    case excludeByPrefix(ExcludeByPrefixStrategy)
    case excludeByPathsByExpandingSubPaths(ExcludeByPathsByExpandingSubPaths)
    
    static func createExcludeByStrategy(options: LintOrAnalyzeOptions,
                                        configuration: Configuration,
                                        fileManager: some LintableFileManager = FileManager.default)
    -> ExcludeByStrategyType {
        if options.useExcludingByPrefix {
            let strategy = ExcludeByPrefixStrategy(excludedPaths: configuration.excludedPaths)
            return .excludeByPrefix(strategy)
        }

        let strategy = ExcludeByPathsByExpandingSubPaths(configuration: configuration, fileManager: fileManager)
        return .excludeByPathsByExpandingSubPaths(strategy)
    }
    
    var strategy: any ExcludeByStrategy {
        switch self {
        case .excludeByPrefix(let strategy):
            return strategy
        case .excludeByPathsByExpandingSubPaths(let strategy):
            return strategy
        }
    }
}
```

And use it like 
` ExcludeByStrategyType.createExcludeByStrategy(options: options, configuration: self).strategy` ? 


It cannot be an extension of the protocol as it fails on call, cannot call static funcs in meta types

rlf-doordash:
> > > > @SimplyDanny Do you happen to know how to get a review on proposed changes?
> > > 
> > > 
> > > I can have a look. But it'll take a while.
> > 
> > 
> > Sounds good, would appreciate it. This is part of a larger improvement I have prepared to allow faster parsing of large codebases :D
> 
> That motivated me come back to this PR faster. 😉

Sorry, I was out and it has taken a while to get back on this :( I am expectin g to be able to put some time into this for the next couple of weeks tho!

Source/SwiftLintFramework/Configuration/Configuration+LintableFiles.swift
```diff
@@ -38,17 +33,12 @@ extension Configuration {
     internal func lintablePaths(
         inPath path: String,
         forceExclude: Bool,
-        excludeBy: ExcludeBy,
+        excludeBy: any ExcludeByStrategy,
```
SimplyDanny, line 36: Can also be `some`, can't it?

Source/SwiftLintFramework/ExcludeByStrategy/ExcludeByStrategyFactory.swift
SimplyDanny: This is what I had in mind:

```swift
protocol ExcludeByStrategy {}

struct ExcludeByPathsByExpandingSubPaths: ExcludeByStrategy {}
struct ExcludeByPrefixStrategy: ExcludeByStrategy {}

extension ExcludeByStrategy {
  static func create(_ b: Bool) -> any ExcludeByStrategy {
    if b {
      ExcludeByPathsByExpandingSubPaths()
    } else {
      ExcludeByPrefixStrategy()
    }
  }
}
```

Source/SwiftLintFramework/ExcludeByStrategy/ExcludeByPathsByExpandingSubPaths.swift
```diff
@@ -0,0 +1,35 @@
+import Foundation
+
+/// Returns an array of file paths after removing the excluded paths as defined by this configuration.
+///
+/// - parameter fileManager: The lintable file manager to use to expand the excluded paths into all matching paths.
+/// - parameter paths:       The input paths to filter.
+///
+/// - returns: The input paths after removing the excluded paths.
+struct ExcludeByPathsByExpandingSubPaths: ExcludeByStrategy {
```
SimplyDanny, line 9:
```suggestion
struct ExcludeByExpandingSubPathsStrategy: ExcludeByStrategy {
```

Source/SwiftLintFramework/ExcludeByStrategy/ExcludeByPrefixStrategy.swift
```diff
@@ -0,0 +1,19 @@
+/// Returns the file paths that are excluded by this configuration using filtering by absolute path prefix.
+///
+/// For cases when excluded directories contain many lintable files (e. g. Pods) it works faster than default
+/// algorithm `filterExcludedPaths`.
```
SimplyDanny, line 4:
```suggestion
/// For cases when excluded directories contain many lintable files (e. g. Pods) it works faster than the
/// default `filterExcludedPaths` strategy.
```

Source/SwiftLintFramework/ExcludeByStrategy/ExcludeByPathsByExpandingSubPaths.swift
```diff
@@ -0,0 +1,35 @@
...
+    let excludedPaths: [String]
+
+    init(configuration: Configuration, fileManager: some LintableFileManager = FileManager.default) {
+        self.excludedPaths = configuration.excludedPaths
```
SimplyDanny, line 13: Can call the other `init`.

Source/SwiftLintFramework/ExcludeByStrategy/ExcludeByStrategyType.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+
+enum ExcludeByStrategyType {
```
SimplyDanny, line 3: The enum was only to introduce a namespace containing the factory method. We don't really need that, especially not the cases, which are sufficiently described by the two strategies conforming to the protocol.

As written in the other comment, the static factory method can be part of the protocol itself via an extension.

rlf-doordash:
Closing as #6342 already fixed what this was supposed to do

PR #6185, opened 2025-08-12
Use SwiftSyntax 6.2 pre-release 2025-08-11

SimplyDanny:


SimplyDanny:
New and fixed violations in the Swift repository are all related to the new `unsafe` keyword the parser is now able to interpret correctly.

Performance improved significantly again. 🎉

PR #6202, opened 2025-08-23
Ignore locally defined `count` identifiers in `empty_count` rule

### LINKED ISSUE #5326: `empty_count` rule triggers on local symbols
### New Issue Checklist

### Describe the bug

`empty_count` violation is triggered on extension over `IteratorProtocol`

Sample code:

```swift
extension IteratorProtocol {
    mutating func skip(_ count: Int) {
        guard count > 0 else { return } // ERROR: Empty Count Violation: Prefer checking `isEmpty` over comparing `count` to zero (empty_count)

        for _ in 0..<count {
            _ = next()
        }
    }
}
```

```bash
$ swiftlint lint
```

### Environment

```yml
# insert yaml contents here
```

SimplyDanny:
Fixes #5326.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -123,6 +123,7 @@ swift_library(
     visibility = ["//visibility:public"],
     deps = [
         ":SwiftLintCore",
+        "@SwiftSyntax//:SwiftLexicalLookup_opt",
     ],
 )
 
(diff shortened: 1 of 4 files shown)


SimplyDanny:
Requires SwiftSyntax >=6.2 to have the library available in Bazel.

PR #6217, opened 2025-08-31
Merge `function_name_whitespace` and `operator_whitespace` rules

SimplyDanny:

PR #6223, opened 2025-09-02
Add `multiline_call_arguments ` rule

GandaLF2006:
### Summary

This PR adds a new opt-in SwiftSyntax rule: `multiline_call_arguments `.

The rule mirrors the behavior of multiline_parameters for declarations, but applies it to function and method call sites (including Type.foo(...) / Enum.foo(...)), enforcing a consistent multiline formatting style for arguments.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -32,6 +32,7 @@ disabled_rules:
   - missing_docs
   - multiline_arguments
   - multiline_arguments_brackets
+  - multiline_call_arguments
   - multiline_function_chains
   - multiline_parameters_brackets
   - no_extension_access_modifier
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
@@ -1,0 +1,114 @@
+import SwiftLintCore
+import SwiftSyntax
+
+@SwiftSyntaxRule(optIn: true)
+struct MultilineCallArgumentsRule: Rule {
+    var configuration = MultilineCallArgumentsConfiguration()
+
+    static let description = RuleDescription(
+        identifier: "multiline_call_arguments",
+        name: "Multiline Call Arguments",
+        description: "Call should have one parameter per line",
+        kind: .style,
+        nonTriggeringExamples: [
+            Example("""
+                    foo(
+                    param1: "param1",
+                        param2: false,
+                        param3: []
+                    )
+                    """, configuration: ["max_number_of_single_line_parameters": 2]),
+            Example("""
+                    foo(param1: 1,
+                        param2: false,
+                        param3: [])
+                    """, configuration: ["max_number_of_single_line_parameters": 1]),
+            Example("foo(param1: 1, param2: false)",
+                    configuration: ["max_number_of_single_line_parameters": 2]),
+            Example("Enum.foo(param1: 1, param2: false)",
+                    configuration: ["max_number_of_single_line_parameters": 2]),
+            Example("foo(param1: 1)", configuration: ["allows_single_line": false]),
+            Example("Enum.foo(param1: 1)", configuration: ["allows_single_line": false]),
+            Example("Enum.foo(param1: 1, param2: 2, param3: 3)", configuration: ["allows_single_line": true]),
+            Example("""
+                    foo(
+                        param1: 1,
+                        param2: 2,
+                        param3: 3
+                    )
+                    """,
+                    configuration: [
+                        "allows_single_line": false
+                    ]),
+        ],
+        triggeringExamples: [
+            Example("↓foo(param1: 1, param2: false, param3: [])",
+                    configuration: [
+                        "max_number_of_single_line_parameters": 2
+                    ]),
+            Example("↓Enum.foo(param1: 1, param2: false, param3: [])",
+                    configuration: [
+                        "max_number_of_single_line_parameters": 2
+                    ]),
+            Example("""
+                    ↓foo(param1: 1, param2: false,
+                         param3: [])
+                    """, configuration: [
+                        "max_number_of_single_line_parameters": 3
+                    ]),
+            Example("""
+                    ↓Enum.foo(param1: 1, param2: false,
+                         param3: [])
+                    """, configuration: [
+                        "max_number_of_single_line_parameters": 3
+                    ]),
+            Example("↓foo(param1: 1, param2: false)", configuration: ["allows_single_line": false]),
+            Example("↓Enum.foo(param1: 1, param2: false)", configuration: ["allows_single_line": false]),
+        ]
+    )
+}
+
+private extension MultilineCallArgumentsRule {
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        override func visitPost(_ node: FunctionCallExprSyntax) {
+            let parameterPositions = node.arguments.map(\.positionAfterSkippingLeadingTrivia)
+            if containsViolation(parameterPositions: parameterPositions) {
+                violations.append(node.calledExpression.positionAfterSkippingLeadingTrivia)
+            }
+        }
+
+        private func containsViolation(parameterPositions: [AbsolutePosition]) -> Bool {
+            guard parameterPositions.isNotEmpty else {
+                return false
+            }
+
+            var numberOfParameters = 0
+            var linesWithParameters: Set<Int> = []
+            var hasMultipleParametersOnSameLine = false
+
+            for position in parameterPositions {
+                let line = locationConverter.location(for: position).line
+
+                if !linesWithParameters.insert(line).inserted {
+                    hasMultipleParametersOnSameLine = true
+                }
+
+                numberOfParameters += 1
+            }
+
+            if linesWithParameters.count == 1 {
+                guard configuration.allowsSingleLine else {
+                    return numberOfParameters > 1
+                }
+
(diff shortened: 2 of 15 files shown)


GandaLF2006:
@SimplyDanny One more PR for review :)

SimplyDanny:
Thanks @GandaLF2006 and sorry for the late response! 

I think this should a separate rule instead of an option, though.

GandaLF2006:
> I think this should a separate rule instead of an option, though.

Hey. Are you sure about it? `function_name_whitespace` created as separate rule, but later merged with `operator_whitespace` [here](https://github.com/realm/SwiftLint/pull/6217)

SimplyDanny:
> > I think this should a separate rule instead of an option, though.
> 
> 
> 
> Hey. Are you sure about it? `function_name_whitespace` created as separate rule, but later merged with `operator_whitespace` [here](https://github.com/realm/SwiftLint/pull/6217)

Yes. Those rules were basically the same. The only difference was the function name being a real name vs. a symbol.

Here, we are talking about function declarations vs. function calls. That's totally different syntax elements even though the idea is the same. We might come up with different options for declarations and calls, which we can't separate when both are mixed in the same rule.

rgoldberg:
Should this possibly be renamed `multiline_arguments`, `multiline_call_arguments`, or `multiline_call_parameters`?

I'd recommend one of the first two, since it's for arguments being passed to parameters, not just the parameters themselves.

Also, the greater the naming difference between the two rules, the less likely people are to mistake them for each other.

If you still want to use `parameters`, `multiline_call_parameters` is more concise than `multiline_parameters_on_call`.

If you really want the new rule to start with the whole name as ``multiline_parameters`, maybe `multiline_parameters_for_call` sounds better.

GandaLF2006:
@rgoldberg Good catch.
Renamed to `multiline_call_arguments`

GandaLF2006:
@SimplyDanny It's done! Please review :)

SimplyDanny:
Please disable the rule in `.swiftlint.yml` for the time being. We should not mix the rule implementation with unrelated code changes. Whether the rule should be activated for SwiftLint is a separate decision.

GandaLF2006:
@SimplyDanny Rule disabled

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
```diff
@@ -0,0 +1,150 @@
...
+
+@SwiftSyntaxRule(optIn: true)
+struct MultilineCallArgumentsRule: Rule {
+    var configuration = MultilineCallArgumentsConfiguration()
+
+    static let description = RuleDescription(
+        identifier: "multiline_call_arguments",
+        name: "Multiline Call Arguments",
+        description: """
+            Arguments in function and method calls should be either on the same line \
```
SimplyDanny: We rather need a concise violation message here. What about "Call should have one parameter per line"?

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
```diff
@@ -0,0 +1,150 @@
...
+                        param2: false,
+                        param3: [])
+                    """, configuration: ["max_number_of_single_line_parameters": 1]),
+            Example("foo(param1: 1, param2: false)",
+                    configuration: ["max_number_of_single_line_parameters": 2]),
+            Example("Enum.foo(param1: 1, param2: false)",
+                    configuration: ["max_number_of_single_line_parameters": 2]),
+            Example("""
+                    foo(param1: 1)
+                    """,
```
SimplyDanny: One-liners can be written like this:
```suggestion
            Example("foo(param1: 1)",
```

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
```diff
@@ -0,0 +1,150 @@
...
+                    configuration: [
+                        "allows_single_line": false
+                    ]),
+            Example("""
+                    Enum.foo(param1: 1)
+                    """,
+                    configuration: [
+                        "allows_single_line": false
+                    ]),
```
SimplyDanny: These can be a bit shorter:

```suggestion
            Example("Enum.foo(param1: 1)",configuration: ["allows_single_line": false]),
```

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
```diff
@@ -0,0 +1,150 @@
...
+        ]
+    )
+}
+
+private extension MultilineCallArgumentsRule {
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        override func visitPost(_ node: FunctionCallExprSyntax) {
+            guard node.arguments.isNotEmpty else { return }
+            guard node.trailingClosure == nil else { return }
```
SimplyDanny:
```suggestion
            guard node.arguments.isNotEmpty, node.trailingClosure == nil else { 
                return
            }
```

SimplyDanny: Why to exclude the trailing closure case?

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
```diff
@@ -0,0 +1,150 @@
...
+
+            if containsViolation(for: node.arguments) {
+                let anchor = node.calledExpression.positionAfterSkippingLeadingTrivia
```
SimplyDanny: Can be inlined.

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
```diff
@@ -0,0 +1,150 @@
...
+                violations.append(anchor)
+            }
+        }
+
+        private func containsViolation(for arguments: LabeledExprListSyntax) -> Bool {
```
SimplyDanny: Let the other method extract the positions and remove this one.

SimplyDanny:
> @SimplyDanny Rule disabled

Could you please also revert all the unrelated code changes?

SimplyDanny (changes requested):
Thanks for implementing a separate rule!

There are a few comments to consider ...

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
GandaLF2006: Good catch! Remoted

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
GandaLF2006: Done

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
GandaLF2006: Fixed

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
GandaLF2006: Updated

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
GandaLF2006: Updated

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
```diff
@@ -0,0 +1,118 @@
...
+            Example("↓foo(param1: 1, param2: false)", configuration: ["allows_single_line": false]),
+            Example("↓Enum.foo(param1: 1, param2: false)", configuration: ["allows_single_line": false]),
+        ]
+    )
+}
+
+private extension MultilineCallArgumentsRule {
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        override func visitPost(_ node: FunctionCallExprSyntax) {
+            guard node.arguments.isNotEmpty else {
```
SimplyDanny: We can even remove this condition as `containsViolation` also works without any parameters.

.swiftlint.yml
```diff
@@ -48,6 +48,7 @@ disabled_rules:
   - trailing_closure
   - type_contents_order
   - vertical_whitespace_between_cases
+  - multiline_call_arguments
```
SimplyDanny: Alphabetically please.

GandaLF2006: Done

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
GandaLF2006: Removed

SimplyDanny (approved):
Just applied some polishing on the examples and description. Everything should be good now.

Thanks @GandaLF2006!

PR #6225, opened 2025-09-02
Include bzl dependency files in release package

sebastianv1:
Bumping to `0.60.0` and trying to run `bazel test "@SwiftLint//Tests:ExtraRulesTests"` revealed the issue
```
ERROR: error loading package '@@swiftlint~//Tests': cannot load '@@swiftlint~//Tests:generated_tests.bzl': no such file
```

as these files aren't packaged into the release. I went ahead and added them as a part of the release package but let me know if there's a better approach instead.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -237,6 +237,8 @@ filegroup(
         "//:DyldWarningWorkaroundSources",
         "//:LintInputs",
         "//Tests:BUILD",
+        "//Tests:generated_tests.bzl",
+        "//Tests:test_macros.bzl",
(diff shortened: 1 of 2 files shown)


SimplyDanny (approved):
Thanks @sebastianv1! That was an oversight of a test split in the last release.

It's good for now. Having the internal test generation scripts available for consumers as part of `release_files` isn't perfect though. All that's needed would be `TestHelpers` and `ExtraRulesTests`. I'll work on that as a follow up.

PR #6235, opened 2025-09-04
Expose only `TestHelpers` and `ExtraRulesTests` as part of Bazel releases

SimplyDanny:
Other test targets don't need to be released.


Tests/BUILD
```diff
@@ -1,44 +1,15 @@
 load("@build_bazel_rules_swift//swift:swift.bzl", "swift_library", "swift_test")
+load("//bazel:copts.bzl", "STRICT_COPTS", "TARGETED_COPTS")
 load(":generated_tests.bzl", "generated_tests")
```
jszumski, line 3: I think this load will fail when given a release archive since it's excluded from `release_files`.  I had to patch locally by adding the file:

```
def generated_tests():
    pass
```

SimplyDanny: The idea is to stop exposing `Tests/BUILD` and with that all test sources. The only relevant test targets should be `ExtraRulesTests` and `TestHelpers`. This then also avoids that `generated_tests.bzl` is required. Is that reasonable?

SimplyDanny: If you like, you can try [bazel.tar.gz](https://github.com/user-attachments/files/22178781/bazel.tar.gz) which would be the archive for the next release.

jszumski: Gotcha, I saw the analysis error on 0.60.0 but that linked archive looks good

PR #6237, opened 2025-09-05
Rewrite `quick_discouraged_call` rule with SwiftSyntax

SimplyDanny:


SimplyDanny:
OSS findings show that the violation location changed for more complex expressions and the rule triggers on a few more cases now. However, it's anyway very fuzzy to decide on what to trigger and what not.

PR #6261, opened 2025-09-28
Add prefer_asset_symbols rule

### LINKED ISSUE #5939: Rule Request: Prefer using Asset Symbols
### New Issue Checklist

### New Rule Request

It would be happy when Image that initialized with string replaced with initializer with asset symbols.

> 1. Why should this rule be added? Share links to existing discussion about what

`UIKit.UIImage(named:)` and `SwiftUI.Image(_:)` contain the risk of bugs due to typos.
Since Xcode 15, Xcode generates codes for images in the Asset Catalog and it can avoid typos. ([ref](https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes#Asset-Catalogs))

### Would trigger:
```swift
// UIKit
UIImage(named: "some_image") // simple
UIImage(named: "some image")  // contains space
UIImage(named: "someImage", in: .module) // contains specified module

// SwiftUI
Image("some_image") // simple
Image("some image")  // contains space
Image("someImage", bundle: .module) // contains specified module
```

### Would not trigger:
```swift
// UIKit
UIImage(resource: .someImage)
UIImage(systemName: "trash")  // using SF Symbols

// SwiftUI
Image(.someImage)
Image(systemName: "trash")  // using SF Symbols
```

Currently no.

Enabled by default on Swift 5.9+.
Asset symbol generation is enabled by default in Xcode, and it needs editing the build setting to disable it.

danglingP0inter:
Implements a new SwiftLint rule that suggests using asset symbols over string-based image initialization to avoid typos and enable compile-time checking.

The rule detects:
- UIImage(named: "string") calls
- SwiftUI Image("string") calls

And suggests using asset symbols like:
- UIImage(resource: .imageName)
- Image(.imageName)

Resolves #5939

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferAssetSymbolsRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferAssetSymbolsRule.swift
@@ -1,0 +1,119 @@
+import SwiftLintCore
+import SwiftSyntax
+
+@SwiftSyntaxRule(optIn: false)
+struct PreferAssetSymbolsRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "prefer_asset_symbols",
+        name: "Prefer Asset Symbols",
+        description: "Prefer using asset symbols over string-based image initialization to avoid typos and enable compile-time checking",
+        rationale: """
+            UIKit.UIImage(named:) and SwiftUI.Image(_:) contain the risk of bugs due to typos. \
+            Since Xcode 15, Xcode generates codes for images in the Asset Catalog and it can avoid typos by providing compile-time checking.
+            """,
+        kind: .idiomatic,
+        minSwiftVersion: .fiveDotNine,
+        nonTriggeringExamples: [
+            // UIKit - using asset symbols
+            Example("UIImage(resource: .someImage)"),
+            Example("UIImage(systemName: \"trash\")"),
+            // SwiftUI - using asset symbols  
+            Example("Image(.someImage)"),
+            Example("Image(systemName: \"trash\")"),
+            // Dynamic strings (variables or interpolated)
+            Example("UIImage(named: imageName)"),
+            Example("UIImage(named: \"image_\\(suffix)\")"),
+            Example("Image(imageName)"),
+            Example("Image(\"image_\\(suffix)\")"),
+            // Bundle-specific initializers
+            Example("UIImage(named: \"someImage\", in: Bundle.main, compatibleWith: nil)"),
+            Example("Image(\"someImage\", bundle: Bundle.main)"),
+        ],
+        triggeringExamples: [
+            // UIKit examples
+            Example("↓UIImage(named: \"some_image\")"),
+            Example("↓UIImage(named: \"some image\")"),
+            Example("↓UIImage.init(named: \"someImage\")"),
+            // SwiftUI examples  
+            Example("↓Image(\"some_image\")"),
+            Example("↓Image(\"some image\")"),
+            Example("↓Image.init(\"someImage\")"),
+        ]
+    )
+}
+
+private extension PreferAssetSymbolsRule {
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        override func visitPost(_ node: FunctionCallExprSyntax) {
+            let calledExpression = node.calledExpression.trimmedDescription
+            
+            // Check for UIImage(named:) calls
+            if isUIImageNamedInit(node: node, name: calledExpression) {
+                violations.append(node.positionAfterSkippingLeadingTrivia)
+            }
+            // Check for SwiftUI Image(_:) calls
+            else if isSwiftUIImageInit(node: node, name: calledExpression) {
+                violations.append(node.positionAfterSkippingLeadingTrivia)
+            }
+        }
+        
+        private func isUIImageNamedInit(node: FunctionCallExprSyntax, name: String) -> Bool {
+            // Check if this is a UIImage or UIImage.init call
+            guard uiImageInits().contains(name) else {
+                return false
+            }
+            
+            // Check if it has a "named" parameter with a string literal
+            guard let namedArgument = node.arguments.first(where: { $0.label?.text == "named" }),
+                  let stringLiteral = namedArgument.expression.as(StringLiteralExprSyntax.self),
+                  stringLiteral.isConstantString else {
+                return false
+            }
+            
+            // Don't trigger if there are additional parameters like "in:" (bundle parameter)
+            let argumentLabels = node.arguments.compactMap(\.label?.text)
+            return argumentLabels == ["named"]
+        }
+        
+        private func isSwiftUIImageInit(node: FunctionCallExprSyntax, name: String) -> Bool {
+            // Check if this is an Image or Image.init call
+            guard swiftUIImageInits().contains(name) else {
+                return false
+            }
+            
+            // For SwiftUI Image, the first parameter is unlabeled for the image name
+            guard let firstArgument = node.arguments.first,
+                  firstArgument.label == nil,
+                  let stringLiteral = firstArgument.expression.as(StringLiteralExprSyntax.self),
+                  stringLiteral.isConstantString else {
+                return false
+            }
+            
+            // Don't trigger if there are additional parameters like "bundle:"
+            let argumentLabels = node.arguments.compactMap(\.label?.text)
+            return argumentLabels.isEmpty || (argumentLabels.count == 1 && argumentLabels.first == nil)
+        }
+        
+        private func uiImageInits() -> [String] {
+            return [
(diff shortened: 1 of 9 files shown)


Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferAssetSymbolsRule.swift
```diff
@@ -0,0 +1,119 @@
+import SwiftLintCore
+import SwiftSyntax
+
+@SwiftSyntaxRule(optIn: false)
```
SimplyDanny: I personally feel like this rule shouldn't be enabled by default.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferAssetSymbolsRule.swift
```diff
@@ -0,0 +1,119 @@
...
+struct PreferAssetSymbolsRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "prefer_asset_symbols",
+        name: "Prefer Asset Symbols",
+        description: "Prefer using asset symbols over string-based image initialization to avoid typos and enable compile-time checking",
```
SimplyDanny: This is also the message that appears in a violation, so it should be shorter:
```suggestion
        description: "Prefer using asset symbols over string-based image initialization",
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferAssetSymbolsRule.swift
```diff
@@ -0,0 +1,119 @@
...
+        rationale: """
+            UIKit.UIImage(named:) and SwiftUI.Image(_:) contain the risk of bugs due to typos. \
+            Since Xcode 15, Xcode generates codes for images in the Asset Catalog and it can avoid typos by providing compile-time checking.
```
SimplyDanny: This will be rendered as Markdown. Let's also be a bit more descriptive here:
```suggestion
            `UIKit.UIImage(named:)` and `SwiftUI.Image(_:)` bear the risk of bugs due to typos in their string arguments.
            Since Xcode 15, Xcode generates codes for images in the Asset Catalog. Usage of these codes
            and system icons from SF Symbols avoid typos and allow for compile-time checking.
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferAssetSymbolsRule.swift
```diff
@@ -0,0 +1,119 @@
...
+            Example("↓Image(\"some image\")"),
+            Example("↓Image.init(\"someImage\")"),
+        ]
+    )
+}
+
+private extension PreferAssetSymbolsRule {
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        override func visitPost(_ node: FunctionCallExprSyntax) {
+            let calledExpression = node.calledExpression.trimmedDescription
```
SimplyDanny: We should't rely on the string representation but instead match on the syntax tree.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferAssetSymbolsRule.swift
```diff
@@ -0,0 +1,119 @@
...
+        }
+        
+        private func isUIImageNamedInit(node: FunctionCallExprSyntax, name: String) -> Bool {
+            // Check if this is a UIImage or UIImage.init call
+            guard uiImageInits().contains(name) else {
+                return false
+            }
+            
+            // Check if it has a "named" parameter with a string literal
+            guard let namedArgument = node.arguments.first(where: { $0.label?.text == "named" }),
```
SimplyDanny: It must be the **first** argument not "any `named` argument anywhere in the call", right?

You can actually do `node.arguments.onlyElement`. That covers all cases in one condition (it ensure that there are no other labels).

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferAssetSymbolsRule.swift
```diff
@@ -0,0 +1,119 @@
...
+                  let stringLiteral = namedArgument.expression.as(StringLiteralExprSyntax.self),
+                  stringLiteral.isConstantString else {
+                return false
+            }
+            
+            // Don't trigger if there are additional parameters like "in:" (bundle parameter)
```
SimplyDanny: For my understanding, why wouldn't we trigger in this case?

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferAssetSymbolsRule.swift
```diff
@@ -0,0 +1,119 @@
...
+            guard let firstArgument = node.arguments.first,
+                  firstArgument.label == nil,
+                  let stringLiteral = firstArgument.expression.as(StringLiteralExprSyntax.self),
+                  stringLiteral.isConstantString else {
+                return false
+            }
+            
+            // Don't trigger if there are additional parameters like "bundle:"
+            let argumentLabels = node.arguments.compactMap(\.label?.text)
+            return argumentLabels.isEmpty || (argumentLabels.count == 1 && argumentLabels.first == nil)
```
SimplyDanny: See my other comment about `onlyElement` ...

SimplyDanny (changes requested):
Thanks! Please consider my comments.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferAssetSymbolsRule.swift
danglingP0inter: Thanks for pointing out! I rechecked and it should trigger the violation. Have updated the logic.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferAssetSymbolsRule.swift
danglingP0inter: Yes. It should be the first argument not "any named argument anywhere in the call". Have made the changes.

I don't think we can use `node.arguments.onlyElement`, as init may additionally have `in: ` (Bundle parameter). We need to raise rule violation trigger in such cases.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferAssetSymbolsRule.swift
```diff
@@ -0,0 +1,138 @@
...
+                  firstArgument.label == nil,
+                  let stringLiteral = firstArgument.expression.as(StringLiteralExprSyntax.self),
+                  stringLiteral.isConstantString else {
+                return false
+            }
+
+            return true
+        }
+
+        private func isUIImageCall(_ expression: ExprSyntax) -> Bool {
```
SimplyDanny: Let's combine this with `isImageCall` passing the class name as a parameter.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferAssetSymbolsRule.swift
```diff
@@ -0,0 +1,138 @@
...
+            if isUIImageNamedInit(node: node) {
+                violations.append(node.positionAfterSkippingLeadingTrivia)
+            }
+            // Check for SwiftUI Image(_:) calls
+            else if isSwiftUIImageInit(node: node) {
+                violations.append(node.positionAfterSkippingLeadingTrivia)
+            }
+        }
+
+        private func isUIImageNamedInit(node: FunctionCallExprSyntax) -> Bool {
```
SimplyDanny: You can combine this with `isSwiftUIImageInit` as the only difference is the check for the argument label.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferAssetSymbolsRule.swift
```diff
@@ -0,0 +1,138 @@
...
+            Example("↓Image(\"someImage\", bundle: .main)"),
+        ]
+    )
+}
+
+private extension PreferAssetSymbolsRule {
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        override func visitPost(_ node: FunctionCallExprSyntax) {
+            // Check for UIImage(named:) calls
+            if isUIImageNamedInit(node: node) {
```
SimplyDanny:
```suggestion
            if isUIImageNamedInit(node: node) || isSwiftUIImageInit(node: node) {
```

danglingP0inter:
> That looks good now. Thanks @danglingP0inter!
> 
> Linting checks but complain about a few trailing spaces in the CHANGELOG.

Fixed. Removed trailing spaces. Thanks, @SimplyDanny!

PR #6269, opened 2025-10-01
Add `legacy_uigraphics_functions` rule

### LINKED ISSUE #6268: Rule Request: "Legacy UIGraphics functions"
We could have a `legacy_uigraphics_functions`, which can already be defined as a custom rule:

```
custom_rules:
  legacy_uigraphics_functions:
    included: ".*\\.swift"
    name: "Legacy UIGraphics functions"
    regex: "(UIGraphicsBeginImageContextWithOptions|UIGraphicsBeginImageContext|UIGraphicsGetImageFromCurrentImageContext|UIGraphicsEndImageContext)\\s*\\("
    message: "This UIGraphics method is deprecated. Use UIGraphicsImageRenderer and related APIs instead."
    severity: warning
```

These 4 methods are already marked `API_TO_BE_DEPRECATED` in UIKit Headers.
This new rule would match existing SwiftLint rule such as `legacy_cggeometry_functions` and `legacy_nsgeometry_functions`

3. configurable: no
4. opt-in: yes, to match existing rules `legacy_*_functions` ? The replacement API (UIGraphicsImageRenderer) exists since iOS 10.

DimDL:
Add new `legacy_uigraphics_functions` rule to encourage the use of modern UIGraphicsImageRenderer instead of legacy UIGraphics{Begin|End}ImageContext.

The modern replacement is safer, cleaner, Retina-aware and more performant.
The legacy API is deprecated

Resolves [#6268](https://github.com/realm/SwiftLint/issues/6268)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,7 +56,8 @@
 
 * Add new default `invisible_character` rule that detects invisible characters
   like zero-width space (U+200B), zero-width non-joiner (U+200C),
-  and FEFF formatting character (U+FEFF) in string literals, which can cause hard-to-debug issues.  
+  and FEFF formatting character (U+FEFF) in string literals, which can cause
+  hard-to-debug issues.  
   [kapitoshka438](https://github.com/kapitoshka438)
   [#6045](https://github.com/realm/SwiftLint/issues/6045)
 
@@ -64,6 +65,12 @@
   an identifier from an outer scope.  
   [nadeemnali](https://github.com/nadeemnali)
   [#6228](https://github.com/realm/SwiftLint/issues/6228)
+  
+* Add `legacy_uigraphics_functions` rule to encourage the use of modern
+  `UIGraphicsImageRenderer` instead of the legacy `UIGraphics{Begin|End}ImageContext`.
+  The modern replacement is safer, cleaner, Retina-aware and more performant.  
+  [Dimitri Dupuis-Latour](https://github.com/DimDL)
+  [#6268](https://github.com/realm/SwiftLint/issues/6268)
 
 ### Bug Fixes
 
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/LegacyUIGraphicsFunctionRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/LegacyUIGraphicsFunctionRule.swift
@@ -1,0 +1,74 @@
+import SwiftSyntax
+
+@SwiftSyntaxRule(optIn: true)
+struct LegacyUIGraphicsFunctionRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "legacy_uigraphics_function",
+        name: "Legacy UIGraphics Function",
+        description: "Prefer using `UIGraphicsImageRenderer` over legacy functions",
+        rationale: "The modern replacement is safer, cleaner, Retina-aware and more performant.",
+        kind: .idiomatic,
+        nonTriggeringExamples: [
+            Example("""
+            let renderer = UIGraphicsImageRenderer(size: bounds.size)
+            let screenshot = renderer.image { _ in
+                myUIView.drawHierarchy(in: bounds, afterScreenUpdates: true)
+            }
+            """),
+
+            Example("""
+            let renderer = UIGraphicsImageRenderer(size: newSize)
+            let combined = renderer.image { _ in
+                background.draw(in: CGRect(origin: .zero, size: newSize))
+                watermark.draw(in: CGRect(origin: .zero, size: watermarkSize))
+            }
+            """),
+
+            Example("""
+            UIGraphicsImageRenderer(size: newSize, format: UIGraphicsImageRendererFormat()).image { _ in
+                image.draw(in: CGRect(origin: .zero, size: newSize))
+            }
+            """),
+        ],
+        triggeringExamples: [
+            Example("""
+            ↓UIGraphicsBeginImageContext(newSize)
+            myUIView.drawHierarchy(in: bounds, afterScreenUpdates: false)
+            let optionalScreenshot = ↓UIGraphicsGetImageFromCurrentImageContext()
+            ↓UIGraphicsEndImageContext()
+            """),
+
+            Example("""
+            ↓UIGraphicsBeginImageContext(newSize)
+            background.draw(in: CGRect(origin: .zero, size: newSize))
+            watermark.draw(in: CGRect(origin: .zero, size: watermarkSize))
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -28,7 +28,11 @@
 
 ### Enhancements
 
-* None.
+* Add `legacy_uigraphics_functions` rule to encourage the use of modern
+  UIGraphicsImageRenderer instead of legacy UIGraphics{Begin|End}ImageContext.
+  The modern replacement is safer, cleaner, Retina-aware and more performant.
```
```suggestion
* Add `legacy_uigraphics_functions` rule to encourage the use of modern
  `UIGraphicsImageRenderer `instead of legacy `UIGraphics{Begin|End}ImageContext`.
  The modern replacement is safer, cleaner, Retina-aware and more performant.  
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/LegacyUIGraphicsFunctionsRule.swift
```diff
@@ -0,0 +1,82 @@
+import SwiftSyntax
+
+@SwiftSyntaxRule
```
SimplyDanny: The rule should be opt-in as you have described in the issue:
```suggestion
@SwiftSyntaxRule(optIn: true)
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/LegacyUIGraphicsFunctionRule.swift
```diff
@@ -0,0 +1,82 @@
...
+
+    static let description = RuleDescription(
+        identifier: "legacy_uigraphics_functions",
+        name: "Legacy UIGraphics Functions",
+        description: "Prefer using `UIGraphicsImageRenderer` over legacy functions",
+        rationale: "The modern replacement is safer, cleaner, Retina-aware and more performant",
+        kind: .idiomatic,
+        nonTriggeringExamples: [
+            Example("""
+            let renderer = UIGraphicsImageRenderer(size: bounds.size)
```
SimplyDanny, line 15: Tests can be a bit easier, as the rule effectively only checks function names. So the `let screenshot = ...` part isn't actually necessary.

SimplyDanny (changes requested):
Thanks for the prompt implementation of your issue!

Please consider my comments ... Have you checked whether the OSS findings reported in the bot comment make sense?

SimplyDanny:
I wasn't actually aware of [`LegacyFunctionRuleHelper`](https://github.com/realm/SwiftLint/blob/main/Source/SwiftLintBuiltInRules/Helpers/LegacyFunctionRuleHelper.swift). Can we make use of it for this rule as well?

Edit: I remove the helper in #6271 so that visitor and rewrite can directly be used.

rgoldberg:
I have created an analyzer rule that reports all uses of any deprecated declarations (marked with `@available`), including soft deprecations, etc.

The types mentioned in this PR all seem like they should be detectable by my analyzer rule.

I used AI to create it, as I don't know SourceKitten, SourceKit, SwiftLint APIs, etc. I am cleaning up the code now, and will try to learn SourceKitten, etc. by reading the code.

Once it's cleaned up, would you review it? If so, I assume I should make a new issue for it (as I haven't found an existing one), and move this discussion to that issue.

Would it supersede this rule? What configuration options would you prefer? What names would you prefer be used? e.g., should it be the `deprecated_uses` rule, as it will report only uses of deprecated declarations, not deprecated declarations themselves?

I assume it would be helpful to filter out deprecated uses where the replacement is not available for the current project (e.g., based on the minimum supported platform versions), but I haven't determined if it's easy to obtain that info. If such filtering were provided, should the rule be able to read the minimum platform versions from a `Package.swift`? What about `Package@….swift`, where `…` is a swift version? What about other build systems?

Should there be some way to globally ignore uses of specified declarations? If so, should they be specified via strings like "URL.init(fileURLWithPath:isDirectory:)"?

Should the rule try to detect the replacement declaration, and output its details in the violation message (I assume it will be too difficult to apply fixes via `--fix`).

Should the rule detect deprecated declarations via additional mechanisms? e.g., should it support manual replacement mappings (e.g., `NSString.replacingOccurrences(of:with:)` > `RangeReplaceableCollection.replacing(_:with:maxReplacements:)`, as the former isn't marked with `@available`)?

Etc.

SimplyDanny:
> I have created an analyzer rule that reports all uses of any deprecated declarations (marked with `@available`), including soft deprecations, etc.
> 
> The types mentioned in this PR all seem like they should be detectable by my analyzer rule.
> 
> I used AI to create it, as I don't know SourceKitten, SourceKit, SwiftLint APIs, etc. I am cleaning up the code now, and will try to learn SourceKitten, etc. by reading the code.
> 
> Once it's cleaned up, would you review it? If so, I assume I should make a new issue for it (as I haven't found an existing one), and move this discussion to that issue.
> 
> Would it supersede this rule? What configuration options would you prefer? What names would you prefer be used? e.g., should it be the `deprecated_uses` rule, as it will report only uses of deprecated declarations, not deprecated declarations themselves?
> 
> I assume it would be helpful to filter out deprecated uses where the replacement is not available for the current project (e.g., based on the minimum supported platform versions), but I haven't determined if it's easy to obtain that info. If such filtering were provided, should the rule be able to read the minimum platform versions from a `Package.swift`? What about `Package@….swift`, where `…` is a swift version? What about other build systems?
> 
> Should there be some way to globally ignore uses of specified declarations? If so, should they be specified via strings like "URL.init(fileURLWithPath:isDirectory:)"?
> 
> Should the rule try to detect the replacement declaration, and output its details in the violation message (I assume it will be too difficult to apply fixes via `--fix`).
> 
> Should the rule detect deprecated declarations via additional mechanisms? e.g., should it support manual replacement mappings (e.g., `NSString.replacingOccurrences(of:with:)` > `RangeReplaceableCollection.replacing(_:with:maxReplacements:)`, as the former isn't marked with `@available`)?
> 
> Etc.

Both rules (and more existing ones) can live side by side. As Analyzer rules run very slow, the syntax-based rules are faster alternatives that might be sufficient for some/most users.

Let's chat about the other questions you have in the PR that you are about to create. A dedicated issue is not needed. 😉

rgoldberg:
Will do. I will create a draft PR later today if you want to look over the code before I have a chance to completely understand SourceKit & ensure that the PR is up to snuff. I'll update the code & switch to a normal PR once I've had the chance to completely look through it all.

I'll also keep everything barebones to get feedback first, then add bells & whistles once everything has been discussed.

rgoldberg:
FYI: the AI code was worse than I expected. I need to completely redo most of it, and to research beforehand. I will get back to it in a bit, then submit a PR once it's much better.

SimplyDanny (approved):

Issue #6283, opened 2025-10-07
`async_without_await` triggers on `@concurrent` functions without `await`


SimplyDanny:
The rule is supposed to flag `async` functions which don't contain any `await` expression/statement. However, `@concurrent` functions in Swift 6.2 require the `async` in any case. The rule needs to exclude `@concurrent` functions.

Example:

```swift
@concurrent
func f() async {} // rule triggers here even though it must not
```

PR #6284, opened 2025-10-07
Add logic to exclude async without await rule for concurrent function

nandhinisubbu:
Fix - https://github.com/realm/SwiftLint/issues/6283

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,8 +66,8 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#6253](https://github.com/realm/SwiftLint/issues/6235)
 
-* Exclude @concurrent function from `async_without_await` rule analysis.
-  @concurrent function requires aysnc in any case
+* Exclude `@concurrent` functions from `async_without_await` rule analysis.
+  `@concurrent` functions requires `aysnc` in any case.  
   [nandhinisubbu](https://github.com/nandhinisubbu)
   [#6283](https://github.com/realm/SwiftLint/issues/6283)
 
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/AsyncWithoutAwaitRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/AsyncWithoutAwaitRule.swift
@@ -19,7 +19,6 @@ private extension AsyncWithoutAwaitRule {
     private struct FuncInfo {
         var containsAwait = false
         let asyncToken: TokenSyntax?
-        let isConcurrent: Bool
     }
 
     final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -66,6 +66,11 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#6253](https://github.com/realm/SwiftLint/issues/6235)
 
+* Exclude @concurrent function from `async_without_await` rule analysis.
+  @concurrent function requires aysnc in any case
```

```suggestion
* Exclude `@concurrent` functions from `async_without_await` rule analysis.
  `@concurrent` functions require `async` in any case.  
```

Source/SwiftLintBuiltInRules/Rules/Lint/AsyncWithoutAwaitRule.swift
```diff
@@ -112,7 +117,7 @@ private extension AsyncWithoutAwaitRule {
             }
             if pendingAsync == node.bindings.onlyElement?.typeAnnotation?.type.asyncToken {
```
SimplyDanny, line 120: There is also the case of async closures, that we need to consider. For example:

```swift
let c: () async -> Int = { @concurrent in 1 }
```

Source/SwiftLintBuiltInRules/Rules/Lint/AsyncWithoutAwaitRuleExamples.swift
```diff
@@ -160,6 +160,26 @@ internal struct AsyncWithoutAwaitRuleExamples {
...
+            func alwaysSwitch() async {
+                // This is valid - @concurrent functions require async even without await
+            }
+        }
+        """),
+        Example("""
+        @concurrent
+        init() async {
+            setup()
+        }
```
SimplyDanny, line 189: Let's wrap that into a type to make it proper Swift.

Source/SwiftLintBuiltInRules/Rules/Lint/AsyncWithoutAwaitRule.swift
```diff
@@ -31,7 +32,9 @@ private extension AsyncWithoutAwaitRule {
             }
 
             let asyncToken = node.signature.effectSpecifiers?.asyncSpecifier
-            functionScopes.push(.init(asyncToken: asyncToken))
+            // @concurrent functions require the async keyword even without await calls
+            let isConcurrent = node.attributes.contains(attributeNamed: "concurrent")
+            functionScopes.push(.init(asyncToken: asyncToken, isConcurrent: isConcurrent))
```
SimplyDanny: Perhaps a bit easier than this would be to add `.init(asyncToken: nil)` to the stack in case there is a `@concurrent` attribute. It would cause the same outcome but avoid the extra `isConcurrent` property.

SimplyDanny (changes requested):
Thank you for the prompt fix! I came up with this other case for closures that we should consider too.

Source/SwiftLintBuiltInRules/Rules/Lint/AsyncWithoutAwaitRuleExamples.swift
nandhinisubbu: Thanks for the suggestion. I have update it

CHANGELOG.md
nandhinisubbu: Thank you, have updated.

Source/SwiftLintBuiltInRules/Rules/Lint/AsyncWithoutAwaitRule.swift
nandhinisubbu: Thank you for the suggestion, have made the changes. Please have a look.

PR #6288, opened 2025-10-09
[WIP]: prefer unknown default rule experiments

jamieQ:


jamieQ:
@SimplyDanny a (low priority) question for you regarding the CI failures and warnings here – this rule was motivated by the discussion [here](https://forums.swift.org/t/anything-similar-to-wswitch-enum-for-swift/82542) and would provide a mechanism to check that switch statements do not have 'bare' `default` cases to encourage explicit handling of enumeration cases where possible (or explicit opt-outs via a lint rule suppression).

the integration tests currently appear to be failing due to this issue:

```
...
/Tests/IntegrationTests/IntegrationTests.swift:58: error: -[IntegrationTests.IntegrationTests testSwiftLintAutoCorrects] : XCTAssertTrue failed - Unexpected corrections have been applied
```

is there test fixture i need to opt out of using this rule when it runs those tests, or what is the expected interaction there?

additionally there are... a lot of Danger warnings about projects for which this rule would introduce new warnings to if enabled. are these expected? i would expect use of this to be fairly niche, and definitely not something enabled by default – i tried to set that up appropriately in the rule's configuration, so wondering if i've missed something. thanks in advance for your time!

SimplyDanny:
> the integration tests currently appear to be failing due to this issue:
> 
> ```
> ...
> /Tests/IntegrationTests/IntegrationTests.swift:58: error: -[IntegrationTests.IntegrationTests testSwiftLintAutoCorrects] : XCTAssertTrue failed - Unexpected corrections have been applied
> ```
>
> is there test fixture i need to opt out of using this rule when it runs those tests, or what is the expected interaction there?

`IntegrationTests`'s job is to run SwiftLint on its own code base respecting its `.swiftlint.yml` configuration. You may let it run locally to see why it fails.

I approved and ran all actions, so that you can now also see (in the "Files changed" tab of the PR) where the rule would trigger in SwiftLint itself. The fix is to disable the rule in the configuration for this repository, as all rules are enabled by default.
 
> additionally there are... a lot of Danger warnings about projects for which this rule would introduce new warnings to if enabled. are these expected? i would expect use of this to be fairly niche, and definitely not something enabled by default – i tried to set that up appropriately in the rule's configuration, so wondering if i've missed something. thanks in advance for your time!

The setup is correct. When we scan OSS repositories, the rule is enabled in any case though. The intention is to get an idea of which effect a code change would have on these projects. It's quite helpful to disclose special cases you haven't thought of in your initial implementation. Important is to verify that all these findings are valid, i.e., they would be expected if one were to enable the rule in the repository.

jamieQ:
looks like we should expect a variant of this to be built into the compiler that will have access to more semantic info and presumably will be able to be conditionally enabled via a warning flag: https://github.com/swiftlang/swift/pull/85168. so, closing this.

jamieQ:
@SimplyDanny i meant to say earlier – it was a nice experience attempting a contribution to this repo. everything was very straightforward, and the test suite stuff was very impressive (showing the results of new rules against a large swath of OSS code is very cool). thank you for your work here! 🙇

SimplyDanny:
> @SimplyDanny i meant to say earlier – it was a nice experience attempting a contribution to this rep. everything was very straightforward, and the test suite stuff was very impressive (showing the results of new rules against a large swatch of OSS code is very cool). thank you for your work here! 🙇

Glad to hear you had an easy start. That's music to my ears. Thank you!

The rule you proposed is clearly hard to implement sufficiently reliably in a syntax-based linter. It's much better hosted in the compiler itself. But feel free to contribute at other ends. 😉

PR #6290, opened 2025-10-10
Add new `incompatible_concurrency_annotation` rule

### LINKED ISSUE #5987: Rule to check for Swift 5-mode-incompatible API
### New Issue Checklist

- [ ] I've Updated SwiftLint to the latest version.

There are a number of ways to break compatibility with Swift 5 mode (strict concurrency off) that are difficult to notice. Here are some examples:

```swift
@MainActor
class MainThreadOnly {}

func takeBlock(_ block: @Sendable () - Void)
func takeGlobalBlock(_ block: @MainActor () - Void)
```

In all of these cases, to preserve compatibility you must use `@preconcurrency`.

```swift
@preconcurrency @MainActor
class MainThreadOnly {}

@preconcurrency 
func takeBlock(_ block: @Sendable () - Void)

@preconcurrency 
func takeGlobalBlock(_ block: @MainActor () - Void)
```

I think this could be a great new rule!

SimplyDanny:
Closes #5987.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/IncompatibleConcurrencyAnnotationRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/IncompatibleConcurrencyAnnotationRule.swift
@@ -125,7 +125,7 @@ private func preconcurrencyRequired(for syntax: some WithModifiersSyntax & WithA
             if required { return true }
         }
 
-        // Check parameters for `@Sendable`, `sending` and global actors.
+        // Check parameters for `@Sendable` and global actors.
         let parameterClause = syntax.as(FunctionDeclSyntax.self)?.signature.parameterClause
             ?? syntax.as(InitializerDeclSyntax.self)?.signature.parameterClause
             ?? syntax.as(SubscriptDeclSyntax.self)?.parameterClause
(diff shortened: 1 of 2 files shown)


mattmassicotte:
This looks super cool! Thanks so much for doing it.

I didn't even think of the generic constraint case! There could be others, like the use of the `sending` keyword. I'm not exactly sure and it would take some experimenting to figure out. But this is wonderful regardless!

SimplyDanny:
> There could be others, like the use of the `sending` keyword. I'm not exactly sure and it would take some experimenting to figure out.

I've added checking for `sending` keywords. However, I think that it's not needed because of the statement about source compatibility in the [introducing proposal](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0430-transferring-parameters-and-results.md):

> In the Swift 5 language mode, `sending` diagnostics are suppressed under minimal concurrency checking, and diagnosed as warnings under strict concurrency checking. The diagnostics are errors in the Swift 6 language mode, as shown in the code examples in this proposal. This diagnostic behavior based on language mode allows `sending` to be adopted in existing Concurrency APIs including `CheckedContinuation`.

Would you agree, @mattmassicotte? I did some experiments as well and haven't seen any changes with or without `@preconcurrency`. That doesn't mean much, though, as I'm not sure whether my examples have been sufficient.

mattmassicotte:
Yeah I wasn't sure about `sending`, parameters in particular. But that text does seem unambiguous, and the biggest culprits are global actors and `@Sendable` anyways. So I think this will provide very good coverage.

Thanks again for doing this, super cool!

PR #6291, opened 2025-10-11
Set 512KiB to thread stack size for static Linux binary

### LINKED ISSUE #6287: The new binaries built with Static Linux SDK crashes
### New Issue Checklist

### Bug Description

> A fully statically linked binary named swiftlint-static that does not require
any dynamic libraries at runtime. Rules requiring SourceKit will be disabled and
reported to the console when running this binary.

https://github.com/realm/SwiftLint/releases/tag/0.61.0

I have tried this new binary for Linux hoping that we can simply run SwiftLint on plain ubuntu env on GitHub Actions.
It crashed with segmentation fault when linting some code. This seems to occur on both AMD64 and ARM64.

```
$ ./swiftlint-static
Linting Swift files in current working directory...
....
Linting 'xxxx.swift' (57/616)
Segmentation fault (core dumped)
```

To reproduce the issue, I did set up Linux environment with Swiftly. I managed to get minimum input file that can reproduce the crash on linting and some backtrace by building static linux sdk binary with 0.61.0 tag locally.

Environment (I used VMware Fusion with Ubuntu Desktop image on macOS so it's ARM env)

```
ainame@ainame-vm:~/repos/SwiftLint$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 24.04.3 LTS
Release:	24.04
Codename:	noble
ainame@ainame-vm:~/repos/SwiftLint$ swiftly list
Installed release toolchains
----------------------------
Swift 6.1.2 (in use) (default)

Installed snapshot toolchains
-----------------------------
ainame@ainame-vm:~/repos/SwiftLint$ swift --version
Swift version 6.1.2 (swift-6.1.2-RELEASE)
Target: aarch64-unknown-linux-gnu
ainame@ainame-vm:~/repos/SwiftLint$ swift sdk list
swift-6.1.2-RELEASE_static-linux-0.0.1
```

```
ainame@ainame-vm:~/repos/SwiftLint$ swift build --product swiftlint -Xswiftc -DSWIFTLINT_DISABLE_SOURCEKIT --swift-sdk aarch64-swift-linux-musl -Xswiftc -g
ainame@ainame-vm:~/repos/SwiftLint$ lldb ./.build/debug/swiftlint
(lldb) target create "./.build/debug/swiftlint"
Current executable set to '/home/ainame/repos/SwiftLint/.build/debug/swiftlint' (aarch64).
(lldb) run --no-cache -- ./Bug.swift
Process 26811 launched: '/home/ainame/repos/SwiftLint/.build/debug/swiftlint' (aarch64)
Linting Swift files at paths ./Bug.swift
Linting 'Bug.swift' (1/1)
Process 26811 stopped
* thread #7, name = 'swiftlint', stop reason = signal SIGSEGV: invalid permissions for mapped object (fault address: 0xfffff7d34dc8)
    frame #0: 0x00000000032202f0 swiftlint`Keyword.init(text=SwiftSyntax.SyntaxText @ 0x0000fffff7d36dc8) at Keyword.swift:228
   225 	  case wrt
   226 	  case yield
   227 	
-> 228 	  @_spi(RawSyntax) public init?(_ text: SyntaxText) {
   229 	    switch text.count {
   230 	    case 2:
   231 	      switch text { 
```

Full backtrace from lldb is here
https://gist.github.com/ainame/9a5851a2fd5dc32b0189d38b6bbe7c57

`Bug.swift` that I used above is this (redacted one from actual source code from our project but still make it crash).

```swift
import Foundation
import UIKit

class CustomTextField: UIView {
    private lazy var stackView: UIStackView = {
        let stack = UIStackView(
            axis: .vertical,
            alignment: .fill,
            distribution: .fill,
            spacing: .s,
            subviews: [
                UIStackView(
                    axis: .vertical,
                    alignment: .fill,
                    distribution: .fill,
                    subviews: []
                ),
            ]
        )
        return stack
    }()
}
```

This of course doesn't make SwiftLint crash on macOS nor Ubuntu with normal debug build binary.

### Environment

* SwiftLint version 0.61.0 / swiftlint-static for ARM64/AMD64 
* Xcode version N/A
* Installation method used from GitHub Release page
* Configuration file: The issue persists with no config file

ainame:
Closes #6287 

# Background

In #6287, I reported crashes occurred with 0.61.0's new fully static Linux binaries. 
Since then I managed to replicate the crashes on GitHub Actions.

In short, it's easy to replicate if you run `swiftlint-static` against https://github.com/apple/swift-nio or https://github.com/pointfreeco/isowords repos.
https://github.com/ainame/swiftlint-static-test/actions/runs/18432502301

I also did debug `swiftlint-static` with lldb and identified the root cause, which was stack exhaustion.
https://github.com/realm/SwiftLint/issues/6287#issuecomment-3388109177

# What 

Static Linux SDK comes with musl and musl has limited thread stack size by default - 128KiB, which is smaller than the default size for Darwin's pthread 512KiB.

> musl provides a default thread stack size of 128k (80k prior to 1.1.21).

https://wiki.musl-libc.org/functional-differences-from-glibc.html 

```c
// Default stack size is 512KB; independent of the main thread's stack size.
#define DEFAULT_STACK_SIZE (size_t)(512 * 1024)
```

https://github.com/apple-oss-distributions/libpthread/blob/1ebf56b3a702df53213c2996e5e128a535d2577e/src/pthread.c#L84-L85

I think SwiftLint running on Linux should have the same default stack size as on macOS, 
so it can lint source files of equivalent complexity.

# How 

> Since 1.1.21, musl supports increasing the default thread stack size via the PT_GNU_STACK program header, which can be set at link time via -Wl,-z,stack-size=N.

Thanks to this we can specify default thread stack size via `swift build` command.
I updated `release.yml` to use 512KiB, like below

```
swift build \
  -c release \
  --product swiftlint \
  --swift-sdk ${{ matrix.swift_sdk }} \
  -Xswiftc -DSWIFTLINT_DISABLE_SOURCEKIT \
  -Xlinker -z -Xlinker stack-size=0x80000 # 512KiB
```

# Test

https://github.com/ainame/swiftlint-static-test/actions/runs/18432924761

There are four jobs in matrix

* 0.61.0's swiftlint-static for x86_64
* 0.61.0's swiftlint-static for aarch64
* Manually built and patched swiftlint-static for x86_64 with stack-size set to 512KiB
* Manually built and patched swiftlint-static for aarch64 with stack-size set to 512KiB
(see details of the compile options and the patch here https://github.com/ainame/swiftlint-static-test/releases/tag/test3)

If `swiftlint-static lint` successfully exit (0 or 2), the job succeeds.
And thanks to the small patch, you can actually see 512KiB stack size is used for each DispatchQueue.

```
Run cd swift-nio-main
[dispatch queue main]: pthread stack size = 526912 bytes (≈ 0.50 MiB)
[dispatch queue 0]: pthread stack size = 526912 bytes (≈ 0.50 MiB)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,7 +86,7 @@
 * `swiftlint-static`, the experimental fully-static Linux binary, now uses 512 KiB
   as thread stack size matching Darwin’s size to prevent stack exhaustion.  
   [ainame](https://github.com/ainame)
-  [#6291](https://github.com/realm/SwiftLint/pull/6291)
+  [#6287](https://github.com/realm/SwiftLint/pull/6287)
 
 ## 0.61.0: Even Fresher Breeze
 


.github/workflows/release.yml
```diff
@@ -145,7 +145,13 @@ jobs:
...
         run: swift sdk install https://download.swift.org/swift-6.1.2-release/static-sdk/swift-6.1.2-RELEASE/swift-6.1.2-RELEASE_static-linux-0.0.1.artifactbundle.tar.gz --checksum df0b40b9b582598e7e3d70c82ab503fd6fbfdff71fd17e7f1ab37115a0665b3b
       - name: Build static binary
-        run: swift build -c release --product swiftlint --swift-sdk ${{ matrix.swift_sdk }} -Xswiftc -DSWIFTLINT_DISABLE_SOURCEKIT
+        run: |
+          swift build \
+            -c release \
+            --product swiftlint \
+            --swift-sdk ${{ matrix.swift_sdk }} \
+            -Xswiftc -DSWIFTLINT_DISABLE_SOURCEKIT \
+            -Xlinker -z -Xlinker stack-size=0x80000 # 512KiB
```
SimplyDanny, line 154: Did you try to run this on a Linux machine? It doesn't work on macOS, because the linker expects different arguments to set the stack size.

ainame: Yes I have created this page's binaries with this command on Ubuntu (VM). 
https://github.com/ainame/swiftlint-static-test/releases/tag/test3

`-z stack-size=VALUE` format is supported by `ld`. Apple's `ld64` doesn't support it unfortunately. But that's okay since the binary will be built on ubuntu-24.04/ubuntu-24.04-arm runner, right? 

```
% lld --help
lld is a generic driver.
Invoke ld.lld (Unix), ld64.lld (macOS), lld-link (Windows), wasm-ld (WebAssembly) instead
% ld64.lld --help | grep stack
%
```

>  stack-size=value
>  Specify  a  stack  size  for  an  ELF  "PT_GNU_STACK" segment.  Specifying zero will override any
>  default non-zero sized "PT_GNU_STACK" segment creation.

https://manpages.ubuntu.com/manpages/plucky/man1/ld.1.html

ainame:
I updated CHANGELOG.md as well🙏

CHANGELOG.md
```diff
@@ -83,6 +83,11 @@
   [nandhinisubbu](https://github.com/nandhinisubbu)
   [#6283](https://github.com/realm/SwiftLint/issues/6283)
 
+* `swiftlint-static`, the experimental fully-static Linux binary, now uses 512 KiB
+  as thread stack size matching Darwin’s size to prevent stack exhaustion.
```
```suggestion
  as thread stack size matching Darwin’s size to prevent stack exhaustion.  
```

.github/workflows/release.yml
SimplyDanny: Yes, that's totally fine. I build the binaries on macOS from time to time for testing purposes. So far they worked without the stack size increase. For the release, that's Ubuntu's job. I may trigger a new release right after merging your PR, so that we can verify the binaries again.

CHANGELOG.md
```diff
@@ -83,6 +83,11 @@
   [nandhinisubbu](https://github.com/nandhinisubbu)
   [#6283](https://github.com/realm/SwiftLint/issues/6283)
 
+* `swiftlint-static`, the experimental fully-static Linux binary, now uses 512 KiB
+  as thread stack size matching Darwin’s size to prevent stack exhaustion.  
+  [ainame](https://github.com/ainame)
+  [#6291](https://github.com/realm/SwiftLint/pull/6291)
```
```suggestion
  [#6287](https://github.com/realm/SwiftLint/pull/6287)
```

.github/workflows/release.yml
ainame: Oh `-Xlinker -z -Xlinker stack-size=0x80000` worked with other project indeed.

I assume that the current toolchains might have applied the linker options for `SwiftLintCoreMacros` target, which is built and run for macOS? and failed. I haven't checked logs or any though.

SimplyDanny: Yes, it exactly failed while linking the macro tool. I'm not sure how to configure the flags in more detail in order to determine to which steps they should be applied to with SPM.

SimplyDanny:
[There](https://github.com/realm/SwiftLint/releases/tag/0.62.0) we go. I noticed that the static binaries lint significantly slower than the dynamic ones. That's also something requiring attention before they can leave their experimental state.

ainame:
I did casual benchmark and tried a potential fix https://github.com/microsoft/mimalloc. It's promising already. 
I can try some investigation if you don't mind

https://github.com/realm/SwiftLint/issues/6298

SimplyDanny:
> I can try some investigation if you don't mind

That'd be highly appreciated! Please go ahead. 🙏

PR #6306, opened 2025-10-16
Fix: `closure_end_indentation` false positive with macro expansion expressions

gibachan:
## Summary

This PR addresses a false positive in the `closure_end_indentation` rule when the closure is part of a macro expansion expression, such as the `#Preview { ... }` macro.

## Details

Prior to this fix, the rule would incorrectly flag an indentation violation for the closing brace (`}`) of a closure used within a  preview macro expansion expression, particularly when arguments were present before the trailing closure and spanned multiple lines.

For example, the following code, which should be valid, was incorrectly flagging an indentation violation on the final }:

```swift
        #Preview("foo",
                 traits: .landscapeLeft) {
            ZStack {}
        }
```

The issue was introduced in SwiftLint 0.60.0, as it worked correctly in version 0.59.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,8 @@
 
 ### Bug Fixes
 
-* Fix false positive in closure_end_indentation rule with #Preview macro
+* Fix false positive with `#Preview` macro in `closure_end_indentation` rule.  
   [gibachan](https://github.com/gibachan)
-  [#6306](https://github.com/realm/SwiftLint/pull/6306)
 
 ## 0.62.1: Generous Drum Volume
 


CHANGELOG.md
```diff
@@ -16,7 +16,9 @@
 
 ### Bug Fixes
 
-* None.
+* Fix false positive in closure_end_indentation rule with #Preview macro
+  [gibachan](https://github.com/gibachan)
+  [#6306](https://github.com/realm/SwiftLint/pull/6306)
```
```suggestion
* Fix false positive with `#Preview` macro in `closure_end_indentation` rule.  
  [gibachan](https://github.com/gibachan)
```

Issue #6304, opened 2025-10-15
Option for `line_length` to ignore regex literals


rgoldberg:
### New Issue Checklist

- [x] I've Updated SwiftLint to the latest version.
- [x] I've searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues).

### Feature or Enhancement Proposal

Add an option for `line_length` to ignore regex literals, maybe named `ignores_regex_literals`.

PR #6307, opened 2025-10-16
Add new option ignores regex literal in line length rule

nandhinisubbu:
Enhancement - https://github.com/realm/SwiftLint/issues/6304


SimplyDanny (approved):
Very good. Thank you @nandhinisubbu!

Issue #6167, opened 2025-07-21
`implicit_return` doesn't trigger for if/switch expressions


lukaskollmer:
### Bug Description

The `implicit_return` rule does not trigger in cases where an `if` or `switch` is used as the only expression in a function, and each branch consists of only a single `return` statement:

```swift
// This does not trigger a violation:
func test() -> String {
    if .random() {
        return "A"
    } else {
        return "B"
    }
}
```

Mention the command or other SwiftLint integration method that caused the issue. Include stack traces or command output.

```bash
$ swiftlint
```

### Environment

* SwiftLint version: 0.59.1
* Xcode version: Xcode 16.4 (16F6)
* Installation method used: built from source
* Configuration file:
    ```yml
    enabled_rules:
      - implicit_return
    ```
- no nested configurations

PR #6310, opened 2025-10-17
Update Implicit return to check if/switch for single expression

nandhinisubbu:
Enhancement - https://github.com/realm/SwiftLint/issues/6167


SimplyDanny (changes requested):
That's cool. Thanks @nandhinisubbu!

Logically, this looks about right. Did you consider `if` expressions as the only statements in `switch` cases and vice versa? I think, there should be some kind of recursion to cover that.

With respect to code structure, I'd add all the functions with a single syntax node argument as computed properties to the syntax node itself. That also separates them a little from the rule's implementation, so that the visitor as the entry point is rather small and only provides the entry points.

Other than that, we should probably make checking of `if` and `switch` expression optional. I can imagine, not everyone wants to have them checked. Then, for SwiftLint itself, keep the option off and enable to only in a follow-up PR to separate the concerns.

Issue #6164, opened 2025-07-14
isolation parameter for modifier_order rule


MartinColja:
### New Issue Checklist

- [x] I've Updated SwiftLint to the latest version.
- [x] I've searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues).

### Feature or Enhancement Proposal

I implemented my own `preferred_modifier_order` and I wanted to specify the order for `isolated` and `nonisolated` keywords, but I didn't find that they were available in the [current implementation](https://github.com/realm/SwiftLint/blob/c1ffdfe8913c12398a3a05fe19844c1031fa7add/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/ModifierOrderConfiguration.swift#L11) or [documentation](https://realm.github.io/SwiftLint/modifier_order.html), and I didn't find any open issues.

I think a parameter like `isolation` would be a welcome addition for fine-tuning the modifier order.

PR #6316, opened 2025-10-21
Add isolation parameter for modifier_order rule

nandhinisubbu:
Enhancement - https://github.com/realm/SwiftLint/issues/6164

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/ModifierOrderConfiguration.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/ModifierOrderConfiguration.swift
@@ -10,9 +10,9 @@ struct ModifierOrderConfiguration: SeverityBasedRuleConfiguration {
     @ConfigurationElement(key: "preferred_modifier_order")
     private(set) var preferredModifierOrder: [SwiftDeclarationAttributeKind.ModifierGroup] = [
         .override,
+        .isolation,
         .acl,
         .setterACL,
-        .isolation,
         .dynamic,
         .mutators,
         .lazy,
--- a/Source/SwiftLintCore/Extensions/NSRegularExpression+SwiftLint.swift
+++ b/Source/SwiftLintCore/Extensions/NSRegularExpression+SwiftLint.swift
@@ -2,7 +2,7 @@ import Foundation
 import SourceKittenFramework
 
 private let regexCacheLock = NSLock()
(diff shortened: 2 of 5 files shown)


Tests/FrameworkTests/ModifierOrderTests.swift
```diff
@@ -388,4 +390,92 @@ final class ModifierOrderTests: SwiftLintTestCase {
...
+                }
+                """),
+            ])
+            .with(corrections: [:])
+
+        verifyRule(descriptionOverride,
+                   ruleConfiguration: ["preferred_modifier_order": ["override", "acl", "isolation", "final"]])
+    }
+
+    func testIsolationModifierCorrections() {
```
SimplyDanny: Should be fine to have these corrections as a part of the previous test.

Tests/FrameworkTests/ModifierOrderTests.swift
```diff
@@ -388,4 +390,92 @@ final class ModifierOrderTests: SwiftLintTestCase {
...
+            ])
+            .with(triggeringExamples: [
+                Example("""
+                @MainActor
+                class Foo {
+                    public nonisolated func bar() {}
+                }
+                """),
+            ])
+            .with(corrections: [:])
```
SimplyDanny: Let's have a correction example here as well.

nandhinisubbu: Added, Thanks for the suggestion!

Tests/FrameworkTests/ModifierOrderTests.swift
nandhinisubbu: Updated. Thanks!

nandhinisubbu:
> Looks good. Thanks @nandhinisubbu! There's only nits ...
> 
> From my experience, `nonisolated` is often first (after `override`, before ACL) in a declaration. Perhaps we should use this as the default.

Thanks for the suggestion @SimplyDanny!  Have given this order for the default test. Followed the previous order for the custom order test so it tests the opposite scenario.

SimplyDanny:
> > Looks good. Thanks @nandhinisubbu! There's only nits ...
> 
> > 
> 
> > From my experience, `nonisolated` is often first (after `override`, before ACL) in a declaration. Perhaps we should use this as the default.
> 
> 
> 
> Thanks for the suggestion @SimplyDanny!  Have given this order for the default test. Followed the previous order for the custom order test so it tests the opposite scenario.

The rule should also follow this order by default. So the configuration needs an update too.

PR #6318, opened 2025-10-24
Fix correction of `sorted_imports` rule for leading comments case

### LINKED ISSUE #6317: sorted_imports in 0.62.0 doesn't correctly distinguish groups
### Bug Description

The rewritten `sorted_imports` rule doesn't correctly distinguish import groups that also contain comments.  Leading comments stay with the import node they're associated with, but I'd expect that to break on newlines.

Bad correction I get when I expected the original code not to trigger a violation:

```diff
+// comment for second group
+import AAA
 //
 // Copyright © Foo
 //
 
 import DDD
 import FFF
-
-// comment for second group
-import AAA
 import BBB
 import CCC
```

Related example where the correction is placed above the first item's trivia when I would expect it to split on the new line:

```diff
+import AAA
 //
 //  Copyright © Foo
 //
 
 import class BBB.View
-import AAA
```

### Environment

* SwiftLint 0.62.1
* Xcode 26.0.1
* installed from source

SimplyDanny:
Fixes #6317.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,11 @@
 * Fix false positive with `#Preview` macro in `closure_end_indentation` rule.  
   [gibachan](https://github.com/gibachan)
 
+* Fix correction of `sorted_imports` rule when comments (with a distance) are present before
+  the first import.  
+  [SimplyDanny](https://github.com/SimplyDanny)
+  [#6317](https://github.com/realm/SwiftLint/issues/6317)
+
 ## 0.62.1: Generous Drum Volume
 
 ### Breaking
--- a/Source/SwiftLintBuiltInRules/Rules/Style/SortedImportsRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/SortedImportsRule.swift
@@ -34,7 +34,7 @@ private extension SortedImportsRule {
             var importBlocks = [[Import]]()
             for `import` in imports {
                 if let lastBlock = importBlocks.last, let lastImport = lastBlock.last {
-                    if `import`.isDirectlyAfter(other: lastImport, in: file) {
+                    if `import`.isDirectlyAfter(previous: lastImport, in: file) {
                         importBlocks[importBlocks.count - 1].append(`import`)
                     } else {
                         importBlocks.append([`import`])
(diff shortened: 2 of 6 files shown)


jszumski:
I'm still getting some unexpected diffs with this change:

```diff
 //
 // Copyright © Foo
 //
 
+import AAA // a comment
 import DDD
 import FFF
-
-// some comment
-import AAA // a comment
 import NNN
 import RRR
 import SSS
```

and

```diff
 //
 // Copyright © Foo
 //
 
 import DDD
-import SSS
 
 // some comment
 import FFF // a comment
+import SSS
```

SimplyDanny:
> I'm still getting some unexpected diffs with this change:
> 
> ```diff
>  //
>  // Copyright © Foo
>  //
>  
> +import AAA // a comment
>  import DDD
>  import FFF
> -
> -// some comment
> -import AAA // a comment
>  import NNN
>  import RRR
>  import SSS
> ```
> 
> and
> 
> ```diff
>  //
>  // Copyright © Foo
>  //
>  
>  import DDD
> -import SSS
>  
>  // some comment
>  import FFF // a comment
> +import SSS
> ```

They should all be fine with the latest update. Are they, @jszumski?

jszumski:
So close! just this one edge case left:

```diff
+import BBB
 //
 // Copyright © Foo
 //
 
 import class CCC
-import BBB
 import LLL
```

I guess if the trivia of the old first item ends with a newline then a replacement into the first position should append to that old trivia?

SimplyDanny:
> I guess if the trivia of the old first item ends with a newline then a replacement into the first position should append to that old trivia?

Yeah, sort of that. 😅 The condition was too vague. The case of an empty line between the comment and the first imports wasn't decently handled.

Could you give it another try? That really helps getting this properly sorted (pun indented).

jszumski:
Also close, but if any replacement is made in a position >1 it chomps the first position's trivia:

```diff
 //
 //  Copyright © Foo
 //
-
 import AAA
-import class CCC2.View
 import CCC1
+import class CCC2.View
```

SimplyDanny:
Next try ...

ebubedimobi:
Good day, when can we expect new release with this fix?

SimplyDanny:
> Good day, when can we expect new release with this fix?

It will be ready within the next few days.

Issue #6161, opened 2025-07-11
JUnit Reporter Missing Test Count


kbarnesMCC:
### New Issue Checklist

- [x] I've Updated SwiftLint to the latest version.
- [x] I've searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues).

### Bug Description

Coming out of [PR #4725](https://github.com/realm/SwiftLint/pull/4725) updates were made to enhance the JUnit reporting with additional details that are necessary for some pipeline engines; specifically Bitbucket.

Unfortunately, one aspect of that PR was seemingly dropped, namely the "total tests count" which we're finding appears to be a requirement for Bitbucket ingestion in particular. For example, we're getting this message as Bitbucket's JUnit parsing logic:

```Merged test suites, total number tests is 0, with 0 failures and 2 errors.```

While it properly detects the presence of failures/errors per that referenced PR's changes; it's detecting 0 tests which we believe it causing them to disable all their UI integrations for reporting from this file.

The [JUnit Complete Reference Example](https://github.com/testmoapp/junitxml?tab=readme-ov-file#complete-junit-xml-example) shows this as simply adding a "tests" attribute to both "testsuites" and "testsuite" XML nodes.

Additionally, I suspect the "testsuite" XML nodes need a "name" property for effective rendering. Most other tools I've seen (e.g. eslint in the Javascript world) would populate this simply as the filename though depending on how Swiftlint marshals this file maybe the linting rule is more appropriate or even just a single "SwiftLint" name if everything is marshalled under a single "testsuite".

### Environment

* SwiftLint version: 0.59.1
* Xcode version: 16.4 Build 16F6
* Installation method used: Homebrew
* Configuration file:

```yml
disabled_rules:
  - file_length
  - function_body_length
  - line_length
  - type_body_length

opt_in_rules:
  - conditional_returns_on_newline
  - contains_over_filter_count
  - contains_over_filter_is_empty
  - contains_over_first_not_nil
  - contains_over_range_nil_comparison
  - empty_collection_literal
  - empty_string
  - explicit_init
  - fatal_error_message
  - first_where
  - flatmap_over_map_reduce
  - identical_operands
  - implicit_return
  - joined_default_parameter
  - last_where
  - legacy_multiple
  - let_var_whitespace
  - multiline_function_chains
  - multiline_literal_brackets
  - object_literal
  - optional_enum_case_matching
  - prefer_self_in_static_references
  - prefer_self_type_over_type_of_self
[...]

PR #6319, opened 2025-10-25
Add test count to JUnit reporter

nandhinisubbu:
Enhancement - https://github.com/realm/SwiftLint/issues/6161

### DIFF
--- a/Source/SwiftLintFramework/Reporters/JUnitReporter.swift
+++ b/Source/SwiftLintFramework/Reporters/JUnitReporter.swift
@@ -11,7 +11,7 @@ struct JUnitReporter: Reporter {
     static func generateReport(_ violations: [StyleViolation]) -> String {
         let warningCount = violations.filter({ $0.severity == .warning }).count
         let errorCount = violations.filter({ $0.severity == .error }).count
-        let testCount = violations.count
+        let testCount = warningCount + errorCount
 
         return """
             <?xml version="1.0" encoding="utf-8"?>


Source/SwiftLintFramework/Reporters/JUnitReporter.swift
```diff
@@ -11,11 +11,12 @@ struct JUnitReporter: Reporter {
     static func generateReport(_ violations: [StyleViolation]) -> String {
         let warningCount = violations.filter({ $0.severity == .warning }).count
         let errorCount = violations.filter({ $0.severity == .error }).count
+        let testCount = violations.count
```
SimplyDanny:
```suggestion
        let testCount = warningCount + errorCount
```
So separate issues always add up to the total number.

SimplyDanny (approved):
Thanks @nandhinisubbu! Just one small improvement ...

Source/SwiftLintFramework/Reporters/JUnitReporter.swift
nandhinisubbu: @SimplyDanny, Does it mean, only the count of tests with severity as .warning and .error to be considered for tests count?

SimplyDanny: The JUnit reporter is a bit strange, because JUnit is actually a testing framework. We run tests in it and they either succeed or fail (or get skipped).

Now to support this format, we need to interpret SwiftLint results as kind of tests. But we only have tests that fail. So the total number of tests is comprised by the number of issues we report, and those are either warnings or errors.

That said, `violations.count` is totally fine as well. Adding warnings and errors makes clearer, though, that the reporter numbers are consistent.

nandhinisubbu: Thanks for the clarification @SimplyDanny!, have pushed the changes.

Issue #6298, opened 2025-10-13
The performance of static Linux binary is bad


ainame:
### New Issue Checklist

- [x] I've Updated SwiftLint to the latest version.
- [x] I've searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues).

### Bug Description

0.61.0 introduced experimental fully static Linux binary with Static Linux SDK. I fixed its crash issue #6287 and 0.62.1 was released. But it looks like the performance is significantly slow https://github.com/realm/SwiftLint/pull/6291#issuecomment-3398992436

This is not formal benchmark but static Linux binary is about 3 time slower than just normal release build on my end (Ubuntu Desktop 24.04 on VMware running on M4 Max with 8 cores/16GB memory allocated).

```
# v0.62.1 binary downloaded from GitHub Release
ainame@ainame-vm:~/repos/SwiftLint$ time ./swiftlint-static --no-cache 2&>1 /dev/null 

real	0m4.300s
user	0m13.376s
sys	0m10.517s

# Normal release build at main branch 32cf983
# swift build --product swiftlint -c release -Xswiftc -DSWIFTLINT_DISABLE_SOURCEKIT
ainame@ainame-vm:~/repos/SwiftLint$ time .build/aarch64-unknown-linux-gnu/release/swiftlint --no-cache  2&>1 > /dev/null

real	0m1.353s
user	0m6.682s
sys	0m0.127s
```

I suppose this is kind of expected that musl's allocator seems to be known as slow and found many Rust devs complaint about it. 

https://nickb.dev/blog/default-musl-allocator-considered-harmful-to-performance/
https://github.com/nextstrain/nextclade/issues/1338

In fact, when I just gave https://github.com/microsoft/mimalloc a try, it is as performant as normal release build. That said, it's better to get some benchmark/profile before introducing something new.

```
sudo apt-get install -y musl-tools cmake
git clone https://github.com/microsoft/mimalloc.git
cd mimalloc/
CC=musl-gcc CFLAGS="-O3" cmake -B build -DMI_BUILD_STATIC=ON -DMI_BUILD_SHARED=OFF
cmake --build build -j
cp build/libmimalloc.a ../SwiftLint/
```

```
ainame@ainame-vm:~/repos/SwiftLint$ swift build --product swiftlint \
    -c release \
    -Xswiftc -DSWIFTLINT_DISABLE_SOURCEKIT \
    --swift-sdk x86_64-swift-linux-musl  \
[...]

PR #6321, opened 2025-10-26
Introduce mimalloc to static Linux binaries

ainame:
# Background 

Closes https://github.com/realm/SwiftLint/issues/6298

Currently static Linux binary distributed isn't as performant as expected. This is because `malloc` in musl libc, which is used in Static Linux SDK, doesn't work well with multiple threaded execution. It uses `futex` syscalls to safely access memory, which in turns, decrease performance in multi-threaded env.

https://git.musl-libc.org/cgit/musl/tree/src/thread/__wait.c?h=v1.1.23&id=7e399fabd3db2c528b5982803eeba2841f547695
http://git.musl-libc.org/cgit/musl/tree/src/malloc/malloc.c?h=v1.1.23&id=7e399fabd3db2c528b5982803eeba2841f547695

# What

To address this issue, I introduced mimalloc for static Linux binary in this PR.
https://github.com/microsoft/mimalloc

mimalloc is performant memory allocator written in C that works as drop-in replacement for `malloc` and supports builds for musl. Compared with other popular allocators like jemalloc or tcmalloc, It think it's fair option. Also, in Swift compiler project (swiftlang/swift), mimaloc has been adopted for the allocator in Windows's toolchain (not runtime!). 
 
https://forums.swift.org/t/se-454-adopt-mimalloc-for-windows-toolchain/

# How 

Since mimalloc is only necessary when building static Linux binary, I added a new step that compiles `libmimalloc.a` to `release.yml` and updated `swift build` command.

```
swift build \
    -c release \
    --product swiftlint \
    --swift-sdk ${{ matrix.swift_sdk }} \
    -Xswiftc -DSWIFTLINT_DISABLE_SOURCEKIT \
    -Xlinker -z -Xlinker stack-size=0x80000 \
    -Xlinker --whole-archive \
    -Xlinker $(pwd)/libmimalloc.a \
    -Xlinker --no-whole-archive
```

The mimalloc README recommends placing mimalloc.o (instead of the .a) at the very beginning of the linker input so its malloc overrides the standard one. With swift build, we can’t reliably force an absolute “first” position.
https://github.com/microsoft/mimalloc?tab=readme-ov-file#static-override

As a practical alternative, I link against libmimalloc.a and wrap it with --whole-archive/--no-whole-archive so all relevant objects are included, provided the library is built with the malloc override enabled. To be fair, the .o approach works fine for now. I just wanted to make sure it will keep linking correctly in the future.

Since mimalloc is MIT license, I've also updated `Makefile` to include mimalloc's license file into the zip files for Linux binary.  `third_party_licenses/mimalloc-LICENSE` is now bundled in repo.

# Test

## GitHub Actions

I ran release.yml on my fork and worked successfully.
https://github.com/ainame/SwiftLint/actions/runs/18823390591 

## Artefacts 

It runs faster with mimalloc. 

[swiftlint-static-arm64](https://github.com/ainame/SwiftLint/actions/runs/18823390591/artifacts/4375891318) with mimalloc from https://github.com/ainame/SwiftLint/actions/runs/18823390591

```
ainame@ainame-vm:~/repos/SwiftLint$ time /home/ainame/Downloads/swiftlint-static-arm64/swiftlint . --no-cache
Linting Swift files at paths .
Linting 'Path+Helpers.swift' (1/670)
...
real	0m1.342s
user	0m6.289s
sys	0m0.088s
```

[swiftlint-static-arm64](https://github.com/ainame/SwiftLint/actions/runs/18819592629/artifacts/4374908255) built from main branch synced with upstream from https://github.com/ainame/SwiftLint/actions/runs/18819592629/artifacts/4374908255

```
ainame@ainame-vm:~/repos/SwiftLint$ time '/home/ainame/Downloads/swiftlint-static-arm64(1)/swiftlint' . --no-cache
Linting Swift files at paths .
Linting 'Path+Helpers.swift' (1/670)
Linting 'SwiftLintCommandPlugin.swift' (2/670)
....
real	0m4.445s
user	0m14.040s
sys	0m9.510s
```

## Symbols

`nm` suggests that malloc/free/calloc/realloc are resolved from mimalloc. They are defined in the final binary (symbol type T).

```
ainame@ainame-vm:~/repos/SwiftLint$ nm -C --defined-only .build/aarch64-swift-linux-musl/release/swiftlint | grep --color=auto -E "(T|W) (malloc|free|calloc|realloc)$"
0000000002878df0 T calloc
0000000002877de0 T free
0000000002878b70 T malloc
0000000002879870 T realloc
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -175,7 +175,9 @@ jobs:
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
           name: ${{ matrix.artifact_name }}
-          path: ${{ env.BINARY_PATH }}
+          path: |
+            ${{ env.BINARY_PATH }}
+            mimalloc-build/LICENSE
 
   build-macos:
     name: Build macOS Binaries
@@ -231,6 +233,9 @@ jobs:
           mv -f swiftlint-linux-arm64/swiftlint swiftlint_linux_arm64
           mv -f swiftlint-static-amd64/swiftlint swiftlint_static_amd64
           mv -f swiftlint-static-arm64/swiftlint swiftlint_static_arm64
+
+          # Just pick one of the licenses.
+          mv -f swiftlint-static-amd64/LICENSE LICENSE.mimalloc
       - name: Make binaries executable
         run: chmod +x swiftlint*
       - name: Create artifacts
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,6 @@
 TEMPORARY_FOLDER?=/tmp/SwiftLint.dst
 PREFIX?=/usr/local
 BUILD_TOOL?=xcodebuild
-MIMALLOC_VERSION?=3.0.10
 
 XCODEFLAGS=-scheme 'swiftlint' \
 	-destination 'platform=macOS' \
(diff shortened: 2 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -144,14 +144,32 @@ jobs:
           persist-credentials: false
       - name: Install SDK
         run: swift sdk install https://download.swift.org/swift-6.2-release/static-sdk/swift-6.2-RELEASE/swift-6.2-RELEASE_static-linux-0.0.1.artifactbundle.tar.gz --checksum d2225840e592389ca517bbf71652f7003dbf45ac35d1e57d98b9250368769378
+      - name: Build mimalloc object
+        run: |
+          set -euxo pipefail
+          MIMALLOC_VERSION=3.0.10
```
SimplyDanny: This should be a global environment variable of the workflow, so we can easily find and update it in the future.

.github/workflows/release.yml
```diff
@@ -144,14 +144,32 @@ jobs:
...
+          curl -sSfL "https://github.com/microsoft/mimalloc/archive/refs/tags/v${MIMALLOC_VERSION}.tar.gz" | tar xz
+          export CC=clang
+          cmake -S "mimalloc-${MIMALLOC_VERSION}" -B mimalloc-build \
+            -DMI_BUILD_SHARED=OFF \
+            -DMI_BUILD_STATIC=ON \
+            -DMI_BUILD_OBJECT=OFF \
+            -DCMAKE_BUILD_TYPE=Release
+          cmake --build mimalloc-build --target mimalloc-static
+          cp mimalloc-build/libmimalloc.a libmimalloc.a
```
SimplyDanny: Can we skip this copy step and directly refer to the library in line 171?

Makefile
```diff
@@ -1,6 +1,7 @@
 TEMPORARY_FOLDER?=/tmp/SwiftLint.dst
 PREFIX?=/usr/local
 BUILD_TOOL?=xcodebuild
+MIMALLOC_LICENSE=third_party_licenses/mimalloc-LICENSE
```
SimplyDanny: We could copy the license file right from the tarball, so we don't need to care about updating it. Just upload it together with the binary in the workflow.

ainame:
Testing the changes here https://github.com/ainame/SwiftLint/actions/runs/18858095838/job/53810600398


-----------------

It ran successfully after those changes!
https://github.com/ainame/SwiftLint/releases/tag/0.0.5

Confirmed mimalloc's license files still bundles.
Screenshot 2025-10-27 at 22 59 20

ainame:
@SimplyDanny Thank you for your review! I've addressed your suggestions and updated CHANGELOG.md!
Feel free to make additional changes or merge this PR. Looking forward to using static Linux binary🙂

PR #6339, opened 2025-11-10
OverrideInExtension: should not trigger inside '@objc @implementation'

jaredgrubb:
I hit this rule violation and I think the rule should be updated to handle the case. Having `override` inside a `@objc @implementation` is common (and required in some cases).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#6332](https://github.com/realm/SwiftLint/issues/6332)
 
-* `override_in_extension` no longer triggers inside `@objc @implementation` extensions.
+* `override_in_extension` no longer triggers inside `@objc @implementation` extensions.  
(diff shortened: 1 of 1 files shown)


jaredgrubb:
I only look for `@implementation` attribute for this patch; I _think_ the same principle would apply for other kinds of this (eg, C or C++ interop), but I have never done anything like that so I can't say for sure.

I could tighten this up to require **both** `objc` and `implementation` to trigger if you prefer.

CHANGELOG.md
```diff
@@ -35,6 +35,9 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#6332](https://github.com/realm/SwiftLint/issues/6332)
 
+* `override_in_extension` no longer triggers inside `@objc @implementation` extensions.
```
```suggestion
* `override_in_extension` no longer triggers inside `@objc @implementation` extensions.  
```

SimplyDanny (approved):
Thanks @jaredgrubb! It makes sense to only check for `@implementation` due to being orthogonal to the language attribute.

PR #6355, opened 2025-11-21
correct incorrect handling of file paths

compnerd:
`URL.path` is not a usable rendering of the path. `URL.path` represents the URI path. If the path is meant to be consumable as a file path, the file system representation (aka FSR) must be retrieved.

Unfortunately, the FSR access is a bit more complicated as the API for that returns a +0 borrowed object with the lifetime bound to `URL`. Because this relies on the toll-free bridged storage of `URL`, Swift does not have this. Instead, we are provided a
`withUnsafeFileSystemRepresentation` which provides a heap allocated buffer that is scoped, which we can copy into a `String`. To abstract away this wart, we introduce an extension on `URL` to access the FSR through the new `filepath` property.

This improves the file path handling and makes additional tests now pass.

### DIFF
--- a/Source/SwiftLintFramework/Extensions/FileManager+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/FileManager+SwiftLint.swift
@@ -39,8 +39,7 @@ extension FileManager: LintableFileManager {
 
         return subpaths(atPath: root.path)?.parallelCompactMap { element -> String? in
             let elementURL = URL(fileURLWithPath: element, relativeTo: root)
-            guard elementURL.pathExtension == "swift" else { return nil }
-            if FileManager.default.isFile(atPath: elementURL.path) {
+            if elementURL.isSwiftFile {
                 return elementURL.standardized.filepath
             }
             return nil


compnerd:
CC: @SimplyDanny @roman-bcny

compnerd:
@SimplyDanny I think that I will need a little bit of help here. The current failure seems to be unrelated to my change and might be a general configuration issue. It does seem that this change is passing the CI and does improve the test coverage on Windows. I'm not sure that brace expansion is required - the musl path does not support that either. Please let me know if there is anything else that this needs.

SimplyDanny:
I think this was just a hiccup in the pipeline. Works now after retriggering it.

I'll have a look at the changes later.

Source/SwiftLintCore/Extensions/URL+SwiftLint.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+public extension URL {
+  var filepath: String {
+    self.withUnsafeFileSystemRepresentation { String(cString: $0!) }
+  }
```
SimplyDanny: We normally use 4 spaces for indentation.

Source/SwiftLintFramework/Extensions/FileManager+SwiftLint.swift
```diff
@@ -31,19 +31,20 @@ public protocol LintableFileManager {
...
-        let absolutePath = path.bridge()
-            .absolutePathRepresentation(rootDirectory: rootDirectory ?? currentDirectoryPath).bridge()
-            .standardizingPath
-
+        let root =
+            URL(fileURLWithPath: path.absolutePathRepresentation(rootDirectory: rootDirectory ?? currentDirectoryPath))
         // if path is a file, it won't be returned in `enumerator(atPath:)`
-        if absolutePath.bridge().isSwiftFile(), absolutePath.isFile {
-            return [absolutePath]
+        if FileManager.default.isFile(atPath: root.path), root.pathExtension == "swift" {
```
SimplyDanny: For `FileManager.default.isFile(atPath:)`, we don't need to use `root.filepath`?

Source/SwiftLintCore/Extensions/URL+SwiftLint.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+public extension URL {
```
SimplyDanny, line 3: We could add `isSwiftFile` that encapsulates `FileManager.default.isFile(atPath: path)` and `pathExtension == "swift"` to this extension.

SimplyDanny (commented):
Apart from a few nits, this looks reasonable to me, even though is a bit strange that one needs to rely on an "unsafe" and unwieldy API with `withUnsafeFileSystemRepresentation`.

Source/SwiftLintCore/Extensions/URL+SwiftLint.swift
compnerd: I like this idea!

Source/SwiftLintCore/Extensions/URL+SwiftLint.swift
```diff
@@ -0,0 +1,14 @@
...
+    var filepath: String {
+        self.withUnsafeFileSystemRepresentation { String(cString: $0!) }
+    }
+
+    var isSwiftFile: Bool {
+        guard FileManager.default.isFile(atPath: filepath), pathExtension == "swift" else {
+            return false
+        }
+        return true
```
SimplyDanny: `isFile` is only defined in `LintableFileManager`, but does just `path.isFile`:
```suggestion
        filepath.isFile && pathExtension == "swift"
```

Source/SwiftLintFramework/Extensions/FileManager+SwiftLint.swift
```diff
@@ -31,19 +31,19 @@ public protocol LintableFileManager {
...
-            let absoluteElementPath = absolutePath.bridge().appendingPathComponent(element)
-            return absoluteElementPath.isFile ? absoluteElementPath : nil
+        if root.isSwiftFile { return [root.standardized.filepath] }
+
+        return subpaths(atPath: root.path)?.parallelCompactMap { element -> String? in
+            let elementURL = URL(fileURLWithPath: element, relativeTo: root)
+            guard elementURL.pathExtension == "swift" else { return nil }
+            if FileManager.default.isFile(atPath: elementURL.path) {
+                return elementURL.standardized.filepath
+            }
```
SimplyDanny:
```suggestion
            if elementURL.isSwiftFile {
                return elementURL.standardized.filepath
            }
```

compnerd:
> Apart from a few nits, this looks reasonable to me, even though is a bit strange that one needs to rely on an "unsafe" and unwieldy API with `withUnsafeFileSystemRepresentation`.

Yes, that is because the model is migrating from ObjC to Swift. The original API vended an inner pointer, which requires that a closure based API is used (i.e. `with.*`), and the use of a raw pointer means that this needs to be `unsafe` API, yielding the `withUnsafe.*` function.

rgoldberg:
@compnerd Sorry to bother you, but seeing the code that this PR updated inspired me to learn how to handle paths correctly.

What are the reasons why you didn't use a [FilePath](https://developer.apple.com/documentation/System/FilePath)? e.g.:

```swift
public extension URL {
    var filepath: String {
        FilePath(self)!.string
    }
}
```

Does that return inapplicable path strings? Is it too slow? Is there a problem using `FilePath` on various platforms? Etc.

`FilePath` should be available from the toolchain for macOS 11+. Is it available from the toolchain on Windows & Linux?

It looks like it is also available from https://github.com/apple/swift-system, whose current version (1.6.4) requires Swift 5.9+ (its main branch requires Swift 6.0+, but no releases have been made with that higher requirement).

If the `FilePath` code above would work, is avoiding the unsafe pointers not worth needing to import another module?

If the `FilePath` code above would work at least as well as the existing implementation, might it return even better paths in certain circumstances?

Does any existing library have an equivalent for `URL.filepath`, so I can use a standardized version?

Thanks for sharing any info about this.

compnerd:
> @compnerd Sorry to bother you, but seeing the code that this PR updated inspired me to learn how to handle paths correctly.
> 
> What are the reasons why you didn't use a [FilePath](https://developer.apple.com/documentation/System/FilePath)? e.g.:
> 
> ```swift
> public extension URL {
>     var filepath: String {
>         FilePath(self)!.string
>     }
> }
> ```
> 
> Does that return inapplicable path strings? Is it too slow? Is there a problem using `FilePath` on various platforms? Etc.
> 
> `FilePath` should be available from the toolchain for macOS 11+. Is it available from the toolchain on Windows & Linux?
> 
> It looks like it is also available from https://github.com/apple/swift-system, whose current version (1.6.4) requires Swift 5.9+ (its main branch requires Swift 6.0+, but no releases have been made with that higher requirement).
> 
> If the `FilePath` code above would work, is avoiding the unsafe pointers not worth needing to import another module?
> 
> If the `FilePath` code above would work at least as well as the existing implementation, might it return even better paths in certain circumstances?
> 
> Does any existing library have an equivalent for `URL.filepath`, so I can use a standardized version?
> 
> Thanks for sharing any info about this.

`FilePath` is not available from the toolchain, it is a separate package (swift-system). Furthermore, FilePath (v1) had some issues that never fully got resolved. Hopefully FilePath v2 will address those (and is currently in the pitch phase). Unfortunately, much of the features in v2 were pared back, so some future version should be sufficiently complete. (Of course, I am partial as I am/was a co-author on FilePath).

rgoldberg:
@compnerd Thanks for all the info.

If my code runs only on macOS, not on any other platforms, is `URL.path(percentEncoded: false)` sufficient to generate correct path strings?

If `URL.path(percentEncoded: false)` isn't correct when limited to macOS, would `System.FilePath` work correctly just for macOS? Or should I also use the unsafe pointer implementation?

Thanks again.

Good luck with FilePath 2!

compnerd:
I would recommend that you always use the unsafe pointer approach as that is the official contract of `URL`. `withUnsafeFileSystemRepresentation` is the only way to get the FSR (file system representation). The URL representation can happen to work, but it is not meant for use with the FS operations, it is meant to be a generic URI.

SimplyDanny:
Another reason, we chose `URL` as the main representation for all files (currently only on the `windows-support` branch) is that it covers two use cases. One is the proper file system representation with `withUnsafeFileSystemRepresentation`, the other is that `URL.path` returns a Unix-like representation of the path even on Windows which we require to match it against include and exclude patterns as defined in `.swiftlint.yml` configuration files. This is important to keep configurations in an OS-independent way.

PR #6358, opened 2025-11-25
Replace colon with underscore in caching filePath (and a few other characters)

roman-bcny:
Without this we end up with a file name `http__:site` which doesn't work on Windows and crashes `ConfigurationTests.testValidRemoteChildConfig` with abort() which fast-exits resulting in partial test results
```
Subcode: 0x7 FAST_FAIL_FATAL_APP_EXIT 
ucrtbase!abort+0x4e
```

With this change I'm able to run swift test on Windows to completion (x64 6.2.1):
```
Test Suite 'All tests' failed at 2025-11-24 18:23:31.026
         Executed 1025 tests, with 91 failures (1 unexpected) in 365.412 (365.412) seconds
```

cc @compnerd

### DIFF
--- a/Source/SwiftLintFramework/Configuration/Configuration+Remote.swift
+++ b/Source/SwiftLintFramework/Configuration/Configuration+Remote.swift
@@ -200,9 +200,12 @@ internal extension Configuration.FileGraph.FilePath {
     }
 
     private func filePath(for urlString: String, rootDirectory: String) -> String {
-        let adjustedUrlString = urlString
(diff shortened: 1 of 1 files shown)


SimplyDanny (commented):
There're even more characters allowed in a URL that must not be used in a "universal" file path. I think those are

```
< > : " / \ | ? *
```

and trailing dots `.`. Since we're already on it, we may replace them too.

roman-bcny:
@SimplyDanny makes sense, updated

SimplyDanny (approved):

PR #6365, opened 2025-12-02
various fixes for Windows path handling

compnerd:
An assortment of fixes that improve the test coverage on Windows. With this set, local testing reveals 1 failure (RemoteCycleDetection) and another set of test failures due to Windows globbing not matching the POSIX semantics.

### DIFF
--- a/Source/SwiftLintCore/Models/Location.swift
+++ b/Source/SwiftLintCore/Models/Location.swift
@@ -23,17 +23,7 @@ public struct Location: CustomStringConvertible, Comparable, Codable, Sendable {
 
     /// The file path for this location relative to the current working directory.
     public var relativeFile: String? {
-        guard let file else { return nil }
-        let fileComponents = URL(fileURLWithPath: file).pathComponents
-        let baseComponents = URL(fileURLWithPath: FileManager.default.currentDirectoryPath, isDirectory: true)
-            .pathComponents
-
-        guard fileComponents.starts(with: baseComponents) else {
-            return file
-        }
-
-        // TODO: update for Windows paths, together with other places where slash is used as a separator
-        return fileComponents.dropFirst(baseComponents.count).joined(separator: "/")
+        file?.replacingOccurrences(of: FileManager.default.currentDirectoryPath + "/", with: "")
     }
 
     /// Creates a `Location` by specifying its properties directly.


Source/SwiftLintCore/Models/Location.swift
```diff
@@ -24,7 +24,7 @@ public struct Location: CustomStringConvertible, Comparable, Codable, Sendable {
     /// The file path for this location relative to the current working directory.
     public var relativeFile: String? {
         let url = URL(fileURLWithPath: FileManager.default.currentDirectoryPath, isDirectory: true)
-        return file?.replacingOccurrences(of: url.filepath, with: "")
+        return file?.replacingOccurrences(of: url.filepath + "/", with: "")
```
compnerd: This feels wrong - we should be computing that by stripping the parent components.

roman-bcny: You mean removing the prefix instead of replacing occurrences? Yes, this is a pre-existing issue unrelated to this PR, where relative path of /swiftlint/src/swiftlint/ with cwd=/swiftlint would break. Diff with origin:
```
     /// The file path for this location relative to the current working directory.
     public var relativeFile: String? {
-        return file?.replacingOccurrences(of: FileManager.default.currentDirectoryPath + "/", with: "")
+        let url = URL(fileURLWithPath: FileManager.default.currentDirectoryPath, isDirectory: true)
+        return file?.replacingOccurrences(of: url.filepath + "/", with: "")
     }
```

compnerd: Yeah, we should change this to use the parent path stripping instead IMO

roman-bcny: Updated!

Source/SwiftLintCore/Models/Location.swift
```diff
@@ -23,8 +23,17 @@ public struct Location: CustomStringConvertible, Comparable, Codable, Sendable {
...
+        let filePath = URL(fileURLWithPath: file).filepath
+        var basePath = URL(fileURLWithPath: FileManager.default.currentDirectoryPath, isDirectory: true).filepath
+        if !basePath.hasSuffix("/") {
+            basePath += "/"
+        }
+
+        if filePath.hasPrefix(basePath) {
+            return String(filePath.dropFirst(basePath.count))
+        }
+        return file
```
compnerd: Sorry, I should've been more clear. We should de-compose the path with `pathComponents` based handling. we would need to ensure that the CWD is entirely present though.

roman-bcny: ah, like this then

compnerd: Almost - but rather than the `.joined(separator: "/")`, use a reduction of `appendingPathComponent` (or if we can use newer APIs, `appending(components:)`).

roman-bcny: Updated to reduce+append in https://github.com/realm/SwiftLint/pull/6365/commits/30b80c48dcf57c3934839be4c023aca145896ab8, apparently `appending` doesn't work as "Swift doesn't allow spreading arrays into variadic parameters" but please lmk if there is a better way

roman-bcny: eh, this doesn't work either because we get `./` at the beginning of the path.. Maybe just keep `.joined(separator: "/")`? It seems like the most straightforward opion.

Source/SwiftLintCore/Models/Location.swift
```diff
@@ -32,9 +32,7 @@ public struct Location: CustomStringConvertible, Comparable, Codable, Sendable {
             return file
         }
 
-        return fileComponents.dropFirst(baseComponents.count).reduce(into: URL(filePath: "")) {
-            $0.append(component: $1)
-        }.relativePath
+        return fileComponents.dropFirst(baseComponents.count).joined(separator: "/")
```
compnerd: This opens us up to issues with different path separators. This needs to be a proper file system representation somehow.

roman-bcny: fwiw there are already a couple other places where `.joined(separator: "/")` is used, and this is normally not really an issue on Windows as it can handle both? I find it really odd that there is no equivalent of python's `os.path.join` on Swift as that would be the obvious choice in cases like this.

compnerd: Those need to be adjusted as well. It _is_ an issue, Windows _cannot_ handle both. Win32 handles both, NT does not. If you have an extended path, it must be in the correct form.

The equivalent is to always just use `URL`.

roman-bcny: I'll look some more but I don't think `URL` is equivalent to `os.path.join` - the latter just joins paths without trying to be smart about it, URL has all these surprises with symlink resolution etc, which we run into in those bazel builds I believe.

btw I didn't realize how easy it was to repro the bazel stuff locally, turns out it's just
% brew install bazelisk
% bazelisk test //Tests:FileSystemAccessTests --test_filter=GlobTests.testMatchesCharacterInRange --test_output=streamed

and I now see the /var/tmp vs /private/var/tmp discrepancy locally. That makes it a bit easier to figure out those surprises :)

roman-bcny: Looking at what was done to ConfigurationTests in https://github.com/realm/SwiftLint/pull/6342/files, these two things stood out for me (the tests themselves were changed to this):
```
XCTAssert(FileManager.default.changeCurrentDirectoryPath(Mock.Dir.exclusionTests))
```
and
```
assertEqual(["directory/File1.swift", "directory/File2.swift", "directory/ExcludedFile.swift"], paths)
```

I think this might be 1) getting around some of the path shenanigans and 2) expecting the forward slash to be returned by e.g. `lintablePaths`.

> Win32 handles both, NT does not. If you have an extended path, it must be in the correct form

I suppose you're talking about the API level, I was thinking more of the general tooling which often hides this, I'm guessing some kind of translation underneath, so you can `cd "c:/"` in cmd or just use forward slashes in python without running into issues (os.path.join still being preferable of course).

I'm not sure what we're trying to do here, as the change above is using forward slashes so maybe we want that to keep behaving the same way and hide the / -> \ as an implementation detail in some lower level?..

cc @SimplyDanny

SimplyDanny: At some point, the string paths returned by `lintablePaths` need to be converted into `URL`s (before we eventually access real files with them). Not sure if and where this is currently happening. For the test, it's fine to work with the string paths as its purpose is to find the correct files in the first place.

roman-bcny: I narrowed down the bazel issue to this part of the change:

```
--- a/Tests/TestHelpers/TestResources.swift
+++ b/Tests/TestHelpers/TestResources.swift
@@ -9,7 +9,6 @@ public enum TestResources {
         }
         return folder
             .appendingPathComponent("Resources")
-            .path
-            .absolutePathStandardized()
+            .filepath
     }
 }
```

Which is due to:
```
folder.filepath: /private/var/tmp/...
folder.path: /private/var/tmp/...
folder.path.absolutePathStandardized(): /var/tmp/...
```

What should be the intent here?

SimplyDanny: I changed `TestResources` already a while back (the PR was laying around for fairly long) and so forgot to update it for the Windows use case. 

Since it's referring to real files, it should return `URL` and later keep using `URL` up to the point when the path doesn't get modified further and is eventually used to access a file (then with the rather new `.filepath`).

roman-bcny: Okay I unbroke tests by switching TestResources path to URL and sprinkling symlink resolution in a few places.

I kept the "/" join as I'm not sure if we want that API to change, or some underlying implementation to change.

@compnerd maybe we could merge this like this (unless there are major non-Windows-specific issues with what I've changed) and iterate on follow up PRs?

Tests/TestHelpers/TestResources.swift
```diff
@@ -2,14 +2,14 @@ import Foundation
 import SwiftLintCore
 
 public enum TestResources {
-    public static func path(_ calleePath: String = #filePath) -> String {
+    public static func path(_ calleePath: String = #filePath) -> URL {
```
SimplyDanny, line 5: Can symlinks be resolved once here? I assume symlinks appear when running with Bazel only?

roman-bcny: Updated. Yes, I've only seen symlinks in Bazel and specifically /var/tmp/ vs /private/var/tmp/. I haven't tested anything else.

Source/SwiftLintCore/Models/Location.swift
roman-bcny: Discussed offline, restored to the original implementation so we can properly fix in a follow up

Tests/BuiltInRulesTests/FileHeaderRuleTests.swift
```diff
@@ -2,7 +2,7 @@
 import TestHelpers
 import XCTest
 
-private let fixturesDirectory = "\(TestResources.path())/FileHeaderRuleFixtures"
+private let fixturesDirectory = "\(TestResources.path().filepath)/FileHeaderRuleFixtures"
```
compnerd, line 5:
```suggestion
private let fixturesDirectory = TestResources.path().appendingPathComponent("FileHeaderRuleFixtures").filepath
```

Tests/BuiltInRulesTests/FileNameNoSpaceRuleTests.swift
```diff
@@ -3,7 +3,7 @@ import SourceKittenFramework
 import TestHelpers
 import XCTest
 
-private let fixturesDirectory = "\(TestResources.path())/FileNameNoSpaceRuleFixtures"
+private let fixturesDirectory = "\(TestResources.path().filepath)/FileNameNoSpaceRuleFixtures"
```
compnerd, line 6:
```suggestion
private let fixturesDirectory = TestResources.path().appendingPathComponent("FileNameNoSpaceRuleFixtures").filepath
```

Tests/CoreTests/YamlSwiftLintTests.swift
```diff
@@ -40,6 +41,6 @@ final class YamlSwiftLintTests: SwiftLintTestCase {
     }
 
     private func getTestYaml() throws -> String {
-        try String(contentsOfFile: "\(TestResources.path())/test.yml", encoding: .utf8)
+        try String(contentsOfFile: "\(TestResources.path().filepath)/test.yml", encoding: .utf8)
```
compnerd, line 44:
```suggestion
        try String(contentsOfFile: TestResources.path().appendingPathComponent("test.yml").filepath, encoding: .utf8)
```

Tests/FileSystemAccessTests/SourceKitCrashTests.swift
```diff
@@ -35,7 +35,7 @@ final class SourceKitCrashTests: SwiftLintTestCase {
     }
 
     func testRulesWithFileThatCrashedSourceKitService() throws {
-        let file = try XCTUnwrap(SwiftLintFile(path: "\(TestResources.path())/ProjectMock/Level0.swift"))
+        let file = try XCTUnwrap(SwiftLintFile(path: "\(TestResources.path().filepath)/ProjectMock/Level0.swift"))
```
compnerd, line 38:
```suggestion
        let file = try XCTUnwrap(SwiftLintFile(path: TestResources.path()..appendingPathComponent("ProjectMock").appendingPathComponent("Level0.swift").filepath)
```

Tests/FrameworkTests/CustomRulesTests.swift
```diff
@@ -9,7 +9,7 @@ import XCTest
 final class CustomRulesTests: SwiftLintTestCase {
     private typealias Configuration = RegexConfiguration<CustomRules>
 
-    private var testFile: SwiftLintFile { SwiftLintFile(path: "\(TestResources.path())/test.txt")! }
+    private var testFile: SwiftLintFile { SwiftLintFile(path: "\(TestResources.path().filepath)/test.txt")! }
```
compnerd, line 12:
```suggestion
    private var testFile: SwiftLintFile { SwiftLintFile(path: TestResources..path().appendingPathComponent("test.txt").filepath)! }
```

Tests/TestHelpers/TestResources.swift
```diff
@@ -2,14 +2,15 @@ import Foundation
...
         let folder = URL(fileURLWithPath: calleePath, isDirectory: false).deletingLastPathComponent()
         if let rootProjectDirectory = ProcessInfo.processInfo.environment["BUILD_WORKSPACE_DIRECTORY"] {
-            return "\(rootProjectDirectory)/Tests/\(folder.lastPathComponent)/Resources"
+            return URL(
+                fileURLWithPath: "\(rootProjectDirectory)/Tests/\(folder.lastPathComponent)/Resources",
```
compnerd, line 9:
```suggestion
                fileURLWithPath: rootProjectDirectory.appendingPathComponent("Tests").appendingPathComponent(folder.lastPathComponent).appendingPathComponent("Resources").filepath,
```

Tests/BuiltInRulesTests/FileNameRuleTests.swift
```diff
@@ -165,7 +166,7 @@ final class FileNameRuleTests: SwiftLintTestCase {
             try validate(
                 fileName: "main.swift",
                 excluded: [],
-                excludedPaths: [".*/FileNameRuleFixtures/.*"]
+                excludedPaths: [#".*[\\/]FileNameRuleFixtures[\\/].*"#]
```
SimplyDanny, line 169: These patterns should be treated independently of what the right format is on the current OS. I propose that to support `/` only in these patterns and internally SwiftLint needs to make sure that they are mapped correctly to the underlying file/path representation.

Does that make sense?

roman-bcny: Well that's exactly what I was talking about earlier... I wasn't sure where you expected to see plain strings with forward slashes vs URLs. I personally find reasoning about POSIX paths much more convenient than generic paths that might have either of the slashes, they're easier to reason about and Windows users aren't that surprised when they see a forward slash. So to me personally plaintext POSIX makes sense as the top-level abstraction, but that's a different approach from "everything is an URL" so we need to pick one or the other. What does the glob support library (you were looking into) use by the way?

SimplyDanny: The [Glob library](https://github.com/ileitch/swift-filename-matcher) only works for Unix paths. It matches on strings using forward slashes.

I think that the current approach (using URLs internally) is quite right. We only need to make sure that when it comes to user input, path patterns will be Unix-style (only) and we need to be able to handle them correctly: probably like URL -> Unix path -> filter with patterns -> URL -> Unix/Windows output paths.

I just need to think more about where the right conversion points are. 😅

roman-bcny: Hmm. I'm not actually sure how you're planning to use that library. It seems to only provide an interface that tells you whether a given filename matches a pattern, but there is no interface for actually efficient globbing. You could list all files and then use that matching but that means traversing the whole tree? The proper implementation would require at least some integration with directory listing of a particular OS.

Example: an efficient traversal of `a123/b456/*/1.json` would only list items under `a123/b456/` and not traverse any other part of the tree, so on Windows it needs to somehow get translated to FindFirstFile/NtQueryDirectoryFile specifically on `a123\b456`

SimplyDanny: With https://github.com/realm/SwiftLint/pull/6366, file traversal is done with a directory enumerator which takes the cases into account that you describe in your example, only that it doesn't work for Windows paths yet. There, I just need to follow my own advice:

> We only need to make sure that when it comes to user input, path patterns will be Unix-style (only) and we need to be able to handle them correctly: probably like URL -> Unix path -> filter with patterns -> URL -> Unix/Windows output paths.

Here, I commented on the patterns only considered by specific rules (`file_name` in this case). These rules also need to respect `/` over `\` on Windows.

roman-bcny: I had a quick look and didn't figure out how it takes care of my example to avoid listing unrelated dirs, but if it does then great!

I feel that the easiest way to avoid thinking about `/` vs `\` at all levels (including those where you're not using URLs but instead dealing with plain strings like you're doing in a few places in the PR you linked) is to make everything work on posix paths and provide a translation layer that converts both ways as needed, plugged at the lowest possible level. This would go in a different direction from the "use URL everywhere" idea - but I'm not sure how well that approach plays with globbing in the first place and it seems to be forcing to choose between plain strings vs URLs in many places.

roman-bcny: @SimplyDanny I'd like us to merge this PR in some form so that it's not staying up hanging. Maybe the PR is doing too much and we could split some of the parts out in that case. Could you have a look at what remains undecided/possibly blocking and what we're ready to commit? Otherwise, would you prefer a different approach like finishing some other changes before we move forward here?

compnerd: Running this locally, it seems that the current version still fails to run the test suite - not just with failing tests but with abnormal termination. I'm not sure I see the value of merging this as is as it isn't really focused on a single issue but tries to fix a number of things. If this were a very targeted fix, I would agree.

SimplyDanny: I started with the "URL everywhere and always" approach. It's quite a lot but looks promising so far. The glob stuff also works as it can operate in the file representation of URL which remains the Unix-style, so the logic doesn't need to change. It's only some back and forth with URL and string path representation. But what's passed around can be URL only.

On my tour through the whole code base, I came across a few more places that also need adaption. Using URL feels more natural now than operating on strings.

I'll need some more time to get this done but will let you know once it is. Performance check are critical then. It's not always obvious which operations on URLs consult the file system.

compnerd: Globing against the URI is reasonable, but, it should be explicitly documented.

Consider something like `file:///C:/**/*.swift` and `file://filer/share/**/*.swift`, `file://filer.local/share/**/*.swift` as patterns (for Windows and UNC paths).

This is also complicated as there are multiple spellings:
- `file:///c:/**/*.swift`
- `file:///C|/**/*.swift`
- `file:C|/**/*.swift`
are all the same pattern (note separators, case changes, and scheme and path joining).

UNC path compatibility (needed for proper windows support) also has the oddity of the empty authority and preserved leader, i.e. `file://///filer.local/share/**/*.swift`.

Finally, note also the need to accept URL encoded patterns as certain characters must be URL encoded as they are invalid, e.g. `$`, which is used on OpenVMS and special paths on windows.

roman-bcny: @SimplyDanny thanks for the clarification!

@compnerd this is quickly escalating towards being worthy a 3rd party library :) which could then be used elsewhere not just SwiftLint

compnerd: Cross-platform globbing is not really a thing for a reason - file paths are not uniform.

What about HFS paths? `Machintosh HD:Users:compnerd:Desktop` wouldn't work very well with the glob patterns for Unix.

Perhaps a VMS path would be a good thing to also account for: `DISK$USERS:[COMPNERD]FILE.TXT;4`. 

How would you account for that in the current globbing mechanism? We really should avoid the globbing or require the user to provide a generic regular expression to match paths.

SimplyDanny:
@roman-bcny, @compnerd: I finally managed to have everything up and running on Windows with #6408. It was quite a fight, but all tests pass and running `swiftlint lint` works now.

The only thing excluded is `?[]` symbols in glob patterns. They just won't be supported on Windows at first. Assuming good test coverage, everything else works as it does on macOS and Linux.

I'll have the PR lie around for a bit for review and to play around with it. Anyway, I need to think about what would be a good strategy to release it for people to test without causing potential issues in official releases. In parallel we might think about release artifacts for Windows.

roman-bcny:
Very exciting! I did a quick build and a test run of that PR on our codebase. No crashes. Tested a lot of files. Complained about a few things that I need to confirm, and whether it is missing something that it'd catch on Mac (which is trickier to verify as we lint during commit so we don't routinely have these examples) - but preliminary looking promising! Will need to allocate some time to check more thoroughly.

roman-bcny:
@SimplyDanny I don't see false positives for our rule set on Windows, and a 100-ish warnings/errors we have in a few places show exactly the same as on Mac so it seems pretty good. It's possible that something e.g. glob related is broken on Windows and we just don't hit it as almost everything is clean, but our globs seem pretty simple so probably not likely. Once we get a statically linked exe I think this is usable for our Windows work!

SimplyDanny:
> @SimplyDanny I don't see false positives for our rule set on Windows, and a 100-ish warnings/errors we have in a few places show exactly the same as on Mac so it seems pretty good. It's possible that something e.g. glob related is broken on Windows and we just don't hit it as almost everything is clean, but our globs seem pretty simple so probably not likely. Once we get a statically linked exe I think this is usable for our Windows work!

That's great news! I suggest to close this PR and continue on #6408. It contains the commits from here as one single squashed commit.

Issue #6359, opened 2025-11-27
`optional_data_string_conversion` does not trigger when `.init(decoding:as:)` is called instead of `String(decoding:as:)`


vapidinfinity:
### New Issue Checklist

- [x] I've Updated SwiftLint to the latest version.
- [x] I've searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues).

### Bug Description

As the title suggests, `optional_data_string_conversion` does not trigger when calling `String.init()` instead of directly calling `String()`. 

Example:
```swift
// This triggers a violation:
let text = String(decoding: data, as: UTF8.self)

// This does not trigger a violation:
let text: String = .init(decoding: data, as: UTF8.self)
```

(swiftlint is in its default configuration, ran by xcode)

### Environment

* SwiftLint version — 0.62.2
* Xcode version — 26.1 (17B55)
* Installation method used (**Homebrew**, CocoaPods, building from source, etc)
* Configuration file:

```yml
# n/a?
```

Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)? If so, paste their
relative paths and respective contents.
`N/A`

SimplyDanny:
This could be supported. However, I'd hide it behind an option since we cannot be 100% sure that `.init(decoding: data, as: UTF8.self)` refers to `String.init(decoding: data, as: UTF8.self)`.

theamodhshetty:
Implemented this in #6508.

I added include_shorthand_init (default false) so shorthand .init(decoding:as:) can be linted when desired, while avoiding ambiguous cases by default.

theamodhshetty:
Opened #6567. This extends `optional_data_string_conversion` to catch the equivalent `String.init(decoding:as:)` and contextual `.init(decoding:as:)` spellings, so they get the same treatment as the existing `String(decoding:as:)` case. I also added explicit examples for both forms and ran the generated rule test plus a strict lint pass on the touched files.

PR #6372, opened 2025-12-09
Add detection of leading-dot when the contextual type is String ` to `optional_data_string_conversion` rule

nadeemnali:
### Summary
 
 The `optional_data_string_conversion` rule now detects additional variant patterns of `String(decoding:as:)` 
initialization:
 
 - Direct calls: `String(decoding: data, as: UTF8.self)` (already supported)
 - Type-qualified initializer: `String.init(decoding: data, as: UTF8.self)` (NEW)
 - Leading-dot initializer: `let text: String = .init(decoding: data, as: UTF8.self)` (NEW)
 - Optional implicit leading-dot: `let text = .init(decoding: data, as: UTF8.self)` (controlled by 
`allow_implicit_init` config)
 
 ### Changes
 
 1. **Refactored visitor logic** (`OptionalDataStringConversionRule.swift`):
    - Changed from `visitPost(_ node: DeclReferenceExprSyntax)` to `visitPost(_ node: FunctionCallExprSyntax)` for 
better detection of all call patterns
    - Added support for detecting `String.init(...)` (MemberAccessExpr with base `String`)
    - Added support for detecting leading-dot `.init(...)` when assigned to a variable with explicit `String` type 
annotation
 
 2. **Added configuration system** (new file: `OptionalDataStringConversionConfiguration.swift`):
    - New configuration option: `allow_implicit_init` (default: `false`)
    - When `false` (default): only flag `.init(...)` with explicit `String` type annotation
    - When `true`: also flag implicit leading-dot `.init(...)` without type annotation
 
 3. **Enhanced rule documentation**:
    - Expanded non-triggering examples to cover various String initializer patterns
    - Added triggering examples for `String.init(...)` and `.init(...)` with explicit type
    - Added example showing `allow_implicit_init` configuration option
 
 4. **Updated configuration registry** (`default_rule_configurations.yml`):
    - Added `allow_implicit_init: false` to default configuration
 
 ### Why
 
 Fixes [#6359](https://github.com/realm/SwiftLint/issues/6359) where cases like `let text: String = .init(decoding:
 data, as: UTF8.self)` were not detected.
 
 ### Testing
 
 Run the rule's example code with:
 ```bash
 swift test --filter OptionalDataStringConversionRule

Or lint against the rule directly:

 ./.build/debug/swiftlint lint Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,12 @@
   [theamodhshetty](https://github.com/theamodhshetty)
   [#5741](https://github.com/realm/SwiftLint/issues/5741)
 
+* Add detection of cases such as `String.init(decoding: data, as: UTF8.self)` and
+  `let text: String = .init(decoding: data, as: UTF8.self)` to
+  `optional_data_string_conversion` rule.  
+  [nadeemnali](https://github.com/nadeemnali)
+  [#6359](https://github.com/realm/SwiftLint/issues/6359)
+
 ### Bug Fixes
 
 * Add an `ignore_attributes` option to `implicit_optional_initialization` so
@@ -72,16 +78,11 @@
   `redundant_self` rule.  
   [SimplyDanny](https://github.com/SimplyDanny)
   [#6553](https://github.com/realm/SwiftLint/issues/6553)
-  
+
 * Respect existing environment variables when setting `BUILD_WORKSPACE_DIRECTORY`
   in build tool plugins.  
   [SimplyDanny](https://github.com/SimplyDanny)
   [#6080](https://github.com/realm/SwiftLint/issues/6080)
-* Add detection of cases such as `String.init(decoding: data, as: UTF8.self)` and 
-* Add detection of cases such as `String.init(decoding: data, as: UTF8.self)` and
-  `let text: String = .init(decoding: data, as: UTF8.self)` to `optional_data_string_conversion` rule.  
-  [nadeemnali](https://github.com/nadeemnali)
-  [#6359](https://github.com/realm/SwiftLint/issues/6359)
 
 ## 0.63.2: High-Speed Extraction
 


Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift
```diff
@@ -34,5 +37,53 @@ private extension OptionalDataStringConversionRule {
                 violations.append(node.positionAfterSkippingLeadingTrivia)
             }
         }
```
SimplyDanny: We should rather implement `visitPost(_ node: FunctionCallExprSyntax)` instead. It can check all cases at the same time. If the called name is either `String`, `String.init` or `.init` and the arguments match, it might trigger.

The `var x: String = .init(...)` case is very special and I don't expect it to appear often. While `String.init` is safe, `.init` could appear everywhere (think of `f(.init(...))`. We might add an option for that (as I've already pointed out in #6359.

Tests/BuiltInRulesTests/OptionalDataStringConversionRuleTests.swift
SimplyDanny, line 1: This file isn't needed. All examples are automatically verified.

Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift
```diff
@@ -17,7 +18,9 @@ struct OptionalDataStringConversionRule: Rule {
             Example("String(decoding: data, encoding: UTF8.self)"),
```
SimplyDanny, line 18: Please add more cases that shouldn't trigger, e.g. `let text: Int = .init(decoding: data, as: UTF8.self)`. 🤷‍♂️

CHANGELOG.md
```diff
@@ -8605,6 +8605,11 @@ This release has seen a phenomenal uptake in community contributions!
   [JP Simard](https://github.com/jpsim)
   [#43](https://github.com/realm/SwiftLint/issues/43)
 
+### Bug Fixes
+
+* optional_data_string_conversion: Detect `String.init(...)` and leading-dot `.init(...)` usages when converting `Data` to `String`. This fixes cases such as `String.init(decoding: data, as: UTF8.self)` and `let text: String = .init(decoding: data, as: UTF8.self)` which previously did not trigger the rule. (Fixes https://github.com/realm/SwiftLint/issues/6359)
```
SimplyDanny: Please add this entry to the current "Main" section at the top of the file.

CHANGELOG.md
```diff
@@ -8605,6 +8605,11 @@ This release has seen a phenomenal uptake in community contributions!
...
+  [Nadeem Ali](https://github.com/nadeemnali)
```
```suggestion
* Add detection of cases such as `String.init(decoding: data, as: UTF8.self)` and 
  `let text: String = .init(decoding: data, as: UTF8.self)` to `optional_data_string_conversion` rule.  
  [Nadeem Ali](https://github.com/nadeemnali)
  [#6359](https://github.com/realm/SwiftLint/issues/6359)
```

Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift
SimplyDanny: I'm missing this example. Please add it.

Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift
```diff
@@ -15,23 +16,91 @@ struct OptionalDataStringConversionRule: Rule {
...
-               let expr = parent.arguments.last?.expression.as(MemberAccessExprSyntax.self),
-               expr.base?.description == "UTF8",
-               expr.declName.baseName.description == "self" {
-                violations.append(node.positionAfterSkippingLeadingTrivia)
+        override func visitPost(_ node: FunctionCallExprSyntax) {
+            // Only consider calls with labels `decoding` and `as`
+            guard node.arguments.map(\.label?.text) == ["decoding", "as"] else { return }
+
+            // Check that the `as:` argument is `UTF8.self`
+            func isUTF8Self(on call: FunctionCallExprSyntax) -> Bool {
```
SimplyDanny: Why do we need this function? It's only called once, isn't it?

Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift
```diff
@@ -15,23 +16,91 @@ struct OptionalDataStringConversionRule: Rule {
...
+            let called = node.calledExpression
+
+            // Case 1: direct `String(...)`
+            if let declRef = called.as(DeclReferenceExprSyntax.self), declRef.baseName.text == "String" {
+                violations.append(called.positionAfterSkippingLeadingTrivia)
+                return
+            }
+
+            // Case 2 and 3: `.init` or `String.init`
+            if let member = called.as(MemberAccessExprSyntax.self), member.declName.baseName.text == "init" {
```
SimplyDanny: Can be a `guard`.

Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift
```diff
@@ -15,23 +16,91 @@ struct OptionalDataStringConversionRule: Rule {
...
+                // Case 3: leading-dot `.init(...)`
+                // This is ambiguous in general. We conservatively only trigger if the call
+                // is used to initialize a variable that has an explicit `String` type annotation:
+                // let x: String = .init(...)
+                // We intentionally do not (yet) match arbitrary contexts like f(.init(...)) or returns,
+                // because `.init` can refer to other types and those cases are frequent and risky.
+                if member.base == nil {
+                    // Walk ancestors to find a VariableDecl or PatternBinding with a type annotation of `String`.
+                    var parent: Syntax? = node.parent
+                    while let ancestor = parent {
```
SimplyDanny: Better to not walk up the tree arbitrarily high. `node.parent.parent` gives you the binding to check for the `String` type.

SimplyDanny (changes requested):

SimplyDanny:
Hey @nadeemnali, is anything blocking you from moving this forward?

nadeemnali:
> Hey @nadeemnali, is anything blocking you from moving this forward?

Hey @SimplyDanny  Apologies, I was busy for past weeks, will update the comments ASAP

nadeemnali:
@SimplyDanny Can you please help me with the danger issue in pipeline, can't seem to find any logs for the failure

SimplyDanny:
> @SimplyDanny Can you please help me with the danger issue in pipeline, can't seem to find any logs for the failure

Danger doesn't like merge commits. Please rebase your branch onto main instead.

CHANGELOG.md
```diff
@@ -185,7 +190,7 @@
 * Reinstate handling of access level imports in `sorted_imports` rule.  
   [Brett-Best](https://github.com/Brett-Best)
   [#6374](https://github.com/realm/SwiftLint/issues/6374)
-
+
```
SimplyDanny: Remove.

Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift
```diff
@@ -15,23 +16,72 @@ struct OptionalDataStringConversionRule: Rule {
...
+            if let baseDecl = member.base?.as(DeclReferenceExprSyntax.self),
+               baseDecl.baseName.text == "String" {
+                violations.append(called.positionAfterSkippingLeadingTrivia)
+                return
+            }
+
+            // Case 3: leading-dot `.init(...)`
+            // This is ambiguous in general. We conservatively only trigger if the call
+            // is used to initialize a variable that has an explicit `String` type annotation:
+            // let x: String = .init(...)
```
SimplyDanny, line 91: We could optionally allow this case as well if there isn't an explicit type. We'd need to introduce a new configuration for the rule. Would you like to add that too?

SimplyDanny (commented):
Thanks for addressing my remarks! There is one more thing about allowing `.init(...)` even without an explicit type optionally.

Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift
nadeemnali: @SimplyDanny I have added the configuration for the rule

Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift
```diff
@@ -1,37 +1,96 @@
...
 struct OptionalDataStringConversionRule: Rule {
-    var configuration = SeverityConfiguration<Self>(.warning)
+    var configuration = OptionalDataStringConversionConfiguration()
 
     static let description = RuleDescription(
         identifier: "optional_data_string_conversion",
         name: "Optional Data -> String Conversion",
-        description: "Prefer failable `String(bytes:encoding:)` initializer when converting `Data` to `String`",
+        description: """
+            Prefer failable `String(bytes:encoding:)` initializer when converting `Data` to `String`. \
```
SimplyDanny: I'd rather keep it as is. Mentioning configuration parameters here increases the maintenance effort.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/OptionalDataStringConversionConfiguration.swift
```diff
@@ -0,0 +1,12 @@
+import SwiftLintCore
+
+@AutoConfigParser
+struct OptionalDataStringConversionConfiguration: SeverityBasedRuleConfiguration { // swiftlint:disable:this type_name
+    @ConfigurationElement(key: "severity")
+    private(set) var severityConfiguration = SeverityConfiguration<Parent>.warning
+
+    // When true, also flag leading-dot `.init(decoding:as:)` without explicit `String` type annotation.
+    // Default is false to preserve conservative behavior.
```
SimplyDanny: Skip that last sentence. Should be clear even without it.

Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift
```diff
@@ -1,37 +1,96 @@
...
+            Example("String(repeating: \"a\", count: 3)"),
+            Example("String(format: \"%d\", 3)"),
+            // Default behavior (allow_implicit_init == false): implicit leading-dot init without explicit type
+            Example("let text = .init(decoding: data, as: UTF8.self)"),
         ],
         triggeringExamples: [
-            Example("String(decoding: data, as: UTF8.self)")
+            Example("String(decoding: data, as: UTF8.self)"),
+            Example("String.init(decoding: data, as: UTF8.self)"),
+            Example("let text: String = .init(decoding: data, as: UTF8.self)"),
```
SimplyDanny: Tests (triggering and non-triggering) are missing for the new option. Just add them here with a configuration.

nadeemnali:
Hi @SimplyDanny I have updated all the comments, can you please review them

Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift
```diff
@@ -15,23 +16,82 @@ struct OptionalDataStringConversionRule: Rule {
...
+            // If configuration.allowImplicitInit is true, we also trigger for the leading-dot
+            // form even when the type is not explicitly annotated as `String`.
+            guard member.base == nil else { return }
+
+            // Walk ancestors to find a VariableDecl or PatternBinding with a type annotation of `String`.
+            if let binding = node.parent?.parent?.as(PatternBindingSyntax.self) {
+                let annotatedString = binding.typeAnnotation?.type.description
+                    .trimmingCharacters(in: .whitespacesAndNewlines) == "String"
+
+                if annotatedString || configuration.allowImplicitInit {
```
SimplyDanny: For this case, we don't need to have a pattern binding as hand. If the expression is allowed without the type (due to `configuration.allowImplicitInit`), we may accept it everywhere no matter the parent.

nadeemnali: Updated to use configuration

PR #6408, opened 2026-01-03
Use URL almost everywhere

### LINKED ISSUE #6351: Fix Glob filtering for included/excluded paths on Windows
SwiftLint can already be built on Windows and the binary even runs. The handling of paths and especially Glob filtering doesn't work yet.

A few tests also fail due to the incomplete support. They should be fixed in the scope of this ticket as well.

### LINKED ISSUE #6352: Fix all tests failing on Windows
#6351 is a prerequisite. The few other tests failing (if still any) can be addressed in the scope of this issue.

Once that's achieved, the build workflow introduced in #6350 shall be extended to ensure all tests remain stable.

SimplyDanny:
Resolves #6351.
Resolves #6352.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -89,8 +89,7 @@ jobs:
           key: windows-${{ matrix.windows-version }}-spm-${{ env.SWIFT_VERSION }}-${{ env.SWIFT_BUILD }}-${{ hashFiles('Package.resolved', 'Package.swift') }}
           restore-keys: windows-${{ matrix.windows-version }}-spm-${{ env.SWIFT_VERSION }}-${{ env.SWIFT_BUILD }}-
           path: .build
-      - name: Build all targets
-        run: swift build --build-tests
-      - name: Run selected tests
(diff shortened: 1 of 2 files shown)


compnerd (commented):
I didn't go through this extremely carefully, but it does seem like it is a significant improvement. Some of this should actually improve performance too (due to the migration to the new API constructs).

Tests/FileSystemAccessTests/GlobTests.swift
```diff
@@ -88,9 +90,23 @@ final class GlobTests: SwiftLintTestCase {
...
+            var root = root
+            var pattern = pattern
+            var filename = filename
+            if root.starts(with: "/") {
+                root = "C:" + root
+            }
+            if pattern.starts(with: "/") {
+                pattern = "C:" + pattern
+            }
+            filename = "C:" + filename
```
compnerd: You should use `SystemDrive`. There is no guarantee that the drive letter is `C:` even for the system. GHA image generally use `D:`, Swift development uses `S:` as the primary area where they are working.

PR #6418, opened 2026-01-07
Fix non-excluded bool literal in optional_enum_case_name

tristan-burnside-anz:
When running with `--fix`, this rule would catch:

```
case (true?, false?)
```

but it should not be catching this case, according to the documentation, and it does not report it in lint mode.


SimplyDanny (approved):
Perfect! Thank you for fixing this inconsistency!

Can you add a `correction` test case that proves the skipped fix?

Issue #6045, opened 2025-04-04
Rule Request: invisible_characters


kapitoshka438:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New Rule Request

We recently faced an issue when some string literals (URLs) were containing a zero-width whitespace (\u200B). The URLs were copypasted from a Confluence document, and the URLs in Confluence, in turn, were copypasted from Swagger specification. This resulted in network request errors in our application. We had to add a custom rule to avoid such situations:
```yml
no_zero_width_spaces:
    name: "No Zero Width Spaces"
    regex: '\u200B'
    message: "Avoid using zero width space characters (U+200B) in your code."
    severity: error
```
I don't know what other invisible characters exist, but application bugs in such situations are hard to track down.
I think it would be useful to add a default rule that would disallow some characters, at least in string literals, and perhaps in all code.
I know at least 3 characters that are not displayed:

- Zero-width space: \u200B
- NULL control character: \u0000
- FEFF Formatting character: \uFEFF

## Triggering Examples
```
let s = "Hello​World" // This is a string with a \u200B character between Hello and World
// \u0048\u0065\u006c\u006c\u006f\u200b\u0057\u006f\u0072\u006c\u0064
```
```
let s = "Hello﻿World" // This is a string with a \ufeff character between Hello and World
// \u0048\u0065\u006c\u006c\u006f\ufeff\u0057\u006f\u0072\u006c\u0064
```
## Non-triggering examples
```
let s = "HelloWorld" // This is a string without invisible characters
// \u0048\u0065\u006c\u006c\u006f\u0057\u006f\u0072\u006c\u0064
```

PR #6424, opened 2026-01-08
Add new invisible_characters rule

kapitoshka438:
https://github.com/realm/SwiftLint/issues/6045
## Summary
  * Adds new `invisible_characters` lint rule that detects invisible characters in string literals:
    * zero-width space U+200B
    * zero-width non-joiner U+200C
    * FEFF formatting character U+FEFF
  * Reports all occurrences with precise violation positions pointing to each invisible character
  * Includes early exit optimization to avoid unnecessary iteration when no invisible characters are present

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/InvisibleCharacterRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/InvisibleCharacterRule.swift
@@ -123,13 +123,9 @@ private extension InvisibleCharacterRule {
                     continue
                 }
                 var utf8Offset = 0
-                var previousScalar: UnicodeScalar?
-                var previousUtf8Size = 0
 
                 for scalar in scalars {
                     defer {
-                        previousScalar = scalar
-                        previousUtf8Size = scalar.utf8.count
                         utf8Offset += scalar.utf8.count
                     }
                     guard violatingCharacters.contains(scalar) else {
@@ -139,40 +135,15 @@ private extension InvisibleCharacterRule {
                     let characterName = InvisibleCharacterConfiguration.defaultCharacterDescriptions[scalar]
                         ?? scalar.escaped(asASCII: true)
 
-                    // Check if this scalar forms a grapheme cluster with the previous one.
-                    // This is needed on Windows and Linux where NSString operations on grapheme clusters
-                    // can delete more than intended when removing a combining character like ZWJ.
-                    let formsCombinedCluster: Bool
-                    if let prev = previousScalar {
-                        let combined = String(prev) + String(scalar)
-                        formsCombinedCluster = combined.count == 1
-                    } else {
-                        formsCombinedCluster = false
-                    }
-
-                    let correctionStart: AbsolutePosition
-                    let replacement: String
-
-                    if formsCombinedCluster, let prev = previousScalar {
-                        // Include previous scalar in the correction range and use it as replacement
-                        correctionStart = stringSegment.content.positionAfterSkippingLeadingTrivia
-                            .advanced(by: utf8Offset - previousUtf8Size)
-                        replacement = String(prev)
-                    } else {
-                        correctionStart = stringSegment.content.positionAfterSkippingLeadingTrivia
-                            .advanced(by: utf8Offset)
-                        replacement = ""
-                    }
-
                     let position = stringSegment.content.positionAfterSkippingLeadingTrivia.advanced(by: utf8Offset)
                     violations.append(
                         ReasonedRuleViolation(
                             position: position,
                             reason: "String literal should not contain invisible character \(characterName)",
                             correction: .init(
-                                start: correctionStart,
+                                start: position,
                                 end: position.advanced(by: scalar.utf8.count),
-                                replacement: replacement
+                                replacement: ""
                             )
                         )
                     )


kapitoshka438:
@SimplyDanny please review. I don't check NULL control character: \u0000 because it gets lost when copypasting so it's almost impossible to have it in string literals by accident.

Source/SwiftLintBuiltInRules/Rules/Lint/InvisibleCharactersRule.swift
```diff
@@ -0,0 +1,101 @@
...
+@SwiftSyntaxRule
+struct InvisibleCharactersRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.error)
+
+    static let description = RuleDescription(
+        identifier: "invisible_characters",
+        name: "Invisible Characters",
+        description: "Disallows invisible characters like zero-width space (U+200B) " +
+                     "and FEFF formatting character (U+FEFF) " +
+                     "in string literals as they can cause hard-to-debug issues",
```
SimplyDanny: You can use a string block.

Source/SwiftLintBuiltInRules/Rules/Lint/InvisibleCharactersRule.swift
```diff
@@ -0,0 +1,101 @@
+import SwiftSyntax
+
+@SwiftSyntaxRule
+struct InvisibleCharactersRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.error)
+
+    static let description = RuleDescription(
+        identifier: "invisible_characters",
```
SimplyDanny: Admittedly not very consistently, but we prefer to use singular for rule names.

Source/SwiftLintBuiltInRules/Rules/Lint/InvisibleCharacterRule.swift
```diff
@@ -0,0 +1,101 @@
...
+                                reason: reason
+                            )
+                        )
+                    }
+                    utf8Offset += String(scalar).utf8.count
+                }
+            }
+        }
+
+        private func characterName(for scalar: Unicode.Scalar) -> String {
```
SimplyDanny: Instead of this `invisibleCharacters` could be a dictionary containing the names as values.

Source/SwiftLintBuiltInRules/Rules/Lint/InvisibleCharacterRule.swift
```diff
@@ -0,0 +1,101 @@
...
+            Hello
+            World
+            \"\"\"
+            """),
+            Example(#"let empty = """#),
+            Example(#"let tab = "Hello\tWorld""#),
+            Example(#"let newline = "Hello\nWorld""#),
+            Example(#"let unicode = "Hello 👋 World""#),
+        ],
+        triggeringExamples: [
```
SimplyDanny, line 33: I'd also be interested in examples like `"abc" + "def"` and `"abc\(name)def"`.

CHANGELOG.md
```diff
@@ -12,7 +12,11 @@
 
 ### Enhancements
 
-* None.
+* Add new default `invisible_characters` rule that detects invisible characters
+  like zero-width space (U+200B) and FEFF formatting character (U+FEFF) in
+  string literals, which can cause hard-to-debug issues.  
+  [kapitoshka438](https://github.com/kapitoshka438)
+  [#6416](https://github.com/realm/SwiftLint/issues/6045)
```
```suggestion
  [#6045](https://github.com/realm/SwiftLint/issues/6045)
```

kapitoshka438:
> Thanks for the contribution!
> 
> I wonder if this rule should catch even more characters of this kind, some maybe optionally. We could check what other linters do.

- This [tool](https://github.com/nhoizey/vscode-gremlins) led me to another character (zero-width non-joiner U+200C). I've added processing it to this PR.
- ESLint [provides a rule](https://eslint.org/docs/latest/rules/no-control-regex) that prevents using some control characters in regular expressions, but I think this is not our case because those characters are visible.
- There is also zero-width joiner character U+200D, but it is used to construct complex symbols, such as some emoji [example](https://github.com/realm/SwiftLint/blob/6405223edbb5eb6c29e55c4835281849f33f3d78/Source/SwiftLintBuiltInRules/Rules/Lint/CommentSpacingRule.swift#L71)

Source/SwiftLintBuiltInRules/Rules/Lint/InvisibleCharacterRule.swift
```diff
@@ -0,0 +1,97 @@
...
+                }) else {
+                    continue
+                }
+
+                // Find all invisible characters and their positions
+                var utf8Offset = 0
+                for scalar in text.unicodeScalars {
+                    if let character = InvisibleCharacterRule.invisibleCharacters[scalar] {
+                        // Calculate absolute position of the invisible character
+                        let characterPosision = stringSegment.content.positionAfterSkippingLeadingTrivia
```
SimplyDanny: Typo in variable name. You can also inline it at its single reference.

Source/SwiftLintBuiltInRules/Rules/Lint/InvisibleCharacterRule.swift
```diff
@@ -0,0 +1,97 @@
...
+                            .advanced(by: utf8Offset)
+                        let reason = "String literal should not contain invisible character: \(character)"
```
SimplyDanny:
```suggestion
                        let reason = "String literal should not contain invisible character \(character)"
```

You might as well inline it.

kapitoshka438:
> Would it be wrong to have the rule auto-correct the findings by just removing the violating characters?
> 
> We could also think about making the rule configurable, so that people can add their own characters with descriptions. That's what the "tool" you cited allows as well.

The second commit makes the rule correctable. I've also made some performance optimizations.
As for custom configurations, let me think about this. I'm sure there may be more symbols we haven't accounted for, but I'm not sure users should be forced to set descriptions for them. Each symbol has a name, and the purpose of this rule is to control the presence of specific special symbols, not just any symbols.

Source/SwiftLintBuiltInRules/Rules/Lint/InvisibleCharacterRule.swift
```diff
@@ -0,0 +1,161 @@
...
+                for scalar in text.unicodeScalars {
+                    defer { utf8Offset += scalar.utf8Length }
+                    guard let characterName = invisibleCharacters[scalar] else {
+                        continue
+                    }
+                    violations.append(
+                        ReasonedRuleViolation(
+                            position: segment.content.positionAfterSkippingLeadingTrivia
+                                .advanced(by: utf8Offset),
+                            reason: "String literal should not contain invisible character \(characterName)"
```
SimplyDanny: Corrections could be a bit easier to achieve. You can add a `correction` parameter here that replaces just ranges in the original code. It's simple but might be sufficient. Rewriters tend to turn quite complex very fast.

Tests/IntegrationTests/Resources/default_rule_configurations.yml
```diff
@@ -546,6 +546,11 @@ invalid_swiftlint_command:
   meta:
     opt-in: false
     correctable: false
+invisible_character:
+  severity: error
+  meta:
+    opt-in: false
+    correctable: false
```
SimplyDanny:
```suggestion
    correctable: true
```

Source/SwiftLintBuiltInRules/Rules/Lint/InvisibleCharacterRule.swift
```diff
@@ -0,0 +1,161 @@
...
+    )
+    // swiftlint:enable invisible_character
+
+    private static let invisibleCharacters: [Unicode.Scalar: String] = [
+        "\u{200B}": "U+200B (zero-width space)",
+        "\u{200C}": "U+200C (zero-width non-joiner)",
+        "\u{FEFF}": "U+FEFF (zero-width no-break space)",
+    ]
+
+    private static let invisibleCharacterSet: Set<Unicode.Scalar> = Set(invisibleCharacters.keys)
```
SimplyDanny: Dictionaries are like sets internally. So I guess that checking whether the dictionary contains a key is as fast as using this separate set.

SimplyDanny:
> I'm sure there may be more symbols we haven't accounted for, but I'm not sure users should be forced to set descriptions for them.

Fine for me to skip the description or have it optional.

kapitoshka438:
> > I'm sure there may be more symbols we haven't accounted for, but I'm not sure users should be forced to set descriptions for them.
> 
> Fine for me to skip the description or have it optional.

I've added the ability to configure additional characters. I thought about it for a long time and initially decided not to allow disabling the validation of default characters. The hardest part was coming up with names for the parameter and variables 😅.  But I'm still having doubts. Please review )

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/InvisibleCharacterConfiguration.swift
```diff
@@ -0,0 +1,39 @@
...
+
+    private static let defaultCharacters = Set(defaultCharacterDescriptions.keys.map(\.hexCode))
+
+    @ConfigurationElement(key: "severity")
+    private(set) var severityConfiguration = SeverityConfiguration<Parent>.error
+    @ConfigurationElement(
+        key: "include_hex_codes",
+        postprocessor: { $0.formUnion(defaultCharacters) }
+    )
+    private(set) var violatingCharacters = Set<String>()
```
SimplyDanny: If you make `UnicodeScalar` conform to `AcceptableByConfigurationElement`, it should end up in a `Set<UnicodeScalar>` right away. This avoids the conversion in `violatingScalars()` and the definition of `defaultCharacters`.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/InvisibleCharacterConfiguration.swift
```diff
@@ -0,0 +1,39 @@
...
+        "\u{200C}": "U+200C (zero-width non-joiner)",
+        "\u{FEFF}": "U+FEFF (zero-width no-break space)",
+    ]
+
+    private static let defaultCharacters = Set(defaultCharacterDescriptions.keys.map(\.hexCode))
+
+    @ConfigurationElement(key: "severity")
+    private(set) var severityConfiguration = SeverityConfiguration<Parent>.error
+    @ConfigurationElement(
+        key: "include_hex_codes",
```
SimplyDanny: We often use `additional_` if the values extend a set of defaults. I'm not sure about "hex codes". Would "code points" be clearer? Then we would end up with `additional_code_points`. Not sure either ...

Source/SwiftLintBuiltInRules/Rules/Lint/InvisibleCharacterRule.swift
```diff
@@ -0,0 +1,129 @@
...
+    var configuration = InvisibleCharacterConfiguration()
+
+    // swiftlint:disable invisible_character
+    static let description = RuleDescription(
+        identifier: "invisible_character",
+        name: "Invisible Character",
+        description: """
+            Disallows invisible characters like zero-width space (U+200B), \
+            zero-width non-joiner (U+200C), and FEFF formatting character (U+FEFF) \
+            in string literals as they can cause hard-to-debug issues
```
SimplyDanny:
```suggestion
            in string literals as they can cause hard-to-debug issues.
```

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/InvisibleCharacterConfiguration.swift
kapitoshka438: Sounds reasonable for me. Done in the last commit.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/InvisibleCharacterConfiguration.swift
kapitoshka438: That's a good point, thanks. I've made changes.

SimplyDanny:
There is something wrong with the tests on Linux and Windows. I can reproduce that on Linux. The test execution is just stuck.

kapitoshka438:
> There is something wrong with the tests on Linux and Windows. I can reproduce that on Linux. The test execution is just stuck.

I can reproduce that on Windows but I wasn't able to debug the issue. The problem is related only to the 0x200D character. For some reason, it just stucks on Windows/Linux if there is an Example with this character. 
I tried turning off `testMultiByteOffsets`, `testOnLinux`, `testOnWindows`, setting `excludeFromDocumentation` to true. Nothing helps.

SimplyDanny:
So that might even be an issue with the compiler. If we were able to create a small reproducer, we could use that to report a bug.

kapitoshka438:
> So that might even be an issue with the compiler. If we were able to create a small reproducer, we could use that to report a bug.

I was able to fix the behavior for Windows/Linux. On Windows, when ZWJ forms a grapheme cluster with the previous character (in this case "o" + ZWJ = one grapheme cluster), the
  NSString.replacingCharacters(in:with:) operation deletes the entire cluster, not just the specified code unit.
  Solution: Added a check in InvisibleCharacterRule.swift: if a violating character (e.g., ZWJ) forms a grapheme cluster with the previous character, then:
  - The correction range is expanded to include the previous character
  - The previous character is used as the replacement (instead of an empty string).

Here is a small piece of code where you can see the difference between Mac OS and Windows/Linux behavior:
```
import Foundation

let s = "Hello\u{200D}World"
print("String: \(s)")
print("Count (graphemes): \(s.count)")
print("UTF-8 count: \(s.utf8.count)")
print("UTF-16 count: \(s.utf16.count)")
print("Unicode scalars count: \(s.unicodeScalars.count)")
print("")

// Check grapheme boundaries
print("Characters (grapheme clusters):")
for (i, char) in s.enumerated() {
     let scalars = char.unicodeScalars.map { $0.value }
     print("  \(i): scalars: \(scalars)")
 }
 print("")

 // Check if "o" and ZWJ form a single grapheme
let utf8 = s.utf8
let idx5 = utf8.index(utf8.startIndex, offsetBy: 5) // After "Hello"
print("UTF-8 index after 5 bytes points to scalar value: \(s.unicodeScalars[idx5].value)")

// Check UTF-16 distance
let utf16 = s.utf16
let utf16Dist = utf16.distance(from: utf16.startIndex, to: idx5)
print("UTF-16 distance to idx5: \(utf16Dist)")

// Now test with NSString
let ns = s as NSString
print("")
print("NSString length: \(ns.length)")

// Test replacement
let range = NSRange(location: 5, length: 1)  // Should be just ZWJ
let result = ns.replacingCharacters(in: range, with: "")
print("After replacing NSRange(5, 1): \(result)")

let range2 = NSRange(location: 4, length: 1)  // Just 'o'
let result2 = ns.replacingCharacters(in: range2, with: "")
print("After replacing NSRange(4, 1): \(result2)")
```

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/InvisibleCharacterConfiguration.swift
```diff
@@ -0,0 +1,39 @@
+import SwiftLintCore
+
+@AutoConfigParser
+struct InvisibleCharacterConfiguration: SeverityBasedRuleConfiguration {
+    static let defaultCharacterDescriptions: [UInt32: String] = [
```
SimplyDanny: This can be `UnicodeScalar` again, can't it?
```suggestion
    static let defaultCharacterDescriptions: [UnicodeScalar: String] = [
```

Source/SwiftLintBuiltInRules/Rules/Lint/InvisibleCharacterRule.swift
```diff
@@ -0,0 +1,183 @@
...
+                        let combined = String(prev) + String(scalar)
+                        formsCombinedCluster = combined.count == 1
+                    } else {
+                        formsCombinedCluster = false
+                    }
+
+                    let correctionStart: AbsolutePosition
+                    let replacement: String
+
+                    if formsCombinedCluster, let prev = previousScalar {
```
SimplyDanny: We can always replace the violating character plus the previous character with just the previous character to avoid two different code paths.

Tests/TestHelpers/TestHelpers.swift
```diff
@@ -198,15 +199,23 @@ public extension Collection where Element == Example {
...
-        markerOffsets.append(markerRange.location)
-        contents = contents.replacingCharacters(in: markerRange, with: "").bridge()
-        markerRange = contents.range(of: violationMarker)
+    var cleanedContents = ""
+    cleanedContents.reserveCapacity(contents.count)
+
+    var iter = contents.makeIterator()
+
+    var offset = 0
+    while let char = iter.next() {
```
SimplyDanny: Can it just be a `for` loop?

Tests/TestHelpers/TestHelpers.swift
SimplyDanny, line 1: Are the changes in this file still necessary?

kapitoshka438: Yeah they are extremely important

Source/SwiftLintBuiltInRules/Rules/Lint/InvisibleCharacterRule.swift
kapitoshka438: I don't think we can. The real issue is not here. This and the changes in TestHelpes are just a workaround for Windows/Linux so that it works correctly for us with the search and replace operation in strings.
I'm not sure, but I think ideally the [correction](https://github.com/realm/SwiftLint/blob/6a4fc9037f0595111d33ab2a3b7a79256f17830c/Source/SwiftLintCore/Protocols/SwiftSyntaxCorrectableRule.swift#L21) work needs to be reworked to avoid incorrect behavior of NSRange / NSString / replacingCharacters in Windows/Linux.
But I wouldn't want to make such critical changes in this pull request.

kapitoshka438:
@SimplyDanny is this ready for merge or anything else you would like to improve?

kapitoshka438:
@SimplyDanny yet another rebase / resolve conflicts / make register is done. Looking forward to merge this PR.

SimplyDanny:
Sorry for the late reaction, @kapitoshka438. I have been quite busy with other tasks.

I tried to address your recommendation with #6570. Could that resolve the issue and harmonize behavior?

kapitoshka438:
> Sorry for the late reaction, @kapitoshka438. I have been quite busy with other tasks.
> 
> I tried to address your recommendation with #6570. Could that resolve the issue and harmonize behavior?

Hey @SimplyDanny . I've applied your changed locally and can prove that it works and allow to remove the Windows/Linux workaround in [InvisibleCharacterRule.swift](https://github.com/realm/SwiftLint/pull/6424/changes#r2756285503).
Changed in TestHelpers.swift are still necessary. Windows test run just stucks when running tests.

SimplyDanny:
> > Sorry for the late reaction, @kapitoshka438. I have been quite busy with other tasks.
> > I tried to address your recommendation with #6570. Could that resolve the issue and harmonize behavior?
> 
> Hey @SimplyDanny . I've applied your changed locally and can prove that it works and allow to remove the Windows/Linux workaround in [InvisibleCharacterRule.swift](https://github.com/realm/SwiftLint/pull/6424/changes#r2756285503). Changed in TestHelpers.swift are still necessary. Windows test run just stucks when running tests.

That's good news for the rule implementation part! Obviously, there are more cleanups necessary in `TestHelpers`. We can do this as a follow-up though.

@kapitoshka438: Would you like to rebase and update your implementation?

kapitoshka438:
> > > Sorry for the late reaction, @kapitoshka438. I have been quite busy with other tasks.
> > > I tried to address your recommendation with #6570. Could that resolve the issue and harmonize behavior?
> > 
> > 
> > Hey @SimplyDanny . I've applied your changed locally and can prove that it works and allow to remove the Windows/Linux workaround in [InvisibleCharacterRule.swift](https://github.com/realm/SwiftLint/pull/6424/changes#r2756285503). Changed in TestHelpers.swift are still necessary. Windows test run just stucks when running tests.
> 
> That's good news for the rule implementation part! Obviously, there are more cleanups necessary in `TestHelpers`. We can do this as a follow-up though.
> 
> @kapitoshka438: Would you like to rebase and update your implementation?

@SimplyDanny done

SimplyDanny:
Thanks for all the effort @kapitoshka438!

I'll check how `TestHelpers` can be improved in a similar way to avoid the other workaround.

PR #6444, opened 2026-01-17
Fix multiline_call_arguments with enum-case patterns.

GandaLF2006:
### Summary

This PR fixes false positives in multiline_call_arguments when enum-case patterns (e.g. .caseOne(...)) are used in pattern-matching positions. SwiftSyntax represents such patterns as FunctionCallExprSyntax, so the rule previously treated them as real calls and produced violations.

### Problem

The rule could incorrectly lint enum-case patterns in pattern-matching constructs, e.g.:
```swift
if case .caseOne(1, 2, 3, 4) = value {
    // ...
}
```
This is a pattern, not a function call, but it was previously linted as a call expression.

### Fix

We now ignore FunctionCallExprSyntax nodes only when they are located inside a pattern subtree (pattern-part only), by walking up the parent chain and checking containment within pattern containers:

- MatchingPatternConditionSyntax (e.g. if/guard/while case ...)
- SwitchCaseItemSyntax (e.g., case .foo(...))
- ForStmtSyntax (e.g., for case let .foo(...) in ...)
- CatchClauseSyntax (e.g. catch EnumCase.foo(...) / catch let EnumCase.foo(...))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,10 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#6080](https://github.com/realm/SwiftLint/issues/6080)
 
+* `multiline_call_arguments` no longer reports violations for enum-case patterns in
+  pattern matching (e.g. if case, switch case, for case, catch).  
+  [GandaLF2006](https://github.com/GandaLF2006)
+
 ## 0.63.2: High-Speed Extraction
 
 ### Breaking
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
@@ -5,105 +5,179 @@ import SwiftSyntax
 struct MultilineCallArgumentsRule: Rule {
     var configuration = MultilineCallArgumentsConfiguration()
 
+    enum Reason {
+        static let singleLineMultipleArgumentsNotAllowed =
+            "Single-line calls with multiple arguments are not allowed"
+
+        static func tooManyArgumentsOnSingleLine(max: Int) -> String {
+            "Too many arguments on a single line (max: \(max))"
+        }
+
+        static let eachArgumentMustStartOnOwnLine =
+            "In multi-line calls, each argument must start on its own line"
+    }
+
     static let description = RuleDescription(
         identifier: "multiline_call_arguments",
         name: "Multiline Call Arguments",
-        description: "Call should have each parameter on a separate line",
+        description: """
+                     Enforces one-argument-per-line for multi-line calls; \
+                     optionally limits or forbids multi-argument single-line calls via configuration
+                     """,
         kind: .style,
-        nonTriggeringExamples: [
-            Example("""
-                foo(
-                param1: "param1",
-                    param2: false,
-                    param3: []
-                )
-                """,
-                configuration: ["max_number_of_single_line_parameters": 2]),
-            Example("""
-                foo(param1: 1,
-                    param2: false,
-                    param3: [])
-                """,
-                configuration: ["max_number_of_single_line_parameters": 1]),
-            Example(
-                "foo(param1: 1, param2: false)",
-                configuration: ["max_number_of_single_line_parameters": 2]),
-            Example(
-                "Enum.foo(param1: 1, param2: false)",
-                configuration: ["max_number_of_single_line_parameters": 2]),
-            Example("foo(param1: 1)", configuration: ["allows_single_line": false]),
-            Example("Enum.foo(param1: 1)", configuration: ["allows_single_line": false]),
-            Example(
-                "Enum.foo(param1: 1, param2: 2, param3: 3)",
-                configuration: ["allows_single_line": true]),
-            Example("""
-                foo(
-                    param1: 1,
-                    param2: 2,
-                    param3: 3
-                )
-                """,
-                configuration: ["allows_single_line": false]),
-        ],
-        triggeringExamples: [
-            Example(
-                "↓foo(param1: 1, param2: false, param3: [])",
-                configuration: ["max_number_of_single_line_parameters": 2]),
-            Example(
-                "↓Enum.foo(param1: 1, param2: false, param3: [])",
-                configuration: ["max_number_of_single_line_parameters": 2]),
-            Example("""
-                ↓foo(param1: 1, param2: false,
-                        param3: [])
-                """,
-                configuration: ["max_number_of_single_line_parameters": 3]),
-            Example("""
-                ↓Enum.foo(param1: 1, param2: false,
-                        param3: [])
-                """,
-                configuration: ["max_number_of_single_line_parameters": 3]),
-            Example("↓foo(param1: 1, param2: false)", configuration: ["allows_single_line": false]),
-            Example("↓Enum.foo(param1: 1, param2: false)", configuration: ["allows_single_line": false]),
-        ]
+        nonTriggeringExamples: MultilineCallArgumentsRuleExamples.nonTriggeringExamples,
+        triggeringExamples: MultilineCallArgumentsRuleExamples.triggeringExamples
     )
 }
 
 private extension MultilineCallArgumentsRule {
     final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        /// Cache line lookups by utf8Offset (stable, cheap key)
+        private var lineCache: [Int: Int] = [:]
+
+        override init(configuration: ConfigurationType, file: SwiftLintFile) {
+            super.init(configuration: configuration, file: file)
+
+            // Most files trigger O(10–100) unique line lookups for this rule.
+            // Reserving a small initial capacity reduces rehashing; it is NOT a hard limit.
+            lineCache.reserveCapacity(64)
+        }
+
         override func visitPost(_ node: FunctionCallExprSyntax) {
-            if containsViolation(parameterPositions: node.arguments.map(\.positionAfterSkippingLeadingTrivia)) {
-                violations.append(node.calledExpression.positionAfterSkippingLeadingTrivia)
-            }
+            // Ignore calls that are part of pattern-matching syntax (patterns only, not bodies).
+            guard !node.isInPatternMatchingPatternPosition else { return }
+
+            let args = node.arguments
+            guard args.count > 1 else { return }
+
+            let argumentPositions = args.map(\.positionAfterSkippingLeadingTrivia)
+            guard let violation = reasonedViolation(argumentPositions: argumentPositions) else { return }
+            violations.append(violation)
         }
 
-        private func containsViolation(parameterPositions: [AbsolutePosition]) -> Bool {
-            var numberOfParameters = 0
-            var linesWithParameters: Set<Int> = []
-            var hasMultipleParametersOnSameLine = false
+        private func reasonedViolation(argumentPositions: [AbsolutePosition]) -> ReasonedRuleViolation? {
+            guard let firstPos = argumentPositions.first else { return nil }
 
-            for position in parameterPositions {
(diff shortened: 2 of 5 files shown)


GandaLF2006:
@SimplyDanny Hey-hey. One more PR to fix rule :)

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
```diff
@@ -5,105 +5,179 @@ import SwiftSyntax
...
+            "In multi-line calls, each argument must start on its own line"
+    }
+
     static let description = RuleDescription(
         identifier: "multiline_call_arguments",
         name: "Multiline Call Arguments",
-        description: "Call should have each parameter on a separate line",
+        description: """
+                     Enforces one-argument-per-line for multi-line calls; \
+                     optionally limits or forbids multi-argument single-line calls via configuration
```
SimplyDanny:
```suggestion
                     optionally limits or forbids multi-argument single-line calls via configuration.
```

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
```diff
@@ -5,105 +5,179 @@ import SwiftSyntax
 struct MultilineCallArgumentsRule: Rule {
     var configuration = MultilineCallArgumentsConfiguration()
 
+    enum Reason {
```
SimplyDanny, line 8: Each of them is only used once, isn't it? Would be fine to inline them.

CHANGELOG.md
```diff
@@ -83,6 +83,9 @@
   in build tool plugins.  
   [SimplyDanny](https://github.com/SimplyDanny)
   [#6080](https://github.com/realm/SwiftLint/issues/6080)
+* `multiline_call_arguments` no longer reports violations for enum-case patterns in
```
SimplyDanny, line 98:
```suggestion

* `multiline_call_arguments` no longer reports violations for enum-case patterns in
```

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRuleExamples.swift
```diff
@@ -0,0 +1,510 @@
...
+
+        Example("""
+            foo(
+                a: (
+                    1,
+                    2
+                ), b: 3
+            )
+            """,
+                configuration: ["max_number_of_single_line_parameters": 10]
```
SimplyDanny: What is this configuration good for?

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRuleExamples.swift
```diff
@@ -0,0 +1,510 @@
...
+                c: 3
+            )
+            """),
+
+        Example("""
+            foo(
+                a: (
+                    1,
+                    2
+                ), b: 3
```
SimplyDanny: Is this really multi-line? 😅

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/MultilineCallArgumentsConfiguration.swift
```diff
@@ -4,19 +4,20 @@ import SwiftLintCore
...
     func validate() throws(Issue) {
-        guard let maxNumberOfSingleLineParameters else {
-            return
-        }
+        guard let maxNumberOfSingleLineParameters else { return }
+
         guard maxNumberOfSingleLineParameters >= 1 else {
             throw Issue.inconsistentConfiguration(
                 ruleID: Parent.identifier,
-                message: "Option '\($maxNumberOfSingleLineParameters.key)' should be >= 1."
```
SimplyDanny, line 19: We are quite inconsistent with that in general, but I like it more with the period. And in this case, the similar message in other rules also use it.

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
```diff
@@ -5,105 +5,179 @@ import SwiftSyntax
...
+    /// not just somewhere inside `if case` / `switch case` bodies.
+    var isInPatternMatchingPatternPosition: Bool {
+        let selfSyntax = Syntax(self)
+        var current: Syntax? = parent
+
+        // Low-level pattern nodes can appear inside multiple contexts; we only need to check each once.
+        var checkedExpressionPattern = false
+        var checkedValueBindingPattern = false
+
+        while let node = current {
```
SimplyDanny: I wonder what an example would be that really requires walking up the whole tree. Isn't it enough to check 1-2 parent levels for specific nodes like `MatchingPatternConditionSyntax` and others?

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRuleExamples.swift
SimplyDanny: Is this really multi-line? 😅

Tests/BuiltInRulesTests/MultilineCallArgumentsRuleTests.swift
```diff
@@ -0,0 +1,102 @@
...
+        )
+    }
+
+    // MARK: - Helpers
+
+    private func makeRule(_ config: [String: Any]) throws -> MultilineCallArgumentsRule {
+        try MultilineCallArgumentsRule(configuration: config)
+    }
+
+    private func validate(_ rule: some Rule, contents: String) -> [StyleViolation] {
```
SimplyDanny: Could accept the config and create the rule instance itself.

SimplyDanny (changes requested):
Thanks for the update! Please consider my comments.

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
GandaLF2006: Added

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/MultilineCallArgumentsConfiguration.swift
GandaLF2006: Added

CHANGELOG.md
GandaLF2006: Fixed

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
GandaLF2006: I kept `Reason` intentionally as a single source of truth for violation messages used by both the rule and the unit tests. This avoids duplicating strings in tests and prevents message drift if we ever tweak wording.

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
GandaLF2006: Good question. The short version is: the relevant “pattern container” isn’t consistently 1–2 parent hops away from the .caseOne(...) node.

For example, in:
```swift
if case let .caseOne(_, _, three, _) = value { }
```

SwiftSyntax models .caseOne(...) as a FunctionCallExprSyntax, but it’s usually wrapped by pattern nodes first. In other words, the parent chain often looks like:

FunctionCallExprSyntax → ExpressionPatternSyntax → ValueBindingPatternSyntax → MatchingPatternConditionSyntax

If we only look 1–2 levels up, we can miss MatchingPatternConditionSyntax and end up linting a pattern as if it were a real call (the exact false positive this PR fixes).

Also, we’re not walking to the root in the normal case: once we hit a top-level container (if/guard/while case, switch case, for case, catch), we check whether we’re inside its pattern subtree and then stop (break). So it stays bounded, but robust.

Tests/BuiltInRulesTests/MultilineCallArgumentsRuleTests.swift
GandaLF2006: Good suggestion — agreed. I’ll update the test helper to accept a config dictionary and instantiate MultilineCallArgumentsRule internally.

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRuleExamples.swift
GandaLF2006: Fixed

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRuleExamples.swift
GandaLF2006: Removed

GandaLF2006:
> Thanks for the update! Please consider my comments.

@SimplyDanny Fixed your comments. Please re-review :)

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
SimplyDanny: Okay. I missed the them being used in unit tests as well.

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
SimplyDanny: Got your point. However, as long as it's only a hand full of cases to check explicitly, I'd still go with them instead of the generic solution for the sake of simplicity.

Also, the `isInside` doesn't have an early exit, has it?

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRuleExamples.swift
```diff
@@ -0,0 +1,529 @@
+// swiftlint:disable file_length
+// swiftlint:disable type_body_length
+internal struct MultilineCallArgumentsRuleExamples {
```
SimplyDanny:
```suggestion
struct MultilineCallArgumentsRuleExamples {
```

Source/SwiftLintBuiltInRules/Rules/Lint/MultilineCallArgumentsRule.swift
GandaLF2006: The pattern subtree can be nested deeper than 1-2 levels. For `case .foo(let a)`, `.foo(...)` is inside ExpressionPatternSyntax → ValueBindingPatternSyntax → MatchingPatternConditionSyntax. We need to walk up to verify we're in the pattern, not the body.

The current code does stop once it finds a container and confirms position — it just uses `isInside` to make that verification simple.

SimplyDanny: Okay, the thing is, if I put
```swift
var isInPatternMatchingPatternPosition: Bool {
    parent?.is(ExpressionPatternSyntax.self) == true
}
```
all tests still pass. So either the current implementation is a little over-engineered or there are test examples missing that would justify it.

GandaLF2006: Good point. Simplified to:

```swift
var isInPatternMatchingPatternPosition: Bool {
    parent?.is(ExpressionPatternSyntax.self) == true
}
```
All tests pass. SwiftSyntax wraps pattern expressions in ExpressionPatternSyntax, so this single check covers all cases.

SimplyDanny (approved):
I'm happy now. 😅 Thanks!

PR #6445, opened 2026-01-20
Add Windows release artifacts

### LINKED ISSUE #6353: Prepare release artifacts for Windows as part of CI
SwiftLint's release steps are fully automated. The Windows build(s) must be made part(s) of the release workflows.

Things to clarify beforehand:

1. Which build artifacts do we need to provide? Just the executable? An installer?
2. Which Windows versions to provide pre-built binaries for?
3. Is the executable allowed to load SourceKit at runtime or do we only provide a fully statically-linked binary that supports only SwiftSyntax-based rules (like we provide a separate static Linux executable)?

SimplyDanny:
Resolves #6353.


SimplyDanny:
@compnerd, @roman-bcny: In order to prepare for a first official (pre-)release including Windows support, I set up this installer build mimicking what you have done for SwiftFormat. Please have a look! 🙏

.github/workflows/windows-release-build.yml
```diff
@@ -0,0 +1,94 @@
...
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  build-windows:
```
SimplyDanny: Needs to be moved into `release.yml` once we are happy with it.

.github/workflows/windows-release-build.yml
```diff
@@ -0,0 +1,94 @@
...
+      - name: Build static executable
+        shell: pwsh
+        env:
+          SWIFTLINT_STATIC_WINDOWS_BUILD: 1
+        run: |
+          $sdk = Join-Path (Split-Path -Path $env:SDKROOT -Parent) 'WindowsExperimental.sdk'
+          $archDir = ('${{ matrix.target-triple }}'.Split('-')[0])
+          $staticLibDir = Join-Path $sdk (Join-Path 'usr\lib\swift_static\windows' $archDir)
+          Copy-Item -Path (Join-Path $staticLibDir 'libcurl.lib') -Destination (Join-Path $staticLibDir 'curl.lib') -Force
+          Copy-Item -Path (Join-Path $staticLibDir 'libxml2s.lib') -Destination (Join-Path $staticLibDir 'xml2.lib') -Force
```
SimplyDanny: This feels very strange. However, it cannot be built without having both names available. I can reproduce this locally.

Platforms/Windows/SwiftLint.wixproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="WixToolset.Sdk/6.0.2">
+  <PropertyGroup>
+    <OutputType>Package</OutputType>
+    <ProductName>SwiftLint</ProductName>
+    <PackageName>SwiftLint</PackageName>
+    <DefineConstants>
+      ProductVersion=$(ProductVersion);
+      ProductArchitecture=$(ProductArchitecture);
```
SimplyDanny: Remove.

Package.swift
```diff
@@ -25,6 +26,19 @@ swiftLintPluginDependencies = [.target(name: "SwiftLintBinary")]
 swiftLintPluginDependencies = [.target(name: "swiftlint")]
 #endif
 
+// Linker flags for static Windows builds.
+let windowsStaticLinkerflags: [LinkerSetting] =
```
SimplyDanny:
```suggestion
let windowsStaticLinkerFlags: [LinkerSetting] =
```

.github/workflows/windows-release-build.yml
```diff
@@ -0,0 +1,94 @@
...
+            wix-platform: x64
+            target-triple: x86_64-unknown-windows-msvc
+          - windows-version: "11-arm"
+            windows-display-name: "11 ARM"
+            arch: arm64
+            arch-display-name: ARM64
+            wix-platform: arm64
+            target-triple: aarch64-unknown-windows-msvc
+    steps:
+      - name: Enable long path support
```
compnerd: IMO this shouldn't be done at all. Foundation should already transparently handle long paths. If you are running into issues, we should try to fix those rather than work around them.

.github/workflows/windows-release-build.yml
roman-bcny: Hmm. Not sure what's up with that. One thing to try could be snapshot from a different date? I had a bunch of those not working last time I checked until I found one that seemed good.

compnerd: This is wrong. I think that the copying here needs to be done because you are missing compile flags. I think that you might need `-Xcc -DCURL_STATICLIB -Xcc -DLIBXML_STATIC`.

.github/workflows/windows-release-build.yml
```diff
@@ -0,0 +1,94 @@
...
+          Copy-Item -Path (Join-Path $staticLibDir 'libxml2s.lib') -Destination (Join-Path $staticLibDir 'xml2.lib') -Force
+          swift build `
+            -c release `
+            --product swiftlint `
+            --triple ${{ matrix.target-triple }} `
+            -debug-info-format none `
+            -Xswiftc -static-stdlib `
+            -Xswiftc -sdk `
+            -Xswiftc $sdk `
+            -Xlinker /LIBPATH:"$staticLibDir"
```
compnerd: This is odd and shouldn't be needed

.github/workflows/windows-release-build.yml
```diff
@@ -0,0 +1,94 @@
...
+      - name: Create portable zip
+        shell: pwsh
+        run: |
+          Compress-Archive -Path "artifacts/swiftlint.exe" -DestinationPath "artifacts/swiftlint-portable-${{ matrix.arch }}.zip" -Force
+      - name: Setup MSBuild
+        uses: microsoft/setup-msbuild@6fb02220983dee41ce7ae257b6f4d8f9bf5ed4ce # v2
+      - name: Build WiX MSI
+        shell: pwsh
+        run: |
+          & msbuild -nologo -restore `
```
compnerd: Do you need `-restore`? Are you using any nuget packages that are being fetched?

.github/workflows/windows-release-build.yml
SimplyDanny: Works indeed: #6449

.github/workflows/windows-release-build.yml
SimplyDanny: It's because of the `PackageReference` in `SwiftLint.wixproj`.

.github/workflows/windows-release-build.yml
SimplyDanny: It exists with
```
[116/118] Write Objects.LinkFileList
lld-link: error: could not open 'xml2.lib': no such file or directory
lld-link: error: could not open 'curl.lib': no such file or directory
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
with only these defines.

SimplyDanny: It's the same with just https://github.com/jpsim/SourceKitten which has Clang dependencies. So I guess some options are not passed on to this lower level and `xml2.lib` is more or less "hard-coded" somewhere, so it needs to be found as such.

compnerd: The problem is that code generation is impacted by that macro and you now have some code assuming static linking and some assuming dynamic linking. I think that we should simply go through and find and address the incorrect expectation.

SimplyDanny: It works with explicit linking options:
```pwsh
-Xlinker /NODEFAULTLIB:curl -Xlinker libcurl.lib
-Xlinker /NODEFAULTLIB:xml2 -Xlinker libxml2s.lib
-Xlinker zlibstatic.lib
-Xlinker brotlidec.lib
-Xlinker brotlicommon.lib
```
I think that's acceptable.

roman-bcny:
🥳

PR #6489, opened 2026-02-11
Make conditional_returns_on_newline rule autocorrectable

ldct:
## Summary

This PR adds autocorrection support to the `conditional_returns_on_newline` rule.

### Changes

- Added a `Rewriter` class to automatically fix violations where return statements are on the same line as conditional keywords (`if`/`guard`)
- The autocorrection:
  - Adds a newline before the return statement
  - Preserves proper indentation (4 spaces from the conditional's indentation)
  - Handles both `if`/`else` and `guard` statements
  - Preserves preceding statements on the same line (e.g., `XCTFail();`)
  - Respects the `if_only` configuration option
- Added test for guard statement corrections

### Example Corrections

```swift
// Before
guard true else { return }

// After
guard true else {
    return
}
```

```swift
// Before
guard condition else { XCTFail(); return }

// After
guard condition else { XCTFail();
    return
}
```

### DIFF
--- a/Tests/BuiltInRulesTests/ConditionalReturnsOnNewlineRuleTests.swift
+++ b/Tests/BuiltInRulesTests/ConditionalReturnsOnNewlineRuleTests.swift
@@ -4,7 +4,6 @@ import TestHelpers
 final class ConditionalReturnsOnNewlineRuleTests: SwiftLintTestCase {
     func testConditionalReturnsOnNewlineWithIfOnly() {
         // Test with `if_only` set to true
-        // guard statements should not trigger or be corrected
         let nonTriggeringExamples = [
             Example("guard true else {\n return true\n}"),
             Example("guard true,\n let x = true else {\n return true\n}"),
@@ -13,6 +12,7 @@ final class ConditionalReturnsOnNewlineRuleTests: SwiftLintTestCase {
             Example("if textField.returnKeyType == .Next {"),
             Example("if true { // return }"),
             Example("/*if true { */ return }"),
+            // guard statements should not trigger or be corrected
             Example("guard true else { return }"),
         ]
         let triggeringExamples = [
@@ -22,7 +22,7 @@ final class ConditionalReturnsOnNewlineRuleTests: SwiftLintTestCase {
             Example("↓if true { return \"YES\" } else { return \"NO\" }"),
         ]
         // Only include `if` corrections - guard corrections should not apply with if_only
-        let corrections = [
+        let expectedCorrections = [
             Example("↓if true { return }"): Example("if true {\n    return\n}"),
             Example("↓if true { break } else { return }"):
                 Example("if true { break } else {\n    return\n}"),
@@ -33,7 +33,7 @@ final class ConditionalReturnsOnNewlineRuleTests: SwiftLintTestCase {
         let description = ConditionalReturnsOnNewlineRule.description
             .with(triggeringExamples: triggeringExamples)
             .with(nonTriggeringExamples: nonTriggeringExamples)
-            .with(corrections: corrections)
+            .with(corrections: expectedCorrections)
 
         verifyRule(description, ruleConfiguration: ["if_only": true])
     }


Source/SwiftLintBuiltInRules/Rules/Style/ConditionalReturnsOnNewlineRule.swift
```diff
@@ -30,6 +31,13 @@ struct ConditionalReturnsOnNewlineRule: Rule {
             Example("""
             ↓guard condition else { XCTFail(); return }
             """),
+        ],
+        corrections: [
+            Example("↓if true { return }"): Example("if true {\n    return\n}"),
+            Example("↓if true { break } else { return }"):
+                Example("if true { break } else {\n    return\n}"),
+            Example("↓if true { return \"YES\" } else { return \"NO\" }"):
+                Example("if true {\n    return \"YES\"\n} else { return \"NO\" }"),
```
SimplyDanny, line 51: Please add another example where the `if` is indented, e.g.
```swift
func f() {
    if true { return }
}
```

Tests/BuiltInRulesTests/ConditionalReturnsOnNewlineRuleTests.swift
```diff
@@ -27,4 +27,19 @@ final class ConditionalReturnsOnNewlineRuleTests: SwiftLintTestCase {
 
         verifyRule(description, ruleConfiguration: ["if_only": true])
     }
+
+    func testGuardCorrection() {
+        // Test guard correction with default configuration
+        let corrections = [
```
SimplyDanny: Can be part of the example list in the rule itself. Make sure, that violating `guard` examples don't get fixed when the `if_only` option is set.

Source/SwiftLintBuiltInRules/Rules/Style/ConditionalReturnsOnNewlineRule.swift
ldct: added as the second example!

Source/SwiftLintBuiltInRules/Rules/Style/ConditionalReturnsOnNewlineRule.swift
```diff
@@ -67,6 +95,95 @@ private extension ConditionalReturnsOnNewlineRule {
...
+            guard isReturn(node.body.statements.lastReturn, onTheSameLineAs: node.guardKeyword) else {
+                return super.visit(node)
+            }
+
+            numberOfCorrections += 1
+            let fixedNode = node.with(\.body, fixCodeBlock(node.body, baseToken: node.guardKeyword))
+            return super.visit(fixedNode)
+        }
+
+        private func isReturn(_ returnStmt: ReturnStmtSyntax?, onTheSameLineAs token: TokenSyntax) -> Bool {
```
SimplyDanny, line 135: Please don't duplicate this method.

Source/SwiftLintBuiltInRules/Rules/Style/ConditionalReturnsOnNewlineRule.swift
```diff
@@ -30,6 +31,33 @@ struct ConditionalReturnsOnNewlineRule: Rule {
...
+            Example("↓if true { break } else { return }"):
+                Example("if true { break } else {\n    return\n}"),
+            Example("↓if true { return \"YES\" } else { return \"NO\" }"):
+                Example("if true {\n    return \"YES\"\n} else { return \"NO\" }"),
+            Example("↓guard true else { return }"):
+                Example("guard true else {\n    return\n}"),
+            Example("""
+                ↓guard condition else { XCTFail(); return }
+                """): Example("""
+                guard condition else { XCTFail();
```
SimplyDanny, line 57: Not sure this is the expected fix here ... and not sure if there is any reasonable fix for that. We should probably restrict fixing to single return statements. That would also simplify the implementation a little.

Alternatively, we could move both statements onto the next line but keeping them together (there is another rule to violate on multiple statements on one line).

Source/SwiftLintBuiltInRules/Rules/Style/ConditionalReturnsOnNewlineRule.swift
```diff
@@ -67,6 +95,95 @@ private extension ConditionalReturnsOnNewlineRule {
...
+            }
+
+            return locationConverter.location(for: returnStmt.returnKeyword.positionAfterSkippingLeadingTrivia).line ==
+                locationConverter.location(for: token.positionAfterSkippingLeadingTrivia).line
+        }
+
+        private func fixCodeBlock(_ block: CodeBlockSyntax, baseToken: TokenSyntax) -> CodeBlockSyntax {
+            // Get the indentation of the base token (e.g., `if` or `guard`)
+            let baseIndentation = baseToken.leadingTrivia.indentation(isOnNewline: true) ?? Trivia()
+            let innerIndentation = Trivia(pieces: baseIndentation.pieces + [.spaces(4)])
```
SimplyDanny, line 147: You can add `Trivia` objects directly.

Source/SwiftLintBuiltInRules/Rules/Style/ConditionalReturnsOnNewlineRule.swift
```diff
@@ -30,6 +31,33 @@ struct ConditionalReturnsOnNewlineRule: Rule {
...
+                Example("if true {\n    return \"YES\"\n} else { return \"NO\" }"),
+            Example("↓guard true else { return }"):
+                Example("guard true else {\n    return\n}"),
+            Example("""
+                ↓guard condition else { XCTFail(); return }
+                """): Example("""
+                guard condition else { XCTFail();
+                    return
+                }
+                """),
```
SimplyDanny, line 60: Make sure to keep leading an trailing comments as well, e.g.
```swift
guard true else { /* leading comment */ return /* trailing comment */ }
```

PR #6490, opened 2026-02-11
Skip force_unwrapping for calls with static string literal args

### LINKED ISSUE #6487: Make force_unwrapping more nuanced
### New Issue Checklist

- [X ] Updated SwiftLint to the latest version
- [ X] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New Rule Request

Modification to an existing rule.

Please describe the rule idea, format this issue's title as `Rule Request: [Rule Name]` and describe:

force_unwrapping flags in cases where it is probably legitimate.

Should not trigger:
let a = Data(hexString: "AABBCCDD")! // argument is a static String
let a = URL(string: "https://www.example.com")! // argument is static String

Should trigger:
let a = Data(hexString: someVar)! // argument is a variable
let a = URL(string: someVar)! // argument is variable

Probably not outside of basing enable/disable. 

4. Should the rule be opt-in or enabled by default? Why?   See [README.md](../README.md#opt-in-rules) for guidelines on when to mark a rule as opt-in.
Whatever current force_unwrapping is.

claudeaceae:
## Summary

Suppresses `force_unwrapping` violations when the force-unwrapped expression is a function call where **all** arguments are static string literals (no interpolation).

Common patterns like these are now allowed:
```swift
let url = URL(string: "https://example.com")!
let data = Data(hexString: "AABBCCDD")!
let image = UIImage(named: "icon")!
```

While these still trigger violations:
```swift
let url = URL(string: variable)!           // variable argument
let url = URL(string: "\(dynamicValue)")!  // string interpolation
```

## Rationale

When a developer writes a function call with a hardcoded string literal, they know the input is valid at compile time. Force unwrapping in this case is a deliberate, informed choice rather than a potential crash risk. This is one of the most common patterns flagged as a false positive by the current rule.

## Changes

- Added `isCallWithOnlyStaticStringArguments` check to the `ForceUnwrapExprSyntax` visitor
- Added `isStaticStringLiteral` helper that verifies a string has no interpolation segments
- Updated `nonTriggeringExamples` and `triggeringExamples` to cover the new behavior
- Moved `NSURL(string: "http://www.google.com")!` from triggering to non-triggering

Closes #6487

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/ForceUnwrappingRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/ForceUnwrappingRule.swift
@@ -37,8 +37,7 @@ struct ForceUnwrappingRule: Rule {
             Example("let url = URL.init(string: \"https://www.example.com\")!"),
             Example(
                 "let result = someFunction(\"constant\")!",
-                configuration: ["ignored_literal_argument_functions": ["someFunction(_:)"]],
-                excludeFromDocumentation: true
+                configuration: ["ignored_literal_argument_functions": ["someFunction(_:)"]]
             ),
         ],
         triggeringExamples: [


claudeaceae:
Added a CHANGELOG entry under Enhancements. Thanks for the CI run — glad to see 0 errors and the real-world violation fixes across projects like Aerial.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/ForceUnwrappingRule.swift
```diff
@@ -63,14 +66,41 @@ struct ForceUnwrappingRule: Rule {
...
+        return arguments.allSatisfy { argument in
+            argument.expression.isStaticStringLiteral
+        }
+    }
+
+    var isStaticStringLiteral: Bool {
+        guard let stringLiteral = `as`(StringLiteralExprSyntax.self) else {
+            return false
+        }
+        return stringLiteral.segments.allSatisfy { $0.is(StringSegmentSyntax.self) }
```
SimplyDanny:
```suggestion
   `as`(StringLiteralExprSyntax.self)?.segments.onlyElement?.is(StringSegmentSyntax.self) == true
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/ForceUnwrappingRule.swift
```diff
@@ -63,14 +66,41 @@ struct ForceUnwrappingRule: Rule {
...
+                return
+            }
             violations.append(node.exclamationMark.positionAfterSkippingLeadingTrivia)
         }
     }
 }
+
+private extension ExprSyntax {
+    var isCallWithOnlyStaticStringArguments: Bool {
+        guard let funcCall = `as`(FunctionCallExprSyntax.self) else {
```
SimplyDanny: I think that this should be restricted to a defined set of function name, because we wouldn't know whether the optionality of `f("abc")!` only depends on the string argument.

SimplyDanny (changes requested):
Thanks! The approach is reasonable, but we should be explicit about the function calls to ignore.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/ForceUnwrappingRule.swift
claudeaceae: Applied — used `onlyElement` and also switched to key path syntax to fix the `prefer_key_path` lint failure. Thanks for the suggestion.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/ForceUnwrappingRule.swift
claudeaceae: That's a fair point. I went with the general heuristic because the set of failable initializers that accept string literals is fairly large across Foundation and UIKit (`URL(string:)`, `NSURL(string:)`, `UIImage(named:)`, `Data(hexString:)`, `NSPredicate(format:)`, `DateFormatter.date(from:)`, etc.) and grows with third-party code too.

The reasoning: if every argument is a static string literal, the developer has full visibility over the inputs and the force unwrap is deliberate. Even if the optionality doesn't solely depend on the string content, the determinism of the call means the developer already knows whether it returns nil.

That said, I can see edge cases where this is too permissive — like `someStatefulObject.lookup("tag")!` where optionality depends on object state, not the string.

Happy to go with a configurable allowlist approach if you prefer. What set of function names would you want to start with?

claudeaceae:
Updated to use an explicit, configurable allowlist per your feedback. The new `static_string_argument_functions` option defaults to:

- `URL(string:)`
- `NSURL(string:)`
- `UIImage(named:)`
- `NSImage(named:)`
- `Data(hexString:)`

Users can extend the set in their `.swiftlint.yml`:

```yaml
force_unwrapping:
  static_string_argument_functions:
    - MyType(identifier:)
```

The postprocessor pattern (from `NoMagicNumbersConfiguration`) ensures defaults are always included — user entries are merged in rather than replacing them.

Also normalizes `Type.init(label:)` to `Type(label:)` so both call forms are matched against the same allowlist entry.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/ForceUnwrappingRule.swift
claudeaceae: Done — switched to a configurable allowlist with sensible defaults. The set uses the `postprocessor` merge pattern from `NoMagicNumbersConfiguration` so user entries extend rather than replace the defaults.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/ForceUnwrappingConfiguration.swift
```diff
@@ -0,0 +1,20 @@
+import SwiftLintCore
+
+@AutoConfigParser
+struct ForceUnwrappingConfiguration: SeverityBasedRuleConfiguration {
+    @ConfigurationElement(key: "severity")
+    private(set) var severityConfiguration = SeverityConfiguration<Parent>(.warning)
+    @ConfigurationElement(
+        key: "static_string_argument_functions",
```
SimplyDanny: We should also encode what they are there for. Should we prefix them with `accepted_` or `ignored_`?

And maybe we should not focus on strings only. This could be likewise implemented for any literal. For new, strings are fine, but we could leave room for a future extension by naming the option more generically.

claudeaceae: Good points on both fronts.

For the prefix, I'd lean toward `ignored_` since the option causes violations to be skipped — it aligns with the rule's perspective ("ignore these cases"). `accepted_` works too but feels slightly more like a value judgment.

For generality, agreed — renaming to drop "string" makes sense so the option name stays valid when we extend to other literals (integer, boolean, etc.). Something like:

- `ignored_literal_functions` — concise, describes what's skipped and why
- `ignored_literal_argument_functions` — more explicit about the mechanism

I'll go with whichever you prefer. My slight preference is `ignored_literal_functions` for brevity, but happy to use the longer form if clarity matters more here.

SimplyDanny: I personally like `ignored_literal_argument_functions` more.

claudeaceae: Sounds good — I'll rename to `ignored_literal_argument_functions` and push the update.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/ForceUnwrappingRule.swift
```diff
@@ -30,13 +30,17 @@ struct ForceUnwrappingRule: Rule {
             Example("func foo() -> [AnyHashable: Any]!"),
             Example("func foo() -> [Int]! { return [] }"),
             Example("return self"),
+            Example("let url = URL(string: \"https://www.example.com\")!"),
+            Example("let data = Data(hexString: \"AABBCCDD\")!"),
+            Example("let image = UIImage(named: \"icon\")!"),
+            Example("let url = NSURL(string: \"http://www.google.com\")!"),
+            Example("let url = URL.init(string: \"https://www.example.com\")!"),
```
SimplyDanny, line 37: We should have one more example proofing that the rule doesn't trigger on an arbitrary function call configured as ignorable.

claudeaceae:
Added — the new non-triggering example uses `someFunction("constant")\!` with `["ignored_literal_argument_functions": ["someFunction(_:)"]]` in the configuration. This mirrors the existing triggering example (`let result = someFunction("constant")↓\!` under default config), so together they prove the configurability end-to-end.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/ForceUnwrappingRule.swift
```diff
@@ -30,13 +30,22 @@ struct ForceUnwrappingRule: Rule {
...
+            Example(
+                "let result = someFunction(\"constant\")!",
+                configuration: ["ignored_literal_argument_functions": ["someFunction(_:)"]],
+                excludeFromDocumentation: true
+            ),
```
SimplyDanny: It's okay to include this into the documentation. The examples show deviating configurations ...
```suggestion
                configuration: ["ignored_literal_argument_functions": ["someFunction(_:)"]]
            ),
```

PR #6495, opened 2026-02-13
Support String.init and .init syntax in optional_data_string_conversion

### LINKED ISSUE #6359: `optional_data_string_conversion` does not trigger when `.init(decoding:as:)` is called instead of `String(decoding:as:)`

claudeaceae:
## Summary

Fixes #6359. The `optional_data_string_conversion` rule now detects `String.init(decoding:as:)` syntax in addition to `String(decoding:as:)`.

A new `include_bare_init` configuration option (default: `false`) enables detection of bare `.init(decoding:as:)` where the `String` type is inferred from context. This is behind an option because SwiftLint's syntactic analysis cannot verify that `.init` actually refers to `String.init`, as noted by @SimplyDanny in #6359.

### Changes

- **Rule implementation**: Rewrote the visitor to match on `FunctionCallExprSyntax` instead of `DeclReferenceExprSyntax`, handling three cases:
  1. `String(decoding:as:)` — always flags (existing behavior)
  2. `String.init(decoding:as:)` — always flags (new)
  3. `.init(decoding:as:)` — flags only when `include_bare_init: true` (new, opt-in)
- **Configuration**: New `OptionalDataStringConversionConfiguration` with `include_bare_init` option, following the same pattern as `ExplicitInitConfiguration`
- **Tests**: Added `OptionalDataStringConversionRuleTests` with `testIncludeBareInit()` for the opt-in behavior
- **CHANGELOG**: Added entry under Enhancements

### Examples

```yaml
# .swiftlint.yml
optional_data_string_conversion:
  include_bare_init: true  # opt-in to detect .init(decoding:as:)
```

```swift
// Always triggers (default config):
String(decoding: data, as: UTF8.self)       // use String(bytes:encoding:) instead
String.init(decoding: data, as: UTF8.self)  // use String(bytes:encoding:) instead

// Only triggers with include_bare_init: true:

// Never triggers:
String(data: data, encoding: .utf8)
String(bytes: data, encoding: .utf8)
```

## Test plan

- [x] Existing triggering/non-triggering examples pass
- [x] New `String.init(decoding:as:)` triggering example added
- [x] `include_bare_init` test verifies opt-in `.init` detection
- [x] Default rule configurations updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,12 @@
 
 ### Enhancements
 
-* None.
+* Support `String.init(decoding:as:)` and `.init(decoding:as:)` syntax in the
+  `optional_data_string_conversion` rule. The bare `.init` form is gated behind
+  a new `include_bare_init` option (default: `false`) since type inference cannot
+  be verified syntactically.
+  [claudeaceae](https://github.com/claudeaceae)
+  [#6359](https://github.com/realm/SwiftLint/issues/6359)
 
 ### Bug Fixes
 
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift
@@ -2,7 +2,7 @@ import SwiftSyntax
 
 @SwiftSyntaxRule
 struct OptionalDataStringConversionRule: Rule {
-    var configuration = SeverityConfiguration<Self>(.warning)
+    var configuration = OptionalDataStringConversionConfiguration()
 
     static let description = RuleDescription(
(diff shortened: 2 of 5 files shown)


SimplyDanny:
Thanks for the PR! However, there is #6372 already which tries to achieve this. I'd prefer to drive that one forward instead.

claudeaceae:
Thanks for pointing that out, @SimplyDanny! I missed #6372 when searching for prior work. Happy to close this in favor of that PR. Apologies for the duplicate.

PR #6503, opened 2026-02-19
Fix `optional_data_string_conversion` not triggering for `.init(decoding:as:)`

### LINKED ISSUE #6359: `optional_data_string_conversion` does not trigger when `.init(decoding:as:)` is called instead of `String(decoding:as:)`

william-laverty:
## Summary

Fixes #6359.

The `optional_data_string_conversion` rule only detected `String(decoding: data, as: UTF8.self)` via `DeclReferenceExprSyntax`. This missed the equivalent implicit member syntax:

```swift
```

## Changes

- Refactored the visitor from `visitPost(DeclReferenceExprSyntax)` to `visitPost(FunctionCallExprSyntax)` to match on the call expression directly
- Added detection for `.init(decoding:as:)` (implicit member access with no base)
- Added triggering and non-triggering test examples for the `.init` pattern

## Before

```swift
// ✅ Triggered
String(decoding: data, as: UTF8.self)

// ❌ Did NOT trigger
```

## After

Both patterns now correctly trigger a violation.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift
@@ -15,23 +15,34 @@ struct OptionalDataStringConversionRule: Rule {
             Example("String(UTF8.self)"),
             Example("String(a, b, c, UTF8.self)"),
             Example("String(decoding: data, encoding: UTF8.self)"),
+            Example("let text: String = .init(data: data, encoding: .utf8)"),
         ],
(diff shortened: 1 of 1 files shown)


SimplyDanny:
There is #6372 already trying to address this. I'd prefer that one to move forward.

william-laverty:
Closing in favor of #6372 as suggested. Thanks for the pointer!

Issue #6407, opened 2026-01-03
Rule Request: Redundant `final` for `actor`s


akuzminskyi:
### New Issue Checklist

- [x] Updated SwiftLint to the latest version
- [x] I searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues)

### New Rule Request

`actor`s in Swift already `final` by default. but developers still sometimes write it as `final actor` and it doesn't produce any compiler warnings. But basically `final` is redundant in this case. similar to `internal` ACL.

I don't see any configuration for the rule. and I think it should be enabled by default as it doesn't change "logic".

```
final actor DataStorage {
```
-> 
```
actor DataStorage {
```

https://github.com/swiftlang/swift-evolution/blob/main/proposals/0306-actors.md#actors-2
> hey are reference types like classes, but do not support inheritance, and therefore do not have (or need) features such as `required` and `convenience` initializers, `overriding`, or class members, `open` and `final`.

PR #6504, opened 2026-02-19
Add `redundant_final_actor` opt-in rule

william-laverty:
## Summary

Implements #6407.

Actors in Swift cannot be subclassed ([SE-0306](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0306-actors.md#actors-2)), so the `final` modifier is always redundant on actor declarations. The compiler doesn't warn about this, but it's unnecessary noise similar to redundant `internal` access control.

## New Rule: `redundant_final_actor`

- **Kind:** Idiomatic
- **Default severity:** Warning
- **Opt-in:** Yes (since it doesn't affect logic, but may surprise developers unfamiliar with actor semantics)
- **Auto-correctable:** Yes — removes the redundant `final` modifier

### Examples

```swift
// Triggering
final actor MyActor {}           // → actor MyActor {}
public final actor DataStore {}  // → public actor DataStore {}

// Non-triggering
actor MyActor {}
final class MyClass {}
```

## Changes

- Added `RedundantFinalActorRule.swift` with visitor + rewriter
- Registered in `BuiltInRules.swift`

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantFinalActorRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantFinalActorRule.swift
@@ -1,6 +1,6 @@
 import SwiftSyntax
 
-@SwiftSyntaxRule(optIn: true)
+@SwiftSyntaxRule(correctable: true, optIn: true)
 struct RedundantFinalActorRule: Rule {
     var configuration = SeverityConfiguration<Self>(.warning)
 
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantFinalActorRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantFinalActorRule.swift
@@ -7,7 +7,12 @@ struct RedundantFinalActorRule: Rule {
     static let description = RuleDescription(
         identifier: "redundant_final_actor",
         name: "Redundant Final on Actor",
-        description: "`final` is redundant on an actor declaration because actors cannot be subclassed",
+        description: "`final` is redundant on an actor declaration and its members because actors cannot be subclassed",
+        rationale: """
+            Actors in Swift currently do not support inheritance, making `final` redundant \
+            on both actor declarations and their members. Note that this may change in future \
+            Swift versions if actor inheritance is introduced.
+            """,
         kind: .idiomatic,
         nonTriggeringExamples: [
             Example("actor MyActor {}"),
@@ -15,6 +20,17 @@ struct RedundantFinalActorRule: Rule {
             Example("""
             @globalActor
             actor MyGlobalActor {}
+            """),
+            Example("""
+            actor MyActor {
+                func doWork() {}
+                var value: Int { 0 }
+            }
+            """),
+            Example("""
+            class MyClass {
+                final func doWork() {}
+            }
             """),
         ],
         triggeringExamples: [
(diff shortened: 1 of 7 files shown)


Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantFinalActorRule.swift
```diff
@@ -0,0 +1,63 @@
...
+                Example("public actor DataStore {}"),
+        ]
+    )
+}
+
+private extension RedundantFinalActorRule {
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        override func visitPost(_ node: ActorDeclSyntax) {
+            if let finalModifier = node.modifiers.first(where: { $0.name.text == "final" }) {
+                violations.append(finalModifier.positionAfterSkippingLeadingTrivia)
```
SimplyDanny: You can append violations objects here which can contain a correction. It's a simple range-replace-based correction, but should be sufficient for this use case. If something is possible at all, it is often much easier than an explicit rewriter.

SimplyDanny (commented):
Good idea!

Have in mind, however, that actors don't support inheritance now. That is open to change in the future. We should perhaps mention that in a rational. `RuleDescription` support a `rational` value we can use to explain that a little and consider it a caveat.

What about `final` methods and properties?

```swift
actor A {
  final func f() {}
}
```

SimplyDanny:
What about `final` methods and properties? Would you like to support linting them as well?

william-laverty:
Good callouts, thanks.

1. **Rationale**: I'll add a `rationale` noting that actors currently don't support inheritance, so `final` is redundant — with a caveat that this may change in future Swift versions.

2. **`final` methods and properties**: Yes, I think we should lint those too. `final func f()` and `final var x` inside an actor are equally redundant for the same reason. I'll extend the visitor to catch those.

I'll push both changes shortly.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantFinalActorRule.swift
```diff
@@ -0,0 +1,125 @@
+import SwiftSyntax
+
+@SwiftSyntaxRule(correctable: true, optIn: true)
+struct RedundantFinalActorRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "redundant_final_actor",
+        name: "Redundant Final on Actor",
+        description: "`final` is redundant on an actor declaration and its members because actors cannot be subclassed",
```
SimplyDanny, line 10: Keep it short:
```suggestion
        description: "`final` is redundant on an actor declaration and its members",
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantFinalActorRule.swift
```diff
@@ -0,0 +1,125 @@
...
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        private var insideActor = false
+
+        override func visit(_ node: ActorDeclSyntax) -> SyntaxVisitorContinueKind {
+            insideActor = true
+            return .visitChildren
+        }
+
+        override func visitPost(_ node: ActorDeclSyntax) {
+            if let finalModifier = node.modifiers.first(where: { $0.name.text == "final" }) {
```
SimplyDanny, line 75: We could have the keyword lookup in an extension on `DeclModifierListSyntax`. You might even add it to `SwiftSyntax+SwiftLint.swift`.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/RedundantFinalActorRule.swift
```diff
@@ -0,0 +1,125 @@
...
+                Example("public actor DataStore {}"),
+            Example("actor MyActor {\n    final func doWork() {}\n}"):
+                Example("actor MyActor {\n    func doWork() {}\n}"),
+        ]
+    )
+}
+
+private extension RedundantFinalActorRule {
+    final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
+        private var insideActor = false
```
SimplyDanny, line 67: This is not sufficient. You would need to have a stack to track "is actor" when entering a type declaration. Otherwise, you cannot correctly handle nested types, e.g.

```swift
struct S {
  final actor A {}
}
```

Another case that justifies a stack is

```swift
actor A {
  final var x: Int { 1 } // Rule triggers here.
  struct S {}
  final var y: Int { 2 } // Would it trigger here as well?
}
```

SimplyDanny (changes requested):
Sorry for the late reply ... here are a few more comments.

Please also add a changelog entry to credit yourself and let others know about the new rule.

PR #6506, opened 2026-02-20
Add `disallow_default_parameter` opt-in rule

### LINKED ISSUE #6488: Rule Request: disallow_default_parameter
### New Issue Checklist

### New Rule Request

(Dis)allow default args in functions based on their access level.

#### Why?

I've often run into cases where a bug slips in because a parameter is not properly forwarded down to the layer that needs it because the compiler doesn't inform me that the argument is necessary for a particular function. 
To combat this, I've made a rule of thumb for myself not to use default arguments on private or internal functions so that the compiler will force me to be explicit about how the functions should be parameterized.

This works well internal to a package because I think it's reasonable to expect the maintainers/developers to know enough about the internals of the package to know how to parameterize functions appropriately. Default arguments are very useful for consumers of the public API though.

#### Example

**Triggering**
```swift
// Depending on configuration
private func foo(_: Int = 5) {}
internal func bar(n: UInt = 1000) -> String { /* ... */ }
package  func bar(n: UInt = 1000) -> String { /* ... */ }
```

**Non-triggering**
```swift
public func foo(n: UInt = 100) -> String { /* ... */ }
```

#### Configurations
- Which access levels to trigger for (`private`, `internal` `package`)

#### Opt In?
Yes - opt-in. It doesn't feel like a hard and fast swift convention - rather something that I've personally been bitten by after years of Swift development.

william-laverty:
### New Rule

Adds a new **opt-in** rule `disallow_default_parameter` that flags default parameter values on functions with certain access levels.

#### Motivation

As described in #6488, default parameter values on non-public functions can mask bugs when parameters aren't properly forwarded. This rule enforces explicit parameterization at internal API boundaries while allowing defaults on public APIs where consumers benefit from convenience.

#### Configuration

- `disallowed_access_levels`: Array of access levels to flag. Default: `["private", "internal"]`
- Supported values: `private`, `fileprivate`, `internal`, `package`

```yaml
disallow_default_parameter:
  disallowed_access_levels:
    - private
    - internal
    - package
```

#### Examples

**Triggering** (default config):
```swift
private func foo(bar: Int = 0) {}  // ⚠️
func foo(bar: Int = 0) {}          // ⚠️ (implicit internal)
internal func foo(bar: Int = 0) {} // ⚠️
```

**Non-triggering**:
```swift
public func foo(bar: Int = 0) {}   // ✅ public is always allowed
private func foo(bar: Int) {}      // ✅ no default
```

Closes #6488

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -21,6 +21,7 @@ disabled_rules:
   - conditional_returns_on_newline
   - contrasted_opening_brace
   - convenience_type
+  - discouraged_default_parameter
   - discouraged_optional_collection
   - explicit_acl
   - explicit_enum_raw_value


Source/SwiftLintBuiltInRules/Rules/Idiomatic/DisallowDefaultParameterRule.swift
```diff
@@ -0,0 +1,63 @@
...
+@SwiftSyntaxRule(optIn: true)
+struct DisallowDefaultParameterRule: Rule {
+    var configuration = DisallowDefaultParameterConfiguration()
+
+    static let description = RuleDescription(
+        identifier: "disallow_default_parameter",
+        name: "Disallow Default Parameter",
+        description: "Default parameter values should not be used in functions with certain access levels",
+        kind: .idiomatic,
+        nonTriggeringExamples: DisallowDefaultParameterRuleExamples.nonTriggeringExamples,
```
SimplyDanny: If there aren't too many examples, it's okay to inline them here.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/DisallowDefaultParameterRule.swift
```diff
@@ -0,0 +1,63 @@
+import SwiftSyntax
+
+@SwiftSyntaxRule(optIn: true)
+struct DisallowDefaultParameterRule: Rule {
+    var configuration = DisallowDefaultParameterConfiguration()
+
+    static let description = RuleDescription(
+        identifier: "disallow_default_parameter",
+        name: "Disallow Default Parameter",
+        description: "Default parameter values should not be used in functions with certain access levels",
```
SimplyDanny:
```suggestion
        description: "Default parameter values should not be used in functions with certain access levels.",
```

and use a shorter reason with every violation mentioning the access level like

```swift
"Default parameter values should not be used in '<visibility>' functions"
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/DisallowDefaultParameterRule.swift
```diff
@@ -0,0 +1,63 @@
...
+        kind: .idiomatic,
```
SimplyDanny: Fits better:
```suggestion
        kind: .lint,
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/DisallowDefaultParameterRule.swift
```diff
@@ -0,0 +1,63 @@
...
+        override func visitPost(_ node: SubscriptDeclSyntax) {
+            guard matchesDisallowedAccessLevel(node.modifiers) else { return }
+            for param in node.parameterClause.parameters where param.defaultValue != nil {
+                violations.append(param.defaultValue!.positionAfterSkippingLeadingTrivia)
+            }
+        }
+
+        private func collectViolations(
+            modifiers: DeclModifierListSyntax,
+            signature: FunctionSignatureSyntax
```
SimplyDanny: If you use `ParameterClauseSyntax` here, you can implement the case for `SubscriptDeclSyntax` exactly like the other two.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/DisallowDefaultParameterRule.swift
```diff
@@ -0,0 +1,63 @@
...
+        ) {
+            guard matchesDisallowedAccessLevel(modifiers) else { return }
+            for param in signature.parameterClause.parameters where param.defaultValue != nil {
+                violations.append(param.defaultValue!.positionAfterSkippingLeadingTrivia)
```
SimplyDanny: I'd prefer to have the default parameter unwrapped in an optional binding instead of relying on the `where` implicitly.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/DisallowDefaultParameterConfiguration.swift
```diff
@@ -0,0 +1,24 @@
...
+
+@AutoConfigParser
+struct DisallowDefaultParameterConfiguration: SeverityBasedRuleConfiguration {
+    @ConfigurationElement(key: "severity")
+    private(set) var severityConfiguration = SeverityConfiguration<Parent>(.warning)
+    @ConfigurationElement(
+        key: "disallowed_access_levels",
+        postprocessor: { $0 = Set($0.sorted()) }
+    )
+    private(set) var disallowedAccessLevels: Set<AccessLevel> = [.private, .internal]
```
SimplyDanny: Why do you suggest these two as defaults?

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/DisallowDefaultParameterConfiguration.swift
```diff
@@ -0,0 +1,24 @@
+import SwiftLintCore
+
+@AutoConfigParser
+struct DisallowDefaultParameterConfiguration: SeverityBasedRuleConfiguration {
+    @ConfigurationElement(key: "severity")
+    private(set) var severityConfiguration = SeverityConfiguration<Parent>(.warning)
+    @ConfigurationElement(
+        key: "disallowed_access_levels",
+        postprocessor: { $0 = Set($0.sorted()) }
```
SimplyDanny: Postprocessor can be removed. Neither order nor uniqueness are important.

Incidentally, with `Set($0.sorted())` sorting would have no effect. What you want is `Set($0).sorted()`, but then `disallowedAccessLevels` would need to be a list.

SimplyDanny (changes requested):
Thanks for the contribution!

There are a few remarks I'd like to have addressed before merging.

william-laverty:
Hi @SimplyDanny, I've addressed all the review feedback in the previous push — description updated, kind changed to `.lint`, examples inlined, optional binding for default parameter, postprocessor removed, and defaults trimmed to just `[.internal]`. Please let me know if there's anything else!

CHANGELOG.md
```diff
@@ -12,6 +12,11 @@
 
 ### Enhancements
 
+* Add `disallow_default_parameter` opt-in rule that flags default parameter
+  values in functions with configurable access levels.  
+  [William-Laverty](https://github.com/William-Laverty)
+  [#6506](https://github.com/realm/SwiftLint/pull/6506)
```
SimplyDanny: Should be the issue number, not the PR's one.

Tests/IntegrationTests/Resources/default_rule_configurations.yml
```diff
@@ -195,6 +195,12 @@ direct_return:
   meta:
     opt-in: true
     correctable: true
+disallow_default_parameter:
+  severity: warning
+  disallowed_access_levels: [internal, private]
```
SimplyDanny: Needs updating.

Source/SwiftLintBuiltInRules/Rules/Lint/DiscouragedDefaultParameterRule.swift
SimplyDanny, line 1: Should be moved to `Rules/Lint`.

SimplyDanny (changes requested):
I'd still like to understand what the reasoning behind `internal` as a default is. Why not `package`, `fileprivate` and `private` as well?

Before you change anything, let's first discuss the "why" and "what".

william-laverty:
Good point. I was being conservative with the default, but you're right — the rationale applies equally to `package`, `fileprivate`, and `private`. If callers are within the same module (or even the same file), explicit arguments are arguably *more* warranted, not less.

I'd propose changing the default to `[.internal, .package, .fileprivate, .private]`, excluding only `public` and `open` where default parameters are a standard API design pattern. Does that align with your thinking?

SimplyDanny:
> Good point. I was being conservative with the default, but you're right — the rationale applies equally to `package`, `fileprivate`, and `private`. If callers are within the same module (or even the same file), explicit arguments are arguably _more_ warranted, not less.
> 
> I'd propose changing the default to `[.internal, .package, .fileprivate, .private]`, excluding only `public` and `open` where default parameters are a standard API design pattern. Does that align with your thinking?

I'd definitely support `internal` and `package`. For `private` and `fileprivate` you could argue that they are not at all exposed. We can call them only inside of a single file. So their scope is small and default arguments are easy to find and know about.

william-laverty:
Good point — agreed. I've updated the default to `[.internal, .package]`, keeping `private` and `fileprivate` excluded since their scope is small enough that default arguments are easy to discover. Updated the examples accordingly.

PR #6508, opened 2026-02-24
Add optional shorthand .init support to optional_data_string_conversion

### LINKED ISSUE #6359: `optional_data_string_conversion` does not trigger when `.init(decoding:as:)` is called instead of `String(decoding:as:)`

theamodhshetty:
## Summary
This updates `optional_data_string_conversion` to:
- detect `String.init(decoding:as:)` in addition to `String(decoding:as:)`
- optionally detect shorthand `.init(decoding:as:)` via a new config option:
  - `include_shorthand_init` (default: `false`)

The shorthand case is behind an option because `.init(...)` can be ambiguous without type context.

## Changes
- Added `OptionalDataStringConversionConfiguration` with:
  - `severity`
  - `include_shorthand_init` (default `false`)
- Updated rule visitor logic and examples
- Added `include_shorthand_init: false` to default rule configurations
- Added changelog entry under `Main > Enhancements`

## Testing
- `SWIFTPM_ENABLE_KEYCHAIN=0 swift test --filter OptionalDataStringConversionRuleGeneratedTests`
- `SWIFTPM_ENABLE_KEYCHAIN=0 swift test --filter RuleConfigurationTests`
- `SWIFTPM_ENABLE_KEYCHAIN=0 swift run swiftlint lint Source/SwiftLintBuiltInRules/Rules/Lint/OptionalDataStringConversionRule.swift`

Closes #6359

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,12 @@
   [ahmadalfy](https://github.com/ahmadalfy)
   [#6499](https://github.com/realm/SwiftLint/issues/6499)
 
+* Add `include_shorthand_init` option to
+  `optional_data_string_conversion` rule to optionally lint shorthand
+  `.init(decoding:as:)` calls as `String` conversions.  
+  [theamodhshetty](https://github.com/theamodhshetty)
+  [#6359](https://github.com/realm/SwiftLint/issues/6359)
+
 ### Bug Fixes
 
 * Ensure that disable commands work for `redundant_nil_coalescing` rule.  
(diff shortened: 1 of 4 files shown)


theamodhshetty:
Aligned this with the guidance in #6359 by keeping shorthand .init(decoding:as:) behind an opt-in configuration flag (include_shorthand_init). Happy to adjust naming or defaults if you prefer a different shape.

SimplyDanny:
Thanks for the PR! However, there is #6372 already which tries to achieve this. I'd prefer to drive that one forward instead.

PR #6513, opened 2026-02-27
implicit_optional_initialization: add ignore_attributes support

### LINKED ISSUE #3998: `redundant_optional_initialization` warns on wrapped nilable property leading to compile errors

theamodhshetty:
## Summary

Adds a targeted escape hatch for `implicit_optional_initialization` when explicit `= nil` is required by specific wrappers/attributes.

Fixes #3998.

## What changed

- Added `ignore_attributes` configuration to `implicit_optional_initialization`.
- Updated rule logic to skip style checks for declarations annotated with ignored attributes (for `style: always`).
- Added rule examples covering both ignored and non-ignored attribute cases.
- Added a changelog entry.

## Why

Some property wrappers/attributes require explicit optional initialization (`= nil`) to compile.
A previous broad approach (skip all custom attributes) was too permissive. This keeps behavior explicit and configurable.

## Validation

- `swift test --filter ImplicitOptionalInitializationRuleGeneratedTests`

## Example config

```yaml
implicit_optional_initialization:
  style: always
  ignore_attributes:
    - Wrapper
```

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRuleExamples.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/ImplicitOptionalInitializationRuleExamples.swift
@@ -139,6 +139,20 @@ enum ImplicitOptionalInitializationRuleExamples { // swiftlint:disable:this type
 
     static let corrections = [
         // never style
+        Example(
+            """
+            @Wrapper("name")
+            var flag: Bool?
+            """,
+            configuration: [
+                "style": "never",
+                "ignore_attributes": ["Wrapper"],
+            ]):
+            Example(
+                """
+                @Wrapper("name")
+                var flag: Bool?
+                """),
         Example("var ↓myVar: Int? // comment", configuration: ["style": "never"]):
             Example("var myVar: Int? = nil // comment"),
         Example("var ↓myVar: Optional<Int> // comment", configuration: ["style": "never"]):
@@ -168,6 +182,20 @@ enum ImplicitOptionalInitializationRuleExamples { // swiftlint:disable:this type
             }
             """),
 
+        Example(
+            """
+            @Wrapper("name")
+            var flag: Bool? = nil
+            """,
+            configuration: [
+                "style": "always",
+                "ignore_attributes": ["Wrapper"],
+            ]):
+            Example(
+                """
+                @Wrapper("name")
+                var flag: Bool? = nil
+                """),
         Example("var ↓myVar: Int? = nil // comment", configuration: ["style": "always"]):
             Example("var myVar: Int? // comment"),
         Example("var ↓myVar: Optional<Int> = nil // comment", configuration: ["style": "always"]):


Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/ImplicitOptionalInitializationConfiguration.swift
```diff
@@ -12,4 +12,6 @@ struct ImplicitOptionalInitializationConfiguration: SeverityBasedRuleConfigurati
     private(set) var severityConfiguration = SeverityConfiguration<Parent>(.warning)
     @ConfigurationElement(key: "style")
     private(set) var style: Style = .always
+    @ConfigurationElement(key: "ignore_attributes")
```
SimplyDanny, line 15: I wonder if we could find a clearer name for this option. We don't ignore attributes but the variables to which the attributes are attached to.

theamodhshetty: Good point. I kept `ignore_attributes` for consistency with other rules (e.g. `redundant_type_annotation`), but I am happy to rename if you prefer a different key.

theamodhshetty:
Thanks for the catch on `style: never`.

I pushed 1da5182 to apply `ignore_attributes` checks for both styles (`always` and `never`) and added examples/tests for the `never` path.

theamodhshetty:
Follow-up fix for the failing Registration step:

- ran rule/reporter registration and committed the generated default config update
- added ignore_attributes: [] under implicit_optional_initialization in Tests/IntegrationTests/Resources/default_rule_configurations.yml

Commit: d9e7257

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/ImplicitOptionalInitializationConfiguration.swift
SimplyDanny: Thanks for the clarification. Consistency beats taste here.

theamodhshetty:
Thanks for calling this out. I added explicit correction coverage to ensure the rewriter follows `ignore_attributes` for both styles:

- `style: never` + ignored attribute remains unchanged
- `style: always` + ignored attribute remains unchanged

I also ran:
`swift test --filter ImplicitOptionalInitializationRuleGeneratedTests`

and it passes locally.

PR #6521, opened 2026-02-28
docs(readme): clarify SwiftLintPlugin path with clonedSourcePackagesDirPath

### LINKED ISSUE #5821: SwiftPlugins setup doesn't work when specifying "clonedSourcePackagesDirPath" in xcodebuild
### New Issue Checklist

- [x ] I've Updated SwiftLint to the latest version.
- [ x] I've searched for [existing GitHub issues](https://github.com/realm/SwiftLint/issues).

### Bug Description

The setup mentioned in the docs for SwiftPlugin doesn't work when executed with `clonedSourcePackagesDirPath`:
`xcodebuild --clonedSourcePackagesDirPath ".build"`

```bash
SWIFT_PACKAGE_DIR="${BUILD_DIR%Build/*}SourcePackages/artifacts"
SWIFTLINT_CMD=$(ls "$SWIFT_PACKAGE_DIR"/swiftlintplugins/SwiftLintBinary/SwiftLintBinary.artifactbundle/swiftlint-*/bin/swiftlint | head -n 1)

if test -f "$SWIFTLINT_CMD" 2>&1
then
    "$SWIFTLINT_CMD"
else
    echo "warning: `swiftlint` command not found - See https://github.com/realm/SwiftLint#installation for installation instructions."
fi
```

Packages are then stored in a that folder, not the `%BUILD_DIR`.

### Environment

* SwiftLint version: any
* Xcode version: 15.0, 16.0
* Installation method used: SwiftLintPlugins

theamodhshetty:
## Summary
This updates the SwiftLintPlugin run script guidance in the README to make custom package directories explicit.

When users build with `xcodebuild -clonedSourcePackagesDirPath <path>`, the previous snippet implied only the default `SourcePackages/artifacts` location. This change makes the script configurable via `SWIFT_PACKAGE_DIR` and documents the expected custom path (`<path>/artifacts`).

## Changes
- README run script now allows overriding `SWIFT_PACKAGE_DIR` via environment variable.
- Added a note for `-clonedSourcePackagesDirPath` usage.
- Added changelog entry under `Main > Enhancements`.

## Why this helps
It removes ambiguity for projects using a non-default package location and prevents false "swiftlint command not found" warnings when the plugin artifacts live outside the default Xcode path.

Closes #5821

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,12 +16,6 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#6501](https://github.com/realm/SwiftLint/issues/6501)
 
-* Clarify SwiftLintPlugin run script guidance in the README for
-  `xcodebuild -clonedSourcePackagesDirPath` builds by documenting how to set
-  `SWIFT_PACKAGE_DIR` to the custom artifacts location.  
-  [theamodhshetty](https://github.com/theamodhshetty)
-  [#5821](https://github.com/realm/SwiftLint/issues/5821)
-
 * Add `ignored_literal_argument_functions` option to the `force_unwrapping` rule
   to skip violations for configurable function calls when all arguments are
   literal values (e.g. `URL(string: "https://example.com")!`). Defaults


README.md
```diff
@@ -351,6 +351,8 @@ fi
 > The `SWIFTLINT_CMD` path uses the default Xcode configuration and has been
 > tested on Xcode 15/16. In case of another configuration (e.g. a custom
 > Swift package path), please adapt the values accordingly.
+> If you run `xcodebuild` with `-clonedSourcePackagesDirPath`, set
+> `SWIFT_PACKAGE_DIR` to `<that path>/artifacts` before running the script.
```
SimplyDanny: Why would we not append the `/artifacts` suffix ourselves in the script?

theamodhshetty: Good point. I updated the snippet so `SWIFT_PACKAGE_DIR` stays at the package root and the script appends `/artifacts` itself. The note below now points `-clonedSourcePackagesDirPath` users at `<that path>` instead of `<that path>/artifacts`.

CHANGELOG.md
```diff
@@ -16,6 +16,12 @@
   [SimplyDanny](https://github.com/SimplyDanny)
   [#6501](https://github.com/realm/SwiftLint/issues/6501)
 
+* Clarify SwiftLintPlugin run script guidance in the README for
+  `xcodebuild -clonedSourcePackagesDirPath` builds by documenting how to set
+  `SWIFT_PACKAGE_DIR` to the custom artifacts location.  
+  [theamodhshetty](https://github.com/theamodhshetty)
+  [#5821](https://github.com/realm/SwiftLint/issues/5821)
```
SimplyDanny: We don't add changelog items for documentation updates. Please remove.

theamodhshetty: Removed in a060316. You were right to call that out; this PR is docs-only, so the changelog entry was unnecessary.

PR #6523, opened 2026-03-02
Add `explicit_return` opt-in rule

### LINKED ISSUE #3632: Rule Request: `implicit_return` should also enforce explicit returns
### New Issue Checklist

### New rule request

`implicit_return` should also enforce explicit returns

#### Why should this rule be added?

When `implicit_return` is configured like:
```yaml
implicit_return:
    included:
        - closure
        - getter
```
I think it should also enforce explicit returns in contexts not included in the config (`function` in this example).

I always assumed it was a bug (and replied in #3029) until I read the source code.

### LINKED ISSUE #3859: Rule Request: `explicit_return`
**New rule request**

`explicit_return` would check that the return is always present.

**Description**

I've found a rule for hiding return where is no needed https://realm.github.io/SwiftLint/implicit_return.html and a discussion about new rule about checking that the return exist https://github.com/realm/SwiftLint/issues/3632.  
There is a discussion about that removing `return` is confusing and make it slow to reed a code. https://forums.swift.org/t/how-is-se-0255-implicit-returns-not-a-violation-of-swifts-design-principles/29267

talanov:
Add `explicit_return` opt-in rule that warns against omitting the `return` keyword inside closures, functions and getters. This rule provides behavior opposite to `implicit_return` and complements it. It's disabled by default and provides configurable return kinds and severity.

resolves #3632 
resolves #3859
based on #3905 

Changes from #3905:

* Rewritten to use `@SwiftSyntaxRule`, `@AutoConfigParser`, `ViolationsSyntaxVisitor` (matching `implicit_return`)
* Added subscript and initializer return kinds
* Initializer checks are scoped to failable (init?) only, since non-failable inits cannot implicitly return expressions
* Tests follow the `ImplicitReturnRuleTests` pattern (per-kind subsets verify absence of cross-kind false positives)

Known limitation:
The closure check doesn't have access to type information, so it can't tell a Void closure (`forEach`, `async`, completion handlers, etc.) apart from one that actually returns a value (see discussion in #3905).

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -25,6 +25,7 @@ disabled_rules:
   - discouraged_optional_collection
   - explicit_acl
   - explicit_enum_raw_value
+  - explicit_return
   - explicit_top_level_acl
   - explicit_type_interface
   - file_types_order
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,13 @@
   built-in rules in accordance with the SARIF specification.  
   [ahmadalfy](https://github.com/ahmadalfy)
   [#6499](https://github.com/realm/SwiftLint/issues/6499)
+  
+* Add `explicit_return` opt-in rule that warns against omitting the `return`
+  keyword inside closures, functions and getters.
+  [m-chojnacki](https://github.com/m-chojnacki)
+  [talanov](https://github.com/talanov)
+  [#3632](https://github.com/realm/SwiftLint/issues/3632)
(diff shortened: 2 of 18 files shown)


rgoldberg:
Would it make more sense to add an option for `implicit_return` that is set to `always` by default, but that can be changed to `never`?

It's nice to be able to enable `all` rules without needing to explicitly disable either the existing `implicit_return` or this proposed `explicit_return`.

It's also nicer not to unnecessarily increase the rule count.

PR #6534, opened 2026-03-09
Improve pattern_matching_keywords SwiftSyntax rule

GandaLF2006:
## Summary

This PR improves the existing opt-in SwiftLint rule `pattern_matching_keywords`.

Compared to the original implementation, the rule now covers more pattern-matching contexts and handles nested tuple/enum associated value patterns more reliably.

## What changed

The original implementation only analyzed `switch` case items and relied on a narrower syntax-based approach.

This update expands the rule to also check:

- `if case`
- `guard case`
- `while case`
- `for case`
- `catch`

It also improves handling of:

- nested tuple patterns
- nested enum associated value patterns
- labeled associated values
- multi-pattern `case` clauses

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,10 @@
   [theamodhshetty](https://github.com/theamodhshetty)
   [#6161](https://github.com/realm/SwiftLint/issues/6161)
 
+* Improve the opt-in `pattern_matching_keywords` rule by extending support
+  beyond `switch case` and refining nested pattern handling.  
+  [GandaLF2006](https://github.com/GandaLF2006)
+
 ### Bug Fixes
 
 * Detect and autocorrect missing whitespace before `else` in `guard`
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/PatternMatchingKeywordsRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/PatternMatchingKeywordsRule.swift
@@ -7,9 +7,12 @@ struct PatternMatchingKeywordsRule: Rule {
     static let description = RuleDescription(
         identifier: "pattern_matching_keywords",
         name: "Pattern Matching Keywords",
-        description: "Combine multiple pattern matching bindings by moving keywords out of tuples",
+        description: """
+            Combine multiple pattern matching bindings by moving keywords out of tuples
+            and enum associated values
+            """,
         kind: .idiomatic,
-        nonTriggeringExamples: [
+        nonTriggeringExamples: switchWrapped(examples: [
             Example("default"),
             Example("case 1"),
             Example("case bar"),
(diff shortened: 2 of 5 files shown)


GandaLF2006:
@SimplyDanny It would be great if you can review it

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PatternMatchingKeywordsRule.swift
```diff
@@ -7,9 +7,12 @@ struct PatternMatchingKeywordsRule: Rule {
     static let description = RuleDescription(
         identifier: "pattern_matching_keywords",
         name: "Pattern Matching Keywords",
-        description: "Combine multiple pattern matching bindings by moving keywords out of tuples",
+        description: """
+            Combine multiple pattern matching bindings by moving keywords out of tuples
+            and enum associated values
```
SimplyDanny: This is going to be the rule's violation message. Perhaps, it should be specific for both cases then. In turn, the description can be a bit more detailed.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PatternMatchingKeywordsRule.swift
```diff
@@ -22,88 +25,238 @@ struct PatternMatchingKeywordsRule: Rule {
...
+        ]
     )
 
-    private static func wrapInSwitch(_ example: Example) -> Example {
-        example.with(code: """
-            switch foo {
-                \(example.code): break
-            }
-            """)
+    private static func switchWrapped(examples: [Example]) -> [Example] {
```
SimplyDanny: Why the change?

CHANGELOG.md
```diff
@@ -80,6 +80,9 @@
   better CI parser compatibility.  
   [theamodhshetty](https://github.com/theamodhshetty)
   [#6161](https://github.com/realm/SwiftLint/issues/6161)
+* Improve the opt-in `pattern_matching_keywords` rule by extending support
+  beyond `switch case` and refining nested pattern handling.  
+  [GandaLF2006](https://github.com/GandaLF2006)
```
SimplyDanny, line 91:
```suggestion

* Improve the opt-in `pattern_matching_keywords` rule by extending support
  beyond `switch case` and refining nested pattern handling.  
  [GandaLF2006](https://github.com/GandaLF2006)
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PatternMatchingKeywordsRule.swift
```diff
@@ -22,88 +25,238 @@ struct PatternMatchingKeywordsRule: Rule {
...
 }
 
-private enum ArgumentType {
+private enum GroupCategory {
     case binding(specifier: TokenSyntax)
     case reference
-    case constant
+    case neutral
 
     var isReference: Bool {
```
SimplyDanny: I'm not a fan of such properties that check exactly one enum case only. Likewise, callers can just compare the enum values directly.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PatternMatchingKeywordsRule.swift
```diff
@@ -22,88 +25,238 @@ struct PatternMatchingKeywordsRule: Rule {
...
-        case .reference: true
-        default: false
+        case .reference:
+            return true
+        default:
+            return false
+        }
+    }
+
+    var bindingSpecifier: TokenSyntax? {
```
SimplyDanny: Same here. Callers would need to check for `nil` now. So they can as well pattern-match and unwrap `.binding` themselves.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PatternMatchingKeywordsRule.swift
```diff
@@ -22,88 +25,238 @@ struct PatternMatchingKeywordsRule: Rule {
...
+            return nil
         }
     }
 }
 
-private extension ExprSyntax {
-    var categorized: ArgumentType {
-        if let binding = `as`(PatternExprSyntax.self)?.pattern.as(ValueBindingPatternSyntax.self) {
+private enum GroupCategoryResolver {
+    static func category(for expression: ExprSyntax) -> GroupCategory {
```
SimplyDanny: This factory method, could be a `from(expression: ExprSyntax) -> Self` method in `GroupCategory` itself.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PatternMatchingKeywordsRule.swift
```diff
@@ -22,88 +25,238 @@ struct PatternMatchingKeywordsRule: Rule {
...
     }
-}
 
-private extension LabeledExprListSyntax {
-    func flatteningEnumPatterns() -> [LabeledExprSyntax] {
-        flatMap { elem in
-            guard let pattern = elem.expression.as(FunctionCallExprSyntax.self),
-                  pattern.calledExpression.is(MemberAccessExprSyntax.self) else {
-                return [elem]
+    private static func collect(from expression: ExprSyntax, into violations: inout [AbsolutePosition]) {
```
SimplyDanny: I don't see the point in all the `collect` redirects. Just having `appendViolationsIfBindingsCanBeLifted(from:into:)` as `collectViolations(from:)` right as a private method inside of the visitor would lead to a bit more code in its body. But without the redirects, the logic would be clearer.

SimplyDanny (changes requested):
Thanks for the extension of the rule! Logic and test cases look good.

I have a few remarks on code structure, though, that would, in my opinion, make the code implementation a bit easier to follow. Please let me know what you think.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PatternMatchingKeywordsRule.swift
SimplyDanny: But please keep the description as before for one Danger run reporting the new and fixed OSS findings. That makes much easier to see what's really new or what has just changed in terms of violation position or wording.

GandaLF2006: Good point. The rule now provides case-specific violation reasons via `ReasonedRuleViolation.reason`
1. Tuples: "Combine multiple pattern matching bindings by moving keywords out of tuples"
2. Enum associated values: "Combine multiple pattern matching bindings by moving keywords out of enum associated values"

The description is kept as before this PR.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PatternMatchingKeywordsRule.swift
GandaLF2006: No strong reason for the rename, reverted.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PatternMatchingKeywordsRule.swift
GandaLF2006: Agreed, removed. Made GroupCategory Equatable so callers can just use `categories.contains(.reference)` directly.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PatternMatchingKeywordsRule.swift
GandaLF2006: Agreed, removed. Callers now pattern-match on `.binding(specifier)` directly with `compactMap`.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PatternMatchingKeywordsRule.swift
GandaLF2006: Done. Moved the factory method into `GroupCategory.from(expression:)` and removed the `GroupCategoryResolver` type.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PatternMatchingKeywordsRule.swift
GandaLF2006: Agreed

GandaLF2006:
@SimplyDanny 
> Thanks for the extension of the rule! Logic and test cases look good.
> 
> I have a few remarks on code structure, though, that would, in my opinion, make the code implementation a bit easier to follow. Please let me know what you think.

Thanks for the review! All remarks addressed, ready for another look.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PatternMatchingKeywordsRule.swift
```diff
@@ -4,12 +4,18 @@ import SwiftSyntax
 struct PatternMatchingKeywordsRule: Rule {
     var configuration = SeverityConfiguration<Self>(.warning)
 
+    enum Reason {
+        static let tuples = "Combine multiple pattern matching bindings by moving keywords out of tuples"
+        static let enumAssociatedValues =
+            "Combine multiple pattern matching bindings by moving keywords out of enum associated values"
+    }
```
SimplyDanny, line 14: Let's have the same (previous reason) for one Danger run so that it doesn't report all the changes in wording only. After assessing them, we can just revert/remove this commit:

```suggestion
    enum Reason {
        static let tuples = "Combine multiple pattern matching bindings by moving keywords out of tuples"
        static let enumAssociatedValues =
            "Combine multiple pattern matching bindings by moving keywords out of tuples"
    }
```

SimplyDanny:
I had a look into a few of the OSS findings. They all seem valid.

@GandaLF2006: Please have a look as well to check whether the changes in behavior match your expectations.

GandaLF2006:
> I had a look into a few of the OSS findings. They all seem valid.
> 
> @GandaLF2006: Please have a look as well to check whether the changes in behavior match your expectations.

@SimplyDanny 

I've checked the OSS findings and verified all the behavior changes from this PR.

**OSS findings** — Ran the rule against Nimble, swift-algorithms (Intersperse, Chain, EitherSequence, Reductions), and Alamofire. All flagged violations are true positives. No false positives found across various patterns including wildcards, literal patterns alongside bindings, labeled associated values, and multi-pattern cases.

**PR-claimed coverage** — Verified each new context and pattern type with dedicated test cases:

| Case | Triggering | Non-triggering |
|------|-----------|---------------|
| `if case` | `(let x, let y)` ✅ flagged | `let (x, y)` and `(foo, let x)` ✅ not flagged |
| `guard case` | `(let x, let y)` ✅ flagged | `let (x, y)` ✅ not flagged |
| `while case` | `(let x, let y)` ✅ flagged | `let (x, y)` ✅ not flagged |
| `for case` | `.foo(let x, let y)` ✅ flagged | `let .foo(x, y)` and `.foo(existingValue, let x)` ✅ not flagged |
| `catch` | `Pattern.error(let x, let y)` and `(let x, let y)` ✅ flagged | `let Pattern.error(x, y)` and `(foo, let x)` ✅ not flagged |
| Nested tuples | `((let x, let y), z)` ✅ flagged | `((let x, y), z)` ✅ not flagged |
| Nested enum assoc values | `.foo(.bar(let x), .bar(let y))` ✅ flagged | `.foo(.bar(let x), .bar(var y))` ✅ not flagged |
| Labeled assoc values | `.foo(bar: let x, baz: let y)` ✅ flagged | `.foo(bar: existingValue, baz: let x)` ✅ not flagged |
| Multi-pattern cases | `.foo(let x, let y), .bar(let x, let y)` ✅ flagged | `let .foo(x, y), let .bar(x, y)` ✅ not flagged |

All behavior changes match expectations.

SimplyDanny (approved):

PR #6535, opened 2026-03-10
Fix false positives in vertical_parameter_alignment_on_call with Unicode

### LINKED ISSUE #6467: False positive in rule `vertical_parameter_alignment_on_call`
version 0.63.0

```
vertical_parameter_alignment_on_call:
    severity: error
```

example to reproduce:
```
    func test() {
        СreateAdUpdateShopHelper.updateShopModels(categoryModels: &categoryModels,
                                                  contactModels: &contactModels, // < - error
                                                  country: response.country!, // < - error
                                                  contact: response.contact, // < - error
                                                  shop: response.shop) // < - error
    }
```

theamodhshetty:
## Summary
- compare argument alignment using character columns instead of raw UTF-8 columns
- add a regression example for a call whose first line contains a multi-byte character before the argument list
- update the changelog

Fixes #6467.

## Testing
- swift test --filter VerticalParameterAlignmentOnCallRuleGeneratedTests
- ./.build/debug/swiftlint lint Source/SwiftLintBuiltInRules/Rules/Style/VerticalParameterAlignmentOnCallRule.swift CHANGELOG.md --strict

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Style/VerticalParameterAlignmentOnCallRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/VerticalParameterAlignmentOnCallRule.swift
@@ -83,6 +83,15 @@ struct VerticalParameterAlignmentOnCallRule: Rule {
                 // completion
             }
             """),
+            Example("""
+            func test() {
+                СreateAdUpdateShopHelper.updateShopModels(categoryModels: &categoryModels,
+                                                          contactModels: &contactModels,
+                                                          country: response.country!,
+                                                          contact: response.contact,
+                                                          shop: response.shop)
+            }
+            """),
         ],
         triggeringExamples: [
             Example("""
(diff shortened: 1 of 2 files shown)


Source/SwiftLintBuiltInRules/Rules/Style/VerticalParameterAlignmentOnCallRule.swift
```diff
@@ -83,6 +83,15 @@ struct VerticalParameterAlignmentOnCallRule: Rule {
                 // completion
             }
             """),
+            Example("""
+            func test() {
+                СreateAdUpdateShopHelper.updateShopModels(categoryModels: &categoryModels,
```
SimplyDanny, line 88: If we do the following, the test case would fail (with adapted parameter indentation):
```suggestion
                СreateAd😵‍💫UpdateShopHelper.updateShopModels(categoryModels: &categoryModels,
```

There might exist code you can reuse. I remember that we have done something similar in other rules already, but can't remember where exactly.

PR #6540, opened 2026-03-14
feat: add --fail-on-unfixable flag for single-pass fix and validation

### LINKED ISSUE #6450: Add --fail-on-unfixable flag for single-pass fix + validation
## Problem

When using `swiftlint --fix`, the command always returns exit code 0, even when unfixable violations remain. This forces users to run SwiftLint twice in CI/pre-commit hooks:

```bash
swiftlint --fix && swiftlint --strict
```

This doubles the execution time for files with violations.

## Proposed Solution

Add a new flag `--fail-on-unfixable` (or `--strict-fix`) that:
1. Performs auto-correction as usual
2. Returns non-zero exit code if any violations couldn't be auto-fixed
3. Optionally outputs the unfixable violations to stderr

## Use Case

Pre-commit hooks and CI pipelines that want to:
- Auto-fix what can be fixed
- Block commits/builds when manual intervention is required
- Do this in a single SwiftLint invocation

## Current Workaround

Using `swiftlint --fix && swiftlint --strict`, but this is suboptimal for large codebases and adds unnecessary latency to developer workflows.

## Example Usage

```bash
# Current (two passes)
swiftlint --fix --config .swiftlint.yml file.swift
swiftlint --strict --config .swiftlint.yml file.swift

# Proposed (single pass)
swiftlint --fix --fail-on-unfixable --config .swiftlint.yml file.swift
```

## Related Issues

- #5387 - Discussion about `--strict --fix` behavior

omar-y-abdi:
Fixes #6450

Problem: `swiftlint --fix` always returns exit code 0, even when unfixable violations
remain. Users in CI/pre-commit hooks must run SwiftLint twice: `swiftlint --fix && swiftlint --strict`,
doubling execution time.

Solution: Add `--fail-on-unfixable` flag that performs autocorrection as usual, then
re-lints to detect remaining violations. If serious violations remain, exits with code 2.
Reuses existing `lintOrAnalyze()` path, so `--strict`/`--lenient` work naturally with the
re-lint pass.

Why: The two-pass approach (fix then lint) reuses existing infrastructure with minimal code
changes. Configuration parsing is cached (O(1) on second pass), and `ignoreCache: true`
ensures stale cache entries from modified files are not used. This is architecturally the
simplest solution that correctly handles all edge cases.

Test:
- `swift build` passes
- `LintOrAnalyzeOptionsTests` passes (0 failures)
- Manual e2e verification:
  - Mixed violations (fixable + unfixable) + `--fail-on-unfixable` → exit 2 → works
  - Same without flag → exit 0 → works
  - Only fixable violations + `--fail-on-unfixable` → exit 0 → works
  - `--fail-on-unfixable` without `--fix` → prints warning → works

Closes #6450


alexey1312 (approved):

swizzlr:
@SimplyDanny hope it's alright to ping you or any other maintainers - this looks good and very useful!

SimplyDanny:
Sorry for the late response. I'm not quite happy with an additional option and would rather prefer a new `--fix-then-lint` option that overrules `--fix`.

The proposed `--fail-on-unfixable` only makes sense with `--fix`. That would need to be validated if we decided for it. One option avoids that. `--fix-then-lint` can go with `--fix`, the first just wins.

What do others think?

SimplyDanny:
Danger doesn't like merge commits. Please rebase your branch onto `main` instead of merging `main` into it.

omar-y-abdi:
Feel free to submit review changes if they are needed in this PR, i'll submit commits with all the changes.

SimplyDanny:
> Feel free to submit review changes if they are needed in this PR, i'll submit commits with all the changes.

Did you read my [first comment](https://github.com/realm/SwiftLint/pull/6540#issuecomment-4147619186)?

PR #6543, opened 2026-03-16
Fix statement_position for guard else without whitespace

### LINKED ISSUE #6153: `statement_position` rule triggers for missing space before `else` in if stmt, but not in guard
### New Issue Checklist

### Bug Description

A clear and concise description of what the bug is. Ideally, provide a small (but compilable) example code snippet that
can be used to reproduce the issue.

```swift
guard foo()else { return } // This should trigger a violation, but doesn't

if foo() {
}else { // This triggers a violation (statement_position)
}
```

Mention the command or other SwiftLint integration method that caused the issue. Include stack traces or command output.

```bash
$ swiftlint test.swift
Linting Swift files at paths test.swift
Linting 'test.swift' (1/1)
/Users/lukas/temp/test.swift:4:1: warning: Statement Position Violation: Else and catch should be on the same line, one space after the previous declaration (statement_position)
```

### Environment

* SwiftLint version: 0.59.1
* Xcode version Xcode 16.4 Build version 16F6
* Installation method used Homebrew
* Configuration file: none (using implicit default config)

theamodhshetty:
## Summary
- detect `guard foo()else` in `statement_position`
- autocorrect it to `guard foo() else`
- add regression examples and a changelog entry

The rule currently only looks for `else` and `catch` after a closing brace, so the `guard` form from #6153 slips through. This keeps the existing brace-based matching in place and adds a small guard-specific matcher that only fires when `else` is tokenized as a keyword.

## Testing
- `swift test --filter StatementPositionRuleGeneratedTests`
- `swift test --filter StatementPositionRuleTests`
- `./.build/debug/swiftlint lint Source/SwiftLintBuiltInRules/Rules/Style/StatementPositionRule.swift CHANGELOG.md --strict`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,11 @@
 
 ### Bug Fixes
 
+* Detect and autocorrect missing whitespace before `else` in `guard`
+  statements for the `statement_position` rule.  
+  [theamodhshetty](https://github.com/theamodhshetty)
+  [#6153](https://github.com/realm/SwiftLint/issues/6153)
+
 * Add an `ignore_attributes` option to `implicit_optional_initialization` so
   wrappers/attributes that require explicit `= nil` can be excluded from
   style checks for both `style: always` and `style: never`.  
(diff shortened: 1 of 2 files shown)


SimplyDanny (approved):
Thank you!

PR #6544, opened 2026-03-17
Fix false positives for unused_enumerated with result member access

### LINKED ISSUE #5881: False positive for  `unused_enumerated`
### New Issue Checklist

### Bug Description

There's a false positive for the rule `unused_enumerated`. The offset given from enumerated is used, just apparently not fast enough.

```swift
// This triggers a unused_enumerated  violation:
let previousValueComponent = "1234"
let currentValueComponent = "1434"

                // Compare previous and current components and find if there are any indexes where:
                // - two characters are digits
                // - these two digits are different
                //
                let firstIndexWithDifferentDigit = zip(previousValueComponent, currentValueComponent)
                    .enumerated()
                    .first {
                        $0.element.0.isNumber &&
                        $0.element.1.isNumber &&
                        $0.element.0 != $0.element.1
                    }?.offset

                return firstIndexWithDifferentDigit
```

Mention the command or other SwiftLint integration method that caused the issue. Include stack traces or command output.

```bash
$ swiftlint lint [--no-cache] [--fix]
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure) 0.57.1
* Xcode version (run `xcodebuild -version` to be sure) 16.0
* Installation method used (Homebrew, CocoaPods, building from source, etc) SPM
* Configuration file:

```yml
# insert yaml contents here
```

theamodhshetty:
## Summary
- avoid flagging `.enumerated()` when a higher-order call uses tuple members like `?.offset` after the closure
- keep tracking `$0.element` / `$0.0` and seed the closure usage from result-member access on the surrounding call
- add regression examples and the required changelog entry

## Testing
- swift test --filter UnusedEnumeratedRuleGeneratedTests
- ./.build/debug/swiftlint lint Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift CHANGELOG.md --strict
- ./.build/debug/swiftlint lint --only-rule unused_enumerated /tmp/unused_enumerated_ok.swift --quiet
- ./.build/debug/swiftlint lint --only-rule unused_enumerated /tmp/unused_enumerated_bad.swift --quiet

Closes #5881

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
@@ -25,6 +25,18 @@ struct UnusedEnumeratedRule: Rule {
             Example("list.enumerated().map { ($0.offset, $0.element) }"),
             Example("list.enumerated().map { ($0.0, $0.1) }"),
             Example("""
+            list.enumerated().first {
+                $0.element.0.isNumber &&
+                $0.element.1.isNumber &&
+                $0.element.0 != $0.element.1
+            }?.offset
+            """),
+            Example("""
+            list.enumerated().max {
+                $0.element < $1.element
+            }?.offset
+            """),
+            Example("""
             list.enumerated().map {
                 $1.enumerated().forEach { print($0, $1) }
                 return $0
@@ -95,14 +107,21 @@ private extension UnusedEnumeratedRule {
         var zeroPosition: AbsolutePosition?
         var onePosition: AbsolutePosition?
 
-        init(enumeratedPosition: AbsolutePosition? = nil) {
+        init(
+            enumeratedPosition: AbsolutePosition? = nil,
+            usesZeroResultMember: Bool = false,
+            usesOneResultMember: Bool = false
+        ) {
             self.enumeratedPosition = enumeratedPosition
+            self.zeroPosition = usesZeroResultMember ? enumeratedPosition : nil
+            self.onePosition = usesOneResultMember ? enumeratedPosition : nil
         }
     }
 
     final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
         private var nextClosureId: SyntaxIdentifier?
         private var lastEnumeratedPosition: AbsolutePosition?
+        private var lastEnumeratedResultMemberUsage = (zero: false, one: false)
         private var closures = Stack<Closure>()
 
         override func visitPost(_ node: ForStmtSyntax) {
@@ -159,6 +178,9 @@ private extension UnusedEnumeratedRule {
             } else {
                 nextClosureId = trailingClosure.id
                 lastEnumeratedPosition = node.enumeratedPosition
+                lastEnumeratedResultMemberUsage =
+                    parent.parent?.as(FunctionCallExprSyntax.self)?
+                    .usedEnumeratedResultMembers ?? (false, false)
             }
 
             return .visitChildren
@@ -166,9 +188,14 @@ private extension UnusedEnumeratedRule {
 
         override func visit(_ node: ClosureExprSyntax) -> SyntaxVisitorContinueKind {
             if let nextClosureId, nextClosureId == node.id, let lastEnumeratedPosition {
-                closures.push(Closure(enumeratedPosition: lastEnumeratedPosition))
-                self.nextClosureId = nil
-                self.lastEnumeratedPosition = nil
+                closures.push(Closure(
+                    enumeratedPosition: lastEnumeratedPosition,
+                    usesZeroResultMember: lastEnumeratedResultMemberUsage.zero,
+                    usesOneResultMember: lastEnumeratedResultMemberUsage.one
+                ))
+                nextClosureId = nil
+                lastEnumeratedPosition = nil
+                lastEnumeratedResultMemberUsage = (false, false)
             } else {
                 closures.push(Closure())
             }
@@ -256,6 +283,27 @@ private extension FunctionCallExprSyntax {
            trailingClosure == nil
         && additionalTrailingClosures.isEmpty
         && arguments.isEmpty
+    }
+
+    var usedEnumeratedResultMembers: (zero: Bool, one: Bool) {
+        var current = parent
+
+        while let currentNode = current {
+            if let memberAccess = currentNode.as(MemberAccessExprSyntax.self) {
+                switch memberAccess.declName.baseName.text {
+                case "offset", "0":
+                    return (true, false)
+                case "element", "1":
+                    return (false, true)
+                default:
+                    return (false, false)
+                }
+            }
+
+            current = currentNode.parent
+        }
+
+        return (false, false)
     }
 }
 
--- a/Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
@@ -187,9 +187,11 @@ private extension UnusedEnumeratedRule {
         }
 
         override func visit(_ node: ClosureExprSyntax) -> SyntaxVisitorContinueKind {
-            if let nextClosureId, nextClosureId == node.id, let lastEnumeratedPosition {
+            if let trackedClosureId = nextClosureId,
+               trackedClosureId == node.id,
+               let trackedEnumeratedPosition = lastEnumeratedPosition {
                 closures.push(Closure(
-                    enumeratedPosition: lastEnumeratedPosition,
+                    enumeratedPosition: trackedEnumeratedPosition,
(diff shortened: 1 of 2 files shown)


theamodhshetty:
Pushed a small follow-up in `eb3575a` after the first CI pass.

The original version reset `nextClosureId` / `lastEnumeratedPosition` inside the same `if let` that introduced shorthand bindings with those names, which newer matrix legs were treating as the shadowed locals instead of the stored properties. This update just renames the local bindings so the property resets stay unambiguous across toolchains.

rgoldberg:
Does anything else need to be done to merge this PR?

theamodhshetty:
I don’t think so from my side. The follow-up for the CI/toolchain issue is already in, and the checks are green now. If there’s anything else you want adjusted before merge, I’m happy to make it.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
SimplyDanny:
```suggestion
                  let parentCall = parent?.parent?.as(FunctionCallExprSyntax.self)
                  let trailingClosure = parentCall.trailingClosure
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -159,16 +178,26 @@ private extension UnusedEnumeratedRule {
             } else {
                 nextClosureId = trailingClosure.id
                 lastEnumeratedPosition = node.enumeratedPosition
+                lastEnumeratedResultMemberUsage =
+                    parent.parent?.as(FunctionCallExprSyntax.self)?
+                    .usedEnumeratedResultMembers ?? (false, false)
```
SimplyDanny:
```suggestion
                lastEnumeratedResultMemberUsage = parentCall.usedEnumeratedResultMembers ?? (false, false)
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -257,6 +286,27 @@ private extension FunctionCallExprSyntax {
         && additionalTrailingClosures.isEmpty
         && arguments.isEmpty
     }
+
+    var usedEnumeratedResultMembers: (zero: Bool, one: Bool) {
+        var current = parent
+
+        while let currentNode = current {
+            if let memberAccess = currentNode.as(MemberAccessExprSyntax.self) {
```
SimplyDanny: If there isn't a member access whatsoever, this will always run up to the file's root. Can we find another condition to stop iteration prematurely?

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -257,6 +286,27 @@ private extension FunctionCallExprSyntax {
...
+                switch memberAccess.declName.baseName.text {
+                case "offset", "0":
+                    return (true, false)
+                case "element", "1":
+                    return (false, true)
+                default:
+                    return (false, false)
```
SimplyDanny:
```suggestion
                    break
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -95,14 +107,21 @@ private extension UnusedEnumeratedRule {
...
 
-        init(enumeratedPosition: AbsolutePosition? = nil) {
+        init(
+            enumeratedPosition: AbsolutePosition? = nil,
+            usesZeroResultMember: Bool = false,
+            usesOneResultMember: Bool = false
+        ) {
             self.enumeratedPosition = enumeratedPosition
+            self.zeroPosition = usesZeroResultMember ? enumeratedPosition : nil
+            self.onePosition = usesOneResultMember ? enumeratedPosition : nil
```
SimplyDanny: This logic looks a bit surprising on first sight. Why would `zeroPosition` become `enumeratedPosition`?

I think that actually, `enumeratedPosition` should always be non-optional, so that we only need to track whether "zero" or "one" was found. In turn, `Closure` must be `Closure?` in the stack.

SimplyDanny (changes requested):
Thanks for the contribution!

The idea looks good. I have a few remarks though ...

theamodhshetty:
Pushed a follow-up in `7ba4ed1` that addresses the review points from `SimplyDanny`.

Changes in this pass:
- collapsed the `parentCall` / `trailingClosure` setup into the cleaner form from the suggestions
- reworked the closure tracking so only tracked closures carry an `enumeratedPosition`, instead of using a nullable position on every stack entry
- tightened the result-member ancestor walk so it stops once we leave the optional/member-access chain, while still continuing past unrelated member names until it can see a later `offset` / `element`

Validation rerun:
- `swift test --filter UnusedEnumeratedRuleGeneratedTests`
- `./.build/debug/swiftlint lint Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift --strict`

If you want this split up differently, I can adjust it.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -210,6 +236,22 @@ private extension UnusedEnumeratedRule {
...
+            closures.pop().flatMap(\.self)
+        }
+
+        private func modifyTrackedClosure(_ modifier: (inout Closure) -> Void) {
+            closures.modifyLast {
+                guard var closure = $0 else { return }
+                modifier(&closure)
+                $0 = closure
+            }
+        }
```
SimplyDanny: I'd prefer to inline them as each of them is only used once.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -257,6 +299,35 @@ private extension FunctionCallExprSyntax {
...
+                case "element", "1":
+                    return (false, true)
+                default:
+                    break
+                }
+            }
+
+            guard parent.is(OptionalChainingExprSyntax.self)
+                || parent.is(ForceUnwrapExprSyntax.self)
+                || parent.is(MemberAccessExprSyntax.self)
```
SimplyDanny: Also consider tuple expressions with single elements (i.e. parenthesized expressions).

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -257,6 +299,35 @@ private extension FunctionCallExprSyntax {
         && additionalTrailingClosures.isEmpty
         && arguments.isEmpty
     }
+
+    var usedEnumeratedResultMembers: (zero: Bool, one: Bool) {
```
SimplyDanny, line 298: You may attach this to `ExprSyntax` instead. So we wouldn't need the `while` at all and can rely on recursion with the types listed in 319.

theamodhshetty:
Thanks for the follow-up review. I pushed `1b21940` with those cleanups.

What changed:
- inlined the one-use closure stack helpers
- moved result-member lookup to `ExprSyntax` and made it recurse through member access, optional chaining, force unwraps, and parenthesized single-expression tuples
- added a non-triggering example for a parenthesized `.first { ... }` result followed by `?.offset`

Validation rerun:
- `swift test --filter UnusedEnumeratedRuleGeneratedTests`
- `./.build/debug/swiftlint lint Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift --strict`

Source/SwiftLintBuiltInRules/Rules/Idiomatic/UnusedEnumeratedRule.swift
```diff
@@ -259,6 +294,41 @@ private extension FunctionCallExprSyntax {
...
+            case "element", "1":
+                return (false, true)
+            default:
+                return parent.usedEnumeratedResultMembers
+            }
+        }
+
+        if parent.as(OptionalChainingExprSyntax.self)?.expression.id == id
+            || parent.as(ForceUnwrapExprSyntax.self)?.expression.id == id
+            || parent.as(TupleExprSyntax.self)?.elements.onlyElement?.expression.id == id {
```
SimplyDanny:
```suggestion
           || parent.as(ForceUnwrapExprSyntax.self)?.expression.id == id
           || parent.as(TupleExprSyntax.self)?.elements.onlyElement?.expression.id == id {
```

SimplyDanny (approved):

PR #6545, opened 2026-03-18
Avoid dropping inline comments in control_statement fixes

### LINKED ISSUE #6207: `control_statement` kills inline comments
### New Issue Checklist

### Bug Description

`control_statement` kills inline comments

Code before:

```c
if (abc == 1/* && cdf == 2*/) {
    print("Hello, world!")
}
```

Run SwiftLint
```bash
swiftlint --fix
```

Code afterwards:
```c
if abc == 1 {
    print("Hello, world!")
}
```

I would prefer it to keep comments and thus be:
```c
if abc == 1/* && cdf == 2*/ {
    print("Hello, world!")
}
```

See repo here: https://github.com/herrernst/swiftlint-control_statement-comments

### Environment

* SwiftLint version (run `swiftlint version` to be sure): 0.59.1
* Xcode version (run `xcodebuild -version` to be sure): Xcode 16.4 (Build version 16F6)
* Installation method used (Homebrew, CocoaPods, building from source, etc): SPM
* Configuration file: none (default configuration)

theamodhshetty:
## Summary
- stop `control_statement` from unwrapping parenthesized conditions when the parentheses contain comments
- add a regression example for the inline-comment case from the issue
- add the required changelog entry

## Why
`swiftlint --fix` was deleting inline comments in cases like `if (abc == 1/* ... */) {}`. This change takes the conservative path and skips those corrections instead of applying a destructive rewrite.

## Testing
- swift test --filter ControlStatementRuleGeneratedTests
- ./.build/debug/swiftlint lint Source/SwiftLintBuiltInRules/Rules/Style/ControlStatementRule.swift CHANGELOG.md --strict
- ./.build/debug/swiftlint lint --fix --only-rule control_statement /tmp/control_statement_comment.swift

Closes #6207

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,11 @@
 
 ### Bug Fixes
 
+* Skip `control_statement` corrections for parenthesized conditions that contain
+  comments, avoiding comment loss during `swiftlint --fix`.  
+  [theamodhshetty](https://github.com/theamodhshetty)
+  [#6207](https://github.com/realm/SwiftLint/issues/6207)
+
 * Add an `ignore_attributes` option to `implicit_optional_initialization` so
   wrappers/attributes that require explicit `= nil` can be excluded from
   style checks for both `style: always` and `style: never`.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,8 +42,8 @@
 
 ### Bug Fixes
 
-* Skip `control_statement` corrections for parenthesized conditions that contain
-  comments, avoiding comment loss during `swiftlint --fix`.  
+* Preserve inline comments when `control_statement` removes parentheses during
+  `swiftlint --fix`.
   [theamodhshetty](https://github.com/theamodhshetty)
   [#6207](https://github.com/realm/SwiftLint/issues/6207)
 
--- a/Source/SwiftLintBuiltInRules/Rules/Style/ControlStatementRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/ControlStatementRule.swift
@@ -20,6 +20,11 @@ struct ControlStatementRule: Rule {
             Example("renderGif(data)"),
             Example("guard condition else {}"),
             Example("while condition {}"),
+            Example("""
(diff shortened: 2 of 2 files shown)


SimplyDanny:
Why not remove the parentheses while keeping the comment as is?

theamodhshetty:
Thanks, that makes sense. I pushed `ba5b107` so the correction removes the outer parentheses and keeps inline comments in place instead of skipping those cases.

I also checked the same trivia path for `guard`, `while`, and `switch` locally so the fix does not only work for the added `if` example.

Validation rerun:
- `swift test --filter ControlStatementRuleGeneratedTests`
- `./.build/debug/swiftlint lint Source/SwiftLintBuiltInRules/Rules/Style/ControlStatementRule.swift CHANGELOG.md --strict`

CHANGELOG.md
```diff
@@ -42,6 +42,11 @@
 
 ### Bug Fixes
 
+* Preserve inline comments when `control_statement` removes parentheses during
+  `swiftlint --fix`.
```
SimplyDanny, line 46: Please append two spaces to the end of the description. They are required to enforce a line break in the rendered Markdown.
```suggestion
  `swiftlint --fix`.  
```

Source/SwiftLintBuiltInRules/Rules/Style/ControlStatementRule.swift
```diff
@@ -64,6 +64,15 @@ struct ControlStatementRule: Rule {
...
             Example("↓if (max(a, b) < c) {}"): Example("if max(a, b) < c {}"),
             Example("""
+            ↓if (abc == 1/* && cdf == 2*/) {
+                print("Hello, world!")
+            }
+            """): Example("""
+                if abc == 1/* && cdf == 2*/ {
+                    print("Hello, world!")
+                }
+                """),
```
SimplyDanny, line 74: I wish there were more correction examples. The implementation touches much more constructs than the test examples cover.

SimplyDanny (changes requested):
Let's have a few more test cases. Quite a few cases are currently not covered.

PR #6552, opened 2026-03-25
Fix prefer_self_in_static_references for shadowed nested types

### LINKED ISSUE #5917: False positive for `prefer_self_in_static_references` when an inner struct has the same name as its owner
### New Issue Checklist

### Bug Description

```swift
struct S1 {
    struct S1 {}
    var s = S1()
}
```

`prefer_self_in_static_references` will report a violation at `var s = S1()`, but S1 here refers to the inner struct, not the outer one, so `var s= Self()` would be different, and would fail if the inner S1 has different properties to the outer one.

As a workaround, 

```swift
struct S1 {
    struct S1 {}
    var s = Self.S1()
}
```

Has the same effect, but does not trigger `prefer_self_in_static_references`.

theamodhshetty:
## Summary
- skip `prefer_self_in_static_references` violations when a nested type shadows its enclosing type name
- add a rule example covering the shadowed nested type case
- add the changelog entry

## Testing
- `swift test --filter PreferSelfInStaticReferencesRuleGeneratedTests`
- `./.build/debug/swiftlint lint Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRule.swift Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRuleExamples.swift CHANGELOG.md --strict`
- manual repro with `only_rules: [prefer_self_in_static_references]` against the `struct S1 { struct S1 {}; var s = S1() }` sample

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRule.swift
@@ -37,15 +37,16 @@ private extension PreferSelfInStaticReferencesRule {
 
     final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
         private var parentDeclScopes = Stack<ParentDeclBehavior>()
+        private var shadowingNestedTypeScopes = Stack<Bool>()
         private var variableDeclScopes = Stack<VariableDeclBehavior>()
 
         override func visit(_ node: ActorDeclSyntax) -> SyntaxVisitorContinueKind {
-            parentDeclScopes.push(.likeClass(name: node.name.text))
+            pushParentDeclScope(.likeClass(name: node.name.text), for: node.name.text, memberBlock: node.memberBlock)
             return .skipChildren
         }
 
         override func visitPost(_: ActorDeclSyntax) {
-            parentDeclScopes.pop()
+            popParentDeclScope()
         }
 
         override func visit(_: AttributeSyntax) -> SyntaxVisitorContinueKind {
@@ -56,12 +57,12 @@ private extension PreferSelfInStaticReferencesRule {
         }
 
         override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
-            parentDeclScopes.push(.likeClass(name: node.name.text))
+            pushParentDeclScope(.likeClass(name: node.name.text), for: node.name.text, memberBlock: node.memberBlock)
             return .visitChildren
         }
 
         override func visitPost(_: ClassDeclSyntax) {
-            parentDeclScopes.pop()
+            popParentDeclScope()
         }
 
         override func visit(_: CodeBlockItemListSyntax) -> SyntaxVisitorContinueKind {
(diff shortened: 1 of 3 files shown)


Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRule.swift
```diff
@@ -224,5 +227,44 @@ private extension PreferSelfInStaticReferencesRule {
...
+            for name: String? = nil,
+            memberBlock: MemberBlockSyntax? = nil
+        ) {
+            parentDeclScopes.push(behavior)
+            let hasShadowingNestedType: Bool
+            if let name, let memberBlock {
+                hasShadowingNestedType = containsSameNamedNestedType(named: name, in: memberBlock)
+            } else {
+                hasShadowingNestedType = false
+            }
```
SimplyDanny:
```suggestion
            let hasShadowingNestedType =
                if let name, let memberBlock {
                    containsSameNamedNestedType(named: name, in: memberBlock)
                } else {
                    false
                }
```

Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRule.swift
```diff
@@ -224,5 +227,44 @@ private extension PreferSelfInStaticReferencesRule {
...
+            shadowingNestedTypeScopes.push(hasShadowingNestedType)
```
SimplyDanny: I don't think we need the separate `shadowingNestedTypeScopes` by just adding `.skipReferences` to `parentDeclScopes`.

Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRule.swift
```diff
@@ -224,5 +227,44 @@ private extension PreferSelfInStaticReferencesRule {
                 )
             }
         }
+
+        private func pushParentDeclScope(
+            _ behavior: ParentDeclBehavior,
+            for name: String? = nil,
```
SimplyDanny: Name can be extracted from `behavior`.

Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRule.swift
```diff
@@ -224,5 +227,44 @@ private extension PreferSelfInStaticReferencesRule {
...
+                    return classDecl.name.text == name
+                }
+                if let enumDecl = member.decl.as(EnumDeclSyntax.self) {
+                    return enumDecl.name.text == name
+                }
+                if let structDecl = member.decl.as(StructDeclSyntax.self) {
+                    return structDecl.name.text == name
+                }
+
+                return false
```
SimplyDanny: Should test for type aliases as well. Shorter:

```suggestion
if member.decl.isProtocol((any DeclGroupSyntax).self) || member.decl.is(TypeAliasDeclSyntax.self) {
    return member.decl.asProtocol((any NamedDeclSyntax).self)?.name.text == name
}
return false
```

Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRuleExamples.swift
```diff
@@ -101,6 +101,12 @@ enum PreferSelfInStaticReferencesRuleExamples {
                 }
             }
             """, excludeFromDocumentation: true),
+        Example("""
+            struct S1 {
+                struct S1 {}
+                var s = S1()
```
SimplyDanny, line 107: Also add
```swift
struct S {
  var s = S()
  struct S {}
}
```

SimplyDanny (changes requested):
Good idea! Thank you.

theamodhshetty:
Thanks for the review. I pushed `fc95301` with the requested cleanup.

What changed:
- removed the separate shadowing stack and folded the skip behavior into `parentDeclScopes`
- derived the checked name from `ParentDeclBehavior`
- included type aliases in the nested-name check
- added the extra non-triggering case where the reference appears before the nested type declaration

Validation rerun:
- `swift test --filter PreferSelfInStaticReferencesRuleGeneratedTests`
- `./.build/debug/swiftlint lint Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRule.swift Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRuleExamples.swift --strict`

SimplyDanny (approved):

PR #6568, opened 2026-03-30
Adds Variable Shadowing Rule

### LINKED ISSUE #6228: Warning for local variable shadowing
### New Issue Checklist

Variable shadowing is an anti-pattern because it makes code harder to understand. It is a compiler error in many languages. I think it would be great to warn against it in SwiftLint. Here is an example:

```
var b: String? = "thank"

test(a:"you")

func test(a: String?) {
    if let b = a {
        print("b:\(b)")
        let b = "wow"
        print("b:\(b)")
    }
}
```

- The line `if let b = a {` shadows the above variable b
- The line `let b = "wow"` redeclares b as a new variable

In my opinion both of these should be covered by SwiftLint.

Image

nadeemnali:
## Summary
Adds a new `variable_shadowing` rule that detects when a variable declaration shadows an identifier from an outer scope.

## Changes
- **New Rule:** `VariableShadowingRule.swift` - flags variable declarations that shadow outer scope identifiers
- **Registration:** Added to `BuiltInRules.swift`
- **Tests:** Added generated tests in `GeneratedTests_10.swift`
- **Configuration:** Updated `default_rule_configurations.yml`
- **Documentation:** Updated `CHANGELOG.md`

## Example
```swift
var outer: String = "hello"
func test() {
    let outer = "world"  // ❌ Violates rule
}
```

Resolves #6228


Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
```diff
@@ -0,0 +1,123 @@
...
+import SwiftSyntax
+
+@SwiftSyntaxRule
+struct VariableShadowingRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "variable_shadowing",
+        name: "Variable Shadowing",
+        description: "Prefer not to shadow variables declared in outer scopes",
```
SimplyDanny: Keep it simpler and more direct:
```suggestion
        description: "Do not shadow variables declared in outer scopes",
```

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
```diff
@@ -0,0 +1,123 @@
...
+        ],
+        triggeringExamples: [
+            Example("""
+            var outer: String = "hello"
+            func test() {
+                let ↓outer = "world"
+                print(outer)
+            }
+            """),
+        ]
```
SimplyDanny, line 263: There should be a few more example involving properties, local variables, parameters, functions, tuples, optional bindings, ...

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
```diff
@@ -0,0 +1,123 @@
...
+    )
+}
+
+private extension VariableShadowingRule {
+    final class Visitor: DeclaredIdentifiersTrackingVisitor<ConfigurationType> {
+        override func visit(_ node: VariableDeclSyntax) -> SyntaxVisitorContinueKind {
+            // Early exit for member blocks (class/struct properties)
+            if node.parent?.is(MemberBlockItemSyntax.self) != true {
```
SimplyDanny:
```suggestion
            if node.parent?.is(MemberBlockItemSyntax.self) == false {
```

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
```diff
@@ -0,0 +1,123 @@
...
+            if let valueBinding = pattern.as(ValueBindingPatternSyntax.self) {
+                checkForShadowing(in: valueBinding.pattern)
+                return
+            }
+
+            // Other pattern kinds are not relevant for shadowing checks here; no action needed.
+        }
+
+        private func isShadowingOuterScope(_ identifier: String) -> Bool {
+            guard !scope.isEmpty, scope.count > 1 else { return false }
```
SimplyDanny:
```suggestion
            guard scope.count > 1 else { return false }
```

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
```diff
@@ -0,0 +1,123 @@
...
+
+            // Use early exit and lazy evaluation for better performance
+            for scopeDeclarations in scope.dropLast() where
+                scopeDeclarations.lazy.contains(where: { $0.declares(id: identifier) }) {
```
SimplyDanny: `lazy` doesn't have an effect for just a single operation:
```suggestion
                scopeDeclarations.contains(where: { $0.declares(id: identifier) }) {
```

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        private func checkForShadowing(in pattern: PatternSyntax) {
+            // Handle direct identifier patterns
+            if let identifier = pattern.as(IdentifierPatternSyntax.self) {
+                let identifierText = identifier.identifier.text
+                if isShadowingOuterScope(identifierText) {
+                    violations.append(identifier.identifier.positionAfterSkippingLeadingTrivia)
+                }
+                return
```
SimplyDanny: Can use `else if` to get away with the `return`s.

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
```diff
@@ -0,0 +1,123 @@
...
+            // Recurse into tuple patterns: e.g., (a, b)
+            if let tuple = pattern.as(TuplePatternSyntax.self) {
+                tuple.elements.forEach { element in
+                    checkForShadowing(in: element.pattern)
+                }
+                return
+            }
+
+            // Recurse into value binding patterns: e.g., `let a`, `var (a, b)`
+            if let valueBinding = pattern.as(ValueBindingPatternSyntax.self) {
```
SimplyDanny: What is a value binding anyway? Do you mean optional bindings? In any case, there should be an example for both.

SimplyDanny (changes requested):
Nice one! Glad to see that `DeclaredIdentifiersTrackingVisitor` is used for more advanced rules.

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
nadeemnali: Yes it is related to optional bindings, I will add examples for it

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
nadeemnali: Examples added for all the use cases

.swiftlint.yml
```diff
@@ -51,6 +51,8 @@ disabled_rules:
   - trailing_closure
   - type_contents_order
   - vertical_whitespace_between_cases
+  - variable_shadowing
+
```
SimplyDanny, line 55:
```suggestion
  - variable_shadowing
```

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
SimplyDanny: What about:

```swift
var a = 1
if let a = b {}
```

```swift
var i = 1
for i in list {}
```

```swift
struct S {
  var a = 1
  var b: Int {
    let a = 2
    return a
  }
}
```

```swift
var a = 1
func a(a: Int) {} // Should we trigger for the function name and/or the parameter (optionally)?
```

nadeemnali: Thanks for highlighting, i have added additional triggering examples, expanding the triggering examples

nadeemnali:
```
var a = 1
func a(a: Int) {} // Should we trigger for the function name and/or the parameter (optionally)?

```
Triggering for parameters is optional, but the current behavior (not triggering) is standard and user-friendly. If you want to add an option to control this, I can add a configuration flag (e.g., ignoreParameters: true/false).

SimplyDanny: > Thanks for highlighting, i have added additional triggering examples, expanding the triggering examples

Sure? I can't find them.

> Triggering for parameters is optional, but the current behavior (not triggering) is standard and user-friendly. If you want to add an option to control this, I can add a configuration flag (e.g., ignoreParameters: true/false).

I'd prefer an option for that as I think some would like that. However, I leave it open to you to implement it or ignore it for now. In the latter case, this can be a follow-up.

nadeemnali: > > Thanks for highlighting, i have added additional triggering examples, expanding the triggering examples
> 
> Sure? I can't find them.
> 

Apologies, didn't push it yet.

> > Triggering for parameters is optional, but the current behavior (not triggering) is standard and user-friendly. If you want to add an option to control this, I can add a configuration flag (e.g., ignoreParameters: true/false).
> 
> I'd prefer an option for that as I think some would like that. However, I leave it open to you to implement it or ignore it for now. In the latter case, this can be a follow-up.

Yes, i can look into this, will tag you once all the changes are pushed

nadeemnali: Hi @SimplyDanny, I have updated the rules with configuration and also added relevent examples, so all changes are pushed now

Tests/IntegrationTests/Resources/default_rule_configurations.yml
```diff
@@ -1383,6 +1383,12 @@ valid_ibinspectable:
   meta:
     opt-in: false
     correctable: false
+variable_shadowing:
+  severity: warning
+  ignore_parameters: true
+  meta:
+    opt-in: false
```
SimplyDanny: Rule should be opt-in.

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
SimplyDanny: I'm still missing
```swift
struct S {
  var a = 1
  var b: Int {
    let ↓a = 2
    return a
  }
}
```

And like `if` expressions, we can also have `while let a = ... {}`
and `guard let a = ... else {}`.

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
```diff
@@ -0,0 +1,313 @@
...
+        ]
+    )
+}
+
+private extension VariableShadowingRule {
+    final class Visitor: DeclaredIdentifiersTrackingVisitor<VariableShadowingConfiguration> {
+        override func visit(_ node: VariableDeclSyntax) -> SyntaxVisitorContinueKind {
+            if node.parent?.is(MemberBlockItemSyntax.self) == false {
+                node.bindings.forEach { binding in
+                    checkForShadowing(in: binding.pattern)
```
SimplyDanny, line 272: No test case fails if `checkForBindingShadowing` is used here. Is `checkForShadowing` really necessary? I actually doubt it. If it is, we need an example.

Tests/BuiltInRulesTests/VariableShadowingRuleTests.swift
```diff
@@ -0,0 +1,15 @@
+@testable import SwiftLintBuiltInRules
+import TestHelpers
+import XCTest
+
+final class VariableShadowingRuleTests: SwiftLintTestCase {
```
SimplyDanny: This test isn't needed. You already check the option in the rule's examples.

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
```diff
@@ -0,0 +1,313 @@
...
+            for scopeDeclarations in scope.dropLast() where
+                scopeDeclarations.contains(where: { $0.declares(id: identifier) }) {
+                return true
+            }
+            return false
+        }
+
+        /// Checks all scope levels including the current one. Used for parameter checking
+        /// since parameters are declared into a child scope, not the current one.
+        private func isShadowingAnyScope(_ identifier: String) -> Bool {
```
SimplyDanny: We might not need this anymore once #6589 is merged.

SimplyDanny (changes requested):

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
nadeemnali: Yes, checkForShadowing (using isShadowingOuterScope / scope.dropLast()) is still necessary for 
 VariableDeclSyntax. The distinction is:
 
 
  - if let / for / while let / guard let bindings are added to a child scope by super.visit(...), so at the 
point we check, the outer variable is still visible in the current scope → hasSeenDeclaration(for:) 
(checking all scopes) is correct.
  - let/var declarations are added to the current scope by super.visitPost(...). If we used 
hasSeenDeclaration here, we'd also match the identifier that's already been added to the current scope 
earlier in the same block, causing false positives for sequential declarations at the same nesting level.
 
 A concrete case: the DisableAllTests suite constructs code with the same name appearing twice at file scope
 — checkForShadowing correctly ignores this; checkForBindingShadowing would wrongly flag it.

Tests/IntegrationTests/Resources/default_rule_configurations.yml
nadeemnali: Done — changed to opt-in: true in default_rule_configurations.yml and also added @SwiftSyntaxRule(optIn:
 true) to the rule declaration itself.

Tests/BuiltInRulesTests/VariableShadowingRuleTests.swift
nadeemnali: Agreed and removed

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
nadeemnali: Updated to reuse the existing one

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
nadeemnali: Updated now with relevent examples

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
SimplyDanny: So please add an example to prove that. Because just using `checkForBindingShadowing` wouldn't let any test fail.

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
SimplyDanny: I have another case:
```swift
let a: Int?
if let a {}
```
should not trigger the rule as this is the idiomatic way to bind an optional variable. This applies also to `guard` and `while` and probably also `let a = a`. However, `let a = b` would be a violation.

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
nadeemnali: Added non-triggering example demonstrating why checkForShadowing is needed:

 do {
     var counter = 0
     var sum = 0      // No shadowing - different name
     counter = 1
 }

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
nadeemnali: Added Idiomatic Optional Binding Detection for shorthand syntax with examples

Source/SwiftLintBuiltInRules/Rules/Lint/VariableShadowingRule.swift
SimplyDanny: Sorry, but all your examples still pass if I replace the only call of `checkForShadowing` ...

Don't get me wrong. I'm not saying, we definitely don't need it. If it was needed, there would have to be an example proving that, though.

nadeemnali: Yes you are technically correct on this, I have highlighted some points below
  
Facts:

   - No current test case fails with the method swap
   - Both methods are functionally equivalent in practice  
   - Swift forbids same-scope variable redeclaration (compiler error), so the subtle timing 
  difference between methods never manifests

  Why it still should exist:

   1. Semantic clarity - explicitly shows intent to check only outer scopes
   2. Design correctness - the distinction documents why optional bindings differ
   3. Future-proof - protects against bugs if scope tracking changes
   4. Code documentation - makes the design intent clear to future maintainers

But having said that, I am fine in unifying both methods. Will update it now

.swiftlint.yml
```diff
@@ -51,6 +51,7 @@ disabled_rules:
   - trailing_closure
   - type_contents_order
   - vertical_whitespace_between_cases
+  - variable_shadowing
```
SimplyDanny:
```suggestion
  - variable_shadowing
  - vertical_whitespace_between_cases
```

SimplyDanny (approved):

PR #6595, opened 2026-04-12
feat: add `redundant_final_actor` rule

### LINKED ISSUE #6407: Rule Request: Redundant `final` for `actor`s

VDurocher:
## Description

Adds a new `redundant_final_actor` rule that warns when a `final` modifier is applied to an `actor` declaration.

Per [SE-0306](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0306-actors.md#actors-2), actors are reference types without support for inheritance, so they are implicitly `final`. Writing `final actor` is therefore redundant, similar to how `internal` is redundant as an access modifier.

Closes #6407

## Examples

**Triggers:**
```swift
final actor DataStorage {}

final actor Counter {
    var count = 0
}

public final actor Scheduler {}
```

**Does not trigger:**
```swift
actor DataStorage {}

public actor Scheduler {}

final class Foo {}  // still valid for classes
```

## Auto-correction

This rule supports auto-correction: it removes the redundant `final` modifier automatically.

## Notes

- `BuiltInRules.swift` updated manually (run `make register` to verify the generated output matches)
- CHANGELOG entry to be added under `### Enhancements` in `## Main`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
 ### Enhancements
 
+* Add new `redundant_final_actor` rule that warns when the `final` modifier is applied to an `actor` declaration. Since actors are implicitly `final` per [SE-0306](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0306-actors.md), the modifier is always redundant.  
+  [VDurocher](https://github.com/VDurocher)
+  [#6407](https://github.com/realm/SwiftLint/issues/6407)
+
 * Print fixed code read from stdin to stdout.  
   [SimplyDanny](https://github.com/SimplyDanny)
   [#6501](https://github.com/realm/SwiftLint/issues/6501)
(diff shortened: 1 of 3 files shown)


SimplyDanny:
Thank you! Unfortunately, I have to tell you that there is [another PR](https://github.com/realm/swiftlint/pull/6504) already that I just haven't found the time to review. I would like to move that one forward.

PR #6609, opened 2026-04-17
Add `prefer_scaled_to_fit` opt-in rule

### LINKED ISSUE #5713: Rule Request: SwiftUI: Prefer scaledToFit()/scaledToFill() over aspectRatio(contentMode:)
### New Issue Checklist

### New rule request

SwiftUI has `scaledToFit()` `scaledToFill()` which are convenience methods for calling `aspectRatio`  with only a content mode.

> This method is equivalent to calling [aspectRatio(_:contentMode:)](https://developer.apple.com/documentation/swiftui/view/aspectratio(_:contentmode:)-5z8qw) with a nil aspectRatio and a content mode of [ContentMode.fit](https://developer.apple.com/documentation/swiftui/contentmode/fit).
https://developer.apple.com/documentation/swiftui/view/scaledtofit()#

Since Apple considered it a common enough use case to have dedicated view modifiers for it, there should be a rule to prefer using them over the more general purpose view modifier.

## Triggering cases

It just needs to be the versions where they're passed a constant.
```swift
view.aspectRatio(contentMode: .fit)
view.aspectRatio(contentMode: .fill)

aspectRatio(contentMode: .fit)
aspectRatio(contentMode: .fill)
```

## Non-triggering cases

If a ratio is being passed or the content mode is not a constant, it should not trigger
```swift
let ratio = CGSize(width: 1, height: 1)
view.aspectRatio(ratio, contentMode: .fit)
view.aspectRatio(ratio, contentMode: .fill)

let contentMode = ContentMode.fit
view.aspectRatio(contentMode: contentMode)

let shouldFit = true
view.aspectRatio(contentMode: shouldFit ? .fit : .fill)
```

## Configuration
I don't think this needs any configuration.

It's always possible someone else would prefer the `aspectRatio` method over the convenience method. If that occurs the rule could be modified to controlling the preferred method, though it could also be a separate rule as well since it would be checking for the opposite case.

## Opt-in?
Based on the criteria in the readme, this could be enabled by default. Though it depends on what the general consensus is.

DemiDevv:
Closes #5713

Adds a new opt-in rule `prefer_scaled_to_fit` that suggests using
`scaledToFit()` or `scaledToFill()` instead of
`aspectRatio(contentMode:)` when the content mode is a constant
(`.fit` or `.fill`).

As noted in Apple's documentation, `scaledToFit()` is equivalent to
`aspectRatio(nil, contentMode: .fit)`, so the shorter form is
preferred when no custom aspect ratio is provided.


DemiDevv:
@SimplyDanny It seems like some CI checks are stuck in "Waiting for status to be reported" for a while. Is there anything I need to do on my end, or is this expected?

DemiDevv:
Gentle ping @SimplyDanny — the 14 pending checks haven't changed in over a week. Could you take a look when you have a moment? Thanks!

SimplyDanny:
@DemiDevv: I rebased your branch. Checks are running now.

DemiDevv:
@SimplyDanny Gentle ping — this has been ready for review for a few weeks now with all checks passing. Would appreciate a look when you have a moment!

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferScaledToFitRule.swift
```diff
@@ -0,0 +1,70 @@
+import SwiftLintCore
+import SwiftSyntax
+
+@SwiftSyntaxRule(optIn: true)
+struct PreferScaledToFitRule: Rule {
```
SimplyDanny: There are a few `Legacy...` rules already. I wonder whether we should name this one in a similar way. What do you think?

All of them are enabled by default (not opt-in) and support auto-correction using `LegacyFunctionVisitor`.

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferScaledToFitRule.swift
```diff
@@ -0,0 +1,70 @@
...
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "prefer_scaled_to_fit",
+        name: "Prefer Scaled To Fit",
```
SimplyDanny: Title style:
```suggestion
        name: "Prefer Scaled to Fit",
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferScaledToFitRule.swift
```diff
@@ -0,0 +1,70 @@
...
+                return
+            }
+
+            guard functionName == "aspectRatio" else {
+                return
+            }
+
+            guard
+                node.arguments.count == 1,
+                let argument = node.arguments.first,
```
SimplyDanny:
```suggestion
                let argument = node.arguments.onlyElement,
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferScaledToFitRule.swift
```diff
@@ -0,0 +1,70 @@
...
+                argument.label?.text == "contentMode"
+            else {
+                return
+            }
+
+            guard
+                let memberValue = argument.expression.as(MemberAccessExprSyntax.self),
+                let valueName = memberValue.declName.baseName.text as String?,
+                valueName == "fit" || valueName == "fill"
```
SimplyDanny:
```suggestion
                ["fit, "fill"].contains(memberValue.declName.baseName.text)
```

Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferScaledToFitRule.swift
```diff
@@ -0,0 +1,70 @@
...
+            guard
+                node.arguments.count == 1,
+                let argument = node.arguments.first,
+                argument.label?.text == "contentMode"
+            else {
+                return
+            }
+
+            guard
+                let memberValue = argument.expression.as(MemberAccessExprSyntax.self),
```
SimplyDanny: Should we also check that `base` is `nil` or `ContentMode`.

SimplyDanny (changes requested):
Thanks for the work and sorry for the delay! Please see my findings (and questions).

DemiDevv:
@SimplyDanny 
  Thanks for the detailed review. I went through the points and I agree with them.

  The rule name `legacy_swiftui_aspect_ratio` better describes the actual intent, so I renamed it and updated the implementation around that. I also made
  it correctable/default-enabled and added corrections for the safe cases:
  `aspectRatio(contentMode: .fit)` -> `scaledToFit()`
  `aspectRatio(contentMode: .fill)` -> `scaledToFill()`

  I kept the matching intentionally narrow to `.fit`, `.fill`, `ContentMode.fit`, and `ContentMode.fill` so that unrelated enum cases are not flagged.

  I also rebased on the latest `upstream/main`, resolved the generated test conflicts, regenerated the rule tests, and force-pushed with `--force-with-
  lease`. CI is currently awaiting maintainer approval.

SimplyDanny (approved):
Thanks @DemiDevv!

PR #6610, opened 2026-04-18
Fix `explicit_self` false positives around string interpolation

### LINKED ISSUE #6611: Bug: `explicit_self` false positive with string interpolation
### New Issue Checklist

### Bug Description

The [`explicit_self`](https://realm.github.io/SwiftLint/explicit_self.html) **analyzer** rule reports a violation (“Instance variables and functions should be explicitly accessed with `self.`”) even when the member is **already** accessed with `self.` inside string interpolation `\(...)` (and the same pattern with raw-string interpolation `\#(...)`).

**Expected:** No violation; `self` is explicit.

**Actual:** `explicit_self` reports a violation (false positive).

`explicit_self` only runs with **`swiftlint analyze`** (compiler invocations required), not `swiftlint lint`.

Minimal example:

```swift
class StringInterpolation {
    let foo = "foo"

    var description: String {
        return "StringInterpolation{foo: \(self.foo)}"
    }

    var rawStringLiteral: String {
        return #"StringInterpolation{foo: \#(self.foo)} using "raw" string literals."#
    }
}
```

Full repro (Swift package, `xcodebuild` log + analyze): https://github.com/jffmrk/swiftlint-explicit-self-string-interpolation

Example command output:

```text
$ swiftlint analyze --compiler-log-path xcodebuild.log
Analyzing 'StringInterpolation.swift' (1/1)
/Users/jmark/src/github/swiftlint-explicit-self-string-interpolation/Sources/ExplicitSelfStringInterpolationBug/StringInterpolation.swift:5:43: error: Explicit Self Violation: Instance variables and functions should be explicitly accessed with 'self.' (explicit_self)
/Users/jmark/src/github/swiftlint-explicit-self-string-interpolation/Sources/ExplicitSelfStringInterpolationBug/StringInterpolation.swift:9:45: error: Explicit Self Violation: Instance variables and functions should be explicitly accessed with 'self.' (explicit_self)
```

(Line/column may vary slightly by Swift/Xcode version.)

### Environment

* **SwiftLint version (run `swiftlint version` to be sure):** 0.63.2
* **Xcode version (run `xcodebuild -version` to be sure):** Xcode 26.4 (Build version 17E192)
* **Installation method used (Homebrew, CocoaPods, building from source, etc):** Homebrew (`/opt/homebrew/bin/swiftlint`)
* **Configuration file:**

```yml
analyzer_rules:
  - explicit_self

explicit_self:
  severity: error
```

**Are you using [nested configurations](https://github.com/realm/SwiftLint#nested-configurations)?** No — single `.swiftlint.yml` at the repository root.

jffmrk:
## Summary
SourceKit can report misleading instance-member refs for `explicit_self` when code uses string literals and `\(...)` interpolation (e.g. literal text that matches a property name, or bogus offsets at the interpolation open paren).

This PR filters those cases using SwiftSyntax static string segments, skips offsets at `\`(`, and verifies explicit `self.<member>` spans in source text.

Fixes #6611 

## Testing
- `swift test --filter ExplicitSelfRuleGeneratedTests`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,6 +132,10 @@
   [theamodhshetty](https://github.com/theamodhshetty)
   [#5917](https://github.com/realm/SwiftLint/issues/5917)
 
+* Fix `explicit_self` false positives around string interpolation.  
+  [jffmrk](https://github.com/jffmrk)
+  [#6611](https://github.com/realm/SwiftLint/issues/6611)
+
 ## 0.63.2: High-Speed Extraction
 
 ### Breaking
--- a/Source/SwiftLintBuiltInRules/Rules/Style/ExplicitSelfRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/ExplicitSelfRule.swift
@@ -1,5 +1,6 @@
 import Foundation
 import SourceKittenFramework
+import SwiftSyntax
 
 struct ExplicitSelfRule: CorrectableRule, AnalyzerRule {
     var configuration = SeverityConfiguration<Self>(.warning)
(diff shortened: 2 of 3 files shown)


SimplyDanny:
Hey @jffmrk, thanks for bringing this up and sorry for the late response.

While reviewing your code, I thought about simplifying the whole rule a lot by making use of your idea to let SwiftSyntax help to find correct positions more reliably. This is what led to #6662. It uses SwiftSyntax to find symbols in the first place and only asks SourceKit about their origin. Doing so, a lot of code could be removed making the rule quite a bit easier to comprehend.

I hope you don't mind if we move forward with #6662. I have retained you as a (co-)author in the commit and the changelog entry.

jffmrk:
Cool. Thanks! I'll try it out on our code tomorrow. This was Cursor helping me out.

SimplyDanny:
> Cool. Thanks! I'll try it out on our code tomorrow. This was Cursor helping me out.

Let me know about the results. I'll cut a new release shortly after.

PR #6614, opened 2026-04-20
Add --strict-baseline flag to fail lint when baselined violations are fixed

### LINKED ISSUE #6511: Linting should fail when baseline violations are fixed
### New Issue Checklist

As with other linter baseline implementations, a fix to a violation that was previously baselined should cause a lint failure that encourages the user to update the baseline. For backwards compatibility, this should probably be a flag, e.g. `--strict-baseline` or something (feel free to bikeshed).

swizzlr:
## Summary

Closes #6511.

Adds a new `--strict-baseline` command-line flag and matching
`strict_baseline` configuration key. When enabled, SwiftLint compares
the stored baseline against violations detected in the current run and
reports an error for each baselined violation that is no longer
detected, encouraging the baseline to be regenerated once violations
have been fixed. This brings SwiftLint's baseline behaviour in line
with other linters.

Behaviour is opt-in and off by default, preserving backwards
compatibility.

## Details

* New `--strict-baseline` flag on `swiftlint lint` / `swiftlint analyze`
  via `LintOrAnalyzeArguments`.
* New `strict_baseline` YAML key wired through `Configuration`,
  `Configuration+Parsing`, and `Configuration+Merging`.
* `LintOrAnalyzeCommand` now retains the loaded baseline and, after
  collection, diffs it against the current run via the existing
  `Baseline.compare` primitive. Each fixed baselined violation is
  surfaced as a `fixed_baseline_violation` error at the original
  location.
* README and CHANGELOG updated.

## Test plan

- [x] `swift test --filter BaselineTests`
- [x] `swift test --filter LintOrAnalyzeOptions`
- [x] Manual end-to-end: recorded a baseline against a file with
      trailing-semicolon warnings, then removed the semicolons and
      re-ran with `--baseline baseline.json --strict-baseline`. Exit
      code `2` and per-violation errors as expected. Without
      `--strict-baseline`, exit code `0`.
- [x] `swift build` clean.

### DIFF
--- a/Source/SwiftLintFramework/LintOrAnalyzeCommand.swift
+++ b/Source/SwiftLintFramework/LintOrAnalyzeCommand.swift
@@ -44,6 +44,7 @@ package struct LintOrAnalyzeOptions {
     let reporter: String?
     let baseline: String?
     let writeBaseline: String?
+    let strictBaseline: Bool
     let workingDirectory: String?
     let quiet: Bool
     let output: String?
@@ -73,6 +74,7 @@ package struct LintOrAnalyzeOptions {
                  reporter: String?,
                  baseline: String?,
                  writeBaseline: String?,
+                 strictBaseline: Bool,
                  workingDirectory: String?,
                  quiet: Bool,
                  output: String?,
@@ -101,6 +103,7 @@ package struct LintOrAnalyzeOptions {
         self.reporter = reporter
         self.baseline = baseline
         self.writeBaseline = writeBaseline
+        self.strictBaseline = strictBaseline
         self.workingDirectory = workingDirectory
         self.quiet = quiet
         self.output = output
@@ -150,6 +153,7 @@ package struct LintOrAnalyzeCommand {
         if let baselineOutputPath = options.writeBaseline ?? builder.configuration.writeBaseline {
             try Baseline(violations: builder.unfilteredViolations).write(toPath: baselineOutputPath)
         }
+        appendFixedBaselineViolations(builder: builder)
         let numberOfSeriousViolations = try Signposts.record(name: "LintOrAnalyzeCommand.PostProcessViolations") {
             try postProcessViolations(files: files, builder: builder)
         }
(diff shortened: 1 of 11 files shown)


Source/SwiftLintFramework/LintOrAnalyzeCommand.swift
```diff
@@ -266,6 +287,24 @@ package struct LintOrAnalyzeCommand {
...
+        )
+        return StyleViolation(
+            ruleDescription: description,
+            severity: .error,
+            location: violation.location,
+            reason: """
+                Violation previously recorded in the baseline for '\(violation.ruleIdentifier)' is no longer \
+                detected. Regenerate the baseline to acknowledge this fix
+                """
+        )
```
swizzlr, line 305: @SimplyDanny is there a more idomatic way to say "baseline is no longer valid"?

mildm8nnered: baseline is "stale" perhaps, but that still doesn't feel quite right.

Source/swiftlint/Common/LintOrAnalyzeArguments.swift
```diff
@@ -46,6 +46,12 @@ struct LintOrAnalyzeArguments: ParsableArguments {
     var baseline: String?
     @Option(help: "The path to save detected violations to as a new baseline.")
     var writeBaseline: String?
+    @Flag(help: """
+        Fail linting if violations previously recorded in the baseline are no longer \
+        detected. Encourages keeping the baseline up-to-date by regenerating it \
+        once baselined violations have been fixed.
```
mildm8nnered, line 52: This feels a little verbose. `Fail linting when violations recorded in the baseline are no longer detected.` might enough here (lifted from the `strictBaseline` property doc comment).

Source/SwiftLintFramework/LintOrAnalyzeCommand.swift
```diff
@@ -266,6 +287,24 @@ package struct LintOrAnalyzeCommand {
         return numberOfWarningViolations >= warningThreshold
     }
 
+    private static func createFixedBaselineViolation(fromBaseline violation: StyleViolation) -> StyleViolation {
```
mildm8nnered, line 290: So reporting these as violations is very convenient, but the `fixed_baseline_violation` "pseudo-rule" is not really a rule - for example, I can't `// swiftlint:disable fixed_baseline_violation` at the top of a file (and `fixed_baseline_violation` isn't easily matched up to `strict_baseline`).

I wonder if these should be reported as `Issue`'s instead, something like

`Issue.fixedBaselineViolation(violation).print()`

ggrell:
@swizzlr Any chance we can get this merged? Will be a great & much needed feature! 🙏

PR #6615, opened 2026-04-21
Add --pretty-baseline flag for reviewable baseline diffs

swizzlr:
## Summary

Adds an **opt-in** `--pretty-baseline` flag (plus matching
`pretty_baseline` configuration key) that writes baseline files as
pretty-printed JSON with sorted keys and unescaped forward slashes.
Violations have always been sorted by file and location, so the
complete opt-in output is deterministic across runs and produces
stable, human-reviewable diffs whenever a baseline is regenerated.

**Default behaviour is unchanged** — existing baselines remain
byte-identical, so this change is fully backwards compatible.

Motivation: the default baseline is emitted as a single minified line
with escaped slashes and (in compact mode) non-deterministic key
order. That makes it nearly impossible to review updates in a PR, and
trivial changes can reorder unrelated entries. With `--pretty-baseline`,
regenerating a baseline after fixing a few violations produces a
minimal, readable diff.

## Details

* New `--pretty-baseline` flag on `swiftlint lint` / `swiftlint analyze`
  via `LintOrAnalyzeArguments`, threaded through `LintOrAnalyzeOptions`
  and `LintOrAnalyzeCommand` to `Baseline.write(toPath:pretty:)`.
* New `pretty_baseline` YAML key wired through `Configuration`,
  `Configuration+Parsing`, and `Configuration+Merging`.
* `Baseline.write(toPath:pretty:)` defaults to `pretty: false`, so the
  existing default output is preserved exactly.
* `--pretty-baseline` without `--write-baseline` is rejected at
  argument-parse time via a `validate()` method on
  `LintOrAnalyzeArguments`.
* Snapshot tests on `BaselineTests` assert both the default (compact,
  one-line) and the opt-in (pretty, multiline, sorted) output using the
  existing `withExampleFileCreated` helper.
* README and CHANGELOG updated.

## Test plan

- [x] `swift test --filter BaselineTests` — all 8 cases pass, including
      a new snapshot test asserting the exact pretty-printed output and
      a companion test asserting the default remains compact.
- [x] `swift test --filter LintOrAnalyzeOptionsTests`.
- [x] Manual: recorded a baseline with `--write-baseline` alone
      (single-line output preserved) and with
      `--write-baseline X.json --pretty-baseline` (pretty, sorted,
      unescaped slashes).
- [x] Manual: `swiftlint --pretty-baseline` without `--write-baseline`
      prints a validation error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,13 @@
 
 ### Enhancements
 
+* Add `--pretty-baseline` flag (and matching `pretty_baseline`
+  configuration key) that writes baseline files as pretty-printed JSON
+  with sorted keys, producing stable, reviewable diffs when the baseline
+  is regenerated. Default behaviour is unchanged so existing baselines
+  remain byte-identical.  
+  [swizzlr](https://github.com/swizzlr)
+
 * Print fixed code read from stdin to stdout.  
   [SimplyDanny](https://github.com/SimplyDanny)
   [#6501](https://github.com/realm/SwiftLint/issues/6501)
--- a/README.md
+++ b/README.md
@@ -742,6 +742,10 @@ baseline: Baseline.json
 # The path to save detected violations to as a new baseline.
 write_baseline: Baseline.json
 
+# If true, the baseline produced by write_baseline is pretty-printed with
+# sorted keys, producing stable diffs when regenerated. Defaults to false.
+pretty_baseline: false
+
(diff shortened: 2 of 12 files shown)


mildm8nnered:
This seems quite nice - the original format was optimized for size over readability - having the option to choose is good.

Is it possible to add a `pretty` subcommand to the `baseline` command as well, which could do this conversion?

PR #6639, opened 2026-05-15
Surface violations of `prefer_self_in_static_references` inside extensions

### LINKED ISSUE #3993: False negatives in prefer_self_in_static_references since 0.46.2
### New Issue Checklist

### Describe the bug

The `prefer_self_in_static_references` is not reporting some issues since in from version 0.46.2 onward. Our CI was still on 0.46.1 and caught many more issues with the older version

Seems like the same issue as
- https://github.com/realm/SwiftLint/issues/3875

```bash
$ swiftlint lint
Linting 'foo.swift' (1/1)

# After downgrading to 0.46.1
$ swiftlint lint
Linting 'foo.swift' (1/1)
/Users/neo/Desktop/swiftlint/foo.swift:6:27: warning: Prefer Self in Static References Violation: Static references should be prefixed by `Self` instead of the class name. (prefer_self_in_static_references)
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.47.1

```yml
opt_in_rules:
```

* Which Xcode version are you using (check `xcodebuild -version`)? Xcode 13.4.1 Build version 13F100

```swift
class Foo {
  static let value = 1
}

extension Foo {
  static let otherValue = Foo.value
}
```

itsybitsybootsy:
## Problem

`prefer_self_in_static_references` currently pushes `.skipReferences` for every `ExtensionDeclSyntax`, which silences the rule inside the entire extension body. The rule misses violations like:

```swift
class Foo {
    static let value = 1
}

extension Foo {
    func helper() -> Int { Foo.value }      // not flagged, should be Self.value
    var alias: Int { Foo.value }            // not flagged, should be Self.value
}
```

Per @SimplyDanny's [comment on #3993](https://github.com/realm/SwiftLint/issues/3993#issuecomment-1937562648):

> Allowing the rule for extensions by treating the type as it was a class, would be achievable as a first step without much effort. This might be a "good first issue".

## Fix

When the extended type is a plain identifier (`extension Foo`) or a member type (`extension Foo.Bar`), push `.likeClass(name: extendedTypeName)` instead of `.skipReferences`. Complex extended-type shapes (generic specialisations, opaque types, function types) continue to fall through to the previous `.skipReferences` behaviour.

The class scope is the most conservative reuse of existing logic — it already correctly handles method bodies, computed properties, and closure invocations on the right-hand side, while still skipping the things class mode skips (stored-property initializers, function signatures) to avoid false positives. This matches the maintainer's "first step" framing: the reporter's exact case (`extension Foo { static let otherValue = Foo.value }` — a stored-property initializer) remains exempt under this change, matching how the rule already treats top-level class declarations.

The `IdentifierTypeSyntax` visitor also gains a guard so that the extension declaration header itself (the extended-type identifier) is excluded from violations — without it, the new scope would rewrite `extension Foo` to `extension Self`.

## Tests

- New non-triggering example: extension of an unrelated type with a reference to a different type (confirms parent-name matching is correct).
- New triggering examples: extension of a class with violations in method body + computed property, and extension of a nested type (`Outer.Inner`) confirming `MemberTypeSyntax` handling.
- New correction: extension method body that auto-fixes `↓C.i` → `Self.i`.
- Existing non-triggering example at the top of the file (`extension T { static let j = S.i + T.i; static let k = { T.j }() }`) continues to pass: class-mode `MemberBlockSyntax` already skips stored-property initializer expressions, so neither `T.i` nor the closure body's `T.j` is reachable.

## Self-corrections

The integration test (`testSwiftLintAutoCorrects`) lints SwiftLint's own source against all rules. With this change, four pre-existing violations in SwiftLint's own code are surfaced and auto-corrected — they're equivalent rewrites of the same shape (`String.operators` → `Self.operators` inside a `private extension String`, etc.) and are bundled here so the integration test passes:

- `Source/SwiftLintBuiltInRules/Rules/Idiomatic/PreferKeyPathRule.swift`: `as ExprSyntax` → `as Self`
- `Source/SwiftLintBuiltInRules/Rules/Performance/EmptyCountRule.swift`: `String.operators` → `Self.operators`
- `Source/SwiftLintBuiltInRules/Rules/Style/ContrastedOpeningBraceRule.swift`: `\IfExprSyntax.elseBody` → `\Self.elseBody`
- `Source/SwiftLintFramework/Reporters/SummaryReporter.swift`: `Int.numberFormatter` → `Self.numberFormatter`

Verified locally with `swift run swiftlint-dev rules register && swift test`. All tests pass.

Resolves #3993.

---

This contribution was developed with the help of Claude Code. The fix, regression tests, edge-case analysis, and the local test suite were reviewed manually before submitting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,11 +12,8 @@
 
 ### Enhancements
 
-* Surface violations of `prefer_self_in_static_references` inside
-  extensions of types whose name is a plain identifier or a member type
-  (e.g. `extension Foo` or `extension Foo.Bar`). Stored-property
-  initializers and function signatures remain exempt — matching the
-  rule's existing behaviour for top-level class declarations.  
+* Treat extensions like classes in the `prefer_self_in_static_references`
+  rule.  
   [itsybitsybootsy](https://github.com/itsybitsybootsy)
   [#3993](https://github.com/realm/SwiftLint/issues/3993)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,14 @@
 
 ### Enhancements
 
+* Surface violations of `prefer_self_in_static_references` inside
+  extensions of types whose name is a plain identifier or a member type
+  (e.g. `extension Foo` or `extension Foo.Bar`). Stored-property
+  initializers and function signatures remain exempt — matching the
+  rule's existing behaviour for top-level class declarations.  
+  [itsybitsybootsy](https://github.com/itsybitsybootsy)
+  [#3993](https://github.com/realm/SwiftLint/issues/3993)
+
 * Print fixed code read from stdin to stdout.  
   [SimplyDanny](https://github.com/SimplyDanny)
   [#6501](https://github.com/realm/SwiftLint/issues/6501)
--- a/Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRule.swift
@@ -137,12 +137,15 @@ private extension PreferSelfInStaticReferencesRule {
                     return .skipChildren
                 }
             }
-            // In a member-type extension (e.g. `extension Foo.Bar`), the only
-            // valid reference to the extended type is the full member path, so
-            // match the access chain structurally rather than a single token.
-            // A path used to call an initializer (`Foo.Bar()`) is left alone.
+            // In a member-type extension (e.g. `extension Foo.Bar`), match the
+            // extended-type access chain structurally; a single token can't
+            // disambiguate `Outer.Inner.x` from `Other.Inner.x`. Skip when the
+            // path is the callee of a call or carries explicit generic
+            // arguments (`Foo.Bar()`, `Foo.Bar<Int>()`), since `Self` doesn't
+            // substitute for those.
             if let components = parentDeclScopes.peek()?.memberTypeComponents,
                node.parent?.is(FunctionCallExprSyntax.self) != true,
+               node.parent?.is(GenericSpecializationExprSyntax.self) != true,
                let tokens = memberAccessChain(node),
                tokens.map(\.text) == components {
                 addMemberTypeViolation(spanning: tokens)
@@ -254,6 +257,12 @@ private extension PreferSelfInStaticReferencesRule {
             }
             // Don't flag the extended type in the extension header itself.
             if node.parent?.is(ExtensionDeclSyntax.self) == true {
+                return
+            }
+            // Mirror the single-identifier rule: a type appearing as a generic
+            // argument is left alone, since `Self` may not be available in that
+            // position (e.g. an extension's own inheritance clause).
+            if node.parent?.is(GenericArgumentSyntax.self) == true {
                 return
             }
             if let tokens = memberTypeChain(node), tokens.map(\.text) == components {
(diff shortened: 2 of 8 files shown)


Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRuleExamples.swift
```diff
@@ -235,6 +244,25 @@ enum PreferSelfInStaticReferencesRuleExamples {
...
+            }
+            """, excludeFromDocumentation: true),
+        Example("""
+            class Outer {
+                class Inner {
+                    static let i = 0
+                }
+            }
+            extension Outer.Inner {
+                func f() -> Int { ↓Inner.i }
```
SimplyDanny: `Inner.i` doesn't compile. It should be `Outer.Inner.i`.

CHANGELOG.md
```diff
@@ -12,6 +12,14 @@
 
 ### Enhancements
 
+* Surface violations of `prefer_self_in_static_references` inside
+  extensions of types whose name is a plain identifier or a member type
+  (e.g. `extension Foo` or `extension Foo.Bar`). Stored-property
+  initializers and function signatures remain exempt — matching the
+  rule's existing behaviour for top-level class declarations.
```
SimplyDanny: Could be shorter. Something like "Treat extensions like classes in ... rule.".

Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRuleExamples.swift
```diff
@@ -277,5 +305,20 @@ enum PreferSelfInStaticReferencesRuleExamples {
...
+                func f() -> Int { ↓C.i }
+            }
+            """): Example("""
+                class C {
+                    static let i = 0
+                }
+                extension C {
+                    func f() -> Int { Self.i }
+                }
+                """),
```
SimplyDanny, line 369: Please add an `Outer.Inner.i` correction example as well.

SimplyDanny (changes requested):
Thanks! That's the right direction with a few remarks ...

itsybitsybootsy:
comment #1 wasn't just a rename. the rule never matched member types on extensions, only single identifiers, so `extension Foo.Bar` flagged nothing and `Outer.Inner.i` slipped through. added structural matching for member types, so it now flags and fixes them to `Self`. that turned up one case in our own Configuration+Remote.swift, fixed in the diff. shortened the changelog, added the Outer.Inner.i correction.

Source/SwiftLintBuiltInRules/Rules/Style/PreferSelfInStaticReferencesRule.swift
```diff
@@ -82,21 +102,52 @@ private extension PreferSelfInStaticReferencesRule {
...
         override func visit(_ node: MemberAccessExprSyntax) -> SyntaxVisitorContinueKind {
             if case .likeClass = parentDeclScopes.peek() {
                 if node.declName.baseName.tokenKind == .keyword(.self) {
                     return .skipChildren
                 }
             }
+            // In a member-type extension (e.g. `extension Foo.Bar`), the only
+            // valid reference to the extended type is the full member path, so
+            // match the access chain structurally rather than a single token.
+            // A path used to call an initializer (`Foo.Bar()`) is left alone.
```
SimplyDanny: `Foo.Bar<Int>()` is not left alone though.

SimplyDanny (changes requested):

itsybitsybootsy:
yeah, `Foo.Bar<Int>()` wraps the member access in a GenericSpecializationExpr so the FunctionCall guard doesn't see it. added GenericSpecializationExpr to that guard, plus the same check on GenericArgumentSyntax for types. example covers `Outer.Inner<Int>()` and `Outer.Inner<Int>.i`.

SimplyDanny (approved):

PR #6648, opened 2026-05-17
fix(missing_docs): treat actors as inheriting Actor

### LINKED ISSUE #5422: missing_docs warns about actors not documenting public protocol properties
### New Issue Checklist

### Describe the bug

If you use the "actor" keyword, it implicitly conforms to the "Actor" protocol, but SwiftLint still warns about not documenting public properties from the Actor protocol. For example, if you provide a public "unownedExecutor", it warns about it not documenting it. But if you explicitly conform to "Actor" then it doesn't warn.

It seems like SwiftLint should treat actors as if they explicitly specified they conform to the "Actor" protocol.

**Generates warnings**
```
/// Documentation for MyActor.
public final actor MyActor {
    public nonisolated var unownedExecutor: UnownedSerialExecutor { ... } // Warns here
}
```

**No warnings**
```
/// Documentation for MyActor.
public final actor MyActor: Actor {
    public nonisolated var unownedExecutor: UnownedSerialExecutor { ... }
}
```

```bash
$ swiftlint lint
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure)? 0.53.0

```yml
opt_in_rules:
  - missing_docs
```

No

Invoking SwiftLint directly.

Pasted above.

wyf027:
## Summary

- When `excludes_inherited_types` is enabled (the default), `missing_docs` now skips members inside `actor` declarations, matching behavior for types with explicit inheritance.
- Actors implicitly conform to the `Actor` protocol, so members like `unownedExecutor` should not require documentation when inherited types are excluded.

Fixes #5422

## Test plan

- [x] Added non-triggering example for documented actor with `nonisolated` member
- [x] Added triggering example when `excludes_inherited_types: false`
- [ ] CI (Buildkite)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,12 @@
 
 ### Bug Fixes
 
+* Treat implicit `Actor` conformance in `missing_docs` when `excludes_inherited_types`
+  is enabled, skipping documentation requirements for known inherited members such as
+  `unownedExecutor` without skipping all actor members.  
+  [leno23](https://github.com/leno23)
+  [#5422](https://github.com/realm/SwiftLint/issues/5422)
+
 * Avoid false positives in `prefer_self_in_static_references` for generic
   constraints and generic parameter bounds such as `where A: P` and `<A: P>`
   in classes and extensions.  
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/MissingDocsRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/MissingDocsRule.swift
@@ -26,7 +26,7 @@ private extension MissingDocsRule {
         }
 
         override func visit(_ node: ActorDeclSyntax) -> SyntaxVisitorContinueKind {
-            if node.inherits, configuration.excludesInheritedTypes {
+            if node.explicitlyConformsToActor, configuration.excludesInheritedTypes {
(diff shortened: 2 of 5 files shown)


SimplyDanny (changes requested):
The changes are causing way too many findings in OSS projects. Please review the scope of your implementation. The rule should only skip known actor properties/functions (by default).

wyf027:
Thanks for the review @SimplyDanny — I've pushed an update that narrows the scope:

- Only skips known inherited `Actor` members (currently `unownedExecutor`) inside actors that do **not** explicitly conform to `Actor`
- Explicit `: Actor` still uses the existing `skipChildren` behavior
- Other actor members are still linted as before

This should avoid the OSS-wide noise from skipping all actor children.

wyf027:
Pushed another update per review:

- Added `excluded_implicit_actor_members` configuration (default: `["unownedExecutor"]`) so only known inherited Actor members are skipped
- Implicit actors still lint other members; explicit `: Actor` keeps existing `skipChildren` behavior

Happy to adjust the default list if you have other common members in mind.

wyf027:
@SimplyDanny The latest push includes `excluded_implicit_actor_members` (default: `unownedExecutor`). Could you re-review when convenient? Once this is approved I can open the remaining queued PRs.

PR #6662, opened 2026-05-24
Fix `explicit_self` false positives around string interpolation

### LINKED ISSUE #6611: Bug: `explicit_self` false positive with string interpolation

SimplyDanny:
Fixes #6611. 

Follow-up of #6610.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -112,6 +112,11 @@
   existing `@TestInjected` and `@TestWeakly` exclusions.  
   [William-Laverty](https://github.com/William-Laverty)
   [#5803](https://github.com/realm/SwiftLint/pull/5803)
+  
+* Fix `explicit_self` false positives around string interpolation.  
+  [jffmrk](https://github.com/jffmrk)
+  [SimplyDanny](https://github.com/SimplyDanny)
+  [#6611](https://github.com/realm/SwiftLint/issues/6611)
 
 * Properly taint variables in tuples for `unneeded_escaping` rule.  
   [SimplyDanny](https://github.com/SimplyDanny)
--- a/Source/SwiftLintBuiltInRules/Rules/Style/ExplicitSelfRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/ExplicitSelfRule.swift
@@ -1,5 +1,6 @@
 import Foundation
 import SourceKittenFramework
+import SwiftSyntax
 
 struct ExplicitSelfRule: CorrectableRule, AnalyzerRule {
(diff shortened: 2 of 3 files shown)


jffmrk (approved):
This works on my test project and our main project. Thanks for simplifying things!

PR #6663, opened 2026-05-25
Tolerate unreadable directories and dedupe globs in linear time

Chupik:
## Introduction

We noticed that SwiftLint sometimes has problems with our very large repository (50,000+ files). The problem is that SwiftLint occasionally fails to enumerate the files to lint. We finally reproduced the problem and used an LLM to investigate and fix it. As a side benefit, the LLM also identified a way to significantly improve linting performance - see the Performance section below. The explanation below was drafted with LLM assistance.

## Problem

`Glob.expandGlobstar` uses `subpathsOfDirectory(atPath:)` to discover nested directories before per-directory globbing. That API is all-or-nothing: it throws on the first item it cannot access, and the catch block discards every directory collected so far. After the throw, only the search root remains in `directories`, so the rest of `expandGlobstar` globs nothing more than the root pattern — every nested file is silently ignored.

Then in `Glob.resolveGlob`, matched paths are deduped with `Array.unique`, whose `Equatable` overload is O(n²) (`Array.contains` linear scan per element). On a 50k-file project that's ~2.5 billion string comparisons; in practice it effectively hangs.

So a large project with a `**` glob and any unreadable subdirectory (permission denied, dangling symlink, file removed mid-scan in CI) is hit by both bugs: most files are dropped, and the few that remain take very long to dedupe.

## Fix

1. Replace `subpathsOfDirectory(atPath:)` with a lazy `FileManager.enumerator(at:includingPropertiesForKeys:options:errorHandler:)`. The `errorHandler` returns `true`, so a single unreadable entry skips only itself rather than aborting the whole scan. Removed the now-unused `Glob.isDirectory(path:)` helper since `.isDirectoryKey` replaces its job (and is pre-cached by the enumerator).
2. Add a `Hashable` overload of `Array.unique` (linear time via `Set.insert`). Swift's overload resolution selects it automatically wherever the element type is `Hashable`, so every existing call site gets the speed-up with no other code changes — including the call inside `Glob.resolveGlob` that motivated this.

## Tests

`testGlobstarToleratesUnreadableSubdirectory` in `Tests/FileSystemAccessTests/GlobTests`: creates a temp tree, chmods one subdirectory `000`, runs `Glob.resolveGlob("**/*.swift")`, asserts that siblings of the unreadable subtree still resolve while the subtree itself stays inaccessible. Skipped under root since permission-bit tests can't exercise the tolerance fix as root.

## Performance

Measured on an internal iOS monorepo (58 133 `.swift` files; 43 643 of them lintable after `excluded:` filtering) using `swiftlint lint --no-cache --quiet --only-rule hardcoded_localizable_string --write-baseline …`. Runs alternated between the two binaries to even out OS page-cache effects.

| Run      | This branch | `realm/SwiftLint` 0.59.1 |  Speedup |
| -------- | ----------: | -----------------------: | -------: |
| 1        |    52.93 s  |                 83.74 s  |   1.58×  |
| 2        |    51.36 s  |                 82.33 s  |   1.60×  |
| 3        |    45.37 s  |                 79.58 s  |   1.75×  |
| **mean** | **49.89 s** |             **81.88 s**  | **1.64×** |
| min      |    45.37 s  |                 79.58 s  |   1.75×  |
| max      |    52.93 s  |                 83.74 s  |   1.58×  |

Baseline output is byte-identical between the two binaries, so the speedup is on the same workload (not a case of the new build doing less work):

| Metric                            |          New |          Old |
| --------------------------------- | -----------: | -----------: |
| Baseline JSON size                | 3 413 444 B  | 3 413 444 B  |
| Violations reported               |       6 072  |       6 072  |
| Distinct files with violations    |       1 750  |       1 750  |

## Notes for review

- **Symlink behavior** (intentional, worth flagging): the URL enumerator does not descend into symlinks to directories, whereas `subpathsOfDirectory` did. For glob expansion this is generally desirable — no symlink-cycle hangs, no double-linting — and matches the direction this codebase has already taken in `FileManager.collectFiles`. Happy to change if you'd rather preserve the old following behavior.
- **`Array.unique` public surface**: this adds a new computed property on `Array where Element: Hashable` alongside the existing `Array where Element: Equatable`. Overload resolution picks the Hashable one transparently, so no source breakage. If you'd prefer to scope the fix to a private `Set`-based dedup in `Glob.swift` and leave `Array.unique` untouched, that's a one-line alternative — let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -186,6 +186,16 @@
   be ordered via the `isolation` entry in `preferred_modifier_order`.  
   [leno23](https://github.com/leno23)
   [#6164](https://github.com/realm/SwiftLint/issues/6164)
+* Make `Glob.expandGlobstar` tolerant of unreadable directory entries on
+  large trees. `subpathsOfDirectory(atPath:)` aborted the entire glob
+  expansion on the first unreadable entry (permission denied, dangling
+  symlink, file removed mid-scan), causing most files in large projects to
+  be silently ignored. Replace the directory walk with a lazy `URL`
+  enumerator that has a per-item error handler so unreadable items are
+  skipped individually. Also dedupe glob results in linear time by adding
+  a `Hashable` overload of `Array.unique`, removing a quadratic dedup that
+  previously effectively hung 50k-file projects.  
+  [Chupik](https://github.com/Chupik)
 
 * Detect and autocorrect missing whitespace before `else` in `guard`
   statements for the `statement_position` rule.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,15 @@
   [leno23](https://github.com/leno23)
   [#5648](https://github.com/realm/SwiftLint/issues/5648)
 
+* Make `Glob.expandGlobstar` tolerant of unreadable directory entries on
+  large trees. `subpathsOfDirectory(atPath:)` aborted the entire glob
+  expansion on the first unreadable entry (permission denied, dangling
+  symlink, file removed mid-scan), causing most files in large projects to
+  be silently ignored. Replace the directory walk with a lazy `URL`
+  enumerator that has a per-item error handler so unreadable items are
+  skipped individually.  
+  [Chupik](https://github.com/Chupik)
+
 * Avoid false positives in `prefer_self_in_static_references` for generic
   constraints and generic parameter bounds such as `where A: P` and `<A: P>`
   in classes and extensions.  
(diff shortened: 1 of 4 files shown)


SimplyDanny:
Hey @Chupik, the directory enumeration part looks reasonable to me. However, the `unique` part ist not clear to me. Where is `.unique` called anyway that is of relevance here?

Would you be willing to rebase your branch first?

SimplyDanny (approved):

SimplyDanny:
> However, the `unique` part ist not clear to me. Where is `.unique` called anyway that is of relevance here?

Removed as it's not being used anymore.

> Would you be willing to rebase your branch first?

Done.

Chupik:
@SimplyDanny hello! Thank you and sorry for the late reply.

I rechecked Glob.swift and you are right - `.unique` is no longer used. I should have removed it during rebasing on fresh main branch.

SimplyDanny:
> @SimplyDanny hello! Thank you and sorry for the late reply.
> 
> I rechecked Glob.swift and you are right - `.unique` is no longer used. I should have removed it during rebasing on fresh main branch.

All good. I fixed that up. It's merged and released now.

PR #6665, opened 2026-05-25
Add sort order option for sorted imports

### LINKED ISSUE #4810: Feature request: lexicographic sorting options for `sorted_imports`

ayushnagpal570-ai:
## Summary
- add a `sort_order` option to `sorted_imports` with `case_insensitive` as the existing default and `lexicographic` as the new alternative
- thread the configured sort order through linting and autocorrection
- cover the new ordering mode in rule examples and the default configuration fixture

Closes #4810.

## Tests
- `swift test --filter SortedImportsRuleGeneratedTests`
- `swift test --filter RuleConfigurationTests`
- `swift test --filter IntegrationTests`
- `git diff --check`

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Style/SortedImportsRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/SortedImportsRule.swift
@@ -25,7 +25,12 @@ private extension SortedImportsRule {
 
         override func visitPost(_ node: ImportDeclSyntax) {
             imports.append(
-                Import.from(importDecl: node, grouping: configuration.grouping, locationConverter: locationConverter)
+                Import.from(
+                    importDecl: node,
+                    grouping: configuration.grouping,
+                    sortOrder: configuration.sortOrder,
+                    locationConverter: locationConverter
+                )
             )
         }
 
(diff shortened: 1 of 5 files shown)


Source/SwiftLintBuiltInRules/Rules/Style/SortedImportsRule.swift
```diff
@@ -183,6 +190,7 @@ private struct Import: Comparable {
     let offset: Int
     let attributes: String
     let modifier: UInt8
+    let sortOrder: SortedImportsConfiguration.SortOrder
```
SimplyDanny, line 193: I don't think this should be a property of an `import`, but of the comparison. What if I set one style for one import and another one for the other? That doesn't happen here, but the API would suggest that it would work ...

SimplyDanny (changes requested):
Thank you for the contribution, @ayushnagpal570-ai!

I wonder if `case_insensitive` and `lexicographic` are the clearest pair of names here. `lexicographic` is technically accurate, but it doesn’t feel parallel with `case_insensitive`, so the distinction may be a little harder to grasp from the config alone. If the practical difference is just whether case affects ordering, would `case_insensitive`/`case_sensitive` be clearer and more consistent?

PR #6667, opened 2026-05-26
Update Yams to version 6.2.2

SimplyDanny:


jpsim:
Needs https://github.com/bazelbuild/bazel-central-registry/pull/8998

Package.swift
```diff
@@ -38,7 +38,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-argument-parser.git", .upToNextMajor(from: "1.6.1")),
         .package(url: "https://github.com/swiftlang/swift-syntax.git", exact: "604.0.0-prerelease-2026-04-21"),
         .package(url: "https://github.com/jpsim/SourceKitten.git", .upToNextMajor(from: "0.37.2")),
-        .package(url: "https://github.com/jpsim/Yams.git", .upToNextMajor(from: "6.0.2")),
+        .package(url: "https://github.com/jpsim/Yams.git", .upToNextMajor(from: "6.2.2")),
```
jpsim: Why constrain consumers? This can stay the same while the `Package.resolved` file updates.

SimplyDanny: Right. Just reflex. Reverted.

PR #6692, opened 2026-05-30
Fix unused_setter_value false positives for empty setter bodies

### LINKED ISSUE #3863: `unused_setter_value` incorrectly gets triggers for an empty setter
### New Issue Checklist

### Describe the bug

The following code gets reported as a vioation of the `unused_setter_value` rule, even though it was one of the examples [explicitly cited](https://github.com/realm/SwiftLint/pull/3653/files#diff-1fbe50f693894fd9cd63ee11f77b80b49944fa39c7a508798626ddbe83b1b1a7R50) as being non-offending.

```swift
override var aValue: String {
  get {
    return Persister.shared.aValue
  }
  set() { }
```

There might be a meta-issue here, about why the example didn't cause a test failure.

```bash
$ swiftlint lint
```

### Environment

* SwiftLint version: `0.46.3`
* Which Xcode version are you using (check `xcodebuild -version`)? `Xcode 13.2.1 Build version 13C100`
* Do you have a sample that shows the issue? Yes:

    ```bash
    echo 'private var abc: Int { get { 123 } set() { } }' | swiftlint lint --no-cache --use-stdin --enable-all-rules`

wyf027:
## Summary

- Skip `unused_setter_value` when the setter accessor body has no statements (e.g. `set { }`, `set() { }`, `set {}`).
- Covers protocol witness stubs and other intentional empty setters discussed in #3863.

## Test plan

- [x] `swift test --filter UnusedSetterValueRuleGeneratedTests`
- [x] Manual lint on issue reproducers (`set() { }`, `set { }` with empty body)

Fixes #3863

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,11 @@
 
 ### Bug Fixes
 
+* Fix `unused_setter_value` false positives for empty setter bodies, including
+  `set() { }` and protocol witness implementations with `set { }`.  
+  [leno23](https://github.com/leno23)
+  [#3863](https://github.com/realm/SwiftLint/issues/3863)
+
 * Avoid false positives in `prefer_self_in_static_references` for generic
   constraints and generic parameter bounds such as `where A: P` and `<A: P>`
   in classes and extensions.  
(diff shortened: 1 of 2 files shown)


SimplyDanny:
@leno23, you either react on and follow my comments in your previous gazillion PRs or I'll just close all of them! 

Neither do I like to be spammed by a robot nor will I accept to be ignored. You are burning resources for nothing. Open source is about collaboration and engagement, not about "just getting shit done", especially with "Good first issues" that are meant for people to learn and familiarize themselves with the project.

There is nothing against AI being used as a tool. But I'll definitely not blindly merge any auto-generated junk without any sensible interaction!

This is your chance to reply. Any new PR you open, leads to all of them being just closed. There is really much more important work to do than dealing with people spamming the project!

PR #6762, opened 2026-06-06
Skip SourceKit requests once sourcekitd is known unavailable

Brett-Best:
Targets the `adopt-swift-testing` branch (#6048).

## Problem

On `adopt-swift-testing` the **macOS CI test jobs hang and are cancelled at the 10‑minute timeout** (Linux/Windows pass). In the cancelled run the whole suite prints `started` in one burst, then `sourcekitd has failed` at ~T+2s, then **zero tests complete for 8+ minutes**.

Root cause: Swift Testing runs every test as a concurrent task on a **bounded cooperative executor** (XCTest used its own per-test threads). Each linted file issues a synchronous, blocking sourcekitd request. When sourcekitd is unavailable, each request **fails fast (~2s)** but there is no global short‑circuit, so all ~1000 files independently re‑issue their own doomed request. Across the few cooperative threads that is >10 minutes of nothing-but-failing-requests, which starves the executor so no other test can make progress.

## Fix

Add a process‑global `SourceKitStatus` latch, enforced centrally in `Request.sendIfNotDisabled()` (so it covers `editorOpen`, `index`, and `cursorInfo` uniformly):

- Once a request fails **and sourcekitd has never produced a successful response in this process**, `isUnavailable` latches `true` and subsequent requests throw `SourceKitUnavailableError` immediately instead of blocking. The cascade collapses to a single failed request.
- The **"never succeeded" guard** is what keeps this safe for normal linting. SwiftLint deliberately handles *per‑file* sourcekitd crashes (`SwiftLintFile.sourcekitdFailed` / `assertHandler`); such a crash happens *after* sourcekitd has already answered other files, so `everSucceeded` is already set and the global latch never engages. On a healthy machine the first request succeeds, so the latch is inert — **no behaviour change**.

Also routes `UnusedDeclarationRule`'s `index` request through `sendIfNotDisabled()` (it previously called `send()` directly, bypassing the latch).

## Testing

- New deterministic regression test (`SourceKitCrashTests.sourceKitUnavailableShortCircuitsWithoutIssuingRequest`): forces the unavailable state for one task tree (isolated from parallel tests via a `@TaskLocal`) and asserts the central choke point **throws without issuing a request**. It fails before the fix and passes after.
- Full local suite green (`1067 tests / 370 suites`), `swiftlint lint --strict` clean.

## Notes / caveats

- The latch only auto‑engages when sourcekitd never came up (the CI failure mode). I could not reproduce that exact mode locally — under a CPU‑throttled container sourcekitd instead *works but is glacially slow*, a different bottleneck this change doesn't target — so the auto‑latch trigger is validated by the regression test and the macOS CI log rather than an end‑to‑end local repro. macOS CI on this PR is the real confirmation.
- One edge: if the very first sourcekitd request (the `SwiftVersion` startup probe) fails *transiently* on an otherwise‑healthy machine, the latch trips for the run. In practice this is rare, and it is actually beneficial on CI (trips early, before file processing).


SimplyDanny:
Thanks for the idea, @Brett-Best.

As it turns out, the reason for the hang is not that SourceKit failed (then we should fix the fail), but that it gets overwhelmed by the number of requests it receives. This can be reproduced even with a small number of parallel threads. Not sure whether this is an issue of SourceKit or due to the way we use it.

In any case, this seems only to be an issue in parallel testing, so running them sequentially for now should be fine until there is a proper solution.

PR #6768, opened 2026-06-09
Add `--parent-config` command-line option

### LINKED ISSUE #5421: Allow parent config as an argument
### New Issue Checklist

### Feature Request

I would like to add `--parent-config` argument to the `lint` command. That way we don't have to specify the `parent-config:` inside of the config file. The use case for this is that the parent config path is not known ahead of time and I need to pass it in from a temporary directory where it is pulled and I don't control the path of the directory.

This is how it would look
```
swiftlint lint --parent-config "path/to/config" 
```

It would add the parent config to the `.swiftlint` config file it reads during execution.

tsushanth:
## Why

Closes #5421.

The \`parent_config:\` entry inside a SwiftLint configuration file lets a project chain a parent ruleset, but the path has to be baked into the child config file at authoring time. The issue describes the case where the parent config lives in a temporary directory whose path is only known at lint time — pulled from a remote, generated by a script, etc. — and the child config can't reasonably hard-code it.

## What

Adds \`--parent-config <path>\` to both \`lint\` and \`analyze\`. Equivalent to declaring \`parent_config:\` inside the resolved config, but supplied at the command line so the parent path can be passed in dynamically. Can be repeated to layer multiple parent configurations.

\`\`\`bash
swiftlint --parent-config /tmp/pulled/shared.yml --config .swiftlint.yml
\`\`\`

In the file graph, each \`--parent-config\` vertex becomes a parent of the first \`--config\` vertex (or of the default \`.swiftlint.yml\` when no \`--config\` is provided). Existing \`parent_config:\` resolution inside YAML configs is unchanged.

The option threads through:

- \`LintOrAnalyzeArguments.parentConfig\` (\`[URL]\`)
- \`LintOrAnalyzeOptions.parentConfigurationFiles\`
- \`Configuration.init(configurationFiles:parentConfigurationFiles:...)\` (default \`[]\` to preserve all existing call sites)
- \`Configuration.FileGraph.init(commandLineChildConfigs:commandLineParentConfigs:...)\` (default \`[]\` for the same reason)

8 files, ~57 LOC. Cache identifier now includes the parent-config list so configurations with different parent paths don't collide.

## Compat

- \`Configuration.init(configurationFiles:...)\` gains an optional \`parentConfigurationFiles\` parameter with a \`[]\` default. All existing call sites compile unchanged.
- \`FileGraph.init\` gains an optional \`commandLineParentConfigs\` parameter with a \`[]\` default. Internal-only initializer; no public consumers.
- \`LintOrAnalyzeOptions.init\` gains a \`parentConfigurationFiles\` parameter with a \`[]\` default. \`package\` access, only \`Lint.swift\` / \`Analyze.swift\` call it directly.
- No deprecation needed.

## Tests

\`swift build\` is clean. Happy to add a fixture-based test exercising the parent-config injection (likely in \`Tests/SwiftLintFrameworkTests/ConfigurationTests/\`) if you'd like — pointing me at the right existing test as the template would help.

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/DiscouragedDefaultParameterRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/DiscouragedDefaultParameterRule.swift
@@ -10,7 +10,7 @@ struct DiscouragedDefaultParameterRule: Rule {
         name: "Discouraged Default Parameter",
         description: "Default parameter values should not be used in functions with certain access levels.",
         rationale: """
-            By discouraging default parameter values in functions, that are exposed to other source files in the module
+            By discouraging default parameter values in functions that are exposed to other source files in the module
             or package and their consumers, we can promote call sites and reduce the likelihood of bugs caused by
             unexpected (or changed) default values being used.
             """,


tsushanth:
Rebased onto current main.

tsushanth:
Rebased onto current main.

PR #6780, opened 2026-06-20
Support Bazel 8 and 9

### LINKED ISSUE #6630: Update to bazel 9 ?
Hello, Wanted to know if you are open to update the project to support bazel 9?

If you are okay, we can open a PR

SimplyDanny:
Resolves #6630.


SimplyDanny:
Could you take a look, @keith? Is it especially okay to remove `max_compatibility_level` and `compatibility_level` if Bazel 7 is still supported?

keith:
unfortunately not safe to drop for the older versions

keith (commented):
lgtm besides the compatibility level changes

keith:
it might work for 7.x but probably not 7.0.0, so it's probably better to just leave them for a bit

MODULE.bazel
```diff
@@ -1,7 +1,7 @@
 module(
     name = "swiftlint",
     version = "0.64.1",
-    compatibility_level = 1,
+    bazel_compatibility = [">=7.0.0"],
```
SimplyDanny, line 5:
```suggestion
    compatibility_level = 1,
    bazel_compatibility = [">=7.0.0"],
```

SimplyDanny:
> it might work for 7.x but probably not 7.0.0, so it's probably better to just leave them for a bit

Thanks for the review. I just re-added them. We can live with the warning for the time being.

PR #6783, opened 2026-06-22
Honor excluded paths for projects under /private-symlinked directories

### LINKED ISSUE #6782: excluded paths ignored when project is under /private (e.g. /tmp, mktemp, CI workspaces) — regression in 0.64.0
### New Issue Checklist

- [x] Updated SwiftLint to the latest version (0.64.0)

### Describe the bug

`excluded` paths are silently ignored, and the files get linted anyway, when the project's real path is located **under `/private`** (which on macOS includes anything under `/tmp`, `/var`, and `/var/folders`, since those are symlinks to `/private/tmp`, `/private/var`, …).

The same project, same config, and same SwiftLint binary work correctly when the project lives under a non-`/private` path such as `/Users/...`.

This is a **regression in 0.64.0**. `0.63.3` excludes the files correctly in the exact same location.

This bites in practice because:

- Many CI systems run jobs out of a workspace under `/private/var/...`.
- Anything created with `mktemp -d` (tests, scripts, scratch repos) lands under `/private/var/folders/...`.

In both cases, files that should be excluded (e.g. generated code) start failing the lint, with no config change.

### Complete output when running SwiftLint

Self-contained reproduction (paste into a terminal on macOS):

```bash
# mktemp -d returns a path under /var/folders -> /private/var/folders
D=$(mktemp -d)
mkdir -p "$D/Sources/Feature/Sources"
printf 'let value = Optional(1)!\n' > "$D/Sources/Feature/Sources/Generated.swift"
cat > "$D/.swiftlint.yml" <<'EOF'
included:
  - Sources
excluded:
  - Sources/**/Generated.swift
opt_in_rules:
  - force_unwrapping
EOF
cd "$D" && swiftlint lint --quiet
```

**Actual (0.64.0):** the excluded file is linted 

```
.../Sources/Feature/Sources/Generated.swift:1:24: warning: Force Unwrapping Violation: Force unwrapping should be avoided. (force_unwrapping)
```

**Expected (and actual on 0.63.3):** no output.`Generated.swift` is excluded.

#### Isolating the trigger

Same project (`Sources/Feature/Sources/Generated.swift`, same `.swiftlint.yml`), SwiftLint 0.64.0, only the location differs:

| Project real path | `excluded` honored? |
|---|---|
| `/Users/<me>/sl-repro` | ✅ yes (no violations) |
| `/private/tmp/sl-repro` (also via the `/tmp` symlink) | ❌ no (file linted) |
| `/private/var/folders/.../sl-repro` (`mktemp -d`) | ❌ no (file linted) |

`0.63.3` honors `excluded` in **all** of these locations. It is not about how you `cd` into the directory (canonical path vs. via the `/tmp` symlink behaves the same), it depends on whether the project's real path is under `/private`.

I tried several config variants under 0.64.0, none restored the exclusion in a `/private` location:

- `Sources/**/Generated.swift` (glob)
- `**/Generated.swift` (leading globstar)
- `Sources/Feature/Sources/Generated.swift` (explicit, no glob)

### Environment

* SwiftLint version: **0.64.0** (regression from **0.63.3**, which works)
* Installation: Homebrew (also reproduced with the `portable_swiftlint.zip` release binaries for both versions)
* Xcode / OS: macOS
* Paths involved are all under `/private` via the standard macOS `/tmp` → `/private/tmp` and `/var` → `/private/var` symlinks

### Likely cause

Looks like a path-normalization mismatch introduced in 0.64.0: the lint-target file paths and the resolved `excluded` patterns end up standardized differently with respect to the leading `/private`. macOS `NSString.standardizingPath` strips a leading `/private` (so `/private/var/x` → `/var/x`), while symlink resolution goes the other way (`/var/x` → `/private/var/x`). If one side is standardized and the other is symlink-resolved, the exclusion comparison never matches for any path under `/private`, which is exactly the observed behavior. Under `/Users` neither transform changes the path, so exclusion still matches.

tumata:
## Summary

Fixes #6782.

`excluded` paths are silently ignored and the files get linted anyway when the linted project's real path is located under a `/private`-symlinked directory. On macOS that includes anything under `/tmp`, `/var`, and `/var/folders` (firmlinks to `/private/tmp`, `/private/var`, …), so it commonly affects CI workspaces and `mktemp` directories. This is a regression introduced in 0.64.0.

## Root cause

Exclusion matchers/prefixes are built from `excludedPaths`, whose URLs are produced by `String.url()`, which applies `.standardizedFileURL`. On macOS, standardization **strips a leading `/private`** (so `/private/var/x` → `/var/x`).

The file paths to be linted, however, come from `FileManager`'s directory enumerator, which yields paths that **keep** the `/private` prefix (`/private/var/x/...`).

So `Excluder.excludes(path:)` compares `/private/...` file paths against `/...` patterns, and the `FilenameMatcher` / prefix check never matches. Under `/Users` there is no `/private` to strip, which is why the bug looks machine-dependent.

## Fix

In `Excluder.excludes(path:)`, compare each candidate against the path **both with and without** a leading `/private` prefix, using string operations only. This bridges the discrepancy without touching the filesystem, and unlike standardizing every candidate it leaves matching behavior for non-`/private` paths exactly as it was.

## Performance

An earlier iteration of this fix standardized every candidate via `standardizedFileURL`, which `stat`s each path and regressed linting performance noticeably. The current string-only approach avoids that. Measured with `tools/oss-check`'s command (`lint --no-cache --enable-all-rules`, 5-run average) on the performance corpus:

| | avg | vs `main` |
|---|---|---|
| `main` | 0.18 s | N/A |
| `standardizedFileURL` approach | 0.28 s | +55% |
| this PR (string-only) | 0.20 s | +10% (within noise) |

Same violation count in all cases.

## Test

Added `ConfigurationTests.testExcludedPathsUnderSymlinkedDirectory`, which creates a small project under the system temporary directory (resolves under `/private/var/folders/…` on macOS) and asserts the excluded file is not linted. Verified it **fails without the fix** (the excluded file leaks into the lintable set) and **passes with it**. The full `FileSystemAccessTests` suite passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,11 @@
 
 ### Bug Fixes
 
-* None.
+* Honor `excluded` paths when the linted project is located under a system directory
+  on macOS that resolves through a symlink, such as `/var` and `/tmp` (which point
+  to `/private/var` and `/private/tmp`).  
+  [tumata](https://github.com/tumata)
+  [#6782](https://github.com/realm/SwiftLint/issues/6782)
 
 ## 0.64.0: All Windows Opened
 
--- a/Source/SwiftLintFramework/Extensions/FileManager+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/FileManager+SwiftLint.swift
@@ -33,15 +33,36 @@ public enum Excluder {
     /// An excluder that does not exclude any files.
     case noExclusion
 
+    private static let privatePrefix = "/private"
+
     func excludes(path: URL) -> Bool {
-        switch self {
-        case let .matching(matchers):
-            matchers.contains(where: { $0.match(filename: path.path) })
-        case let .byPrefix(prefixes):
-            prefixes.contains(where: { path.path.hasPrefix($0) })
-        case .noExclusion:
-            false
+        if case .noExclusion = self {
+            return false
         }
+
+        // Exclusion matchers and prefixes are derived from standardized URLs (see `String.url()`),
+        // which on macOS drops a leading `/private` from paths under firmlinks such as `/tmp` and
+        // `/var`. Paths produced by `FileManager`'s directory enumerator keep that prefix, so a
+        // file under a `/private`-resolved directory (e.g. CI workspaces or `mktemp` directories)
+        // would never match. Compare against the path with and without the `/private` prefix to
+        // bridge the gap without touching the filesystem; standardizing every candidate via
+        // `standardizedFileURL` would `stat` each path and is measurably slower.
+        let fullPath = path.path
+        let strippedPath = fullPath.hasPrefix(Self.privatePrefix + "/")
+            ? String(fullPath.dropFirst(Self.privatePrefix.count))
+            : nil
+        return switch self {
+            case let .matching(matchers):
+                matchers.contains { matcher in
+                    matcher.match(filename: fullPath) || strippedPath.map { matcher.match(filename: $0) } == true
+                }
+            case let .byPrefix(prefixes):
+                prefixes.contains { prefix in
+                    fullPath.hasPrefix(prefix) || strippedPath?.hasPrefix(prefix) == true
+                }
+            case .noExclusion:
+                queuedFatalError("Unreachable case; should have been handled at the beginning of the method")
+            }
     }
 }
 
--- a/Tests/FileSystemAccessTests/ConfigurationTests.swift
+++ b/Tests/FileSystemAccessTests/ConfigurationTests.swift
@@ -293,6 +293,33 @@ final class ConfigurationTests: SwiftLintTestCase { // swiftlint:disable:this ty
         assertEqual(["directory/File1.swift", "directory/File2.swift"], paths)
     }
 
+    // On macOS the system temporary directory resolves through a `/private` symlink (e.g.
+    // `/var/folders/…` -> `/private/var/folders/…`). The directory enumerator yields file
+    // paths carrying the `/private` prefix, whereas exclusion patterns are standardized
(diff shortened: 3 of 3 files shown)


tumata:
@SimplyDanny please take a look at your earliest convenience. This is a bit of a blocker for us, because we use homebrew to install `SwiftLint` on our CI and are unable to pin to a specific version. All I can do for now is disable the lint checks on our repos, which is not ideal. Thanks!

SimplyDanny:
This apparently causes a slowdown in one of the OSS repositories intended to focus exactly on exclude performance:

> Linting SwiftLintPerformanceTests with this PR took 0.23 s vs 0.16 s on main (43% slower).

If we found a way that would roughly align with the numbers from `main`, it would be great. I'll need some time to think about it further.

tumata:
@SimplyDanny I've pushed a different approach for this. Instead of standardizing every path (which stats the filesystem and regressed performance), it just matches the candidate with and without the `/private` prefix using plain string comparison, so it's back within noise of main while still fixing the exclusion. 

Let's wait and see if Danger is happier.

EDIT:
image

SimplyDanny (approved):

tumata:
Thank you for merging the changes @SimplyDanny. Would you have a target date for patch release?

SimplyDanny:
> Thank you for merging the changes @SimplyDanny. Would you have a target date for patch release?

[There](https://github.com/realm/SwiftLint/releases/tag/0.64.1) you go. Homebrew will be ready in a bit as well.

tumata:
Brew finally updated, and I was able to successfully test end to end. 
Thank you for your prompt help in getting this PR merged @SimplyDanny! 🙏

PR #6796, opened 2026-07-01
Add opt-in inconsistent_default_argument_in_file rule

### LINKED ISSUE #6668: Add `inconsistent_default_argument` rule
### New Issue Checklist

### New Rule Request

Please add a `inconsistent_default_argument` rule (it could be named `coordinate_default_arguments` if you want an imperative name for the fix instead of a descriptive name for the problem, or something else).

In certain circumstances, it would ensure that parameters with the same label & type that have a default argument use the same default argument.

It could have a few options to specify the applicable circumstances. I'll leave it to you to decide what circumstances (& if the whole rule) should be enabled by default (also, you can rename any options, add options, etc.) (I'd prefer the rule & all options to be enabled by default, but I don't know how much code would violate the maximal rule, so I don't want to cause a tsunami of violations). Boolean options:

- `call_default_arguments`: If functions F & G both have a default argument for parameter P, if F calls G passing its value of P to to G's P, then it is an error if the default arguments for P for both F & G are not the same.
- `overload_default_arguments`: If methods F & G for the same type with the same name but with different signatures both have a default argument for parameter P, then it is an error if the default arguments for P for both F & G are not the same.
- `override_default_arguments`: If method F exists for type T, and overrides a method G from some type/protocol, S, that T extends / conforms to, and both F & G  have a default argument for parameter P, then it is an error if the default arguments for P for both F & G are not the same.
- etc.

nenadvulic:
## Summary

Adds a new **opt-in** rule `inconsistent_default_argument_in_file` (kind: `lint`), as discussed in #6668.

It flags overloaded functions declared on the **same type** — the type's primary declaration **and any extensions of it in the same file** — when a shared `(label, type)` parameter uses **different default values**. This catches accidental *default drift* across overloads that callers reasonably expect to behave the same, e.g.:

```swift
struct Network {
  func fetch(_ url: URL, retries: Int = 3) {}
  func fetch(_ url: URL, headers: [String: String], retries: Int = 0) {} // ← silently drops to 0
}
```

## Scope

Per the agreement with @SimplyDanny and @rgoldberg in #6668, this is the intentionally small, purely syntactic, single-file variant:

- **Opt-in** (disabled by default) — teams that want the "overloads share defaults" invariant get it; others aren't bothered.
- Behavior = `across_one_type`: a type's primary declaration plus all extensions of it **in the same file** are treated as one scope (matched via the fully-qualified lexical type name, e.g. `Foo.Bar`).
- **Nested types form their own scope** (`separately`): `Parent.f(…)` and `Parent.Nested.f(…)` are functions on different types, so they are never compared — comparing them would produce false positives.
- The offending parameter is the one whose default **diverges from the first declaration** of that shared parameter, in source order.

Deliberately **out of scope** here (left to a potential future analyzer rule, so no name clash):

- `call_default_arguments` and `override_default_arguments` — both need type/symbol resolution unavailable to a syntactic rule.
- Cross-file cases.

The `_in_file` suffix keeps the name honest about the limitation and leaves room for a fully-general analyzer rule later. Config options floated in the discussion (`container_scope`, `nested_containers`, `parameter_grouping`, …) are intentionally **not** included yet; they can be layered on additively without breaking the v1 defaults.

## Examples

Triggering:
```swift
struct S {
  func f(x: Int = 1) {}
  func f(↓x: Int = 2, y: Int) {}
}
```
```swift
struct S {
  func f(x: Int = 1) {}
}
extension S {
  func f(↓x: Int = 2, y: Int) {}
}
```

Non-triggering (same default, one-sided default, different type, different scope, nested type):
```swift
struct S {
  func f(x: Int = 1) {}
  func f(x: Int = 1, y: Int) {}
}
struct Outer {
  func f(x: Int = 1) {}
  struct Inner {
    func f(x: Int = 2) {} // different scope
  }
}
```
Closes #6668.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,14 @@
 
 ### Enhancements
 
-* None.
+* Add a new opt-in rule `inconsistent_default_argument_in_file` that flags
+  overloaded functions declared on the same type (its primary declaration and
+  any extensions of it in the same file) when a shared parameter uses different
+  default values, which is usually an accidental drift callers do not expect.
+  Nested types form their own scope, and the cross-file/semantic cases are left
+  to a potential future analyzer rule.  
+  [nenadvulic](https://github.com/nenadvulic)
+  [#6668](https://github.com/realm/SwiftLint/issues/6668)
 
 ### Bug Fixes
 
--- a/Source/SwiftLintBuiltInRules/Models/BuiltInRules.swift
+++ b/Source/SwiftLintBuiltInRules/Models/BuiltInRules.swift
@@ -94,6 +94,7 @@ public let builtInRules: [any Rule.Type] = [
     ImplicitlyUnwrappedOptionalRule.self,
     InclusiveLanguageRule.self,
     IncompatibleConcurrencyAnnotationRule.self,
+    InconsistentDefaultArgumentInFileRule.self,
     IndentationWidthRule.self,
     InvalidSwiftLintCommandRule.self,
     InvisibleCharacterRule.self,
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/InconsistentDefaultArgumentInFileRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/InconsistentDefaultArgumentInFileRule.swift
@@ -1,0 +1,201 @@
+import SwiftSyntax
+
+@SwiftSyntaxRule(optIn: true)
+struct InconsistentDefaultArgumentInFileRule: Rule {
+    var configuration = SeverityConfiguration<Self>(.warning)
+
+    static let description = RuleDescription(
+        identifier: "inconsistent_default_argument_in_file",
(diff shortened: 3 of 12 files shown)


nenadvulic:
@SimplyDanny the rule behaves exactly as scoped in this thread [6668](https://github.com/realm/SwiftLint/issues/6668) : grouping by (label, type) within a type's declaration + its extensions in the same file, nested types isolated.

All matches are intra-type (no cross-type false matches). The mix of real catches and intentional differences is expected and is the reason for the opt-in default; teams that want the "overloads share defaults" invariant enable it, others aren't affected.

rgoldberg:
@SimplyDanny Must the PR be modified before it can be merged?

IMO, ", this is usually an accidental drift that callers do not expect" should probably be excised from the violation messages.

I don't know if the violations reported for the various tests must be fixed, etc.

Thanks.

PR #6814, opened 2026-07-09
Add dark mode styling to docs

Brett-Best:
## Summary
- add dark-mode styling to the Jazzy-generated documentation
- cover page chrome, tables, inline/code blocks, search/typeahead, blockquotes, asides, and syntax highlighting
- use APCA-audited Tailwind palette values so checked text pairs clear Lc 60

## Validation
- ruby -e 'require "yaml"; YAML.load_file(".jazzy.yaml")'
- git diff --check
- APCA audit script: 28 checked pairs >= Lc 60
- swift run swiftlint generate-docs
- bundle exec jazzy --build-tool-arguments=--disable-sandbox

### DIFF
--- a/.jazzy.yaml
+++ b/.jazzy.yaml
@@ -8,6 +8,83 @@ swift_build_tool: spm
 theme: fullwidth
 clean: true
 copyright: '© 2023 [JP Simard](https://jpsim.com) under MIT.'
+head: |
+  <style>
+  @media (prefers-color-scheme: dark) {
+    /* Colors below are from the Tailwind CSS default palette:
+       https://tailwindcss.com/docs/colors#default-color-palette-reference */
+    :root { color-scheme: dark; }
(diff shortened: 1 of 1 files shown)


.jazzy.yaml
```diff
@@ -8,6 +8,83 @@ swift_build_tool: spm
 theme: fullwidth
 clean: true
 copyright: '© 2023 [JP Simard](https://jpsim.com) under MIT.'
+head: |
```
SimplyDanny, line 11: I wonder if this can go into a separate file for easier access, syntax highlighting, ... Did you try that or check for an option?

SimplyDanny: While there are options to have the styles in a dedicated file, duplication would be necessary with the commands handling documentation generation. In terms of "single source of truth", we should keep it like proposed in the PR.

SimplyDanny (approved):
Thanks @Brett-Best!

PR #6824, opened 2026-07-13
Ignore unowned(unsafe) captures by default

### LINKED ISSUE #6817: `unowned_variable_capture` fires for `unowned(unsafe)` captures
the `unowned_variable_capture` rule currently fires not only for `unowned` captures, but also for `unowned(unsafe)` captures.

IMO this is excessive, as the `unsafe` addition already makes clear the programmer's intent to trade safety for performance here, and guarantee that the captured value's lifetime will exceed that of the closure.

suggestion: `unowned_variable_capture` should fire only for `unowned` captures, but not for `unowned(unsafe)` ones.

looking at [the rule's documentation](https://realm.github.io/SwiftLint/unowned_variable_capture.html), it seems like this should be the behaviour anyway, as the "triggering examples" section lists only `unowned` captures but no `unowned(unsafe)` ones.

Yurii201811:
## Summary

Ignore `unowned(unsafe)` captures by default while preserving an opt-in `include_unsafe` setting.

## Background

The explicit unsafe spelling communicates the intended lifetime tradeoff.

Closes #6817

## Changes

- add `include_unsafe`, defaulting to `false`
- distinguish unsafe capture specifiers from `unowned` and `unowned(safe)`
- cover default and configured behavior
- update generated configuration metadata and changelog

## Verification

- [ ] generated rule tests
- [ ] full Swift tests
- [ ] SwiftLint self-lint
- [ ] remaining platform gates or CI

### DIFF
--- a/Source/SwiftLintBuiltInRules/Rules/Lint/UnownedVariableCaptureRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Lint/UnownedVariableCaptureRule.swift
@@ -3,7 +3,7 @@ import SwiftSyntax
 
 @SwiftSyntaxRule(optIn: true)
 struct UnownedVariableCaptureRule: Rule {
-    var configuration = SeverityConfiguration<Self>(.warning)
+    var configuration = UnownedVariableCaptureConfiguration()
 
     static let description = RuleDescription(
         identifier: "unowned_variable_capture",
@@ -15,6 +15,7 @@ struct UnownedVariableCaptureRule: Rule {
             "foo { [weak self] param in _ }",
             "foo { [weak bar] in _ }",
             "foo { [weak bar] param in _ }",
+            "foo { [unowned(unsafe) self] in _ }",
             "foo { bar in _ }",
             "foo { $0 }",
             """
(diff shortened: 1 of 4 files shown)


Source/SwiftLintBuiltInRules/Rules/Lint/UnownedVariableCaptureRule.swift
```diff
@@ -30,15 +31,24 @@ struct UnownedVariableCaptureRule: Rule {
...
             "foo { [bar, ↓unowned self] in _ }",
+            "foo { [↓unowned(unsafe) self] in _ }"
+                .asExample(configuration: ["include_unsafe": true]),
         ])
     )
 }
 
 private extension UnownedVariableCaptureRule {
     final class Visitor: ViolationsSyntaxVisitor<ConfigurationType> {
         override func visitPost(_ node: TokenSyntax) {
```
SimplyDanny: Let's have a `visitPost(_ node: ClosureCaptureSpecifierSyntax)` instead which allows us to check for `specifier` and `detail` directly.

Source/SwiftLintBuiltInRules/Rules/RuleConfigurations/UnownedVariableCaptureConfiguration.swift
```diff
@@ -0,0 +1,9 @@
+import SwiftLintCore
+
+@AutoConfigParser
+struct UnownedVariableCaptureConfiguration: SeverityBasedRuleConfiguration {
+    @ConfigurationElement(key: "severity")
+    private(set) var severityConfiguration = SeverityConfiguration<Parent>(.warning)
+    @ConfigurationElement(key: "include_unsafe")
+    private(set) var includeUnsafe = false
```
SimplyDanny: Let's call it `allow_explicit_unsafe_unowned`. It should be `false` by default to match the current behavior. It should be opt-in.

SimplyDanny (changes requested):
Thanks @Yurii201811 for taking this!

SimplyDanny (approved):

PR #6825, opened 2026-07-15
Revive excludedPaths by stopping double resolution

### LINKED ISSUE #6795: `excluded` seems to be ignore in 0.65.0
### New Issue Checklist

### Bug Description

In **0.64.1**, `swiftlint` passes.
In **0.65.0**, `swiftlint` reports thousands of violations, all in folders that have been filed under `excluded` in the `swiftlint.yml` file

```bash
$ swiftlint lint
```

### Environment

* SwiftLint version: working in **0.64.1**, not working in **0.65.0**
* Xcode version: 26.6 (17F113)
* Configuration file:

```yml
excluded:
  - DerivedData
  - Pods
```

tomotakatakahashi:
## Summary

This commit stops resolving a path to an absolute path in `Configuration.groupFiles`. This will resolve the issue mentioned in comments in #6795 that the `excluded` configuration in .swiftlint.yml is ignored when file paths are passed as arguments.

## Context
The `excluded` configuration is sometimes ignored since v0.64.0  (https://github.com/realm/SwiftLint/issues/6795#issuecomment-4953871888). The issue seems to happen when file paths are passed as arguments (for example, running`swiftlint Foo.swift Bar.swift` instead of `swiftlint .` or `swiftlint`).

Using a debugger, I found that `excludedRelativePath` in the diff was actually not a relative path, because it had  already been resolved in [Configuration+Parsing.swift](https://github.com/realm/SwiftLint/blob/e4d82f13dcd25583c33caec013b084376db5125e/Source/SwiftLintFramework/Configuration/Configuration%2BParsing.swift#L104). As a result, `let excludedPath` in the diff was being doubly-resolved to an invalid path like `"file:///Users/username/projects/foo/Users/username/projects/foo/MyPackage/Package.swift"`.

## Test
- [x] With [this reproduction setup](https://github.com/realm/SwiftLint/issues/6795#issuecomment-4953871888) and with `Arguments Passed On Launch` `NoProblem.swift MyPackage/Package.swift`, running the swiftlint scheme:
  - [x] Before this PR: Failed
  - [x] After this PR: Succeeded

## Comment
- I couldn't find an appropriate place to put unit tests. Any suggestions would be appreciated.
- I'm new to this codebase. Sorry if something is wrong.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@
   [#6798](https://github.com/realm/SwiftLint/issues/6798)
   [#6799](https://github.com/realm/SwiftLint/issues/6799)
 
+* Fix `excluded` configuration being ignored when file paths are passed as arguments.  
+  [Tomotaka Takahashi](https://github.com/tomotakatakahashi)
+  [#6795](https://github.com/realm/SwiftLint/issues/6795)
+
 ## 0.65.0: Fresh Folded Fixtures
 
 ### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@
   [Hokila](https://github.com/Hokila)
   [#6897](https://github.com/realm/SwiftLint/issues/6897)
 
+* Fix `excluded` configuration being ignored when file paths are passed as arguments.  
+  [Tomotaka Takahashi](https://github.com/tomotakatakahashi)
+  [#6795](https://github.com/realm/SwiftLint/issues/6795)
+
 ## 0.65.1: Fresh Folded Fixtures
 
 ### Breaking
--- a/Source/SwiftLintFramework/Configuration+CommandLine.swift
+++ b/Source/SwiftLintFramework/Configuration+CommandLine.swift
@@ -118,11 +118,9 @@ extension Configuration {
 
         return files.parallelFilterGroup { file in
             let fileConfiguration = configuration(for: file)
-            let fileConfigurationRootPath = fileConfiguration.rootDirectory
 
             // Files whose configuration specifies they should be excluded will be skipped
-            let shouldSkip = fileConfiguration.excludedPaths.contains { excludedRelativePath in
-                let excludedPath = fileConfigurationRootPath.appending(path: excludedRelativePath.relativePath)
+            let shouldSkip = fileConfiguration.excludedPaths.contains { excludedPath in
                 let filePathComponents = file.path?.pathComponents ?? []
                 let excludedPathComponents = excludedPath.pathComponents
                 return filePathComponents.starts(with: excludedPathComponents)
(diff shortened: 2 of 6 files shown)


SimplyDanny (changes requested):
This change is causing changes in scanned OSS repositories. That's definitely unexpected and should be revisited.

tomotakatakahashi:
Thank you for your feedback. I wasn't entirely clear on the comment from SwiftLintBot, but I now understand that I need to address the warnings. I will try to look into this, though my time might be a bit limited.

tomotakatakahashi:
@SimplyDanny After a little investigation, I now think the changes in the scanned OSS repositories are *expected*.

The files listed under "This PR fixed a violation":

1. `/SharedPackages/BrowserServicesKit/scripts/assets/Swift File For Package.xctemplate/___FILEBASENAME___.swift`
2. `/SharedPackages/BrowserServicesKit/Package.swift`
3. `/SharedPackages/BrowserServicesKit/Sources/RemoteMessaging/Model/AnyDecodable.swift`
4. `/macOS/DuckDuckGo/Common/Localizables/UserText.swift`

are all excluded via .swiftlint.yml files ([1, 2, 3](https://github.com/duckduckgo/apple-browsers/blob/0e044c2d1fc0bb386867565071962492cfe9d170/SharedPackages/BrowserServicesKit/.swiftlint.yml#L70), [4](https://github.com/duckduckgo/apple-browsers/blob/0e044c2d1fc0bb386867565071962492cfe9d170/macOS/.swiftlint.yml#L78)) in that repository. So if the main branch were working correctly, these files shouldn't produce any violations at all. Therefore "This PR fixed a violation" messages are desirable and expected.

My guess is that the double-resolution bug was introduced by PR #6636, but *somehow* SwiftLintBot couldn't catch it.

Could you take a look?

SimplyDanny:
I had another look at it, @tomotakatakahashi, and I think you are right and the fix is decent. Thank you and sorry for my doubts! I got so much used to be extra careful these days with so many automated low-quality contributions that I didn't believe in this rather simple fix and the OSS findings appeared to me like a warning signal.

As for testing, there is `ConfigPathResolutionTests` that tries to mimic the resolution logic for parent/child configs. It would be a good place to set up another scenario like we found in the said OSS repositories. For the command-line case, there's not really something similar.

Would you be willing to rebase as well? The builds seem to be stuck for whatever reason. A rebase should fix that.

tomotakatakahashi:
Thank you for your review. Your doubts are completely understandable, especially in the AI era. I appreciate you maintaining SwiftLint.

I've rebased the branch onto the latest main and force-pushed it.

I'll take a look at `ConfigPathResolutionTests` this weekend, hopefully.

tomotakatakahashi:
Sorry for the delay.

While I was adding test cases, Claude Code reported a potential regression. I'm investigating it.

I'll keep working on this, but I may not be very responsive, so feel free to take over this PR if you'd rather not wait.

PR #6826, opened 2026-07-16
Reduce lint hot-path overhead

Brett-Best:
## Summary

This PR reduces lint and correction hot-path overhead across rule execution, syntax classification, file caching, and parallel scheduling. In SwiftLint's OSS-check suite, the combined series improves end-to-end lint time by 9–35%.

- move the remaining lint-mode rules and SwiftSyntax-mode custom rules off SourceKit by deriving their structure and syntax-kind information from SwiftSyntax
- skip syntax-tree rewriting for correctable rules when a file has no violations
- replace repeated full-tree and classification work with targeted scans, cheap pre-checks, and per-file caches
- remove repeated per-token, per-node, per-rule, and per-file invariant work across frequently executed rules
- reduce cache synchronization and violation post-processing overhead
- use file-level parallelism more effectively by running rules serially per file when appropriate and dispatching the largest files first
- fix sorted-index and SourceKit-compatible comment-range behavior

## Motivation

Profiling a large repository showed that a substantial portion of runtime was spent outside rule-specific analysis: SourceKit `editorOpen` requests, repeated syntax-tree walks, classification of unrelated tokens, full-tree correction rewrites on clean files, filesystem and cache synchronization, repeated source-line and term-list construction, task-local bookkeeping, violation-region partitioning, and concurrent walks contending on the same syntax arena.

Most of this work is invariant for a rule type or file, can be cached, or can be avoided entirely after a cheap candidate check. This series moves invariant work out of repeated paths and makes expensive analysis conditional on a file having relevant content.

### Remove unnecessary SourceKit work

- port `indentation_width`, `literal_expression_end_indentation`, `multiline_function_chains`, `multiline_parameters_brackets`, `statement_position`, and the vertical-whitespace brace rules to SwiftSyntax-backed implementations
- add a shared SwiftSyntax-derived syntax map for rules that still need syntax-kind filtering
- execute custom rules configured for SwiftSyntax mode without consulting SourceKit, including inherited `default_execution_mode`
- preserve SourceKit-compatible line-comment ranges for LF, CRLF, and CR line endings without extending block comments

### Avoid unnecessary correction and rule work

- validate correctable rules first and skip rewriter construction and serialization when a file has no violations
- skip adjusted body-line analysis when the physical body span is already within the configured threshold
- gate `comment_spacing` regex work behind a cheap UTF-16 scan and match directly against the file's string view
- hoist per-file invariants and add cheap preconditions in `colon`, `comma`, `closure_end_indentation`, and `redundant_self`
- skip violation-region partitioning and identifier-set construction when no violations exist or the file has one enable-everywhere region
- reuse the rule identifier already resolved by `lint` instead of copying its `RuleDescription` again in `performLint`

### Cache and narrow repeated analysis

- replace the per-file artifact cache's concurrent queue with an `NSLock` while retaining computation deduplication
- cache comment byte ranges per file and compute them with one position-tracking visitor pass
- speed up `CommentLinesVisitor` by avoiding empty-trivia re-lexing, reversed trivia copies, repeated token-kind checks, and per-token converter lookups
- collect `indentation_width` syntax data in one tree walk
- cache source lines once per file for `vertical_parameter_alignment`
- precompute `inclusive_language` term lists once per file
- cache `allowedSymbolsAndAlphanumerics`, minimum-Swift-version support, environment state, file emptiness, and targeted file metadata
- avoid general syntax classification and full-tree traversal where a token, root-node, line, or regex-candidate scan is sufficient
- memoize nested function-chain suffixes to avoid quadratic re-walking

### Improve scheduling

- skip collection dispatch when no collecting rule is enabled
- when the number of files is at least the active processor count, run rules serially per file and rely on file-level parallelism; few-file invocations retain per-file rule parallelism
- sort parallel lint work by descending file size so expensive files start early instead of leaving one core processing the largest file at the end

## Correctness fixes

- make `RandomAccessCollection.firstIndexAssumingSorted(where:)` return `nil`, rather than `endIndex`, when no element matches
- honor custom-rule `execution_mode` and inherited `default_execution_mode`
- extend bridged line-comment ranges correctly across LF, CRLF, and CR endings while keeping block-comment extents unchanged
- invalidate cached comment byte ranges with the rest of a file's cached artifacts

## Impact

Large multi-file lint runs should spend less CPU time in SourceKit, syntax-tree bookkeeping, filesystem access, cache synchronization, rule post-processing, and shared-arena contention. Clean `--fix` runs avoid reconstructing the syntax tree for every correctable rule. Single-file and editor integrations retain concurrent rule execution. No configuration migration is required, and SourceKit execution remains available where explicitly selected or required.

## Validation

- the earlier 13-commit tree passed 101 focused Swift tests covering the SourceKit-free ports, custom-rule modes, comment/line handling, file caching, collecting behavior, and affected built-in rules
- the 15-commit follow-up applied cleanly and was pushed without waiting for another test run; CI provides validation for the final tree
- metadata rewriting preserved the applied tree, and all 28 commits use `Brett-Best <brettbest@live.com.au>` as author and committer

`git diff --check` reports only the two intentional Markdown hard breaks in this PR's changelog entries.


Brett-Best:
Some other benchmarks which also include the fix https://github.com/jpsim/SourceKitten/pull/846
```
Linting Aerial with this PR took 0.29 s vs 0.43 s on `main` (32% faster).
Linting Alamofire with this PR took 0.43 s vs 0.58 s on `main` (25% faster).
Linting Brave with this PR took 2.51 s vs 3.76 s on `main` (33% faster).
Linting DuckDuckGo with this PR took 10.73 s vs 13.83 s on `main` (22% faster).
Linting <<work project>> with this PR took 3.97 s vs 7.52 s on `main` (47% faster).
Linting Firefox with this PR took 3.99 s vs 5.59 s on `main` (28% faster).
Linting Kickstarter with this PR took 2.77 s vs 4.47 s on `main` (38% faster).
Linting Moya with this PR took 0.19 s vs 0.28 s on `main` (32% faster).
Linting NetNewsWire with this PR took 0.98 s vs 1.3 s on `main` (24% faster).
Linting Nimble with this PR took 0.24 s vs 0.32 s on `main` (25% faster).
Linting PocketCasts with this PR took 2.43 s vs 6.02 s on `main` (59% faster).
Linting Quick with this PR took 0.17 s vs 0.23 s on `main` (26% faster).
Linting Realm with this PR took 1.31 s vs 2.58 s on `main` (49% faster).
Linting Sourcery with this PR took 0.63 s vs 1.16 s on `main` (45% faster).
Linting Swift with this PR took 1.64 s vs 2.39 s on `main` (31% faster).
Linting SwiftLintPerformanceTests with this PR took 0.13 s vs 0.14 s on `main` (7% faster).
Linting VLC with this PR took 0.51 s vs 0.69 s on `main` (26% faster).
Linting Wire with this PR took 5.6 s vs 8.17 s on `main` (31% faster).
Linting WordPress with this PR took 3.72 s vs 5.16 s on `main` (27% faster).
```

PR #6830, opened 2026-07-21
Fix closure_end_indentation corrupting CRLF files

### LINKED ISSUE #6598: closure_end_indentation rule corrupts files with CRLF line endings
## Description

  The `closure_end_indentation` rule corrupts Swift files that have CRLF line endings (`\r\n`) instead
  of Unix LF line endings (`\n`).

  ## Environment

  - SwiftLint version: 0.63.2
  - macOS: Tahoe (Darwin 25.1.0)
  - Installation method: Homebrew

  ## Steps to Reproduce

  1. Create a simple SwiftUI file with CRLF line endings:

  ```swift
  import SwiftUI

  struct TestView: View {
      var body: some View {
          HStack {
              VStack {
                  Text("Hello")
                  Spacer()
              }
          }
      }
  }
  ```

  2. Convert the file to CRLF line endings (e.g., using `unix2dos` or saving with Windows line endings)

  3. Run `swiftlint --fix` on the file

  ## Expected Behavior

  The file should remain properly formatted, or SwiftLint should report an error about the line
  endings.

  ## Actual Behavior

  The `closure_end_indentation` rule moves closing braces to incorrect positions, corrupting the file:

  ```swift
  import SwiftUI

  struct TestView: View {
      var body: some View {
          HStack {
              VStack {
                  Text("Hello")
                  Spacer()
  }
  }
      }
  }
  ```

  ## Additional Context

  - The issue does not occur with Unix LF line endings
  - Converting the file to LF line endings (`sed -i '' 's/\r$//' file.swift`) fixes the issue
  - The corruption gets worse with more deeply nested SwiftUI view builders
  - Other rules like `trailing_whitespace` and `comma` work correctly with CRLF files

  ## Workaround

  Convert files to Unix line endings before running SwiftLint:

  ```bash
  sed -i '' 's/\r$//' file.swift
  ```

sjh9714:
## Summary

Fixes #6598.

Running `swiftlint --fix` with `closure_end_indentation` on a file with CRLF
(`\r\n`) line endings moved the closing braces of multi-line closures to
column 1, corrupting the file. LF files were unaffected.

The rule computed the expected indentation by indexing SourceKitten's
`file.lines` with a line number produced by SwiftSyntax's
`locationConverter`. The two disagree on CRLF: SwiftSyntax counts `\r\n` as
one newline, while `StringView` splits it into `\n` and `\r` separately,
inserting a phantom empty line after every real line. On CRLF files the
anchor-line lookup therefore hit the wrong (often empty) line, making the
expected indentation column 0.

## Changes

- `ClosureEndIndentationRule.Visitor.getFirstNonWhitespaceColumn(onLine:)`
  now reads the anchor line from `locationConverter.sourceLines`, the same
  line-numbering domain that produced the line number. No behavior change for
  LF files.
- Regression test with CRLF examples: a correctly indented closure no longer
  triggers, and a genuinely misindented closing brace is corrected to the
  right column while CRLF endings are preserved. (`testWrappingInString` is
  disabled for the triggering example because the test helper's string
  wrapping escapes only `\n`, leaving bare `\r` bytes in the literal.)
- CHANGELOG entry.

## Testing

- `swift test --filter ClosureEndIndentationRule` — new CRLF test fails
  before the fix (closing braces pulled to column 1) and passes after;
  the generated default-examples suite also passes.
- `swift test` — full suite locally (1063 tests), except `MacroTests`, which
  cannot run in this environment (no XCTest under Command Line Tools). One
  unrelated suite (`UnusedDeclarationRuleGeneratedTests`) fails identically
  with and without this change in my environment.
- Manual check of the reproduction from #6598: on the corrupted file the fix
  restores the correct indentation; on the correctly indented CRLF file
  `swiftlint --fix` makes no changes and `swiftlint` reports 0 violations,
  with CRLF endings preserved throughout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,11 @@
   [#6798](https://github.com/realm/SwiftLint/issues/6798)
   [#6799](https://github.com/realm/SwiftLint/issues/6799)
 
+* Fix `closure_end_indentation` moving closing braces to column 1 in files
+  with CRLF line endings when running `swiftlint --fix`.  
+  [sjh9714](https://github.com/sjh9714)
+  [#6598](https://github.com/realm/SwiftLint/issues/6598)
+
 ## 0.65.0: Fresh Folded Fixtures
 
 ### Breaking
--- a/Source/SwiftLintBuiltInRules/Rules/Style/ClosureEndIndentationRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Style/ClosureEndIndentationRule.swift
@@ -195,10 +195,13 @@ private extension ClosureEndIndentationRule {
 
         /// Calculates the column of the first non-whitespace character on a given line.
         private func getFirstNonWhitespaceColumn(onLine lineNumber: Int) -> Int {
-            guard lineNumber > 0, lineNumber <= file.lines.count else {
+            // Use the lines seen by `locationConverter`, which produced `lineNumber`.
+            // `file.lines` counts CRLF line endings differently, corrupting files.
+            let sourceLines = locationConverter.sourceLines
+            guard lineNumber > 0, lineNumber <= sourceLines.count else {
                 return 1 // Should not happen
             }
-            let lineContent = file.lines[lineNumber - 1].content
+            let lineContent = sourceLines[lineNumber - 1]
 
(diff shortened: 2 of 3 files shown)


SimplyDanny (approved):
LGTM! Thanks @sjh9714.

PR #6841, opened 2026-07-28
Don't flag color components of `Color` and `NSColor` initializers

lechuckcaptain:
## Summary

- `no_magic_numbers` exempts the color-component arguments of a `UIColor` initializer, but the
  check hardcoded that one type name, so SwiftUI's `Color` and AppKit's `NSColor` got no exemption
  at all. All three are now recognised, and the `opacity` label is accepted alongside `alpha`.
- AppKit encodes the color space in the label of the first component, so recognising the type name
  alone was not enough: `NSColor(srgbRed: 0.1, green: 0.42, blue: 0.7, alpha: 1)` would have had
  its last three components exempted and the first still reported. The `calibratedRed`,
  `deviceRed`, `srgbRed`, `calibratedWhite`, `deviceWhite`, `genericGamma22White`, `calibratedHue`
  and `deviceHue` variants are covered too, as are the CMYK components of
  `deviceCyan:magenta:yellow:black:alpha:`.
- The exemption also required the literal to be the *immediate* argument, so `UIColor(red: 0.6, …)`
  was exempt while `UIColor(red: 0x19 / 255, …)` was not. The enclosing argument is now found by
  walking up through arithmetic, parentheses and numeric conversions, bailing out on anything else
  so an exemption cannot leak past the argument it was granted for.
Every literal is still judged by its own label, so an unrecognised label is exempted for nothing but
itself: in Wire's `UIColor(rgb: 0x33373A, alpha: 0.16)` the `rgb:` value stays reported while
`alpha:` is exempt, since `0.16` is a color component whatever its sibling happens to be. There is
a triggering example pinning that.

## Validation

- `swift test --filter NoMagicNumbersRuleGeneratedTests` passes, as do the `RuleDocumentation`,
  `DefaultRuleConfiguration`, `RuleNameConsistency` and `Example` suites. No existing example was
  edited, reordered or removed; the manual bit-packing example `Int(r * 255) << 16` and
  `Color.primary.opacity(isAnimate ? 0.1 : 1.5)` still trigger.
- `swift run swiftlint --strict` clean on the changed file.
- OSS-check: 861 false positives removed across 18 public projects, 0 introduced.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,13 @@
   `unowned_variable_capture` rule accept explicit `unowned(unsafe)` captures.  
   [Yurii Bakurov](https://github.com/Yurii201811)
   [#6817](https://github.com/realm/SwiftLint/issues/6817)
+* Fix `no_magic_numbers` reporting color-component arguments of SwiftUI `Color`
+  and AppKit `NSColor` initializers, which are now exempt as `UIColor` already
+  was, including the color-space-prefixed components of `NSColor` as in
+  `NSColor(srgbRed:green:blue:alpha:)`. The exemption further covers computed
+  component values as in
+  `Color(red: 0x19 / 255, green: 0x7A / 255, blue: 0x3C / 255)`.  
+  [lechuckcaptain](https://github.com/lechuckcaptain)
 
 * Fix baseline writing to store file locations as paths relative to the current working directory,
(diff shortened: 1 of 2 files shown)


lechuckcaptain:
Closing in favour of a smaller series. This PR bundled three separate changes; I have split them so each can be reviewed on its own, starting with the type-name fix. Filed #6869 for the bug and will open the first PR against it shortly.

The branch stays, so nothing here is lost.

PR #6842, opened 2026-07-28
Don't flag typed factory constants in `no_magic_numbers`

lechuckcaptain:
> [!NOTE]
> Stacked on #6841. This branch is based on that one, so the diff below includes its commit too —
> the second commit is the change described here. Happy to rebase onto `main` if you would rather
> review them independently, or if #6841 is not wanted.

## Summary

- `no_magic_numbers` exempts a literal bound directly to a name, so `let x: Int = 5` is clean while
  `let x: Duration = .seconds(5)` — the same constant, better typed — is flagged: the factory call
  breaks the grandparent chain `isMagicNumber` looks for. The rule effectively penalises the
  stronger API.
- A literal is now exempt when it is the sole argument of a factory call forming the whole
  initializer of a declaration whose declared type is listed in the new `definitional_types`
  option, defaulting to `Duration` and `Angle`.
- The type is read from the syntax — a member-access base (`Duration.seconds(5)`) or the
  declaration's type annotation (`let x: Duration = .seconds(5)`) — since the rule has no type
  information. That is the trade-off `test_parent_classes` already makes for superclass names; a
  `typealias` is not followed.

Gating on the declared type rather than on the call shape is what keeps arbitrary single-argument
factories flagged: in `let x: Int = .factorial(20)` and `let d: Data = .randomBytes(64)` the name
labels the result of a computation, not a unit. Requiring the call to *be* the initializer value
keeps use sites such as `clock.sleep(for: .seconds(30))` flagged.

`Measurement` is deliberately absent from the default. Every construction is either
`.init(value:unit:)` or `Measurement(value:unit:)`, excluded by the `init` check and the
sole-argument requirement respectively, so listing it would be a no-op.

## Validation

- `swift test --filter NoMagicNumbersRuleGeneratedTests` passes. No existing example was edited,
  reordered or removed; the new option is covered by `configuration:` examples.
- `swift run swiftlint-dev rules register` run, so `default_rule_configurations.yml` is up to date.
- `swift run swiftlint` clean on the changed files.
- Sweep of a real 341-file package tree against a build of the base commit: 141 → 137 violations,
  4 removed, 0 introduced. The four are three `Duration` constants and one `Angle.degrees(90)`.

Note for adopters: because this pattern is usually silenced with a disable comment today, projects
on `--strict` may see `superfluous_disable_command` fire after upgrading where a block existed only
for this. That is the intended payoff, but the upgrade is not silent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,14 @@
   with CRLF line endings when running `swiftlint --fix`.  
   [sjh9714](https://github.com/sjh9714)
   [#6598](https://github.com/realm/SwiftLint/issues/6598)
+* Don't flag a numeric literal in the `no_magic_numbers` rule when it is the sole argument of a
+  factory call that forms the whole initializer of a named constant whose declared type is listed
+  in the new `definitional_types` option, defaulting to `Duration` and `Angle`. This treats
+  `static let interval: Duration = .seconds(5)` as definitional just like
+  `static let intervalSeconds: Int = 5` already was, while arbitrary factories such as
+  `let x: Int = .factorial(20)` and inline use sites such as `clock.sleep(for: .seconds(30))`
+  keep triggering.  
+  [lechuckcaptain](https://github.com/lechuckcaptain)
 
 ## 0.65.0: Fresh Folded Fixtures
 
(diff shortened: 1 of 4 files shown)


lechuckcaptain:
Closing for now. This was stacked on #6841, which I have closed in favour of a smaller series, so the diff here was misleading.

The change itself is still worth doing. I will raise it as a proposal issue first to agree the option name and default before reopening a PR, rather than leaving a draft sitting here.

PR #6843, opened 2026-07-31
Fix line_length URL detection for property accesses

### LINKED ISSUE #6811: line_length: ignores_urls strips property accesses like `post.id` (NSDataDetector matches bare domains with valid TLDs)
# New Issue Checklist

# Bug Report

`line_length` with `ignores_urls: true` silently skips lines containing ordinary property accesses like `post.id` or `user.name`, because `NSDataDetector` recognizes them as links (`.id`, `.name`, `.info`, … are valid TLDs).

### Complete output when running SwiftLint, including the stack trace and command used

```bash
$ cat Repro.swift
```
```swift
// Both lines are exactly 124 characters long. Only the second one is reported.
        guard let response = try? await qoqaService2.performRequest(GetUserBookmarkRequest(postId: post.id)) else { return }
        guard let response = try? await qoqaService2.performRequest(GetUserBookmarkRequest(postId: postXid)) else { return }
```
```bash
$ cat .swiftlint.yml
line_length:
  ignores_urls: true

$ swiftlint lint --no-cache Repro.swift
Repro.swift:3:1: warning: Line Length Violation: Line should be 120 characters or less; currently it has 124 characters (line_length)
```

Line 2 (`post.id`) produces no violation; line 3 (`postXid`), identical in length, is flagged.

## Environment

* SwiftLint version: 0.65.0
* Xcode version: 26.x
* Configuration file: see above (`ignores_urls: true` is the only relevant option)

### Root cause

`String.strippingURLs` in `LineLengthRule.swift` uses `NSDataDetector(types: .link)` and removes **every** match before measuring the line:

https://github.com/realm/SwiftLint/blob/main/Source/SwiftLintBuiltInRules/Rules/Metrics/LineLengthRule.swift

`NSDataDetector` intentionally matches bare domains such as `example.com` — but in Swift source, member accesses like `post.id`, `user.name` or `offer.info` have the same shape (`identifier.validTLD`), so they are stripped too. In a codebase using `ignores_urls: true`, any over-long line that happens to contain such an expression passes the rule.

### Possible fix

Only strip matches that look like actual URLs, e.g. require a scheme (`match.url?.scheme != nil` and not the implicit `http` NSDataDetector adds to bare domains), or require the match to contain `://` / start with `www.`. The Linux regex path already roughly does this by requiring a scheme, `www.`, or a trailing `/`.

thliu21:
## Summary

- only strip `NSDataDetector` matches that have explicit URL syntax
- keep scheme, `www.`, and domain-with-path URLs aligned with the Linux/Windows behavior
- add regressions for property accesses whose member names are valid top-level domains

Fixes #6811

## Testing

- `swift test --filter LineLengthRuleTests.lineLengthWithIgnoreURLsEnabled`
- `swift test --filter LineLength`
- `swift test --quiet`
- strict SwiftLint on the two changed Swift files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,11 @@
   [Yurii Bakurov](https://github.com/Yurii201811)
   [#6817](https://github.com/realm/SwiftLint/issues/6817)
 
+* Fix `line_length` with `ignores_urls` enabled incorrectly ignoring property
+  accesses whose member names are valid top-level domains.  
+  [Arthur Liu](https://github.com/thliu21)
+  [#6811](https://github.com/realm/SwiftLint/issues/6811)
+
 * Fix baseline writing to store file locations as paths relative to the current working directory,
   restoring baseline portability and avoiding absolute `file://` paths in generated baseline files.  
   [SimplyDanny](https://github.com/SimplyDanny)
--- a/Source/SwiftLintBuiltInRules/Rules/Metrics/LineLengthRule.swift
+++ b/Source/SwiftLintBuiltInRules/Rules/Metrics/LineLengthRule.swift
@@ -252,7 +252,22 @@ private extension String {
             guard let urlDetector = try? NSDataDetector(types: types) else {
                 return self
             }
-            return urlDetector.stringByReplacingMatches(in: self, options: [], range: range, withTemplate: "")
+            var result = self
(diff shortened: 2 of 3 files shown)


SimplyDanny (approved):
LGTM. Thanks @thliu21!

PR #6851, opened 2026-08-01
Reduce per-file filesystem and gating overhead

Brett-Best:
Linting each file currently does two avoidable pieces of fixed work:

- cache validation obtains the complete file attribute dictionary, including extended attributes, just to read modification date;
- rule gating repeatedly compares each rule's minimum Swift version and can repeatedly acquire the file-contents synchronization path just to test emptiness.

This asks the URL for only `contentModificationDateKey`. Swift-version eligibility is computed once per shared configuration and carried alongside each selected rule. Before reading `file.isEmpty`, the linter checks whether any version-compatible rule requires empty-file gating, so the value is computed zero or one time per file. Autocorrect keeps the read lazy behind the same version and empty-file gates.

### Verification

- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer swift test`: passed.
- `swift test --filter EmptyFileTests`: 3 tests passed, including a deferred nonexistent-file regression proving an unsupported rule does not read the file during autocorrect.
- `make bazel_test_tsan`: 18/18 targets passed; the affected `//Tests:FrameworkTests` target was rerun after adding the regression test.
- `.build/debug/swiftlint lint --strict --quiet --no-cache` on all four changed files: clean.
  - DuckDuckGo, default configuration: 84 before, 84 after.
  - realm-swift with `--enable-all-rules`: 64,834 before, 64,834 after.
  - Cache-enabled parity also matched: 84/84 on DuckDuckGo and 64,834/64,834 on realm-swift.
- Comparisons sorted by file, line, character, rule, reason, and severity; raw JSON order is not stable for parallel linting.

### Measurement

`--only-rule trailing_newline` on DuckDuckGo (6,747 linted files), chosen so that fixed per-file
cost dominates rather than rule work. Both binaries interleaved, best of three, on an otherwise
idle machine:

| | baseline | change |
| --- | ---: | ---: |
| warm cache — cache validation runs for every file | 1.01s | 0.92s |
| empty cache on each run | 2.47s | 2.21s |

```bash
swiftlint lint --quiet --only-rule trailing_newline --cache-path <dir>
```

The warm-cache row is the one that exercises the attribute change directly, since validating a
cached entry is what reads the modification date.

A separate seven-run `--no-cache` check of the version-aware gating revision was neutral:
2.125s before versus 2.126s after. For a full `--no-cache` lint the end-to-end difference is
within machine noise and is not presented here as a speedup.

### DIFF
--- a/Source/SwiftLintFramework/Extensions/FileManager+SwiftLint.swift
+++ b/Source/SwiftLintFramework/Extensions/FileManager+SwiftLint.swift
@@ -133,6 +133,6 @@ extension FileManager: LintableFileManager {
     }
 
     public func modificationDate(forFileAtPath path: URL) -> Date? {
-        (try? attributesOfItem(atPath: path.filepath))?[.modificationDate] as? Date
+        try? path.resourceValues(forKeys: [.contentModificationDateKey]).contentModificationDate
     }
 }
--- a/Source/SwiftLintFramework/Models/Linter.swift
+++ b/Source/SwiftLintFramework/Models/Linter.swift
@@ -18,6 +18,11 @@ private struct LintResult {
     let deprecatedToValidIDPairs: [(String, String)]
 }
 
+private struct LinterRule: Sendable {
+    let rule: any Rule
+    let supportsCurrentSwiftVersion: Bool
+}
+
 private extension Rule {
     func superfluousDisableCommandViolations(regions: [Region],
                                              superfluousDisableCommandRule: SuperfluousDisableCommandRule?,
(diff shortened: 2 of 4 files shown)


Source/SwiftLintFramework/Extensions/FileManager+SwiftLint.swift
```diff
@@ -133,6 +133,9 @@ extension FileManager: LintableFileManager {
     }
 
     public func modificationDate(forFileAtPath path: URL) -> Date? {
-        (try? attributesOfItem(atPath: path.filepath))?[.modificationDate] as? Date
+        // `attributesOfItem` assembles the full attribute dictionary including an enumeration of
```
SimplyDanny: The comment might be more confusion without the context of the replaced line.

Source/SwiftLintFramework/Models/Linter.swift
```diff
@@ -68,13 +68,19 @@ private extension Rule {
     }
 
     func shouldRun(onFile file: SwiftLintFile) -> Bool {
+        shouldRun(onFile: file, fileIsEmpty: file.isEmpty)
+    }
```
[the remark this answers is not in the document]
SimplyDanny: Valid one.

CHANGELOG.md
```diff
@@ -14,6 +14,12 @@
 
 ### Enhancements
 
+* Reduce fixed per-file linting overhead: read a file's modification date for
+  cache validation via a single URL resource value instead of building the full
+  file attribute dictionary, and compute whether a file is empty once per file
+  rather than once per rule.
```
SimplyDanny: Not quite clear/relevant to users.

SimplyDanny (changes requested):

Source/SwiftLintFramework/Models/Linter.swift
```diff
@@ -98,6 +104,7 @@ private extension Rule {
     // As we need the configuration to get custom identifiers.
     // swiftlint:disable:next function_parameter_count
     func lint(file: SwiftLintFile,
+              fileIsEmpty: Bool,
```
SimplyDanny, line 115: Should also be a closure.

Source/SwiftLintFramework/Models/Linter.swift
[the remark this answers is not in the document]
Brett-Best: Still seeing if this can actually be addressed, will move to draft temporarily.

PR #6872, opened 2026-08-15
Fix baseline duplicate replacements

### LINKED ISSUE #6871: Baseline: a same-count swap of duplicate-text violations within a rule is invisible
Hi, an observation about the baseline filter (Source/SwiftLintCore/Models/Baseline.swift), not a bug report. It comes from a formal result on separation criteria I deposited recently.

A baseline entry is matched first by exact identity (rule, file, line, character, reason, and the source line's text), and any exact match is removed from both sides before anything else happens. What's left is grouped by rule, and within a rule, grouped again by a key that's just the source line's text plus the reason string, with no location in it. For each key, if the current count of violations sharing that key is greater than the baseline's count for that key, all of them get reported; if it's less than or equal, none do.

That second part means two violations of the same rule with the exact same text and the exact same reason, at different locations, are indistinguishable once they're grouped, because the key drops the location. I reimplemented the algorithm from source (a small script, not the real binary) and ran this: a baseline with two identical TODO violations at line 10 and line 20 (same text, same reason). Then I fix line 10 for real, and a genuinely new TODO with the identical text and reason shows up at line 30, unrelated to either original. The current violations, after removing exact matches, are just line 20 (unchanged) and line 30 (new). Grouped by key they're both in the same bucket as the one leftover baseline entry (line 10's, now unmatched), count 1 vs count 1, so nothing gets reported. The new line-30 violation disappears.

This needs duplicate text and duplicate reason to collide, so it won't happen for rules whose reason includes specifics (a value, a name), but it will for rules with a fixed, generic reason and short repeated lines. todo is the obvious one, and I'd guess mark, orphaned_doc_comment, and some of the pattern-matching rules on very short or common expressions have the same shape.

Would it make sense to include location, or at least something coarser like file, when deciding whether two same-key violations are the same reported instance, instead of relying on key count alone once the exact-identity pass has already run? I know that's basically re-adding what the initial exact-match pass already checks, but the count-only fallback for the collision case is what loses it.

Reference, if useful: Closure Trivialization in Typed Separation Systems, https://doi.org/10.5281/zenodo.21908527, section 10.

LizunovSergey:
Fixes #6871

## Summary
- report a same-count replacement of identical baseline violations when another duplicate still matches exactly
- preserve the existing behavior for uniformly shifted duplicate violations

## Root cause
After exact matches are removed, the fallback comparison used only the number of remaining violations with the same text and reason. A replacement could therefore be hidden by an unmatched baseline duplicate with the same key.

An unchanged duplicate now acts as an anchor: if the remaining counts are equal within that group, the unmatched current violations are reported. A group with no exact match still follows the existing shift-tolerant behavior.

## Tests
- `swift test --filter BaselineTests`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,10 @@
 
 ### Bug Fixes
 
+* Fix baseline duplicate replacements.  
+  [LizunovSergey](https://github.com/LizunovSergey)
+  [#6871](https://github.com/realm/SwiftLint/issues/6871)
+
 * Add an opt-in `allow_explicit_unsafe_unowned` option to let the
   `unowned_variable_capture` rule accept explicit `unowned(unsafe)` captures.  
   [Yurii Bakurov](https://github.com/Yurii201811)
--- a/Source/SwiftLintCore/Models/Baseline.swift
+++ b/Source/SwiftLintCore/Models/Baseline.swift
@@ -156,6 +156,8 @@ public struct Baseline: Equatable {
             return []
         }
 
+        let exactViolations = Set(relativePathViolations).intersection(baselineViolations)
+
         let violationsByRuleIdentifier = relativePathViolations.groupedByRuleIdentifier(
             filteredBy: baselineViolations
         )
@@ -181,6 +183,11 @@ public struct Baseline: Equatable {
                     continue
                 }
                 if ruleViolations.count > baselineViolations.count {
+                    filteredViolations.formUnion(ruleViolations)
+                } else if ruleViolations.count == baselineViolations.count,
+                          exactViolations.contains(where: {
+                              $0.violation.ruleIdentifier == ruleIdentifier && $0.key == key
+                          }) {
                     filteredViolations.formUnion(ruleViolations)
                 }
             }
--- a/Tests/FileSystemAccessTests/BaselineTests.swift
+++ b/Tests/FileSystemAccessTests/BaselineTests.swift
@@ -50,6 +50,21 @@ struct BaselineTests {
         Baseline(violations: ruleDescriptions.violations(for: filePath))
     }
 
+    private static func lines(withDuplicatesAt duplicateLines: Set<Int>) -> [String] {
+        (1...30).map { line in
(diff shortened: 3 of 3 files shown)


yunusdim:
Author of #6871. The anchor approach is the right call: an unchanged duplicate is the evidence that the group did not shift, which is exactly the distinction the old fallback could not make. Keeping the shift-tolerant path where no exact match survives leaves the intentional behaviour intact.

One residual worth stating, not a request to change anything. When every duplicate in a group is replaced in the same run, no anchor survives, the group falls through to the shift-tolerant path, and the swap stays invisible. That is narrower than what #6871 reported — it needs the whole group to turn over at once — but it is the same shape, and it is better known now than discovered later.

Reason and line text remain the only keys on that path, so the residual is bounded by how generic those are. `todo`, `mark` and `orphaned_doc_comment` are where it would show.

Checks are green apart from the ones waiting on maintainer approval.

LizunovSergey:
Thanks — and the residual is real. Worth stating why it stays open rather than getting folded in here.

The anchor is the only evidence available. Reason and line text are the whole key on that path, so a
group where every duplicate turned over in one run looks byte-for-byte identical to the same group
having shifted: same rule, same texts, same count, nothing unchanged to pin it to. There is no
signal left to separate the two. Treating a same-count, zero-anchor group as new would report every
shifted group as a fresh violation, which is exactly the false positive the shift-tolerant path
exists to avoid — a strictly worse trade than the miss it would close.

So it is bounded the way you describe: it needs the whole group to turn over at once, and it only
bites where the key is generic enough for a group to exist at all — `todo`, `mark`,
`orphaned_doc_comment`. Closing it properly needs a stronger key than reason plus line text, which
is a different change from this one.

yunusdim:
Agreed, and that is the right line to draw. A key stronger than reason plus line text is a different change from this one; folding it in here would report every shifted group as new, which is the false positive the shift-tolerant path exists to avoid. This PR closes #6871 as reported. The whole-group-turnover case is worth its own issue against a stronger key, not a rider on this one.

PR #6885, opened 2026-08-22
Fix legacy_swiftui_aspect_ratio not preserving whitespace (#6883)

### LINKED ISSUE #6883: Applied fix for `legacy_swiftui_aspect_ratio` violations doesn't preserve newlines
### New Issue Checklist

### Bug Description

When running `swiftlint lint --fix` on the following

```swift
someView
    .tint(.red)
    .aspectRatio(contentMode: .fit)
```

The expected behaviour is that it would keep the view modifier on it's own line

```swift
someView
    .tint(.red)
    .scaledToFit()
```

The actual behaviour is that it removes the leading whitespace/newline so it's moved onto the previous line, potentially not following the surrounding view modifiers.

```swift
someView
    .tint(.red).scaledToFit()
```

It can be reproduced with the following.

```bash
echo "someView\n.tint(.red)\n.aspectRatio(contentMode: .fit)" | swiftlint lint --use-stdin --only-rule legacy_swiftui_aspect_ratio --fix ;
```

### Environment

* SwiftLint version (run `swiftlint version` to be sure) 0.65.1
* Xcode version (run `xcodebuild -version` to be sure) Xcode 26.5 Build version 17F42
* Installation method used (Homebrew, CocoaPods, building from source, etc) brew and mist
* Configuration file: default

CraigSiemens:
Fixes #6883 by editing the `baseName` of the `MemberAccessExprSyntax` to preserve the existing trivia (whitespace) around the function call.

Previously it was creating a new expression for the function call so the result would always be on one line.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 
 * Keep the formatting of multiline member chains when correcting
   `legacy_swiftui_aspect_ratio`. The trivia between the base expression and
-  `.aspectRatio` is no longer collapsed onto a single line.
+  `.aspectRatio` is no longer collapsed onto a single line.  
   [CraigSiemens](https://github.com/CraigSiemens)
 
 ## 0.65.1: Fresh Folded Fixtures
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,10 @@
 
 ### Bug Fixes
 
-* None.
+* Keep the formatting of multiline member chains when correcting
+  `legacy_swiftui_aspect_ratio`. The trivia between the base expression and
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -16,7 +16,10 @@
 
 ### Bug Fixes
 
-* None.
+* Keep the formatting of multiline member chains when correcting
+  `legacy_swiftui_aspect_ratio`. The trivia between the base expression and
+  `.aspectRatio` is no longer collapsed onto a single line.
```
```suggestion
  `.aspectRatio` is no longer collapsed onto a single line.  
```

SimplyDanny (approved):
Thanks @CraigSiemens!

PR #6893, opened 2026-08-26
Add `#example` macro allowing for compiled example code

SimplyDanny:
This adds an `#example` (singular) macro that accepts a single trailing closure body which describes the example to test. `/*>*/` marks the position where a violation is expected to trigger.

By having the example code compiled, we can make sure that they describe valid Swift code avoiding examples only passing tests because of a parsing issue.

### DIFF
--- a/Source/SwiftLintCoreMacros/Example.swift
+++ b/Source/SwiftLintCoreMacros/Example.swift
@@ -1,0 +1,78 @@
+import SwiftBasicFormat
+import SwiftLintBase
+import SwiftSyntax
+import SwiftSyntaxBuilder
+import SwiftSyntaxMacros
+
+// swiftlint:disable:next blanket_disable_command
+// swiftlint:disable fatal_error
+
+struct Example: ExpressionMacro {
+    static func expansion(
+        of node: some FreestandingMacroExpansionSyntax,
+        in context: some MacroExpansionContext
+    ) -> ExprSyntax {
+        guard let fileID = context.location(of: node, at: .afterLeadingTrivia, filePathMode: .fileID),
+              let filePath = context.location(of: node, at: .afterLeadingTrivia, filePathMode: .filePath) else {
+            context.diagnose(SwiftLintCoreMacroError.invalidSourceLocation.diagnose(at: node))
+            fatalError(SwiftLintCoreMacroError.invalidSourceLocation.message)
+        }
+
+        guard let trailingClosure = node.trailingClosure else {
+            context.diagnose(SwiftLintCoreMacroError.missingExampleBody.diagnose(at: node))
+            fatalError(SwiftLintCoreMacroError.missingExampleBody.message)
+        }
+
+        let spacesCount = trailingClosure.rightBrace.leadingTrivia.countSpaces
+        let example = """
+            Example(
+                code: \"\"\"\(trailingClosure.statements.asExampleBody(unindentedBy: spacesCount))
+                \"\"\",
+                \(node.exampleArguments),
+                fileID: \(fileID.file),
+                file: \(filePath.file),
+                line: \(filePath.line)
+            )
+            """.indent(by: spacesCount)
+        return ExprSyntax(stringLiteral: example)
+    }
+}
+
+private extension FreestandingMacroExpansionSyntax {
+    func argumentValue(named name: String) -> String? {
+        arguments.first { $0.label?.text == name }?.expression.description
+    }
+
+    var exampleArguments: String {
+        """
+        configuration: \(argumentValue(named: "configuration") ?? "[:]"),
+        testMultiByteOffsets: \(argumentValue(named: "testMultiByteOffsets") ?? "true"),
+        testWrappingInComment: \(argumentValue(named: "testWrappingInComment") ?? "true"),
+        testWrappingInString: \(argumentValue(named: "testWrappingInString") ?? "true"),
+        testDisableCommand: \(argumentValue(named: "testDisableCommand") ?? "true"),
+        testOnLinux: \(argumentValue(named: "testOnLinux") ?? "true"),
+        testOnWindows: \(argumentValue(named: "testOnWindows") ?? "true"),
+        excludeFromDocumentation: \(argumentValue(named: "excludeFromDocumentation") ?? "false")
+        """.indent(by: 4, skipFirst: true)
+    }
+}
+
+private extension CodeBlockItemListSyntax {
+    func asExampleBody(unindentedBy spaces: Int) -> String {
+        CodeIndentingRewriter(style: .unindentSpaces(spaces))
+            .visit(self)
+            .description
+            .replacing("/*>*/", with: "↓")
+    }
+}
+
+private extension Trivia {
+    var countSpaces: Int {
+        pieces.compactMap { piece in
+            switch piece {
+            case let .spaces(number): number
+            default: nil
+            }
+        }.reduce(0, +)
+    }
+}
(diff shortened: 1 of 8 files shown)


SimplyDanny:
I'd like to hear your opinion on this, @ZevEisenberg.

Source/SwiftLintCoreMacros/Example.swift
```diff
@@ -0,0 +1,78 @@
...
+
+struct Example: ExpressionMacro {
+    static func expansion(
+        of node: some FreestandingMacroExpansionSyntax,
+        in context: some MacroExpansionContext
+    ) -> ExprSyntax {
+        guard let fileID = context.location(of: node, at: .afterLeadingTrivia, filePathMode: .fileID),
+              let filePath = context.location(of: node, at: .afterLeadingTrivia, filePathMode: .filePath) else {
+            context.diagnose(SwiftLintCoreMacroError.invalidSourceLocation.diagnose(at: node))
+            fatalError(SwiftLintCoreMacroError.invalidSourceLocation.message)
```
ZevEisenberg: Why a fatal error inside the macro code? I would think that diagnosing and then early-returning, or diagnosing and then dropping back to some placeholder code, would be preferable.

Source/SwiftLintCoreMacros/Example.swift
```diff
@@ -0,0 +1,78 @@
...
+
+        guard let trailingClosure = node.trailingClosure else {
+            context.diagnose(SwiftLintCoreMacroError.missingExampleBody.diagnose(at: node))
+            fatalError(SwiftLintCoreMacroError.missingExampleBody.message)
+        }
+
+        let spacesCount = trailingClosure.rightBrace.leadingTrivia.countSpaces
+        let example = """
+            Example(
+                code: \"\"\"\(trailingClosure.statements.asExampleBody(unindentedBy: spacesCount))
```
ZevEisenberg: Consider a raw string here to avoid noisy escaping

Source/SwiftLintCoreMacros/Example.swift
```diff
@@ -0,0 +1,78 @@
...
+                \"\"\",
+                \(node.exampleArguments),
+                fileID: \(fileID.file),
+                file: \(filePath.file),
+                line: \(filePath.line)
+            )
+            """.indent(by: spacesCount)
+        return ExprSyntax(stringLiteral: example)
```
ZevEisenberg: Is there a version of this init that doesn't take a string literal? It probably works, but semantically, this is supposed to be used for _literal_ literals 😂

Source/SwiftLintCoreMacros/Example.swift
```diff
@@ -0,0 +1,78 @@
...
+    }
+}
+
+private extension FreestandingMacroExpansionSyntax {
+    func argumentValue(named name: String) -> String? {
+        arguments.first { $0.label?.text == name }?.expression.description
```
ZevEisenberg: Is `description` the best way to get what we need here? It's often documented not to be stable. I forget if there's a more appropriate thing though.

Source/SwiftLintCoreMacros/Example.swift
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    var exampleArguments: String {
+        """
+        configuration: \(argumentValue(named: "configuration") ?? "[:]"),
```
ZevEisenberg: Including all of these here is going to expand the generated macro call site to include all params. Would it be possible to omit any that aren't passed explicitly? That would also allow you to fall back to the real default values without having to repeat them in both places.

Tests/MacroTests/ExampleTests.swift
```diff
@@ -0,0 +1,110 @@
...
+private macro example(body: () -> Void) -> SwiftLintCore.Example = #externalMacro(
+    module: "SwiftLintCoreMacros",
+    type: "Example"
+)
+
+@Suite
+struct ExampleTests {
+    @Test
+    func exampleMacro() {
+        assertMacroExpansion(
```
ZevEisenberg, line 22: I forget if I used the Point-Free macro testing stuff in my earlier PR, and can't easily check now, but it makes stuff like this file a little easier to read, and a lot easier to develop. (Mega facepalm for me if this _is_ that stuff and I just don't recognize it.)

ZevEisenberg (commented):
This is a great idea! Do you have a sense of how many cases it could apply to? I wonder if there's any world where `#examples` and `#corrections` macros could be modified, even temporarily, to help diagnose potential refactors, and even help to suggest changes?

Tests/MacroTests/ExampleTests.swift
SimplyDanny: These are the basic helpers provided by the SwiftSyntax library for testing. Any changes here if considered worthwhile I'd like to postpone to a separate change.

Source/SwiftLintCoreMacros/Example.swift
SimplyDanny: Good call. Safes a second duplication of default arguments, leaving them in the `Example` initializer and the macro declaration.

Source/SwiftLintCoreMacros/Example.swift
SimplyDanny: I think it is. But actually, the conversion to a string is only necessary to get the violation marker `/*>*/` replaced.

Source/SwiftLintCoreMacros/Example.swift
SimplyDanny: Changed to a direct parsing step plus indentation adjustments via rewriter.

Source/SwiftLintCoreMacros/Example.swift
SimplyDanny: With a raw string there is no string interpolation `\(...)`, or is that possible somehow?

Source/SwiftLintCoreMacros/Example.swift
SimplyDanny: True. Returning just `""` now. The diagnostic is enough to stop compilation.

SimplyDanny:
> Do you have a sense of how many cases it could apply to?

Using the compiled example snippets should be possible for many examples, given that they don't require a lot of setup. For example, in the rule I updated for demonstration reasons,

```swift
List($history) { $historyItem in
  Foo()
}
```

cannot easily made compilable.

> I wonder if there's any world `#examples` and `#corrections` macros could be modified, even temporarily, to help diagnose potential refactors, and even help to suggest changes?

Not sure I get this. There is a ticket, that asks for every example to report parsing issues and let the test fail. There even exists code to prepare for this. However, all examples would need to be reviewed and fixed first to activate that. That would only be "parsing" though. `#example { ... }` is about compilation, which is nice to have but not super important for what SwiftLint does.

ZevEisenberg:
The existing examples/corrections macros could supply fixits that transform the current string-based examples into live compiling examples. You could apply the fixits once and then revert the modifications to the macros. Or, keep them, if you want people to be able to use them in future so they default to the most expressive macros available by default. If that's still lot clear, I can try to sketch it out with some code examples.

SimplyDanny:
> The existing examples/corrections macros could supply fixits that transform the current string-based examples into live compiling examples. You could apply the fixits once and then revert the modifications to the macros. Or, keep them, if you want people to be able to use them in future so they default to the most expressive macros available by default. If that's still lot clear, I can try to sketch it out with some code examples.

Working on it. Thanks for the idea! There will still be use cases for all these kinds of expression example code snippets.

ZevEisenberg:
I've had package traits on the brain recently, so: you could have a package trait that is read by the macro target. When enabled, all `#example` macros in the code would switch over to "suggest fixits for me" mode. Then, when you're done migrating all you want to migrate, disable the trait again before committing.

Just a random thought. I'm eager to see where you end up with this.

